PR #328, opened 2013-09-13
Adding Time To Control MovieClip Animations

MikkoH:
This pull request is an update to the Pull Request: https://github.com/GoodBoyDigital/pixi.js/pull/322

I've made ```Time``` not be a static object but rather now each renderer will have an instance of time. It is set to a new property on ```Stage``` called time.


englercj:
Things like this are moving into separate modules, we will not be making core modifications for them. Feel free to write a user-land module for this when we get there.

PR #537, opened 2014-02-02
Protect against NaN rotation values

cacheflowe:
Throws an Error if a DisplayObject's rotation was set to NaN. Without this check, Sprite's textures could disappear silently, with no reference to an improper rotation value as the culprit.

### DIFF
--- a/src/pixi/display/DisplayObject.js
+++ b/src/pixi/display/DisplayObject.js
@@ -402,6 +402,9 @@ PIXI.DisplayObject.prototype.updateTransform = function()
     // TODO OPTIMIZE THIS!! with dirty
     if(this.rotation !== this.rotationCache)
     {
+        if(isNaN(parseFloat(this.rotation)))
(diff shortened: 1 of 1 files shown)


Jacq:
This problem took me some time today because it was mixed with the container type, I'm using pixi.js 3.0.6
Using NaN in rotation for Sprites in PIXI.Container show them ok but if you change to a PIXI.ParticleContainer all sprites dissapear without warning/errors in the console.

englercj:
@Jacq Correct, all we do is upload the value to the GPU. Uploading NaN could do a lot of weird things. Taking a significant performance hit to ensure you didn't have a NaN value (when you can just breakpoint and debug) is not worth it.

Potentially we could do some meta programming to have the "debug" version of pixi include guards like this and the "production" version not include it (similar to how STL works in many places) but that is not likely to happen soon. I opened a feature request for it in #2032.

PR #565, opened 2014-02-11
Simple implementation of getElementsByTagName when no DOMParser API is present to enable cross-platform bitmap fonts

wojciak:
This allows bitmap fonts to properly load with the asset loader and render with PIXI.BitmapText on Ejecta and possibly CocoonJS - removes strict DOMParser API dependency.


englercj:
This should be in a separate module from core, and even separate from the new coming loader module. We won't be putting polyfills in PIXI code very much going forward. See #1296

PR #618, opened 2014-03-01
Created a transparency-based hitArea implementation

peci1:
So here we have it! The transparency-based hitArea finally rewritten to the new renderer structure and (hopefully) complying to the contribution guide =) Comes with tests and a new example :) Enjoy!


englercj:
This should be a user-land module separate from PIXI's core. See #1296 for more details.

PR #696, opened 2014-04-08
Added proper canvas drop shadow support

mowday:
Uses the proper canvas drop shadow support, and also added blur.
Do note though that using the blur option is very heavy on the CPU and should be used with caution.

Always good to have the option though :)


englercj:
We can review this change when Text has been split into a separate module. See #1296.

PR #717, opened 2014-04-24
Added ability to change BitmapText tint and alpha per-character

lewster32:
Added ability to specify tint and alpha per-character on BitmapText through the use of (optional) codes in the format {#ffffff} (RGB), {#ffffffff} (RGBA) and {#} to revert to the default color of the text. If codes are detected then the text is parsed appropriately, otherwise the text behaves as normal.

Example available here: http://rotates.org/phaser/bitmaptextcolor.htm

Parsing could be more efficient, however the above example shows that even with relatively frequent updates it appears to perform well, even using canvas on mobile.

### DIFF
--- a/src/pixi/text/BitmapText.js
+++ b/src/pixi/text/BitmapText.js
@@ -67,13 +67,79 @@ PIXI.BitmapText.prototype.setStyle = function(style)
 };
 
 /**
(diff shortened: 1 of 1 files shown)


englercj:
BitmapText will be moving to a separate module after the rewrite, for more info checkout #1296. Also, it may make sense to have a user-land module that implements text differently if you so choose.

PR #773, opened 2014-06-01
Adding AnimationManager entity for animations and delayed/repeated calls maintenance.

dimalev:
there is still some more work to do. currently, this patch is just to make sure i do things right

thankyou


englercj:
This seems cool but after the rewrite really makes more sense as a user-land module. Check out #1296 for more details.

Issue #573, opened 2014-02-13
Why does PIXI use Object.defineProperty ?


tleunen:
Just a global question to know the reason why PIXI uses that... Because it's isn't compatible with the obfuscation process as the variable is defined with a string, but not when we access it later in the code.

If you want to use that, you should use the variable as a string everywhere else in the code. In that case, it won't break the obfuscated code.

Or you want to prevent users to obfuscate their code :D

cdunlap:
> I have a new work around that fix this issue for me without changing the initial code so it's perfect.

@tleunen Does your workaround allow uglify to complete the obfuscation and minification processes without any problems? If so, I'm interested.

tleunen:
@cdunlap: I didn't test with uglify but we were using Google Closure Compiler and it worked well.
Unfortunately, it was done in my previous company and I don't work there anymore so I don't have access to the custom code I did at that time.

The code was also based on Pixi v2, now that it's v3, it might not be needed anymore

PR #866, opened 2014-07-17
Remove trailing commas for Google Closure compatibility

tapeinosyne:
By default, the Google Closure Compiler is rather pedantic about backward compatibility and chokes on trailing commas. Removing them would allow Pixi to be compiled with Closure, and improve stylistic consistency. (As far as I see, trailing-less is already the dominant style in your codebase; I suppose the occasional trailing comma springs up because JSHint does not detect it.)

It seems that Pixi does not support advanced Closure compilation (#573), but basic compatibility would be nice nonetheless.

For context, I should add that I am writing a Pixi wrapper in Clojurescript, which uses Closure. Compiling the wrapped library is [generally desirable](http://lukevanderhart.com/2011/09/30/using-javascript-and-clojurescript.html).

### DIFF
--- a/src/pixi/extras/Spine.js
+++ b/src/pixi/extras/Spine.js
@@ -1286,7 +1286,7 @@ spine.AtlasRegion.prototype = {
     index: 0,
     rotate: false,
     splits: null,
-    pads: null,
+    pads: null
 };
 
 spine.AtlasReader = function (text) {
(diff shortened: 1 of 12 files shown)


martinklepsch:
Sorry for hijacking this thread! Any chance someone here has complete extern files for pixi.js? 
I'd like to package it up for Clojurescript (@cljsjs) and extern files are required for things to work across all optimisation modes in Clojurescript.

PR #911, opened 2014-08-05
Remove colorAttribute from the attributes array in ComplexPrimitiveShader.init method

FedeOmoto:


englercj:

PR #924, opened 2014-08-14
A little patch to CanvasTinter.js to fix a performance issue (ipad/ipad mini crash)

alaa-eddine:
the situation : when you handle a lot of textures with tints changing overthe time, the performance drop.
I noticed that document.createElement("canvas") is done for each new tint.
this patch only save the initially created canvas so it'll be used for all future texture tinting.

on ipad/ipad mini I had crashes without the patch and 10~30 FPS ... with the patch I get 55~60 FPS and no more crash :)


englercj:
Closing this for now since I think there is an error in user-land.

PR #941, opened 2014-08-29
Safari Pixel Rounding Fix

jiminychris:
Safari automatically anti-aliases when not drawing to the canvas with pixel-perfect (integer) coordinates. I imagine this should be fixed by setting the CanvasRenderer.renderSession.roundPixels = true, but this can result in weird behavior if a Sprite frame offset is not an integer (eg. x-anchor at .5 on a 15-pixel wide sprite results in a dx of -7.5).

This might not be the best way to fix this, but I'm just truncating the decimal from the dx and dy frame offsets. It seems to work for me.

Before: http://jsfiddle.net/jiminychris/z3srL5dy/

After: http://jsfiddle.net/jiminychris/mygLLr9c/

Note that these should look exactly the same in Chrome, and this was only an issue in Safari, as far as I can tell.

### DIFF
--- a/src/pixi/display/Sprite.js
+++ b/src/pixi/display/Sprite.js
@@ -326,6 +326,10 @@ PIXI.Sprite.prototype._renderCanvas = function(renderSession)
     {
         renderSession.context.globalAlpha = this.worldAlpha;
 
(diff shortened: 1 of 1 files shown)


englercj:
This was merged into the codebase at some point but then never closed. Probably my fault!

PR #956, opened 2014-09-08
Added viewport culling to render process

alexh00:
Here is frustum culling again, added to dev branch not master this time, sorry about that!

### DIFF
--- a/src/pixi/display/DisplayObject.js
+++ b/src/pixi/display/DisplayObject.js
@@ -441,6 +441,25 @@ Object.defineProperty(PIXI.DisplayObject.prototype, 'cacheAsBitmap', {
     }
 });
 
+/**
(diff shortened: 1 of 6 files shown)


englercj:
Thanks for all the work @alexh00, with the upcoming rewrite a lot of things are changing and whether or not it makes sense for us to put this in core will be one of them. I think it might make more sense for a separate module to manage the visibility of items instead of writing core code to do culling.

alexh00:
Ok cool, thanks for letting me know. I think making PIXI more modular sounds like a great approach to take, and as far as viewport culling goes from my point of view as long as the feature is available that is a win.

englercj:
Think we may keep this out of core for now, unless @GoodBoyDigital disagrees and wants to reopen this.

PR #969, opened 2014-09-15
Optimalization of updateTransform in Spine, condition has been met every time, attachment.name does not exist

frankiex:
Condition with attachment.name changed to attachment.rendererObject.name, because attachment.name never exists


englercj:

PR #1015, opened 2014-10-04
Browserify conversion (fixed target)

Qard:
> This is the fix of #1014 to point to the dev branch, rather than master.

I am working on converting the build process to use browserify, so eventually better modularization can be achieved.

I've looked through the previous attempts at https://github.com/drkibitz/node-pixi and https://github.com/pixijs, but found they suffered from being too radically divergent. Unlike those, I'm avoiding making radical changes to the codebase. I hope to be able to track upstream with minimal conflicts, allowing gradual conversion to occur.

Yes, this method is a little bit ugly for commonjs. But it allows the current code to work almost entirely unchanged while still splitting it up into unique modules in the browserify bundle. This should allow files to be converted to commonjs one at a time, rather than in one big rewrite.

I could use some feedback. Does this seem like a reasonable approach?

A possible next step to avoid the dependency in each file of a global "PIXI" variable being present would be to add `var PIXI = require('path/to/module.js')` at the top of each file. I've held off on that for now because it would break the existing build process, leaving browserify the only option. I want to get feedback before I continue.


englercj:
Awesome work @Qard Pixi has been browserified in the rewrite branch, for more info check out #1296.

Boycce:
I think this switch to browserify isn't the greatest. Browserify was built for enabling server code running on the client side. Pixi is a pure client side library.

When using Pixi as a dependency for libraries, a granular control over the files you want included in the build process is no longer possible. You are limited to Pixi's controlled module bundles through `--exclude`

Here is an example of file daisy picking in `1.6.1`
https://github.com/Boycce/pixi.js/blob/e4579b1515e7cc884c0b447ff791ae52e62f8c16/Gruntfile.js

englercj:
@Boycce Still possible with browserify:

```js
// loads only the Container.js file (and dependencies).
require('pixi.js/src/core/display/Container');
```

Boycce:
Cheers @englercj. Got it working in the end through browserify's `b.pipeline.get('deps')...)` & `b.exclude('src')` . Although not as convenient.

englercj:
I agree, we are looking at doing ES6 modules in v4 which has the feature of being able to pull in only specific parts. Also, and extreme lean towards modularity to more granular control over what your bundle is.

PR #1025, opened 2014-10-07
Don't set the style.width and style.height on resize.

agamemnus:


englercj:
I think in the rewrite we will be completely decoupling the renderer from the DOM, and putting DOM handling into user-land. Thanks for this.

PR #1052, opened 2014-10-16
[FILTERS] Renderbuffers should always be reseted after use.

darionco:
Fixes filter rendering issue in Ejecta. Filters would lose the render targets and the screen would freeze.


englercj:

PR #1127, opened 2014-11-06
Adds 2 new properties to DisplayObject: scaleMin and scaleMax.

photonstorm:
These are Point objects that allow you to set a lower and upper bounds on the scale value the display object can transform to. The updateTransform method now checks these two properties (if set) and enforces them.

A small helper method setScaleMinMax has been added to allow the devs to quickly set these two properties, but if you feel it's not needed then you can remove it and it'll carry on working happily with it out.

I have found these 2 properties utterly invaluable in some responsive bbc games we've been building. For example we had a case where the whole game scaled responsively, but the top-level UI was never allowed to go below a certain size (or it becamse impossible to touch). Setting just the scaleMin allowed this, but it could still scale up happily on larger displays.

### DIFF
--- a/src/pixi/display/DisplayObject.js
+++ b/src/pixi/display/DisplayObject.js
@@ -28,6 +28,24 @@ PIXI.DisplayObject = function()
     this.scale = new PIXI.Point(1,1);//{x:1, y:1};
 
     /**
+     * Set the minimum scale this Display Object will scale down to.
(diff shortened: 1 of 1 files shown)


englercj:
Where did we land on this @GoodBoyDigital, do we want this in core or userland?

GoodBoyDigital:
userland! :+1: cheers all!

PR #1141, opened 2014-11-08
MovieClip animations helper

elsassph:
Add a static `Texture.fromFrames(prefixId)` tool function which creates an `Array<Texture>` from `TextureCache` where the `frameId` starts with a given prefix.

Useful to create a MovieClip animation based on an arbitrary number of frames named like "ship01.png", "ship02.png",... without having to create long lists of strings.

PS: this might be worth creating a "cache" for this method's results.


englercj:
This is neat, but I don't think it is necessary or super useful in core. Especially with the rewrite coming up (#1296) we have decided to not put things into core that are not used in core, and MovieClip is not in core.

PR #1171, opened 2014-11-13
Add extra checks to handle context lost.

glafarge:
I have some context lost errors those days.
Not a big change but I hope it can help...

** Seen on [Handling Context Lost](https://www.khronos.org/webgl/wiki/HandlingContextLost).


englercj:

PR #1176, opened 2014-11-14
RenderTexture - preserve world transform

pnstickne:
This addresses issue https://github.com/GoodBoyDigital/pixi.js/issues/1174 - as a bonus it also fixes and expands documentation.

The renderCanvas and renderWebGL modify the worldTransform, worldAlpha,
and update the transformations of the children. This code adds in a
parameter to the these methods (true by default now, but that can be
changed) to indicate that these transformations should be undone when the
rendering is complete.

While restoring the previous transformations does add a little bit of
overhead this is adventageous in two important ways:

- The default (or requested) behavior works in an "unsurprising manner"
  and does not leave a temporary transformation applied permamently when
  the render operation completes.
- Even if the restore/preserve operation is the documentation and addition
  of method parameter makes this behavior more apparent.

### DIFF
--- a/src/pixi/textures/RenderTexture.js
+++ b/src/pixi/textures/RenderTexture.js
@@ -9,19 +9,19 @@
  *
  * A RenderTexture takes a snapshot of any Display Object given to its render method. The position and rotation of the given Display Objects is ignored. For example:
  *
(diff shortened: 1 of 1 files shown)


englercj:

PR #1186, opened 2014-11-17
Make video work in Firefox.

tobireif:


englercj:
Is this needed anymore? This thread is hard to follow, should this be merged in?

tobireif:
I tested
http://tobireif.com/pixijs_video_example_copy/examples/example%2025%20-%20Video/
on Mac in FF and Chrome, and on Windows in FF and Chrome. (video, filtering, sound)

Mat had written "seems to not work for me in FF on my mac". Ideally Mat would do the same test he did back then and confirm that it works for him now.

Also, it would be great if someone could (then) review the code, I'm not sure whether some of the old code (and the bunny pic in the example 25 dir) could be removed.

Then, please merge.

P.S.
The type attribute is not required per the spec, but it speeds up content selection, and also might save people headache when their serverside MIME type settings don't take effect immediately (or when content gets moved to another server - it's just much more robust).

englercj:
Merged this in with a few changes.

englercj:
> Will the WebM file (and the code your merged into your branch) be merged into GoodBoyDigital:dev ?

Yes we will merge this code when the rewrite is a little further along.

> Also, in your branch there's no examples dir and no index.html of example 25

The examples are going to be written from scratch in a new repository.

> I tried to test your code by uploading the two video files, an index.html where two videos files are supplied per your spec

The rewrite build does not run yet, there is a lot more work to do before it does. See #1296 for more info.

PR #1192, opened 2014-11-18
Allow filter areas to originate off screen. Clip filter areas to on screen portion.

giraluna:
Clamping filter area origin to 0 means filters cant go partially off screen due to camera scrolling etc, but stick to the edge in full size.

This guy had the same problem http://www.html5gamedevs.com/topic/9638-webgl-filters-rendering-offset/

This commit should clip the filter area to the visible part of the screen, but an alternative fix might be just removing the 4 lines here https://github.com/GoodBoyDigital/pixi.js/blob/master/src/pixi/renderers/webgl/utils/WebGLFilterManager.js#L106-110


englercj:
merged into the rewrite branch, see #1296.

PR #1206, opened 2014-11-21
Add 'pixelRatio' property

alesveselka:
Using text on Retina displays (or displays with 'devicePixelRatio' > 1) results in blurred text. Also blurred images. I understand that the new 'resolution' property should solve this issue, but I can't get as sharp text/images when set the resolution via PIXI's renderer options. When I scale the canvas manually  (i.e. canvas.style.width = w; canvas.width = w * devicePixelRatio;) I get much sharper text and images, even though the final canvas' HTML element has same dimensions as if using PIXI's resolution. I can't figure the reason why. 

So I introduced property 'pixelRatio' for correct pointers coordinates when the resolution is set to 1, but the canvas is scaled manually outside of PIXI.

### DIFF
--- a/src/pixi/InteractionManager.js
+++ b/src/pixi/InteractionManager.js
@@ -30,6 +30,14 @@ PIXI.InteractionManager = function(stage)
     this.mouse = new PIXI.InteractionData();
 
     /**
+     * The touch data
+     *
+     * @property touchData
+     * @type InteractionData
+     */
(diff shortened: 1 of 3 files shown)


englercj:
We are currently rewriting pixi with an emphasis on modularizing and rewriting components like this. We will keep these issues in mind when we move to a new solution.

PR #1207, opened 2014-11-22
[SPINE] Updated to latest version + added skinned mesh attachment support.

darionco:
This is a pretty big change to the spine components but self explanatory:
- Separates the Spine.js file into two files
  - Spine.js which contains the raw spine runtime directly from esoteric software (the needed changes for pixi.js were merged into the main branch https://github.com/EsotericSoftware/spine-runtimes/pull/315)
  - PIXISpine.js implements the pixi.js specific details about how to load and draw spine animations. This new implementation supports region attachments and skinned mesh attachments.
- JsonLoader implements the new loading system required by Spine 2.0 files (free update to spine users)
- Adds the ability to select between drawing GL_TRIANGLE_STRIP or GL_TRIANGLES buffers to Strip.js, after this change it feels like "Mesh" would be a better name for the class. Also uses the padding variable when rendering in canvas rather than WebGL.

Adding more attachment types at this stage is trivial, I will add them as I need them. Here's an example; these are two sprites animated using skinned mesh attachments (rendered with pixi.js in safari with WebGL enabled):

### DIFF
--- a/src/pixi/display/Sprite.js
+++ b/src/pixi/display/Sprite.js
@@ -36,8 +36,8 @@ PIXI.Sprite = function(texture)
      * @property texture
      * @type Texture
      */
-    this.texture = texture || PIXI.Texture.emptyTexture;;
-    
+    this.texture = texture || PIXI.Texture.emptyTexture;
+
     /**
      * The width of the sprite (this is initially set by the texture)
      *


JaredSartin:
Don't mean to necro this, any meshed examples (like the dancing animals)?

Issue #1264, opened 2014-12-11
stage.setInteractionDelegate doesn't work


beeglebug:
It looks like there area few problems with it:

It doesn't set `stage._interactiveEventsAdded`, so when the renderer runs, it calls `stage.interactionManager.setTarget(this)`, which wipes out the newly added delegate.

Also, because it calls `interactionManager.setTargetDomElement` directly, rather than `setTarget`, the `target` and `resolution` properties are not set.

I have an idea of how to fix it, and potentially add some functionality at the same time, i'll do a PR later.

GoodBoyDigital:
This should work correctly in v3 now ->
```renderer.plugins.interactiveManager.setTargetElement( element )```
cheers! :+1:

PR #1265, opened 2014-12-11
Fixes for setInteractionDelegate

beeglebug:
See #1264 for original issue.

I've put some comments below to explain the changes.


englercj:
I merged your commits to `InteractionManager` manually to the rewrite branch, but just know that we are going to rewrite the interaction code anyway here soon and the renderers are no longer coupled to interaction at all. See #1296 for more info.

PR #1274, opened 2014-12-13
Added error listeners to the ImageLoader and the JsonLoader

gilamran:


englercj:
I merged your commit manually, but just know that we are going to rewrite the loaders anyway here soon. See #1296 for more info.

Issue #1296, opened 2014-12-26
Pixi v3.0.0


englercj:
### Current Status

**Total Rewrite Progress:** 100% complete

**Reminders for myself:**
- PolyK should be replaced, it has a few bugs ([earcut](https://github.com/mapbox/earcut)?)

# Proposal

## Introduction

Hello all, I decided to write this up to inform the community what I am working on and to gather some feedback. The biggest thing to realize about what follows is that I am working on it right now, and that it is no longer theoretical. Also, keep in mind this has not yet be "blessed" by the rest of the Pixi team so this may not make it in, I am just looking for feedback from the team and the community.

You can watch the work being done in a branch on my fork: https://github.com/englercj/pixi.js/tree/dev-rewrite

If you read this proposal and want to help, fork my branch and submit PRs to the `dev-rewrite` branch and lets get this done!

That being said, v3.0.0 will be a rewrite of PIXI. I have it split into distinct phases with clear goals to accomplish a clean, modular library that will make it easier for us to maintain and for others to contribute.

### Phase 0 - Cleanup/Meta
**Status**: Completed

Phase 0 is to cleanup the code repository, removing things that shouldn't be in the main repo. For example: `examples/`, `docs/`, `trim/`. Each of those should be separate and maintained on their own. It also converts the project to use gulp, reorganizes the tasks for the project, and generally prepare the project for the coming rewrite.

### Phase 1 - Code Modularization
**Status**: Completed
**Related issues**: #307

Phase one is to convert the codebase to use browserify, remove dead code, convert yuidoc to jsdoc comments, and cleanup some style inconsistencies. This is perhaps the most time consuming phase because it requires combing through every source file in pixi and changing a lot. This phase will make it much easier to refactor the library and accomplish later phases.

### Phase 2 - Component Modularization
**Status**: Completed

Once the library has been revamped from a code-quality standpoint the next step is to remove circular dependencies and split the core library into multiple discrete modules. Each module will be in a separate folder within the `src/` directory of the project, but should be split into separate repos at a later date.

The modules I have in mind are:
- `core` - Core renedering and display (math, DO, DOC, Sprite, SpriteBatch, Graphics, textures, renderers).
- `spine` - Adds spine support.
- `loader` - Adds loaders for the different things pixi supports.
- `filters` - Adds a pack of different filters.
- `text` - Adds text support.
- `extras` - Adds TilingSprite, and MovieClip support (could use a better name).
[...]

PR #1300, opened 2014-12-27
more Garbage Collector friendly changes

Kamil93:


englercj:
:+1: I will make sure changes like this make it to the rewrite branch (see: #1296).

PR #1301, opened 2014-12-27
Change the way array is cleared to zero elements

Kamil93:
interactiveItems now is cleared by length = 0; instead of creating new array every time =[]; It is more efficient for garbage collector.


englercj:
Went ahead and merged this in to the rewrite, but we will be completely rewriting the manager before this is over.

PR #1303, opened 2015-01-01
Also log the type for non-Chrome browsers

gre:
The initial "sayHello" log is not consistent across browsers. There is no way to get the WebGL/Canvas information from the console of Firefox for instance (only works on Chrome).


gre:
oops sorry. I'll redo it on `dev`

PR #1307, opened 2015-01-02
Prefer to expose the unminified version to NPM

gre:
this is to improve the dev experience when working with PIXI using browserify.
(and up to the final user to minify the resulting build as it is the convention in NPM in general)

Thanks


gre:
oops sorry. will redo another PR (on dev)

PR #1309, opened 2015-01-02
Prefer to expose the unminified version to NPM

gre:
this is to improve the dev experience when working with PIXI using browserify.
(and up to the final user to minify the resulting build as it is the convention in NPM in general)


englercj:
:+1: I think this is already fixed in the rewrite branch, once it is merged I will likely close this.

PR #1313, opened 2015-01-03
Allow canvas graphics renderer to process multiple paths per mask

ianbjorndilling:
I was recently in need of using multiple mask paths to dynamically mask multiple areas of a sprite. This works fine on the WebGL renderer, but the canvas renderer could only handle one mask path per graphics object. Judging by the pre existing for loop, it seems like it may have been originally intended that the canvas renderer be able to mask multiple paths.

I added a single .beginPath() call just before the execution of the for loop. All other .beginPath() calls were removed. I made sure to leave the calls to .closePath(). Since properties such as fillStyle and lineWidth aren't taken into account by canvas clipping, only one call to .beginPath() is needed. This way the canvas can clip using multiple paths.

### DIFF
--- a/src/pixi/renderers/canvas/CanvasGraphics.js
+++ b/src/pixi/renderers/canvas/CanvasGraphics.js
@@ -212,20 +212,15 @@ PIXI.CanvasGraphics.renderGraphicsMask = function(graphics, context)
 
     if(len === 0) return;
 
-    if(len > 1)
-    {
-        len = 1;
-        window.console.log('Pixi.js warning: masks in canvas can only mask using the first path in the graphics object');
-    }
+    context.beginPath();
 
-    for (var i = 0; i < 1; i++)
+    for (var i = 0; i < len; i++)
     {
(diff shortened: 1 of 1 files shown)


englercj:
Do you have a test for this, or have you tried it and made sure it works so I can merge it in?

ianbjorndilling:
Yes, I checked to make sure it worked for every possible shape. Here's a [fiddle](http://jsfiddle.net/jxnh3enk/2/) of it in action.

The only issue I had was using .drawRoundedRect. When a rounded rectangle shape is used in a mask graphic, shape.points is undefined in .renderGraphicsMask. I tried it with the latest build and got the same error. I opened an issue #1321.

englercj:
Interesting, I assigned this issue to myself and will take a look at it tomorrow.

englercj:
Yeah looks like I was able to fix that issue, here is a fiddle of it working with my custom build:

http://jsfiddle.net/jxnh3enk/3/

I'll merge this in.

Issue #1305, opened 2015-01-01
BitmapText bugs with `tint`


gre:
I have found 2 bugs in BitmapText when used with tint.

- The first is that if you update `tint` after creating a `BitmapText`, it won't be refresh until you re-set the text, or you explicitly call `.updateText()` (missing a dirty state).

- The second is that it is not possible to use the black tint `0x000000` because of [this code](https://github.com/GoodBoyDigital/pixi.js/blob/master/src/pixi/text/BitmapText.js#L169) which fallbacks to **white** – this behavior is inconsistant to the behavior of a `Sprite`.

marcusstenbeck:
@gre Could you make a [jsFiddle](http://jsfiddle.net/) that shows the bug in action?

I'll try to reproduce it on my machine, but to make sure I'm looking at the same problem you're looking at I'd love to get some running code showing the problem.

—Marcus

gre:
http://jsfiddle.net/n61590gc/

marcusstenbeck:
### Updating tint
Thanks for making the jsFiddle. I noticed that you're setting the `BitmapText.tint`. Right now it looks like Pixi doesn't support setting the attribute directly like you're doing (in fact, tint isn't even metioned in the docs for `BitmapText`). Even so, setting tint directly by assigning a value to the attribute directly makes it impossible for Pixi to know that you've updated (without adding some pretty slow code that keeps track of earlier states). The method [BitmapText.setStyle()](http://www.goodboydigital.com/pixijs/docs/classes/BitmapText.html#method_setStyle) is what you want to be doing. Unfortunately, [BitmapText.setStyle()](http://www.goodboydigital.com/pixijs/docs/classes/BitmapText.html#method_setStyle) won't let you set just one attribute, so you have to copy the previous style on your bitmap text, and change the thing you want to change, and then apply it afterwards. Example below.

```
// Copy the current style
var newStyle = myBitmapText.style;

// Update the tint
newStyle.tint = 0x123456;

// Apply the new style to myBitmapText
myBitmapText.setStyle(newStyle);
```

This is kinda painful, so I'll do a little bit of work to change it so you can specify just one attribute in [BitmapText.setStyle()](http://www.goodboydigital.com/pixijs/docs/classes/BitmapText.html#method_setStyle), changing the needed code to the following.

```
myBitmapText.setStyle({ tint: 0x123456 });
```

What do you think?

### Using black tint
Yeah. That's a bug. I'll get that fixed.

[...]

PR #1316, opened 2015-01-04
Check if tint is in acceptable range, otherwise set a default value.

marcusstenbeck:
From issue #1305, if you set the tint to black (`0x000000`) it'll fail the first test, even though it's a valid color.


marcusstenbeck:
I realized I didn't PR to dev branch. I'll reopen a new PR and do it right.

PR #1319, opened 2015-01-04
Check if tint is in acceptable range, otherwise set a default value.

marcusstenbeck:
From issue #1305, if you set the tint to black it'll fail the first test, even though it's a valid color.


englercj:
:+1: I will merge this in to the rewrite branch shortly.

PR #1315, opened 2015-01-04
Allow setStyle() to accept a variable number of style updates.

marcusstenbeck:
I've updated the BitmapText class to better handle updating style via the `BitmapText.setStyle()` method. Basically, whenever a user calls the method it builds the style from either passed in arguments, the current style, or a default style (when applicable).

From issue #1305.

PR #1320, opened 2015-01-04
Improve BitmapText.setStyle() and add BitmapText.tint attribute.

marcusstenbeck:
`BitmapText.setStyle()` should handle not having every value be defined, and not overwrite existing values with defaults. This is done by checking the input to the function.

`BitmapText.tint` is added, and the behavior is modeled from `Sprite.tint`.

From issue #1305, previous bad PR #1315

### DIFF
--- a/src/pixi/text/BitmapText.js
+++ b/src/pixi/text/BitmapText.js
@@ -50,7 +50,7 @@ PIXI.BitmapText = function(text, style)
     /**
      * @property style
      * @type Object
-     * @private
+     * @readOnly
      */
     this.style = {};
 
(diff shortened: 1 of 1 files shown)


englercj:
:+1: I will merge this in to the rewrite branch shortly.

marcusstenbeck:
@englercj So convention goes against methods like `setStyle()`?

And should we have `tint` in the `style` object? It's a inconsistent with how the `Text` class styles.

englercj:
@marcusstenbeck Yes going forward if you write `get*()` or `set*()` then it *probably* should just be a property setter (with some exceptions). As far as tint, I only put it in `_style` for internal class tracking, the external API is `BitmapText#tint = ...` so it should be consistent. I haven't gotten around to rewriting text yet :)

marcusstenbeck:
If you could point me into the direction of the style of rewrite you're looking I'm happy to get the text classes in shape. Where do I start?

englercj:
@marcusstenbeck #1296 describes it as best as I can, I would read through that doc as closely as possible and then hit me up with specific questions.

marcusstenbeck:
@englercj Thanks. I wanted to make sure that there wasn't some resource of description that I'm just because of being new to the project. ✌️

Issue #1331, opened 2015-01-08
IE 10/11 indexSizeError when unloading image asset and IE tries to render it right after


JanAmbrozic:
There is an existing issue that looks similar to the already closed #614. 

I have a scene switching components that unloads/loads assets depending on the scene but it is using the same pixi object. So for a brief period of time there is a "placeholder" 1x1 image (our asset loader adds that if the loading is not done when the image is trying to be rendered) that is  being rendered for some reason in IE (works in all other desktop/mobile browsers - the image is actually there on time).

I only tested CanvasRendered because IE10 only supports that but the fix should be good for all WebGL as well.

I have a fix for it so I'll add a PR in few minutes.

englercj:
I think this has to do with how you are "unloading" assets as I mentioned in the PR.

GoodBoyDigital:
Pretty sure this is licked in v3. Please let us know if it still persists! thanks :+1:

danieldunderfelt:
Hi there! Sorry to open this issue, but it is still present. We have an app that doesn't work in any IE version because it throws the `IndexSizeError` when it tries to draw an image to a CanvasRenderer on ie 10 and 11. We are running Pixi.js v3.0.11. Here's the code that throws:

```
renderer.context.drawImage(
                texture.baseTexture.source,
                texture.crop.x * resolution,
                texture.crop.y * resolution,
                width * resolution,
                height * resolution,
                dx  * renderer.resolution,
                dy  * renderer.resolution,
                width * renderer.resolution,
                height * renderer.resolution
            );
```

The texture exists, all values are over 0 except `dx` and `dy`. `resolution` is 1.

Is there any way to circumvent this?

themoonrat:
The issue is occuring because the dimensions it's calculating are incorrect. Other browsers just handle it, IE throws an error. For example, if it tries to draw from the source image thinking it's 100x100 pixels, but the source image is actually only 90x90 pixels, that error this thrown. 

Are the resolutions of the assets all 1 as well? I've had the issue when using resolutions like 0.5 for textures / rendering - and I changed that code to do 
```
renderer.context.drawImage(
    texture.baseTexture.source,
    Math.floor(texture._frame.x * resolution),
    Math.floor(texture._frame.y * resolution),
    Math.floor(width * resolution),
    Math.floor(height * resolution),
    Math.floor(dx * renderer.resolution),
    Math.floor(dy * renderer.resolution),
[...]

PR #1332, opened 2015-01-08
Dev

JanAmbrozic:
Fix for #1331


englercj:
I'm curious what situation occurs that causes the crop to be wrong? Also this code creates a new object each frame which is then discarded. We definitely don't want to do that.

JanAmbrozic:
Yeah I was thinking about that object, I'll remove it later and issue a new PR. Wasn't sure if you have any preferred way of "normalising" variables in cases like that.

I will create a test case tomorrow but essentially if you drop image out of memory this happens. And only way to really remove the image out of memory (that we know off) is to replace it with 1x1 px "fake"  image. And when this happens the crop width/height are still from original image and not 1x1 like they should be. That is only the case in IE's though, works as expected in other browsers.

englercj:
I'm more interested in what pixi code you are running to do that. If you are setting `texture.baseTexture.source` directly that isn't how that is supposed to work.

In v3 there are better methods for making a  texture be removed from gl memory, and for loading new sources.

JanAmbrozic:
I spent a bit more time investigating this and I will decline my PR. The problem is actually not there but the real problem is because the width and height don't get updated no matter what I do.

We have a pool of images that we add to the PIXI.TextureCache and when one of them is unloaded we just change the source of the DOM image element to the 1x1px fake image. The only thing I want to achieve now is to update crop width and height. I tried setting "requiresUpdate" of baseTexture to true, I tried using "updateSourceImage" method and so on. 

I checked the code of baseTexture/Texture and I've seen that crop is only being updated in the "setFrame" method. I'm just wondering if that is intentionally or are you maybe missing crop update somewhere.

If I update the crop manually the issue disappears and everything is fine
```javascript
PIXI.TextureCache[imageAsset].crop.width = 1;
PIXI.TextureCache[imageAsset].crop.height = 1;
```

englercj:
So obviously our code can't handle it if you do `PIXI.TextureCache[imageAsset].baseTexture.source.src = ''`. Instead you should probably delete the texture (`texture.destroy(true)`) and load something else later.

JanAmbrozic:
It's not just changing the src to '', but to the new image. But regardless I fixed it with that crop adjustment and since it only affects IE10/11 it's not the best place to put it in sprite.js like you said.

Than you for all the help and sorry for wasting your time :)

PR #1340, opened 2015-01-09
fix example script path

mrspeaker:
Don't know if you want a PR for this (feel free to close it if not relevant) - but the include path for `pixi.dev.js` is incorrect (so the example doesn't work). It's pointing one level too high.


englercj:
I'm going to merge code that deletes the examples from this repo so going to close this for now :)

Thanks though!

PR #1351, opened 2015-01-14
Reduced several lines

rivertam:
As far as I know, this makes the code equally efficient and slightly more readable.

### DIFF
--- a/src/pixi/display/Sprite.js
+++ b/src/pixi/display/Sprite.js
@@ -187,11 +187,7 @@ PIXI.Sprite.prototype.getBounds = function(matrix)
     var tx = worldTransform.tx;
     var ty = worldTransform.ty;
 
-    var maxX = -Infinity;
-    var maxY = -Infinity;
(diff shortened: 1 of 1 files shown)


rivertam:
Apologies if I've done something wrong: I've pretty much never even used Pixi.js or contributed to open source, but this seemed like an innocuous change.

englercj:
Can you create a jsperf that proves this is the same or faster? Looks like a good change if it is, but I'm not sure that the function call overhead is cheaper..

rivertam:
Apologies. I looked this up beforehand and found a StackOverflow question and attached jsperf which revealed very similar results, but I created a jsperf that directly compares what this code is actually doing and found that the current way is much better than the proposed way.

JSPerf here: http://jsperf.com/math-min-vs-ternary-operators

englercj:
All good, I had a hunch that might have been the case. Thanks for taking the time to look into this!

Issue #1325, opened 2015-01-05
TilingSprite doesn't use worldAlpha


JiDW:
Hi,

If I try to modify the alpha value of a DOC, the tilingsprites inside won't have their alphas affected.

Ex : http://www.serenity-ff.com/images/screenshots/screenshot155740001.jpg

Two TilingSprite (under the second text and the OK button) inside a DOC with alpha set to 0.5
Every other DO are correctly affected.

Modifying the alpha values of the tilingSprites works fine. So I guess this have something to do with worldAlpha.

Render : webgl

Regards.

marcusstenbeck:
Hey @JiDW, does the same thing work as you expect it to work when using the canvas renderer?

If so, could you provide a [jsFiddle](http://jsfiddle.net/) showing the error in action. If both Canvas and WebGL renderers behave the same I suggest we wait for @GoodBoyDigital or @englercj to give their 2c. Otherwise I'll risk wasting time fixing a bug that might not be a bug. :laughing:

marcusstenbeck:
@JiDW Thanks for the excellent fiddle. I could reproduce the weird behavior.

@GoodBoyDigital @englercj I'll have a look at this next up. Please assign this to me (is that possible?).

englercj:
@marcusstenbeck Nope! Thanks for looking into this, if you do end up with a fix it will make more sense to send the PR to my dev-rewrite branch since that is going to be merged into dev here in a few weeks.

marcusstenbeck:
@englercj Sure thing. `dev-rewrite` noted. I might not get to it until the weekend. That's my OS dev time slot. :wink:

marcusstenbeck:
@englercj This doesn't show up in the dev-rewrite branch. However I found that the default color in the Canvas- and WebGL-renderers aren't the same, so I opted for black in both (from reading source it looked like it was the intended color).

GoodBoyDigital:
looks like this is cool for v3 so closing for now. Thanks everyone!

PR #1359, opened 2015-01-18
Change to black as default renderer color.

marcusstenbeck:
The behavior is different between Canvas and WebGL renderers. I found more instances of black as default color, so I opted to go for that.

#1325


englercj:
Sure why not, need to refactor the renderers to use a common base so simple things like this are never different.

marcusstenbeck:
Agreed, there's a bunch of duplicate or very similar code all over the place there.

PR #1360, opened 2015-01-19
fix mask for CanvasGraphics

hameleonick:
I found bug with mask in current (2.2.3) version of PIXI. Just FYI.

Best regards,
Nickolay

### DIFF
--- a/bin/pixi.dev.js
+++ b/bin/pixi.dev.js
@@ -11419,12 +11419,12 @@ PIXI.CanvasGraphics.renderGraphicsMask = function(graphics, context)
         else if (data.type === PIXI.Graphics.RREC)
         {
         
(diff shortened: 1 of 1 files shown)


englercj:
@hameleonick Thanks! This is already fixed in the `dev-rewrite` branch: https://github.com/GoodBoyDigital/pixi.js/blob/dev-rewrite/src/core/renderers/canvas/utils/CanvasGraphics.js#L273

Thanks for contributing, make sure to read the [Contributing Guide](https://github.com/GoodBoyDigital/pixi.js/blob/master/CONTRIBUTING.md) for next time! The file you edited is generated at build time, so changes made in it are overwritten by the source files. We also don't accept pull requests to the `master` branch.

PR #1364, opened 2015-01-20
Added support to different retina prefixes (not only @2x)

krzysztof-o:
You may want to use different resolutions then @2x - for example I use @0.5x to use low quality textures for low-end devices.


englercj:
mind rebasing this onto the new v3 dev?

krzysztof-o:
I don't mind, but where is the new v3 dev branch? :)

englercj:
`dev` :)

GoodBoyDigital:
Nice @krzysztof-o ! Very handy to have :+1:  I just added it to v3. Thanks for this :+1:

PR #1388, opened 2015-01-28
Simplify InteractionManager into a working state, ignored touch events for now

serprex:
If there's interest in this pull request, I'll complete the touch events

If there's interest in removing buttonMode with a static function InteractionManager.setupButton that would take a displayobject & add events to it to achieve the same result, then I'll add that too. That'd imply turning the callbacks into events

To explain the code changes: remove dirty stuff & the bookkeeping that comes with it. Remove the update method. Remove __hit from DisplayObjects. Remove __isDown & __isRightDown. Instead we have a variables over & down of InteractionManager which tracks the target of those events

Instead of iterating over an array of interactiveItems, we have a method visitChildren. Instead of tracking interactiveChildren ourselves, we allow users to explicitly set interactiveChildren to false if they do not want visitChildren to iterate an element's children (as a way to optimize those cases, or to turn off a set of interactive children without turning off the individual interactive flags of those children)

visitChildren can be preemptively aborted with a non-false value. This is used once we've found a target

visitChildren works from deepest child up, so a child on top of a parent will have priority of being target. I maintained iterating children backwards

I also added a second optional parameter to the constructor for passing the renderer to the InteractionManager to have it call setTargetElement on immediately

### DIFF
--- a/src/interaction/InteractionData.js
+++ b/src/interaction/InteractionData.js
@@ -16,13 +16,6 @@ function InteractionData()
     this.global = new core.math.Point();
 
     /**
-     * The target Sprite that was interacted with
-     *
-     * @member {Sprite}
-     */
-    this.target = null;
-
-    /**
      * When passed to an event handler, this will be the original DOM Event that was captured
      *
      * @member {Event}
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -1,10 +1,6 @@
 var core = require('../core'),
     InteractionData = require('./InteractionData');
 
-// TODO: Obviously rewrite this...
-var INTERACTION_FREQUENCY = 30;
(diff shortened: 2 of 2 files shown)


englercj:
Thanks @serprex, but I'm going to leave this up to @GoodBoyDigital. He is currently rewriting the interaction manager so not sure how much he wants to pull in for that.

If you are looking for a good place to help with v3, I would cruise [the milestone](https://github.com/GoodBoyDigital/pixi.js/milestones/v3.x) and fix specific bugs!

GoodBoyDigital:
Hi there @serprex . There was some really nice ideas in this pr that influenced the new interaction manager that I have just finished. You can check it out here if ya fancy: https://github.com/GoodBoyDigital/pixi.js/blob/dev/src/interaction/InteractionManager.js 

 Thanks very much for sharing this! :+1:

PR #1389, opened 2015-01-29
Update extras to use texture setter, update Text to use getter/setter for style/text

serprex:
I took the liberty to remove setText while making since change since Sprite dropped setTexture

### DIFF
--- a/src/text/Text.js
+++ b/src/text/Text.js
@@ -44,8 +44,8 @@ function Text(text, style)
 
     core.Sprite.call(this, core.Texture.fromCanvas(this.canvas));
 
-    this.setText(text);
-    this.setStyle(style);
+    this.text = text;
+    this.style = style;
 }
 
 // constructor
@@ -66,7 +66,6 @@ Object.defineProperties(Text.prototype, {
             if (this.dirty)
             {
                 this.updateText();
-                this.dirty = false;
             }
 
             return this.scale.x * this.texture.frame.width;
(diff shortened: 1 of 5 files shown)


src/text/Text.js
```diff
@@ -44,8 +44,8 @@ function Text(text, style)
 
     core.Sprite.call(this, core.Texture.fromCanvas(this.canvas));
 
-    this.setText(text);
-    this.setStyle(style);
+    this.text = text;
+    this.style = style;
```
englercj: Even though these will technically set the underscored tracker variables, we should put them in the ctor explicitly as well and document them as private members.

src/text/Text.js
```diff
@@ -161,20 +176,15 @@ Text.prototype.updateText = function ()
...
 
-    var outputText = this.text;
-
     // word wrap
     // preserve original text
-    if (this.style.wordWrap)
-    {
-        outputText = this.wordWrap(this.text);
-    }
+    var outputText = this.style.wordWrap ? this.wordWrap(this.text) : this.text;
```
englercj: Use `_style` internally to prevent object getter call overhead.

englercj:
Thanks for this! Will merge it as soon as the small notes I had are fixed. Really appreciate the work you have been doing for v3!

serprex:
I don't know if github notifies when a pull request receives new commits

Addressed small notes

englercj:
:+1:

Issue #1286, opened 2014-12-19
Setting resolution > 1 and using BlurFilter does not work correctly


vanillabass:
The blur filter only works as expected when the renderer resolution is 1.  Any quick fixes for this?

englercj:
Closing as duplicate of #1215.

PR #1397, opened 2015-01-30
Add resolution property to WebGLFilterManager

### LINKED ISSUE #1215: Filters and Custom Renderer Resolution Bug
Hello, so I've found a rather weird issue concerning filters and custom resolutions with Pixi.

See [here](http://jsfiddle.net/amadeus/j5c5wtq5/) for a quick example - click to toggle the filter on and off.

Essentially it seems like the entire scene gets scaled back down to 1x resolution, and attached to the bottom left corner. I've tried it with a couple other filters so it's not specific to the invert filter.

It's rather simple code - the only thing perhaps out of the ordinary is how I am instantiating my renderer:

```js
renderer = new PIXI.WebGLRenderer(
	446,
	250,
	{
		view: document.getElementById('canvas'),
		resolution: 3
	}
);
```

If I change resolution to 1, no problems at all.

amadeus:
This fixes #1215 and possibly #1286.

I made some assumptions about how to set things up, based on what I've seen before, however the big change is that I did add an optional param to the `WebGLFilterManager` constructor that may or may not be how you guys would like to make this change.

### DIFF
--- a/src/pixi/renderers/webgl/WebGLRenderer.js
+++ b/src/pixi/renderers/webgl/WebGLRenderer.js
@@ -192,7 +192,7 @@ PIXI.WebGLRenderer = function(width, height, options)
      * @property filterManager
      * @type WebGLFilterManager
      */
-    this.filterManager = new PIXI.WebGLFilterManager();
+    this.filterManager = new PIXI.WebGLFilterManager(this.resolution);
 
     /**
      * Manages the stencil buffer
--- a/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
+++ b/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
@@ -6,7 +6,7 @@
 * @class WebGLFilterManager
 * @constructor
 */
-PIXI.WebGLFilterManager = function()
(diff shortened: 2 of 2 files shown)


edankwan:
It doesn't resolve the problem completely. It only fixed the position and scaling. The resolution seems to be low with high resolution value

edankwan:
And I guess you can access the renderer "resolution" value by using "this.renderSession.resolution". Although, there should be a separated "resolution" property to handle some certain filters for better better performance.

amadeus:
@edankwan you're absolutely right that it doesn't scale the filters at all. I was mostly going for something that wouldn't completely bork the scene. I am open to more suggestions on this, but I am unsure of how to actually scale the filter in these cases.

Tomorrow I will look into using `renderSession.resolution` instead of passing it through the constructor, that seems to be more the 'pixi' way.

englercj:
Aye we would rather use the values on the renderer in case they change instead of passing them around.

amadeus:
@englerc in can make that change and clean up this PR so it can be re-used. I'll do that shortly.

PR #1398, opened 2015-01-30
RenderTexture fixes

### LINKED ISSUE #1322: Resolution + RenderTexture Bug
I have found yet another issue when using the resolution property with RenderTexture class. I believe this issue may be somewhat related to issue #1215 since the results are somewhat the same (some botched transformations).

The premise is quite simple. I created a two sprites. One from an image, the other based on a RenderTexture that was rendered from the sprite.

If I leave the RenderTexture resolution to the default of 1, everything works as expected:

working example - resolution @ 1

However, if I change the resolution to 2 when I instantiate the RenderTexture, I get this result:

broken example - resolution @ 2

Here is the example fiddle, I left a comment on the line that can be played with:

http://jsfiddle.net/amadeus/s87s1c3w/

amadeus:
This PR fixes #1322

I also did a bit of cleanup in the `resize` method as I noticed it was logically doing things differently from the constructor, which could lead to a lot of confusion if anyone ends up using it.

### DIFF
--- a/src/pixi/textures/RenderTexture.js
+++ b/src/pixi/textures/RenderTexture.js
@@ -153,8 +153,8 @@ PIXI.RenderTexture.prototype.resize = function(width, height, updateBase)
 
     if (updateBase)
     {
-        this.baseTexture.width = this.width;
-        this.baseTexture.height = this.height;
+        this.baseTexture.width = this.width * this.resolution;
+        this.baseTexture.height = this.height * this.resolution;
     }
 
     if (this.renderer.type === PIXI.WEBGL_RENDERER)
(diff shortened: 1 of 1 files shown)


amadeus:
Any feedback on this? Is it perhaps too many changes to make it an easy merge?

englercj:
It isn't that, @GoodBoyDigital is rewriting the resolution code so much of this may not make sense once he is done.

amadeus:
@englercj ahh I see. Is this resolution rewrite for the 2.x branch? or for 3.0?

englercj:
3.0, for sure. We are only doing bug fixes for v2.x as needed.

I just realized you pulled this to v2.x >.<

amadeus:
You're the best! Thanks! :+1:

PR #1399, opened 2015-01-30
Fix WebGLFilterManager to use renderSession.resolution

### LINKED ISSUE #1215: Filters and Custom Renderer Resolution Bug

amadeus:
Weird, I thought if I fixed the branch and force pushed, it should've updated #1397.

Perhaps because it was closed?

Anyways, this has been updated as per the discussion.

And for reference, it fixes #1215

### DIFF
--- a/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
+++ b/src/pixi/renderers/webgl/utils/WebGLFilterManager.js
@@ -286,7 +286,7 @@ PIXI.WebGLFilterManager.prototype.popFilter = function()
 
     gl.bufferSubData(gl.ARRAY_BUFFER, 0, this.uvArray);
 
-    gl.viewport(0, 0, sizeX, sizeY);
+    gl.viewport(0, 0, sizeX * this.renderSession.resolution, sizeY * this.renderSession.resolution);
 
     // bind the buffer
     gl.bindFramebuffer(gl.FRAMEBUFFER, buffer );


englercj:
Looks good to me, +review @GoodBoyDigital

amadeus:
@edankwan had a good point about the filter's not scaling. Is there an easy fix I could add to this to make that happen? Or is that something far more in-depth?

amadeus:
Just a quick bump on this.

englercj:
Same comments as #1398, even if I merged it now. @GoodBoyDigital may take it out during his resolution rewrite.

GoodBoyDigital:
All fixed up in v3 :+1:
Default resolution for filters is the same as the renderer

amadeus:
@GoodBoyDigital This PR was meant for the `2.x` branch since I figured 3.0 would've fixed it in a different way. Is there a plan to backport a fix like this to `2.x`?

englercj:
Didn't even notice this was a PR to the 2.x branch, I'll merge it in.

amadeus:
You guys rock, thanks!

PR #1404, opened 2015-02-01
Auto line break with max width in BitmapText

mickez:
I needed this feature locally so I thought I might as well submit a pull request in case the feature is appreciated by others.


englercj:
Very cool, thanks!

PR #1412, opened 2015-02-04
'use strict';

serprex:
Haven't tested, did initial changes, wondering if there's a reason pixi.js doesn't use strict mode


englercj:
Because I would rather use something like this:  https://www.npmjs.com/package/strictify to insert user-strict and not have to remember to type it everywhere.

I'm waiting on this bug from jshint before using that: https://github.com/jshint/jshint/issues/924

PR #1418, opened 2015-02-08
Make use of module.exports & exports being predefined

serprex:


englercj:
Thanks but I would rather be consistent throughout the library. There isn't really any significant benefit to switching to make some use exports and others override the full module exports.

PR #1419, opened 2015-02-08
Supply optional webgl-enabled parameter

serprex:
webgl-enabled 1.0.3 introduces an optional parameter to avoid creating an unnecessary canvas

Also delete src/core/graphics/index.js as it appears to be an unintentional duplicate of src/core/index.js


englercj:
:+1:

englercj:
Looking at this now, this actually will break a lot of stuff. If we do this, then the next context we try to get will ignore our options and return the context that was created in this step. See:

```js
var canvas = document.createElement('canvas');

// create a context with no params
var gl1 = canvas.getContext('webgl');

gl1.getContextAttributes().stencil; // false

// create another from the same canvas with params
var gl2 = canvas.getContext('webgl', { stencil: true });

gl2.getContextAttributes().stencil; // false

// create another from a fresh canvas with params
var gl3 = document.createElement('canvas').getContext('webgl', { stencil: true });

gl2.getContextAttributes().stencil; // true
```

So if we use the one passed in by the user, we can't actually use any options that are not default. I have to revert this.

PR #1420, opened 2015-02-08
Fix temp text fix

serprex:
Follow up to https://github.com/GoodBoyDigital/pixi.js/issues/1254

I believe the descender was being drawn off the canvas

Calculating the descender could be optimized by scanning backwards for it, since the q is at the end of the string


GoodBoyDigital:
nice on @serprex !

PR #1421, opened 2015-02-09
Add listeners for touchleave and touchcancel.

chalgoman:


drkibitz:
(Sorry reposting on PR)
touchleave is part of an old draft, I wouldn't count on it. touchcancel is a good addition though.
http://www.w3.org/TR/touch-events/#list-of-touchevent-types

chalgoman:
Yes, it's part of an old draft, but some browsers may still dispatch it instead of touchcancel and touchend.

englercj:
I manually merged this to the `v2.x` branch. Please don't submit PRs to master in the future.

PR #1439, opened 2015-02-16
Matrix.applyInverse bug

DarkDev-:
Caches x and y from point so that the calculation is correct even when pos and newPos is the same PIXI.Point.


GoodBoyDigital:
nice catch! :+1:

PR #1441, opened 2015-02-17
Matrix.apply bug

serprex:
Have apply be consistent with applyInverse change by https://github.com/GoodBoyDigital/pixi.js/pull/1439


GoodBoyDigital:
nice! :+1:

Issue #1449, opened 2015-02-20
Set pixi.dev.js as value for 'main' within package.json


munkychop:
When using build tools, such as Grunt, alongside Browserify or Uglify it is preferable to target a non-minified PIXI file in order to avoid double minification, since this should be left to an individual's build implementation. So it would be better if the path to the non-minified version of PIXI was set as the value for the 'main' field within the package.json.

Current:
```JSON
"main": "bin/pixi.js"
```

Proposed:
```JSON
"main": "bin/pixi.dev.js"
```

Is there really any reason to use the minified file?

englercj:
Nope, and we don't in v3: https://github.com/GoodBoyDigital/pixi.js/blob/dev/package.json#L10

munkychop:
Cool, yeah I noticed it in v3, but you guys have been doing tons of work there, so I thought I'd ask before submitting a pull request.

englercj:
Yes, to get things into a v2 maintenance release pull them to `v2.x`, `dev` is the next release (v3).

PR #1452, opened 2015-02-20
Set pixi.dev.js as value for 'main' within package.json

munkychop:
Resolves issue #1449

When using build tools, such as Grunt, alongside Browserify or Uglify it is preferable to target a non-minified PIXI file in order to avoid double minification, since this should be left to an individual's build implementation. So it would be better if the path to the non-minified version of PIXI was set as the value for the 'main' field within the package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "type": "git",
     "url": "https://github.com/GoodBoyDigital/pixi.js.git"
   },
-  "main": "bin/pixi.js",
+  "main": "bin/pixi.dev.js",
   "scripts": {
     "test": "grunt travis --verbose"
   },


englercj:
Thanks! :+1:

drkibitz:
@englercj what would you think of actually making the main `src/index.js`? I'm probably not considering all the implications, but I'm posing the question since it's a rewrite and major version increase anyway.

munkychop:
@drkibitz this is going to be happening in v3 AFAIK. Take a look at the package.json within the dev branch.

drkibitz:
@munkychop Thanks :+1:

englercj:
@drkibitz It is pointing to the source in v3, but in v2 if we pointed to src browserify et al. wouldn't work. Our built v2 file uses UMD so browserify or r.js can pick it up.

PR #1459, opened 2015-02-24
Bit of docs, but more importantly some api fixes

alvinsight:
The containsPoint method exists so that if people want to build their own interaction manager they don't need to touch some pixi stuff such as worldTransform.applyInverse, etc...


englercj:
You're awesome!

Issue #1386, opened 2015-01-28
Double-click support


cmdelatorre:
Hello,

I wanted to know if there is any plan to provide a dbl-click interaction.

Currently, I've implemented it outside of PIXI, roughly following http://stackoverflow.com/questions/6330431/jquery-bind-double-click-and-single-click-separately/7845282#7845282

But I think it'd be great if PIXI provided the capability to distinguish click/dbl-click events, maybe even with a configurable delay.

Any ideas?

englercj:
Closing this since we decided not to do it.

PR #1462, opened 2015-02-25
Double-click support Issue #1386

adelciotto:
Added a double click event.

Example use:

```javascript
// set the delay between clicks
sprite.dblClickDelay = 200; // defaults to 300ms

// set the callback function
sprite.dblclick = function() {
....
};

// or

sprite.on('dblclick', function() {
....
});
```

**In the below gif, the tint changes when the double click event fires**
pixi_dblclick

Issue: https://github.com/GoodBoyDigital/pixi.js/issues/1386

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -5,6 +5,7 @@ var core = require('../core'),
 // TODO: Obviously rewrite this...
 var INTERACTION_FREQUENCY = 10;
 var AUTO_PREVENT_DEFAULT = true;
+var CLICK_DELAY = 700;
 
 /**
  * The interaction manager deals with mouse and touch events. Any DisplayObject can be interactive
@@ -75,6 +76,11 @@ function InteractionManager( renderer )
     /**
      * @member {Function}
      */
+    this.onDblClick = this.onDblClick.bind(this);
+
+    /**
+     * @member {Function}
+     */
     this.onMouseMove = this.onMouseMove.bind( this );
     this.processMouseMove = this.processMouseMove.bind( this );
 
@@ -120,6 +126,22 @@ function InteractionManager( renderer )
      * @private
      */
     this._tempPoint = new core.math.Point();
+
+    /**
+     * A click counter, used to check for double clicks.
+     * @member {Number}
+     * @private
+     */
+    this._clickCount = 0;
+
+    /**
+     * The timer ID for the double click timer.
+     * This timer is used to measure how long between each click
+     * to determine if the user double clicked.
+     * @member {Number}
+     * @private
+     */
+    this._doubleClickTimer = null;
 
     /**
      * The current resolution
@@ -174,6 +196,7 @@ InteractionManager.prototype.addEvents = function ()
 
     this.interactionDOMElement.addEventListener('mousemove',    this.onMouseMove, true);
     this.interactionDOMElement.addEventListener('mousedown',    this.onMouseDown, true);
+    this.interactionDOMElement.addEventListener('dblclick',     this.onDblClick, true);
     this.interactionDOMElement.addEventListener('mouseout',     this.onMouseOut, true);
 
     this.interactionDOMElement.addEventListener('touchstart',   this.onTouchStart, true);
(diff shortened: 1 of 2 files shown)


src/interaction/InteractionManager.js
```diff
@@ -414,12 +436,45 @@ InteractionManager.prototype.processMouseDown = function ( displayObject, hit )
...
+        {
+            this._clickCount++;
+
+            if(this._clickCount === 1)
+            {
+                this._dblClickTimer = setTimeout(function() {
+                    displayObject[ isRightButton ? '_isRightDown' : '_isLeftDown' ] = true;
+                    this.dispatchEvent( displayObject, isRightButton ? 'rightdown' : 'mousedown', this.eventData );
+                    this._clickCount = 0;
+                }.bind(this), displayObject.dblClickDelay);
```
englercj: creating a bound function every mouse event will be very expensive. Use a scope variable for context instead.

src/interaction/InteractionManager.js
```diff
@@ -414,12 +436,47 @@ InteractionManager.prototype.processMouseDown = function ( displayObject, hit )
 
     if(hit)
     {
-        displayObject[ isRightButton ? '_isRightDown' : '_isLeftDown' ] = true;
-        this.dispatchEvent( displayObject, isRightButton ? 'rightdown' : 'mousedown', this.eventData );
+        // only when the double click callback is defined we should perform
+        // the logic for checking double clicks
+        if(displayObject.dblclick)
```
englercj: This won't work if I do `displayObject.on('dblclick', ...);` since that doesn't set the event property, just adds a listener.

src/interaction/interactiveTarget.js
```diff
@@ -2,6 +2,7 @@ var core = require('../core');
 
 
 core.DisplayObject.prototype.interactive = false;
+core.DisplayObject.prototype.dblClickDelay = 700;
```
englercj: This seems like a really long time, basically if I am listening for clicks and doubleclicks the regular clicks are *significantly* delayed. 700ms is noticeable by the end user.

adelciotto:
I agree the 700ms is far too long, I have changed it to 300ms but we need some sort of delay to differentiate between a user performing a single and a double click, if they have both events defined and we were just simply using the browsers dblclick event, every time the user double clicks it would fire the single click event also.

englercj:
I'm trying to think if there is a way we can leverage browser behavior instead of emulating it. I'm still thinking on this one, sorry for the delay. Hopefully I can get some of this stuff done Saturday before I go out of town.

adelciotto:
It's all good take your time. I have some free time today, so I'll also try to think of something.

englercj:
Finally have some time this weekend to work on some projects (some life events have been coming at me lately). I'll make sure to check this one out and have a solution by next week.

Sorry for the delay @adelciotto I appreciate all your hard work!

src/interaction/InteractionManager.js
```diff
@@ -414,12 +437,52 @@ InteractionManager.prototype.processMouseDown = function ( displayObject, hit )
...
+        {
+            this._clickCount++;
+
+            if(this._clickCount === 1)
+            {
+                var self = this;
+
+                this._dblClickTimer = setTimeout(function() {
+                    displayObject[ isRightButton ? '_isRightDown' : '_isLeftDown' ] = true;
+                    self.dispatchEvent( displayObject, isRightButton ? 'rightdown' : 'mousedown', self.eventData );
```
wojciak: Code duplication with 466.

src/interaction/InteractionManager.js
```diff
@@ -414,12 +437,52 @@ InteractionManager.prototype.processMouseDown = function ( displayObject, hit )
...
+                    self._clickCount = 0;
+                }, displayObject.dblClickDelay);
+            }
+            else
+            {
+                clearTimeout(this._dblClickTimer);
```
wojciak: Not properly debounced - clearTimeout should happen on each click event, this will reduce CC.

src/interaction/InteractionManager.js
```diff
@@ -414,12 +437,52 @@ InteractionManager.prototype.processMouseDown = function ( displayObject, hit )
...
+                this.dispatchEvent( displayObject, 'dblclick', this.eventData );
+                this._clickCount = 0;
```
wojciak: This is not true as for 3 clicks we set the clickCount to 0. This violates the open-close principle if we want to ever extend the module.

src/interaction/InteractionManager.js
```diff
@@ -75,6 +75,11 @@ function InteractionManager( renderer )
     /**
      * @member {Function}
      */
+    this.onDblClick = this.onDblClick.bind(this);
```
wojciak: This should have a processDoubleClick method for handling the event.

wojciak: onDblClick should be renamed to onDoubleClick if we want to stick to it

wojciak:
Just a quick review, but from my experience this should be handled as a plugin or strictly on the consumer side. If we allow doubleClick, then we'll have:
- doubleTap
- pinch/zoom
- pincho/rotate
- fast doubleClick/tap
- swipe
- rotate
- pan

and we're writing http://hammerjs.github.io in the core.
Sorry for the complaint, here's a sad potato:
sad_potato

englercj:
@wojciak I kind of agree with that sentiment which is why I haven't merged it in. Been debating quite a bit over this one. I will be needing double click in an upcoming app so I will see how easy it is to extend this in a plugin to include more "gestural" type things.

bryanwillis7:
There is clearly a demand for double click support so it's sad to see that you guys have chosen to abandon it all together. Is there no plugin or any option at all for those of us who need to include it in our game?

englercj:
@bryanwillis7 We decided to let this live in user land, if all you need is double click it is pretty easy to use the click event and a timeout to do it. Otherwise, feel free to be the one that makes a plugin adding the multiple necessary events for full cross-platform support like mentioned here: https://github.com/pixijs/pixi.js/pull/1462#issuecomment-102520873

PR #1466, opened 2015-02-26
Added fromImage and fromFrame helpers to tiling sprites

alvinsight:
Also added a configuration file for Inch Ci

### DIFF
--- a/src/core/renderers/webgl/utils/StencilMaskStack.js
+++ b/src/core/renderers/webgl/utils/StencilMaskStack.js
@@ -6,8 +6,25 @@
  */
 function StencilMaskStack()
 {
+	/**
+     * The actual stack
+     *
+     * @member {Array}
+     */
     this.stencilStack = [];
+
+    /**
+     * TODO @alvin
+     *
+     * @member {boolean}
+     */
     this.reverse = true;
+
+    /**
+     * Internal count
+     *
+     * @member {number}
+     */
     this.count = 0;
 }
 
--- a/src/extras/TilingSprite.js
+++ b/src/extras/TilingSprite.js
@@ -520,5 +520,5 @@ TilingSprite.fromFrame = function (frameId,width,height)
  */
 TilingSprite.fromImage = function (imageId, width, height, crossorigin, scaleMode)
 {
-    return new Sprite(Texture.fromImage(imageId, crossorigin, scaleMode));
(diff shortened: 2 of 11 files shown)


src/extras/TilingSprite.js
```diff
@@ -483,3 +483,42 @@ TilingSprite.prototype.destroy = function () {
...
+ * @param imageId {String} The image url of the texture
+ * @param width {number}  the width of the tiling sprite
+ * @param height {number} the height of the tiling sprite
+ * @param [crossorigin=(auto)] {boolean} if you want to specify the cross-origin parameter
+ * @param [scaleMode=scaleModes.DEFAULT] {number} if you want to specify the scale mode, see {@link SCALE_MODES} for possible values
+ * @return {TilingSprite} A new TilingSprite using a texture from the texture cache matching the image id
+ */
+TilingSprite.fromImage = function (imageId, width, height, crossorigin, scaleMode)
+{
+    return new Sprite(Texture.fromImage(imageId, crossorigin, scaleMode));
```
englercj: Is this correct? I feel like `TilingSprite.fromImage` should return a `TilingSprite`, no?

src/core/renderers/webgl/utils/StencilMaskStack.js
```diff
@@ -6,8 +6,25 @@
  */
 function StencilMaskStack()
 {
+	/**
```
englercj: You don't have editorconfig installed do you...grr...

You should install it, it will enforce consistent whitespaces for us, otherwise you get stuff like this where there are tabs instead of spaces.

src/interaction/interactiveTarget.js
```diff
@@ -1,4 +1,5 @@
-var core = require('../core');
+var core = require('../core'),
+	extras = require('../extras');
```
englercj: ditto on the whitespace.

src/extras/TilingSprite.js
alvinsight: Argh, caught me red-handed, just fixed it

src/core/renderers/webgl/utils/StencilMaskStack.js
alvinsight: You're right I don't, I'll look into that

PR #1470, opened 2015-02-27
Bitmap text, cache as bitmap fixes, and deprecation

alvinsight:


src/deprecation.js
```diff
@@ -0,0 +1,37 @@
+var core = require('./core'),
+    math = core.math,
+    // plugins:
+    extras         = require('./extras'),
+    filters        = require('./filters'),
+    interaction    = require('./interaction'),
+    loaders        = require('./loaders'),
+    spine          = require('./spine'),
+    text           = require('./text');
+
```
englercj: I bet jshint is giving you unused variable warnings in this file, I think only 2 of these requires are being used. Make sure to run jshint before committing.

src/deprecation.js
```diff
@@ -0,0 +1,37 @@
...
+core.Stage = function ()
+{
+        window.console.warn("You don't need to use a PIXI Stage any more, you can simply render any container");
+        return new core.Container();
+};
+
+core.SpriteBatch = function ()
+{
+        window.console.warn("SpriteBatch doesn't exist any more, please use the new ParticleContainer instead");
```
englercj: Why do these warn, but not perform the previous action like Stage does? This needs to also have `return new core.ParticleContainer();`

src/deprecation.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+core.Sprite.prototype.setTexture = function ()
+{
+    window.console.warn('setTexture is now deprecated, please use the texture property, e.g : sprite.texture = texture; ');
+};
```
englercj: Need to add `this.texture = texture;` as well as the texture param here.

src/deprecation.js
```diff
@@ -0,0 +1,37 @@
...
+
+text.BitmapText.prototype.setText = function ()
+{
+    window.console.warn(" setText is now deprecated, please use the text property, e.g : myBitmapText.text = 'my text'; ");
+};
```
englercj: Need to add `this.text = text` as well as the text param here.

src/deprecation.js
```diff
@@ -0,0 +1,37 @@
...
+
+text.Text.prototype.setText = function ()
+{
+    window.console.warn(" setText is now deprecated, please use the text property, e.g : myText.text = 'my text'; ");
+};
```
englercj: Same as above.

src/deprecation.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.exports = {};
```
englercj: Is this necessary?

src/extras/cacheAsBitmap.js
```diff
@@ -17,6 +17,14 @@ DisplayObject.prototype._cachedSprite = null;
 
 Object.defineProperties(DisplayObject.prototype, {
 
+
+    /**
+     * Set this to true if you want this display object to be cached as a bitmap.
+     * This basically takes a snap shot of the display object as it is at that moment. It can provide a performance benefit for complex static displayObjects.
+     * To remove simply set this property to 'null'
+     * @member
```
englercj: `@member {boolean}` I think is what is wanted here.

src/extras/cacheAsBitmap.js
```diff
@@ -60,12 +68,16 @@ Object.defineProperties(DisplayObject.prototype, {
...
-                this.containsPoint = this._originalContainesPoint;
+                this.containsPoint = this._originalContainsPoint;
             }
         }
     }
 });
-
+/**
+* Renders a cached version of the sprite with WebGL
+* @param renderer {WebGLRenderer} the WebGL renderer
```
englercj: Make sure to have a line of space between the description and the jsdoc tags, makes it a lot easier to read.

src/index.js
```diff
@@ -1,9 +1,36 @@
...
+};
+
+function bundleUp(object1,object2){
+
+    for(var keys in object2){
+        object1[keys] = object2[keys];
+    }
+}
+
+module.exports = core;
```
englercj: I really liked the namespacing, when did we decide to do away with it? I made it easy to understand the plugin system and what things are optional and what are plugins.

Even if we were to do it this way, why build an array manually then loop through it? Why not just call `bundleUp` on the require directly?

src/loaders/bitmapFontParser.js
```diff
@@ -84,6 +85,8 @@ module.exports = function ()
 
             resource.bitmapFont = data;
 
+            text.BitmapText.fonts[data.font] = data;
```
englercj: This puts a hard dependency of the `text` plugin into `loader` now. If I exclude text from my pixi bundle with a command line param, this code will make it still be bundled anyway.

I think we should do this a different way. Like instead of bitmap text looking some global cache, just pass the loaded data into it.

src/deprecation.js
GoodBoyDigital: We could do this, but ParticleContainer is a different beast entirely. We could return a container though as that would work exactly the same.

src/deprecation.js
GoodBoyDigital: good shout!

src/deprecation.js
GoodBoyDigital: @alvinsight you cool to tweak this matey?

src/deprecation.js
GoodBoyDigital: yup!

src/index.js
GoodBoyDigital: Keen to get your thoughts on this one @englercj . I agree that the namespacing is nice for modular usage. Very keen to let people build the custom versions of pixi too (omitting things they don't need)

I also think that doing the above will make the transition from v2 - v3 super easy when it comes to general usage. Its a small thing but ```PIXI.Text``` is less effort than ```PIXI.text.Text``` with the added bonus that people who use v2 will already know how to use v3.

Maybe we make it optional? What do ya think the best way to achieve both these things are?

src/loaders/bitmapFontParser.js
GoodBoyDigital: Good shout! Will have a think about this one...

src/index.js
englercj: I like having things in their plugin namespace. If you want to alias things to the root object, it should be in the compatibility layer not the default v3 build imho. If the goal is easier compat with v2, it makes sense to be in the compat layer which people can opt **into** (not out of).

englercj:
:+1:

src/index.js
drkibitz: :+1: with @englercj

PR #1478, opened 2015-03-02
Namespacing discussion 

alvinsight:
I'm just opening this as a separate PR so we have a place to talk about it and get people's feedback.
Obviously the code can be shortened, I did it on purpose so people can look at the file and instantly see what it does.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,12 +1,36 @@
-var core        = require('./core'),
-    deprecation = require('./deprecation');
+var core = require('./core'),
+    math = core.math,
+    // plugins:
+    extras         = require('./extras'),
+    filters        = require('./filters'),
+    interaction    = require('./interaction'),
+    loaders        = require('./loaders'),
+    spine          = require('./spine'),
+    text           = require('./text'),
+    deprecation    = require('./deprecation');
 
-// plugins:
-core.extras         = require('./extras');
-core.filters        = require('./filters');
-core.interaction    = require('./interaction');
-core.loaders        = require('./loaders');
-core.spine          = require('./spine');
-core.text           = require('./text');
+var modules = [];
+
+modules.push(math);
+modules.push(extras);
+modules.push(filters);
+modules.push(interaction);
+modules.push(loaders);
+modules.push(spine);
+modules.push(text);
+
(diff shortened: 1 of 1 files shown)


englercj:
My comments oreviously:

https://github.com/GoodBoyDigital/pixi.js/pull/1470#discussion_r25544796

Fishrock123:
PIXI isn't _too_ large, so I think this could be reasonable. :)

tobireif:
PIXI.Text instead of PIXI.text.Text definitely would be an improvement.

englercj:
I have a feeling I'm the only one who cares about making plugins look like plugins, so merging to a single namespace.

ATN31:
Chad how feel about the new Virtual Internet?

englercj:
Just realized that this actually causes breakage for non-reference types. For example, the constants being copied to the core object actually copies the values over for a lot of them making a few constants (like RETINA_PREFIX) that people need to change unchangeable.

drkibitz:
-1 for const being copied.

If I were to try and appease both this issue and the other, I might try this:
```javascript
var core = module.exports = Object.create(require('./const'), {
    utils: {value: require('./utils')},
    math: {value: require('./math')},
    // ....
});
```
or just:
```javascript
var core = module.exports = Object.create(require('./const'));
core.utils = require('./utils');
core.math = require('./math');
// ....
```

As for users defining their own const values, I'm really not sure how this would be achieved in any case without building their own lib.

drkibitz:
Am I supposed to be able to do this?
```javascript
var pixi = require('pixi');
pixi.SCALE_MODES.DEFAULT = pixi.SCALE_MODES.NEAREST;
```

englercj:
Well instead of creating an object from the const object, we can extend it:

```js
var core = module.exports = require('./const');

core.utils = require('./utils');
core.math = require('./math');
```

But this only works if we never put any non-reference value in other core files. It really just makes more sense for these to be namespaced.

> Am I supposed to be able to do this?

Yes, that is necessary.

drkibitz:
> It really just makes more sense for these to be namespaced.

Fair enough, but get ready for the complaints ;)

PR #1484, opened 2015-03-04
Mesh and rope docs, typos

alvinsight:


GoodBoyDigital:
good stuff thanks @alvinsight !

src/deprecation.js
```diff
@@ -16,6 +17,16 @@ core.AssetsLoader = function () {
     window.console.warn("The loader system was overhauled in pixi v3, please see the new PIXI.Loader class.");
 };
 
+core.DisplayObjectContainer = function () {
+    window.console.warn("The Strip class has been renamed to Mesh, please use Container from now");
```
englercj: Pretty sure that is a typo!

src/extras/mesh/Mesh.js
```diff
@@ -384,10 +378,10 @@ Mesh.prototype.getBounds = function (matrix)
  * @static
  * @constant
  * @property {object} DrawModes
- * @property {number} DrawModes.TRIANGLE_Mesh
+ * @property {number} DrawModes.TRIANGLE_MESH
  * @property {number} DrawModes.TRIANGLES
  */
 Mesh.DrawModes = {
```
englercj: `DrawModes` looks like a class, can we change it to `DRAW_MODES`?

englercj:
:+1: Looks awesome dude, keep up the great work!

src/extras/mesh/Mesh.js
alvinsight: Yes, you're right we can, added in #1486

PR #1486, opened 2015-03-04
Renamed The DrawModes consts

alvinsight:
Even though I prefer it the other way, it matches the V3 changes on the other consts like BLEND_MODES, etc...


GoodBoyDigital:
on fire!

englercj:
:+1:

PR #1489, opened 2015-03-05
add some methods to Colormatrix filter

clementchenebault:


GoodBoyDigital:
Sweet! Great job @clementchenebault 
We have been in need of that LSD function for ages. Glad its finally arrived :+1:

englercj:
Nice! This is super awesome!

PR #1490, opened 2015-03-05
FXAA vert and frag filters were the wrong way around

woutercommandeur:
The fragment and vertex shader for the FXAA filter were the wrong way around, it worked, but their naming was rubbing me the wrong way :)


woutercommandeur:
This is me, nitpicking back at you :)

englercj:
Ha! Nice catch!

PR #1491, opened 2015-03-05
jshint housekeeping

woutercommandeur:
This fixes most of the complaints jshint gave. It also fixes some weird issue in SpriteMaskFilter where the fs module could only be required inside the constructor for browserify to work.

### DIFF
--- a/src/core/renderers/webgl/filters/SpriteMaskFilter.js
+++ b/src/core/renderers/webgl/filters/SpriteMaskFilter.js
@@ -1,6 +1,5 @@
 var AbstractFilter = require('./AbstractFilter'),
-    math =  require('../../../math'),
-    fs = require('fs');
+    math =  require('../../../math');
 
 /**
  * The SpriteMaskFilter class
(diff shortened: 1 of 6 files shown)


src/core/renderers/webgl/filters/SpriteMaskFilter.js
```diff
@@ -13,6 +12,7 @@ var AbstractFilter = require('./AbstractFilter'),
 function SpriteMaskFilter(sprite)
 {
     var maskMatrix = new math.Matrix();
+    var fs = require('fs');
```
englercj: Why was this moved out of the top requires? Should be outside the scope of this fn.

woutercommandeur: Because browserify failed and all other filters do it the same way

englercj: Not sure why browserify would fail, and others was me writing an example of how to do it and it got copy-pasted everywhere. fs should be required at the top of the file then used later on, not run in each constructor.

If there really is a browserify error, I'll fix it.

woutercommandeur: Try the current dev. It won't build. It does build when moving it into the constructor for some reason.

woutercommandeur: Want me to move it back for this pull request?

woutercommandeur:
moved it back to the top where it should be.

PR #1496, opened 2015-03-07
Move require('fs') out of the constructor to the top.

woutercommandeur:


GoodBoyDigital:
nice one @woutercommandeur :+1:

PR #1502, opened 2015-03-09
Moved the mesh folder to avoid the having to type pixi.extras.mesh.Mesh, plus tried getter-setters

alvinsight:


src/deprecation.js
```diff
@@ -17,15 +13,33 @@ core.AssetLoader = function () {
...
+
+    Strip: {
+        get: function ()
+        {
+            window.console.warn('The Strip class has been renamed to Mesh, please use Mesh from now on');
+            return new mesh.Mesh(texture);
+        }
+    }
+
+});
```
englercj: Is there a reason we want these to be non-writable?

englercj:
:+1:

GoodBoyDigital:
nice one Alvin! :+1:

englercj:
Why did we change some deprecated things to be non-writable properties and leave the other ones as-is?

GoodBoyDigital:
Yo! Its not to make them private. The purpose is so that anything that extends a DisplayObjectContainer would break. The below code now works ( but still outputs the deprecation message )

```
var myThing = function()
{
    PIXI.DisplayObjectContainer.call(this);
}

myThing.prototype = Object.create( PIXI.DisplayObjectContainer );
```

englercj:
@GoodBoyDigital Huh? That doesn't make sense, defining it as a property or adding it as a property shouldn't be any different in this case (also object.create should use the prototype). I find it hard to believe that this change fixes anything.

GoodBoyDigital:
Good catch @englercj . @alvinsight are you cool to tweak? We need to remove the 'new' keywords from the getters. Cheers dude!

PR #1507, opened 2015-03-10
Tidied up the destroy methods and moved the ticker class

alvinsight:


englercj:
:+1:

src/core/display/Container.js
```diff
@@ -544,3 +544,12 @@ Container.prototype.renderCanvas = function (renderer)
...
     }
 };
+
+/**
+ * Destroys the container
+ *
+ */
+Container.prototype.destroy = function ()
+{
+    this.children = null;
```
drkibitz: I might be missing an obvious case here, but I thought I ask anyway ;)

Could be be recursive, or maybe recursive with an arg `Container.prototype.destroy = function (recursive)`, or maybe another method?

At the very least, should this not call `removeChildren`?

englercj: It should definitely call the destroy method of children, and probably remove them as well.

GoodBoyDigital: Hey guys. I think we could add a 'destroyChildren' property to the function (much like the destroy texture function) That we we have both angles covered.

GoodBoyDigital: We should deffo call ```removeChildren``` too. That way we get rid of the parent references from the children.

englercj: Yeah, a param just like texture would be awesome.

PR #1508, opened 2015-03-10
More flexibility with multitouch gesture events (Advances, Optimization, ConconJS)

SirKnightDragoon:
Sorry for this new PR, but now it's very better.

* Add the global position on all changedTouches object
* Best way to gesture the double tap event on touchstart and getting each position.
* add optimisation and speed up the code
* add compatibility support for CocoonJS
* add more optimization


englercj:
We do not accept PRs to master. If this is a change you want to make to v2, then please do it in the `v2.x` maintenance branch. `dev`/`master` will both be v3 soon.

PR #1510, opened 2015-03-10
Only prepend textureUrl with baseUrl if not a Data URI

### LINKED ISSUE #1509: Sprite sheets don't work with Data URIs
If I try and load a sprite sheet JSON file, and in my image path I use a Data URI, the JSONLoader incorrectly prepends that Data URI with the url path, creating an invalid uri.

For example, if my spritesheet is hosted at /images/sprites/, and it contains

    {
         ...
         "meta": {
               "image": "data:image/png;base64,iVBO..."
         }
    }

Then the ImageLoader will make a request for "**/images/sprites/**data:image/png;base64,iVBO..."

This can be seen at [JSONLoader:117](http://www.goodboydigital.com/pixijs/docs/files/src_pixi_loaders_JsonLoader.js.html#l117).

    // sprite sheet
    var textureUrl = this.baseUrl + this.json.meta.image;          // <------
    var image = new PIXI.ImageLoader(textureUrl, this.crossorigin);

I think it would be a simple fix to first check if it's a Data URI, and, if it is, pass it through unchanged. Do you want me to create a PR for this?

sfentress:
When the JSONLoader loads an image, it should only prepend the baseUrl if the image url is not a Data URI.

This fixes #1509

### DIFF
--- a/src/pixi/loaders/JsonLoader.js
+++ b/src/pixi/loaders/JsonLoader.js
@@ -128,7 +128,10 @@ PIXI.JsonLoader.prototype.onJSONLoaded = function () {
     if(this.json.frames && this.json.meta && this.json.meta.image)
     {
         // sprite sheet
-        var textureUrl = this.baseUrl + this.json.meta.image;
+        var textureUrl = this.json.meta.image;
+        if (textureUrl.indexOf('data:') === -1) {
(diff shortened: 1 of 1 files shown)


englercj:
Nice catch, I need to add this to the loader in v3 as well!

PR #1516, opened 2015-03-12
Do not require raf polyfill, but instead add it to bundle

drkibitz:
- Also including new version of polyfill based on all references listed in header comment.
- Allowing browserfy to wrap and pass global reference

This is my stab at it. I looked far and wide all over the interwebs for a decent raf polyfill that was good for the purpose of pixi, and honestly wasn't completely happy with any of them. I also tried out 2 different polyfill services, as well as integration with said services through gulp plugins. Through all of that, I just thought wow, this is overkill.

Although cringing a little, I finally settled on looking at all of the references listed, and creating a custom version for pixi, and only including it in the bundle.

I went this direction because any polyfill should be included at the highest level possible, and for this lib, it's the distributed bundle. If anyone is using the bundle as is, it is a good assumption to think that they just want things to work without worry, and in all of the lib's supported browsers. It's only when users are building the library that they are opting out of the bundle provided, and polyfills should already be a part of this pipeline. I'm quoting somebody here (sorry can't remember who off hand), but "requiring" polyfills as modules is a anti-pattern. Although it is kind of a "module" here, it is technically an entry file, and included and executed at the beginning of the bundle.


englercj:
Only problem with this is if I `require('pixi.js')` in my app I don't get the polyfill since it is dynamically added by out bundle script. Meaning people using this in browserify will have to add their *own* polyfill which I don't think will be a good idea.

I know requiring polyfills is an anti-pattern, but why don't we require it in our entry file, but don't expose it.

```js
// use only side effects of this code running
require('./polyfills/...');

module.exports = {
    //....
};
```

woutercommandeur:
I agree I would like the polyfill there when I require pixi and not use the bundle.

drkibitz:
@englercj @woutercommandeur That wouldn't be too bad, it's nearly the same. I just didn't include the people NOT building their own versions yet still bundling, with the group using the bundle that just want things to work in supported browsers. But will make this change.

Side note, here's 2 services to get polyfills on your pages with User-Agent sniffing (polyfills will be different based on browser):
- https://cdn.polyfill.io/v1/polyfill.js
- https://cdn.jsdelivr.net/polyfills/polyfill.js

Example forced raf:
- https://cdn.polyfill.io/v1/polyfill.js?features=requestAnimationFrame&flags=always,gated

englercj:
Wow that cdn service is neat, never knew about it.

PR #1518, opened 2015-03-12
VideoBaseTexture bug fix

adireddy:
VideoBaseTexture.baseTextureFromVideo not avaliable, so changed it to VideoBaseTexture.fromVideo


englercj:
:+1: Thanks!

PR #1519, opened 2015-03-12
`copy` and `equals` methods for the Point object

miriti:
Hello!
I suggest to add `copy` and `equals` methods to the Point.
Why can't we have such useful methods in the Point object?
I followed the development before and saw that you didn't want to overload this object with unnecessary methods and suggested to create own extended Point instead. But...
Maybe it's just me but I use such operations A LOT. For example I use Point to represent position on the tilemap and in this case `equals` method could be very useful. But it seems to be a minor overkill to create own object for that purpose.
Also when you create some object and want to place it on the same position as the "creator" like this:

```
var o = new Entity();
o.x = this.x;
o.y = this.y;
```

I should be more efficent to make it like this

```
o.position.copy(this.position);
```

instead of 

```
o.position = this.position.clone();
```

Tell me if I'm wrong about all this stuff.

### DIFF
--- a/src/core/math/Point.js
+++ b/src/core/math/Point.js
@@ -36,6 +36,25 @@ Point.prototype.clone = function ()
 };
 
 /**
+ * Copies x and y from the given point
+ *
+ * @param p {Point}
+ */
+Point.prototype.copy = function (p) {
+    this.set(p.x, p.y);
+};
+
+/**
+ * Returns true if the given point is equal to this point
+ *
+ * @param p {Point}
+ * @returns {boolean}
+ */
+Point.prototype.equals = function (p) {
+    return (p.x === this.x) && (p.y === this.y);
+};
+
+/**
  * Sets the point to a new x and y position.
  * If y is omitted, both x and y will be set to x.
  *


drkibitz:
:+1: on equals. :-1: on `copy`, it sounds like a synonym to `clone` but is just a set. I would be happier if `set` accepted an object as well as coords.

miriti:
But `clone` creates a new object and `copy` will just copy `x` and `y` to the existing one. Probably it is not a significant impact on the memory usage but still..

miriti:
But you're right that it would be better if `set` accepted object instead of creating a new method...

drkibitz:
@miriti it was not a concern about memory, it was a concern about the name of the function, and say my preference would be to just modify `set` to do what you have `copy` doing.

englercj:
I'm a little torn. I like the idea of set doing both, and I also like the idea of it not being polymorphic so these fns can be highly optimized (many people will use point methods in tight loops).

drkibitz:
@englercj if that's the case, maybe adding a method like @miriti has done is best, but something other than `copy`? It's growing on me now.

miriti:
But what's wrong with `copy`? :) You don't like the name?

drkibitz:
There's a couple physics libs I've used that have these methods separated by `verb()` and `verbTo()`, such as `set` and `setTo`, `move` and `moveTo`. This is where I'm coming from I guess, but I get confused because in the case of those physics libs, the `verbTo()` was always took coordinates, while the `verb()` always took the objects. Also in said physics lib, `copy` is a special method where it returns a reference to one of the underlying objects in the pool. Anyway, just saying a bit about there I'm coming from here.

englercj:
@drkibitz Ah, I see. Personally I think having set/copy are OK, and IIRC consistent with what some game engines are doing that are built on pixi. Gonna merge this in.

Thanks for the thoughts all, was good discussion.

miriti:
@drkibitz now I see what you're talking about :)
@englercj thanks for merging!!

Issue #1522, opened 2015-03-12
Text rounding param


englercj:
Remind to add in text rounding param circa #1070. Make it optional via style/font params.

englercj:
PR merged to put this into v2.x, but still need it for v3

PR #1525, opened 2015-03-12
Fix for #1522 for v3

JanAmbrozic:
Patch that was merged at #1524 for v3.


englercj:
I think there is a little more required for v3 since styles are handled differently. Also looks like this PR is pretty dirty, not sure what happened!

JanAmbrozic:
Ok, I'll have a closer look tomorrow and issue a nicer PR :). Haven't used v3 yet so this is a perfect chance for that. Thanks for the tip!

PR #1526, opened 2015-03-12
Fix recursive destroy children

woutercommandeur:


englercj:
:+1:

alvinsight:
I'm not sure about this, if the child is a sprite (which is very likely) then the destroyChildren argument becomes a destroyTexture boolean.
The fact that we destroy a parent container doesn't mean we don't want to reuse the textures that its children use.

woutercommandeur:
Hmm.. Then maybe only call destroy on container and not sprites?

alvinsight:
I think we don't want to go down this path, if we do we'd need to check the type of every child to call the appropriate function, and have a special case for a sprite, or a graphics, or even a rope, assuming all of them have a destroy function.
To me this is looks quite convoluted just for a destroy function, @GoodBoyDigital , @englercj what do you think ?

woutercommandeur:
Then maybe the first argument of the Sprite destroy should be destroyChildren to keep it inline with it's parent class the Container.

englercj:
I agree with @woutercommandeur. Destroying children is almost necessary or people will leak and not understand why. But I also agree with @alvinsight that we don't want it to be too complex...

I feel like anything that inherits from `Container` should at least be able to have a destroy that acts like container's destroy, so first param is `destroyChildren` like in container that just makes sense from an oop point of view.

Following that, the challenge for me is then how do we allow players to destroy Textures as they are destroying children of a container? Which basically means we need to have the sprite version of destroy (texture/basetexture destroy params) also be in the Container method, and they are passed around until someone uses them. How do you feel about that?

woutercommandeur:
3 parameters with sane defaults sound fine with me.

Issue #1511, opened 2015-03-11
EventEmitter2


drkibitz:
I finally got around to trying to get a couple of PRs done and building things.

I noticed resouce-loader depends on EventEmitter2 which is included in the bundle, and wondering if we can just use that as the emitter pattern. Similarly to how pixi.js now shares a dependency on async with resource-loader. I know this may be unwanted for perf concerns, but I was concerned about having both `src/core/utils/eventTarget.js` and eventemitter2.js lingering around in the bundle.

englercj:
:+1: Do want.

ivw:
Maybe consider [EventEmitter3](https://github.com/primus/eventemitter3), it has better performance and the same API, but without some unnecessary functions.

drkibitz:
@ivw this issue is really more about code duplication, but if that's truly the case, maybe @englercj would consider using it here:
https://github.com/englercj/resource-loader/blob/master/package.json#L24

englercj:
@drkibitz Updated resource loader to use ee3 in v1.3.0, and bumped pixi in `dev` to use v1.3.0.

drkibitz:
@englercj that was fast ;)

englercj:
Closed by #1532

drkibitz:
@englercj I was actually working on this :( oh well, my bad not saying anything.

I did a take a different approach though. I kept eventTarget.js and pull in EventEmitter proto methods, but still implemented a different emit method to keep bubble prevention. So how is bubbling going to work with EE3?

englercj:
@drkibitz We took bubbling out of the eventTarget anyway. We decided we didn't want it to happen for arbitrary events, just for the interaction stuff. So the interaction manager actually does the bubbling of events, other events are normal observer pattern.

Sorry to rip it out from under you like that!

drkibitz:
@englercj got it, thanks

PR #1532, opened 2015-03-13
Convert the lib to use EE3

englercj:
This replaces the custom event implementation with [EventEmitter3](https://github.com/primus/eventemitter3).


englercj:
If merged, this would close #1511.

GoodBoyDigital:
Nice nice! Think we should also go through the code base and remove any ```bind()``` instances as this emitter class lets you pass in a scope property as a third param :+1:

englercj:
@GoodBoyDigital Good call, didn't even think about that. I'll do it now.

GoodBoyDigital:
:+1:

PR #1533, opened 2015-03-14
missed on, and use once

drkibitz:
ee3 doesn't have those methods


drkibitz:
I did a search for more across src and couldn't find any more.

englercj:
:+1: Boom! Nice catch!

Issue #1468, opened 2015-02-26
rendering spine animations that specify draworder keyframes result in misplaced bones


mreinstein:
I have an animation that I'm exporting from the spine editor program. It has many animations in it. All of the animations play correctly when previewed in spine. But when I run them through pixi, some of them have very screwed up bone positions. I've noticed that all animations which specify a `draworder` keyframe have this problem, while the animations that lack that type of keyframe run perfectly. 

I'm using pixi `v2.2.7`, and the newest spine editor (`2.1.11` essential)

If there's any other information I can post to make this easier to debug I'm happy to help!

mreinstein:
@englercj does the fact that it's been labeled as a bug mean this behavior has been determined to be a bug of pixi? I'm not sure if it is. Just curious if this has been investigated.

englercj:
No determinations, means "not operating correctly". The tag doesn't assign blame.

mreinstein:
ah ok, thanks for the clarification!

mreinstein:
fixed in 2.2.8! awesomeness!!

oarmstrong:
I'm seeing a similar issue, even when using 2.2.8 and with an animation without any draw order keyframes.  @mreinstein Would you mind sharing generic pseudo-code of everything between `new PIXI.Spine` to `stage.addChild()`?

mreinstein:
@oarmstrong I'm happy to help but given that your animations are giving you problems regardless of the presence of draw order keyframes, it's likely a separate issue. Please create another issue and ping me, we can pick up from there?

PR #1534, opened 2015-03-14
Fix Spine drawOrder.

ivanpopelyshev:
Issue #1468
Also, skeleton.drawOrder now is permutation of in integers, it is easier to use.


englercj:
merged into v2.x branch.

PR #1540, opened 2015-03-16
 More flexibility with multitouch gesture events

SirKnightDragoon:
Give the possibility to get the local position with your own global position.
Add the global position on all changedTouches object
Best way to gesture the double tap event on touchstart and getting each position.
This feature give us the possibility to manage multitouch event in the save object ex: Resizing Gesture event with two touch moving in the same time.
add optimisation and speed up the code


SirKnightDragoon:
It's done for you mister @englercj :) I love the new framework of the code, it took me a few minutes to understand what had changed between v2 and v3...

englercj:
Thanks for all the hard work, I really appreciate it!

The part I'm not crazy about in this PR is that it generates a lot of throw away objects. Each time a touch starts or ends we create a new Point object.

Also, isn't `getTouchData` already doing what you added in the touch start/end events?

SirKnightDragoon:
Ok @englercj right now it's very better ! :)

The idea is to send global position with all changedTouches. Currently the e.data.global just give the position of one touch in the elements, but if you want do some gesture event, like pressing double touch on one element and extend out to resize (pinch), you need to give also the global position of each others changedtouche.

So when you check e.data.originalEvent.changedTouches[0].globalX now with my modification you get the global position to a specific touch, and if you have double, you can get it on e.data.originalEvent.changedTouches[1].globalX. 

So for gesture you need to use:
* "e.data.originalEvent.touches.length" to know how many touches are in the elements.
* "e.data.originalEvent.changedTouches" to know which touches have ben changed
*  "e.data.originalEvent.changedTouches[0,1,2...].globalX and Y" to get the global position of the touch
*  "e.data.originalEvent.changedTouches[0].identifier to identify who really is the touch, because the order can be differences denpending of the traking of your tocuhes.

I'M currently working on a app using pixijs, and I have gesture event, zoom in and out, moving in a area with bounds limit, bouces and velocity depend on avanced speed traking of you touchMove etc.

Currently, I am on v2, but I'll pass v3 when it is more stable.

PR #1545, opened 2015-03-17
Fix an issue with loadFramedSpriteSheet

alaa-eddine:
loadFramedSpriteSheet will throw an error if the texture is not already loaded


englercj:
This looks like you tried to fix something in v2, but accidently made the PR to v3? This file doesn't exist in v3, if you want to PR something to the v2 maintenance branch use the `v2.x` branch.

alaa-eddine:
oh soory :)
was not aware about the v3 branch !
I'll submit the PR to v2

PR #1546, opened 2015-03-17
RetinaPrefix as part of rendering options

adireddy:
- added retinaPrefix as part of rendering options so that it can be customised

Not sure if this the right way to do it, but I thought providing a way to specify regex for retina assets is very useful.

Please review and let me know if this is ok.

I did try overwriting RETINA_PREFIX in constants and it works but it's a hack as it was defined as a static constant and you don't want to overwrite that at runtime.

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -128,6 +128,7 @@ module.exports = {
     DEFAULT_RENDER_OPTIONS: {
         view: null,
         resolution: 1,
+        retinaPrefix: /@(.+)x/,
         antialias: false,
         forceFXAA: false,
         autoResize: false,
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -94,15 +94,6 @@ module.exports = {
         NEAREST:    1
     },
 
-    /**
-     * The prefix that denotes a URL is for a retina asset
-     *
-     * @static
-     * @property {string} RETINA_PREFIX
-     */
-    //example: '@2x',
-    RETINA_PREFIX: /@(.+)x/,
-
     RESOLUTION:1,
 
     FILTER_RESOLUTION:1,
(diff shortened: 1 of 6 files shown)


englercj:
This actually doesn't make sense because the param doesn't have anything to do with the renderer, and causes unexpected behavior with two renderers, and having the renderer reach out and change utils based on a ctor param is a very leaky abstraction.

There isn't anything wrong with changing the pixi "constants" when your app start up. In fact, many of the values are there so that you can change them for your app if needed. For example, changing `SCALE_MODES.DEFAULT` to what your app will use most of the time is te expected way to use NEAREST filtering for all textures by default. Same with the retina prefix.

adireddy:
Cool. Will change in the application startup then.

Minor issue related to that. 

During the start up all the constants are being assigned to PIXI (https://github.com/GoodBoyDigital/pixi.js/blob/dev/src/core/index.js)

And for the retina prefix to work correctly I need to change PIXI.CONST.RETINA_PREFIX and the value assigned to PIXI.RETINA_PREFIX will not change.

Any issues with that?

englercj:
Yeah, that is an issue, you will need to change the .CONST one.

drkibitz:
@englercj yeah, why did this happen? https://github.com/GoodBoyDigital/pixi.js/commit/c2b7d3a62cf8854ab67237af8bb6c77c4daa0d7e

englercj:
@drkibitz Because I lost the fight of leaving everything in its own namespace and not moving it to the global export.

#1478 was open for about a week and I'm the only one who argued in favor of separate namespaces, so changes like this stayed in. I only just now realized that this will cause breakage for non-reference types.

englercj:
@adireddy This should be fixed in the latest dev branch now, we are exporting the const values properly so if you change PIXI.RETINA_PREFIX it should work fine.

adireddy:
Great. Thanks for letting me know.

Issue #1547, opened 2015-03-18
DisplayObjects using filters cause WebGL errors when moved off screen in v3.0.0-rc2


baseten:
See this fiddle: http://jsfiddle.net/arxx5zxj/3/

Appears to be because `this.frame` inside `RenderTarget` contains negative dimensions when attempting to render a `DisplayObject` that has been moved outside of canvas bounds, when using a filter on it (and obviously using `WebGLRenderer`).

Error is: `WebGL: INVALID_VALUE: viewport: size < 0`

Everything still renders correctly but the error causes a slower framerate.

GoodBoyDigital:
nice catch!

GoodBoyDigital:
all fixed now. Thanks for sharing! :+1:

PR #1550, opened 2015-03-18
clamped filterArea dimensions to positive values

baseten:
This prevents the WebGL error discussed in #1547 by clamping the viewport dimensions to positive values. I suspect you'd want to prevent the render pass completely if the viewport has 0 width and height, but I've not dug that deep myself yet.


englercj:
+@GoodBoyDigital For review.

PR #1551, opened 2015-03-18
Fix for loadFramedSpriteSheet crash when the texture is not loaded yet

alaa-eddine:
this was posted by error in v3 branch :)

### DIFF
--- a/src/pixi/loaders/ImageLoader.js
+++ b/src/pixi/loaders/ImageLoader.js
@@ -78,25 +78,63 @@ PIXI.ImageLoader.prototype.onLoaded = function()
 PIXI.ImageLoader.prototype.loadFramedSpriteSheet = function(frameWidth, frameHeight, textureName)
 {
     this.frames = [];
-    var cols = Math.floor(this.texture.width / frameWidth);
-    var rows = Math.floor(this.texture.height / frameHeight);
-
-    var i=0;
-    for (var y=0; y<rows; y++)
+    if(!this.texture.baseTexture.hasLoaded)
     {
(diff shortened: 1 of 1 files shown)


Moeberg:
I have had trouble with this function. Would some one care to give an example of correct usages of this function?

alaa-eddine:
I suggested this function for Pixi v1 because I needed it but didn't found any equivalent.
it mainly load a spritesheet and split it so you can use each sprite as separate pixi texture.

here is the original post with an example, it's quite old but will show you hot to use it http://www.html5gamedevs.com/topic/1278-loading-uniform-sprite-sheets-not-texturepacker-based/

hope it helps

Moeberg:
Yes I stumbled upon this thread earlier and it works correctly under those circumstances when frames are passed trough a callback function, but the function you wrote for the framework don't have that options which is what throws me off. 

Also the example you linked returns a empty array so I am not able to define in variable which is what I am trying to do.

alaa-eddine:
oh sorry here is an example :)


```javascript
var image = new PIXI.ImageLoader(image_url);
image.loadFramedSpriteSheet(width, height, 'mySpriteSheet');
//image.frames contains the loaded frames but you can't be sure that they are all available for this you need to listen for loaded event

pixiData.image.addEventListener("loaded", function (event) {
	var frames = image.frames; 

       //example of sprite instantiation using a the sprite at index 0
       var sprite = new PIXI.Sprite(frames['mySpriteSheet-0']);

}
```

Moeberg:
Thank you that helped a lot!

PR #1553, opened 2015-03-19
Feature/object assign

drkibitz:
Changed from extend module in favor of Object.assign polyfill using [this package](https://github.com/sindresorhus/object-assign). I know no syntactical ES6 features, but this one is an easy IE9+ polyfill and is always useful.

Also minor change to use `core.utils` ref instead of `utils` ref in core/index.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "async": "^0.9.0",
     "brfs": "^1.2.0",
     "eventemitter3": "^0.1.6",
-    "extend": "^2.0.0",
+    "object-assign": "^2.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "async": "^0.9.0",
     "brfs": "^1.2.0",
     "eventemitter3": "^0.1.6",
-    "extend": "^2.0.0",
+    "object-assign": "^2.0.0",
```
drkibitz: coincidence

englercj:
:+1: Nice!

PR #1554, opened 2015-03-19
Feature/docs

drkibitz:
Small tweaks to docs task.

I was starting to work on some fixes to the docs themselves, but it looks like things may still be in flux here and there so I didn't want to go assuming too much just yet. Ultimately would love a quick overview of what the intention is with the docs. A quick note about how I usually do it, ensure things are documented the public api rather than internal access.


englercj:
:+1:

PR #1563, opened 2015-03-19
Proper classes for the deprecation, that's hopefully going to settle the thing down :)

alvinsight:


src/deprecation.js
```diff
@@ -24,51 +24,56 @@ core.AssetLoader = function () {
...
+ * @see {@link PIXI.Container}
+ * @deprecated since version 3.0
+ */
+core.Stage = function (){
+
+    window.console.warn('You do not need to use a PIXI Stage any more, you can simply render any container.');
+
+    core.Container.call(this);
+
+    return this;
```
drkibitz: Don't return here.

src/deprecation.js
```diff
@@ -24,51 +24,56 @@ core.AssetLoader = function () {
...
+ * @name PIXI.DisplayObjectContainer
+ * @see {@link PIXI.Container}
+ * @deprecated since version 3.0
+ */
+core.DisplayObjectContainer = function (){
+    window.console.warn('DisplayObjectContainer has been shortened to Container, please use Container from now on');
+
+    core.Container.call(this);
+
+    return this;
```
drkibitz: Don't return here.

src/deprecation.js
```diff
@@ -24,51 +24,56 @@ core.AssetLoader = function () {
...
+ * @name PIXI.Strip
+ * @see {@link PIXI.Mesh}
+ * @deprecated since version 3.0
+ */
+core.Strip = function (){
+    window.console.warn('The Strip class has been renamed to Mesh, please use Mesh from now on');
+
+    mesh.Mesh.call(this);
+
+    return this;
```
drkibitz: Don't return here.

src/deprecation.js
```diff
@@ -24,51 +24,56 @@ core.AssetLoader = function () {
...
+core.DisplayObjectContainer.prototype.constructor = core.DisplayObjectContainer;
+
+
+/**
+ * @class
+ * @name PIXI.Strip
+ * @see {@link PIXI.Mesh}
+ * @deprecated since version 3.0
+ */
+core.Strip = function (){
```
drkibitz: +core.Strip = function (texture){

src/deprecation.js
```diff
@@ -24,51 +24,56 @@ core.AssetLoader = function () {
...
+    window.console.warn('The Strip class has been renamed to Mesh, please use Mesh from now on');
+
+    mesh.Mesh.call(this);
```
drkibitz: mesh.Mesh.call(this, texture);

src/deprecation.js
```diff
@@ -24,51 +24,56 @@ core.AssetLoader = function () {
...
-    }
-
-});
+/**
+ * @class
+ * @name PIXI.Stage
+ * @see {@link PIXI.Container}
+ * @deprecated since version 3.0
+ */
+core.Stage = function (){
```
drkibitz:
```js
core.Stage = function (backgroundColor) {
    global.warn('You do not need to use a PIXI Stage any more, you can simply render any container.');
    if (backgroundColor) {
        global.console.warn('In pixi.js 3.0, backgroundColor is a property of the renderer.');
    }
```

englercj:
:+1: To all of @drkibitz's comments.

Also, curious why this change came up. Was the other getter based method not working?

alvinsight:
Closing this to sync my branch properly and add other changes, but I'll reference it back :)

PR #1577, opened 2015-03-22
Feature/cleanup

drkibitz:
No need to define `module.exports = {}` for modules that don't export, modules are always provided with a default empty object via `exports`. Also no need to further export an empty module for up through the dependency tree just requiring will suffice.

The extra modules were getting confusing in how they were bypassing the separation of concerns. I didn't think they should be reaching into the core modules like that. The only one I left was the TextureUv reach in, but I was also tempted to expose that from core to get all of them. What I am trying to help here is maintaining the api contract set by core/index. Use what and how it exports its internal modules, rather than bypassing it.


drkibitz:
Went ahead and expose TextureUvs

englercj:
Looks great to me!  +@GoodBoyDigital for review

GoodBoyDigital:
looks good to me man! Nice work @drkibitz :+1:

PR #1580, opened 2015-03-22
RenderTexture now checks whether the display object its rendering has children.

adelciotto:
RenderTexture now checks whether the display object its rendering has children or not.

Issue: https://github.com/GoodBoyDigital/pixi.js/issues/1537

### DIFF
--- a/src/core/textures/RenderTexture.js
+++ b/src/core/textures/RenderTexture.js
@@ -270,7 +270,7 @@ RenderTexture.prototype.renderWebGL = function (displayObject, matrix, clear, up
 
         // Time to update all the children of the displayObject with the new matrix..
         var children = displayObject.children;
-        if (children)
+        if (children && children.length > 0)
         {
             var i, j;
 
@@ -329,7 +329,7 @@ RenderTexture.prototype.renderCanvas = function (displayObject, matrix, clear, u
 
     // Time to update all the children of the displayObject with the new matrix..
     var children = displayObject.children;
-    if (children)
+    if (children && children.length > 0)
     {
         var i, j;
 
--- a/src/core/textures/RenderTexture.js
+++ b/src/core/textures/RenderTexture.js
@@ -270,11 +270,18 @@ RenderTexture.prototype.renderWebGL = function (displayObject, matrix, clear, up
 
         // Time to update all the children of the displayObject with the new matrix..
         var children = displayObject.children;
-        var i, j;
-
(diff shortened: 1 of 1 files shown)


englercj:
@GoodBoyDigital Looks good to me, thoughts?

src/core/textures/RenderTexture.js
```diff
@@ -270,11 +270,18 @@ RenderTexture.prototype.renderWebGL = function (displayObject, matrix, clear, up
 
         // Time to update all the children of the displayObject with the new matrix..
         var children = displayObject.children;
-        var i, j;
+        if (children)
```
drkibitz: Do you mean for this condition to allow for `children.length == 0` ?

src/core/textures/RenderTexture.js
```diff
@@ -322,11 +329,18 @@ RenderTexture.prototype.renderCanvas = function (displayObject, matrix, clear, u
 
     // Time to update all the children of the displayObject with the new matrix..
     var children = displayObject.children;
-    var i, j;
+    if (children)
```
drkibitz: Do you mean for this condition to allow for `children.length == 0` ?

englercj:
Good catch @drkibitz I missed that.

adelciotto:
yeah cheers @drkibitz, silly mistake by me.

GoodBoyDigital:
hey there! thanks for this PR @adelciotto . Im not 100% sure what problem this is solving mind? the displayObject being rendered should not have it own transform called as it would override the matrix passed to it IYSWIM?

adelciotto:
Hey @GoodBoyDigital, I do see what you mean, I was attempting to solve this issue: https://github.com/GoodBoyDigital/pixi.js/issues/1537 . Although now I think what I have done here may be pointless, as DisplayObject is an abstract base class, all the renderable objects in the framework like Text, Sprite, Graphics etc all inherit from container, and therefore will always have a defined children array. I believe if something simply extends DisplayObject and and size of it is changed and is rendered to a RenderTexture, the boundaries won't update. Please correct me though if i'm wrong though, I'm still very much learning my way around the framework.

englercj:
@GoodBoyDigital Looks like you assigned the linked bug to yourself, so assigning this PR to you as well to do with as you will :)

GoodBoyDigital:
I think v3 should have all this stuff licked so closing for now. Thanks @adelciotto

Issue #1567, opened 2015-03-20
support Graphics.{beginBitmapFill, lineBitmapStyle}


FlorianLudwig:
API:

 * lineBitmapStyle(bitmap:BitmapData, matrix:Matrix = null, repeat:Boolean = true)
 * beginBitmapFill(bitmap:BitmapData, matrix:Matrix = null, repeat:Boolean = true)

as implemented in [flash](http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/Graphics.html)

Implementation is [WIP](https://github.com/GreyRook/pixi.js/tree/dev).

The GraphicsData class is growing quite a bit through this so we are going to introduce a `Brush` class.  Is this ok?

Adding everything to the constructor of GraphicsData we get:
```
function GraphicsData(lineWidth, lineColor, lineAlpha, lineBitmap, lineBitmapMatrix, lineBitmapRepeat, fillColor, fillAlpha, fillBitmap, fillBitmapMatrix, fillBitmapRepeat, fill, shape)
```
So we are looking to go for:
```
function GraphicsData(lineWidth, lineBrush, fillBrush, fill, shape)
```

with a new class `Brush` with the properties 
 * color
 * alpha
 * bitmap
 * bitmapMatrix
 * bitmapRepeat

And will grow further when we go for lineGradientStyle and beginGradientFill next ;)

*Edited* for clarification about the GraphicsData change and format fixed

GoodBoyDigital:
this looks interesting!

FlorianLudwig:
Implementation branch moved to [GreyRook:dev-bitmaps](https://github.com/GreyRook/pixi.js/tree/dev-bitmaps) based upon PR #1587 for refactoring `GraphicsData` with `Brush` classes.

mattkenefick:
+1

kasparsj:
+1

englercj:
Closing in favor of the Brush Refactoring PR, which is serving as my reminder to do this in v5.

PR #1587, opened 2015-03-24
Brush Classes refactoring

FlorianLudwig:
In preparation to implement more different line and fill styles we would like to refactor GraphicsData to use a new class to store brush details. See #1567


drkibitz:
My initial reaction is :+1:, just with a couple questions...

- What's the reason to use a getter for `canvasBrush`? Would a possible micro optimization (similar to caching the value), be setting `_canvasBrush` in a `tint` setter?

- Would using `"rgba(x,x,x,x)"` format for `context.strokeStyle` and `context.fillStyle` simplify anything here for _canvasBrush?

MishaAmashukeli:
Good points, thanks.
The getter for canvasBrush was there for lazy evaluation, because I did not want anything canvas-related to happen when the renderer is webgl-renderer. But for SolidBrush that does not really make sense, so I changed that.
I initially considered creating canvas and webgl related subclasses  but then decided that would create too many classes and too deep hierarchy. 

As for rgba vs setting context.globalAlpha - I'll try that out.

MishaAmashukeli:
Setting globalAlpha does not seem to have any performance issues. Using RGBA would force us to regenerate the value on each frame if we're animating the graphics alpha, so I think we don't need it.

MishaAmashukeli:
tint arg is not used by SolidBrush, but it will be used by the TextureBrush:

    TextureBrush.prototype.setTint = function (tint)
    {
        var texture = (tint === 0xFFFFFF)
            ? this.texture.baseTexture.source
            : CanvasTinter.getTintedTexture({texture: this.texture}, tint);
    ......

We can remove tint arg in SolidBrush.setTint and add a condition in CanvasGraphics.updateGraphicsTint to pass only tintR/tintG/tintB if the brush is SolidBrush, but imho this is not necessary... 
As far as jshint is concerned, this is not an error.

src/core/graphics/brushes/SolidBrush.js
```diff
@@ -0,0 +1,70 @@
...
+/*
+ * Generates _canvasBrush based on the tint color
+ *
+ * @param tint {Number} the tint color
+ * @param tintR {Number} the R component of the tint color
+ * @param tintG {Number} the G component of the tint color
+ * @param tintB {Number} the B component of the tint color
+ *
+ */
+SolidBrush.prototype.setTint = function (tint, tintR, tintG, tintB)
```
englercj: This seems wrong, and definitely does not follow the pattern in the rest of pixi.

Most likely this should just be a tint color param that uses `utils.hex2rgb()`. Having both params doesn't really make sense.

MishaAmashukeli:
Ok, changed it.

englercj:
I like the direction of this, @GoodBoyDigital can you review as well?

FlorianLudwig:
@englercj @GoodBoyDigital I would really like to see this merged, is there something left I could help with to move this forward?

stbaer:
+1 for getting this in.

GoodBoyDigital:
Hi @FlorianLudwig , apologies for not getting on this sooner. Busy times at the moment. I will get round to checking this out soon. In the mean time feel free to push forward with your branch as I'm keen to see where this goes!

englercj:
@FlorianLudwig Have you worked on this any more? I'm interested in merging this in, just curious if there has been more movement on it and if we should start a new PR that is off the latest develop with your latest brush-related code.

FlorianLudwig:
@englercj The intention was to enable bitmapfill and radiantfill in the long run.  The project where we needed those was solved with what is already available in pixi, so we did not pursue it any further.  But I still believe those APIs would be useful within pixi.  I will look if I can schedule someone to take a look at rebasing it and finishing the bitmapfill work.

englercj:
Anyway this can rebased onto the lastest `dev` branch?

```
git checkout brushes
git rebase dev
git push -f origin brushes
```

FlorianLudwig:
@englercj done

Kran67:
@FlorianLudwig Can you make an example to use gradient brush in graphics shape?

In this exemple : http://pixijs.github.io/examples/index.html?s=demos&f=graphics-demo.js&title=Graphics

Thanks

englercj:
This is something I want to do in v5, but I don't think it will happen in v4. I'll keep this open as a reminder and reference implementation though. Thanks!

bigtimebuddy:
I'm closing this since @englercj added #3123 which references this.

Issue #1591, opened 2015-03-24
New home for Spine


englercj:
Hey all,

Mat and I have decided we want to split the spine implementation to a plugin module, outside of the main repository. It would live at https://github.com/pixijs/pixi-spine and be initially based upon the current spine implementation.

We are looking for someone passionate about spine in pixi who would take this plugin and drive it forward from now on. Anyone interested in working on that?

Thanks!

drkibitz:
Sorry, haven't actually used the spine implementation in pixi for a project yet, but thought I'd give a :+1: for the direction ;)

ivanpopelyshev:
Ok, let me try it. I can watch for pull requests and update it for new versions of pixi if its necessary.

englercj:
I'll update this thread when I have the new plugin repo up and running.

GoodBoyDigital:
You the man @ivanpopelyshev :+1:

ivanpopelyshev:
Hey, Mat!
Nice to see you here :) Remember OGS2013?

GoodBoyDigital:
Absolutely! Mr Bombermine :D Hope you are well!

englercj:
I've setup the start of the new spine repo:

https://github.com/pixijs/pixi-spine

It is the v3 version of spine with the changes that @ivanpopelyshev has made recently merged in. Next steps are to remove spine from this repo in the v3 branch, and to publish the other spine implementation to be used here via npm.

ivanpopelyshev:
Stupid question: tabs or spaces? This files have both.

englercj:
I'll add an editorconfig, we prefer spaces.

ivanpopelyshev:
Ok, I'll look at it today or tomorrow. Got a busy weekend :)

englercj:
This repo has been updated to use pixi-spine now.

PR #1592, opened 2015-03-25
Fix for NaN duration values for 'looping' fix

adamdanielsdesign:


ivanpopelyshev:
That fix is for animations that change ik constraints, and its certainly working :)

englercj:
:+1: Are either of you interested in #1591 ?

adamdanielsdesign:
Maybe, I just started with my client, so <<< total noob more or less. However, if things work out I would be working with both spine and pixi on the reg. So I'd be happy lend a comment here and there.

PR #1609, opened 2015-03-27
Spine FlipX/FlipY timeline fix

ivanpopelyshev:
Again, my animation crashed against small error. Even webstorm shown that thing in warnings :) Who wants to check other modules with WebStorm?


englercj:
jshint would've also caught that :) (which is probably what webstorm is doing).

PR #1611, opened 2015-03-27
Fix resource caching only by url and never by name / key if provided

Inateno:
This wasn't working

```javascript
PIXI.loader.add( { name: "my_image", url: "complex/url/of/death.png" } )
PIXI.loader.load();

var sprite = new PIXI.Sprite.fromImage( "my_image" ); // only complex/url/of/death.png was working
```

Of course using the other way with url inside sprite still work if you didn't set a name or key when using add.

### DIFF
--- a/src/loaders/textureParser.js
+++ b/src/loaders/textureParser.js
@@ -8,8 +8,9 @@ module.exports = function ()
         if (resource.data && resource.data.nodeName && resource.data.nodeName.toLowerCase() === 'img')
         {
             resource.texture = new core.Texture(new core.BaseTexture(resource.data, null, core.utils.getResolutionOfUrl(resource.url)));
-            // lets also add the frame to pixi's global cache for fromFrame and fromImage fucntions
-            core.utils.TextureCache[resource.url] = resource.texture;
+            // lets also add the frame to pixi's global cache for fromFrame and fromImage functions
+            // retrieve resource.name to let possibility using key or raw url
+            core.utils.TextureCache[resource.name] = resource.texture;
         }
 
         next();


englercj:
While this seems a nice improvement, it actually isn't what `fromImage()` is for. It is for creating something from an image url.

If you want to use your key just do it:

```js
PIXI.loader
    .add('my_image', 'complex/url/of/death.png')
    .load(onAssetsLoaded);

funciton onAssetsLoaded() {
    var texture = new PIXI.Texture(new PIXI.BaseTexture(PIXI.loader.resources.my_image.data));
}
```

Inateno:
It's much longer and less clear when you have to deal with tons of Sprite :/

Also using this way doesn't change anything for those who still want use url am I wrong ?

Or maybe we could add a Sprite.fromTextureCache taking key as argument if you prefer separate them ?

englercj:
Honestly? It is like 15 more characters, and it *way* more explicit and obvious what you are doing instead of just throwing a magic string into pixi.

> Also using this way doesn't change anything for those who still want use url am I wrong ?

Yes, you are wrong. You're PR makes it cache by key not URL, so if I pass a URL in to fromImage for a loaded image it would load it again since the cache is by key instead of URL. You are taking about tightly coupling the loader implementation within pixi core which is something we really want to avoid.

URL is flexible, versatile, and easy to remember. We also build mechanisms to make it even easier to access. I really don't see what the big deal about using this:

```js
var sprite = PIXI.Sprite.fromImage(PIXI.loader.resources.my_image.url);
```

Or even better, use the textures the loader makes for you:

```js
var sprite = new PIXI.Sprite(PIXI.loader.resources.my_image.texture);
```

That really isn't "much longer" and is definitely more clear than magic strings.

englercj:
What I am trying to get accross is that URL will always work whether you use the loader or not. This PR makes it tied to how you load via the loader, but what you do with the loader shouldn't effect the core API.

Inateno:
No prob this is something we use for convenience, we often rename files or change url so this is helpful in our case.

About the PR I just thought it was a fix cause there is a key and name inside the loader, and also inside examples in comments but cannot use it at any moment when we need images, except by going manually inside the loader.

Just close the PR if this doesn't match with your usage.

englercj:
> we often rename files or change url so this is helpful in our case.

Then just use you key like I showed you:

```js
var sprite = new PIXI.Sprite(PIXI.loader.resources.my_image.texture);
```

I honestly don't understand why that isn't just as acceptable as what this PR tries to do? You can use the key just like I did here and not have to update code when you change your URL. You can access the loader from anywhere, and access the loaded resources from anywhere, is that not good enough?

I'm not trying to shut you down, I'm trying to understand why what we have isn't good enough.

Inateno:
The deal is I'm migrating our engine on PIXI, we use this kind of declaration, in the team there is no only core programmers, and if I have to explain they have to write "NS.PIXI.loader.resources[ "my_image" || my_var ].texture or .url, then they are lost, it is just less user friendly for us and I prefer to create an abstraction inside the loader, then you configure your loader once with a big json data, and just use key.
new Sprite( my_var ) is just faster and more maintainable if there is any change in the loader or urls. (imagine if later it's not this anymore but loader.resourcesHandler.cache for example, 2000 declarations to change in each games, *happy*).

But this is not the reason why I made this PR, I can do an overlay for my own needs, without discussing.
I made this PR because I thought this was the good way to do it in PIXI 'cause I read the comments over loader, talking about key, see in the code there is key attribute but cannot use it at the end (except with the way you show before) I thought it was a mistake, if this isn't the way you want for Sprite declaration (what I understand perfectly) maybe you should talk less about key in examples (or there is an other important usage I mess ? I'm also pretty new with PIXI and don't handle all the code yet).

englercj:
Why would your devs have to write out the full pixi stuff?

```js
NS.getResource = function (key) {
    return NS.PIXI.loader.resources[key];
};

var sprite = new PIXI.Sprite(NS.getResource(key).texture);
```

Or

```js
NS.createSprite = function (key) {
    return new PIXI.Sprite(PIXI.loader.resources[key].texture);
};

var sprite = NS.createSprite(key);
```

If you are already building abstractions, then let the core deal with how to get it out of pixi and present it to your devs however you want.

Inateno:
Yup, that's what I did. As I said my PR was not in this intention (I did abstraction before PR).

PR #1631, opened 2015-04-02
Added blend modes for particle renderer

TomSlezakowski:
Simple addition to ParticleRenderer to allow BlendModes to be used.

It will only interrogate the first particle, assuming they are all using the same.


woutercommandeur:
This can probably be closed, and doesn't neeed merging, @GoodBoyDigital added this yesterday.

PR #1645, opened 2015-04-09
cacheAsBitmap should respect .visible property

FletcherD:
It seems to me that even when an object is set with the cacheAsBitmap property, it should be able to respond to changes in its .visible property instantly, just as it already does with the .alpha property. This change checks the .visible property before rendering a cached bitmap, like other objects already do.


GoodBoyDigital:
nice one @FletcherD :+1:

PR #1651, opened 2015-04-11
Full 5x5 colorMatrix support

spaceribs:
Modified the colorMatrix fragment shader to support the full 5x5. An example of this fragment shader is implemented here:

http://codepen.io/spaceribs/pen/zxVLoe?editors=001


spaceribs:
I'm still a little confused about the difference between the 5th column which represents the offset for each channel in some of the documentation I've looked at, and the 5th row which is also an offset for each channel here: https://msdn.microsoft.com/en-us/library/windows/desktop/ms533875(v=vs.85).aspx

Seems like a lot of inconsistency anyway, we could just simplify this to a 5x4?

englercj:
@GoodBoyDigital Can you review this one? My matrix math needs some refreshing before I feel comfortable saying this is correct.

GoodBoyDigital:
Hi @spaceribs. bit late to the party here - but yes we would accept a pull request for 5x4 . cheers man!

spaceribs:
This should be good to go!

src/filters/color/ColorMatrixFilter.js
```diff
@@ -113,10 +109,10 @@ ColorMatrixFilter.prototype._multiply = function (out, a, b) {
 /**
  * Create a Float32 Array and normalize the offset component to 0-1
  *
- * @param matrix {array} (mat 5x5)
- * @return m { array } (mat 5x5) with all values between 0-1
+ * @param matrix {Array.<number>} (mat 5x4)
```
englercj: Prefer the `number[]` syntax, easier to read! :)

src/filters/color/ColorMatrixFilter.js
```diff
@@ -156,18 +150,19 @@ ColorMatrixFilter.prototype.brightness = function(b, multiply)
...
         scale, scale, scale, 0, 0,
-        0,0,0,1,0,
-        0,0,0,0,1
+        0, 0, 0, 1, 0
     ];
 
     this._loadMatrix(matrix, multiply);
 };
+//Americanized alias
+ColorMatrixFilter.prototype.grayscale = ColorMatrixFilter.prototype.greyscale;
```
englercj: Ha! I actually laughed when I saw this, good call!

.jshintrc
```diff
@@ -33,11 +33,10 @@
     "trailing"      : true,     // Prohibit trailing whitespaces.
     "maxparams"     : 8,        // Prohibit having more than X number of params in a function.
     "maxdepth"      : 8,        // Prohibit nested blocks from going more than X levels deep.
-    "maxstatements" : false,    // Restrict the number of statements in a function.
-    "maxcomplexity" : false,    // Restrict the cyclomatic complexity of the code.
     "maxlen"        : 220,      // Require that all lines are 100 characters or less.
     "globals"       : {         // Register globals that are used in the code.
-        "performance": false    // Depends on polyfill, simplifies usage
+        "performance": false,   // Depends on polyfill, simplifies usage
+        "PIXI": false
```
englercj: What was this added for?

spaceribs: Isn't PIXI the libraries' global? I added it to cover any example files.

src/filters/color/ColorMatrixFilter.js
spaceribs: cool, i'll switch that over now.

.jshintrc
englercj: Examples are in a completely different repo. This config is to lint the pixi source code which should not be using a global pixi object.

src/filters/color/ColorMatrixFilter.js
spaceribs: I was half thinking I should now go through and make colour aliases...maybe a little extreme though

Issue #1652, opened 2015-04-11
New frag based blurring is horizontally biased 


spaceribs:
The more you blur, the more the horizontal blurring takes precedence, here's an extreme example:

http://codepen.io/spaceribs/pen/yydxOd?editors=001

englercj:
Closing this since #1654 was merged.

PR #1654, opened 2015-04-12
Gaussian Blur concept

spaceribs:
This isn't in a great place yet, but I'm submitting it for feedback. BlurX and BlurY deprecated in favor of BlurDir, which will allow some nice motion blurs later. This should also be a partial fix (and inspired by) for #1652 but it's still:
1. A little too griddy.
2. Strangely still based on the viewport.
Anyway, I'm putting this in for more eyes and feedback.


spaceribs:
Leaving these 2 shadertoy filters here for some reference material:
* https://www.shadertoy.com/view/XdfGDH
* https://www.shadertoy.com/view/Mtl3Rj

src/filters/blur/BlurDirFilter.js
```diff
@@ -0,0 +1,143 @@
...
+BlurDirFilter.prototype.constructor = BlurDirFilter;
+module.exports = BlurDirFilter;
+
+BlurDirFilter.prototype.applyFilter = function (renderer, input, output, clear) {
+
+    var shader = this.getShader(renderer);
+
+    this.uniforms.strength.value = this.strength / 4 / this.passes * (input.frame.width / input.size.width);
+
+    if (this.passes === 1) {
```
englercj: Please fix the braces in this block to use the newline-first style used throughout the code base :)

src/filters/blur/BlurDirFilter.js
```diff
@@ -0,0 +1,143 @@
...
+
+BlurDirFilter.prototype = Object.create(core.AbstractFilter.prototype);
+BlurDirFilter.prototype.constructor = BlurDirFilter;
+module.exports = BlurDirFilter;
+
+BlurDirFilter.prototype.applyFilter = function (renderer, input, output, clear) {
+
+    var shader = this.getShader(renderer);
+
+    this.uniforms.strength.value = this.strength / 4 / this.passes * (input.frame.width / input.size.width);
```
englercj: Magic number `4` here, prefer a constant that describes what it is for.

src/filters/blur/BlurFilter.js
```diff
@@ -26,12 +35,14 @@ BlurFilter.prototype.applyFilter = function (renderer, input, output)
 {
     var renderTarget = renderer.filterManager.getRenderTarget(true);
 
-    this.blurXFilter.applyFilter(renderer, input, renderTarget);
-    this.blurYFilter.applyFilter(renderer, renderTarget, output);
+    for (var e = 0; e < this.blurFilters.length; e++) {
```
englercj: `e` seems like a random iterator, why not `i`?

englercj:
Looks good to me, had some minor nitpicks about style I mentioned. Also, wouldn't this supersede the BlurX/Y filters? We should remove them if so.

spaceribs:
Hey engle, thanks for taking a look!

Yeah I was definitely thinking about just taking them out entirely, the other option is hooking them up to 2 BlurDirs in opposite directions if you wanted to keep that.

I'll fix those style issues today, but I also kinda want to implement one of these high quality blurs as well (linked in the shadertoys). Let me know if you think we should make it an entirely separate filter.

GoodBoyDigital:
Nice work @spaceribs!How does this perform compared to our current implementation?

spaceribs:
I can set that up, do you guys have something for benchmarking I can use?

englercj:
@spaceribs No, we don't have a framework setup for benchmarking specifically.

englercj:
This looks good to me, if you remove the old filters that this replaces I will merge it.

GoodBoyDigital:
I would like to keep both for now peeps! At least until we have done some perf tests. ta!

englercj:
Done!

spaceribs:
Hey sorry, I've been working on other stuff all day, last weekend I also
worked on a separate fine blur based on the shaders linked above, I'll open
up another PR for it though

PR #1662, opened 2015-04-15
fixes #1617 Use bitmap font texture from spritesheet

### LINKED ISSUE #1617: Use bitmap font texture from spritesheet
To optimize number of draw calls I need to have all UI images (and bitmap font) on a single spritesheet.
Currently bitmapFontParser always forces you to load dedicated texture.

krzysztof-o:


JiDW:
Great !

Does it manage the resolution of the base texture ? I had extra code for that but I may be missing something you did.

Btw about the following line : 

 texture: core.utils.TextureCache[charCode] = new core.Texture(texture.baseTexture, textureRect)

It exists in PIXI too but do you see the point of "core.utils.TextureCache[charCode]" ? Because it can't manage multiple fonts and I don't see where it can be useful.

krzysztof-o:
I think resolution should work fine, but I'm not very experienced with that.

src/loaders/bitmapFontParser.js
```diff
@@ -51,68 +107,23 @@ module.exports = function ()
...
-            }
-
-            resource.bitmapFont = data;
-
-            // I'm leaving this as a temporary fix so we can test the bitmap fonts in v3
-            // but it's very likely to change
-            extras.BitmapText.fonts[data.font] = data;
-
+        var file = resource.data.getElementsByTagName('page')[0].getAttribute('file');
+        if (utils.TextureCache[file]) {
```
englercj: I don't think the cache checking logic is right, it should be using the full url not the url relative to the xml. So move line 117 to use below `var file` and check that way.

src/loaders/bitmapFontParser.js
```diff
@@ -1,8 +1,64 @@
 var Resource = require('resource-loader').Resource,
     core = require('../core'),
+    utils = require('../core/utils'),
     extras = require('../extras'),
     path = require('path');
 
+
+var parse = function(resource, texture) {
```
englercj: Why is this `var parse = function()` instead of `function parse()`?

src/loaders/bitmapFontParser.js
krzysztof-o: but url inside XML is usually relative to itself. So let's say we have font.xml and font.png in side in some folder, the path will be relative to xml.
I can easily change it but I checked both options and this one felt better.

englercj:
> Because it can't manage multiple fonts and I don't see where it can be useful.

I think that was a tag along from v2, we should just remove that assignment.

src/loaders/bitmapFontParser.js
krzysztof-o: I wasn't sure about your preferences. I'll change it.

englercj: the expression here is evaluated are runtime, the other is evaluated at parse time. Always prefer the named function unless you need the expression for some reason.

src/loaders/bitmapFontParser.js
JiDW: I used a way to know if the loaded png is from a frame even before checking the cache to prevent this.

Frame's name are not a true URL (because of TexturePacker) while classic .FNT file name is a relative URL.

englercj: Yes it is relative to the XML, meaning this check will be checking for `font.png` which will never have been loaded before, or worse it will have been loaded before but it wont be what you expect.

Consider:

```
/
`--font.png
    |--bitmapFonts/
    |   |--font.xml
    |   `--font.png
```

If you load `/font.png` then `/bitmapFonts/font.xml` the XML load will check for `font.png` and actually get back `/font.png` when you wanted `/bitmapFonts/font.png`.

The reason the code you have here doesn't work is *because* the url is relative to itself. The cache will be keyed by url of the item that was loaded (see [textureParser.js][tp]). Does that make sense?

[tp]: https://github.com/GoodBoyDigital/pixi.js/blob/dev/src/loaders/textureParser.js#L12

englercj:
:+1:

PR #1668, opened 2015-04-16
Use removeListener instead of remove in PIXI.Texture.destroy.

elmoeleven:
### Related To:
https://github.com/GoodBoyDigital/pixi.js/issues/1667

### Comments:
- Use `removeListener` over `remove`.


GoodBoyDigital:
Good catch @elmoeleven :+1:

PR #1673, opened 2015-04-17
Another short hand way for creating movieclips

chalgoman:
It's useful when you create a movieclip from sprites exported by https://www.codeandweb.com/texturepacker or other similar software and all frames for the movie clip are in single json file and has similar frame names. For example: frame0000.png, frame0001.png, frame0002.png, frame0004.png, etc.
Then you can just call PIXI.MovieClip.fromSettings(5, 'frame', '0000', 0, '.png');
Maybe the name "fromSettings" is not the best choise, but I could not think of a better name, so any suggestions for changes in the comment, function name, parameters names, etc are appreciated.

### DIFF
--- a/src/pixi/display/MovieClip.js
+++ b/src/pixi/display/MovieClip.js
@@ -205,3 +205,37 @@ PIXI.MovieClip.fromImages = function(images)
 
     return new PIXI.MovieClip(textures);
 };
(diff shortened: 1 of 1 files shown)


englercj:
:+1:

PR #1675, opened 2015-04-17
fix wrong uniform for loaded textures

gordonnl:


englercj:
If you are trying to submit to the v2 maintenance branch, please submit to `v2.x`. Otherwise submit to `dev` for v3 changes.

PR #1680, opened 2015-04-18
Refactor PIXI.extras.Ticker

drkibitz:
This refactor changes ticker to only request frames when necessary. Whether the ticker has been started and has listeners are the conditions that need to be met in order to request a frame.

The API has also changed and the ticker composes itself around an emitter instead of extending the class. This not only minimizes and simplifies the API, it also allows for the previous functionality without being overly complex.

`PIXI.extras.Ticker#autoStart` property was introduced, and `PIXI.extras.Ticker` no longer exports a singleton, but instead exposes `PIXI.extras.Ticker.sharedTicker` which `PIXI.extras.MovieClip` uses. This should allow easier testing (and hopefully general usage) in the future.

### DIFF
--- a/src/core/textures/RenderTexture.js
+++ b/src/core/textures/RenderTexture.js
@@ -434,11 +434,13 @@ RenderTexture.prototype.getCanvas = function ()
  */
 RenderTexture.prototype.getPixels = function ()
 {
+    var width, height;
+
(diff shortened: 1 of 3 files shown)


GoodBoyDigital:
Tidy! Nice work @drkibitz :+1:

PR #1682, opened 2015-04-19
More Ticker changes

drkibitz:
These additional Ticker refactor changes include:
- Separate update() from requesting an animation frame
    - This allows calling update() with a completely separate animation/ticking API if necessary or one chooses
- Complete all TODOs (documentation and more)
    - Added const.TARGET_FPMS
    - Added readonly property ticker.FPS
    - Added readwrite property ticker.minFPS
- added polyfilled "performance" to jshint globals
    - removed `global` in usages of it

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -37,6 +37,7 @@
     "maxcomplexity" : false,    // Restrict the cyclomatic complexity of the code.
     "maxlen"        : 220,      // Require that all lines are 100 characters or less.
     "globals"       : {         // Register globals that are used in the code.
+        "performance": false    // Depends on polyfill, simplifies usage
     },
 
     // == Relaxing Options ================================================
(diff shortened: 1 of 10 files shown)


drkibitz:
So this turned to be more than I thought, but hopefully they are welcome changes.
Overall what I was going for was a more centralized update loop and completely opt out from internal request animation frame ticks.

Basically a ticker is separate from what does the "ticking". I think there's more more work to be done here, but InteractionManager and MovieClip were the biggest offenders.

Let me know what you think when you get a chance.

drkibitz:
Also, while I was rooting around in InteractionManager, I thought I might do some refactoring/optimizations there, but since they would most likely would depend on the changes in this PR, so I'll hold off until I get some feedback.

woutercommandeur:
Looks good to me

GoodBoyDigital:
Thanks for the update @drkibitz :+1: 
Definitely keen on the idea of getting the ticker in Interaction Manager too. Nice and consistent :)

PR #1686, opened 2015-04-20
added blendModes to depracation

krzysztof-o:


englercj:
Mind adding scaleModes as well since you're in there?

englercj:
Love it, you're the best!

PR #1688, opened 2015-04-20
Update dependencies

drkibitz:
Been testing with this package.json for a few days so figure it would be safe.

The big jump is major 8 to major 9 with browserify.


drkibitz:
and 2 to 3 with watchify

englercj:
Thank you so much for doing this for me :)

PR #1693, opened 2015-04-21
movieclip loop fix

Noctivagus:
When one MovieClip has loop property set to false and ended its animation - it stops correctly but it also stops every other MovieClip on stage even those that have loop set to true. The playing property stays with true value but the update call has been cancelled.

```javascript
createExplosion(explosionTextures,true,100,100);
setTimeout(function(){
		createExplosion(explosionTextures,false,400,100);
}, 2000);

function createExplosion(textures,loop,_x,_y){

	var explosion = new PIXI.MovieClip(textures);
	explosion.loop = loop;
	explosion.test = true;
				
	explosion.position.x = _x;
	explosion.position.y = _y;
	explosion.anchor.x = 0.5;
	explosion.anchor.y = 0.5;
				
	explosion.rotation = Math.random() * Math.PI;
	explosion.scale.x = explosion.scale.y = 0.75 + Math.random() * 0.5
				
	explosion.gotoAndPlay(0);
				
	stage.addChild(explosion);
}
```

### DIFF
--- a/src/extras/MovieClip.js
+++ b/src/extras/MovieClip.js
@@ -71,6 +71,8 @@ function MovieClip(textures)
      * @readonly
      */
     this.playing = false;
+
+    this.update = this.update.bind(this);
 }
 
 // constructor


krzysztof-o:
+1 This is critical for us.

englercj:
I see, when it removes it removes the prototype method, which they all share. While this does fix the issue by making each unique, bound functions are really slow and I'm not convinced that this is the *best* way of fixing this issue.

I wonder if there is a better way of handling this... @drkibitz, thoughts?

drkibitz:
I agree bound functions are totally fine for most things this day and age, except for critical code paths like update loops, they are still just too much of an overhead. I would suggest creating a small scoped private method in constructor, which is still way faster in most if not all browsers, and use that with ticker e.g.:
```js
var _this = this;
this._update = function (time) {
    _this.update(time);
};
```

I would also like to note that I think this is a problem with [eventemitter3](https://github.com/primus/eventemitter3) as I would expect to have a third argument to remove by context, but it just doesn't have it.

There is an event api I still try to maintain, take a look at the [off](http://drkibitz.github.io/qi-events/api/latest/module-qi-events.Events.html#toc7) method there. Must more versitile in my opinion, which is why I still maintain it. I think the eventemitter3 api should work the same, especially since context already exists! Will file an issue there, and link to this issue.

englercj:
We shouldn't need to use a private `_this` scope, since `this` will be correct in the update method. But most likely we should create the update method in the ctor so it is unique to every object.

englercj:
If you ensure you have the latest dependencies (make sure to do `npm install` in the pixi.js folder) and then rebuild pixi this should now work. EventEmitter3 added API to remove a function by context so we are using that now, and it should work.

Let me know if you have any problems.

Noctivagus:
This did the trick. Thanks!

PR #1696, opened 2015-04-22
MovieClip.currentFrame renamed  because of confusing name

gprzybylowicz:
https://github.com/GoodBoyDigital/pixi.js/issues/1694

### DIFF
--- a/src/extras/MovieClip.js
+++ b/src/extras/MovieClip.js
@@ -58,11 +58,9 @@ function MovieClip(textures)
     /**
      * Elapsed time since animation has been started, used internally to display current texture
      *
-     * @member {number}
-     * @default 0
-     * @readonly
+     * @private
      */
-    this.currentTime = 0;
+    this._currentTime = 0;
 
     /**
      * Indicates if the MovieClip is currently playing
@@ -111,7 +109,7 @@ Object.defineProperties(MovieClip.prototype, {
         {
             this._textures = value;
 
-            this.texture = this._textures[Math.floor(this.currentTime) % this._textures.length];
+            this.texture = this._textures[Math.floor(this._currentTime) % this._textures.length];
         }
     },
 
@@ -124,7 +122,7 @@ Object.defineProperties(MovieClip.prototype, {
     currentFrame: {
         get: function ()
         {
-            return Math.floor(this.currentTime) % this._texture.length;
+            return Math.floor(this._currentTime) % this._texture.length;
         }
     }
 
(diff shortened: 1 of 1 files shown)


RoryPicko:
surely currentFrame makes more sense as it is the current frame being rendered and not the totalTime that movieclip will take to play through. Your merge request doesn't make sense to me unless you rename totalTime to currentTime

gprzybylowicz:
"doesn't make sense to me unless you rename totalTime to currentTime" - yep, maybe currentTime is better name, 
"only benefit is having an integer back instead of a float" - no, imagine that you have loop animation with 8 textures, after 1min currentTime is 60 but currentFrame is 4

RoryPicko:
I don't know where my comment has gone!? but the variable makes more sense now and you raise a valid point about having frames/time as seperate entities!

krzysztof-o:
I stumbled upon currentFrame yesterday and I confirm it is confusing. :+1:

src/extras/MovieClip.js
```diff
@@ -56,13 +56,13 @@ function MovieClip(textures)
     this.onComplete = null;
 
     /**
-     * The MovieClips current frame index (this may not have to be a whole number)
+     * Elapsed time since animation has been started, used internally to display current texture
```
drkibitz: If it is used internally, want to just make it internal? e.g. _currentTime, mark with @private

gprzybylowicz: i was wondering about that, but you can still set/get this value that's why i used name without underscore.

drkibitz: @gprzybylowicz this is a a convention in JavaScript in general. Marking with private and putting an underscore is means it is internal, and does not merrit a major version increase if the name changes, or any special treatment in deprecation etc..

gprzybylowicz: @drkibitz . I know this convention you mentioned but I meant I wasn't sure if this var should be private or not:)   ok, i However  I've rethought and agree with you, it makes sense to make it privete,  will change it soon

src/extras/MovieClip.js
```diff
@@ -56,13 +56,11 @@ function MovieClip(textures)
...
      *
-     * @member {number}
```
englercj: Still want the `@member {number}` tag, even when private.

gprzybylowicz: tag added

drkibitz:
@gprzybylowicz thanks a lot, much less confusing IMO ;)

PR #1697, opened 2015-04-22
SmartBlurFilter to use delta as uniform rather than const

superguigui:
Hey there !
I was toying with the SmartBlur filter and I noticed the delta param was used as a const in the fragment shader and I think it should rather be a uniform as you can end up with interesting results playing with this variable. What do you guys think ?
Thanks for the great work, I can't wait for the release of V3 :)


englercj:
I think it makes sense, though this change also changes the default value from `0.1, 0` to `0.1, 0.1` was that intentional?

superguigui:
Ah no sorry these were the values I was using in my exploration, let me reset it to the previous value. I restored to previous value :)

englercj:
noice!

PR #1700, opened 2015-04-22
Allow access to resource by its name

rayfranco:
Since ResourceLoader [is giving `resource.url` as a `resource.name` if none is given](https://github.com/englercj/resource-loader/blob/master/src/Loader.js#L212), I think it should be allowed to call a texture by it's name. I came across this use case, where only the Loader is aware of paths and it became tricky to give each objects its textures root path.

### DIFF
--- a/src/loaders/textureParser.js
+++ b/src/loaders/textureParser.js
@@ -9,7 +9,7 @@ module.exports = function ()
         {
             resource.texture = new core.Texture(new core.BaseTexture(resource.data, null, core.utils.getResolutionOfUrl(resource.url)));
             // lets also add the frame to pixi's global cache for fromFrame and fromImage fucntions
-            core.utils.TextureCache[resource.url] = resource.texture;
+            core.utils.TextureCache[resource.name] = resource.texture;
         }
 
         next();


englercj:
This has come up before with #1611, I didn't merge it because it hacks up what `.fromImage()` is really for.

`.fromImage(url)` is to build a sprite/texture/base-texture from a url. With this change it convolutes what this is really for, and the extra addition doesn't make sense without using our specific loader. The whole point of v3 was to split things away from eachother, so for example `.fromImage()` should have nothing to do with the internals of our loader (to the end user).

The fact that we cache urls for you so they don't load again is a bonus, but the user doesn't have to know that to use any non-loader APIs. However, this change makes the user have to understand how our loader interacts with internal caches to be able to use an API that has nothing to do with the loader.

Like I said in the previous thread, if you want to use the name just do it:

```js
var sprite = new Sprite(resources.myResource.texture);
```

That is why I am hesitant to make this change. I think it makes this function more confusing than helpful.

rayfranco:
@englercj I understand that it can be confusing at some point. I thought it was just a mistype, sorry I missed the other PR. Thanks for taking time to answer.

englercj:
@rayfranco No problem, thanks for being understanding. I'm willing to discuss it further if you feel strongly about it. I'm never afraid to change my opinion :)

rayfranco:
@englercj since I'm already overriding the Loader, it's easy for me to abstract this and have a `getTexture` from it. As @Inateno said in the previous thread, it is not very _user friendly_, but it makes sense. My steps before looking into the code was to:

* Try to put the name in `Texture.fromImage(textureName)`
* Checked out if no `Texture.fromName` method exists
* Then: `new BaseTexture(resource.data)` but thought is was aweful
* Then finally, while looking at the examples, I tried `Texture.fromFrame(textureName)` just in case
* After seing your comment, I chose to have an abstract getter in my custom Loader: `Loader.get(name).texture` which is just a little sugar :smile: 

Maybe a method `Texture.fromName` or `Texture.fromResourceName` could be more friendly to most of us but it will be linked to the internal of your loader, yet.

So for now, your solution is fine to me, but it should be documented somehow, because lot of people I've worked with so far uses names in manifests to load their assets and will probably face this problem, event if it is not one really.

englercj:
@rayfranco Interesting, I wonder if the loader plugin can add sugar methods to the texture prototype to make it easier to load from the resource cache.

If you want to add a `.get(key)` method to the loader for getting a resource by key a little shorter, just send me a PR and I'll merge it :)

https://github.com/englercj/resource-loader

PR #1702, opened 2015-04-22
Remove pixi-spine dependency. Tested on real project :)

ivanpopelyshev:
Dont merge it until we understand what to write in gulpfile to expose pixi.js.


ivanpopelyshev:
I tested pixi-spine build that used "require" => it included pixi.js in it. Now I need to expose pixi.js and use "external" in pixi-spine.
I didnt commit it in pixi-spine yet.

englercj:
Your bundle should mark pixi as external `bundle.external('pixi.js')` so it knows :)

ivanpopelyshev:
Where exactly do I have to insert to this line?

ivanpopelyshev:
What about resource-loader and async? Are they exposed? Do I have to include them? That way they'll appear in both pixi.js and plugin

englercj:
You put it in the `gulp/util/bundle.js` file where the bundler is created.

Resource loader is exposed via `PIXI.loaders.Loader`, but `async` is not exposed. That shouldn't be necessary in your plugin though, and if it is we can expose it via `PIXI.utils`. Now that I think of it we should probably expose ee3 via `PIXI.utils` as well for plugins to use without duplicating...

ivanpopelyshev:
Ok, just when I was ready to push changes into pixi-spine, I kind of pressed wrong button and deleted all the changes :( Now I have to redo it.

ivanpopelyshev:
Ok, pixi-spine works with this commit. Now you can merge it :+1:

ivanpopelyshev:
Yeah, and atlasSpineLoader was moved to pixi-spine, and pixi-spine has its own PIXI.spine.loaders.Loader that uses it.

ivanpopelyshev:
Ok, I updated pixi-spine in its repo, now it uses these two exposed classes:
PIXI.loaders.Resource
PIXI.utils.async

Binary file of pixi-spine does not contain that modules and it works with my demo.

PR #1718, opened 2015-04-24
Small inconsistency between Text and BitmapText

JiDW:
Setting a text now use toString(), like PIXI.Text.
Atm, creating a BitmapText from a number fails silently.

Note : first PR ever, so if I did something wrong, let me know.


englercj:
Nice catch, your PR is perfect :)

PR #1731, opened 2015-04-27
Fix touch datas for cocoonjs default configuration

JiDW:
When CocoonJS use a scale mode (default), global.x and global.y need to be divided by
the resolution of the interaction manager.

If you specify "no scale", it won't work properly. However, there is to my knowledge no way to detect the scale mode of CocoonJS and by default it is set to "ScaleToFill".

I believe that the default conf should work with PIXI.

Furthermore, CocoonJS defines "No scale" as a "Do it yourself" mode.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -779,6 +779,12 @@ InteractionManager.prototype.getTouchData = function (touchEvent)
     touchData.identifier = touchEvent.identifier;
     this.mapPositionToPoint( touchData.global, touchEvent.clientX, touchEvent.clientY );
 
+    if(navigator.isCocoonJS)
+    {
+        touchData.global.x = touchData.global.x / this.resolution;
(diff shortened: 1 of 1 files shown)


englercj:
I thought we were considering this a cocoonJS bug?

http://support.ludei.com/hc/communities/public/questions/202007585-Inconsistent-behavior-with-events-when-using-a-scale-mode?locale=en-us

JiDW:
At first I wanted to wait for an official answer from CocoonJS but they are not very active at this moment (no new build since 3 months) and we can safely assume that they will correct this bug by making "no scale" act the same way as "scale to fill", "scale aspect fill" and "scale aspect fit".

So I propose to make this change to PIXI to be compatible with the most used mode/default mode.

englercj:
Makes sense to me, thoughts @GoodBoyDigital ?

englercj:
le merge.

PR #1735, opened 2015-04-28
Fix e.data.getLocalPosition() weird behaviour. Fixes #1734

### LINKED ISSUE #1734: e.data.getLocalPosition() weird behaviour
Hi,

Lets say that I have a sprite at [0,0], when my mouse cursor is 100px on the right from the sprite then e.data.getLocalPosition(sprite) should return [100,0]. Instead it returns [100,~26]

Is this the expected behaviour? (Pixi 3)

wojciak:


englercj:
Good catch, sorry about that!

PR #1739, opened 2015-04-29
Fix in InteractionManager

naranjamecanica:
A display object can be destroyed before
InteractionManager.prototype.processInteractive is called, resulting in
an error because there’s no check if children still exist.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -352,6 +352,11 @@ InteractionManager.prototype.processInteractive = function (point, displayObject
     }
 
     var children = displayObject.children;
+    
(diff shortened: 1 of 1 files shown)


englercj:
Is this caused by destroying the object that you were rendering so that `processInteractive` gets called with the lastRendered object, which you destroyed?

Or am I misunderstanding this?

naranjamecanica:
Maybe it was an edge case, I'm creating and killing a lot of sprites, but I can't reproduce it right now anymore, lets close this and reopen if I run into it again.

englercj:
Ok, just let me know!

naranjamecanica:
Could this be related to #1739?

PR #1740, opened 2015-04-29
Simple support for using large or unbounded sets of textures over time

aschampion:
This PR fixes two issues when loading a large number of textures over time. For example, our use case is browsing massive image stacks of microscopy data, so while the size of textures in use at any given time is reasonable, a browsing session over several hours may cover dozens of GB of image data.

1. The `ImageLoader` for each texture is retained, which can accumulate memory and crash the browser.
2. The default texture caching strategy is cumulative, which while appropriate for games with a known, in-memory limit on the total number of textures required, does not fit our use case. Again this results in an OOM browser crash.

The current implementation solving 2, creating a LRU cache manager wrapping `BaseTextureCache`, is a bit rough, but I wanted to gauge interest in merging before cleaning it up since it meets our needs as is.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -63,6 +63,7 @@ module.exports = function(grunt) {
             '<%= dirs.src %>/extras/PIXISpine.js',
             '<%= dirs.src %>/textures/BaseTexture.js',
             '<%= dirs.src %>/textures/Texture.js',
+            '<%= dirs.src %>/textures/TextureCacheManager.js',
             '<%= dirs.src %>/textures/RenderTexture.js',
             '<%= dirs.src %>/textures/VideoTexture.js',
             '<%= dirs.src %>/loaders/AssetLoader.js',
--- a/src/pixi/loaders/ImageLoader.js
+++ b/src/pixi/loaders/ImageLoader.js
@@ -48,7 +48,7 @@ PIXI.ImageLoader.prototype.load = function()
 {
     if(!this.texture.baseTexture.hasLoaded)
     {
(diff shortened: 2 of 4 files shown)


englercj:
Interesting changes, sounds like you are not destroying the base textures when you are done with them though (which removes them from the cache).

https://github.com/GoodBoyDigital/pixi.js/blob/v2.x/src/pixi/textures/BaseTexture.js#L166

You will want to do `myTexture.destroy(true);` so the base texture is destroyed as well and both are cleaned out of the cache.

I like the `.once()` update, but I'm not conviced the LRU is the way to do this. Or more importantly that it seems out of Pixi's scope.

(P.S. you should upgrade to v3, it is easy and many app even "just work" and [warn you on what to change][dep]).

[dep]: https://github.com/GoodBoyDigital/pixi.js/blob/master/src/deprecation.js

aschampion:
Right, the textures are not destroyed because they may be quickly needed again if the user browses back over that part of the image stack (i.e., which textures will be reused has some locality but is not predictable), hence the need for a caching manager. But I agree it's probably outside of Pixi's scope, and it's easy to hack this into Pixi for those who do need it.

Should I close this PR and open one with just the `ImageLoader` fix?

[We've](https://github.com/acardona/CATMAID) been very happy with Pixi and will definitely upgrade to v3 in a future relase. Thanks for the great software.

GoodBoyDigital:
Hey guys - this all looks quite interesting. My two pence.. this could make a nice module / plugin for pixi?

PR #1753, opened 2015-05-05
Added better tap detection to the InteractionManager

naranjamecanica:
I’ve added a threshold of number of pixels a touch can move between a
down and up event, before a tap event is emitted.

In the game I’m making right now i have a draggable world, and every
object get’s an tap event when I start&stopping dragging on the object.
Using a threshold prevents is.

Let me know what you think and if it needs adjustment.


src/core/math/Point.js
```diff
@@ -66,3 +66,16 @@ Point.prototype.set = function (x, y)
...
+ *
+ * @param p {Point}
+ * @returns {boolean}
+ */
+Point.prototype.distance = function (p) {
+    var dx = p.x - this.x,
+        dy = p.y - this.y;
+    
+    return Math.sqrt((dx * dx) + (dy * dy));
+};
```
drkibitz: nitpick, newline.

src/interaction/interactiveTarget.js
```diff
@@ -29,7 +31,12 @@ module.exports = {
     /**
      * @todo Needs docs.
      */
-    defaultCursor: 'pointer',
+    defaultCursor: 'pointer',    
+    
+    /**
+     * The threshold in px used to check if a tap occured on the interactive object.
+     */
+    tapThreshold: 8,
```
drkibitz: Just curious about where you got this number?

src/interaction/interactiveTarget.js
```diff
@@ -42,5 +49,11 @@ module.exports = {
...
      * @private
      */
-    _touchDown: false
+    _touchDown: false,
+    
+    /**
+     * Used to track the global start position of a touchdown event on this target.
+     * @private
+     */
+    _touchDownPoint: new core.Point()
```
drkibitz: Pretty sure this object reference will be shared among all instances, see:
https://github.com/naranjamecanica/pixi.js/blob/dev/src/interaction/InteractionManager.js#L5-L8

Is this ok?

drkibitz: Actually answering my own question, since this is not a great pattern to promote. I would suggest making `touchDownPoint` a privately scoped static object within InteractionManager.js, and use that instead. No need to have it accessible on all display objects.

naranjamecanica: What if we have multiple touches, can we attach it to event data?

src/interaction/interactiveTarget.js
naranjamecanica: Experience.. feeling.. we make a lot of apps for kids, who always move their fingers before releasing, but are intending making a tap. So only 2px is to little.

src/interaction/interactiveTarget.js
naranjamecanica: Do you think it would be better to put it inside InteractionData?

drkibitz: Git it a shot ;)

My suggestion was based on the current implementation, but your new idea would require a bit more refactoring because currently the InteractionManager is trying to abstract the difference in touch and mouse away from the end API. I don't necessarily agree with this, so curious what you may have in mind.

drkibitz: Moving to a privately scoped object in InteractionManager is the same implementation you have now without a shared prototype reference. Moving to InteractionData would require more thought.

drkibitz:
Actually `getTouchData` confused me with a argument misnomer, it accepts a `Touch` object, not a `TouchEvent`, so it actually does process individual touches for each eventData.

This means you would need to put your point object specifically in `onTouchStart`, but you would have to figure the best way to work with the `interactiveDataPool` array while modifying objects within it on specific events.

But the problem is interactiveDataPool is shared with all events, so you'd have to ensure that you null out the point

drkibitz:
The main issue here is to basically add support for Gestures ;)

Would require using the `touches` array or the `touch.identifier`, instead of just `changedTouches`.

naranjamecanica:
I've dived a bit deeper in the inner workings, and see that the InteractionData objects have no state, and right now can not be tracked back to an earlier touch/click. 

Is it an idea to save the flags (_over, _touchDown) and other stuff like the touchstart position in this InteractionData? Or create an new data type TouchData which is only used internally.

drkibitz:
Currently InteractionData represents the data from a single touch, and it would not be the right places to store these references. A very quick brainstorm is leading me to creating a separate GestureData object to store references to specific InteractionData objects, and the InteractionData object can have a circular reference back to GestureData. This way from the event callback you have access to the GestureData through InteractionData, and any other InteractionData objects that are a part of this GestureData.

Again this is a very quick brainstorm ;)

drkibitz:
Maybe the proper place for an object like this would be directly on eventData...
Like I said, more thought... ;)

drkibitz:
GestureData would be more of a GestureRecognizer (hint, hint, UIKit)

naranjamecanica:
Yes, that would be very nice, all though I don't know if gestures are in the scope of PIXI, waht do you think?

Right now i'm going to solve this outside of PIXI in my click event handler of my Sprite instance.

JiDW:
Hmmm that would explain some "missed" tap we get on iOS, with the feeling that interaction isn't reliable. I didn't consider that the ".tap" event could be canceled by a unintentional move. Thanks for sharing !

drkibitz:
@englercj ping.

IMO, gestures are outside the scope of PIXI, but currently much of what InteractionManager provides I also believe is also outside the scope of PIXI, though in it's current form it is in a much better state than pixi v2. What I would like is for interaction manager to simply be a proxy to native events, and display objects themselves should handle the basic and abstracted `start`, `move` and `end` events. Which can be abstracted further in a gesture plugin. A gesture plugin would most likely be specific to pixi where a gesture recognizer would be added to a displayObject (in application code), and internally it would use the InteractionManager provided `start`, `move`, and `end` events to dispatch it's own gesture based events. Then the application level code can manage the gesture recognizer as needed.

src/interaction/interactiveTarget.js
englercj: I like the idea of this, but it should be called "radius" not "threshold" to be clear about what it does.

src/core/math/Point.js
englercj: :+1: Newlines at ends of file pl0x.

src/interaction/InteractionManager.js
```diff
@@ -701,7 +705,13 @@ InteractionManager.prototype.processTouchEnd = function ( displayObject, hit )
         if( displayObject._touchDown )
         {
             displayObject._touchDown = false;
-            this.dispatchEvent( displayObject, 'tap', this.eventData );
+            
+            // calculate the distance since the tap
+            var tapDistance = displayObject._touchDownPoint.distance(this.eventData.data.global);
```
englercj: What are the performance implications of running sqrt() so often?

src/interaction/interactiveTarget.js
englercj: Not sure I understand why the touchdown point is on the object being touched? Seems like this makes more sense to be a property in the manager.

englercj:
Definitely want to merge this as soon as the changes the comments mention are made.

src/interaction/InteractionManager.js
JiDW: I believe none.
https://jsperf.com/cheap

src/interaction/interactiveTarget.js
drkibitz: @englercj the issue is that if you do not relate the object based on the touch itself then multitouch  might make this flakely, if only have slightly unexpected behavior. But I don't know to what extent without testing myself.

naranjamecanica: Yes this is because of multitouch, although there's no check right now if it was the same touch event... An idea would be to make a pool of data objects which can store the touch related info and are related to an touch event through the touch event identifier. We can then fill them with info on touchstart, update them on touchmove and release them on touchend after we used the info.

src/interaction/InteractionManager.js
englercj: :+1:

englercj:
Any chance these changes we noted can be fixed so I can merge this?

drkibitz:
Based on our comment thread, I don't know if this should be merged in, @naranjamecanica. Were you still investigating a different solution? Want to close this one for now until you are ready with something different?

src/core/math/Point.js
```diff
@@ -66,3 +66,16 @@ Point.prototype.set = function (x, y)
     this.x = x || 0;
     this.y = y || ( (y !== 0) ? this.x : 0 ) ;
 };
+
+/**
+ * Returns the distance between the given and this point
+ *
+ * @param p {Point}
+ * @returns {boolean}
```
wojciak: Return value comment is incorrect, should be a number.

src/interaction/InteractionManager.js
wojciak: As for @englercj comment in the next file, I think we should keep consistency and make this a tapRadius.

englercj:
I'm going to close this since there were a few things that needed to change that never did, in favor of being part of #2071 .

PR #1759, opened 2015-05-07
updates text before calling getLocalBounds

gprzybylowicz:
when text has been changed properties width/height had wrong value, to fix it we need to update text first and then calculate local bounds


englercj:
Does `Text` need a change similar to this as well?

gprzybylowicz:
Don't think so, Text updates itself in width/height properties and bounds are calculated based on this._texture so all fields have correct values

englercj:
Awesome, thanks for this!

PR #1762, opened 2015-05-07
Fixed MovieClip when using negative animationSpeed

snc:
If you initially set the animationSpeed to a negative value (to play the animation backwards), the current implementation fails. This PR fixes this issue.

See http://jsfiddle.net/xLzgoqbL/ to reproduce the error.


englercj:
:+1:

PR #1766, opened 2015-05-08
update indices data before drawing

gprzybylowicz:
indices should be updated before drawing, otherwise i get error:
```[.WebGLRenderingContext]GL ERROR :GL_INVALID_OPERATION : glDrawElements: range out of bounds for buffer```


englercj:
:+1:

PR #1768, opened 2015-05-08
Rope supports textures from spritheet

gprzybylowicz:
Rope can use texture from spriteheets, not only standalone image

### DIFF
--- a/src/mesh/Rope.js
+++ b/src/mesh/Rope.js
@@ -47,6 +47,12 @@ function Rope(texture, points)
      */
     this.indices = new Uint16Array(points.length * 2);
 
+    /*
+     * @member {TextureUvs} Current texture uvs
+     * @private
+     */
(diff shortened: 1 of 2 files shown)


src/mesh/Rope.js
```diff
@@ -124,6 +116,34 @@ Rope.prototype.refresh = function ()
...
+ * Returns texture UVs
+ *
+ * @private
+ */
+
+Rope.prototype._getTextureUvs = function()
+{
+    if(!this._textureUvs)
+    {
+        this._textureUvs = new core.TextureUvs();
```
englercj: This changes the hidden class of Rope since it adds a property that didn't exist after the ctor ran. Please add this property at least with a null value to the ctor.

gprzybylowicz: done

englercj:
:+1: Nice!

Issue #1665, opened 2015-04-16
Graphics drawRoundedRect method radius parameter


adireddy:
http://jsfiddle.net/adireddy/qoxsr8t2/

In `graphics.drawRoundedRect` method, If I set the radius value with half the height (30 in this case), it's not working.
```js
graphics.drawRoundedRect(10, 90, 200, 60, 30); //not working`
graphics.drawRoundedRect(10, 90, 200, 60, 20); //working`

Uncaught TypeError: Cannot read property 'length' of null
```

stbaer:
Heres another fiddle: http://jsfiddle.net/f1gm72wa/

The error is gone without fill, but the rounded rect is not being drawn correctly if the radius is too big.

AleBles:
I also have this on v2.2.9

englercj:
Is this both canvas and webgl?

adireddy:
webgl only.

englercj:
This will get fixed by using #1772, this was a bug in PolyK.

adireddy:
:+1:

craigbennett1981:
Is there any update as to whether anyone is working on a fix for this issue? I am still seeing it in 3.0.9.

Issue #1755, opened 2015-05-06
Anonymous hole created inside rectangle


Priyank30:
I want to add holes inside graphics object and I have tried several ways but still not getting how to do it. Meanwhile I found some strange behaviors and could not able to figure out.

            var renderer = PIXI.autoDetectRenderer(1200, 800);
			document.body.appendChild(renderer.view);
			
			// create the root of the scene graph
			var stage = new PIXI.Container();
			
			var graphicsObject1 = new PIXI.Graphics();
			graphicsObject1.beginFill(0X60b044);	
			graphicsObject1.drawRect(100, 100, 600, 600);  //Green rectangle
			graphicsObject1.endFill();			
			stage.addChild(graphicsObject1);
			
			//An overlapping graphics object with hole.
			var graphicsObject2 = new PIXI.Graphics();
			graphicsObject2.beginFill(0XF41A1A);		
			
			graphicsObject2.moveTo(0, 0); //Red rectangle
			graphicsObject2.lineTo(0, 600);
			graphicsObject2.lineTo(600, 600);			
			graphicsObject2.lineTo(600, 0);
			graphicsObject2.lineTo(0, 0);
			
            //Creates hole in object (Don't know how !!!)		
     		graphicsObject2.arc(120, 100, 25, 0, 2 * Math.PI);
			
			graphicsObject2.endFill();
			
			graphicsObject2.position.x = 80;
			graphicsObject2.position.y = 5;			
						
			stage.addChild(graphicsObject2);			
			
			renderer.render(stage);

Above is a simple test code which gives output as below:
![issue](https://cloud.githubusercontent.com/assets/2055200/7491827/3585916e-f40b-11e4-8f2f-488ae6e7fa2c.png)

As shown in output an arc creates a hole inside red rectangle but when I draw another counterclockwise rectangle inside red rectangle it fills out with black color but it suppose to create a hole. To proper simulate hole test I have created another green rectangle behind red rectangle.

Priyank30:
@englercj 
I want to create a hole inside any 2d shape but wondering  how to do that. I have referred [1639](https://github.com/GoodBoyDigital/pixi.js/issues/1639) but not getting idea how to deal with 2d shapes. 
I have explained my situation in detail at [stackoverflow](http://stackoverflow.com/questions/30120352/pixijs-add-hole-into-graphics-object). Please help me out to move forward. I'm stuck at this point.

englercj:
[...]

PR #1772, opened 2015-05-08
Replace PolyK with Earcut

englercj:
Want to get this merged in, will leave open for a bit for review if anyone wants to help me out :)

@mourner - Finally happening, sorry it took so long!

Fixes: #1665


GoodBoyDigital:
nice! will take a look - do we know if this is faster?

mourner:
@englercj awesome, looking forward to see this in action! Would love to see some before/after benchmarks too. :) I'm sure it's faster, especially on shapes with lots of points, but a proof will be comforting.

@GoodBoyDigital other than performance, two more benefits you should be aware of:

- it's much more robust — e.g. won't crash on bad data and much more likely to triangulate correctly when dealing with degenerate geometries
- it supports holes in case you want to add support for them in future

englercj:
@GoodBoyDigital Yes, it is faster than PolyK plus the things that @mourner mentioned.

I want this to go in so we can draw polygons with holes in `Graphics` objects. I've gotten a couple asks for that now.

Priyank30:
@mourner 
I have previously used Earcut library with threejs to improve better triangulation calculation and performance and,  indeed it is a better in both prospects then threejs native triangulation algorithm. 
But in some of my test it fails on two case:
1) When holes are intersects
2) Hole entirely does not belongs to the shape.
Is there any fix for it?  
It might not be the correct place to ask about earcut solution but now I'm using PixiJs instead of Threejs and I found this thread because @englercj  has referenced it in my question #1755 so just want address some issue.

mourner:
@Priyank30 report the specific test cases in Earcut issues and I'll look if it can be helped. As far as I'm aware, no triangulation library can handle situations like that gracefully. 

The only reliable solution I can think of is to preprocess the data with something like http://sourceforge.net/projects/jsclipper/, correcting the topological errors with holes.

mourner:
@GoodBoyDigital @englercj updated the perf table to include PolyK: https://github.com/mapbox/earcut. Earcut is roughly 3-4 times faster.

Also released version 2.0.1 which adds support for Steiner points (additional vertices inside polygons).

GoodBoyDigital:
oooh! this is great news :D
Im convinced, thanks for making such a nice tool!

PR #1784, opened 2015-05-12
Fix Mesh Rendering to enable pixi-spine Mesh to render properly

simonTadram:
Mesh Rendering was not working at all (at least when created via pixi-spine)
With That fix Spine animation Mesh & skinned mesh will work (I need to create a pull request to the pixi-spine as well there were couple broken thing there too)

### DIFF
--- a/src/mesh/webgl/MeshRenderer.js
+++ b/src/mesh/webgl/MeshRenderer.js
@@ -1,5 +1,6 @@
 var ObjectRenderer = require('../../core/renderers/webgl/utils/ObjectRenderer'),
-    WebGLRenderer = require('../../core/renderers/webgl/WebGLRenderer');
+    WebGLRenderer = require('../../core/renderers/webgl/WebGLRenderer'),
+    Mesh = require('../Mesh');
 
 /**
  * @author Mat Groves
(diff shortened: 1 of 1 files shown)


englercj:
Wonderful!

+@GoodBoyDigital for review.

krzysztof-o:
Awesomeness!

ivanpopelyshev:
https://github.com/pixijs/pixi-spine/issues/10 - @Shadowstep33 confirmed that this version works for spine meshes.

gprzybylowicz:
your push destroys Mesh rendering e.g Rope doesn't work. 
This line ```  gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER,0, mesh.indices); ``` is needed

englercj:
@gprzybylowicz Did some git bisecting and it seems like this commit was the one that broke it: 49135508622ba99d72dbeeb5dabfb63d80f4c4dd.

Looking into fixing now.

gprzybylowicz:
@englercj  without a code line i mentioned and when you are working with multiple Rope objects, you will get WebGL error. However i'm not master of low-level webgl so maybe there is another way to fix it.

englercj:
Fixed these issues in 3b13a85ca7dbba286ae848db0b0ad629afb961d4, latest dev builds include these fixes.

krzysztof-o:
Now we're getting:
```javascript
points is undefined
if (points.length < 1 || !this.texture._uvs)
```

englercj:
Good catch! Fixed in 4bc6541.

krzysztof-o:
It doesn't throw any errors but it draws the shape only between first 2 points. It used to work before this commit (https://github.com/simonTadram/pixi.js/commit/4c037c4d6a4c67aa9023e6d4a2a4b17e2a44e39e). Any ideas?

englercj:
Moving the ctor call down is what caused the drawing errors, let me look at it a bit.

englercj:
Fixed in e8eb1c97b144062009d362fc347a277cbac5956a.

krzysztof-o:
I confirm, works like a charm!

fifobuffers:
Looks like there is still a problem with just rendering a simple rope mesh, no errors are being thrown but the mesh image is not being drawn (even using the code directly copy/pasted from the demo: http://pixijs.github.io/examples/index.html?s=basics&f=textured-mesh.js&title=Textured%20Mesh ), I'm having to use a slightly older v. of pixie right now and it works fine.

englercj:
The link to our example is working for me, do you have an example of one that is broken we can look into? If you do, would you mind opening a new issue so I can track it on the next release?

PR #1789, opened 2015-05-13
Don't try to grab new WebGL context if we have one already.

### LINKED ISSUE #1788: WebGL context mismatch on firefoxOS
Hi,

After leaving and resuming a game on firefoxOS I get a black screen and:
"Error: WebGL: Retrieving a WebGL context from a canvas via a request id ('webgl') different from the id used to create the context ('experimental-webgl') is not allowed."

I think the culprit is this line in the ._initContext call (which is called by handleContextRestored):
```
 var gl = this.view.getContext('webgl', this._contextOptions) || this.view.getContext('experimental-webgl', this._contextOptions);
```

I'm posting a quick solution by doing:
```
var gl = this.gl || this.view.getContext('webgl', this._contextOptions) || this.view.getContext('experimental-webgl', this._contextOptions);
```

But please check if it makes sens and doesn't break anything ;)

wojciak:
Probably fixes #1788

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -168,7 +168,7 @@ WebGLRenderer.glContextId = 0;
  */
 WebGLRenderer.prototype._initContext = function ()
 {
-    var gl = this.view.getContext('webgl', this._contextOptions) || this.view.getContext('experimental-webgl', this._contextOptions);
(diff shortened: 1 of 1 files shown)


englercj:
Thanks for bringing this up @wojciak. Unfortunately this is the wrong branch for PRs and I want to do a  little more than this to fix it up. I'll get something in asap to fix #1788.

PR #1791, opened 2015-05-13
Moved ticker to core

drkibitz:
Take it or leave it ;)

This was just getting annoying to me since it's a shared module across extras and interaction modules.


drkibitz:
Dang it

PR #1792, opened 2015-05-13
Moved ticker to core

drkibitz:
Take it or leave it ;)

This was just getting annoying to me since it's a shared module across extras and interaction modules.

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -13,6 +13,7 @@ var core = module.exports = Object.assign(require('./const'), require('./math'),
     // utils
     utils: require('./utils'),
     math: require('./math'),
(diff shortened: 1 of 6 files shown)


englercj:
I think this makes sense, @GoodBoyDigital?

englercj:
@drkibitz Can you put something in the deprecation file with this since it is changing namespaces? If you put that in I will merge it :)

drkibitz:
@englercj if you look closer, you'll notice that it actually didn't change namespaces :neckbeard:. It just moved module location (to reflect it's usage + namespace).

englercj:
oh you got me there!

Issue #1790, opened 2015-05-13
[v3] Uncaught TypeError: CanvasGraphics.renderGraphics is not a function


miguelmoraleda:
I just migrated to v3 and I'm getting this error. 

The renderGraphics function is being called in Graphics.js 823.

The renderGraphics function doesn't exist on CanvasGraphics.

miguelmoraleda:
The issue is only present when using CanvasRenderer. Is it deprecated?

englercj:
It most certainly does:

https://github.com/GoodBoyDigital/pixi.js/blob/master/src/core/renderers/canvas/utils/CanvasGraphics.js#L16

Do you have a running example of this?

miguelmoraleda:
You are totally right, it does.

Not sure why I'm getting this error when trying to use CanvasRenderer, no issues when ussing WebGLRenderer. 

I'm gonna publish an example of this ASAP and let you know.

englercj:
That method is only called with canvas renderer, it is how graphics are drawn to the canvas. You won't see that call in WebGL rendering because it doesn't use the CanvasGraphics object at all for that.

Looking forward to you test case.

wojciak:
You guys! Look!

englercj:
This was a regression, got it fixed in latest dev. Will hotfix shortly.

miguelmoraleda:
@englercj Thanks! Sorry, I couldn't build the test case.

Issue #1778, opened 2015-05-11
[v3] Sprite.prototype._renderCanvas chop size bug


ZebraRoy:
I found that the sprite render incorrectly in CanvasRenderer at retina. And I found a code in Sprite.prototype._renderCanvas:

```javascript
if (this.tint !== 0xFFFFFF)
        {
            if (this.cachedTint !== this.tint)
            {
                this.cachedTint = this.tint;

                // TODO clean up caching - how to clean up the caches?
                this.tintedTexture = CanvasTinter.getTintedTexture(this, this.tint);
            }

            renderer.context.drawImage(
                this.tintedTexture,
                0,
                0,
                width * resolution * renderer.resolution,
                height * resolution * renderer.resolution,
                dx,
                dy,
                width * renderer.resolution,
                height * renderer.resolution
            );
        }
        else
        {
            renderer.context.drawImage(
                texture.baseTexture.source,
                texture.crop.x * resolution,
                texture.crop.y * resolution,
                width * resolution * renderer.resolution,
                height * resolution * renderer.resolution,
                dx,
                dy,
                width * renderer.resolution,
                height * renderer.resolution
            );
        }
```
While the chop width and height need to * resolution * renderer.resolution. Is it the texture.crop.x and texture.crop.y need to * resolution * renderer.resolution too?
Also, is it dx and dy should * renderer.resolution too?
Here is the changed version that I found working in my local copy:
```javascript
if (this.tint !== 0xFFFFFF)
        {
            if (this.cachedTint !== this.tint)
            {
                this.cachedTint = this.tint;

                // TODO clean up caching - how to clean up the caches?
                this.tintedTexture = CanvasTinter.getTintedTexture(this, this.tint);
            }

            renderer.context.drawImage(
                this.tintedTexture,
                0,
                0,
                width * resolution * renderer.resolution,
                height * resolution * renderer.resolution,
                dx * renderer.resolution,
                dy * renderer.resolution,
                width * renderer.resolution,
                height * renderer.resolution
            );
        }
        else
        {
            renderer.context.drawImage(
                texture.baseTexture.source,
                texture.crop.x * resolution * renderer.resolution,
                texture.crop.y * resolution * renderer.resolution,
                width * resolution * renderer.resolution,
                height * resolution * renderer.resolution,
                dx * renderer.resolution,
                dy * renderer.resolution,
                width * renderer.resolution,
                height * renderer.resolution
            );
        }
```

[...]

PR #1793, opened 2015-05-13
Make Canvas Renderer functional again.

wojciak:
- fix sprite texture square (#1778)
- fix missing export in CanvasGraphics (#1790)

### DIFF
--- a/src/core/renderers/canvas/utils/CanvasGraphics.js
+++ b/src/core/renderers/canvas/utils/CanvasGraphics.js
@@ -7,6 +7,7 @@ var CONST = require('../../../const');
  * @memberof PIXI
  */
 var CanvasGraphics ={};
+module.exports = CanvasGraphics;
 
 /*
  * Renders a Graphics object to a canvas.
@@ -299,7 +300,7 @@ CanvasGraphics.renderGraphicsMask = function (graphics, context)
  *
  * @private
  * @param graphics {Graphics} the graphics that will have its tint updated
- * 
+ *
  */
 CanvasGraphics.updateGraphicsTint = function (graphics)
 {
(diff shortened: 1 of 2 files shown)


src/core/renderers/canvas/utils/CanvasGraphics.js
```diff
@@ -7,6 +7,7 @@ var CONST = require('../../../const');
  * @memberof PIXI
  */
 var CanvasGraphics ={};
+module.exports = CanvasGraphics;
```
englercj: Ha! Was this seriously the problem? Hilarious!

src/core/sprites/Sprite.js
```diff
@@ -490,12 +490,12 @@ Sprite.prototype._renderCanvas = function (renderer)
         {
             renderer.context.drawImage(
                 texture.baseTexture.source,
-                texture.crop.x * resolution,
-                texture.crop.y * resolution,
+                texture.crop.x * resolution * renderer.resolution,
```
englercj: I'm confused by this. If this is the proper calculation then shouldn't it just be `x * texture.baseTexture.resolution`?

Right now `resolution` is defined as:

```js
var resolution = texture.baseTexture.resolution / renderer.resolution;
```

And then you are multiplying by `renderer.resolution`. Sounds like what should happen is that `resolution` should be just `var resolution = texture.baseTexture.resolution;` and all these multiplications become `* resolution`.

englercj: I'm also confused why this would use `renderer.resolution` for dx/dy but the drawImage just above this doesn't?

englercj:
Closing, managed to get these changes into dev.

wojciak:
Not 100% sure why this works - I'm not that saavy with pixi internals yet :)

PR #1804, opened 2015-05-15
Fix typo in cacheAsBitmap.js, also respect .visible property

FletcherD:


FletcherD:
I accidentally added another change I intended to be a separate pull request... 
"cacheAsBitmap should respect .visible property"
As it was, a DisplayObject that had cacheAsBitmap set won't change when you change .visible until you unset cacheAsBitmap. Since this should be a simple matter of just not drawing the cached bitmap I reused the code from Container.js to just not draw the object when necessary.

englercj:
Thanks for the codex @FletcherD, we always appreciate contributions. Unfortunately we do not accept PRs to the `master` branch, that is our latest release.

Please submit PRs to the `dev` branch as laid out in our [Contributing Guide](https://github.com/GoodBoyDigital/pixi.js/blob/master/CONTRIBUTING.md).

PR #1806, opened 2015-05-15
cacheAsBitmap should respect .visible property

FletcherD:
This just causes cacheAsBitmap objects to not draw their cached selves when .visible is false, instead of having to update the cache for it to take effect.


englercj:
:+1:

PR #1807, opened 2015-05-15
Fix TilingSprite tilePosition + tileScale bug

sweetmandm:
Demonstration: http://www.davidsweetman.com/pixi/demo/

When the tilePosition was set to be greater than the texture
width while a tileScale was applied, taking the tilePosition
modulo the texture width would cause the resulting texture
coordinates to drift off of whole steps of the texture.

This fix takes the tileScale into account while applying the
tilePosition.


englercj:
:+1:

PR #1812, opened 2015-05-17
Docs configuration file updated to use jaguar.js

JiDW:
Modify dev dependencies to use jaguar.js instead of docstrap.
update readme to replace docstrap by jaguar.js
Update jsdoc conf to use the new template


englercj:
Noice! :+1:

Issue #1813, opened 2015-05-17
Sprite.destroy can sometimes error due to sprite.children being undefined


khalilravanna:
Specific steps to repro I have in my code:

1) Create sprite via `PIXI.Sprite.fromFrame`
2) Call `sprite.destroy()` (`Sprite.prototype.destroy`)
3) This calls `Container.prototype.destroy`
4) Which calls `Container.prototype.removeChildren`
5) Which references `this.children.length` even though `this.children` is `undefined`

Will look a little further to see if it's something specifically to do with `PIXI.Sprite.fromFrame` because I haven't seen this error up until today with my code which calls destroy all over the place.

Edit: It looks like it's nothing to do with `PIXI.Sprite.fromFrame` as that just calls `new Sprite()`. It looks like this could easily resolved by just not calling `Container.prototype.removeChildren` if `!this.children` or escape out of `removeChildren` for the same check. I'll go ahead and open a PR for this now.

englercj:
Hmm, a Container (which Sprite inherits from) should always have a `children` property, unless it had been previously destroyed...

Do you have a repro test case I can look at?

khalilravanna:
> unless it had been previously destroyed

Aha it would make sense that it's an error in my code somewhere :) but I couldn't find any glaring deletion of the children happening on my end in my initial search. I'll have to look into places where I might be calling `destroy` more than once.

That being said, it seems odd that calling `destroy` on a sprite that has already been destroyed would cause an error instead of being a noop.

englercj:
Well destroy clears the internal variables, that's all. It is up to user code to then lose your reference of the sprite. If you call destroy, then keep your reference around not only will GC never clean it up, but you may get errors like this.

khalilravanna:
Ok that's totally valid. Thanks for the quick response!

PR #1814, opened 2015-05-17
Only call Container.removeChildren if there are children

khalilravanna:
This fixes an error when calling `Sprite.destroy` as per #1813.

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -563,7 +563,11 @@ Container.prototype.destroy = function (destroyChildren)
         }
     }
 
-    this.removeChildren();
+    // Since this is called by Sprite.destroy and sprites don't have children double check to see if there are children first
+    if (this.children)
+    {
+        this.removeChildren();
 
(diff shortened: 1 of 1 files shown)


serprex:
But sprites do have children..

coreybadass:
at do you mean

coreybadass:
i maen waht

GoodBoyDigital:
@serprex is correct, sprites do have children also :+1:

englercj:
Hence my comment on the original bug:

https://github.com/GoodBoyDigital/pixi.js/issues/1813#issuecomment-102880669

:)

khalilravanna:
Derp yeah it looks like the cause is that I'm most likely calling `destroy` more than once on a sprite which looks like it would cause an error. I feel like calling `destroy` after the first time should be a noop instead of causing a hard error but that's an entirely separate issue.

Figured I'd open the PR immediately just in case it wasn't an error on my end ;)

Closing now.

PR #1821, opened 2015-05-18
Fix an error when an event occurs on deleted obj

JiDW:
fix https://github.com/GoodBoyDigital/pixi.js/issues/1817

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -346,7 +346,7 @@ InteractionManager.prototype.mapPositionToPoint = function ( point, x, y )
  */
 InteractionManager.prototype.processInteractive = function (point, displayObject, func, hitTest, interactive )
 {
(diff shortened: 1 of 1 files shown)


englercj:
I feel like this just hides the symptom, I want to track down the cause and fix it.

JiDW:
I think you're right. Thinking about it, I find it odd to see this error reported on the same day that it occured to me on a old code.

I did change things, however, like setting the shared ticker update at the start of my main loop.

For now, the temp "fix" is visibily enough but it is worth investigating.

drkibitz:
Objects in pixi that can be destroyed in the render loop could use a `.destroyed` property flag. Would help with situations like this, and potentially more we haven't found or encountered yet.

endel:
I'm curious to see a minimal example to reproduce this problem. Do you guys have one? It would be great to write some test cases to avoid this situation.

PR #1829, opened 2015-05-20
Add convenience event 'clicktap' that combines events for 'click' and 'tap'

marekventur:
Disclaimer: I'm not sure this is the best way of doing it.

In our code we have a few places where we do
```javascript
container.once("click", doSomething);
container.once("tap", doSomething);
```

This PR adds a new event called "clicktap" that combines the two:
```javascript
container.once("clicktap", doSomething);
```

I've run testem (some of the tests timed out), ~~but I couldn't figure out how to run jshint~~ (It's ```gulp jshint```). The CONTRIBUTING.md is a bit out-of-date, so I apologise if I missed a step.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -488,7 +488,12 @@ InteractionManager.prototype.processMouseUp = function ( displayObject, hit )
         if( displayObject[ isDown ] )
         {
             displayObject[ isDown ] = false;
-            this.dispatchEvent( displayObject, isRightButton ? 'rightclick' : 'click', this.eventData );
+            if (isRightButton) {
+                this.dispatchEvent( displayObject, 'rightclick', this.eventData );
+            } else {
+                this.dispatchEvent( displayObject, 'click', this.eventData );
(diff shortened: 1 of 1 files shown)


JiDW:
I wasn't even aware of this way to bind events.

We just do this : 

container.tap = container.click = function(){};

marekventur:
yes, but that only works well if you have only one listener, the on/once approach is a lot more flexible and allows multiple listeners.

marekventur:
But this PR would also allow you to do 
```
container.clicktap = function() {}
```

englercj:
I don't think this is a good idea, if we wanted to combine events we would implement the pointer event API. This is totally non-standard.

marekventur:
Fair enough. Anything I can do to help to move in that direction?

englercj:
There is a lot of work that needs to get done in the interaction manager IMO. I'm thinking of what the best way to do that would be.

In the meantime if you want to unify events like this, just use the standard pointer event names and I would be ok with it, at least temporarily.

PR #1833, opened 2015-05-20
Fix tilingsprite from a spritesheet

JiDW:
Fix https://github.com/GoodBoyDigital/pixi.js/issues/1787

When rendering a tilingsprite, we should use the frame and not the trimmed texture.

Thanks @greewi for this one.

I used tileScale to determine if the sprite is a tilingsprite. Tell me if you know a better way.

### DIFF
--- a/src/core/sprites/webgl/SpriteRenderer.js
+++ b/src/core/sprites/webgl/SpriteRenderer.js
@@ -221,9 +221,9 @@ SpriteRenderer.prototype.render = function (sprite)
 
     var w0, w1, h0, h1;
 
-    if (texture.trim)
+    if (texture.trim && sprite.tileScale === undefined)
     {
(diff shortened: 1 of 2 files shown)


englercj:
Wow, very interesting. @GoodBoyDigital is there a better check to see if it is a tilingsprite?

GoodBoyDigital:
Hmm, I think there is a nicer way to do this to avoid the extra check. Leave it with me chaps!

GoodBoyDigital:
Will have a think about this one.. wound be nice to find a way that does not involve modifying the spriteRenderer.  We should move the check to the tilingSprite class if we can..

JiDW:
Any updates on this please ? Could be nice to be able to update our project to the latest build

englercj:
Merging this for now, if @GoodBoyDigital has a better way we can refactor it.

PR #1834, opened 2015-05-21
Define package files, up devDependencies

drkibitz:
@englercj instead of commenting yet again here https://github.com/GoodBoyDigital/pixi.js/issues/1832

I just made a PR :wink: 

This defines just 2 dirs in files `bin/` and `src/`.
`package.json`, `README.md`, and `LICENSE` are included by default.

Run the following to ensure:
```sh
npm pack
```

Also updated dev dependencies. Again like last time, browserify was a major update, tested new bin with examples repo everything seems a ok, also testem still passes. Build is just slightly faster :rocket:

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,10 @@
     "test": "gulp && testem ci",
     "docs": "jsdoc -c ./gulp/util/jsdoc.conf.json -R README.md"
   },
+  "files": [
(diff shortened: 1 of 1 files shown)


englercj:
:+1: Always just what I needed, thanks @drkibitz !

PR #1836, opened 2015-05-21
Chore/cleanup

drkibitz:
This is a PR made of general cleanup. Major things were the removing the filter aliases and renaming uuid to uid.

---

I have another pending PR where I have done the best I can to make pixi.js an isomorphic library. There are only 2 issues with the PR.
1. I made the changes on top of these changes, so waiting to hear about this one.
2. The resource-loader is actually the biggest offender in terms of environment dependent code (at module execution time), so would still need some work to get it over the finish line.

--- 
Anyway, enough about a PR yet to be submitted, this one is hopefully straight forward.


drkibitz:
You may notice that a lot of this work enforces the dependency separate so it will make splitting up the code base easier in the future.

englercj:
I will definitely take some time to review this, thanks @drkibitz!

> The resource-loader is actually the biggest offender in terms of environment dependent code (at module execution time), so would still need some work to get it over the finish line.

This is totally true, originially I was going to use superagent to keep it agnostic and work in both places. But unfortunately they dropped XDR/IE9 support so I ended up having to do it myself :(

src/deprecation.js
```diff
@@ -232,3 +232,59 @@ core.Texture.prototype.setFrame = function(frame) {
...
+    /**
+     * @class
+     * @private
+     * @name PIXI.filters.FXAAFilter
+     * @see {@link PIXI.FXAAFilter}
+     * @deprecated since version 3.0.6
+     */
+    FXAAFilter: {
+        get: function() {
+            console.warn('filters.AbstractFilter is an undocumented alias, please use FXAAFilter from now on.');
```
englercj: Copy pasta error here in the text.

drkibitz: I like that :spaghetti: , and good :eyes:

drkibitz:
@englercj @GoodBoyDigital I was almost going to call this "done" from double checking any more paste errors etc.. but then I wanted to fix one more thing.

I'm not a fan of publicly made aliases, as you can assume, it makes the API surface unnecessarily larger to cover, and more difficult to update (deprecations, etc.).

The last piece I wanted to address was the items within and the namespace itself of `PIXI.math`. For me it's totally fine to maintain and math/index module that is internal to the core, but then mix it in to core itself to make items public. It is also ok with me to not mix it in, and expose the math/index module as a namespace on core itself (like core.utils). What bothers me is doing both.

So I am will to do this work in one way or another. I would like to do one of the following:
1. Deprecate PIXI.math namespace, and favor the member of math mixed into PIXI
2. Deprecate the members that are accessible on PIXI in favor of accessing them on PIXI.math

Just let me know what you think. Personally I'm in favor of #2 since it is more in line with the direction of the rest of the 3.0 API.

englercj:
@GoodBoyDigital and some others feel strongly about having math functions on the main object.

Personally I agree and think `PIXI.math` is the correct way to use it since that is what `PIXI.utils` and other modules are as well. Maybe people have changed their minds recently?

Pinging a few interested parties:
@GoodBoyDigital @photonstorm @alvinsight

alvinsight:
Well, at first, thank you @drkibitz for this ! Pixi needed a spring-clean :)

So regarding the PIXI.math and mixing modules in, I would always prefer having the things that people use outside of PIXI (math, blend modes, etc) in the global namespace.
It's very straightforward (you want a point? then PIXI.Point, etc...) and in line with the ease of use and the level of abstraction that pixi provides.
I don't mind having the texture cache in the Utils and having the Ticker in tickers.Tickers because they are classes that people don't normally touch when they are just creating a product with pixi.
If users want to play with the internals of the library then they are obviously willing to discover how these internals work.

drkibitz:
@englercj @alvinsight @GoodBoyDigital  Since this has been sitting a couple weeks, I think I'm going to go with deprecating the math namespace.

drkibitz:
Rebased, and added https://github.com/GoodBoyDigital/pixi.js/commit/f684a1424506af656e02d0a0320c179a5fce44be

drkibitz:
Like I said, I was not really married to one way or another, I just didn't want both. I don't believe in maintaining synonymous members in a single API. Yes, it bothers me on node's EventEmitter with `on` and `addListener`, and the worst offender I've come across is https://github.com/Reactive-Extensions/RxJS. This comes from experience maintaining internal APIs at various employers, and being a long term consumer of various open source libraries. Compatibility is fine and usually the only time I'm fine with it, just as long as it always comes with deprecation. Without that, it is just a compromise, and is just the result of not coming to a decision. Synonymous members always become troublesome to maintain long term, and ultimately lead to unnecessary confusion one way or another by consumers of the API. This all may seem like a small thing when it comes to a single method or namespace, but my hope is that it comes with a longer term contract for the future ;)

englercj:
This all looks good to me, if @GoodBoyDigital is OK with it we can merge.

GoodBoyDigital:
Looks good to me! Thanks for this one @drkibitz :+1:

alvinsight:
Nice one @drkibitz, thank you for your efforts :)

PR #1864, opened 2015-06-04
Improved Particle Shader performance by discarding transparent pixels

bchevalier:
On my machine the impact was a 50% fps boost on the bunnymark with 50,000 bunnies:
-Before: 30fps
-After: 45fps

The Improvement makes sense considering that roughly 30% of the pixels of the bunny texture are transparent and the bottleneck is blending the fragments together.

I do not know about other hardware but on my machine I still get 30fps while using completely opaque textures. Meaning that executing ```if (color.a == 0.0) discard;``` has no measurable performance impact and could be added to other fragment shaders.

### DIFF
--- a/src/core/particles/webgl/ParticleShader.js
+++ b/src/core/particles/webgl/ParticleShader.js
@@ -49,7 +49,9 @@ function ParticleShader(shaderManager)
             'uniform float uAlpha;',
 
             'void main(void){',
-            '   gl_FragColor = texture2D(uSampler, vTextureCoord) * vColor * uAlpha ;',
+            '  vec4 color = texture2D(uSampler, vTextureCoord) * vColor * uAlpha;',
+            '  if (color.a == 0.0) discard;',
+            '  gl_FragColor = color;',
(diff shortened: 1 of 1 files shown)


englercj:
Yeah this is a great call out, I use it a bit with the light shaders and it helps improve perf measurably:

https://github.com/pixijs/pixi-lights/blob/ce588d5f29ab349a3b96b08d47cd274c4cea2801/src/lights/_shared/loadNormals.glsl#L5
https://github.com/pixijs/pixi-lights/blob/master/src/lights/pointLight/point.frag#L25

Thanks for the PR!

GoodBoyDigital:
This is a really interesting topic! After chatting with some webGL peeps the general consensus is not to do it - or it makes very little difference. It can cause slowdown if you use things like the depthbuffer but again its hard to test! 

Happy to have it in particle container as it makes sense (and clearly faster), but will not be adding it to the main shader unless we have some real hard evidence that it improves things. 

Very interesting stuff though..

PR #1870, opened 2015-06-05
Trigger "added" and "removed" events when adding children on Container

endel:
Hello there,

This is similar to Flash's `Event.ADDED_TO_STAGE` behavior.
I've added a simple test for that. 

**Usage example:**

```javascript
class Something extends PIXI.Container {

  constructor () 
  {
    super();

    this.on('added', this.onAdded, this);
    this.on('removed', this.onRemoved, this);
  }

  onAdded (to) 
  {
    console.log("I'm now attached to some Container!", to)
  }

  onRemoved (from) 
  {
    console.log("I was removed from the Container. :(", from)
  }

}
```

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -131,7 +131,7 @@ Container.prototype.addChildAt = function (child, index)
 
         this.children.splice(index, 0, child);
 
-        child.emit('added');
+        child.emit('added', this);
 
         return child;
     }
@@ -250,7 +250,7 @@ Container.prototype.removeChildAt = function (index)
     child.parent = null;
     this.children.splice(index, 1);
 
-    child.emit('removed');
+    child.emit('removed', this);
 
     return child;
 };
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -130,6 +130,9 @@ Container.prototype.addChildAt = function (child, index)
         child.parent = this;
 
         this.children.splice(index, 0, child);
+
+        child.emit('added');
+
         return child;
     }
     else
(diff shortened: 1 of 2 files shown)


englercj:
I like the idea of an event happening when a child is added!

I don't like the change to the parent property. This property is used a few times in rendering paths, so making it a getter will likely cause a perf hit. Does it not make sense to emit this event in the addChild methods?

endel:
@englercj Thanks for your feedback!

I've made the changes you suggested and squashed the commits into a single one.

Now it makes more sense, since when the event is triggered it's possible to access container's children in a consistent state. I've also added a test for that.

englercj:
Thanks! I also suggest sending the container that it was added/remove to/from as the param of the event, so people can access that if they need it.

endel:
Nice! I've added it as the param. Also updated the test. Sounds good for you? :)

englercj:
Looks great to me! If @GoodBoyDigital is OK with it, lets merge it in.

GoodBoyDigital:
Will give this a review over the weekend chaps. Good stuff

On Saturday, 6 June 2015, Chad Engler <notifications@github.com> wrote:

> Looks great to me! If @GoodBoyDigital <https://github.com/GoodBoyDigital>
> is OK with it, lets merge it in.
>
> —
> <https://github.com/GoodBoyDigital/pixi.js/pull/1870#issuecomment-109534490>
> .
>


-- 
Mat Groves

*Technical Partner*


GoodBoyDigital:
looking good. This could turn out to be quite handy! thanks @endel :+1:

PR #1871, opened 2015-06-05
Update CONTRIBUTING guide.

endel:
- Removed `grunt` references
- Avoid `gulp` on the guide, since `npm run build` now can do the same.

If there is a need to replace gulp with something else in the future, there is no need to change it in the contributing guide. :)


englercj:
:+1:

PR #1874, opened 2015-06-06
Fix em-dash in README

upisfree:


GoodBoyDigital:
thanks @upisfree :+1:

PR #1877, opened 2015-06-08
Mesh interaction

ivanpopelyshev:
Mesh Interaction and optimizing spaces (I'm sorry)

### DIFF
--- a/src/mesh/Mesh.js
+++ b/src/mesh/Mesh.js
@@ -1,4 +1,6 @@
-var core = require('../core');
+var core = require('../core'),
+    tempPoint = new core.Point(),
+    tempPolygon = new core.Polygon();
 
 /**
  * Base mesh class
@@ -29,9 +31,9 @@ function Mesh(texture, vertices, uvs, indices, drawMode)
      * @member {Float32Array}
      */
     this.uvs = uvs || new Float32Array([0, 1,
-                                 1, 1,
-                                 1, 0,
-                                 0, 1]);
+        1, 1,
+        1, 0,
+        0, 1]);
 
     /**
      * An array of vertices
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
This commit will fix pixi-spine issue https://github.com/pixijs/pixi-spine/issues/38

englercj:
:+1: Looks good to me, @GoodBoyDigital thoughts?

src/mesh/Mesh.js
```diff
@@ -366,55 +368,103 @@ Mesh.prototype._onTextureUpdate = function ()
...
+ * Tests if a point is inside this mesh. Works only for TRIANGLE_MESH
+ *
+ * @param point {Point} the point to test
+ * @return {boolean} the result of the test
+ */
+Mesh.prototype.containsPoint = function( point ) {
+    if (!this.getBounds().contains(point.x, point.y)) {
+        return false;
+    }
+    this.worldTransform.applyInverse(point,  tempPoint);
```
endel: Two spaces on function arguments

endel:
Wouldn't it be nice to have some tests on this? 

It may have some assertions that `containsPoint` does what it is supposed to do. It seems implemented perfectly, but having tests it will ensure that future changes doesn't break existing functionality.

englercj:
I 100% agree, I don't think it should stop the merge of the feature, but we should definitely add some if at all possible.

ivanpopelyshev:
Ok, I'll make some tests :) Btw, I only did test with vertex indices, other branch of "if" wasnt tested.

GoodBoyDigital:
looks good to me! happy to roll in!

PR #1886, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Per request for `dev` branch

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
     "name": "pixi.js",
-    "version": "3.0.6",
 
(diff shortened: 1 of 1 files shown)


englercj:
Thanks so much for this!

kkirsche:
Yep!

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

Issue #1884, opened 2015-06-10
mousemove fires on all children


PierBover:
So I have a container with a bunch of children. I set the `mousemove` event on each children, and when the mouse moves over one of the children, the event is fired on all children.

Is this a bug?

englercj:
Sounds like it!

endel:
Maybe this is related with https://github.com/GoodBoyDigital/pixi.js/issues/1837

PierBover:
Yes it sounds similar to #1837 

I'm using 3.0.6 with WebGL renderer.

wojciak:
+1 for this ;]

endel:
I'll try to figure out what is happening to fix this, ok?

endel:
It was simpler than I thought. Here is a working example with the patch: https://jsfiddle.net/endel/upz3t2a0/1/

pavan4:
This works great but when you have overlapping objects, event is fired only on the object on top. how can it made to fire on both the objects?

I opened an issue here at https://github.com/GoodBoyDigital/pixi.js/issues/1891 for this and could not solve it yet.

stephenweaver:
added a comment to the PR https://github.com/pixijs/pixi.js/pull/1894
this issues is not resolved with that fix. I think if we can figure out why targetTouches includes all touches on the canvas rather than just the pixi object, the events will get sent to the correct objects. maybe?

6174:
@endel   when   use  mousemove event for drag , your patch cannot work, because mouse move is faster than rendering,  when mouse miss hit the drag element , drag can't work. 

here is the enhance code 

```javascript
  PIXI.interaction.InteractionManager.prototype.processMouseMove = function ( displayObject, hit )
  {
      if(hit || displayObject.isDragging )
      {
          this.dispatchEvent( displayObject, 'mousemove', this.eventData);
      }
      this.processMouseOverOut(displayObject, hit);
  }
```

6174:
@endel   when   use  mousemove event for drag , your patch cannot work, because mouse move is faster than rendering,  when mouse miss hit the drag element , drag can't work. 

here is the enhanced code 

```javascript
  PIXI.interaction.InteractionManager.prototype.processMouseMove = function ( displayObject, hit )
  {
      if(hit || displayObject.isDragging )
      {
          this.dispatchEvent( displayObject, 'mousemove', this.eventData);
      }
      this.processMouseOverOut(displayObject, hit);
  }
```

[...]

PR #1894, opened 2015-06-13
dispatch mousemove/touchmove event only inside hit area

### LINKED ISSUE #1837: touchmove always being triggered [V3]
It seems all touchmove events are being triggered regardless of the touch position. In the below example the touchmove event is always being triggered for both boxes whenever and wherever a movement is made. 

https://jsfiddle.net/shinybob/hgk6469x/

endel:
Fixes #1837 and #1884

Live example with the patch: https://jsfiddle.net/endel/upz3t2a0/1/

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -540,7 +540,10 @@ InteractionManager.prototype.onMouseMove = function (event)
  */
 InteractionManager.prototype.processMouseMove = function ( displayObject, hit )
 {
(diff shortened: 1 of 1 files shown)


endel:
Hey guys,

The actual problem here was that the `mousemove` event were being triggered even without the mouse inside the element. You can watch the logs in the console without the patch here: https://jsfiddle.net/endel/g14e2mrw/1/

Issue #1763, opened 2015-05-08
Click in HTML can't be stopped from being triggered in PIXI


georgiee:
See here:
http://jsfiddle.net/georgie/gnka1jne/

1. Click on the red rect, this will trigger an expected event 'html triggered' 
2. But the yellow rect located in pixi triggers itself an event and there is no way to stop the event from outside of pixi.

Pixi registers an event listener on window at the [capture phase](https://github.com/GoodBoyDigital/pixi.js/blob/df0ee19d3e76c0935cfe06d5447befb2d20826bb/src/interaction/InteractionManager.js#L217), that's the earliest possible position where a mouse event can be listened to, isn't it ? So there can't be a solution in HTML to stop an event from being triggered in pixi ?

This might be the same?
https://github.com/GoodBoyDigital/pixi.js/issues/1725

Thanks!

georgiee:
Here are two other problems with interaction. This seems to be a mix of a regression with 3.0.3 (following @LuQano's description that it was working before) and a normal bug in V3.

Here is my second testcase:
http://jsfiddle.net/georgie/zqzbnmf4/

Two problems: 
1. Clicking is not blocked, so that an underlying element gets triggered, even if an object on top already consumed the click. It's shouldn't be necessary to stop the propagation manually as this is a behaviour someone would expect I guess?
2. Click and drag on top of another target. You will trigger a mouseup on that target which wasn't under the mouse during initiation.

Reproducing:
1. Click on the rect in the middle (which is a small zone where both larger rects are overlapping) You get two outputs. One from each rectangle.
2. Click on the yellow rect, hold and go down to the blueish one. Release. You will get the message pixi 2 triggered. Which was dispatched from the blueish rect. This is not a correct mouseup behaviour.

Related:
https://github.com/GoodBoyDigital/pixi.js/issues/1765

krzysztof-o:
+1

englercj:
@GoodBoyDigital Can you comment on why capture phase is being used?

jwdunn1:
I'm not convinced this is a bug. The "new and improved" default behavior makes sense if you want to grab a stack of interactive elements and move them around together. To make my app work in the same manner as it did in v3.0.2, I simply added event.stopPropagation(); to the mousedown and touchstart event handlers. So, now the top object in the stack is the only one that handles the event. You can do that in your test case also.

georgiee:
[...]

PR #1902, opened 2015-06-17
allow events to be canceled before coming into PIXI

endel:
This fixes the issue reported here: #1763 

Basically I'm checking the `originalEvent.defaultPrevented` to flag events as `stopped` and binding `mousemove` events on `interactionDOMElement ` instead of `window` (which is the first level of event propagation and can't be prevented/stopped)

Live example of this patch: http://jsfiddle.net/endel/xnmy239h/1/

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -206,15 +206,14 @@ InteractionManager.prototype.addEvents = function ()
         this.interactionDOMElement.style['-ms-touch-action'] = 'none';
     }
 
-    window.document.addEventListener('mousemove',    this.onMouseMove, true);
+    this.interactionDOMElement.addEventListener('mousemove',    this.onMouseMove, true);
     this.interactionDOMElement.addEventListener('mousedown',    this.onMouseDown, true);
     this.interactionDOMElement.addEventListener('mouseout',     this.onMouseOut, true);
+    this.interactionDOMElement.addEventListener('mouseup',      this.onMouseUp, true);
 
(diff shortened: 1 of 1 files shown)


src/interaction/InteractionManager.js
```diff
@@ -414,7 +412,7 @@ InteractionManager.prototype.onMouseDown = function (event)
 {
     this.mouse.originalEvent = event;
     this.eventData.data = this.mouse;
-    this.eventData.stopped = false;
+    this.eventData.stopped = event.defaultPrevented;
```
englercj: What is this supposed to do? `preventDefault()` is not the same as `stopPropagation()` (which sets `.stopped`).

endel: As I've searched there is no way to check if an `Event` was stopped via `stopPropagation()`, method, and the only property that is exposed on the API is `defaultPrevented` property.

So this allow users to cancel the event via `preventDefault` method, before it enters in PIXI's InteractionManager, as the [jsfiddle](http://jsfiddle.net/endel/xnmy239h/1/) demonstrates.

What are your thoughts on this?

englercj: I think that is super hacky, instead of trying to handle a stopped event we should just not use the capturing phase.

I feel like doing that and solving problems we have there will be the best solution. I'm looking to try and rewrite the Interaction Manager in the coming weeks. Stay tuned.

englercj:
Closing this based on my comment feedback.

endel:
Got it, thanks for your feedback!

PR #1907, opened 2015-06-20
Added a clamping to prevent gaps

JiDW:
...in tillingsprites when using a frame from a spritesheet and a scaled
container. fix https://github.com/GoodBoyDigital/pixi.js/issues/1787


GoodBoyDigital:
Nice fix man! this ones been floating around for a while! :+1:

georgiee:
I just discovered this commit. This fixes all of my local examples, in both scale modes nearest & linear. 
Thank you, great addition!

PR #1915, opened 2015-06-24
[Fix/Hack] Displacement filter blurry on iphone4, iphone4s and ipod5

bchevalier:
Added hack to displacement filter shader to avoid pixelized effect on some devices.
Tested devices that have the issue: iphone4, iphone4s, ipod5, ipad2
Tested devices that do not have the issue: Samsung Galaxy s3/s4, iphone 5s, iphone 6, ipad air

I know this does not look clean in the code but it has the effect of solving a very ugly bug (probably hardware related):

Before the hack:
bunnies_pixel

After the hack:
bunnies_smooth

PS: I stole your bunnies for this small demo but be assured that no harm was done to them (beside being exploded into 10 smaller copies of themselves)

### DIFF
--- a/src/filters/displacement/displacement.frag
+++ b/src/filters/displacement/displacement.frag
@@ -17,5 +17,5 @@ void main(void)
    map -= 0.5;
    map.xy *= scale;
 
-   gl_FragColor = texture2D(uSampler, vec2(vTextureCoord.x + map.x, vTextureCoord.y + map.y));
+   gl_FragColor = texture2D(uSampler, vec2(vTextureCoord.x + 0.999 * map.x, vTextureCoord.y + 0.999 * map.y));
 }
--- a/src/filters/displacement/displacement.frag
+++ b/src/filters/displacement/displacement.frag
@@ -1,4 +1,4 @@
-precision lowp float;
+precision mediump float;
 
 varying vec2 vMapCoord;
 varying vec2 vTextureCoord;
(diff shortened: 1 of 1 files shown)


englercj:
Interesting, does this change anything about how it looks on other devices and desktop?

bchevalier:
Visually, this fix does not seem to change anything about the behaviour of other devices as they still render correctly (desktop, laptop, androids...).

englercj:
What does it look like if you change the shader to use `highp` instead of doing this multiplication trick? I wonder if it is just a precision issue.

bchevalier:
Thanks, it worked. It also works with precision set to mediump.
I did not pay attention to that before.

englercj:
I had a feeling, I'd rather just bump up the precision and let the GPU deal with doing it the right way. If `mediump` works fine would you mind make that change instead and I will merge it in?

bchevalier:
Here you go.
By the way, is there really a performance difference between using lowp and mediump? If there is I could never measure it.
Another question, how is it possible that multiplying by ```0.999``` had an effect on the precision?

englercj:
> By the way, is there really a performance difference between using lowp and mediump? If there is I could never measure it.

A little bit of a performance difference (depending on hardware), but definitely a compatibility difference (depending on hardware). Some older mobile devices don't have hardware that can do `highp`, and even others have hardware that can't do `mediump`. It is always a balancing act :P

> Another question, how is it possible that multiplying by 0.999 had an effect on the precision?

It didn't, but for your case it corrected a rounding error that happened due to precision loss. Or more accurately *looked* like it was correcting a rounding error. It probably was still wrong, but less so enough that the result looked correct, which is all you can ask for when dealing with normalized values.

Issue #1926, opened 2015-06-27
Sprite batching causes GC pauses


davedx:
Hey guys,

We saw some behaviour yesterday when working with sprite batching that led us to believe the flush() was causing GC cycles due to things being allocated/thrown away again. After some hunting in the source I found the following line:

https://github.com/GoodBoyDigital/pixi.js/blob/07e4e26f8fd404524a7ebaa73c210de3d8f2cf25/src/core/sprites/webgl/SpriteRenderer.js#L342

Here you see that we do a subarray on the vertex float data, send the result to the GPU, then throw it away again.

I removed this line to make it send all of the data every time. It seems like rendering still works fine, I think it just means we're sending a bit more data to the GPU, but we prevent GC pauses that can interfere with smooth animation. (Especially on our slow set top boxes where a GC every 15 seconds causes a 150-200ms freeze D:)

I think it might be possible to make the sprite batch just dynamically resize the float array *when needed* instead of every frame (I had "downsizing" working but not "upsizing") -- maybe this is the optimum? Game devs learn pretty early on that you should try to avoid constantly making/removing game objects from a scene, so the resizing should not happen often. For us, the temp fix of just removing the subarray is working nicely.

Any thoughts on this?

englercj:
Do you have an example app the reproduces this behavior?

davedx:
The app we're working on is closed source. I'll try and put together something that reproduces the behaviour.

davedx:
Here's a demo repo: https://github.com/davedx/pixi_gc_demo

The code I patched is: https://github.com/davedx/pixi_gc_demo/blob/master/pixi.js#L17496

There's a screenshot of a sawtooth memory timeline in that repo too: https://raw.githubusercontent.com/davedx/pixi_gc_demo/master/pixi3.png

The weird thing is, if I remove the subarray and run it on our device (a set top box), we see almost no GC cycles if the app is idle (rendering every frame, no animations). But in Chrome on Win 8, I actually still see the memory sawtooth whether I have the subarray or not.

Any insights into this?

englercj:
Looking to implement a method like this:

https://github.com/GoodBoyDigital/pixi.js/pull/1939#issuecomment-118461829

I think this will work well to help solve this issue as well.

bchevalier:
I was wondering if that behaviour was consistant across all the browsers or if this was only happening on chrome?

bchevalier:
[...]

PR #1939, opened 2015-07-03
Improved performance of particle container for small amounts

bchevalier:
The particle container always upload a full batch size (15000 particles). When the number of particles is smaller it becomes a huge performance drawback in terms of CPU.
In my use case: maximum of roughly 500 sprites, bufferSubData accounts for 25% of my whole program cpu usage. After this optimisation it accounts only for 5%.
This is far from negligible when developing for smartphones where the maximum number of particles will max at 1000.


JiDW:
Ohhh so that's why is was slower than Container. Good job !

englercj:
This looks awesome, but according to #1926 it can cause GC stalls. We need to figure out of subarray is the cause of this before merging.

GoodBoyDigital:
We could create an array that 'grows' based on how much is required (it its too small we double the size)

bchevalier:
The problem is that, if the buffer does not get downsized, someone who empties his particle container might wonder why he still gets poor performance. Would it be acceptable in terms of performance to make the vertex buffer shrink as well whenever the number of particles gets too small?

davedx:
IMHO the sensible way to use a particle system is to create a fixed amount of particles that you then switch on/off as you need them. In other words, you allocate how ever many you need for your buffers, and take the hit to resize them if on the off chance you need to add/remove some permanently in that frame.

So what I reckon is the buffer should always be exactly the size needed to render the current amount of sprites/particles.

bchevalier:
Well I would like this container to be as flexible as possible. As a user I do not want to think about the number of particles I might need nor worry about the performance drawback if I do not use the full buffer.
In my use case I have a particle container to display from 50 to 500 particles (could be more, there is no actual upper bound). But in average it is around 100. If I have to decide beforehand the container size, I would put something like 700 to be safe, but then I get performance 7 times slower than what I need in average.

englercj:
What if we let the user specify an initial size (optionally), we grow the array dynamically (doubling when full), and we shrink dynamically (halfing when < 1/2 used).

We have a sensible default for the common case, but the initial size specification allows you to tweak it to be just right for your application.

bchevalier:
That seems good to me.

davedx:
I like the initial size specification.

For our use case, as long as the subarray isn't done every frame no matter what the conditions are, we're happy ;)

bchevalier:
I created a new pull request that solves the problem differently: https://github.com/GoodBoyDigital/pixi.js/pull/1990/files

If the new solution is fine I suggest we close this PR.

englercj:
Closing this in favor of the other PRs we have improving ParticleContainer.

PR #1942, opened 2015-07-03
ensure that sprite width and height are never negative

tiriana:
I found out that ```sprite.width``` is negative when ```sprite.x``` is negative (same for ```height``` and ```scale.y```). Here's a fix for that, and some tests :)


englercj:
This looks like a good change, but please read the Contributing Guide for making PRs:

https://github.com/GoodBoyDigital/pixi.js/blob/master/CONTRIBUTING.md#making-changes

PRs should hit the `dev` branch, and please do not include built files.

tiriana:
thx - I will open another PR for that.

PR #1944, opened 2015-07-03
Ensure that sprite width and height are never negative

tiriana:
I found out that sprite.width is negative when sprite.x is negative (same for height and scale.y). Here's a fix for that, and some tests.

This is the same change as in my previous PR - https://github.com/GoodBoyDigital/pixi.js/pull/1942, but my previous was to wrong branch, contained built files etc.

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -224,6 +224,16 @@ var utils = module.exports = {
     },
 
     /**
+     * Retunrs sign of number
+     *
+     * @param n {number}
+     * @returns {number} 0 if n is 0, -1 if n is negative, 1 if n i positive
+     */
+    sgn: function (n){
+        return n === 0 ? 0 : n < 0 ? -1 : 1;
+    },
+
+    /**
      * @todo Describe property usage
      * @private
      */
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -229,8 +229,8 @@ var utils = module.exports = {
      * @param n {number}
      * @returns {number} 0 if n is 0, -1 if n is negative, 1 if n i positive
      */
-    sgn: function (n){
-        return n === 0 ? 0 : n < 0 ? -1 : 1;
+    sign: function (n) {
+        return n ? (n < 0 ? -1 : 1) : 0;
     },
 
     /**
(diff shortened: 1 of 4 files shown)


src/core/utils/index.js
```diff
@@ -224,6 +224,16 @@ var utils = module.exports = {
     },
 
     /**
+     * Retunrs sign of number
+     *
+     * @param n {number}
+     * @returns {number} 0 if n is 0, -1 if n is negative, 1 if n i positive
+     */
+    sgn: function (n){
```
adelciotto: Perhaps just call the function *sign*, or *getSign*. Don't see the need to abbreviate it by just one character.

src/core/utils/index.js
```diff
@@ -224,6 +224,16 @@ var utils = module.exports = {
...
+        return n === 0 ? 0 : n < 0 ? -1 : 1;
```
adelciotto: perhaps:

```javascript
function sign(n) {
    return n ? (n < 0 ? -1 : 1) : 0;
}
```
But meh, up to you @tiriana. I'm being very nit picky I guess :P

src/core/utils/index.js
englercj: :+1:

src/core/utils/index.js
tiriana: @adelciotto Hi. I will use implementation you send, because it's faster then mine. And about the name - I can change it, but [```sgn```](https://en.wikipedia.org/wiki/Sign_function) is widely used in mathematics, that's why I used it. And it's abbreviation of 'signum' not 'sign', btw :) . 
But on the other hand - there is [```Math.sign```](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) in ```ES6```, so developers might be more familiar with ```sign```.  I'm not sure - I guess I'll wait a bit with this decision, maybe more people will express their opinion about the name.

@englercj  Was that :+1: for name change or function itself ? :)

englercj: The name change

tiriana: @adelciotto @englercj I changed the name.

englercj:
Thanks so much for this @tiriana, and a huge thank you for writing tests for it!

PR #1969, opened 2015-07-09
make sure fromUrl uses the the iterator to find the right src

pr1ntr:
reopened PR for correct branch. Couldn't find anywhere to change the old one.


pr1ntr:
ref: https://github.com/GoodBoyDigital/pixi.js/issues/1959

englercj:
Nice ty!

PR #1987, opened 2015-07-22
[Bug Fix + Enhancement] Particle container refactoring

bchevalier:
* Fixed bugs related to children being added or removed and the static data of the particle container not being re-uploaded.
* Improved consistency/readability by leaving the state of the properties (either static or dynamic) on the container and not attaching them on the particle renderer, which is a singleton (this was leading to confusion when reading the code and I think I fixed a bug in the process but did not try to produce it to see if there actually was a bug or if I fixed it)
* Improved performance of the particle container as the user can now specify a batch size for the particles (Especially useful when developing for slow hardware such as mobile devices).


bchevalier:
Actually let me make 3 different PRs for this.

PR #1988, opened 2015-07-22
[FIX] Particle Container children addition and removal

bchevalier:
* Fixed bugs related to children being added or removed and the static data of the particle container not being re-uploaded.
* Also correctly emit the "removed" event when child are batched removed


src/core/display/Container.js
```diff
@@ -95,6 +95,13 @@ Object.defineProperties(Container.prototype, {
 });
 
 /**
+ * Overridable method that can be used by Container subclasses whenever the children array is modified
+ *
+ * @private
+ */
+Container.prototype.handleChildrenChange = function () {};
```
englercj: Rename this method to `onChildrenChange` and this looks good to me!

englercj:
Rebased and merged manually.

PR #1989, opened 2015-07-22
[Enhancement + Fix] Improved consistency/readability of ParticleRenderer

bchevalier:
Improved consistency/readability by leaving the state of the properties (either static or dynamic) on the container and not attaching them on the particle renderer, which is shared by many containers. The reason is that it was leading to confusion when reading the code and it was the cause of a possible bug that I possibly fixed, I think (I did not try to produce it to see if there actually was a bug).


src/core/particles/webgl/ParticleRenderer.js
```diff
@@ -210,7 +205,7 @@ ParticleRenderer.prototype.render = function ( container )
             return;
         }
 
-        if(!this.properties[0].dynamic || !this.properties[3].dynamic)
+        if(!container._properties[0] || !container._properties[3])
```
bchevalier: This is where the bug would have occurred as the status of the properties of the renderer was checked rather than the status of the properties of the container.
The renderer being shared by several containers the status might not be corresponding to the current container and lead to the static data not being uploaded.

englercj:
Great catch, I'll rebase and merge this in.

englercj:
Manually rebased and merged.

bchevalier:
Thanks for the reactivity,
By the way why the rebasing?

englercj:
There are a bunch of merge commits and changes in these PRs that I don't want in the history. Much cleaner to have a single commit that represents the change. That way the history only includes the necessary changes and not all these merges that are unecessary.

bchevalier:
ah I see, thanks

PR #1990, opened 2015-07-22
[Enhancement] Added possibility to specify a batch size to the Particle Container

bchevalier:
It Improves performance as the user can now specify a batch size for the particles. It is especially useful when developing for slow hardware such as mobile devices.

Note: I plan on doing another optimisation on top of it that should (crossing fingers) allow developers to greatly improve performance when it comes to adding and removing particles from the container (it should show good results on the bunnymark).


src/core/particles/ParticleContainer.js
```diff
@@ -25,7 +25,7 @@ var Container = require('../display/Container'),
  * @extends PIXI.Container
  * @memberof PIXI
  *
- * @param [size=15000] {number} The number of images in the SpriteBatch before it flushes.
+ * @param [totalSize=15000] {number} The maximum number of particles that can be renderer by the container.
```
englercj: `size` or `maxSize` is better naming for this I think.

src/core/particles/ParticleContainer.js
```diff
@@ -33,10 +33,25 @@ var Container = require('../display/Container'),
  * @param [properties.uvs=false] {boolean} When true, uvs be uploaded and applied.
  * @param [properties.alpha=false] {boolean} When true, alpha be uploaded and applied.
  */
```
englercj: Need to add a line for `batchSize`.

englercj:
Good start, this adds the code necessary for initializing the buffers to a different size if you want. It doesn't add the dynamic resizing that we talked about yet, but that is OK we can add it later.

src/core/particles/ParticleContainer.js
```diff
@@ -33,10 +33,25 @@ var Container = require('../display/Container'),
...
-function ParticleContainer(size, properties)
+function ParticleContainer(totalSize, properties, batchSize)
 {
     Container.call(this);
 
+    batchSize = batchSize || 15000; //CONST.SPRITE_BATCH_SIZE; // 2000 is a nice balance between mobile / desktop
```
englercj: Why did the default change to 15k? The problem this is trying to solve is not updating a large buffer that is mostly unused, but this actually will make that worse for most people by default.

src/core/particles/ParticleContainer.js
bchevalier: yes, good catch

src/core/particles/ParticleContainer.js
bchevalier: I will change to maxSize

src/core/particles/ParticleContainer.js
bchevalier: It was already 15k but it used to be in the ParticleRenderer component (see ```this.size = 15000```).

bchevalier:
I realised that there would be quite some changes to do in order to make the buffer size grow. This solution was more elegant and it fits my needs so I guess it fits the needs of people with similar performance requirements. I think this can substitute to a dynamically resizing buffer.

englercj:
Yeah like I said, I think this is a great first step in that direction. This alone is very helpful.

src/core/particles/ParticleContainer.js
englercj: Oh my bad, you are right!

englercj:
Looks good to me, would you mind squashing these commits and I'll merge it in?

```
git rebase -i HEAD~6
git push -f origin particleContainerRefactor3
```

Should do the trick!

bchevalier:
After the rebase it tells me that everything is already up to date so I cannot push anything

englercj:
It should say that your branches have diverged and make you need to force push.

Thats fine, I can rebase it and merge it manually.

englercj:
Manually rebased and merged.

src/core/particles/webgl/ParticleRenderer.js
```diff
@@ -270,9 +269,9 @@ ParticleRenderer.prototype.generateBuffers = function ( container )
         this.properties[i].dynamic = container._properties[i];
     }
 
-    for (i = 0; i < size; i += this.size)
+    for (i = 0; i < size; i += batchSize)
     {
-        buffers.push( new ParticleBuffer(gl,  this.properties, this.size, this.shader) );
+        buffers.push( new ParticleBuffer(gl,  this.properties, batchSize, this.shader) );
```
bchevalier: I believe a small mistake was made when resolving the conflict here, if you look at the current dev version ```this.size``` hasn't been replaced by ```batchSize```

englercj:
Good catch, thanks for that! Let me know if you see anything else.

PR #1992, opened 2015-07-23
[Enhancement] Improved performance of adding and removing particles

bchevalier:
Whenever a particle was being added or removed, the static data of every particle buffer was being uploaded. Now, only the static data of the children whose index have changed are being uploaded.

This shows good performance boost on the bunnymark when chosing moderate batch size (= 3000). The demo does not lag anymore whenever bunnies are being added.

However, the bunnymark is a perfect showcase for this optimisation as elements are only being added. Therefore only one particle buffer needs to be updated whenever a particle is being added. Actually, in average, half the particle buffer would need to be updated whenever a particle is being either added or removed. And in the worst case all the buffers have to be uploaded, as it was already the case.

### DIFF
--- a/src/core/particles/ParticleContainer.js
+++ b/src/core/particles/ParticleContainer.js
@@ -83,7 +83,7 @@ function ParticleContainer(maxSize, properties, batchSize)
      * @member {boolean}
      * @private
      */
-    this._updateStatic = false;
+    this._bufferToUpdate = 0;
 
     /**
      * @member {boolean}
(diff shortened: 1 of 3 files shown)


src/core/particles/ParticleContainer.js
```diff
@@ -83,7 +83,7 @@ function ParticleContainer(maxSize, properties, batchSize)
      * @member {boolean}
```
englercj: think these are wrong now.

englercj:
Looks nice dude, thanks for spending so much time making ParticleContainer the best.

bchevalier:
My pleasure,
by the way, I fixed the js doc to ```number```.

Issue #1777, opened 2015-05-11
setBackgroundColor?


kittykatattack:
Hello,

Is there a replacement for  `stage.setBackgroundColor` in v3?

miriti:
Hey, @kittykatattack !
Try this:

```javascript
renderer = new PIXI.WebGLRenderer(800, 600);
renderer.backgroundColor = 0x061639;
```

kittykatattack:
Thank you!
Maybe I'm losing my mind (very possible!!) - but that doesn't work for me.
Your code produces a blank canvas on my system (I'm on v3.0.3)

Also, `backgroundColor` isn't a property of the renderer object, according the current API docs?
Here's the code I'm currently using, which just produces a standard black WeGL context.
```
var renderer = PIXI.autoDetectRenderer(256, 256, {antialiasing: false, transparent: false, resolution: 1});
document.body.appendChild(renderer.view);
renderer.backgroundColor = 0x061639;
```

miriti:
To see the effect you need to call `render` method of the renderer like this (I changed the color to red to make it more obvious):

```javascript
var renderer = PIXI.autoDetectRenderer(256, 256, {antialiasing: false, transparent: false, resolution: 1});
    document.body.appendChild(renderer.view);
    renderer.backgroundColor = 0xff0000;

    var scene = new PIXI.Container();

    var render = function() {
        renderer.render(scene);
        requestAnimationFrame(render);
    }

    render();
```

Here is the working code with PIXI 3.0.3: http://miriti.github.io/pixi-background/

kittykatattack:
Thanks for your help, that works!

unwitting:
How come this can't be seen in the documentation @miriti ? I searched for this for ages before turning to the issues!

miriti:
@unwitting good question! I don't know.. I just looked it up in the code. For some reason it hasn't got a doc: https://github.com/GoodBoyDigital/pixi.js/blob/master/src/core/renderers/SystemRenderer.js#L148

unwitting:
@miriti PRd :)

englercj:
Fixed in 4827cea4bfd4af99bb363c2230da1393ed8221c4

jrheard:
[...]

PR #1995, opened 2015-07-23
Add documentation: SystemRenderer.backgroundColor

unwitting:
As in #1777, noticed this isn't documented in the public docs since only the private field has been commented.


miriti:
Lol. I just did the same pull-request. But in a different way: https://github.com/GoodBoyDigital/pixi.js/pull/1996

miriti:
Turned out that the documentation is already there. It's not added to the documentation due to lack of the PIXI namespace in the jsdoc

englercj:
Ha! Too slow!

4827cea4bfd4af99bb363c2230da1393ed8221c4

PR #2001, opened 2015-07-26
Added 'durationMultipliers' to MovieClip

zOadT:
Hello everyone,

I implemented 'durationMultipliers' in MovieClip.prototype. Initially it's an array with the same length as MovieClip._textures and all values set to 1. By changing a value you can cause individual frames to be shown longer or shorter (e.g. movie.durationMultipliers = [1, 2, 1]; will show the second frame twice as long as normally)

As long as you don't set or get the durationMultipliers property, the update function is using the old algorithm to determine the current frame, so there should be no no performance impact.

Any feedback is welcome :)

regards

### DIFF
--- a/src/extras/MovieClip.js
+++ b/src/extras/MovieClip.js
@@ -39,6 +39,18 @@ function MovieClip(textures)
     this.animationSpeed = 1;
 
     /**
+     * @private
+     */
+    this._durationMultipliers = null;
+
+    /**
+     * Elepsed Time of current frame (Only active if durationMultipliers is used)
+     *
+     * @private
+     */
+    this._lag = 0;
+
+    /**
      * Whether or not the movie clip repeats after playing.
      *
      * @member {boolean}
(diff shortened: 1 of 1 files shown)


englercj:
Hmm, this seems like a hack to try and do heterogeneous timed frames. I'd rather buff the MovieClip to be able to specify a time in ms to spend on each frame, for maximum flexibility.

zOadT:
Okay sorry, I thought this is the most compatible way to implement it, because it works good in combination with _currentTime and animationSpeed. (Aren't heterogeneous timed frames the goal? :D )
I don't completely understand your suggestion, can you maybe provide a short example code showing how you would like to see the API and how it works in combination with animationspeed. Also what exactly to you mean by "buff the MovieClip".
Sorry for asking, but I want to do my best ;)

englercj:
> Okay sorry, I thought this is the most compatible way to implement it, because it works good in combination with _currentTime and animationSpeed

No problem, you are probably right. This is the most compatible way to make this happen. I just feel like it will complicate things too much.

> Also what exactly to you mean by "buff the MovieClip".

Sorry, I just mean "make it better" to support something it doesn't now.

> I don't completely understand your suggestion, can you maybe provide a short example code showing how you would like to see the API and how it works in combination with animationspeed.

I'm suggesting that you can use animation speed *or* specify a ms time that each frame is displayed. Not necessarily both at the same time.

For example:

```js
var frames = [
    { texture: ..., time: 20 },
    { texture: ..., time: 50 }
];

var aniamtion = new MovieClip(frames);
```

Something along those lines where each texture gets a time to be displayed associated with it.

zOadT:
Thanks for the response :) I will implement it like you suggest, so the durationMultipliers gets private and I use them only internally. But I think it's important to be possible to change the speed of an animation, so I will still provide an animationSpeed property (default: 1) but it will be handled differently internal, so the provided time actually corresponds to the time the frame is shown if animationSpeed is 1.
Just one last question: How do I need to define the parameter of MovieClip for the documentation?

englercj:
> How do I need to define the parameter of MovieClip for the documentation?

We use jsdoc.

What I am saying about the frame timing is that I don't think `durationMultipliers` are necessary, I think we can track those in ms instead. `animationSpeed` then just becomes a multiplier of the time a frame is meant to be shown.

zOadT:
Yes, this is the way I want it to implement (sorry for the confusing text :D) durationMultipliers are now just the durations per frame (I will rename them accordingly) I will open a new PR when I am finished.

englercj:
Awesome, sounds good!

PR #2008, opened 2015-07-28
Only add "update" listener to texture if baseTexture has a source

Fyrd:
baseTextures without a source will have `hasLoaded` set to false and never update. As a result, without this change "update" listeners are added but never get called, which can result in a memory leak. This leak is seen in our app, where the `._events.update` array is populated with hundreds of thousands of entries resulting in memory usage of > 250MB.

This is working on the assumption that "update" will only be called for textures with a source. If that's not necessarily the case then maybe there should be an alternate solution to the problem.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -164,7 +164,7 @@ Object.defineProperties(Sprite.prototype, {
                 {
                     this._onTextureUpdate();
                 }
-                else
+                else if (value.baseTexture.source)
                 {
                     value.once('update', this._onTextureUpdate, this);
                 }


englercj:
Is this to handle a WebGL RenderTexture? What case do you have that 1) generates that many textures and 2) they are all textures without sources.

englercj:
Bump, I'm not sure what this is solving. It would be helpful to see a test case of the issue.

rybon:
This is a relevant problem. Sometimes you'll want to create a pool of empty `Pixi.Sprite` objects (for memory / GC efficiency reasons) and fill them with images later on. When instantiating Sprites without an image source the code will default to `Texture.EMPTY`, which creates empty `Texture` and `BaseTexture` objects. When you add an image source later on, the `EE` context is bound to both the 'empty' `Texture` and `BaseTexture` objects and the new `Texture` and `BaseTexture` objects for that image source. Only the latter will be cleaned up eventually in the EventEmitter code, while the empty ones will keep hanging around in the listener arrays. Rinse and repeat and you'll have a memory leak.

englercj:
In that case this doesn't sound like the right solution. It sounds like it would be best to remove events a sprite has on a texture when the texture is changing.

GoodBoyDigital:
Perf wise, that would be a shame to add / remove events each time a texture is changed. Especially if there are lots of animations going on...

rybon:
Would it be possible to refactor Sprite just a little bit to allow for an empty image source? I think we can get away with just removing the `Texture.EMPTY` fallback and some extra checks for updating and destroying the texture / base texture? This would make some optimization scenarios such as pooling and reusing objects a bit easier and performant, especially on limited hardware.

Thoughts?

englercj:
> Perf wise, that would be a shame to add / remove events each time a texture is changed. Especially if there are lots of animations going on...

This is why we need a benchmark suite; so we can test different ways of changing things and examine the perf regressions and improvements.

thesunlover:
@rybon your refference isn't proper

englercj:
Closing due to inactivity.

PR #2012, opened 2015-07-30
Buff MovieClip: Added support for timed frames

zOadT:
Examle:

    var animation = new MovieClip([
        { texture: Texture(...), time: 100 },    // show 100 ms
        { texture: Texture(...), time: 0 },      // skip frame
        { texture: Texture(...), time: 500 }     // show 500 ms
    ]);

    console.log(animation.animatinSpeed);        // --> 1
    animation.animationSpeed = 2;                // play animation twice as fast

Like last time feedback is welcome ;)

regards

### DIFF
--- a/src/extras/MovieClip.js
+++ b/src/extras/MovieClip.js
@@ -19,16 +19,25 @@ var core = require('../core');
  * @class
  * @extends PIXI.Sprite
  * @memberof PIXI.extras
- * @param textures {Texture[]} an array of {Texture} objects that make up the animation
+ * @param textures {PIXI.Texture[]|Object[]} an array of {PIXI.Texture} or frame objects that make up the animation
+ * @param textures[].texture {PIXI.Texture} the {PIXI.Texture} of the frame
+ * @param textures[].time {number} the duration of the frame in ms
  */
 function MovieClip(textures)
 {
-    core.Sprite.call(this, textures[0]);
+    core.Sprite.call(this, textures[0] instanceof core.Texture ? textures[0] : textures[0].texture);
 
     /**
      * @private
      */
-    this._textures = textures;
+    this._textures = null;
+
+    /**
+     * @private
+     */
+    this._durations = null;
+
+    this.textures = textures;
 
     /**
      * The speed that the MovieClip will play at. Higher is faster, lower is slower
@@ -37,6 +46,13 @@ function MovieClip(textures)
      * @default 1
      */
     this.animationSpeed = 1;
+
+    /**
+     * Elepsed Time of current frame in ms (Only active if timed frames are used)
+     *
+     * @private
+     */
+    this._lag = 0;
 
     /**
      * Whether or not the movie clip repeats after playing.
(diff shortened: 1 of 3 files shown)


src/extras/MovieClip.js
```diff
@@ -19,16 +19,25 @@ var core = require('../core');
  * @class
  * @extends PIXI.Sprite
  * @memberof PIXI.extras
- * @param textures {Texture[]} an array of {Texture} objects that make up the animation
+ * @param textures {PIXI.Texture[]|Object[]} an array of {PIXI.Texture} or frame objects that make up the animation
```
englercj: Make these `{@link PIXI.Texture}` so they will link properly please!

src/extras/MovieClip.js
```diff
@@ -123,7 +151,12 @@ Object.defineProperties(MovieClip.prototype, {
     currentFrame: {
         get: function ()
         {
-            return Math.floor(this._currentTime) % this._textures.length;
+            var currentFrame = Math.floor(this._currentTime) % this._textures.length;
+            if (currentFrame < 0)
```
englercj: I'm confused by this, what is it here for? Looks like it is making it loop, but starting at the end of the list?

src/extras/MovieClip.js
```diff
@@ -170,8 +203,16 @@ MovieClip.prototype.gotoAndStop = function (frameNumber)
 
     this._currentTime = frameNumber;
 
-    var round = Math.floor(this._currentTime);
-    this._texture = this._textures[round % this._textures.length];
+    if (this._durations !== null)
```
englercj: Is this needed at gotoAndStop? Feel like this should just click over to the texture immediately and be done.

src/extras/MovieClip.js
```diff
@@ -170,8 +203,16 @@ MovieClip.prototype.gotoAndStop = function (frameNumber)
...
+    {
+        this._lag = frameNumber % 1 * this._durations[this.currentFrame];
```
englercj: Doesn't this always equal `0`? Anything mod 1 is 0, and since mod and multiplication are in the [same operator precedence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence) this just runs left-to-right making it `0 * value`.

src/extras/MovieClip.js
zOadT: It just handles the case that currentTime is negative (this is the case if animationSpeed (or deltaTime). e.g if currentTime is -4 and your animation has 3 frames it returns: (-4 % 3) + 3 = 2

src/extras/MovieClip.js
zOadT: You need to update the _lag property because you change currentTime. In case you set the frame to 1.5 the lag property is set so it's the half of the duration of frame 1

src/extras/MovieClip.js
zOadT: `frameNumber % 1` returns always `0` if framenumber is an integer. `5.5 % 1` returns `0.5`, so you can stop the animation between frames.
Maybe the `_lag` property is a bit confusing, I can remove the `_lag` property and read it from the fraction part of `_currentTime` if you want, I just thought it may be imprecise

src/extras/MovieClip.js
```diff
@@ -191,40 +217,59 @@ MovieClip.prototype.gotoAndPlay = function (frameNumber)
  */
 MovieClip.prototype.update = function (deltaTime)
 {
+    var elapsed = this.animationSpeed * deltaTime;
 
-    this._currentTime += this.animationSpeed * deltaTime;
+    if (this._durations !== null)
+    {
+        var lag = this._currentTime % 1 * this._durations[this.currentFrame];
```
englercj: Same comment as before, isn't this always 0?

zOadT: `_currentTime` is a float. `1.24 % 1` returns `0.24`. It's a shortcut for `_currentTime - _currentTime | 0` I want to determine how long the current frame is already be shown.

englercj:
:+1:

zOadT:
Thanks for you patience dude ;)

englercj:
Of course!

PR #2013, opened 2015-07-30
Mouseover bug fix

tengotengo:
Added cursor lock feature. set 
 renderer.mouseLock = true
to activate mouse lock within the next mouse click

fixed mouseover bug when mouseout event would fire while cursor is still in the hit area

### DIFF
--- a/src/core/renderers/canvas/CanvasRenderer.js
+++ b/src/core/renderers/canvas/CanvasRenderer.js
@@ -113,6 +113,13 @@ function CanvasRenderer(width, height, options)
         worldAlpha: 1
     };
 
+    /**
+     * If set to true, requests the mouse cursor lock inside the canvas.
+     *
+     * @member {boolean}
+     * @default false
+     */
+    this.mouseLock = false;
 
     this.resize(width, height);
 }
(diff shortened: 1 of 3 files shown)


drkibitz:
:+1:

src/core/renderers/canvas/CanvasRenderer.js
```diff
@@ -113,6 +113,13 @@ function CanvasRenderer(width, height, options)
         worldAlpha: 1
     };
 
+    /**
+     * If set to true, requests the mouse cursor lock inside the canvas.
+     *
+     * @member {boolean}
+     * @default false
+     */
+    this.mouseLock = false;
```
adelciotto: As this variable is in both the CanvasRenderer and WebGLRenderer; perhaps place it in the SystemRenderer base class.

englercj:
Why is there a property related to interation on the renderer?

drkibitz:
@englercj makes a good point with that question. Maybe change your thinking that you have to operate on renderer.view and look at using the manager's reference to interactionDOMElement.

englercj:
Closing this in favor of #2071

PR #2018, opened 2015-07-31
Bug fix with canvas context state stack and masking

RoryDouglas7:
On canvas when masking and changing the compositeOperation, the renderers currentBlendMode property will get out of sync with the compositeOperation (popping the canvas state returns the compositeOperation to it's previous state). 

This fix tracks the compositeOperation variable rather than currentBlendMode.


englercj:
Can you rebase this and exclude the build files please? We do not merge in changes to build files:

https://github.com/pixijs/pixi.js/blob/master/CONTRIBUTING.md

Thanks!

PR #2028, opened 2015-08-04
Fix TilingSprite offset issue with CanvasRenderer

sweetmandm:
There is a bug in the CanvasRenderer of TilingSprite when
tilePosition + tileScale are both set.

The WebGL version of this fix was in 6d5bedd, however at the
time I didn't realize that it didn't actually fix the
CanvasRenderer version. This resolves the issue, as
demonstrated here:

http://www.davidsweetman.com/pixi/canvas-tiling-sprite-fix/


englercj:
:+1:

MichaelHoste:
Nice! I came here to post this exact issue and it's already fixed. It's a nice day :-)

Issue #2031, opened 2015-08-05
Destroying texture loaded through assetLoader, won't remove texture from Base/TextureCache


autumn8:
Calling destroy(true, true) on a sprite that has a texture that was loaded using the assetLoader, doesn't remove the texture from TextureCache or BaseTextureCache as imageURL is not populated when texture is created.

Libary version: Pixi.js 3.0.7

Example usage:

```javascript
   //aset loading
   loader
    .add('bird', './bird.jpg')
    .load(onAssetsLoaded);
 ```

```javascript
  //on load
  function onAssetsLoaded() {
     bird = new PIXI.Sprite(loader.resources.bird.texture);
     stage.addChild(bird);
  }
```
```javascript
   //destroy
   bird.destroy(true, true);
 ```

```javascript
BaseTexture.prototype.destroy = function ()
{
    if (this.imageUrl)   //condition never met
    {
       
        delete utils.BaseTextureCache[this.imageUrl];
        delete utils.TextureCache[this.imageUrl];
        .....

    }
    .....
};
```

englercj:
Thanks for the report!

autumn8:
You're welcome. Looks like it should be a fairly easy fix in the textureParser  (also takes care of adding texture to the BaseTextureCache)

```javascript
    var baseTexture = new core.BaseTexture(resource.data, null, core.utils.getResolutionOfUrl(resource.url));
    baseTexture.imageUrl = resource.url;
    resource.texture = new core.Texture(baseTexture);    
    core.utils.BaseTextureCache[resource.url] = resource.texture;
    core.utils.TextureCache[resource.url] = resource.texture;
```

I'm happy to edit and send a pull request?

englercj:
Sounds good, that makes my life a lot easier!

PR #2033, opened 2015-08-06
Fix for issue #2031

autumn8:
1) Added imageUrl to baseTexture in textureParser so that texture can be properly removed from BaseTextureCache and TextureCache when BaseTexture.prototype.destroy() is called.

2) Added texture to BaseTextureCache.


englercj:
:+1: Nice job, looks great!

PR #2036, opened 2015-08-10
Add tint to tiling sprites in canvas renderer

berthise:


englercj:
Nice! We should also add it to WebGL so that they are consistent.

berthise:
It's already in WebGL renderer because it use the same backend function that Sprite

englercj:
> It's already in WebGL renderer because it use the same backend function that Sprite

Well, no; but I do see that the color attribute already exists in the TilingSprite shader so it is supported.

PR #2037, opened 2015-08-11
Added DisplayObject.setParent

mjhasbach:
Because sometimes I would like to do this:

```
var container = new pixi.Container(),
    graphic = new pixi.Graphics().setParent(container);
```

instead of this:

```
var container = new pixi.Container(),
    graphic = new pixi.Graphics();

container.addChild(graphic);
```


mjhasbach:
Accidentally selected master branch...

PR #2038, opened 2015-08-11
Added DisplayObject.setParent

mjhasbach:
Because sometimes I would like to do this:

```
var container = new pixi.Container(),
    graphic = new pixi.Graphics().setParent(container);
```

instead of this:

```
var container = new pixi.Container(),
    graphic = new pixi.Graphics();

container.addChild(graphic);
```

### DIFF
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -431,6 +431,25 @@ DisplayObject.prototype.generateTexture = function (renderer, scaleMode, resolut
 };
 
 /**
+ * Set the parent Container of this DisplayObject
+ *
+ * @param container {Container} The Container to add this DisplayObject to
+ * @return {DisplayObject} This DisplayObject
+ */
+DisplayObject.prototype.setParent = function (container)
+{
+    if (!container || !container.addChild)
+    {
+        throw new Error('setParent: Argument must be a Container');
+    }
+    else
+    {
+        container.addChild(this);
+        return this;
+    }
+};
+
+/**
  * Base destroy method for generic display objects
  *
  */
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -434,7 +434,7 @@ DisplayObject.prototype.generateTexture = function (renderer, scaleMode, resolut
  * Set the parent Container of this DisplayObject
  *
  * @param container {Container} The Container to add this DisplayObject to
- * @return {DisplayObject} This DisplayObject
+ * @return {Container} The Container that this DisplayObject was added to
  */
 DisplayObject.prototype.setParent = function (container)
 {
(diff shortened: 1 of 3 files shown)


src/core/display/DisplayObject.js
```diff
@@ -431,6 +431,25 @@ DisplayObject.prototype.generateTexture = function (renderer, scaleMode, resolut
...
+DisplayObject.prototype.setParent = function (container)
+{
+    if (!container || !container.addChild)
+    {
+        throw new Error('setParent: Argument must be a Container');
+    }
+    else
+    {
+        container.addChild(this);
+        return this;
```
englercj: `addChild` returns the child, you can just do `return container.addChild(this)`.

mjhasbach: Nice catch. Is there any reason `Container.addChild` returns the child instead of the `Container`? The `Container` returning itself would likely be more useful (for chaining). If you're amenable, I'll submit a PR for changing the return value of `Container.addChild`, otherwise I'll go ahead and make the requested changes to `DisplayObject.setParent`.

englercj: It is the way it is now after a discussion on a previous issue and other chats outside it. Right now we will likely keep it the way it is because the community felt that this is more common:

```js
var container = new PIXI.Container(),
    child = container.addChild(new PIXI.Sprite);
```

than this:

```js
var container = new PIXI.Container().addChild(sprite);
```

Here is the issue for reference: https://github.com/pixijs/pixi.js/issues/299

mjhasbach:
Cool. In the spirit of following convention, do you think I should refactor `DisplayObject.setParent` to return the `Container` instead of the `DisplayObject`,  or go ahead and make the originally requested changes?

englercj:
sure, it also means that your original use-case isn't as valid as it was because you still need to do it on a new line; in which case you could just use addChild :)

mjhasbach:
True. I think it would be best to stick with a consistent return pattern. There are still some cases where `DisplayObject.setParent` would be useful. For example:

```
container.addChild(
    new pixi.Graphics()
        .beginFill(color)
        .lineStyle(width, color)
        .drawRect(x, y, width, height)
        .endFill()
);
```

I feel this is cleaner:

```
new pixi.Graphics()
    .beginFill(color)
    .lineStyle(width, color)
    .drawRect(x, y, width, height)
    .endFill()
    .setParent(container)
```

Assignment and chaining is still possible like this.

```
var rect = new pixi.Graphics()
    .beginFill(color)
    .lineStyle(width, color)
    .drawRect(x, y, width, height)
    .endFill()
    .setParent(container)
    .children[container.children.length - 1]
```

englercj:
Makes sense!

Issue #1928, opened 2015-06-30
can i change totalSegs???


monpy:
Now big scale circle is not smoothly, because in build circle define `totalSegs = 40`.
I want to change this value on drawCircle Method.

englercj:
Good call, that is currently not possible.

aeauseth:
`totalSegs = Math.floor(30 * Math.sqrt(circleData.radius))` seems to work OK as a temporary fix. Got this idea from http://slabode.exofire.net/circle_draw.shtml.  I also noticed that under extreme zoom the ends of the circle don't meet up perfectly.
![capture](https://cloud.githubusercontent.com/assets/13549766/8950247/cbba8b5e-3573-11e5-9608-2ae87aa040e2.PNG)

stbaer:
+1

englercj:
Closing since the PR was merged.

PR #2040, opened 2015-08-11
Changed totalSegs in GraphicsRenderer buildCircle …

stbaer:
…from 40 to a dynamic value that depends on the radius of the circle, see issue #1928 - the new value was proposed by @aeauseth 

totalSegs = 40:
http://jsfiddle.net/n4w4w7yj/

totalSegs = Math.floor(30 * Math.sqrt(circleData.radius));
http://jsfiddle.net/7aLhx5r9/

### DIFF
--- a/src/core/graphics/webgl/GraphicsRenderer.js
+++ b/src/core/graphics/webgl/GraphicsRenderer.js
@@ -515,7 +515,7 @@ GraphicsRenderer.prototype.buildCircle = function (graphicsData, webGLData)
         height = circleData.height;
     }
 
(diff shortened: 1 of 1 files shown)


giraluna:
This breaks Graphics.drawEllipse. graphicsData.shape can also be of type Ellipse which has no radius property leading totalSegs to be NaN.

http://jsfiddle.net/7aLhx5r9/1/
outputs no image and "GL ERROR :GL_INVALID_OPERATION : glDrawElements: attempt to access out of range vertices in attribute 0" in console

Issue #2006, opened 2015-07-28
Canvas Renderer + Retina (resolution - 2) + Tint not working


adireddy:
http://jsfiddle.net/adireddy/uf1zggL2/

When I tried to apply tint on a retina asset, the image is being cropped.

It works on non-retina asset.

englercj:
Thanks @adireddy certainly looks like an issue, will look into it.

alebianco:
it seems that on a mobile devices (tested ipad4 ios 6.1 so far)  it doesn't even render the first quarter of the sprite, it fully disappears

alebianco:
I had a quick dig, in both cases it's because the CanvasTinter, when using tintWithPerPixel (mobile) or tintWithMultiply (desktop), does not consider the baseTexture.resolution
I need to do more tests but I could have a patch coming soon

stbaer:
+1 for getting this fixed

dohomi:
+1 for fixing

PR #2041, opened 2015-08-11
fixes wrong jsdoc param name

stbaer:


englercj:
Is there a reason you closed this?

stbaer:
Just thought it's more than a month now and there are other PR's that are more important / I'd prefer to see reviewed (#2006 for example), feels like they are stacking up at the moment.

englercj:
They are stacking up quite a bit, unfortunately I just don't have the time to focus on Pixi :(

PR #2045, opened 2015-08-12
Add calls to super destructors and other extra cleanup work.

davedx:
Hi guys,

We've been doing some intensive memory usage testing here, and found that when running our app with Pixi for 12 hours for a KPI test we had some JavaScript memory leaks, some of which we tracked down to Pixi. The problems are:

* Super destructors (destroy) not always being called - this can result in hanging event subscriptions
* Not all managers being destroyed and nulled
* A couple of event subscriptions on Texture not being unsubscribed

What's your thoughts?

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -310,6 +310,7 @@ WebGLRenderer.prototype.setObjectRenderer = function (objectRenderer)
     }
 
     this.currentRenderer.stop();
+    this.currentRenderer.destroy();
     this.currentRenderer = objectRenderer;
     this.currentRenderer.start();
 };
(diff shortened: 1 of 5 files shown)


src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -310,6 +310,7 @@ WebGLRenderer.prototype.setObjectRenderer = function (objectRenderer)
     }
 
     this.currentRenderer.stop();
+    this.currentRenderer.destroy();
```
englercj: I don't think this is desired, this would destroy the object renderers everytime you switched to a different renderer. For example, if you render a sprite then a graphics object when you render the graphics object the sprite renderer will be destroyed.

The object renderers are registered as plugins of the webgl renderer and should be destroyed in the destroy method when all other plugins are destroyed.

src/mesh/webgl/MeshRenderer.js
```diff
@@ -209,4 +209,7 @@ MeshRenderer.prototype.start = function ()
  */
 MeshRenderer.prototype.destroy = function ()
 {
+    if (this.renderer) {
```
englercj: Is this check necessary? When is there not a renderer value in this object renderer?

src/core/renderers/webgl/WebGLRenderer.js
davedx: Aha, understood. I'll revisit this in our code base and push an update tomorrow hopefully.

davedx:
@englercj I have updated the PR by removing the bits you had comments on, as we're not 100% sure about why we needed them at this point. We need to find more time to re-run our memory tests with and without those changes. Does the PR look any better/make sense now? :)

Issue #1620, opened 2015-03-31
graphics lineStyle thickness bug


NikkiKoole:
version : "3.0.0-rc4"

When I set the lineStyle smaller then 1 and then do a drawRect the outline isn't aligned properly to the shape fill:
![screen shot 2015-03-31 at 12 25 56 pm](https://cloud.githubusercontent.com/assets/453055/6917011/1b72840c-d7a1-11e4-9e59-310403e90689.png)
```
var thing = new PIXI.Graphics();
thing.lineStyle(.1, 0x000000);
thing.beginFill(0xffffff);
thing.drawRect(0,0,20,20);
stage.addChild(thing);
```

here's a fiddle showing the issue:
http://jsfiddle.net/332gdcna/8/

I've played around with that fiddle myself:
and separated the lineStyle and beginFill to different Graphics objects.
It looks like the fill part should be drawn at 0.5, 0.5 then the fill and outline parts do align properly.
http://jsfiddle.net/332gdcna/9/

adelciotto:
Just thought I'd this bug only appears when using the WebGL renderer.

NikkiKoole:
I've fiddled around a little more: http://jsfiddle.net/332gdcna/12/

actually it seems the fill part in the webgl is done properly, it looks like all the lines should be positioned += -0.5, -0.5.

adelciotto:
ahhh ok, I get ya. so would simply adding to that if statement like so:

```javascript
if (graphicsData.lineWidth % 2 && graphicsData.lineWidth >= 1)
{
    ....
}
```

be appropriate ? Because % 2 with any floating point number would never return 0 anyway.

NikkiKoole:
I've tried that, and also a third check to see if the lineWidth is an int to begin with:
```
if (graphicsData.lineWidth % 2 && graphicsData.lineWidth >= 1 && (graphics.lineWidth%1===0)) {...}
```
but I am not sure if the whole test should be there to begin with... ;) In my current project just commenting the whole lot out looks better ;)

gaelduplessix:
I think this issue is related to #711

englercj:
Just to confirm, does this issue still exist in the latest v3 release?

gaelduplessix:
It probably does, since the condition is still present:
https://github.com/GoodBoyDigital/pixi.js/blob/dev/src/core/graphics/webgl/GraphicsRenderer.js#L583

simonbw:
[...]

PR #2055, opened 2015-08-18
webgl/GraphicsRenderer buildline change: fixes #711 and #1620

### LINKED ISSUE #711: Graphics line position is off when inside scaled DisplayObjectContainer
When placing a Graphics object inside a DisplayObjectContainer that has been scaled up.

Looks fine for the CanvasRenderer but the position of the line is off when using the WebGL renderer. Tested on Chrome and IE (Win8)

http://jsfiddle.net/LyDx9/5/

NikkiKoole:


chrispine:
This also fixes an incompatibility between WebGL and Canvas rendering. In other words: as it was before this PR, if you fixed your pixel alignment the *right* way in your Canvas rendering, the above code would **break it in WebGL**.

This change is For The Good. :+1:

PR #2056, opened 2015-08-18
graphics update

wayfu:
fixes a number of issues with open and closed paths


alvinsight:
Hi there, thanks for overcoming all the woes that github's put in your way ! 
It's definitely a tricky system, and a total mindfuck in the beginning, but it gets easier after a short while, keep the faith ! 
We'll review this PR shortly :)

PR #2069, opened 2015-08-26
Fix #2006 by applying texture resolution to crop size

### LINKED ISSUE #2006: Canvas Renderer + Retina (resolution - 2) + Tint not working

alebianco:
From the tests that i could do, both on desktop, emulators and devices, this seems to fix the issue without side effects ...

### DIFF
--- a/src/core/renderers/canvas/utils/CanvasTinter.js
+++ b/src/core/renderers/canvas/utils/CanvasTinter.js
@@ -69,15 +69,13 @@ CanvasTinter.tintWithMultiply = function (texture, color, canvas)
     var resolution = texture.baseTexture.resolution;
 
     var crop = texture.crop;
-    crop.width *= resolution;
-    crop.height *= resolution;
 
-    canvas.width = crop.width;
-    canvas.height = crop.height;
+    canvas.width = crop.width * resolution;
+    canvas.height = crop.height * resolution;
 
     context.fillStyle = '#' + ('00000' + ( color | 0).toString(16)).substr(-6);
 
-    context.fillRect(0, 0, crop.width, crop.height);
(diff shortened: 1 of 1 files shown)


stbaer:
Hi @alebianco - thank you for looking into this. 

I just tested the pr and it doesn't fix it yet for me. Looks like image positions are off, for example if the anchor is set to 0.5.

alebianco:
looks like i've been to optimistic in reducing the number of multiplications.
i've made explicit multiplications where needed now. not sure if that's preferable to cloning the texture crop or not...

stbaer:
:+1: noice

stbaer:
Looks good for normal images. But the combination CanvasRenderer + Retina + Sprite-sheet doesn't work, neighboring images are overlapping (this might be a different issue though).

alebianco:
any chance you have an example of that?

stbaer:
Am travelling atm, but will try to make a fiddle this weekend or early next week.

alebianco:
I managed to test it, is this what you saw? https://jsfiddle.net/alebianco/vkjeumkb/
it does happen regardless of resolution and i'd say it's a separate issue, even if it's always caused by tinting

stbaer:
Not sure if it's the same problem, but I'd also say it's a separate issue. Will take a closer look soon.

alebianco:
Applying a tint to a spritesheet animation will still display incorrectly, but _static_ images taken from spritesheets should be ok now

stbaer:
Yes, +1 for merging / reviewing this pr.

PR #2076, opened 2015-09-01
documentation fixes #2066

### LINKED ISSUE #2066: clearBeforeRender has no effect in webgl?
Not sure I'm going the wrong way with this but it definitely works different compared to canvas renderer.

I want to stamp sprites to scene and want them to remain on the screen after I remove them from the scene graph. This works just fine with canvas renderer but webgl somehow cleans the screen each frame.

Demo:

Canvas (expected behaviour): https://jsfiddle.net/a2oar38g/6/
WebGL (everything clears at each frame): https://jsfiddle.net/a2oar38g/7/

mreinstein:


englercj:
I feel like you `dev` is out of date:

https://github.com/pixijs/pixi.js/blob/dev/src/core/textures/RenderTexture.js#L20-L21

PR #2085, opened 2015-09-08
Fixed the AsciiFilter Y Flip

SkuliOskarsson:
AsciiFilter no longer flips the image by it's Y Axis


englercj:
This fixes the symptom but I had that TODO in there to find out the cause. I'm not convinced this is the "correct" fix for this issue.

SkuliOskarsson:
Yeah your might right about that. I figured it might be better to have this fix instead of having it upside down. I'm going to look some more into this soon.

I should have added more information in the pull request description about that.

GoodBoyDigital:
sorted in the new pixi filters branch I believe!

PR #2087, opened 2015-09-08
Fix renderer plugin destruction exceptions

aschampion:
Closes pixijs/pixi.js#2086. Trivial fix, so I included another that popped up for `MeshRenderer`. [Fixed jsFiddle](http://jsfiddle.net/ho6qdzxk/4/).


englercj:
:+1:

Issue #1968, opened 2015-07-09
Destroyed WebGLRenderer still listens to dispose events of base textures


aschampion:
Currently, multiple renderers/contexts may make use of the same `BaseTexture` because they each manage their own GL textures in `updateTexture`. However, if a renderer is destroyed it does not unregister its listeners to textures' `dispose` events. If the texture is later destroyed because it is no longer needed, the listener for the destroyed renderer is called, which excepts [here](https://github.com/GoodBoyDigital/pixi.js/blob/master/src/core/renderers/webgl/WebGLRenderer.js#L426) because `this.gl` is null in the destroyed renderer.

[Contrived JSFiddle with v3.0.7, which excepts](http://jsfiddle.net/jckw77qd/). A fixed version would not throw and the canvas would be blank rather than show the bunny.

This can occur even with one renderer, if the code managing texture destruction outside of Pixi destroys a texture after a renderer has been destroyed. The single renderer case could reasonably be viewed as not Pixi's responsibility, though.

Either the renderer should unregister its texture listeners, or `destroyTexture` should check that the renderer has a context (i.e., `gl` is not null). The former's advantage is it also allows the renderer to be GC'd, but would require the destroyed renderer checking every texture in the base texture cache for its context in `_glTextures`, even textures it has never used. The latter is easy but does not help clean up references in listeners that may keep the destroyed renderer from being GC'd.

englercj:
Good catch! Will have to think about the best way to clean this up.

aschampion:
I have a question related to this issue. Is it intended that `utils.BaseTextureCache` contains only objects of type `BaseTexture`, not `Texture`? I ask because this would be expected from the name (and the existence of a separate`utils.TextureCache`), and some code ([example](https://github.com/pixijs/pixi.js/blob/f5abaa7060bd1347cfd0506ab2b60db9fc8c9f19/src/core/renderers/webgl/WebGLRenderer.js#L461)) seems to assume that is the case. However, in practice `Texture` objects often end up in this cache. One cause is [here](https://github.com/pixijs/pixi.js/blob/f5abaa7060bd1347cfd0506ab2b60db9fc8c9f19/src/loaders/textureParser.js#L14) due to #2031/#2033, though I'm still checking for others.

If the cache is intended to only contain `BaseTexture`, I'll identify and fix cases of insertion of `Texture` as part of this PR for this issue. Otherwise I will identify and fix cases where the type is assumed to be `BaseTexture`.

PR #2089, opened 2015-09-08
WebGLRenderer texture event listener cleanup

aschampion:
[Fixed jsFiddle](http://jsfiddle.net/q3byr0wz/2/) that includes these changes for #1968 on top of those from #2087. While touching every texture's listeners twice can be expensive, destroying a renderer should never be in performance-critical code paths anyway.

The JSHint failure is fixed by #2087.

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -475,6 +475,12 @@ WebGLRenderer.prototype.destroy = function (removeView)
     this.view.removeEventListener('webglcontextlost', this.handleContextLost);
     this.view.removeEventListener('webglcontextrestored', this.handleContextRestored);
 
+    for (var key in utils.BaseTextureCache) {
+        var texture = utils.BaseTextureCache[key];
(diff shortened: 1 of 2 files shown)


src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -475,6 +475,12 @@ WebGLRenderer.prototype.destroy = function (removeView)
     this.view.removeEventListener('webglcontextlost', this.handleContextLost);
     this.view.removeEventListener('webglcontextrestored', this.handleContextRestored);
 
+    for (var key in utils.BaseTextureCache) {
+        var texture = utils.BaseTextureCache[key];
+        texture.off('update', this.updateTexture, this);
+        texture.off('dispose', this.destroyTexture, this);
+    }
```
englercj: This is a fantastic change, but just want to point out that it is not garuenteed to capture all textures that this renderer put a listener on. You can create textures without them being put into that cache, and in those cases they won't be caught by this. But this is much better than not having it at all.

PR #2091, opened 2015-09-10
Fixed resolution not being respected when set externally - fixes #1867

### LINKED ISSUE #1867: PIXI 3 - Texture does not represent dimensions correctly after baseTexture resolution change
I have come across an issue with PIXI.Texture since moving across to version 3.  We need to support SD assets for our projects and one of the ways we go about this in the last few versions of PIXI is by simply adjusting the baseTexture resolution for each texture when requested.

var texture = PIXI.Texture.fromImage("image");    *// image is 1024x1024*
texture.baseTexture.resolution = 0.5;
texture.baseTexture.update();

**Example OLD**
console.log(texture.baseTexture.width);    *// 1024 correct* 
console.log(texture.baseTexture.realWidth);    *// 512 correct*
console.log(texture.width);    *// 1024 correct*

**Example CURRENT**
console.log(texture.baseTexture.width);    *// 1024 correct*
console.log(texture.baseTexture.realWidth);    *// 512 correct*
console.log(texture.width);    *// 512 wrong*

I am assuming this is a bug but if this IS desired behaviour can you point us in a better way of doing things - maybe?

Thanks

andrewstart:
This fixes inconsistencies with how Texture & BaseTexture width/height is used with the BaseTexture's resolution.
An [updated JSFiddle](http://jsfiddle.net/qhyh3otr/5/) from #1867 shows that it works properly for both asset resolutions as well as Canvas and WebGL rendering.
I haven't tested it, but this might also fix any issues with Canvas tinting at resolutions other than 1.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -109,11 +109,11 @@ Object.defineProperties(Sprite.prototype, {
     width: {
         get: function ()
         {
-            return Math.abs(this.scale.x) * this.texture._frame.width;
+            return Math.abs(this.scale.x) * this.texture.width;
         },
         set: function (value)
         {
-            this.scale.x = utils.sign(this.scale.x) * value / this.texture._frame.width;
+            this.scale.x = utils.sign(this.scale.x) * value / this.texture.width;
             this._width = value;
         }
     },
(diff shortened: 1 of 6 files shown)


ivanpopelyshev:
That's not how resolution is supposed to work. Texture frame must be the same for 512x512x1 , 1024x1024x2 .

englercj:
Closing due to inactivity.

Issue #2093, opened 2015-09-13
gulp running clean after build and exits


brean:
Hi,

I just installed a new node with all required modules and tried to build pixi.js but the build process just called jshint, build, clean and exits, so I do not have a bin/-folder:
```bash
$ gulp
[13:08:41] Using gulpfile ~/coding/gown/src/pixi.js/gulpfile.js
[13:08:41] Starting 'jshint'...
[13:08:41] Starting 'build'...
[13:08:41] Starting 'clean'...
[13:08:45] Finished 'jshint' after 3.79 s
```

When I call done directly in clean (without doing any actual cleaning) it also runs other tasks:
```javascript
gulp.task('clean', function (done) {
    done();
});
```

```bash
$ gulp
[13:12:52] Using gulpfile ~/coding/gown/src/pixi.js/gulpfile.js
[13:12:52] Starting 'jshint'...
[13:12:52] Starting 'build'...
[13:12:52] Starting 'clean'...
[13:12:52] Finished 'clean' after 94 μs
[13:12:52] Starting 'scripts'...
[13:13:05] Finished 'scripts' after 13 s
[13:13:05] Finished 'build' after 13 s
[13:13:06] Finished 'jshint' after 15 s
[13:13:06] Starting 'default'...
[13:13:06] Finished 'default' after 5.24 μs
```

I am running ubuntu 14-04 with node v4.0.0 (installed via nodeenv inside a python-virtualenv)

englercj:
This is because you got a new version of `del` that is backwards incompatible.

englercj:
Fixed in the latest dev.

PR #2094, opened 2015-09-13
clean task fix (for del versions >= 2)

stbaer:
This should fix issue #2093


stbaer:
The return also doesn't hurt versions of del < 2 (currently ^1.2.0 in pixie's package.json). This makes it a bit more future proof.

englercj:
For `del` > v2 you should also remove the `done` param.

stbaer:
True, removed done and updated the del version.

PR #2097, opened 2015-09-15
Cleanup primitive and complex shader during ShaderManager destroy

SanderSpies:
Not doing this causes more shaders to be created then destroyed.


englercj:
:+1:

PR #2118, opened 2015-09-24
drawOneSideRoundedRect for Graphics

Mgonand:
Draw a one side rounded rectangle, like rounded bars in charts.

Usage:
graphics.drawOneSideRoundedRect(x, y, width, height, radius, PIXI.POSITION);


englercj:
This looks cool, there a couple things I wanted to point out though:

1. No need to add things to the .gitignore :)
2. Do not commit build files (bin/)
3. This should be in a plugin

This is an awesome addition but there is no reason it *needs* to be in the core lib. I would rather you made a plugin and then added it to our plugins page:

https://github.com/pixijs/pixi.js/wiki/Pixi-v3-Plugins

Thanks for the awesome contribution!

Mgonand:
I just did it for my own project and I wanted to share it.

I don't know if I have time to do the plugin, but i'll try in my free time :)

Thanks for the awesome work, keep going!

englercj:
Well you basically did the plugin in this PR :) Just need to move those to a separate file and push to GitHub :)

PR #2134, opened 2015-09-30
Add blurred text shadow; Fix angle and distance never getting set to 0

onion-soup:
* Adds the canvas' `shadowBlur` feature in order to render blurred shadow behind the text.
* Fixes the constructor so that we can set `dropShadowAngle` and `dropShadowDistance` to `0`. Before it would just overwrite it with the default values.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -31,6 +31,7 @@ var Sprite = require('../sprites/Sprite'),
  * @param [style.dropShadowColor='#000000'] {string} A fill style to be used on the dropshadow e.g 'red', '#00FF00'
  * @param [style.dropShadowAngle=Math.PI/4] {number} Set a angle of the drop shadow
  * @param [style.dropShadowDistance=5] {number} Set a distance of the drop shadow
+ * @param [style.dropShadowBlur=0] {number} Set a shadow blur radius
  * @param [style.padding=0] {number} Occasionally some fonts are cropped on top or bottom. Adding some padding will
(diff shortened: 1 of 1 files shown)


src/core/text/Text.js
```diff
@@ -196,8 +198,9 @@ Object.defineProperties(Text.prototype, {
 
             style.dropShadow = style.dropShadow || false;
             style.dropShadowColor = style.dropShadowColor || '#000000';
-            style.dropShadowAngle = style.dropShadowAngle || Math.PI / 6;
-            style.dropShadowDistance = style.dropShadowDistance || 5;
+            style.dropShadowAngle = style.dropShadowAngle !== undefined ? style.dropShadowAngle : Math.PI / 6;
+            style.dropShadowDistance = style.dropShadowDistance !== undefined ? style.dropShadowDistance : 5;
+            style.dropShadowBlur = style.dropShadowBlur !== undefined ? style.dropShadowBlur : 0; //shadowBlur is '0' by default according to HTML
```
drkibitz: Probably not really part of this PR, but since I just saw this, thought I would comment that it would be great if this just used `Object.assign` here:
```js
this._style = Object.assign({}, Text.defaultStyle, style);
```
Yes, `Text.defaultStyle` does not exist just yet, but you get the idea.

GoodBoyDigital: Thats a great idea @drkibitz !

GoodBoyDigital:
Thanks for this pr @onion-soup ! I think we could definitely look into @drkibitz  idea about using object.assign when we get some time too!

PR #2162, opened 2015-10-22
Added setTransform convenience method

bigtimebuddy:
@GoodBoyDigital For the PIXI.flash work


GoodBoyDigital:
Nice one @bigtimebuddy :+1: 
This will come in handy for many peeps!

Issue #2159, opened 2015-10-22
Sprite doesn't display after sprite.width = 0 and sprite.width = nonzero number


ayamada:
See http://plnkr.co/edit/iuk2n6fSVmQsbokd1A4p?p=preview .

`sp.width = number` work well.
But, this code doesn't work after to run `sp.width = 0`.
`sp` is not display, if any i ran `sp.width = nonzero_number` !

Sorry for my poor english ...

englercj:
https://github.com/pixijs/pixi.js/blob/master/src/core/sprites/Sprite.js#L116

Because of this line. `sign(0) === 0` which causes it never to be a non-zero value. Thanks for bringing this up.

englercj:
Closed by #2166.

PR #2166, opened 2015-10-24
Issue #2159: Force to sign=1 when scale=0

ayamada:
Thanks to response issue #2159, @englercj san!

I try to fix this problem, and i confirmed this fix works well in my environment.

But, i am not sure, this fix are best fit to this problem.

As well, https://github.com/pixijs/pixi.js/blob/master/src/core/sprites/Sprite.js#L186 has codes like `utils.sign(this.scale.x)` too, but there are not need to convert zero sign, i seem.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -113,7 +113,8 @@ Object.defineProperties(Sprite.prototype, {
         },
         set: function (value)
         {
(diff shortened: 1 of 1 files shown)


englercj:
:+1: Nailed it imo!

ayamada:
Thank you!

PR #2169, opened 2015-10-26
Standardized the class construction and inheritance

bigtimebuddy:
## Overview

* Adds `PIXI.utils.extend`, for instance `PIXI.utils.extend(MyClass, EventEmitter);`
 * Adds the `constructor` property to the prototype
 * Extends the prototype from the parent to the child using `Object.create(parent.prototype)`
 * Adds a static `extend` function to the child class for extending further
 * Gives a return value of the child class (first parameter)
* Where applicable, replaced class constructor/inheritance with short-hand (e.g.,
`DisplayObject.extend(Container)`). This does the same thing as calling `PIXI.utils.extend`.

## Example Usage 

```js
function MySprite(texture)
{
	Sprite.call(this, texture);
}
Sprite.extend(MySprite);
```

Or alternately: 

```js
var MySprite = Sprite.extend(function(texture)
{
	Sprite.call(this, texture);
});
```

As opposed to the more verbose:

```js
function MySprite(texture)
{
	Sprite.call(this, texture);
};
MySprite.prototype = Object.create(Sprite.prototype);
MySprite.prototype.constructor = MySprite;
```

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -28,10 +28,7 @@ function Container()
     this.children = [];
 }
 
-// constructor
-Container.prototype = Object.create(DisplayObject.prototype);
-Container.prototype.constructor = Container;
-module.exports = Container;
+module.exports = DisplayObject.extend(Container);
 
 Object.defineProperties(Container.prototype, {
     /**
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -1,4 +1,5 @@
 var math = require('../math'),
+    utils = require('../utils'),
     RenderTexture = require('../textures/RenderTexture'),
     EventEmitter = require('eventemitter3'),
     CONST = require('../const'),
@@ -150,10 +151,7 @@ function DisplayObject()
     this._mask = null;
 }
 
-// constructor
-DisplayObject.prototype = Object.create(EventEmitter.prototype);
-DisplayObject.prototype.constructor = DisplayObject;
-module.exports = DisplayObject;
+module.exports = utils.extend(DisplayObject, EventEmitter);
 
 Object.defineProperties(DisplayObject.prototype, {
     /**
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -158,10 +158,7 @@ function Graphics()
      */
 }
 
-// constructor
-Graphics.prototype = Object.create(Container.prototype);
-Graphics.prototype.constructor = Graphics;
-module.exports = Graphics;
+module.exports = Container.extend(Graphics);
 
 /**
  * Creates a new Graphics object with the same values as this one.
--- a/src/core/graphics/GraphicsData.js
+++ b/src/core/graphics/GraphicsData.js
@@ -1,3 +1,5 @@
+var utils = require('../utils');
+
 /**
  * A GraphicsData object.
  *
@@ -62,8 +64,7 @@ function GraphicsData(lineWidth, lineColor, lineAlpha, fillColor, fillAlpha, fil
     this.type = shape.type;
 }
 
-GraphicsData.prototype.constructor = GraphicsData;
-module.exports = GraphicsData;
+module.exports = utils.extend(GraphicsData);
 
 /**
  * Creates a new GraphicsData object with the same values as this one.
--- a/src/core/graphics/webgl/GraphicsRenderer.js
+++ b/src/core/graphics/webgl/GraphicsRenderer.js
@@ -31,9 +31,7 @@ function GraphicsRenderer(renderer)
     this.maximumSimplePolySize = 200;
 }
 
-GraphicsRenderer.prototype = Object.create(ObjectRenderer.prototype);
-GraphicsRenderer.prototype.constructor = GraphicsRenderer;
-module.exports = GraphicsRenderer;
+module.exports = ObjectRenderer.extend(GraphicsRenderer);
 
 WebGLRenderer.registerPlugin('graphics', GraphicsRenderer);
 
--- a/src/core/graphics/webgl/WebGLGraphicsData.js
+++ b/src/core/graphics/webgl/WebGLGraphicsData.js
@@ -1,3 +1,5 @@
+var utils = require('../../utils');
+
 /**
  * An object containing WebGL specific properties to be used by the WebGL renderer
  *
@@ -67,8 +69,7 @@ function WebGLGraphicsData(gl) {
     this.glIndices = null;
 }
 
-WebGLGraphicsData.prototype.constructor = WebGLGraphicsData;
-module.exports = WebGLGraphicsData;
+module.exports = utils.extend(WebGLGraphicsData);
 
 /**
  * Resets the vertices and the indices
--- a/src/core/math/Matrix.js
+++ b/src/core/math/Matrix.js
@@ -2,7 +2,8 @@
 // should either fix it or change the jshint config
 // jshint -W072
 
-var Point = require('./Point');
+var Point = require('./Point'),
+    utils = require('../utils');
 
 /**
  * The pixi Matrix class as an object, which makes it a lot faster,
@@ -53,8 +54,7 @@ function Matrix()
     this.ty = 0;
 }
 
-Matrix.prototype.constructor = Matrix;
-module.exports = Matrix;
+module.exports = utils.extend(Matrix);
 
 /**
  * Creates a Matrix object based on the given array. The Element to Matrix mapping order is as follows:
--- a/src/core/math/Point.js
+++ b/src/core/math/Point.js
@@ -1,3 +1,5 @@
+var utils = require('../utils');
+
 /**
  * The Point object represents a location in a two-dimensional coordinate system, where x represents
  * the horizontal axis and y represents the vertical axis.
@@ -22,8 +24,7 @@ function Point(x, y)
     this.y = y || 0;
 }
 
-Point.prototype.constructor = Point;
-module.exports = Point;
+module.exports = utils.extend(Point);
 
 /**
  * Creates a clone of this point
--- a/src/core/math/shapes/Circle.js
+++ b/src/core/math/shapes/Circle.js
@@ -1,4 +1,5 @@
 var Rectangle = require('./Rectangle'),
+    utils = require('../../utils'),
     CONST = require('../../const');
 
 /**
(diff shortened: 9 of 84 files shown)


GoodBoyDigital:
Nice! Thanks @bigtimebuddy :+1:  
This looks pretty awesome to me! @englercj looks cool to you too?

englercj:
Looks good to me, I'd rather the plugin mixin wasn't in the extend method though. I still feel that should be separate since it is a separate concern.

If you remove the plugin mixin part of `extend` it looks good to me! Thanks for this! Very cool.

bigtimebuddy:
@englercj Good call, I removed the pluginTarget mixin stuff from utils.extend.

drkibitz:
Nothing against the great work here @bigtimebuddy, but having been bit many times by *standard* yet framework specific classical inheritance implementations that tend to bleed all over the place, I'm a :-1: on this. Contributors tend to feel they can't implement a class without `utils`, or forget, or maybe prefer to forget how JavaScript works. Plugins with classes must decide when and how to hook into the core utils. Overall it usually just ends up encouraging bad, or at least, non-standard practices.

If people are all for less verbosity, I would suggest one of following 2 things.

1. Just do this:
```javascript
function MySprite(texture)
{
    Sprite.call(this, texture);
}
MySprite.prototype = Object.create(Sprite.prototype, {constructor: {value: Sprite}});
```
It's not really all that bad IMO, and is the actual pre-ES6 standard.

2. Use a transpiler (e.g. [Babel](https://babeljs.io/)), which conforms to next-gen JavaScript syntax:
```javascript
class MySprite extends Sprite  {
    constructor() {
        super();
    }
}
```
Talk about less verbosity :wink:  This was discussed at some point this year (pretty sure it was before June). If a PR like this one is actually being considered for merge, I believe it should be up for discussion again. It's not like we'd be targeting an incomplete [spec](http://www.ecma-international.org/ecma-262/6.0/) at this point.

bigtimebuddy:
I think it would be helpful to have some context. The motivation for adding this and several other new features recently (`DisplayObject.setTransform`, `DisplayObject.skew`) is that we're laying groundwork for creating a publishing pipeline for Flash. There are some essential ingredients missing from Pixi to provide compatibility with Flash as well as some design choices we've made, like the need to export JavaScript from Flash that is vanilla Pixi. 

There is a strong need to create _some_ native Pixi inheritance model/utility that is concise and readable. With Flash, it's not realistic at this point to be targeting ES6 for the output JS and arguably the single-line (`MySprite.prototype = Object.create(Sprite.prototype, {constructor: {value: Sprite}});`) is clever but isn't _that_ more useful size-wise than creating two lines (96 vs 85 characters). We'll need to be creating export Flash files that contain 100s of extending library items (e.g., [Flash output example](https://github.com/CloudKidStudio/PixiFlash/blob/dev/examples/color_effects/ColorEffects.js)). Having an extending convenience utility will really help.

@GoodBoyDigital I defer to you. The alternative solution is that I'm okay adding this functionality to `PIXI.flash` instead, which could provide this exact inheritance by adding static methods for Sprites, Container, Graphics, DisplayObjects, etc.

englercj:
So to be clear, I am OK with using this since the majority of people will likely enjoy using it more than `Object.create`.

That being said @drkibitz is 100% correct. The idiomatic and "correct" way is what he posted. The way this PR implements is custom, non-standard, and requires users to know how to "extend in PIXI" instead of just doing normal JS prototype extension.

I do not subscribe to the notion that this or any change at all is necessary for PIXI Flash, nor do I agree that our current implementation is not readable. It is prototype extension, it is JavaScript; this is just how it works in this language.

I still feel like this PR is OK, if people want it, because people who understand JS less can easily extend one of our objects. But as an experienced JS developer, nothing is worse than opening a lib and seeing some weird custom extension method. It almost makes me want to give up on trying to figure out the lib.

Finally, after all that, I would suggest that we compromise. Do not use a custom extension method, but do use an inherit method. One that has been around for a while, is pretty common and people are familiar with. We are using browserify, so you can `require('util')` and just use the built-in node.js [utils.inherits][nd].

Obviously the best would be to write ES6 and just run it through babel, and I'm not sure what your concerns about the output of PIXI flash being ES6 are but we should talk about that (not in this thread). I definitely see the future of pixi going to ES6.

-------

Specifically about this PR and not the concept of prototype extension methods in general; there is some inconsistencies in this PR that makes it difficult to immediately understand whats going on.

examples:

Some sites use `utils.extend()` while others use `Parent.extend()`. I understand *why* but that is only because I understand the implementation of this function. If you have to understand the implementation of an API to use it, I feel like we can do better.

There are a couple places like this:

```js
module.exports = utils.extend(GraphicsData);
```

Which doesn't make sense to me. We aren't extending anything so why is it there? I realize we were setting constructor before and maybe this is here to do that, but setting constructor was superfluous and this is confusing. It looks like a typo since we are calling `extend` but not extending anything.

I think if we standardize on `util.inherits()` and use that everywhere we are truly consistent and the inherits method will [do the right thing][il]. Hopefully that sounds agreeable.

[nd]: https://nodejs.org/docs/latest/api/util.html#util_util_inherits_constructor_superconstructor
[il]: https://github.com/isaacs/inherits/blob/master/inherits_browser.js

bigtimebuddy:
@englercj Okay. Just so I'm clear, you're suggesting simply using Node's built-in `util.inherits` and removing `PIXI.utils.extend` and ignore the former sites where only the constructor was being set? 

<hr>
I'm happy to retool this PR, but it doesn't _exactly_ solve the initial problem we had of being able to minimally express inheritance in the Flash output that we're hoping to generate. 

Just for reference, this is how CreateJS expresses inheritance in it's Flash output, which is problematic for a number of reasons:
```js
(lib.MyContainer = function() {
	this.initialize();
}).prototype = p = new cjs.Container();
```
We were hoping to do something a little more cleanly (which doesn't have to live at the Pixi-level). We are planning on creating `PIXI.flash` as a runtime (a la PIXI.spine) and because this is specific to the Flash output, I can move the static `extend` methods there.
```js
lib.MyContainer = Container.extend(function() {
    Container.call(this);
});
```
I welcome your suggestions, thoughts.

drkibitz:
@bigtimebuddy I'm sorry, but I disagree with almost everything in that comment. There is no *need* for anything here except convenience. I have preferred PIXI for more than just one reason but rather a combination, of course for performance, but also clarity and the speed of innovation. But specifically over CreateJS because it always seemed like innovation in EaselJS and family were blocked by their support for Adobe product export, (e.g. their sponsors). Adobe themselves has some of the worst JavaScript exported from their own products, it is the worst to try and include in your app or library. One thing being a utility library to make JavaScript look like ActionScript. For the 17th time, lets stop trying to do this please. ActionScript3 was great in theory, but it's not JavaScript, and if there is any possible way not to encourage and propagate the abuse of classical inheritance the way ActionScript does, lets go that route. To be honest, the explanation that these changes are brought on because of supporting some sort of flash export, this just makes me believe that PIXI will eventually go the way of EaselJS for me.

@englercj Many good points, and I actually thought about mentioning `util.inherits`, but even this is unnecessary. So I can't be ok with that and not with this PR. I'm still a solid :-1: on this. I've been on the side of PIXI modularity since I don't know when. Adding a util (or PIXI.utils) dependency on nearly every class in the library is just unnecessary, especially since it's not there now. I prefer vanilla JS, yes in my own development which is not important, but especially in OpenSource. It encourages good development practices, which in turn attracts good contributors to the project. This trumps any other *need* that I could think of.

But to be fair, my opinion probably doesn't matter that much. I haven't contributed nearly as much as I probably should have with code, but I feel like sharing my opinions here and there is a form of contribution regardless how people take them. To be clear, I've been using JavaScript for over 16 years, and ActionScript over 8 (as2 and 3). I've lived the evolution of Flash, and the reinventions of JS. Hopefully my experience has generated some useful opinions, and not things that have sounded like hollow rants here and there. I'm going to stand strong on this opinion, but don't feel bad if you simply don't agree or don't care. It's just JavaScript anyway :wink:.

bigtimebuddy:
@drkibitz Whoa, really hit a nerve there with the word _Flash_. I'm fine to have an discussion off this PR thread about that topic and our aims to build better tooling for Pixi. But thanks for sharing your thoughts and I get where you're coming from--our backgrounds are very similar.

@englercj Let me know how you'd like me to proceed with this PR.

englercj:
I'm still with @drkibitz on this one, I haven't seen anything here as to why we **need** this for PIXI.flash, or what it has to do with it at all?

What we do now is idiomatic JS, why does that need to change? Why does the flash plugin require a different inheritance mechanism? 

@drkibitz made a really good point here:

> Adding a util (or PIXI.utils) dependency on nearly every class in the library is just unnecessary, especially since it's not there now.

That is a great point, why do we need to introduce a new dependency (and new code, etc) for this?

bigtimebuddy:
Okay I agree, I think this is superfluous for what the original intention was. I will close the PR. 

The main problem we're solving for is JavaScript bloat. With auto-generate JavaScript (as you can see with CreateJS) deeply nested, complex and enormous display lists (like creating long animations), produce unreasonably large files with many library items that require inheritance (not just instantiation).

We have considered other formats to represent timeline animations (JSON, binary formats, etc), but the consensus from all though working on this project was that the code should 1) be easily readable, 2) be as vanilla Pixi as possible, and 3) be easy for coding and non-coding people to modify. The implication of auto-generating those few lines of prototype-Object.create inheritance, while more pure-JS, will mean huge output file sizes (~10-20% larger). 

I don't expect you to agree with this approach, but we will probably still find someway to abstract the prototype chain inheritance for brevity's sake.

drkibitz:
@bigtimebuddy @englercj Thanks, I appreciate the civilized debate on this one. @bigtimebuddy If it is specifically to reduce bloat for autogenerated code, I would be a lot less against utility methods like this if they remain specific to their needs, e.g. a PIXI.flash.extends or PIXI.flash.inherits sounds a lot more reasonable to me, as it would not extend beyond those boundaries, but hopefully help in the way the purpose of this originally intended.

englercj:
Yeah using a method like that only inside of autogenerated code, only for that plugin seems extremely reasonable.

I'd be careful saying it would be 10-20% larger with those lines. They are really more like 1-5% of the filesize of an average pixi file. I feel like this change won't be as big of an impact as you think it will.

For example, measuring the effect this PR has on pixi.js:

```
// pure JS prototype extension
pixi.js        810,381 bytes
pixi.min.js    286,652 bytes

// using this PR for a util method
pixi.js        805,293 bytes   (0.628% file size reduction)
pixi.min.js    283,780 bytes   (1.000% file size reduction)
```

The change isn't all that noticable. I would expect you to see similar results in your generated code. Also I would make the argument that using the same extension method (that is pure prototype extension) that pixi.js itself uses makes it consistent with what people who have been using pixi.js are familiar with!

PR #2176, opened 2015-10-30
inline Sprite.destroy logic with Text.destroy

shabanovd:
When Sprite.destroy called from Container.destroy 2nd parameter is undefined and should be interpreted as true.
Opposite solution to add destroyBaseTexture parameter to all destroy functions.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -522,7 +522,7 @@ Sprite.prototype.destroy = function (destroyTexture, destroyBaseTexture)
 
     if (destroyTexture)
     {
-        this._texture.destroy(destroyBaseTexture);
+        this._texture.destroy(destroyBaseTexture === undefined ? true : destroyBaseTexture);
     }
 
     this._texture = null;


src/core/sprites/Sprite.js
```diff
@@ -522,7 +522,7 @@ Sprite.prototype.destroy = function (destroyTexture, destroyBaseTexture)
 
     if (destroyTexture)
     {
-        this._texture.destroy(destroyBaseTexture);
+        this._texture.destroy(destroyBaseTexture === undefined ? true : destroyBaseTexture);
```
drkibitz: You are changing the default from false to true, so wondering why. Also should change the doc block along with this change.

shabanovd: The problem that memory is not free up when call Container.destroy because there no way to set destroyBaseTexture parameter and by default it treat as false.

There are two options: add second parameter to all destroy function or change destroyBaseTexture default value. While I looking through code found Text.destroy and because it simplest change I decided to do it, but personally I prefer 2nd parameter change here.

How docs should be updated? Should Text.destroy be updated too?

bigtimebuddy:
Hello! Pull requests should be made to the **dev** branch and not the **master** branch. Also, PR #2592 resolve some inconsistencies with the destroy functionality.

PR #2184, opened 2015-11-04
Webgl scissor manager

TannerRogalsky:
A Scissor Mask for the WebGLRenderer.

API:
```javascript
displayObject.mask = new PIXI.Rectangle(0, 0, 20, 20);
```

### DIFF
--- a/src/core/renderers/webgl/managers/MaskManager.js
+++ b/src/core/renderers/webgl/managers/MaskManager.js
@@ -1,5 +1,6 @@
 var WebGLManager = require('./WebGLManager'),
-    AlphaMaskFilter = require('../filters/SpriteMaskFilter');
+    AlphaMaskFilter = require('../filters/SpriteMaskFilter'),
+    math =  require('../../../math');
 
 /**
  * @class
@@ -33,6 +34,9 @@ MaskManager.prototype.pushMask = function (target, maskData)
     {
         this.pushSpriteMask(target, maskData);
     }
+    else if (maskData instanceof math.Rectangle) {
+        this.pushScissorMask(target, maskData);
+    }
     else
     {
         this.pushStencilMask(target, maskData);
@@ -51,6 +55,9 @@ MaskManager.prototype.popMask = function (target, maskData)
     if (maskData.texture)
     {
         this.popSpriteMask(target, maskData);
+    }
+    else if (maskData instanceof math.Rectangle) {
+        this.popScissorMask(target, maskData);
     }
     else
     {
@@ -109,5 +116,27 @@ MaskManager.prototype.pushStencilMask = function (target, maskData)
 MaskManager.prototype.popStencilMask = function (target, maskData)
 {
     this.renderer.stencilManager.popMask(maskData);
+};
+
+/**
+ * Applies the Mask and adds it to the current filter stack.
+ *
+ * @param target {PIXI.RenderTarget}
+ * @param maskData {PIXI.Rectangle}
+ */
+MaskManager.prototype.pushScissorMask = function (target, maskData)
+{
+    this.renderer.scissorManager.pushMask(maskData);
+};
+
+/**
+ * Removes the last filter from the filter stack and doesn't return it.
+ *
+ * @param target {PIXI.RenderTarget}
+ * @param maskData {PIXI.Rectangle}
+ */
+MaskManager.prototype.popScissorMask = function (target, maskData)
+{
+    this.renderer.scissorManager.popMask(maskData);
 };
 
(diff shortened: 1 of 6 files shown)


englercj:
Very cool, I'm curious what benefit this gives you over using a Graphics object since you can draw a rectangle there as well.

TannerRogalsky:
@englercj It's just a matter for performance. It's a lot less data to transfer to the GPU if I don't need to bind a buffer and can just use send 4 floats instead. Fewer draw calls and less modified GPU state.

englercj:
@TannerRogalsky Fair enough, I would much rather use the shapes (Polygon/Rectangle/Circle/etc) for masks rather than Graphics. The big benefit of graphics is the easy-to-use drawing APIs to create polygons without knowing vertices. But I may want to make that a v4 thing, rather than mix-and-matching in v3.

@GoodBoyDigital I'll leave it up to you whether to merge this in or not.

TannerRogalsky:
@englercj I can see that. In general I like the API for setting masks using Graphics and I think it makes a lot of sense for setting the stencil buffer but was a little dismayed by the number of draw calls required when I just wanted to mask a simple rectangle.

Maybe it should be noted that I didn't support the CanvasRenderer with this change but that's only because I wanted to keep the PR smaller. I can't think of any technical limitation that should keep this from working with both renderers (although I haven't spent much time with the PIXI Canvas stuff). If this gets merged and there's interest, I would be happy to look into that as well.

englercj:
Yeah you are right, and I think it is a great candidate for v4 :)

GoodBoyDigital:
Hey guys!

Thanks for this PR @TannerRogalsky! I 100% agree that scissor masking is something we need to get into pixi as often a scissor test is way faster for simple rectangles. 

@englercj - I think the cool thing about masks extending display objects is that they can be easily animated / transformed. Using shapes would probably be a little more restrictive.

How about instead of passing in a mat.Rectangle, we keep the API the same and if a user uses a graphics object as a mask we check to see if the mask is:

1: A rectangle shape.
2: It does not have any rotation / skew.

If the above is true then we can automatically use scissor masking instead. Would be pretty neat! 
What do ya think?

TannerRogalsky:
@GoodBoyDigital I thought about that and I think it's a great idea but I wasn't sure if it would be seen as an abuse of the Graphics object. Do you imagine just checking if `graphicsData` is length one and the resembles a rectangular polygon? It might be nice to maintain the `Rectangle` API in addition to that because it's a little more succinct to just create a rectangle directly but I can make the change to check for the Graphics case.

TannerRogalsky:
@GoodBoyDigital Is this something like what you were thinking? 838bc81

GoodBoyDigital:
pretty much! will aim to review this over the weekend, looks good though man! Thanks for your efforts so far @TannerRogalsky :+1:

TannerRogalsky:
My pleasure! :+1:

src/core/renderers/webgl/managers/MaskManager.js
```diff
@@ -52,6 +69,22 @@ MaskManager.prototype.popMask = function (target, maskData)
...
     }
+    else if (maskData.type === CONST.SHAPES.RECT) {
+        this.popScissorMask(target, maskData);
+    }
+    else if (maskData.graphicsData && maskData.graphicsData.length === 1) {
+        var graphicsData = maskData.graphicsData[0];
+        if (graphicsData.type === CONST.SHAPES.RECT && maskData.rotation === 0 &&
+            maskData.skew.x === 0 && maskData.skew.y === 0)
+        {
+            // The graphics object has no rotation or skew and only a single rectangle in it
```
englercj: The check above doesn't guarantee there is only a single rectangle in the Graphics object. Just that the first one is a Rectangle.

TannerRogalsky: @englercj Doesn't the `maskData.graphicsData.length === 1` ensure that there's only one shape?

GoodBoyDigital:
hey there! Looking good @TannerRogalsky ! A  few thoughts for the PR..

- I think it would be best to check the world transform of the object, that way if the rectangle is in a rotated/skewed container it would default to regular masking. Something like the below maybe?

```
var wt = maskData.worldTransform
if( !(wt.b % CONST.PI_2) && wt.c === wt.b)
{
      // only scale and transform....
}

```

- We should also make sure to apply the translation and the scale of the worldTransform to the scissor rect.

- Im guessing we can only push one scissor rect onto the stack? As you can only have one scissor rect active at one time?

Thanks all!

GoodBoyDigital:
New thought! 

How about - Maybe for the first mask in the stack, we could Always use a scissor rect? We could get the bounds of the mask shape, and scissor rect that. Then if the mask is not a valid rect we also use the stencil buffer. Could give us a performance boost for all masks as we only render what we know will be seen?

src/core/renderers/webgl/managers/MaskManager.js
englercj: DERP, missed that. Carry on! :see_no_evil:

GoodBoyDigital:
hello! this feature has landed in v4. Thanks for the original effort @TannerRogalsky :+1:

PR #2194, opened 2015-11-15
Dev accessibility

GoodBoyDigital:
Added an accessibility layer to Pixi enabling displayObjects to be tabbed and read by screenreaders.

All that needs to be added:

```myButton.accessible = true;```

Example here: 

http://www.goodboydigital.com/pixijs/accessibility/accessibility.html

Now Pixi content can be accessed by all!


englercj:
This looks great, but I feel like it should be a plugin you can package with pixi not something we ship to everyone by default.

GoodBoyDigital:
This did cross mind! Initially it was a separate part. But then after thinking about it...

Accessibility is a really important feature to have (as important as interaction) and no other canvas based renderers support it. Means we can make cool stuff and easily make it useable for people with certain disabilities who may not be able to navigate content with a mouse or even see it?

The other thing worth mentioning is that unless someone actually starts using it there is zero overhead.

This is the reason I made it a PR though as thought it was worth discussing :)

englercj:
I see that the overhead is negligible, I'm more concerned with the general idea of "it seems useful so lets ship it by default". One of the biggest things holding pixi back in my opinion is that it is getting pretty bloated with a bunch of stuff people maybe might use in certain situations.

How often is this going to be used? Really it is only useful for UI type interactions, right? As a library with no UI toolkit, what value does this really bring to the core library that it needs to be shipped in every bundle?

Seems like this would be an important feature, for a small subset of projects. But more and more feature slike this make pixi less and less a rendering engine and more and more an application framework. I feel like the core should be focused on rendering and solving that problem. If the problem isn't rendering related it belongs in an external plugin. Interaction and this are both things that are important to an application framework, but not to rendering. They don't belong here IMO.

GoodBoyDigital:
Sone very good points man! Will add it as an extension :+1:

photonstorm:
I completely agree this ought to be an extension, but I completely disagree it's an "edge case". It *ought* to be used in _every_ game, and not just for UI either, but for actual game play.

GoodBoyDigital:
@photonstorm  :+1:

englercj:
>  It ought to be used in every game, and not just for UI either, but for actual game play.

@photonstorm That is an interesting thing to say, I'm curious what it actually means in practice. Consider any generic side scroller, what purpose does the mouse serve besides UI? It may be my ignorance, but I can only think of a few games where mouse interaction is key *and* this would help alleviate the problem of being unable to use a pointer.

photonstorm:
@englercj when you've worked on as many BBC games as I have, you learn to appreciate just how to get accessibility into any sort of game, side scrollers included. From dwell timers, to single and two button switch support, to grid positioning layouts. There are myriad ways, but it all has to start with an extension like this.

englercj:
I guess I just don't have the experience to understand the problem fully, I'll defer to you two then. Thanks for the info.

GoodBoyDigital:
@photonstorm this update is actually for the BBC. Just wanted to make it easier for us and for others to make things accessible too. Then got a bit altruistic started thinking that the easier it is for people to implement the more likely people would use it on projects by default.

photonstorm:
Yeah I figured it was for the BBC :)

Did you update the docs or are they still all 'TODO' ! You ought to make sure you carefully explain that it's generating `<div>`s all over the place, because that will be a nightmare for some site layouts :) (and is going to be interesting to see where they end up in scaled canvases!)

GoodBoyDigital:
Good point man, docs do need updating! But things such as scaling and
various layouts should have all been taken care off. If I missed anything I
will make sure to address before it lands in Master mind  (we are currently
fitting it into a few games to ensure it works smoothly).

On Thu, Nov 19, 2015 at 11:54 AM, Richard Davey <notifications@github.com>
wrote:

> Yeah I figured it was for the BBC :)
>
> Did you update the docs or are they still all 'TODO' ! You ought to make
> sure you carefully explain that it's generating <div>s all over the
> place, because that will be a nightmare for some site layouts :) (and is
> going to be interesting to see where they end up in scaled canvases!)
>
> —
> <https://github.com/pixijs/pixi.js/pull/2194#issuecomment-158034540>.
>



-- 
Mat Groves

*Technical Partner*

Issue #2071, opened 2015-08-28
Rewrite Interaction Manager


englercj:
I think it becoming aparant that the interaction manager is due for an update. There are many bugs and feature requests that need to be taken care of:

https://github.com/pixijs/pixi.js/issues?q=is%3Aopen+is%3Aissue+label%3A%22Interaction+Manager%22

This issue is to track completion of a new interaction manager that fixes all (or at least most) of these issues.

---

I'm pretty busy with work right now, but I have opened a bounty on this issue. If we reach the goal then I will make time to complete this feature. You can post to the bounty on [Bountysource](https://www.bountysource.com/issues/26281649-rewrite-interaction-manager).

tobireif:
Just in case it's of potential use for the InteractionManager incl. Pointer Events incl. "pointerup":

Perhaps one of these libs could be used / forked / built on / looked at?

https://github.com/Deltakosh/handjs
https://twitter.com/Inateno/status/641677865038315520

https://github.com/Rapid-Application-Development-JS/Pointer

drkibitz:
Where's this work being done? Fork, feature branch, privately?

englercj:
It isn't being done, we haven't reached the goal I set or gotten close enough for me to have the time to start :)

cvissie:
Also another thing we might need to add to this.

enabled=false

Our project has HTML interaction capabilities as well as PIXI interaction capabilities.

Sometimes I would like to disable PIXI interaction and re-enable it later.

The way I have implemented it is by simply adding a this.enabled=true field in InteractionManager and then checking for it in the update method as well as the processInteractive method.

PR #2195, opened 2015-11-18
Add pointermove, pointerdown, pointerout, and pointerup events to the…

stephenweaver:
… interaction manager

To enable multi-touch to work in IE 11 and other browsers that pointer events.

From pointermove, pointerdown, pointerout, and pointerup browser events; the following displayObject events are available: 

pointerdown, pointerup, pointerupoutside, pointerover, pointerout, pointercommanddown, pointercommandup, pointercommandtap, pointertap, pointercommandupoutside, pointerupoutside

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -54,6 +54,13 @@ function InteractionManager(renderer, options)
     this.mouse = new InteractionData();
 
     /**
+     * The pointer data
+     *
+     * @member {PIXI.interaction.InteractionData}
+     */
+    this.pointer = new InteractionData();
+
+    /**
      * An event data object to handle all the event tracking/dispatching
      *
      * @member {object}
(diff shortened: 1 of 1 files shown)


drkibitz:
I'm a fan of the pointer event implementation, but why maintain 2 virtual objects, 1 for the mouse and 1 for the pointer? Wouldn't it makes sense to abstract everything to pointer events, e.g. mouse events and touch events?

drkibitz:
Also you should take note of this issue https://github.com/pixijs/pixi.js/issues/2071 by @englercj 
I for one would think a rewrite on pointer events would be one option.

But I am also biased toward a proposal I made a year or two ago where pixi does not attempt to maintain any event listeners whatsoever. It should simply contain helper methods to transform and project screen/page/client/element coordinates to pixi coordinates, and apply a hittest to the pixi object tree. The hittest can be abstracted further to apply a particular virtual pixi event to the tree. In this case, you would still need the appropriate event abstraction to go with, and like I said, I'm a fan of pointer events ;)

drkibitz:
Think of something like this:
```js
var point = PIXI.interaction.pointFromDomEvent(domEvent, containerObject); // containerObject is pixi scope
PIXI.interaction.applyEventAtPoint("pointerdown", point, containerObject); // containerObject is pixi scope
```
Ok probably not the best place ot post these comments, more appropriate on #2071

stephenweaver:
Good feedback. While it would be good to abstract everything to pointer events as suggested, I see this solution as a intermediate implementation until issue #2071 is done. I also see your point that it would be best for PIXI to not handle the events. Initially i didn't want to modify the PIXI source code, but it turned out to be cleaner than duplicating/implementing my own hit testing code.

stephenweaver:
Could this be merged in?

englercj:
@GoodBoyDigital Can you take a look at this? I don't have a touch device to test with.

stephenweaver:
Is there an eta on this? If it's going to be while, I'll just use my forked repo; but I'd rather not if I don't have to.

englercj:
Closing in favor of #2476.

PR #2199, opened 2015-11-19
renderTexture canvas fallback

ivanpopelyshev:
Issue 2197: renderTexture canvas fallback bug


GoodBoyDigital:
nice one @ivanpopelyshev you are on a roll! :+1:

PR #2203, opened 2015-11-20
Changed cacheAsBitmap help to instruct cache to be set to false instead of null

Excaa:


GoodBoyDigital:
Nice one @Excaa :+1:

englercj:
Thanks!

PR #2204, opened 2015-11-21
replaced native Array.splice to reduce gc churn

mreinstein:
Javascripts' built-in `Array.splice` is pretty slow, and generates garbage because it returns a newly created array containing the removed objects.

I've only replaced `Array.splice` where 1 item is removed. That seems to be the most common code path. Might be able to use this replacement function for removing multiple indices but I haven't looked into it.


englercj:
Wow: http://jsperf.com/manual-splice

That is very interesting, thanks for bringing this up! It looks good to me. The only problem is that you committed the built files! Would you mind removing those and force pushing this branch?

mreinstein:
> Would you mind removing those and force pushing this branch?

@englercj sure, I'll do it this afternoon. There are also a few places where native Array.splice is still used.

mreinstein:
see #2205 instead

PR #2205, opened 2015-11-21
replaced native Array.splice to reduce gc churn

mreinstein:
Javascripts' built-in `Array.splice` is pretty slow, and generates garbage because it returns a newly created array containing the removed objects.

I've only replaced `Array.splice` where 1 item is removed. That seems to be the most common code path. Might be able to use this replacement function for removing multiple indices but I haven't looked into it.

### DIFF
--- a/src/accessibility/AccessibilityManager.js
+++ b/src/accessibility/AccessibilityManager.js
@@ -201,7 +201,7 @@ AccessibilityManager.prototype.update = function()
 		{
 			child._accessibleActive = false;
 
-			this.children.splice(i, 1);
+			core.utils.spliceOne(this.children, i);
 			this.div.removeChild( child._accessibleDiv );
 			this.pool.push(child._accessibleDiv);
 			child._accessibleDiv = null;
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -1,4 +1,5 @@
 var math = require('../math'),
+    utils = require('../utils'),
     DisplayObject = require('./DisplayObject'),
     RenderTexture = require('../textures/RenderTexture'),
     _tempMatrix = new math.Matrix();
@@ -233,8 +234,8 @@ Container.prototype.setChildIndex = function (child, index)
 
     var currentIndex = this.getChildIndex(child);
 
-    this.children.splice(currentIndex, 1); //remove from old position
-    this.children.splice(index, 0, child); //add at new position
+    utils.spliceOne(this.children, currentIndex); // remove from old position
+    this.children.splice(index, 0, child); // add at new position
     this.onChildrenChange(index);
 };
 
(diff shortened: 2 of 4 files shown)


adelciotto:
This is nice! :+1: 

Perhaps for removing multiple indices, something like this could be used:

```js
function splice(arr, startIdx, deleteCount) {
    if (startIdx >= arr.length || startIdx+deleteCount >= arr.length) {
        return;
    }

    for (var i = startIdx, len = arr.length-deleteCount; i < len; ++i) {
        arr[i] = (i > start+deleteCount ? arr[i + 1] : arr[i + deleteCount]);
    }

    arr.length = len;
}
```

```js
array = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
splice(array, 5, 3) // => [0, 1, 2, 3, 4, 8, 9, 10]
```

mreinstein:
@adelciotto looks good to me on first glance. Care to submit a few unit tests for this?

mreinstein:
`removeItem` or `removeElement` might be a better name choice than `splice` or `spliceOne`, because splice implies it offers the ability to add and remove elements in one call.

adelciotto:
good point @mreinstein. Should i just submit a PR with this new function in ```utils```, along with some unit tests ?

englercj:
@adelciotto Can you change `l` to `len` or `length`? It took me a while to realize it was an `l` and not a `1`.

mreinstein:
@adelciotto sounds good! 

one thing to be aware of; if you grep the pixi codebase you'll find `splice` is used in at least 1 place to add a new element to the array, and at least 1 place where the return result of `splice` is actually used. So you'll want to keep those as-is.

mreinstein:
I'd vote for `removeItems` as it's slightly shorter than `removeElements` :)

mreinstein:
somewhat on a tangent, this is the blog post that inspired me to move away from native `splice`. It has some other interesting ideas around efficient array handling. Might be worth a read:

https://gamealchemist.wordpress.com/2013/05/01/lets-get-those-javascript-arrays-to-work-fast/

englercj:
@mreinstein Thanks for starting this conversation and for sharing the blog post. I'm going to be looking hard at things like this for v4!

grovesNL:
While we're on this topic, it may be worthwhile to look at certain methods within [fast.js](https://github.com/codemix/fast.js) for potential speed improvements.

mreinstein:
> I'm going to be looking hard at things like this for v4!

@englercj is there a tentative roadmap for v4 in the works? Curious what you're thinking of right now. v3 emphasized modularization so I'm surprised that a major version number is planned. The biggest thing I can foresee needing is lighting/shadows. Above all else that's what is the most painful for me is not having lighting primitives.


@grovesNL fast looks interesting. One thing that I'd like to see is fast trig functions. The current pixi.js based code I'm working on is a particle engine, and it seems like Javascript's built-in `sin`, `cos` and other functions are the bottleneck. It would be great to tradeoff *some* accuracy for speed.

I have not at all tested or verified this fast trig gist, but on first inspection it looks correct:

https://gist.github.com/imbcmdth/6338194

englercj:
@mreinstein I'll publish something soon, I have some goals and things right now but I want to flesh it out a bit more.

mreinstein:
closing this in favor of #2209

PR #2208, opened 2015-11-24
Fix InteractionData.getLocalPosition populate existing point

twhitbeck:
Includes test

### DIFF
--- a/src/interaction/InteractionData.js
+++ b/src/interaction/InteractionData.js
@@ -38,10 +38,10 @@ module.exports = InteractionData;
  *
  * @param displayObject {PIXI.DisplayObject} The DisplayObject that you would like the local coords off
  * @param [point] {PIXI.Point} A Point object in which to store the value, optional (otherwise will create a new point)
- * param [globalPos] {PIXI.Point} A Point object containing your custom global coords, optional (otherwise will use the current global coords)
+ * @param [globalPos] {PIXI.Point} A Point object containing your custom global coords, optional (otherwise will use the current global coords)
  * @return {PIXI.Point} A point containing the coordinates of the InteractionData position relative to the DisplayObject
  */
 InteractionData.prototype.getLocalPosition = function (displayObject, point, globalPos)
 {
-    return displayObject.toLocal(globalPos ? globalPos : this.global, point);
+    return displayObject.toLocal(globalPos || this.global, null, point);
 };
--- a/test/unit/interaction/InteractionData.test.js
+++ b/test/unit/interaction/InteractionData.test.js
@@ -1,0 +1,16 @@
+describe('PIXI.interaction.InteractionData', function() {
+    describe('getLocalPosition', function() {
+        it('should populate second parameter with result', function() {
+            var data = new PIXI.interaction.InteractionData(),
+                displayObject = new PIXI.DisplayObject(),
+                point = new PIXI.Point();
+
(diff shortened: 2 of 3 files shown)


test/unit/interaction/InteractionData.test.js
```diff
@@ -0,0 +1,16 @@
...
+            var data = new PIXI.interaction.InteractionData(),
+                displayObject = new PIXI.DisplayObject(),
+                point = new PIXI.Point();
+
+            data.getLocalPosition(displayObject, point);
+
+            expect(point).to.exist;
+            expect(point).to.be.an.instanceof(PIXI.Point);
+            expect(point.x).to.equal(0);
+            expect(point.y).to.equal(0);
```
englercj: This is only the most basic case, I would add one that actually requires calculation so you can ensure it is correct.

src/interaction/InteractionData.js
```diff
@@ -38,10 +38,10 @@ module.exports = InteractionData;
...
  * @param displayObject {PIXI.DisplayObject} The DisplayObject that you would like the local coords off
  * @param [point] {PIXI.Point} A Point object in which to store the value, optional (otherwise will create a new point)
- * param [globalPos] {PIXI.Point} A Point object containing your custom global coords, optional (otherwise will use the current global coords)
+ * @param [globalPos] {PIXI.Point} A Point object containing your custom global coords, optional (otherwise will use the current global coords)
  * @return {PIXI.Point} A point containing the coordinates of the InteractionData position relative to the DisplayObject
  */
 InteractionData.prototype.getLocalPosition = function (displayObject, point, globalPos)
 {
-    return displayObject.toLocal(globalPos ? globalPos : this.global, point);
+    return displayObject.toLocal(globalPos || this.global, null, point);
```
englercj: This is a change in functionality, is that on purpose?

twhitbeck:
Perhaps I should have included more information in my pull request. The issue that this pull request fixes is that previously passing a `Point` as the second parameter to `getLocalPosition` did nothing.

src/interaction/InteractionData.js
twhitbeck: Yes, this was part of the fix.
Correct me if I'm wrong, but `globalPos ? globalPos : this.global` and `globalPos || this.global` are semantically equivalent, right?

test/unit/interaction/InteractionData.test.js
twhitbeck: This test only checks the thing which was broken and is fixed by the pull request. I suppose lots more tests could be added, but I'd add them with other pull requests/commits

twhitbeck: nvm, test needs to be rewritten

src/interaction/InteractionData.js
englercj: Yes, but I was talking about the fact that now you pass `null` for the `from` parameter instead of `point`.

test/unit/interaction/InteractionData.test.js
englercj: Yeah this doesn't actually test if it is set, because the default is zero and nothing would change it from zero.

twhitbeck:
Ok, the test actually tests this now.

englercj:
Looks good, unfortunately the PR is to the `master` branch so I will have to merge to `dev` manually when I get home.

PR #2211, opened 2015-11-25
Remove texture from cache if it will be added by fromImage

### LINKED ISSUE #1741: delete from texture cache should be in texture
Right now only destroying a base texture will remove it from the cache:

https://github.com/GoodBoyDigital/pixi.js/blob/dev/src/core/textures/BaseTexture.js#L319

But that can cause problems if you destroy a texture, then try to create one again with `.fromImage()` you will get a broken texture back out of the cache.

Texture should remove itself from the TextureCache and not rely on BaseTexture to do it.

zont:
Problem:
```javascript
var IMAGE_URL = 'any-image.png';

// create any container with sprite, use... then destroy
var sprite = new PIXI.Sprite.fromImage(IMAGE_URL); 

container.addChild(sprite);
container.destroy(true);

// fail to create sprite from image with url IMAGE_URL
var sprite2 = new PIXI.Sprite.fromImage(IMAGE_URL); 
// Uncaught TypeError: Cannot read property 'hasLoaded' of null
```

fixed #1741

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -317,7 +317,6 @@ BaseTexture.prototype.destroy = function ()
     if (this.imageUrl)
     {
         delete utils.BaseTextureCache[this.imageUrl];
-        delete utils.TextureCache[this.imageUrl];
 
         this.imageUrl = null;
 
--- a/src/core/textures/Texture.js
+++ b/src/core/textures/Texture.js
@@ -250,6 +250,11 @@ Texture.prototype.destroy = function (destroyBase)
 {
     if (this.baseTexture)
     {
+        if (this.baseTexture.imageUrl)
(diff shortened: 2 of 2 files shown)


englercj:
Good catch, with this change the removal in BaseTexture seems superfluous:

https://github.com/pixijs/pixi.js/blob/master/src/core/textures/BaseTexture.js#L320

May want to just remove that one too.

jrowny:
Just ran into this issue, I think this fix is good enough to commit.

englercj:
Unfortunately this PR is to the `master` branch not the `dev` branch and needs to be merged manually or the PR redone.

GoodBoyDigital:
hello! I f you could resubmit this for the dev branch that would be ace. Thanks!

PR #2214, opened 2015-11-30
fix(InteractionData): populate second param of getLocalPosition

twhitbeck:
Supercedes #2208


twhitbeck:
Here it is against `dev`

englercj:
:+1:

PR #2222, opened 2015-12-04
blendMode can be changed when filter applies.

ivanpopelyshev:
blendMode can be changed when filter applies. No changes in current filters are required.


ivanpopelyshev:
However, demo from https://github.com/pixijs/pixi.js/issues/1598 wont work on that, because Multiply operation is not additive. Those small squares in intersections are black if you dont apply filter:

Alt text

ivanpopelyshev:
how it works:

```javascript
container.filters = [bloomFilter];
container.filterBlendMode = PIXI.BLEND_MODES.ADD;

ivanpopelyshev:
That patch has solved this case: http://www.html5gamedevs.com/topic/19029-cant-set-blendmode-with-filters/

ivanpopelyshev:
By the way, PIXI.Container doesnt have blendMode on its own , so filterBlendMode can be good solution for two other cases:

* blendModes that arent supported by glBlendFunc

* use of globalCompositionMode in canvas fallback on whole container

englercj:
This PR can't be merged for a couple reasons:

1) It creates new properties on the object during runtime, which destroys the hidden class optimizations. Any property on the object must be created at constructor time.
2) I don't think this is the right way to do this, You have render target access in the apply methods of filters, I wonder if it makes sense for the filter to be applying a blend mode there.

ivanpopelyshev:
Help me with 1, I know you can change PR before you apply it.

As for 2, I dont want to change other filters, and for example, BlurFilter does apply two times.

I thought about that stuff for 5 days already, and I dont see other solutions. BlendMode is using data from renderTarget  , its not a filter, its a special case.

Also, to implement all standart blendModes we'll have to use "readPixels", we cant avoid that.

GoodBoyDigital:
what do you guys think about putting the filter blend mode property as a property on the filter as it feels like its more of a filter specific thing?

ivanpopelyshev:
It makes sence only for last filter in the list, and we have to implement that thing inside FilterManager, so filters will be able to call that special method which is taking care of blendMode.

PR #2225, opened 2015-12-07
IE fix, see: https://github.com/pixijs/pixi.js/issues/1930

jberube:
Signed-off-by: Julien Bérubé <julienb@playtika.com>


jberube:
Sorry, that wasn't mean to be submited, yet.

PR #2226, opened 2015-12-08
try optimize updateTransform

zzm2q:
PS: This pull request is just for discussion because the code is incomplete.

`updateTransform` is called each frame, i found this would cost a lot of time with the following profiling:

In simple terms, the worldTransform of DisplayObject should not be update if it's local transform(position, scale, rotation, anchor ...etc) and the worldTransform of it's parent wasn't changed. This could be done with a flag `transformDirty` in calling chain, review the code of this pull request.

After optimized:

both tow profiling use the Container example of `pixijs/example`

### DIFF
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -140,6 +140,8 @@ function DisplayObject()
      * @private
      */
     this._mask = null;
+
+    this._transformDirty = true;
 }
 
 // constructor
@@ -148,6 +150,16 @@ DisplayObject.prototype.constructor = DisplayObject;
 module.exports = DisplayObject;
 
 Object.defineProperties(DisplayObject.prototype, {
+    transformDirty: {
+        get: function()
+        {
+            return this._transformDirty;
+        },
+        set: function(value)
+        {
+            this._transformDirty = value;
+        }
+    },
     /**
      * The position of the displayObject on the x axis relative to the local coordinates of the parent.
      *
@@ -161,6 +173,8 @@ Object.defineProperties(DisplayObject.prototype, {
         },
         set: function (value)
         {
+            if(value === this.position.x) return;
+            this.transformDirty = true;
             this.position.x = value;
         }
     },
@@ -178,6 +192,8 @@ Object.defineProperties(DisplayObject.prototype, {
         },
         set: function (value)
         {
+            if(value === this.position.y) return;
+            this.transformDirty = true;
             this.position.y = value;
         }
     },
--- a/src/extras/index.js
+++ b/src/extras/index.js
@@ -8,6 +8,7 @@
 require('./cacheAsBitmap');
 require('./getChildByName');
 require('./getGlobalPosition');
+require('./optimizeUpdateTransform');
 
 /**
  * @namespace PIXI.extras
--- a/src/extras/optimizeUpdateTransform.js
+++ b/src/extras/optimizeUpdateTransform.js
@@ -1,0 +1,30 @@
+var core = require('../core'),
+    DisplayObject = core.DisplayObject,
+    Container = core.Container;
+
+DisplayObject.prototype.updateTransform = function (transformDirty)
+{
+    transformDirty = transformDirty || this.transformDirty || false;
+    if(transformDirty) {
+        this.displayObjectUpdateTransform();
+        this.transformDirty = false;
+    }
+};
+
+Container.prototype.updateTransform = function(transformDirty) {
+    if (!this.visible)
+    {
+        return;
+    }
+
+    transformDirty = transformDirty || this.transformDirty || false;
+    if(transformDirty) {
+        this.displayObjectUpdateTransform();
+        this.transformDirty = false;
+    }
+
+    for (var i = 0, j = this.children.length; i < j; ++i)
+    {
+        this.children[i].updateTransform(transformDirty);
+    }
+};


ivanpopelyshev:
That will cause too many problems. For example, pixi-spine updateTransform also calls "update()" function.

Default container behaviour is "call updateTransform every frame for visible objects".

There's ParticleContainer that optimizes that doesnt call its children updateTransform(), its optimal for large quantities of sprites in it.

There's pixi-spine gpupatch that does the same but for spine animations with meshes and stuff.

If you want to use dirtyflag, make your own container and use it.

zzm2q:
I don't think this will cause so many problem, every subclass could override `updateTransform` if put dirty flag check in core module(DisplayObject, Container), and these subclasses(for example the Spine class of pixi-spine)  can ignore dirty flag and implements it's logic by itself.

Let's think this again:

```
the worldTransform of DisplayObject should not be update if it's local transform(position, scale, rotation, anchor ...etc) and the worldTransform of it's parent wasn't changed. 
```

why?  In the upper case, whatever you don't call `updateTransform`, the values of the worldTransform will not be changed.

zzm2q:
There are a lot objects is static in a game, it's worth to do dirty check.

englercj:
I think this is something worth investigating. It is definitely a backwards incompatible change to not call updateTransform every frame, but I like the idea for v4.

ivanpopelyshev:
I think that's good idea to use multiple container types in scene: PIXI.Container, PIXI.ParticleContainer, and this one can be called PIXI.DirtyFlagContainer

ivanpopelyshev:
Also, this approach makes impossible some things, like using one position per multiple objects

zzm2q:
well, it's not bad to use DirtyFlagContainer, but it will not apply to those objects extends from DisplayObject such as Sprite, Text...etc

ivanpopelyshev:
Store all previous position of childrens in that container, like angular.js does with watches. Performance will be the same, code will be much cleaner, you wont have to take into account that stuff while writing your own object class.

GoodBoyDigital:
Hey chaps!

This PR is definitely a great discussion point! I have been thinking about quite a bit recently too. 

The current pixi set up assumes things are always moving. this keeps things nice and simple and works really well for the majority of the fast dynamic content people tend to create.

For elements that do not often move I tend to use cacheAsBitmap as it can go a long way to simplify the rendering of static objects.

I am quite keen on the idea of exploring a special type of 'static' mode for pixi objects that we disable by default as I am pretty sure we can have our cake and eat it too for this situation.

ivanpopelyshev:
@GoodBoyDigital 

What do you think about "If you want to change behaviour of standart elements - do it in container" policy? ParticleContainer is a good example of that.

GoodBoyDigital:
yes, that does sound like a really nice idea! We could make a new 'Static' Container. This hi jacks the updateTransform for all the children, and caches as much detail as possible! (for webGL we can even upload a the transform of the static object, so it can move without having to refresh all the children matrix's!)

ivanpopelyshev:
@GoodBoyDigital that's exactly how ParticleContainer works, right?

GoodBoyDigital:
yesir!

ivanpopelyshev:
That wont make canvas fallback faster, because drawImage takes much more time than this calculations.

As for WebGL, particleContainer is already doing good job on that. And you know that Im working on special container for tilemaps and spine animations.

grovesNL:
In a Pixi project I am working on, everything is static but still interactive (`cacheAsBitmap` is not very useful here). For example, dragging a Sprite may cause a Graphics object to be redrawn from scratch because of the Sprite's new position

 I am doing a similar optimization during the render call (a dirty flag decides whether the frame is rendered), but I am definitely interested in optimizations within the library.

zzm2q:
ParticleContainer works well with a lots of dynamic objects on WebGL mode.
`static/dirty` Container override updateTransform of it's children, this's also a good way for v3.
My only question is whether changing an object's method affects hidden class or not. 
(Sorry, I have got carried away)

In general, I want to find a way to optimize these few moving objects.

dcodeIO:
Came here through profiling.

10/10

GoodBoyDigital:
'static/dirty' <- this will be in v4, performance gains are quite impressive! :+1:

dcodeIO:
Appreciated! Is it already sane to give the v4 branch a try, for development I mean?

GoodBoyDigital:
na - its under heavy development at the mo. I would wait a little :)

ivanpopelyshev:
@GoodBoyDigital please dont change default behaviour of the container.

GoodBoyDigital:
@ivanpopelyshev - don't worry man - its all pretty much the same!

GoodBoyDigital:
aiming to keep core API consistent

ivanpopelyshev:
@GoodBoyDigital I have enough time now to review your code and make examples, if you need help :)

GoodBoyDigital:
I certainly will soon! I will keep you all posted on that for sure though - The new WebGL core stuff is nearly ready for review.

ivanpopelyshev:
We'll need bunnymark as first example.

ivanpopelyshev:
By the way, the most awesome thing that can be done in Container easily - is to add variables like "overrideRotation" and "overrideScale" which will override parent rotation and scale if set. That will be VERY handy, its easy and its really a problem. The only problem is that you wont be able to use Identity matrix in getLocalBounds(), you have to preserve the scale.

dcodeIO:
One more thought: Would it be wise to check if the value actually differs, let's say [here](https://github.com/pixijs/pixi.js/blob/dev-v4.0.0/src/core/display/DisplayObject.js#L236) etc.?

```js
var epsilonFloat = 1.0 / (1 << 23)

function approximatelyEqual(a, b) {
    if (a === b)
        return true
    return Math.abs(a - b) < epsilonFloat
}
```

ivanpopelyshev:
Can you please NOT to add this thing to the core?

zzm2q:
@ivanpopelyshev, could you give a example project for me? i really want to know your cases. This behavior looks very transparent to outside of core.

And, so if we can cache the quad info of Sprite if not dirty?

englercj:
@ivanpopelyshev I'm confused by your resistance to this change, if we optimize updateTransform to not run every frame, how does that prevent you from doing pre-frame updates on your own without patching into updateTransform (which was always a hack that was recommended against anyway).

ivanpopelyshev:
@englercj I bet that calculating if there's difference or not will take almost the same time as updateTransform() itself. Also if I add special parameters like "overrideScale" I'll have to change both updateTransform and diff calculation. That's not something that new users will understand. Also, I worked with angular.js too much, and I use diffs in bomber for synchronization, so I understand that they will come with some bugs and some new invariants that we'll have to satisfy. Current pixi behaviour is easy to understand, but with diffs it will be a pain to explain.

ivanpopelyshev:
Actually, can anyone look how threejs handles these?

photonstorm:
three.js uses a property called `matrixAutoUpdate` which controls if the transform matrix is updated each frame or not (and if not you have to do it yourself). It also uses two matrices, one local and one world, and keeps track of when the _world_ matrix needs updating in a boolean.

Personally, for all its hyperbole, I think Famous has a really nice Transform implementation. It's fully 3D, so overkill for Pixi, but has anchor points for rotation, scale and origin, a properly split between local and world matrices, booleans to detect if the position, rotation or scale has changed since the last update and a nice feature called `breakpoint` which can be set on any transform and allows that transform to then override any parent influence.

GoodBoyDigital:
That sounds pretty cool! Currently i have refactored pixi so it now has one world and one local and it caches both. perf wise I think this will be quite nice..

My question would be, whats the best way to invalidate the transform and flag it as dirty? The options as I see it are:

1: Change API so we have x, y, scaleX, scaleY, skewX, skeyY etc.. on the displayObject directly. Each of those properties will be a setter / getter that when changed flags the matrix as dirty.

2: Keep the API and put the getters and setters on the points them selves (position, scale, skew)

3: Use three.js' ```matrixAutoupdate``` style property to move the responsibility over to the user.

For what its worth I am leaning towards option two as it means no API change and we keep the ``set`` function which I think is really useful.

What do you guys think?

photonstorm:
I assume the problem is that in option 2 the `Point` objects have no relationship to the transform, yes? I.e. they are just purely un-observed children of the `displayObject`?

GoodBoyDigital:
yes, option two means I would need to make the point observable something like this:

```
var CachePoint = function(doc){
    
    this.doc = doc;

    this.x = 0;
    this.y = 0;
}

Object.defineProperties(CachePoint.prototype, {
    x: {
        get: function ()
        {
            return this.doc._x;
        },
        set: function (value)
        {
            this.doc.position._x = value;
            this.doc.dirtyTransform = true;
        }
    },

    y: {
        get: function ()
        {
            return this.doc.position._y;
        },
        set: function (value)
        {
            this.doc.position.y = value;
            this.doc.dirtyTransform = true;
        }
    }
});
```

This would mean people could switch to v4 without any major refactoring..

photonstorm:
MelonJS does the same as this too - it uses a `Vec2 Observable` data type, which is basically the same as a Pixi Point just with a parent reference. I would suggest you do something similar (`Observable Point`) and retain `Point` as well, then the API doesn't change, but equally devs using `Point` everywhere don't suddenly get `dirtyTransform` baggage. I guess that is what `CachePoint` above means? Although the name is slightly misleading imho, unless they are truly cached somewhere? :)

ivanpopelyshev:
Lets move everything into displayObject.transform, and make fallback getters/setters in DisplayObject.

GoodBoyDigital:
sounds like a plan! Will look at now. thanks chaps!

ivanpopelyshev:
My proposition:

1. Use Vec2 instead of Point
2. Do not use observable Points, because sprite.position = sprite2.position is VERY USEFUL THING
3. Move all transform into "sprite.transform", that way it will be possible to add/remove fields and use custom transformations. "position, scale and rotation" getters/setters in displayobject can use fields from that transform.
4. Add position/scale/rotation overrides, also skewX and skewY to special advanced transform.
5. if transform.autoUpdate is false, then dont update matrices of container and its children, let the user handle that.

If we do that changes will be minimal, no additional ifs will slow down the game, old code will work just fine, and it will be possible to use custom set of fields in transforms.

photonstorm:
1. I agree it ought to be a Vec2 instead of a Point too, but changing it won't add any benefits - as you'd have to retain Point for API compatibility you're effectively just adding to the bundle size (a tiny bit).

2. Bit of an edge-case (and a very poor design imho), but also true. Having an actual Transform object would be much more sensible.

3/4/5. Agreed, this would be ideal situation. Although I fail to see how 3 new overrides and a boolean `autoUpdate` is going to avoid adding new ifs into the flow ;) It's also API breaking for sure, but then if v4 can't do that, what can?

GoodBoyDigital:
hey chaps! I have updated the v4.0.0 https://github.com/pixijs/pixi.js/tree/dev-v4.0.0 
A shiny new Transform class now exists. there are shortcuts to the points on the displayObject itself, but they are to maintain the current API more than anything else.

There is also a class called TransformStatic which uses observable points and only updates if the user changes a property.

Its certainly very cool having a transform object as it will let people create there own custom ones (such as a transform object featuring 3,4,5 from the list) 

Whilst i'm cool with flipping to using vec2s or gl-matrix, I figured that I would rather finish the pixi structure refactor first and then do a conversion sweep (If we feel its necessary.)

Let me know what ya think!

ivanpopelyshev:
May be its better to make getters/setters for position and scale? Its not possible to use different transform class at this point without changing all that fields too, right?

GoodBoyDigital:
Yo! Do you mean getter setters on the displayobject?

On Tuesday, 19 January 2016, Ivan Popelyshev <notifications@github.com>
wrote:

> May be its better to make getters/setters for position and scale? Its not
> possible to use different transform class at this point without changing
> all that fields too, right?
>
> —
> <https://github.com/pixijs/pixi.js/pull/2226#issuecomment-172694408>.
>


-- 
Mat Groves

*Technical Partner*


ivanpopelyshev:
Yes, this part is what I dont like:

```javascript
this.position = this.transform.position;
```

Its better if it will be the valid way to use:

```javascript
sprite.position = pos1;
```

GoodBoyDigital:
i see, we could use getter setters for the points? That would that add an extra layer of getter / setters mind? Think that would make a diff on perf?

ivanpopelyshev:
That wont affect performance much, and it is backward compatible. We wont have to explain people  "do not assign those things or you'll break transform". If you change positions of 100k elements, do it in the transform.

GoodBoyDigital:
Sounds good! So you think we should have getter setter for position/scale/pivot and also getter setters x/y/scaleX/scaleY/pivotX/pivotY for ease of use?

dcodeIO:
What if...
```js
sprite.position = sprite2.position
```
```js
Object.defineProperty(Point.prototype, "position", {
   get: function() {
       return this.transform.position;
   },
   set: function(value) {
       this.transform.position.x = value.x;
       this.transform.position.y = value.y;
   }
})
```
Pretty much just a shorthand with a stable GC footprint.

ivanpopelyshev:
Guys! Very very significant thing: Dont use booleans for dirty flags. Use ints and increase them. That way some other things will notice that matrix has changed, bounding box will be able to update.

ivanpopelyshev:
https://github.com/pixijs/pixi.js/pull/2303 - thats what Im talking about. Are you sure that you know how correctly use dirty flags?

GoodBoyDigital:
We are looking intot creating a static transform object for v4. Thanks for starting this @zzm2q :+1:

PR #2235, opened 2015-12-11
Properly map position to point when using different target element

twhitbeck:
When using different interactionDOMElement, if the element doesn't have `width` or `height` specified, the `mapPositionToPoint` method fails. This fixes by using `clientWidth` and `clientHeight` instead.


englercj:
Ping @GoodBoyDigital for review.

themoonrat:
the normal interactionDOMElement is the canvas, which has the .width and .height properties, which are effected by the resolution. If you used a different non-canvas based interactionDOMElement, the clientWidth and clientHeight are not effected by the resolution, so (without checking) this PR wouldn't work?

PR #2239, opened 2015-12-11
Fix for using the loader when loading images

naranjamecanica:
If I use the loader to load images manually and gave them a name, the
images are not added to the texture cache with that name.

We have to use the original url to get them. In could be annoying
because if you load images that could be retina (@2x) you have to acces
them also with the retina url. If I can use the normal url as an
identifier I don’t have to have retina logic to get the right url
spread throughout my code.

For my this fixes this issue, but I don’t know if this will break
existing usage of the loader and cache…

### DIFF
--- a/src/loaders/textureParser.js
+++ b/src/loaders/textureParser.js
@@ -11,8 +11,10 @@ module.exports = function ()
             baseTexture.imageUrl = resource.url;
             resource.texture = new core.Texture(baseTexture);
             // lets also add the frame to pixi's global cache for fromFrame and fromImage fucntions
-            core.utils.BaseTextureCache[resource.url] = baseTexture;
-            core.utils.TextureCache[resource.url] = resource.texture;
+            // we use the provided name if available, otherwise the url of the resource
+            var id = resource.name || resource.url; 
+            core.utils.BaseTextureCache[id] = baseTexture;
+            core.utils.TextureCache[id] = resource.texture;
         }
 
         next();


ivanpopelyshev:
Thats easy solution, but I think we need something better: programmable image loader that will try to load retina images or compressed textures if its possible.

ivanpopelyshev:
something like that:

```
loader.add('test, 'images/test', ["@2x.dds", "@2x.png", ".dds", ".png"])
```

It will look at supported formats, choose one and load it. In your case, ["@2x.png", ".png"].

naranjamecanica:
Sure even better, but how do you access the loaded texture in the game? Because right now there's no way as far as I can tell to acces a texture by name. My pull request is this first step, so the texture is registered under the provided name.

naranjamecanica:
Right now I made a quick fix without patching the PIXI source by creating my own middleware:

    PIXI.loaders.Loader.addPixiMiddleware(function () {
        return function (resource, next) {
            // check if it is an image and we need to add it to the cache under the provided name
            if (resource.data && resource.isImage && resource.texture && resource.name)
            {
                PIXI.utils.BaseTextureCache[resource.name] = resource.texture.baseTexture;
                PIXI.utils.TextureCache[resource.name] = resource.texture;
            }

            next();
        }
    });

ivanpopelyshev:
Im testing new library for pixi that will include whis functional, I'll post it here in few hours.

englercj:
This has come up before: #1611, #1700.

What this PR proposes is a breaking change in how the TextureCache and the `.fromX()` methods work. It is not something that we will be doing in v3. The loader is extensible enough to where if your app needs this, create a custom middleware.

ivanpopelyshev:
I almost fixed it in new lib :)

naranjamecanica:
Ok, clear, sorry I missed that one, stopped using the fromImage methods and directly creating sprite objects, works perfectly!

englercj:
@naranjamecanica :+1: Nice!

ivanpopelyshev:
@naranjamecanica LOL! I just made this thing, helps to track @2x textures and stuff: https://github.com/pixijs/pixi-compressed-textures

naranjamecanica:
@ivanpopelyshev cool I'll check it out!

ivanpopelyshev:
@naranjamecanica works only with latest pixi dev

PR #2263, opened 2015-12-29
Make brfs a devDependency

ArtskydJ:
Since it is used in the build step, `brfs` should be a devDependency, not a dependency.


englercj:
It is a dependency because if you build with browserify it won't properly pull in that dependency since if you "npm install pixi.js" it doesn't pull dev dependencies.

PR #2269, opened 2016-01-01
create baseTexture in the middleware before loading starts

ivanpopelyshev:
this code will work:

loader.add('something.png');
loader.load();
 //loading started but we want to use that NOW
var tex = Texture.fromImage('something.png');

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "earcut": "^2.0.7",
     "eventemitter3": "^1.1.1",
     "object-assign": "^4.0.1",
-    "resource-loader": "^1.6.4"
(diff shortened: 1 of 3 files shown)


englercj:
Couple notes:

1) Don't change the package.json version in PRs
2) This change makes texture loader run both before and after, which I don't understand

ivanpopelyshev:
We had loooong chat in gitter and I admit that this thing needs work. I just wanted fromImage to work through new shiny loader with middlewares and I dont care about that invariant: "all objects must be created only when resources are ready". I thought that BaseTexture can be some kind of promise for an image.

PR #2276, opened 2016-01-07
Update Plane.js

whizzkid:
Mapped UVs for rectangular textures, I'm not sure if nonrectangular textures ever happen? if so, then this method will not suffice.


whizzkid:
weird? the test fails on a piece of code that I haven't touched?   somebody else did a faulty commit maybe?

GoodBoyDigital:
Hmm that failing build check is weird, the PR looks good though!

PR #2285, opened 2016-01-13
Fix InteractionManager fatal: If an interaction leads to a child being removed this loop could potentially throw an error

marekventur:
There is an error in the newest version of PIXI that causes an error when a child of a class removes itself during an interaction event: http://jsbin.com/zoyeyugede/edit?js,output

This patch stops the intermediate error, but it would be probably cleaner to defer all calls to func() to the next tick. Any suggestions or thoughts?


englercj:
This has come up a few times, right now the recommended approach is to not change the scene tree during the interaction loop (in events).

Instead delay it until outside the loop by adding a simple `setTimeout(..., 1);` when you will remove/destroy an object.

https://github.com/pixijs/pixi.js/issues/2270#issuecomment-168524412

Issue #2279, opened 2016-01-08
Does sprite mask NOT support against CanvasRenderer?


holymonson:
Hi, I try using a `Sprite` to mask a `Container` on `CanvasRenderer` but seems no effect, while 
a) on `WebGLRenderer` it does work previously;
b) using `SpriteMaskFilter` doesn't work; 
c) changing the mask from `Sprite` to `Graphics` does work.

Then I found in source code http://pixijs.github.io/docs/core_renderers_canvas_utils_CanvasMaskManager.js.html#line39
```js
    //TODO suport sprite alpha masks??
    //lots of effort required. If demand is great enough..
    if(!maskData.texture)
    {
        CanvasGraphics.renderGraphicsMask(maskData, renderer.context);
        renderer.context.clip();
    }
```
Does it mean sprite mask NOT support against CanvasRenderer? Since we are using easel.js and preparing to move to pixi.js, a bitmap mask is great in demand... (`WebGLRenderer` is out of choice due to #2278 )

englercj:
Correct, alpha masking is not available on the CanvasRenderer, it uses a shader to do it and therefore only works with webgl.

You can still use vector masking on canvas using the `Graphics` class.

holymonson:
IMHO, if this notice is explicitly placed in doc instand of in code would better, but thanks anyway.

englercj:
Feel free to add a PR that adds it to the docs of the `.mask` property. Our docs are generated from source.

PR #2294, opened 2016-01-17
Add doc hint on using Sprite mask on canvas

holymonson:
For the moment, PIXI.CanvasRenderer doesn't support PIXI.Sprite as mask. #2279


holymonson:
mistake

PR #2301, opened 2016-01-19
Update InteractionManager.js

PoulsQ:
For some unknown reason a child can be removed from the initial array between 2 recursive loops.


PoulsQ:
I don't know why it fails, I just added a test for a child existence ...
@GoodBoyDigital, any ideas ?

englercj:
It is failing because of a different test that has been failing for a while.

As far as this PR, I've answered this a few times now: 

https://github.com/pixijs/pixi.js/pull/2285#issuecomment-171657836

PR #2303, opened 2016-01-20
Correct usage of dirty flags in transformStatic

ivanpopelyshev:
That's not the only place where dirty flags can be used. Bounds will use it too, right? And when two objects are tracking this thing, the only way is to use ints. Also, there were no protection against changes in the parent.


ivanpopelyshev:
Also, that's why I dont like dirty flags. Its difficult to use them correctly. Code with parentId and parentVersionId actually reminds me double CAS checks.

ivanpopelyshev:
@GoodBoyDigital Are you here? Did you look at it?

PR #2314, opened 2016-01-25
Dev v4.0.0 better textures

ivanpopelyshev:
Correct usage of dirty flags in transformStatic

Texture rotations


GoodBoyDigital:
nice work @ivanpopelyshev :+1:

Issue #2310, opened 2016-01-25
`mousemove` event dispatches when moving anywhere


sktt:
Greetings, thanks for a great tool...

I discovered that the `mousemove` event is fired even though the mouse movement happens outside of the targeted DisplayObject.. This may also be true for the `touchmove` event.

I assume it's only about checking if a hit occured and don't fire unless so..

Is this correctly assumed? Should I open a pr?

alclub:
Some kinds of drap/drog wouldn't work if mousemove wasn't called while the cursor is outside the object though, for eg, imagine manipulating a control knob. Once the knob was clicked, it shouldn't let go until mouseup.

sktt:
Valid point! But I'm not really sure if this use case justifies emitting
mousemove events on to all interactive displayobjects all the time...?

For your example it would actually be wiser to use the pointer lock api
https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API to fix
the position while mouse button is pressed. If you for instance look at how
ableton live handles their knobs, they actually do this.

Otherwise ux would be flawed anyway because if your window is small you'd
have to release and move the cursor back every time the it leaves the
window. on top of that the mouseup event will probably not even be captured
cause the cursor is not even inside the browser window anymore..

NikkiKoole:
"on top of that the mouseup event will probably not even be captured
cause the cursor is not even inside the browser window anymore.." 

thats why there's a mouseupoutside, 

here's a similar issue  https://github.com/pixijs/pixi.js/issues/1250

sktt:
Since there's already similar issue I guess that has to mean that other developers are also surprised by the behaviour... It's however true that it can easily be checked for so its not really a big deal. I just don't think it makes sense. I could just as well put the `mousemove` listener to the root container itself

NikkiKoole:
No it's intended behaviour ;) it does make sense too imo, imagine you putting a mousemove handler on the root, it could work sure, but then you'd need to keep track of the currently dragging item in the root (to be set in the mouseenter handler of the item), now imagine using touch and doing that for 10+ items ;)

Just set some property 'isDown' or something when you start a mousedown on an item, then in your mousemove do the moving and set the isDown = false in the mouseup(outside) of that item

GoodBoyDigital:
Hello! We have decided to offer the best of both worlds for this issue.
there is a new property in the interactionManager called ```moveWhenInside```
Setting to true will make things work more in line with how the DOM verison works.
[...]

PR #2316, opened 2016-01-25
Don't fire interaction event when not interacting with object

sktt:
- Don't fire mousemove or touchmove unless hit is true
- Fixes issue described in #2310

Maybe its a conscious choice not to respect the `hit` flag for these event. Let me know if so.......

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -565,7 +565,10 @@ InteractionManager.prototype.onMouseMove = function (event)
  */
 InteractionManager.prototype.processMouseMove = function ( displayObject, hit )
 {
-    this.dispatchEvent( displayObject, 'mousemove', this.eventData);
+    if(hit)
+    {
+        this.dispatchEvent( displayObject, 'mousemove', this.eventData);
+    }
     this.processMouseOverOut(displayObject, hit);
 };
 
@@ -780,8 +783,10 @@ InteractionManager.prototype.onTouchMove = function (event)
  */
 InteractionManager.prototype.processTouchMove = function ( displayObject, hit )
 {
-    hit = hit;
-    this.dispatchEvent( displayObject, 'touchmove', this.eventData);
+    if(hit)
+    {
+        this.dispatchEvent( displayObject, 'touchmove', this.eventData);
+    }
 };
 
 /**


ivanpopelyshev:
Wow, that's one serious issue

sktt:
But it appeara in the refenced issue to be a necessary feature for some of us.

I've currently detatched the interaction manager in my pixi project since I my single use case for pixi is just rendering anyway.

There is not a cause of this though but also because events don't propagate from parent to children (should be another issue for this)

It's also because it's potentially a performance killer since it frequently calls getBoundingClientRect which causes a reflow on the dom tree

akbr:
+1 to mousemove and touchmove firing only on hit. Very odd to have them firing all the time on multiple sprites.

englercj:
This was done this was on purpose:

https://github.com/pixijs/pixi.js/commit/f7ccb8b1de524d0fd45f9073f275df0e27fb213a

akbr:
Having reviewed [https://github.com/pixijs/pixi.js/commit/f7ccb8b1de524d0fd45f9073f275df0e27fb213a](https://github.com/pixijs/pixi.js/commit/f7ccb8b1de524d0fd45f9073f275df0e27fb213a), I think this deserves further consideration.

This is getting reported again and again because Pixi.js's implementations of `mousemove` and `touchmove` do not conform to the W3C's DOM specs on [UI Events](https://www.w3.org/TR/uievents/) and [Touch Events](https://www.w3.org/TR/touch-events/). Both specs say that events should be triggered on motion over a _target element_. (I think this makes intuitive sense, since one is listening to a particular target!)

Pixi.js doesn't need to follow the DOM specs, but it's confusing not to. As a user, I don't like juggling multiple meanings for `mousemove` and `touchmove` as I move between the DOM and the scene graph.

I hear @GoodBoyDigital that it's useful to have nonstandard versions of these events in Pixi.js. I just think it's problematic for these nonstandard versions to use standardized names.

ivanpopelyshev:
@englercj Look in the PR closely. Its a bug.

ivanpopelyshev:
@englercj @GoodBoyDigital I dont understand why did you do that. I thought you fixed the other issue in that commit.

GoodBoyDigital:
Hello! We have decided to offer the best of both worlds for this issue.
there is a new property in the interactionManager called ```moveWhenInside```
Setting to true will make things work more in line with how the DOM verison works.
Setting to false will mean the ```mousemove``` and ```touchmove``` will be dispatched regardless of weather the cursor is over the object.

The default of this property is true. :+1:

akbr:
:beers:

sktt:
:+1:

stbaer:
:sake:

PR #2321, opened 2016-01-26
BlendModes premultipliedAlpha hell.

### LINKED ISSUE #2248: Incorrect Transparency Amount With Additive Sprites
Originally I posted this as a bug with pixi-spine before realizing that it is actually an issue with pixi itself: https://github.com/pixijs/pixi-spine/issues/68

Below is an image and source to reproduce the problem. It appears as if alpha is being applied twice to sprites with additive blend modes and as a result a faded out sprite can appear barely visible, while the same values in other programs (photoshop, spine) would still be clearly visible. The top bar uses normal blend mode, and the bottom two additively blended bars should appear exactly the same as the top since the background is black.

I created it by making 11 white sprites and tinting the alpha of each at increments of 10 percent and setting the blend mode to additive.

You can see the issue here: http://codepen.io/mgard/pen/MKKWLR

download

ivanpopelyshev:
Trying to fix #2248. ADDITIVE is working like intended, but I dont know how to fix MULTIPLY yet :)


GoodBoyDigital:
nice one man!

PR #2340, opened 2016-02-05
Multiple transformations with different pivot point in DisplayObject

louisatome:


GoodBoyDigital:
Hi there, if you could refactor this pr for v4 we can create a new transform. Thanks!

Issue #2342, opened 2016-02-07
InteractionManager doesn't yet implement event bubbling


akbr:
Pixi.js should bubble interaction events up the scene graph. (See [#625](https://github.com/pixijs/pixi.js/issues/625).)

Previously, bubbling was handled by EventTarget.js. However, after EventTarget.js was deprecated in favor of EventEmitter3, bubbling disappeared. (See [#914](https://github.com/pixijs/pixi.js/pull/914) and [#1532](https://github.com/pixijs/pixi.js/pull/1532#event-287332803).)

InteractionManager is likely the best place to handle bubbling (particularly `InteractionManager.prototype.dispatchEvent`, where it looks like implementation was contemplated, but never completed).

Do we want _both_ callbacks and `.emit`s to bubble?

If desired, happy to submit a patch.

GoodBoyDigital:
Agreed, a patch would be super! :+1:

endel:
Having global interaction events (https://github.com/pixijs/pixi.js/pull/2658) solves this for interactivity, but I'm still missing bubbling for custom `.emit` events. 

As a matter of code structure sometimes it's needed for a DisplayObject to emit an event and handle it on its parent. Maybe we could implement this as a plugin? Any ideas?

ivanpopelyshev:
Look at how pixi-display changes it:  https://github.com/pixijs/pixi-display/blob/master/src/InteractionManagerMixin.js
You can do something similar to pass event not only to specific element but to its parents that are interactive as well.

englercj:
Closing as the original request was implemented.

PR #2358, opened 2016-02-13
implemented event bubbling

akbr:
See #2342.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -309,26 +309,27 @@ InteractionManager.prototype.update = function (deltaTime)
 };
 
 /**
- * Dispatches an event on the display object that was interacted with
+ * Dispatches an event on the display object that was interacted with, and then bubbles up the scene graph
  *
  * @param displayObject {PIXI.Container|PIXI.Sprite|PIXI.extras.TilingSprite} the display object in question
- * @param eventString {string} the name of the event (e.g, mousedown)
+ * @param eventString {string} the name of the event (e.g., mousedown)
  * @param eventData {object} the event data object
  * @private
  */
-InteractionManager.prototype.dispatchEvent = function ( displayObject, eventString, eventData )
+InteractionManager.prototype.dispatchEvent = function (displayObject, eventString, eventData)
 {
-    if(!eventData.stopped)
+    eventData.target = displayObject;
+    eventData.type = eventString;
+
+    var bubbleTarget = displayObject;
+    while (bubbleTarget && !eventData.stopped) 
     {
-        eventData.target = displayObject;
-        eventData.type = eventString;
-
-        displayObject.emit( eventString, eventData );
-
-        if( displayObject[eventString] )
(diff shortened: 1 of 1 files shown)


gurmukhp:
Would it be possible to see an example of this working? I've tried this approach but the event.target value is always the parent DisplayObject and not the children, although looking in event.children[1], I can see the one that I clicked.

A working example would be great, thanks!

akbr:
Hello @gurmukhp! Here is an example for you: http://jsbin.com/noboliqamo/1/edit?html,js,console,output

Click events show up in the console.

Let me know if this helps!

akbr:
Ok, I think I see your issue.

My example works great _unless_ the outer Pixi container is set as interactive. If it is, it prevents its children from hearing the click.

I'll see if I can find time to take a closer look soon.

gurmukhp:
Thank you! That's really useful! Do you know if event bubbling from within a Pixi Canvas brings any performance benefit? 

In the HTML world the more events bound, the slower it can be, do you know if this also applies to Pixi?

englercj:
@gurmukhp It will of course be slower, it will have to emit events many more times than before.
.

gurmukhp:
Thanks, so to confirm, we should try to limit the number of event listeners possible.

Any idea when this will make it to the public release?

englercj:
> Thanks, so to confirm, we should try to limit the number of event listeners possible.

@gurmukhp No, if an event happens it will emit from the target through all the parents. It doesn't matter how many listeners you add. Aside from the overhead of actually calling your callbacks, there isn't any significant difference between 10 listeners and 100. It still walks the same number of nodes, and emits the same number of events.

akbr:
Yep: Any performance penalty would come from deeply-nested targets bubbling through lots of parents, _not_ the volume of interactive targets. (If you want to use event bubbling in a deep scene graph, you can use `stopPropagation` to mitigate this.)

I use event bubbling not for performance reasons, but because it can help me write cleaner code. In some cases, I think it's best to separate event logic from individual sprites.

> Any idea when this will make it to the public release?

This doesn't seem to be a high priority issue. (I might be the only one who really cares!)

Given the potential performance implications, and the fact most Pixi devs don't seem to use bubbling, I'm wondering if bubbling should ultimately be implemented as an optional toggle (defaulting to "off").

bigtimebuddy:
@akbr Where there is clear trade-off for performance v. usability, having an opt-in toggle for usability is probably a better approach.

endel:
About performance, I don't think this would be an overhead for any PIXI application. Some benchmarks would be necessary to confirm exactly how much time it would consume having to bubble in a long display list.

akbr:
I'm closing this in light of https://github.com/pixijs/pixi.js/pull/2658, which adds global events to the `InteractionManager`. As @andrewstart notes, global events allow for "some of the use cases of event bubbling without the performance hit of deep display lists." (:beers:)

If there is a desire for "proper" event bubbling -- which should probably be opt-in, as discussed above -- someone else can pick this up.

PR #2363, opened 2016-02-18
Add breakWords option to PIXI.Text

swanson:
Related: https://github.com/pixijs/pixi.js/issues/1906

Adds `breakWords` option to `PIXI.Text` that allows for breaks to be added in the middle of words if they are wider than `wordWrapWidth`. Defaults to false and maintains current behavior.

With `breakWords` option:
image

With current behavior (notice some characters are cut off):
image


GoodBoyDigital:
Lovely stuff! Thanks @swanson :+1:

PR #2381, opened 2016-03-08
Using resolution in TilingSprite canvas renderer

LTcolombo:
Tiling sprite doesn't respect pixel ratio when rendering on canvas.

In case of using a single image for the texture it appears bigger.
In case of using atlases the bug is much more severe, as a wrong frame is being cut out from the base texture.


GoodBoyDigital:
thanks for this @LTcolombo :+1:

PR #2396, opened 2016-03-14
Dev v4 flip

ivanpopelyshev:
This stuff is needed for flip.


ivanpopelyshev:
This is needed for ability to change transform on fly. Transform can be re-created to Transform3d. Also, rotation will point to "euler.z" in that case

ivanpopelyshev:
Also things like "sprite.position = p" will actually copy stuff instead of assigning that way :) We can write "copy" in the property.

Issue #1342, opened 2015-01-10
ColorMatrixFilter not multiplying RGB on A


MishaAmashukeli:
http://codepen.io/mishaa/pen/pvRLGJ
        var filter = new PIXI.ColorMatrixFilter();
        filter.matrix = [
            1, 0, 0, 0,
            0, 1, 0, 0,
            0, 0, 1, 0,
            0, 0, 0, 0
        ];
It's all zeroes in the alpha row but the sprite is still visible. The culprit is the blend mode.
From WebGLRenderer.js:
`PIXI.blendModesWebGL[PIXI.blendModes.NORMAL]        = [gl.ONE,       gl.ONE_MINUS_SRC_ALPHA];`

With this blend mode, if you set pixel's alpha to 0 in a shader, the pixel won't get fully transparent. You should multiply the RGB value to alpha value to get the correct transparency. But that's not intuitive and not all PIXI filters do that. 
So, should not it be `[gl.SRC_ALPHA,       gl.ONE_MINUS_SRC_ALPHA]` ?

MishaAmashukeli:
Ok I read some more about this and appearantly this approach is called premultiplied alpha and it's a good thing. So it's the filters that need fixing, not blending.

GoodBoyDigital:
Should be all good now in v3 I believe! :+1:

mattlima:
Sorry to jump in on a closed issue, I've done a lot of research before posting here. I'm working on some recursive filtering in a project and I'm having a similar issue with alpha; There's not a ton of material on the behavior of the alpha channel in the color matrix filter; I've looked at the source and it seems very straightforward in that the alpha channel of pixels resulting from a matrix filter where indexes 15-19 are zero should be zero, i.e., the result should be a transparent sprite. This isn't the case.  I've updated the original codepen to pixi v3: http://codepen.io/mattlima/pen/LGOBmv – I'm sure you're insane busy but if you could point me in the right direction I'd really appreciate it.

Issue #2302, opened 2016-01-19
ColorMatrixFilter does not operate correctly on alpha channel


mattlima:
This is a follow up to https://github.com/pixijs/pixi.js/issues/1342 - I've done a lot of research before posting here. I'm working on some recursive filtering in a project and I'm having a similar issue with alpha; There's not a ton of material on the behavior of the alpha channel in the color matrix filter; I've looked at the source and it seems very straightforward in that the alpha channel of pixels resulting from a matrix filter where indexes 15-19 are zero should be zero, i.e., the result should be a transparent sprite. This isn't the case.  I've updated the original codepen to pixi v3: http://codepen.io/mattlima/pen/LGOBmv – I'm sure you're insane busy but if you could point me in the right direction I'd really appreciate it.

jackie-t:
I am affected by this too - it seems that transparent parts are heavily mixed with black. Even if I only change the hue by 1 (with the official method, so I'm not playing around with custom matrices), the color is distorted.

Look here for an example that overlaps two pictures - one with the ColorMatrixFilter applied, and one without. As it is, it makes the ColorMatrixFilter almost completely useless.

http://codepen.io/anon/pen/JGVpyP

Here's a screenshot of the result - yes, the black "bar" is the same green you see above, with the hue shifted by 1°: https://gyazo.com/12af2b1c8f8d50b87f94f46d1f680b48

jackie-t:
Just to reiterate, the referenced issue does not fix this bug. Changing the hue by even 1° causes issues with images that rely on the alpha channel. The problem in https://gyazo.com/12af2b1c8f8d50b87f94f46d1f680b48 (that two images with a 1° hue shift are black when merged) exists despite the fix.

PR #2397, opened 2016-03-16
Fixed the alpha for the ColorMatrixFilter

bigtimebuddy:
I believe this fixes issues #1342 and #2302. Alpha now is handled correctly when applying additive color. Credit to @andrewstart for [this fix](https://github.com/CloudKidStudio/PixiFlash/blob/master/src/ColorMatrixFilter.js).

### Before
screen shot 2016-03-15 at 9 37 36 pm

### After
screen shot 2016-03-15 at 9 37 04 pm

### Example Code
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Example</title>
    <link rel="stylesheet" href="style.css">
    <script src="pixi.js/bin/pixi.js"></script>
  </head>
  <body>
    <canvas id="stage" width="200" height="200"></canvas>
    <script>
      var renderer = new PIXI.autoDetectRenderer(200, 200, {
        backgroundColor: 0x000000,
        antialias: true
      });
      var filter = new PIXI.filters.ColorMatrixFilter();
      filter.matrix = [
         0.4, 0, 0, 0, 0.4, 
         0, 0.4, 0, 0, 0.4, 
         0, 0, 0.4, 0, 0.4, 
         0, 0, 0, 1, 0
      ];
      var gfx = new PIXI.Graphics()
        .beginFill(0xff0000, 1)
        .drawCircle(100, 100, 50);
      gfx.filters = [filter];
      gfx.alpha = 1;
      stage.addChild(gfx);
      function update() {
        renderer.render(stage);
      }
      update();
    </script>
  </body>
```

### DIFF
--- a/src/filters/color/colorMatrix.frag
+++ b/src/filters/color/colorMatrix.frag
@@ -13,24 +13,24 @@ void main(void)
         gl_FragColor.r += (m[1] * c.g);
         gl_FragColor.r += (m[2] * c.b);
         gl_FragColor.r += (m[3] * c.a);
-        gl_FragColor.r += m[4];
+        gl_FragColor.r += m[4] * c.a;
 
     gl_FragColor.g = (m[5] * c.r);
         gl_FragColor.g += (m[6] * c.g);
         gl_FragColor.g += (m[7] * c.b);
         gl_FragColor.g += (m[8] * c.a);
-        gl_FragColor.g += m[9];
+        gl_FragColor.g += m[9] * c.a;
 
      gl_FragColor.b = (m[10] * c.r);
         gl_FragColor.b += (m[11] * c.g);
         gl_FragColor.b += (m[12] * c.b);
         gl_FragColor.b += (m[13] * c.a);
-        gl_FragColor.b += m[14];
+        gl_FragColor.b += m[14] * c.a;
 
      gl_FragColor.a = (m[15] * c.r);
         gl_FragColor.a += (m[16] * c.g);
         gl_FragColor.a += (m[17] * c.b);
         gl_FragColor.a += (m[18] * c.a);
-        gl_FragColor.a += m[19];
+        gl_FragColor.a += m[19] * c.a;
 
 }


GoodBoyDigital:
nice one @bigtimebuddy :+1:

jackie-t:
Hmm - I tried re-running my tests using the file at https://github.com/pixijs/pixi.js/blob/dev/bin/pixi.min.js and it still behaves as before (https://gyazo.com/ec10e6589fbcc3692ec07ecd01970f5d - the graphics with the CMF are from the same file, with the hue changed by 1 - so they should look virtually the same as the "No CMF" ones).

Do I need to do anything else to verify it? The codepen is at http://codepen.io/anon/pen/JGVpyP - I copied the whole pixi.min.js file from the dev-branch into it, because I don't know how to directly link files from github.

t-mw:
Was this fix definitely correct? It makes the 5th column of the matrix completely redundant, since the 4th column already multiplies by alpha. The screenshots don't show a bug, because the 5th column is meant to ignore alpha (see [here](http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/filters/ColorMatrixFilter.html)). To fix the example code you would move the values in the 5th column to the 4th column:

```javascript
filter.matrix = [
         0.4, 0, 0, 0.4, 0, 
         0, 0.4, 0, 0.4, 0, 
         0, 0, 0.4, 0.4, 0, 
         0, 0, 0, 1, 0
];
```

andrewstart:
It is my understanding that the 4th column is not an offset, but a multiplier of existing values just like the first 3 columns. The problem in the screenshots is that filter used is not offsetting the alpha at all, so it is counterintuitive to have transparent areas stop being transparent just because you were offsetting colors - that is not the behavior used by Flash, probably because of a difference between an expectation of pre-multiplied alpha vs not pre-multiplied alpha.

t-mw:
I agree that's what the 4th column should do, and it already does. The filter used in the screenshots doesn't use the 4th column. This commit changes the behaviour of the 5th column and actually diverges from the behaviour used by Flash (according to the docs I linked previously).

Issue #2328, opened 2016-01-29
No `tint` property for `PIXI.Container`


AndyBarron:
Is it possible to implement `tint` for `PIXI.Container`? Since `alpha` is already supported, `tint` makes sense.

I know it would technically be possible to loop through the children and tint them, but this won't work e.g. if you have a `Container` with instances of both `Sprite` and `Container` as children. And in my use case, the children of the `Container` change often, so I don't want them to have a residual tint after calling `removeChild`.

ivanpopelyshev:
so, separate it to tint and worldTint, like alpha and worldAlpha?

AndyBarron:
Well, `worldAlpha` is the read-only computed alpha based on a `DisplayObject`'s parents. So it would make sense to add `worldTint` in the same fashion if this went through.

rejemy:
I work around this by using a ColorMatrixFilter, which you can set on a container, but it's overkill for most tinting use cases. Having tint be supported by all display objects along the chain would definitely be more user friendly, and is certainly technically possible, if implemented along the same lines as alpha currently is.

GoodBoyDigital:
we can make a little plugin for this I think!

englercj:
Moving to v5.

praleks:
tint is color multiplication, how about to make color addition e. g. color offset?

ivanpopelyshev:
@praleks  here: https://github.com/gameofbombs/pixi-heaven

PR #2398, opened 2016-03-16
Added tint to DisplayObject that supports Container tinting - #2328

andrewstart:
Changed existing tint functionality to pull from the new worldTint so
that it respects the parent tint.

### DIFF
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -108,6 +108,39 @@ function DisplayObject()
      * @member {PIXI.Rectangle}
      */
     this.filterArea = null;
+    
+    /**
+     * The tint applied to the DisplayObject (and its children). This is a hex value. A value of
+     * 0xFFFFFF will remove any tint effect.
+     *
+     * @member {number}
+     * @default 0xFFFFFF
+     */
+    this.tint = 0xFFFFFF;
+    
+    /**
+     * The multiplied tint of the DisplayObject.
+     *
+     * @member {number}
+     * @readOnly
+     */
+    this.worldTint = 0xFFFFFF;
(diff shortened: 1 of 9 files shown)


bigtimebuddy:
@GoodBoyDigital could you take a look at Andrew's PR and let us know what you think?

ivanpopelyshev:
Ok guys, a interruption here: that kind of stuff needs to be separated into rendering context: all "world" variables have to be calculated into a component, otherwise it will get messy. I'm already having a lot of problems with "world" stuff in my fork

GoodBoyDigital:
I had a really great thought about how we could implement this kind of functionality as a plugin.
A way that does not involve changing anything but adding the extra variables to a displayObject. Imagine pixi used a series of passes through the scene graph.

So the default pass is what we have - updates the transform + the alpha
We could add a new tint pass that would go through the scene graph and calculate the world tint of all the objects.
Further down the line we could add new passes for extra functionality
After all passes are complete we can then do the render pass.

What do you guys think?

ivanpopelyshev:
even alpha logic is not that easy, filters and masks can break it. It needs to be separated from DisplayObject, in a component. And yes, different pass.

ivanpopelyshev:
Proper way would be addign special RenderingContext component that takes care of "tint", "alpha", "renderable". Possible to add this component as second parameter to "renderWebGL" and "renderCanvas". That will allow many smart hacks and performance optimizations.

bigtimebuddy:
@GoodBoyDigital Like your suggestion here.

englercj:
Closing this since we will be redesigning how this works a bit in v5, and it probably wont make it in v4.

PR #2401, opened 2016-03-16
Correct a reference error in the basic example

markfinger:
The basic example in README breaks on the first call to `animate`. 

These changes alter the scope of the `bunny` variable such that it works as expected.


GoodBoyDigital:
ta!

englercj:
This was already fixed a while ago in `dev`:

https://github.com/pixijs/pixi.js/commit/db953c3810a424d6423ce8810cf32c9f05fb3edb

PR #2414, opened 2016-03-26
Dev v4.0.0 geom

ivanpopelyshev:
This is temporary pull request for v4.1, please DONT MERGE IT YET.

Only simple objects like Sprites, Graphics and Text are tested.

build for this version is at my dropbox, https://dl.dropboxusercontent.com/u/46636530/pixijs/v4.1/pixi.js


gvgv13:
hola

ivanpopelyshev:
basically, all objects in "c2d" folder have uid and version

and if object is computed on top of two others, it remembers their uid and version

so it wont be updated if versions weren't changed

DisplayObject now have three transform objects

1) transform, which is local one

2) computedTransform - based on local transform and parent computedTransform

3) projectedTransform - based on computedTransform and worldProjection. It is used by CanvasRenderer if there's a Camera

also, three "geometry" objects

geometry is just set of points, and sometimes indices

geometry, computedGeometry, projectedGeometry

they have bounds and bounds ARE CACHED

everything is cached :)

so there are MANY objects that are computed based on other twos

also some "update" actions have two or three steps, and if one of input objects is 3d, it will be able to override the action to make result object also 3d

all bounds stuff is moved to geometries and DisplayObject, and most common behaviours of bounds is taken care of
even mouse global coords are DisplayPoint now, which projects to Raycast2d, which will become another Raycast2d for particular object
and its cached :)))

no need of hit testing the object with old coords and old camera

that's madness

one of the most sophisticated things I made :)

and whats strange, it worked from first run

also, width/height of Sprite is based on "size" which is ObservablePoint

if texture/size/anchor is changed, then geometry of sprite will be re-calculated

its what you did in "calculateVertices" thing in sprite

and SIZE DOES NOT AFFECT SCALE

 its awesome :) people wanted that

ivanpopelyshev:
How many bunnies do you see there? http://ivanpopelyshev.github.io/examples/index.html?s=camera&f=mirror.js&title=Mirror

PR #2417, opened 2016-03-30
Renamed Const to const

frankbaele:
require was failing on unix based systems


GoodBoyDigital:
thanks for that! :+1:

PR #2418, opened 2016-03-30
Don't pass Container's destroyChildren parameter to a Sprite's destroyTexture parameter

Nick-Riggs:
I don't think it's intended that a Container's destroyChildren parameter is passed to the Sprite's s destroyTexture parameter. This leads to the unintentional destruction of shared textures.

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -628,8 +628,7 @@ Container.prototype.destroy = function (destroyChildren)
 {
     DisplayObject.prototype.destroy.call(this);
 
-    if (!destroyChildren) {
-    } else {
+    if (destroyChildren) {
         for (var i = 0, j = this.children.length; i < j; ++i) {
             var child = this.children[i];
             var isSprite = !!child._texture; //TODO: Need a better way to detect if it's a sprite
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -628,11 +628,20 @@ Container.prototype.destroy = function (destroyChildren)
 {
     DisplayObject.prototype.destroy.call(this);
 
-    if (destroyChildren)
-    {
-        for (var i = 0, j = this.children.length; i < j; ++i)
-        {
-            this.children[i].destroy(destroyChildren);
+    if (!destroyChildren) {
+    } else {
+        for (var i = 0, j = this.children.length; i < j; ++i) {
+            var child = this.children[i];
+            var isSprite = !!child._texture; //TODO: Need a better way to detect if it's a sprite
+
+            if (isSprite)
+            {
+                child.destroy();
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


src/core/display/Container.js
```diff
@@ -628,11 +628,20 @@ Container.prototype.destroy = function (destroyChildren)
 {
     DisplayObject.prototype.destroy.call(this);
 
-    if (destroyChildren)
-    {
-        for (var i = 0, j = this.children.length; i < j; ++i)
-        {
-            this.children[i].destroy(destroyChildren);
+    if (!destroyChildren) {
+    } else {
```
englercj: ??

src/core/display/Container.js
```diff
@@ -628,11 +628,20 @@ Container.prototype.destroy = function (destroyChildren)
...
+        for (var i = 0, j = this.children.length; i < j; ++i) {
+            var child = this.children[i];
+            var isSprite = !!child._texture; //TODO: Need a better way to detect if it's a sprite
```
englercj: `X instanceof Sprite`

src/core/display/Container.js
Nick-Riggs: Whoops sorry, I think that's a hold over from some older pixi code I was working in. I'll correct it.

src/core/display/Container.js
Nick-Riggs: Can we do that here? We'd have to require Sprite and it requires Container - circular reference.

englercj: Yeah I don't think we can, which is why this idea of checking here isn't great. A parent knowing about how its children override a function is a crazy concept in OOP.

Ideally we would fix the discrepency, but changing this behavior (or the function signature) is a breaking change which is why I haven't done it yet :(

Nick-Riggs: So any chance of getting some fix merged in? Right now in our game, we are monkey patching the fixed function in to get around the bug. Do you recommend sticking with that for now?

Nick-Riggs:
Looks like this bug still exists in v4. Would it be acceptable to add a isSprite property to test if a Container is a sprite? Would that be more acceptable than looking to see if the Container has a texture?

englercj:
I'd rather have type constants and test those, similar to what phase does:

```js
const TYPES = {
    DISPLAY_OBJECT: 1,
    CONTAINER: 2,
    SPRITE: 3,
    //....
};

function Sprite() {
    this.type = TYPES.SPRITE;
}
```

andrewstart:
Since we are doing a major version bump, now would be the time to make a breaking change to Sprite's destroy function. I made a suggestion in https://github.com/pixijs/pixi.js/issues/2392 about that.

bigtimebuddy:
Fixed with #2592

PR #2428, opened 2016-03-31
Updated WebGL blend modes to v3.0.10 versions

digibo:
Updated WebGL blend modes to their v3.0.10 versions, as they seem to be the proper implementations


digibo:
Working to get blend modes working as goog as in v3. This should fix the issues I've noticed so far.

GoodBoyDigital:
thanks @digibo !

PR #2449, opened 2016-04-04
Add jshint to dev-dependencies

ylambda:
Previously this wasn't required because `gulp-jshint` would install it. With `gulp-jshint@^2.0.0`  `jshint` is now a peer-dependency.

https://github.com/spalger/gulp-jshint/tree/v2.0.0


englercj:
:+1:

PR #2450, opened 2016-04-04
BaseTexture.scaleMode default should be PIXI.SCALE_MODES.DEFAULT, not…

Hachitus:
… PIXI.SCALE_MODES.LINEAR

I know it's insignificantly tiny typo, but still :).


GoodBoyDigital:
every little helps! thanks @Hachitus

Issue #2034, opened 2015-08-06
Include helpful guards in pixi debug version


englercj:
It would be pretty cool if our debug build of pixi was instrumented with helpful debug guards. For example, ensuring you don't assign `NaN` to matrix values. This could be done via some metaprogramming that ensured the production build did not include any of the guards for maximum performance.

devdoomari:
:+1 on this...
I just got burnt by having NaNs creep out on transform matrices.
I couldn't figure out why my someGraphics.rotate didn't work
 -- turned out I set someGraphics.x = undefined somewhere.
I've lost like 7 hours on this :(

Anyway, are 'isNaN' calls expensive?
Also, what are you thinking about code inclusion/exclusion?
I don't have any ideas on what pixify is, so I'm wondering what pixify can do

englercj:
pixify is just a collection of browserify transforms and build scripts that we use.

PR #2451, opened 2016-04-04
Better, more useable deprecation warnings

bigtimebuddy:
### Overview

Added support for a collapsable stack trace in deprecation warnings. Makes it easier to track-down usage of deprecated methods/properties/classes. Added some styling to the warning to clearly indicate deprecation (as opposed to WebGL warnings or other console.log warnings). For browsers that don't support `groupCollapsed`, fallback to regular `console.warn`.

**Before**
screen shot 2016-04-04 at 4 29 52 pm

**After**
screen shot 2016-04-04 at 4 30 18 pm

### Test
```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8">
        <title>Deprecation</title>
    </head>
    <body>
        <script src="pixi.js/bin/pixi.min.js"></script>
        <script>
                backgroundColor: 0xffffff,
                antialias: true
            });
            var stage;
            function create() {
                stage = new PIXI.Stage();
                stage = new PIXI.DisplayObjectContainer();
            }
            create();
            function update() {
                renderer.render(stage);
            }
            update();
        </script>
    </body>
</html>
```

### DIFF
--- a/src/deprecation.js
+++ b/src/deprecation.js
@@ -9,19 +9,25 @@ var core = require('./core'),
 // useful for tracking-down where deprecated methods/properties/classes
 // are being used within the code
 function warn(msg) {
-    var stack = new Error().stack; 
+    var stack = new Error().stack;
 
-    // chop off the stack trace which includes pixi.js internal calls
-    stack = stack.split('\n').splice(3).join('\n');
-
-    if (console.groupCollapsed) {
-        console.groupCollapsed('%cDeprecation Warning: %c%s', 'color:#614108;background:#fffbe6', 'font-weight:normal;color:#614108;background:#fffbe6', msg);
-        console.warn(stack);
-        console.groupEnd();
+    // Handle IE < 10 and Safari < 6
+    if (typeof stack === 'undefined') {
+        console.warn('Deprecation Warning: ', msg);
     }
     else {
-        console.warn('Deprecation Warning: ', msg);
-        console.warn(stack);
+        // chop off the stack trace which includes pixi.js internal calls
+        stack = stack.split('\n').splice(3).join('\n');
+
+        if (console.groupCollapsed) {
+            console.groupCollapsed('%cDeprecation Warning: %c%s', 'color:#614108;background:#fffbe6', 'font-weight:normal;color:#614108;background:#fffbe6', msg);
+            console.warn(stack);
+            console.groupEnd();
+        }
+        else {
+            console.warn('Deprecation Warning: ', msg);
+            console.warn(stack);
+        }
     }
 }
 
--- a/src/deprecation.js
+++ b/src/deprecation.js
@@ -5,6 +5,26 @@ var core = require('./core'),
     extras = require('./extras'),
     filters = require('./filters');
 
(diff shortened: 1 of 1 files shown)


englercj:
Before I even start to review this I want to say a couple things:

1. This is a beautiful PR. Full explanation, test code, resulting screenshots...just wonderful.
2. This is a great change, I'm sure every user of pixi will appreciate this.

GoodBoyDigital:
I agree with @englercj, nice PR 👍

src/deprecation.js
```diff
@@ -5,6 +5,26 @@ var core = require('./core'),
...
     filters = require('./filters');
 
+// provide method to give a stack track for warnings
+// useful for tracking-down where deprecated methods/properties/classes
+// are being used within the code
+function warn(msg) {
+    var stack = new Error().stack; 
+
+    // chop off the stack trace which includes pixi.js internal calls
+    stack = stack.split('\n').splice(3).join('\n');
```
englercj: Is it guaranteed to only be 3 stack levels? I feel like this might be too fragile. Maybe detect if a line is internal to pixi and filter it instead?

bigtimebuddy: It's at least 3 calls always, the `console.warn`, the inline `warn` function and the deprecation call below where `warn()` is invoked. Its removes those 3 calls.

bigtimebuddy:
@englercj Thanks, I'm a big fan of clear and thorough PRs. Hopefully other people will follow as an example.

bigtimebuddy:
In another project that I built, it's using this same deprecation pattern as Pixi. We are actually removing deprecation warnings in release mode and only leaving them in debug/un-minified version. Using the [gulp-preprocess plugin](https://www.npmjs.com/package/gulp-preprocess). Would be pretty easy to add here too, if it was felt deprecation warnings weren't necessary in release mode. Would remove some fat from the release build. 

For instance, would implement like this:

```js
      DisplayObjectContainer: {
          get: function()
          {    
              //@if DEBUG     
              warn('DisplayObjectContainer has been shortened to Container, please use Container from now on.');
              //@endif
              return core.Container;
          }
      }
```

GoodBoyDigital:
@bigtimebuddy looks good to me!

src/deprecation.js
JiDW: stack is introduced in IE > 9 and Safari >5. We should check if stack is not undefined.

bigtimebuddy: Good catch @JiDW, thanks.

JiDW: I checked on Edge and it needs a "4" splice parameter to work. The stack property being non standard, I think that @englercj idea make sense here.

bigtimebuddy:
@GoodBoyDigital I'm going to do the preprocess stuff as a separate PR since it requires some more complicate gulp re-configuration. I can work with @englercj on that. This PR should be ready to merge though.

GoodBoyDigital:
good stuff!

englercj:
Preprocessor removal is a good idea, I even suggested it before (#2034). I would prefer it completely removed this file rather than just the warn message too.

PR #2463, opened 2016-04-06
trying to fix clear()

ivanpopelyshev:
Behaviour of renderTexture is a bit different


GoodBoyDigital:
👍

src/core/renderers/canvas/CanvasRenderer.js
```diff
@@ -178,16 +178,18 @@ CanvasRenderer.prototype.render = function (displayObject, renderTexture, clear,
         context.clear();
     }
 
-    if( clear || this.clearBeforeRender)
+    if(clear !== undefined || this.clearBeforeRender)
```
englercj: If `clear` is `false`, this check will pass...

PR #2465, opened 2016-04-06
awesome implementation for size/anchor abstraction

ivanpopelyshev:
Huge improvement of previous PR that was rejected.

1. spriteFrame is independent from Sprite, many sprites (particles) can use one frame. That removes some calculations from ParticleContainer, SpriteRenderer (both webgl and canvas). 
2. Its possible to use one spriteFrame per different textures.
3. renderers are simplified, they dont know about all that internal problems with crop/trim 
4. sprite.size.x = 100 will set sprite width to 100, preserving aspect ratio. It wont affect scale.
5. If size.x or size.y is 0 and the other one has non-zero value, aspect ratio will be preserved.
6. sprite.width and sprite.height works like before, affecting scale

The only one side effect - you have to call sprite.onTextureUpdate() after trim/crop or something was changed. And PIXI.Text didnt do that, so I fixed it.

Works with all examples.


ivanpopelyshev:
Ok, I see the problem with rotations

ivanpopelyshev:
And now its fixed.

ivanpopelyshev:
Side effect removed.

Now rectangles inside texture are observable, and sprites will automatically detect if texture frame version has been changed.

PR #2467, opened 2016-04-07
Improved text style API

AMorporkian:
With these changes, the text style API is more functional. You can now set individual style properties, and have the changes reflect immediately.

For example, to change the fill in the past, you would need to do:

```js
textObject.style.fill = PIXI.utils.hex2string(0xFFC321);
textObject.dirty = true;
```

With this PR, you're now able to do the same with:
```js
textObject.style.fill = 0xFFC321;
```

Additionally, commit exposes a new method, Text.resetStyle(), which resets all values in the style object to their defaults.


bigtimebuddy:
While I think this addresses some important issues with the Text object, I'm not crazy about the implementation. The binding/calling to create the setters and getters feels a little too clever and hard to read and harder to document. Also, `_createStyleObject` seems like it doesn't need to be it's own function, why not have this be the style setter?

I think we're missing an opportunity to create a TextStyle class which can externalize all of the style settings from the Text object. This TextStyle class could be shared between Text fields. But because a new styleBase is created every style set, it makes it trickier to share references.

```js
// Create a shared/style
var style = new PIXI.TextStyle({
    font: 'bold 12pt Arial',
    fill: "#ff0000"
});
var text1 = new PIXI.Text("some text", style);
var text2 = new PIXI.Text("more text", style);

// or alternatively
text1.style = text2.style = style;
```

If TextStyle extends EventEmitter, a "change" event could be emitted whenever a style property is changed it it would mark all Text objects as dirty.

To support backward-compatibility for plain Objects the Text style setter might look like this:

```js
Object.defineProperty(Text.prototype, 'style', {
    get: function() {
        return this._style;
    },
    set: function(style) {
        if (this._style)
            this._style.off('changed', this.onStyleChanged);
        // support for plain Objects
        if (!(style instanceof TextStyle)) {
            if (!this._style)
                this._style = new TextStyle(style);
            else
                this._style.reset(style);
            style = this._style;
        }
        this._style = style;
        this._style.on('changed', this.onStyleChanged);
    }
});

AMorporkian:
I feel like an event emitter might be overcomplicating it. We could just make Text.dirty a getter that checks for `TextStyle.dirty || this.dirty`.

AMorporkian:
Wait, I see the issue with that. If multiple `Text`s are using the same style, you'd never know if it's safe to unset dirty. I'll take a crack at a version with an EventEmitter.

AMorporkian:
Closing in favor of incoming PR.

PR #2468, opened 2016-04-07
TextStyle, a more flexible and fluent interface for modifying text styling

AMorporkian:
This commit introduces `PIXI.TextStyle`. The API for `PIXI.Text` is left mostly unchanged. `TextStyle` extends EventEmitter, and emits a `styleChanged` event every time it is changed. `Text` hooks onto that event and marks `Text.dirty = true`. 

Additionally, `TextStyle` exposes `TextStyle.clone()`, which allows an end user to clone a previously defined text style.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -92,10 +92,7 @@ Object.defineProperties(Text.prototype, {
     width: {
         get: function ()
         {
-            if (this.dirty)
-            {
-                this.updateText();
-            }
+            this.updateText(true);
 
             return this.scale.x * this._texture._frame.width;
         },
@@ -115,10 +112,7 @@ Object.defineProperties(Text.prototype, {
     height: {
         get: function ()
         {
-            if (this.dirty)
-            {
-                this.updateText();
-            }
+            this.updateText(true);
 
             return  this.scale.y * this._texture._frame.height;
         },
@@ -187,11 +181,14 @@ Object.defineProperties(Text.prototype, {
 
 /**
  * Renders text and updates it when needed
- *
+ * @param respectDirty {boolean} Whether to abort updating the text if the Text isn't dirty and the function is called.
  * @private
  */
-Text.prototype.updateText = function ()
+Text.prototype.updateText = function (respectDirty)
 {
+    if (!this.dirty && respectDirty) {
+        return;
+    }
     var style = this._style;
     this.context.font = style.font;
 
@@ -353,12 +350,7 @@ Text.prototype.updateTexture = function ()
  */
 Text.prototype.renderWebGL = function (renderer)
 {
-    if (this.dirty)
-    {
-        //this.resolution = 1//renderer.resolution;
-
-        this.updateText();
-    }
+    this.updateText(true);
 
     Sprite.prototype.renderWebGL.call(this, renderer);
 };
@@ -371,10 +363,7 @@ Text.prototype.renderWebGL = function (renderer)
  */
 Text.prototype._renderCanvas = function (renderer)
 {
-    if (this.dirty)
-    {
-        this.updateText();
-    }
+    this.updateText(true);
 
     Sprite.prototype._renderCanvas.call(this, renderer);
 };
@@ -537,10 +526,7 @@ Text.prototype.wordWrap = function (text)
  */
 Text.prototype.getBounds = function (matrix)
 {
-    if (this.dirty)
-    {
-        this.updateText();
-    }
+    this.updateText(true);
 
     return Sprite.prototype.getBounds.call(this, matrix);
 };
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -57,14 +57,13 @@ function Text(text, style, resolution)
      * @private
      */
     this._style = null;
-
     /**
-     * Last rendered version number of the style object.
+     * Private listener to track style changes.
      *
      * @member {Function}
      * @private
      */
-    this._styleVersion = null;
+    this._styleListener = null;
 
     var texture = Texture.fromCanvas(this.canvas);
     texture.trim = new math.Rectangle();
@@ -93,7 +92,11 @@ Object.defineProperties(Text.prototype, {
     width: {
         get: function ()
         {
-            this.updateText(true);
+            if (this.dirty)
+            {
+                this.updateText();
+            }
+
             return this.scale.x * this._texture._frame.width;
         },
         set: function (value)
@@ -112,7 +115,11 @@ Object.defineProperties(Text.prototype, {
     height: {
         get: function ()
         {
-            this.updateText(true);
+            if (this.dirty)
+            {
+                this.updateText();
+            }
+
             return  this.scale.y * this._texture._frame.height;
         },
         set: function (value)
@@ -135,6 +142,10 @@ Object.defineProperties(Text.prototype, {
         },
         set: function (style)
         {
+            if (this._style) {
+                this._style.off(CONST.TEXT_STYLE_CHANGED, this._onStyleChange, this);
+            }
+
             style = style || {};
             if (style instanceof TextStyle)
             {
@@ -143,7 +154,8 @@ Object.defineProperties(Text.prototype, {
             {
                 this._style = new TextStyle(style);
             }
-            this._styleVersion = -1;  // Force it to re-render and grab the version number.
+            this._style.on(CONST.TEXT_STYLE_CHANGED, this._onStyleChange, this);
+            this.dirty = true;
         }
     },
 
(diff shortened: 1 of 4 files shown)


src/core/text/TextStyle.js
```diff
@@ -0,0 +1,315 @@
...
+        this[property] = properties[property];
+    }
+}
+
+TextStyle.prototype = Object.create(EventEmitter.prototype);
+TextStyle.prototype.constructor = TextStyle;
+module.exports = TextStyle;
+
+// Default settings. Explained in the constructor.
+TextStyle.prototype._defaults = {
```
bigtimebuddy: Consider making this a static variable `TextStyle.DEFAULTS`, so this object isn't created whenever you making a new TextStyle object.

bigtimebuddy:
Great work :+1:. This is so much better. I'm going to leave a few specific code comments.

src/core/text/TextStyle.js
andrewstart: Stylistically, that might be nice, but objects defined on the class prototype merely makes new instances of a class with a reference to that object - the object itself is not duplicated for each new instance.

src/core/text/Text.js
```diff
@@ -175,42 +143,17 @@ Object.defineProperties(Text.prototype, {
         set: function (style)
         {
```
bigtimebuddy: The event listener should be cleared each time so that they don't continue to accumulate if the style is set repeatedly: `if (this._style) this._style.removeListener(this._styleListener);`

src/core/text/TextStyle.js
```diff
@@ -0,0 +1,315 @@
...
+TextStyle.prototype.clone = function ()
+{
+    var clonedProperties = {};
+    for (var key in this._defaults)
+    {
+        clonedProperties[key] = this[key];
+    }
+    return new TextStyle(clonedProperties);
+};
+
```
bigtimebuddy: How do you feel about creating

src/core/text/TextStyle.js
bigtimebuddy: Oh yah, makes sense. Thanks @andrewstart

src/core/text/TextStyle.js
```diff
@@ -0,0 +1,315 @@
...
+            }
+        }
+    },
+    fill: {
+        get: function ()
+        {
+            return this._fill;
+        }, set: function (fill)
+        {
+            var outputColor = typeof fill === 'number' ? utils.hex2string(fill) : fill;
```
bigtimebuddy: You _could_ create that local function somewhere here (`convertColor`?) from the first PR. Could keep from having to repeat the same ternary operator a few times.

src/core/text/TextStyle.js
```diff
@@ -0,0 +1,315 @@
...
+    font: {
+        get: function ()
+        {
+            return this._font;
+        }, set: function (font)
+        {
+            if (this._font !== font)
+            {
+                this._font = font;
+                this.emit('styleChanged');
```
bigtimebuddy: Maybe create a local constant at the top of the page `const STYLE_CHANGED = 'styleChanged';` so we don't have to repeat this literal string a dozen times. Make it easier to change. I'm also cool with `'changed'` to keep it simple.

src/core/text/TextStyle.js
AMorporkian: I did have that before I made the ternary, it just felt awkward having a lone hanging function. I'll put it back in.

src/core/text/TextStyle.js
AMorporkian: If we're going to go that route, I think it'd be better to put it in `CONST`, since it's used in two files.

src/core/text/Text.js
```diff
@@ -175,42 +143,17 @@ Object.defineProperties(Text.prototype, {
...
-
-            style.lineJoin = style.lineJoin || 'miter';
-            style.miterLimit = style.miterLimit || 10;
-
-            this._style = style;
-            this.dirty = true;
+            this._styleListener = this._style.on('styleChanged', function ()
+            {
+                this.dirty = true;
+            }.bind(this));
```
englercj: Couple things here:

1. `.on()` returns the object you are binding to, not the listener function
2. `.on()` can take 3 params, where the third is the context which is much faster on calls than `.bind()`

I would do this instead:

```js
Text.prototype._onStyleChange = function ()
{
    this.dirty = true;
};

// ...

this._style.on('styleChanged', this._onStyleChange, this);

// ...

this._style.off('styleChanged', this._onStyleChange, this);
```

src/core/text/TextStyle.js
```diff
@@ -0,0 +1,315 @@
...
+/**
+ * A TextStyle Object decorates a Text Object. It acts as an event emitter, and can be shared between
+ * multiple Text objects.
+ *
+ * @class
+ * @extends EventEmitter
+ * @memberof PIXI
+ * @param [style] {object} The style parameters
+ * @param [style.font='bold 20pt Arial'] {string} The style and size of the font
+ * @param [style.fill='black'] {String|Number} A canvas fillstyle that will be used on the text e.g 'red', '#00FF00'
```
englercj: Use lower-case names of types `String` -> `string`.

src/core/text/TextStyle.js
```diff
@@ -0,0 +1,315 @@
...
+ *      or increase the spikiness of rendered text.
+ */
+function TextStyle(style)
+{
+    EventEmitter.call(this);
+    var properties = Object.assign({}, this._defaults, style);
+    for (var property in properties)
+    {
+        this[property] = properties[property];
+    }
```
englercj: Instead of `Object.assign()` into a new object, then copy into `this` could you just assign into `this`?

```js
Object.assign(this, this._defaults, style);
```

AMorporkian: Good call, I don't know how I missed that.

src/core/text/TextStyle.js
AMorporkian: Will do. I just copied that from the previous Text.js, but I'll correct it.

englercj:
Well, looks good to me. Unless @GoodBoyDigital has any thoughts I will merge this soon.

GoodBoyDigital:
Looks good to me, great work everyone! 👍

AMorporkian:
Actually, hold up a few, I can simplify the reset method.

bigtimebuddy:
@AMorporkian I did some QA this morning on this feature. The only bug I found was that you should set `dirty` to be `true` whenever the Text `style` setter is called. This will make it possible to hot-swap styles. Here's the text example I used:

```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8">
        <title>text-style</title>
    </head>
    <body>
        <script src="text-style/bin/pixi.min.js"></script>
        <script>
                backgroundColor: 0xffffff,
                antialias: true
            });
            var style = new PIXI.TextStyle();
            var style2 = new PIXI.TextStyle({
                font: 'bold 14pt Arial',
                fill: "#ff0000" 
            });
            var text = new PIXI.Text("Something", style);
            stage.addChild(text);
            text.interactive = true;
            text.on('click', function()
            {
                text.style = text.style === style ? style2 : style;
                text.dirty = true; // <-- should not be required
            });

            function update() {
                renderer.render(stage);
            }
            update();
        </script>
    </body>
</html>
```

AMorporkian:
@bigtimebuddy Good catch. Resolved.

ivanpopelyshev:
Ok, guys, I dont like EventEmitter because we can leak Text's with it. May be its better to use versions, like i do in https://github.com/pixijs/pixi.js/pull/2465 ? Every time you change style, increase its version, and in text remember previous version of that style. In text updateTransform() check if version has been changed.

AMorporkian:
What do you mean we could leak Text's with it?

ivanpopelyshev:
Easy: if you share text style between many objects and  forget to destroy them, they all will be accumulated as "listeners".

AMorporkian:
I'd be fine switching it to versioned if there are no objections.

englercj:
> Easy: if you share text style between many objects and forget to destroy them, they all will be accumulated as "listeners".

I'm not sure I follow this logic, it literally applies to everything. If you don't destroy it, it will leak. Why is this a special case?

bigtimebuddy:
I agree with @englercj, I think converting to versions is optimizing for something that isn't a problem.

ivanpopelyshev:
> I'm not sure I follow this logic, it literally applies to everything. If you don't destroy it, it will leak. Why is this a special case?

Right now destroy() API is pain. The less type of objects need to be managed by developer, the better.

AMorporkian:
Should I revert?

ivanpopelyshev:
@bigtimebuddy @englercj There's PR #2465 which works on versions, and all new objects in my branch v4.1.0 are using versions pattern and not emitters.

bigtimebuddy:
There's a larger discussion here about embracing a new design pattern. My preference is to merge it with the EventEmitter (both @englercj, @GoodBoyDigital  and I were cool with) AND it is currently consistent with the rest of Pixi.js. Then we can make an additional PR to move over to versions pattern if that makes sense moving forward.

AMorporkian:
Copy that, I'll revert for now and keep a local branch at the ready in case we switch to versioning.

bigtimebuddy:
@AMorporkian This was inadvertently closed because the **dev-v4.0.0** branch was deleted. All PRs should be going to **dev** anyways. Can you make the request again.

PR #2474, opened 2016-04-08
Stop ticker from being able to report a negative deltaTime

themoonrat:
When you start a stopped timer, the very first timeDelta can sometimes be negative. This PR just adds a Math.max to ensure that in this case it would report 0 rather than a negative number.

Example: https://jsfiddle.net/mL4qx95n/9/
Open up the console. It will log when the value of the timeDelta is negative.


GoodBoyDigital:
nice! 👍

PR #2476, opened 2016-04-09
Adding support for pointer events

themoonrat:
A reworked version of the PR: https://github.com/pixijs/pixi.js/pull/2195

This also adds support for the pointer events, but there are a couple of differences

1. The default behavior is that this is an additional event to listen to. The other PR would stop all mouse and touch events from being emitted if the browser supported pointer events, which is breaking behavior

2. This PR allows the 'normalization' of interaction events via the normalizeEvents flag so that only pointer events are ever emitted.
 - Browsers that support pointer events emit them as normal.
 - Browsers that support touch events have these touch events converted to pointer events.
 - All other browsers have mouse events converted to pointer events.
In all cases, only the single pointer event is emitted. So an old Android browser that emits a touch event and a mouse down event will have their touch event converted to a pointer event, and the mouse event ignored.

I have a wide variety of mobile devices at work that I have tested this on - windows/android/ios :)


niclaslindberg:
Hi! What's the status on this PR? Will it make it in to pixijs dev?

src/interaction/InteractionManager.js
```diff
@@ -341,6 +459,56 @@ InteractionManager.prototype.dispatchEvent = function ( displayObject, eventStri
...
+ * This allows a developer to just listen for emitted pointer events on interactive sprites
+ *
+ * @param eventString {string} the name of the event (e.g, mousedown)
+ * @private
+ */
+InteractionManager.prototype.normalizeEventString = function ( eventString )
+{
+    if (!this.supportsPointerEvents)
+    {
+        if (this.supportsTouchEvents && Device.any)
```
andrewstart: is Device.any needed? I would think that anything that is outputting touch events you'd want to convert to pointer events, even if it isn't a recognized Android or iOS device. That would also allow you to turn this function into a straight dictionary check instead of switch statements.

andrewstart:
I know we don't have it for touch events, but could you add support for multi-touch, since different pointer identifiers is part of the Pointer spec?

themoonrat:
I've slightly changed strategy on this for normalisation which I'll update this with... instead of converting a mousedown event string name to pointerdown, I add the pointerdown function callback to the mouse event itself. This allows the normalisation to occur without effecting regular events; normalisation can be switched on without effecting v3 ported games.

The Device.all is using the mobile detection lib, so it's referring to allowing for all mobile devices. The reason it's not any browser that supports touch is because of chrome on touch screen devices like the ms surface. You want to convert the mouse in that case, not the touch (had this as a bug at work before!)

I'll take a look at multi pointer events. I'll pr my latest changes this morning so that it doesn't delay a potential merge, but multi pointer definitely needs to be there. Hopefully i can use similar code to multi touch

themoonrat:
@andrewstart could you take a new look please? 'Normalized' mouse and touch events now get all of the properties that a real pointer event contains, including pointer identifiers.

src/interaction/InteractionManager.js
themoonrat: Replied below. Device.any is essentially 'any mobile device'. It can't be _any_ device that supports touch, because that would cover using Chrome on a touch screen laptop... and then any mouse/trackpad clicks wouldn't work! mouse events are still triggered when the screen is touched in this scenario however, so you're not losing that functionality.

andrewstart:
I've never used Chrome on a touch-screen enabled computer, or a mouse on Android, but is it possible for someone to trigger both touch events and mouse events in one session?

themoonrat:
Yes. Last time I tested, with a surface laptop using chrome, clicking the mouse created a mouse event; touching the screen created a mouse event and a touch event.

andrewstart:
It looks like your code normalizes mouse events or touch events, but won't do both at the same time, which would be an issue if you wanted just pointer events on a Surface in Chrome.

themoonrat:
It picks one or the other to normalise, but it leaves the original events to fire as normal. So if you want the normalised event, just listen to the pointer events. If you want to know the touch or mouse events, they are there still too if the browser emits them.

andrewstart:
Right - if you want normalized events, and it is normalizing mouse events, and then your user touches the screen of their laptop, then you don't get an event.

themoonrat:
That's solved for desktops; it can normalise both at the same time. 
Mobiles I still look for just touch if it has it; primarily because of older android aosp browsers that emit both mouse and touch events. I want the 'normalised' events to be sane, and getting 2 of everything isn't sane.

andrewstart:
Psh, sanity is overrated. That was the last of my concerns, so this looks good to me. I can see it fitting in nicely in my current projects.

englercj:
Can you rebase this onto the latest `dev` branch please?

themoonrat:
Closing with new PR in favour of rebasing: https://github.com/pixijs/pixi.js/pull/2997

PR #2479, opened 2016-04-09
moveWhenInside property changed to true by default

themoonrat:
As warned by inline JSDOCs that the behaviour would change in a future version... v4 seems a good moment to make this change.


englercj:
Probably not going to do this for v4, and I feel like I need to rewrite interactions in v5 anyway. Thanks for this though!

PR #2480, opened 2016-04-09
JSHint fixes

AMorporkian:
This resolves every JSHint error/warning, save for three files. 

There is a function in bezierCurveTo.js that has 9 parameters, and the JSHint settings don't like that one bit. I didn't want to change the signature on my own, if indeed it should be changed at all.

Additionally, TextureUvs.js has a bunch of properties that aren't camel cased. Interestingly, 4 of them don't seem to be used at all and could probably be removed (`this.xy0_uint32` through `this.xy3_uint32`), leaving `this.uvs_uint32` as the only one that would need changed, which would also resolve the issue in SpriteRenderer.js. 

Thoughts?

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -1,8 +1,6 @@
 var math = require('../math'),
     utils = require('../utils'),
-    DisplayObject = require('./DisplayObject'),
-    RenderTexture = require('../textures/RenderTexture'),
-    _tempMatrix = new math.Matrix();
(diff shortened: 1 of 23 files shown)


GoodBoyDigital:
Hi @AMorporkian, Thank you for the PR - this branch is no longer in use as it has been merged into Dev. I need to unpublish the v4.0.0 branch, sorry for the confusion!

PR #2481, opened 2016-04-09
Forcing the loss of the WebGL context after detecting WebGL capabilities

themoonrat:
As discussed here: https://github.com/pixijs/pixi.js/issues/2233
No-one actually submitted a PR for that so here it is!


GoodBoyDigital:
super! 👍

PR #2475, opened 2016-04-08
Updated and fixed JSDocs

AMorporkian:
There were a lot of problems with the JSDocs; missing parameters, parameters that didn't exist, references to types that didn't exist, that sort of thing. I fixed most or all of them. Some parameters may need better explanations, but I did the best I could on that end.

PR #2482, opened 2016-04-09
Updated and fixed JSDocs

AMorporkian:
Recreation of #2475 

There were a lot of problems with the JSDocs; missing parameters, parameters that didn't exist, references to types that didn't exist, that sort of thing. I fixed most or all of them. Some parameters may need better explanations, but I did the best I could on that end.


AMorporkian:
Hold off on merging this, I just discovered a bunch more problems.

AMorporkian:
Ok, I'd appreciate it if someone else could look this over. I corrected everywhere I could, but I feel like there's a chance I might have gotten a few of the params wrong, particularly when it comes `core` stuff.

AMorporkian:
Also, I feel like all the math shapes should inherit from some common object, just for param simplicity if nothing else.

themoonrat:
I notice in some of the changes that you changed the JSDocs from
PIXI.<ClassName> to 
core.<ClassName>

For example:
 - * @param renderTexture {PIXI.RenderTexture}	
+ * @param displayObject {core.DisplayObject} the object to be rendered

I think the PIXI variant is better; it's consistent with the rest of pixi, and it works in WebStorm! It also makes more sense in the generated JSDocs.

Speaking of which, creating the docs no longer works with a number of "unable to parse a tag's type expression" errors.

AMorporkian:
@themoonrat, I actually changed those because WebStorm was complaining.

themoonrat:
hmmm, yeah, there are some places I can see that whatever way you choose, WebStorm will complain.

But it must be PIXI.DisplayObject or it harms the JSDoc html output

GoodBoyDigital:
hello! I agree with @themoonrat, priority should be a correct JSDoc html output. great pr!

PR #2484, opened 2016-04-10
JSDoc for interactiveTarget

themoonrat:
Added JSDoc typing to the interactiveTarget.js file
Moved hitArea out as a property from DisplayObject, since it's an interactive only property
Allowed JSDoc to now show within the DisplayObject page that there can be the interactiveTarget mixin. Previously there was no way of knowing these properties existed as interactiveTarget wasn't generated anywhere in the documentation at all.

DisplayObject now shows:
it1

And clicking on that link takes you to:
it2

### DIFF
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -73,13 +73,6 @@ function DisplayObject()
     this.filterArea = null;
 
     /**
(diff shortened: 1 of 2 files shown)


GoodBoyDigital:
👍

PR #2488, opened 2016-04-11
TextStyle, a more flexible and fluent interface for modifying text styling

AMorporkian:
Recreation of #2468.

This commit introduces PIXI.TextStyle. The API for PIXI.Text is left mostly unchanged. TextStyle extends EventEmitter, and emits a styleChanged event every time it is changed. Text hooks onto that event and marks Text.dirty = true.

Additionally, TextStyle exposes TextStyle.clone(), which allows an end user to clone a previously defined text style, and TextStyle.reset(), which allows resetting to default parameters.


GoodBoyDigital:
nice work sir! 👍

PR #2493, opened 2016-04-12
Fixed linting and async build problems.

bigtimebuddy:


src/core/graphics/webgl/utils/buildPoly.js
```diff
@@ -51,8 +51,6 @@ var buildPoly = function (graphicsData, webGLData)
 
         var vertPos = verts.length / 6;
 
-        var i = 0;
```
englercj: Instead of relying on function-scope hoisting for the above `var i` can we just move its decl to the top of the function for clarity?

englercj:
Give it some time for review before merging your own PR. I generally follow the rule of thumb that I don't merge my own PRs at all, or at the very least not until I get a :+1: from someone else.

bigtimebuddy:
Sure, makes sense. Sorry about that.

PR #2494, opened 2016-04-12
Dev

mattstyles:
Not sure if there was a reason why Pixi needed to spill to global, the browserify standalone UMD handles this nicely, is there a separate reason?

Also killed the unnecessary, and potentially harmful, assignment in there. Deprecation monkey patches core rather than exporting an object to patch over so the `Object.assign` is just waiting for someone to export something stupid from `deprecation.js` and bring the whole house down.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -25,6 +25,3 @@ core.loader = new core.loaders.Loader();
 
 // mixin the deprecation features.
 Object.assign(core, require('./deprecation'));
-
-// Always export pixi globally.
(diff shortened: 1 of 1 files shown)


englercj:
>  is there a separate reason?

Yes, we want the global to *always* be created; even in AMD/CommonJS environments. UMD doesn't do that. Some plugins rely on it.

mattstyles:
Thanks for the clarification Chad, gotcha!

PR #2502, opened 2016-04-14
Workaround for Atom's unsafe-eval error in pixi-gl-core

bigtimebuddy:
_**NOTE:** The [PR found here](https://github.com/pixijs/pixi-gl-core/pull/6) to **pixi-gl-core** needs to be considered first before this change can be merged._

### Overview

This explicitly states which of Browserify's builtins polyfills to use. The advantage is that it _does not_ include the `vm` polyfill, which makes it easier to workaround the unsafe-eval error in Atom by attempting to require `vm` and then falling-back to use `new Function` instead.

### DIFF
--- a/gulp/util/bundle.js
+++ b/gulp/util/bundle.js
@@ -62,6 +62,7 @@ function createBundler(args) {
     args = args || {};
     args.debug = false;
     args.standalone = 'PIXI';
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Found another solution which doesn't involve any changes.

PR #2507, opened 2016-04-14
use gl-matrix and mat4

ivanpopelyshev:
One of commits that prepares us for upcoming v4.1

1. projectionMatrix and transformMatrix in all shaders have type mat4
2. other uniforms can use mat3
3. its possible to use mat3 as uniform value in filters

I didnt test it on all filters yet, only on pixi examples

Bunnymark shows the same FPS for me.

So far I expose ONLY glMat.mat4, we dont need other stuff yet


ivanpopelyshev:
Matrix.copy was working like copyTo, while other objects copy worked like copyFrom
I added deprecation warning there

PR #2511, opened 2016-04-16
Dev 4.1.0

gentwolf:


ivanpopelyshev:
Just... Why?

englercj:
Probably a bot

photonstorm:
It's shockingly easy to do this by mistake when taking your own fork of a repo. Personally I think GitHub should find a way to prevent it (I close several of these a month personally)

PR #2519, opened 2016-04-18
Update Container.destroy documentation

renarsvilnis:
Updated the `Container.destroy` documentation very slightly to make it clear what the function will do if optional parameter passed.


bigtimebuddy:
Pull requests should be made to **dev** and not **master* branch. Also, we have recently updated the destroy API, see #2592  for more information.

PR #2523, opened 2016-04-21
Documentation update in Text.js for style.fill

evilangelist:
The style.fill property supports a CanvasGradient (https://developer.mozilla.org/en-US/docs/Web/API/CanvasGradient) so it would be good if the documentation reflected that.


bigtimebuddy:
👍 Good catch. What do you think about adding link in the description to the MDN API reference you provided? Might be helpful for developers.

evilangelist:
@bigtimebuddy Good idea, the link to MDN is now there. I've expanded it a bit more to include CanvasPattern since that's also an option.

bigtimebuddy:
Thanks for your help @evilangelist!

bigtimebuddy:
Sorry my bad, @evilangelist can you make this PR to Dev instead of master?

PR #2525, opened 2016-04-22
Mobile accessability fix

GoodBoyDigital:
Added a touch 'hook' div to activate the accessibility layer if a screen reader is being used on mobile.

### DIFF
--- a/src/accessibility/AccessibilityManager.js
+++ b/src/accessibility/AccessibilityManager.js
@@ -93,6 +93,7 @@ function AccessibilityManager(renderer)
      * @private
      */
    	this.isActive = false;
+   	this.isMobileAccessabillity = false;
 
    	// let listen for tab.. once pressed we can fire up and show the accessibility layer
    	window.addEventListener('keydown', this._onKeyDown, false);
@@ -116,6 +117,7 @@ AccessibilityManager.prototype.createTouchHook = function()
 
     hookDiv.addEventListener('focus', function(){
 
+    	this.isMobileAccessabillity = true;
     	this.activate();
     	document.body.removeChild(hookDiv);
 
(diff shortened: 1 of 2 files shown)


src/accessibility/AccessibilityManager.js
```diff
@@ -124,6 +152,7 @@ AccessibilityManager.prototype.activate = function()
  */
 AccessibilityManager.prototype.deactivate = function()
 {
+	return;
```
englercj: ?

src/accessibility/AccessibilityManager.js
```diff
@@ -1,4 +1,5 @@
 var core = require('../core');
+var  Device = require('ismobilejs');
```
englercj: New dependency? Probably should add it to `package.json`.

src/accessibility/AccessibilityManager.js
GoodBoyDigital: good catch! fixed now..

GoodBoyDigital:
@andyish is there a way to check if a mobile device has a screen reader enabled?

src/accessibility/AccessibilityManager.js
GoodBoyDigital: yup its added to the package.json

andyish:
@GoodBoyDigital No, not that we're aware of. Could leave it to the developer thats using it check. 

With accessibility in games on mobile devices we have a setting to enable it

src/accessibility/AccessibilityManager.js
englercj: Now I see it in dev branch package.json, ignore me :P

PR #2528, opened 2016-04-24
Easy addition of gradients on text

themoonrat:
Instead of providing a single string value for a fill colour, you can provide an array. This array is then turned into the Canvas Text gradient stops.

I made this change over other ways to provide gradient text because
1. It's a re-use of an already existing fill setting.
2. The vertical gradient works for each individual line on word wrapped text. Previous examples I've seen just extend the one gradient over all of the lines.

On the downside, you can't create complex gradients; the examples have a straight linear or vertical, with evenly distributed stop points. However, there's no reason why you can't create your own complex one and pass it through, as documentation added via PR: https://github.com/pixijs/pixi.js/pull/2523

jsfiddle example: https://jsfiddle.net/themoonrat/d3roaLmf/


bigtimebuddy:
I'm not crazy about this approach for adding gradients. I think it's piling on too much functionality to the Text class. Why not create a separate API which simplifies CanvasGradient and provides some nice convenience? This could be some static methods (like below) or by inheriting the CanvasGradient class (e.g. `new PIXI.TextGradient`).
 
```js
var text = new PIXI.Text('blah', {
     fill: PIXI.TextGradient.linear(['#fff', '#fcc'], 0) // 2nd arg: angle? isVerticle?
});
```

themoonrat:
The problem with setting the gradient as a fill style is with multi line text. That one gradient will cover all lines, so each line would look different. I have a requirement that each line should look the same, with the same looking gradient. To do that I need to know how many lines of text there are to repeat the gradient that many times, and I only know that during updateText. 

That's the rationale behind why I went with this method anyway :)

bigtimebuddy:
Why not create a getter to know the numbers of lines?

themoonrat:
So, for my scenario, the games I make need to be translated into 20 odd different languages, so text could go from 1 line in Chinese to 3 for German/Greek/Russian. I don't know how many lines of text each translation will actually take up until it's already on screen. 

As I understand your question, I could have a getter for the number of lines, but that would be gaining knowledge of the text after the Text canvas has already been drawn to; a 2nd run through the updateText would then be required with a gradient set just for that number of lines which would be inefficient?

rkrasky:
+1

englercj:
Looks good to me. :+1: 

Note to merger: Please use the "Squash and Merge" option (Merge Pull Request, then select down arrow on right side of confirm button).

rkrasky:
Could someone merge it please?

bigtimebuddy:
@themoonrat, Great work, tested this out and it works well. One small change before merge, could you make `style.fill` also accept an array of uints? This would be consistent with the changes in TextStyle, to allow for broader support for colors: e.g. `"#FF0000"` or `0xFF0000`

themoonrat:
Done :boom:

bigtimebuddy:
👍

PR #2532, opened 2016-04-26
Fixes to TextStyle

themoonrat:
To get this previous PR working https://github.com/pixijs/pixi.js/pull/2488 
As @AMorporkian himself worked out, there were a couple of style settings that weren't being set in TextStyle that were trying to be used in Text, thus causing the display issues.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -1,8 +1,8 @@
 var Sprite = require('../sprites/Sprite'),
     Texture = require('../textures/Texture'),
     math = require('../math'),
-    utils = require('../utils'),
-    CONST = require('../const');
+    CONST = require('../const'),
+    TextStyle = require('./TextStyle');
 
 /**
  * A Text Object will create a line or multiple lines of text. To split a line you can use '\n' in your text string,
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
So glad this is back, thanks @themoonrat! 3rd time's a charm. I'm going to QA this today.

themoonrat:
np, it's so useful :)

As a note, it includes the extra jsdoccing of https://github.com/pixijs/pixi.js/pull/2523 - not intentional, I have my own build of pixi that has extra stuff that I then pruned back for this PR, and missed that change to keep the PR as slim as possible. Useful to have in there, but I can remove if you wish.

bigtimebuddy:
@themoonrat That's fine about the CanvasGradient note for fill, thanks for pointing it out.

AMorporkian:
Thanks for fixing it up @themoonrat, I really appreciate it. Damned real life getting in the way of important stuff like Pixi.

bigtimebuddy:
Looks pretty good to me. @GoodBoyDigital, you had some issues with the original PR. Can you test to see if you're still getting the same issues? Here's my test for TextStyle:

### Text Example
```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8">
        <title>text-style</title>
    </head>
    <body>
        <script src="text-style/bin/pixi.min.js"></script>
        <script>
                backgroundColor: 0xffffff
            });

            // Swap out TextStyle objects on the fly
            function hotSwappable() {
                var style = new PIXI.TextStyle();
                var style2 = new PIXI.TextStyle({
                    font: 'bold 20pt Arial',
                    fill: "#ff0000" 
                });
                var text = new PIXI.Text("Hot-swap styles", style);
                text.interactive = true;
                text.on('mouseover', function() {
                    text.style = style2;
                });
                text.on('mouseout', function() {
                    text.style = style;
                });
                return text;
            }

            // Swap out TextStyle objects on the fly
            function shareStyles() {
                var style = new PIXI.TextStyle({
                    font: '20pt monospace',
                    fill: "#009900",
                    dropShadow: true,
                    dropShadowDistance: 2,
                    dropShadowBlur: 2
                });
                var text = new PIXI.Text("Share styles", style);
                var text2 = new PIXI.Text("Share styles copy", text.style);

                text2.y += 30;
                return text2;
            }

            // Use the old legacy way of setting styles
            function objectBased() {
                var text = new PIXI.Text("Legacy styles", {
                    font: '12pt Times New Roman',
                    fill: "#000099",
                    letterSpacing: 5
                });
                text.y += 70;
                text.x += 20;
                text.rotation = Math.PI / 2;
                return text;
            }

            // Use the old legacy way of setting styles
            function wrapping() {
                var text = new PIXI.Text("Can I wrap some words please?", {
                    font: '12pt Verdana',
                    fill: "black",
                    wordWrap: true,
                    wordWrapWidth: 120,
                    lineHeight: 40
                });
                text.alpha = 0.5;
                text.y += 60;
                text.x += 30;
                return text;
            }

            // Use the old legacy way of setting styles
            function rightAlign() {
                var style = new PIXI.TextStyle({
                    font: 'italic 2em sans-serif',
                    fill: 0x009900,
                    wordWrap: true,
                    wordWrapWidth: 120,
                    align: 'right',
                });
                var text = new PIXI.Text("Wrap words on the right side.", style);
                text.y += 60;
                text.x += 200;
                return text;
            }

            stage.addChild(
                hotSwappable(),
                objectBased(),
                shareStyles(),
                wrapping(),
                rightAlign()
            );

            function update() {
                renderer.render(stage);
            }
            update();
        </script>
    </body>
</html>
```
### Rendering
screen shot 2016-04-26 at 12 55 29 pm

endel:
Nice work here guys. Congrats.

One thing I'd like to add on this discussion is that this feature could be completely framework-agnostic. I've extracted a tiny bit of what was implemented in PIXI v3 to create a three.js library to deal with 2d text: https://github.com/gamestdio/three-text2d

As you can see, the [canvas](https://github.com/gamestdio/three-text2d/blob/master/src/CanvasText.js) part could be extracted into its own project and be used as a dependency on other frameworks, such as PIXI, threejs, and any other. My implementation doesn't cover word-wrapping and complex stuff, but it would be great if we could benefit from a single implementation.

Cheers!

bigtimebuddy:
@endel Interesting idea! I think your idea is outside the scope of the PR, but it would be cool if you start a new issue. There would need to be more discussion about how the decoupling might happen and if it seems appropriate moving forward for Pixi.

bigtimebuddy:
Looking for a merge here. @englercj or @GoodBoyDigital, thoughts?

src/core/text/Text.js
```diff
@@ -246,11 +183,14 @@ Object.defineProperties(Text.prototype, {
 
 /**
  * Renders text and updates it when needed
- *
+ * @param respectDirty {boolean} Whether to abort updating the text if the Text isn't dirty and the function is called.
  * @private
  */
-Text.prototype.updateText = function ()
+Text.prototype.updateText = function (respectDirty)
```
englercj: Looks like most calls to `updateText` pass `true`. Would it make more sense to have this be something like `skipDirtyCheck` so that the default respects dirty, but you can override that behavior in rare cases?

englercj:
Made one last comment, but aside from that it looks good to me!

GoodBoyDigital:
Looks good to me too! HIT it 👍

src/core/text/Text.js
AMorporkian: It's been a while since the original discussion, but I believe I added that in to preserve backwards compatibility at another dev's request.

bigtimebuddy: I believe the expectation for developers with `updateText()` is that it forces a redraw regardless of `dirty`. I think this is a much more typical use-case. Without changing the call signature of `updateText`, another simpler approach is to make an internal update method which is used instead of the `updateText(true)`:

```js
Text.prototype.safeUpdateText = function() {
    if (this.dirty) {
        this.updateText();
    }
};
```

englercj: Rather not put the dirty check everywhere. Explanation of user-land expecting it to force seems fine.

PR #2473, opened 2016-04-08
Container.destroy() api unification

AMorporkian:
As brought up repeatedly (#2392, #2457, #2418), the API for the various descendants of `Container` have different method signatures for `.destroy()`, leading to very confusing results. This unifies the API signatures of the various `.destroy()` methods.

Issue #2457, opened 2016-04-05
Sprite.prototype.destroy not destroying children


Kouty:
**Container** class has a method:  **destroy(destroyChildren)** which allows the caller to destroy also the children of the container.
A **Sprite** is also a Container, but it's destroy signature is:
**destroy(destroyTexture, destroyBaseTexture)** which does not allow its children to be destroyed. This is not coherent with it's super class (Container) behavior. In addition, calling container.destroy(true) do not guarantee that all children are destroyed, since children of Sprite will not be destroyed. It's an example of violating Liskov substitution principle.

This is a structural cause of memory leaks in PIXI, since there is not an easy way to remove a container and all it's children, so that browser memory is freed accordingly.

Moreover: Sprite.destroy(destroyTexture, destroyBaseTexture) it's cofunsing. What does sprite.destroy(false, true) mean? I have destroyed the Sprite, the baseTexture, but not the texture. But I think that texture, having it's baseTexture destroyed, is not valid anymore.

Probably a good solution is:
- enhance Container with destroy(destroyChildren, destroyTextures), which let you decide if letting or not all textures available for reuse.
- Change Texture to have a no parameter destroy(), that also destroys its base texture.
- make Sprite.destroy to have the same signature of Container: destroy(destroyChildren, destroyTextures)

This issue is related to:
Container#destroy will pass down the first parameter regardless if the child is a Sprite instance, is this right? #2392

AMorporkian:
There is an attempt to fix this in #2418. It looks like it will likely be fixed in v4.0.0.

GoodBoyDigital:
closing as this is fixed in v4 👍

Issue #2392, opened 2016-03-13
Container#destroy will pass down the first parameter regardless if the child is a Sprite instance, is this right?


inetfuture:
```js
/**
 * Destroys the container
 * @param [destroyChildren=false] {boolean} if set to true, all the children will have their destroy method called as well
 */
Container.prototype.destroy = function (destroyChildren)
{
    DisplayObject.prototype.destroy.call(this);

    if (destroyChildren)
    {
        for (var i = 0, j = this.children.length; i < j; ++i)
        {
            this.children[i].destroy(destroyChildren);
        }
    }

    this.removeChildren();

    this.children = null;
};
```

When call `container.destroy(true)` and if one of the children is a `Sprite`, its `destroy` method will also be called with `true`, which means destroy the current texture of the sprite as well, is it intentional?

I'm developing a single page app, where when route changed, I will create and destroy the stage and renderer accordingly. When I use `stage.destroy(true)` and recreate it later when user navigates back, sprites didn't show, there is an error `Cannot read property 'hasLoaded' of null` of Sprite.js, I'm guessing it's because their texture cache has been destroyed. So I put an empty function on sprite's `destroy` after I created it, then it can show up again.

ivanpopelyshev:
Yeah, that's a problem :)

inetfuture:
> Yeah, that's a problem :)

So what should we do?

mike-tomko-codebaby:
Also encountering this issue in a single page app experience.

While I agree ```Container.destroy(true)``` shouldn't cause child sprites to destroy their textures by default, I think the underlying issue is that the ```Texture.destroy``` method doesn't remove the texture from the ```textureCache```, so there's a destroyed texture still in cache. This means that when you try to recreate the texture with the same imageId, it sees that it's already cached and tries to use the already-destroyed texture instance.

```BaseTexture``` properly deletes itself from the ```baseTextureCache``` in its ```destroy()``` method, which is what ```Texture.destroy()``` should be doing.

I made a jsFiddle that represents what I think is the situation @inetfuture and I, are running into. https://jsfiddle.net/sjs2ye2u/2/ It renders a stage with a sprite, then after 5s it simulates a single page app page change by telling the stage to destroy, and then reinitializing the stage and sprite. As you'll see in the console, when it simulates the SPA page change, it throws the error @inetfuture references (at least on Chrome, different wording in Firefox).

andrewstart:
[...]

PR #2536, opened 2016-04-27
Container destruction API unification

AMorporkian:
Recreation of #2473 

As brought up repeatedly (#2392, #2457, #2418), the API for the various descendants of Container have different method signatures for .destroy(), leading to very confusing results. This unifies the API signatures of the various .destroy() methods.


bigtimebuddy:
As discussed before this is a breaking change. An alternative, non-breaking change, would be to have destroy options:

```js
// Current usage (this destroys everything)
container.destroy(true);

// The above would be short-hand for the 
// following options, all would default to false
container.destroy({
    children: true,
    texture: true,
    baseTexture: true
});
```

englercj:
@bigtimebuddy A breaking change in a major version is fine.

src/core/display/Container.js
```diff
@@ -606,17 +606,19 @@ Container.prototype.renderCanvas = function (renderer)
 
 /**
  * Destroys the container
- * @param [destroyChildren=false] {boolean} if set to true, all the children will have their destroy method called as well
+ * @param [destroyChildren=false] {boolean} Should it destroy any children?
+ * @param [destroyTexture=false] {boolean} Should it destroy the current texture of any children?
+ * @param [destroyBaseTexture=false] {boolean} Should it destroy the base texture?
  */
-Container.prototype.destroy = function (destroyChildren)
+Container.prototype.destroy = function (destroyChildren, destroyTexture, destroyBaseTexture)
```
englercj: While this works, it makes me uncomfortable because it makes container be knowledgable about how child objects are setup.

Instead using an object as the param means you can put any keys int he object you want and only the children need to be knowledgeable about them. See this comment: https://github.com/pixijs/pixi.js/pull/2536#issuecomment-215141153

bigtimebuddy:
Yah, what @englercj said.

marekventur:
I've stumbled across this a few times now as well. I'm happy to implement the suggested changes if @AMorporkian is away at the moment, I'd love to have this for v4.

bigtimebuddy:
@marekventur I'd say go for it. This would be great to get in.

AMorporkian:
It would probably be best if you did, I am indeed away.

englercj:
Closing in favor of a new PR, coming soon from @marekventur .

PR #2539, opened 2016-04-29
Text: Splitting font setting into it's various style components

themoonrat:
Instead of just setting font, and having to put the settings for bold / size etc. in the right order, these style settings are split up into the 5 style components, which are joined together automatically for you.
There is now:
* fontFamily
* fontSize
* fontStyle
* fontVariant
* fontWeight

Before any merge is done... I was wondering if the defaults should be changed. I have just taken the 'bold 20pt Arial' and taken those settings as the default. However, because a developer isn't forced to override all style settings for each font they want anymore....

1. I do not think 'bold' should be enabled by default.
2. I think that the default size should move from 'pt' to an equivilant 'px'. 'px' is easier to understand in a rendering to a canvas that is made up of pixels context.

Thoughts?


bigtimebuddy:
@themoonrat, you're really killing it with **PIXI.Text** this week. 

1. I agree
2. I'm in favor of 'px' and would never use 'pt' except for print CSS

How do you feel about a shorthand where `fontSize` can be a pixel integer as well? For instance, a value of `10` would be the same as `"10px"`. Any other measurements would need to be expressed as strings, e.g., `"1em"`, `"80%"`, `"12pt"`. I like how the TextStyle now support ints for fill, this would be consistent with that.

themoonrat:
Changes made :)

bigtimebuddy:
Could you add some `TextStyle.font` deprecation code to **src/deprecation.js** and parse the font string as the arguments mentioned here? This would provide better upgrading support for developers.

themoonrat:
Done.
Not pretty, what with the various guises and information that could or could not be there....but done.

bigtimebuddy:
I did some testing on this feature and it works well. Love the flexibility. Thanks for all the good work @themoonrat. 

@englercj, unless you have any concerns, I like this and think it's good to merge.

englercj:
Aside from the fact that it removes our build files, changes look good.

themoonrat:
bah, in the battle to not commit updated build files I won by removing them. D'oh. Fixing

englercj:
Looks good to me, squash up (git rebase -i) and I will merge it in!

themoonrat:
My git skills suck, can't get that rebase working, so submitted a new clean PR :)
https://github.com/pixijs/pixi.js/pull/2553

PR #2543, opened 2016-04-30
Updated the style.fill documentation to reflect gradient options

evilangelist:
@englercj I'm guessing the squash wasn't working because of the way I did the upstream merge, but I'm not 100% sure. Here it is again in a single commit.


bigtimebuddy:
Thanks! Good to merge @englercj

PR #2544, opened 2016-05-01
Add anchor to BitmapText

### LINKED ISSUE #1769: BitmapText: Anchor does not work as expected
When setting a ```BitmapText``` anchor to (0.5, 0.5) I would expect the text to be centered on its position -- at least ```Text``` does exactly that. 

For some reason the anchor seems to remain in the top left corner:

image

http://jsbin.com/micazuqalu/2/edit?js,output

(I'm having the same problem with 3.0.3, but I couldn't find a good cdn for it, so the jsbin is usind 3.0.2)

marekventur:
Fixes #1769 

```BitmapText``` (unlike ```Text```) is inheriting from ```Container``` rather than ```Sprite``` and therefore ```BitmapText.anchor``` does currently not exist. This is slightly confusing and makes aligning BitmapTexts harder than it needs to be. There are workarounds to add centring behaviour (see #1769), but they all rely on manually changing the position. 

This PR adds an ```anchor``` to ```BitmapText``` that behaves similar to ```anchor``` for ```Text```. 

Since every anchor is a ```PIXI.Point``` we can't just use a getter/setter to set ```dirty``` but instead have to regularly compare ```anchor``` against a private copy. I would have rather avoided this, but it's the only solution that I can currently see to detect those changes. Suggestions welcome!

### Performance considerations

In case of ```anchor``` being left at (0, 0) the only costs is a ```PIXI.Point.equals``` every frame to check whether it has changed.  It's a relatively cheap check, so I hope it's not going to be a problem.

If ```anchor``` is set to anything else then there's an additional loop over every glyph during ```#updateText```.

### Backwards compatibility

Unless someone has set (the currently ineffective) ```anchor``` property this change is backwards compatible.

### DIFF
--- a/src/extras/BitmapText.js
+++ b/src/extras/BitmapText.js
@@ -1,4 +1,5 @@
-var core = require('../core');
+var core = require('../core'),
+    math = require('../core/math');
 
 /**
  * A BitmapText object will create a line or multiple lines of text using bitmap font. To
@@ -103,6 +104,24 @@ function BitmapText(text, style)
      * @member {number}
      */
     this.maxLineHeight = 0;
+
+    /**
+     * The anchor sets the origin point of the text.
+     * The default is 0,0 this means the text's origin is the top left
+     * Setting the anchor to 0.5,0.5 means the text's origin is centered
+     * Setting the anchor to 1,1 would mean the text's origin point will be the bottom right corner
+     *
+     * @member {PIXI.Point}
+     */
+    this.anchor = new math.Point();
+
+    /**
+     * Private tracker for the anchor. This allows us to check whether there has been a change
+     *
+     * @member {PIXI.Point}
+     * @private
+     */
+    this._anchor = this.anchor.clone();
 
     /**
      * The dirty state of this object.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by GoodBoyDigital** [comment]
This is a nice PR! one that I think people (including myself! ) will find very useful. 
My one suggestions is that we use ```PIXI.ObservablePoint``` so that we can set the dirty flag whenever anchor is changed. That way we only loop through the characters if the anchor changes. What do you think?

**Comment by ivanpopelyshev** [comment]
@GoodBoyDigital that's what my SpriteFrame was about.

**Comment by marekventur** [comment]
I only found out about ```ObservablePoint``` last week, but I haven't gotten around to update the PR yet. Will do so today or tomorrow (I hope). 

The only problem I can see is how am I going to deal with someone doing ```text.anchor = new PIXI.Point(0.5, 0.5)```? Silently convert it into an ObservablePoint?

@ivanpopelyshev Have you done something similar? Could you point me at those changes?

**Comment by marekventur** [comment]
@ivanpopelyshev nm, I found it. will update.

**Comment by marekventur** [comment]
@GoodBoyDigital @ivanpopelyshev updated. I had to pull in some changes from #2465 to make it work, I hope that's not a problem.

**Comment by ivanpopelyshev** [comment]
Nice! @GoodBoyDigital what do you think?

File: src/extras/BitmapText.js [hunk 1]
```diff
@@ -343,6 +374,16 @@ BitmapText.prototype.updateText = function ()
 
     this.textWidth = maxLineWidth * scale;
     this.textHeight = (pos.y + data.lineHeight) * scale;
+
+    // apply anchor
+    if (this.anchor.x !== 0 || this.anchor.y !== 0)
```
englercj: Would it not just be `this._glyphs[i].anchor.copy(this._anchor);`?

marekventur: That would only move the glyphs by ```anchor*glyphWidth``` rather than ```anchor*textWidth```, wouldn't it?

englercj: Oic, good point.

marekventur:
Is there anything outstanding for this PR or something I can do to help it get merged?

PR #2553, opened 2016-05-06
Text: Splitting font setting into it's various style components (take 2)

themoonrat:
Instead of just setting font, and having to put the settings for bold / size etc. in the right order, these style settings are split up into the 5 style components, which are joined together automatically for you.
There is now:

fontFamily
fontSize
fontStyle
fontVariant
fontWeight

A re-submit of https://github.com/pixijs/pixi.js/pull/2539 - sorry, my lack of git skills couldn't do the rebasing you asked for, and the resident git expert couldn't help either, so was much easier to just submit a new PR!

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -65,6 +65,14 @@ function Text(text, style, resolution)
      */
     this._styleListener = null;
 
+    /**
+     * Private tracker for the current font.
+     *
+     * @member {string}
+     * @private
+     */
+    this._font = '';
+
     var texture = Texture.fromCanvas(this.canvas);
     texture.trim = new math.Rectangle();
     Sprite.call(this, texture);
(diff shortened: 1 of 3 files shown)


SebastianNette:
I don't really see the benefit of this.
What exactly is the advantage of splitting a font string into its components?

All it seems to do is creating bloated code.

ivanpopelyshev:
May be to use the same thing for BitmapFont later?

SebastianNette:
Okay that is true. It might be more convenient to be able to change the font size directly without changing the whole font string.

But then I would remove at least the warning and instead just run a regexp over the font property. Then it is up to the dev, which way he prefers.

I just don't really like declaring 5 different properties if a simple one liner can handle all that for me.

ivanpopelyshev:
Yeah, regexp can be improved :)

englercj:
Having a TextStyle that represents the style of the font gives us a structured way to talk about the style of a font. A string that represents the font is not a very structured way to talk about a font style.

It might be a good way to transmit it (i.e. pass it around) but that isn't the goal here. This makes styles reusable, shareable, and mutable. It makes them easier to interact with, understand, and use. It also empowers us to work with components of the font without splitting the string, changing parts, rejoining the string each time. All of those actions require allocation, if we store just a single string.

There are many good-api benefits that come from splitting these properties, but at the cost, as you noted, of some code bloat. Since the additional code isn't all that much, I'm OK with the tradeoff of code for usability here.

PR #2567, opened 2016-05-13
return to legacy behaviour https://github.com/pixijs/pixi.js/issues/2565

ivanpopelyshev:


ivanpopelyshev:
#2567 Sprites width/height must be in sync with getLocalBounds

ivanpopelyshev:
And now by mistake, it fixes TilingSprite blendMode too.

PR #2591, opened 2016-05-19
Fixes loading embedded bitmap fonts from data URIs (both FNT and Image)

khaitu:
PIXI currently fails to load bitmap fonts from data URIs because while it can correctly parse a FNT XML document from a data URI, it fails to load the image specified inside the FNT file because it tries to prepend `xmlUrl` to the file URI.

This will successfully load the FNT file:
```
PIXI.loader.add(['data:application/xml;base64,...'])
```

If the FNT has an embedded data URI in the form:
```
<page id="0" file="/images/font.png"/>
```

This will fail because PIXI will try to prepend the data URI to it when it passes the new URL to the loader.

Fixing this issue will also allow data URIs to be used inside the FNT file to embed the image.

### DIFF
--- a/src/loaders/bitmapFontParser.js
+++ b/src/loaders/bitmapFontParser.js
@@ -80,25 +80,27 @@ module.exports = function ()
             return next();
         }
 
-        var xmlUrl = path.dirname(resource.url);
+        var xmlUrl = !resource.isDataUrl ? path.dirname(resource.url) : '';
 
-        if (xmlUrl === '.') {
-            xmlUrl = '';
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@khaitu Thanks for the PR. If you can please not make PRs to the `master` branch, PRs are only accepted to the `dev` branch. 
Please read through our [Contributing Guide](https://github.com/pixijs/pixi.js/blob/master/CONTRIBUTING.md) which details how to make good PRs and Issues.

PR #2592, opened 2016-05-20
Container destruction API unification

marekventur:
Based on #2536 

This is a breaking change for V4, the signature for ```Container.destroy()```. ```Sprite.destroy()``` and ```Text.destroy()``` will now only accept an optional object with the following fields. Passing a bool in (v3 behaviour) will cause an error to be thrown.

```javascript
var container = new Container();
container.destroy({children: true, texture: true, baseTexture: true});
```

All parameters default to ```false``` and will be passed on in case of ```children == true```.

As an aside, since I couldn't find anything: Is there a changelog or some sort of update instructions where breaking changes are outlined?

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -606,17 +606,22 @@ Container.prototype.renderCanvas = function (renderer)
 
 /**
  * Destroys the container
- * @param [destroyChildren=false] {boolean} if set to true, all the children will have their destroy method called as well
+ * @param [options] {object} Options parameter
+ * @param [options.children=false] {boolean} if set to true, all the children will have their destroy
+ *      method called as well. 'options' will be passed on to those calls.
+ *
  */
-Container.prototype.destroy = function (destroyChildren)
+Container.prototype.destroy = function (options)
 {
+    options = options || {};
+
     DisplayObject.prototype.destroy.call(this);
 
-    if (destroyChildren)
+    if (options.children)
     {
         for (var i = 0, j = this.children.length; i < j; ++i)
         {
-            this.children[i].destroy(destroyChildren);
+            this.children[i].destroy(options);
         }
     }
 
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -606,18 +606,16 @@ Container.prototype.renderCanvas = function (renderer)
 
 /**
  * Destroys the container
- * @param [options] {object} Options parameter
+ * @param [options] {object|boolean} Options parameter. A boolean will act as if all options have been set to that value
  * @param [options.children=false] {boolean} if set to true, all the children will have their destroy
  *      method called as well. 'options' will be passed on to those calls.
- *
  */
 Container.prototype.destroy = function (options)
 {
-    options = options || {};
-
     DisplayObject.prototype.destroy.call(this);
 
-    if (options.children)
+    var destroyChildren = typeof options === 'boolean' ? options : options && options.children;
+    if (destroyChildren)
     {
         for (var i = 0, j = this.children.length; i < j; ++i)
         {
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -407,20 +407,25 @@ Sprite.prototype.containsPoint = function( point )
 
 
 /**
- * Destroys this sprite and optionally its texture
+ * Destroys this sprite and optionally its texture and children
  *
- * @param [destroyTexture=false] {boolean} Should it destroy the current texture of the sprite as well
- * @param [destroyBaseTexture=false] {boolean} Should it destroy the base texture of the sprite as well
+ * @param [options] {object} Options parameter
+ * @param [options.children=false] {boolean} if set to true, all the children will have their destroy
+ *      method called as well. 'options' will be passed on to those calls.
+ * @param [options.texture=false] {boolean} Should it destroy the current texture of the sprite as well
+ * @param [options.baseTexture=false] {boolean} Should it destroy the base texture of the sprite as well
  */
-Sprite.prototype.destroy = function (destroyTexture, destroyBaseTexture)
+Sprite.prototype.destroy = function (options)
 {
-    Container.prototype.destroy.call(this);
+    options = options || {};
+
+    Container.prototype.destroy.call(this, options);
 
     this.anchor = null;
 
-    if (destroyTexture)
+    if (options.texture)
     {
-        this._texture.destroy(destroyBaseTexture);
+        this._texture.destroy(!!options.baseTexture);
     }
 
     this._texture = null;
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -409,7 +409,7 @@ Sprite.prototype.containsPoint = function( point )
 /**
  * Destroys this sprite and optionally its texture and children
  *
- * @param [options] {object} Options parameter
+ * @param [options] {object|boolean} Options parameter. A boolean will act as if all options have been set to that value
  * @param [options.children=false] {boolean} if set to true, all the children will have their destroy
  *      method called as well. 'options' will be passed on to those calls.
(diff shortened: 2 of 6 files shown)


src/core/display/Container.js
```diff
@@ -606,17 +606,23 @@ Container.prototype.renderCanvas = function (renderer)
...
- * @param [destroyChildren=false] {boolean} if set to true, all the children will have their destroy method called as well
+ * @param [opts] {object} Options parameter
+ * @param [opts.children=false] {boolean} if set to true, all the children will have their destroy
+ *      method called as well. 'opts' will be passed on to those calls.
+ *
  */
-Container.prototype.destroy = function (destroyChildren)
+Container.prototype.destroy = function (opts)
 {
+    opts = opts || {};
```
bigtimebuddy: I'd strongly recommend being verbose in names and avoid abbreviations, e.g., `options` vs `opts`. 

Alternative approach for setting defaults: 
```js
options = Object.assign({
    children: false
}, options || {});
```

src/core/sprites/Sprite.js
```diff
@@ -407,20 +407,27 @@ Sprite.prototype.containsPoint = function( point )
...
+ * @param [opts.baseTexture=false] {boolean} Should it destroy the base texture of the sprite as well
  */
-Sprite.prototype.destroy = function (destroyTexture, destroyBaseTexture)
+Sprite.prototype.destroy = function (opts)
 {
-    Container.prototype.destroy.call(this);
+    opts = opts || {};
+    var texture     = opts.hasOwnProperty('texture')     ? opts.texture     : false;
+    var baseTexture = opts.hasOwnProperty('baseTexture') ? opts.baseTexture : false;
+
```
bigtimebuddy: Alternate using Object.assign:
```js
options = Object.assign({
    texture: false,
    baseTexture: false
}, options || {});
```

src/core/display/Container.js
bigtimebuddy: There is value is still passing in a simple boolean. I don't think we should deprecate because it's simple and has the added benefit of breaking less. Would recommend checking for boolean:
```js
if (typeof options === "boolean") {
    options = { children: options };
}
```

englercj: :+1:

src/core/display/Container.js
```diff
@@ -606,17 +606,23 @@ Container.prototype.renderCanvas = function (renderer)
...
+    var children = opts.hasOwnProperty('children') ? opts.children : false;
```
englercj: I don't think this is necessary, you are doing a falsey check below and the default is false. `if (options.children)` is sufficient.

src/core/sprites/Sprite.js
englercj: Similar feedback to above, if the default is false then just do the check or use `!!` to normalize to a boolean if you *have* to. In this case you don't, just do the check normally and skip this extra work.

ivanpopelyshev:
May be pass a mask FLAG_CHILDREN + FLAG_TEXTURE + FLAG_BASE_TEXTURE ?

```js
const = {
...
FLAG_CHILDREN = 1;
FLAG_TEXTURE = 2;
FLAG_BASE_TEXTURE  = 3;
}
```

i dont know how to name it though, may be DB_ (destroy bit) or DF_ (destroy flag)

Pass integer instead of new object - it will be more effective, but really we dont do destroy() often

src/core/display/Container.js
marekventur: I've changed ```opts``` to ```options```. 

I'm can also allow booleans to be passed, the reason I didn't do it in the first place is because I didn't know how to do it for Text/Sprite. Shall I try to emulate the different old signature for each of them or always treat a boolean as "destroyChildren"? 

If we stick with the old signature, shall I throw a deprecation warning? I'm not sure it's a good idea to keep the old (and broken) signatures around forever.

src/core/sprites/Sprite.js
marekventur: done

src/core/display/Container.js
marekventur: done

marekventur:
@ivanpopelyshev: I guess you meant 1, 2, 4? Personally, it doesn't feel very javascript-y to me. For example it's harder to debug with developer tools - you have to resolve the number manually rather than having a full object to look at.

I'm not sure this is a case where I'd trade of efficiency for readability, it's only going to be called rarely and the difference in speed will be tiny.

ivanpopelyshev:
@marekventur Yeah, I'm for objects too. Just posted it there because we discussed it with @GoodBoyDigital

englercj:
Problem with flags is that plugins can't add their own, or they will almost certainly conflict. Names can conflict too, but are much less likely.

src/core/display/Container.js
marekventur: Looks like this is the last outstanding question to this PR. Shall I add support for the old signature with a deprecation warning?

bigtimebuddy:
Flags don't seem to be as popularly practiced in JavaScript (as opposed to say PHP). Also, I don't like API approaches that are unique. If they were used throughout Pixi, maybe that would be an acceptable approach here. For that and the reasons @englercj said, I'd say no to flags.

src/core/display/Container.js
bigtimebuddy: @marekventur I think we should add a boolean as an alternative function signature and _not_ deprecate this functionality.  There is an open question of whether passing `true` into `destroy()` would be a shortcut for `{children: true}` or `{children: true, texture: true, baseTexture: true}`. I'm in favor of the latter because it doesn't break current implementations and seems more convenient. Thoughts?

ivanpopelyshev: I'm for alternative function signature as well

marekventur: @bigtimebuddy I think both your option would be breaking behaviour (Sprite and Text never destroyed children), I was more leaning towards ```{children: false, texture: <passedInBool1>, baseTexture: <passedInBool2>}``` for ```Sprite.destroy``` and ```{children: false, texture: <passedInBool1>, baseTexture: false}``` for ```Text.destroy```. Happy to do so if that's what you meant.

What are your thoughts on the deprecation warning?

bigtimebuddy: @marekventur We are not deprecating just providing an alternative (shortened) API for destroy.  No warning necessary. This is going to be a subtle breaking change. Providing a boolean is a convenience that says, "aggressively destroy as much as possible"

Here's what I'm thinking:

```js
// These are equivalent for Sprite.prototype.destroy
sprite.destroy(true);
sprite.destroy({baseTexture: true, texture: true});

// These are equivalent for Text.prototype.destroy
text.destroy(true);
text.destroy({baseTexture: true, texture: true});

// These are equivalent for Container.prototype.destroy
contrainer.destroy(true);
container.destroy({children: true, baseTexture: true, texture: true});
```

englercj: I agree with @bigtimebuddy. Only thing I want to clarify is that you shouldn't translate the boolean into an object explicitly. Just pass along the boolean and each object checks if it is true or has its flag to do deletion.

Something along the lines of:

```js
Sprite.prototype.destroy = function (options)
{
    Container.prototype.destroy.call(this, options);
    var destroyTexture = typeof options === 'boolean' ? options : options && options.texture);
    // ....
}
```

That way you destroy the texture if it is `true` or if `options.texture` is true. Same logic for other flags:

```js
```

Maybe someone else can simplify that.

bigtimebuddy: @marekventur Thoughts? Would love to get this merged.

marekventur: Updated with both your suggestions. Thanks!

bigtimebuddy:
@marekventur Great work! And 👍 for the added unit tests.

marekventur:
happy to help!

englercj:
Wonderful, thanks for this @marekventur!!

PR #2614, opened 2016-05-28
Better SVG support

RanzQ:
SVG images are loaded using canvas which allows pre-texture scaling. Added 4th parameter sourceScale to BaseTexture (and Texture) which is used for scaling the SVG before creating the texture.

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -107,6 +107,33 @@ class BaseTexture extends EventEmitter
         this.source = null; // set in loadSource, if at all
 
         /**
+         * The image source that is used to create the texture. This is used to
+         * store the original Svg source when it is replaced with a canvas element.
+         *
+         * TODO: Currently not in use but could be used when re-scaling svg.
+         *
+         * @member {Image}
+         * @readonly
+         */
+        this.origSource = null; // set in loadSvg, if at all
+
+        /**
+         * Type of image defined in source, eg. `png` or `svg`
+         *
+         * @member {string}
+         * @readonly
+         */
+        this.imageType = null; // set in updateImageType
+
+        /**
+         * Scale for source image. Used with Svg images to scale them before rasterization.
+         *
+         * @member {number}
+         * @readonly
+         */
+        this.sourceScale = 1.0;
+
+        /**
          * Controls if RGB channels should be pre-multiplied by Alpha  (WebGL only)
          * All blend modes, and shaders written for default value. Change it on your own risk.
          *
@@ -190,13 +217,17 @@ class BaseTexture extends EventEmitter
      */
     update()
     {
-        this.realWidth = this.source.naturalWidth || this.source.videoWidth || this.source.width;
-        this.realHeight = this.source.naturalHeight || this.source.videoHeight || this.source.height;
+        // Svg size is handled during load
+        if (this.imageType !== 'svg')
+        {
+            this.realWidth = this.source.naturalWidth || this.source.videoWidth || this.source.width;
+            this.realHeight = this.source.naturalHeight || this.source.videoHeight || this.source.height;
 
-        this.width = this.realWidth / this.resolution;
-        this.height = this.realHeight / this.resolution;
+            this.width = this.realWidth / this.resolution;
+            this.height = this.realHeight / this.resolution;
 
-        this.isPowerOfTwo = bitTwiddle.isPow2(this.realWidth) && bitTwiddle.isPow2(this.realHeight);
+            this.isPowerOfTwo = bitTwiddle.isPow2(this.realWidth) && bitTwiddle.isPow2(this.realHeight);
+        }
 
         this.emit('update', this);
     }
@@ -237,9 +268,15 @@ class BaseTexture extends EventEmitter
         this.source = source;
 
         // Apply source if loaded. Otherwise setup appropriate loading monitors.
-        if ((this.source.complete || this.source.getContext) && this.source.width && this.source.height)
+        if ((source.src && source.complete || source.getContext) && source.width && source.height)
         {
-            this._sourceLoaded();
+            this._updateImageType();
+
+            if (this.imageType === 'svg'){
+                this._loadSvgSource();
+            } else {
+                this._sourceLoaded();
+            }
         }
         else if (!source.getContext)
         {
@@ -251,6 +288,7 @@ class BaseTexture extends EventEmitter
 
             source.onload = function ()
             {
+                scope._updateImageType();
                 source.onload = null;
                 source.onerror = null;
 
@@ -261,6 +299,11 @@ class BaseTexture extends EventEmitter
 
                 scope.isLoading = false;
                 scope._sourceLoaded();
+
+                if (scope.imageType === 'svg'){
+                    scope._loadSvgSource();
+                    return;
+                }
 
                 scope.emit('loaded', scope);
             };
@@ -285,11 +328,18 @@ class BaseTexture extends EventEmitter
             // NOTE: complete will be true if the image has no src so best to check if the src is set.
             if (source.complete && source.src)
             {
-                this.isLoading = false;
 
                 // ..and if we're complete now, no need for callbacks
                 source.onload = null;
                 source.onerror = null;
+
+                if (scope.imageType === 'svg')
+                {
+                    scope._loadSvgSource();
+                    return;
+                }
+
+                this.isLoading = false;
 
                 if (source.width && source.height)
                 {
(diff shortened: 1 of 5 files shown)


RanzQ:
A demo: https://raw.githack.com/RanzQ/svg-pixi-demo/master/index.html

The texture is re-created on window resize (there's a bug that it requires a click after, renders empty screen otherwise). And yes, it's forked from bunnymark. :)

bigtimebuddy:
@RanzQ Thanks for doing this PR. Would you not commit build files, please see the [contributing guidelines](https://github.com/pixijs/pixi.js/blob/master/CONTRIBUTING.md#making-changes).

RanzQ:
@bigtimebuddy Sorry, didn't mean to. Haven't done many PRs and didn't know a commit would be added afterwards.

GoodBoyDigital:
Hey peeps! Firstly thanks for this PR @RanzQ, this has been an issue for a while! I have been reviewing this and had an idea that I would like to get your opinion on.

How about instead of adding the code into the base texture constructor we instead create a new static function called ```fromSVG``` That way we can add the scaling logic into there rather than into the main body. 

We can then leverage the new ```from``` method to identify if the source is an svg and run the correct method. 

What do ya think?

RanzQ:
Good suggestion, the API was a bit clumsy, I can modify the PR. I'll also do some testing on other browsers than Chrome and the demo seems to be broken now (after a Chrome update I guess).

GoodBoyDigital:
super! Yes a PR would be ace 👍

RanzQ:
@GoodBoyDigital I tried to modify this towards your suggestion but actually the scaling related code is not in the constructor but in `fromImage` and `update`.

All I do in `fromImage` is that I store `sourceScale` and `imageType` (which is read from extension), then `update` handles the "load using canvas" logic.

So would it be ok to separate that logic in `updateSVG`? And where should the scaling parameter be in the API?

RanzQ:
I updated it a bit. Also noticed that the SVG image must have `width` and `height` defined, otherwise the Image size defaults to 150x150 in Chrome and 0x0 in Firefox. They could be given as parameter too but I decided to throw an error.

src/core/textures/BaseTexture.js
```diff
@@ -195,10 +214,79 @@ BaseTexture.prototype.update = function ()
...
+};
+
+/**
+ * Updates texture from SVG image using canvas. The original SVG Image is stored
+ * as `origSource` and the created canvas is is now `source`. The SVG is scaled
+ * using `sourceScale`.
+ */
+BaseTexture.prototype.updateSVG = function ()
+{
+    if (this.source.type !== 'svg') {
```
englercj: This should probably use `this.isSVG()`

src/core/textures/BaseTexture.js
```diff
@@ -195,10 +214,79 @@ BaseTexture.prototype.update = function ()
...
+        // Do nothing if source is not svg
+        return;
+    }
+
+    if (!this.realWidth || !this.realHeight || this.realWidth <= 1 || this.realHeight <= 1 ||
+        (this.realWidth === 150 && this.realHeight === 150)) {
```
englercj: So this means if I try to draw an image of width/height 150 pixi will throw an error?

src/core/textures/BaseTexture.js
RanzQ: Yeah, that should be fixed.

src/core/textures/BaseTexture.js
RanzQ: @englercj basically yes. I didn't find a neat way to determine if the SVG has size properties or not. With inline svg that can be checked but when loaded using Image API I don't know a way to access the original width and height of the svg. So now it relies on the fact that Chrome sets width and height to 150. One should anyway use pow2 textures, right?

englercj: I don't think this check is the right thing to do, I wonder if we can use the `source` to check for properties in the svg instead.

Texture do not *have* to be power of two. Non-power of two textures work just fine for most drawing (I was going to bring that up in your next code that forces power of two).

RanzQ: I can investigate this a bit further but the source seems to have no info of svg size properties if they were missing. Chrome set it to 150x150 and Firefox to 0x0.

About the power of two, I set the canvas to be power of two and draw the correctly sized image into middle of it. I think I had some problems with non-pow2 canvases. I can check that too why I ended up with that solution.

RanzQ: Actually that isn't a good way to do it since corner anchors would be misaligned... Have to fix that.

RanzQ:
Ok, `this.isPowerOfTwo` solved the issue with non-pow2 canvases. I still didn't find a better solution for svgs without width and height. I added a better comment to clarify the problem.

src/core/utils/index.js
```diff
@@ -84,6 +84,17 @@ var utils = module.exports = {
...
 
     /**
+     * Determine if the given url is an SVG image.
+     *
+     * @param url {string} the image path
+     * @return {boolean}
+     */
+    getImageTypeOfUrl: function (url)
+    {
+        return CONST.IMAGE_TYPE.exec(url)[1];
```
englercj: This can throw for non-matching urls. Proper guards need to be here.

src/core/textures/BaseTexture.js
```diff
@@ -405,15 +499,24 @@ BaseTexture.fromImage = function (imageUrl, crossorigin, scaleMode)
             image.crossOrigin = determineCrossOrigin(imageUrl);
         }
 
+        var imageType = getImageTypeOfUrl(imageUrl);
+
+        if (imageType) {
+            image.type = imageType;
```
englercj: Are the expando properties necessary here? Looks like this is only used in one place, can't the getImageTypeOfUrl call be there instead?

src/core/textures/BaseTexture.js
```diff
@@ -405,15 +499,24 @@ BaseTexture.fromImage = function (imageUrl, crossorigin, scaleMode)
...
+        }
+
         baseTexture = new BaseTexture(image, scaleMode);
         baseTexture.imageUrl = imageUrl;
 
-        image.src = imageUrl;
-
-        utils.BaseTextureCache[imageUrl] = baseTexture;
+        baseTexture.sourceScale = sourceScale;
```
englercj: What is this property for? I don't see it used, or declared in the constructor.

src/core/textures/BaseTexture.js
RanzQ: You're right. I can remove that.

src/core/textures/BaseTexture.js
RanzQ: That is the scale for the svg image. One can use it to scale the svg before creating the texture. For example in the demo I scale the image based on screen width like: `beeTexture = new PIXI.Texture.fromImage('bee.svg', undefined, undefined, width / 600)`.

src/core/textures/BaseTexture.js
```diff
@@ -405,15 +499,24 @@ BaseTexture.fromImage = function (imageUrl, crossorigin, scaleMode)
...
 
         // if there is an @2x at the end of the url we are going to assume its a highres image
         baseTexture.resolution = utils.getResolutionOfUrl(imageUrl);
+        baseTexture.imageType = image.type;
```
RanzQ: I guess you meant to add the comment here, that is unnecessary line.

src/core/textures/BaseTexture.js
englercj: Sorry I misclicked, I was referring to the lines below this:

```js
baseTexture.imageType = image.type;
```

`imageType` was the property my comment was about.

RanzQ:
Rebased to latest dev.

src/core/textures/Texture.js
```diff
@@ -346,16 +346,21 @@ Texture.prototype._updateUvs = function ()
  * @static
  * @param imageUrl {string} The image url of the texture
  * @param [crossorigin] {boolean} Whether requests should be treated as crossorigin
+<<<<<<< HEAD
  * @param [scaleMode=PIXI.SCALE_MODES.DEFAULT] {number} See {@link PIXI.SCALE_MODES} for possible values
+=======
+ * @param [scaleMode] {number} See {@link PIXI.SCALE_MODES} for possible values
+ * @param [sourceScale=(auto)] {number} Scale for the original image, useful with SVG images.
+>>>>>>> Better SVG support. Images are loaded using canvas which allows pre-texture scaling.
```
englercj: You had a conflict during your rebase.

src/core/textures/BaseTexture.js
```diff
@@ -104,6 +105,24 @@ function BaseTexture(source, scaleMode, resolution)
...
+    this.origSource = null; // set in update, if at all
+
+
+    /**
+     * Scale for source image. Used with SVG images to scale them before rasterization.
+     *
+     * @member {number}
+     * @readonly
+     */
+    this.sourceScale = null;
```
englercj: I would default this to `1` and remove the `if (this.sourceScale)` later on.

src/core/textures/Texture.js
RanzQ: Oh, forgot to add the file.

src/core/textures/BaseTexture.js
RanzQ: Change committed for this.

englercj:
Cool! This all looks great to me, only thing left to do is  figure out a way to do the width/height checks without hard-coding `150` as invalid. Maybe `.getAttribute()` helps?

RanzQ:
I'll try to figure out iff that can be done somehow. The problem is that the svg in this case is an external document and I think you can't access its properties. With inline svg it would be easy.

trsh:
Wow. I hope this makes to the dev branch. Or is it already?

englercj:
> Wow. I hope this makes to the dev branch. Or is it already?

Not yet.

RanzQ:
@englercj There might be a way... Reading the svg first as an xml file and then manually creating the dom element the properties should be accessible. Like here:

http://stackoverflow.com/a/21289967

I'll try it soon.

RanzQ:
Another possible solution, using `<embed>` element instead of `Image`: 

image

trsh:
One another idea:

var exLoadedDocument = soemthing.source;

var div = document.createElement('svgwrap');
div.innerHTML = exLoadedDocument;
var elements = div.childNodes; // You can access SVG DOM

So before you feed Canvas with it, add desired width and height (instead of reading it, because it does not make sense with SVG) with standard JS dom manipulation.

RanzQ:
@trsh You would still need to create a new Image using the data-uri method (which doesn't work in all browsers) cause the canvas API needs an Image (svg element can't be drawn).

trsh:
@RanzQ i don't know pixi code very well. But sure at some point the SVG doc is fetched from external source as it is - XML.  **Right after it**, we need to read the width and height (or set it) and continue as now.

http://jsfiddle.net/8ztNU/77/

trsh:
> "But sure at some point the SVG doc is fetched"

 not so sure about this after checking the code :D

trsh:
@RanzQ to be honest, I googled now for SVG images, and hoped to get one without width and height (for tests). After 40 I gave up. These are like rare cases that we can choose to NOT support?

RanzQ:
In the case where you have `</img>` you cannot modify the svg. It's just loaded (in native code) and then the image can be scaled. For drawing to canvas the image element is needed.

Actually now when thinking about it, I'm not sure if the canvas is even needed. Maybe we could just pre-scale the Image and then texturize it. 

@trsh there are some without width and height but as discussed but it's hard to detect (they just bug when drawing to canvas). So that's why I was willing to just throw an Error.

It seems that I was unlucky since I tested with one of these: http://emojione.com/releases/. The old versions don't have width and height but new ones do. Maybe they realized too that it's bad practise. :D

RanzQ:
Here's a codepen about inconsistencies within the `drawImage` API: http://codepen.io/RanzQ/pen/PzGNEO

2016-06-18 12_03_17-2016_06_18_11_53_14_a_pen_by_ranzq png - paint

So I'll try just to scale the Image and see if that works.

RanzQ:
I will finish this during the next few weeks, just been too busy (celebrating summer).

tobireif:
Would this support responsive pages? Would eg an SVG circle stay scalable == grow/shrink during viewport resize?

RanzQ:
@tobireif That is the plan, I'm currently working on this again.

trsh:
@RanzQ wasn't the initial plan to convert SVG to standard texture object during the load resource phase. And then it's used just like like it would come from PNG. I like the scenario, because it promises same rendering speed.

tobireif:
@trsh That could still be an option I'd guess.

RanzQ:
@trsh @tobireif Yeah, I'm still figuring how to pameterize it, but what I had in mind was that you could define the sprite size in fraction of the viewport and when the viewport is updated the texture would be generated again. Of course re-uploading the texture to GPU is a demanding process but normally you don't resize your viewport so often.

I've experimented with this a bit and it works.

RanzQ:
@englercj Regarding [this](#issuecomment-226575217), there's a way to read the size from the SVG using `<object>` but it doesn't support CORS obviously.

Size of same origin svgs can be read but cross-origin is more difficult, I can't read the width and height from the svg but object `clientWidth` and `clientHeight` gives something:

image

Left is Chrome, object size is updated after timeout (not sure after which event) not on load. Right is Firefox, object size is updated on load. Anyway the SVG has no size properties and there's no way to tell if it has or not. So the object gets maximum size that fits the window.

 What do you think, should we throw an error for cross-origin svgs and otherwise for svgs with non-defined size?

RanzQ:
@GoodBoyDigital Should I squash the commits into a single before merge? I just committed a much better version. I've still have to write tests and do some testing on IE since it throws an Error.

src/core/textures/BaseTexture.js
```diff
@@ -210,69 +219,158 @@ module.exports = BaseTexture;
...
     }
 
-    // Store source size for canvas drawImage
-    var sourceWidth = this.realWidth;
-    var sourceHeight = this.realHeight;
+    this._loadSvgSourceUsingString(svgString);
+};
+
+/**
+ * Creates an SVG document using XHR and then calls `_loadSvgSourceUsingString`.
```
RanzQ: This has invalid description, SVG document is not used anymore.

src/core/utils/index.js
```diff
@@ -100,13 +135,42 @@ var utils = module.exports = {
...
+     * Typedef for Size object.
+     *
+     * @typedef {object} Size
+     * @property {width} Width component
+     * @property {height} Height component
+     */
+
+    /**
+     * Get size from an svg string using regexp. This could be used to optimize
+     * SVG size parsing, currently not in use.
```
RanzQ: Actually is in use....

RanzQ:
After doing some more tests, IE11 throws a SecurityError (see [gl.SECURITY_ERR in Exceptions](https://msdn.microsoft.com/en-us/library/dn302435(v=vs.85).aspx)).

image

This happens with image if it has an svg source and with canvas if it has an svg image drawn to it (`drawImage()` has been used to fill pixels). No matter if it's same domain or not. Related to https://github.com/pixijs/pixi.js/issues/2928.

RanzQ:
More IE problems. This "draw svg to canvas" method I've used has problems on IE10 too with canvas fallback: http://stackoverflow.com/a/29370038

Seems like even if loaded, the svg might not be ready to be drawn.

RanzQ:
This is now ready to be merged in my opinion. Improvement suggestions welcome!

englercj:
I'm willing to try and merge this if you can rebase onto the latest `dev` to get the ES6 changes.

RanzQ:
@englercj Done. Also squashed the previous commits.

src/core/textures/BaseTexture.js
```diff
@@ -236,9 +267,15 @@ class BaseTexture extends EventEmitter
         this.source = source;
 
         // Apply source if loaded. Otherwise setup appropriate loading monitors.
-        if ((this.source.complete || this.source.getContext) && this.source.width && this.source.height)
+        if ((source.src && source.complete || source.getContext) && source.width && source.height)
         {
-            this._sourceLoaded();
+            this._updateImageType();
+
+            if (this.imageType === 'svg'){
```
englercj: there are a few placed here where the brace-style doesn't match the rest of the file. Please place braces on a newline

RanzQ: This can be linted btw `"brace-style": [1, "allman"]`, but got 300 warnings when enabled. Fixed the files in this PR anyway. :)

trsh:
So...??!!  Still excited about this.

RanzQ:
@trsh In case you want to try it out, I've got a build [here](https://github.com/RanzQ/svg-pixi-demo/blob/master/js/pixi.min.js) and a [codepen](https://codepen.io/RanzQ/pen/YGkQYj) that uses it.

MartinBspheroid:
@RanzQ It works fine, but seems like there is problem with parsing svg files, as it's expecting `width` and `height` parameters without `px` in main `<svg>` tag.

RanzQ:
@MartinBspheroid Oh, you're right. Actually the spec says unit can be `"em" | "ex" | "px" | "in" | "cm" | "mm" | "pt" | "pc" | "%"`. Any suggestions how the different units should be handled?

englercj:
parse pixels, warn it won't work otherwise?

bigtimebuddy:
@RanzQ merge conflicts? Otherwise, I think this looks good. Great work.

trsh:
I also think for start support px, and show warning otherwise. Can improve/upgrade in time, if there's a point to it.

RanzQ:
It now supports values with `px`. I'll do some tests with different units later but I think mostly people work in pixels when they do vector graphics for web.

MartinBspheroid:
Hey, that is great.  I'm sorry that I couldn't help more than just raising flag, I'm new to Javascript and I would probably come up with really cumbersome solution.

PR #2615, opened 2016-05-31
Fixes JSDoc not running to completion

themoonrat:
The *[] syntax to describe a member that is an array of any type throws errors when trying to generate the docs.

I've changed this to a similar meaning Array<*> which it is happy with


bigtimebuddy:
Good catch. Thanks @themoonrat.

PR #2616, opened 2016-05-31
Adds missing destroy to Prepare, supports multiple simultaneous uploads

bigtimebuddy:
Was crashing when trying to destroy renderer because plugin had not destroy method. 
Also, adds support for multiple simultaneous uploads (calling upload twice in a row).


GoodBoyDigital:
nice! 👍

PR #2617, opened 2016-05-31
Dev static transform

GoodBoyDigital:
Added some static transforms - should be faster if we consider most things do not move too much in a scene…


ivanpopelyshev:
OH GOD what is that?

PR #2626, opened 2016-06-05
Stopping Ticker from being able to return a negative delta time

themoonrat:
As noted by @drkibitz - my PR to stop negative delta times contained a glaring assignment error. Sorry all.

### DIFF
--- a/src/core/ticker/Ticker.js
+++ b/src/core/ticker/Ticker.js
@@ -332,7 +332,7 @@ Ticker.prototype.update = function update(currentTime)
     // Allow calling update directly with default currentTime.
     currentTime = currentTime || performance.now();
     // Save uncapped elapsedMS for measurement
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
:+1

drkibitz:
I'm ok, but noting the comment thread here:
https://github.com/pixijs/pixi.js/commit/52e4a749d4a6c7150422cabe1288a9cf884808ef#commitcomment-17744980

drkibitz:
I can make a PR with what I'm thinking if anyone is interested ;) I just want to make sure a comment is make about the issue in code, and the appropriate things are skipped when necessary. For instance, if the elapsedMS is NOT greater than zero (a result of, but is not limited to, this issue), then there is no delta, and therefore we should not emit anything in this case, as it wouldn't result in any change anyway.

englercj:
Closing in favor of #2627.

PR #2627, opened 2016-06-05
Comment about ticker timing issue. Do not trigger update for zero delta.

drkibitz:
This is my preferred approached to fixing the issue for this PR https://github.com/pixijs/pixi.js/pull/2626. 

It was originally addressed here https://github.com/pixijs/pixi.js/commit/52e4a749d4a6c7150422cabe1288a9cf884808ef#commitcomment-17744980, with information on the issue in the comment thread.

The comment in the diff describes things.

### DIFF
--- a/src/core/ticker/Ticker.js
+++ b/src/core/ticker/Ticker.js
@@ -331,19 +331,42 @@ Ticker.prototype.update = function update(currentTime)
 
     // Allow calling update directly with default currentTime.
     currentTime = currentTime || performance.now();
-    // Save uncapped elapsedMS for measurement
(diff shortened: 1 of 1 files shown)


drkibitz:
@themoonrat

Here is an updated fiddle with this change monkey patched:
https://jsfiddle.net/ud05btt5/9/

themoonrat:
Much better than my plastering over the issue :)

GoodBoyDigital:
Nice! Thanks @drkibitz, clean PR and well documented - just way we like em 👍

PR #2628, opened 2016-06-05
Update dev dependencies and fix container test

drkibitz:
Watchify was already pulling in Browserify 13, so this just brings pixi dep in line with that, as well as updating a bunch of other minor modules.

FYI, [gulp-sourcemaps](https://github.com/floridoo/gulp-sourcemaps) still needs a big upstream update, as it is still depending on an older version [graceful-fs](https://github.com/isaacs/node-graceful-fs). gulp-sourcemaps, [testem](https://github.com/testem/testem), and [mocha](https://github.com/mochajs/mocha) dependencies, are where most of the [npm deprecation warnings](https://docs.npmjs.com/cli/deprecate) are coming from.


GoodBoyDigital:
nice, thanks @drkibitz 👍

Issue #2631, opened 2016-06-08
v4 wrong width/height behaviour of PIXI.Graphics and PIXI.Container


rkrasky:
```js
var renderer = new PIXI.WebGLRenderer(800, 600);

document.body.appendChild(renderer.view);

var stage = new PIXI.Container();

var g = new PIXI.Graphics();
g.beginFill(0xff00ff, 0.5);
g.drawRect(0, 0, 100, 100);
g.endFill();
stage.addChild(g);

setInterval(function(){ g.width = 200; console.log(g.width); }, 1000);

animate();

function animate() {
    requestAnimationFrame(animate);

    renderer.render(stage);
}
```

Expected result: after one second the graphics object resizes to width 200 and is not resizing anymore
Actual result: the graphics object width jumping from 100 to 200 every second

Edit: also PIXI.Container has problems with width and height:

```js
var renderer = new PIXI.WebGLRenderer(800, 600);

document.body.appendChild(renderer.view);

var stage = new PIXI.Container();

PIXI.loader.add("test", "bunny.png").load(function() {

    var c = new PIXI.Container();
    stage.addChild(c);

    console.log("container w/h", c.width, c.height); // correct (0,0)

    bunny = PIXI.Sprite.fromImage("bunny.png");

    console.log("bunny w/h", bunny.width, bunny.height); // correct

    c.addChild(bunny);

    console.log("container w/h", c.width, c.height); // incorrect (0,0) should be the same size as bunny

    setTimeout(function() {
        console.log("container w/h", c.width, c.height); // correct
    }, 500);

    animate();
});

function animate() {
    requestAnimationFrame(animate);

    renderer.render(stage);
}
```

ivanpopelyshev:
WOW that's a good bug!

whizzkid:
I'm experiencing the same. 
width is not set correctly on containers before the 1st render call.

whizzkid:
vertexData is all zeroes when you do getbounds before a render() so the getbounds function doesnt work properly

ivanpopelyshev:
So we need to lazily compute vertexData, even without render

whizzkid:
I've been stepping through the code a bit
vertexdata gets calculated for _renderWebGL 
but canvasrenderer also gives a wrong result.

whizzkid:
scratch that. my test is using webgl even when i switched to canvas mode.
gonna try to see if real canvas rendering also has this problem

whizzkid:
[...]

PR #2635, opened 2016-06-09
fix bug: sprite bounds are incorrect until 1st render

whizzkid:
This fix calls calculateVertices() when a texture is set (and loaded)
so users can add stuff to the screen, and use their sizes before having to call render()

fixes displayObject and sprite bugs in  #2631


whizzkid:
there's still something wrong, cancelling this pullrequest

PR #2636, opened 2016-06-10
Fix hue operation in ColorMatrixFilter

georgiee:
The difference:
screen shot 2016-06-10 at 09 02 55

See here the origin of the PR:
http://www.html5gamedevs.com/topic/22808-problem-with-hue-color-shift/

And the demo (skip the shader, this PR is based on the fixed color matrix):
http://codepen.io/gkey/pen/yJeVVy


georgiee:
sorry wrong branch

PR #2638, opened 2016-06-10
Fix hue operation in ColorMatrixFilter

georgiee:
_(I don't want to spam you with this simple commit. My first PR was for the wrong branch, the next based on the master instead of the latest dev, now my last try. I hope this works now)_

The difference:
screen shot 2016-06-10 at 09 02 55

See here the origin of the PR:
http://www.html5gamedevs.com/topic/22808-problem-with-hue-color-shift/

And the demo (skip the shader, this PR is based on the fixed color matrix):
http://codepen.io/gkey/pen/yJeVVy

### DIFF
--- a/src/filters/colormatrix/ColorMatrixFilter.js
+++ b/src/filters/colormatrix/ColorMatrixFilter.js
@@ -186,21 +186,45 @@ ColorMatrixFilter.prototype.blackAndWhite = function (multiply)
 ColorMatrixFilter.prototype.hue = function (rotation, multiply)
 {
     rotation = (rotation || 0) / 180 * Math.PI;
-    var cos = Math.cos(rotation),
-        sin = Math.sin(rotation);
+   
+    var cosR = Math.cos(rotation),
+        sinR = Math.sin(rotation),
+        sqrt = Math.sqrt;
+ 
+    /*a good approximation for hue rotation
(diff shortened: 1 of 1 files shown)


englercj:
please do not commit builds in a PR, otherwise this looks interesting. Thanks for taking the time to put it together.

georgiee:
Hi Chad,
sorry for including the build I wondered why it's not excluded by the gitignore file so I was unsure about excluding it. Is it fine for this PR? I will exclude it in any following PR then.

ivanpopelyshev:
@georgiee because of reasons. I'm discussing it right now with @GoodBoyDigital, I want it removed from pixi source repo and added to separate repo with all other plugins binary files

englercj:
@georgiee We include builds so they are easily accessible to people landing on the page or cloning the project. However, we only generate them once in a while, and as our [Contributing Guide](https://github.com/pixijs/pixi.js/blob/master/CONTRIBUTING.md) points out, we prefer not to take them in PRs (just dirties the merge, and often conflicts).

georgiee:
Sorry I read the contribution guide beforehand but I must have missed that part about the build. I just reverted the build files so this PR only consists of the changed file from now on. Thanks.

GoodBoyDigital:
N I C E 👍

PR #2640, opened 2016-06-11
Dev static transform

GoodBoyDigital:
Pretty sure this is good to go now! @bigtimebuddy does this work ok with pixi-animate?


ivanpopelyshev:
Go for it.

bigtimebuddy:
Little late on this, but it worked well in my pixi-animate tests.

PR #2644, opened 2016-06-12
Dev

Hachitus:
Only text / documentation changes, except file: src/interaction/InteractionManager.js

Correct or not for this PR: But it was using 1 as a default and I changed it to core.RESOLUTION. As I see it better to get that default from the constants, than be hard-coded. In my simple preliminary tests, it seems to work ok. But if needed, it is at it's own commit and can be dropped easily.


Hachitus:
Sorry. It seems there is accidentally commits there, that shouldn't be there.

PR #2645, opened 2016-06-12
pixijs#2541 resolution means pixel ratio too

Hachitus:
Should consist of only text / documentation changes.


GoodBoyDigital:
nice one @Hachitus 👍

PR #2647, opened 2016-06-14
Takes into account stroke thickness when drawing dropShadow

themoonrat:
Currently, a drop shadow only takes into account the actual text, and not the stroke. This can make the shadow look odd on large stroke thickness as the shadow appears much thinner than it should be.This PR fixes this issue

As a before and after example: especially note the difference on the 'e' of 'Stroke'

http://www.moonrat.co.uk/html5/pixi_before/
http://www.moonrat.co.uk/html5/pixi_after/


englercj:
@themoonrat Looks good to me. :+1:

GoodBoyDigital:
👍

bigtimebuddy:
Nice, good work @themoonrat

PR #2650, opened 2016-06-14
RFC: add sprite mask support for canvas renderer

uorbe001:
We are using sprite masks on some of our games and the lack of support for canvas is becoming a bit of an issue. I thought I'd give it a try and try to implement it myself, but even though this solution works (for the cases I've tested so far), I'm not 100% happy with it. 

There is some things in this PR that I'm sure can be cleaned-up, but I'm not familiar with the pixi codebase, so I could use some advice if someone can spot things that are easy to cleanup (specifically, I don't quite like the `_copyCanvas` I had to add and some of the code around it). 

I'm also not convinced this is the best solution, as I had to create 2 different canvases (one to render the stuff that will be masked on, and the other one to apply the mask to that content which is then applied to the 'real' canvas). So if anyone has an alternate solution, tips/advice would be more than welcome.  I had to add the `_makeBlackTransparent` function because this wasn't doing the same as the webgl mask does (even though it seemed arguably correct), which is a nasty I'm not fond of either.

I have tested this using some code based on the examples repo one: 

```

stage.interactive = true;

var bg = PIXI.Sprite.fromImage('_assets/background.png');

bg.anchor.x = 0.5;
bg.anchor.y = 0.5;

bg.position.x = renderer.width / 2;
bg.position.y = renderer.height / 2;

stage.addChild(bg);

container.position.x = renderer.width / 2;
container.position.y = renderer.height / 2;

// add a bunch of sprites

var bgFront = PIXI.Sprite.fromImage('_assets/SceneRotate.jpg');
bgFront.anchor.x = 0.5;
bgFront.anchor.y = 0.5;

container.addChild(bgFront);

stage.addChild(container);

// let's create a moving shape
var thing = PIXI.Sprite.fromImage('_assets/stage_mask_alpha_inv.png');
stage.addChild(thing);
thing.anchor.set(0.5);
thing.position.x = renderer.width / 2;
thing.position.y = renderer.height / 2;


container.mask = thing;

var count = 0;

stage.on('click', onClick);
stage.on('tap', onClick);

function onClick()
{
    if(!container.mask)
    {
        container.mask = thing;
    }
    else
    {
        container.mask = null;
    }
}

var help = new PIXI.Text('Click to turn masking on / off.', { font:'bold 12pt Arial', fill: 'white' });
help.position.y = renderer.height - 26;
help.position.x = 10;
stage.addChild(help);

animate();

function animate()
{
    renderer.render(stage);
}
```

As the mask, I have used 3 different images:
stage_mask_alpha_coloured
stage_mask_alpha_inv
stage_mask_alpha

For this code, using these images I get the same results with a webgl-enabled browser and a webgl-disabled browser, although there might be some cases that don't work with this.

Could someone give me some help identifying issues/alternate solutions to be able to get this merged in dev?

### DIFF
--- a/src/core/renderers/canvas/utils/CanvasMaskManager.js
+++ b/src/core/renderers/canvas/utils/CanvasMaskManager.js
@@ -14,6 +14,48 @@ export default class CanvasMaskManager
     constructor(renderer)
     {
         this.renderer = renderer;
+        this._maskStates = [];
+    }
+
+    /**
+     * @param {PIXI.CanvasRenderer} renderer - The canvas renderer.
+     * @returns {Canvas} a new canvas object with the same attributes as the one used in the rendering context.
+     */
+    _copyCanvas(renderer)
+    {
+        const canvas = document.createElement('canvas');
+
+        canvas.width = renderer.context.canvas.width;
+        canvas.height = renderer.context.canvas.height;
+
+        return canvas;
+    }
+
+    /**
+     * This function sets up and returns a context in the given canvas copying the
+     * data from the canvas used in the renderer class.
+     *
+     * @param {Canvas} canvas - The canvas to setup the context for.
+     * @param {PIXI.CanvasRenderer} renderer - The canvas renderer to copy the setup from.
+     * @param {object} maskData - the maskData that will be pushed
+     * @returns {Context} a new 2d rendering context created to draw in the given canvas.
+     */
+    _setupContext(canvas, renderer, maskData)
+    {
+        const context = canvas.getContext('2d');
+        const transform = maskData.worldTransform;
+        const resolution = renderer.resolution;
+
+        context.setTransform(
+            transform.a * resolution,
+            transform.b * resolution,
+            transform.c * resolution,
+            transform.d * resolution,
+            transform.tx * resolution,
+            transform.ty * resolution
+        );
+
+        return context;
     }
 
     /**
@@ -40,9 +82,18 @@ export default class CanvasMaskManager
             transform.ty * resolution
         );
 
-        // TODO suport sprite alpha masks??
-        // lots of effort required. If demand is great enough..
-        if (!maskData._texture)
+        if (maskData._texture)
+        {
+            const maskState = {
+                originalContext: renderer.context,
+                maskableContext: this._setupContext(this._copyCanvas(renderer), renderer, maskData),
+                maskData,
+            };
+
+            this._maskStates.push(maskState);
+            renderer.context = maskState.maskableContext;
+        }
+        else
         {
             this.renderGraphicsShape(maskData);
             renderer.context.clip();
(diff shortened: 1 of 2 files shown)


GoodBoyDigital:
oh! this looks nice interesting indeed! I will review and get back to you 👍

englercj:
Can you rebase your changes onto the latest `dev` branch please?

uorbe001:
@englercj sorry, I've been away for a few days and I just got the time to read this. I'll do it *soon* (the rebase is already done, just need to solve linter errors and check that it still works)

englercj:
Cool, np. Thanks for taking the time to contribute this!

src/core/renderers/canvas/utils/CanvasMaskManager.js
```diff
@@ -14,6 +14,48 @@ export default class CanvasMaskManager
...
+        this._maskStates = [];
+    }
+
+    /**
+     * @param {PIXI.CanvasRenderer} renderer - The canvas renderer.
+     * @returns {Canvas} a new canvas object with the same attributes as the one used in the rendering context.
+     */
+    _copyCanvas(renderer)
+    {
+        const canvas = document.createElement('canvas');
```
englercj: These need to be pooled, creating one (or more) each frame is going to be a lot of GC.

uorbe001: good point, will change it as soon as possible.

ivanpopelyshev:
We need a pool like in FilterManager, but for Canvas2d

src/core/renderers/canvas/utils/CanvasMaskManager.js
ivanpopelyshev: Please look at pool implementation for FilterManager. Pot2 and all that stuff

src/core/renderers/canvas/utils/CanvasMaskManager.js
```diff
@@ -153,12 +208,63 @@ export default class CanvasMaskManager
...
+            maskContext.drawImage(maskState.maskableContext.canvas, -maskCanvas.width / 2, -maskCanvas.height / 2);
+            this._pool.freeElement(maskState.maskableContext.canvas);
+
+            // Restore the original context
+            renderer.context = maskState.originalContext;
+            // draw the masked content into the original context
+            renderer.context.drawImage(maskCanvas, -maskCanvas.width / 2, -maskCanvas.height / 2);
+            // freeing the maskCanvas here makes it create weird artifacts for some reason I haven't
+            // been able to figure out, if someone knows of a better solution, please fix this.
+            this._clearOnNextFrame.push(maskCanvas);
```
uorbe001: I don't like this, but I haven't been able to figure out why this canvas is needed after this point (I don't think it should be necessary) and I couldn't think of any other point to free it.

uorbe001:
@englercj @ivanpopelyshev I added the pool. I still feel like this is more complex than it ought to be, if anyone has any ideas on how to improve it, I'm all ears.

bigtimebuddy:
@englercj could you review the canvas pooling?
@uorbe001 could you please updated this PR to the latest **dev** branch?

src/core/renderers/canvas/utils/CanvasMaskManager.js
```diff
@@ -153,12 +208,63 @@ export default class CanvasMaskManager
...
+    }
+
+    /**
      * Restores the current drawing context to the state it was before the mask was applied.
      *
      * @param {PIXI.CanvasRenderer} renderer - The renderer context to use.
      */
     popMask(renderer)
     {
+        this._freeCanvasOnNextFrame = this._freeCanvasOnNextFrame || [];
```
englercj: create this property in the ctor so that this doesn't cause a hidden class transition.

src/core/renderers/canvas/utils/CanvasMaskManager.js
```diff
@@ -153,12 +208,63 @@ export default class CanvasMaskManager
...
+        this._freeCanvasOnNextFrame.forEach((canvas) => this._pool.freeElement(canvas));
```
englercj: Need to set `this._freeCanvasOnNextFrame.length = 0`, or you are going to free extra canvases more than likely.

src/core/renderers/canvas/utils/CanvasMaskManager.js
```diff
@@ -153,12 +208,63 @@ export default class CanvasMaskManager
...
+
+        if (this._maskStates.length > 0)
+        {
+            const maskState = this._maskStates.pop();
+            const key = this._pool.getKey(renderer.context.canvas);
+            const maskCanvas = this._pool.popElementOrCreate(key, this._copyCanvas.bind(null, renderer));
```
englercj: Creating a bound function every frame isn't ideal. Either create one bind per renderer, or just add support in the popElement function to take args that it should pass to its factory.

englercj: Also looks like these aren't cleared.

src/core/renderers/canvas/utils/CanvasMaskManager.js
englercj: `this._freeCanvasOnNextFrame` is used above, and `this._clearOnNextFrame` is used here. Are they different?

bigtimebuddy:
Seems like there's been some pending review feedback since March. I'm going to close because of inactivity. If you @uorbe001 would like to re-submit or re-open, please do.

PR #2651, opened 2016-06-14
Docs update

themoonrat:
So, I've spent by evening going through every file generated by JSDoc, and updating the code to produce better where I could. Mostly it was adding / modifying param types and descriptions, changing @readOnly that doesn't work to @readonly that does, etc. etc, but there's a few other changes to get the docs looking good. For example, there were 2 classes called 'Prepare' even tho they were within WebGLPrepare.js and CanvasPrepare.js. So I update those class names to match the filename, and thus fixing the jsdocs.

I made notes of things that I didn't feel comfortable changing / looking at..... which are....

GraphicsData
member this.holes
function addHole 

mapWebGLBlendModesToPixi
constructor params

mapCanvasBlendModesToPixi
constructor params

mapWebGLDrawModesToPixi
constructor params

MaskManager
Number of function params needing a description
function pushScissorMask

Matrix
members a,b,c,d,tx,ty to be documented as to what they represent
function set

Quad
constructor stat param

deprecation.js

Texture.js
@event update valid?

SpriteMaskFilter
function apply

StencilMaskManager
filename and class name does not match
all functions

TransformStatic
function onChange
function updateSkew

WebGLRenderer
function render

PIXI.filters
esp ColorMatrixFilter - lot of functions that should be private?

PIXI.mesh
 
PIXI.particles

### DIFF
--- a/src/accessibility/AccessibilityManager.js
+++ b/src/accessibility/AccessibilityManager.js
@@ -81,7 +81,9 @@ function AccessibilityManager(renderer)
    	this.children = [];
 
    	/**
-     * pre bind the functions..
+     * pre-bind the functions
+	 *
+ 	 * @private
      */
    	this._onKeyDown = this._onKeyDown.bind(this);
    	this._onMouseMove = this._onMouseMove.bind(this);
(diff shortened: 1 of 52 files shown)


GoodBoyDigital:
You sir are a legend 👍

PR #2653, opened 2016-06-15
fix MovieClip texture update issue

ivanpopelyshev:
That bug appeared in my fork a few month ago, and @GoodBoyDigital made the same mistake with adding texture versions


GoodBoyDigital:
👍

PR #2656, opened 2016-06-17
Mesh Tint

ivanpopelyshev:
This had been tested on pixi examples.


ivanpopelyshev:
Fixing the tree

GoodBoyDigital:
we do indeed! 👍

PR #2669, opened 2016-06-23
Assignment by value for position/scale/skew/pivot

ivanpopelyshev:
http://www.html5gamedevs.com/topic/23380-position-in-latest-dev-build/

### DIFF
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -501,11 +501,7 @@ DisplayObject.prototype.setTransform = function(x, y, scaleX, scaleY, rotation, 
  */
 DisplayObject.prototype.destroy = function ()
 {
-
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
Please help me to solve it:

I want to destroy or set transform to null in displayObject.destroy();

1. Either we use "transform && transform.position" in getters
2. Either we modify the tests somehow

First way probably loses some performance.
Second way, any call of "displayObject.position" will throw an error after destroy().

I do not know what to do with it.

GoodBoyDigital:
Yo! If you can tweak the test so it passes - I can merge this one in 👍

ivanpopelyshev:
@GoodBoyDigital done.

GoodBoyDigital:
thanks man! 👍

PR #2675, opened 2016-06-27
Multiple ParticleContainers no longer incorrectly share properties, fixes #1953

### LINKED ISSUE #1953: Two particle containers are rendered incorrectly, if rotation is enabled
The following fiddle displays two bunnies side-by-side in two separate ParticleContainer instances:
http://jsfiddle.net/sergebat/wd03Lxt8/1/

However, if second particle container allows dynamic rotation for its kids, things in the first ParticleContainer got distorted in Chrome and Firefox at least on OSX:
http://jsfiddle.net/sergebat/6w9m9hft/1/

In canvas mode everything works like a charm:
http://jsfiddle.net/sergebat/yxzpcsp9/1/

Chrome version 43.0.2357.130 (64-bit) on OSX 10.9.5). Version: Pixi.js 3.0.7.

Screenshot:
screenshot 2015-07-06 00 12 41

rejemy:


rejemy:
I realized this bugfix probably deserves a little more explanation. The actual root of the problem is in ParticleRenderer.prototype.onContextChange, where the properties list is initialized:

```javascript
this.properties = [
        // verticesData
        {
            attribute:this.shader.attributes.aVertexPosition,
            size:2,
            uploadFunction:this.uploadVertices,
            offset:0
        },
        // positionData
        {
            attribute:this.shader.attributes.aPositionCoord,
            size:2,
            uploadFunction:this.uploadPosition,
            offset:0
        },
        // rotationData
        {
            attribute:this.shader.attributes.aRotation,
            size:1,
            uploadFunction:this.uploadRotation,
            offset:0
        },
        // uvsData
        {
            attribute:this.shader.attributes.aTextureCoord,
            size:2,
            uploadFunction:this.uploadUvs,
            offset:0
        },
        // alphaData
        {
            attribute:this.shader.attributes.aColor,
            size:1,
            uploadFunction:this.uploadAlpha,
            offset:0
        }
    ];
```

Javascript unwisely decided in this case to create one global singleton of this list of properties, and just assign a reference to this.properties each time this code is run. This means that every instance of ParticleRenderer shares the same list of properties. This becomes a problem when offsets are calculated and assigned, because whatever ParticleRenderer has been initialized most recently will overwrite the offsets of every other ParticleRenderer. If they have the same static and dynamic properties, the offsets are the same and no damage is done, thus why the problem only manifests if you have a ParticleRenderer with different dynamic properties than the rest.

It's possible there's a more elegant way to clone that properties list so that each ParticleRenderer has its own copy, but I just went with a simple solution. Any suggestions for improvement are welcome of course.

GoodBoyDigital:
nice! thanks @rejemy 👍

PR #2678, opened 2016-06-27
FastRect problem

ivanpopelyshev:
https://jsfiddle.net/whizzkid74/1wkpkspy/3/

After 1 second rectangle becomes "fast" (due to updateBounds()), and it stops moving

### DIFF
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -501,11 +501,7 @@ DisplayObject.prototype.setTransform = function(x, y, scaleX, scaleY, rotation, 
  */
 DisplayObject.prototype.destroy = function ()
 {
-
-    this.position = null;
-    this.scale = null;
-    this.pivot = null;
-    this.skew = null;
+    this.transform = null;
 
     this.parent = null;
 
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -208,6 +208,7 @@ Object.defineProperties(DisplayObject.prototype, {
 
     /**
      * The pivot point of the displayObject that it rotates around
+     * Assignment by value since pixi-v4.
      *
      * @member {PIXI.Point|PIXI.ObservablePoint}
      * @memberof PIXI.DisplayObject#
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -175,6 +175,7 @@ Object.defineProperties(DisplayObject.prototype, {
 
     /**
      * The coordinate of the object relative to the local coordinates of the parent.
+     * Assignment by value since pixi-v4.
      *
      * @member {PIXI.Point|PIXI.ObservablePoint}
      * @memberof PIXI.DisplayObject#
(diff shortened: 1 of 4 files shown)


ivanpopelyshev:
Added fix for observablePoint in the same branch, I'm sorry.

ivanpopelyshev:
@englercj merge it. Its just two files difference. pleeeeeease

src/core/math/ObservablePoint.js
```diff
@@ -66,6 +67,15 @@ Object.defineProperties(ObservablePoint.prototype, {
 });
 
 /**
+ * Creates a clone of this point. Clone is not observable!
+ *
+ * @return {PIXI.Point} a copy of the point
+ */
+ObservablePoint.prototype.clone = function() {
+    return new Point(this._x, this._y);
```
englercj: SHouldn't `ObservablePoint#clone()` return an `ObservablePoint`?

ivanpopelyshev:
@englercj That thing must be backwards-compatible, users dont even know that we started use Observables as a default implementation. Also, there's a question, observable by whom?

clone() works as a value copying.

var p = displayObject.scale.clone();
displayObject2.scale = p;

it will work just fine with point.

englercj:
But that doesn't mean the API is correct, if an `ObservablePoint` can masquerade as a `Point` then there should be no issue with returning an `ObservablePoint` from the clone method.

ivanpopelyshev:
Which callback will it have? This will be ugly. ObservablePoint always has the owner.

```js
this.cb.call(this.scope);
```

englercj:
I assume, since it is a clone, that it would be identical?

ivanpopelyshev:
@englercj Since we are talking about APIs, let me reference https://api.jquery.com/clone/ - by default, cloned object have empty list of listeners.

In our implementation, Point with listener is ObservablePoint, and  ObservablePoint without listener is Point. Thus, cloning must always return the Point.

englercj:
Luckily we don't have to worry about DOM here, and aren't working on jQuery. Nor are we talking about *listeners* (a totally different pattern).

I see the options as:

1. Make it return an identical observable point, including the callback property.
2. Make it return an observable point with no callback.
 - Adjustment to the class would be needed to support this case, which IMO it should support anyway because it is *observable* and not *observed*.
3. Make it return a Point object.

`1` follows the Principle of Least Surprise and is exactly what I would expect.
`2` still makes some sense, but is no longer a clone but just a `cloneXY` or `createFrom(point)`.
`3` makes no sense to me at all, and I fail to see a good reason for it. The argument "we didn't finish the api for ObservablePoint to be able to do all the things it may need to" is not a good excuse to me.

ivanpopelyshev:
This is our problem: http://pixijs.github.io/examples/index.html?s=filters&f=displacement-map.js&title=Displacement%20Map&v=dev . I think we can just remove "clone()" then, we already changed behaviour that way 

```js
var p = obj1.position;
obj2.position = p;
```

is not working like before. We can survive without clone()

clark-stevenson:
Part of me has a vote for Option 1 by @englercj 

If I clone something, I expect an identical result.  Be that Point ObservablePoint or Dolly The Sheep.

In saying that, I have no idea of the use case for wanting to clone that inner ObservablePoint callback other than to get unexpected duplicated behavior.

ivanpopelyshev:
Then we'll have to add more ifs in ObservablePoint

PR #2681, opened 2016-06-28
bitmap font resource loading issue

adireddy:
Not sure why `resource.isDataUrl` was added but it's breaking the bitmap font resource url.

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -297,10 +297,10 @@ var CONST = {
      * @static
      * @constant
      * @property {object} PRECISION
-     * @property {number} PRECISION.DEFAULT='mediump'
-     * @property {number} PRECISION.LOW='lowp'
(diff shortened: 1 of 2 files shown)


adireddy:
Related to another PR #2680.

englercj:
If you are ever curious why some line of code exists, you can check the blame:

https://github.com/pixijs/pixi.js/blame/dev/src/loaders/bitmapFontParser.js

Which will leave you to this commit:

https://github.com/pixijs/pixi.js/commit/5e94eb4caf6dfeff9324ce013d7541980f041306

Which leads to this PR:

https://github.com/pixijs/pixi.js/pull/2630

Would you mind adding some unit tests for these (both data and non-data urls) so we can ensure they don't break again in the future?

adireddy:
Sure thing.

Closing this for now and will raise another PR after checking the previous commits and with unit tests.

Issue #2677, opened 2016-06-27
Plane and Texture.fromFrame not working ?


whizzkid:
Was trying to make a 9slice scaling example and found out Plane and Texture arent working together nicely.

When I create a Plane directly with a Texture, it doesn't work.

When I use a RenderTexture, it does work..

example:
using latest PIXI dev (4.x)
using @ivanpopelyshev 's version

https://jsfiddle.net/whizzkid74/8p2s3jz8/8/

GoodBoyDigital:
we have a nine slice :D

PR #2688, opened 2016-06-29
Added NineSlicePlane.js to allow 9Slice scaling

whizzkid:
Hope the doc's are OK, I couldnt get width and height to be displayed. the override tag is not working.


ivanpopelyshev:
That will work only if texture is not in atlas. Also, sometimes texture might be rotated. I can add special function in TextureUvs that can take "uvs" as param and modify it that way it actually corresponds to texture and not its base.

It will look something like that snippet from pixi-spine:

```js
//this.regionUVs are your UVs
//r is textureUVs object
//this.uvs is result
for (var i = 0; i < n; i += 2)
{
   var u = this.regionUVs[i], v = this.regionUVs[i+1];
   u = (u * w2 - x) / w1;
   v = (v * h2 - y) / h1;
   this.uvs[i] = (r.x0 * (1 - u) + r.x1 * u) * (1-v) + (r.x3 * (1 - u) + r.x2 * u) * v;
   this.uvs[i+1] = (r.y0 * (1 - u) + r.y1 * u) * (1-v) + (r.y3 * (1 - u) + r.y2 * u) * v;
}
```

ivanpopelyshev:
Never mind, its not your problems, I will change Mesh for that later :)

whizzkid:
Cool.
I've used a RenderTexture in my development because Plane did not work with Texture.fromFrame #2677

I've also noticed a weird alpha bug, where setting alpha on the original sprite I used to render a texture from after It's been rendered to texture, changes the alpha on both the original and the clone.
but i haven't had time yet to cleanly reproduce this bug.
wanted to finish this 9slice first :)

src/mesh/NineSlicePlane.js
```diff
@@ -0,0 +1,240 @@
...
+ *    +---+----------------------+---+
+ *    |   |                      |   |
+ *  D | 4 |          5           | 6 |
+ *    |   |                      |   |
+ *    +---+----------------------+---+
+ *  E | 7 |          8           | 9 |
+ *    +---+----------------------+---+
+
+ *  When changing this objects width and/or height:
+ *     areas 1 3 7 and 8 will remain unscaled.
```
englercj: 1, 3, 7, and 9 I think you mean.

src/mesh/NineSlicePlane.js
```diff
@@ -0,0 +1,240 @@
...
+ * </pre>
+ *
+ * @class
+ * @extends PIXI.mesh.Plane
+ * @memberof PIXI.mesh
+ * @param {PIXI.Texture} texture - The texture to use on the NineSlicePlane.
+ * @param {int} [leftWidth=10] size of the left vertical bar (A)
+ * @param {int} [topHeight=10] size of the top horizontal bar (B)
+ * @param {int} [rightWidth=10] size of the right vertical bar (C)
+ * @param {int} [bottomHeight=10] size of the bottom horizontal bar (D)
```
englercj: Not sure if the names, descriptions, and letters actually match your diagram above.

src/mesh/NineSlicePlane.js
```diff
@@ -0,0 +1,240 @@
...
+ *
+ */
+function NineSlicePlane(texture, leftWidth, topHeight, rightWidth, bottomHeight)
+{
+	Plane.call(this, texture,4,4);
+	var DEFAULT_BORDER_SIZE= 10;
```
englercj: pull this constant out of the class and to the top of the file.

src/mesh/NineSlicePlane.js
```diff
@@ -0,0 +1,240 @@
...
+ * @param {PIXI.Texture} texture - The texture to use on the NineSlicePlane.
+ * @param {int} [leftWidth=10] size of the left vertical bar (A)
+ * @param {int} [topHeight=10] size of the top horizontal bar (B)
+ * @param {int} [rightWidth=10] size of the right vertical bar (C)
+ * @param {int} [bottomHeight=10] size of the bottom horizontal bar (D)
+ *
+ */
+function NineSlicePlane(texture, leftWidth, topHeight, rightWidth, bottomHeight)
+{
+	Plane.call(this, texture,4,4);
```
englercj: Fix the spacing with these `4`s please.

src/mesh/NineSlicePlane.js
LukaszGrela: Nice, but could this accept `PIXI.Rectangle` to define the slicing?

whizzkid: It could, but that would not really be user friendly.
you'd have to calculate the widths that you want yourself before entering,

var slicer = new PIXI.mesh.NineSlicePlane(myTex, new PIXI.Rectangle(10,20, myTex.width - 10 - actualSizeOfRightBorder, myTex.height - 20 - actualSizeOfBottomRow));


also nitpicky minimally slower and less memory efficient too :package:

whizzkid: It would also make changing the sizes of the borders after creation more difficult.

src/mesh/NineSlicePlane.js
```diff
@@ -0,0 +1,241 @@
...
+	uvs[4] = uvs[12] = uvs[20] = uvs[28] = 1 - this._uvw * rightWidth;
+	uvs[9] = uvs[11] = uvs[13] = uvs[15] = this._uvh * topHeight;
+	uvs[17] = uvs[19] = uvs[21] = uvs[23] = 1 - this._uvh * bottomHeight;
+
+	/**
+	 * The width of the left column (a)
+	 *
+	 * @member {number}
+	 */
+	this.leftWidth = leftWidth || DEFAULT_BORDER_SIZE;
```
bigtimebuddy: If `leftWidth` is `0` then this will always default to 10. This should support 3 or 6 slice scaling.

whizzkid: good call, changed it. now defaults only get selected if values are undefined.

rkrasky:
@ivanpopelyshev sorry... me again. Did you rewrite the mesh class so texture from atlas will work?
@whizzkid in the width and height setter there are these lines:

``this._uvw = 1 / value * value / this._origWidth;``

respectively

``this._uvw = 1 / value * value / this._origHeight;``

Why don't write `1 / this._origWidth` and `1 / this._origHeight`?

Thanks a lot, both of you for your great work!

whizzkid:
@rkrasky you are absolutely correct. I've changed the code. now its only set once in the constructor.

I've also created a little jsFiddle using my branch of pixi to demonstrate.

https://jsfiddle.net/whizzkid74/1wkpkspy/

rkrasky:
Great! Thanks a lot! Do you have a clue how to get textures from atlas to work?

whizzkid:
I think @ivanpopelyshev is working on that.
right now you can use a rendertexture as a workaround.

ivanpopelyshev:
Yeah, I'll add special function and PR for that

bigtimebuddy:
@whizzkid Could you please not check-in the bin (built) files per the contribution guidelines?

whizzkid:
ah I didnt know it would automatically include it in the pr because I didn't include it in the original.. live and learn... will remove soon.

englercj:
@whizzkid FYI: Anything you commit to your `dev_whizzkid_NineSlicePlane` branch will show up in this PR.

bigtimebuddy:
@whizzkid, We don't actually want to delete the bin files, we just don't want those changes. Maybe you should rollback 2 commits and then manually re-add your changes?

whizzkid:
Ok rollbacks dont work. can't commit anything after i try. I'm gonna make a new branch with the latest version and just make a new PR.

whizzkid:
new PR #2715

Issue #2683, opened 2016-06-28
Pixi Accessibility for mobile


scottmcdonnell:
Pixi accessibility in v3 is not working for VoiceOver on Safari or Talkback for Android as the invisible dom divs are only added when the Tab button is pressed.

Pixi v4 adds a new `createTouchHook` method which adds a new div and when it gets focus it then activates the additional button divs. Nice.

This works for IOS VoiceOver but not for Android as TalkBack [does not fire a focus event](https://www.paciellogroup.com/blog/2014/04/javascript-touchpointer-events-research-notes-screenreaders-faked-mouse-events-touchscreens/) until button activation via double tap.

To get the accessibility working for Android the divs need to be created and onscreen all of the time. Maybe there could be an accessiblity mode that can be switched between the levels required?
`PIXI.accessibility.level = 'none' | 'focus' | 'always'`

Also should the [pixijs/pixi-accessiblity](https://github.com/pixijs/pixi-accessibility) repo be closed if the code is now in the main repo? Confused me anyway!

PR #2694, opened 2016-07-01
Update for IOS Voiceover,  Android Talkback and Kindle VoiceView

scottmcdonnell:
Update for accessibility to address issue #2683
There is a [demo here](http://scottmcdonnell.github.io/pixi-accessibility/) 

This always adds the divs on mobiles that do not support focus events in accessibility mode.
Additionally it has an updateTabOrders to correctly order the divs for tabbing.

This AccessibilityManager has been tested with PIXI v4 and PIXI v3 across:
- IOS VoiceOver for IOS
- Android TalkBack
- Kindle VoiceView


GoodBoyDigital:
Thanks for this PR @scottmcdonnell ! 
Nice to see you got it working on all 3 devices 👍 
One thing to quickly check - the way the PR is set up, does that mean that the accessibility layer will always be there on android devices?

scottmcdonnell:
Yes, we tried a load of different ways and for android only "always on" worked for us as there is no way we could find of detecting the screen reader enabled. 

The alternative I would consider in the future would be to roll all HUD buttons in plain html over the pixi layer so it's fully accessible without any potential performance hit of post render action every frame.

JiDW:
This will however continue to break on WebGL views that does not support DOM, like CocoonJS.

https://github.com/pixijs/pixi.js/issues/2660

Is it possible to add a way to disable to accessibility manager at least please, if you can't detect when it's needed or not ?

scottmcdonnell:
I also think it would be a good idea to be able to switch the accessibility off, that would work for Cacoonjs or your own html HUD.  I didn't add this to this PR to keep it focused but +1 for adding to pixi.js

englercj:
Can you rebase this onto the latest `dev` branch please @scottmcdonnell? Thanks!

scottmcdonnell:
Sorry was busy on another project, rebased now I to latest dev.

GoodBoyDigital:
Closing this for now. Rolling a new Accessibility manager for v5 👍

PR #2698, opened 2016-07-04
sprite json data "sourceSize" change to a optional data

1056ng:


ivanpopelyshev:
Hm. Why is it optional?

1056ng:
@ivanpopelyshev 
coz, I has recognized that Texture class's `orig` is optional, right?

https://github.com/pixijs/pixi.js/blob/dev/src/core/textures/Texture.js#L99

ivanpopelyshev:
I do not know if it is optional in spritesheet format.

1056ng:
> I do not know if it is optional in spritesheet format.

incidentally, `v3` is working.
https://github.com/pixijs/pixi.js/blob/master/src/loaders/spritesheetParser.js#L85

ivanpopelyshev:
I do not understand what's the bug and what are you trying to fix. Can you be more clear please?

1056ng:
Sorry for that, I tried to reduce in json size. But, it's not good technique.

ivanpopelyshev:
Ah, that's fine. Just make your own parser for your format, and ask me if you will have problems with creating textures

PR #2709, opened 2016-07-08
text padding possible fix

ivanpopelyshev:
need to test it

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -407,14 +407,14 @@ Text.prototype.updateTexture = function ()
 
     texture.baseTexture.width = this.canvas.width / this.resolution;
     texture.baseTexture.height = this.canvas.height / this.resolution;
-    texture.orig.width = texture._frame.width = this.canvas.width / this.resolution;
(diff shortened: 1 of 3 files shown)


src/core/text/Text.js
```diff
@@ -125,14 +128,15 @@ Object.defineProperties(Text.prototype, {
...
         },
         set: function (value)
         {
             this.updateText(true);
 
-            this.scale.y = value / this._texture._frame.height;
-            this._height = value;
+            var sign = utils.sign(this.scale.x) || 1;
+            this.scale.x = sign * value / this.texture.orig.width;
+            this._width = value;
```
flashist: Is it ok that we change values of the _width and scale.x properties inside of the height setter? Was it planned this way? It looks like this is a typo and this part of the setter was just copy-pasted from the width setter.

GoodBoyDigital: Great catch @flashist aka Mr Eagle eyes! Just pushed a fix and a unit test, we should be covered now. Thanks! 👍

flashist: Thanks for the fastest fix ever!

PR #2715, opened 2016-07-13
Add NineSlicePlane.js to enable 9Slice scaling, 

whizzkid:
without the bin folder this time.

### DIFF
--- a/src/mesh/NineSlicePlane.js
+++ b/src/mesh/NineSlicePlane.js
@@ -1,0 +1,241 @@
+var DEFAULT_BORDER_SIZE= 10;
+
+var Plane = require('./Plane');
+
+/**
+ * The NineSlicePlane allows you to stretch a texture using 9-slice scaling. The corners will remain unscaled (useful
+ * for buttons with rounded corners for example) and the other areas will be scaled horizontally and or vertically
+ *  
+ *```js
+ * var Plane9 = new PIXI.NineSlicePlane(PIXI.Texture.fromImage('BoxWithRoundedCorners.png'), 15, 15, 15, 15);
+ *  ```
+ * <pre>
+ *      A                          B
+ *    +---+----------------------+---+
+ *  C | 1 |          2           | 3 |
+ *    +---+----------------------+---+
+ *    |   |                      |   |
+ *    | 4 |          5           | 6 |
+ *    |   |                      |   |
+ *    +---+----------------------+---+
+ *  D | 7 |          8           | 9 |
+ *    +---+----------------------+---+
+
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
This continues the work on #2688

GoodBoyDigital:
This PR is great @whizzkid! - Keen to roll in, but have one thought.. We should probably create a 'renderCanvas' part to this. Mainly as the plane rendering implementation uses triangle / cropping in and can be quite heavy. A much lighter implementation could be used for canvas (using drawImage) What do ya think?

whizzkid:
I need to do checks for a size of minimal 1, otherwise drawImage draws nothing, or sometimes draws faint lines when when using slices with 0 width 

So i've made the drawSegments  a separate function for readability.
but, I see in the original Mesh I also need to multiply the source coords with Base.Resolution.

Now I wanted to add this to the function but then JSLint says I have too many parameters :P

I could halve the indexes that I use and use  index+1  to get the y-index of a coordinate..
or just provide the base, and get texturesource, width and height inside of drawSegment, but that would also be a little slower...

I could also do everything inline, which would make _renderCanvas bigger and uglier.

or change the projects jslint setting so that you're allowed to use a few more parameters  :)


And another optimisation could be to prerender the uvs * width, but that would require a little more memory allocated per NineSlicePlane...


What do you guys think? :)

GoodBoyDigital:
Nice! this looks all good to me :)

staff0rd:
I've found two issues with this. Now that it's merged, what's the process, add the issues as a continuation of this (closed) issue or create a new one?

### 1. Test warning when building;
~~~
src\mesh\NineSlicePlane.js
--------------------------
 W098  Line 4, Col 5  'Mesh' is defined but never used.
× 1 problem in 1 file
~~~

### 2. At runtime;
`Uncaught TypeError: Cannot read property 'x0' of null`, [here](https://github.com/pixijs/pixi.js/blob/3c3ce1200e5d0ffceb7da70c52365d841741d844/src/mesh/Plane.js#L78) because `texture._uvs` is undefined.

whizzkid:
@staff0rd I get no such errors? Just pulled the dev branch and built it.. 0 errors or warnings.

something wrong with your setup maybe? for example,  Mesh is used in line 23 so I don't know why you'd get that error.

And the runtime error is related to a separate issue #2677 which has to do with Plane not working properly with Texture.fromFrame.

I use a renderTexture as a workaround. but the issue is not in the NineSlicePlane.

whizzkid:
Removed the line. and made a PR #2751

PR #2720, opened 2016-07-14
BaseTexture used for text now reports correct realWidth and realHeight

themoonrat:
Was trying to work out how much texture space I was taking up with Text, and noticed that measuring that with a resolution of 1 was the same as resolution of 2. Turns out that realWidth and realHeight weren't being updated, so were always reported the default canvas creation size.


GoodBoyDigital:
nice 👍

PR #2730, opened 2016-07-18
Create .doclets.yml

lipp:
Adds documentation to doclets.io service. Future tags and master branch will be published **automatically**.

[Preview generated with my account](https://doclets.io/lipp/pixi.js/add-to-doclets)

Doclets also supports "live/editable" examples for the browser, like [this](https://doclets.io/lipp/acme-jsdoc-example/master#dl-greet)! Just replace the @example tag with @playground tag.

@adireddy    If you like this, you are very welcomed to login to https://doclets.io with GitHub OAuth. If you don't like it, I'd be happy to hear what you don't like about it :)

Steps required (3min):
1) @adireddy   logs in to doclets.io
2) @adireddy   adds pixi.js
3) @adireddy   merges this PR

Doclets loads big projects like pixi.js slow as it renders all API on one single page. This will change soon!

### DIFF
--- a/.doclets.yml
+++ b/.doclets.yml
@@ -1,0 +1,3 @@
+dir: src
+articles:
+  - Readme: README.md


ivanpopelyshev:
And it slowed down my new MSI laptop :) Hope it will change.

englercj:
Seems like a cool service, if they fix that obvious issue I don't see a problem with putting pixi in there. Until then, it seems like not that useful.

Additionally, I don't think we would change our `@example` tags to `@playground` since it is non-standard. Would be great if they just supported the standard though!

lipp:
@ivanpopelyshev  @englercj sorry for the performance / load issue (pixi.js is one of the biggest projects so far... nice work on the docs by the way!).  The "one page" approach just doesn't work very well for projects this big, so doclets will split docs around modules, categories and explicit/implicit namespaces. 

@englercj The standard @example tag works too, of course.  😄

englercj:
Closing this since we will not be doing this. We want custom doc output, not a hosted service. Thanks for bringing this up though!

PR #2732, opened 2016-07-19
simple transform refactoring

ivanpopelyshev:
TransformBase is the parent

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -20,7 +20,6 @@ var core = module.exports = Object.assign(require('./const'), require('./math'),
     Transform:              require('./display/Transform'),
     TransformStatic:        require('./display/TransformStatic'),
     TransformBase:          require('./display/TransformBase'),
-    TransformManual:        require('./display/TransformBase'),
 
     // sprites
     Sprite:                 require('./sprites/Sprite'),
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -19,7 +19,8 @@ var core = module.exports = Object.assign(require('./const'), require('./math'),
     Container:              require('./display/Container'),
     Transform:              require('./display/Transform'),
     TransformStatic:        require('./display/TransformStatic'),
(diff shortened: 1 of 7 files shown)


src/core/index.js
```diff
@@ -19,7 +19,8 @@ var core = module.exports = Object.assign(require('./const'), require('./math'),
     Container:              require('./display/Container'),
     Transform:              require('./display/Transform'),
     TransformStatic:        require('./display/TransformStatic'),
-    TransformManual:        require('./display/TransformManual'),
+    TransformBase:          require('./display/TransformBase'),
+    TransformManual:        require('./display/TransformBase'),
```
englercj: Is this needed?

englercj:
No builds please.

src/core/index.js
ivanpopelyshev: Yes, temporarily, for pixi-spine, its using TransformManual right now. I'll remove that as soon as i change pixi-spine to TransformBase.

ivanpopelyshev:
Fixed, thank you!

src/core/index.js
bigtimebuddy: Maybe a deprecation for TransformManual is more appropriate here?

ivanpopelyshev: Its v4-rc2 thing, it is used only in pixi-spine, we dont have to deprecate it, we'll just remove it before release of v4. I just dont want pixi-spine to be broken for a few days.

bigtimebuddy: I hear you, but it would be much better to add it to src/deprecation.js. This won't break pixi-spine and will actually give warnings for you to update pixi-spine. We can remove this specific deprecation with v4 release, but if someone forgets, it's not a big deal.

englercj:
Looks good to me :+1:

ivanpopelyshev:
@englercj WOW! Thank you! Can you please look at 2733 too?

GoodBoyDigital:
👍

Issue #2734, opened 2016-07-20
Weird issue when setting Sprite.visible to false


RanzQ:
When using a sprite sheet and `Sprite.fromFrame` I noticed that if I set `visible = false` for a single sprite, all other sprites using the same texture aren't rendered. And the weird thing is that this doesn't happen on my desktop Chrome but on a RPi with [WebKit](https://github.com/Metrological/WebKitForWayland/) (WebGL enabled).

Setting visibility for Container works ok, just in case someone ends up with a similar problem.

Any idea why this could happen?

ivanpopelyshev:
pixi v3 or v4?

RanzQ:
@ivanpopelyshev v4, 8fdb17a92f8f1fe015e59d153274c1be0b7ee6f8 to be exact

ivanpopelyshev:
It looks like issue with Sprite._textureID, somehow other sprites dont see that it was updated.. What if you preload stuff with PIXI.loader?

RanzQ:
@ivanpopelyshev I'm using the loader and start rendering after all textures (3 json sheets now) have loaded. I noticed I was initializing one sprite like this:
```
  this.addGreenFrameToStage = () => {
    let greenFrame = PIXI.Sprite.fromFrame('green-frame')
    // greenFrame.visible = false
    greenFrame.alpha = 0.0

    this.stage.addChild(greenFrame)
  }
```
Commenting that line made all the other sprites render again. Touch events were still handled for the invisible sprites, so I guessed it must be some visibility issue.

ivanpopelyshev:
Do you use any esoterics like ParticleContainer?

RanzQ:
No, just `new PIXI.Container()` as stage and some child containers as well.

ivanpopelyshev:
can you make simple demo of that? I think its something about multitexturing. I just need to see the actual code that works on PC but doesnt work correctly on mobiles.

ivanpopelyshev:
Waaait, i actually have an idea. Wait just one moment..

ivanpopelyshev:
Try that one: https://dl.dropboxusercontent.com/u/46636530/pixijs/multutex/pixi.js

RanzQ:
@ivanpopelyshev that works! :)

GoodBoyDigital:
fixed 👍

Issue #2731, opened 2016-07-19
Pixi v4 - iOS flickering spritesheets


guilanier:
Hi there,
I have an issue on ios with the latest pixi build (dev branch): playing a spritesheet on ios either by setting the frame of the texture manually or by using the built-in PIXI.MovieClip is making the sprite flicker each frame. The problem doesn't happen on desktop or any other devices not running on iOS. You can see this issue there if you are on iOS : https://dl.dropboxusercontent.com/u/29261094/pixi-spritesheet/index.html

ivanpopelyshev:
Try this one please: https://dl.dropboxusercontent.com/u/46636530/pixijs/multutex/pixi.js

guilanier:
Hey @ivanpopelyshev I've just tried it and put it there : https://dl.dropboxusercontent.com/u/29261094/pixi-spritesheet2/index.html
The problem is remaining on iOs devices

ivanpopelyshev:
pasting from the other issue, try it 

PIXI.glCore.VertexArrayObject.FORCE_NATIVE = true;

guilanier:
That is looking better: https://dl.dropboxusercontent.com/u/29261094/pixi-spritesheet3/index.html

 thanks a lot for that, it is still flickering though but less than previously

ivanpopelyshev:
And if you use both my pixi.js and that line?

guilanier:
yes absolutely, that's what happen on the last link

englercj:
Can you retest this with the latest dev build and see if that has fixed your issue?

You can see how to download the latest build [in the FAQs](https://github.com/pixijs/pixi.js/wiki/FAQs#where-can-i-get-a-build).

Thanks!

themoonrat:
@guilanier as above.. could you check if this is still an issue on the latest PIXI please? Thanks

GoodBoyDigital:
Hi there! Closing this issue for now due to its inactivity. Feel free to give us a poke if you would like this issue reopened. Thanks 👍

Issue #2726, opened 2016-07-16
iOS 10 beta - WebGLRenderer renders PNGs with alpha incorrectly.


POlITIC:
Create texture from any png image that has transparent areas in it.
Create a sprite from this texture and render it with WebGLRenderer.
Run this on iOS10 beta and observe results:
<img src="https://cloud.githubusercontent.com/assets/1701819/16894466/b79c1df6-4b5f-11e6-9d0d-7ecaec8f326b.PNG" height="200">

Result is that WebGLRenderer draws alpha areas with some artifacts.

The same code is run on iOs 9.3.2 - works fine:
<img src="https://cloud.githubusercontent.com/assets/1701819/16894467/b79d185a-4b5f-11e6-8e48-a0a283ff722b.jpg" height="200">

Code snippet: https://jsfiddle.net/xshn8nvu/1/

Or just run this(http://www.goodboydigital.com/pixijs/examples/16/) pixi sample on iOs10.

GoodBoyDigital:
Hello! Thanks for sharing, I noticed this issue too when I installed the beta a little while back. Im 99% sure its a safari bug that will be rectified by the time iOS10 is officially released. Lets keep this open until its sorted.

ivanpopelyshev:
Try this one please: https://dl.dropboxusercontent.com/u/46636530/pixijs/multutex/pixi.js

POlITIC:
Seems like this issue is now fixed on iOS side.Can be tested on latest beta. So I'm closing this issue.

PR #2735, opened 2016-07-20
Multitexturing DERP

ivanpopelyshev:
Precision errors can be very strange on some devices. Especially if you are rendering 3d transform. This is part of gameofbombs/pixi.js fork, I did not think that it will be needed in 2D scenes too.

Related to #2734, may be #2731 #2726 too.


englercj:
I was wondering if interpolation would cause problems when I was the ID being passed in as a float. Nice catch.

ivanpopelyshev:
It did cause problems with 3D transforms in my fork, and I just didn't think that it can affect 2D stuff. That's how I noticed it that fast :)

GoodBoyDigital:
good catch @ivanpopelyshev 👍

PR #2736, opened 2016-07-20
extra conditions for bounds

ivanpopelyshev:
When I made previous boundsBuilder PR, I wanted these conditions to be met, but I didnt specify it. @GoodBoyDigital added many tests (THANK YOU! :+1:), introduced "skipUpdate" parameter, but he also sweeped all my conditions as a garbage.

In the future we will have to deal with one big performance problem: culling. I believe that it will require these conditions.

BoundsBuilder: Use it only as a temporary object per one calculation or per one module. If browser profiler shows more than 10 elements of that in memory, something is clearly wrong

DisplayObject.transform: transform shows global coordinates that were calculated after last updateTransform() pass. It may contain some local stuff only if a parent is cachedAsBitmap or inside generateTexture. getLocalBounds() does not use it.

DisplayObject.calculateBounds: does not affect anything except passed builder and may be some cached or pooled bounds for Mesh or Graphics.

### DIFF
--- a/src/core/display/BoundsBuilder.js
+++ b/src/core/display/BoundsBuilder.js
@@ -6,6 +6,9 @@ var math = require('../math'),
  * Axis-Aligned Bounding Box
  * It is not a shape! Its mutable thing, no 'EMPTY' or that kind of problems
  *
+ * Condition: Use it only as a temporary object per one calculation or per one module.
+ * If browser profiler shows more than 10 elements of that in memory, something is clearly wrong
+ *
  * @class
  * @memberof PIXI
  */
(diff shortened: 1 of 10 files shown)


ivanpopelyshev:
Yes, I see that my implementation of Container.getBounds() is a bit big, but please take into account that previous implementation that method took care of _currentBounds caching of ALL CHILDREN, while @GoodBoyDigital edition cant actually do that. 

It will affect performance of culling and filters.

ivanpopelyshev:
@bQvle said its broken. I'm trying to fix it.

PR #2739, opened 2016-07-21
Add exception handling to `utils.canUseNewCanvasBlendModes`

nikolowry:
Multiple sites using pixi have been throwing the following error:
`Uncaught TypeError: Cannot read property 'data' of undefined`.

----

Chromium v53 x64 Linux  Kernel 4.6

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -93,8 +93,15 @@ var utils = module.exports = {
         context.globalCompositeOperation = 'multiply';
         context.drawImage(magenta, 0, 0);
         context.drawImage(yellow, 2, 0);
-
-        var data = context.getImageData(2,0,1,1).data;
+        
+        try 
+        {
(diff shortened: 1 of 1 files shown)


nikolowry:
Also want to add some context to this - I use a few extensions to prevent canvas fingerprinting and they seem to be the cause of my error. 

Would prefer to keep my privacy, while also being able to view sites that use pixi.

englercj:
Instead of try catch:

```js
var imageData = context.getImageData(2,0,1,1);

if (!imageData)
{
    return false;
}

var data = imageData.data;

return //...
```

englercj:
Oh shoot just realized this PR was to `master`, we only accept PRs to `dev`:

https://github.com/pixijs/pixi.js/blob/master/CONTRIBUTING.md

nikolowry:
I'll close and reopen another - thanks

PR #2740, opened 2016-07-22
Add exception handling to `canUseNewCanvasBlendModes`

nikolowry:
Multiple sites using pixi have been throwing the following error:
`Uncaught TypeError: Cannot read property 'data' of undefined.`

Additional context - I use a few extensions to prevent canvas fingerprinting and they seem to be the cause of my error.


ivanpopelyshev:
exception in getImageData also ruins PIXI.Text

englercj:
https://github.com/pixijs/pixi.js/blob/dev/src/core/text/Text.js#L496

ivanpopelyshev:
TorBrowser does not like that ;)

GoodBoyDigital:
nice! 👍

PR #2741, opened 2016-07-22
Able to find out the real width and height of a base render texture

themoonrat:
Previously these weren't updated so reported back as 100/100


GoodBoyDigital:
nice catch @themoonrat  👍

PR #2749, opened 2016-07-28
fix for lineHeight

rkrasky:
fix for lineHeight, canvas is actually too small if the line height is smaller than the font size.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -280,7 +280,7 @@ Text.prototype.updateText = function ()
     // calculate text height
     var lineHeight = this.style.lineHeight || fontProperties.fontSize + style.strokeThickness;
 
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
Hi @rkrasky ! Thanks for this, would you mind sharing a fiddle of this in action please? Just so I can confirm the fix. Cheers man! 👍

rkrasky:
Sure! Here a two fiddles, one with current v4 version and the other with modified height calculation:

working (current v4 with height fix):
https://jsfiddle.net/hgqab1j1/

not working (current v4):
https://jsfiddle.net/oo4jwc6v/

GoodBoyDigital:
perfect! thanks for that - looking good 👍

PR #2752, opened 2016-07-29
MaskManager - Sprite Mask Resolution

siwel:
Currently when applying a mask to a -x2 Sprite, the Sprite resolution is lowered to that of the mask.


GoodBoyDigital:
nice catch! 👍

PR #2755, opened 2016-08-01
Added webgl call to setBlendMode to apply mesh blending effects

neutrino84:


ivanpopelyshev:
Yeah! So many problems will be solved when @GoodBoyDigital merge it :)

GoodBoyDigital:
POW! 👍

PR #2756, opened 2016-08-01
Adding performance caveat context option

rkrasky:
http://www.html5gamedevs.com/topic/16425-ie11-software-rendering/#comment-92760
https://msdn.microsoft.com/en-us/library/dn798645(v=vs.85).aspx

If IE or Edge shows up the message "WEBGL11258: Temporarily switching to software rendering to display webGL content", WebGL detection should fail because the software rendering is slow as hell.


GoodBoyDigital:
slick!

PR #2760, opened 2016-08-03
Dev bounds tweak

GoodBoyDigital:
tidied up bounds - less variable
fixed caching of bounds if called multiple times in with skipUpdate


bigtimebuddy:
Like this cleanup 👍

PR #2763, opened 2016-08-04
Adding CanvasRender worldAlpha

rkrasky:
Got an issue of the alpha value of NineSlicePlane. Adding  worldAlpha value to context if rendered via CanvasRenderer.


GoodBoyDigital:
nice one @rkrasky 👍

PR #2764, opened 2016-08-04
Adds .npmignore file

bigtimebuddy:
**Overview**

* Adds a list of files that should be excluded from NPM (mostly Gulp, tests and other build files).

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,12 @@
+bower.json
+.gitignore
+.editorconfig
+.jshintrc
+node_modules
+.travis.yml
+gulpfile.js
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
nice one @bigtimebuddy 👍

englercj:
This isn't necessary, we already specify a whitelist in package.json; don't also need a blacklist:

https://github.com/pixijs/pixi.js/blob/master/package.json#L24-L31

bigtimebuddy:
Oops, sorry missed that @englercj! I'll revert.

PR #2782, opened 2016-08-08
Preprocess

bigtimebuddy:
## Overview

Updates to [pixify@v1.2.0](https://github.com/pixijs/pixify/releases/tag/v1.2.0) which adds support for using preprocess to add comment blocks.  In the future, this feature could more broadly be used to provide better debugging features in the un-minified library (**bin/pixi.js**) and remove unnecessary code and logging from release library (**bin/pixi.min.js**). 

This PR utilizes this feature to remove the deprecations from the release build of the library.

### Debug Example
```js
// @if DEBUG
console.error('only show in debug mode');
// @endif
```
### Release Example
```js
// @if RELEASE
throw 'fail hard';
// @endif
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "mkdirp": "^0.5.1",
     "mocha": "^2.5.3",
     "parallelshell": "^2.0.0",
-    "pixify": "^1.0.0",
+    "pixify": "^1.2.0",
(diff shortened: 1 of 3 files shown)


GoodBoyDigital:
👍

PR #2783, opened 2016-08-09
Floor'ed all values passed to canvas.drawImage() as using non-integers

peteward44:
on IE and PhantomJS can cause INDEX_ERR errors to be thrown.

Especially prevalent if you are using a graphics resolution of <1


GoodBoyDigital:
Thanks for this PR! Unfortunatly this solution forces the canvas renderer to always round pixels which in most use cases is undesired. However this does highlight the fact that roundPixels needs to be added to the canvas renderer. Setting it to true once implemented would solve your problem. Cheers 👍

Issue #2679, opened 2016-06-28
incorrect names/description in mesh.Plane


whizzkid:
* @param {int} segmentsX - The number ox x segments
 * @param {int} segmentsY - The number of y segments

these are actually the number of vertices.  not segments..

GoodBoyDigital:
Thanks for sharing, fancy putting a PR together? 👍

GoodBoyDigital:
👍

PR #2784, opened 2016-08-10
Dev 2679 rename params in plane

whizzkid:
params, and vars renamed to reflect their true content/function


whizzkid:
for #2679

GoodBoyDigital:
slick 👍

PR #2788, opened 2016-08-11
fix call to getBounds

staff0rd:
A possible response to [this](https://github.com/pixijs/pixi.js/commit/2565f7a555b895b79c8dc77c723d946fa83879ca#commitcomment-18600748).


GoodBoyDigital:
Nice catch @staff0rd  👍

PR #2789, opened 2016-08-11
initialize _localID

staff0rd:
@ivanpopelyshev I think this is missing because the conditionals [here](https://github.com/staff0rd/pixi.js/blob/9aa8533338255c0cb9e038364045d0d763b8a778/src/core/display/TransformStatic.js#L78) and [here](https://github.com/staff0rd/pixi.js/blob/9aa8533338255c0cb9e038364045d0d763b8a778/src/core/display/TransformStatic.js#L113) eventually start comparing NaN because `_localID` is undefined.

### DIFF
--- a/src/core/display/TransformStatic.js
+++ b/src/core/display/TransformStatic.js
@@ -47,7 +47,8 @@ function TransformStatic()
     this._sy  = Math.sin(0);//skewY);
     this._nsx = Math.sin(0);//skewX);
     this._cx  = Math.cos(0);//skewX);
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
sweet! Another great catch, thanks @staff0rd 👍

ivanpopelyshev:
I think I wanted to move it in TransformBase, but then something distracted me.

GoodBoyDigital:
good shout!

PR #2790, opened 2016-08-11
fix 'different rectangle if getting local after global bounds' test

staff0rd:
Result of `getLocalBounds()` was not being `expect()`ed.


GoodBoyDigital:
I think I may have named the test wrong here! This is actually intended. The issue was if you called getBounds and then getLocalBounds,  the getBounds result would change to local bounds as they were written to the same rectangle.

staff0rd:
Makes sense!

PR #2798, opened 2016-08-12
added enabled to filters

### LINKED ISSUE #2797: Add an enabled property to filters
Some filters can be skipped under certain conditions. Eg a blur with 0 strength. Lets add that!

GoodBoyDigital:
fixes #2797

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -433,43 +433,7 @@ Container.prototype.renderWebGL = function (renderer)
     // do a quick check to see if this element has a mask or a filter.
     if (this._mask || this._filters)
     {
-        renderer.currentRenderer.flush();
-
-        // push filter first as we need to ensure the stencil buffer is correct for any masking
-        if (this._filters && this._filters.length)
-        {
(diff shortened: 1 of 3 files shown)


src/core/display/Container.js
```diff
@@ -428,59 +428,85 @@ Container.prototype.renderWebGL = function (renderer)
...
-        renderer.currentRenderer.flush();
+        this._enabledFilters.length = 0;
 
-        if (this._mask)
+        for (i = 0; i < filters.length; i++)
         {
-            renderer.maskManager.popMask(this, this._mask);
+            if(filters[i].enabled)
+            {
+                this._enabledFilters.push( filters[i] );
```
bigtimebuddy: Why not use Array.filter?

 this._enableFilters = filters.filter(function(filter){ return filter.enabled;});

GoodBoyDigital: Good question! Even though it's cleaner to look at , it generates a new
array each time which is something I'm keen to avoid in the core render
loop..

bigtimebuddy: Makes sense

PR #2802, opened 2016-08-14
fastRect problem, again

ivanpopelyshev:
Compare results for rounded rect the fast one:

```js
var renderer = PIXI.autoDetectRenderer(800, 600, { antialias: true });

var clr = [0xFF700B, 0x0000FF];
var step = 0;

stage.interactive = true;

stage.addChild(graphics);

// run the render loop
animate();

function animate() {
    graphics.clear();
    graphics.beginFill(clr[Math.floor(step/5)], 1);
    //graphics.drawRoundedRect(50, 250, 120, 120, 5);
    graphics.drawRect(50, 250, 120, 120, 5);
    graphics.endFill();
    step = (step+1)%10;    

    renderer.render(stage);
    requestAnimationFrame( animate );
}

```

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -735,10 +735,9 @@ Graphics.prototype._renderSpriteRect = function (renderer)
         }
 
         this._spriteRect = new Sprite(Graphics._SPRITE_TEXTURE);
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
👍

Issue #2806, opened 2016-08-15
Wrong getBounds() on trimmed Sprites inside a Container 


staff0rd:
Two sprites are in [this atlas](https://rawgit.com/staff0rd/pixi-tilesprite/c9427eab8d104bed1a06d543bb5798704dd01666/spritesheet-2.json).  The first, 50x50, the second, 100x50 including 25px transparency right and left.

Placing these sprites inside containers results in correct render layout, however inspecting the width of the containers **incorrectly reports the width of the second sprite as 50px rather than 100px**.

![image](https://cloud.githubusercontent.com/assets/6289998/17658372/a7a4aea6-630e-11e6-9161-883cc50ffe7b.png)

https://jsfiddle.net/staff0rd/5h2ybm92/

@ivanpopelyshev, this is possibly from your bounds update, as I didn't see this issue prior, however it doesn't look like [this code](https://github.com/pixijs/pixi.js/blob/0abd932991470dcc8348ef895a58537dbd6f853b/src/core/sprites/Sprite.js#L232-L241) has changed.  I'm not certain whether it's that code, or somewhere else, that causes this issue.

ivanpopelyshev:
That's strange.

Bounds are calculated in next method: https://github.com/pixijs/pixi.js/blob/0abd932991470dcc8348ef895a58537dbd6f853b/src/core/sprites/Sprite.js#L271

ivanpopelyshev:
Yes, it is a bug. We need to call "calculateBoundsVertices" , and then pass vertexData[8,9,..] to calcBounds. 

@GoodBoyDigital that's the part of my PR that you changed. Please look at it and decide what to do with it.

whizzkid:
as a sidenote, I see in the comment there that we need to be compatible with pixiv3..... why?   this is v4 :)

ivanpopelyshev:
Exactly because people are complaining that width must be 100px and not 50px :)

ivanpopelyshev:
That's build: https://dl.dropboxusercontent.com/u/46636530/pixijs/bounds666/pixi.js

GoodBoyDigital:
Gonna throw this out there... @whizzkid makes an interesting point, should bounds care about trim forbounds? v4 could be different...

ivanpopelyshev:
@GoodBoyDigital people are using it. There were other issues about that.

staff0rd:
[...]

PR #2808, opened 2016-08-15
fix boundsData problem

ivanpopelyshev:
Fix for #2806


ivanpopelyshev:
Do not merge yet, it does not pass tests

ivanpopelyshev:
May be remove that "if"? Its too big, I dont know if it actually give us speed there.

Issue #2800, opened 2016-08-13
Cannot read property 'transform' of null on MovieClip


dan335:
I'm using the latest 4.  If I call obj.position.y on a MovieClip I get.

```
Cannot read property 'position' of null
```

If I call getBounds I get 

```
Cannot read property 'updateTransform' of null
```

If I do console.log on the MoveClip then transform is null. Is this a bug or am I doing something wrong?

GoodBoyDigital:
Thanks for sharing! If you could provide a fiddle for us to debug that would be ace 👍

dan335:
I tried making one but it has a transform.  https://jsfiddle.net/0guok9de/  Not sure why mine doesn't.

GoodBoyDigital:
Random Guess, are you using the 'new' keyword?

dan335:
Here's how I'm using it.

```
export class MapFlagBuilding extends MyMovieClip {
    constructor() {}
}

export class MapFlagCity extends MapFlagBuilding {
    constructor() {}
}

this.flag = new MapFlagCity();
```

Would the new keyword break the transform?

GoodBoyDigital:
Its more of a case that not using it usually creates the issues

dan335:
Ah, I'll check but I think I'm using it.

dan335:
I'm setting the position when it's created and it works.  Somehow it gets set to null though.  I'll keep digging, thanks.

dan335:
This is when it happens.  When the flag color changes I call this function.  Before calling it it has a transform.  Afterwards it's null.

```
drawFlag() {
    if (this.flag) {
      this.flag.destroy();
    }

    this.flag = new MapFlagCity(this.canvas, this._id, this.coord, this);
}
```

dan335:
I got it working.  I needed to call removeChild on the parent container before calling destroy.  Not sure if it's a bug or not?

ivanpopelyshev:
Not a bug, just badly documented API ;) Yes, you should call removechild on parent before destroy.

GoodBoyDigital:
Good shout, one of you chaps fancy making a PR to the docs to save others
from going through the same issue?

dan335:
Should it be on MovieClip.destroy, Sprite.destroy, Container.destroy or DisplayObject.destroy or all of the above?

GoodBoyDigital:
All of the above would work well - as a thought,
As an alternative solution - do you guys think we should remove the object
automatically when destroy is called?

[...]

PR #2809, opened 2016-08-15
Destroying a DisplayObject will remove itself

bigtimebuddy:
Child removes itself upon being destroyed. 
Responds to issue #2800

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -574,13 +574,17 @@ Container.prototype.destroy = function (options)
 
     var destroyChildren = typeof options === 'boolean' ? options : options && options.children;
 
+    var oldChildren = this.children;
+    this.children = null;
+
     if (destroyChildren)
     {
-        for (var i = this.children.length - 1; i >= 0; i--)
+        for (var i = oldChildren.length - 1; i >= 0; i--)
         {
-            this.children[i].destroy(options);
+            var child = oldChildren[i];
+            child.parent = null;
+            child.emit('removed', this);
+            child.destroy(options);
         }
     }
-
-    this.children = null;
 };
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -563,7 +563,8 @@ Container.prototype.renderCanvas = function (renderer)
 };
 
 /**
- * Destroys the container
+ * Removes all internal references and listeners as well as remove current children from the display list. 
+ * Do not use a Container after calling `destroy`.
  * @param [options] {object|boolean} Options parameter. A boolean will act as if all options have been set to that value
  * @param [options.children=false] {boolean} if set to true, all the children will have their destroy
  *      method called as well. 'options' will be passed on to those calls.
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -583,7 +583,6 @@ Container.prototype.destroy = function (options)
         {
             var child = oldChildren[i];
             child.parent = null;
-            child.emit('removed', this);
             child.destroy(options);
         }
     }
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -572,16 +572,15 @@ Container.prototype.destroy = function (options)
 {
     DisplayObject.prototype.destroy.call(this);
 
-    this.removeChildren();
-
     var destroyChildren = typeof options === 'boolean' ? options : options && options.children;
     
     if (destroyChildren)
     {
-        for (var i = 0, j = this.children.length; i < j; ++i)
+        for (var i = this.children.length - 1; i >= 0; i--)
         {
             this.children[i].destroy(options);
         }
     }
+
     this.children = null;
 };
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -572,7 +572,10 @@ Container.prototype.destroy = function (options)
 {
     DisplayObject.prototype.destroy.call(this);
 
+    this.removeChildren();
+
     var destroyChildren = typeof options === 'boolean' ? options : options && options.children;
+    
     if (destroyChildren)
     {
         for (var i = 0, j = this.children.length; i < j; ++i)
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
NONONONONONO WHYYY!

Do you realize what will happen in container.destroy(true) ?

GoodBoyDigital:
interesting conundrum that!

src/core/display/Container.js
```diff
@@ -572,16 +572,16 @@ Container.prototype.destroy = function (options)
 {
     DisplayObject.prototype.destroy.call(this);
 
+    this.removeChildren();
```
GoodBoyDigital: unless i'm mistaken - would this line make ```this.children``` empty?

bigtimebuddy:
DOH! You're right. Destroying the children in reverse order should be fine.

ivanpopelyshev:
And now, gentlemans, we have destroy() method that works O(N^2). Nice :)

bigtimebuddy:
If someone wants to implement another approach, I'm cool with that. I think this functionality is important though and would like to see it in v4.

ivanpopelyshev:
I think its better to leave it as it is for now. There are also things like "remove" event in child, and at the same time we dont have "parentDetached" or anything like that... 

The thing is, any small change may break some code, its better if we make big PR that deals with that.

bigtimebuddy:
@ivanpopelyshev I disagree that we should leave it. There's never a case where you'd want to destroy and not remove that child from it's display list. Expecting developers to remember that is a poor design choice.

ivanpopelyshev:
Then destroying whole subtree must be separated to different method, like we have with renderWebGL and _renderWebGL. We still don't have a convention which events must be fired in case of whole subtree is removed, or whole subtree is destroyed.

ivanpopelyshev:
Hm. Yes, good idea: we need to separate destroying this element from destroying whole subtree, also remove all children before we ask their destruction, something like that.

GoodBoyDigital:
Interesting debate going on here!

 I think @bigtimebuddy makes a good point about removing an object from its parent if it has been destroyed. If its going to be destroyed we know it will throw errors if its not removed. @ivanpopelyshev - what kind of use cases are you thinking about where this is undesirable behaviour?

Thanks!

ivanpopelyshev:
Mass destruction. We have to separate methods for killing only this element and killing whole subtree.

GoodBoyDigital:
I see what ya mean, thats a different issue than removing children though?

ivanpopelyshev:
No, it's not. You cant just add "removeChild" in destroy() because of performance reasons in mass destroy.  Its either big hack, either architecture improvement.

GoodBoyDigital:
is removeChild really expensive?

bigtimebuddy:
Maybe Container should removeChildren before it tries to destroy them. Caching the children list before removing wouldn't create so many removeChild calls within DisplayObject.

ivanpopelyshev:
It has both indexOf and removeAt inside. That means its O(N). If we do N removeChilds, its O(N^2). Try 100k bunnies and your CPU will be fried.

GoodBoyDigital:
Thanks for sharing the details @ivanpopelyshev , I guess to solve this issue do we even need to do a recursive remove child operation? We only need to remove the object being destroyed from its parent (rather than all its children)?

ivanpopelyshev:
@GoodBoyDigital 

OK, may be that will work:

```js
//this like bigtimebuddy's
DisplayObject.prototype.destroy = function ()
{
    if (this.parent)
    {
        this.parent.removeChild(this);
    }
    this.transform = null;

    this.parent = null;

    this._bounds = null;
    this._currentBounds = null;
    this._mask = null;

    this.filterArea = null;
};

//this is mine
Container.prototype.destroy = function (options)
{

    var oldChildren = this.children;
    this.children = null;
   
    if (destroyChildren)
    {
        for (var i = oldChildren.length - 1; i >= 0; i--)
        {
            var child = oldChildren[i];
            child.parent = null;
            child.emit('removed', this);
            child.destroy(options);
        }
    }
};
```

bigtimebuddy:
Updated with @ivanpopelyshev Container destroy.

ivanpopelyshev:
That's a benchmark, but i dont know how to write this. Basically, if it takes more than 1s on any machine, then its wrong. Problem is, its synchronous code:

```js
describe('destroy', function () {
        it('should destroy 100k rabbits without a problem', function () {
            var texture = PIXI.RenderTexture.create(10, 10);
            var topContainer = new PIXI.Container();
            var bottomContainer = new PIXI.Container();

            topContainer.addChild(bottomContainer);
            for (var i=0;i<100000;i++) {
                bottomContainer.addChild(sprite);
            }

            topContainer.destroy({children: true, texture:true});
        });

    });
```

GoodBoyDigital:
One thing - was wandering if we need -> ```child.emit('removed', this);``` ?
Let me know if i'm completely off the mark!

ivanpopelyshev:
I really don't know, its up to you :) But we just had someone ask if "removed" will be called when whole subtree is destroyed, somewhere in the issues, please find him.

GoodBoyDigital:
What do you think @bigtimebuddy ?

bigtimebuddy:
@GoodBoyDigital I think I'm not a big fan of emitting here. When you destroy something, it means that you have no desire to use that thing anymore. If you're still responding to callback, then you clearly should not be destroying it. Instead, I actually would consider `this.removeAllListeners()` in DisplayObject's `destroy`.

GoodBoyDigital:
I like that! Intention is definitely to nuke the object and it would help ensure that it completely free's it from getting caught in events memory!

ivanpopelyshev:
Maybe add special removeChildrenRecursive then?

bigtimebuddy:
Okay, now removes listeners and doesn't emit remove on destroy.

ivanpopelyshev:
Can we write it down somewhere in docs? We really need explanation of how removeChild/destroy works

bigtimebuddy:
Added more documentation for destroy

GoodBoyDigital:
nice work chaps!

PR #2810, opened 2016-08-15
Adds outputName to pixify call

### LINKED ISSUE #2804: nw.js and browserify problem
Something wrong happened in pull request #2769 , commit 4d03034cadb4185c92ce2cab4537a48f62768b7d.

nw.js, both new version and the one that is used in rpgmaker MV are failing. I even removed "loader" module so right now the only pixi dependency is eventemitter.

What i see in the console is "PIXI is not defined". Sometimes it crashes, sometimes i can see that "global.PIXI" is set while "window.PIXI" lacks.

How to reproduce: download latest nw.js binaries, put these files in the same directory, launch it.

"pixi.js" is good version, just before 2769 was merged. 
"pixi_fail.js" is one of latest versions, stripped of PIXI.loader so we can pinpoint the problem better.

[nwjs.zip](https://github.com/pixijs/pixi.js/files/417293/nwjs.zip)

By the way, if someone needs good version that works with nw.js - just take it from the archive.

bigtimebuddy:
Fixes #2804. 
Standalone name with Browserify (`PIXI`) needs to be different from the output name (`pixi`). Also, changed watchify to be `src` instead of `src/index.js` (same difference, less verbose).


bigtimebuddy:
@ivanpopelyshev Can you verify this fix? Please make sure you update your **pixify** dependency.

bigtimebuddy:
Ivan confirmed with me on Slack, this change resolves his issue.

GoodBoyDigital:
nice one team! 👍

PR #2813, opened 2016-08-15
Integrates pixify watch command

bigtimebuddy:
Simplifies and optimizes the watch command for building (e.g., `npm start`)
* Removes **watchify** dependency
* pixify watch only builds in debug (not release) mode
* Watch now externalizes the source-maps, adds headers and apply preprocess


ivanpopelyshev:
nice! watchify was taking browserify with it , and now only pixify depends on it.

GoodBoyDigital:
👍

PR #2817, opened 2016-08-16
sprite bounds tweaks

GoodBoyDigital:
A variation of @ivanpopelyshev PR
This PR favours non trimmed Textures (As most will not be trimmed)

- It basically checked for trim, if its not then business as usual

- Renamed a few functions to make it easier to understand that we are calculating trimmed bounds

- Slight refactor that hopefully should make it easier to follow

- vertexTrimmedData only created when required


ivanpopelyshev:
I like it :+1:

PR #2826, opened 2016-08-17
cleaning up WebGLState

ivanpopelyshev:
I think we dont use push/pop, but its not good that they had that bug. Also, blendMode can be 2, so resetToDefault had a problems too


ivanpopelyshev:
That means WebGLRenderer.reset() was a bit wrong.

GoodBoyDigital:
nice!

PR #2828, opened 2016-08-17
InteractionManager: Add target and currentTarget to eventData

endel:
I've extracted `eventData` into a new class called `InteractionEvent`. 

- what previously was  `event.target` is now `event.currentTarget` (object currently calling the callback handler)
- `event.target` now have the reference to the object that **triggered** the event

The idea here is to mimic the [DOM Event](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) spec.

Is [InteractionData.target](https://github.com/pixijs/pixi.js/blob/0446c7829e9727439349d5dbbdd85f57dd54d5ea/src/interaction/InteractionData.js#L18-L23) being used right now? I guess somebody removed the usage of this at some point and it doesn't work anymore.

Cheers!

### DIFF
--- a/src/interaction/InteractionEvent.js
+++ b/src/interaction/InteractionEvent.js
@@ -1,0 +1,68 @@
+/**
+ * Event class that mimics native DOM events.
+ *
+ * @class
+ * @memberof PIXI.interaction
+ */
+class InteractionEvent
+{
+
+    constructor()
+    {
+        /**
+         * Which this event will continue propagating in the tree
+         *
+         * @member {boolean}
+         */
+        this.stopped = false;
+
+        /**
+         * The object to which event is dispatched.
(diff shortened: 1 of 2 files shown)


endel:
Just fixed the `target` to use the deepest hit element found. It wasn't working properly on the very first implementation. Now it's fine! :)

GoodBoyDigital:
Thanks for this! We will review and get back to ya. 👍

englercj:
Can you rebase onto the latest `dev` please so I can review/merge this?

endel:
Hey @englercj,

Thanks. The diffs went crazy after this major code reorganization. I made the changes again and rewrote the history back into my branch. 

Please let me know if you can test it. Cheers!

PR #2830, opened 2016-08-18
Fix for Text.width setter not updating if it's dirty before applying …

rfenner:
Fix for Text not updating it's width when setting if teh Text is dirty. Cause the scaled value not be what was expected.


GoodBoyDigital:
Thanks for this @rfenner! - we just merged in v4 so this is now sorted on master 👍

PR #2839, opened 2016-08-19
Fix browserify-versionify dependency

BrianRosamilia:
As is the case in the current Pixi 

https://github.com/pixijs/pixi.js/blob/master/package.json#L35

browserify-versionify is picked up when it's a dependency, not a dev dependency.  Tried against the latest version of browserify.


bigtimebuddy:
Good catch @BrianRosamilia 👍

PR #2845, opened 2016-08-23
Tweak to realWidth / realHeight in base render texture

themoonrat:
Making sure that the resolution used for realWidth & realHeight will work if nothing has been passed in as a constructor parameter


GoodBoyDigital:
👍

PR #2847, opened 2016-08-23
Fixes #2840

### LINKED ISSUE #2840: Hunting a Memory Leak: Are PIXI Entities regarded as "Nodes" in the Webkit Timeline Inspector?
Hi

I've been hunting a memory leak in my game for 2 days, both Electron and Chrome memory usage grows at the same rate, while Javascript Heap is < 100MB (30MB's now), the "Nodes" count increases, it can't be GC collected, the memory usage slowly increases, I've seen it reach 1.3GB's in <2 hours (so when memory usage of "Electron" or "Chrome" is 1.3GB's, the Heap is insignificantly small)

I'm not even sure whether this "Nodes" increase accounts for the memory usage, it might be something else

I've inspected all my code and found 2 major possible leak causes, I wasn't calling .destroy() on 2X Sprite types, I was just removing them from their parents, I hope this was the cause of the issue, however it wasn't the major cause I was looking for, I think I just fixed the "actual memory leak" - while what I'm hunting is the "Nodes leak" - or something else - after this fix, the Javascript Heap is ~25MB's, however the "Nodes" growth and memory usage growth is still there

I've inspected all my code several times, currently I either have a regular DOM leak, or I might be hunting something WebGL related I can't observe

I'm currently running another experiment, yet each experiment takes 1-3 hours, I just wanted to get some input from the PIXI Community, is there any known PIXI-related causes I should look for?

(My PIXI Is V4, I built it 5 days ago from the source)

GoodBoyDigital:
Default destroy options for text are now all true.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -7,6 +7,7 @@ var Sprite = require('../sprites/Sprite'),
 
     var defaultDestroyOptions = {
             texture:true,
+            children:false,
             baseTexture:true
     };
 /**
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Looks good 👍

PR #2853, opened 2016-08-25
Added 1 extra check in the procesInteractive

whizzkid:
to check if children exists before attempting to call children.length

this would've thrown an error if you remove stuff in an onMouseClick for example.


whizzkid:
not in master, stupid git...

PR #2855, opened 2016-08-25
Fix process interactive

whizzkid:
Turns out processInteractive didnt need a fix if we Destroyed the DisplayObject correctly.
by setting interactive and interactiveChildren to false there,  it will no longer throw exceptions when an object gets destroyed in an onMouseEvent for example.


GoodBoyDigital:
nice!

PR #2864, opened 2016-08-25
Support for built, deployable documentation

bigtimebuddy:
## Overview

Supports building documentation to the S3 bucket. This can be useful for a couple of reasons:
- Always have a distributed up-to-date version of the documentation represented on a particular branch or release tag.
- Documentation is build and stored (archived) with each tag, so older versions of the docs are readily accessible. This makes the docs more future-forward, without having to so thoroughly capture old, deprecated APIs.

### Example 

Here's the documentation for this branch that was build by Travis:
http://s3-eu-west-1.amazonaws.com/pixi.js/cdn-docs/docs/index.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
     - npm run docs
 
 deploy:
-    provider: s3
+  - provider: s3
     access_key_id: $S3_ACCESS_KEY_ID
     secret_access_key: $S3_SECRET_ACCESS_KEY
     bucket: "pixi.js"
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
As an aside, one improvement we could make to the documentation theme (jaguarjs-jsdoc) is to add the version number _somewhere_ on the header and/or title. Would make it easier to know which version you're on.

bigtimebuddy:
@GoodBoyDigital any thoughts?

GoodBoyDigital:
I think this is a fantastic idea!

PR #2870, opened 2016-08-26
Fix CanvasRenderer resolution.

YayoiAyame:
CanvasRenderer was not able to change the resolution, from v4.
Removed "rootResolution" that seems to be no effect.


englercj:
Can you rebase this onto the latest `dev` branch please?

PR #2872, opened 2016-08-27
Update README.md

xeoneux:


bigtimebuddy:
Thanks for the PR. Could you make sure to make all PRs to dev and not master. Also, I think we got this fix already.

PR #2894, opened 2016-09-01
Fix the first released versions link

mortonfox:


englercj:
Looks like this was fixed elsewhere, thanks!

PR #2904, opened 2016-09-02
Fix for graphics fast rect with multiple renderers

RoryDouglas7:
This is due to a bug where graphics fast rects do not render when there are graphics on multiple renderers.

Changed _SPRITE_TEXTURE to array with the key based on renderer.CONTEXT_UID

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -177,7 +177,7 @@ function Graphics()
      */
 }
 
-Graphics._SPRITE_TEXTURE = null;
+Graphics._SPRITE_TEXTURES = [];
 
 // constructor
 Graphics.prototype = Object.create(Container.prototype);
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
what exactly is wrong in current implementation? If we cant use the same sprrite in multiple renderers, its a bug. do you have a case where latest version reenders something wrong? we just fixed a few of these isssues a week aor two ago

RoryDouglas7:
The following example is using the latest released pixi and is based on the [example here] (http://pixijs.github.io/examples/#/basics/graphics.js):

https://jsfiddle.net/kd6gL10q/3/
https://jsfiddle.net/kd6gL10q/2/

In the first example both are fast rectangles, and only one appears. In the second example the second graphic is not a fast rectangle as it has two shapes so it does appear.

I was also able to replicate this using the latest dev version of Pixi.

ivanpopelyshev:
So, we have to find why that bug happens first, then fix it. Do you have an explanation for that?

ivanpopelyshev:
OK, got it, your solution is right but its better to actually use bindRenderTarget() mechanism for that

RoryDouglas7:
Ok, I will be able to take a look in a few days. Right now I am facing a deadline so don't have the time to look properly.

ivanpopelyshev:
No problem, I'm just making alternative PR for that :) @GoodBoyDigital will decide the fate of the issue ;)

englercj:
Looks like #2963 was merged which should fix this issue.

PR #2920, opened 2016-09-06
Optimization: Less calls to gl.bindTexture() & gl.enable(gl.STENCIL_TEST)

ddude:
This is a resubmit of https://github.com/pixijs/pixi.js/pull/2919 to the proper branch this time :)


englercj:
Going to close this for the reasons we mentioned on the related gl-core thread. We may do something similar in the future, but probably differently in implementation.

PR #2936, opened 2016-09-09
Implementing classes using the new ES6 syntax

leonardo-silva:
Hi,

In my PhD research, I am using ``pixi.js`` in some experiments.

As a result, I identified the existence of several "classes" in the current codebase (Container, Shader, CanvasPrepare, WebGLManager, etc)

So, I decided to rewrite the classes to follow the new syntax provided by ECMAScript 2016, which I think is much more readable. All changes are just syntactical. I was also very careful to preserve the code style.
Finally, the tests are green in the migrated code.

Do you think it is worth to integrate these changes to the project?

Thank you

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -34,6 +34,7 @@
     "maxparams"     : 8,        // Prohibit having more than X number of params in a function.
     "maxdepth"      : 8,        // Prohibit nested blocks from going more than X levels deep.
     "maxlen"        : 220,      // Require that all lines are 100 characters or less.
+    "esversion"     : 6,        // ECMAScript 2015 syntax allowed
     "globals"       : {         // Register globals that are used in the code.
         "performance": false   // Depends on polyfill, simplifies usage
     },
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "browserify-versionify": "^1.0.6"
   },
   "devDependencies": {
+    "babel-preset-es2015": "^6.14.0",
     "babelify": "^7.3.0",
     "del": "^2.2.0",
     "electron-prebuilt": "^1.3.2",
--- a/package.json
+++ b/package.json
@@ -51,9 +51,11 @@
     "browserify-versionify": "^1.0.6"
   },
   "devDependencies": {
+    "babel-preset-es2015": "^6.14.0",
     "del": "^2.2.0",
     "electron-prebuilt": "^1.3.2",
     "floss": "^0.7.1",
+    "gulp-babel": "^6.1.2",
     "jaguarjs-jsdoc": "^1.0.1",
     "jsdoc": "^3.4.0",
     "jshint": "^2.9.2",
--- a/package.json
+++ b/package.json
@@ -51,11 +51,10 @@
     "browserify-versionify": "^1.0.6"
   },
   "devDependencies": {
-    "babel-preset-es2015": "^6.14.0",
+    "babelify": "^7.3.0",
     "del": "^2.2.0",
     "electron-prebuilt": "^1.3.2",
     "floss": "^0.7.1",
-    "gulp-babel": "^6.1.2",
     "jaguarjs-jsdoc": "^1.0.1",
     "jsdoc": "^3.4.0",
     "jshint": "^2.9.2",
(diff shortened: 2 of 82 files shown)


ivanpopelyshev:
Yeah, we just discussed it a day ago. Also we have to ask @bigtimebuddy to add babelify to pixify, so there wont be direct dependency there

GoodBoyDigital:
Holy monkeys!! Thanks so much for this! Out drinking beers at the moment..
because Friday.. but will check this out properly tomorrow :)

Thanks for taking the time for this pr!

leonardo-silva:
Hi @GoodBoyDigital, you are welcome! :)

bigtimebuddy:
Awesome work! It is really great timing because we were planning on doing this very soon anyways. You'll need to include the babel-preset-es2015 module in the package.json devDependencies (Travis is failing). 

There are a bunch of things we would need to be sure of before we could accept this, namely, performance, documentation.

I think it makes more sense for us to make a new Dev branch and start working on this conversion there (starting by merging this PR).  I'd like to make additional passes on this for const/let usage, fat arrows instead of binds, statics and other ES6 features. 

@GoodBoyDigital what do you think about that approach?

bigtimebuddy:
Also @ivanpopelyshev this approach is correct. Adding babelify to the package.json is sufficient for transpiling. No additional pixify changes required. Eventually we might move bablel to pixify, but for now it'll keep pixify slim.

ivanpopelyshev:
@bigtimebuddy ack.

OK, so right now we have about 15MB for pixify, 34MB of dependencies of example plugin in total.

PIXI setup has more than extra 100MB for electon and tests

How much do we add with babelify? Imagine that we have a simple plugin there, we run "npm install" and get like 34MB of things. How much will babel cost us?

bigtimebuddy:
Unfortunately, heavier dev dependencies are the cost right now for creating more maintainable code that's transpiled. Babel is pretty big and other tech like TypeScript, coffeescript, Haxe, etc have tradeoffs too.

ivanpopelyshev:
Yeah, and that's for every plugin. That's why i want it to be separate dependency - we can "npm link" pixify and stuff, and it wont get more disk space

photonstorm:
It's a slight shame the overall size of 'installing' the project will increase, but honestly, who cares given the benefits of this? Drive space can't be used as a reason to not do this. 35MB is nothing for anyone who does any serious web development these days, let alone game development. That total alone is a fraction of the size of just one of our game PSDs!

The only real considerations should be the built lib size, and performance. And I guess the overhead of maintaining this, because it's not like the standards and Babel aren't changing on a daily basis :)

ivanpopelyshev:
@photonstorm 35MB vs 120MB. That's without electron of course. Stuff like electron and babel must be downloaded once per project, not per plugin.

There are also troubles for IDE, I have to make sure only one instance of pixi node_modules exists in build paths, otherwise indexing is a pain.

Just imagine that "pixi.js", "pixi-spine", "pixi-filters" are modules of one big project.

photonstorm:
You've obviously never had to work on an Angular project! I would _kill_ for a project size as small as 120MB there. It used to be over twice that. Just for 'hello world'.

npm 3 helped a lot though, getting rid of dupes and flattening the modules folder out. And when Babel moved to using [Lerna](https://github.com/lerna/lerna) it helped even more (and I'd strongly recommend it for Pixi).

It might suck that this is the state of web development today, but it's so common and expected it also makes it a non-issue. It also only impacts an (unknown) percentage of Pixi devs. Lots just grab the build files and run off, not caring about the packages. This change will only impact those left.

englercj:
> ust imagine that "pixi.js", "pixi-spine", "pixi-filters" are modules of one big project. And there are "pixi-extra-filters", "pixi-tilemap", "pixi-picture", "pixi-display", there will be updated version of "pixi-lights" soon. Every one of them will need this thing.

This isn't true, npm dedupes. Only one version is installed as long as they are compatible.

> 35MB vs 120MB. That's without electron of course. Stuff like electron and babel must be downloaded once per project, not per plugin.

Most of my day-job-related projects have 3GB+ of dependencies, meh.

bigtimebuddy:
@leonardo-silva Could you please close and make this PR against the newly-created branch, **dev-es6**? We would like to do some longer-term development migrating to ES6 that involves more work beyond this single PR. Thanks for kicking this off!

leonardo-silva:
Hi @bigtimebuddy, I edited this PR and changed the base branch to dev-es6. Is it ok ?

Issue #2941, opened 2016-09-11
Gradient PIXI.Text doesn't update in Safari (iOS,OSX) and Chrome (iOS)


hendrysadrak:
Check out this sample code: http://codepen.io/hendrysadrak/full/yaYxxg/
Works fine in Chrome (OSX)
In Safari (iOS,OSX) the text updates only when texts width is different from which it was before (example: "short text" -> "long text") 

My workaround for this right now is setting the text to something much longer before setting text to right value:
```
var text = new PIXI.Text('10', {
  fill: [0xf1c40f,0x2ecc71]
});
// text.text === '10'

// ugly fix
text.text = 'aaaaaaaaaaaaaaaaaaaaaa';
text.updateTransform();
text.updateText();
text.updateTexture();

// update text
text.text = '11';
```

Would help fixing this but don't know where to start.

hendrysadrak:
Adding gifs of the problem:


Expected result: (as it is in Chrome)
![sep-12-2016 12-07-35](https://cloud.githubusercontent.com/assets/4000963/18430543/ad58c6fa-78e1-11e6-9382-189cf2af272d.gif)
Safari:
![sep-12-2016 12-07-29](https://cloud.githubusercontent.com/assets/4000963/18430544/ad597460-78e1-11e6-9342-a5ac71c27c47.gif)

ivanpopelyshev:
> Would help fixing this but don't know where to start.

https://github.com/pixijs/pixi.js/blob/master/src/core/text/Text.js

If you want to fix it, write something like 

```js
PIXI.Text.prototype.updateText = function() {
//your hax code
}
```

I dont know whats wrong there yet :(

hendrysadrak:
Thx! I will go through the code and check if I can find smth.

ivanpopelyshev:
May be some tweaks must be done with "updateTexture" and not "updateText"

Also, which renderer is used in safari, Canvas or WebGL?

hendrysadrak:
@ivanpopelyshev Renderer type is WebGL

hendrysadrak:
@ivanpopelyshev  
I tried by removing `if` [here](https://github.com/pixijs/pixi.js/blob/master/src/core/text/Text.js#L273-L277) 

Leaving only `this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);` and it fixed it. 

Should i make a PR for this? Is that a OK fix?

ivanpopelyshev:
[...]

PR #2945, opened 2016-09-12
Issue #2941

hendrysadrak:
#2941 Gradient PIXI.Text doesn't update in Safari (iOS,OSX) and Chrome (iOS)

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -273,7 +273,10 @@ Text.prototype.updateText = function (respectDirty)
     if (navigator.isCocoonJS)
     {
         this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);
-
+    }
+    else
+    { // some browsers, like safari requires any update of any kind here to allow text to update properly
+        this.context.clearRect( 0, 0, 1, 1 );
     }
 
 //    this.context.fillStyle="#FF0000";


src/core/text/Text.js
```diff
@@ -273,7 +273,10 @@ Text.prototype.updateText = function (respectDirty)
     if (navigator.isCocoonJS)
```
englercj: Wonder if we should just remove this check and always clear?

themoonrat: It was an option as discussed https://github.com/pixijs/pixi.js/issues/2941 - My fault for suggesting the less pretty, but potentially faster option :)
I wish jsperf.com was still around to quickly benchmark a clearRect of 1x1 vs a clearRect of say 1920x1080!

englercj: Ha, yeah true. We are basically banking on the width/height assignments clearing the canvas without this clear here. But, like you found here that doesn't always work. I wonder if it might be a case of preferring "correctness" over bleeding perf. I don't particularly feel strongly either way.

themoonrat:
I did a basic benchmark test, just doing a loop over calling clearrect, and even going upto a million clearRects, can not see a difference, on a canvas of 1920x1080, between a clearRect of 1x1 and a clearRect of the full 1920x1080. Tried on Firefox, IE and Chrome on the desktop, Cocoon on an android device with Canvas+.

So I've changed my mind and am now leaning towards going the cleaner route and clear the entire canvas all of the time at that point ie. lose the if statement of only doing it for cocoon.

src/core/text/Text.js
themoonrat: I did a basic benchmark test, just doing a loop over calling clearrect, and even going upto a million clearRects, can not see a difference, on a canvas of 1920x1080, between a clearRect of 1x1 and a clearRect of the full 1920x1080. Tried on Firefox, IE and Chrome on the desktop, Cocoon on an android device with Canvas+.

So I've changed my mind and am now leaning towards going the cleaner route and clear the entire canvas all of the time at that point ie. lose the if statement of only doing it for cocoon.

themoonrat:
Last thing @hendrysadrak  - could you submit this PR to the dev branch, and not to the master please. Then we're good I think, thanks!

themoonrat:
The branch change looks good to me :)

themoonrat:
@englercj - one thing this did was show that the readme and license have been changed in master but not dev!

englercj:
Yeah I'll merge down this weekend :P

PR #2948, opened 2016-09-13
Allows a callback to be set on when the frame changes on a MovieClip

themoonrat:
https://github.com/pixijs/pixi.js/issues/2947

Adds a callback when the frame changes
Added some protection to the code to make sure that the texture is only changed when the frame changes. For example, if a playing animation is currently on frame 7, then gotoAndStop(7) is called, the onFrameChange function will not be called.


whizzkid:
Might I suggest adding the currentframe to the callback so that you don't have to do get a reference to your movieclip and get its currentframe yourself all the time.

so:

`this.onFrameChange(this.currentFrame);`

englercj:
Rebase required!

themoonrat:
Done

PR #2959, opened 2016-09-16
CanvasPrepare updated to simulate WebGLPrepare

themoonrat:
When rendering canvas games with large textures, there can be micro-stutters when a texture is drawn for the first time. Looking at Chrome dev tools I found that sometimes this is due to the texture being decoded on the GPU.

With PIXI, this is covered in WebGLPrepare, which can use the WebGL texture manager to update base textures and graphics. 
CanvasPrepare was just a stub. I have filled out this stub to try and make the behaviour match. The act of calling drawImage onto a canvas requires that the GPU has the image drawn; and thus is an indirect way to force textures to be uploaded and decoded.

Whereas WebGLPrepare works for both base textures and graphics, this PR for CanvasPrepare is just for base textures


englercj:
Rebase on `dev` required D: Sorry!

themoonrat:
And also done :)

PR #2961, opened 2016-09-17
Linted using ESLint and other cleanups

jayands:
I used [Overcommit](https://github.com/brigade/overcommit), ESLint and Babel to help find some interesting "errors" in your source code consistency, albeit mostly small stuff, like quote consistency and mixed white space (man, THAT one was "fun"; there were tabs in the source code!). I did, however, manage to find one actual error, in `test/core/Bounds.js`, in the process: it was potentially giving a false positive with a `no-unused-vars` error showing up during linting (c.f. that file).

Other than that, it was just cleanup and enforcing the linting rules. I know that there are a lot of files changed, but I would still ask that you go over all of them; sometimes I removed an unused variable, and other times, I told it to ignore it. Overcommit wouldn't let me just halfway do it, after all. I didn't include any files that weren't already in the system, but I will attach my extras with the PR if you want to add them in. The ESLint files were created with [`jshint2eslint`](https://github.com/jrajav/jshint2eslint), too, so they should be pretty consistent. 
[extras.zip](https://github.com/pixijs/pixi.js/files/477989/extras.zip)

Signed-off-by: Jonathan Sifuentes <jayands2k11@ymail.com>


bigtimebuddy:
Thanks for this! Unfortunately, we are already in the process converting to ESLint in our **dev-es6** branch (as well as converting to ES6) and have fixed many of these issues you pointed out. If you want to take a pass at that source to make sure we didn't miss anything, that'd be useful!

Issue #2934, opened 2016-09-09
anchor = new Point breaks the render


mathieuanthoine:
The two syntaxes work:
mySprite.scale.set(0.3,0.3);
mySprite.scale= new PIXI.Point(0.3,0.3);

but not with anchor:
mySprite.anchor.set(0.5,0.5); // ok
mySprite.anchor= new PIXI.Point (0.5,0.5); // breaks the render

GoodBoyDigital:
You are correct 👍 
This is on our to fix list :)

englercj:
Should be fixed by #3017

PR #2966, opened 2016-09-17
Add getter and setter to Sprite.anchor

tcreeds:
anchor is now a property for _anchor. Setter ensures that _anchor is set
to a valid ObservablePoint, and sets it with Point values if a Point is
passed in.

References #2934


ivanpopelyshev:
I see potential side-effects. It must be done the same way "position", "scale" and "skew". Also we have to make transform read-only. I'll make new PR for that.

englercj:
Closing this for the reasons @ivanpopelyshev mentioned. Did you ever make that PR ivan?

PR #2968, opened 2016-09-17
Converts src to ES6

bigtimebuddy:
### Overview

Modernizes the source code by converting to ES6 and using Babel es2015 (loose mode) to transpile to ES5. Also replaces jshint with eslint.


englercj:
Merging to master?

bigtimebuddy:
Oops, fixed! I also need to merge in changes from Dev.

englercj:
Just rebase onto lastest dev

bigtimebuddy:
@drkibitz Not sure I follow. Can you explain?

drkibitz:
The comment was made before the `_tick` method was actually meant to be bound, the commit to change to an ES6 class was what made it make no sense, this change just fixes that commit's functionality. The comment is basically saying, this is an internal closure because Function.bind is still slow, but now it's bound using Function.bind. It doesn't make sense. Either remove the comment or don't use Function.bind. Keep in mind, yes, it is almost 2 years, later and Function.bind is still slow.

bigtimebuddy:
@drkibitz Cool, I think I address your concern. Is that okay with you?

src/core/ticker/Ticker.js
```diff
@@ -107,34 +107,31 @@ class Ticker
...
+         * Internal tick method bound to ticker instance.
+         * This is because in early 2015, Function.bind
+         * is still 60% slower in high performance scenarios.
+         * Also separating frame requests from update method
+         * so listeners may be called at any time and with
+         * any animation API, just invoke ticker.update(time).
+         *
+         * @private
+         */
+        this._tick = (time) => {
```
drkibitz: This `() => {}` is still the same as Function.bind

alvinsight: Yes, arrow functions (`() => {}`)retain the scope they were declared in, and prevent the `this` headache that people used to get and forced us to use `Function.bind` all over the place.

bigtimebuddy: Babel does not transpile fat arrow functions into binds, so this is, in fact not the same as Function.bind. Here's how it get's [transpiled](http://s3-eu-west-1.amazonaws.com/pixi.js/dev-es6/pixi.js), which is the same as currently on dev.

```js
var _this = this;
this._tick = function (time) {

    _this._requestId = null;

    if (_this.started) {
        // Invoke listeners now
        _this.update(time);
        // Listener side effects may have modified ticker state.
        if (_this.started && _this._requestId === null && _this._emitter.listeners(TICK, true)) {
            _this._requestId = requestAnimationFrame(_this._tick);
        }
    }
};
```

drkibitz:
@bigtimebuddy in that case :+1:

PR #2978, opened 2016-09-20
Jsdoc fixes - marking relevant functions as @private

themoonrat:
As a side note, the property ._frame on Texture, and the property _activeRenderTarget on WebGLRenderer both have the leading underscore as if they are to be private, but are publicly accessed all over the shop.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -57,6 +57,8 @@ function Sprite(texture)
      */
     this._height = 0;
 
+    this._tint = null;
+    this._tintRGB = null;
     /**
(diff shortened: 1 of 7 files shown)


src/core/sprites/Sprite.js
```diff
@@ -63,9 +63,9 @@ function Sprite(texture)
      * @member {number}
      * @default 0xFFFFFF
      */
+    this.tint = 0xFFFFFF;
     this._tint = null;
     this._tintRGB = null;
-    this.tint = 0xFFFFFF;
```
englercj: This will cause a bug, the private versions need to be set *before* the setter runs.

themoonrat: Fixed

PR #2981, opened 2016-09-21
webpack temp fix

GoodBoyDigital:
This will mean things will work with web pack - we can address properly in a future release


englercj:
😭 <-- MFW built files as main

GoodBoyDigital:
@englercj haha! It's only temporary I promise :)

PR #2983, opened 2016-09-21
Update Contributing Guide with new branch flow.

englercj:
Added more details, links to GitHub docs about forking, and the new
branching flow. As a bonus, also changed some minor things in the
editorconfig.

Best viewed in Side-by-Side view.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,8 +9,5 @@ trim_trailing_whitespace = true
 indent_style = space
 indent_size = 4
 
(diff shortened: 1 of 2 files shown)


themoonrat:
You might want to comment on which node version is required. The pixify stuff has object.assign for example, which meant my work required 0.12.x wouldn't compile pixi

englercj:
Good call, I will update this when I get home.

PR #2986, opened 2016-09-22
Improve Bower Workflow

bigtimebuddy:
## Overview

This change removes the **bin** files from version control and create a process with allows for Bower releases to be published on Github. 

This requires the publishing to be done with NPM _first_ and then the Github release is created (this is a change to our existing workflow @GoodBoyDigital ). To do a release, first we publish to NPM:

```bash
# "patch" can also be "minor" or "major"
npm run publish:patch
```

This script bumps the release, rebuilds and pushes everything in current branch to the "release" branch. NOTE: when a Github release is created, it needs to reference the "release" branch (not "master") in order to work for Bower. 

Builds are versioned only in the "release" branch. For folks looking for the latest build, they can use/download from our [S3 bucket](https://s3-eu-west-1.amazonaws.com/pixi.js/dev/pixi.js).

### Changes

* Cleaned up the Bower ignore list
* Updated Travis to not build twice


.travis.yml
```diff
@@ -25,7 +24,7 @@ before_script:
   - Xvfb :99 -screen 0 1024x768x24 -extension RANDR &
 
 script:
-    - xvfb-maybe npm run build
+    - xvfb-maybe npm install
```
bigtimebuddy: `npm install` also triggers the `build` command.

englercj: Why not let the install script do this, and run tests here?

package.json
```diff
@@ -29,7 +29,12 @@
     "prebuild": "npm run lint && npm run clean",
     "build": "pixify -n PIXI -o pixi",
     "postbuild": "npm test",
-    "docs": "jsdoc -c scripts/jsdoc.conf.json -R README.md"
+    "docs": "jsdoc -c scripts/jsdoc.conf.json -R README.md",
+    "publish:patch": "npm version patch --no-git-tag-version && npm publish",
+    "publish:minor": "npm version minor --no-git-tag-version && npm publish",
+    "publish:major": "npm version major --no-git-tag-version && npm publish",
```
englercj: Why no git tag? Does that happen in the `postpublish` script?

.travis.yml
englercj: Why not let the install script do this, and run tests here?

.gitignore
```diff
@@ -26,4 +26,6 @@ docs/
 examples_old/
 
 # jetBrains IDE ignores
-.idea
\ No newline at end of file
+.idea
+
+/bin
```
englercj: Would you mind moving the jeyBrains stuff above "project ignores" and adding `bin/` to project ignores?

package.json
```diff
@@ -29,7 +29,12 @@
...
+    "prepublish": "npm run build",
```
englercj: Can people who clone and run `npm install` succeed? Does this run *before* installing deps, or after?

package.json
englercj: Reread your PR post, are you using the GitHub Releases page to create the tags?

.travis.yml
bigtimebuddy: The tests are run here too. NPM installs dependencies, the build is run, then tests are run (just calling `npm install`). I had split this into two command, but you only need to call `npm install`.

package.json
bigtimebuddy: If you let `npm version` do the Git tag, it tags the wrong commit/branch. It needs to be done with GitHub Releases to make sure it tags the **release** branch.

.gitignore
bigtimebuddy: Done.

package.json
bigtimebuddy: Testing reveals it runs after installing dependencies. From the NPM docs about **prepublish**: `prepublish: Run BEFORE the package is published. (Also run on local npm install without any arguments.)`

.travis.yml
englercj: Not sure how I feel about the tests being in postbuild, I don't feel like tests are part of building I think those should be separated.

bigtimebuddy: Okay, I detached tests from the build process. Should now all be really explicit.

bigtimebuddy:
@GoodBoyDigital could you weigh in or merge? I think this is a good publish process update and removes **bin** files from versioning.

GoodBoyDigital:
Let's do it!

PR #2995, opened 2016-09-22
Update Container.js

kaansoral:
If one is reusing Sprite's before removing them from their deceased parent, this modification is needed, as `this.children` will be null

Around the 4.0.1 release, I started getting exceptions with the same code that I was using, I don't know what caused it, but this fixes it

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -213,6 +213,9 @@ class Container extends DisplayObject
         }
         else
         {
+            if(!this.children) return;
+            // If one is reusing Sprite's before removing them from their deceased parent, this is needed, as this.children will be null
+            
             let index = this.children.indexOf(child);
 
             if (index === -1)


englercj:
Calling `.removeChild` on a destroyed container is an error. If you are hitting this error, it is a signal that your code is doing things out of order. I recommend fixing the cause of this error in your code, not hiding the error with a null check.

kaansoral:
I'm not calling .removeChild, but .addChild is automatically calling .removeChild

kaansoral:
It's also a Regression, as the same usage worked before - what is the intended way of reusing Sprite's of a deceased parent?

Looking at the code, I might add a manual "freeChildren" routine similar to this, that doesn't destroy but reset the `parent` property:

```

    var oldChildren = this.children;
    this.children = null;

    if (destroyChildren)
    {
        for (var i = oldChildren.length - 1; i >= 0; i--)
        {
            var child = oldChildren[i];
            child.parent = null;
            child.destroy(options);
        }
    }
```

englercj:
You do not reuse objects you call `.destroy` on. Destroy means "I am done with this object, remove it from memory completely".

If you want to reuse the sprites in a container, but are throwing the container object away then just tell it not to destroy children:

```js
const saveTheSprites = container.children;

container.destroy({ children: false });

// all the sprites in saveTheSprites are still valid and can be used for stuff
```

Any object that has `.destroy()` called on it is done, and should never be used again. In fact you should lose any references to it as soon as you call destroy so that it can be GC'd.

kaansoral:
To clarify, I'm not destroying and re-using anything, this was my custom fix:

```function free_children(sprite)
{
	if(!sprite.children) return;
	for(var i=0;i<sprite.children.length;i++)
	{
		sprite.children[i].parent=null;
	}
}
```

I agree with the second part of your comment, it seems the `.parent` properties need to be nullified even if they are not destroyed

PR #2997, opened 2016-09-22
Adding support for pointer events

themoonrat:
And for devices that don't support pointer events, it normalises mouse / touch events to give the same data that a pointer event would

Re-submit of https://github.com/pixijs/pixi.js/pull/2476

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -640,7 +640,7 @@ class InteractionManager extends EventEmitter
         }
 
         // if the user move the mouse this check has already been dfone using the mouse move!
-        if(this.didMove)
+        if (this.didMove)
         {
             this.didMove = false;
             return;
@@ -672,16 +672,16 @@ class InteractionManager extends EventEmitter
      * @param eventData {object} the event data object
      * @private
      */
-    dispatchEvent(displayObject, eventString, eventData )
+    dispatchEvent(displayObject, eventString, eventData)
     {
-        if(!eventData.stopped)
+        if (!eventData.stopped)
         {
             eventData.target = displayObject;
             eventData.type = eventString;
 
             displayObject.emit(eventString, eventData);
 
-            if(displayObject[eventString] )
+            if (displayObject[eventString])
             {
                 displayObject[eventString](eventData);
             }
@@ -696,7 +696,7 @@ class InteractionManager extends EventEmitter
      * @param  {number} x     the x coord of the position to map
      * @param  {number} y     the y coord of the position to map
      */
-    mapPositionToPoint(point, x, y )
+    mapPositionToPoint(point, x, y)
     {
         let rect;
         // IE 11 fix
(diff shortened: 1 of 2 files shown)


kaansoral:
I just discovered a critical issue that prevents all mouse events for a small subset of PC users

One player I worked with had touch-action: none on his canvas

The issue wasn't on 4.0.2, but it was on the dev 4.0.3 build I was using, I wonder whether if this is the cause

staff0rd:
So this causes #3167, and more specifically does not support any backwards compatibility, in that `mousedown`, `mouseup` etc are no longer called on devices that support pointer events.  Is this expected?

themoonrat:
Not expected, no :/ It appears that browsers that support the pointet events will not send out mouse events if the pointer ones are listened to. I shall investigate further.

GoodBoyDigital:
Hey chaps! This one has turned into a pretty big issue for one of our projects thats about to launch. IE11 what a pain huh?

@themoonrat is this something that ya think is a straight forward issue to address? Ideally I would love to have this sorted for a minor release on Wednesday. If its a bit of a tricky one I suggest that we revert this commit so that the issue is resolved for users. We can then take our time to understand how and why IE11 is being a pain. 

Let me know your thoughts!

Cheers

themoonrat:
@GoodBoyDigital I'll take a look tonight, and let you know if I get anywhere / can get anywhere / cry

GoodBoyDigital:
good luck man! Il be there with the hanky if needed 👍

kaansoral:
I would appreciate a PostMortem too, this issue hurt me when it mattered the most, while my game was receiving initial attention, I've lost a lot of potential players: https://github.com/pixijs/pixi.js/issues/3158 (Not blaming anyone, but just saddens me)

Also on https://github.com/pixijs/pixi.js/issues/3167 There's some advice related to using "pointerdown" to replace "mousedown"+"touchdown"

So basically, assuming the issue is fixed or patched, do we use "pointerdown" along with "mousedown"+"touchdown"? (I assume so)
Anyway, to sum up, a usage scenario+example would be nice

themoonrat:
Hi @kaansoral 
First of all, very sorry about the bug.... I switched my game over to just use pointer events so never noticed the issue :/ 

Regarding usage: I personally just use pointer events now. For browsers that support them natively, those are sent through. For browsers that don't support them natively, mouse&touch events get converted into a pointer events with the required properties from the pointer spec added in.

However, the regular mouse and touch events are still sent through (now, after this PR with the fix).

So my advice is EITHER 
listen to mouse AND touch events
OR
listen to JUST pointer events.

kaansoral:
Ah, thanks

I started refactoring now, basically replacing "mousedown","touchstart" with "pointerdown"

However, "rightdown" worried me, when the issue happened, both left and right clicks weren't working, and as far as I checked, there is no "rightdown" replacement in the pointer event-set?

(I also read the Code, checked for a documentation, but came up empty)

themoonrat:
This is the page you are looking for: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events
Search for 'Determining button states'

To get to this info on a PIXI event for pointerdown, you'd look at
event.data.originalEvent.button or event.data.originalEvent.buttons

kaansoral:
Thanks for the info, is it verified that "rightdown" works as before on devices that have pointer events enabled? (I mean testing-wise, I can't test it myself, as I don't have such a device/system)

themoonrat:
'rightdown' should still be sent on devices that support pointer events. I'll verify this on some devices in a few hours time when I have access to them and the little test page I made.

(ps. the latest chrome versions now support pointer events, so it's not just an IE & Edge thing anymore)

kaansoral:
I deployed 4.1.1 - People started reporting the inability to click again, so I reversed to 4.0.2 :(

themoonrat:
4.1.1 didn't include anything other than a blank .npmignore file from 4.1.0 as described in the release notes https://github.com/pixijs/pixi.js/releases/tag/v4.1.1
I believe a 4.2.0 release is due soonish which would contain this pointer event fix

kaansoral:
Ah, apologies then, I assumed the minor release @GoodBoyDigital mentioned was 4.1.1

Looking forward to 4.2.0 then :)

englercj:
We use semantic versioning (http://semver.org), where the version is defined as `Major.Minor.Patch`.

PR #3000, opened 2016-09-22
fixes autoDetectRenderer in core/index.js

stbaer:
`core.` being used inside `core = {...}` and small jsdoc change.


stbaer:
Actually I don't think it's a real bug, for me it's just cleaner to use WebGLRenderer, CanvasRenderer and utils directly there.

PR #3017, opened 2016-09-26
Allow anchors to be asssigned without breaking ObservablePoint

marekventur:
While trying to switch our games to PIXI 4 we came across loads of cases where we did

```js
sprite.anchor = new PIXI.Point(0.5, 0.5);
```

which causes v4 to break in weird ways. 

This PR uses the same approach used in https://github.com/pixijs/pixi.js/blob/dev/src/extras/BitmapText.js#L366 to make assignments save again and to make updating from v3 to v4 a little easier.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -33,8 +33,9 @@ class Sprite extends Container
          * Setting the anchor to 1,1 would mean the texture's origin point will be the bottom right corner
          *
          * @member {PIXI.ObservablePoint}
+         * @private
          */
-        this.anchor = new math.ObservablePoint(this.onAnchorUpdate, this);
+        this._anchor = new math.ObservablePoint(this.onAnchorUpdate, this);
 
         /**
          * The texture that the sprite is using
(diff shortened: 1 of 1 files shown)


marekventur:
rebased, I hope this fixes the eslint failures

bigtimebuddy:
@marekventur looks like the [unit-tests are failing](https://travis-ci.org/pixijs/pixi.js/builds/162804267#L411-L419). You can run `npm run test:debug` to use DevTools to debug what's happening there.

ivanpopelyshev:
Yep, that's what I talked about long time ago :)

marekventur:
I'm sorry for the failing tests, I've forgot to update one reference.

GoodBoyDigital:
Looks good @marekventur !
My one comment would be, how do you guys feel about the (typeof value === 'number') check as this is not consistent with our other point setters we have in pixi.

marekventur:
Happy to change it (both here and in BitmapText) if that's preferred.

ivanpopelyshev:
I think its optional, but why not :) Anchors aren't set that often

marekventur:
Updated to only do the ```copy()``` bit without the ```typeof value === 'number'``` check

GoodBoyDigital:
Nice one @marekventur 👍

marekventur:
thanks!

PR #3018, opened 2016-09-26
Fix linting issues

bigtimebuddy:
### Fixed

* Minor eslint errors


englercj:
OOPS, we were editing tags at the same time

PR #3020, opened 2016-09-26
Fix readonly errors when normalizing touch events on ios

themoonrat:
ios10 puts some readonly properties that aren't valid touchevent spec but are in the pointer spec. Trying to overwrite these causes a failure in the normalization code, so there is now a check for each property before changing it.

also slightly tweaked the name of the flag to enable the normalization and made them public; so devs can find and switch the behaviour off if they wish.


englercj:
Can we just use our own intermediate objects instead of assigning to the event object? I feel like that might be a better way to go about this.

themoonrat:
Yeah sorry, started out as wanting to modify the bare minimum, and now its an ugly cludge. Will refactor into something nicer.

themoonrat:
I know it's ugly, but I'm struggling to come up with an alternative option :/
It is possible to clone an event via
`let newEvent = new originalEvent.constructor(originalEvent.type, originalEvent);`
But there are side effects. Stop propagation isn't going to work if you have a different event object. And the .isTrusted flag on the cloned object is false. Suggestions welcome!

englercj:
I meant something like this:

```js
class OurEventWrapper
{
    constructor(event)
    {
        this.button = event.button || event.touches.length ? 1 : 0;
        // ...
    }
}

// ..........

if (this.normalizeTouchEvents && event.changedTouches)
{
    const evt = new OurEventWrapper(event);
}
```

I do this in Fae by abstracting events into a [`Pointer`](https://github.com/Fae/fae/blob/master/plugins/interaction/src/Pointer.js) class which can be setup from an event:

https://github.com/Fae/fae/blob/master/plugins/interaction/src/Pointer.js#L406

I plan to do this with the new v5 interaction manager, so I'll just merge this for now and redo it in v5.

Issue #2882, opened 2016-08-30
Cannot read property 'indexOf' of null


Dadibom:
Getting this error: Cannot read property 'indexOf' of null

in this part of pixi: 
```
i.prototype.removeChild = function(t) {
                var e = arguments.length;
                if (e > 1)
                    for (var r = 0; r < e; r++)
                        this.removeChild(arguments[r]);
                else {
                    var i = this.children.indexOf(t);
                    if (i === -1)
                        return;
                    t.parent = null ,
                    n.removeItems(this.children, i, 1),
                    this.onChildrenChange(i),
                    t.emit("removed", this)
                }
                return t
            }
```
after calling sprite.destroy();

this.texture is null when this happens, so I'm guessing that the problem is that the parent sprite was destroyed before the child sprite. But I shouldn't have to worry about this, a destroy call should work regardless.

ivanpopelyshev:
So it seems that we have a new bug after we patched "removeChild" and "destroy()" methods

Dadibom:
Should be as simple as adding a if(!this.children){return;} ?

PR #3021, opened 2016-09-26
Fix for "Cannot read property 'indexOf' of null #2882"

PatrickGormley:
removeChild function checks if there is anything to remove first. #2882

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -202,6 +202,11 @@ class Container extends DisplayObject
     {
         const argumentsLength = arguments.length;
 
+        // break out if there are not children to remove
+        if (!this.children)
+        {
+            return child; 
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
Hi @PatrickGormley children should always be there?

englercj:
The only time you should get this error is if you are trying to remove a child from a previously destroyed container, which isn't something that should happen unless you have an error in your code somewhere.

In fact, in the latest `dev` we never even set `.children` to null at all. Not even on destroy.

PatrickGormley:
Yep, looking though the code and the destroy() functions again I came to the same realization. Guessing the bug (#2882) should be closed out then?

englercj:
Closed!

Issue #2940, opened 2016-09-11
Setting BaseTexture.resolution requires BaseTexture.update() prior to any effect


staff0rd:
In [this pen](https://codepen.io/staff0rd/pen/qaOKWQ), a 100x100 and a 200x200 image is loaded via resource loaded, with the latter set to resolution=2 and the former defaulting to resolution=1.  Displaying sprites from these textures displays two equally sized textures on a renderer that has resolution=2.

I then use a PIXI.Graphics to generate the same sized textures and set resolution=2 on the second's baseTexture.  Without calling baseTexture.update() after setting the resolution, the second texture is rendered at the wrong size.

<img src="https://cloud.githubusercontent.com/assets/6289998/18415886/7549aeec-7845-11e6-9752-93feb00eb6e3.png" width="400" height="400" />

I think that his means that either BaseTexture.resolution should be a property setter that calls this, or, the documentation on the field should be updated to advise the required call to `.update()`, or, the field should be readonly.

ivanpopelyshev:
Will this work for you?

```js
generateTexture(myGraphics, undefined, 2)
```

staff0rd:
Bah - I didn't even check that function's parameters (which is `generateCanvasTexture` FYI).  

However, as seen in [this pen](https://codepen.io/staff0rd/pen/wzKXAg?editors=0010), updating to use the following still doesn't work;
```
g.generateCanvasTexture(undefined, 2);
```

ivanpopelyshev:
```js
g.beginFill(color);
  g.drawRect(0, 0, 100*resolution, 100*resolution);
  g.endFill();
```

remove the multiplication.

staff0rd:
I don't think that's correct - the resulting baseTexture has realWidth and realHeight set to 100, unlike the @2x image from the loader which has those values set to 200.

staff0rd:
Sorry - I was looking at the wrong pen, removing the multiplication does seem to render it correctly.  However it kind of puts it in a weird state - realWidth and realHeight are 200, but so are width and height.  Also, if I then call baseTexture.update() on it, it shrinks smaller and width/height sets to 100.

ivanpopelyshev:
Yeah, its a weird state. So

- add a note to resolution in jsdocs
- fix the weird state

staff0rd:
After the fix from #3022 I can properly use generateCanvasTexture with resolution instead of multiplication.  Thanks @ivanpopelyshev.

[...]

PR #3022, opened 2016-09-27
Properly set resolution

staff0rd:
Sets the generated texture's resolution properly and calls `texture.update()`.

Resolves the underlying issue that caused #2940.

I wonder if BaseTexture.resolution docs also need a comment noting that if you set it manually, you must call BaseTexture.update() for it to be effected?


GoodBoyDigital:
pow pow! thanks @staff0rd 👍

PR #3023, opened 2016-09-27
Attach tint docs to correct property

staff0rd:
Will stop PIXI.Sprite generated docs from having a `_tint` member as the first entry, and rather a `tint` member instead.


englercj:
Can't do this, the private assignments will override the setter value.

Instead that comment should be about the private item, and the public one should be on the getter.

PR #3039, opened 2016-09-30
Fixes a destroy error in Sprites with a shared texture

themoonrat:
If you have 2 sprites, with a shared texture, then try to destroy(true) both of them, the first one destroys correctly, but the second one will error because it's texture has already been destroyed. This PR adds a flag to fix that.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -397,7 +397,7 @@ export default class Sprite extends Container
 
         const destroyTexture = typeof options === 'boolean' ? options : options && options.texture;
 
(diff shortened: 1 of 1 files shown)


themoonrat:
Actually looking at it now, it might be because I had a reference to a sprite that I called .destroy(true) on, but also had called .destroy(true) on it's parent container.

In any case, there's no harm in having this extra flag check IMO :)

englercj:
I actually like having it error when things like that happen. Just like if you delete a memory address twice, it is UB.

We could add some `// @if DEBUG` checks in here and let the user know what they did know though.

PR #3041, opened 2016-09-30
Coverage using Floss & Istanbul

bigtimebuddy:
### Added

* Adds support for code coverage using istanbul via the latest version of [floss](https://github.com/jiborobot/floss/releases/tag/v1.2.0). Can be generated using `npm run coverage`. Open **coverage/index.html** to see report.


bigtimebuddy:
Example coverage report:
screen shot 2016-09-30 at 4 02 59 pm

englercj:
Should we deploy the coverage results to S3 so we can track improvements over time?

bigtimebuddy:
@englercj good idea, I can add.

bigtimebuddy:
See report here: https://pixi-builds.s3-eu-west-1.amazonaws.com/coverage/coverage/index.html

englercj:
Nice! This is so cool, last step would be to not cover `node_modules`. That should reduce build/test time and cleanup the output too.

Amazing work so far!!

bigtimebuddy:
Okay, I got around this by building Pixi without externals, running coverage and then rebuilding the bundle again (for deployment). Slightly longer build time, but works well. Istanbul is a jerk to tool with their API, so doing something more low-level would take some time.

.travis.yml
```diff
@@ -25,7 +25,6 @@ before_script:
   - Xvfb :99 -screen 0 1024x768x24 -extension RANDR &
 
 script:
-  - npm run clean
```
englercj: I had to add this in because it was uploading old builds even when new ones were created. I think we need to keep this in or we may have travis uploading incorrect builds again.

englercj: I never took the time to figure out what the problem exactly was, but without cleaning first S3 was definitely getting builds uploaded that were not the current one that was done on travis.

bigtimebuddy: Doing a build also does a clean. It's in the prebuild script.

englercj: Hmm, I wonder why that was happening then...

PR #3042, opened 2016-10-01
Adding npm command 'lintfix'

themoonrat:
I find it extremely useful to have an easy to access npm command to do the automatic eslint fixes


englercj:
Do we need this at all, or can we just use `npm run lint -- --fix`?

themoonrat:
You don't _need_ it, but you could argue the visibility to devs is good that it's possible, and, in my ultimate laziness, this ensures it shows up in WebStorm as an npm command I can just double click :P

PR #3043, opened 2016-10-01
VideoBaseTexture fixes and changes

themoonrat:
- https://github.com/pixijs/pixi.js/commit/8d39e810c595b0afc48e542b2d4cad9c204cc7d2 didn't fully solve the video loading issues; the examples page for videos still doesn't work. I've tracked that down to the static 'fromUrl' function, which called video.play(). This was a tad early, resulting in width and height being 0. Removing the .play here doesn't matter, as it is still automatically called within _onCanPlay()

- autoUpdate changed to a getter/setter. It's doc states that it could be used to control when auto updating happened, but this wasn't true, it was actually just an internal variable to check whether it was currently auto updating or not. I've create a couple of private tracking variable for this state, whilst the public getter / setting now truly controls autoUpdate as per docs, and can be changed on the fly.

- moved this class to use the ticker. My expectation as a dev is that if I stop the shared ticker, PIXI would stop anything internal requiring timing, but VideoBaseTexture was an exception as it used rAF directly.

### DIFF
--- a/src/core/textures/VideoBaseTexture.js
+++ b/src/core/textures/VideoBaseTexture.js
@@ -1,5 +1,6 @@
 import BaseTexture from './BaseTexture';
 import { uid, BaseTextureCache } from '../utils';
+import * as ticker from '../ticker';
 
 /**
  * A texture of a [playing] Video.
@@ -54,13 +55,8 @@ export default class VideoBaseTexture extends BaseTexture
         this.width = source.videoWidth;
         this.height = source.videoHeight;
 
-        /**
-         * Should the base texture automatically update itself, set to true by default
-         *
-         * @member {boolean}
-         * @default true
-         */
-        this.autoUpdate = false;
+        this._autoUpdate = true;
+        this._isAutoUpdating = false;
 
         /**
          * When set to true will automatically play videos used by this texture once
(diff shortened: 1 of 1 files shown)


src/core/textures/VideoBaseTexture.js
```diff
@@ -54,13 +55,8 @@ export default class VideoBaseTexture extends BaseTexture
...
 
-        /**
-         * Should the base texture automatically update itself, set to true by default
-         *
-         * @member {boolean}
-         * @default true
-         */
-        this.autoUpdate = false;
+        this._autoUpdate = true;
+        this._isAutoUpdating = false;
```
englercj: Do you need both flags here? I feel like you could get away with just using `_autoUpdate`.

englercj:
> autoUpdate changed to a getter/setter. It's doc states that it could be used to control when auto updating happened, but this wasn't true

It checked the flag in the update loop. If you set it to false, the update loop would stop (since it was using requestAnimationFrame from within the flag check. Having the getter/setter now is important that you use shared ticker.

src/core/textures/VideoBaseTexture.js
themoonrat: I had 2 because i was worried about a certain set of circumstances that could lead to the ticker being joined twice, such as using the autoPlay setter before the canplay comes through from the video source. Thus one flag for the getter/setter on whether it _should_ autoplay, and another for if it _is_ autoplaying

themoonrat:
But the original flag couldn't be used to control the autoPlay functionality. You could only use it to control autoPlay after canplay from the video was emitted, as that would turn on then autoplaying flag no matter what. The change in the pr means you can control that functionality straight after the base texture creation.

Issue #3049, opened 2016-10-02
Clearing RenderTexture on render() only uses r,g,b,a = 0,0,0,0


lordnoriyuki:
See https://jsfiddle.net/zvxydco8/

It draws a red circle into a RenderTexture and then a white circle onto the frame buffer, then draws the sprite with the RenderTexture onto the frame buffer. I'm passing true as the third argument for rendering to the RenderTexture, which should clear the RenderTexture. It clears it to transparent, not to the background color of the renderer. My expected output is the RenderTexture should clear to the background color of the renderer.

One fix would be to insert
`renderTarget.clearColor = this._backgroundColorRgba;`
after setting the renderTarget in [WebGLRenderer](https://github.com/pixijs/pixi.js/blob/master/src/core/renderers/webgl/WebGLRenderer.js) in bindRenderTexture (line 367).

I'm not sure whether that's the best place, and it means that the RenderTexture's RenderTarget inherits the WebGLRenderer's backgroundColor. That'd be better than always using clear with r,g,b,a=0,0,0,0, but maybe there should be a way of setting clearColor on a RenderTexture.

I can put in a PR for the simple fix if that looks ok.

PR #3050, opened 2016-10-02
Proposed fix for issue #3049

lordnoriyuki:


ivanpopelyshev:
Yeah, that was real pain to discover

englercj:
Not sure about this, it may break the current expectations of some filters. I think most of them expect to be using a buffer that when cleared will be transparent. Clearing it to a solid color would make a lot of them act strangely.

englercj:
Anyone else have thoughts about this?

alvinsight:
I agree with @englercj a lot of existing filters would be weird, we probably want to make it an "opt-in" rather than an "opt-out" option.

GoodBoyDigital:
As mentioned above I do not think this is quite the solution.

We could let a RenderTexture / baseRenderTexture have its own clear color? RenderTargets do after all!

englercj:
Closing this PR, but the discussion can continue.

PR #3058, opened 2016-10-04
Improvement of PR #3053

ivanpopelyshev:
Its ok to add child in the end of array.

Actually, it will work even without this "if", <0 will have same result as 0, and >length will be the same as length, its how splice works :)

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -87,7 +87,7 @@ export default class Container extends DisplayObject
      */
     addChildAt(child, index)
     {
-        if (index < 0 || (index > 0 && index >= this.children.length))
+        if (index < 0 || index > this.children.length)
         {
             throw new Error(`${child}addChildAt: The index ${index} supplied is out of bounds ${this.children.length}`);
         }


ivanpopelyshev:
When @englercj merged #3053 he didnt saw my request. Shame on you!

GoodBoyDigital:
Interesting stuff! @ivanpopelyshev - taking what you mentioned into account, what should the line be?

photonstorm:
@ivanpopelyshev you're right, checking `> array.length` is pointless for a splice operation, as it'll default to `length` if over the limit. But checking `< 0` is essential - if they pass in a negative index by mistake it'd splice it that many elements from the _end_ of the array, causing potentially highly unpredictable results.

englercj:
> When @englercj merged #3053 he didnt saw my request. Shame on you!

I saw it, I just disagreed. This check is here to notify people if they are performing out-of-bounds operations. And if they are, it is likely because they have an error in their code. I planned to switch these checks to be `// @if DEBUG` and didn't want to allow for invalid entries in debug, though in production builds when compiled out it would just work.

The negative values are like @photonstorm said, they actually will cause undesired behavior most likely.

GoodBoyDigital:
Very good points @englercj and @photonstorm! Does this PR seem like a pretty decent all-round solution? Happy to merge?

photonstorm:
I think in cases like this you should always honor the contract. The method says it will add the child _at that specific index_. If the index given is out of bounds, then it is impossible to fulfill the contract, so it should abort.

englercj:
@photonstorm I also agree, which is why I didn't make the change that was suggested the first time. I also don't feel like `length` vs `length - 1` is a sword I feel like dying on. I'll let y'all revert and change if you feel it is appropriate.

ivanpopelyshev:
@englercj @photonstorm 

Let us look at openJDK:

```java
class ArrayList {
private void rangeCheckForAdd(int index) {
    if (index > size || index < 0)
    throw new IndexOutOfBoundsException(outOfBoundsMsg(index));
}
}
```

Issue #3059, opened 2016-10-04
Invalid image type in URL


englercj:
When given a url witha  query string, the IMAGE_TYPE regex fails:

https://github.com/pixijs/pixi.js/blob/dev/src/core/const.js#L299

Imported from [this comment](https://github.com/pixijs/pixi.js/issues/2903#issuecomment-251063579):

```
Error: Invalid image type in URL.
at BaseTexture._updateImageType (http://xxxx/js/pixi.js?4:22379:23)
at HTMLImageElement.source.onload (http://xxxx/js/pixi.js?4:22283:27)

(url: "src/img/../img/npc/xxxx.png?v=0.4.0.R42180")
```

This bug was caused by #2614.

Dadibom:
Still doesn't work in 4.1.0

englercj:
Can you clarify what "doesn't work" means please? When I wrote that fix it did in-fact fix the error for me, and it is currently working in the examples.

Dadibom:
Sorry, I just saw the same error pop up and assumed that it was still broken. But in fact, it wasn't the exact same error. This time it's because I load images from a source without .png in the url. I would really like a way to keep doing this.

Dadibom:
You want a new issue for that?

englercj:
Yes, please.

Dadibom:
Alright, sorry for the confusion. #3159

PR #3060, opened 2016-10-04
Replace url extension regex

englercj:
Fixes issue #3059.

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -296,7 +296,7 @@ export const DEFAULT_RENDER_OPTIONS = {
  * @type {RegExp|string}
  * @example `image.png`
  */
-export const IMAGE_TYPE = /\.(gif|jpe?g|tiff|png|svg)$/i;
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
Doesn't browserify add polyfills for Node APIs? Seems like you could do this with `url` and `path` modules:

```js
const url = require('url');
const path = require('path');

const fullUrl = 'http://i.imgur.com/Mvv4bx8.jpg?querystring=true';
const result = path.extname(url.parse(fullUrl).pathname); // '.jpg'
```

englercj:
yes, but the `url` polyfill is pretty big. I ended up removing it from resource-loader due to the size. Since we don't need full URL parsing I figured this would be a lighter solution.

bigtimebuddy:
Makes sense, I'm cool with this then. Good improvement and nice unit-tests.

GoodBoyDigital:
Nice man! 👍 for the tests too!

PR #3064, opened 2016-10-05
Complete addChildAt test coverage

staff0rd:
Some other functions also remove a child from parent, so I included those also.


englercj:
You're the best!

PR #3067, opened 2016-10-05
Removed arrow functions as they mess with scope (this. is undefined)

GoodBoyDigital:


GoodBoyDigital:
Weird not sure what this eslint error means?

  351:48  error  Missing function expression name  func-names

themoonrat:
It's a requirement that every function is named, so that when you debug through a call stack, nothing is anonymous.

So
`let printHelloWorld = function() { console.log('hello world'); }; 
`becomes
`let printHelloWorld = function printHelloWorld() { console.log('hello world'); };`

englercj:
The linter wants function expressions to have a name, or if they are anonymous to be arrow functions. You just need to put a name for the function (this helps debugging).

GoodBoyDigital:
Gotcha! Thanks for clarifying chaps!

englercj:
Looks like you got the names in, but changed the template strings to be single quotes for some reason?

GoodBoyDigital:
ah yes! the ` slashes completely broke how it sublime views the file. Hopefully not a big one?

englercj:
Well they definitely need to change back, because ` is a template string, and single quote is not.

```
var a = 10;

`${a}` === '10'
```

GoodBoyDigital:
oh nice! everydays a school day!

PR #3069, opened 2016-10-05
Renames S3 bucket from "pixi-builds" to "pixi.js"

bigtimebuddy:
### Changed

* Renames the S3 bucket to be “pixi.js” instead of “pixi-builds”. This change makes it easier to remember, consistent with the name of the module and bucket names for other modules (like **pixi-gl-core** and **pixi-filters**).

_Note: links will need to be updated in the Wiki: https://s3-eu-west-1.amazonaws.com/pixi.js/dev/pixi.js_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ deploy:
   - provider: s3
     access_key_id: $S3_ACCESS_KEY_ID
     secret_access_key: $S3_SECRET_ACCESS_KEY
-    bucket: "pixi-builds"
+    bucket: "pixi.js"
     skip_cleanup: true
     acl: public_read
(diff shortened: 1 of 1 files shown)


englercj:
Did you also copy over all the builds in the `pixi-builds` bucket over to the new bucket? I'm having trouble loading S3 right now so I can't check.

bigtimebuddy:
Yes, I copied over all the contents of pixi-builds

englercj:
Note that after this change you cannot use SSL with the bucket subdomain URLS because it has a `dot` in the name which is invalid for the SSL cert. You will *have* to use the `amazonaws.com` and CloudFront URLs.

PR #3071, opened 2016-10-05
Bypass the getter / setter for anchor inside Sprite

GoodBoyDigital:


englercj:
We can do this in TilingSprite in a different PR.

PR #3077, opened 2016-10-07
Custom deployment domain: pixi.download

bigtimebuddy:
### Changed

* Modified the S3 deployment to make it accessible from a custom domain (http://pixijs.download). This is easier to remember for devs and more portable if we ever want to use another deployment option over AWS. 

Example: http://pixijs.download/pixi.js/s3-bucket/pixi.js

All the files and release from the old bucket have been moved over.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,39 +34,39 @@ deploy:
   - provider: s3
     access_key_id: $S3_ACCESS_KEY_ID
     secret_access_key: $S3_SECRET_ACCESS_KEY
-    bucket: "pixi.js"
+    bucket: "pixijs.download"
     skip_cleanup: true
     acl: public_read
     region: eu-west-1
     cache_control: "max-age=60"
     local_dir: bin
-    upload-dir: "$TRAVIS_BRANCH"
+    upload-dir: "pixi.js/$TRAVIS_BRANCH"
     on:
         all_branches: true
         condition: -z $TRAVIS_TAG
   - provider: s3
     access_key_id: $S3_ACCESS_KEY_ID
     secret_access_key: $S3_SECRET_ACCESS_KEY
-    bucket: "pixi.js"
+    bucket: "pixijs.download"
     skip_cleanup: true
     acl: public_read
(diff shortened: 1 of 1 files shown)


englercj:
I'm not sure I understand this, we can put any CNAME we want on top of the CloudFront distribution without changing the bucket name at all.

Why does the bucket name need to change here?

bigtimebuddy:
Yeah, CNAME entry was my first approach.  I created a CNAME entry to the S3 domain (pixi.js.s3-eu-west-1.amazonaws.com). Unfortunately in S3, in order to make this work, [you need to have the bucket match the name of the domain](http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingCustomURLs). 

If you have some insight about how to make this work, would be great. I can share ownership of the domain with you on Namecheap (if you have an account).

englercj:
But you don't for CloudFront, we can use any domain we want for a CloudFront distribution.

If you point your nameservers for that domain over to AWS I can take care of it completely in AWS. Just change your nameservers to these:

```
ns-1675.awsdns-17.co.uk
ns-1104.awsdns-10.org
ns-722.awsdns-26.net
ns-436.awsdns-54.com
```

I'll manage the DNS in Route 53 and point it to the CloudFront distro I already have setup.

englercj:
In fact I went ahead and set it up, after you change the nameservers at your domain host and it propagates, you should be able to use a URL like the one I posted previously.

bigtimebuddy:
DONE. Thanks @englercj!  Should take a bit for the DNS to propagate. Could you grant me access to this in AWS in case we need to add sub-domains (for other modules)?

englercj:
Already did :)

bigtimebuddy:
@englercj You really need to stop anticipating my every move

bigtimebuddy:
How awesome are these links: http://pixijs.download/dev/pixi.js

englercj:
Looks like this is working:

http://pixijs.download/dev/pixi.js

PR #3083, opened 2016-10-09
Small refactoring of TilingSprite

ivanpopelyshev:
Small refactoring of TilingSprite, allows to use renderer parts external plugins (pixi-picture)

### DIFF
--- a/src/extras/webgl/TilingSpriteRenderer.js
+++ b/src/extras/webgl/TilingSpriteRenderer.js
@@ -46,10 +46,46 @@ export class TilingSpriteRenderer extends core.ObjectRenderer {
     }
 
     /**
+     * whether or not tilingSprite can be drawn without transforms and clamps in fragment shader
      *
-     * @param {PIXI.extras.TilingSprite} ts tilingSprite to be rendered
+     * @param {TilingSprite} ts tiling sprite
+     * @returns {boolean} whether or not its simple
      */
-    render(ts)
+    isSimpleSprite(ts)
+    {
+        const renderer = this.renderer;
+        const tex = ts._texture;
+        const baseTex = tex.baseTexture;
+        let isSimple = baseTex.isPowerOfTwo
+            && tex.frame.width === baseTex.width && tex.frame.height === baseTex.height;
+
+        // auto, force repeat wrapMode for big tiling textures
+        if (isSimple)
+        {
+            if (!baseTex._glTextures[renderer.CONTEXT_UID])
+            {
+                if (baseTex.wrapMode === WRAP_MODES.CLAMP)
+                {
+                    baseTex.wrapMode = WRAP_MODES.REPEAT;
+                }
+            }
+            else
+            {
+                isSimple = baseTex.wrapMode !== WRAP_MODES.CLAMP;
+            }
+        }
+
+        return isSimple;
+    }
+
+    /**
+     * For extra pixi plugins use
+     *
+     * @param {PIXI.TilingSprite} ts tiling sprite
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
Its not serious, I just need one review for that

src/extras/webgl/TilingSpriteRenderer.js
```diff
@@ -94,9 +75,41 @@ export class TilingSpriteRenderer extends core.ObjectRenderer {
...
 
-        renderer.bindShader(shader);
+    /**
+     * For extra pixi plugins use
+     *
+     * @param {PIXI.TilingSprite} ts tiling sprite
+     * @param {boolean} isSimple is it repeating texture or not
+     * @param {PIXI.Shader} shader, must be set accordingly to isSimple
+     */
+    renderWithShader(ts, isSimple, shader)
```
englercj: Why is this necessary over just setting the `.shader` property and calling `.render()` ?

src/extras/webgl/TilingSpriteRenderer.js
```diff
@@ -130,17 +143,36 @@ export class TilingSpriteRenderer extends core.ObjectRenderer {
...
 
         const color = tempArray;
+        const alpha = ts.worldAlpha;
 
         core.utils.hex2rgb(ts.tint, color);
-        color[3] = ts.worldAlpha;
+        color[0] *= alpha;
+        color[1] *= alpha;
+        color[2] *= alpha;
+        color[3] = alpha;
```
englercj: What is the benefit of multiplying here instead of in the fragment shader?

src/extras/webgl/TilingSpriteRenderer.js
ivanpopelyshev: so external code can set shader and blendMode and call that function.

ivanpopelyshev:
The thing is, I have that code in separate plugin: https://github.com/pixijs/pixi-picture/blob/master/src/PictureRenderer.js#L184 , and I really don't want to copy existing TilingSprite rendering code there. I hope to move entire library into PIXI a bit later ;)

src/extras/webgl/TilingSpriteRenderer.js
englercj: Well as far as I can tell this is the only place where we multiply in code and not in the shader, so I would say this breaks the current convention.

src/extras/webgl/TilingSpriteRenderer.js
ivanpopelyshev: because blendMode that needs to be set must be different from sprite's blendMode. Of course I can just store it temporarily in other variable...

also shader needs to be bound before that to set some custom uniforms

ivanpopelyshev: Also, which shader to choose depends on isSimple variable.

src/extras/webgl/TilingSpriteRenderer.js
ivanpopelyshev: But SpriteRenderer does that in vertex shader, which is different. I can o it there too, passing it to varying. Do you think that will be better?

src/extras/webgl/TilingSpriteRenderer.js
ivanpopelyshev: Actually, we cant just set ANY shader to TilingSprite and Sprite. What if we make some container for uniforms too, like we do with filter?

ivanpopelyshev:
OK, screw this PR , I will make new one later

PR #3089, opened 2016-10-10
Further Container test coverage

staff0rd:


test/core/Container.js
```diff
@@ -352,6 +358,175 @@ describe('PIXI.Container', () =>
...
+
+        it('should reset scale', () =>
+        {
+            const container = new PIXI.Container();
+
+            container.scale.y = 2;
+            container.height = 5;
+
+            expect(container.height).to.be.equals(0);
+            expect(container.scale.y).to.be.equals(1);
```
englercj: This was unexpected for me, why does setting `height` or `width` reset the scale to `1`? Shouldn't it just be read only or something?

staff0rd: Yeah I dunno - I wrote the reset tests to cover [this functionality](https://github.com/pixijs/pixi.js/blob/185b8db01c268bb06161458a1c8b1ac32b5ab6d2/src/core/display/Container.js#L547-L550).

What should I do?

englercj: This is some seriously old code that unfortunately doesn't include a comment as to why it was added:

https://github.com/pixijs/pixi.js/commit/ecb0d1c5c8932a1217e1c953482b500c24b151ac#diff-aadfaf0cb7e68405ff1d57f9972aaf79

I'm going to merge this now but will put this on my list of things that should be checked on. Thanks!

Issue #2791, opened 2016-08-11
Latest dev returns globalPosition == localPosition


staff0rd:
Having difficulty isolating this one - its not broken in b4c148d95184ec605bb56d214d0e4a57f4426170 which was July 9.  Although I can only repro it via spine, a change in pixijs/pixi.js dev is causing this.

**REPRO**

1. Go to the [dragon spine dev example](https://pixijs.github.io/examples/index.html?s=spine&f=dragon.js&title=Dragon&plugins=pixi-spine&v=dev)
2. Immediately before the `animate()` call in `onAssetsLoaded()` add this code, run it, and check the console;

~~~
var sprite = dragon.skeleton.findSlot("L_rear_leg").currentSprite;
console.log(sprite.getGlobalPosition());
console.log(sprite.position);
~~~

**EXPECTED**
Global and local positions returned;
~~~
Point {x: 366.0400075151632, y: 395.0811625845801}
ObservablePoint {_x: 67.29, _y: 12.62, scope: TransformStatic}
~~~

**ACTUAL**
Local position is returned for both global and local positions;
~~~
Point {x: 67.29, y: 12.62}
ObservablePoint {_x: 67.29, _y: 12.62, scope: TransformStatic}
~~~

Note that [rc3 example](https://pixijs.github.io/examples/index.html?s=spine&f=dragon.js&title=Dragon&plugins=pixi-spine&v=v4.0.0-rc3) produces the expected output.

staff0rd:
It would seem that calling `sprite.getBounds()` immediately before `sprite.getGlobalPosition()` is a temporary work around.

GoodBoyDigital:
looks like we need to add an updateTransform to ```getGlobalPosition()``` with the option of a skipUpdate! (similar to how ```getBounds``` works)

ivanpopelyshev:
Done

PR #3092, opened 2016-10-11
Fixes for #2791 , with tests

ivanpopelyshev:
Fixes for #2791 , with tests


bigtimebuddy:
@ivanpopelyshev looks like build is failing: https://travis-ci.org/pixijs/pixi.js/builds/166717065#L520

PR #3094, opened 2016-10-11
TilingSprite BaseTexture resolution

rkrasky:
I ran into a problem if the base texture resolution is not 1 and the renderer is not WebGL. The resolution auf the base texture is nowhere considered. This is how i solved it. The solution is probably not the best...


ivanpopelyshev:
Actually that looks OK

PR #3096, opened 2016-10-12
More container tests

staff0rd:


englercj:
You're mocking enough that we should use sinon and sinon-chai to do that for you. Feel free to add it in next time you do stuff like this.

bigtimebuddy:
sinon and chai [already exist](https://github.com/jiborobot/floss/blob/master/electron/renderer.js#L17-L22) in the testing suite using **floss**, you can get them globally.

englercj:
Oh, perfect. Then we should clean these up to use sinon mocks/spys instead of manually mocking them out like this.

PR #3098, opened 2016-10-12
TilingSprite bounds fixes

ivanpopelyshev:
and a test

### DIFF
--- a/test/core/TilingSprite.js
+++ b/test/core/TilingSprite.js
@@ -1,6 +1,6 @@
 'use strict';
 
-describe('PIXI.Sprite', function ()
+describe('PIXI.TilingSprite', function ()
 {
     describe('getBounds()', function ()
     {
--- a/test/core/TilingSprite.js
+++ b/test/core/TilingSprite.js
@@ -1,0 +1,27 @@
+'use strict';
+
+describe('PIXI.Sprite', function ()
+{
+    describe('getBounds()', function ()
+    {
+        it('must have correct value according to _width and _height', function ()
+        {
+            const parent = new PIXI.Container();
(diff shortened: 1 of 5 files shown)


test/core/TilingSprite.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+describe('PIXI.Sprite', function ()
```
englercj: TilingSprite

test/core/TilingSprite.js
```diff
@@ -0,0 +1,27 @@
...
+{
+    describe('getBounds()', function ()
+    {
+        it('must have correct value according to _width and _height', function ()
```
englercj: Looks like this test is actually failing:

```
1) PIXI PIXI.Sprite getBounds() must have correct value according to _width and _height:
      AssertionError: expected -350 to equal -150
      + expected - actual
      --350
      +-150
      
      at Context.<anonymous> (/home/travis/build/pixijs/pixi.js/test/core/TilingSprite.js:21:33)
      at runCallback (timers.js:574:20)
```

ivanpopelyshev: That means we have bugs in PIXI.Sprite too :) I'm fixing it.

englercj: The failing test is in the TilingSprite test case, not the Sprite one. It was just incorrectly named, but you can see it comes from the TilingSprite test:

```
/home/travis/build/pixijs/pixi.js/test/core/TilingSprite.js:21:33
```

ivanpopelyshev: I fixed both. Also renamed testset.

Issue #3107, opened 2016-10-14
[Suggestion] move the code about Children to a standalone function


finscn:
Like:

```
Container.prototype.updateTransformChildren = ...
Container.prototype.renderWebGLChildren  = ...
Container.prototype.renderCanvasChildren = ...
```

And call them in   updateTransform /  renderWebGL / renderCanvas .

Then we could extend the  Container via override  xxxxChildren .

finscn:
And share one my extension about Pixi.js

```
Container.prototype.renderWebGLChildren = function(renderer) {
    // In one game loop , the render is the last thing about an entity normally.
    // So we could remove the entity in render function.
    var i = 0,
        len = this.children.length;
    while (i < len) {
        var child = this.children[i];
        if (child._to_remove) {
            len--;
            child.parent = null;
            this.children.splice(i, 1);
            this.onChildrenChange(index);
            child.emit('removed', this);
            continue;
        }
        child.renderWebGL(renderer);
        i++;
    }

};

```

When I want to remove a child, don't do it immediately . ( set  child._to_remove=true ) .
When render children , do that  together , it's more fast and reliable.

endel:
Do you have measurements of how faster your approach could be?

finscn:
Hi @endel , performance is not the key point of this thread .
Because my game is special , I just use pixi  as a  render engine , not a all-in-one engine , So my test case maybe useless for other user.

the key is  https://github.com/pixijs/pixi.js/issues/3106

GoodBoyDigital:
In our game engine, we have this logic in our app rather than pixi itself. Main reason being is that the style you are using is not applicable to all situations. Hope that helps!

finscn:
@GoodBoyDigital  , yes ,  you are right .
The logic in the second floor  ( https://github.com/pixijs/pixi.js/issues/3107#issuecomment-253720685 )  is not applicable to all situations. 
But , please FORGET it  . the key of this issue is 
```
Move the code about Children to a standalone function, please.
``` 
 "Move the code about Children to a standalone function" is important for Flexibility and Extensibility.

If don't do this, developer will hard to extend Container for their own situation.

finscn:
I've create a PR  https://github.com/pixijs/pixi.js/pull/3108  for this issues .

themoonrat:
As the PR was rejected, I'm going to close down this issue

PR #3108, opened 2016-10-14
Update Container.js about #3107

finscn:
I hope this PR could  make my point clear to you. @GoodBoyDigital  @englercj 

https://github.com/pixijs/pixi.js/issues/3107

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -375,6 +375,11 @@ Container.prototype.updateTransform = function ()
     //TODO: check render flags, how to process stuff here
     this.worldAlpha = this.alpha * this.parent.worldAlpha;
 
+    this.updateTransformChildren();
+};
+
+Container.prototype.updateTransformChildren = function()
+{
     for (var i = 0, j = this.children.length; i < j; ++i)
     {
         this.children[i].updateTransform();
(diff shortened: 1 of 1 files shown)


englercj:
I see what you are looking for, what I'm not clear on is the use-case. What is the situation where you need to change only the loop iterating children?

finscn:
@englercj , example: I want to sort children before render , or do something like https://github.com/pixijs/pixi.js/issues/3107#issuecomment-253720685 , or something else.

The key point is NOT what I want to do now , is  when I want to do I could do.

Flexibility and Extensibility are very important for a Framework / Engine

englercj:
> Flexibility and Extensibility are very important for a Framework / Engine

I absolutely agree, but we have to balance that against performance. Adding a function call to the hottest function in the engine is something we only want to do for a very specific reason. If you want to sort children you can do that in a couple other hook locations, or just run it as part of your updateTransform.

finscn:
This   function call  is not the hotspot of performance , if you very care about "function call" ,
I think there are many unnecessary "function call"  in PIXI.js , 
e.g.  utils.sign , utils.removeItems  ,  the getter / setter of  width / height / tint / texture  ....

englercj:
> utils.sign , utils.removeItems , the getter / setter of width / height / tint / texture

Getters/setters are inlined when trivial, as is util.sign. And we already talked about util.removeItems, it isn't adding a fn call but is removing GC trash.

finscn:
I still think THIS "function call" is not the hotspot of performance.

GoodBoyDigital:
Hi @finscn! Thanks for the PR. 

I guess there are many different ways to skin a cat :)

The situation being solved here is considered an edge case and I believe it can be sorted out by overriding the functions to ensure pixi does what you need it to do.

Cheers!

PR #3133, opened 2016-10-17
Fix for tilingsprite destroy

naranjamecanica:
When a tiling sprite was destroyed the setter was called, which would
set the point value to null. This gave an error while Point.copy does
not like null values.

My fix is to bypass the setters and set the object properties to null.
I’ve added the uvTransform also.


ivanpopelyshev:
Thank you!

PR #3138, opened 2016-10-18
Integration Tests

bigtimebuddy:
## Overview

Builds a structure for creating rendering/integrations tests that are run with the rest of the unit tests. These are rendered in WebGL and Canvas renderers and then compared against a pre-existing solution.

Rendering solutions are generated by using this tool: `npm run renders`


bigtimebuddy:
Travis CI seems to not be able to run WebGL content natively (anyone have experience with this?), but it's still valuable to run WebGL renders locally. Code coverage is going to be slightly different in each case: local coverage is about 51.24%, on Travis: 47.92%.

englercj:
Try specifying the Firefox version to see if a more up-to-date version works:

```
addons:
  firefox: "49.0"
```

If that doesn't work, you can use headless-gl: https://github.com/stackgl/headless-gl#how-can-i-use-headless-gl-with-a-continuous-integration-service

bigtimebuddy:
This approach isn't perfect and I hear you about the Mocha-inconsistency with the other tests. The trade-off here is that it is very easy to generate (or re-generate) a rendered solution from a bit of code because the rendering tests are authored as somewhat standalone. Having some GUI tooling for generating the test makes it easier to debug and preview the render tests. 

It would be easy for these render tests to support sinon, chai, etc. That's a good suggestion.

I could also potentially move the pixi-tests-tool code/functionality here too.

bigtimebuddy:
Okay, I integrated the app for creating render solutions (`npm run renders`) and provided full-parity with Floss for running the render tests: (including sinon, chai, etc.).

englercj:
> full-parity with Floss for running the render tests: (including sinon, chai, etc.)

looks like the tests are still not using mocha/chai? Is that intended?

bigtimebuddy:
Yes, the render test code is now using chai's `assert` (they were using Node's before). It is intended that the render code be externalized so the solutions can be re-generated. 

Someone could create additional integration Mocha tests, but with these I really wanted to validate that the output render is exactly what it's suppose to be. So these render Mocha test simply load a snippet of code and load a solution and compares the two. If they are within a tolerance, the test passes.

englercj:
Why is it using assert there instead of expect-style like the rest of the tests?

bigtimebuddy:
Because I'm lazy :) I can change to expect

Issue #3152, opened 2016-10-19
[Suggestion] Don't use dynamic parameters of ES6


finscn:
the function with  dynamic parameters  like  this (ES 6) : 
```
   // in  class Container
    addChild(...childs)
    {
        // ......
    }
```
will be transformed to  (ES 5): 
```
    Container.prototype.addChild = function addChild() {
        for (var _len = arguments.length, childs = Array(_len), _key = 0; _key < _len; _key++) {
            childs[_key] = arguments[_key];
        }
        // ......
    }
```

The performance of the porting code is bad.

GoodBoyDigital:
ouch, that is slow :/

finscn:
```
var childs = Array.prototype.slice.call(arguments, 0);
```
is enough.

And why use childs  , children is a good name, I think.

GoodBoyDigital:
thats still creating a new array. The fastest way is how we had it before:

```
var argumentsLength = arguments.length;

    // if there is only one argument we can bypass looping through the them
    if(argumentsLength > 1)
    {
        // loop through the arguments property and add all children
        // use it the right way (.length and [i]) so that this function can still be optimised by JS runtimes
        for (var i = 0; i < argumentsLength; i++)
        {
            this.addChild( arguments[i] );
        }
    }     
    else
    {
          //add child code
    }
```

finscn:
I think best way is separate it to 2 functions:
addChild
addChildren

englercj:
> var childs = Array.prototype.slice.call(arguments, 0);

You should not use that, it causes an argument leak deoptimization. The compiled ES5 from babel is the correct way to convert `arguments` to an array (and is faster that slice anyway).

In this case, not creating the array at all is the obvious solution. However, if the function would be creating an array from arguments anyway then rest-params and the associated transpiled code is perfectly fine.

PR #3153, opened 2016-10-19
faster add child #3152

GoodBoyDigital:
No new arrays being created here! Also tweaked the eslintsrc to   allow
us to not have to use rest.


src/core/display/Container.js
```diff
@@ -64,18 +74,14 @@ export default class Container extends DisplayObject
 
             child.parent = this;
 
-            // ensure a transform will be recalculated..
-            this.transform._parentID = -1;
```
englercj: Why was this removed? Pretty sure it is necessary to reset the transform/bounds Ids when adding the child to a new parent.

GoodBoyDigital:
Arrrg! Where would we be without your eagle eyes! Great catch dude

englercj:
Fixed!

Issue #3155, opened 2016-10-19
Custom hitArea for graphics with 4.1.0 broken?


stbaer:
Hi, I just updated to 4.1.0. I have many graphics with custom hit areas and they all seem to be broken. Can't really provide a reduced test case / fiddle at the moment (no time right now, maybe in the next days) and didn't follow the latest changes enough to have any idea why this is happening. Are there any breaking changes for hitArea's?

englercj:
Not that I am aware of. Will likely need a test case to investigate :(

stbaer:
Here's a simple example that can be run on [pixijs.github.io/examples](http://pixijs.github.io/examples/), just toggle between 4.0.3 and 4.1.0 to see the difference.

```
var renderer = PIXI.autoDetectRenderer(800, 600, { antialias: true });
document.body.appendChild(renderer.view);

var stage = new PIXI.Container();
var graphics = new PIXI.Graphics();
var hitArea = new PIXI.Polygon(190,190, 410,190, 410,210, 190,210);
graphics.lineStyle(2, 0xffd900, 1);

// draw a shape
graphics.moveTo(200,200);
graphics.lineTo(400, 200);

graphics.hitArea = hitArea;
graphics.interactive = true;
graphics.buttonMode = true;

// just for making the hit area visible
graphics.lineStyle(1, 0xff0000, 0.8);
graphics.drawPolygon(hitArea); 

stage.addChild(graphics);

animate();
function animate() {
    renderer.render(stage);
    requestAnimationFrame( animate );
}
```

Custi:
Do you have any idea when this fix will be available via npm?

stbaer:
Hi @Custi - if you want you can use `"pixi.js": "git+https://github.com/stbaer/pixi.js.git#tmp-release"` until this is fixed in a new release in pixi. It's a build of pixi 4.1.0 + this fix.

Issue #3151, opened 2016-10-19
Interactivity problem with masked elements (v4.1.0)


jolelievre:
Hello, I just updated pixi with the last update (from 4.0.3 to 4.1.0) and I immediately noticed a bug with elements using masks, the interactivity (over, click, even the cursor changing) don't work when a mask is assigned. I use a PIXI.Graphics object as the mask.

GoodBoyDigital:
Thanks for sharing! If you could provide a fiddle for us to debug that would be ace 👍

jordanhaven:
Here's a simple fiddle: https://jsfiddle.net/tt4zd79p/

Hopefully that helps.

jordanhaven:
Narrowed it down, the bug appears to have been introduced in this commit: https://github.com/pixijs/pixi.js/commit/475c66b5da11abf367ef0b78d043fa3c9cf3905d

jordanhaven:
One part of the problem appears to be here: https://github.com/pixijs/pixi.js/blob/master/src/core/graphics/Graphics.js#L858

If you remove that conditional, the interaction is detected (cursor changes, click events get all the way through to `dispatchEvent` in InteractionManager, and the correct `displayObject` is identified).

Past that, there are no visible errors with the `displayObject.emit(eventString, eventData);` or `displayObject[eventString](eventData);` calls in `dispatchEvent`, but the handler on the `displayObject` still doesn't fire :/

jordanhaven:
So yeah, just adding a fill to the mask (`mask.beginFill(0xFFFFFF); mask.drawRect(0,0,w,h); mask.endFill()` works around the problem for us. Don't need to set a hitArea or interactive/buttonMode on the mask, either. New (working) fiddle: https://jsfiddle.net/u0z7yasd/

staff0rd:
pen via @ricklamers; http://codepen.io/ricklamers/pen/ygVqaB
reduced; http://codepen.io/staff0rd/pen/qRRmpB?editors=0010

As above, the workaround is to use `beginFill`.

Another workaround is (should be) to generate a texture and use a sprite, although strangely this looks as though the interaction is preserved but the masked item is no longer displayed (BUG);
http://codepen.io/staff0rd/pen/xggdYv?editors=0010

Setting a `filterArea` or `hitArea` on either the mask or the masked item also does not solve this (BUG); http://codepen.io/staff0rd/pen/rjjmQg?editors=0010.

Tests incoming.

[...]

PR #3165, opened 2016-10-20
fixes issue #3155 - Polygon.contains not working

stbaer:
This might also fix issue #3151


ivanpopelyshev:
@xerver you broke it! https://github.com/pixijs/pixi.js/commit/22c5248a51059979278b7cefc11e3a7e7078c008

Please check all other changes in that commit!

ivanpopelyshev:
@englercj why had it "++i" instead of "i++" everywhere? I looked in commit, that was the only affected place. But WHY? Do we have a new lint rule about ++i ?

PR #3171, opened 2016-10-20
Document skipHello, deprecates _saidHello

bigtimebuddy:
### Overview

In v4.1.0, `PIXI.utils._saidHello = true` was removed in favor of `PIXI.utils.skipHello()`. This PR provides the backward compatibility. It also documents skipHello.

### DIFF
--- a/src/deprecation.js
+++ b/src/deprecation.js
@@ -622,8 +622,8 @@ core.utils.canUseNewCanvasBlendModes = () =>
 let saidHello = true;
 
 /**
- * @method
  * @name PIXI.utils._saidHello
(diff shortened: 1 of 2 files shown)


src/deprecation.js
```diff
@@ -618,3 +618,27 @@ core.utils.canUseNewCanvasBlendModes = () =>
 
     return core.CanvasTinter.canUseMultiply;
 };
+
+let saidHello = true;
+
+/**
+ * @method
```
englercj: method?

bigtimebuddy:
Seems good, @englercj?

PR #3172, opened 2016-10-20
Fix typo

marciopuga:
degredation -> degradation


ivanpopelyshev:
oops, we have to make it in DEV first. My bad, going to merge it now

Issue #3005, opened 2016-09-23
Is Mobile


antag0n1st:
The code is already part of PIXI (using isMobile.js) , can that be exposed ?

stbaer:
This can be closed now that #3175 is merged.

PR #3175, opened 2016-10-21
expose isMobile (PIXI.utils); added tests to core/util

stbaer:
See issue #3005

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -9,7 +9,27 @@ let saidHello = false;
 /**
  * @namespace PIXI.utils
  */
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
I think we should provide documentation for `isMobile` if we are going to expose it.

englercj:
We don't document the Loader or EventEmitter but we expose those. I think just having something that says "You can find docs about this thing here: <url>" is good enough. You don't want to duplicate all the docs from a dependency.

stbaer:
Ok, I added jsdoc comments for EE3 and isMobile with links to the github repos. Also added a doc comment for pluginTarget.

PR #3185, opened 2016-10-24
ES6 updates for tests

staff0rd:


englercj:
Looks like you may need to rebase to remove these conflicts.

staff0rd:
hmm, that didn't turn out as I expected.  1e1ae36 is in fact the merge that cleans up the conflict.  Should I close this and open another to clean the commits up?

englercj:
If there is a commit that says `rebase` then I don't think you actually rebased :P

Make sure your local dev branch is up-to-date then checkout your PR branch, then use `git rebase dev`. You should end up with a branch that is just a couple commits ahead of dev. It should also require that you use a force push, so be careful and specific `git push -f <your-repo-remote> tests_es6`.

staff0rd:
Yeah, was my first rebase, my bad ;)  Actually figured out properly in #3184.  With this one I didn't force the push (as I see you've mentioned also), so I thought I had to pull first, which resulted in the extra merges.  It seems doing it again (properly this time) has magically fixed it however ;)

Thanks for your help!

englercj:
Nice, next approval will likely merge as well (provided they have no comments).

test/core/Bounds.js
```diff
@@ -1,19 +1,17 @@
 'use strict';
 
-describe('getBounds', function ()
+describe('getBounds', () =>
 {
-    it('should register correct width/height with a LOADED Sprite', function ()
+    it('should register correct width/height with a LOADED Sprite', () =>
```
bigtimebuddy: By converting all the tests to use fat arrows, we lose the Mocha context for doing things like `this.timeout(3000)`, `this.slow(500)` as well as passing object and functions between tests. While I like the `var` converting to `const`, I think the test callbacks should remain as functions.

test/core/Bounds.js
```diff
@@ -88,17 +83,15 @@ describe('getBounds', function ()
...
     });
 
-    it('should register correct width/height with an empty Container', function ()
+    it('should register correct width/height with an empty Container', () =>
     {
-        var parent = new PIXI.Container();
+        const parent = new PIXI.Container();
 
-        var container = new PIXI.Container();// Graphics().beginFill(0xFF0000).drawCircle(0, 0, 10, 10);//texture);
+        const container = new PIXI.Container();// Graphics().beginFill(0xFF0000).drawCircle(0, 0, 10, 10);//texture);
```
bigtimebuddy: Are the inline comments here really needed? Can we just remove them (as well as all the subsequence copy/pasted version of this)?

staff0rd:
@bigtimebuddy updated

PR #3187, opened 2016-10-24
add missing default exports.

ericvandever:


src/extras/cacheAsBitmap.js
```diff
@@ -13,7 +13,7 @@ DisplayObject.prototype._cacheData = false;
  * @class
  * @ignore
  */
-class CacheData
+export default class CacheData
```
englercj: Does this need to be exported? I didn't export it because it seemed to only be used in this file.

src/extras/index.js
```diff
@@ -8,5 +8,5 @@ export { default as TilingSpriteRenderer } from './webgl/TilingSpriteRenderer';
 export { default as BitmapText } from './BitmapText';
 
 export { default as cacheAsBitmap } from './cacheAsBitmap';
-export { default as getChildByName } from './getChildByName';
-export { default as getGlobalPosition } from './getGlobalPosition';
+import './getChildByName';
+import './getGlobalPosition';
```
englercj: Can you add a comment that these just extend the prototype and don't have any exports?

src/extras/cacheAsBitmap.js
ericvandever: It's referenced in the extras/index.js file, but I don't see any other references, so that reference could be removed and the import switched to a side-effect only import like the other two.

Issue #3183, opened 2016-10-24
npm Install pixi.js - missing bin folder


ddennis:
First and foremost - Thank you for a fantastic library.

Doing **npm install** does not include the **bin** folder. Tested with version 4.1.0.
which makes my require and imports fails - **cannot find module pixi.js**

The package.json also seems to point to the bin folder:

`"main": "./bin/pixi.min.js",`

I might be wrong, but i would expect all files to be included. If i need to install directly from the release branch, the documentation needs to be updated. 

`npm install https://github.com/pixijs/pixi.js.git#release -D`

Thanks
/Dennis

bigtimebuddy:
Hey @ddennis. 

Unfortunately, we no longer include the *bin* folder in our dev or master branches. It was causing issues whenever people trying to make PRs and is generally not a good version control practice. The *bin* folder is available on the release branch, as you've found out. Alternatively, you can also include directly from our download links, e.g. http://pixijs.download/v4.1.0/pixi.js or build pixi.js locally `npm install && npm run build`. 

We'd definitely welcome a PR for updating documentation that mentions this. Thanks.

ddennis:
I will try to do a PR.

I have tried to figure out how it would be possible to make `npm install pixi.js` pull from the release branch and not master. But it seems you manually have to npm publish from the release folder 
Maybe someone else knows how to fix this.

englercj:
It should be included because when running the publish commands it also builds everything then publishes:

https://github.com/pixijs/pixi.js/blob/master/package.json#L38

Since `bin/` is in the `files` array, it should already include the bin folder when we publish. If this is not happening, something strange is going on...

RanzQ:
There should be an empty `.npmignore` file, otherwise it uses the `.gitignore`, see here: https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

I guess that `.gitignore` overrides whatever you have in `files`.

englercj:
> I guess that .gitignore overrides whatever you have in files.

Well that is super dumb, I was under the impression if we had a `files` array specified in package.json that was the entire whitelist and nothing else was considered...

I'll touch an .npmignore file so we can fix this.

englercj:
Try it now, just released v4.1.1 that has a blank npmignore.

RanzQ:
@englercj Now the `bin/` is included. :)

PR #3193, opened 2016-10-25
Updated docs: #3183

ddennis:
npm install should be done from the release branch, since package.json points to the bin folder, which is not included on master.

This is my first PR, i have read the Contributing Guide but did not find anything regarding commit messages guide lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ Alternatively, Pixi.js can be installed with [Bower](https://bower.io/#getting-s
 $> bower install pixi.js
 ```
 
-#### NPM Install
-
+#### npm Install
+Pixi.js is currently not released to npm. You will have to install it directly from the github release branch, to get lastest version.
 ```
-$> npm install pixi.js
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@ddennis This is not accurate. **pixi.js** module *is* published on NPM. However, if you want to install directly from Github, you can only use the *release* branch.

ddennis:
Sorry - in my search I only found the [pixi](https://www.npmjs.com/package/pixil) - just found [pixi.js](https://www.npmjs.com/package/pixi) - sorry for wasting your time. 

But if you already are publishing to npm. The correct solution would be to publish to npm from the release branch, since this includes everything (bin and src). Which would eliminate the need to change the docs. 

If i am totally wrong here - you can just close the thread, i don't want to waste your time.

bigtimebuddy:
I think this is working exactly how we want. Our publishing process publish builds the library, publishes to npm and then pushes to the release branch (to support Bower). Thanks for looking into this though.

Issue #2662, opened 2016-06-20
touchendoutside events only gets dispatched for the first release.


karmacon:
In a multitouch application, when you touch an object with two fingers, and then drag off and release both fingers, the 'touchendoutside' event is only dispatched once.

http://stackoverflow.com/questions/37882401/pixijs-not-all-touchend-events-triggering-outside-of-displayobject?noredirect=1#comment63288886_37882401

englercj:
Looks like you made a commit that fixed it for you. Can you retest this with the latest dev build and see if that has fixed your issue?

You can see how to download the latest build [in the FAQs](https://github.com/pixijs/pixi.js/wiki/FAQs#where-can-i-get-a-build).

If it hasn't would you mind sending your fix as a PR?

Thanks!

starakaj:
Had this same issue, can confirm that this change fixes it. I tried making a pull request as well, thanks!

andrewstart:
Now that #3546 has been merged, this should be all set.

PR #3198, opened 2016-10-25
Fix issue #2662

starakaj:
Added a touch counter instead of flag to ensure the correct touch count


themoonrat:
You'll want to update interactiveTarget.js with the new property

PR #3200, opened 2016-10-25
	Fix issue #2662

starakaj:
Added a touch counter instead of flag to ensure the correct touch count


src/interaction/InteractionManager.js
```diff
@@ -1327,7 +1327,11 @@ export default class InteractionManager extends EventEmitter
     {
         if (hit)
         {
-            displayObject._touchDown = true;
+            if (typeof displayObject._touchCount === 'undefined')
```
GoodBoyDigital: Do we need this check? Other than that it looks good to me!

englercj: Yeha looks like this `if` block should just be removed, mixing interaction target should set this to 0.

starakaj:
Cool, I see that you are correct --- the mixin should make this check unnecessary. Made a new commit, should be good now. Thanks!

englercj:
I think your branch is out of date, pulling in a lot of other stuff.

PR #3201, opened 2016-10-25
Dev texture states cacheing

GoodBoyDigital:
Evening All! So this turned out to be more complicated than I thought - but finally here we are :)

The idea here is that we basically cache textures when they are bound - so if a texture is at a location already it is not bound again saving on GPU calls.

Taking the a step further I modified the sprite batch algorithm to try to reuse textures that are already currently bound.

The results mean that un less you have lots of textures in your scene we get a lot less GPU calls binding textures and locations (often none!)

This PR is not 100% complete (I have a small bit to do with filters left), but I wanted peeps to get eyes on for feedback!

Cheers!


GoodBoyDigital:
This is good to go now from my perspective! Combined with the vao pr it has brought our total number of GL calls down from 80s - 40s!

ivanpopelyshev:
Again, please look at ALL places with _activeTextureLocation. Search it through all the sources. We had something in filters too

GoodBoyDigital:
Did a search and could not see any more? Mind pointing me in the right direction? ta!

ivanpopelyshev:
So, there's no _activeTextureLocation anymore?

GoodBoyDigital:
nope!

src/core/renderers/webgl/TextureManager.js
```diff
@@ -61,11 +61,19 @@ export default class TextureManager
      * Updates and/or Creates a WebGL texture for the renderer's context.
      *
      * @param {PIXI.BaseTexture|PIXI.Texture} texture - the texture to update
+     * @param {uint} location - the location the texture will be bound to.
```
englercj: `number` instead of `uint`

src/core/renderers/webgl/TextureManager.js
```diff
@@ -61,11 +61,19 @@ export default class TextureManager
...
      * @return {GLTexture} The gl texture.
      */
-    updateTexture(texture)
+    updateTexture(texture, location)
     {
-        texture = texture.baseTexture || texture;
+        // TODO - add default to location?
+        // probably ok!
```
englercj: looks like you have a default location below (`0`), is this comment relevant?

src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -435,33 +452,65 @@ export default class WebGLRenderer extends SystemRenderer
      * @param {number} location - the texture location
      * @return {PIXI.WebGLRenderer} Returns itself.
      */
-    bindTexture(texture, location = 0)
+    bindTexture(texture, location)
     {
+        location = location || 0;
```
englercj: Why move this default out of the parameter?

src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -354,21 +377,15 @@ export default class WebGLRenderer extends SystemRenderer
...
-            {
-                // the texture needs to be unbound if its being rendererd too..
-                this._activeTextureLocation = baseTexture._id;
-                gl.activeTexture(gl.TEXTURE0 + baseTexture._id);
-                gl.bindTexture(gl.TEXTURE_2D, null);
+                // bind the current texture
+                this.textureManager.updateTexture(baseTexture, 0);
             }
 
+            this.unbindTexture(baseTexture);
```
englercj: This looks weird, you are binding the texture then immediately unbinding it?

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -108,16 +109,13 @@ export default class SpriteRenderer extends ObjectRenderer
...
-        this.shaders = new Array(this.MAX_TEXTURES);
-        this.shaders[0] = generateMultiTextureShader(gl, 1);
-        this.shaders[1] = generateMultiTextureShader(gl, 2);
+        const shader = this.shader = generateMultiTextureShader(gl, this.MAX_TEXTURES);
 
         // create a couple of buffers
         this.indexBuffer = glCore.GLBuffer.createIndexBuffer(gl, this.indices, gl.STATIC_DRAW);
 
         // we use the second shader as the first one depending on your browser may omit aTextureId
         // as it is not used by the shader so is optimized out.
```
englercj: comment relevant anymore?

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -138,6 +136,8 @@ export default class SpriteRenderer extends ObjectRenderer
 
         this.vao = this.vaos[0];
         this.currentBlendMode = 99999;
+
+        this.boundTextures = new Array(this.MAX_TEXTURES);
```
englercj: Initialize this in the constructor so that it exists at construction time.

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -108,16 +109,13 @@ export default class SpriteRenderer extends ObjectRenderer
         // step 2: check the maximum number of if statements the shader can have too..
         this.MAX_TEXTURES = checkMaxIfStatmentsInShader(this.MAX_TEXTURES, gl);
 
-        this.shaders = new Array(this.MAX_TEXTURES);
-        this.shaders[0] = generateMultiTextureShader(gl, 1);
-        this.shaders[1] = generateMultiTextureShader(gl, 2);
+        const shader = this.shader = generateMultiTextureShader(gl, this.MAX_TEXTURES);
```
englercj: Doing this means we will always use maximum branching. Is that desired?

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -317,20 +349,23 @@ export default class SpriteRenderer extends ObjectRenderer
...
         }
 
         currentGroup.size = i - currentGroup.start;
 
-        this.vertexCount++;
+        // keep an eye out here!
+        // TODO - check this out for mobile!
 
+        /*
         if (this.vaoMax <= this.vertexCount)
```
englercj: Why are we creating/storing the multiple VAOs if they are no longer used? Looks like only 1 is used now (since we only have 1 shader) so can we remove the loops creating these and change it to a single VAO like you did with shader?

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -341,28 +376,30 @@ export default class SpriteRenderer extends ObjectRenderer
...
 
-        this.vertexBuffers[this.vertexCount].upload(buffer.vertices, 0);
-        this.vao = this.vaos[this.vertexCount].bind();
+        // this.vertexCount++;
+        // set textures..
+        */
+
+        this.vertexBuffers[this.vertexCount].upload(buffer.vertices, 0, false);
+
+        for (i = 0; i < MAX_TEXTURES; i++)
```
englercj: `++i` *wink*

src/core/renderers/webgl/TextureManager.js
GoodBoyDigital: good catch!

src/core/renderers/webgl/WebGLRenderer.js
GoodBoyDigital: it does! we need to make sure the render texture itself is not bound as at this point we are rendering to it. WebGL throws lots of errors if you are rendering to a frame buffer whilst also having its texture bound. 
unbindTextutre makes sure that it happens.

src/core/renderers/webgl/WebGLRenderer.js
GoodBoyDigital: Its faster than the babel output. This is a real hot piece of code and figured every little helps :)

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: In all my tests, this does not seem to make a difference (which is great!) The main reason I set up the shaders like this was to mitigate a webGL error that was introduced about half a year ago where it complained about using textures that are not bound. Fortunately this restructure solves it in a different way.

For some mobile devices we still set max textures to 4 as even though they claim to be able to have more that, performance begins to drop because of the branching. 

Correct me if I'm wrong here too. But I believe that the style of these if statements work pretty fast as they is good Cache coherence.

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: nope! will delete that 👍

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: Unfortunatly we cant, we only know MAX_TEXURES once we have create the context as we need to read that value from it.

I could add the variable though ```this.boundTextures = null```?

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: Good catch, will look into this. I think it may still be required for ios. Either way I need to tweak a few bits..

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: lol, old habits init :)

src/core/sprites/webgl/SpriteRenderer.js
englercj: As long as the branching is consistent per pixel it should be fine. Its when you actually do get up to using multiple textures (4, 5, 6, ...) that the branching becomes a concern because it will be very different and the parallelism of the GPU will be shot.

> where it complained about using textures that are not bound

Is that why you stuff dummy textures into the state, so that when this shader runs there are always texture sin each slot even if we don't use them?

src/core/sprites/webgl/SpriteRenderer.js
englercj: Yeah the property just needs to exist at construction time. Without doing `this.boundTextures = null` or something, it will perform a Hidden Class transition here, and we want to avoid that.

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: yup! the dummy textures mean that we always use the maximum sized shader. This is essential as it may be that a texture ends up being stored in say slot 1 and 6.

src/core/renderers/webgl/TextureManager.js
GoodBoyDigital: fixed 👍

src/core/renderers/webgl/TextureManager.js
GoodBoyDigital: Fixed 👍

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: Fixed 👍

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: Fixed 👍

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: Fixed 👍

src/core/sprites/webgl/SpriteRenderer.js
GoodBoyDigital: Sorted -kept them in. Still faster on IOS

Issue #3180, opened 2016-10-23
MovieClip.gotoAndPlay skips first frame when called for the second time.


Tade0:
Hello everyone, first of all thank you for creating such a magnificent library.

Second thing: I haven't found any similar reported issue, so please take a look at this jsFiddle:
https://jsfiddle.net/zxz0se93/6/

**Expected behaviour**
`gotoAndPlay` always starts with the frame number provided in the parameter.

**Actual behaviour**
When `gotoAndPlay` is called for the second time it stays one frame duration at `4` and then starts with `1`.

Tested and reproduced in Chrome and Firefox.

staff0rd:
Fixed in master; https://jsfiddle.net/szdb6gjc/

PR #3203, opened 2016-10-26
Update MovieClip frame immediately with gotoAndPlay #3180

hanenbro:
Employs the same approach for updating the current frame and invoking the `onFrameChange` callback as `gotoAndStop()`

### DIFF
--- a/src/extras/MovieClip.js
+++ b/src/extras/MovieClip.js
@@ -143,13 +143,7 @@ export default class MovieClip extends core.Sprite
 
         if (previousFrame !== this.currentFrame)
         {
-            this._texture = this._textures[this.currentFrame];
-            this._textureID = -1;
-
-            if (this.onFrameChange)
-            {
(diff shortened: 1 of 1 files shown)


themoonrat:
Looks good to me. 
If I could change anything, it would be to turn part of that into a function

```
if (previousFrame !== this.currentFrame)
 {
   this._texture = this._textures[this.currentFrame];
   this._textureID = -1;

   if (this.onFrameChange)
   {
     this.onFrameChange(this.currentFrame);
   }
}
```
That part of code is called 3 times now.
Would be nicer if it read something like
```
if (previousFrame !== this.currentFrame)
{
  changeFrame();
}
```
On each occasion. That way if we need to update the changing frame behaviour, we're not fixing it in 3 places!

hanenbro:
Thanks @themoonrat - good suggestion! Update committed.

GoodBoyDigital:
nice work fellas 👍

PR #3207, opened 2016-10-26
Allow extension-less image urls

staff0rd:
The adding of SVG support added a requirement of `imageType`, which is determined by the file extension, and was throwing were no extension available. There doesn't seem to be a way to determine an `HTMLImageElement`'s image type, so this PR just allows non-svg images to be extension-less as per prior functionality.

The caveat is that svg image types cannot be loaded without an extension.  A couple of options to enable it would be;

1) Extend `*.fromImage()` to allow a user to specify an `imageType` override (easy)
2) xhr images as byte arrays instead of Image, determine type from data, convert to image (hard)

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -412,7 +412,7 @@ export default class BaseTexture extends EventEmitter
 
             if (!imageType)
             {
-                throw new Error('Invalid image type in URL.');
+                imageType = 'png';
             }
         }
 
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
Can you also look at https://github.com/pixijs/pixi-compressed-textures ? Its using that kind of feature

englercj:
> also wanted to deprecate fromImage, so, do we really have to change fromImage() code?

While I do, I don't think we ever will. Though, what I might do is make them use the loader or something.

englercj:
I'm going to merge this for now, but would you mind doing option #1 @staff0rd ?

> Extend *.fromImage() to allow a user to specify an imageType override (easy)

PR #3210, opened 2016-10-27
Update Rectangle.js small typo in api doc

markknol:
Rectanle should be Rectangle


bigtimebuddy:
@markknol good catch! Could you make this PR against Dev instead of master? Thanks!

markknol:
see https://github.com/pixijs/pixi.js/pull/3214

PR #3214, opened 2016-10-28
Update Rectangle.js small typo in api doc

markknol:
Rectanle should be Rectangle


ivanpopelyshev:
I dont think we need people to review that :)

PR #3221, opened 2016-10-30
Fixes issue where rope breaks if it changes size

GoodBoyDigital:
- Recreate Types arrays in rope if the size changes
- Fixed a little rope dirty issue too where it was being set to false
  rather than incremented.
- Pass the data to buffer when calling upload as this will resize the
  buffers automatically

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -103,7 +103,7 @@
         "no-unused-labels": 1,
         "no-useless-call": 1,
         "no-useless-concat": 1,
-        "no-useless-escape": 1,
+        "no-useless-escape": 0,
         "no-void": 1,
         "no-warning-comments": 0,
         "no-with": 1,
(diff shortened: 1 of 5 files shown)


.eslintrc.json
```diff
@@ -103,7 +103,7 @@
         "no-unused-labels": 1,
         "no-useless-call": 1,
         "no-useless-concat": 1,
-        "no-useless-escape": 1,
+        "no-useless-escape": 0,
```
englercj: Why was this changed?

englercj: Maybe, doesn't look like this affects any changes here; so not sure what it is referring to. Might have been left over from when you had changed things in deprecation file?

englercj: If we can revert this one line I'm good to merge this.

GoodBoyDigital: slick - will do that later tonight

englercj:
Merging this, I'll manually revert the eslint change.

PR #3233, opened 2016-10-31
Fixed interaction issues with pointer events

themoonrat:
* Fixed issue where browsers that supported pointer events would no longer fire mouse events.
* Changed ordering of listeners, so that they are normalized events are sent in the same order as non-normalized events. pointer event first, mouse/touch event 2nd

https://github.com/pixijs/pixi.js/pull/2997
https://github.com/pixijs/pixi.js/issues/3167
https://github.com/pixijs/pixi.js/issues/3158


ivanpopelyshev:
Is #3200 related?

themoonrat:
I don't believe so, no

PR #3196, opened 2016-10-25
Fix polygon bounds

staff0rd:
Fixes #3168 

Before;
https://codepen.io/staff0rd/pen/rrRdpj?editors=0010
image

After;
https://codepen.io/staff0rd/pen/ZpPxvK?editors=0010
image

PR #3234, opened 2016-10-31
Fix polygon bounds

ivanpopelyshev:
Custom version of #3196 without trigonometric functions

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -974,8 +974,8 @@ export default class Graphics extends Container
                             theta = (x2 - x) / (y2 - y);
                         }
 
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
OK, your tests helped me, thank you.

Basically,

```js
sinTheta = (y2 - y) / w;
cosTheta = (x2 - x) / w;
```

atan2 is dividing stuff on "w" then calculating the angle by its sin and cos. But whats the point if we already have them? :)

ivanpopelyshev:
Also, in that commit I found one interesting thing: now we actually have convention about order of operators, i cant just write 

```js
a * b + c * d
```

staff0rd:
Confirmed; https://codepen.io/staff0rd/pen/qaeZjv?editors=0010
image


Thanks @ivanpopelyshev

PR #3239, opened 2016-11-01
Update package.json main prop

LucianoGraziani:
It would be nice if, while developing, it imports the source code, instead the minified one. For example, this code:

```js
import { Container } from 'pixi.js';
```
Using atom and js-hyperclick, you can Ctrl+click `pixi.js` and go into the code. Right now, it opens the minified. With this change, it opens the `src/index.js`.

But the main problem will occur when something goes wrong. Error stacks will be easier to read and trace when using the source code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pixi.js",
-  "version": "4.1.1",
+  "version": "4.1.0",
   "description": "Pixi.js is a fast lightweight 2D library that works across all devices.",
   "author": "Mat Groves",
   "contributors": [
@@ -9,7 +9,7 @@
     "Chad Engler <chad@pantherdev.com>",
     "Richard Davey <rdavey@gmail.com>"
   ],
-  "main": "./bin/pixi.min.js",
+  "main": "./src/index.js",
   "homepage": "http://goodboydigital.com/",
   "bugs": "https://github.com/pixijs/pixi.js/issues",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pixi.js",
-  "version": "4.1.2",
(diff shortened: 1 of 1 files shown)


englercj:
Definitely, unfortunately there is some context behind why it is the way it is right now.

https://github.com/pixijs/pixi.js/pull/2981

There are issues with webpack when pointing to source. Hopefully we can fix those up if/when we switch back to pointing at source for `main`.

P.S. Please do not modify the `version` field in PRs.

bigtimebuddy:
Why not `import { Container } from 'pixi.js/src'`?

LucianoGraziani:
Hmm... it could be. There is a reason to point to the minified version?
If there is a reason, it would be good to say that when using ES6, you have to import from `pixi.js/src`.

LucianoGraziani:
@englercj thank you for pointing the version modification. When you say "some context", you're talking about this errors?

```
WARNING in ./~/pixi.js/src/deprecation.js
56:0 export 'SpriteBatch' (imported as 'core') was not found in './core'

WARNING in ./~/pixi.js/src/deprecation.js
70:0 export 'AssetLoader' (imported as 'core') was not found in './core'

WARNING in ./~/pixi.js/src/deprecation.js
573:19 export 'AbstractFilter' (imported as 'core') was not found in './core'
```
Or this one?

```
ERROR in ./~/pixi.js/src/index.js
Module build failed: SyntaxError: 'import' and 'export' may appear only with 'sourceType: module' (2:0)
```

englercj:
Not sure there is a particular reason we point to the minified over unminified version. @GoodBoyDigital has more experience trying to get pixi to work with webpack than I do though.

englercj:
Closing this in favor of #3289

PR #3248, opened 2016-11-03
Fix the Polygon contains method

twhitbeck:
We couldn't click on things. This change (to align with the mentioned algorithm from substack) fixed it for us.


englercj:
Good catch, I think this deserves a hotfix (master merge).

themoonrat:
FYI, this change is already in the dev branch via: https://github.com/pixijs/pixi.js/commit/93714a9dd46c353a4a51b00c6b287d3bae24d392#diff-fddecedb5f4e12741286f6a3c3211feb

englercj:
Ha, oops. Forgot we hit this already. My bad! Good catch @themoonrat.

Issue #2824, opened 2016-08-17
arc called after lineTo makes the line disappear


afarber:
This code works well with PIXI 3.0.10 and draws a line and an arc:

```
this.bgGraphics.lineStyle(4, 0xFF00FF, 1);
this.bgGraphics.moveTo(15, 0);
this.bgGraphics.lineTo(60 - 14, 0);
//this.bgGraphics.lineTo(60 - 15, 0);
this.bgGraphics.arc(45, 15, 15, -Math.PI / 2, 0);
```

![screen shot 2016-08-17 at 17 10 52](https://cloud.githubusercontent.com/assets/1884458/17742064/d7ce294a-649e-11e6-9721-a2a5aca54f6e.png)

But change just 1 pixel and the horizontal line is suddenly not displayed:

```
this.bgGraphics.lineStyle(4, 0xFF00FF, 1);
this.bgGraphics.moveTo(15, 0);
//this.bgGraphics.lineTo(60 - 14, 0);
this.bgGraphics.lineTo(60 - 15, 0);
this.bgGraphics.arc(45, 15, 15, -Math.PI / 2, 0);
```

![screen shot 2016-08-17 at 17 11 14](https://cloud.githubusercontent.com/assets/1884458/17742081/e426bffe-649e-11e6-90f7-fe9946825ca9.png)

ivanpopelyshev:
Something wrong happens when point appears two times in current shape. In this base its (45,0). I think you can also omit "lineTo(64-14,0)"

Also, arcTo is not working properly. We need to fix both issues before v4 release.

PR #3249, opened 2016-11-03
Fix anticlockwise sweep calculation

twhitbeck:
This changed from v3 to v4, we reverted back and it fixed arcs in our application.

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -480,7 +480,7 @@ export default class Graphics extends Container
             startAngle += Math.PI * 2;
         }
 
-        const sweep = anticlockwise ? (startAngle - endAngle) : (endAngle - startAngle);
+        const sweep = anticlockwise ? (startAngle - endAngle) * -1 : (endAngle - startAngle);
         const segs = Math.ceil(Math.abs(sweep) / (Math.PI * 2)) * 40;
 
         if (sweep === 0)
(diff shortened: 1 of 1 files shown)


englercj:
Did you check the blame on this? This was an intentional change, what issue are you having?

https://github.com/pixijs/pixi.js/commit/1bcc6b4c2f51566bc438d7966d33f21f3a5aee21
https://github.com/pixijs/pixi.js/issues/2824

twhitbeck:
The issue we were having was the arcs being drawn in mirror position to what we were expecting.

twhitbeck:
Interestingly enough, we were also experiencing the issue described in #2824, but it has been resolved (even with this revert back to `* - 1`)

twhitbeck:
After upgrading to v4, we saw this:

current

with this change, it cleared up and went back to:

reverted

englercj:
So unless I am mistaken, this revert means that clockwise and anti-clockwise are actually the same because `(startAngle - endAngle) * -1 === (endAngle - startAngle)`. So something is seriously messed up.

englercj:
Oops didn't mean to close.

GoodBoyDigital:
good catch @twhitbeck! I guess we could simplify this line to basically be:

```const sweep = endAngle - startAngle;```

twhitbeck:
@GoodBoyDigital ok I simplified the calculation.

GoodBoyDigital:
Nice work team! 👍

PR #3258, opened 2016-11-06
Dev texture states

GoodBoyDigital:
Fixed a little issue where the same texture could end up being bound in
different locations.

Added a smartBindTexture function that finds a the best texture slot
(its pretty basic at this stage!)

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -168,7 +168,8 @@ export default class WebGLRenderer extends SystemRenderer
          * @member {PIXI.RenderTarget}
          */
         this._activeRenderTarget = null;
-        this._activeTexture = null;
+
+        this._nextTextureLocation = 0;
 
         this.setBlendMode(0);
     }
@@ -249,6 +250,8 @@ export default class WebGLRenderer extends SystemRenderer
         {
             return;
         }
+
+        this._nextTextureLocation = 0;
 
         if (!renderTexture)
         {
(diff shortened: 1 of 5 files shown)


ivanpopelyshev:
So instead of texture bind, we'll have uniform1i call, right?

GoodBoyDigital:
Yup! As I understand it uniform setting is cheapest (let me know if i'm way off the mark here!). Plus once we get a uniform caching layer in, might be that there are no GPU calls at all!

ivanpopelyshev:
That layer is not caching. It does not compare with previous value.

GoodBoyDigital:
Thats true. The caching layer is not in yet, something for a future update 👍

englercj:
Why are `smartBindTexture` and `bindTexture` separate? Couldn't they just be the same function?

ivanpopelyshev:
@englercj because it has different behaviour, which requires more brains to make a renderer plugin. Assume that I want to experiment with something fast, do I really have to think about that optimization or multitexturing while prototyping?

englercj:
Its only different behavior if you don't specify the location or the texture is already bound. I'm wondering what is the case for having these separate and not the same behind a param.

ivanpopelyshev:
If it has bugs, only smartBindTexture calls will be affected. We can force is as default behaviour when we'll be sure its safe

englercj:
> If it has bugs, only smartBindTexture calls will be affected. We can force is as default behaviour when we'll be sure its safe

I'm not sure what you mean by that, but "it may not work" is a bad reason to build a confusing API.

GoodBoyDigital:
Howdy! Sorry for delay on this! Been moving house.. fun.

Gonna look into this now...

As a heads up - the reason for smart bind logic is that it protects the textures states so that a texture is never bound twice. This assumption means less checks are required by the sprite renderer in the core loop. As an added bonus it will also result in less GPU texture bind and location calls too.

But yeah either way, we need something there as currently if you bind the same texture to different locations, you get some really interesting visual bugs :)

GoodBoyDigital:
should be all good now! Although @englercj not sure what an interactive rebase is?

englercj:
I'll do the rebase, but just FYI `git rebase -i` is an interactive rebase.

GoodBoyDigital:
cheers dude!

englercj:
Rebased to a single commit, please read through and ensure I didn't mess anything up in the merging. But if it looks good, then you're good to merge!

GoodBoyDigital:
looks good! thanks all 👍

PR #3259, opened 2016-11-07
More flexible preparation limits

andrewstart:
* Combines the shared code of WebGLPrepare and CanvasPrepare into one BasePrepare class.
* Makes the rate limiting functionality more configurable by changing it to be class based and adding a class for limiting based on time spent preparing.
* Adds hooks for both WebGL and Canvas for preparing PIXI.Text and PIXI.TextStyle objects.
* Adds unit tests for the rate limiting and basic functionality of BasePrepare.

I did some testing of GPU upload times on a mobile device and found that texture upload time was quite variable and generally expensive, so I wanted to be able to limit preparation to a specified amount of time to attempt to preserve the framerate instead of loading as quickly as possible.
I also found that calculating the font properties was generally worse than uploading any single texture to the GPU, so added the hooks to prepare that stuff ahead of time using the same system.

### DIFF
--- a/src/deprecation.js
+++ b/src/deprecation.js
@@ -455,12 +455,16 @@ core.Text.prototype.setStyle = function setStyle(style)
  * @method
  * @name PIXI.Text#determineFontProperties
  * @see PIXI.Text#calculateFontProperties
+ * @deprecated since version 4.2.0
  * @private
  * @param {string} fontStyle - String representing the style of the font
  * @return {Object} Font properties object
  */
 core.Text.prototype.determineFontProperties = function determineFontProperties(fontStyle)
 {
+    warn('determineFontProperties is now deprecated, please use the static calculateFontProperties method, '
+        + 'e.g : Text.calculateFontProperties(fontStyle);');
+
     return Text.calculateFontProperties(fontStyle);
 };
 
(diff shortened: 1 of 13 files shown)


ivanpopelyshev:
I think we also need PIXELS per frame. 2048x2048 and that kind off stuff

andrewstart:
That's a little more complex than I wanted to get into. It also can't be a primary limiting metric, as it is only valid for textures.

ivanpopelyshev:
@themoonrat there's something about fonts, can you check it?

src/deprecation.js
```diff
@@ -642,3 +656,49 @@ Object.defineProperty(core.utils, '_saidHello', {
...
+    set()
+    {
+        warn('PIXI.CanvasPrepare.UPLOADS_PER_FRAME has been removed. Please set '
+            + 'renderer.plugins.prepare.limiter.maxItemsPerFrame on your renderer');
+        // because we don't have a reference to the renderer, we can't actually set
+        // the uploads per frame, so we'll have to stick with the warning.
+    },
+    get()
+    {
+        return 4;
```
englercj: Same as above

andrewstart:
Deprecations have been fixed.

PR #3270, opened 2016-11-10
sfinktah - use addChild to bringToFont without overhead [time 08:25]

sfinktah:
Well, there was no 'next' branch, so it's ending up here.

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -59,20 +59,40 @@ export default class Container extends DisplayObject
             // if the child has a parent then lets remove it as Pixi objects can only exist in one place
             if (child.parent)
             {
-                child.parent.removeChild(child);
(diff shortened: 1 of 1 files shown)


sfinktah:
pixi-what-now?  i'll check it out.  i did read a discussion about the proposed introduction of a z-index, and i figured if manipulating the order of children was going to become a more frequently performed operation, that there ought to be some specific code to optimize such things

bigtimebuddy:
@sfinktah Thanks for the PR and I appreciate all the inline comments. Two requests: 1). Could you resolve conflicts with dev? and 2). Could you make sure that these cases are covered by the unit tests? (if not please add a tests or two).

bigtimebuddy:
@sfinktah I'm closing this PR because of inactivity. If you would like it to be reconsidered, please feel free to re-open and respond to my comments above. Thanks.

PR #3275, opened 2016-11-11
Bumped version

vanillajonathan:


bigtimebuddy:
Are we sure we want this to go to master, @englercj?

englercj:
> Are we sure we want this to go to master, @englercj?

Don't think it matters, `master` is the default branch so it is what people see first. Additionally, this wouldn't cause a release so I think it's fine to merge to master, then merge down to `dev`.

PR #3280, opened 2016-11-13
SpriteRenderer.destroy fix

staff0rd:
SpriteRenderer.destroy() was failing - looks related to #3201

### DIFF
--- a/src/core/sprites/webgl/SpriteRenderer.js
+++ b/src/core/sprites/webgl/SpriteRenderer.js
@@ -452,13 +452,7 @@ export default class SpriteRenderer extends ObjectRenderer
 
         super.destroy();
 
-        for (let i = 0; i < this.shaders.length; i++)
-        {
-            if (this.shaders[i])
-            {
-                this.shaders[i].destroy();
-            }
-        }
+        this.shader.destroy();
 
         this.vertexBuffers = null;
         this.vaos = null;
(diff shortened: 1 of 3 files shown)


staff0rd:
Hmm, no webgl on travis CI.  Thoughts?

englercj:
You could try just passing in a canvas renderer? Since you're not actually doing anything with it but destroying it.

staff0rd:
mocked

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -452,13 +452,7 @@ export default class SpriteRenderer extends ObjectRenderer
...
         super.destroy();
 
-        for (let i = 0; i < this.shaders.length; i++)
-        {
-            if (this.shaders[i])
-            {
-                this.shaders[i].destroy();
-            }
-        }
+        this.shader.destroy();
```
englercj: check for existence, construction and immediate destruction without context creation would make this error out.

bigtimebuddy: Also, make sure to nullify. `this.shader = null`

staff0rd:
I don't know how to get rid of the merge at the end - does it matter?

englercj:
Nah, I'll squash it out when I merge.

src/core/sprites/webgl/SpriteRenderer.js
staff0rd: @englercj construction and immediate destruction without context creation was broken anyway - fixed with cc58b0e

PR #3285, opened 2016-11-14
Add jsdocs for renderer.extract property

giraluna:
The position of the doc block is a bit awkward, but I'm not sure there's really a better place for it.

Declaring it where the plugin is registered [(here)](https://github.com/pixijs/pixi.js/blob/master/src/extract/webgl/WebGLExtract.js#L223) would clarify why the property is being set for the renderer, but obscure what's being set in the first place.


bigtimebuddy:
Looks good. @englercj good to merge.

PR #3286, opened 2016-11-14
Spritesheet Scale

bigtimebuddy:
### Added

* Adds support for using the `meta.scale` property of spritesheet format to determine the resolution rather than relying on the filename (e.g., `@2x.png`), for conditions when renaming the file might not be appropriate. 

_Note: this is also needed to support an optimization feature in PixiAnimate_

### DIFF
--- a/src/loaders/spritesheetParser.js
+++ b/src/loaders/spritesheetParser.js
@@ -42,7 +42,17 @@ export default function ()
 
             const frames = resource.data.frames;
             const frameKeys = Object.keys(frames);
-            const resolution = core.utils.getResolutionOfUrl(resource.url);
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
Hey chaps! this change seems to be breaking our games :/
Gonna investigate now..

ivanpopelyshev:
Instead of taking "scale" from data.meta i suggest to use our loader's "metadata" field and add resolution there.

englercj:
It should probably do both.

bigtimebuddy:
@GoodBoyDigital added a fix #3297 which addresses the problem. Scale is only considered if the resolution from the filename is not 1.

PR #3287, opened 2016-11-14
Fixes a bug on SpriteRenderer's destroy

bigtimebuddy:
### Overview

Resolves a bug when trying to destroy the  SpriteRenderer. Found when running `npm test` locally.

cc @GoodBoyDigital


staff0rd:
FYI #3280

PR #3289, opened 2016-11-14
Build System, Babel and ES6 src

bigtimebuddy:
## Overview

Removes the ES6 src directory from npm and replaces with a lib folder, which is now the main point for the package.json's `main` field. Creates bundled files in the "dist" folder and individual files in the "lib" folder. 

This should help resolve issues with devs using webpack and other build systems that are more friendly with unbundled modules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "lint": "eslint scripts src test --max-warnings 0",
     "lintfix": "npm run lint --fix",
     "prebuild": "npm run lint",
-    "build": "babel src --out-dir lib",
+    "build": "babel src --out-dir lib -s",
     "bundle": "pixify -d dist -n PIXI -o pixi",
     "docs": "jsdoc -c scripts/jsdoc.conf.json -R README.md",
     "publish:patch": "npm version patch --no-git-tag-version && npm publish",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "Chad Engler <chad@pantherdev.com>",
     "Richard Davey <rdavey@gmail.com>"
   ],
-  "main": "./bin/pixi.min.js",
+  "main": "./lib/index.js",
   "homepage": "http://goodboydigital.com/",
   "bugs": "https://github.com/pixijs/pixi.js/issues",
   "license": "MIT",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -18,31 +18,33 @@
...
-    "coverage": "npm test -- -c bin/pixi.js -s -h",
+    "precoverage": "rimraf coverage && npm run build",
+    "coverage": "npm test -- -c dist/pixi.js -s -h",
     "lint": "eslint scripts src test --max-warnings 0",
     "lintfix": "npm run lint --fix",
-    "prebuild": "npm run lint && npm run clean",
-    "build": "pixify -n PIXI -o pixi",
+    "prebuild": "npm run lint",
+    "build": "babel src --out-dir lib -s",
+    "bundle": "pixify -d dist -n PIXI -o pixi",
```
englercj: Instead of using the babelify and other transforms could we just run browserify on the ES5 files to get a bundle? Something like `browserify -e lib/index.js -o dist/pixi.js -s PIXI` and then `uglifyjs dist/pixi.js -c -o dist/pixi.min.js`.

Do we even need `pixify` anymore? Seems like it would be simpler to just use the cli for the couple things we need and just drop babelify/pixify?

package.json
```diff
@@ -52,14 +54,15 @@
     "bit-twiddle": "^1.0.2",
     "earcut": "^2.0.7",
     "eventemitter3": "^2.0.0",
-    "glslify": "^5.0.2",
     "ismobilejs": "^0.4.0",
     "object-assign": "^4.0.1",
     "pixi-gl-core": "^1.0.3",
     "resource-loader": "^1.8.0",
     "browserify-versionify": "^1.0.6"
```
englercj: We use this to replace the version string, I assume that doesn't happen in the ES5 files. How can we inject the string into those files (before browserify runs).

package.json
englercj: I wonder if we can keep sourcemaps and non re-babel now that you mention it...

bigtimebuddy: Did a test. Unfortunately, the sourcemaps maintain the lib folder source and not the original files. Build time is much faster though.

bigtimebuddy:
Replaced Browserify's version insert with one based on Babel so it work for **lib** and **dist**.

package.json
bigtimebuddy: I thought that Browserify might respect sourcemaps for content it pulls in, but seems to disregard them. Can't seem to locate any transform or plugin to mitigate this.

englercj: if you use gulp source maps you can tell it to "load sourcemaps" so that it honors what is in the file. Don't think you can do that same using browserify itself though

bigtimebuddy: How do you want to proceed? I feel like this sourcemap thing to get a faster bundle-build from the lib folder is a wishlist item we can look at later. For now it builds lib and dist from the src.

PR #3304, opened 2016-11-16
Dev geometry update

GoodBoyDigital:
A sneaky peak at the mesh restructure I have been playing around with.

Main changes to so far:

## Shaders
- Refactored Shader to be a Pixi level Object. Added a shader manager that deals with creating shaders and binding them
- Added a caching layer. BUT I think this needs looking at as underneath the hood I think they should share a GLShader.
- Filter now extends Shader

## Mesh
- Now expects 3 params (geometry, shader, draw style)

## Geometry
- Geometry is kind of an abstract VAO. It is built up of GeometryData and GeometryStyle 

## GeometryData 
- Contains the Buffers

## GeometryStyle
- Describes the structure of the geometry and how the data is used.

## Buffer
- A abstract wrapper for a GLBuffer

The above should enable us a tonne of flexibility to create custom ways to render whatever we want using a nice clean API! This can also act as a base for some of our other elements such as tilingSprite

Actions speak louder than words:
http://dev.goodboydigital.com/client/goodboy/pixi-geometry/
Make sure to view source!

## Stuff still left to Do
- Shaders information should be pulled out from a program not the src code
- Create a default geometry style (position, uvs, tint)
- Plane / Rope will need a little refactoring
- Need a nice way to deal with other gl states e.g. DepthTest (do they sit on shader?)

All feedback welcome at this stage!


ivanpopelyshev:
version in dev-geometry-update isnt working. I'm using this build with examples and it fails horribly:   https://d157l7jdn8e5sf.cloudfront.net/dev-geometry-update/pixi.js

GoodBoyDigital:
Ah yes, that build has an incompatible version of glcore. We merged it this
afternoon - you may need to build it locally for right now. Cheers!

ivanpopelyshev:
@GoodBoyDigital its not easy to test examples on local file, there's no such option :(

ivanpopelyshev:
Can you please bump it that way it recompiles with new pixi-gl-core?

ivanpopelyshev:
I checked with examples, everything is working.

ivanpopelyshev:
Someone isnt sleeping in friday night? :)

bigtimebuddy:
@GoodBoyDigital shall we close this?

GoodBoyDigital:
Yes please! We are very far from here now :)

PR #3306, opened 2016-11-17
Support for beefy by removing browserify config

bigtimebuddy:
### Fixed

* Adds support for using [**beefy**](https://www.npmjs.com/package/beefy) with `require(‘pixi.js’)`

cc @GoodBoyDigital


GoodBoyDigital:
nice one @bigtimebuddy 👍

PR #3309, opened 2016-11-17
move CAN_UPLOAD_SAME_BUFFER to settings

ivanpopelyshev:
I really dont know what to do with TODOs

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -1,5 +1,3 @@
-import canUploadSameBuffer from './utils/canUploadSameBuffer';
-
 /**
  * String of the current PIXI version.
  *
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
Demo: 

```js
var renderer = PIXI.autoDetectRenderer(800, 600,{backgroundColor : 0x1099bb});

stage.addChild(container);

for (var j = 0; j < 5; j++) {

    for (var i = 0; i < 5; i++) {
        var bunny;
        if ((i+j)%4 == 1) bunny =  PIXI.extras.TilingSprite.fromImage('required/assets/basics/bunny.png', 10, 10);
      else bunny =  PIXI.Sprite.fromImage('required/assets/basics/bunny.png');
        bunny.x = 40 * i;
        bunny.y = 40 * j;
        container.addChild(bunny);
    };
};
/*
 * All the bunnies are added to the container with the addChild method
 * when you do this, all the bunnies become children of the container, and when a container moves,
 * so do all its children.
 * This gives you a lot of flexibility and makes it easier to position elements on the screen
 */
container.x = 100;
container.y = 60;

// start animating
animate();

function animate() {

    // render the root container
    renderer.render(stage);
}

```

now enter the console, switch to examples iframe, place breakpoint in if(!CAN_UPLOAD... and do the following
```js
_const.CAN_UPLOAD_SAME_BUFFER = false
```

If you apply #3309 first, then you can just add this into example instead of using debugger magic:
```js
PIXI.settings.CAN_UPLOAD_SAME_BUFFER = false
```

ivanpopelyshev:
That "API" was added in 11th of november.

GoodBoyDigital:
Thats a good point dude!

PR #3328, opened 2016-11-20
add nativeLine method in graphics

DavidBM:
Hi everyone!

Some time ago I was trying to make a galactic game map with pixi and I had a problem. The width of the lines increase with the scale of the container. The only way to maintain the correct width in screen was to recreate the line every time. After some research I created a patch for pixiv3 that used gl.LINE to draw the 1px lines https://github.com/DavidBM/pixi-nativeGLline

Now, with the version 4 is impossible to apply this kind of patches. Then I thought, "hey! why a path? Better a pull request!"

And this is the pull request :D

This patch adds a new method to the graphics object (drawNative) that accepts only one parameter (native). If this parameter is true, the next lines will be draw using GL_LINE instead of GL_TRIANGLE_STRIP. 

This have an advantage, is faster than drawing a complete graphic mesh and the lines are scree-width-fix, but have several drawbacks. The line will be always 1px with in screen in windows (limitation of the OpenGL ES implementation in browsers), the fill and other functions may not work (I changed the way that the line is created, now are only a sequence of points without indexes). 

It make sense for (for example, as I made in my game) draw connection between solar system of always the same with independent of the scale. Is a corner use case, but it doesn't break the API. 

The point is: **All working code now will continue working**

There are other method that includes the vertex shader, but I still don't had enough time for implementing it. (See the "Expanding in a Vertex Shader" demo in https://mattdesl.svbtle.com/drawing-lines-is-hard ,  http://artgrammer.blogspot.com/2011/05/drawing-nearly-perfect-2d-line-segments.html and https://www.mapbox.com/blog/drawing-antialiased-lines/)

Maybe this pull request doesn't fit your idea of API, but in any case, consider expand the line functionality for having efficient screen-fix width lines in Pixi.js :)

If all of you agree that it can be a good functionality I can get time for adding this functionality in a way that likes everyone.

Have a good sunday! :dancer:

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -28,8 +28,9 @@ export default class Graphics extends Container
 {
     /**
      *
+     * @param {boolean} [nativeLines=false] - If true the lines will be draw using LINES instead of TRIANGLE_STRIP
      */
-    constructor()
+    constructor(nativeLines = false)
     {
         super();
 
@@ -48,6 +49,13 @@ export default class Graphics extends Container
          * @default 0
          */
         this.lineWidth = 0;
+
+        /**
+         * If true the lines will be draw using LINES instead of TRIANGLE_STRIP
+         *
+         * @member {boolean}
+         */
+        this.nativeLines = nativeLines;
 
         /**
          * The color of any lines drawn.
@@ -1026,6 +1034,7 @@ export default class Graphics extends Container
             this.fillColor,
             this.fillAlpha,
             this.filling,
+            this.nativeLines,
             shape
         );
 
--- a/src/core/graphics/GraphicsData.js
+++ b/src/core/graphics/GraphicsData.js
@@ -14,14 +14,19 @@ export default class GraphicsData
      * @param {number} fillColor - the color of the fill
      * @param {number} fillAlpha - the alpha of the fill
      * @param {boolean} fill - whether or not the shape is filled with a colour
+     * @param {boolean} nativeLines - the method for drawing lines
      * @param {PIXI.Circle|PIXI.Rectangle|PIXI.Ellipse|PIXI.Polygon} shape - The shape object to draw.
      */
-    constructor(lineWidth, lineColor, lineAlpha, fillColor, fillAlpha, fill, shape)
+    constructor(lineWidth, lineColor, lineAlpha, fillColor, fillAlpha, fill, nativeLines, shape)
     {
         /**
          * @member {number} the width of the line to draw
          */
         this.lineWidth = lineWidth;
+        /**
+         * @member {boolean} if true the liens will be draw using LINES instead of TRIANGLE_STRIP
+         */
+        this.nativeLines = nativeLines;
 
         /**
          * @member {number} the color of the line to draw
@@ -85,6 +90,7 @@ export default class GraphicsData
             this.fillColor,
             this.fillAlpha,
             this.fill,
+            this.nativeLines,
             this.shape
         );
     }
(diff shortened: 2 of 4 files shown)


ivanpopelyshev:
@DavidBM 

Actually your code has to work, except only one thing: since 4.2.2 you have to use "renderer.bindVao"

DavidBM:
@ivanpopelyshev are you speaking about the pull request or the code in my patch repo?

In the repo, the main problem is that it can't access to the function to overwrite because that function is accessed from the context and not from the PIXI global object. 

In my pull request I'm calling the "renderer.bindVao". Is just before the IF `if (graphics.drawNativeLine)` in the line 104 for selecting the draw method. https://github.com/pixijs/pixi.js/pull/3328/files#diff-b43b3774ef63e967ad5078272f216d57R104

m-ahmadi:
Hi,
My understanding of the underlying WebGL stuff is not much,
But Can I ask what is the drawbacks of adding this to Pixi? Or is there any?
To me, it seems like a feature, If someone wants it, they can use it.
I think it's useful for graphing projects.
So I ask this: Why wouldn't they use it? What is the argument for not using this?

DavidBM:
As far I know there is two drawbacks:

- The user can activate the drawNative and expect that fill functionality and more than one 1px width lines work and find that not. Is not a drawback itself, but for new people can be confusing.
- In my opinion the lines draw in graphics object must be enhanced. I mean, lines is a hard topic in graphics in general and there are many corners cases and uses. There isn't a one good general solution. For a lot of use cases a specific solution is required for having a decent performance. For that a more general API for drawing lines with several methods and techniques will be a very nice to have. The problem of adding this is that maybe in the future for adding the enhanced version of lines a break change will be necessary. Maybe is better to think now what we want and then start to add more things.

Summarizing, maybe is better to think on what is the correct solution and start to build this solution instead of adding this patch and hope that someone in the future create a better solution.

The different methods and techniques that I was saying are explained in the links in my pull request description. Some examples:
- Draw lines with native lines in webgl: The most performant for small segments, but need general anti-aliasing for being anti-aliased and that is a big performance penalty.
- Draw lines with 4 vertices (2 polygons) and using the vertex shader for doing the width: Very fast, a lot of control on the width but need general anti-aliasing for being anti-aliased and that is, again, a big performance penalty. 
- Draw lines with 6 vertices (6 polygons) and using the vertex shader for doing the width: Fast, a lot of control on the width and antialiased allways in a "cheap" way (very performant compared with global antialiasing) but needs more memory and (6 points every segment instead of 2 or 4) and the vertex shader is more complex and need more information (not so much).
- Draw lines with 4 vertices (2 polygons) without using the vertex shader for anything: A bit faster on redering (no vertex shader involved) but no control of the width ones created (need to be recreated every time the width change) and not possible to escalate the lines maintaining the width in pixels in the screen without rebuilding all the poligons every frame.

src/core/graphics/Graphics.js
```diff
@@ -50,6 +50,14 @@ export default class Graphics extends Container
         this.lineWidth = 0;
 
         /**
+         * If true the liens will be draw using LINES instead of TRIANGLE_STRIP
```
bigtimebuddy: `liens` => `lines`

bigtimebuddy:
@DavidBM thoughts on my comments here? Would love to get this new feature in our upcoming 4.4 release.

DavidBM:
Hi! Right now I'm not available, but this weekend I will dedicate time to
it. In general term I agree with you. If this weekend I don't see any
problem I will implement it. :)

DavidBM:
I updated the pull request. I think that, as you said, with the constructor argument is more clear :)

Any more required change, tell me!

bigtimebuddy:
Hey @DavidBM, terrific work. This is really looking great. However, it doesn't look like you added a constructor argument for Graphics. Was this missed?

```js
constructor(nativeLines = false)
{
}
```

DavidBM:
Yups! My bad. I was accessing to `nativeLines` property from outside of the object for testing and I totally forgot. Fixed.

bigtimebuddy:
Thanks @DavidBM for a great pull request, the Pixi community thanks you!

DavidBM:
Thanks! Is a pleasure! :)

PR #3331, opened 2016-11-20
Reworked cursor handling to expand cursor options.

### LINKED ISSUE #3301: Moving mouse out of pixijs canvas sets inline cursor style
Easily observable in every pixijs project, even the one on pixijs website. First move mouse cursor on canvas and then move it on some other element outside of canvas, or outside browser window. As a result "cursor: inherit;" inline style is permanently set for <canvas> element. This prevents setting custom cursors for canvas using CSS classes.

andrewstart:
Changed how cursors are handled:
* `InteractionManager` has a `cursorStyles` dictionary, which is a number of CSS cursor styles, callbacks upon cursor change, or dictionaries of CSS properties.
* `InteractionManager` tracks the cursor mode, and then handles it according to its `cursorStyles` dictionary.
* `interactiveTarget.cursor` is the cursor mode to be used when moused over. It is not limited to just 'default' and 'pointer'.
* `interactiveTarget.buttonMode` now just sets cursor to 'pointer'.

Also, added some mixin utilities so that the Accessibility and Interaction mixins can use accessors and can be affected by deprecations.

This does not directly fix #3301, but allows users to specify callbacks for cursor changes _instead_ of changing CSS, so that their callbacks can change CSS classes.

Additionally, fixed bugs in the InteractionManager where I found them, related to mouseover/out handling, and #3662.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -6,8 +6,8 @@ import EventEmitter from 'eventemitter3';
 import interactiveTarget from './interactiveTarget';
 import MobileDevice from 'ismobilejs';
 
-// Mix interactiveTarget into core.DisplayObject.prototype
-Object.assign(
+// Mix interactiveTarget into core.DisplayObject.prototype, after deprecation has been handled
+core.utils.mixins.delayMixin(
     core.DisplayObject.prototype,
     interactiveTarget
 );
@@ -220,20 +220,32 @@ export default class InteractionManager extends EventEmitter
         this.onPointerOver = this.onPointerOver.bind(this);
 
         /**
-         * Every update cursor will be reset to this value, if some element wont override it in
-         * its hitTest.
-         *
-         * @member {string}
-         * @default 'inherit'
+         * Dictionary of how different cursor modes are handled. Strings are handled as CSS cursor
+         * values, objects are handled as dictionaries of CSS values for interactionDOMElement,
+         * and functions are called instead of changing the CSS.
+         * Default CSS cursor values are provided for 'default' and 'pointer' modes.
+         * @member {Object.<string, (string|Function|Object.<string, string>)>}
          */
-        this.defaultCursorStyle = 'inherit';
+        this.cursorStyles = {
+            default: 'inherit',
+            pointer: 'pointer',
+        };
 
         /**
-         * The css style of the cursor that is being used.
+         * The mode of the cursor that is being used.
+         * The value of this is a key from the cursorStyles dictionary.
          *
          * @member {string}
          */
-        this.currentCursorStyle = 'inherit';
+        this.currentCursorMode = null;
+
+        /**
+         * Internal cached let.
+         *
+         * @private
+         * @member {string}
+         */
+        this.cursor = null;
 
         /**
          * Internal cached let.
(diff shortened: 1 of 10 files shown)


src/interaction/InteractionManager.js
```diff
@@ -263,20 +263,32 @@ export default class InteractionManager extends EventEmitter
...
+         * TODO: Finish documenting this.
+         */
+        this.cursorStyles = {
+            default: 'inherit',
+            pointer: 'pointer',
+        };
+
+        /**
+         * The mode of the cursor that is being used.
+         * The value of this is a key from the cursorStyles dictionary.
```
andrewstart: I'm not sure how to properly document this. It is a dictionary/map of the form `<string, string|Function|{[cssProp:string]:string}>`.

englercj: Something like `Object.<string, (string|Function|Object.<string, string>)>`

src/interaction/InteractionManager.js
```diff
@@ -681,14 +693,46 @@ export default class InteractionManager extends EventEmitter
...
+    setCursorMode(mode)
+    {
+        mode = mode || 'default';
+        this.currentCursorStyle = mode;
+        const style = this.cursorStyles[mode];
+
+        // only do things if there is a cursor style for it
+        if (style)
+        {
+            if (typeof style === 'string')
```
englercj: I'd do a switch-case here.

bigtimebuddy: @andrewstart follow-up here?

bigtimebuddy:
@andrewstart could you resolve conflicts?

andrewstart:
Feedback and conflicts have been taken care of.

src/interaction/InteractionManager.js
```diff
@@ -683,14 +695,46 @@ export default class InteractionManager extends EventEmitter
...
+    setCursorMode(mode)
+    {
+        mode = mode || 'default';
+        this.currentCursorStyle = mode;
+        const style = this.cursorStyles[mode];
+
+        // only do things if there is a cursor style for it
+        if (style)
+        {
+            switch (typeof stye)
```
bigtimebuddy: `stye` => `style`?

andrewstart: These are not the typos you are looking for. Move along. (fixed)

andrewstart:
I've wiped the old branch and reapplied my changes to get rid of any merge conflicts, as well as fixed any bugs I found and added a number of unit tests, so this needs another look.

bigtimebuddy:
@andrewstart should `delayMixin` be private? Doesn't seem like that would work externally without manually calling `performMixins`. Also, is it possible to accomplish this cursor stuff without creating `delayMixin`?

andrewstart:
Yeah, it should be private - there is no reason for anything outside the library to use it. `delayMixin` is required to have `deprecation.js` affect any mixin - without that change, mixins are applied to prototypes as soon as the code applying the mixin is imported. For example, importing the interaction code in `deprecation.js` causes the mixin to be applied to `DisplayObject.prototype`, before you could attempt to modify the mixin.

bigtimebuddy:
Oops @andrewstart looks like a test conflict. Can you resolve? I will merge this after that.

andrewstart:
Conflicts should be solved.

PR #3341, opened 2016-11-21
Allow '0x' notation of colours via strings to work in Text class

themoonrat:
I know internally, canvas likes "#FF0000" for setting colours, but I find it easier to have all values as "0xFF0000" in my settings json. "0x" notation already works in places such as tinting, so I like being able to have this consistency into TextStyle

### DIFF
--- a/src/core/text/TextStyle.js
+++ b/src/core/text/TextStyle.js
@@ -443,7 +443,7 @@ function getSingleColor(color)
     }
     else if ( typeof color === 'string' )
     {
-        if ( color.includes('0x') )
+        if ( color.indexOf('0x') === 0 )
         {
             color = color.replace('0x', '#');
         }
--- a/src/core/text/TextStyle.js
+++ b/src/core/text/TextStyle.js
@@ -435,22 +435,43 @@ export default class TextStyle
  * @param {number|number[]} color
  * @return {string} The color as a string.
  */
-function getColor(color)
+function getSingleColor(color)
 {
     if (typeof color === 'number')
     {
         return hex2string(color);
     }
-    else if (Array.isArray(color))
+    else if ( typeof color === 'string' )
     {
-        for (let i = 0; i < color.length; ++i)
+        if ( color.includes('0x') )
         {
-            if (typeof color[i] === 'number')
-            {
-                color[i] = hex2string(color[i]);
-            }
+            color = color.replace('0x', '#');
         }
     }
 
     return color;
+}
+
+/**
+ * Utility function to convert hexadecimal colors to strings, and simply return the color if it's a string.
+ * This version can also convert array of colors
+ *
+ * @param {number|number[]} color
+ * @return {string} The color as a string.
+ */
+function getColor(color)
+{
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
It seems a little weird to me that `"0xFF0000"` is the desired format here, since `0xFF0000` is a hexidecimal format. The hex color string `"#FF0000"` is useful because it's already handled by external APIs  and `0xFF0000` is useful because it's already a uint. I don't see the advantage here.

kaansoral:
On the opposite direction, I wish "#FF0000"'s were internally converted to 0xFF0000's for other usages

Using the "Color Sublime" plugin, I always end up converting colors to the hex format, modify them, then re-convert them to 0x's

But in both cases, it's usually a minor nuisance, if I was pulling values from a json, I would probably add a string/eval based solution on my end

themoonrat:
Hi @bigtimebuddy - the reason I came to this solution as the PR was....

I was using tints in my game code, and so was using 0xFF0000 style format. I then wanted to put these tint values into a JSON settings file; and thus used '0xFF0000' which works. Finally, I came to put text colours into the JSON settings file.... and '0xFF0000' didn't work. Me, being a man of consistency, wanted to solve this, and so had 2 options:

1. Allow text to take the '0xFF0000' syntax (this PR)
2. Allow tint to take the '#FF0000'

I chose option 1 because then it meant, across both settings and code, everything was in 0xFF0000 style, whether within JSON as a string, or in code as a hex number. Mmmm, yummy consistency.

I totally get why this wouldn't be accepted though, as '#FF0000' is what is used by the canvas api. Thought I'd put it out there for discussion.

src/core/text/TextStyle.js
```diff
@@ -435,22 +435,43 @@ export default class TextStyle
...
 {
     if (typeof color === 'number')
     {
         return hex2string(color);
     }
-    else if (Array.isArray(color))
+    else if ( typeof color === 'string' )
     {
-        for (let i = 0; i < color.length; ++i)
+        if ( color.includes('0x') )
```
bigtimebuddy: A polyfill is required for `includes` as it's not supported in IE. Suggestion is to use `color.indexOf('0x') === 0`

bigtimebuddy:
I'm open, but would like other to weigh-in on the utility of this. With issues like this, I really wish we could standardize across all the API instead of making small tweaks (including utility methods in PIXI.utils). There are other places with colors like Graphics as well as Text, tint and Filter?

Also, saw one small code issue.

src/core/text/TextStyle.js
bigtimebuddy: The ES6 API way to do this would be color.startsWith('0x'), but we aren't there yet.

englercj:
@themoonrat Can't you store the tint as `0xFF0000` in json (the number, not the string) then use hex2rgb to send it into APIs?

themoonrat:
Hex and Oct numbers are not valid JSON, only Dec numbers are

englercj:
Wow, how dumb. Didn't know that!

PR #3342, opened 2016-11-21
Fix for the texture garbage collector

themoonrat:
Do not unload generated textures when manually running the texture garbage collector. This check is already in place for when it is run automatically. The check is there, as otherwsie the texture to not be able to be renderered again.

Fixes: https://github.com/pixijs/pixi.js/issues/3215


GoodBoyDigital:
nice! approved :) - rolled it in as its quite small too!

Issue #3349, opened 2016-11-23
Slabo 27px Not Rendering


ClickSimply:
Hey  Guys,

I've ran into an odd issue, PixiJS won't render the Google Font "Slabo 27px".  Doesn't make any sense why this would be the case but it is.  I threw together a codepen:

http://codepen.io/clicksimply/pen/vymRZM

For some reason the font is always shown at like 2px size Arial, even though the browser is rendering it fine.  I am also using WebFont loader to make sure the fonts are available before painting them to the canvas.

bigtimebuddy:
Ahh. I think I see what's going on here. We are neglecting to put quotes around the fontFamily when the text style string is generated for Canvas. Should be an easy fix.

PR #3352, opened 2016-11-23
Fix for Text's fontFamily

bigtimebuddy:
### Fixed

* Fixes issue #3349 by putting quotes around fontFamiy when setting the font style.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -735,7 +735,7 @@ export default class Text extends Sprite
         // build canvas api font setting from individual components. Convert a numeric style.fontSize to px
         const fontSizeString = (typeof style.fontSize === 'number') ? `${style.fontSize}px` : style.fontSize;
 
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
After Travis finishes running, could you test your example with this version of Pixi: 
http://pixijs.download/font-family-fix/pixi.js

ClickSimply:
Confirm the patch is working, swapped out the pixi JS library with the fix one:
http://codepen.io/clicksimply/pen/yVbQGP

bigtimebuddy:
Awesome, thanks @ClickSimply for helping to make Pixi better.

simnalamburt:
This change prevents users from using multiple font-family with pixi.js. We need to revert this change or make another method allowing users to use multiple font-family.

themoonrat:
Yeps, on my todo list. Been swamped recently, sorry :( https://github.com/pixijs/pixi.js/issues/3406

PR #3353, opened 2016-11-24
use high precision float instead of medium in texture shader

Icemic:
Fiddle: https://jsfiddle.net/Icemic/wmdruwLc/show/light/

Device: MEIZU MX5 ( Android 5.1 )
Browser: Chromium
CPU: MTK Helio X10 Turbo (MTK6795)
GPU: PowerVR G6200

Detail: Draw a 800\*600 image to a 800\*600 canvas, there will be a 1px width line at right. After some test, I find that the image is scaled to 799\*600. It must be to do with precision, maybe a GPU bug.

image

Solution: add `precision highp float;` to `pixi.js/src/core/sprites/webgl/texture.vert` (default is `mediump`)

Note: Need more test for detecting how many CPUs/GPUs effected by this bug.

### DIFF
--- a/src/core/sprites/webgl/texture.vert
+++ b/src/core/sprites/webgl/texture.vert
@@ -1,3 +1,4 @@
+precision highp float;
 attribute vec2 aVertexPosition;
 attribute vec2 aTextureCoord;
 attribute vec4 aColor;


themoonrat:
A setting exists to change this i believe.

Take a look here: http://pixijs.download/v4.2.2/docs/PIXI.settings.html#.PRECISION

themoonrat:
(and the reason it's not highp by default is due to compatibility issues on some devices if i remember correctly)

Icemic:
Thanks for the new docs!

But I think it is a issue only to do with sprite texture, solving it by editing global setting may not a good choise.

Besides, I did fix this problem by using `highp`. If it is a compatibility issue, how to explain it?

themoonrat:
Some mobile devices do not support highp according to this comment: https://github.com/pixijs/pixi.js/issues/1922#issuecomment-115292500

Icemic:
Oh, I misunderstood your words before. 

The issue shows that `highp` may cause compatibility issue, and according to http://stackoverflow.com/questions/31496639/webgl-highp-fragment-support, even iPhone5S and Nexus 5 have this issue.

Well, I have to admit the pr is not appropriate, I'll close it.

For this issue, detecting `highp` support by `getShaderPrecisionFormat()` and set `PIXI.settings.PRECISION` will work.

Icemic:
No...No, I'm a bit confused. 😭 

It is said that use `highp`in **fragment shaders** will cause problem because `highp` support for fragment is optional, but not in **vertex shaders**.

So, I think my pr will not cause compatibility issue, and however fix another compatibility issue.

themoonrat:
My bad!

finscn:
Merged in my fork. Thank you.

PR #3361, opened 2016-11-25
Adds tintRgbFloat property to Sprite (fixes #3229)

### LINKED ISSUE #3229: [Suggestion] Compute hex2rgb of Sprite when set tint
I found that  for Mesh : 
```
{
    key: 'tint',
    get: function get() {
      return core.utils.rgb2hex(this.tintRgb);
    }

    /**
     * Sets the tint the mesh uses.
     *
     * @param {number} value - The value to set.
     */
    ,
    set: function set(value) {
      this.tintRgb = core.utils.hex2rgb(value, this.tintRgb);
    }
  }

```

when set tint , the this.tintRgb ( an array)  be set.

but for Sprite : 
```
{
        key: 'tint',
        get: function get() {
            return this._tint;
        }

        /**
         * Sets the tint of the sprite.
         *
         * @param {number} value - The value to set to.
         */
        ,
        set: function set(value) {
            this._tint = value;
            this._tintRGB = (value >> 16) + (value & 0xff00) + ((value & 0xff) << 16);
        }
```

only compute   this._tintRGB (an integer), the this.tintRgb not .
but when render sprite ,  we need  this.tintRgb sometimes.

jojuntune:

Currently setter for tint in Sprite class sets variable _tintRGB, which is integer. As requested in #3229, this commit adds a getter and a setter for tintRgbFloat, which is an array of three floats representing the rgb value of tint. This corresponds the tintRgb value in Mesh class.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -1,5 +1,5 @@
 import { Point, ObservablePoint, Rectangle } from '../math';
-import { sign, TextureCache } from '../utils';
+import { sign, TextureCache, hex2rgb, rgb2hex } from '../utils';
 import { BLEND_MODES } from '../const';
 import Texture from '../textures/Texture';
 import Container from '../display/Container';
@@ -89,6 +89,15 @@ export default class Sprite extends Container
          * @member {PIXI.Filter|PIXI.Shader}
          */
         this.shader = null;
+
+        /**
+         * The tint applied to the mesh. This is a [r,g,b] value. A value of [1,1,1] will remove any
+         * tint effect.
+         *
+         * @member {number}
+         * @memberof PIXI.mesh.Mesh#
+         */
+        this._tintRgbFloat = new Float32Array([1, 1, 1]);
 
         /**
          * An internal cached value of the tint.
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
@jojuntune, Thanks for the PR. 

Overall, I like the idea of creating API consistency with Mesh, I'm not sure this is the right approach. 

This PR creates two alternative nearly identical APIs on Sprite (`tint` and `tintRgbFloat`). The name of the setter is different than Mesh (`tintRgb` vs `tintRgbFloat`) so it is not actually create a corresponding API. Lastly, with this PR developers need reassign the value if changing  any value of `tintRgbFloat`, which is inconvenient. 

```js
var sprite = new PIXI.Sprite();
sprite.tintRgbFloat[0] = 0.5;
sprite.tintRgbFloat = sprite.tintRgbFloat;
```

jojuntune:
Hi @bigtimebuddy and thanks for the reply. I think those are good remarks that you presented. My approach clearly needs some rework.

bigtimebuddy:
I'm going to close this for now. If you decide to rework it, feel free to reopen.

Issue #3343, opened 2016-11-22
Graphics arc issue


istersin:
When creating an arc I get the following: "Uncaught TypeError: Cannot read property 'shape' of null" in PIXI.Graphics.arc, on line var points = this.currentPath.shape.points;

The same code works fine one 4.0.3. Also it might be related to #1086.

ggarek:
Hey, have same issue.

Here is a small [fiddle](https://jsfiddle.net/avprqboq/), which is based on official example for Graphics module.
It shows the problem. My guess is that there is a bug in begin/end -Fill methods.

- The example first draws a triangle, then an arc. If you run it as is, then script wont fail but the arc will be connected to the triangle, despite there are proper begin/end -Fill calls.
<img width="317" alt="screen shot 2016-11-26 at 20 21 54" src="https://cloud.githubusercontent.com/assets/5501062/20642032/04e3a61e-b416-11e6-9b8e-c57c9b0cd2b7.png">


- Then if you try to uncomment a code which draws rounded rectangle, the script will fail with the error mentioned by @istersin

staff0rd:
@ggarek your first issue, where the arc is connected to the previous shape is because `beginFill` and  `endFill` do not actually define the start/finish of a shape, rather they just change the current fill settings for the current shape.  Thus, arc (and arcTo) just continue drawing to the current shape.

I'm not sure what code changes need/should occur, however calling `moveTo` is a workaround; https://codepen.io/staff0rd/pen/aBqmEW?editors=0010

staff0rd:
Closed via #3363

ArronFerguson:
Thank you for this fix! Appreciate it!

PR #3363, opened 2016-11-26
fix/graphics: get points for newly created shape (#3433)

ggarek:
Fixed the issue #3343 
Added a test case, it catches the bug, but it is not very obvious as test should be. So if you have any suggestions i will modify/remove it.

All tests pass, new build works (use for my project now), and lint is satisfied. 😄

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -491,6 +491,7 @@ export default class Graphics extends Container
         const startX = cx + (Math.cos(startAngle) * radius);
         const startY = cy + (Math.sin(startAngle) * radius);
 
+        // If the currentPath exists, take its points. Otherwise call `moveTo` to start a path.
         let points = this.currentPath ? this.currentPath.shape.points : null;
 
         if (points)
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -491,9 +491,9 @@ export default class Graphics extends Container
         const startX = cx + (Math.cos(startAngle) * radius);
         const startY = cy + (Math.sin(startAngle) * radius);
 
-        const points = this.currentPath.shape.points;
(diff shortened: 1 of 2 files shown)


test/core/Graphics.js
```diff
@@ -128,4 +128,33 @@ describe('PIXI.Graphics', () =>
             expect(graphics.containsPoint(point)).to.be.false;
         });
     });
+
+    describe('complex drawing #1: draw triangle, rounder rect and an arc (issue #3433)', () =>
```
bigtimebuddy: `describe` should be the API, since this is a test for the change you made, would suggest this be `arc`. Please include you're current description as an inline-comment.

src/core/graphics/Graphics.js
```diff
@@ -491,9 +491,9 @@ export default class Graphics extends Container
         const startX = cx + (Math.cos(startAngle) * radius);
         const startY = cy + (Math.sin(startAngle) * radius);
 
-        const points = this.currentPath.shape.points;
+        let points = this.currentPath ? this.currentPath.shape.points : null;
```
bigtimebuddy: Can you please inline a comment about why you're checking for this case?

test/core/Graphics.js
ggarek: Ok, just added requested changes.

But as i described in the issue, it is possible that something else is broken within Graphics module. I just found a case, when things blow up on `.arc(..)` call, but i also described another case when things do not blow up, but the outcome is not correct (as one can see on the attached screenshot).

So, may be it is a good idea to add more tests for the module. I could do it, probably as another PR, but i do not know how the API is meant to work internally.

src/core/graphics/Graphics.js
ggarek: Sure, its also done.

ivanpopelyshev:
@ggarek congrats!

Issue #3356, opened 2016-11-25
Pixi 4.2.2 cacheAsBitmap shifts container in Canvas renderer


FFankias:
I have container full of sprites, for performance set  container.cacheAsBitmap = true;
WebGl renderer works fine, but with in Canvas container being shifted.

valtver:
Hi,

Got the same issue.
Reason is that rendered texture with old transformation is cut by renderer. So, _skipUpdateTransform_ .
**pixi.js/src/extras/cacheAsBitmap.js**
#_`Line 249:         renderer.render(this, renderTexture, true, m, false);`_
**`Line 249:         renderer.render(this, renderTexture, true, m, true);`**

valtver:
Sorry, wrong one...

286:  m.tx -= bounds.x;
 287   m.ty -= bounds.y;

to

 **286:  m.tx = -bounds.x;
 287   m.ty = -bounds.y;**

sivael:
This will not fix the issue, it will simply change where it appears.
I've created a PR #3494 for this.

GoodBoyDigital:
Looks all good here!

brenwell:
I have the same bug

brenwell:
I have a way to small reproducible version if you need it

GoodBoyDigital:
@brenwell yes please!

brenwell:
sure 1 second

brenwell:
This seems to highlight the issue for me

PIXI v4 + CanvasRender + PixelRatio > 1 + CacheAsBitmap = a strange rendering in regard to position. Change the 2 `const` to see the difference

https://jsfiddle.net/brenwell/krok0es6/5/

GoodBoyDigital:
perfect! this will definitely speed up the bug hunting! :D

themoonrat:
That fiddle is not using the latest version. Change it to 4.3.1 or above and you can see that the issue is fixed :)

brenwell:
Yep it 4.3.1+ looks good, thanks guys. Sorry for the run-around

bigtimebuddy:
@brenwell thank _you_ for wanting to make PixiJS better!

PR #3366, opened 2016-11-27
Fix Issue #3356 

valtver:
Pixi 4.2.2 cacheAsBitmap shifts container in Canvas renderer.

- Transform Matrix elements should be overwritten

Fixes issue: #3356


sivael:
The problem is that on my other project part of the elements renders properly with the pre-patch solution, and part works with post-patch solution.
After the change the issue remains, it's just for the other type of items.

sivael:
Look at #3494 for the proper fix, this is not fixing the issues, it fixes half of them while causing issues to pop up i the other half if use cases..

englercj:
Closing in favor of #3494.

Issue #3149, opened 2016-10-19
looks need touchcancel and unique identifier 


SeminYun:
hi.

i think InteractionData need Identifier and 
touchstart ,touchmove, touchend  have to use same inteactiondata 
or can check identifier to use multtouch

and touchcancel need becuase sometimes touchend is disappeared. (ios is happened often)

if attack button's touchend is gone, there give me big problem.

ios touch identifier is increased continuously.

so if pixi don't have own identifier to manage touch, ,, hmm..

i don't have any solution with pixi yet..

3.x 4.x same..

i think touch event operate like phaser touch input
http://phaser.io/examples/v2/input/multi-touch

thx..

themoonrat:
There is a 'touchendoutside' which is often useful to bind to the same function that you call on 'touchend'.   The description according to the docs is

> Fired when a touch point is removed outside of the display object that initially registered a [touchstart]

starakaj:
I'm also finding that I need touchcancel and pointercancel events. Is there any reason that InteractionManager doesn't currently attach to these events?

andrewstart:
Unfamiliarity with the events, I would imagine (first I've heard of them). Do you think it would make sense for those events to fire `touchendoutside` and `pointerendoutside` events within PIXI to streamline what people have to listen for?

starakaj:
In practice that's usually what I end up doing, but it's not always correct. For example, if you're passing your touch events off to a gesture recognizer. Though I agree that it's super annoying to have to listen to three different "touch is over now" events.

NicolasVereenoghe:
@SeminYun there was a pixi.interaction.InteractionData::identifier property, back in pixi 3.1.x.

It was indeed very handy for multitouch / gesturte gameplay. You could distinguish and manage differently each gesture.

I still hope it could be adapted for 4.x, there's nothing that replaced this concept of touch identification.

For 4.x, to manage multitouch environment, and distinguish a touch from another, I update a map of "active" touches, with their coordinates. If something moves, I update the nearest registerd "active" touch coordinates. In this maner, I can give some identification to all registerd touches ... it's quite reliable, but not 100%

andrewstart:
#3546 has been merged, so this should be solved.

PR #3375, opened 2016-11-29
See #3149 --- Add support for touchcancel and pointercancel

starakaj:
Doesn't fix all of the concerns in #3149, but does add touchcancel and pointercancel events

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -825,7 +825,10 @@ export default class InteractionManager extends EventEmitter
          * Normalized events, however, may have the double mousedown/touchstart issue on the native android browser,
          * so still need to be prevented.
          */
-        if (this.autoPreventDefault && (originalEvent.pointerType === 'touch' || originalEvent.pointerType === 'mouse'))
+
+        // Guaranteed that there will be at least one event in events, and all events must have the same pointer type
+
+        if (this.autoPreventDefault && (events[0].pointerType === 'touch' || events[0].pointerType === 'mouse'))
         {
             originalEvent.preventDefault();
         }
@@ -917,13 +920,13 @@ export default class InteractionManager extends EventEmitter
 
             this.emit(cancelled ? 'pointercancel' : 'pointerup', interactionEvent);
 
-            if (originalEvent.pointerType === 'mouse')
+            if (event.pointerType === 'mouse')
             {
                 const isRightButton = event.button === 2 || event.which === 3;
 
                 this.emit(isRightButton ? 'rightup' : 'mouseup', interactionEvent);
             }
-            else if (originalEvent.pointerType === 'touch')
+            else if (event.pointerType === 'touch')
             {
                 this.emit(cancelled ? 'touchcancel' : 'touchend', interactionEvent);
             }
@@ -1048,7 +1051,7 @@ export default class InteractionManager extends EventEmitter
     {
         const events = this.normalizeToPointerData(originalEvent);
 
-        if (originalEvent.pointerType === 'mouse')
+        if (events[0].pointerType === 'mouse')
         {
             this.didMove = true;
 
@@ -1075,9 +1078,9 @@ export default class InteractionManager extends EventEmitter
                 this.processPointerMove,
                 interactive
             );
-            this.emit('pointermove', this.eventData);
-            if (originalEvent.pointerType === 'touch') this.emit('touchmove', interactionEvent);
-            if (originalEvent.pointerType === 'mouse') this.emit('mousemove', interactionEvent);
+            this.emit('pointermove', interactionEvent);
+            if (event.pointerType === 'touch') this.emit('touchmove', interactionEvent);
+            if (event.pointerType === 'mouse') this.emit('mousemove', interactionEvent);
         }
 
         if (events[0].pointerType === 'mouse')
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -1,6 +1,7 @@
 import * as core from '../core';
 import InteractionData from './InteractionData';
 import InteractionEvent from './InteractionEvent';
+import InteractionTrackingData from './InteractionTrackingData';
 import EventEmitter from 'eventemitter3';
 import interactiveTarget from './interactiveTarget';
 import MobileDevice from 'ismobilejs';
@@ -885,7 +886,7 @@ export default class InteractionManager extends EventEmitter
 
         if (hit)
         {
-            displayObject._pointerIdentifiers.add(id);
+            displayObject._trackedPointers[id] = new InteractionTrackingData(id);
             this.dispatchEvent(displayObject, 'pointerdown', interactionEvent);
 
             if (e.type === 'touchstart')
@@ -896,7 +897,15 @@ export default class InteractionManager extends EventEmitter
             {
                 const isRightButton = e.button === 2 || e.which === 3;
 
-                displayObject[isRightButton ? '_isRightDown' : '_isLeftDown'] = true;
+                if (isRightButton)
+                {
+                    displayObject._trackedPointers[id].rightDown = true;
+                }
+                else
+                {
+                    displayObject._trackedPointers[id].leftDown = true;
+                }
+
                 this.dispatchEvent(displayObject, isRightButton ? 'rightdown' : 'mousedown', interactionEvent);
             }
         }
@@ -967,9 +976,9 @@ export default class InteractionManager extends EventEmitter
 
         const id = interactionEvent.data.identifier;
 
-        if (displayObject._pointerIdentifiers.has(id))
+        if (displayObject._trackedPointers[id] !== undefined)
         {
-            displayObject._pointerIdentifiers.delete(id);
+            delete displayObject._trackedPointers[id];
             this.dispatchEvent(displayObject, 'pointercancel', interactionEvent);
 
             if (e.type === 'touchcancel')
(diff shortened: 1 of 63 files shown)


andrewstart:
This looks pretty straightforward overall. How would you feel about emitting `Xendoutside` events on display objects, and `Xcancel` events from the InteractionManager? I feel like that would get us all of the ease of use for button states and such, but also allow gesture management to differentiate between events.

src/interaction/InteractionManager.js
```diff
@@ -1287,226 +1195,96 @@ export default class InteractionManager extends EventEmitter
...
      * Ensures that the original event object contains all data that a regular pointer event would have
      *
      * @private
-     * @param {TouchEvent|MouseEvent} event - The original event data from a touch or mouse event
+     * @param {TouchEvent|MouseEvent|PointerEvent} event - The original event data from a touch or mouse event
      */
     normalizeToPointerData(event)
     {
-        if (this.normalizeTouchEvents && event.changedTouches)
+        const type = typeof event;
```
andrewstart: I don't think that typeof works this way - it would probably just return `"object"`.

starakaj: Good catch, I definitely meant `instanceof`

starakaj:
Still workin'. I like this idea for the left/right mouse flags. I still don't know about the outside vs. cancel issue. Is the issue backwards compatibility? FWIW, InteractionManager wasn't event listening to xcancel events before, so adding them and then emitting xcancel events shouldn't break anyone's old code.

andrewstart:
Not backwards compatibility, but I imagine that the browser sends a `touchcancel` event instead of a `touchend` event, so it would be nice if everyone's `touchendoutside` listeners didn't have to change and would work as expected with the new event. It is definitely a design decision worth some consideration and discussion though.

src/interaction/interactiveTarget.js
```diff
@@ -88,18 +88,10 @@ export default {
...
-
-    /**
-     * Internal check to detect the number of active touches on the displayObject
-     *
-     * @inner {number}
+     * @inner {Set}
      * @private
      */
-    _touchCount: 0,
+    _pointerIdentifiers: new Set(),
```
andrewstart: If not, then perhaps Map would be suitable, as the ES5 compatibility fallback should just be `{}`.

bigtimebuddy: Map and Set are both not supported on IE 9 & 10. Would suggest added a polyfill to fallback to Object.

ivanpopelyshev:
Are pointer identifiers strings or what?

starakaj:
Pointer Identifiers are integers, mapped from DOM events. DOM PointerEvents have a pointerId property, and the Touch elements of a TouchEvent all have an identifier property. For mouse events there's just a constant pointer identifier MOUSE_POINTER_ID.

ivanpopelyshev:
Why not make use array as a set then?

starakaj:
Okay, so I'm feeling pretty good about where this is at. I think I've address the changes that @andrewstart wanted, among other stuff. Let me know if there's anything else.

src/interaction/InteractionTrackingData.js
```diff
@@ -0,0 +1,136 @@
...
+ */
+export default class InteractionTrackingData
+{
+    /**
+     * @param {number} pointerId - Unique pointer id of the event
+     */
+    constructor(pointerId)
+    {
+        this._pointerId = pointerId;
+        this._flags = this.constructor.FLAGS.NONE;
```
andrewstart: I'd suggest using `InteractionTrackingData.FLAGS`, for one less variable depth, and in case someone does something weird like subclass InteractionTrackingData.

src/interaction/InteractionTrackingData.js
```diff
@@ -0,0 +1,136 @@
...
+     *
+     * @param {boolean} yn - Is the right mouse button down?
+     */
+    set rightDown(yn)
+    {
+        this._doSet(this.constructor.FLAGS.RIGHT_DOWN, yn);
+    }
+
+    /**
+     * Is the left mouse button down?
```
andrewstart: Is this intended to be a flag for if the button is down _anywhere_, or just if there was a mouse down on the display object?

src/interaction/interactiveTarget.js
```diff
@@ -8,41 +8,39 @@
...
  *      function MyObject() {}
  *
- *      Object.assign(
- *          MyObject.prototype,
- *          PIXI.interaction.interactiveTarget
- *      );
+ *      PIXI.interaction.interactiveTarget.call(MyObject.prototype)
  */
-export default {
+export default function ()
```
andrewstart: I'd keep this as an object instead of a function - as a function you are instantiating `getTrackedPointers` for every InteractiveTarget. Also, it is easier to deprecate things on InteractiveTarget if it is an object we can attach properties to.

src/interaction/InteractionManager.js
```diff
@@ -864,286 +804,184 @@ export default class InteractionManager extends EventEmitter
...
-     *
-     * @private
-     * @param {MouseEvent} event - The DOM event of a mouse button being released
-     */
-    onMouseUp(event)
-    {
-        this.mouse.originalEvent = event;
-        this.eventData.data = this.mouse;
-        this.eventData._reset();
+            const interactionData = this.getIteractionDataForPointerId(event.pointerId);
```
andrewstart: Should this be `getInteractionDataForPointerId` (note the n in Interaction)? This typo appears to be in a few places.

src/interaction/InteractionManager.js
```diff
@@ -1155,392 +993,383 @@ export default class InteractionManager extends EventEmitter
...
 
     /**
-     * Processes the result of a touch move check and dispatches the event if need be
+     * Get InteractionData for a given pointerId. Store that data as well
      *
      * @private
-     * @param {PIXI.Container|PIXI.Sprite|PIXI.extras.TilingSprite} displayObject - The display object that was tested
-     * @param {boolean} hit - the result of the hit test on the display object
+     * @param {number} pointerId - Identifier from a pointer event
+     * @return {InteractionData} - Interaction data for the givent pointer identifier
```
andrewstart: givent -> given

src/interaction/InteractionManager.js
```diff
@@ -675,9 +591,29 @@ export default class InteractionManager extends EventEmitter
         // Resets the flag as set by a stopPropagation call. This flag is usually reset by a user interaction of any kind,
         // but there was a scenario of a display object moving under a static mouse cursor.
         // In this case, mouseover and mouseevents would not pass the flag test in dispatchEvent function
-        this.eventData._reset();
+        for (const k in this.activeInteractionData)
+        {
+            if (Object.prototype.hasOwnProperty.call(this.activeInteractionData, k))
```
andrewstart: This would be more efficient, if slightly less safe, if you used `this.activeInteractionData.hasOwnProperty(k)`

starakaj:
@andrewstart Oh yeah, I guess those never do get removed. They shouldn't be, for mouse and pointer events, since those are still traced back to a single source event after "up" events, but definitely for touches they should be removed.

src/interaction/InteractionManager.js
starakaj: Yes!

src/interaction/InteractionManager.js
starakaj: Agreed. I only changed it after the linter yelled at me.

src/interaction/interactiveTarget.js
starakaj: Hmm hmm. Okay, I may need to do something more involved that Object.assign to apply the mixin then, since otherwise I can't think of an easy way to give each interactiveTarget a unique object to use with _trackedPointers

src/interaction/InteractionTrackingData.js
starakaj: Only if it's over the display object. I believe this is used to differentiate between pointerup and pointerupoutside events.

src/interaction/InteractionTrackingData.js
starakaj: I'm sorry, I'm not sure I understand. Wouldn't that assign this._flags to be an object?

andrewstart: Sorry, the point being swapping out `this.constructor` for `InteractionTrackingData` - you would still set it to `FLAGS.NONE`.

src/interaction/InteractionTrackingData.js
andrewstart: Yeah, I just found the description confusing.

src/interaction/InteractionTrackingData.js
starakaj: Gotcha!

starakaj:
Alright, not sure what to do about the hasOwnProperty thing, since if I change it I can't get pixi to lint, but other than that this should be better.

andrewstart:
For the `hasOwnProperty` thing, you can put `// eslint-disable-next-line no-prototype-builtins` on the line above it - InteractionManager created and owns the object in question, so can be pretty sure that it has the `hasOwnProperty` method.

starakaj:
Good call

starakaj:
Hmm, this is failing because of c05f15b, which added InteractionManager tests based on the un-refactored code

src/interaction/InteractionManager.js
```diff
@@ -1033,7 +1033,7 @@ export default class InteractionManager extends EventEmitter
 
         const isTouch = (e.type === 'touchend' || e.pointerType === 'touch');
 
-        const isMouse = (e.type.startsWith('mouse') || e.pointerType === 'mouse');
+        const isMouse = (/^mouse/.test(e.type) || e.pointerType === 'mouse');
```
bigtimebuddy: Consider `e.type.indexOf('mouse') === 0` as it is faster than using RegExp.

bigtimebuddy:
@themoonrat any follow-up here?

Also, @starakaj can you resolve merge conflicts?

bigtimebuddy:
@starakaj ping

starakaj:
Sorry I'm confused. Where should I be seeing merge conflicts?

bigtimebuddy:
@starakaj I apologize, I mess this up this merge by trying to resolve conflicts. Can you create a new PR for this? Rebase with dev so that you can pull in the MockPointer changes.

starakaj:
Sorry, so to do that do I need to fork dev again?

bigtimebuddy:
You can just create a new pull request from your dev branch fork.

starakaj:
Alright, good?

bigtimebuddy:
I don't see anything. I think you might need to re-fork :/

englercj:
@bigtimebuddy Did you revert this? Github has it marked as merged.

bigtimebuddy:
@englercj I did a force push to dev instead of a revert. What that not the right thing to do?

englercj:
Usually a revert is nice because we can see it was removed, normally avoid force pushing unless absolutely necessary.

PR #3376, opened 2016-11-29
Adds tintRgb property to Sprite (#3229)

jojuntune:

Currently setter for tint in Sprite class sets variable _tintRGB, which is an integer. As requested in #3229, this commit adds a getter for _tintRgb, which is an array of three floats representing the rgb value of tint. This corresponds the _tintRgb value in Mesh class. 

The tint can only be set through _tint, which has setter and getter. _tintRgbInt has only getter which returns an integer by value (not by reference!). Also only getter for _tintRgb is provided, which returns an immutable array of three floats. The reasoning for this is as follows. If for example tintRgb returned a reference to the array, the contents of the array could be changed by assigning a new value to an individual element of the array (e.g. tintRgb()[2] = 0.5). But changing the contents of _tintRgb array should in that case require that the values of _tint and _tintRgbInt are also updated in order to keep the tint values consistent. Implementing that would be tricky and made the implementation of the class difficult.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -735,7 +735,7 @@ export default class Text extends Sprite
         // build canvas api font setting from individual components. Convert a numeric style.fontSize to px
         const fontSizeString = (typeof style.fontSize === 'number') ? `${style.fontSize}px` : style.fontSize;
 
-        return `${style.fontStyle} ${style.fontVariant} ${style.fontWeight} ${fontSizeString} ${style.fontFamily}`;
+        return `${style.fontStyle} ${style.fontVariant} ${style.fontWeight} ${fontSizeString} "${style.fontFamily}"`;
     }
 
     /**


ivanpopelyshev:
Its possible to postpone it until we make "tint" even for containers? Does anyone think we should merge this version first?

bigtimebuddy:
I agree with you Ivan. Tinting needs some unification for the API. While this gives an alternative API that might be useful in some cases, I would like to see a PR that improves the tinting for everything.

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -351,7 +351,7 @@ export default class SpriteRenderer extends ObjectRenderer
             uint32View[index + 12] = uvs[2];
             uint32View[index + 17] = uvs[3];
 
-            uint32View[index + 3] = uint32View[index + 8] = uint32View[index + 13] = uint32View[index + 18] = sprite._tintRGB + (sprite.worldAlpha * 255 << 24);
+            uint32View[index + 3] = uint32View[index + 8] = uint32View[index + 13] = uint32View[index + 18] = sprite.tintRgbInt + (sprite.worldAlpha * 255 << 24);
```
bigtimebuddy: `_tintRGB` was calculated only once here when changing `tint`. Your addition will calculate `tintRgbInt` everytime. Not sure this is an improvement.

src/core/sprites/Sprite.js
```diff
@@ -555,7 +553,33 @@ export default class Sprite extends Container
...
+     * @member {Float32Array}
+     * @memberof PIXI.Sprite#
+     * @default [1.0,1.0,1.0]
+     */
+    get tintRgb()
+    {
+        return Object.freeze(hex2rgb(this._tint));
+    }
+
+        /**
```
bigtimebuddy: Spacing indent issue with the start of jsdoc block here.

jojuntune:
Oops, I inadvertently executed force push on my repo, so this pull request got closed.

However, I noticed the changes you requested @bigtimebuddy , and I will look into those.

PR #3379, opened 2016-11-30
Add attribute "rendererName" to Sprite Class

finscn:
Add attribute ```rendererName``` to Sprite Class , and the default value is "sprite"

Via ```rendererName```, users could  custom renderer of Sprite without creating a new Class extend Sprite , and also could change renderer in app running .

see https://github.com/pixijs/pixi.js/issues/3330

.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -28,6 +28,8 @@ export default class Sprite extends Container
     {
         super();
 
+        this.rendererName = 'sprite';
+
         /**
          * The anchor sets the origin point of the texture.
          * The default is 0,0 this means the texture's origin is the top left
@@ -282,8 +284,8 @@ export default class Sprite extends Container
     {
         this.calculateVertices();
 
-        renderer.setObjectRenderer(renderer.plugins.sprite);
-        renderer.plugins.sprite.render(this);
+        renderer.setObjectRenderer(renderer.plugins[this.rendererName]);
+        renderer.plugins[this.rendererName].render(this);
     }
 
     /**
@@ -294,7 +296,7 @@ export default class Sprite extends Container
     */
     _renderCanvas(renderer)
     {
-        renderer.plugins.sprite.render(this);
+        renderer.plugins[this.rendererName].render(this);
     }
 
     /**


bigtimebuddy:
My suggestion would be to rename to `pluginType` instead of `rendererName` because it is a selection of which `plugin` on the renderer.

finscn:
```pluginType```  is too abstract and macroscopic.

I think we should classify plugin.

**Comment by bigtimebuddy** [comment]
I do not see the value in this addition for the community as a whole. If you need to do some custom or advanced Sprite rendering, you should consider using **pixi-gl-core** and building something custom over that or consider adding a monkey-patch instead:

```js
PIXI.Sprite.prototype. _renderWebGL  = function(renderer) {
         this.calculateVertices();
        renderer.setObjectRenderer(renderer.plugins.customSprite);
        renderer.plugins.customSprite.render(this);
};
```

finscn:
@bigtimebuddy , if use 
```
PIXI.Sprite.prototype._renderWebGL = a new function
```
All Sprite will be changed.

If we don't want to Change All Sprite ,  the right way is create a subclass of Sprite Class. like this ： 

export default class CustomSprite extends Sprite
{
   // override the method in parent class 
    _renderWebGL(renderer)
    {
        this.calculateVertices();
        renderer.setObjectRenderer(renderer.plugins.customSprite);
        renderer.plugins.customSprite.render(this);
    }
}
```

 If  when we just want change the render （it's a simple thing and a common request in a real game） , but  we have to create create a new Class  . I think this is not a good way.

And there is a other question :  If use a standalone class for custom renderer , We will hard to change renderer  dynamically  in app running.

Maybe you would think "Why do we need change renderer  dynamically?" . 
Let me show your a case:

Because SpriteRenderer is a common and fast (support batch draw) renderer .
So  in most scenes , I could choose it for my sprites.
But in some special scenes , I want do some special things to the sprites , So I would want to change the renderer dynamically (e.g. LightSpriteRenderer , PerspectiveRenderer ).

bigtimebuddy:
Hello, seems like you have a few valid options for implementing this functionality in your projects without a change to Pixi. For instance, you could create a class which inherits Sprite and add the rendererName property and use that instead. We appreciate the PR, but I'm going to close this for now.

ivanpopelyshev:
@bigtimebuddy I kinda need that thing for my plugins too, lets ask others :)

finscn:
@bigtimebuddy , You could ignore this issue ,  my request & my case.
Let's tall about  the issue  from another perspective, from design pattern & refactoring.

What's ```plugin``` ? Why we call it ```plugin ``` ? 
Because it is a extension, it is not required,  it is not a core, it is optional, it is just a option.
A option should not be hard coded in a core module of PIXI.
It should be defined/set  at a place , make (or let? F**king my bad english)  it could be changed easily.
 
So , I think the option for "which renderer should be used in Sprite"  is like a config. ```sprite``` is just a default value of it. By convention(or tradition? habit?), we should config it at constructor of Sprite Class.

ivanpopelyshev:
If we change renderer for Sprite by extending it into CustomSprite class, we'll have to do the same thing to Text. Instead we just make a component that specifies which renderer should be used on this particular Sprite instance. We can do same thing with Mesh and TilingSprite.

ivanpopelyshev:
OK, summary: change rendererName to pluginName and make add that thing to Mesh and TilingSprite. That version I'll approve.

bigtimebuddy:
Before being approved, this needs to be documented as well. Since @finscn's dominant language is not English, someone should help describe `pluginName`

ivanpopelyshev:
"Plugin that is responsible for rendering this element. Allows to change the rendering process without overriding 'renderWebGL' method."

At the same time, in deprecation of PIXI.Sprite :

"Shader that is responsible for rendering this element. Deprecated, please register the renderer plugin and use 'pluginName' field"

ivanpopelyshev:
@finscn don't worry buddy, I'll do it :)

Issue #3163, opened 2016-10-19
strange order of matrix operations


gludion:
Because of a "bug" in pixi-flump-runtime, I've spent some time to figure out the effect of scaling and skewing on a DisplayObject in PIXI, by reading your code here:
https://github.com/pixijs/pixi.js/blob/66462aa91c8df826ace7bc3168cfcb3df71ecd1a/src/core/math/Matrix.js
From your formulas in DisplayObject.setTransform, I understand that what you do is:
- first, you apply a skew matrix 
- then you apply a scale matrix
- then a rotation 
- then a translation

If you choose for example skew.x = 45° and scale.x = 2, you would expect that the object is made twice wider, **and** "skewed" by an angle of 45°. I mean, a real 45°, neither sharper nor smoother.
But what happens actually in the current implementation is that the scale actually **increase** the skew effect and the observed angle is much sharper than 45°, which confirms the order of operations described above.

Instead one would expect:
- first, scale the object 
- then skew the result
- then rotate the result
- then translate the result

This choice preserves a independency of the various effects applied to the object.

Therefore, I tried to find another way to change the transform of the object... but could't find any. Unless I missed something, there is no way besides using scale, skew, and rotation. 

One would expect a displayObject method like
`disp.setTransformMatrix(a,b,c,d,tx, ty)`
Such method can be very useful when you clearly know the matrix of the transform you want to achieve, that's why it's usually provided in display list paradigm.
(side note: from what I understood from the source code, but I may be wrong, I see that there is no localTransform stored anywhere, the actual stored data is globalTransform: weird, what happens if a container is modified do you recursively update all the world transforms of the children? seems cumbersome and probably not very CPU friendly)

Sorry if I missed something, tips and workarounds are welcome.

Some suggestions:
- consider changing the order of operations in setTransform() (but it may break existing apps usings skew extensively)
- consider adding another method: disp.setTransformMatrix(a,b,c,d,tx,ty)

ivanpopelyshev:
Well, i think you are right about skew. 

In Spine they are in different order.

gludion:
I did'nt try Spine, but as far as I remember from tutorials, their IDE focuses more on rotation than skew...
Being stuck with skew issues is rare, because trying to achieve a visual squashing effect is usually ok even if skew is not accurate...
The real problem comes when using third-part tools (animation tools, game engines) built on top of PIXI: they will normally need a proper set transform somewhere...
[...]

PR #3382, opened 2016-11-30
Fix for order of skew and scale in transform, plus a test

ivanpopelyshev:
This is the fix for #3163

### DIFF
--- a/src/core/display/Transform.js
+++ b/src/core/display/Transform.js
@@ -67,10 +67,10 @@ export default class Transform extends TransformBase
      */
     updateSkew()
     {
-        this._cx = Math.cos(this.rotation - this.skew._y);
-        this._sx = Math.sin(this.rotation - this.skew._y);
-        this._cy = Math.cos(this.rotation + (Math.PI / 2) + this.skew._x);
-        this._sy = Math.sin(this.rotation + (Math.PI / 2) + this.skew._x);
+        this._cx = Math.cos(this.rotation + this.skew._y);
+        this._sx = Math.sin(this.rotation + this.skew._y);
+        this._cy = -Math.sin(this.rotation - this.skew._x); // cos, added PI/2
+        this._sy = Math.cos(this.rotation - this.skew._x); // sin, added PI/2
     }
 
     /**
--- a/src/core/display/Transform.js
+++ b/src/core/display/Transform.js
@@ -54,25 +54,23 @@ export default class Transform extends TransformBase
          */
         this._rotation = 0;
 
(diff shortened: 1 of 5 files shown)


ivanpopelyshev:
OK, what do I think about it: we should use SHEAR, and not SKEW. Make skew a property. Look at the code, its horrible! Game developers won't understand what is skew, while shear is just a rotation of axis X and axis Y

ivanpopelyshev:
Oh, I forgot to say: it is a breaking change. I don't want to add "doSkewBeforeScale" flag, its obvious that it must be made that way.

src/core/display/TransformStatic.js
```diff
@@ -69,16 +67,16 @@ export default class TransformStatic extends TransformBase
...
     updateSkew()
     {
-        this._cy = Math.cos(this.skew._y);
-        this._sy = Math.sin(this.skew._y);
-        this._nsx = Math.sin(this.skew._x);
-        this._cx = Math.cos(this.skew._x);
+        this._cx = Math.cos(this.rotation + this.skew._y);
+        this._sx = Math.sin(this.rotation + this.skew._y);
+        this._cy = -Math.sin(this.rotation - this.skew._x); // cos, added PI/2
+        this._sy = Math.cos(this.rotation - this.skew._x); // sin, added PI/2
```
GoodBoyDigital: hows about - ```this._rotation``` rather than ```this.rotation```

src/core/display/Transform.js
```diff
@@ -156,7 +148,6 @@ export default class Transform extends TransformBase
     set rotation(value)
     {
         this._rotation = value;
-        this._sr = Math.sin(value);
-        this._cr = Math.cos(value);
+        this.updateSkew();
     }
```
GoodBoyDigital: should we rename updateSkew to something more meaningful (as it deals with rotation now too) Not sure what though!

bigtimebuddy: What about `updateRotation`?

ivanpopelyshev: lets do it in next iteration. I also want to add "shear" as the real field , rather than "skew"

ivanpopelyshev: @bigtimebuddy I don't want to change that right now, because I'm gonna change it in next iteration anyway. "updateSkew" is as good as "updateRotation", because it uses both params

bigtimebuddy: I'm cool with that approach for now

bigtimebuddy:
I need to test against pixi-animate, will get back with results.

PR #3392, opened 2016-12-02
cocoonJS interaction position fix

ivanpopelyshev:
Suggested by @bqvle. Tested on iphone,

getClientBoundingRect and coords are screwed, that's why its different by "resolution^2"

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -735,8 +735,16 @@ export default class InteractionManager extends EventEmitter
             rect = this.interactionDOMElement.getBoundingClientRect();
         }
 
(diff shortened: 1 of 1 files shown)


bQvle:
this fixes the behaviour of Cordova canvas+ when devicePixelRatio != 1

ivanpopelyshev:
Tested on android, it works.

ivanpopelyshev:
@andrewstart merged with your suggestion.

JiDW:
Did you test on Webview+ ?

bQvle:
It works normal with Webview and Webview+, its because Canvas+ sizes the canvas differently from a normal webview. I'm sure that Webview and Webview+ isn't triggering navigator.isCocoonJS

However when you resize your canvas to be fullscreen in Canvas+ it will be 4x the size in webview if the device pixelratio = 2

PR #3396, opened 2016-12-03
Graphics.arc fix

### LINKED ISSUE #3343: Graphics arc issue

staff0rd:
Fixes #3343, adds test.


staff0rd:
Dupe of #3363

PR #3411, opened 2016-12-05
Add "predist" to package.json

finscn:
Remove the previous version output files before run dist.
Because maybe there are other files in the dist folder, so  we shouldn't remove the folder.

Related issues : 
https://github.com/pixijs/pixi.js/issues/3403
https://github.com/pixijs/pixify/pull/8

.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "lintfix": "npm run lint --fix",
     "prebuild": "npm run lint",
     "build": "npm run dist",
-    "predist": "rimraf dist/pixi*.js dist/pixi*.map",
+    "predist": "rimraf dist/pixi.*js dist/pixi.*.map",
     "dist": "pixify -d dist -n PIXI -o pixi -t babelify",
     "lib": "babel src --out-dir lib -s",
     "docs": "jsdoc -c scripts/jsdoc.conf.json -R README.md",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "lintfix": "npm run lint --fix",
     "prebuild": "npm run lint",
     "build": "npm run dist",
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Why not delete the entire `dist` folder?

finscn:
Because maybe there are other files in the dist folder, so we shouldn't remove the folder.

example:  
I put my plugin's  output files in dist folder too, but I don't want to remove & rebuild them.

.

bigtimebuddy:
I don't understand why you're adding files to **dist**. Is this in your fork of **pixi.js**? If so, I would add a `predist` script just in your fork. Otherwise, the behavior here is that `predist` should delete the whole **dist** folder.

finscn:
Ok ， I agree with you ， and I've changed the PR. please review again.
Now, use ``` rimraf dist/*.*```.

But please don't remove the entire ```dist``` folder.
In some IDE/editors  , remove a folder will take some little problems, even recreate immediately.

bigtimebuddy:
Can you be more specific about the error? What IDE/editor has a problem removing the whole folder? I have never had a problem with this on other projects.

finscn:
Example 
 in macOS's  Finder  ( it's not an IDE editor), I've opened the ```dist``` folder.
If at this moment ，remove the ```dist``` folder , the Finder will goto the parent folder.
After  build , I have to find & open ```dist``` folder again.

in an IDE (I forget it's name) , remove a folder in a project , it will pop a warning dialog and the IDE will get the focus and be the top window.

in some text editors , remove a folder, will auto remove it from the resource list , but when you create a new one , it won't auto add the folder to the resource list.

bigtimebuddy:
Replacing with #3441

PR #3413, opened 2016-12-06
Customize rendering process with "pluginName" field.

ivanpopelyshev:
Revamped #3379

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -120,6 +120,13 @@ export default class Sprite extends Container
 
         this._transformID = -1;
         this._textureID = -1;
+
+        /**
+         * Plugin that is responsible for rendering this element. Allows to customize the rendering process without overriding 'renderWebGL' method.
+         *
+         * @member {string}
+         */
+        this.pluginName = 'sprite';
     }
 
     /**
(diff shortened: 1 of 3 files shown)


finscn:
I think the comment 
```without overriding 'renderWebGL' method```
could be changed to 
```without overriding '_renderWebGL' & '_renderCanvas'  method```.

**Comment by ivanpopelyshev** [comment]
@finscn done

**Comment by ivanpopelyshev** [comment]
Look at that thing. We have to add this in 3 places, and there are things like "tint" and "texture" that exist for all three classes. I think we need one more class between Container and Sprite.

**Comment by finscn** [comment]
A Class named "BaseSprite" ?

**Comment by ivanpopelyshev** [comment]
or "SpriteBase" , yep :)

**Comment by bigtimebuddy** [comment]
Good work @ivanpopelyshev. I attempted to generate the docs and pluginName did not show up. Not sure why this is, could be a larger issues with jsdoc.

**Comment by bigtimebuddy** [comment]
`blendMode` is also not showing up on PIXI.Sprite, so I think it's a jsdoc issue. However, you should set a `@default` doc tag:

```js

/**
 * Plugin that is responsible for rendering this element.
 * Allows to customize the rendering process without overriding '_renderWebGL' & '_renderCanvas' methods.
 *
 * @member {string}
 * @default sprite
 */
 this.pluginName = 'sprite';
```

ivanpopelyshev:
Done, please check it

ivanpopelyshev:
Bunnymark shown no performance drop. I doubt that 100k gets in dictionary with several elements can affect anything.

PR #3415, opened 2016-12-06
Add Feature: update transform as orphan

finscn:
I've told many things  about  "updateTransform" , at 
https://github.com/pixijs/pixi.js/issues/3370
https://github.com/pixijs/pixi.js/issues/3389
https://github.com/pixijs/pixi.js/issues/3388

My english is too weak to explain it  in further detail.

There are just few things to add ： 

### Why named the function as ```updateTransformAsOrphan``` ？
Maybe It's not the best name , I named it just by  convention of web development.
In web development, We call a no-parent DOM node ```orphan node```(in China).

### Why don't check parent in ```updateTransform``` , but add a new function ?
I want do it. But I'm afraid that if I do it , you wouldn't accept it for stability of current version.

### When we need this feature  ？
Many times , we can't ensure any display object & transform object has a parent. 
When we need update  orphan object manually ， if no the function like this , there will be an error.

~

I think we could solve the problem about no-parent transform first.
Solving the same problem about no-parent display object  would be the next step.

### DIFF
--- a/src/core/display/Transform.js
+++ b/src/core/display/Transform.js
@@ -96,8 +96,6 @@ export default class Transform extends TransformBase
      */
     updateTransform(parentTransform)
     {
-        const pt = parentTransform.worldTransform;
-        const wt = this.worldTransform;
         const lt = this.localTransform;
 
         lt.a = this._cx * this.scale._x;
@@ -109,6 +107,9 @@ export default class Transform extends TransformBase
         lt.ty = this.position.y - ((this.pivot.x * lt.b) + (this.pivot.y * lt.d));
 
         // concat the parent matrix with the objects transform.
+        const pt = parentTransform.worldTransform;
+        const wt = this.worldTransform;
+
         wt.a = (lt.a * pt.a) + (lt.b * pt.c);
         wt.b = (lt.a * pt.b) + (lt.b * pt.d);
         wt.c = (lt.c * pt.a) + (lt.d * pt.c);
--- a/src/core/display/Transform.js
+++ b/src/core/display/Transform.js
@@ -120,6 +120,33 @@ export default class Transform extends TransformBase
     }
 
     /**
+     * Updates local matrix & world matrix without parent
+     */
+    updateTransformAsOrphan()
+    {
+        const lt = this.localTransform;
+
+        lt.a = this._cx * this.scale._x;
+        lt.b = this._sx * this.scale._x;
+        lt.c = this._cy * this.scale._y;
+        lt.d = this._sy * this.scale._y;
+
+        lt.tx = this.position._x - ((this.pivot._x * lt.a) + (this.pivot._y * lt.c));
+        lt.ty = this.position._y - ((this.pivot._x * lt.b) + (this.pivot._y * lt.d));
+
+        const wt = this.worldTransform;
+
+        wt.a = lt.a;
+        wt.b = lt.b;
+        wt.c = lt.c;
+        wt.d = lt.d;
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
I think we need another branch on big transform refactoring and include that thing there. If we add one more method, we'll have total 6 methods that do the same thing with local transform :(

ivanpopelyshev:
What if we fork pixi and add collabs like you to it? We can support different branches that way and provide a way to test each experiment in pixi-examples.

finscn:
In my opinion  if don't call   ```updateLocalTransform``` in other  update transform, we could remove it.
I've no idea when users would use it.
If   6 methods too many , we could add checking to ```updateTransform```

```
    updateTransform(parentTransform)
    {
        const lt = this.localTransform;

        lt.a = this._cx * this.scale._x;
        lt.b = this._sx * this.scale._x;
        lt.c = this._cy * this.scale._y;
        lt.d = this._sy * this.scale._y;

        lt.tx = this.position.x - ((this.pivot.x * lt.a) + (this.pivot.y * lt.c));
        lt.ty = this.position.y - ((this.pivot.x * lt.b) + (this.pivot.y * lt.d));

        // concat the parent matrix with the objects transform.
        const wt = this.worldTransform;

        if (parentTransform)
        {
            const pt = parentTransform.worldTransform;

            wt.a = (lt.a * pt.a) + (lt.b * pt.c);
            wt.b = (lt.a * pt.b) + (lt.b * pt.d);
            wt.c = (lt.c * pt.a) + (lt.d * pt.c);
            wt.d = (lt.c * pt.b) + (lt.d * pt.d);
            wt.tx = (lt.tx * pt.a) + (lt.ty * pt.c) + pt.tx;
            wt.ty = (lt.tx * pt.b) + (lt.ty * pt.d) + pt.ty;
        }
        else
        {
            wt.a = lt.a;
            wt.b = lt.b;
            wt.c = lt.c;
            wt.d = lt.d;
            wt.tx = lt.tx;
            wt.ty = lt.ty;
        }

        this._worldID ++;
    }
```

finscn:
I think update an orphan node is a common request and a core feature.
Testing  it in a new branch or fork is OK , but we should add it to the master branch finally.

finscn:
OK , with my pleasure.

ivanpopelyshev:
@finscn

By the way, one of ideas is to introduce (Spine transform)[https://github.com/pixijs/pixi-spine/blob/master/src/core/Bone.ts] .

finscn:
@ivanpopelyshev ,  I think PIXI need a pure matrix module,  it could be used in many place besides  calculating transform.
e.g. , in my extension PerspectiveRenderer.

 I will try it later.

finscn:
@ivanpopelyshev , I found  there are too many differences  between origin PIXI.js and your fork version. 
It's too hard to understand by me.  Tears...

ivanpopelyshev:
@finscn you dont have to use my "master" , just fetch my repo, and commit your branch there. "master" is 3d stuff, dont touch it, while dev1-4 i use for small PRs to pixi.

ivanpopelyshev:
About matrix, yes, "a,b,c,d,tx,ty" is much better than Float32Array, also because it uses DOUBLE precision. That way we can handle coords with good precision JS, and then upload normal stuff to GPU.

GoodBoyDigital:
Hi @finscn , Apprechiate the time you took to send us this. Ultimately I not feel this is not required as its an bit of an edge case for your specific needs. 
Cheers!

PR #3422, opened 2016-12-08
Add pluginName to  Plugin

finscn:
A Plugin should know its own name.

If use
```
ctor.prototype.name = pluginName;
```
The ```name``` is too common , there may be conflict in Plugin. So I used pluginName.

### DIFF
--- a/src/core/utils/pluginTarget.js
+++ b/src/core/utils/pluginTarget.js
@@ -23,6 +23,7 @@ function pluginTarget(obj)
     obj.registerPlugin = function registerPlugin(pluginName, ctor)
     {
         obj.__plugins[pluginName] = ctor;
+        ctor.prototype.pluginName = pluginName;
     };
 
     /**


bigtimebuddy:
Can you explain why this is necessary? What's the use case?

finscn:
```
if ( sprite.pluginName === plugin.pluginName ) {
 // do something
}
```

ivanpopelyshev:
No, we still don't see it :(

finscn:
@ivanpopelyshev  , I real code : 

```
if (sprite.pluginName !== render.currentRenderer.pluginName)
{
      render.currentRenderer.start();
      // ......
}
```

When I customize PIXI.js , I need this feature. Though , I could add it in my fork version ,
 but I think  ```One named object should know its own name```.

Another reason  is  `it should know its own name, though never been used`.
That's like : 

- Transform has `uploadLocalTransform`, though never been used.
- FilterManager has `shaderCache`, though never been used.

bigtimebuddy:
I still don't see the utility beyond this very specific use-case.

src/core/utils/pluginTarget.js
```diff
@@ -23,6 +23,7 @@ function pluginTarget(obj)
     obj.registerPlugin = function registerPlugin(pluginName, ctor)
     {
         obj.__plugins[pluginName] = ctor;
+        ctor.prototype.pluginName = pluginName;
```
bigtimebuddy: This property should also be read-only.

```js
obj.registerPlugin = function registerPlugin(pluginName, ctor)
{
    obj.__plugins[pluginName] = ctor;
    Object.defineProperty(ctor.prototype, 'pluginName', {
        get() {
             return pluginName;
        }
    }
};
```

finscn:
done with https://github.com/pixijs/pixi.js/commit/dcaf8ffe1f5d1ab02e09dc76fcc5522744f18af9

PR #3429, opened 2016-12-08
Control whether use core.ticker.shared to update

finscn:
Add a argument to control whether use core.ticker.shared to update animation.

Related Issue :  https://github.com/pixijs/pixi.js/issues/3428

Default value is true.
```
new AnimatedSprite(textures)  ---> use core.ticker.shared to update, as same as current version.

new AnimatedSprite(textures, false)   ---> don't use core.ticker.shared to update.
```

### DIFF
--- a/src/extras/AnimatedSprite.js
+++ b/src/extras/AnimatedSprite.js
@@ -32,18 +32,11 @@ export default class AnimatedSprite extends core.Sprite
     /**
      * @param {PIXI.Texture[]|FrameObject[]} textures - an array of {@link PIXI.Texture} or frame
      *  objects that make up the animation
-     * @param {boolean} [selfUpdate=true] - Whether use PIXI.ticker.shared to self-update animation time.
+     * @param {boolean} [autoUpdate=true] - Whether use PIXI.ticker.shared to auto update animation time.
      */
-    constructor(textures, selfUpdate)
+    constructor(textures, autoUpdate)
     {
         super(textures[0] instanceof core.Texture ? textures[0] : textures[0].texture);
-
-        /**
-         * `true` uses PIXI.ticker.shared to self-update animation time.
-         * @type {boolean}
-         * @default true
-         */
-        this.selfUpdate = selfUpdate !== false;
 
         /**
          * @private
@@ -56,6 +49,13 @@ export default class AnimatedSprite extends core.Sprite
         this._durations = null;
 
         this.textures = textures;
+
+        /**
+         * `true` uses PIXI.ticker.shared to auto update animation time.
+         * @type {boolean}
+         * @default true
+         */
+        this.autoUpdate = autoUpdate !== false;
 
         /**
          * The speed that the AnimatedSprite will play at. Higher is faster, lower is slower
@@ -118,7 +118,7 @@ export default class AnimatedSprite extends core.Sprite
         }
 
         this.playing = false;
-        if (this.selfUpdate)
+        if (this.autoUpdate)
         {
             core.ticker.shared.remove(this.update, this);
         }
@@ -136,7 +136,7 @@ export default class AnimatedSprite extends core.Sprite
         }
 
         this.playing = true;
-        if (this.selfUpdate)
+        if (this.autoUpdate)
         {
             core.ticker.shared.add(this.update, this);
         }
--- a/src/extras/AnimatedSprite.js
+++ b/src/extras/AnimatedSprite.js
@@ -54,6 +54,7 @@ export default class AnimatedSprite extends core.Sprite
          * `true` uses PIXI.ticker.shared to auto update animation time.
          * @type {boolean}
          * @default true
+         * @private
          */
         this.autoUpdate = autoUpdate !== false;
 
--- a/src/extras/AnimatedSprite.js
+++ b/src/extras/AnimatedSprite.js
@@ -137,7 +137,7 @@ export default class AnimatedSprite extends core.Sprite
         }
 
         this.playing = true;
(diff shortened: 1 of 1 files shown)


src/extras/AnimatedSprite.js
```diff
@@ -32,11 +32,14 @@ export default class AnimatedSprite extends core.Sprite
...
      * @param {PIXI.Texture[]|FrameObject[]} textures - an array of {@link PIXI.Texture} or frame
      *  objects that make up the animation
+     * @param {boolean} [autoUpdate=true] - Whether use core.ticker.shared to update animation.
      */
-    constructor(textures)
+    constructor(textures, autoUpdate)
     {
         super(textures[0] instanceof core.Texture ? textures[0] : textures[0].texture);
 
+        this.autoUpdate = autoUpdate !== false;
```
bigtimebuddy: `this.autoUpdate` requires a jsdoc:

```js
/**
 * `true` uses PIXI.ticker.shared to self-update animation time.
 * @type {boolean}
 * @default true
 */
```

src/extras/AnimatedSprite.js
```diff
@@ -32,11 +32,14 @@ export default class AnimatedSprite extends core.Sprite
     /**
      * @param {PIXI.Texture[]|FrameObject[]} textures - an array of {@link PIXI.Texture} or frame
      *  objects that make up the animation
+     * @param {boolean} [autoUpdate=true] - Whether use core.ticker.shared to update animation.
```
bigtimebuddy: Please replace `core.ticker.shared` with `PIXI.ticker.shared` in comment.

finscn:
@bigtimebuddy ,  done.

ivanpopelyshev:
If we are editing this file, I'd rather to make it a switch. Imagine that you have private field _autoUpdate and public property autoUpdate. When you change the property, it adds a handler to the ticker. 

That way we'll be able to switch it in the middle of the game. If you think its good, make another PR? :)

finscn:
@ivanpopelyshev  , I think the option about auto-update  should not be changed after animation be created， otherwise the logic in the game will be confusional.

e.g. ,  if could change  auto-update :

```
// case A:
var anim = new AnimatedSprite(textures, true);
anim.play();
anim.autoUpdate = false;
anim.stop();

// case B:
var anim = new AnimatedSprite(textures, false);
anim.play();
anim.autoUpdate = true;
anim.stop();

// case C:
var anim = new AnimatedSprite(textures, true);
anim.play();
anim.autoUpdate = false;
anim.destroy();

// case D:
var anim = new AnimatedSprite(textures, true);
anim.autoUpdate = false;
anim.play();
anim.autoUpdate = true;
anim.destroy();

//  ...... etc. 


``` 
The different  orders of  calling ```set-autoUpdate  , play(),  stop(),  destroy()```  will make  different sets of results.

If  we could change autoUpdate , we have to add many codes for manage the relation  between  animation and PIXI.ticker.shared.

-------

Let me talk more about ```auto update in PIXI.ticker.shared```.


In fact, I hope the default value of autoUpdate  could be `false` , even remove this feature.
This feature is good for develop the demo/example/mini game of PIXI.
But I think  the most of  formal games would manage the  animation update in their own game ticker/game loop.

The animation sprite is not a pure display object , sometimes it could be influenced by other logic in the game ，and it could influence other things（e.g. sound-playing , stage-event) .

The animation sprite is not a complete entity in game logic，most time， it is considered part of a entity object（e.g.  an instance of Player Class).
I think nobody like  manage player in game ticker , but  manage player.animation in a other ticker.


If someones want auto update animation in PIXI.ticker.shared , they could do it use 
```PIXI.ticker.shared.add(anim) ```  by themselves, it's not complex.


Anyhow  I'm firmly convinced that AnimatedSprite  should NOT be coupled to the PIXI.ticker.shared，and  should NOT require PIXI.ticker.shared.
Even PIXI.ticker.shared should be a standalone tool/helper ， it should NOT be coupled to any other module similar to  DisplayObject(Container, Sprite),  Renderer , Filter,  etc. 


.

.

finscn:
I will create a new Animation Class for PIXI.

I hope my implementation could express my ideas as clearly as possible.

.

bigtimebuddy:
@themoonrat I would suggest renaming to `_autoUpdate` and mark as `@private` in the docs.

finscn:
There is a demo of my new Animation  Class : 
http://fattyboy.cn/anim/

source code :  https://github.com/finscn/pixi.js/blob/master/src/extensions/Animation.js

The features of  my new Animation : 
* it's a normal Class  , not the subclass of DisplayObject.
* it's could apply any display object with texture , e.g. Sprite , Rope
* could implement any feature of  current AnimatedSprite , include time scaling & backward playing.
* a better & more clearly logic for updating animation then current AnimatedSprite


I hope you could consider let it be a part of official pixi.

bigtimebuddy:
@finscn could you address the changes requested by @themoonrat and @ivanpopelyshev

finscn:
@bigtimebuddy  , I prefer to do that `Mark this.autoUpdate as private in jsdocs`

if allow change it , will ... see  https://github.com/pixijs/pixi.js/pull/3429#issuecomment-265963396

themoonrat:
Err, @finscn - the suggestion from bigtimebuddy was to mark the property as private but also change to a leading underscore

so this._autoUpdate rather than this.autoUpdate

otherwise it's good

finscn:
@ivanpopelyshev ， changed.  Review again,  please.

GoodBoyDigital:
This looks good to go to me!

Issue #3439, opened 2016-12-10
【Bug】About  `_renderSpriteRect` in Graphics


finscn:
When fill a rect  via Graphics ,  PIXI.js will use a shared canvas Graphics._SPRITE_TEXTURE for performance.

but when I fill multiple rects  ,  the last one will override the previous ones ， at last there will be only one rect on the canvas.

When use `_renderSpriteRect` , maybe need flush() after render.

PR #3440, opened 2016-12-10
Do `flush()` after `_renderSpriteRect()`

finscn:
Fixed https://github.com/pixijs/pixi.js/issues/3439 

Why don't I put ```renderer.flush();``` into  ```_renderSpriteRect``` ?
`_renderSpriteRect`  just do something for  `rendering sprite rect` , the "flush" is another thing.
And maybe sometimes we need _renderSpriteRect but flush.

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -731,6 +731,7 @@ export default class Graphics extends Container
         if (this._fastRect)
         {
             this._renderSpriteRect(renderer);
+            renderer.flush();
         }
         else
         {


finscn:
@ivanpopelyshev  ,  you're wrong.
There is no changing current object renderer after `_renderSpriteRect()`

when call _renderSpriteRect() , the  current object renderer is SpriteRenderer.

when call another  _renderSpriteRect() , the  current object renderer still is SpriteRenderer.

Please see the code  in `Graphics.js`
```
    _renderWebGL(renderer)
    {
        // if the sprite is not visible or the alpha is 0 then no need to render this element
        if (this.dirty !== this.fastRectDirty)
        {
            this.fastRectDirty = this.dirty;
            this._fastRect = this.isFastRect();
        }

        // TODO this check can be moved to dirty?
        if (this._fastRect)
        {
            this._renderSpriteRect(renderer);
            renderer.flush();
        }
        else
        {
            renderer.setObjectRenderer(renderer.plugins.graphics);
            renderer.plugins.graphics.render(this);
        }
    }
```

When I fill many rects , no anything else , the  current object renderer is always be SpriteRenderer.


Please debug it in browser step by step.

And I've test it.  my patch is correct.

finscn:
My use case is too complex ， and it's a commercial project ， I can't post the code.

But please trust me ， add  `renderer.flush();` every thing is ok， without it ， there will be a bug like #3439

finscn:
@ivanpopelyshev  ,  I create a new test case : 
```
var width = 640, height = 480;

var renderer;

var stage;

// Shared graphics
var graphics;

window.onload = function() {

    renderer = new PIXI.WebGLRenderer(width, height, {
        clearBeforeRender: false,
        backgroundColor: 0xFFFFFF
    });

    stage = new PIXI.Container();
    graphics = new PIXI.Graphics();
    stage.addChild(graphics);

    start();
}

function start() {

    // bind default render texture
    renderer.bindRenderTexture();

    fillRect(graphics, 15, 10, 200, 80, 0xFF0000);

    fillRect(graphics, 15, 10 + 100, 200, 80, 0x00FF00);

    fillRect(graphics, 15, 10 + 100 * 2, 200, 80, 0x0000FF);

    renderer.flush();
}


function fillRect(graphics, x, y, width, height, color) {
    graphics.clear();
    graphics.beginFill(color);
    graphics.drawRect(x, y, width, height);
    graphics.endFill();

    graphics.updateTransform();
    graphics.renderWebGL(renderer);
}

```

I hope it could make you understand me.

For some reason , I  don't use  ```WebGLRenderer.render``` to  render display object.

If add the add `renderer.flush();` , you will see 3 rects, if on it , there will only blue one.

ivanpopelyshev:
If I approve that, then when I render 100k rects it wont be optimized because I'm calling flush every time.

ivanpopelyshev:
Canonical way for pixi is that you render some renderTextures, then render whole stage into main canvas once per frame. 

According to your PR's , you are making the other kind of solution. I understand why are you doing it, and I have a big project that does not uses one big stage too. But that particular PR cripples our main use-case, which is very bad.

ivanpopelyshev:
In your case, you have to call "renderer.flush()" after you are done with all the stage components. And "renderer.gl.flush()" because of that bug.

And no hard feelings, you just had to read my conversations with other maintainers ;) I was in your situation some time ago.

finscn:
@ivanpopelyshev  ,  

I know I could call `renderer.flush` by myself, but  `_renderSpriteRect`  is a hack/magic method ，and it has to be  `flush()` . 
We shouldn't tell users do that by themselves like this :  

> Please call  renderer.flush()  after you render a fill-rect .
> You don't need to call renderer.flush() after rendering other things.

The hack method should be a black box for users.

ivanpopelyshev:
All render methods are supposed to be called from other render methods or the root "renderer.render". If you don't follow that, even "sprite.renderWebGL' call will be wrong.

finscn:
If add `renderer.flush()` after `_renderSpriteRect` ,  we could use `sprite.renderWebGL` anywhere after updateTransform.

Just one line code `renderer.flush()` could make PIXI be more flexible ， Why not enjoy doing it ？

finscn:
I create a WebGLRenderer Wrapper ， use like this： 

```
// the game Loop/ticker :

renderWrapper.begin()
game.update();
game.render();
renderWrapper.end()
```

then , in  game.render , I can use 
```
renderWrapper.render(displayObject)
```
anywhere.

And I also could use   `renderWrapper.render(stage) ` like current PIXI.

It's very flexible.

The only problem is  ```_renderSpriteRect()```.

ivanpopelyshev:
I still don't understand you, and I don't think you understand me. Test 10k rectangles moving around (like bunnies) and you'll see that this change will crush our performance.

finscn:
"Test 10k rectangles moving around (like bunnies) and you'll see that this change will crush our performance."

No , I've  test , the performance no different. ( both bad)

I think you didn't try to debug it step by step.

finscn:
All fill-rect graphics object  share ONE Canvas , if no  renderer.flush() , they can't be display correctly at all.
So, current PIXI call `renderer.flush()` too ,  but it call `renderer.flush()` after render one graphics.

I just want to move  `renderer.flush()` to  the place after `_renderSpriteRect()`.

finscn:
When render many fill-rect graphics , the flow is this : 

```
render  rect-0 
       the currentRenderer  is set as  GraphicsRenderer 
                call _renderSpriteRect()
                           the currentRenderer  is set as SpriterRenderer 


render  rect-1
       the currentRenderer  is set as  GraphicsRenderer 
       ( *** at this time , currentRenderer changed , so , call ```renderer.flush()```)
                call _renderSpriteRect()
                           the currentRenderer  is set as SpriterRenderer 


```

finscn:
@ivanpopelyshev  , sorry I think I can't understand you.
 Learning English is more important than learning PIXI.js  for me.

I ask  one my friend , he translate your words for me ， I know where is my wrong.

Thanks & Sorry.

finscn:
@ivanpopelyshev  , because my english is very bad ,so  I can't find out the polite ways to express my opinion， I hope you could forgive me for this.

finscn:
@ivanpopelyshev   for the example at  https://github.com/pixijs/pixi.js/pull/3440#issuecomment-266204673 

I found a way to solve the problem ：

add  
`
this.currentPath = null;
this._spriteRect = null; 
`

to Graphics.prototype.clear .

Do you have some advices ？

Could I do that ？

ivanpopelyshev:
I mean 10k = 10000 , they will be optimized by using _spriteRect hack. If you add flush() there, then it won't be better than Graphics, and you can just remove "_spriteRect", it will be the same :)

ivanpopelyshev:
Oh, I understand :) You are rendering same Graphics multiple times, and you think that's without flush() that behaviour is bug, but please note that every graphics has its own spriteRect and in normal PIXI scene tree it appears only one time so its ok.

finscn:
@ivanpopelyshev  ,  I create a new PR.  https://github.com/pixijs/pixi.js/pull/3445

It could solve this `problem` (maybe not problem , just my request) , and doesn't crush our performance , and could fix some other bugs.

PR #3441, opened 2016-12-10
Adds predist and prelib

bigtimebuddy:
Replaces #3411


themoonrat:
Do we want to do the same thing with docs?

bigtimebuddy:
@themoonrat also added docs, good call

PR #3447, opened 2016-12-12
Fixing graphic = new PIXI.Graphic(); graphic.generateTexture(); throwing errors when using webgl renderer

lincsanders:
Was having a nightmare migrating a v3 codebase to v4 until I made this change. Perhaps not the most elegant, but hopefully enough to prompt someone with a better understanding into fixing the generateCanvasTexture function on an un-rendered instance of Graphics while using a WebGL renderer?

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -10,7 +10,6 @@ import Bounds from '../display/Bounds';
 import bezierCurveTo from './utils/bezierCurveTo';
 import CanvasRenderer from '../renderers/canvas/CanvasRenderer';
 
-let canvasRenderer;
 const tempMatrix = new Matrix();
 const tempPoint = new Point();
 const tempColor1 = new Float32Array(4);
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
``` 13:5  warning  'canvasRenderer' is defined but never used  no-unused-vars ```

englercj:
Does just resizing the shared CanvasRenderer when this function is called work? If so, I'd rather do that than create a new one each time this function is called.

lincsanders:
I've tried and tried with all variations on resizing the CanvasRenderer/CanvasBuffer/etc but can't get it working without reinstansiating the whole class unfortunately.

lincsanders:
Here is a fiddle of essentially copy/pasted of the parts causing the problem. There should be random-ish sized boxes scattered across the canvas in a working example.

http://jsfiddle.net/1py7jgua/2/

Investigations since I reported this have found the culprit, which is that I have a "plugin" i was using which overrided the render function on the webgl and canvas render prototypes. I have a small piece of functionality which is *right now at least* dependent on overriding that render function, and for some reason is fixed by re-instantiating the canvasRenderer every time.

I would gladly settle for a workaround if this pull request is too detrimental to performance with the re-instantiation change.

Cheeeeeeeeeeers!

lincsanders:
Appreciate the help from you awesome guys, and apologies for taking up your time on what is effectively a hack from my part. Alternatively, an undocumented hook into the renderer's render() function would pretty much solve this for me!

GoodBoyDigital:
the renderers to fire ```prerender``` and ```postrender``` events when you call render. Is that something that would help?

lincsanders:
I think it will! Go ahead and close this, my initial suspicions were that it was a bug with rendering a graphics texture in a webglrender, however it's firmly in the realm of "I can change things on my end" now.

Big ups, you guys are awesome and thanks again for the help! Been using pixi since v2 and won't be stopping any time soon.

You can go ahead and close this 👍 ❤️

GoodBoyDigital:
Your most welcome!

PR #3455, opened 2016-12-13
Rebind vao before glData upload.

finscn:
Fix  https://github.com/pixijs/pixi.js/issues/3453 .


ivanpopelyshev:
@GoodBoyDigital look at that please. Do we need vao bound before upload?

PR #3457, opened 2016-12-13
Remove needless `Number()` call.

finscn:
Call Number() once when create ` factor`  that better than call `Number()` multi times in for-loop.


ivanpopelyshev:
Why is there a "Number" anyway?

staff0rd:
nulls?

ivanpopelyshev:
But "null-null===0", so its obsolete here

ivanpopelyshev:
Oh, wait, its "undefined - undefined" !!! But then, what about first thing? x1 - x0?

englercj:
Looks good, but please make the PR to the `dev` branch; not `master`.

PR #3460, opened 2016-12-13
Adds tintRgb property to Sprite (#3229)

jojuntune:

Currently setter for tint in Sprite class sets variable _tintRGB, which is an integer. As requested in #3229, this commit adds a getter for _tintRgb, which is an array of three floats representing the rgb value of tint. This corresponds the _tintRgb value in Mesh class.

The tint can only be set through _tint, which has setter and getter. _tintRgbInt has only getter which returns an integer by value (not by reference!). Also only getter for _tintRgb is provided, which returns an immutable array of three floats. The reasoning for this is as follows. If for example tintRgb returned a reference to the array, the contents of the array could be changed by assigning a new value to an individual element of the array (e.g. tintRgb()[2] = 0.5). But changing the contents of _tintRgb array should in that case require that the values of _tint and _tintRgbInt are also updated in order to keep the tint values consistent. Implementing that would be tricky and made the implementation of the class difficult.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -1,5 +1,5 @@
 import { Point, ObservablePoint, Rectangle } from '../math';
-import { sign, TextureCache } from '../utils';
+import { sign, TextureCache, hex2rgb } from '../utils';
 import { BLEND_MODES } from '../const';
 import Texture from '../textures/Texture';
 import Container from '../display/Container';
@@ -70,9 +70,26 @@ export default class Sprite extends Container
          * @member {number}
          * @default 0xFFFFFF
          */
-        this._tint = null;
-        this._tintRGB = null;
-        this.tint = 0xFFFFFF;
+        this._tint = 0xFFFFFF;
+
+        /**
+         * The tint applied to the sprite. This is a [r,g,b] value.
+         *
+         * @member {number}
+         * @memberof PIXI.sprites.Sprite#
+         * @default [1,1,1]
+         */
+        this._tintRgb = new Float32Array([1, 1, 1]);
+
+        /**
+         * The tint applied to the sprite. This is an integer value
+         * which corresponds to the hex value of tint.
+         *
+         * @private
+         * @member {number}
+         * @default 16777215
+         */
+        this._tintRgbInt = 16777215;
 
         /**
          * The blend mode to be applied to the sprite. Apply a value of `PIXI.BLEND_MODES.NORMAL` to reset the blend mode.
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
Nice, am I correct that all that things are common for Mesh and TilingSprite also? Lets accept this one and then make super-class for Sprite/Mesh/TilingSprite

jojuntune:
Yes, to me it seems tint is common to Sprite/Mesh/TilingSprite, as they have a common parent class, Container.

finscn:
I propose  you to  change 
```
this._tintRgbInt = (this._tint >> 16) + (this._tint & 0xff00) + ((this._tint & 0xff) << 16);
```
to
```
this._tintRgbInt = (value >> 16) + (value & 0xff00) + ((value & 0xff) << 16);
```
 It could offer a bit little performance gains.

finscn:
And in  Mesh.js , it use  this.tintRgb .
```
    /**
     * The tint applied to the mesh. This is a hex value. A value of 0xFFFFFF will remove any tint effect.
     *
     * @member {number}
     * @memberof PIXI.mesh.Mesh#
     * @default 0xFFFFFF
     */
    get tint()
    {
        return core.utils.rgb2hex(this.tintRgb);
    }

    /**
     * Sets the tint the mesh uses.
     *
     * @param {number} value - The value to set.
     */
    set tint(value)
    {
        this.tintRgb = core.utils.hex2rgb(value, this.tintRgb);
    }
```


In your code, you  use  `this._tintRgb` , I think they should be named by the same way .

jojuntune:
Thanks for your remarks @finscn , I will change this._tintRgbInt as you suggested and uniform the naming of tintRgb.

finscn:
@jojuntune  ， maybe  your `this._ tintRgb` better than `this. tintRgb` , because tintRgb shouldn't be public.

src/core/sprites/Sprite.js
```diff
@@ -70,9 +70,26 @@ export default class Sprite extends Container
...
+        this._tint = 0xFFFFFF;
+
+        /**
+         * The tint applied to the sprite. This is a [r,g,b] value.
+         *
+         * @member {number}
+         * @memberof PIXI.sprites.Sprite#
+         * @default [1,1,1]
+         */
+        this._tintRgb = new Float32Array([1, 1, 1]);
```
englercj: Underscored properties are private, mark it as such with `@private`.

src/mesh/Mesh.js
```diff
@@ -114,7 +114,7 @@ export default class Mesh extends core.Container
          *
          * @member {number}
          */
-        this.tintRgb = new Float32Array([1, 1, 1]);
+        this._tintRgb = new Float32Array([1, 1, 1]);
```
englercj: Underscored properties are private, mark it as such with @private.

finscn:
@jojuntune  , I changed my mind.
I think  `tintRgb` should be public , and users could set it directly.
And in SpriteRenderer.shader, use `tintRgb` , not `_tintRGB` .
If do this , the tint will be more flexible. 

see https://github.com/pixijs/pixi.js/issues/3569

finscn:
I have some new point of this issue.

The tint of Mesh is different from tint of Sprite.
The value  in  Mesh.tint could be  `>1.0` , but Sprite.tint only supports `0---1.0`.

If make them look the same （both has tintRgb） ，maybe make users are confusing.

bigtimebuddy:
@jojuntune I'm closing this because of inactivity. Feel free to re-open when you can address @englercj's requested changes.

Issue #3458, opened 2016-12-13
Base Texture Resolution Issue (0.5 when should be 1)


megmut:
After Updating from 4.1.0, to 4.2.3, the textures are all double the size they should be when the device pixel ratio is 1. This still works fine when it is 2. When creating a new PIXI.Texture, if I log the resolution in either the constructor or in BaseTexture.prototype.update
    -- DPR1 --
    console.log(mySprite.texture.baseTexture.resolution) = 0.5
    -- DPR2 --
    console.log(mySprite.texture.baseTexture.resolution = 2

After much digging around in the BaseTexture, BaseRenderTexture and RenderTexture (constructors or update prototype), I am unable to find the source of this. If I am to console.log the resolution set in the BaseTexture, it shows as 1. I can't find where the resolution is being overriden, and I have tried overriding it myself by setting (mySprite.texture.baseTexture.resolution - 1) however this has no effect at all.

I even tried adding the src@1.png to my atlas's in an attempt to override this. Just as a note, this works fine for every version prior to 4.2.X.

staff0rd:
I'm struggling to remember how/where I discovered something similar, but I did and the issue was due to #3297.  Looking in my generated sprite sheets json, I strangely found (can't remember how I generated it or what tool I used) something like this;
~~~
"meta": {
    "image": "myspritesheet.png",
    "size": {"w":972,"h":1780},
    "scale": "0.5"
}
~~~

I'm not sure how the `0.5` got there in scale but switching it to `"1"` magically solved this problem for me, as it wasn't until #3297 that PIXI was referencing that field.

staff0rd:
Found it - it was a [codepen](https://codepen.io/staff0rd/pen/PGPgXr?editors=0010) for [this issue](https://github.com/pixijs/pixi.js/issues/2748#issuecomment-264641320) that broke (was pointing at latest pixi) so to fix it I did [this](https://github.com/staff0rd/pixi-tilesprite/commit/1393a4075971e6b010cfffd915406b1f00eb3329).

Prior to the fix the left side images were displayed in double size.

megmut:
You are right, we found the issue much later last night. I don't know why PIXI is parsing the scale object in the texture packer json file, but I think it probably shouldn't be. Now I have to make sure I manually update all my textures after exporting them :(

Thanks for taking the time to find the issue! Do we leave this as open for the dev's to fix or close this?

staff0rd:
[...]

PR #3464, opened 2016-12-14
Fixes logic for spritesheet resolution

bigtimebuddy:
Addresses #3458

This PR tweaks the logic for when to respect a spritesheet's internal `scale`.

The current implementation reads the filename but defaults to `1` and then overrides with the JSON scale if it's defined and not `1`.

This change prioritizes the filename resolution (e.g. @2x). If that isn't set, it falls back to the `scale` value set in the JSON. Otherwise, resolution of `1`. 

For example, now if a file is named `atlas@1x.png` with a corresponding scale value of `0.5` in the JSON data, the resolution will be `1` instead of `0.5`. 

### Added

* Adds an additional argument for `PIXI.utils.getResolutionOfUrl` to define the default value.

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -115,7 +115,7 @@ export function getResolutionOfUrl(url, defaultValue)
         return parseFloat(resolution[1]);
     }
 
-    return defaultValue || 1;
+    return defaultValue !== undefined ? defaultValue : 1;
 }
 
 /**
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -103,9 +103,10 @@ export function rgb2hex(rgb)
  * @memberof PIXI.utils
  * @function getResolutionOfUrl
  * @param {string} url - the image path
+ * @param {number} [defaultValue=1] - the defaultValue if not filename prefix is set.
  * @return {number} resolution / device pixel ratio of an asset
  */
-export function getResolutionOfUrl(url)
+export function getResolutionOfUrl(url, defaultValue)
 {
     const resolution = settings.RETINA_PREFIX.exec(url);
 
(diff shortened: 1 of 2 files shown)


src/core/utils/index.js
```diff
@@ -103,9 +103,10 @@ export function rgb2hex(rgb)
  * @memberof PIXI.utils
  * @function getResolutionOfUrl
  * @param {string} url - the image path
+ * @param {number} [defaultValue=1] - the defaultValue if not filename prefix is set.
```
staff0rd: No rather than not?

staff0rd:
Nice one, fixed this; https://codepen.io/staff0rd/pen/yVRWgE?editors=0010

src/core/utils/index.js
```diff
@@ -114,7 +115,7 @@ export function getResolutionOfUrl(url)
         return parseFloat(resolution[1]);
     }
 
-    return 1;
+    return defaultValue !== undefined ? defaultValue : 1;
```
themoonrat: This here checks against undefined for default value

src/loaders/spritesheetParser.js
```diff
@@ -43,14 +43,20 @@ export default function ()
             const frames = resource.data.frames;
             const frameKeys = Object.keys(frames);
             const baseTexture = res.texture.baseTexture;
-            let resolution = core.utils.getResolutionOfUrl(resource.url);
+            let resolution = core.utils.getResolutionOfUrl(resource.url, null);
```
themoonrat: But this here passes in null as the default value.

null !== undefined, so won't it try to return null rather than 1?

themoonrat:
PR looks god
The 
`let resolution = core.utils.getResolutionOfUrl(resource.url, null);`
confused me a tad, since then you go into that function, and it's comparing against undefined. At first I thought that was a mistake, but it's actually intentional to force returning of null. Maybe it's worth a comment by the code here to make it clear that this is the desired behaviour

PR #3465, opened 2016-12-14
Interaction fix

starakaj:
Some testing on android confirms that using 1 for `MOUSE_POINTER_ID` in InteractionManager is bad, since android can use 1 for touch identifiers. Using a constant 'MOUSE' fixes the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,10 +72,10 @@ $> npm install pixi.js
 #### CDN Install (via cdnjs)
 
 ```html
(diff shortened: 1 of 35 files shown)


starakaj:
thanks to @fde31 for identifying the problem!

andrewstart:
Could you add this to https://github.com/pixijs/pixi.js/pull/3375 instead, as we haven't merged that yet?

starakaj:
Sure thing. I merged this branch and pushed to dev, did it work?

andrewstart:
Yes it did! Thanks!

Issue #3450, opened 2016-12-12
Bug with mouse events on overlapping interactive elements


edclement:
Using  Pixi.js 4.2.3 - ✰ WebGL ✰

I have two sprites that overlap partially. They have the following code attached for example purposes (I cut out some excess and this will need to be modified slightly to actually work):
```js
let marker = new pixi.Sprite(this._texture);
marker.interactive = true;
marker.on('mouseover', this._onMarkerMouseOver);
marker.on('mouseout', this._onMarkerMouseOut);

notifyMouseOverMarker (spriteMarker) {
    console.log('over', spriteMarker.identifier, this._markersUnderMouse);
}

notifyMouseOutMarker (spriteMarker) {
    console.log('out', spriteMarker.identifier, this._markersUnderMouse);
}
```

This is what I get in my console:
![image](https://cloud.githubusercontent.com/assets/5667395/21113434/9860669e-c077-11e6-81b7-317eac06461f.png)

Notice that the `6cb255` marker gets the `mouseout` event twice, once before it even gets the `mouseover`event... ??

edclement:
This is in chrome 54.0.2840.99 btw

themoonrat:
Hi @edclement - could you create a codepen or jsfiddle demonstrating this please to make it easier for us to investigate? thank you!

edclement:
I'll try to do so today. Thanks.

edclement:
Here is an example of the issue: https://jsfiddle.net/gf2sabv7/
Check the console to see how the wrong sprite is passed as the event.data.target

staff0rd:
Possibly using the wrong property?  Switching this to `currentTarget` looks to get the functionality you're expecting; https://jsfiddle.net/pa4x26eb/

Issue #3416, opened 2016-12-06
JSdoc Bug with Members


bigtimebuddy:
Some members are not showing up in the docs. For instance: [PIXI.Sprite#blendMode](https://github.com/pixijs/pixi.js/blob/dev/src/core/sprites/Sprite.js#L77-L84). 

**Example:**
http://pixijs.download/dev/docs/PIXI.Sprite.html

This issue was noticed with #3413.

staff0rd:
It seems no public members are displaying by default.  To have them displayed you currently need to use `@memberOf` which makes `@private` seem a bit redundant.  See #3471.

themoonrat:
Indeed, I've found that with jsdoccing my game, which also uses es6 classes. Public functions are output, but not public members. @private isn't redundant as it'll correctly assist IDEs. Maybe there's some JSDoc configuration somewhere to pull in the public members

themoonrat:
Annoyingly, even if you set "private" to true in scripts/jsdoc.conf.json - that outputs private functions to the docs... but still no members at all. :8ball:

bigtimebuddy:
Is this a bug with jsdoc?

themoonrat:
I believe so, can't see a config option for it

themoonrat:
I've raised an issue on JSDoc github (https://github.com/jsdoc3/jsdoc/issues/1301) - we'll see what they say

Native-Coder:
REALLY wish we could get the ES6 Class Memebers to document correctly...

Native-Coder:
I'm having the same problem, but it looks as though it was reported over a month ago (https://github.com/jsdoc3/jsdoc/issues/1293). According to this report it has to do with using @export in the module that contains the class.

bigtimebuddy:
Fixed with #3542

PR #3470, opened 2016-12-16
InteractionEvent.target clarification

staff0rd:
An attempt to clarify issues like #3450 with respect to accesing `target` rather than `currentTarget`.

Property name change would probably greatly benefit here, but would  be breaking.


staff0rd:
That's this issue; #3416

PR #3474, opened 2016-12-16
fix for invalid texture

GoodBoyDigital:
does what it says on the tin!
Please give this a test!

### DIFF
--- a/src/core/renderers/webgl/TextureManager.js
+++ b/src/core/renderers/webgl/TextureManager.js
@@ -83,7 +83,7 @@ export default class TextureManager
         // if the location is undefined then this may have been called by n event.
         // this being the case the texture may already be bound to a slot. As a texture can only be bound once
         // we need to find its current location if it exists.
-        if (location == null)
+        if (location === undefined)
         {
             location = 0;
 
--- a/src/core/renderers/webgl/TextureManager.js
+++ b/src/core/renderers/webgl/TextureManager.js
@@ -83,13 +83,13 @@ export default class TextureManager
         // if the location is undefined then this may have been called by n event.
         // this being the case the texture may already be bound to a slot. As a texture can only be bound once
         // we need to find its current location if it exists.
-        if (location === undefined)
+        if (location == null)
         {
             location = 0;
 
             // TODO maybe we can use texture bound ids later on...
             // check if texture is already bound..
-            for (let i = 0; i < boundTextures.length; i++)
+            for (let i = 0; i < boundTextures.length; ++i)
             {
                 if (boundTextures[i] === texture)
                 {
--- a/src/core/renderers/webgl/TextureManager.js
+++ b/src/core/renderers/webgl/TextureManager.js
@@ -68,7 +68,6 @@ export default class TextureManager
     {
         // assume it good!
         // texture = texture.baseTexture || texture;
-        location = location || 0;
 
         const gl = this.gl;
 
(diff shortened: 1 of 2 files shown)


GoodBoyDigital:
Totally agree! Just wanted to see if this solves the flickering issue first.

finscn:
In my MacBook , 
the   `gl.getParameter(gl.MAX_TEXTURE_IMAGE_UNITS)`==16 ,  `settings.SPRITE_MAX_TEXTURES`==32 .

So I think the correct way is using these codes : 
```
const glMaxTextures = gl.getParameter(gl.MAX_TEXTURE_IMAGE_UNITS);
const maxTextures = Math.min(glMaxTextures, settings.SPRITE_MAX_TEXTURES);
```

.

finscn:
I test my case , still has 
```
WebGL: INVALID_ENUM: activeTexture: texture unit out of range
```

finscn:
I think the key point is not  the number of `maxTextures`.

It's  about  https://github.com/pixijs/pixi.js/issues/3373#issuecomment-265764333 

* why group.ids[j] could be -1 ? 
* why all `t._enabled === TICK` ?  https://github.com/pixijs/pixi.js/blob/dev/src/core/sprites/webgl/SpriteRenderer.js#L275-L295

GoodBoyDigital:
Interesting! Please can you share the code for your case?

finscn:
@GoodBoyDigital  , my test case is very complex , it's a real game project.

It will render many sprites, meshes, graphics (include rect circle etc.)   in one game tick.

finscn:
after  https://github.com/pixijs/pixi.js/pull/3474/files#diff-b87a2a9840fecac0080b87cdd72de184R93 
you miss a  `break;` , I think

GoodBoyDigital:
Good catch! Will tweak when I get home :)

finscn:
@GoodBoyDigital  , the new patch still didn't fix this bug.

finscn:
For my test case , this error message is throw at WebGLRenderer.prototype.bindTexture
```
            // bind the current texture
            this.boundTextures[location] = texture;

// ****** this line ******
            gl.activeTexture(gl.TEXTURE0 + location);

            gl.bindTexture(gl.TEXTURE_2D, glTexture.texture);
```

finscn:
#3474

GoodBoyDigital:
Thanks for letting me know @finscn ! 
The issue occurs when the same texture is bound to two different texture locations. This situation can result in -1 being passed in as the texture location. Getting closer though! :)

If you could share you code I would be happy to debug. That way I can get to the bottom of the issue much quicker! (maybe on a private channel if its sensitive stuff?)

finscn:
@GoodBoyDigital  , I've modified my reply . Your patch is OK for me.

GoodBoyDigital:
wooo! 👍

finscn:
@GoodBoyDigital  I think it's time to merge some PRs  which have been  approved , and release a new version of PIXI.js  as a gift for  New Year.

BTW : why there is an error of travis in this PR ? Could you fix it ?

And 

Happy New Year.

GoodBoyDigital:
@staff0rd fixed :D

bigtimebuddy:
@ivanpopelyshev do you have any follow-up here?

src/core/renderers/webgl/TextureManager.js
```diff
@@ -79,6 +78,29 @@ export default class TextureManager
             return null;
         }
 
+        const boundTextures = this.renderer.boundTextures;
+
+        // if the location is undefined then this may have been called by n event.
+        // this being the case the texture may already be bound to a slot. As a texture can only be bound once
+        // we need to find its current location if it exists.
+        if (location === undefined)
```
englercj: Should we do `== null` here to catch `undefined` and `null`?

src/core/renderers/webgl/TextureManager.js
```diff
@@ -79,6 +78,29 @@ export default class TextureManager
...
+        {
+            location = 0;
+
+            // TODO maybe we can use texture bound ids later on...
+            // check if texture is already bound..
+            for (let i = 0; i < boundTextures.length; i++)
```
englercj: `++i` :)

PR #3479, opened 2016-12-17
Remove  the needless checking for `orig`

finscn:
In Texture , the `orig` is always exist after new Texture object . 

 If `orig` is not accessible , there will be a bug in the App.
The developers must fix it , not ignore and return `0`.

### DIFF
--- a/src/core/textures/Texture.js
+++ b/src/core/textures/Texture.js
@@ -517,7 +517,7 @@ export default class Texture extends EventEmitter
      */
     get width()
     {
-        return this.orig ? this.orig.width : 0;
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
May be , ill think about it

themoonrat:
Indeed - if no orig is provided, it uses frame. And if no frame is provided, a frame of 0,0,1,1 is created. So something really bad must be happening to get into a state of no orig!

bigtimebuddy:
@finscn, please do not make PRs against **master**, only **dev**

PR #3480, opened 2016-12-17
Use `this._texture.*`to instead of`this.texture.*`

finscn:
In get/set width/height method , we don't need use  `this.texture` .

We could use `this._texture.*` to instead of `this.texture.*` , and it could offer a bit little performance gains.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -479,7 +479,7 @@ export default class Sprite extends Container
      */
     get width()
     {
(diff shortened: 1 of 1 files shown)


finscn:
There are many `foo.texture.*`  could be replace by  `foo._texture.*` in PIXI.
I will find , modify & test them  one by one , then pull some PRs later.

I don't want to  put  them into one PR, it's dangerous.

bigtimebuddy:
@finscn, please do not make PRs against **master**, only **dev**

PR #3487, opened 2016-12-19
An useable Plane with some new features

finscn:
The current Plan of PIXI has some problems (see :  https://github.com/pixijs/pixi.js/issues/3456 ).

So, I create a better one.
The demo ： http://fattyboy.cn/anim/     （the Plane Sprite is the bottom flight)

New Features:
* Users could do mesh transform via  `plane.points` easily (it's a 2d array).
* Support texture changing. ( I could create animation via my Animation easily , see :  https://github.com/pixijs/pixi.js/issues/3469 )
* Support `anchor`

### DIFF
--- a/src/mesh/Plane.js
+++ b/src/mesh/Plane.js
@@ -72,10 +72,12 @@ export default class Plane extends Mesh
         for (let i = 0; i < this.verticesY; i++)
         {
             const row = [];
+
             this.points.push(row);
             for (let j = 0; j < this.verticesX; j++)
             {
                 const point = new Point(j, i);
+
                 point.originalX = point.x;
                 point.originalY = point.y;
                 row.push(point);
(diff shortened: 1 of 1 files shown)


finscn:
@ivanpopelyshev  , I know you are trying to create a new Plane & Rope , but I need them recently ， So I do it by myself.

I hope my PR could give you some inspirations.

Expect you to bring us better works.

antag0n1st:
Why this did not merge ?
@ivanpopelyshev  can the mesh.Plane be manipulated like the mesh.Rope ?
The docs for the mesh.Plane are still not good!

ivanpopelyshev:
That mesh works in heaven: https://github.com/gameofbombs/pixi-heaven/tree/master/src/mesh

You can add heaven plugin to your app and try it.

PR #3491, opened 2016-12-20
Add `anchor` to Mesh.js

finscn:
Mesh always has texture , so it should has `anchor`.

If it has `anchor`, it will needs  `_onAnchorUpdate` `refresh()` `_ready` too.
Mesh is an abstract class ， so it doesn't  use them , just define them.

### DIFF
--- a/src/mesh/Mesh.js
+++ b/src/mesh/Mesh.js
@@ -31,6 +31,17 @@ export default class Mesh extends core.Container
         this._texture = null;
 
         /**
+         * The anchor sets the origin point of the texture.
+         * The default is 0,0 this means the texture's origin is the top left
+         * Setting the anchor to 0.5,0.5 means the texture's origin is centered
+         * Setting the anchor to 1,1 would mean the texture's origin point will be the bottom right corner
+         *
+         * @member {PIXI.ObservablePoint}
+         * @private
+         */
+        this._anchor = new core.ObservablePoint(this._onAnchorUpdate, this);
+
+        /**
          * The Uvs of the Mesh
          *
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
"anchor" is not about texture. "uvs" are. Wait until I implement texture rotations, do some workaround for your project temporarily.

There's no sence in adding anchor to Mesh, unless we are talking about polygon-packing

finscn:
I make a mesh as animation ， like a sprite ， so need anchor.

In my opinion , mesh like a sprite ,  it's a DisplayObject with texture , sprite has anchor , mesh should has too.

ivanpopelyshev:
No, it doesnt. In a way, its not a mesh, its a sprite. Please copy Mesh class and do your R&D there. Adding anchor in PIXI.mesh will only bring us more problems.

finscn:
I've add anchor in  https://github.com/finscn/pixi.js/blob/master/src/mesh/Plane.js  
There is no problem in my game. 

And there is a simple demo http://fattyboy.cn/anim

 In mesh , the anchor is like  offsetX/offsetY , just "move" vertices  enough.

ivanpopelyshev:
It is a problem, you just dont understand what "Mesh" means. There's no way of applying anchor to a generic mesh, unless its a Plane or polygon-packed Sprite

englercj:
Anchor is a quick way of changing UVs for quads, that is squares. Mesh unfortunately doesn't necessarily (or even usually) a quad and anchor doesn't really make sense here. Just modify the UVs directly.

Issue #3485, opened 2016-12-19
cacheAsBitmap not working properly for nested elements


sivael:
Hi.

I'm working on a component using pixi and am encountering a weird issue as illustrated here:

https://codepen.io/sivael/pen/rWPwpg?editors=0011

Basically when I do a pixi equivalent of:
```
<scene>
  <g>
    <g x:10 y:10 ::cachethis>
      <roundedBox />
      <text>1,1</text>
    </g>
  </g>
</scene>
```
All works well, and there's no issues with cacheAsBitmap. (the green boxes in example)

BUT when I do:
```
<scene>
  <g>
    <g x:10 y:10>
      <g  ::cachethis>
        <roundedBox />
        <text>1,1</text>
      </g>
    </g>
  </g>
</scene>
```

Then anything beyond the first box doesn't render properly(the red boxes example)

It looks like the function call to:
` m.invert()`
caused my 
`m.tx = -3999.5`

And that prevented the proper rendering of the component.

Note, however, that canvas has the following code:
```    m.invert();

    m.tx -= bounds.x;
    m.ty -= bounds.y;
```

while webGL has this:
```var m = _tempMatrix;

    m.tx = -bounds.x;
    m.ty = -bounds.y;
```

Now the implementations are probably different, and changing canvas to
```    m.invert();

    m.tx = -bounds.x;
    m.ty = -bounds.y;
```
actually breaks the rendering of green elements while fixing reds.

Any ideas?

FFankias:
https://github.com/pixijs/pixi.js/pull/3366

sivael:
It doesn't fix the issue - that change only makes the currently working properly elements stop working.

sivael:
#3494 I've made a PR to fix all use cases.

PR #3494, opened 2016-12-21
Bugfix for #3356 and #3485

sivael:
https://github.com/pixijs/pixi.js/pull/3366
^ This PR is wrong, it takes into the account only one of the two situations (breaks half of the use cases while fixing the other half)

After changing from worldTransform to localTransform, caching is proper for all cases.

### DIFF
--- a/src/extras/cacheAsBitmap.js
+++ b/src/extras/cacheAsBitmap.js
@@ -280,7 +280,7 @@ DisplayObject.prototype._initCachedDisplayObjectCanvas = function _initCachedDis
     // need to set //
     const m = _tempMatrix;
 
-    this.transform.worldTransform.copy(m);
+    this.transform.localTransform.copy(m);
     m.invert();
 
     m.tx -= bounds.x;


staff0rd:
What effect (if any) does this have on [this pen](http://codepen.io/staff0rd/pen/WReJdo?editors=0010)? 

re: #3505

sivael:
I would assume it'd be fixed.
I will test it once i get to work today.

sivael:
@staff0rd 
No, it will not fix it. It only touches the CacheAsBitmap that is used in Canvas.
WebGL must have a different error as in WebGL caching works in all my use cases without changes, and is connected to the masking You're doing.

finscn:
<Flag> merged in my fork.

staff0rd:
@sivael any chance you have any code snippets for your use cases? I'd be interested in writing a suite of tests to cover them and other CacheAsBitmap/Mask issues to help with reproduction and regression.

sivael:
https://github.com/pixijs/pixi.js/issues/3485
This has a test for the different nesting issue that caused my issue initially.
The one with masking is WebGL specific and as I am not using WebGL in my project (I can't - need to use canvas) it's difficult for me to see the reason.

I'll look at it a bit more once I start on the project at work.

PR #3501, opened 2016-12-22
Fix #3373  : WebGL: INVALID_ENUM: activeTexture: texture unit out of range

### LINKED ISSUE #3373: pixi.js:15891 WebGL: INVALID_ENUM: activeTexture: texture unit out of range
?

finscn:
The bug https://github.com/pixijs/pixi.js/issues/3373 .

The key of this bug  is that in WebGLRenderer.js : 
```
                if (rendererBoundTextures[group.ids[_j]] !== currentTexture) {
                  this.renderer.bindTexture(currentTexture, group.ids[_j], true);
                }
```
Sometimes, the  `group.ids[_j] === -1`.

I found the reason is 
```
                        for (let j = 0; j < MAX_TEXTURES; ++j)
                        {
                            const tIndex = (j + TEXTURE_TICK) % MAX_TEXTURES;

                            const t = boundTextures[tIndex];

                            if (t._enabled !== TICK)
                            {
                                TEXTURE_TICK++;

                                t._virtalBoundId = -1;

                                nextTexture._virtalBoundId = tIndex;

                                boundTextures[tIndex] = nextTexture;
                                break;
                            }
                        }
```
The code above can't ensure find a new `tIndex` for  nextTexture._virtalBoundId.
In this situation , We should call TICK++ & find again.

### DIFF
--- a/src/core/sprites/webgl/SpriteRenderer.js
+++ b/src/core/sprites/webgl/SpriteRenderer.js
@@ -291,6 +291,15 @@ export default class SpriteRenderer extends ObjectRenderer
                                 boundTextures[tIndex] = nextTexture;
                                 break;
                             }
+                            /**
+                             * If can't find the correct new `tIndex` in current for-loop,
+                             *     then TICK++ , and find again.
+                             */
+                            if (j + 1 === MAX_TEXTURES)
+                            {
+                                TICK++;
+                                j = -1;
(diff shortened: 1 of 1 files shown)


finscn:
This patch is just a hack.    Maybe The #3474  is the best way .

Issue #3116, opened 2016-10-15
Update to the latest loader (v2) 


englercj:
There are breaking changes in v2, but it fixes progress issues pixi users are experiencing.

PR #3512, opened 2016-12-26
Update to Resource loader v2; fixes #2876

### LINKED ISSUE #2876: loader progress over 100% in v4
occur status : 
- use spritesheets ( generated by texture packer )
- use browser-sync open html (works normal if open html directly.)
- v4  ( works normal at v3 )

example code : 
[pixi-loader-test.zip](https://github.com/pixijs/pixi.js/files/442093/pixi-loader-test.zip)

step : 
  1. npm install
  2. browser-sync start --server
  3. check consloe.log

englercj:
I put in backward compat handling, or deprecation warning, for all v2 changes except for one:

- If a middleware adds more resources, it must pass in the parent resource in options for .add().

I've updated our middleware to follow this rule, but if there are other ones floating around out there they will need to do this for child resources as well. Otherwise, they will get an error.

Also closes: #3116

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "ismobilejs": "^0.4.0",
     "object-assign": "^4.0.1",
     "pixi-gl-core": "^1.0.3",
-    "resource-loader": "^1.8.0"
+    "resource-loader": "^2.0.3"
   },
   "devDependencies": {
     "babel-cli": "^6.18.0",
--- a/src/deprecation.js
+++ b/src/deprecation.js
@@ -4,6 +4,7 @@ import * as particles from './particles';
 import * as extras from './extras';
 import * as filters from './filters';
 import * as prepare from './prepare';
+import * as loaders from './loaders';
 
 // provide method to give a stack track for warnings
 // useful for tracking-down where deprecated methods/properties/classes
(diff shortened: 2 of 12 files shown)


ivanpopelyshev:
Something wrong with travis, I suppose?

bigtimebuddy:
@englercj tests seem to be failing, could you run `npm run test:debug` locally to see what's going on?

englercj:
Sorry about that, was an issue with an import. Corrected it!

bigtimebuddy:
@englercj Could of questions:
* Why are the deprecations (e.g., `isAudio`, `isVideo`, `before`, etc) not living in **resource-loader**? IMO, that seems like a more appropriate place for it than in Pixi's deprecation layer. 
* Could you describe the strategy for EE? This PR allows for two alternative APIs and does not actually deprecate-warn the EE methods on Loader. We should not move to signals until v5. If this is true, we should introduce a PR at the next minor version to provide deprecation warnings for EE methods and give people time to upgrade. Also, we would need to provide PRs to adjust any documentation, examples, etc which utilize events of Loader.

englercj:
> Why are the deprecations (e.g., isAudio, isVideo, before, etc) not living in resource-loader? IMO, that seems like a more appropriate place for it than in Pixi's deprecation layer.

No particular reason, they can be moved.

> Could you describe the strategy for EE? 

For all of pixi? Haven't decided yet, but IMO it should be a hard break with a backwards-compat plugin that adds back in EE3 for people who still need it. I didn't deprecate warn the EE methods on loader, because having them there makes it consistent across pixi. I don't think it makes sense for some features to be signals and others to be EE3. Either we do all or nothing.

bigtimebuddy:
Makes sense, thanks for clarifying. 

I believe resource loader has unit tests, right? Is there any opportunity with this PR to add a unit test for pixi loader integration? Loading a texture, font, etc? Even something super minimal.

Otherwise looks good

englercj:
Yeah definitely, I'll do that tomorrow. Would be really good to test the middlewares we have and ensure the  API doesn't break.

englercj:
Rebased to latest dev, and adding some simple tests.

PR #3529, opened 2017-01-02
Improved documentation of the destroy method

themoonrat:
Documents the .texture and .baseTexture parameters for the destroy method settings object. They may not be used for the class itself, but could get propagated down to children.
Fixes https://github.com/pixijs/pixi.js/issues/3407

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -523,6 +523,10 @@ export default class Container extends DisplayObject
      *  have been set to that value
      * @param {boolean} [options.children=false] - if set to true, all the children will have their destroy
      *  method called as well. 'options' will be passed on to those calls.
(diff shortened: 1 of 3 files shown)


themoonrat:
@staff0rd how about

     * @param {boolean} [options.texture=false] - Only used for child Sprites if options.children is set to true
     *  Should it destroy the texture of the child sprite
     * @param {boolean} [options.baseTexture=false] - Only used for child Sprites if options.children is set to true
     *  Should it destroy the base texture of the child sprite

bigtimebuddy:
@staff0rd look good?

staff0rd:
all good

GoodBoyDigital:
👍

englercj:
Might be best to move these into a typedef for DestroyOptions or something, then use that everywhere. Instead of maintaining separate lists on each method impl.

PR #3530, opened 2017-01-02
Fixes CanvasRenderer getting stuck at the last renderTexture resolution

themoonrat:
Fixes https://github.com/pixijs/pixi.js/issues/3135 and https://github.com/pixijs/pixi.js/issues/3359
Still allows requirement for dynamic changing of resolution which introduced this bug in the first place: https://github.com/pixijs/pixi.js/pull/2870


GoodBoyDigital:
👍

PR #3531, opened 2017-01-02
Use `tex` instead of `this.texture`

finscn:
In `update` method , there is already  `tex` , we shouldn't use `this.texture` any more.

### DIFF
--- a/src/extras/TextureTransform.js
+++ b/src/extras/TextureTransform.js
@@ -74,7 +74,7 @@ export default class TextureTransform {
      */
     update(forceUpdate)
     {
-        const tex = this.texture;
+        const tex = this._texture;
 
         if (!tex || !tex.valid)
         {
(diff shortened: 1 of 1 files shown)


themoonrat:
I don't see the need to cache a short and well described member variable into an even shorter, less clear local variable that isn't used much. Can't everything be this._texture?

bigtimebuddy:
@finscn I have a request. You have been making a lot of code cleanup pull requests, like this one. It's very helpful and we appreciate it, but it generates a lot of work for the team to field every PR. We would be very interested in seeing a larger PR of trivial changes rather than many trivial PRs.

finscn:
@bigtimebuddy   , I know your mean.
But  I said at  https://github.com/pixijs/pixi.js/pull/3480#issuecomment-267782119  .
I'm afraid that if I change many file in different modules ， it's too dangerous.

finscn:
@themoonrat  ， in `update` method of  TextureTransform.js ,  there are many places need `this._texture` , So I think use a short name is worth.  
But I don't like  `tex` ,  in other files of PIXI.js , the `texture` is used generally.

bigtimebuddy:
@finscn I mostly disagree that it is too dangerous. Your changes, like the _texture fixes, would be better as one thorough PR rather than one at a time. We have to balance risk with time required to review each PR. Please consider bundling changes that are similar.

finscn:
@bigtimebuddy  , I see.
In the future , I'll make a larger PR for the similar trivial PRs as possible as I can. ( If I won't find the issues at the same time , I can't pull the larger PR ).

But  I hope this issue and  issue https://github.com/pixijs/pixi.js/pull/3532 could be merged.

staff0rd:
@themoonrat I agree, but take a look at the file and you'll see that without the abbreviations (more than just `tex`) those fomulas are probably way too verbose.  I gave a refactor a shot, wasn't pretty - wanna give it a try?

GoodBoyDigital:
👍

Issue #3419, opened 2016-12-07
posssible InteractionManager propagation bug


uorbe001:
We have stumbled upon what I think is a bug, but might be intended behaviour for some reason I don't comprehend. I have prepared a basic example that I've used to replicate the issue:

```js
var renderer = PIXI.autoDetectRenderer(800, 600,{backgroundColor : 0x1099bb});
document.body.appendChild(renderer.view);

// create the root of the scene graph
var stage = new PIXI.Container();

var parent = new PIXI.Container();
parent.on("click", onStageDown);
parent.interactive = true;
stage.addChild(parent);

var sprite = PIXI.Sprite.fromImage('required/assets/basics/bunny.png');

sprite.position.set(230,264);
sprite.interactive = true;
sprite.on('click', onSpriteDown);

parent.addChild(sprite);

var shadow = PIXI.Sprite.fromImage('required/assets/basics/bunny.png');
shadow.position.set(230,264);
parent.addChild(shadow);

function onSpriteDown (eventData) {
    console.log("Sprite clicked");
}

function onStageDown (eventData) {
    console.log("Stage clicked");
}
// start animating
animate();

function animate() {

    requestAnimationFrame(animate);

    // render the root container
    renderer.render(stage);
}
```

With a structure similar to this, I would expect both the `onSpriteDown` and `onStageDown` listeners to be called, but what I get with that code (tested with pixi 4 in the examples pages) is that only the `onStageDown` is getting called. 

I've debugged this and I know that when [`processInteractive`](https://github.com/pixijs/pixi.js/blob/dev/src/interaction/InteractionManager.js#L819) iterates over the children, when it reaches `shadow` this element is considered as hit, and by the time it processes `sprite` hitTest is set to false, so it never actually tries to test it and call its function.

Is this expected behaviour? If it isn't, I can probably find some time to fix it 😇

staff0rd:
Pen; https://codepen.io/staff0rd/pen/YpjozL

staff0rd:
@uorbe001 I've added some tests that display this functionality, and I agree it seems inconsistent.  Waiting on feedback for #3535 to determine whether this is expected functionality or not.

uorbe001:
@staff0rd thanks for doing that! I'm still waiting for confirmation before spening time trying to fix it.

staff0rd:
Heya @uorbe001, team members agree that the expectation in #3535 is correct.  Please feel free to commit a fix that makes that test pass.

[...]

Issue #2921, opened 2016-09-06
What could cause event propagation to fail?


kaansoral:
Hi

I've noticed a propagation issue today, I've been debugging it for hours, it has consumed me, I've inspected the PIXI source, tested various things, but no luck

I've tested various PIXI versions, including the latest dev, they also respond the same. I've also tried to reproduce the issue, however I couldn't

I suspect pixi-display could be causing the issue, but it seems unlikely - however, since the interaction code is complicated (I couldn't grasp the bubbling logic in a short span of time), it might be so

**The issue:** There is a character, under the character, there is a monster, I right click the character, the event isn't stopped, yet the right click doesn't bubble down, .stopped is false, the event just doesn't bubble :(

I've tested a normal click too, no matter what I do, I can't make the event bubble once a callback is fired (so basically, normally, when the event.stopPropagation isn't called, the event should bubble down on it's own)

ivanpopelyshev:
That's not how its bubbling. I thought its bubbling into the parent? You can't just ask character to ignore the event. If its interactive, it will receive all events, and stop his "brothers" from doing that.

kaansoral:
Very basically, there are 2 objects on the map

When one object overlaps the other, I want the event to bubble down, it doesn't bubble down

They are on the same displayGroup, they are different/separate entities

I just managed to reproduce the issue by the way, simply, when you overlap 2 Sprite's that are on the same displayList, the event bubbles down to the map, yet the event doesn't trigger on both Sprite's, as it should trigger

ivanpopelyshev:
Can you prepare an example that we can put to http://pixijs.github.io/examples/index.html?s=basics&f=container.js&title=Container&v=dev#/basics/basic.js ?

kaansoral:
For example, on this example: http://pixijs.github.io/examples/index.html?s=basics&f=container.js&title=Container&v=dev#/display/zorder.js

There is no `event.stopPropagation` on `onDragStart` - yet the event stops, I think, when you drag(or specifically, click) something on an overlapping point, you should be dragging multiple entities at the same time

ivanpopelyshev:
event bubbles to parents, not to brothers. Its not possible to drag everything :( @GoodBoyDigital any ideas on that?

kaansoral:
Well that clears things up, thank you, as I suspected, it wasn't a pixi-display issue after all, just re-reproduced it without pixi-display, and as you said, there is no reproduction needed, since I wanted/expected sibling propagation :)

[...]

Issue #2364, opened 2016-02-19
[3.0.9] Interaction event triggered on multiple overlapping objects


alebianco:
http://jsfiddle.net/alebianco/2brjt234/

The example shows how two overlapping objects (red square, yellow circle) could be triggered at the same time with a click. This happens when the two objects are added to sibling parents, if they are part of the same parent tree, everything seems to work as expected.

This is definitely a change of behaviour from v3.0.8

alebianco:
bump, no feedback on this?

mathieuanthoine:
it seems more a bug for me than a behavior change. What is the real interest of this behavior ? It completely breaks UI principles.

Here's my post about this issue: http://www.html5gamedevs.com/topic/21164-mouse-behavior-change-between-308-and-higher/

GoodBoyDigital:
yes, seems like a bug! Pretty busy at the mo but will get to it soon :+1:

mathieuanthoine:
Great ! Thank you Mat :)

FNevesFMQ:
Can we expect to see this bug fixed in the next pixi iteration?

alebianco:
I think it was fixed in 3.0.11

FNevesFMQ:
But last release isn't 3.0.10?

alebianco:
it's 3.0.11
https://github.com/pixijs/pixi.js/releases/tag/v3.0.11
but they forgot to bump the version in the code, so it says 3.0.10

FNevesFMQ:
Oh ok!
Thanks

staff0rd:
Looks solved per the tests in #3535

Issue #1901, opened 2015-06-16
v3 issues with interaction


nt-pch:
Hi folks, first... thanks for such a great library! 

I'm noticing differences in how v3 works in comparison to v2. I'm running Chrome 43.0.2357.124 on a Max OS 10.10.3.

I have a graphic red box in a container. The red box is listening to mouse events. http://jsfiddle.net/ar4pjetx/

1) If there is another non-interactive graphic on top of the red box and the red box's parent container has set ‘interactive = true', the red box no longer responds to events where the green box is overlapping, i.e., when mousing over both boxes, if the green box is present, then the red box thinks that the mouse is no longer over it. My expectation is that the box events should not be blocked by a non-interactive object on top of it. This is how v2 worked.
- If I remove line 32 (stage.interactive = true), then the events fire as expected.

2) When logging events in Chrome, it shows an object with ‘target: Graphics’, and logging the object’s given name shows ‘bg’. But on opening the event object, it shows an object with ‘target: Container’ and a name of ‘stage’ (the parent container). I have each event logging two things, the name, and the object. Open the object.target and you can see the name is different inside.
- If I remove line 32 (stage.interactive = true), then the logging of objects shows up as expected. Both the object overview and the individual items in the object show both ‘target: Graphics’ and ‘name: “bg”’.

3) The red box’s ‘over’ event fires the first time this fiddle is run even though the mouse is not over the red box. 

4) The red box’s ‘out’ event fires the first time the mouse moves into the canvas area even though the mouse was never over the red box. (This is most likely because the ‘over’ event was incorrectly fired so it probably thinks the mouse is over the red box to begin with.)

5) Right clicking causes the ‘over’ and ‘out’ events to fire when the right click content box shows up. Expectation: is that the context box is not part of the canvas and should not trigger events.

endel:
Also related with https://github.com/GoodBoyDigital/pixi.js/issues/1837 and https://github.com/GoodBoyDigital/pixi.js/issues/1884

There is a patch/pull-request for this here: https://github.com/GoodBoyDigital/pixi.js/pull/1894

stephenweaver:
added a comment to the PR https://github.com/pixijs/pixi.js/pull/1894
this issues is not resolved with that fix. I think if we can figure out why targetTouches includes all touches on the canvas rather than just the pixi object, the events will get sent to the correct objects. maybe?

englercj:
Is this issue still active on latest v4 `dev`?

[...]

PR #3535, opened 2017-01-03
Added overlapping children tests

staff0rd:
I've added the following tests to help document the expectations of the InteractionManager specific to when the scene includes children that overlap.  The tests set up a scene similar to [this pen](https://codepen.io/staff0rd/pen/MJWJdy?editors=0010) and then fire clicks at various points on it.

I have left one test failing because it _feels_ off, considering the expectation is valid and passes when the parent is non-interactive.  I need some feedback regarding a decision on whether;

1. ~The expectation is wrong, and the test should be changed or;~
2. The expectation is correct, and the API needs to change ✔️ 

Upon which the following issues at least are then possibly fixed or are a step closer to resolution; #3419, #2921, #2364, #1901.

### When parent is non-interactive
* when both children are interactive
 :white_check_mark: should callback front child when clicking front child
 :white_check_mark: should callback front child when clicking overlap
 :white_check_mark: should callback behind child when clicking behind child
* when front child is non-interactive
 :white_check_mark: should not callback when clicking front child
 :white_check_mark: should callback behind child when clicking overlap
 :white_check_mark: should callback behind child when clicking behind child
* when behind child is non-interactive
 :white_check_mark: should callback front child when clicking front child
 :white_check_mark: should callback front child when clicking overlap
 :white_check_mark: should not callback when clicking behind child
### When parent is interactive
* when both children are interactive
 :white_check_mark: should callback parent and front child when clicking front child
 :white_check_mark: should callback parent and front child when clicking overlap
 :white_check_mark: should callback parent and behind child when clicking behind child
* when front child is non-interactive
 :white_check_mark: should callback parent when clicking front child
:x: should callback parent and behind child when clicking overlap
 :white_check_mark: should callback parent and behind child when clicking behind child
* when behind child is non-interactive
 :white_check_mark: should callback parent and front child when clicking front child
 :white_check_mark: should callback parent and front child when clicking overlap
 :white_check_mark: should callback parent when clicking behind child

### DIFF
--- a/test/interaction/InteractionManager.js
+++ b/test/interaction/InteractionManager.js
@@ -364,6 +364,7 @@ describe('PIXI.interaction.InteractionManager', function ()
                     expect(scene.parentCallback).to.have.been.calledOnce;
                 });
 
+                /* TODO: Fix #3596
                 it('should callback parent and behind child when clicking overlap', function ()
                 {
                     const stage = new PIXI.Container();
@@ -381,6 +382,7 @@ describe('PIXI.interaction.InteractionManager', function ()
                     expect(scene.frontChildCallback).to.not.have.been.called;
                     expect(scene.parentCallback).to.have.been.calledOnce;
                 });
+                */
 
                 it('should callback parent and behind child when clicking behind child', function ()
                 {
--- a/test/interaction/InteractionManager.js
+++ b/test/interaction/InteractionManager.js
@@ -81,4 +81,382 @@ describe('PIXI.interaction.InteractionManager', function ()
             expect(clickSpy).to.not.have.been.called;
         });
     });
+
+    describe('overlapping children', function ()
+    {
+        function getScene(callbackEventName)
+        {
+            const behindChild = new PIXI.Graphics();
+            const frontChild = new PIXI.Graphics();
+            const parent = new PIXI.Container();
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
2. Its correct and we have to change it.

ivanpopelyshev:
I think after those tests we can integrate pixi-display at last :)

ivanpopelyshev:
Currently element interaction is described by two vars: interactive and interactiveChildren.

Its not obvious that "interactive=false" element can be clicked and click bubbles to "interactive=true" parent.

GoodBoyDigital:
Interesting stuff! Is there a convention we can follow here to help us point in the right direction? I originally tried to follow the flash way. It might be better for us to follow something like the DOM interactive model? How do they deal with this issue?

GoodBoyDigital:
poke @staff0rd , is there a way we can roll this in without the failing test? I think this is a great set of tests. But not sure how easy the fix will be for the failing test right now. Maybe we could open up another issue for that? Cheers!

staff0rd:
@GoodBoyDigital I've commented out the failing test and opened #3596 for the fix.

GoodBoyDigital:
Three greats in one comment... I have been staring at git hub issues for too long :P

PR #3538, opened 2017-01-04
Fixes #3536

### LINKED ISSUE #3536: Bug with TilingSprite anchor behaviour
Changing the anchor of a TilingSprite causes the tiling to be misplaced in 4.x using WebGL.
In 3.x or 4.x with Canvas, the Tiling would stay the same. The sprite would be positioned and rotated relative to the anchor like one would expect.

Drawing a circle TilingSprite in 4.x with WebGL with anchor at (0.5,0.5):

Drawing a circle TilingSprite in 3.x (or 4.x with Canvas) with anchor at (0.5,0.5):

This behaviour is important when drawing textures such as a rope.

Behaviour in 4.x WebGL:

Comment/uncomment <script> tag to load 3.0.7 or use PIXI.CanvasRenderer to see how it used to be.
https://jsfiddle.net/Lfozgwfd/

staff0rd:
Fix for #3536

ping @ivanpopelyshev

### DIFF
--- a/src/extras/TilingSprite.js
+++ b/src/extras/TilingSprite.js
@@ -73,7 +73,7 @@ export default class TilingSprite extends core.Sprite
 
         /**
          * Whether or not anchor affects uvs
-         * 
+         *
          * @member{boolean}
          * @default false
          */
--- a/src/extras/TilingSprite.js
+++ b/src/extras/TilingSprite.js
@@ -74,7 +74,7 @@ export default class TilingSprite extends core.Sprite
         /**
          * Whether or not anchor affects uvs
          *
-         * @member{boolean}
+         * @member {boolean}
          * @default false
          */
         this.uvRespectAnchor = false;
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
To understand difference, paste this in example and try V3 vs latest V4:

```js
var renderer = PIXI.autoDetectRenderer(800, 600);

// create a texture from an image path
var texture = PIXI.Texture.fromImage('required/assets/p2.jpeg');

/* create a tiling sprite ...
 * requires a texture, a width and a height
 * in WebGL the image size should preferably be a power of two
 */
var tilingSprite = new PIXI.extras.TilingSprite(texture, renderer.width/2, renderer.height/2);
stage.addChild(tilingSprite);

var count = 0;

animate();

function animate() {

    count += 0.005;

    tilingSprite.anchor.set(-Math.abs(1 - count%2.0))
    // render the root container
    renderer.render(stage);

}
```

GoodBoyDigital:
nice work guys!

PR #3543, opened 2017-01-05
Ensure that polyfills are import first

themoonrat:
Fixes https://github.com/pixijs/pixi.js/issues/3291

Exporting the polyfills is a way of making sure they are also imported straight away.


GoodBoyDigital:
👍

PR #3545, opened 2017-01-05
added check for root

### LINKED ISSUE #3505: Problem with render when cacheAsBitmap and mask is used together
If inside container is element, which have set mask on it and this element position is not 0,0, then this element will be not rendered if we set '**cacheAsBitmap**' to true on the container wherein the element is located.

Link to example: http://codepen.io/kiture/pen/bByGKg

If we remove this line: 
_container.cacheAsBitmap = true;_
Then all works correctly.

GoodBoyDigital:
fixes #3505

### DIFF
--- a/src/core/renderers/webgl/managers/MaskManager.js
+++ b/src/core/renderers/webgl/managers/MaskManager.js
@@ -40,6 +40,7 @@ export default class MaskManager extends WebGLManager
         }
         else if (this.enableScissor
             && !this.scissor
+            && this.renderer._activeRenderTarget.root
             && !this.renderer.stencilManager.stencilMaskStack.length
             && maskData.isFastRect())
         {
--- a/src/core/renderers/webgl/managers/MaskManager.js
+++ b/src/core/renderers/webgl/managers/MaskManager.js
@@ -34,6 +34,10 @@ export default class MaskManager extends WebGLManager
      */
     pushMask(target, maskData)
     {
+        // TODO the root check means scissor rect will not
+        // be used on render textures more info here:
+        // https://github.com/pixijs/pixi.js/pull/3545
+
         if (maskData.texture)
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
Thanks for the feedback dude! Not sure what problem you are solving here? The root property changes depending on if you are rendering to a frameBuffer or not?

ivanpopelyshev:
No, the thing is cacheAsBitmap and masking are changing renderTarget.transform but scissor doesn't know about that. As a result I have to use a shift somewhere:

```js
shift.x = Math.round((projectionMatrix.tx + 1) / 2 * rt.sourceFrame.width);
shift.y = Math.round((projectionMatrix.ty + 1) / 2 * rt.sourceFrame.height);
```

GoodBoyDigital:
ScissorMasking is disabled on anything other than root, Mainly as renderTargets use scissor rects to crop where they render too. With that in mind the shift should not be required? 

This issue should not have anything todo with coordinates but more when to use fastRect? Maybe what you mention is a separate issue from this?

ivanpopelyshev:
May be its separate, but I'd like to make a fix for it later. Add "enableScissor" to renderTarget please.

GoodBoyDigital:
Until we have a new fix I believe this PR is good to go. Adding extra variables for PRs that will happen later seems a little redundant?

If you want to make a new PR later down the line that would be ace. Then at that point we can add the extra variables and refactor as necessary.

GoodBoyDigital:
Does this PR break your existing code?

ivanpopelyshev:
Please make a commentary in the code about it and I'll approve it.

GoodBoyDigital:
righto skippy! @ivanpopelyshev note added, you happy to roll this one in? 👍

PR #3547, opened 2017-01-05
Improved documentation

themoonrat:
Class member documentation improved via removal of unnecessary @memberofs - move members now show up in the docs.
Class getter/setter documentation improved using technique above, and also by removing documentation for setters. Documentation was only showing the setter docs, whereas the getter is where the important information is. JSDoc maintainers themselves recommend just doccing the getter https://github.com/jsdoc3/jsdoc/issues/973
Documented a few classes to let devs know that they are automatically created by PIXI and added to renderer.plugins, like interactionManager and prepare


bigtimebuddy:
Great work @themoonrat. I wonder if you could make a new branch/PR on pixi.js (instead of your own fork) so that we could actually preview the docs via Travis?

themoonrat:
done :)
https://github.com/pixijs/pixi.js/pull/3548

PR #3551, opened 2017-01-06
Creates new PIXI.Application class

bigtimebuddy:
## Overview

Setting up a new PIXI scene often requires a few steps: setting up a `requestAnimationFrame` or `Ticker`, creating a `PIXI.Container` and creating the renderer. I would like to improve this experience by creating a very simple API for setting up PIXI, as well as create something that is extendable and object-oriented.

```js
// creates the renderer, root and ticker
 
// Add the view to the DOM

// ex, add display objects
const graphic = new PIXI.Graphics();
graphic.beginFill(0xff0000).drawRect(0, 0, 100, 100);
app.stage.addChild(graphic);
```

### Documentation

http://pixijs.download/application/docs/PIXI.Application.html

### Example

https://jsfiddle.net/9s002r34/

### DIFF
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -1,0 +1,90 @@
+import { autoDetectRenderer } from './autoDetectRenderer';
+import Container from './display/Container';
+import Ticker from './ticker/Ticker';
+
+/**
+ * Convenience class to create a new PIXI application.
+ * This class automatically creates the renderer, ticker
+ * and root container.
+ *
+ * @example
+ * // Create the application
+ * const app = new PIXI.Application();
+ *
+ * // Add the view to the DOM
+ * document.body.appendChild(app.view);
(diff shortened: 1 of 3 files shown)


src/core/Application.js
```diff
@@ -0,0 +1,90 @@
...
+     */
+    start()
+    {
+        this.ticker.start();
+    }
+
+    /**
+     * Read-only reference to the canvas element.
+     * @member {HTMLCanvasElement}
+     */
```
themoonrat: Add @readonly jsdoc tag here

bigtimebuddy: Done, good call.

ivanpopelyshev:
Something about ticker... can you check that shared ticker wont be used in this config?

PR #3554, opened 2017-01-06
Do check as same as `calculateVertices`

finscn:
I already ask a question about this  https://github.com/pixijs/pixi.js/issues/3408  , but no one reply .
I think opening PR is better than asking question.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -123,7 +123,7 @@ export default class Sprite extends Container
 
         this._transformTrimmedID = -1;
         this._textureTrimmedID = -1;
-        
+
         /**
          * Plugin that is responsible for rendering this element.
          * Allows to customize the rendering process without overriding '_renderWebGL' & '_renderCanvas' methods.
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -237,14 +237,13 @@ export default class Sprite extends Container
      */
     calculateTrimmedVertices()
     {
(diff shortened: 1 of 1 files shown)


finscn:
@staff0rd  ， I can't understand what's your mean.
What should I do to this PR ?

staff0rd:
Instead of this;
~~~javascript
if (this._transformID === this.transform._worldID && this._textureID === this._texture._updateID) { return; }
...
if (this._transformID === this.transform._worldID && this._textureID === this._texture._updateID) { return; }
~~~

Self-documenting and reduced;

~~~javascript
isAlreadyUpdated() {
    return this._transformID === this.transform._worldID && this._textureID === this._texture._updateID;
}

if (isAlreadyUpdated()) { return; }
...
if (isAlreadyUpdated()) { return; }
~~~

finscn:
@staff0rd but it will create a new function call. In this place I think performance is very important.

staff0rd:
Ah I see - I wasn't aware javascript had a cost associated with function declaration.  TIL.

englercj:
If you're lucky it gets inlined, but not guaranteed. That is one of the shitty parts of working with such a dynamic lang.

englercj:
I want @ivanpopelyshev to see this before we merge, since he did a lot of trim work.

ivanpopelyshev:
I don't think we can use same temporary variables in two different functions. If you update one of them, second wont be updated.

finscn:
@ivanpopelyshev  ， you're right . I changed them.

GoodBoyDigital:
good catch @ivanpopelyshev !

PR #3563, opened 2017-01-10
High density displays interaction had wrong point calculated.

sivael:
InteractionManager on higher resolution densities was thinking it pointed at 1/resolution point instead of target point.
After this change it will point properly.
Tested on:
Galaxy Tab S2
iPad Air 2
Chrome emulator
Samsung J7


sivael:
Invalidated by #3530

PR #3564, opened 2017-01-10
Fix for resolution in high density devices.

sivael:
Resolution here changes when te canvasrenderer is used to render to texture and it's reset to 1.
This prevents the overwrite of desired resolution variable and uses it properly during device orientation change.

Previously after resize resolution would snap back to 1 (and stuff would be upscaled)


sivael:
invalidated by #3530

Issue #3476, opened 2016-12-17
Text align center doesn't work in v4.0.0


starInEcust:
I used Graphics.line to draw a border for the Text. The code is as follows
```
  const textText = new Text('0 0 0 0 0 0 0', {
    fontSize: 18,
    fill: 0x000000,
    align: 'center',
  });
  textText.y = -100;
  line.x = textText.x;
  line.y = textText.y;
  line.moveTo(0, 0);
  line.lineTo(textText.width, 0);
  line.lineTo(textText.width, textText.height);
  line.lineTo(textText.x, textText.height);
  line.lineTo(0, 0);
```
![image](https://cloud.githubusercontent.com/assets/5434131/21284786/8e4aecfe-c461-11e6-8cd3-3db4b789a7a0.png)

The rendered word is not centered. More blank areas to the right. And I do not know how to get the width of this blank area. I also found that Text.height is not the true height of the text,  but I don't know how to get the height of the blank area (above and below texts ).

version: 4.0.0

themoonrat:
If you set the x anchor of the text to 0.5 i believe you'll achieve what you want :)

starInEcust:
@themoonrat There are some fonts that do not appear in this case,
So I do not know when to use your method

themoonrat:
If you could create a jsfiddle or codepen demonstrating your issue that would help us a huge amount, thanks!

starInEcust:
@themoonrat 
The code only has different font-family

https://jsfiddle.net/starCoding/p49u34hd/4/

![image](https://cloud.githubusercontent.com/assets/5434131/21749528/f6361d2c-d5da-11e6-826b-6ec2171d16b9.png)

staff0rd:
I've opened PR #3565 as a possible solution.

Note that if that PR doesn't get accepted, you can just extend PIXI.Text to have the same functionality as the PR as i've done here with a `TrimmedText` class; https://codepen.io/staff0rd/pen/NdxGag?editors=0010

@starInEcust here's your fiddle pointing at the PR's changeset; https://jsfiddle.net/4h4otaeb/

PR #3565, opened 2017-01-10
Trimmed text

staff0rd:
Allows `PIXI.Text` to be trimmed of transparency for better layouts.

Opt in only;
~~~javascript
new PIXI.Text('my text', null, null, true);
~~~

https://codepen.io/staff0rd/pen/KaVdXZ?editors=0010

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -6,6 +6,7 @@ import { sign } from '../utils';
 import { TEXT_GRADIENT } from '../const';
 import settings from '../settings';
 import TextStyle from './TextStyle';
+import trimCanvas from '../utils/trimCanvas';
 
 const defaultDestroyOptions = {
     texture: true,
@@ -328,11 +329,11 @@ export default class Text extends Sprite
     {
         if (this._style.trim)
         {
-            const trimmed = this.getTrimmed(this.canvas);
+            const trimmed = trimCanvas(this.canvas);
 
-            this.canvas.width = trimmed.trimWidth;
-            this.canvas.height = trimmed.trimHeight;
-            this.context.putImageData(trimmed.trimmed, 0, 0);
+            this.canvas.width = trimmed.width;
+            this.canvas.height = trimmed.height;
+            this.context.putImageData(trimmed.data, 0, 0);
         }
 
         const texture = this._texture;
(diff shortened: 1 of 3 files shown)


src/core/text/Text.js
```diff
@@ -103,6 +104,13 @@ export default class Text extends Sprite
         this.style = style;
 
         this.localStyleID = -1;
+
+        /**
+         * Trim transparent borders
+         *
+         * @member {boolean}
+         */
+        this.trim = trim;
```
bigtimebuddy: How about `this.trim = trim || false;` or set `trim = false` as the default param to make sure defaults to an actually bool and not undefined.

bigtimebuddy:
I think it makes sense to add `trim` as a member of TextStyle. Seems similar to the `padding` option.

staff0rd:
Yeah - i'll move it in there if required - I wasn't sure of whether style was supposed be reflecting css style.  Would certainly be nicer pass via style however.  Will make the changes if there's a desire to merge, thought this might be too hacky.

bigtimebuddy:
@staff0rd I hear you. Text is already super slow, adding a couple `getImageData` calls and looping through the pixels seems even more expensive. Not sure if the juice is worth the squeeze in this case. Can you describe situation where trimming would be really important?

staff0rd:
1. `Text.texture`s take into account ascenders & descenders, even when the current text value does not use them, sometimes resulting in skewed padding and odd layout.  Most visible at larger font sizes, or with "old style" numbers (see number positions of [Raleway](https://fonts.google.com/specimen/Raleway))
2. `calculateFontProperties` is not super accurate, particularly on irregular fonts.
3.  Aligning any given text (with any given font) flush (to edges) on any given side without manually tweaking is not possible without trimming.
4. Bordering text values with equal padding for any font (see #3476)

ivanpopelyshev:
Yeah, I dont know whether to leave it there or move into TextStyle.

themoonrat:
In the unknown of deciding between it being in TextStyle or Text... I'll throw in a third option and say neither. Isn't this trim functionality more of a util that any sprite could use? Or a util that any generator of a texture could use?

staff0rd:
@themoonrat I agree that the actual trim functionality could at some stage be pulled out and go somewhere more generic, however I'm trying to keep the scope small to start with to encourage merging.

bigtimebuddy:
Love the utility approach. Seems this could be generalized and trimming is not specific to Text.

themoonrat:
@staff0rd I understand; I just feel uneasy about yet another optional constructor parameter for Text; I was a bit meh on passing a canvas through, but this one is a tipping balance from me I'm afraid! And it almost, but doesn't quite, fit within TextStyle remit.

bigtimebuddy:
I agree with @themoonrat on adding additional constructor args

bigtimebuddy:
How do you want to proceed with this @staff0rd?

staff0rd:
I've moved it into TextStyle so it looks like this now;

~~~javascript
const text = new PIXI.Text("my text", { trim: true });
~~~
Updated pen; https://codepen.io/staff0rd/pen/KaVdXZ?editors=0010

Let me know where if you want the trimming function pulled out and put elsewhere.

src/core/text/Text.js
```diff
@@ -354,6 +363,79 @@ export default class Text extends Sprite
...
+        }
+
+        const trimHeight = bound.bottom - bound.top + 1;
+        const trimWidth = bound.right - bound.left;
+        const trimmed = this.context.getImageData(bound.left, bound.top, trimWidth, trimHeight);
+
+        return {
+            trimHeight,
+            trimWidth,
+            trimmed,
```
bigtimebuddy: Could you rename these output properties to `height`, `width` and `data`? The "trim" part of the names is implied.

ivanpopelyshev: +1

staff0rd:
@bigtimebuddy applied those changes

PR #3570, opened 2017-01-11
expose mini-signals

ivanpopelyshev:
it must be done if we are going to remove ee3 later
at this point, that will help with typescript definitions

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "earcut": "^2.0.7",
     "eventemitter3": "^2.0.0",
     "ismobilejs": "^0.4.0",
+    "mini-signals": "^1.1.1",
     "object-assign": "^4.0.1",
     "pixi-gl-core": "^1.0.3",
     "resource-loader": "^2.0.3"
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -1,6 +1,7 @@
 import { DATA_URI, URL_FILE_EXTENSION, SVG_SIZE, VERSION } from '../const';
 import settings from '../settings';
 import EventEmitter from 'eventemitter3';
+import Signal from 'mini-signals';
 import pluginTarget from './pluginTarget';
 import * as isMobile from 'ismobilejs';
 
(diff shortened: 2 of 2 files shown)


bigtimebuddy:
@ivanpopelyshev now that pixi-spine is fixed, is this good to close?

GoodBoyDigital:
Hi @ivanpopelyshev , is this still required? cheers!

bigtimebuddy:
It's my understanding that @ivanpopelyshev fixed his issues with pixi-spine. Feel free to re-open if this is incorrect.

drkibitz:
@ivanpopelyshev @bigtimebuddy  @GoodBoyDigital although this is a closed issue, and I've been off the project for a long while. I wanted to post something related to https://github.com/pixijs/pixi.js/issues/3835. Since I was the original author of the current internal ticker implementation, I thought it would only be courteous to chime in there, and hopefully be able to contribute something.

I was curious if anyone had tried to use signals in some way and did a search and found this (I personally prefer signals over emitters, but this isn't important). I have a local refactor of Ticker that uses a private home grown implementation of a Signal, which closely resembles the ee3 implementation with some caveats. I took a look at *mini-signals*, and did notice it is still using `Function.apply`, which in my opinion is unnecessary, as well as multiple optional arguments for dispatching. Because in general, for production ready high performance code, this can always be avoided, and is only necessary for special circumstances. So I'm on the side to not optimize for convenience, but still allow for it, hence make adding and removing easy, but the performance of dispatching not burdened by that convenience. I can only find 2 implementations of signals in JavaScript, mini-signals, and of course js-signals, both can be improved in the performance area.

So I was wondering (small poll):
1. Is anyone still interested in a Signal implementation?
2. Would anyone be interested in a new Signal implementation?
3. Would anyone be interested in a dependency on mini-signals or js-signals with the hope that a few PRs would be accepted to either of those projects?

FYI, my current local private Signal implementation is priority-based, does not use `Function.apply`, does not have the concept of bindings (has basically the same args as an emitter, minus event names), and only allows dispatching a single optional argument. The API currently looks like this:
- `signal.add(new Listener(fn, [context], [priority=0], [once=false]))`
- `signal.remove([fn], [context], [priority=0], [once=false])`
- `signal.dispatch([arg])`

ivanpopelyshev:
@drkibitz that was posted in our slack 2 days ago, I hope @GoodBoyDigital answers when he's free: https://github.com/GoodBoyDigital/mini-runner

drkibitz:
@GoodBoyDigital @ivanpopelyshev thanks, I'll take this discussion to an issue on that repo.

PR #3576, opened 2017-01-12
Ability to set stop points on Text with a gradient

themoonrat:
Currently, doing
fill: ["#FF0000", "#00FF00", "#0000FF"]
evenly spaces those gradient points out, but sometimes you might want more control over each stop point. With this PR you can optionally adjust these points; for example
fillGradientStops: [0.1, 0.3, 0.9]

### DIFF
--- a/src/core/text/TextStyle.js
+++ b/src/core/text/TextStyle.js
@@ -241,7 +241,7 @@ export default class TextStyle
     }
     set fillGradientStops(fillGradientStops)
     {
-        if (this._fillGradientStops !== fillGradientStops)
+        if (!areArraysEqual(this._fillGradientStops,fillGradientStops))
         {
             this._fillGradientStops = fillGradientStops;
             this.styleID++;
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
@themoonrat, can we see a demo of this?

themoonrat:
http://www.moonrat.co.uk/html5/pixi-gradient-stop-text/

src/core/text/TextStyle.js
```diff
@@ -232,6 +235,19 @@ export default class TextStyle
         }
     }
 
+    get fillGradientStops()
+    {
+        return this._fillGradientStops;
+    }
+    set fillGradientStops(fillGradientStops)
+    {
+        if (this._fillGradientStops !== fillGradientStops)
```
bigtimebuddy: While it's simple, I think there might be a better comparison than strict check. Since it is an array, might be useful if you compare the actual array values. For instance:

```js
text.style.fillGradientStop = [0, 0.2, 0.4];
text.style.fillGradientStop = [0, 0.2, 0.4]; // shouldn't re-render
```

bigtimebuddy:
Thanks for the demo, looks great. This will be useful for folks. Also, I appreciated that you used the new Application class in your demo 👍

src/core/text/TextStyle.js
themoonrat: Will do; also think I've seen an issue with the stop gradients settings not working nice enough over multiple lines, so I'll take a look when I get time :)

bigtimebuddy:
@themoonrat Any follow-up here with multiple lines or a better dirty check for array?

themoonrat:
I'll have time to look at it this weekend :)

themoonrat:
Done for both. http://www.moonrat.co.uk/html5/pixi-gradient-stop-text/ is updated with the latest example... with a new comparison for single line versus multi line, to show that they now match

PR #3578, opened 2017-01-13
Rectangle test updates

out-of-band:
Converted to arrow functions
Added new tests and new cases to existing tests
Removed unreachable ternary in Rectangle.pad()
Fixed invalid comparison in Rectangle.enlarge()
Changed param name in Rectangle.enlarge(rectangle) to match other methods

Can anyone help me with the expected functionality of the Rectangle.fit(rectangle) method?  I'm not sure I understand its purpose sufficiently to write a test.

### DIFF
--- a/src/core/math/shapes/Rectangle.js
+++ b/src/core/math/shapes/Rectangle.js
@@ -163,7 +163,7 @@ export default class Rectangle
     pad(paddingX, paddingY)
     {
         paddingX = paddingX || 0;
-        paddingY = paddingY || ((paddingY !== 0) ? paddingX : 0);
+        paddingY = paddingY || paddingX;
 
         this.x -= paddingX;
         this.y -= paddingY;
@@ -222,19 +222,23 @@ export default class Rectangle
     /**
      * Enlarges this rectangle to include the passed rectangle.
      *
-     * @param {PIXI.Rectangle} rect - The rectangle to include.
+     * @param {PIXI.Rectangle} rectangle - The rectangle to include.
      */
-    enlarge(rect)
+    enlarge(rectangle)
     {
-        if (rect === Rectangle.EMPTY)
+        if (rectangle.left === 0
+            && rectangle.top === 0
+            && rectangle.right === 0
+            && rectangle.bottom === 0
+            && rectangle.type === SHAPES.RECT)
         {
             return;
         }
 
-        const x1 = Math.min(this.x, rect.x);
-        const x2 = Math.max(this.x + this.width, rect.x + rect.width);
-        const y1 = Math.min(this.y, rect.y);
-        const y2 = Math.max(this.y + this.height, rect.y + rect.height);
+        const x1 = Math.min(this.x, rectangle.x);
+        const x2 = Math.max(this.x + this.width, rectangle.x + rectangle.width);
+        const y1 = Math.min(this.y, rectangle.y);
+        const y2 = Math.max(this.y + this.height, rectangle.y + rectangle.height);
 
         this.x = x1;
         this.width = x2 - x1;
--- a/src/core/math/shapes/Rectangle.js
+++ b/src/core/math/shapes/Rectangle.js
@@ -163,7 +163,7 @@ export default class Rectangle
     pad(paddingX, paddingY)
     {
         paddingX = paddingX || 0;
-        paddingY = paddingY || paddingX;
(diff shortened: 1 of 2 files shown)


src/core/math/shapes/Rectangle.js
```diff
@@ -163,7 +163,7 @@ export default class Rectangle
     pad(paddingX, paddingY)
     {
         paddingX = paddingX || 0;
-        paddingY = paddingY || ((paddingY !== 0) ? paddingX : 0);
+        paddingY = paddingY || paddingX;
```
GoodBoyDigital: Tricky one this line! This one will need to change back as if you set ```pad(10, 0)``` the paddingY would end up being set to 10 also

src/core/math/shapes/Rectangle.js
```diff
@@ -222,19 +222,23 @@ export default class Rectangle
     /**
      * Enlarges this rectangle to include the passed rectangle.
      *
-     * @param {PIXI.Rectangle} rect - The rectangle to include.
+     * @param {PIXI.Rectangle} rectangle - The rectangle to include.
      */
-    enlarge(rect)
+    enlarge(rectangle)
     {
-        if (rect === Rectangle.EMPTY)
```
GoodBoyDigital: this whole check we can be remove.

GoodBoyDigital:
the ```fit``` function should basically shrink a rectangle so that its bounds fit inside another.

for example if you have a rectangle(0, 0, 20, 20)

and fit it into rectangle(10, 0, 20, 20)

the original rectangle would end up being cropped on the left to become rectangle(10,0,20,20)

Hope that makes sense!

src/core/math/shapes/Rectangle.js
out-of-band: OK I'll remove the check.

src/core/math/shapes/Rectangle.js
out-of-band: Oops.  I should have written a test for that.  Good catch.

out-of-band:
Thanks for reviewing and the info on the Rectangle.fit() method!

The PR has been updated accordingly.

GoodBoyDigital:
almost there! looks like one of the tests is failing?

out-of-band:
I didn't update the Rectangle.enlarge() tests to account for removing the Rectangle.EMPTY check from the method.  It's fixed.  Sorry about that.

The build is still failing, though, but on test `  1) PIXI PIXI.prepare.TimeLimiter should limit to stop after time from beginFrame():` which does not appear to be related to this PR.

GoodBoyDigital:
weird, tried running them again and they passed that time! Good job @out-of-band 👍

test/core/Rectangle.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
-describe('PIXI.Rectangle', function ()
+describe('PIXI.Rectangle', () =>
 {
-    it('should create a new rectangle', function ()
+    it('should create a new rectangle', () =>
```
bigtimebuddy: Could you please revert converting these to fat-arrows? We intentionally kept as `function` so that we could retain the mocha scope, which is more useful here.

staff0rd: @bigtimebuddy I was gonna note that, but I couldn't remember why we do it.  Is it worth having that demonstration of the mocha scope vs fat arrows that you posted at some stage in the wiki?

bigtimebuddy: https://github.com/pixijs/pixi.js/wiki/Unit-Tests

out-of-band: I can certainly change them back to `function`--I just saw the arrow functions in some other tests (e.g., `Graphics.js`, `Container.js`, `DisplayObject.js`) and assumed it was OK.

I can go through and change other tests back to the `function` syntax in another PR if you'd like.

bigtimebuddy: Let's change those back in another PR, but thanks for pointing it out!

staff0rd: @out-of-band probably my bad.

@bigtimebuddy it's only in the `describe` and `it` functions right?  `should`s and other inline stuff can still be fat arrows right?

bigtimebuddy: Yup @staff0rd, exactly.

Issue #3588, opened 2017-01-15
There is a bug about `RenderTexture` in document or  codes.


finscn:
The doc says that  :
```
* A RenderTexture takes a snapshot of any Display Object given to its render method. 
* The position and rotation of the given Display Objects is ignored.
```

But  I found when I render a sprite to an RenderTexture,  it couldn't ignore `position` & `rotation`.

My test code is very simple : 
```
        var img = ResourcePool.get("rmbp");
        var texture = PIXI.Texture.from(img);
        var sprite = PIXI.Sprite.from(texture);
        sprite.tint = 0xFF0000;
        sprite.rotation = 0.8;
        sprite.position.set(50,50);
        sprite.scale.set(0.5, 0.5);
        sprite.anchor.set(0.5, 0.5);

        var renderTexture = PIXI.RenderTexture.create(800, 600);
        game.renderer.render(sprite, renderTexture);


        this.stage = new PIXI.Container();
        var tempSprite = PIXI.Sprite.from(renderTexture);
        this.stage.addChild(tempSprite);

        game.renderer.render(this.stage);
```

~~Only  `position` is ignored.~~

Wrong document or wrong codes ?

.

finscn:
I found it doesn't ignore position & rotation both.

finscn:
BTW , I love current behavior.  :)  

If  ignore position & rotation , it will make some problems in some shaders, like this : 
https://github.com/pixijs/pixi.js/issues/3590

GoodBoyDigital:
Hi @finscn, this is correct behaviour. I have modify the docs to reflect this.

GoodBoyDigital:
sorted 👍

PR #3591, opened 2017-01-16
Updated to docs #3588

GoodBoyDigital:


finscn:
the doc : 
```
 * The Sprite in this case will be rendered to a position of 0,0. To render this sprite at its actual
 * position a Container should be used:
 *
 * ```js
 * let doc = new PIXI.Container();
 *
 * doc.addChild(sprite);
 *
 * renderer.render(doc, renderTexture);  // Renders to center of renderTexture
 * ```
``` 
should be removed too.

finscn:
@GoodBoyDigital  , the doc 
```
 * The Sprite in this case will be rendered to a position of 0,0. To render this sprite at its actual
 * position a Container should be used:
 *
 * ```js
 * let doc = new PIXI.Container();
 *
 * doc.addChild(sprite);
 *
 * renderer.render(doc, renderTexture);  // Renders to center of renderTexture
 * ```
```
is an example for  the situation that `ignore position`.

In the current PIXI.js ,  we don't have to add the sprite to a container for  ` render this sprite at its actual  position`.

GoodBoyDigital:
Ah, i forgot to push!

PR #3601, opened 2017-01-17
Adds destroy method to Application

bigtimebuddy:
### Added

* Adds a `destroy` method to the new Application API to cleanup the container, renderer and ticker instances.

### DIFF
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -54,13 +54,18 @@ export default class Application
          */
         this.ticker = new Ticker();
 
-        this.ticker.add(() =>
-        {
-            this.renderer.render(this.stage);
-        });
+        this.ticker.add(this.render, this);
 
         // Start the rendering
         this.start();
+    }
+
+    /**
+     * Render the current stage.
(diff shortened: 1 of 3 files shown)


src/core/Application.js
```diff
@@ -54,16 +54,21 @@ export default class Application
...
+        this.ticker.add(this.render, this);
 
         // Start the rendering
         this.start();
     }
 
     /**
+     * Render the current stage.
+     */
+    render()
```
themoonrat: I'm not sure how I feel about render being a public function, when part of the appeal of the Application class is to hide boiler plate stuff away from developers. Would they ever want to call render() manually when the ticker is already calling it when it can automatically?

bigtimebuddy: Maybe you want to call `app.stop()` to stop self-updating and then manually render `app.render()`. Could be useful in low-performance environment or where another ticker is driving the updates.

themoonrat: A convincing case :)

PR #3604, opened 2017-01-17
Fix for #3505

GoodBoyDigital:


GoodBoyDigital:
some comment tweaks I forgot to push!

PR #3612, opened 2017-01-18
User can decide whether  auto-set the projection

finscn:
`bindShader` always  set the projection matrix  automatically . in most cases ,  that's OK.

But When I create some new renderers of PIXI ,  sometimes it would make some problems.
e.g. When : 
* shader no projectionMatrix
* shader need use a custom projectionMatrix (not  _activeRenderTarget's)
* shader use a cached array (don't want to call `toArray` every time)

So I add a new argument to `bindShader` , and the default value is true, so  it would NOT break any current code of PIXI

.

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -440,9 +440,10 @@ export default class WebGLRenderer extends SystemRenderer
      * Changes the current shader to the one given in parameter
      *
      * @param {PIXI.Shader} shader - the new shader
+     * @param {boolean} [autoProject=true] - Whether automatically set the projection matrix
      * @return {PIXI.WebGLRenderer} Returns itself.
      */
-    bindShader(shader)
+    bindShader(shader, autoProject)
     {
         // TODO cache
         if (this._activeShader !== shader)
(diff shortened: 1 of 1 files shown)


src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -440,18 +440,22 @@ export default class WebGLRenderer extends SystemRenderer
      * Changes the current shader to the one given in parameter
      *
      * @param {PIXI.Shader} shader - the new shader
+     * @param {boolean} [autoProject=true] - Whether automatically set the projection matrix
      * @return {PIXI.WebGLRenderer} Returns itself.
      */
-    bindShader(shader)
+    bindShader(shader, autoProject)
```
bigtimebuddy: Small stylistic change: could you set a default parameter `autoProject = true` and then  do `if (autoProject)` below. Better to be more explicit than assume true and `!== false`.

finscn: if use   default parameter  `autoProject = true` , the real code generated by babel will has bad performance.

bigtimebuddy: That make sense, but because this is an anti-pattern, it's important that you leave a comment explaining that this is a performance optimization with Babel. In the future we may not use Babel for transpiling and would change this to be default params.

finscn: 'leave a comment explaining something' is very difficulty for  me , you know my english is very bad.   tears...

bigtimebuddy: No problem. Leave an inline comment:

```js
// `autoProject` normally would be a default parameter set to true
// but because of how Babel transpiles default parameters
// it hinders the performance of this method.
```

finscn: Done. Thanks.

PR #3614, opened 2017-01-19
Fix resolution setting of PIXI.Filter, and another two wrong codes on resolution importing

Icemic:
Set `this.resolution` to `settings.RESOLUTION` instead of `1`, otherwise texture will be blurred in retina screens (where `settings.RESOLUTION` may be setting to `2` or above).

### DIFF
--- a/src/core/renderers/canvas/utils/CanvasRenderTarget.js
+++ b/src/core/renderers/canvas/utils/CanvasRenderTarget.js
@@ -1,5 +1,4 @@
 import settings from '../../../settings';
-const { RESOLUTION } = settings;
 
 /**
  * Creates a Canvas element of the given size.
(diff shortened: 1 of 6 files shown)


themoonrat:
There are other areas where the default resolution is set to 1, rather than grabbing from settings. Worth going through and changing all those too in this PR? Or do that separately?

Icemic:
I find three other areas about filter, I think they could be combined to this pr, I would do this later.

image

And there are some others about `InterationManager`, `Graphic`, maybe more. I think they should be fixed separately after checking if they have side effects.

Icemic:
Committed. Fixes default value of resolution about BlurFilter, and another bug on importing the `RESOLUTION` constant

Issue #3589, opened 2017-01-15
Graphics.hitArea not triggering interaction when mask attached


ricklamers:
I think I found a bug in pixi.js. 

If I add a mask like this:

```
this.container = new PIXI.Container();

var mask = new PIXI.Graphics();
//mask.beginFill(0x000000); <-- note I'm not calling this line
mask.drawRect(0,0,500,500);

this.addChild(mask);
this.container.mask = mask;
```

When I run this, I do see that the mask is clipping the content but the interaction is not working, only after I added beginFill(0x000000) it started clipping **and** allowing interaction.

Maybe it's just a mistake on my side, but I spent almost an hour figuring this out. I initially thought I had to set filterArea (which according to the comments in the source is also responsible for interaction clipping in v4, which was hard to find in the docs) in order to get interaction to work.

Maybe this is not so much of a bug as a documentation issue, I just didn't know how to get masking + interaction working. Turns out, I just needed to call beginFill(0x000000) on the masking object.

staff0rd:
Are you able to supply a [pen](https://codepen.io) that displays this issue you're having against the latest [dev build](//d157l7jdn8e5sf.cloudfront.net/dev/pixi.js) for us to mess with/debug?

ricklamers:
Hello staff0rd, I've implemented the bug in a codepen. 

Here's the result: 

http://codepen.io/ricklamers/pen/ygVqaB

To find the core problem, Ctrl+F for 'ISSUE'. You'll see this line:

`// mask.beginFill(0xFFFFFF, 1);`

Uncomment it to make interaction work through the mask. 

Kind regards,

staff0rd:
Hey @ricklamers, thanks for the pen.

It would seem the reason that interactive does not work when you don't use a fill is because of [this](https://github.com/pixijs/pixi.js/blob/e1b98404a982467811ee655de8372bd5d1d77177/src/core/graphics/Graphics.js#L863-L866), which is called by `InteractionManager` when [testing hits on masks](https://github.com/pixijs/pixi.js/blob/e1b98404a982467811ee655de8372bd5d1d77177/src/interaction/InteractionManager.js#L794) (note the adjacent code for specifying either `hitArea` or `filterArea` to achieve similar results.

Considering fill is an expectation of `Graphics.containsPoint` (I've opened #3608 to confirm this), I'd say that if there's any bug here it's that the masking works when the fill is not evident.

[...]

PR #3615, opened 2017-01-19
Mask/Fill interaction tests + fix

### LINKED ISSUE #3151: Interactivity problem with masked elements (v4.1.0)

staff0rd:
Prior to the fix being added, the test state was this;

✔️ should trigger interaction callback when no mask present
✔️ should trigger interaction callback when mask uses beginFill
✔️ should not trigger interaction callback when mask doesn't use beginFill
❌  should trigger interaction callback when mask doesn't use beginFill but hitArea is defined
❌ should trigger interaction callback when mask doesn't use beginFill but filterArea is defined
✔️ should trigger interaction callback when mask is a sprite

All tests pass after 9658fc6.

Resolves #3151 and the duplicate #3589


src/interaction/InteractionManager.js
```diff
@@ -788,19 +788,19 @@ export default class InteractionManager extends EventEmitter
...
+            if (!displayObject.filterArea.contains(point.x, point.y))
             {
                 hitTest = false;
             }
         }
 
-        // it has a filterArea! Same as mask but easier, its a rectangle
-        if (hitTest && displayObject.filterArea)
+        // it has a mask! Then lets hit test that before continuing if no hitArea or filterArea is present
+        if (hitTest && displayObject._mask && !displayObject.hitArea && !displayObject.filterArea)
```
GoodBoyDigital: do we need this last bit? ```&& !displayObject.hitArea && !displayObject.filterArea``` ?

GoodBoyDigital:
Also what @bigtimebuddy said! Big respect for constantly adding tests and bringing more stability to PIXI :D What would we do without you!

src/interaction/InteractionManager.js
staff0rd: @GoodBoyDigital actually, that part _is_ the fix.  Without it, [this](https://github.com/pixijs/pixi.js/blob/ccb99cff69889c228e820969cf8e01af699c82e6/src/core/graphics/Graphics.js#L863-L866) code will be hit (for masks not using `beginFill`) and `hitTest` will be set to false, causing the relevant events to not be emitted.

Prior to the fix, a mask overrides all other hitTests.  After the fix, priority is given to the existence of `hitArea` or `filterArea` over the presence of a mask.

Does this make sense?

GoodBoyDigital: Ahh that does! Just gonna throw this out there - but wander if we should even test for filter area. The purpose was to optimise filters rather than interaction. 

The majority of the time the filter area is set to something massive too, offering very little in way of optimisation here..

I think we should remove the check.. 

What are your thoughts?

Cheers!

staff0rd: @GoodBoyDigital yeah I don't use filters so I can't really comment.  I'll pull it out if you want?

bigtimebuddy: I agree with @GoodBoyDigital, `filterArea` is not intended to be used as interaction area. I think we should pull that out.

staff0rd: yanked :v:

bigtimebuddy:
@GoodBoyDigital ping. Could you take a look at this again?

PR #3617, opened 2017-01-19
Marks the interaction events as being of the type PIXI.interaction.InteractionData

themoonrat:
For the docs


staff0rd:
@themoonrat out of interest, why is both @memberof and @type needed?

themoonrat:
Because it's an @event I believe
Without @memberof the entry will not show in the docs
I tried just without @memberof, and I tried without @memberof but changing @type to @member, and neither worked.
I don't know why, and I don't tend to question how jsdoc works anymore, I just let out a little cheer inside when it outputs what I want.

PR #3620, opened 2017-01-20
CanvasTinter canvas size rounds up

D-Hartley:
In my project I use resolution as an easy way to use smaller assets on mobile devices. This means that occasionally the `crop.width` or `crop.height` is a decimal. When you set a canvas to a decimal number it seems to always round down (at least it does for me). This sometimes results in the tinted image being very slightly cut off.

Rounding the canvas size up has fixed this issue for me.

### DIFF
--- a/src/core/sprites/canvas/CanvasTinter.js
+++ b/src/core/sprites/canvas/CanvasTinter.js
@@ -75,8 +75,8 @@ const CanvasTinter = {
         crop.width *= resolution;
         crop.height *= resolution;
 
-        canvas.width = crop.width;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@D-Hartley thanks for the PR. Do you have a screenshot of what this visual bug looks like?

D-Hartley:
Sure, I managed to recreate it using BitmapText. You can see the bug on the far right of some of the characters, especially the O. The gif isn't great quality but it also effects the top right of N and bottom right of Q.

bug

PR #3628, opened 2017-01-22
Allow non-interactive children to NOT trigger interactive parents

### LINKED ISSUE #3596: Interactive parents should not make non-interactive children block interactive children
When parent is interactive and front child is non-interactive clicking an overlap should cause the behind (interactive) child to be hit.

See the failing test in #3535's description.  [This test has been commented out](https://github.com/pixijs/pixi.js/blob/15881774d4b0a0bb305b428990f0a303989ad72b/test/interaction/InteractionManager.js#L367-L385) to allow merging of the passing tests, however, pixi is still displaying the bug that causes the test to fail. Bug needs to be fixed, test needs to be uncommented.

Example pen; https://codepen.io/staff0rd/pen/BpQaxg?editors=0010

In the example pen above, click the blue square (half hidden behind the green square) and note the console fires the handler.  Click the right-hand side of the green square (the overlap) and note the console fires the handler.  Now, uncomment line 38 and perform the test again;  Clicking blue results in both parent & behind handlers firing.  Clicking the overlap results in **only the parent** firing, which is the bug.  In this scenario (and commented test above), it is expected that the behind handler should also fire - as it does when the parent is not interactive.

staff0rd:
Resolves #3596 and the broken test in #3535

Adds the somewhat verbose, but hopefully descriptive property to `interactiveTarget`;

~~~javascript
/**
 * Whether hits on non-interactive children should cause events to be raised
 * against their interactive parent
 *
 * @inner {boolean}
 */
nonInteractiveChildrenTriggerInteractiveParent: true,
~~~

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -877,6 +877,10 @@ export default class InteractionManager extends EventEmitter
 
                 func(displayObject, hit);
             }
+            else if (displayObject.parent && !displayObject.parent.nonInteractiveChildrenTriggerInteractiveParent)
+            {
+                hit = false;
+            }
         }
 
         return hit;
--- a/src/interaction/interactiveTarget.js
+++ b/src/interaction/interactiveTarget.js
@@ -30,6 +30,14 @@ export default {
     interactiveChildren: true,
 
     /**
+     * Whether hits on non-interactive children should cause events to be raised
+     * against their interactive parent
+     *
+     * @inner {boolean}
+     */
+    nonInteractiveChildrenTriggerInteractiveParent: true,
+
+    /**
      * Interaction shape. Children will be hit first, then this shape will be checked.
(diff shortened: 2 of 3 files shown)


bigtimebuddy:
The verbose name is outta control. There must be something more succinct here: useInactiveChildren

uorbe001:
Hey, @staff0rd I just saw this! 

I opened [#3637](https://github.com/pixijs/pixi.js/pull/3637) fixing the same bug, my fix is *very* simple so it's possible I broke some case that I haven't noticed :grin:. Did you find anything that would break if you didn't add that extra property?

uorbe001:
I updated my branch again (#3638) - I noticed the 2 failing tests I had were pretty easy to solve without having to add a new property. I referenced this there as I think it might be worth looking at the alternative (admittedly, there might be some untested cases that are broken in my branch).

staff0rd:
Superseded by #3535

staff0rd:
Superseded by #3638

Issue #3558, opened 2017-01-08
VertexArrayObjects and MultiTexture Batching on 'old' devices


GoodBoyDigital:
Hey Pixi peeps,

I have noticed that there seems to be a few issues on old devices when using the above (which are enabled by default). 

Im thinking we could introduce a new prop to the settings / a render option called something like 'old'. Setting this prop to true would disable VAOs and set Max textures to 1. 

We could also set them to be that by default, but most modern devices really benefit from this stuff so I think its best to not to.

What do you think?

Cheers!

Mat

themoonrat:
Definitely. A couple of thoughts
1. Should we do some device detection ourselves for greatest compatibility out of the box? So if we know a Galaxy tab requires 'compatibility mode' we turn it on by default? Or always leave it down to the developer to choose?
2. Could this scale the other way? For example, I believe there were issues with highp with fragment shaders for some devices, but for some it'd be nicer to have it available? Could this flag go multiple ways? So -1 = compatibility, 0 = default, 1 = bleeding edge?

staff0rd:
Any examples or specific areas that these issues appear?  Could possibly help devs determining whether using the flag can be a solution for them.

GoodBoyDigital:
@staff0rd , this tends to be an issue than can pop up when rendering content on older generation devices. It manifests itself as crazy flickering / broken textures. I believe this is due to some old GPUs really struggle with if statements in shaders. Whilst others claim to support VBOs, but in reality do not.. 

@themoonrat I do like the idea of a compatibility mode. I think it could scale the other way too :)

artsjedi:
Is possible to disable disable VAOs in v4? how can I do it?

artsjedi:
Meybe... is there something that we could findout if the device will support the MultiTexture Batching. some gl flag or something?

artsjedi:
Is possible to disable disable VAOs in v4? how can I do it? Meybe... is there something that we could findout if the device will support the MultiTexture Batching. some gl flag or something?

@staff0rd  here is an example https://www.youtube.com/watch?v=t6wVYRm731I

artsjedi:
One important thing that i figure find out. 

I made some test with packing all textures in one big (really big) PNG. **this way the issue is gone.**
When I add more than one texture file, somehow the software swap the correct texture that it should render for each sprite.

[...]

PR #3630, opened 2017-01-22
Added legacy option to webGL renderer

GoodBoyDigital:
Setting ```legacy``` to true now forces VertexArrayObjects and
MultiTexture Batching to be disabled. We also store the variable on the
renderer just incase we want to use it to force settings in anything
else. #3558

also added a test

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -49,10 +49,19 @@ export default class WebGLRenderer extends SystemRenderer
      *  enable this if you need to call toDataUrl on the webgl context.
      * @param {boolean} [options.roundPixels=false] - If true Pixi will Math.floor() x/y values when
      *  rendering, stopping pixel interpolation.
+     * @param {boolean} [options.legacy=false] - If true Pixi will aim to ensure compatibility
+     * with older / less advanced devices. If you experiance unexplained flickering try setting this to true.
      */
     constructor(width, height, options = {})
     {
         super('WebGL', width, height, options);
+
+        this.legacy = !!options.legacy;
+
+        if (this.legacy)
+        {
+            glCore.VertexArrayObject.FORCE_NATIVE = true;
+        }
 
         /**
          * The type of this renderer as a standardised const
--- a/test/core/WebGLRenderer.js
+++ b/test/core/WebGLRenderer.js
@@ -1,0 +1,14 @@
+'use strict';
+
+describe('PIXI.WebGLRenderer', function ()
+{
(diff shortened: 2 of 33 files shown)


GoodBoyDigital:
hmm... test is failing because no webgl? Shall i just remove it?

src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -49,11 +49,20 @@ export default class WebGLRenderer extends SystemRenderer
      *  enable this if you need to call toDataUrl on the webgl context.
      * @param {boolean} [options.roundPixels=false] - If true Pixi will Math.floor() x/y values when
      *  rendering, stopping pixel interpolation.
+     * @param {boolean} [options.legacy=false] - If true Pixi will aim to ensure compatibility
+     * with older / less advanced devices. If you experiance unexplained flickering try setting this to true.
```
bigtimebuddy: `experiance` => `experience`

bigtimebuddy:
@GoodBoyDigital Travis does not test WebGL, and you can only test it locally. Just check for WebGL before doing those tests.

src/core/renderers/webgl/WebGLRenderer.js
bigtimebuddy: Could you also add this doc for legacy to the `src/core/Application.js` class?

GoodBoyDigital: done! 👍

test/core/WebGLRenderer.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function isWebGLSupported(fn)
+{
+    return PIXI.utils.isWebGLSupported() ? fn : function () {}; // eslint-disable-line no-empty-function
```
englercj: This will still output different tests for each environment. I would have this return `undefined` instead of an empty function. Then use it on the `it()` blocks.

That way each `it()` test registers, and it will output "skipped" instead of not appearing at all.

artsjedi:
Is possible to know when the legacy mode should be turned on?
I mean. There is some old devices that don't deal very well with MultiTextureBatching. 
Instead of turn "leagacy" on for everyone, is possible to determine that, for the current device, the legacy mode should be turned on?

bigtimebuddy:
It is difficult detect an appropriate device for legacy. The only real option is to create a user agent sniffer to compare against a black list of devices. However this is going to be somewhat complicated and hard to maintain and not that accurate. In the end, having a toggle felt like an okay option and left it up to developers to determine what devices they want to run this option on.

PR #3632, opened 2017-01-22
Fix multiple filters clearing renderTarget

Frugality:
Hi -

I was having an issue with applying multiple filters to an object in my scene. When I did so, many of my other objects would disappear. As I discovered, it's because the FilterManager was clearing the target buffer when popping the last filter, clearing anything rendered before it. Repro steps are pretty easy, take the Container Pivot demo, and add the following:

```
var filter = new PIXI.filters.ColorMatrixFilter();
filter.brightness(0.5);
container.children[12].filters = [filter, filter];
```

Once you do this, you'll see that any bunnies rendered before this one won't show up, and bunnies rendered after it do show up.

Tracking it down was pretty easy. Thanks for this incredible library!

Cheers,
-Courtland


GoodBoyDigital:
nice work @Frugality! This library only gets better with people like yourself helping out 👍

PR #3638, opened 2017-01-23
Uncomment & fix failing test from #3535.

### LINKED ISSUE #3419: posssible InteractionManager propagation bug

uorbe001:
Resolves #3419 and #3628.

I just saw #3628 after I started working on this and I opened another PR (wrong) about the same bug #3628 looks good to me but adds an extra property, which I personally tend to dislike. This might have some side effects for cases that are untested, but I thought it's worth opening as an alternative to #3628.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -799,6 +799,8 @@ export default class InteractionManager extends EventEmitter
             }
         }
 
+        let keepHitTestingAfterChildren = hitTest;
+
         // ** FREE TIP **! If an object is not interactive or has no buttons in it
         // (such as a game scene!) set interactiveChildren to false for that displayObject.
         // This will allow pixi to completely ignore and bypass checking the displayObjects children.
@@ -831,14 +833,19 @@ export default class InteractionManager extends EventEmitter
                     // This means we no longer need to hit test anything else. We still need to run
                     // through all objects, but we don't need to perform any hit tests.
 
-                    // {
-                    hitTest = false;
-                    // }
+                    keepHitTestingAfterChildren = false;
+
+                    if (child.interactive)
+                    {
+                        hitTest = false;
+                    }
 
                     // we can break now as we have hit an object.
                 }
             }
         }
+
+        hitTest = keepHitTestingAfterChildren;
 
         // no point running this if the item is not interactive or does not have an interactive parent.
         if (interactive)
--- a/test/interaction/InteractionManager.js
+++ b/test/interaction/InteractionManager.js
@@ -639,7 +639,6 @@ describe('PIXI.interaction.InteractionManager', function ()
                     expect(scene.parentCallback).to.have.been.calledOnce;
                 });
 
-                /* TODO: Fix #3596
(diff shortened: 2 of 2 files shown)


staff0rd:
I've looked at this while stepping through the tests and am feeling it's a better direction to go than #3628.

Interestingly, this PR moves the code back towards a state prior to changes made via ae2888b42598d18052328a2f4308fbf49a8b2d3e, 806681efa3c61ff5c9a591cfa489857946a5277c.  It's not clear why those changes were neccessary, perhaps @GoodBoyDigital can shed some light, however now that we have tests over this it's possibly ok.

Additionally, because of ac743bb34e0fdaf01bc3d033af6863713fb8ef4c and e46b3652be6922464007b75210271ddaf4531365, all children are being partially inspected anyway, but I can write some further tests to investigate what should/should not be triggered.

staff0rd:
This also passes #3665

staff0rd:
Hey @uorbe001, if you can make that variable name change I'll approve this :v:

bigtimebuddy:
Ping @uorbe001

uorbe001:
@staff0rd sorry, I didn't see the notification for this. Do you have a better name for that variable in mind? I suck at naming things lol

uorbe001:
updated the name, but I'm not sure it's a good one either, if someone has a better idea I'm all for it.

uorbe001:
not entirely sure why that build failed, it doesn't look like anything I changed?

staff0rd:
Build failing was an unrelated (now fixed) issue.

PR #3641, opened 2017-01-23
Additional convenience getters for Application

bigtimebuddy:
Adds `width` and `height` convenience getters to the Application class. There are lots of references in the examples to `renderer.width` and `renderer.height`, this would help to simplify even more.

### DIFF
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -95,6 +95,26 @@ export default class Application
     }
 
     /**
+     * Convenience for the renderer's width.
+     * @member {Number}
+     * @readonly
+     */
+    get width()
+    {
+        return this.renderer.width;
+    }
+
+    /**
+     * Convenience for the renderer's height.
+     * @member {Number}
+     * @readonly
+     */
+    get height()
+    {
+        return this.renderer.height;
+    }
+
+    /**
      * Destroy and don't use after this.
      * @param {Boolean} [removeView=false] Automatically remove canvas from DOM.
      */


themoonrat:
One of the confusions people have with the dimensions is between the renderer width/height (PIXI), and the view width/height (HTML). I'm worried that the getters may not help that confusion. Explicitly accessing renderer.width hits home what width you're actually accessing.

ivanpopelyshev:
May be name it cssWidth, cssHeight and pixelWidth, pixelHeight? Don't forget about resolution too!

bigtimebuddy:
How about `viewWidth` and `viewHeight` for the actual canvas size, and `width` and `height` for the renderer size?

ivanpopelyshev:
that's not that verbose. I have to explain to people meaning of "width" and "height". There should be prefixes.

themoonrat:
I'm with Ivan; if anything is named just 'width' you are inevitable going to come to the question of "width of what?"

I don't see the point in cssWidth / pixelsWidth either. Then you're creating 4 getters, 2 for renderer, 2 for view. You may as well just let people do app.renderer.width and app.view.width and the clarity is there in the variable names.

ivanpopelyshev:
Thank you, @themoonrat! Getters for shortcuts are good, if they named correctly. Personally, I think there's always not enough getters. I was crying when "isJson" was removed.

bigtimebuddy:
I agree with @ivanpopelyshev about more getters/setters. 

I _still_ want to make a case for height/width/viewHeight/viewWidth because I feel it reflects the most common use-case. The parameter order and generic names should be reserved how people will actually use those APIs. Width/height as pass-thrus for renderer's dimensions reflect that those values are actually more important. When you are rendering items, defining filter sizes areas, draw graphics, you are concerned with the renderer width/height, not the CSS pixels.

ivanpopelyshev:
@bigtimebuddy 

> defining filter sizes areas, draw graphics, you are concerned with the renderer width/height, not the CSS pixels.

Actually, filterArea and graphics are measured in CSS pixels. That's the problem ;)

bigtimebuddy:
# 💩 

That _is_ a problem.

ivanpopelyshev:
@GoodBoyDigital we summon you!

Its a real problem with API. Right now we can fix it by naming those properties correctly in Application. The question is how to name them.

Also, I'd like to see the property "screenRectangle" or something like that, which is (0,0,renderer.width/renderer.resolution, renderer.height/renderer.resolution) and updated automagically. It will be a great shortcut

GoodBoyDigital:
i guess we should aim to hide the idea of resolution from users. If i set resolution to 2 then nothing should change other than the pixel density. I should not have to modify ANY code to make it work. 

Aside from `filterArea` is there any other elements that are resolution dependent?

bigtimebuddy:
I'm going to close this PR. I'm more interested in helping to address the proposal @GoodBoyDigital suggested--which I agree with. Resolution and sizes shouldn't be so cumbersome. I'd be interested in seeing a PR which unifies and simplifies these things in Pixi.

ivanpopelyshev:
@GoodBoyDigital @bigtimebuddy Everything depends on it. Users specify "renderer.width" to position stage elements and ask "whats wrong" when they change resolution.

PR #3647, opened 2017-01-24
Add an easy way to clear RenderTexture

finscn:
In current PIXI, clear a RenderTexture is complex : 
```
renderer.bindRenderTexture(myRenderTexture);
renderer.clear();
// restore the previous renderTexture
renderer.bindRenderTexture();
```
And bind renderTexture is a high-cost operation . We shouldn't call it for cleaning render texture.

So I  add this method.

relatived : https://github.com/pixijs/pixi.js/issues/3595

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -376,6 +376,23 @@ export default class WebGLRenderer extends SystemRenderer
     }
 
     /**
+     * Erases the render texture and fills the drawing area with a colour
+     *
+     * @param {PIXI.RenderTexture} renderTexture - The render texture to clear
+     * @param {number} [clearColor] - The colour
+     */
+    clearRenderTexture(renderTexture, clearColor)
+    {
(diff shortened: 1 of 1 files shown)


src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -376,6 +376,23 @@ export default class WebGLRenderer extends SystemRenderer
...
+    clearRenderTexture(renderTexture, clearColor)
+    {
+        const baseTexture = renderTexture.baseTexture;
+        const renderTarget = baseTexture._glRenderTargets[this.CONTEXT_UID];
+
+        if (renderTarget)
+        {
+            renderTarget.clear(clearColor);
+        }
+    }
```
bigtimebuddy: Could you `return this;` to be consistent with the other methods in the Renderer? Also, don't forget the JSDoc for return (can be copied from the other methods).

bigtimebuddy:
Thanks @finscn!

Issue #3645, opened 2017-01-24
TilingSprite uvRespectAnchor for CanvasRenderer


ivanpopelyshev:
Following #3642: we have to implement uvRespectAnchor for CanvasRenderer. Seems like people are using it.

PR #3648, opened 2017-01-24
uvRespectAnchor for CanvasRenderer

ivanpopelyshev:
#3645

It turns out, people are using that mode, so we need implementation for Canvas2d.

I tested it in pixi-examples window, its safe to merge.


bigtimebuddy:
@ivanpopelyshev It's required there be two approvals before merging. Please don't merge prematurely.

ivanpopelyshev:
@bigtimebuddy Oh, right, I counted mine approval by mistake :)

PR #3650, opened 2017-01-25
fix unexpected 0 text behaviour

uorbe001:
We got a weird glitch upgrading from pixi 3 to 4, setting the text in a PIXI.Text to 0 results in the label actually saying ' '. This seems like unexpected behaviour considering all the other numbers work :)

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -744,7 +744,7 @@ export default class Text extends Sprite
 
     set text(text) // eslint-disable-line require-jsdoc
     {
-        text = String(text || ' ');
+        text = String(text === '' || text === null || text === undefined ? ' ' : text);
 
         if (this._text === text)
         {
--- a/test/core/Text.js
+++ b/test/core/Text.js
@@ -112,4 +112,42 @@ describe('PIXI.Text', function ()
             expect(text.width).to.equal(300);
         });
     });
+
+    describe('text', function ()
+    {
+        it('should convert numbers into strings', function ()
+        {
+            const text = new PIXI.Text(2);
+
+            expect(text.text).to.equal('2');
+        });
+
+        it('should not change 0 to \'\'', function ()
+        {
+            const text = new PIXI.Text(0);
+
+            expect(text.text).to.equal('0');
+        });
+
+        it('should prevent setting null', function ()
+        {
+            const text = new PIXI.Text(null);
+
+            expect(text.text).to.equal(' ');
+        });
+
+        it('should prevent setting undefined', function ()
+        {
+            const text = new PIXI.Text();
+
+            expect(text.text).to.equal(' ');
+        });
+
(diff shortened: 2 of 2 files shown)


bigtimebuddy:
Thanks for your pull-request @uorbe001 and helping to make Pixi better.

Since the text setter only accepts a type of `string` (as in the docs) it is not part of the API to pass in numbers. Allowing implicit conversion of types gets into issues, for instance, what if you pass in a value of `0x1`. If you are passing in a number, I suggest you cast it as a string first: `myText.text = String(num)`. It's for this reason, that I don't believe we should merge. 

If others felt strongly that this was an appropriate PR, an alternative approach would be to explicitly check for empty strings and null which would otherwise create issues for the text rendering. For example:

```js
text = String(text === '' || text === null ? ' ' : text);
```

This would handle your use-case with zero.

themoonrat:
I agree that the PR solution isn't correct, and prefer @bigtimebuddy 's suggestion, but I do feel that this should be added in. I know the API says to pass in a string, but the reality is, if you pass in any number the text _does_ show. Apart from 0; so I'd like 0 to also also show.

uorbe001:
@bigtimebuddy I get your point about the docs saying it only accepts strings, but the fact is that the other numbers *do* work and 0 used to work pre v4, which is what caused a bug in our codebase. I fixed the case we had in our codebase with the very same solution you proposed, but it still feels like it *should* work (or at least error) if the rest of the numbers are valid and 0 isn't. 

I amended my commit with your proposed approach 👍

GoodBoyDigital:
even though @bigtimebuddy is logically correct ( my brain agrees), my heart believes that if i pass in a 0 it should just work :)

src/core/text/Text.js
```diff
@@ -744,7 +744,7 @@ export default class Text extends Sprite
 
     set text(text) // eslint-disable-line require-jsdoc
     {
-        text = String(text || ' ');
+        text = String(text === '' || text === null ? ' ' : text);
```
andrewstart: Could this be changed to `text == null`? It may require disabling the ESLint strict equality check for that line, but it would make it so that `undefined` is handled the same as `null`.

bigtimebuddy: Here is the line that @andrewstart is suggesting. Requires disabling two rules in order to use the loose `null` comparison.
```js
text = String(text === '' || text == null ? ' ' : text); // eslint-disable-line no-eq-null, eqeqeq
```
I would suggest we use the more explicit (and shorter) version:
```js
test = String(text === '' || text === null || text === undefined ? ' ' : text);
```

uorbe001: updated this adding the undefined check 👍

bigtimebuddy:
@andrewstart, could you approve?

bigtimebuddy:
Thanks @uorbe001!

PR #3653, opened 2017-01-26
initial commit for ultimate renderer dimensions solution

ivanpopelyshev:
According to #3641, we really have a problem. Not even core developers know what does "width" or "height" mean.

Proposition:

1. screen.width and screen.height are used for the stage
2. view.width and view.height are size of canvas, thus they are measured in real pixels

If this one will be approved, I will move forward to

1. ultimate resize logic: resizeView, resizeScreen, resizeCss with autoUpdate tying them to each other. resizeBack for overriding by canvas&webgl. onResize event/signal/whatever
2. possible accessability fixes, I just saw a number of uses of "renderer.width" there which aren't that clear
3. jsdocs for all "width"s and "height"s, explaining which are actually screen coords and which are physical pixels. Most of them will be screen'ed, of course.
4. possible deprecation of "renderer.width", "renderer.height", just to be clear. I won't press for that, but may be team decides so
5. EXAMPLES FOR ALL THE POSSIBLE CASES of autoUpdate, setting width, height and resolution logic.

### DIFF
--- a/src/core/renderers/SystemRenderer.js
+++ b/src/core/renderers/SystemRenderer.js
@@ -1,5 +1,5 @@
 import { sayHello, hex2string, hex2rgb } from '../utils';
-import { Matrix } from '../math';
+import { Matrix, Rectangle } from '../math';
 import { RENDERER_TYPE } from '../const';
 import settings from '../settings';
 import Container from '../display/Container';
@@ -21,8 +21,8 @@ export default class SystemRenderer extends EventEmitter
 {
     /**
      * @param {string} system - The name of the system this renderer is for.
-     * @param {number} [width=800] - the width of the canvas view
-     * @param {number} [height=600] - the height of the canvas view
+     * @param {number} [screenWidth=800] - the width of the screen
+     * @param {number} [screenHeight=600] - the height of the screen
      * @param {object} [options] - The optional renderer parameters
      * @param {HTMLCanvasElement} [options.view] - the canvas to use as a view, optional
      * @param {boolean} [options.transparent=false] - If the render view is transparent, default false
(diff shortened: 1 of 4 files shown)


bQvle:
Sounds good to me

staff0rd:
I'm happy to see both this PR and #3641 discuss this matter.  In my personal experience I found (and sometimes still find) the relationship between the two dimensions a bit confusing (and have to hit up the docs), especially when it comes scaling over multiple device sizes and resolutions.

I think however that the answer here is to not try and hide the inner workings of resolution as suggested in the other PR (resolution is not private in the API) , but rather, to firmly communicate that there are two dimensions to be aware of.  I think this is best communicated by _not_ having a `width` or `height` convenience getters.  Something like `app.screen.width` and `app.view.width` clearly communicates that there are two dimensions worth knowing about and pretending that there's only one just confuses things.

src/core/renderers/SystemRenderer.js
```diff
@@ -192,23 +185,48 @@ export default class SystemRenderer extends EventEmitter
...
+     * @param {number} screenWidth - the new width of the screen
+     * @param {number} screenHeight - the new height of the screen
+     */
+    resize(screenWidth, screenHeight)
+    {
+        this.screen.width = screenWidth;
+        this.screen.height = screenHeight;
+
+        this.view.width = screenWidth * this.resolution;
+        this.view.height = screenHeight * this.resolution;
```
bigtimebuddy: Shouldn't this be `screenWidth / this.resolution`? For example, if someone sets the width and height of the screen to be 800 x 600 and a resolution of 2, that means the view width and height should be 400 x 300 (in CSS pixels), correct?

GoodBoyDigital: 800 x 600 with resolution of 2 means the view number of pixels will be 1600 x 1200 but the css will maintain the 800 x 600 size

ivanpopelyshev: @bigtimebuddy screen is the stage, its (0,0, 800, 600), its what we supply to resize

PR #3656, opened 2017-01-26
Added a fix for the Chrome on android

GoodBoyDigital:
If max textures is one it can omit the attribute. This check that it exists..


GoodBoyDigital:
smoooooth 👍

PR #3657, opened 2017-01-26
initial commit for ultimate renderer dimensions solution

ivanpopelyshev:
According to #3641, we really have a problem. Not even core developers know what does "width" or "height" mean.

Proposition:

1. screen.width and screen.height are used for the stage
2. view.width and view.height are size of canvas, thus they are measured in real pixels

If this one will be approved, I will move forward to

1. ultimate resize logic: resizeView, resizeScreen, resizeCss with autoUpdate tying them to each other. resizeBack for overriding by canvas&webgl. onResize event/signal/whatever
2. possible accessability fixes, I just saw a number of uses of "renderer.width" there which aren't that clear
3. jsdocs for all "width"s and "height"s, explaining which are actually screen coords and which are physical pixels. Most of them will be screen'ed, of course.
4. possible deprecation of "renderer.width", "renderer.height", just to be clear. I won't press for that, but may be team decides so
5. Examples for all possible cases of autoUpdate, setting width, height and resolution logic.

### DIFF
--- a/src/core/renderers/SystemRenderer.js
+++ b/src/core/renderers/SystemRenderer.js
@@ -1,5 +1,5 @@
 import { sayHello, hex2string, hex2rgb } from '../utils';
-import { Matrix } from '../math';
+import { Matrix, Rectangle } from '../math';
 import { RENDERER_TYPE } from '../const';
 import settings from '../settings';
 import Container from '../display/Container';
@@ -21,8 +21,8 @@ export default class SystemRenderer extends EventEmitter
 {
     /**
      * @param {string} system - The name of the system this renderer is for.
-     * @param {number} [width=800] - the width of the canvas view
-     * @param {number} [height=600] - the height of the canvas view
+     * @param {number} [screenWidth=800] - the width of the screen
+     * @param {number} [screenHeight=600] - the height of the screen
      * @param {object} [options] - The optional renderer parameters
      * @param {HTMLCanvasElement} [options.view] - the canvas to use as a view, optional
      * @param {boolean} [options.transparent=false] - If the render view is transparent, default false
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
Build is here: http://pixijs.download/dev-screen-width/pixi.js

bigtimebuddy:
I would suggest we deprecate SystemRenderer.width and SystemRenderer.height getters. These can easily be added to src/deprecation.js and provide pass-thru for `this.view.width` and `this.view.height`. Per our earlier discussions, this would help to further clarify the view size vs the screen size.

ivanpopelyshev:
I don't know what to do about "renderer.width/height", deprecation is kinda a big move. I was thinking something about SETTINGS, may be we can add a switch there, whether or not its VIEW or SCREEN.

Also, people are using it too often, if we deprecate it only one message should be shown.

ivanpopelyshev:
@bigtimebuddy you can commit deprecation into that branch, I dont mind

bigtimebuddy:
I thought about this some more, and I think you're right @ivanpopelyshev, it's a radical change. Let's leave it as-is for now until someone else feels strongly. Maybe we can revisit in future PRs.

ivanpopelyshev:
@bigtimebuddy yeah, its for v5.

src/core/renderers/SystemRenderer.js
```diff
@@ -69,20 +69,13 @@ export default class SystemRenderer extends EventEmitter
...
-        /**
-         * The height of the canvas view
+         * Its safe to use as filterArea or hitArea for whole stage
          *
-         * @member {number}
-         * @default 600
+         * @member {PIXI.Rectangle}
          */
-        this.height = height || 600;
+        this.screen = new Rectangle(0, 0, screenWidth || 800, screenHeight || 600);
```
bigtimebuddy: Could you make sure that you set `this.screen` to null when the SystemRenderer is destroyed?

ivanpopelyshev: ok, ill do it a bit later :)

ivanpopelyshev: @bigtimebuddy Actually it was taken care of, you just missed it :)

PR #3659, opened 2017-01-27
Improve error message for out-of-bounds Textures

marekventur:
Every now and then we stumble across this error in our logs:
```Texture Error: frame does not fit inside the base Texture dimensions [object Object]```

It's most likely a problem with some of our sprite packing, but it's always a hard one to debug. It doesn't help that the ```[object Object]``` bit is very unhelpful ;)

I know this error is not very common, but if one does come across it it's going to be a massive help to have a bit more information that helps pinpoint the culprit.

### DIFF
--- a/src/core/textures/Texture.js
+++ b/src/core/textures/Texture.js
@@ -459,7 +459,9 @@ export default class Texture extends EventEmitter
 
         if (frame.x + frame.width > this.baseTexture.width || frame.y + frame.height > this.baseTexture.height)
         {
-            throw new Error(`Texture Error: frame does not fit inside the base Texture dimensions ${this}`);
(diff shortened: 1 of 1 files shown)


kaansoral:
Is it possible to pull the file name too if the texture originates from a file?

My usage of PIXI has user facing sections, users read the logs, yet can't make sense of these errors either, the best solution is to prevent them, but having improved logs would be an easier solution too :)

marekventur:
I don't think there's an easily accessible name, but I'm happy to add it if someone can point me in the right direction.

ivanpopelyshev:
If you require a name, its spritesheetParser problem. Btw, it would be good to decouple it from pixi loader, because people just love to include json's in the build

PR #3664, opened 2017-01-27
FilterManager shaderCache reset as object

stevewillard:
The shaderCache is initially created as an object. When the Filter Manager gets destroyed, it should be reset to an empty object, not an array.


bigtimebuddy:
Thank you @stevewillard for making Pixi better!

Issue #3635, opened 2017-01-23
How to inject a sprite-sheet data object directly into the loader?


Milewski:
I have been doing this approach so far which pretty much will request this json data with an additional http request... which works!

```js
 Loader.add('../build/animation-sprite.json');
```

however what i want to do is to give the json object directly to the loader....

```js
let animationSprite = require('../build/animation-sprite.json');
Loader.add(animationSprite);
```

which doesn't seem to work.. how do i inject this data directly to the loader to by pass the http request it will further do?

megmut:
If you are using webpack or some kind of external loader, then there is no need to add the JSON to the loader. When you specify the require, it will trigger the file to be bundled in with the rest of your module.. something like this maybe.

`module: {
        loaders: [
            {
                test: /\.json$/,
                exclude: /node_modules/,
                loader: 'json-loader'
            }
        ],
}
`

You can just call your JSON object directly in your code. If you need this data to persist, then you may want to put a global object to store your json in some simple caching system. If you REALLY want to, though I can't recommend, you can force an object into the loader resources. You  may be able to create an instance of the Resource class. Resource type for JSON is type 1. Set the STATUS_FLAGS to 1 << 1, which is complete so it doesn't make a new xmlhttp request. I don't know what the side effects of doing this are though.

<img width="468" alt="screen shot 2017-01-30 at 10 46 32" src="https://cloud.githubusercontent.com/assets/10081046/22418467/79da0134-e6d9-11e6-9a0f-3b19cfe48234.png">

Milewski:
I didn't quite understand what you meant... 

My json data is dynamic generated... But I would like it to be bundled along my js file.. Because it's so small.. It not really necessary to make a new http request just to fetch this such small json data... 

However what I found annoying is that I can't manually tell the Loader 'look, I have this sprite sheet data, please set up it for me' 

Forcing my assets into the loader resource.. How would I define the offset of each asset.. My sprite sheet has at least 30 assets.. If i manually insert those resource.. I wouldn't be aligning them accordingly.

megmut:
Yes, in order to bundle your JSON along with your js file, you need to use an appropriate JSON loader. With Webpack, it will detect that you have required a JSON file, and inject it into your js bundle there.  [JSON Loader](https://github.com/webpack-contrib/json-loader)

[...]

PR #3676, opened 2017-01-30
Refactors the spritesheet, texture and bitmap font loaders

bigtimebuddy:
## Overview

Externalizes the construction/parser code in the loader middleware to be use-able without the Loader or for use with another loader system. 

### Added

* `PIXI.Texture.fromLoader(source, imageUrl, name)` Creates a texture and automatically adds it's contents to the BaseTextureCache and TextureCache.
* `PIXI.extras.BitmapText.registerFont(xml, texture)` Register a bitmap font from XML data and a Texture.
* `new PIXI.Spritesheet(baseTexture, data)` Parses the spritesheet JSON and asynchronously returns the map of textures.

### Fixed

* Fully documents `PIXI.loaders.Resource` and `PIXI.loaders.parseBitmapFontData` which previously were undocumented.

### Preview

* **Documentation** [here](http://pixijs.download/dev-loaders/docs/index.html)
* **Build** [here](http://pixijs.download/dev-loaders/pixi.js)


src/core/textures/Texture.js
```diff
@@ -413,6 +413,36 @@ export default class Texture extends EventEmitter
     }
 
     /**
+     * Create a texture from a source and add to the cache.
+     *
+     * @static
+     * @param {HTMLImageElement|HTMLCanvasElement} source - The input source.
+     * @param {String} imageUrl - File name of texture, for cache and resolving resolution.
+     * @param {String} name - Optional human readible name for the texture cache.
```
andrewstart: Should this be `[name]`?

src/extras/BitmapText.js
```diff
@@ -455,6 +455,71 @@ export default class BitmapText extends core.Container
...
+        data.size = parseInt(info.getAttribute('size'), 10);
+        data.lineHeight = parseInt(common.getAttribute('lineHeight'), 10);
+        data.chars = {};
+
+        // parse letters
+        const letters = xml.getElementsByTagName('char');
+
+        for (let i = 0; i < letters.length; i++)
+        {
+            const charCode = parseInt(letters[i].getAttribute('id'), 10);
```
andrewstart: Would it be worth caching `letters[i]` here to eke out a slightly faster parsing time?

bigtimebuddy:
While there are the necessary two approvals for merging, please wait 24 hours to allow others to chime in.

staff0rd:
Does this address #3635?

bigtimebuddy:
@staff0rd I wasn't explicitly trying to solve the Loader issue, but @Milewski should be able to do something like this instead of using the Loader: 

```js
const url = './build/spritesheet.json';
const data = require(url);
const texture = PIXI.Texture.fromImage(data.meta.image);
PIXI.loaders.parseSpritesheetData(url, data, texture, () => {
  // finished preparing spritesheet textures
});
```

src/core/textures/Spritesheet.js
```diff
@@ -0,0 +1,269 @@
...
+
+    /**
+     * Destroy Spritesheet and don't use after this.
+     */
+    destroy()
+    {
+        for (const i in this.textures)
+        {
+            this.textures[i].destroy();
+            delete this.textures[i];
```
andrewstart: Since we are cutting `this.textures` loose instead of retaining it, setting the value to `null` should be faster but just as good as clearing references.

src/core/textures/Spritesheet.js
```diff
@@ -0,0 +1,269 @@
...
+            {
+                this._parseComplete();
+            }
+        }, 0);
+    }
+
+    /**
+     * Destroy Spritesheet and don't use after this.
+     */
+    destroy()
```
andrewstart: Should we `destroy()` the base texture, or provide an optional parameter to do so?

bigtimebuddy:
Thanks @andrewstart

ivanpopelyshev:
Can we not force parsing at constructor? So we can delay it or implement another parsing, like for spine?

Milewski:
Could also not enforce the parse function to obligatorily receive a callback? currently i just want to cache the sprite to be used later... and i end up doing something like this

```js
new PIXI.Spritesheet(resource.texture, {...}).parse(() => {}); //<- note i don't intend to do anything right now but cache it...
```

Shafley:
@bigtimebuddy 
Your code
```js
const data = require('./build/spritesheet.json');
const baseTexture = new PIXI.BaseTexture(data.meta.image, null, 1);
const spritesheet = new PIXI.Spritesheet(baseTexture, data);
spritesheet.parse(function (textures) {
   // finished preparing spritesheet textures
});
```
Pixi.js version 4.5.3
The class `PIXI.BaseTexture(source, scaleMode, resolution)` the params `source`  required `HTMLImageElement | HTMLCanvasElement`, but the `data.meta.image` of `const data = require('./build/spritesheet.json');`  is a string.

Thank your code help me resolve then problem of useing spritesheet with webpack.

dulldave:
@Shafley (and to anyone finding this): The alternative looks to be to use `BaseTexture.from(data.meta.image, null, 1);` instead which can accept a string and output a BaseTexture.

kwiss:
How can i use it with 
```
let anim = new PIXI.extras.AnimatedSprite(spritesheet)
```

i have 

```
Uncaught TypeError: Cannot read property 'texture' of undefined

ivanpopelyshev:
@kwiss spritesheet has a map from names to textures inside. AnimatedSprite accepts an array. Convert one into another, I'm sure you know JS enough ;)

Issue #3687, opened 2017-02-02
[question] Render loop with new Ticker vs rendering with shared ticker


CharlesHolbrow:
What are the implications of using a `new Ticker` for the main render loop vs using a `PIXI.ticker.shared`?

I noticed the [Application](https://github.com/pixijs/pixi.js/blob/dev/src/core/Application.js) helper class creates a new Ticker instead of using `PIXI.ticker.shared`. 

Doesn't this mean that pointer events from the interaction manager (which uses the shared ticker) might go out of sync with the rendering ticker? 

In the [dragging example](https://pixijs.github.io/examples/#/demos/dragging.js), the Application class is used at the same time as the events triggered by the interaction manager. Is this intentional?

themoonrat:
I would make the argument that the render loop _should_ be a separate ticker to PIXI.ticker.shared. For example, imagine you wanted to pause everything on screen by stopping the ticker; you wouldn't want interactions to also be paused, you'd want them to carry on as normal.

bigtimebuddy:
That was the intention (^@themoonrat) with the Application class. But the use-case you describe is more advanced. There's an argument that using the shared ticker would help unify all the loops used by PIXI, therefore making it easier to pause everything at once. I created a PR #3688 to allow the shared Ticker's usage in Application.

PR #3688, opened 2017-02-02
Allow option to make Application ticker the shared ticker

bigtimebuddy:
### Added

* Adds an addition Application constructor argument to allow the ticker to be `PIXI.ticker.shared` instead of creating `new PIXI.ticker.Ticker()`

This option provides a performance benefit by having fewer `requestAnimationFrame` calls going on simultaneously.

In response to #3687


bigtimebuddy:
There is probably an argument here that the default should be to use the shared ticker. And the more advanced use-cases would want rendering using it's own Ticker.

PR #3700, opened 2017-02-06
Fixed #3698 (Matrix constructor now takes parameters)

### LINKED ISSUE #3698: Possible issue with Matrix constructor
While creating tests for Matrix.js, I noticed the constructor can only create matrices with the default values. 
i.e: `new PIXI.Matrix(2, 2, 2, 2, 100, 100)` will create a matrix with the default values of a = 1, b = 0, c = 0, d = 0, tx = 0 and ty = 0.

This is not necessarily an issue, as custom Matrix values can then easily be set using set(), but as this differs from all the other PIXI object constructors I looked at, I wasn't sure whether or not it was an intentional design choice.

JessicaNeary:
Fixes #3698 

Matrix can now be constructed with optional parameters for its a, b, c, d, tx and ty properties. If no parameters are given, a matrix with the default values is created.

### DIFF
--- a/src/core/math/Matrix.js
+++ b/src/core/math/Matrix.js
@@ -15,43 +15,43 @@ export default class Matrix
     /**
      *
      */
-    constructor()
+    constructor(a = 1, b = 0, c = 0, d = 1, tx = 0, ty = 0)
     {
         /**
          * @member {number}
          * @default 1
          */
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Good addition, thanks @JessicaNeary

Issue #3668, opened 2017-01-28
Touch events are no long working after MockPointer changes for #3375 (#3546) in Cocoon.


artsjedi:
Today I tried the dev breanch with MockPointer changes #3375 (#3546). after it touch events are no long working.

I got those error messages when I touch the screen.

I tried on ludei's Cocoon packing , in iOS and Android.

here are the results:

(iOS)
![ios](https://cloud.githubusercontent.com/assets/2263037/22397862/679df138-e562-11e6-8870-1b7e1e22d8f2.jpeg)

(Andoid)
![android](https://cloud.githubusercontent.com/assets/2263037/22397861/679dc06e-e562-11e6-8e3f-6d7245807c6b.jpeg)

andrewstart:
Does Cocoon not have a TouchEvent class?

GoodBoyDigital:
Hi , touch /mouse seemed to break when I updated to latest dev branch on an
app we are developing. Not 100% sure why yet!

artsjedi:
@andrewstart  Yes Cocoon does have a TouchEvent class.

andrewstart:
Does changing https://github.com/pixijs/pixi.js/blob/dev/src/interaction/InteractionManager.js#L1387 to `if (this.supportsTouchEvents && event instanceof window.TouchEvent)` fix the problem? It shouldn't, but maybe Cocoon works in ways that I am unaware of.

artsjedi:
sorry, i didint find the line you mentioned.

andrewstart:
Oh, merges messing up the line numbers 🙄. The relevant line is now https://github.com/pixijs/pixi.js/blob/dev/src/interaction/InteractionManager.js#L1444

artsjedi:
Nothing changes, by changing that line.  however, thanks to some "log events" the error is clear now.

![screenshot_2017-02-06-23-37-19](https://cloud.githubusercontent.com/assets/2263037/22674157/969c25d0-ecc5-11e6-9d7e-5372dd84017c.png)

artsjedi:
OOOooKaay ,  I found the bug..

`event `is not defined in this scope.
` if (navigator.isCocoonJS && event.pointerType === 'touch') {`

interactionEvent is the correct name for this variable
`if (navigator.isCocoonJS && interactionEvent.pointerType === 'touch') {`

solve the error  here #3703

andrewstart:
That was the only problem? Wasn't the error thrown elsewhere?

[...]

PR #3703, opened 2017-02-07
Fix undefined "event" error

artsjedi:
"event" was undefined in this scope,  fixed to the correct variable name.
refer to this issue #3668

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -1430,7 +1430,7 @@ export default class InteractionManager extends EventEmitter
         // This is the way InteractionManager processed touch events before the refactoring, so I've kept
         // it here. But it doesn't make that much sense to me, since mapPositionToPoint already factors
         // in this.resolution, so this just divides by this.resolution twice for touch events...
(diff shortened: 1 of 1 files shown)


artsjedi:
refer to this issue #3668

src/interaction/InteractionManager.js
```diff
@@ -1430,7 +1430,7 @@ export default class InteractionManager extends EventEmitter
         // This is the way InteractionManager processed touch events before the refactoring, so I've kept
         // it here. But it doesn't make that much sense to me, since mapPositionToPoint already factors
         // in this.resolution, so this just divides by this.resolution twice for touch events...
-        if (navigator.isCocoonJS && event.pointerType === 'touch')
+        if (navigator.isCocoonJS && interactionEvent.pointerType === 'touch')
```
andrewstart: It actually should be `pointerEvent` - that is the one with a `pointerType` property.

artsjedi: ok, I will change, But both of them works on cocoon.

andrewstart:
Oh, this is targeting my bugfix branch, which I had completely removed the commits that included logs from. I'll just add this to the branch directly.

PR #3707, opened 2017-02-07
Fast rect blendMode problem

ivanpopelyshev:
It seems that fastrected graphics doesn't have blendMode. Tested it on pixi-examples, there is a problem.


cursedcoder:
got rect!

PR #3711, opened 2017-02-07
LocalBounds are calculated through getBounds if worldTransform is easy

ivanpopelyshev:
We don't have to ruin transforms if they are not rotated to calculate local bounds. 

Tests cover all of getLocalBounds() possible usage.

This may break if cacheAsBitmap , because it uses getLocalBounds() to actually change world transforms. CacheAsBitmap is not covered by tests, lets check it on examples.

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -309,3 +309,19 @@ export const TEXT_GRADIENT = {
     LINEAR_VERTICAL: 0,
     LINEAR_HORIZONTAL: 1,
 };
+
+/**
+ * Constants that define the type of gradient on text.
+ *
+ * @static
+ * @constant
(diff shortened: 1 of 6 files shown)


ivanpopelyshev:
Confirmed: cacheAsBitmap is broken with it. Just run cacheAsBitmap example with rotation=0

@GoodBoyDigital can we please not to rely on side-effects of getLocalBounds() ?

staff0rd:
Yeah there are some bugs already existing with cacheAsBitmap, I'll look at adding tests for those soon.

bigtimebuddy:
@ivanpopelyshev can we close this? Have we moved on?

ivanpopelyshev:
I can re-open it later.

PR #3717, opened 2017-02-09
addChild/removeChild transform IDs were wrong

ivanpopelyshev:
Emergency!

Please, don't just copy&paste code to use in tests if you are working with update ID's. Try to make them fall with real getBounds().

Also, changes that I applied in Container were rewinded by some of your refactoring.

Test "should recalculate added child correctly" contains simplest case in which it fails to update the child transform and bounds. 

Line after "// Oops, that can happen sometimes!" might be removed because right now it happens every time in this test :) That line is needed if someone changes implementation - that way it'll break for sure instead of some rare cases.

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -73,11 +73,12 @@ export default class Container extends DisplayObject
             }
 
             child.parent = this;
+            // ensure child transform will be recalculated
+            child.transform._parentID = -1;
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
That thing can also affect cacheAsBitmap and generateTexture and other things.

PR #3721, opened 2017-02-09
fix slot #0 texture

ivanpopelyshev:
If some texture already exists in slot #0, and syncUniforms is trying to use it - mask wont be rendered at all.

Fiddle: https://jsfiddle.net/jdngw3xy/


bigtimebuddy:
Here's the fiddle using this PR:
https://jsfiddle.net/jdngw3xy/1/

ivanpopelyshev:
Sorry, made wrong move :) its ok now

PR #3722, opened 2017-02-09
addChild/removeChild transform IDs were wrong

ivanpopelyshev:
I thought that's critical, but i think it waits for 4.4, we can discuss it.

 "child.transform._parentID" can be anything at the time of adding the child, thus, collisions may happen.

Also I removed some of double-tests, that checked both addChild and addChildAt the same way.

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -288,13 +288,12 @@ export default class Container extends DisplayObject
             for (let i = 0; i < removed.length; ++i)
             {
                 removed[i].parent = null;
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
@GoodBoyDigital Detached container :( That's one of tests: addChild, update it, parentID will be 0. remove child, ask getBounds(), you'll receive previous bounds instead of new ones.

GoodBoyDigital:
awesome! That makes sense!

PR #3730, opened 2017-02-12
More thorough Texture/BaseTexture cleanup

andrewstart:
* Added `name` property to Texture for tracking what name the Texture was added to the TextureCache under.
* Added a `destroy` event to BaseTexture for taking Textures based off that BaseTexture with it when destroyed.
* Made Texture and BaseTexture work together to try to make sure that everything is cleaned out of the texture cache when they are destroyed.

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -597,6 +597,7 @@ export default class BaseTexture extends EventEmitter
             }
         }
         // An svg source has both `imageUrl` and `__pixiId`, so no `else if` here
+        // this does cleanup for SVG and canvas
         if (this.source && this.source._pixiId)
         {
             delete BaseTextureCache[this.source._pixiId];
@@ -605,6 +606,9 @@ export default class BaseTexture extends EventEmitter
         this.source = null;
 
         this.dispose();
+
+        // tell any Textures that reference this that it is time to go
+        this.emit('destroy', this);
     }
 
     /**
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -217,6 +217,14 @@ export default class BaseTexture extends EventEmitter
          * @event error
          * @memberof PIXI.BaseTexture#
          */
+
+        /**
+         * Fired when destroy() is called.
+         *
+         * @protected
+         * @event destroy
+         * @memberof PIXI.BaseTexture#
+         */
     }
 
     /**
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -579,8 +579,15 @@ export default class BaseTexture extends EventEmitter
     {
         if (this.imageUrl)
         {
-            delete BaseTextureCache[this.imageUrl];
-            delete TextureCache[this.imageUrl];
+            // remove this texture from the global texture cache
+            if (BaseTextureCache[this.imageUrl] === this)
+            {
+                delete BaseTextureCache[this.imageUrl];
+            }
+            if (TextureCache[this.imageUrl] && TextureCache[this.imageUrl].baseTexture === this)
+            {
+                delete TextureCache[this.imageUrl];
+            }
 
             this.imageUrl = null;
 
--- a/src/core/textures/Texture.js
+++ b/src/core/textures/Texture.js
@@ -156,6 +156,13 @@ export default class Texture extends EventEmitter
          * @type {Object}
          */
         this.transform = null;
+
+        /**
+         * Name used to store the texture in PIXI.utils.TextureCache. Primarily used for automatic
+         * cleanup.
+         * @type {string}
+         */
+        this.name = null;
     }
 
     /**
@@ -188,6 +195,7 @@ export default class Texture extends EventEmitter
         }
 
         this.baseTexture.on('update', this.onBaseTextureUpdated, this);
+        this.baseTexture.on('destroy', this.destroy, this);
         this.emit('update', this);
     }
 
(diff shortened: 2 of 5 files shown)


src/core/textures/Texture.js
```diff
@@ -140,6 +140,7 @@ export default class Texture extends EventEmitter
         {
             baseTexture.once('loaded', this.onBaseTextureLoaded, this);
         }
+        baseTexture.on('destroy', this.destroy, this);
```
ivanpopelyshev: Good morning, memory leak! 

Now people will be forced to destroy temp Texture objects, not only temp BaseTextures.

ivanpopelyshev: I realize that we still have problem if Texture is in TextureCache and BaseTexture was destroyed, so.. lets scan TextureCache periodically? :)

andrewstart: I'm confused as to how that would be a memory leak if they are already destroying their BaseTextures.

ivanpopelyshev: Yep, that's better :)

src/core/textures/BaseTexture.js
```diff
@@ -579,8 +587,15 @@ export default class BaseTexture extends EventEmitter
...
         if (this.imageUrl)
         {
-            delete BaseTextureCache[this.imageUrl];
-            delete TextureCache[this.imageUrl];
+            // remove this texture from the global texture cache
+            if (BaseTextureCache[this.imageUrl] === this)
+            {
+                delete BaseTextureCache[this.imageUrl];
+            }
+            if (TextureCache[this.imageUrl] && TextureCache[this.imageUrl].baseTexture === this)
```
englercj: I feel that this should move to `Texture.destroy`. If you destroy textures, but not base textures the texture lives on in the cache.

src/core/textures/Texture.js
```diff
@@ -216,15 +224,17 @@ export default class Texture extends EventEmitter
     {
         if (this.baseTexture)
         {
-            if (destroyBase)
+            // remove BaseTexture stored under the texture name if destroyBase is true or
+            // is our base texture (in which case this is the event callback for 'destroy')
+            if ((destroyBase === true || destroyBase === this.baseTexture)
```
englercj: Not sure I understand what this is for. The docs weren't updated so I can only assume this is checking if you pass in a base texture? Why do we need to do this delete here and not just call baseTexture.destroy()? What do we buy from passing in a base texture as a param here, since it only does anything if the owned one is passed?

andrewstart: The `destroy` event passes the base texture that was destroyed as a param, so detecting that means detecting when this is an event callback.

src/core/textures/BaseTexture.js
andrewstart: Because `fromLoader()` adds both BaseTexture and Texture to their respective caches under both `name` and `imageUrl` keys. I was hoping to avoid adding a bunch of new variables just to have each track both keys for later removal.

src/core/textures/Texture.js
englercj: So this is destroying itself when the base texture is destroyed? I don't like that very much, seems totally unexpected that killing a child object somewhere in my code would then also destroy some other object.

Destroying a texture doesn't destroy the sprite it is on.

andrewstart: Do you often reuse a Texture that was generated from `Texture.fromImage()` after destroying its BaseTexture? This is primarily to take care of the textures generated automatically from Spritesheets and `Texture.fromImage()`, `Texture.fromLoader()`, etc. I feel that reusing the Texture instances in those cases is not a common (or even uncommon) use case.

ivanpopelyshev: This is only to take them out from cache.

englercj: I totally get that it isn't often you reuse the texture. I just don't like that some code for removing BaseTexture from the cache is in Texture and some code for removing Texture from the cache is in BaseTexture.

These two classes are supposed to be separate things, but this code ties them together so much that they almost shouldn't be separate.

englercj:
I think there are better ways to do this, but they involve deep refactoring so I'll abstain from this review.

bigtimebuddy:
I tend to agree with @englercj, I think this could be rethought to be cleaner. The `name` property addition seems useful as it's own PR. 

Overall, there's a little too much coupling going on between BaseTexture and Texture classes. Another approach might be to have the BaseTextureCache and TextureCache be actual classes (probably instances of the same class) which detects whether it's been destroyed and provides some more convenience methods for clearing the cache, etc. Potentially a rethink of the caches and how stuff is auto-removed when it's destroyed.

themoonrat:
1. I like the idea of .name property - I added it to my build under the property "textureCacheId" (for more clarity, and to make sure it didn't clash with any future pixi additions!). It's insanely useful, for example, for when I was debugging some slow renders on canvas, and I wanted to know what actual image was being drawn.

2. I think textures, baseTexture and the texture cache needs a full discussion to re-work certain aspects. Sometimes a baseTeture can contain .imageUrl, sometimes a ._pixiId, I've added a .textureCacheId; any of these could be used as keys for certain things.

3. In the end for texture cleanup I added something like
```
for (const prop in TextureCache)
{
  if (TextureCache[prop] === this)
  {
    this.textureCacheId = '';
    delete TextureCache[prop];
  }
}
```
A kinda 'I don't care what property you're looking for, just find and remove yourself in the cache'!

4. I think not enough things are stored in the texture cache. I think any base texture and texture added should go there and be identifiable. But sometimes things like text have a generic 'canvas_19' added to the BaseTextureCache, but nothing added to TextureCache

bigtimebuddy:
@andrewstart how would you like to proceed?

andrewstart:
I think I'll close this, and make another PR with 1 and 3 from @themoonrat's comment.

Issue #3681, opened 2017-01-31
PIXI.Container size changes when cacheAsBitmap is set to true


krojew:
When setting cacheAsBitmap to true, the size of the container (width, height) changes for some reason. The extra space is black. When not using caching, the size is correctly computed from its children.

staff0rd:
Hey @krojew, are you able to supply a [pen](https://codepen.io) that displays this issue you're having against the latest [dev build](//d157l7jdn8e5sf.cloudfront.net/dev/pixi.js)?

krojew:
For some reason I'm unable to make pixi draw anything in a pen. I did some investigation and this seems to be related somehow to container pivot. Here are the steps to reproduce:

1. Create a container with 135x135 grid of 32x32 sprites.
1. Cache it as bitmap.
1. Add it to the stage container.
1. Increase the stage pivot point coordinate.

In this case, the stage container gets its width and/or height increased as the pivot coordinates increase. If no caching as bitmap occurs, both width and height remain unaffected.

staff0rd:
Here's a template; https://codepen.io/pen?template=XjmggJ

krojew:
Here you go: http://codepen.io/anon/pen/WRMbyO

If you move the mouse over the view with the left button down, you should see the stage size in the console. Notice how it's changing with the pivot.

If you comment out the container.cacheAsBitmap = true; line, the size will not change with the pivot.

staff0rd:
Simplified; http://codepen.io/staff0rd/pen/wgyMrz?editors=0010

staff0rd:
Seems specific to WebGL.

WebGL: http://codepen.io/staff0rd/pen/YNOEzb?editors=0010
Canvas: http://codepen.io/staff0rd/pen/rjZYBJ?editors=0010

Note the pivot distance is visible in the WebGL render, but not in the Canvas render.

staff0rd:
Occurs on Canvas also provided you don't render between setting `cacheAsBitmap = true` and changing the pivot: http://codepen.io/staff0rd/pen/mRgeNG?editors=0010

BossLevel:
Just as an FYI, I'm managing to recreate this issue without any manipulation of the pivot - simply calling cacheAsBtmap can cause it to happen.

staff0rd:
@BossLevel pen?

BossLevel:
[...]

PR #3739, opened 2017-02-16
Started cacheAsBitmap tests

staff0rd:
Currently failing due to #3681


ivanpopelyshev:
btw, "isWebGLSupported" is changed to "withGL"

test/extras/cacheAsBitmap.js
```diff
@@ -0,0 +1,59 @@
...
+
+            expect(stage.width).to.be.equal(20);
+        });
+
+        it('on webgl without cacheAsBitmap', withGL(function ()
+        {
+            const renderer = new PIXI.WebGLRenderer(150, 150);
+            const stage = getScene(renderer, false);
+
+            expect(stage.width).to.be.equal(20);
```
englercj: Do we have a way to test that what we expect to be drawn is actually drawn?

bigtimebuddy:
@staff0rd could you fix the linting error?

bigtimebuddy:
I would love to get these tests in. @ivanpopelyshev could you look at why the CanvasRenderer test for cacheAsBitmap is failing? Looks like some math problem considering the pivot point.

bigtimebuddy:
I'm going to close this because of inactivity. Need a solution for handling `cacheAsBitmap` with pivot.

Issue #3741, opened 2017-02-16
[v4.4] Local tests failing: PIXI > renders > should draw a rectangle


staff0rd:
On my local (Win10) machine, the `should draw a rectangle` test fails with the following;

~~~
AssertionError: WebGL results do not match.
    at renderer.compare (C:\git\pixi.js\test\renders\index.js:20:21)
    at run (C:\git\pixi.js\test\renders\lib\Renderer.js:211:17)
    at done (C:\git\pixi.js\test\renders\lib\Renderer.js:168:21)
    at Renderer.run (C:\git\pixi.js\test\renders\lib\Renderer.js:181:17)
    at Renderer.compare (C:\git\pixi.js\test\renders\lib\Renderer.js:197:14)
    at Context.validate (C:\git\pixi.js\test\renders\index.js:16:27)
    at Context.<anonymous> (C:\git\pixi.js\test\renders\index.js:42:14)
    at callFnAsync (C:\git\pixi.js\node_modules\mocha\mocha.js:4494:8)
    at Test.Runnable.run (C:\git\pixi.js\node_modules\mocha\mocha.js:4446:7)
    at Runner.runTest (C:\git\pixi.js\node_modules\mocha\mocha.js:4934:10)
~~~

This only occurs when all tests run.  Running the test on its own, or running the `renders` tests only causes the test to pass.  This test _did_ last time I ran tests locally (about 1-2 weeks ago).

Any thoughts @bigtimebuddy?

ivanpopelyshev:
Same for me . Its actually blank, doesn't render anything.

bigtimebuddy:
I'm debugging this now.

bigtimebuddy:
It appears the commenting out this test: https://github.com/pixijs/pixi.js/blob/dev/test/core/Graphics.js#L334-L363 makes the render test pass. @ivanpopelyshev you created that test, any idea why it might be failing. 

I tried destroying the WebGLRenderer, but that didn't do anything. Puzzling.

ivanpopelyshev:
I think we have a major problem guys. That test uses fastRect! It means that texture that is cached globally is wrong again. I'm assigning this task to me.

englercj:
Looks like #3758 was merged, closing this now.

PR #3743, opened 2017-02-16
Resolves issue with fastRect

bigtimebuddy:
Address #3741 by not creating a single static texture for fast rectangle in Graphics.

Allows the WebGL tests to pass on Windows.


bigtimebuddy:
Replaced with #3758

PR #3747, opened 2017-02-17
The `orig` should be as same as `trim`: use relative region [v5]

finscn:
When we set `orig` ,  use `local/relative region` will be easy. 
And It also should be as same as `trim`.

And if  `orig` use absolute value of  the image,  when users repack the images in a sprite-sheet , they have to recompute the `orig`.


finscn:
BTW : In current PIXI, there is no any code use `orig.x` or `orig.y` , 
so this PR is not a breaking changing.

ivanpopelyshev:
There will be problems with it.  Some users are using only "frame" and they expect that when "frame.width" is changed , "orig.width" will change too.

finscn:
Ok, I see.  I think you should say something about this in document.
The current document is like this :  
```
     * @param {PIXI.Rectangle} [orig] - The area of original texture
     * @param {PIXI.Rectangle} [trim] - Trimmed rectangle of original texture
```

They both use  `of original texture`, but `orig` is absolute value ,  `trim` is relative value.
Too confusing.

.

ivanpopelyshev:
Yeah, lets leave it at that for now. But its better than it was in v3 anyway ;)

ivanpopelyshev:
I understand that we have to change it to "trim.x - orig.x" to fix that, right?

finscn:
@ivanpopelyshev  , please don't forget this pr

bigtimebuddy:
I'm going to close this again @finscn, if you want this for v5, the **next** branch is ready for PRs if you want to submit there for consideration.

finscn:
I see your mean. 
But , I think If you often use `close` instead of `add a tag about  [v5]`(or sth. else)  , I think many good suggestions/PRs will be forgotten.

bigtimebuddy:
I added v5 tag. Will sort through these when we are further along.

PR #3749, opened 2017-02-17
Fix precision setting change not being picked up

themoonrat:
PIXI.settings.PRECISION = 'highp';
Wasn't getting picked up in my game.

It turns out, the style of using the PRECISION setting meant it was a static value. This has been changed so that it's looking to the reference of the settings object to pick up the live changed value


themoonrat:
Fixes: https://github.com/pixijs/pixi.js/issues/3742

bigtimebuddy:
Good find!

PR #3750, opened 2017-02-17
Allows elapsedMS to show up in the docs

themoonrat:
Fixes: https://github.com/pixijs/pixi.js/issues/3732


themoonrat:
For some reason, the 1 / TARGET_FPMS part. Maybe it looked too much like code? Failing at trying to do the actual maths? No idea, but changing to an actual number fixed it

Issue #3742, opened 2017-02-16
Wobbling spritesheet animations on iOS


themoonrat:
Example: http://moonrat.co.uk/html5/pixi-ios-animation-bug/

What you see is 3 sizes of an animation, x0.5 at the top, then x0.75 in the middle and the x1 at the bottom. On the left hand side the regular rendering path is taken, on the right hand side the three animations use pixi-picture.

Now, I _only see this issue on iOS devices in WebGL mode_. Any other browser, even Safari on a Mac, works absolutely fine. iOS in canvas mode, absolutely fine. But iOS (iPads and iPhones) devices using WebGL... wobble

@ivanpopelyshev suggested giving pixi-picture a go, for more accurate rendering, and it does help on most devices, but not on all, and it very rarely eliminates the issue

Here is a list of how all of the iOS devices I have look on the above example

iPad Mini 2: 10.0.2 - pixi shakes, pixi-picture doesn't
iPad 4: 8.4 - pixi shakes, pixi-picture does, but less so
iPad Air: 10.2 - pixi shakes, pixi-picture doens't
iPad Air: 10.3 - pixi shakes, pixi-picture doesn't
iPad Air2: 8.3 - pixi shakes, pixi-picture does, but less so

iphone 4s: 9.1 - pixi shakes, pixi-picture does, but less so
iphone 4s: 9.2.1 - pixi shakes, pixi-picture does, but less so
iphone 5: 9.1 - pixi shakes, pixi-picture does too
iphone 5: 10.1 - pixi shakes, pixi-picture does too
iphone 5s: 10.0.2 - pixi shakes, pixi-picture doesn't
iphone 6: 8.2 - pixi shakes, pixi-picture does, but less so
iphone 6: 10.0.2 - pixi shakes, pixi-picture does, but less so
iphone 6+: 8.4 - pixi shakes, pixi-picture does, but less so
iphone 6+: 9.2.1 - pixi shakes, pixi-picture does, but less so
iphone se: 10.2 - pixi shakes, pixi-picture doesn't
iphone 7: 10.2 - pixi shakes, pixi-picture does, but less so

Any ideas?

themoonrat:
Additional info; this wobble does NOT occur on the last pixi v3: http://moonrat.co.uk/html5/pixi-ios-animation-bug-v3/

ivanpopelyshev:
v3 is better? that's interesting

LePhasme:
[...]

PR #3753, opened 2017-02-18
[v4.4] different settings for vertex and fragment shader

ivanpopelyshev:
Follow #3742 #3749

Best practice: use highp on vertex shaders, mediump on fragment. 

Use https://dl.dropboxusercontent.com/u/46636530/pixijs/wobble/pixi.js if you want to test it with bunnymark. Sorry, I prefer not to create branches that I forget to delete later ;)

### DIFF
--- a/src/core/Shader.js
+++ b/src/core/Shader.js
@@ -1,7 +1,7 @@
 import { GLShader } from 'pixi-gl-core';
 import settings from './settings';
 
-function checkPrecision(src)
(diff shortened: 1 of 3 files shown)


1N50MN14:
@ivanpopelyshev This change broke pixi-plugin-example

ivanpopelyshev:
@1N50MN14 oops. Thank you, will fix it :)

PR #3756, opened 2017-02-18
Fix #3755

### LINKED ISSUE #3755: [BUG] about `utils.getResolutionOfUrl`
If there are multi `@` in url ,  the result will be wrong.

finscn:
If there are multi '@' in url ,  only reserve the last one,  then exec RETINA_PREFIX.

### DIFF
--- a/src/core/utils/index.js
+++ b/src/core/utils/index.js
@@ -114,13 +114,14 @@ export function getResolutionOfUrl(url, defaultValue)
     {
         const lastIndex = url.lastIndexOf('@');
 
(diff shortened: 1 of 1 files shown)


themoonrat:
Surely the solution to this is to fix the default RETINA_PREFIX regex? Your fix assumes that an @ sign is going to be included in that setting, but it doesn't have to be.

finscn:
@themoonrat  ,  `@` + `number` + `x` is  common & customary.
If there is no `@` ,  users should override the `getResolutionOfUrl` by themselves.

Issue #3757, opened 2017-02-18
There is always an error when `npm run test`


finscn:
```
2) PIXI renders should draw a rectangle:
     AssertionError: WebGL results do not match.
      at renderer.compare (/Users/rocky/DEV/@git-fav/_pixijs.com/pixi.js/test/renders/index.js:20:21)
      at run (/Users/rocky/DEV/@git-fav/_pixijs.com/pixi.js/test/renders/lib/Renderer.js:211:17)
      at done (/Users/rocky/DEV/@git-fav/_pixijs.com/pixi.js/test/renders/lib/Renderer.js:168:21)
```

I test the `pixi.js dev` ,not my fork version.

Dimous:
#3741

finscn:
@Dimous , thanks.

PR #3758, opened 2017-02-18
[v4.4] fix SpriteRenderer for multiple webgl renderers, fix tests, add "withGL"

ivanpopelyshev:
Fixes problem in SpriteRenderer, TICK was set to 0 every time webglrenderer was created. That was a cause that our tests were failing.

Yes, I know that TICK++ will overflow bunnymark after few years of work, I can add "if tick>2^52, tick=0" there. Same with our transforms, so lets not address that issue there ;)

This is cumulative patch, it also adds some APIs that can be used in tests and in runtime:

- added Texture.WHITE, 10x10 texture is used by Graphics for spriteRect
- cleaned spriteRect code
- added clearTextureCache() and destroyTextureCache()
- added withGL transformer to simplify webgl tests

### DIFF
--- a/test/withGL.js
+++ b/test/withGL.js
@@ -1,23 +1,8 @@
 'use strict';
 
-function withGL(init, fn)
+function withGL(fn)
 {
-    if (!PIXI.utils.isWebGLSupported())
-    {
-        return undefined;
-    }
-
-    const renderer = init();
-
-    try
-    {
-        return fn(renderer);
-    }
-    finally
-    {
-        PIXI.utils.destroyTextureCache();
-        renderer.destroy();
-    }
+    return PIXI.utils.isWebGLSupported() ? fn : undefined;
 }
 
 module.exports = withGL;
--- a/test/withGL.js
+++ b/test/withGL.js
@@ -1,0 +1,23 @@
+'use strict';
+
+function withGL(init, fn)
+{
+    if (!PIXI.utils.isWebGLSupported())
+    {
+        return undefined;
+    }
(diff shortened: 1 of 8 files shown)


ivanpopelyshev:
Fixes #3757

ivanpopelyshev:
Fixes #3741

test/withGL.js
```diff
@@ -0,0 +1,23 @@
...
+    const renderer = init();
+
+    try
+    {
+        return fn(renderer);
+    }
+    finally
+    {
+        PIXI.utils.destroyTextureCache();
+        renderer.destroy();
```
bigtimebuddy: I am not crazy about this abstraction. It obfuscates more than provides a useful utility. Please just add the renderer and destroy functions within the test code. This seems unnecessary. Also you can use the `afterEach` method in Mocha to always destroy the TextureCache.

ivanpopelyshev: That is one less tab-level, and that kind of abstractions are used widely in modern languages. Its like try-with-resources from java/scala.

ivanpopelyshev: Also, it is possible that I'll add more checks in "finally" later, to check for some wrong renderer states

ivanpopelyshev: And I kinda don't like isWebGLSupported() function ;)

ivanpopelyshev: One more example: we can change some renderer options on fly and run test multiple times instead of one, like what I do in "container.addChild" tests.

PR #3759, opened 2017-02-18
Improve the accuracy of RETINA_PREFIX

finscn:
Sometimes ，this PR could fix https://github.com/pixijs/pixi.js/issues/3755 .
Maybe this PR is better than https://github.com/pixijs/pixi.js/pull/3756.

### DIFF
--- a/src/core/settings.js
+++ b/src/core/settings.js
@@ -77,7 +77,7 @@ export default {
      *
      * @static
      * @memberof PIXI.settings
-     * @type {RegExp|string}
+     * @type {RegExp}
      * @example `@2x`
      * @default /@(.+)x/
      */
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
I think this is a better approach.

finscn:
@bigtimebuddy , I found the doc says : 
```
     * @type {RegExp|string}
```
But if  RETINA_PREFIX is string , the  `RETINA_PREFIX.exec` will take an error .

Is there  a fault in the doc ?

bigtimebuddy:
Yes, seems like it should be a RegExp only.

src/core/settings.js
```diff
@@ -81,7 +81,7 @@ export default {
      * @example `@2x`
      * @default /@(.+)x/
```
staff0rd: `@default` needs to be updated

finscn: Done.

PR #3766, opened 2017-02-20
Fix `the fault after merge`

finscn:


ivanpopelyshev:
Wait a minute, how the heck "true" was there instead of false? I didn't do it!

finscn:
@ivanpopelyshev  , I think it's a github's issue.

At first  , it  is `true`. (version 1 )
Then I  open this PR ,  it is still `true` (version 2)
Then Someone change `true` to `false` (version 3)

When merge my PR , the github can't merge correctly.

finscn:
@ivanpopelyshev , I think github should report `conflict` , but it didn't.

bigtimebuddy:
Either way, thanks for catching this @finscn

Issue #3770, opened 2017-02-20
[poll] Should we drop Bower Support?


bigtimebuddy:
Some of the PixiJS team were having a [conversation](https://twitter.com/rolnaaba/status/793507889486770177) about dropping support for Bower. We feel strongly that supporting this dependency manager hurts us more than it helps. Looking for some feedback from the community before we make a change like this.

## Approach
Propose that we maintain the current **release** branch as is, but future releases will be tagged against the **master** branch instead. This would effectively maintain Bower support for the current release and below to be as non-disruptive as possible. Future versions would be available only as direct downloads, from CDNs and through NPM/Yarn.

### Pros
* Would allow us to completely remove our custom deployment system which uses snapshots on a 'release' branch to keep from uploading build files. Would simplify things.
* NPM/Yarn is more robust/less brittle dependency management system and has more features than Bower.
* It would be easier to see version diffs (with commits) instead of comparing snapshots. This would help make release notes easier and easier to do binary searches.

### Cons
* Bower PixiJS consumers would need to switch to NPM, CDN downloads (e.g., [cdnjs.com](https://cdnjs.com/libraries/pixi.js/)) or some other method in their projects.

**If you are in favor, please give it a** 👍 
**If you are **not** in favor, give it a** 👎 

Or please leave your thoughts.

macguffin:
(y)

neutrino84:
I currently use bower, and there is no reason for my pixi.js code to be coming in through bower. I support dropping bower.

staff0rd:
Thanks for voting everyone, [bower is out](https://github.com/pixijs/pixi.js/issues/3772).

PR #3772, opened 2017-02-21
Drops support for Bower

bigtimebuddy:
Follow-up to #3770.

This PR deprecates the support for Bower. It removes the **bower.json** file and the custom release/publish script to copy all the contents to the **release** branch.

All future releases will be git-tagged against the **master** branch.


bigtimebuddy:
@englercj Cool. I added S3 deployment for "release", which is a good suggestion.

Issue #3774, opened 2017-02-22
JSDoc warnings


bigtimebuddy:
Following warnings are generated when `npm run docs` is run. Could someone look into this (cc @themoonrat)

```
WARNING: The @type tag does not permit a description; the description will be ignored. File: InteractionTrackingData.js, line: 39
WARNING: The @type tag does not permit a description; the description will be ignored. File: InteractionTrackingData.js, line: 43
WARNING: The @private tag does not permit a value; the value will be ignored. File: interactiveTarget.js, line: 87
WARNING: The @private tag does not permit a value; the value will be ignored. File: interactiveTarget.js, line: 92
```

LucianoGraziani:
I'll do it!

themoonrat:
Please do, would appreciate that a lot! I'm on holiday atm and not back til Friday :)

bigtimebuddy:
Thanks @LucianoGraziani!

lgraziani2712:
It didn't close automatically :(

PR #3776, opened 2017-02-22
Fixed issue #3774

LucianoGraziani:
Done!

By the way, eslint has this rule: http://eslint.org/docs/rules/valid-jsdoc for (simple) jsdoc validation. If you want, I can do a PR with it :)

### DIFF
--- a/src/interaction/InteractionTrackingData.js
+++ b/src/interaction/InteractionTrackingData.js
@@ -36,7 +36,7 @@ export default class InteractionTrackingData
 
     /**
      * @readonly
-     * @type {number} Unique pointer id of the event
+     * @return {number} Unique pointer id of the event
      */
     get pointerId()
     {
--- a/src/interaction/InteractionTrackingData.js
+++ b/src/interaction/InteractionTrackingData.js
@@ -35,8 +35,10 @@ export default class InteractionTrackingData
     }
 
     /**
+     * Unique pointer id of the event
+     *
      * @readonly
-     * @return {number} Unique pointer id of the event
+     * @member {number}
      */
     get pointerId()
     {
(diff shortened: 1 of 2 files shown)


src/interaction/InteractionTrackingData.js
```diff
@@ -36,7 +36,7 @@ export default class InteractionTrackingData
 
     /**
      * @readonly
-     * @type {number} Unique pointer id of the event
+     * @return {number} Unique pointer id of the event
```
themoonrat: Do not agree with this one. Getters are functions to replicate member variables. It should say member rather than type. Definitely not return

src/interaction/interactiveTarget.js
```diff
@@ -87,7 +87,8 @@ export default {
     /**
      * Map of all tracked pointers, by identifier. Use trackedPointers to access.
      *
-     * @private {Map<number, InteractionTrackingData>}
+     * @private
+     * @type {Map<number, InteractionTrackingData>}
```
themoonrat: Agree with this tho

src/interaction/InteractionTrackingData.js
LucianoGraziani: I was in doubt about using member or return. Didn't knew his purpose 👍

LucianoGraziani:
Updated!

bigtimebuddy:
I'm pretty sure we are already using valid-jsdoc rule in ESLint.

LucianoGraziani:
Yes, you do. I searched at the end of the eslint file thinking were ordered alphabetically.

bigtimebuddy:
@themoonrat you good here?

PR #3785, opened 2017-02-26
Don't call  getter , use  `texture.orig.*` directly.

finscn:
Don't call  getter of `texture.width`/`texture.height`.

 in other places in PIXI , when need texture's size , PIXI use texture.orig.width/height . Why only this place use texture.width/height ?

I've said : " Do similar things in similar way" .
So , whatever use texture.width/height or texture.orig.width/height , The key point is "please use one style".

~~Why don't use `texture.width/height`  :  
see https://github.com/pixijs/pixi.js/issues/3782 . So I think if there is need  `orig.size` , we should use `orig.size` directly.~~

~~The  4 sizes :~~
~~* orig.size~~
~~* frame.size~~
~~* trim.size~~
~~* texture.size~~

~~are different things.~~


ivanpopelyshev:
texture size is the same as orig size

finscn:
@ivanpopelyshev  , in other places in PIXI , when need texture's size , PIXI use `texture.orig.width/height` . Why only this place use `texture.width/height` ?

I've said : " Do similar things in similar way" .
So , whatever use `texture.width/height`  or `texture.orig.width/height` , The key point is "please use one style".

finscn:
@ivanpopelyshev  , please re-think about this issue , thanks.

finscn:
@ivanpopelyshev , I hope the code in the same style.  like  `sprite._texture`

ivanpopelyshev:
You can close this one, i'll do those changes in dev-texture

bigtimebuddy:
@finscn could you update with the latest from dev?

finscn:
@bigtimebuddy  , done.

bigtimebuddy:
@finscn Thanks for all your pull-requests concerning code formatting, consistency and minor optimization issues. We would _love_ to see a larger more substantial pull-request with a bunch of these things together rather than many minor pull-requests.

PR #3789, opened 2017-02-26
Small Texture Cache enhancements

andrewstart:
From the dicussion of #3730:

* added `textureCacheId` to Texture, for giving a useful name to Texture objects that hopefully matches what they are in the TextureCache as.
* Textures and BaseTextures do a thorough search of their respective caches in `destroy()` to remove themselves, so that subsequent Texture.fromX calls don't attempt to reuse a destroyed object.
* Removed some event listener cleanup in Texture.destroy() that wasn't actually cleaning up anything that ever got added.

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -579,7 +579,6 @@ export default class BaseTexture extends EventEmitter
     {
         if (this.imageUrl)
         {
-            delete BaseTextureCache[this.imageUrl];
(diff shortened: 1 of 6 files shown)


bigtimebuddy:
This looks great @andrewstart. One small requests: could you possibly create some simple unit-tests for making sure a BaseTexture/Texture with multiple aliases is removed? As well as making sure textureCacheId is set correctly?

andrewstart:
New tests have been added.

Issue #3790, opened 2017-02-27
[v4.4.0] Click event of a Sprite is sent even if a DOM element is above


manudurgoni:
Hey !

Since the 4.4.0, I observed that when I click on a DOM element which is above a sprite on the pixi's canvas. The click event of the sprite is still fired.

I made a quick example here : 
v4.4.0
https://jsfiddle.net/ku55deL6/5/

v4.3.5
https://jsfiddle.net/ku55deL6/4/

And thanks guys for your awesome work !

manudurgoni:
Thanks for your fix. But the issue is not completely fixed.

If you click multiple times on a DOM element above a Sprite, I can reproduce the issue.

Example with v4.4.1 : https://jsfiddle.net/ku55deL6/8/

ivanpopelyshev:
Nice :) the adventure continues!

andrewstart:
Okay, https://github.com/pixijs/pixi.js/pull/3800 should fix another problem that was causing this.

bigtimebuddy:
@andrewstart could you verify that fiddle with your build?

andrewstart:
Yep, it works. No matter how many times I click the yellow banner, the bunny stays put.

FNevesFMQ:
Hey guys!

I tried with version 4.4.2 and it seems that it's still not fixed for every situation.. if I have two sprites in the same container it works fine, but if sprites are in different containers, this problem still occurs.

See this: https://jsfiddle.net/FNevesFMQ/uwx24bm8/

andrewstart:
At this point I think you're sneaking errors into the code to make work for me 😛 
I'll have a chance to take a look at this this weekend. I know we have unit tests to enforce not getting pointer events for sprites that are overlapping, so I'll have to dig into why different containers changes that.

FNevesFMQ:
I wish 😛 my app uses this situation a lot, and unfortunately this cause me a lot of bugs..

andrewstart:
I've got this issue solved with https://d157l7jdn8e5sf.cloudfront.net/fix-interaction-issues/pixi.js, but it's probably a good idea to have you use it to see if there are any other problems we don't know about.

hazardsoft:
The patch did not work for me.
The case is the following (mobile, safari):
1. there are 2 containers, both interactive and overlap each other
2. when I click the highest one click handler is fired in the lower too

andrewstart:
@hazardsoft We currently have passing tests for the following conditions:
```
[...]

PR #3800, opened 2017-02-28
Fix another issue that was causing #3790

andrewstart:
Fixed a different issue that was helping cause mouseups to trigger clicks without mousedowns.


ivanpopelyshev:
I believe that, and "gl.flush" could be good for 4.4.2.

PR #3801, opened 2017-03-01
cacheAsBitmap is leaking objects

sivael:
When destroying objects that had cacheAsBitmap, the internal objects were being retained - options parameter decides whether the children should be destroyed.
If you pass true to a containers destroy that's being cached as bitmap, the containers contents would not be destroyed, unless we pass the parameter to the original destroy method here.

To observe:
1. Create an object containing other objects inside.
2. Cache it to bitmap
3. Destroy the object
4. Goto 1

It will leak the containers sub objects and in the end crash the browser(observable on mobile mostly where memory is much more limited)

### DIFF
--- a/src/extras/cacheAsBitmap.js
+++ b/src/extras/cacheAsBitmap.js
@@ -357,6 +357,7 @@ DisplayObject.prototype._destroyCachedDisplayObject = function _destroyCachedDis
 /**
  * Destroys the cached object.
  *
+ * @private
  * @param {object|boolean} [options] - Options parameter. A boolean will act as if all options
  *  have been set to that value. 
  *  Used when destroying containers, see the Container.destroy method.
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks @sivael! I would like to release this ASAP, so I'm manually pushing this change to **master** (c1903f94d5fe4cd965fb68fcbf128f7c2640376c). Will close this.

sivael:
@bigtimebuddy What was preventing this PR from merging? 
Initially it was LINT that prevented it but I've fixed it since.

bigtimebuddy:
I'm not positive, but it looks the Travis never completed (or started) the comment commits. That's needed in order to be merged.

sivael:
Ok, thanks:) Thought i did something wrong there.

bigtimebuddy:
No worries, thanks for the addition. I made sure to add your name the commit message, so you got credit for this.

sivael:
No problem, glad to help:)

PR #3802, opened 2017-03-01
Fix preparation of destroyed items

andrewstart:
* Added `isDestroyed` to DisplayObject, BaseTexture, which is automatically set in `destroy()`.
* The prepare plugins will automatically skip destroyed items, by checking the `isDestroyed` property.

### DIFF
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -113,6 +113,14 @@ export default class DisplayObject extends EventEmitter
          * @private
          */
         this._mask = null;
+
+        /**
+         * If the object has been destroyed via destroy(). If true, it should not be used.
+         *
+         * @member {boolean}
+         * @readonly
+         */
+        this.isDestroyed = false;
     }
 
     /**
@@ -403,6 +411,8 @@ export default class DisplayObject extends EventEmitter
 
         this.interactive = false;
         this.interactiveChildren = false;
+
+        this.isDestroyed = true;
     }
 
     /**
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -118,9 +118,10 @@ export default class DisplayObject extends EventEmitter
          * If the object has been destroyed via destroy(). If true, it should not be used.
          *
          * @member {boolean}
+         * @private
          * @readonly
          */
-        this.isDestroyed = false;
+        this._destroyed = false;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


englercj:
I've seriously resisted the `isDestroyed` markers for a while. destroy() should just remove itself from the pixi ecosystem of things, and users should drop references. Anything else is an error and should be corrected, not worked around by checking if something is destroyed, imo.

GoodBoyDigital:
I agree with chad on tho one!

ivanpopelyshev:
That marker is better than events or signals. Of course it is only for internal usage or dirty production hacks. We don't have to promote it as solution for user problems. We have many hacks that we made for performance causes already.

ivanpopelyshev:
Can we mark it in docs like "for internal usage" ?

andrewstart:
I agree with the sentiment about what `destroy()` should do, but this is a situation where the thing being destroyed has no access to the renderer that might be preparing it, which limits us to setting a variable or emitting an event. We certainly could make the property internal to the library, if that would assuage your concerns.

bigtimebuddy:
My preference would be to have this be an internal flag only (e.g., `_destroyed`) documented as private, since it is only used by internal classes.

bigtimebuddy:
Also, @andrewstart this PR should go to **dev** not **master**.

bigtimebuddy:
@andrewstart could you describe your use-case more specifically why this flag is necessary? Why are destroyed items being sent to the prepare plugin that cannot be mitigated by better resource management dev-side?

andrewstart:
The use case is when one is preparing one or more items to be used, it could take a bit of time due to rate limiting to avoid hitting the framerate too hard. Sometimes, due to user input, the items being prepared need to be discarded (out of view, exiting state to a different game state, etc) before they can be prepared.

englercj:
@andrewstart My follow question would be: why can't the destroy method send out a notification that the prepare system uses to know to remove that item from its list of items.

andrewstart:
Our options are emitting an event or marking it with a private variable. My preference is the variable, because it should be slightly lighter weight, but if the consensus is that this should use events, I'll change it.

bigtimebuddy:
@englercj or @GoodBoyDigital do you want to weigh-in on Andrew's options? I don't have a strong preference on way or another, but overall, I support this PR.

englercj:
Like I mentioned, I'm abstaining :)

bigtimebuddy:
I say private variable @andrewstart!

andrewstart:
It has now been changed to `_destroyed` and is marked as private.

bigtimebuddy:
**Please don't merge this right away.** Need to confirm *dev* vs *master* with @andrewstart.

PR #3807, opened 2017-03-05
Warn on null view

### LINKED ISSUE #3792: autoDetectRenderer should trigger an error if supplied view was not found
```javascript
var myView = document.getElementById('fooMap');
console.log(myView); //is null
var stage = new Container(),
renderer = autoDetectRenderer(window.innerWidth, window.innerHeight, {view: myView});
```

Using pixi.js 4.4.0.

I had a small typo failure (Id of my canvas was "barMap").
And wondered that nothing happends even dont get any errors.
It's a small thing but maybe it would be nice if this case triggers an error...

staff0rd:
Resolves #3792

### DIFF
--- a/src/core/renderers/SystemRenderer.js
+++ b/src/core/renderers/SystemRenderer.js
@@ -82,6 +82,11 @@ export default class SystemRenderer extends EventEmitter
          */
         this.screen = new Rectangle(0, 0, options.width, options.height);
 
+        if (options.view === null)
+        {
+            console.warn('View was null, will create.'); // eslint-disable-line no-console
+        }
+
         /**
          * The canvas element that everything is drawn to
          *


englercj:
Isn't not passing an explicit canvas a common use-case?

ivanpopelyshev:
@englercj 
That case is "undefined".

staff0rd:
@englercj right, this is an unexpected null check, not undefined which is expected.

bigtimebuddy:
So null check is useful for checking if nothing is returned from getElementById or querySelector? I'm not convinced we should solve this problem, which is a user-error querying the DOM incorrectly.

bigtimebuddy:
I'm going to close this for reasons mentioned above:
* `view` is always defaulted to `null` (not `undefined`), which invalidates the utility of this PR
* checking for a user's valid DOM query is outside scope of PIXI

PR #3812, opened 2017-03-07
prepend-middleware

masarakki:
I want to insert middleware before all of presets.
For example, in my case, I must decode DRM at first.

### DIFF
--- a/src/loaders/loader.js
+++ b/src/loaders/loader.js
@@ -90,6 +90,17 @@ export default class Loader extends ResourceLoader
     {
         Loader._pixiMiddleware.push(fn);
     }
+
+    /**
+     * Prepends a default middleware to the pixi loader.
+     *
+     * @static
+     * @param {Function} fn - The middleware to prepend.
+     */
+    static prependPixiMiddleware(fn)
+    {
+        Loader._pixiMiddlware.unshift(fn);
+    }
 }
 
 // Copy EE3 prototype (mixin)


bigtimebuddy:
Thank you for this @masarakki. I can see the use-case here. 

I've always somewhat disliked the `addPixiMiddleware` API name (the "Pixi" part seems superfluous). Maybe we can take this opportunity to improve and create a single install middleware method here, instead of creating two separate methods. Something like this:

```js
PIXI.loaders.Loader.installMiddleware(func, prepend = false);
```

We'd need to deprecate `addPixiMiddleware`, but that's pretty straightforward in the **src/deprecation.js** file.

Thoughts?

englercj:
I feel like we should move this up a level and fix this at the resource loader level:

https://github.com/englercj/resource-loader/issues/74

If resource loaded supported default middlewares on a loader pixi wouldn't have to worry about this at all (except to use it to add the pixi-specific middleware). Thoughts?

masarakki:
@englercj  I think it's a good idea

But actually, I want to use middleware?(or just a function) each resource loading. like this:

```js
const resources = [
   {url: 'file_a', drmkey: 'abc'},
   {url: 'file_b', drmkey: 'xyz'},
   {url: 'file_c', drmkey: null}
];

resources.each(resource => {
   const { url, drmkey } = resource;
   if (drmkey) {
     Pixi.loader.add(url, {
       beforeMiddleware: (res => decodeDrm(res, drmkey))
     }
   } else {
      Pixi.loader.add(url);
   }
}
```

is It `resource-loader` 's scope?

englercj:
This is already supported by the loader:

```
// resources, the `metadata` key is passed through the resource so we can check it in the middleware
const resources = [
   { url: 'file_a', metadata: { drmkey: 'abc' } },
   { url: 'file_b', metadata: { drmkey: 'xyz' } },
   { url: 'file_c', metadata: { drmkey: null } }
];

// Simple middleware that runs decodeDrm on each resource after loading it from the url.
// If you need this to happen *before* loading a resource, use .pre() instead of .use().
Pixi.loader.use(function (res, next) {
    if (!res.metadata || !res.metadata.drmkey)
        return next();

    decodeDrm(res, res.metadata.drmkey);

    return next();
});

// you can pass an array of resources
Pixi.loader.load(resources);
```

bigtimebuddy:
@masarakki I'm going to close this since @englercj provided a good workaround and we will handle installing global middleware in 2.1 of **resource-loader**.

masarakki:
ok, thanks!

PR #3819, opened 2017-03-09
Use texture transform for Mesh

ivanpopelyshev:
Test it here: https://dl.dropboxusercontent.com/u/46636530/pixijs/mesh/pixi.js

My tests work for Canvas and WebGL.

Yes, I know there's upcoming dev-geometry-update , but I'm waiting for it for really long time, and people are asking for this feature.

### DIFF
--- a/src/mesh/Mesh.js
+++ b/src/mesh/Mesh.js
@@ -1,4 +1,5 @@
 import * as core from '../core';
+import { default as TextureTransform } from '../extras/TextureTransform';
 
 const tempPoint = new core.Point();
 const tempPolygon = new core.Polygon();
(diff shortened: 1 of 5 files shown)


ivanpopelyshev:
It turns out, "PIXI.mesh.Plane" had some calculations based on texture, which is not right. I will think whether or not add "uvRespectFrame" variable that enables that mode for mesh.

bigtimebuddy:
@ivanpopelyshev Could you describe (or preferably show) more about what this change does? I'm not as familiar with the TextureTransform and what you're trying to achieve.

Also, @GoodBoyDigital wondering what your thoughts are on this. Does this work with your current refactor? I don't want to create a collision for v4.5.

ivanpopelyshev:
I introduced TextureTransform back when I did big TilingSprite refactoring. Users often bake tilingSprite and Mesh textures into atlases, so its better if the conversion of UVs will be on shader side.

I'm waiting for @GoodBoyDigital dev-geometry-update to start my own shader/geometry refactoring. My future patch will add support for both modes, it will depend on shader.

 I think I'll close this one but open new one in separate branch of main repo.

ivanpopelyshev:
This is WIP for #3819 , I realized we need radical solution :)

PR #3820, opened 2017-03-09
Dev texture update

ivanpopelyshev:
Rewamping Mesh usage of texture.

Test it here: http://pixijs.download/dev-texture-update/pixi.js

1. Texture transform is not applied to mesh on every texture changes
2. For manual meshes, user can call "multiplyUvs" if they use atlas
3. Generated meshes are overriding special method "refresh", they have to call "multiplyUvs" to apply texture transform.
4. Texture transform can be uploaded to GPU with "uploadUvTransform" setting of mesh

I didn't implement tests yet, I just checked examples, including some custom atlas stuff.

Yes, I know there's upcoming dev-geometry-update , I was waiting for it for really long time, and people are asking for this feature.

Tests in examples

open [textured-mesh](http://pixijs.github.io/examples/?v=dev-texture-update#/basics/textured-mesh.js)

1. add "if (strip.texture.valid) strip.texture.rotate = 4;" in the ticker, snake will look the other way! 12 works too,
2. add "strip.uploadUvTransform=true", and it works too! Of course its better to call "refresh(true)" after that, but texture isn't loaded yet, so its ok ;)

### DIFF
--- a/src/mesh/Rope.js
+++ b/src/mesh/Rope.js
@@ -1,5 +1,4 @@
 import Mesh from './Mesh';
-import * as core from '../core';
 
 /**
  * The rope allows you to draw a texture across several points and them manipulate these points
@@ -52,13 +51,11 @@ export default class Rope extends Mesh
         this.indices = new Uint16Array(points.length * 2);
 
         /**
-         * Tracker for if the rope is ready to be drawn. Needed because Mesh ctor can
-         * call _onTextureUpdated which could call refresh too early.
-         *
+         * refreshes vertices on every updateTransform
          * @member {boolean}
-         * @private
+         * @default true
          */
-        this._ready = true;
+        this.autoUpdate = true;
 
         this.refresh();
     }
@@ -67,7 +64,7 @@ export default class Rope extends Mesh
      * Refreshes
      *
      */
-    refresh()
+    _refresh()
     {
         const points = this.points;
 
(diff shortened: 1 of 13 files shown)


ivanpopelyshev:
This is WIP for #3819 , I realized we need radical solution :)

cursedcoder:
+1

but I would update docblocks formatting, should follow "Real sentence rules."

ivanpopelyshev:
@finscn my PR fixes it - you can just rotate the texture.

bigtimebuddy:
Why is TextureTransform being moved to core (from extras) if it's only used by Mesh/Plane/Rope?

ivanpopelyshev:
Because originally it was used in TilingSprite, and someone (i dont remember) asked me to not include it to the core yet. Now its time :)

bigtimebuddy:
That was @GoodBoyDigital https://github.com/pixijs/pixi.js/pull/3048#discussion_r81526922, would like him to weigh-in on whether TextureTransform belongs in core. Future PR should probably re-organized the junk-drawer that is **src/extras**.

src/mesh/Rope.js
```diff
@@ -52,13 +51,11 @@ export default class Rope extends Mesh
...
-         * Tracker for if the rope is ready to be drawn. Needed because Mesh ctor can
-         * call _onTextureUpdated which could call refresh too early.
-         *
+         * refreshes vertices on every updateTransform
          * @member {boolean}
-         * @private
+         * @default true
          */
-        this._ready = true;
+        this.autoUpdate = true;
```
bigtimebuddy: Could you explain what `autoUpdate` is necessary as a public API? When would you want to set this to `false`?

ivanpopelyshev: When rope is static. Like a railroad, as an example.

ivanpopelyshev:
@bigtimebuddy @GoodBoyDigital I moved TextureTransform back, and I tested it with my examples again.

finscn:
@ivanpopelyshev  , how to do something like this http://fattyboy.cn/anim/  with new mesh ?

SebastienFPRousseau:
I'm working with the v4.5.1 installed by npm and i can't find TextureTransform.
PIXI.TextureTransform is not a constructor,
PIXI.extras.TextureTransform is not a constructor. 
I can't see it in the namespace PIXI too. 
May you help me ?

ivanpopelyshev:
That's a bug, It should be in extras but i think we removed it by mistake

SebastienFPRousseau:
Ok, thanks you. 
So i'm waiting for the fix.

Issue #3829, opened 2017-03-13
AnimatedSprite single BaseTexture, prepare plugin issue


brenwell:
Hey there,

I have an issue I have been trying to solve for a day now, and I think its time to call in the big guns.

I am loading 3 medium-sized spritesheets for 1 animation, created by TexturePacker, I collect all the frames and create an AnimatedSprite and play it. However the first time playing the animation is very unsmooth, and almost jumps immediately to the end, from then on it plays really smooth. The lag is caused by the time taken to upload the textures to the GPU. However using the PIXI prepare plugin to upload them prior to playback, doesn't appear to function, at aleast as I would expect.

```javascript
function loadSpriteSheet(callback){
  let loader = new PIXI.loaders.Loader()
  loader.add('http://mysite.fake/sprite1.json')
  loader.add('http://mysite.fake/sprite2.json')
  loader.add('http://mysite.fake/sprite3.json')
  loader.once('complete', callback)
  loader.load()
}  

loadSpriteSheet(function(resource){
  // helper function to get all the frames from multiple textures
  let frameArray = getFrameFromResource(resource)
  let animSprite = new PIXI.extras.AnimatedSprite(frameArray)
  this.renderer.plugins.prepare.upload(){
    stage.addChild(animSprite)
    animSprite.play()  
  }
})
```
After some snooping in the code I found A solution, not THE solution though - renderer.bindTexture(texture). When the PIXI loader receives a sprite atlas url e.g. my_sprites.json it automatically downloads the image file and names it as mysprites.json_image in the loaders resources. So I grab that, make a texture and upload it to the GPU. Here is the updated code:

```javascript
let loader = new PIXI.loaders.Loader()
loader.add('http://mysite.fake/sprite1.json')
loader.add('http://mysite.fake/sprite2.json')
loader.add('http://mysite.fake/sprite3.json')
loader.once('complete', callback)
loader.load()


function uploadToGPU(resourceName){
  resourceName = resourceName + '_image'
  let texture = new PIXI.Texture.fromImage(resourceName)
  this.renderer.bindTexture(texture)
}

loadSpriteSheet(function(resource){
  uploadToGPU('http://mysite.fake/sprite1.json')
  uploadToGPU('http://mysite.fake/sprite2.json')
  uploadToGPU('http://mysite.fake/sprite3.json')

  // helper function to get all the frames from multiple textures
  let frameArray = getFrameFromResource(resource)
[...]

PR #3830, opened 2017-03-14
Prepare plugin support for objects with multiple BaseTextures

brenwell:
The entire discussion for this task and its development is found at #3829 

Changes:

- Adds support for AnimatedSprites with more than 1 source texture to the Prepare plugins
- Removes the duplicate `function findBaseTextures()` found in both WebGL & CanvasPrepare
- Moves `function findBaseTextures()` to BasePrepare.js
- Splits the `BasePrepare.register(addHook,uploadHook) into 2 separate methods
- Updates unit tests

I am not thoroughly please with the naming of the `findBaseTexure` and `findTexture` methods, hopefully there are some better suggestions

### DIFF
--- a/src/prepare/BasePrepare.js
+++ b/src/prepare/BasePrepare.js
@@ -225,7 +225,7 @@ export default class BasePrepare
      * Adds hooks for finding items.
      *
      * @param {Function} [addHook] - Function call that takes two parameters: `item:*, queue:Array`
-              function must return `true` if it was able to add item to the queue.
+     *          function must return `true` if it was able to add item to the queue.
      * @return {PIXI.CanvasPrepare} Instance of plugin for chaining.
      */
     registerFindHook(addHook)
@@ -242,7 +242,7 @@ export default class BasePrepare
      * Adds hooks for uploading items.
      *
      * @param {Function} [uploadHook] - Function call that takes two parameters: `prepare:CanvasPrepare, item:*` and
-     *        function must return `true` if it was able to handle upload of item.
+     *          function must return `true` if it was able to handle upload of item.
      * @return {PIXI.CanvasPrepare} Instance of plugin for chaining.
      */
     registerUploadHook(uploadHook)
--- a/src/prepare/BasePrepare.js
+++ b/src/prepare/BasePrepare.js
@@ -385,6 +385,51 @@ function findTexture(item, queue)
 
     return false;
 }
+/**
+ * Built-in hook to draw PIXI.Text to its texture.
+ *
+ * @private
+ * @param {PIXI.WebGLRenderer|PIXI.CanvasPrepare} helper - Not used by this upload handler
+ * @param {PIXI.DisplayObject} item - Item to check
+ * @return {boolean} If item was uploaded.
+ */
+function drawText(helper, item)
+{
+    if (item instanceof core.Text)
+    {
+        // updating text will return early if it is not dirty
+        item.updateText(true);
+
+        return true;
+    }
+
+    return false;
+}
+
+/**
+ * Built-in hook to calculate a text style for a PIXI.Text object.
+ *
+ * @private
+ * @param {PIXI.WebGLRenderer|PIXI.CanvasPrepare} helper - Not used by this upload handler
+ * @param {PIXI.DisplayObject} item - Item to check
+ * @return {boolean} If item was uploaded.
+ */
+function calculateTextStyle(helper, item)
+{
+    if (item instanceof core.TextStyle)
+    {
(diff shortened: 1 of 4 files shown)


src/prepare/BasePrepare.js
```diff
@@ -213,21 +221,31 @@ export default class BasePrepare
...
             this.addHooks.push(addHook);
         }
 
+        return this;
+    }
+
+    /**
+     * Adds hooks for uploading items.
+     *
+     * @param {Function} [uploadHook] - Function call that takes two parameters: `prepare:CanvasPrepare, item:*` and
```
andrewstart: The `uploadHook` parameter should not be marked as optional.

src/prepare/BasePrepare.js
```diff
@@ -213,21 +221,31 @@ export default class BasePrepare
     }
 
     /**
-     * Adds hooks for finding and uploading items.
+     * Adds hooks for finding items.
      *
      * @param {Function} [addHook] - Function call that takes two parameters: `item:*, queue:Array`
```
andrewstart: The `addHook` parameter should not be marked as optional (the `[]`)

brenwell:
Whats the etiquette when making changes to pull requests. update the branch or make a new one?

ivanpopelyshev:
Just "rebase" the branch, it wont be big change.

src/deprecation.js
```diff
@@ -918,6 +918,24 @@ Object.defineProperty(core.utils, '_saidHello', {
...
+ *        function must return `true` if it was able to handle upload of item.
+ * @return {PIXI.CanvasPrepare} Instance of plugin for chaining.
+ * @deprecated since 4.4.1
+ */
+prepare.BasePrepare.prototype.register = function register(addHook, uploadHook)
+{
+    warn('renderer.plugins.prepare.register is now deprecated, '
+        + 'please use renderer.plugins.prepare.registerFindHook & renderer.plugins.prepare.registerUploadHook');
+
+    return this.registerFindHook(addHook).registerUploadHook(uploadHook);
```
bigtimebuddy: You should probably check for the hook, as both are optional:

```js
if (addHook) {
  this.registerFindHook(addHook);
}
if (uploadHook) {
  this.registerUploadHook(uploadHook);
}
return true;
```

src/deprecation.js
```diff
@@ -918,6 +918,24 @@ Object.defineProperty(core.utils, '_saidHello', {
 });
 
 /**
+ * Adds hooks for finding and uploading items.
+ *
+ * @param {Function} [addHook] - Function call that takes two parameters: `item:*, queue:Array`
```
bigtimebuddy: I like to remove parameters from the deprecation, so that people don't use it. Also this needs a proper JSDoc:
```js
/**
 * @method
 * @name PIXI.prepare.BasePrepare#register
 * @see PIXI.prepare.BasePrepare#registerFindHook
 * @deprecated since version 4.4.2
 */
```

You can test this by running `npm run docs` and opening **docs/index.html**

src/deprecation.js
brenwell: Sure no problemo, both methods check internally but better safe than sorry. I should probably return `this` though to preserve chaining

src/deprecation.js
brenwell: Doesn't seem to like that

```
  920:1  warning  Missing JSDoc for parameter 'addHook'     valid-jsdoc
  920:1  warning  Missing JSDoc for parameter 'uploadHook'  valid-jsdoc
```

I checked and other methods such as `PIXI.Texture.setFrame(frame)`  have the params .

Its quite possible that I implemented the deprecation method in a less than ideal way. Took me a few mins to work out whats going on in there

src/prepare/BasePrepare.js
```diff
@@ -290,6 +320,84 @@ export default class BasePrepare
...
+ * @return {boolean} if a PIXI.Texture object was found.
+ */
+function findMultipleBaseTextures(item, queue)
+{
+    // Objects with mutliple textures
+    if (item && item._textures && item._textures.length)
+    {
+        for (let i = 0; i < item._textures.length; i++)
+        {
+            const baseTexture = item._textures[i].baseTexture;
```
bigtimebuddy: It would probably be safer to check that `item._textures[i]` is an instanceof Texture, much like the `findTexture` method below. Also, should only `return true` if something is found.

brenwell:
Awesome!

No problem, I enjoyed it. I'd be happy to help in the future if you ever need some extra hands.

PR #3834, opened 2017-03-16
Improvements to how and what is stored in the texture caches.

themoonrat:
Built upon the good work from https://github.com/pixijs/pixi.js/pull/3789
+ For Text, the base textures were in the BaseTextureCache, but textures weren't in the TextureCache. This has been fixed
+ Now able to tell the BaseTexture the origin of where the canvas being passed through has come from. ie. was it from the text class? was it from the graphic class? was it from caching as a bitmap? More information now to track where these canvases (canvii??) have been generated from
+ BaseTexture has addBaseTextureToCache and removeBaseTextureFromCache functions, to match regular Texture. Any adding or removing to the BaseTextureCache is now through these functions
+ Protection around adding and remove items to the caches using the functions

! Breaking change. When you call Texture.fromTextureFromCache - it no longer automatically removes it's base texture from the base texture cache. Not sure how people feel about this

? I don't like lines 233 to 236 in Texture.js, and line 591 in BaseTexture.js - I don't see why they are there or why they are needed, but I chickened out on removing them!

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -798,7 +798,7 @@ export default class BaseTexture extends EventEmitter
      */
     static removeFromCache(id)
     {
-        const baseTexture = TextureCache[id];
+        const baseTexture = BaseTextureCache[id];
 
         if (baseTexture)
         {
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -698,12 +698,12 @@ export default class BaseTexture extends EventEmitter
      * @static
      * @param {HTMLCanvasElement} canvas - The canvas element source of the texture
      * @param {number} scaleMode - See {@link PIXI.SCALE_MODES} for possible values
-     * @param {string} [origin] - A string origin of who created the base texture
+     * @param {string} [origin='canvas'] - A string origin of who created the base texture
      * @return {PIXI.BaseTexture} The new base texture.
      */
-    static fromCanvas(canvas, scaleMode, origin)
+    static fromCanvas(canvas, scaleMode, origin = 'canvas')
     {
-        const name = origin ? `${origin}_` : 'canvas_';
+        const name = `${origin}_`;
 
         if (!canvas._pixiId)
         {
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -697,11 +697,9 @@ export default class BaseTexture extends EventEmitter
      */
     static fromCanvas(canvas, scaleMode, origin = 'canvas')
     {
-        const name = `${origin}_`;
-
         if (!canvas._pixiId)
         {
-            canvas._pixiId = `${name}${uid()}`;
+            canvas._pixiId = `${origin}_${uid()}`;
         }
 
         let baseTexture = BaseTextureCache[canvas._pixiId];
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -212,7 +212,7 @@ export default class BaseTexture extends EventEmitter
          *
          * @member {string[]}
          */
-        this.textureCacheId = [];
+        this.textureCacheIds = [];
 
         // if no source passed don't try to load
         if (source)
@@ -612,7 +612,7 @@ export default class BaseTexture extends EventEmitter
         this.dispose();
 
         BaseTexture.removeFromCache(this);
-        this.textureCacheId = null;
+        this.textureCacheIds = null;
 
         this._destroyed = true;
     }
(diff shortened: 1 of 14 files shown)


bigtimebuddy:
I'm in favor of consistent & shorter API names:
* `*.removeFromCache` instead of `BaseTexture.removeBaseTextureFromCache` and `Texture.removeTextureFromCache`
* `*.addToCache` instead of `BaseTexture.addBaseTextureToCache` and `Texture.addTextureToCache`

This will require deprecating `Texture.addTextureToCache` and `Texture.removeTextureFromeCache`

themoonrat:
I agree.. doing that would also get around the breaking change problem, as the deprecated removeTextureFromCache function can carry on also removing the base texture from its cache.

ivanpopelyshev:
What about switching off mipmaps for all "fromCanvas" methods?

andrewstart:
Should `*.addToCache` warn (in debug builds only, perhaps) when something in the cache is going to be overridden?

themoonrat:
Do we have a way of doing that yet? If not, if someone wanted to PR adding support for something like https://github.com/jsoverson/preprocess - then some warning code as you described would be the perfect fit.

bigtimebuddy:
You can already use preprocessor in the builds, however the lib version does not respect preprocess statements.

themoonrat:
Done. 
Also allowed removing from either cache by actual instance... so if a texture was added to the cache under multiple ids, this would remove them all.
Also added some unit tests for the new functionality (which also cover legacy behaviour)

src/core/textures/Texture.js
```diff
@@ -452,51 +454,85 @@ export default class Texture extends EventEmitter
...
+        }
+        else
+        {
+            for (const prop in TextureCache)
+            {
+                if (TextureCache[prop] === texture)
+                {
+                    delete TextureCache[prop];
+                }
+            }
```
englercj: Is this loop because a texture can be stored under multiple `textureCacheIds`? Could we just use that property to lookup the texture rather than looping through all textures?

If a texture can have multiple ids, just make `textureCacheId` an array and when it is added via `addToCache()` just add the new ID.

src/core/textures/BaseTexture.js
```diff
@@ -602,13 +611,15 @@ export default class BaseTexture extends EventEmitter
 
         this.dispose();
 
+        this.textureCacheId = null;
         for (const prop in BaseTextureCache)
```
englercj: This should use `removeFromCache()` rather than do it manually.

src/core/textures/BaseTexture.js
```diff
@@ -754,4 +768,72 @@ export default class BaseTexture extends EventEmitter
...
+            {
+                baseTextureFromCache.textureCacheId = null;
+                delete BaseTextureCache[baseTexture];
+
+                return baseTextureFromCache;
+            }
+        }
+        else
+        {
+            for (const prop in BaseTextureCache)
```
englercj: Same questions as the Texture cache, can we store all the ids on the object and just do lookups here rather than a loop.

src/core/textures/Texture.js
englercj: Then `removeFromCache()` when given an ID only removed that ID, and when given a texture removes all the IDs.

src/core/textures/BaseTexture.js
```diff
@@ -687,21 +698,24 @@ export default class BaseTexture extends EventEmitter
      * @static
      * @param {HTMLCanvasElement} canvas - The canvas element source of the texture
      * @param {number} scaleMode - See {@link PIXI.SCALE_MODES} for possible values
+     * @param {string} [origin] - A string origin of who created the base texture
      * @return {PIXI.BaseTexture} The new base texture.
      */
-    static fromCanvas(canvas, scaleMode)
+    static fromCanvas(canvas, scaleMode, origin)
```
bigtimebuddy: What about setting a default param here: `origin = "canvas"`, would remove the need for the ternary operator line below.

themoonrat:
Finally got around to this - textureCacheId is now textureCacheIds, and is an array.
Unit tests updated to reflect this change

src/core/textures/BaseTexture.js
```diff
@@ -687,21 +692,24 @@ export default class BaseTexture extends EventEmitter
...
      */
-    static fromCanvas(canvas, scaleMode)
+    static fromCanvas(canvas, scaleMode, origin = 'canvas')
     {
+        const name = `${origin}_`;
+
         if (!canvas._pixiId)
         {
-            canvas._pixiId = `canvas_${uid()}`;
+            canvas._pixiId = `${name}${uid()}`;
```
englercj: Looks like `name` is only used here, can we just remove this `name` and do `${origin}_${uid()}` here instead? That will reduce the string concats by one.

themoonrat: snuck that change in

englercj:
LG2M 🚢

PR #3842, opened 2017-03-18
Mesh update of average size, perfect for v4.5

ivanpopelyshev:
This WIP PR is dedicated to @finscn

This PR is also one step forward towards polygon packing.

Add those lines in the end of assetsLoaded of [Spritesheet example](http://pixijs.github.io/examples/?v=dev-mesh-update#/basics/spritesheet.js)

```js
    // here are the changes:
  
const mesh = new PIXI.mesh.Plane(anim.texture, 3, 3)
mesh.anchor.set(0.5);
mesh.position.set(anim.x - 200, anim.y);
app.stage.addChild(mesh);
app.ticker.add(function() {
    mesh.texture = anim.texture;
    mesh.vertices[9] = Math.cos(anim.rotation * 10)*30;
});
```

The case for mesh trim:
1. we have 1000 sprites
2. we want to apply filters on several of them , and we can spend performance on that. 
3. we also want to apply geometry effect - swap some sprites to meshes.
4. API of meshes is the same as sprite, same logic applies + some distortion on them.

In that case, developer doesn't have to repack stuff into separate atlas without trim just because he spontaneously wants to use some "effect".

### DIFF
--- a/src/core/math/GroupD8.js
+++ b/src/core/math/GroupD8.js
@@ -111,13 +111,14 @@ const GroupD8 = {
     rotate180: (rotation) => rotation ^ 4,
 
     /**
-     * I dont know why sometimes width and heights needs to be swapped. We'll fix it later.
+     * Direction of main vector can be horizontal, vertical or diagonal.
+     * Some objects work with vertical directions different.
      *
      * @memberof PIXI.GroupD8
      * @param {number} rotation - The number to check.
-     * @returns {boolean} Whether or not the width/height should be swapped.
+     * @returns {boolean} Whether or not the direction is vertical
      */
-    isSwapWidthHeight: (rotation) => (rotation & 3) === 2,
+    isVertical: (rotation) => (rotation & 3) === 2,
 
     /**
      * @memberof PIXI.GroupD8
--- a/src/mesh/Mesh.js
+++ b/src/mesh/Mesh.js
@@ -31,6 +31,11 @@ export default class Mesh extends core.Container
          */
         this._texture = texture;
 
+        if (!texture.baseTexture.hasLoaded)
+        {
+            texture.once('update', this._onTextureUpdate, this);
+        }
+
         /**
          * The Uvs of the Mesh
          *
@@ -74,7 +79,7 @@ export default class Mesh extends core.Container
         this.indexDirty = 0;
 
         /**
-         * The blend mode to be applied to the sprite. Set to `PIXI.BLEND_MODES.NORMAL` to remove
+         * The blend mode to be applied to the mesh. Set to `PIXI.BLEND_MODES.NORMAL` to remove
          * any blend mode.
          *
          * @member {number}
@@ -130,7 +135,7 @@ export default class Mesh extends core.Container
          * @member {PIXI.extras.TextureTransform}
          * @private
          */
-        this._uvTransform = new TextureTransform(texture);
+        this._uvTransform = new TextureTransform(texture, 0);
 
         /**
          * whether or not upload uvTransform to shader
@@ -151,6 +156,17 @@ export default class Mesh extends core.Container
     }
 
     /**
+     * Updates the object transform for rendering
+     *
+     * @private
+     */
+    updateTransform()
+    {
+        this.refresh();
+        this.containerUpdateTransform();
+    }
+
+    /**
      * Renders the object using the WebGL renderer
      *
      * @private
@@ -158,7 +174,6 @@ export default class Mesh extends core.Container
      */
     _renderWebGL(renderer)
     {
-        this.refresh();
         renderer.setObjectRenderer(renderer.plugins[this.pluginName]);
         renderer.plugins[this.pluginName].render(this);
     }
@@ -171,7 +186,6 @@ export default class Mesh extends core.Container
      */
     _renderCanvas(renderer)
     {
-        this.refresh();
         renderer.plugins[this.pluginName].render(this);
     }
 
@@ -209,7 +223,7 @@ export default class Mesh extends core.Container
     {
         if (this._uvTransform.update(forceUpdate))
         {
-            this._refresh();
+            this._refreshUvs();
         }
     }
 
@@ -217,7 +231,7 @@ export default class Mesh extends core.Container
      * re-calculates mesh coords
      * @protected
      */
-    _refresh()
+    _refreshUvs()
     {
         /* empty */
     }
--- a/src/mesh/NineSlicePlane.js
+++ b/src/mesh/NineSlicePlane.js
@@ -46,40 +46,6 @@ export default class NineSlicePlane extends Plane
     {
         super(texture, 4, 4);
 
-        const uvs = this.uvs;
-
-        // right and bottom uv's are always 1
-        uvs[6] = uvs[14] = uvs[22] = uvs[30] = 1;
-        uvs[25] = uvs[27] = uvs[29] = uvs[31] = 1;
-
-        this._origWidth = texture.orig.width;
-        this._origHeight = texture.orig.height;
-        this._uvw = 1 / this._origWidth;
-        this._uvh = 1 / this._origHeight;
-
-        /**
-         * The width of the NineSlicePlane, setting this will actually modify the vertices and UV's of this plane
-         *
-         * @member {number}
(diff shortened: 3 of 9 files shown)


finscn:
Thanks. I found some bugs. 
But I know it's WIP, so  Waiting for your finished product.

ivanpopelyshev:
Added trim. Gimme the bugs!

finscn:
Great work!    the bugs went away. 

 I'll do some deep tests later.

I think if I'm a girl , I wanna marry you.

finscn:
If the Rope could support  define real vertical points , that's will be better . LOL

finscn:
the `sprite` in https://github.com/pixijs/pixi.js/pull/3842/files#diff-b668501f3bb23914a2f68146e903ddd4R77  should be `mesh`.

finscn:
Some Requests:
* Rope could support define real vertical points
* Rope could support anchor
* If Plane uses  `vertices[]` , Rope should uses too. For unified style/interface.

finscn:
And a question :  
In pixi.spine , there is also mesh-feature , that feature in spine could use this new mesh ?

ivanpopelyshev:
spine has to be compatible even with 3.0.11, so no, it can't :( As for trimming, yes, pixi supports 
trimming for every mesh now, automatically.

ivanpopelyshev:
1. points are not vertices - vertices are the buffer, calculated based on points. They dont exist in Plane , of course.
2. RopeSprite has fourth parameter, which is 0 - horizontal, 2 - vertical.
3. RopeSprite has shift[] which shifts rope normals. Its easier to use than points, but you can modify points too.

```
const mesh = new PIXI.mesh.RopeSprite(anim.texture, 3, 2, 0)
mesh.refresh(true);
mesh.anchor.set(0.5);
mesh.position.set(anim.x - 200, anim.y);
app.stage.addChild(mesh);
app.ticker.add(function() {
    mesh.texture = anim.texture;
   //try this one:
    mesh.points[1].y = Math.sin(anim.rotation*5) * 40
   //or this one
    mesh.shift[1] = Math.sin(anim.rotation*5)* 40
});
```

finscn:
Amazing! 

There is only one problem :  
`RopeSprite` is neither `subclass of Rope`  nor  `subclass of Sprite`. 
But called  `RopeSprite` , it's unreasonable.

finscn:
```
const mesh = new PIXI.mesh.RopeSprite(anim.texture, 3, 2, 0)
mesh.refresh(true);
```
Why don't call `refresh` in constructor automatically?

ivanpopelyshev:
Oh right, no need for refresh() here.

ivanpopelyshev:
I'll add special thing, "RopeNode", like a point but has offset, scale and may be even tint/alpha!

finscn:
In Rope , why use `mesh.shift` , not `mesh. vertices` ?

ivanpopelyshev:
Vertices are generated from points and shifts. Shift is just a adjustment for normals.

finscn:
Is there  shift in Plane ?
Is there  vertices in Rope ?

ivanpopelyshev:
There are no shifts in Plane, because there are no normals. There are vertices in rope, same as in plane, but they are calculated based on rope points.

finscn:
I see . Thanks.   That's mean I also could use  `vertices` in RopeSprite  , Am I right ?

ivanpopelyshev:
renamed shift->pointShift, added pointScale. Please use larger "verticesY" for scale, otherwise it'll be distorted much.

No, you can only read vertices in Rope , but that doesn't mean we have to hide old vertices and rename, points to vertices, they even have different types (Float32Array vs Point[])

I dont have time anymore , I spent totally 12 hours on that issue in last two days. I dont even have time to make examples, so you'll have to make them, otherwise this update wont be merged.

finscn:
OK,  I will do.

finscn:
A Suggestion about name : 

the new `Plane`  named  `PlaneMesh` 
the new `RopeSprite`  named :  `RopeMesh`  .

What do you think ?

ivanpopelyshev:
better demo for snake:

```js
var app = new PIXI.Application();

var count = 0;

var strip = new PIXI.mesh.RopeSprite(PIXI.Texture.fromImage('required/assets/snake.png'), 25, 2, 0);
var points = strip.points;


strip.anchor.set(0, 0.5);
strip.width = 600;
strip.height = 70;
strip.position.set(100, 300)

app.stage.addChild(strip);

var g = new PIXI.Graphics();
g.x = strip.x;
g.y = strip.y;
app.stage.addChild(g);

// start animating
app.ticker.add(function() {

    count += 0.1;
  
  for (var i = 0; i < points.length; i++) {
      strip.pointShift[i] = Math.sin((i * 0.5) + count) * 10;
      //strip.pointScale[i] = Math.sin((i * 0.5) + count+ 1)*0.2 + 1;
  }

    // make the snake
    for (var i = 0; i < points.length; i++) {
        /*points[i].y = Math.sin((i * 0.5) + count) * 30;
        points[i].x = i * strip.width/(points.length-1) + Math.cos((i * 0.3) + count) * 20;*/
    }
    renderPoints();
});

function renderPoints () {

    g.clear();

    g.lineStyle(2,0xffc2c2);
    g.moveTo(points[0].x,points[0].y);

    for (var i = 1; i < points.length; i++) {
        g.lineTo(points[i].x,points[i].y);
    }

    for (var i = 1; i < points.length; i++) {
        g.beginFill(0xff0022);
        g.drawCircle(points[i].x,points[i].y,5);
        g.endFill();
    }
}
```

finscn:
Is the `pivot`  in your plan ?

finscn:
A issue about  `anchor`  : 
When I change the anchor , I have to call `refresh()` manually.
I think maybe we should do something in `_onAnchorUpdate`

finscn:
Could we need a `origVertices` for saving the original vertices ?
Some times , we need change the vertices around the `original vertices`.
e.g:
```
vertices[idx * 2] = origVertices[idx * 2] + Math.sin(timeLine / 100) * 10;
```
of course , I know I could save it by myself in code like this : 
```
var plane = new Plane(.....);
plane.origVertices = plane.vertices.slice();
```

The key point is if we should supply `origVertices` in framework ?

finscn:
A issue about NineSlicePlane : 
now NineSlicePlane can't work.  it only renders left-top corner.

finscn:
A doc issue about Plane : 

The example in annotation should be update 
```
/**
 * The Plane allows you to draw a texture across several points and them manipulate these points
 *
 *```js
 * for (let i = 0; i < 20; i++) {
 *     points.push(new PIXI.Point(i * 50, 0));
 * };
 * let Plane = new PIXI.Plane(PIXI.Texture.fromImage("snake.png"), points);
 *  ```
 *
 * @class
 * @extends PIXI.mesh.Mesh
 * @memberof PIXI.mesh
 *
 */
```

finscn:
I think you should change the title of this PR. Because it's not only about mesh.plane. It's a new Mesh system.

finscn:
Hi, I'm working on the examples of new Mesh. But I found a bug about  `direction`.
the direction is not always expected in Rope.
example:  don't support  horizontal rope ,  don't support MIRROR_HORIZONTAL ...

finscn:
A bug about Plane's vertices & calculatedVertices.

In Rope , when new an instance , the points and calculatedPoints would be created.
But in Plane,  vertices & calculatedVertices would only be created when texture is loaded.
I think the way of Rope is correct.

The code about Plane will take some mistakes: 
```
var texture = PIXI.Texture.fromImage('required/assets/displacement_BG.jpg');
var strip = new PIXI.mesh.Plane(texture, 10, 10);
var vertices = strip.vertices;
var calculatedVertices = strip.calculatedVertices;
//  calculatedVertices is undefined
```

finscn:
@ivanpopelyshev  , Is there a plan about this feature ?

ivanpopelyshev:
Oh shit, i forgot about that thing :) I still need to put your examples (that you gave me, thank you!) and make it separate lib for now.

bigtimebuddy:
Chatted with @GoodBoyDigital about this, and we are going to close. We will create a more complex Mesh class in v5 with your help but will say no to major refactors or features in v4.

PR #3845, opened 2017-03-19
ticker listeners with priority, uses mini-runner

drkibitz:
- Removes internal emitter implementation and manages listeners using mini-runner

Addresses https://github.com/pixijs/pixi.js/issues/3835

---

Please note this is a WIP. It uses a fork or mini-runner with priority insertion, which is in PR here https://github.com/GoodBoyDigital/mini-runner/pull/6

I'm not set on the implementation, but just trying to contribute something. The name `onUpdate` has been fairly common in my past experience, so I just went with it. Could have also just went with `tick`, but since the name `update` means something different depending on which class you look at, I decided against it. Also `update` is a verb, and `onUpdate` is an event, where the first would most likely trigger the second. I also completely removed the `addOnce` from ticker, as the only thing using it was `BasePrepare` and a test. I have not added deprecations for any of this, but it should be possible to do for `addOnce`, but for `add`, which changes the method signature it will be more difficult, but should be doable.

As for the setting `UPDATE_PRIORITY`, that is definitely a first pass. I figured having specific priorities for everything used by PIXI would be helpful. Also not set on the names, or the members themselves. It made sense that `default` was somewhere in between `interactive` and `utility`. The others are just guesses. The goal there was just to make sure consumers of PIXI would get the default, and when using `PIXI.Application`, that their handlers would be called before rendering.

---

Hope people have a chance to take a look, and as always, I welcome feedback.

Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "jaguarjs-jsdoc": "^1.0.1",
     "js-md5": "^0.4.1",
     "jsdoc": "^3.4.2",
+    "mini-runner": "https://github.com/drkibitz/mini-runner/archive/dev/priorities.tar.gz",
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
     "parallelshell": "^2.0.0",
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -97,12 +97,23 @@ export default class Application
         this._ticker = ticker;
         if (ticker)
         {
-            ticker.add(this.render, this);
+            ticker.add(this, settings.UPDATE_PRIORITY.low);
         }
     }
     get ticker() // eslint-disable-line require-jsdoc
     {
         return this._ticker;
+    }
+
+    /**
+     * Handle tick update
+     *
+     * @private
+     * @param {number} deltaTime - Time since last tick.
+     */
+    onUpdate()
+    {
+        this.renderer.render(this.stage);
     }
 
     /**
@@ -156,6 +167,7 @@ export default class Application
     destroy(removeView)
     {
         this.stop();
+        this.ticker.remove(this);
         this.ticker = null;
 
         this.stage.destroy();
--- a/src/core/settings.js
+++ b/src/core/settings.js
@@ -17,6 +17,29 @@ export default {
     TARGET_FPMS: 0.06,
 
     /**
+     * Listener priorities that are both the actual priorities of
+     * internal listeners added to {@link PIXI.ticker.Ticker}
+     * as well as a guideline to priorities of new listeners.
+     *
+     * @static
+     * @constant
+     * @memberof PIXI.settings
+     * @type {object}
+     * @property {number} interaction=50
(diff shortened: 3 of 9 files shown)


bigtimebuddy:
I like this implementation and it improves a lot on the issues with priority of the ticks and will help clear confusion, I think, for developers. 

My biggest concern is that this approach has no deprecation strategy, it is a hard breaking change. Going from A implementation to B implementation we need a period of time where both APIs can coexist (let's call that "A-prime"). I would look at how Chad did the conversion of Loader to mini-signals. Both event emitter and mini-signal APIs work, this is A-Prime. Eventually we could drop the EE APIs.

How could you create A-Prime for this ticker change? What would that look like. I good place to start would be making sure the old test still works.

englercj:
We can't use mini-runner in pixi, it will cause CSP errors that have cleaned out for other people. I also don't like using EE3, mini-signals, AND mini-runner in the same codebase.

drkibitz:
@bigtimebuddy no worries on the deprecation strategy, I did mention it above. It's totally doable, I just didn't include it in a WIP pull request because no use putting the effort into something that won't be accepted. That comes last ;)

@englercj Pardon me, I should have just made a new issue with the following discussion points, but they are strewn across various threads after I commented on things as I came across them. If you can try to follow my thought process through these links, and respond, that would be helpful:
- https://github.com/pixijs/pixi.js/issues/3835#issuecomment-287284349
- https://github.com/pixijs/pixi.js/pull/3570#issuecomment-287529747
- https://gist.github.com/drkibitz/79df1f458aa0e0db02eed338de1a25f8
- https://github.com/GoodBoyDigital/mini-runner/pull/6

With that, I hope you can see why I just went with mini-runner out of the gate here (I did not locally). About the 3 things in the same codebase, I've got to disagree with the general sentiment given there, it's really not about using something because it's there, it's using something that is right for the job. I'm with you with cringing a little with 3 things in the same codebase, but not that concerned, that's what happens in the evolution of a codebase, it'll eventually be cleaned up. As for the CSP errors, that's a "back to the drawing board" kind of a deal breaker. So I'm back to my original questions on the various links above. Would appreciate any sort of discussion on them in this thread now that there's a PR.

bigtimebuddy:
I created a pull request that got merged this week (currently not published) which removes the CSP restrictions for **mini-runner**. https://github.com/GoodBoyDigital/mini-runner/pull/7 I feel strongly that we shouldn't utilize this module in PIXI until it's more stable and production-ready.

There have been various discussions here about moving away from EventEmitter3 in favor of a signal-type API (e.g. **mini-signals**). I'm not opinionated about EE vs signals as long as it's done wholesale across the entire code-base, has proper deprecation support, and is part of a major version bump (e.g., v5).

ivanpopelyshev:
yeah, into v5 it will go.

drkibitz:
Closing in favor of #3859

PR #3855, opened 2017-03-21
Support that cache Sprite as Bitmap directly.

finscn:
In the current pixi.js , when we want to cache a sprite as Bitmap, we have to put the sprite into a container , like this : 
```js
var texture = PIXI.Texture.from(img);
sprite.cacheAsBitmap = true;
var renderTexture = PIXI.RenderTexture.create(sprite.width, sprite.height);

container.addChild(sprite);
renderer.render(container, renderTexture);
```

With this PR , user could cache the sprite directly : 
```js
var texture = PIXI.Texture.from(img);
sprite.cacheAsBitmap = true;
var renderTexture = PIXI.RenderTexture.create(sprite.width, sprite.height);

renderer.render(sprite, renderTexture);
```

### DIFF
--- a/src/extras/cacheAsBitmap.js
+++ b/src/extras/cacheAsBitmap.js
@@ -227,7 +227,17 @@ DisplayObject.prototype._initCachedDisplayObject = function _initCachedDisplayOb
 
     this.transform._parentID = -1;
     // restore the transform of the cached sprite to avoid the nasty flicker..
-    this.updateTransform();
+
+    if (!this.parent)
+    {
+        this.parent = renderer._tempDisplayObjectParent;
+        this.updateTransform();
+        this.parent = null;
+    }
+    else
+    {
+        this.updateTransform();
+    }
 
     // map the hit test..
     this.containsPoint = cachedSprite.containsPoint.bind(cachedSprite);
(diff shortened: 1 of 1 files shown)


macguffin:
Just wondering is there a performance benefit from caching a single bitmap? What is the use case for this?

I thought cacheAsBitmap was most useful if you had lots of elements in a container you wanted to be treated as a single image. 

Thanks in advance for explaining.

ivanpopelyshev:
@macguffin 
I think that cacheAsBitmap is evil. But @finscn working with single elements a lot, he has coding style that somehow depends on it.

j-funk:
@ivanpopelyshev I saw your comment "I think that cacheAsBitmap is evil".  Are you able to elaborate on that or point to somewhere that explains why?   Thanks!

Reloecc:
cacheAsBitmap is such an awesome feature I have seen in pixi.js only - you can achieve this in other frameworks, but no by single function call. I feel really uncomfortable rendering even 50 sprites even in opengl, although I know opengl (webgl) is capable of rendering thousands sprites, I just feel pain back in the head doing that, because I am used to redraw regions and sprite merging from the "past age" of flash.

Issue #3851, opened 2017-03-20
Interaction Issue - Button clicks go through to buttons below.


bob-builder-ngg:
Hi,

We have recently updated our code to use pixi 4.4.0 from 4.3.5 and have come across an issue with mouse/touch events firing on multiple interactive elements when they overlap. This issue is still present in V4.4.2.

We were able to reproduce this issue with this simple PIXI example: https://jsfiddle.net/apodgg2c/2/

Red box is drawn below the blue box. 

Click the red box and you get a console output "RED FIRED"

Click the blue box and you get a console output "BLUE FIRED"

Click the overlapping region and both RED and BLUE will fire. 

In the previous version (4.3.5) only the BLUE will fire when you click the overlapping region. (as would be expected)

A little more information:
- If we add two interactive elements (buttons) as siblings to a container then they work fine and they will not click through when they overlap.
- If you create two containers as children of a parent container and place a button inside each of the containers then it fails. 

Here are some examples of scene graph hierarchies and which ones work (no click through) and which ones fail (clicks through)

```
OK

	Stage
	/  \
     Btn1   Btn2


Fail
	Stage
	/   \
 container container
      /       \  
    Btn1      Btn2


Fail 
(Btn1 added first)
	Stage
	/   \
      Btn1 container
              \  
              Btn2


OK 
(Btn1 added last)
	Stage
	/   \
 container  Btn1
      /         
    Btn2      

```

Any suggestions for a work around or is this something that will like be fixed in future versions?

This issue may be related to the DOM elements also clicking through?
https://github.com/pixijs/pixi.js/issues/3847
https://github.com/pixijs/pixi.js/issues/3790

andrewstart:
Do you have any problems with https://d157l7jdn8e5sf.cloudfront.net/fix-interaction-issues/pixi.js?

bob-builder-ngg:
Hi, thanks for the quick reply.  Yes this version of PIXI seems to resolve the issue. :) Will this be rolled into a future release?

andrewstart:
Yeah, we'll make a patch as soon as I work out some other issues people have reported.

bigtimebuddy:
I'll make this a high priority release if @andrewstart you can make a PR to master.

Tigerwulf:
Hey, 
that bug is not fixed yet. In some specific cases the will still bubble the event over all displayObjects.
For that a short fiddle example. https://jsfiddle.net/d41h5g4d/8/
[...]

PR #3857, opened 2017-03-22
Fix some interaction issues

### LINKED ISSUE #3844: Extreme InteractionManager performance issues with touch in 4.4.x
I originally thought this was an issue with my game or engine, but I think I've isolated it to Pixi's InteractionManager class in versions 4.4.0 and higher.

## The Problem

I'm not sure of the cause, but there is a direct correlation between how many times you have touched the screen, and the severity of a spike in frametime that happens about 3 times a second (on my devices).

### More Info

The spikes do not go away even after you are no longer touching the screen. Although moving the mouse causes [this check](https://github.com/pixijs/pixi.js/blob/v4.4.2/src/interaction/InteractionManager.js#L617) to pass, which results in the problem stopping until you stop moving the mouse. So it would seem the problem is somewhere in [this block](https://github.com/pixijs/pixi.js/blob/v4.4.2/src/interaction/InteractionManager.js#L626-L652). I guess in `configureInteractionEventForDOMEvent()` and/or `processInteractive()`? Both of which seem to be implicated respectively in profiles of the jsfiddle (see below) and my game.

The issue happens every time, though it takes longer to have an effect on something faster like a desktop (but they aren't really affected by a touch issue anyway)

On my Windows 10 tablet in Chrome and in Edge, after a few hundred touches my otherwise performant game becomes unplayable. Same in Chrome on my Android phone.

I guess there's some kind of leak involving touch/pointer events filling up an iterator somewhere.

## Reproducing

...is kind of a pain in the ass. Here is the jsfiddle I wrote: https://jsfiddle.net/1g3kvd5r/1/

Unfortunately you'll have to manually touch the screen a few hundred times to see any effect. If you really hate yourself you can do it on your desktop using chrome's device preview. I tried using dispatchEvent() to simulate the PointerEvents, but couldn't get it to cause the bug.

I suggest throttling your CPU as low as possible and using the profiler to compare before and after doing the touches. You may wish to comment out the text drawing bits so they don't clutter up your profiles. Chrome's timeline profiler is more useful here, since the problem is intermittent. (Speaking of that, I'm not sure what it's tied to. It's very regular but seems to be less frequent than `InteractionManager.interactionFrequency`.)

It's worth making explicit that nothing except for starting pixi and touching the screen is actually needed to reproduce this.

Here are some pictures of profiles I did while trying to figure this out. For each of them I followed these steps:

1. Touched the screen a bunch (and cried a little)
2. Started the profiler
3. Did nothing for a few seconds
4. Stopped the profiler

[jsfiddle before touches](http://i.imgur.com/VpKRMIZ.png)
[jsfiddle after touches](http://i.imgur.com/InSLT68.png)

[real-world before touches](http://i.imgur.com/471Q6VO.png)
[real-world after touches](http://i.imgur.com/C3wag9K.png)

All tests done with 4.4.2

## Thoughts

After several hundred touches, framerate drops to unusable levels multiple times each second in an empty scene. This is a fairly serious problem.

I think that @starakaj wrote most of the changes to interaction in 4.4.x. Maybe he knows something that will help figure this out.

andrewstart:
* Fixes #3844 - memory leak on touch end.
* Fixes another component of #3790, #3851 - tracking that an interactive child was hit was forgotten more than one level up.
* Adds unit tests for both of these fixes.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -1268,6 +1268,12 @@ export default class InteractionManager extends EventEmitter
         {
             this.emit('mouseout', interactionEvent);
         }
+        else
+        {
+            // we can get touchleave events after touchend, so we want to make sure we don't
+            // introduce memory leaks
+            this.releaseInteractionDataForPointerId(interactionData.identifier);
+        }
     }
 
     /**
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -13,6 +13,12 @@ core.utils.mixins.delayMixin(
 
 const MOUSE_POINTER_ID = 'MOUSE';
 
+// private constants for use in processInteractive - tracks whether we hit anything at all, or an
+// actual interactive child, so that we can keep that state going back up the display tree
+const HIT_NONE = 0;
+const HIT_ANY = 1;
+const HIT_INTERACTIVE = 2;
+
 /**
  * The interaction manager deals with mouse, touch and pointer events. Any DisplayObject can be interactive
  * if its interactive parameter is set to true
@@ -761,7 +767,7 @@ export default class InteractionManager extends EventEmitter
      *  interactionEvent, displayObject and hit will be passed to the function
      * @param {boolean} [hitTest] - this indicates if the objects inside should be hit test against the point
      * @param {boolean} [interactive] - Whether the displayObject is interactive
-     * @return {boolean} returns true if the displayObject hit the point
+     * @return {number} returns 1 or 2 if the displayObject hit the point, 0 if not
      */
     processInteractive(interactionEvent, displayObject, func, hitTest, interactive)
     {
@@ -787,7 +793,7 @@ export default class InteractionManager extends EventEmitter
 
         interactive = displayObject.interactive || interactive;
 
-        let hit = false;
+        let hit = HIT_NONE;
         let interactiveParent = interactive;
 
         // if the displayobject has a hitArea, then it does not need to hitTest children.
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Great work @andrewstart! Thanks for tackling these interaction bugs.

src/interaction/InteractionManager.js
```diff
@@ -761,7 +767,7 @@ export default class InteractionManager extends EventEmitter
      *  interactionEvent, displayObject and hit will be passed to the function
      * @param {boolean} [hitTest] - this indicates if the objects inside should be hit test against the point
      * @param {boolean} [interactive] - Whether the displayObject is interactive
-     * @return {boolean} returns true if the displayObject hit the point
+     * @return {number} returns 1 or 2 if the displayObject hit the point, 0 if not
```
bigtimebuddy: How important do you think it is to distinguish between 1 and 2 in the docs here?

ivanpopelyshev:
I think its wrong to pass this kind of thing through processInteractive. Can we make an array of all things that were hit and then calculate which one was actually hit?

src/interaction/InteractionManager.js
andrewstart: Since it's a private method, I don't think it's too important to do so. For those looking at the code, I feel like the name of the constants should be sufficient.

ivanpopelyshev:
I understand how algorithm works but I'm afraid that it became to complicated to be changed in future. We'll have to rewrite it at some point. I approved to move things forward.

GoodBoyDigital:
I think this one is very important and we should do a release asap 👍

PR #3858, opened 2017-03-22
Problem by padding and trim in Text

ayamada:
See http://embed.plnkr.co/hgf5Tb/

Display to Text instances and those rectangles.
`t1`, `t2`, `t3` is OK, but `t4` is not OK.

I fixed it.

(Trimmed texture has been reduced padding, don't reduce padding more)

Sorry for my poor english.


bigtimebuddy:
Thanks @ayamada!

PR #3870, opened 2017-03-24
Fix Graphics memory leak (#3869)

### LINKED ISSUE #3869: Memory leak in Graphics.clear
I investigated this and figured out a fix already; I just wanted to file an issue first before making a pull request.

Repro: this is caused by repeatedly clearing a Graphics object and then calling `drawRect` from WebGL. Rendering the Graphics object will call `_renderSpriteRect`, which makes a new Texture that never gets destroyed, so the TextureUvs object that gets created is never dereferenced.

JSFiddle repro link: http://jsfiddle.net/yabm091r/3/

Solution: add this code to Graphics.prototype.clear:
```js
if (this._spriteRect) {
    this._spriteRect.destroy({texture: true});
    this._spriteRect = null;
}
```

P.S. here's a clip of me accidentally filing this bug with the title "tv" (my ctrl key wasn't working): https://clips.twitch.tv/ArbitraryConcernedHamCharlietheUnicorn

Adam13531:
This fixes #3869. Tests are passing. Comments about the fix are in #3869 directly.

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -711,7 +711,7 @@ export default class Graphics extends Container
 
         if (this._spriteRect)
         {
-            this._spriteRect.destroy({texture: true});
+            this._spriteRect.destroy({ texture: true });
             this._spriteRect = null;
         }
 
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Closed in favor of #3872

PR #3871, opened 2017-03-24
Fix for #3869

ivanpopelyshev:
We didn't truly remove all the handlers from our texture constants.


ivanpopelyshev:
Moving to master branch

Issue #3873, opened 2017-03-24
Filter issue with sourceFrame.fit


Adam13531:
I talked to @ivanpopelyshev offline about this and he said to file an issue.

Here's the JSFiddle that I've made: http://jsfiddle.net/xbmhh207/ (edit: updated to Ivan's JSFiddle that loads Pixi.js as an external resource)

It's a combination of code from other resources that I've found. When the bunny goes off-screen, you can see the yellow circle get "squished" into the visible portion of the sprite even though the shader attempts to position this at the center of the sprite. This is due to [this line of code in FilterManager](https://github.com/pixijs/pixi.js/blob/58ccf3381566c009c418aa305924539f372281f3/src/core/renderers/webgl/managers/FilterManager.js#L111). If I comment that out, things "work" perfectly (as in it renders how I'd expect, but I don't know the full repercussions of removing that line). Ivan mentioned that Pixi may need to provide two different modes so that you can choose whether you want sourceFrame.fit to be called based on the filter.

ivanpopelyshev:
Looking @ it

ivanpopelyshev:
That is build together with #3872: https://www.dropbox.com/s/3a98x273twrnans/pixi.js?dl=0

Please test it.

```
filter.dontFit = true;
```

ivanpopelyshev:
@Adam13531 please, next time add pixi to external resources and not to HTML. Your fiddle appears only after a half of minute.

ivanpopelyshev:
Also, http://www.html5gamedevs.com/topic/29327-guide-to-pixi-v4-filters/

Adam13531:
> @Adam13531 please, next time add pixi to external resources and not to HTML. Your fiddle appears only after a half of minute.

@ivanpopelyshev Ah, I didn't realize; I'd modified a JSFiddle that I found through the forums where they made the same mistake. I updated my original post to use your JSFiddle.

> Also, http://www.html5gamedevs.com/topic/29327-guide-to-pixi-v4-filters/
👍 I'll have to take a closer look at this during the week if I still see any problems with my filters.

englercj:
Hey @Adam13531 have you had a chance to take a look at this?

@ivanpopelyshev Do you have a PR or anything related to this issue?

Adam13531:
@englercj sorry for the incredibly delayed response. Using the new `filter.autoFit = false;` works for me!

PR #3874, opened 2017-03-24
Dont fit filters for some cases

ivanpopelyshev:
That one isn't critical. I don't know how to fix "fit" performance feature yet, but we can add flag for it.

#3873

### DIFF
--- a/src/core/renderers/webgl/filters/Filter.js
+++ b/src/core/renderers/webgl/filters/Filter.js
@@ -87,6 +87,13 @@ export default class Filter
          * @member {boolean}
          */
         this.enabled = true;
+
+        /**
+         * If enabled, pixi wont fit the filter area into boundaries, it helps for shaders with mapping
+         *
+         * @member {boolean}
+         */
+        this.dontFit = false;
     }
 
     /**
--- a/src/core/renderers/webgl/filters/Filter.js
+++ b/src/core/renderers/webgl/filters/Filter.js
@@ -85,15 +85,17 @@ export default class Filter
          * If enabled is true the filter is applied, if false it will not.
          *
          * @member {boolean}
+         * @default true
          */
         this.enabled = true;
 
         /**
-         * If enabled, pixi wont fit the filter area into boundaries, it helps for shaders with mapping
+         * If enabled, pixi will fit the filter area into boundaries, it helps for shaders with mapping
          *
          * @member {boolean}
+         * @default true
          */
-        this.dontFit = false;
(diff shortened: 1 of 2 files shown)


andrewstart:
Tested this in http://jsfiddle.net/xbmhh207/1/ and it seems to work, although I am not familiar enough with filters to know why we are fitting them in the first place. API-wise, I think having `fitToDestination` that defaults to true would be preferred, rather than have a negating setting.

ivanpopelyshev:
OK, that's good idea. I'm investigating on how to make real fix, but for now it fits.

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -106,7 +106,7 @@ export default class FilterManager extends WebGLManager
 
             // TODO we should fit the rect around the transform..
         }
-        else
+        else if (filters[0].fitToDestination)
```
andrewstart: It wouldn't be ideal for performance, but should we check all the filters that are getting added, in case there are multiple and one does not want to be fitted?

ivanpopelyshev: Look at 10 lines above - same strategy applies for other parameters, like resolution. I tried to make working solution without that hack, but so far i failed.

andrewstart: Got it.

src/core/renderers/webgl/filters/Filter.js
```diff
@@ -85,8 +85,17 @@ export default class Filter
          * If enabled is true the filter is applied, if false it will not.
          *
          * @member {boolean}
+         * @default true
          */
         this.enabled = true;
+
+        /**
+         * If enabled, pixi will fit the filter area into boundaries, it helps for shaders with mapping
```
bigtimebuddy: Could you add more here about _why_ you would disable this property?

src/core/renderers/webgl/filters/Filter.js
```diff
@@ -85,8 +85,17 @@ export default class Filter
...
+         *
+         * @member {boolean}
+         * @default true
+         */
+        this.fitToDestination = true;
```
bigtimebuddy: Not crazy about the name of the property. It's a little confusing and verbose. Isn't this more accurate as `fitToSource`? Or maybe, simply: `autoFit`?

ivanpopelyshev:
Closed in favor of new thing.

PR #3877, opened 2017-03-25
Further fix interaction issues related to nesting and display tree positioning

andrewstart:
Fixed further problem as reported in #3851. Also undid the constants added in #3857 because there was actually a simpler way to handle that.

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -13,12 +13,6 @@ core.utils.mixins.delayMixin(
 
 const MOUSE_POINTER_ID = 'MOUSE';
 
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
The saga continues... This seems good to me, but I want to make sure that we have actually tested and verified that it addresses these cluster of issues once and for all (a la jsfiddle). Also, are there any other related issues before we'd make a release for this?

ivanpopelyshev:
I said that constants aren't needed there! I'm a prophet!

andrewstart:
I did confirm the fix in the reporting jsfiddle. I don't know of any other related issues - the one about clicks happening on the DOM below is a different problem entirely.

PR #3878, opened 2017-03-25
Added public hit test method to InteractionManager

andrewstart:
Adds a stable, public method to manually hit test against the display tree, and unit tests to support it.


bigtimebuddy:
@themoonrat is anything you want to comment on before I merge?

themoonrat:
Nopes, looks good to me and something I can already think of uses for :)

PR #3881, opened 2017-03-25
fix  brace-style

finscn:
avoid `warning  Opening curly brace appears on the same line as controlling statement  brace-style`


ivanpopelyshev:
I've already pushed it.

PR #3890, opened 2017-03-29
Expose normalized event properties.

### LINKED ISSUE #3886: Bug with Touch in Chrome (android)
Hello

I have a bug in Chrome - Android:
I can't get touches in touchmove event:
event.data.originalEvent
touches - is undefined. 
changedTouches - is undefined.

But this work in Safari - iOS.
I want get data from touches.

andrewstart:
Fixes #3886 by exposing normalized pointer event data on InteractionData:
**Properties from PointerEvent:**
`isPrimary`, `button`, `buttons`, `width`, `height`, `tiltX`, `tiltY`, `pointerType`, `pressure`, `twist`, `tangentialPressure`.
**Properties from Touch:**
`rotationAngle`
**Properties from Touch normalized to PointerEvent standards:**
* `radiusX` -> `width`
* `radiusY` -> `height`
* `force` -> `pressure`

### DIFF
--- a/src/interaction/InteractionData.js
+++ b/src/interaction/InteractionData.js
@@ -40,6 +40,93 @@ export default class InteractionData
          * @member {number}
          */
         this.identifier = null;
+
+        /**
+         * Indicates whether or not the pointer device that created the event is the primary pointer.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/isPrimary
+         * @type {Boolean}
+         */
+        this.isPrimary = false;
+        /**
+         * Indicates which button was pressed on the mouse or pointer device to trigger the event.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button
+         * @type {number}
+         */
+        this.button = 0;
+        /**
+         * Indicates which buttons are pressed on the mouse or pointer device when the event is triggered.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons
+         * @type {number}
+         */
+        this.buttons = 0;
+        /**
+         * The width of the pointer's contact along the x-axis, measured in CSS pixels.
+         * radiusX of TouchEvents will be represented by this value.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/width
+         * @type {number}
+         */
+        this.width = 0;
+        /**
+         * The height of the pointer's contact along the y-axis, measured in CSS pixels.
+         * radiusY of TouchEvents will be represented by this value.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/height
+         * @type {number}
+         */
+        this.height = 0;
+        /**
+         * The angle, in degrees, between the pointer device and the screen.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/tiltX
+         * @type {number}
+         */
+        this.tiltX = 0;
+        /**
+         * The angle, in degrees, between the pointer device and the screen.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/tiltY
+         * @type {number}
+         */
+        this.tiltY = 0;
+        /**
+         * The type of pointer that triggered the event.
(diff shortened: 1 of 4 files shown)


src/interaction/InteractionData.js
```diff
@@ -58,4 +145,44 @@ export default class InteractionData
...
+        this.buttons = event.buttons;
+        this.width = event.width;
+        this.height = event.height;
+        this.tiltX = event.tiltX;
+        this.tiltY = event.tiltY;
+        this.pointerType = event.pointerType;
+        this.pressure = event.pressure;
+        this.rotationAngle = event.rotationAngle;
+        this.twist = event.twist || 0;
+        this.tangentialPressure = event.tangentialPressure || 0;
```
bigtimebuddy: What about doing these assignments with assign. Probably need to delete isPrimary, but something like this:

```js
Object.assign(this, { twist: 0, tangentialPressure: 0 }, event); 
```

src/interaction/InteractionData.js
```diff
@@ -40,6 +40,93 @@ export default class InteractionData
...
+         * @type {number}
+         */
+        this.rotationAngle = 0;
+        /**
+         * Twist of a stylus pointer.
+         * @see https://w3c.github.io/pointerevents/#pointerevent-interface
+         * @type {number}
+         */
+        this.twist = 0;
+        /**
```
bigtimebuddy: Very small stylistic thing, but could you add a new-line after the property assignment and before the nest Jsdoc block?

src/interaction/InteractionData.js
andrewstart: The event is a full MouseEvent or Touch object, so has a number of other properties we are not interested in.

bigtimebuddy: 👍

bigtimebuddy:
Ping @themoonrat would like to get your thoughts about this. Eager to get v4.4.4 out the door.

themoonrat:
Had a long look. I sort of agree with the premise of this PR, sort of don't. Lemme explain

When I first put together the almighty hack for pointer events and normalising them (sorry everyone!) I made it so that events even if a device did support pointer events, it also listened to mouse and touch events. That way, if you listened to 'mousedown', you saw an actual MouseEvent in originalEvent; if you listened to 'touchstart' you got actual TouchEvent in originalEvent. If you listened to 'pointerdown', you either got an actual PointerEvent, or a MouseEvent/TouchEvent 'polyfilled' with properties added to be accessed as if it was a PointerEvent. This meant anyone using 'mousedown' or 'touchstart' wouldn't see any changes to how they were working at all; only the newest API got polyfilled in.

In the simplifying and removal of code duplication since, if you listen to 'touchstart', you are no longer guaranteed to have a TouchEvent. It will either be a PointerEvent if supported, or a TouchEvent. This feels... wrong. Any developer wanting to just access touch events should get a TouchEvent.  Any developer wanting to access mouse events should get a MouseEvent. These have been around for many years now, and supported everywhere, so it feels wrong that these are now hobbled.  - that problem. This PR doesn't solve https://github.com/pixijs/pixi.js/issues/3886 - because on listening to a touch event, he still can't access the properties that a TouchEvent contains. 

Having said that... this PR how I feel the future of interaction in PIXI should be. Our own object containing normalised properties that removes the requirement of even needing the originalEvent. Only 1 way to listen to events, pointer events, with no listeners available for mouse & touch specific events.

So... yeah. I sorta approve, sorta don't!

themoonrat:
As a clarification after the discussion on slack, my thoughts are now.

The bug at https://github.com/pixijs/pixi.js/issues/3886 should be solved by event.data.originalEvent.touches existing if listening to 'touch' events, however that is done. I've softened my stance on 'it must be an actual TouchEvent they are looking at'; I don't think that is a requirement. But if you look at the originalEvent for a 'touch' callback, all of the required TouchEvent properties must be there, or else we've broken existing code.

I agree with this PR in the longer term thinking of 'only allow listening to pointer events, no need to expose originalEvent, just add the required data to event.data', but that is a breaking API change.

andrewstart:
Sounds like a plan. I'll close this and make a new PR that restores old listening behavior.

PR #3894, opened 2017-03-29
autoFit temporary fix

ivanpopelyshev:
previously known as "dontFit" and "fitToDestination"


src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -107,6 +107,7 @@ export default class FilterManager extends WebGLManager
             // TODO we should fit the rect around the transform..
         }
         else
+        if (filters[0].autoFit)
```
andrewstart: Should `else` and `if` be on the same line?

ivanpopelyshev: fixed

PR #3896, opened 2017-03-30
Fix for TilingSprite.tilePosition with textureResolution > 1 on CanvasRenderer

### LINKED ISSUE #3892: TilingSprite CanvasRenderer @2x images tilePosition bug
Hello everyone

I have noticed that when animating the `tilePosition.x` of an instance of `PIXI.extras.TilingSprite` with the canvasRenderer and the sprite is built from a source retina @2x image e.g. myImage@2x.png there is a jump when the new tile is added to the left, I think its a simple miscalculation error.

I have made an example, simply watch a wait until the tilePosition.x is a factor of  256px

http://jsfiddle.net/brenwell/kn6zvpnL/12/

I looked through the issues and there seemed to be many similar issues but not this exact one

Thanks for any help in advance

brenwell:
Fixes #3892 a small calculation issue with tilingsprites.

You can see the bug in action here

http://jsfiddle.net/brenwell/kn6zvpnL/13/

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,15 +43,7 @@ Pixi.js has full [WebGL](https://en.wikipedia.org/wiki/WebGL) support and seamle
 
 ### Setup ###
 
-It's easy to get started with Pixi.js! Simply grab the pre-built versions from here:
-
-Release Branch - Nice and stable Pixi.js
-- Unminified: [http://pixijs.download/release/pixi.js]
-- Minified: [http://pixijs.download/release/pixi.min.js]
-
-Develop Branch - The bleeding edge version of Pixi.js
-- Unminified: [http://pixijs.download/dev/pixi.js]
-- Minified: [http://pixijs.download/dev/pixi.min.js]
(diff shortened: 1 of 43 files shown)


brenwell:
Oh shit it contains all my previous commit messages

brenwell:
My previous commits have been accepted and merged into Dev already I believe.
How can I prevent these messages from appearing here and only having my last commit mentioned in the pull request?

bigtimebuddy:
@brenwell don't worry about it, we'll squash these commits when it is merged in.

brenwell:
Ok Cheers @bigtimebuddy. But for future reference how can I prevent that. Should I be using *rebase* instead?

englercj:
You should branch from dev to make changes, one branch per PR. Also consider rebasing onto `dev` if a bunch of changes make it in, to ensure your PR is on the latest version.

Either way, it is fine; we will squash these out later.

brenwell:
Ok, will do in future. Cheers

Issue #3876, opened 2017-03-25
Module / namespace / class documentation


drkibitz:
PIXI's documentation has always been great. This is just an issue to make it better. There really needs to be a description of the purpose and usage of a module, namepace, and class.

One in particular is `PIXI.{const name}` versus `PIXI.settings`. What's the difference, why are certain values in one, and others in the other? I would like some documentation to tell me, otherwise I will be going on forums, chat, or create an issue to ask.

Other good examples are just about every available namespace, especially:
- PIXI.prepare: http://pixijs.download/release/docs/PIXI.prepare.html
- PIXI.extras: http://pixijs.download/release/docs/PIXI.extras.html
- PIXI.utils: http://pixijs.download/release/docs/PIXI.utils.html
- PIXI.ticker: http://pixijs.download/release/docs/PIXI.ticker.html (I'm a guilty party for this one)
- and more...

I know people who have been really into PIXI probably think they are self-explanatory. This is at least one person saying they are really not 🦆 .

bigtimebuddy:
Great suggestion. Anyone wanna out there in Pixi-land wanna take a crack at these? Would definitely welcome PR for v4.5.0.

bigtimebuddy:
@drkibitz I added PR #3907 to start to add namespace docs. Would love your feedback on that.

drkibitz:
@bigtimebuddy sorry never got a chance to look at it before merging.

drkibitz:
Everyone else's comments were mine ;) thanks!

drkibitz:
Ok one comment, if `PIXI.settings` is "User customizable", how? If you want to `Object.freeze` it here https://github.com/pixijs/pixi.js/issues/3875#issuecomment-289424833, what is the usage of this? Maybe a little more usage examples would help to understand some of these namespaces, like the settings for one, and other plugins.

bigtimebuddy:
@drkibitz started a new PR for you:  #3876

Also, PIXI.settings should use seal, not freeze. Consts should use freeze, since they are static and don't change. I took a pass at this, but it's complicated because of the deprecations. So I just left them alone for now.

drkibitz:
@bigtimebuddy With that PR I think you've taken care of the goal of this issue.

PR #3907, opened 2017-04-03
Documentation Updates: namespaces & plugins

bigtimebuddy:
### Added

* Adds documentation for missing namespaces like `PIXI.extract`, `PIXI.extras`, etc.
* Adds documentation for `plugins` on the CanvasRenderer and WebGLRenderer.

### Changed

* Added CanvasExtract and WebGLExtract classes to `PIXI.extract` namespace instead of on the global `PIXI` object (mirrors how `prepare` is documented).
* Disables `hardwrap` JSdoc setting because ESLint enforces a line-length limit.

This addresses #3876

#### Preview

https://pixijs.download/docs/docs/index.html

### DIFF
--- a/src/accessibility/index.js
+++ b/src/accessibility/index.js
@@ -1,4 +1,8 @@
 /**
+ * This namespace contains a renderer plugin for interaction accessibility for end-users
+ * with physical impairments which require screen-renders, keyboard navigation, etc.
+ *
+ * Do not instanciate this plugin directly. It is available from the `plugins` renderer property.
  * @namespace PIXI.accessibility
  */
 export { default as accessibleTarget } from './accessibleTarget';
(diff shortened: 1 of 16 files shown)


bigtimebuddy:
Good suggestions @themoonrat, done.

src/accessibility/index.js
```diff
@@ -1,4 +1,9 @@
 /**
+ * This namespace contains a renderer plugin for interaction accessibility for end-users
+ * with physical impairments which require screen-renders, keyboard navigation, etc.
+ *
+ * Do not instanciate this plugin directly. It is available from the `renderer.plugins` property.
```
andrewstart: instantiate, not instanciate, right?

src/core/ticker/index.js
```diff
@@ -53,6 +53,10 @@ shared.destroy = () =>
 };
 
 /**
+ * This namespaces contains an API for interacting with PIXI's internal global loop.
```
andrewstart: 'namespace', singular. It might just be me, but I feel like there should be another word between global and loop, like 'global update loop'. Then again, that sounds like a little more of a mouthful than it should be.

bigtimebuddy:
Thanks @andrewstart, fixed.

PR #3908, opened 2017-04-04
Added TextStyle.measure function

darkfrog26:
Added a measure function to `TextStyle`. It's been a long time since I've written much JavaScript, so let me know if I missed something.  Ideally, I'd like to get rid of the necessity of the `canvas` parameter if you have any ideas?

Applies to https://github.com/pixijs/pixi.js/issues/3903).


themoonrat:
Firstly, I think this function should be in Text, not TextStyle. It references the Text class (which creates a circular dependency between the two) and it also is just a copy of code already in Text. It's much neater if the Text class used your split out function; you go down to just 1 version of the code, and your static function can now access other static functions within the Text class freely.

Second, there are a couple of key lines in this function that do exist in the Text class version that are critical to accurately measuring the text. One is actually setting the font on the canvas to measure, the other is wordWrapping, which greatly effects the potential dimensions of the text
```
this.context.font = this._font;
const outputText = style.wordWrap ? this.wordWrap(this._text) : this._text;
```

As for the not having to pass a canvas through; how about the canvas parameter becomes the second one, and optional. If one is passed through, it uses it, if it not, there is a special small 'measureCanvas' that is created file scope within Text.js, and can be re-used in this circumstance.

darkfrog26:
Do you have a problem with it being a static function in `Text` instead of an instance function?

themoonrat:
No problem with that at all

darkfrog26:
The only issue is that `wordWrap` function also must become static for me to be able to utilize it.

darkfrog26:
Updated, but I haven't modified the `updateText` function to utilize it yet.  I wanted to get some additional feedback before I did so.

darkfrog26:
The only problem I have integrating this into `updateText` is that it utilizes `lines`, `lineWidths`, `maxLineWidth`, and `fontProperties`.  Would you like me to return an object that includes all of this information instead of `Rectangle` or do you have another suggestion of how to handle this?

src/core/text/Text.js
```diff
@@ -277,6 +277,61 @@ export default class Text extends Sprite
...
+     * @param {HTMLCanvasElement} canvas - optional specification of the canvas to use for measuring.
+     * @return {Rectangle} measured width and height of the text.
+     */
+    static measure(text, style, wordWrap, canvas)
+    {
+        if (!canvas)
+        {
+            this.canvas = this.canvas || document.createElement('canvas');
+            canvas = this.canvas;
+        }
```
darkfrog26: @bigtimebuddy fixed

src/core/text/Text.js
```diff
@@ -411,16 +466,23 @@ export default class Text extends Sprite
...
      * @return {string} New string with new lines applied where required
      */
-    wordWrap(text)
+    static wordWrap(text, style, canvas)
     {
+        if (!canvas)
+        {
+            this.canvas = this.canvas || document.createElement('canvas');
+            canvas = this.canvas;
+        }
```
darkfrog26: @bigtimebuddy fixed

darkfrog26:
@themoonrat or @bigtimebuddy I'm not sure what I'm doing wrong.  I added a `MeasuredText` object but it apparently cannot be found by `Text` and I'm not sure why.  I haven't ever worked on a structured project like this in JavaScript so some help would be greatly appreciated.

darkfrog26:
Nevermind, I figured it out.  I hadn't noticed the import functionality.  If you guys have no further suggestions for changes I believe this is ready for merge.

bigtimebuddy:
Organizationally, I'm not feeling 100% on this PR, but we're getting closer!

I would suggest moving MeasuredText class to **src/core/text** (alongside Text and TextStyle) and renaming it to `TextMetrics`. Potentially, you can move `Text.measure` and `Text.wordWrap` to this new TextMetrics class.

TextMetrics should contain all functionality related to measuring and determining text. Text is responsible for rendering it. TextStyle maintains the settings.

darkfrog26:
@bigtimebuddy okay, I implemented your suggestions.  I agree that's a better route.  In fact, that's very similar to what I wanted in the first place. :)

darkfrog26:
Okay, now I'm really done.

src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,229 @@
...
+        );
+    }
+
+    /**
+     * Applies newlines to a string to have it optimally fit into the horizontal
+     * bounds set by the Text object's wordWrapWidth property.
+     *
+     * @param {string} text - String to apply word wrapping to
+     * @param {TextStyle} style - the style to use when wrapping
+     * @param {HTMLCanvasElement} canvas - optional specification of the canvas to use for measuring.
```
bigtimebuddy: Couple of doc things here:
* `{TextStyle}` should be `{PIXI.TextStyle}`
* `canvas` should be `[canvas]` to denote optional

bigtimebuddy:
Also, to expose TextMetrics, you need to add to core. Like here:
https://github.com/pixijs/pixi.js/blob/dev/src/core/index.js#L27-L28

darkfrog26:
@bigtimebuddy oops, sorry about that...it's fixed now.

bigtimebuddy:
Could you explain the need for `clone` and `equals`? They are not used internally and I would rather not include them, unless there's a clear reason. You could use something like lodash's `_.isEqual` for comparing.

darkfrog26:
@bigtimebuddy removed. I added them because I was modeling based on `Point` and I thought perhaps it was a standard requirement for them to be present so I went ahead and added them just in case.

bigtimebuddy:
What have you been using to test this against? I might suggest running against these two examples below. I want to make sure this change doesn't break anything (we don't have good unit-test coverage for Text):

http://pixijs.github.io/examples/#/demos/text-demo.js
http://pixijs.github.io/examples/#/basics/text.js

src/core/text/Text.js
```diff
@@ -130,48 +131,11 @@ export default class Text extends Sprite
...
-            maxLineWidth = Math.max(maxLineWidth, lineWidth);
-        }
-
-        let width = maxLineWidth + style.strokeThickness;
-
-        if (style.dropShadow)
-        {
-            width += style.dropShadowDistance;
-        }
+        const measured = TextMetrics.measure(this._text, this._style, this._style.wordWrap, this.canvas);
```
bigtimebuddy: You _may_ want to consider destructuring the metrics object here since you're already doing it with `width` and `height`.

```js
const { width, height, lines, lineHeight, lineWidths, maxLineWidth, fontProperties } = TextMetrics.measure(
    this._text, this._style, this._style.wordWrap, this.canvas
);
```

src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Measures the supplied string of text and returns a Rectangle.
+     *
+     * @param {string} text - the text to measure.
+     * @param {PIXI.TextStyle} style - the text style to use for measuring
+     * @param {boolean} [wordWrap] - optional override for if word-wrap should be applied to the text.
+     * @param {HTMLCanvasElement} [canvas] - optional specification of the canvas to use for measuring.
+     * @return {PIXI.TextMetrics} measured width and height of the text.
+     */
+    static measure(text, style, wordWrap, canvas)
```
bigtimebuddy: Maybe `create` is a better name than `measure`?

src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Applies newlines to a string to have it optimally fit into the horizontal
+     * bounds set by the Text object's wordWrapWidth property.
+     *
+     * @param {string} text - String to apply word wrapping to
+     * @param {PIXI.TextStyle} style - the style to use when wrapping
+     * @param {HTMLCanvasElement} [canvas] - optional specification of the canvas to use for measuring.
+     * @return {string} New string with new lines applied where required
+     */
+    static wordWrap(text, style, canvas)
```
bigtimebuddy: `wordWrap` doc should have a `@private` flag

src/core/text/Text.js
darkfrog26: I was actually leaning the other direction toward getting rid of the `width` and `height` references to make it a bit more clear.

src/core/text/TextMetrics.js
darkfrog26: I don't think so as `create` has less of an implication of "work" being done and `measure` is very clear what its function is.

src/core/text/TextMetrics.js
darkfrog26: Why should `wordWrap` be private?  It's perfectly valid to use externally.  Currently, I don't have a use-case for that, but I could imagine that it might be useful.  No reason to hide something that *could* have potential benefits without an explicit reason to do so.

src/core/text/Text.js
bigtimebuddy: The advantage is in the minified file size and performance. If you have multiple instances, it's a little better to cache the variable. here's an example:

```js
const measure = TextMetrics.measure();
const width = measure.width;
const height = measure.height;
```
Would minify to:
```js
const a = TextMetrics.measure();
const b = a.width;
const c = a.height;
````

The more you reference `measure.width`, etc the less compressed the source becomes after minification and the more times it needs to use the dot-accessor.

File: src/core/text/TextMetrics.js (hunk 5 above)
Comment (bigtimebuddy) [inline src/core/text/TextMetrics.js]: Makes sense.

File: src/core/text/TextMetrics.js (hunk 6 above)
Comment (bigtimebuddy) [inline src/core/text/TextMetrics.js]: I understand, but we make a conscious effort to minimize our API as small as possible. One less method is one less thing that needs to be changed/deprecated, one less thing to unit-test, etc. I would like this to be private until someone actually has a use-case for it.

File: src/core/text/Text.js (hunk 4 above)
Comment (darkfrog26) [inline src/core/text/Text.js]: Done

File: src/core/text/TextMetrics.js (hunk 6 above)
Comment (darkfrog26) [inline src/core/text/TextMetrics.js]: @bigtimebuddy done

**Comment by darkfrog26** [comment]
@themoonrat is there something you're waiting on to approve these changes?

**Comment by themoonrat** [comment]
I just need a bit of time to take another look through after the changes you and bigtimebuddy have gone through, which I hope to do at some point this weekend :)

**Comment by darkfrog26** [comment]
@themoonrat okay, just hoping to be able to leverage this functionality in my project soon. :)

**Comment by themoonrat** [comment]
Don't worry, me too! I really appreciate the work you've put in to this... the version of pixi I uses actually modified certain functions around that area, and every time a commit goes into PIXI.Text and I have to do a merge it's a nightmare resolving it, so really looking forward to that being easier to deal with after this PR

File: src/core/text/TextMetrics.js [hunk 7]
```diff
@@ -0,0 +1,191 @@
...
+        if (!canvas)
+        {
+            TextMetrics._canvas = TextMetrics._canvas || document.createElement('canvas');
+            canvas = TextMetrics._canvas;
+        }
+
+        wordWrap = wordWrap || style.wordWrap;
+        const font = Text.getFontStyle(style);
+        const fontProperties = Text.calculateFontProperties(font);
+        const context = canvas.getContext('2d');
```
themoonrat: Could we move this to straight after getting the canvas please, as it makes sense to get the context straight after knowing what canvas you have. 

In fact, there is almost a case for passing a context for a canvas through, rather than a canvas itself, so this class isn't doing the same getting of the context over and over (when this function was in Text, it had a cached context) but I don't know how others feel about that. @bigtimebuddy?

src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * @param {string} text - the text to measure.
+     * @param {PIXI.TextStyle} style - the text style to use for measuring
+     * @param {boolean} [wordWrap] - optional override for if word-wrap should be applied to the text.
+     * @param {HTMLCanvasElement} [canvas] - optional specification of the canvas to use for measuring.
+     * @return {PIXI.TextMetrics} measured width and height of the text.
+     */
+    static measure(text, style, wordWrap, canvas)
+    {
+        if (!canvas)
```
themoonrat: Is it worth just creating this dummy canvas element up front, so you could do a much simpler
`canvas = canvas || TextMetrics._canvas`
In the two places you currently do that?

You may also want to check if you could set the width of this temp canvas to a small width, like it is by default for Text canvases, to take up less memory. If it lets the canvas be, say, 3x3 whilst still measuring text correctly, then perfect!

src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,191 @@
+import Text from './Text';
```
themoonrat: This feels... wrong, as there's now a circular dependency between Text and TextMetrics.

The only thing you are using from there are 2 static functions, getFontStyle and calculateFontProperties.
Either those could be moved here, or perhaps better, we could have another new file which just contains text utils? This could be where TextMetrics and measure live, and getFontStyle, calculateFontProperties and wordWrap could go in textUtils.js? @bigtimebuddy again?

darkfrog26: @themoonrat there's nothing wrong with a circular dependency here.  I think those methods perfectly validly fit in `Text`.  If at some point you'd like to move them to text utils that's fine, but that seems out-of-scope for this PR.

src/core/text/TextMetrics.js
darkfrog26: I don't think it's worth creating up-front considering that unless you are using `measure` directly like I intend to from my library, then a canvas would never be created at all.

As for creating a 3x3 canvas I wasn't aware it would make any difference if you aren't drawing to it.

src/core/text/TextMetrics.js
themoonrat: Hey. I'm sorry you feel that way. I don't think any of my comments are a "must change" which is why I've referenced a fellow maintainer to see if my views are shared or not.

Please don't take the reviews as 'criticised to death', I've long learnt to never take code reviews personally! We really appreciate you helping; but as you say, we're just trying to maintain good code in a consistent style, in an area that is unfortunately poorly unit tested right now.

bigtimebuddy: Hey @darkfrog26, saw your comment here. 

First off, thanks for getting this PR so far! It's really awesome, and I know it will be useful for devs for doing things like calculating text-fitting. I know the process can feel a little nit-picky sometimes, but with so many people collaborating and using such a big library, we're working hard to keep things consistent, stable and organized. And, this is a pretty sizable refactor of Text, so just wanna make sure we nail it.

How would you feel if we merge this in and then @themoonrat and I can continue to make a few small edits? Not everyone has infinity time to collaborate on PRs, so we are happy to take it off your hands. Let me know.

darkfrog26: @themoonrat I don't mind the criticisms, I never take them personally.  It's simply that I'm ready to be done with this so I can move on to other things. :)

@bigtimebuddy yes, that would be the preferred course.  I don't mind any changes you make after the merge as long as I can still access the measurement of text without drawing anything. :)

bigtimebuddy: Sounds good! Thanks again for the help.

darkfrog26:
Any idea when I can expect a release with this functionality?

bigtimebuddy:
This will most likely be in 4.5.0, which is slated for this coming week, if everything goes well.

PR #3913, opened 2017-04-06
Change set tint implementation on BitmapText

verochan:
When using bitmaptext and trying to edit the tint of an already created bitmaptext like this:
        ` this.myCustomTextObject().getCustomPixiText().tint="0x021c4a";`
There's no way of changing the color, it doesn't matter if the color is passed like 0x021c4a or #021c4a, or trying other colors, it always gets the 0xFFFFFF color of the tint function assigned:

```
set tint(value) // eslint-disable-line require-jsdoc
    {
        this._font.tint = (typeof value === 'number' && value >= 0) ? value : 0xFFFFFF;

        this.dirty = true;
    }
```

On my project I've modified the pixi library with the below change, and now it's updating correctly the color of the bitmaptext, so I suggest to apply this change on the pixi library for all users not to suffer this problem:

```
set tint(value) // eslint-disable-line require-jsdoc
    {
        this._font.tint =value;

        this.dirty = true;
    }
```

### DIFF
--- a/src/extras/BitmapText.js
+++ b/src/extras/BitmapText.js
@@ -326,7 +326,7 @@ export default class BitmapText extends core.Container
 
     set tint(value) // eslint-disable-line require-jsdoc
     {
-        this._font.tint = (typeof value === 'number' && value >= 0) ? value : 0xFFFFFF;
+        this._font.tint = value;
 
         this.dirty = true;
     }


englercj:
Can you show me an example of this not working when you assign it a number? Note that the code you posted (`this.myCustomTextObject().getCustomPixiText().tint="0x021c4a";`) is incorrect, it is assigning a string. What it should be is `this.myCustomTextObject().getCustomPixiText().tint=0x021c4a;`.

verochan:
As a number it works ok, but at first I was expecting it to work with strings just like Pixi.Text and at BitmapText creation, I mean, why when reassigning a tint value it needs to be a number and when creating a bitmaptext it takes it as a string? It's confusing, for example:

```
var styles={font: "11px MikadoBold", tint: "0xdeb4ce"};
var textString="my text string";
var message=new PIXI.extras.BitmapText(textString, styles);
```

When creating the BitmapText and passing the tint as a string (like passing the fill property to Pixi.Text as a string) it works ok, I think that in order to not to lose "coherence", tint should also accept the value as a string, just like Pixi.Text also accepts the related color parameters as strings.

themoonrat:
I can see where verochan is coming from. I find it very handy that tinting of sprites works with strings such as "0x123456", as it means I can just use .json configuration through ( hex numbers of the style 0x123456 is not valid json ).

englercj:
I'm pretty surprised that passing a string tint in the constructor works at all. I'd like to see an example of that, because I can't see how it would.

Tint is a number everywhere, we don't use string tints anywhere. You can use a string color for the font of `Text` because it actually uses CSS for styling the font. However, `tint` is a sprite property and `BitmapText` is just creating a sprite per glyph of font. When you set the `tint`, it just passes it down into the underlying sprites and therefore must be a number.

I don't mind expanding the accepted values for tint to include other types, but we must do it everywhere and it needs to just convert into a number under the hood anyway.

verochan:
You can see an example here:

http://pixijs.github.io/examples/#/demos/text-demo.js

On the line that is created bitmapFontText, add the tint color like this:

`var bitmapFontText = new PIXI.extras.BitmapText('bitmap fonts are\n now supported!', { font: '35px Desyrel', align: 'right', tint:'0xFF0000' });`

You can check the 'bitmap fonts are now supported' turns red:
test

As you say, the change should be done everywhere or it doesn't make sense, if you do a different Pr can you paste here the link to follow the state of the task? Thanks very much.

PR #3921, opened 2017-04-08
Measure Text API, expanded

bigtimebuddy:
Expands on the work of #3908.

### 🎁 Added

* Adds new `PIXI.TextMetrics` class for calculating Text and Font layout sizes. Makes it possible to measure fitting size of text and and a TextStyle for doing auto-fitting of text. #3921 #3908
  * `PIXI.TextMetrics.measureText` method for calculating the size of text fill.
  * `PIXI.TextMetrics.measureFont` method for calculating the font max and min dimensions.

### 🚽 Deprecated

* `PIXI.Text.calculateFontProperties` becomes `PIXI.TextMetrics.measureFont`  #3921
* `PIXI.Text.getFontStyle` becomes `PIXI.TextStyle#toFontString`  #3921
* Deprecates the following private static properties on Text: `fontPropertiesCache`, `fontPropertiesCanvas`, `fontPropertiesContext`  #3921


src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,191 @@
+import Text from './Text';
```
englercj: Does this actually work though with the circular dependency? Normally, when you have a circular import like this one of them just ends up being an empty object and it doesn't work (wichever comes second). I think this either needs to be removed or have unit tests proving it works.

bigtimebuddy:
@themoonrat: I moved a few things around here. Namely, added getFontStyle to TextStyle and moved the calculate stuff to TextMetrics. Feel free to review and/or add whatever else. Would like to get this into 4.5.0. Also, @englercj, please review.

src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,315 @@
...
+                const wordWidth = context.measureText(words[j]).width;
+
+                if (style.breakWords && wordWidth > wordWrapWidth)
+                {
+                    // Word should be split in the middle
+                    const characters = words[j].split('');
+
+                    for (let c = 0; c < characters.length; c++)
+                    {
+                        const characterWidth = context.measureText(characters[c]).width;
```
andrewstart: Would it be reasonable to cache measured characters (including space) for the duration of this method? That assumes that the hit from making a new object and reading from that cache is lower than `context.measureText()`.

src/core/text/TextMetrics.js
```diff
@@ -0,0 +1,315 @@
...
+ * @property {number} fontSize - Font size from ascent to descent
+ */
+
+/**
+ * Cached canvas element for measuring text
+ * @memberof PIXI.TextMetrics
+ * @type {HTMLCanvasElement}
+ * @private
+ */
+TextMetrics._canvas = document.createElement('canvas');
```
themoonrat: Were you able to see if the metric calculations still worked if we shrank this canvas to a ridiculously small size to save a little memory? Should still work I think.

src/core/text/TextMetrics.js
bigtimebuddy: It seems like caching these characters makes some difference (9ms vs 4ms in my demo), especially if there's more duplicate characters..

src/core/text/TextMetrics.js
bigtimebuddy: You mean, doing something like this:

```js
TextMetrics._canvas.width = 1;
TextMetrics._canvas.height = 1;
```

themoonrat: Yeah. I'd try 10x10 because i don't trust browser vendors, but that'd be the idea.

bigtimebuddy: Do you know if this really makes a difference, even if it's detached from the DOM? I dug a little online and couldn't find anything.

andrewstart: I imagine it uses a little less memory for pixel data. I do recall reading about a minimum size in some browsers that GPU acceleration doesn't kick in under.

bigtimebuddy: Okay, I changed to be 10x10. It doesn't seem to hurt.

bigtimebuddy:
Ping @themoonrat or @andrewstart. Good?

Issue #3920, opened 2017-04-08
Lack of Event Documentation


darkfrog26:
This has been driving me crazy for a while, but the utter lack of documentation regarding events is very frustrating.  I see that `Texture` does fire an `update` event, but what does that event look like?  What properties does it contain?  Why is it not documented?

englercj:
Because someone hasn't written it yet :)

You can see what is emitted here: https://github.com/pixijs/pixi.js/blob/dev/src/core/textures/Texture.js#L200

Its just itself as the only parameter.

darkfrog26:
Is that the case with most events?  Do they typically just fire with a reference to the owner instead of an actual event object?

axelvaindal:
If I remember correctly, events are based on `EventEmitter` provided by the Node.js implementation, which you can find here : https://nodejs.org/api/events.html. 
This is probably why the base about events is not documented, this would be redundant.

However, based on source code, many fired event in Pixi are given with an object containing :

- the name of the event
- the data about the event (thus, you get the event target and position)

This is actually the case for all `DisplayObject` ( such as `Container` or `Sprite` for instance), as you can see there : https://github.com/pixijs/pixi.js/blob/dev/src/interaction/InteractionManager.js 

For specific case, such as `Texture`, I'm afraid you will have to jump into the source code to see which data are actually provided during the event emitting considering no documentation about this exist at the moment.

englercj:
> Is that the case with most events? Do they typically just fire with a reference to the owner instead of an actual event object?

No, just the one you mentioned. The event can emit anything we pass in, and we will pass along what is relevant for that event. Best thing to do is to read the source and improve the docs as you go along. @bigtimebuddy has already put in a PR to improve many of them.

bigtimebuddy:
Fixed with #3922

darkfrog26:
@bigtimebuddy thanks. :)

PR #3922, opened 2017-04-10
Documents events for DisplayObject, renderers, Texture and BaseTexture

bigtimebuddy:
Addresses #3920 

### Added

* Adds additional event documentation including parameters for:
  * Texture
  * BaseTexture
  * DisplayObject
  * WebGLRenderer
  * CanvasRenderer

#### Demo

https://pixijs.download/event-docs/docs/index.html

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -357,6 +357,7 @@ export default class InteractionManager extends EventEmitter
          * Fired when the operating system cancels a pointer event
          *
          * @event PIXI.interaction.InteractionManager#pointercancel
+         * @param {PIXI.interaction.InteractionData} event - Interaction data
          */
 
         /**
@@ -413,6 +414,7 @@ export default class InteractionManager extends EventEmitter
          * Fired when the operating system cancels a touch
          *
          * @event PIXI.interaction.InteractionManager#touchcancel
+         * @param {PIXI.interaction.InteractionData} event - Interaction data
          */
 
         /**
@@ -550,6 +552,7 @@ export default class InteractionManager extends EventEmitter
          * DisplayObject's `interactive` property must be set to `true` to fire event.
          *
          * @event PIXI.DisplayObject#pointercancel
+         * @param {PIXI.interaction.InteractionData} event - Interaction data
          */
 
         /**
@@ -614,6 +617,7 @@ export default class InteractionManager extends EventEmitter
          * DisplayObject's `interactive` property must be set to `true` to fire event.
          *
          * @event PIXI.DisplayObject#touchcancel
+         * @param {PIXI.interaction.InteractionData} event - Interaction data
          */
 
         /**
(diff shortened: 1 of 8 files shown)


themoonrat:
I believe you can do the events as something like
`@event PIXI.WebGLRenderer#postrender`
rather than
```
@event postrender
@memberof PIXI.WebGLRenderer#
```
Just in case you weren't aware. It's personal preference between the cleanliness of an event name on it's own line, or reducing a line used for document an event. I'm not bothered either way.

There are also now some places where we could add the @listens jsdoc, for example, onBaseTextureLoaded for PIXI.BaseTexture#loaded, and onBaseTextureUpdated for PIXI.BaseTexture#updated within Texture.js

bigtimebuddy:
We should do `@fires` and `@listens` for another PR. I mostly wanted to get the missing events documented.

darkfrog26:
@bigtimebuddy it seems there's a ton of missing information regarding events like mouse and keyboard events.  I would expect mouse events to be listed on `DisplayObject`. Is that not a valid assumption?

bigtimebuddy:
Okay, I also added the touch/mouse/pointer events to the DisplayObject (these were documented with the InteractionManager).

Also, there were some missing properties in the documentation like `interactive`, `interactiveChildren` and `hitArea`, all of which are relevant to interaction events.

src/interaction/InteractionManager.js
```diff
@@ -256,210 +256,419 @@ export default class InteractionManager extends EventEmitter
...
+         * @event PIXI.interaction.InteractionManager#pointerup
+         * @param {PIXI.interaction.InteractionData} event - Interaction data
          */
 
         /**
          * Fired when the operating system cancels a pointer event
          *
-         * @event pointercancel
-         * @memberof PIXI.interaction.InteractionManager#
+         * @event PIXI.interaction.InteractionManager#pointercancel
```
andrewstart: `pointercancel` and `touchcancel` should have a parameter like the other events.

bigtimebuddy: I was wondering about that. Okay, fixed that.

bigtimebuddy:
@darkfrog26 I think I see part of the problem. Namely, that DisplayObject has a bunch of mixins (there's a link at the top of the DisplayObject doc for interactiveTarget). But this is pretty hidden. Properties like `hitArea`, `cursor`, `interactive`, etc were somewhat obscured and definitely weren't showing up in the list on the lift. My recent commit fixes this. 

Also, all the events should now be showing up for all DisplayObject.

darkfrog26:
It would also be beneficial if you'd specify the `originalEvent` type beyond just `Event`.  I presume the `originalEvent` for a mouse `mousedown` would be a `MouseEvent` (https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent)?

darkfrog26:
@bigtimebuddy I haven't checked to verify if this code is different than 4.4.4, but `click` on a `Sprite` doesn't fire an `InteractionData` it fires an `InteractionEvent` instead.

bigtimebuddy:
You are correct @darkfrog26. Thanks for pointing this out. See PR #3923.

PR #3923, opened 2017-04-11
Fixes event docs for Interaction

bigtimebuddy:
Fixes the docs for mouse/touch/pointer events. These events return PIXI.interaction.InteractionEvent not InteractionData.

Follow up to #3922

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -257,7 +257,7 @@ export default class InteractionManager extends EventEmitter
          * object.
          *
          * @event PIXI.interaction.InteractionManager#mousedown
-         * @param {PIXI.interaction.InteractionEvent} event - Interaction data
+         * @param {PIXI.interaction.InteractionEvent} event - Interaction event
          */
 
         /**
@@ -265,7 +265,7 @@ export default class InteractionManager extends EventEmitter
          * on the display object.
          *
          * @event PIXI.interaction.InteractionManager#rightdown
-         * @param {PIXI.interaction.InteractionEvent} event - Interaction data
+         * @param {PIXI.interaction.InteractionEvent} event - Interaction event
          */
 
         /**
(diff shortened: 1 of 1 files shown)


darkfrog26:
Ugh, `InteractionEvent` is almost useless. Instead of updating the docs to reference it the code should be changed to point to `InteractionData`.  How are you supposed to determine things like mouse position, key typed, etc. for an event if it's sending `InteractionEvent` which has almost no information?

bigtimebuddy:
`event.data`, `event.data.originalEvent`, etc

darkfrog26:
@bigtimebuddy am I missing something?  https://pixijs.download/event-docs/docs/PIXI.interaction.InteractionEvent.html doesn't mention `data`.

bigtimebuddy:
Looks like a bug with JSDoc because it's definitely documented here: https://pixijs.download/event-docs/docs/interaction_InteractionEvent.js.html#line48

darkfrog26:
@bigtimebuddy thanks...I'm sure you can understand how frustrating it is to find holes like this when trying to use PIXI.

bigtimebuddy:
I know, I understand. Dealing with a poorly documented project sucks and is endlessly frustrating. Yet, none of us are getting paid to develop Pixi. We do it because we love it. I ask that you please show some patience if things are 100% as you expect. We do our best given that we all have day jobs doing other things. We are not Facebook, we are not Google. 

I don't know why those properties are not rendering, but I'd appreciate if you filed a bug so it's recorded. If someone wants to jump down the JSDoc rabbit hole, it's there.

Also, please consider making a PR for doc changes or additions you find. We LOVE those PRs (and they usually get merged faster than code-PRs) and they make Pixi better for everyone.

darkfrog26:
@bigtimebuddy I apologize, I wasn't trying to be critical, I was simply trying to outline the importance of fixing the documentation to be a proper representation of the code.  I'm a bit swamped at the moment but I'll do my best to submit a PR if I can.

andrewstart:
`InteractionData.identifier` is in fact a number, except in the case of the mouse, which doesn't get an identifier normally so we use `MOUSE` because that would never interfere with a real pointer id.

PR #3940, opened 2017-04-16
Fix not respecting target of mouseup events.

### LINKED ISSUE #3466: mouseup should check for originalEvent target
This might be very specific, but imagine this scenario:

You have a fullscreen canvas with a `DisplayObjectContainer` with some background.
If I add the `mouseup` event to it, when I "click" on it it fires normally as it should.

However,
If I add a `domElement` occupying a certain area of my fullscreen canvas and click on it, I expect that the event will not fire inside pixi because I clicked on another element not the `<canvas>` itself.

What happens is that the element does fire, and I could fix this in 2 ways

a)
compare the `e.data.originalEvent.target` of the event and if its not a `<canvas>` element proceed with the rest of my logic. Otherwise if its a `<canvas>`, return.
```
foo.on('mouseup', (e) => {
    // mouse up detected
    // check the original event target to see if it was a canvas 
    if (e.data.originalEvent.target.nodeName !== 'CANVAS') {
        return;
    }
    // if you do click on a canvas handle some logic.
});
```
Now, this creates a problem. the test will always pass if you click on a canvas element.
The test should check if you clicked on the domElement used on the pixi renderer.

b)
store a global reference to the renderer domElement and explicitly compare it.

Any ideas @GoodBoyDigital ?

andrewstart:
Fixes #3466, which was also reported in #3790.
The issue was not looking at the event target on our window listeners for `mouseup`/`pointerup`. This change looks at that and emits `mouseupoutside`/`pointerupoutside` as appropriate.

For the MockPointer changes, it turns out that `target` is not a parameter for Events, and is a read only property. It is configurable, however, which is why I am setting the target in a roundabout way.

### DIFF
--- a/src/core/ticker/TickerListener.js
+++ b/src/core/ticker/TickerListener.js
@@ -91,12 +91,12 @@ export default class TickerListener
             this.fn(deltaTime);
         }
 
+        const redirect = this.next;
+
         if (this.once)
         {
-            this.destroy();
+            this.destroy(true);
         }
-
-        const redirect = this.next;
 
         // Soft-destroying should remove
         // the next reference
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -1217,6 +1217,10 @@ export default class InteractionManager extends EventEmitter
 
         const eventLen = events.length;
 
+        // if the event wasn't targeting our canvas, then consider it to be pointerupoutside
+        // in all cases (unless it was a pointercancel)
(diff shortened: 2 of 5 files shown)


bigtimebuddy:
Looks good. Is there a fiddle that you can demo a before and after?

andrewstart:
Before: https://jsfiddle.net/sbxoya8x/
After: https://jsfiddle.net/e7ujs4a0/

bigtimebuddy:
These both seem to be using 4.4.4

andrewstart:
Ugh, unsaved edits. Link should be fixed now.

Issue #3119, opened 2016-10-15
Fix noise filter


englercj:
Currently the noise filter is incorrectly dealing with premultiplied alpha. I fixed this [in fae](https://github.com/Fae/fae/blob/master/plugins/filters-pack/src/noise/noise.frag) and just need to port it over.

hd719:
Hello I am new to open source and I would like to contribute to this project. 

-Thank you!

englercj:
Merged #3943 which fixes this issue!

PR #3941, opened 2017-04-16
Fixes noise filter (#3119)

YahorPadlucki:
Fixes [#3119](https://github.com/pixijs/pixi.js/issues/3119)

### DIFF
--- a/src/filters/noise/noise.frag
+++ b/src/filters/noise/noise.frag
@@ -15,11 +15,7 @@ void main()
 {
     vec4 color = texture2D(uSampler, vTextureCoord);
 
-    float diff = (rand(gl_FragCoord.xy) - 0.5) * noise;
-
-    color.r += diff;
-    color.g += diff;
-    color.b += diff;
+    color *= rand(gl_FragCoord.xy * noise);
 
     gl_FragColor = color;
 }


bigtimebuddy:
I'm not sure this fixes the problem. Maybe @englercj or @GoodBoyDigital could weigh-in. While NoiseFilter in this PR does in-fact respect alpha, The rendering doesn't look quite right.

### Example

```js
    backgroundColor: 0x006600,
    width: 200,
    height: 200
});
const graphic = new PIXI.Graphics()
    .beginFill(0xff0000)
  .drawCircle(100, 100, 50);
graphic.filters = [ new PIXI.filters.NoiseFilter() ];
app.stage.addChild(graphic);
```

### Current
screen shot 2017-04-16 at 9 10 39 am

### With This PR
screen shot 2017-04-16 at 9 11 17 am

englercj:
The "with this PR" show is what I would expect with a noise filter. What makes you feel it is wrong?

bigtimebuddy:
I would expect something similar to the current filter but maintaining alpha. This filter should be taking the pixels and applying a little darkness and a little lightness to each pixel not removing alpha from random pixels.

englercj:
Opened an alternative fix if you'd like to check that one @bigtimebuddy.

bigtimebuddy:
@YahorPadlucki Thanks a lot for your help here, but we are going to consider #3941 as better approach to this issue.

PR #3942, opened 2017-04-16
Debugging Application

bigtimebuddy:
### Added

* Adds new class `DebugApplication` which extends `Application`. This new class has a `showBounds` property, when enabled (default is `true`) draws a bounds rect for everything in the display list. This can be useful for debugging display performance problems, where somethings are overlapping or not visible.

### Example

http://pixijs.github.io/examples/?v=debug-app#/basics/container-pivot.js in the code change "Application" to "DebugApplication" and refresh the preview.

screen shot 2017-04-16 at 12 56 36 pm

### DIFF
--- a/src/core/DebugApplication.js
+++ b/src/core/DebugApplication.js
@@ -1,0 +1,134 @@
+import Container from './display/Container';
+import Application from './Application';
+import Graphics from './graphics/Graphics';
+import Rectangle from './math/shapes/Rectangle';
+import { UPDATE_PRIORITY } from './const';
+
+/**
+ * Adds visual tools for debugging the {@link PIXI.Application}.
+ * This should not be using in a production environment.
+ * @example
+ * const app = new PIXI.DebugApplication();
+ * app.showBounds = true;
+ * @class
+ * @extends PIXI.Application
+ * @memberof PIXI
+ */
+export default class DebugApplication extends Application
+{
+    /**
+     * @param {Object} [options] See {@link PIXI.Application} for options.
+     */
+    constructor(options)
+    {
+        super(options);
+
+        /**
+         * Graphics to render the bounds
+         * @member {PIXI.Graphics}
+         * @private
+         */
+        this._bounds = null;
+
+        /**
+         * Local variable for `showBounds`
+         * @member {boolean}
+         * @private
+         */
+        this._showBounds = true;
+
+        /**
+         * `true` to show the bounds around all display objects.
+         * @member {boolean}
+         * @default true
+         */
+        this.showBounds = true;
+    }
+
+    set showBounds(showBounds) // eslint-disable-line require-jsdoc
+    {
+        const stage = this.stage;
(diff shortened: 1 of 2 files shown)


themoonrat:
eeeeee, you have no idea how much I like this concept. Do you want to just get this base in now and accept ideas and PRs later, or are you up for some suggestions to implement now before it goes in?

bigtimebuddy:
My biggest question was about how to approach. 

Should this be it's own class or mixin functionality to the existing Application? My thinking was that we could provide some more debugging tools on this class at the expense of performance and not bloat Application.

The other idea I thought was should we implement this on the renderer itself as a plugin or debug option instead of a standalone class.

I guess I'm not solid on the implementation, but I think the idea is important.

themoonrat:
If I was living in a dream world, it would be enabled via an option flag in Application. That way it's easy to link up to a build environment var to enable or not.

It being on the application level makes more sense to me, with the ideas I have not always being about rendering. For example, I'd want to:
- Have a mode that lets you click anywhere on the screen, and an array of any display objects whose boundaries are within that click area are returned in the console.
- Be able to measure how long the last .render() call took.
- Count how many visible objects were rendered last .render();
- Find out how many unique assets are in the Base Texture Cache / Texture Cache, and how much memory they are likely to be taking up.

A couple of those are renderer specific, a couple of those aren't, but could be found out at application level.

bigtimebuddy:
These are awesome ideas, and would be pretty easy to implement.

There isn't really much in PIXI in terms of "developer-mode" or debugging functionality right now, so I'm struggling with how best to organize this stuff. Should these just be debug mixins to existing classes that all live together **src/debug**? Or should it be a separate optional plugin? Not sure what the best developer experience would be. 

Another idea might be to preprocess debugging functionality out of the minified build. The build system supports this functionality but we aren't really using it.

bigtimebuddy:
Another idea might actually be to develope a custom panel for Chrome DevTools which would give us a lot more information about PIXI: https://developer.chrome.com/extensions/devtools

englercj:
I like the last one better than integrating debugging tools into core!

themoonrat:
Mentioned before, but relevant here: someone already took a stab at a chrome extension. I'd also be happy going that route for these debugging options

GoodBoyDigital:
Oh a panel that showed:
- draw calls
- number of objects (inc containers)
- number of masks
- number of filters
- number of sprites
- number of graphics
- total texture usage

armed with that info people would be very empowered!

themoonrat:
I've never properly used Construct 2, but the information it has in it's debug panel is incredibly useful.

bigtimebuddy:
Okay, I'm going to close this. I _may_ make a **pixi-debug** plugin which supports this, but for now, we should find some opportunity to invest in a larger approach to QA and debugging tools. I might help upgrade that Chrome plugin to v4. Lots of great ideas here tho!

PR #3943, opened 2017-04-17
Fix #3199 by honoring alpha

### LINKED ISSUE #3199: Input not registering using IE11
So I have a little project set up which utilizes the built in "click" functionality, It works fine on firefox, chrome, opera, android browser, android chrome, android firefox and IE10, but when trying to use it in IE11 it stops regestering clicks and I cant figure out why.

There are no errors being reportedand, the only console output is "Pixi.js 4.1.0 - WebGL - http://www.pixijs.com/"

englercj:
Proposed to fix #3199, as an alternative to #3941.

Using this test code:

```js
'use strict';

    backgroundColor: 0x006600,
    width: 512,
    height: 512,
    view: document.getElementById('view'),
});

const graphic = new PIXI.Graphics()
    .beginFill(0xff0000)
    .drawCircle(256, 256, 200);

graphic.filters = [new PIXI.filters.NoiseFilter()];

app.stage.addChild(graphic);
```

Produces this output:

image

### DIFF
--- a/src/filters/noise/NoiseFilter.js
+++ b/src/filters/noise/NoiseFilter.js
@@ -28,22 +28,22 @@ export default class NoiseFilter extends core.Filter
             readFileSync(join(__dirname, './noise.frag'), 'utf8')
         );
 
-        this.noise = 0.5;
+        this.noise = Math.random();
     }
 
     /**
-     * The amount of noise to apply.
+     * The amount of noise to apply, this value should be in the range (0, 1].
      *
      * @member {number}
      * @default 0.5
      */
     get noise()
     {
-        return this.uniforms.noise;
+        return this.uniforms.uNoise;
     }
 
     set noise(value) // eslint-disable-line require-jsdoc
     {
-        this.uniforms.noise = value;
+        this.uniforms.uNoise = value;
     }
 }
--- a/src/filters/noise/NoiseFilter.js
+++ b/src/filters/noise/NoiseFilter.js
@@ -17,7 +17,8 @@ import { join } from 'path';
 export default class NoiseFilter extends core.Filter
 {
     /**
-     *
+     * @param {number} noise - The noise intensity, should be a normalized value in the range [0, 1].
+     * @param {number} seed - A random seed for the noise generation. Default is `Math.random()`.
      */
     constructor(noise = 0.5, seed = Math.random())
     {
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
I tested this out and adjusting the `noise` property doesn't seem to do anything. I believe this is suppose to be a float from 0 to 1 of the intensity of the noise. For instance:

noise

englercj:
I used it as a seed, not as intensity, so you could do this:

```
app.ticker.add(() => {
    graphic.filters[0].noise = Math.random();
});
```

Which will result in:

bigtimebuddy:
It's just that this is a breaking change. That `noise` setter currently is the amount of noise, not the randomness/seed. I would think that should be a separate property.

englercj:
Alrighty, I'll try to fix it up Soon.

englercj:
Updated, now adjusting the `noise` property looks like:

adjust-noise

And I added a `seed` property, which when adjusted gives you some actual noise:

adjust-seed-and-noise

bigtimebuddy:
Wow, so awesome!

src/filters/noise/NoiseFilter.js
```diff
@@ -29,21 +29,37 @@ export default class NoiseFilter extends core.Filter
         );
 
         this.noise = 0.5;
```
bigtimebuddy: One very small thing that would be nice is to allow a single constructor argument for the noise value with a default for `0.5`. 

```js
// Current, non-default value
const filter = new PIXI.filters.NoiseFilter();
filter.noise = 0.1;
displayObject.filters = [ filter ];

// With constructor arg
displayObject.filters = [ new PIXI.filters.NoiseFilter(0.1) ];
```

bigtimebuddy: We could do this as another PR too, but thought I'd toss it in.

englercj:
Added ctor args for `noise` and `seed`.

bigtimebuddy:
@englercj https://travis-ci.org/pixijs/pixi.js/builds/222841155#L345-L346

englercj:
Damn I suck, I'll fix it when I get home :P

englercj:
Fixed the jsdoc issue, and the branch is updating now. Then should be good to go!

PR #3944, opened 2017-04-17
Fix #3252, add an interpolation value for the color matrix result

### LINKED ISSUE #3252: SepiaFilter deleted in v4
The former SepiaFilter available in pixi v3 seems to have been deleted. Is there another filter we can use in v4 to achieve the same result?

englercj:
Add an interpolation value that allows you to interpolate between a texel's original color and the result of applying the color matrix.

This should fix #3252 where the user wanted to easily define a scale of sepia to apply.

### DIFF
--- a/src/filters/colormatrix/ColorMatrixFilter.js
+++ b/src/filters/colormatrix/ColorMatrixFilter.js
@@ -36,6 +36,8 @@ export default class ColorMatrixFilter extends core.Filter
             0, 1, 0, 0, 0,
             0, 0, 1, 0, 0,
             0, 0, 0, 1, 0];
+
+        this.interpolation = 1;
     }
 
     /**
@@ -574,6 +576,26 @@ export default class ColorMatrixFilter extends core.Filter
     set matrix(value) // eslint-disable-line require-jsdoc
     {
         this.uniforms.m = value;
+    }
+
+    /**
+     * The interpolation value to use when mixing the original and resultant colors.
+     *
+     * When the value is 0, the original color is used without modification.
+     * When the value is 1, the result color is used.
+     * When in the range (0, 1) the color is interpolated between the original and result by this amount.
+     *
+     * @member {number}
+     * @default 1
+     */
+    get interpolation()
+    {
+        return this.uniforms.uInterpolation;
+    }
+
+    set interpolation(value) // eslint-disable-line require-jsdoc
+    {
+        this.uniforms.uInterpolation = value;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/filters/colormatrix/ColorMatrixFilter.js
```diff
@@ -575,6 +577,26 @@ export default class ColorMatrixFilter extends core.Filter
...
+     * The interpolation value to use when mixing the original and resultant colors.
+     *
+     * When the value is 0, the original color is used without modification.
+     * When the value is 1, the result color is used.
+     * When in the range (0, 1) the color is interpolated between the original and result by this amount.
+     *
+     * @member {number}
+     * @default 1
+     */
+    get interpolation()
```
bigtimebuddy: What do you think about something less verbose than `interpolation`, such as `amount`? Alternatives: `alpha`, `blend`, etc.

englercj: Usually I use the term `alpha`, but was worried that would confuse people since it could also refer to opacity. `blend` also can be confused with `blendMode` so not sure about that one either.

`amount` might work?

bigtimebuddy: Other thought, what about "strength" it's used on some of the other filters.

amount is fine, just a little generic. I agree that alpha and blend are problematic.

bigtimebuddy: Thoughts @englercj?

englercj: I read `strength` as intensity. I think I like `interpolation` best of all  these choices so far :/

andrewstart: Isn't intensity basically what it represents though? The intensity of the effect that we are applying, from 0-100%?

englercj: Intensity of the effect and the interpolation you use are different values. For example, the `night` effect already has an intensity: `filter.night(intensity, multiply);`.

Setting intensity, then intensity again somewhere else would be weird. `intensity` usually acts as a multiplier increasing or decreasing how strong the affect is, and while interpolation in this case could be seen as similar it is a distinct value.

bigtimebuddy: I'm still hung-up on the verbosity of "interpolation". It's not really a term we use elsewhere in PIXI. And, in general, I love to avoid more technical terms like this. I think it makes the API more accessible. TBH, I think this is `alpha` or `opacity` and makes it easier to understand what changing that value would do. 

Does anyone else feel strongly about this?

englercj: I do not, feel free to change it to whatever you think is the most usable name.

ivanpopelyshev:
Technically, it is alpha. That's how result of advanced blendMode is formed.

PR #3945, opened 2017-04-17
fix(GraphicsRenderer): fix graphics render with fill and line when na…

begininend:
…tiveLines=true

### DIFF
--- a/src/core/graphics/webgl/GraphicsRenderer.js
+++ b/src/core/graphics/webgl/GraphicsRenderer.js
@@ -103,7 +103,7 @@ export default class GraphicsRenderer extends ObjectRenderer
 
             renderer.bindVao(webGLData.vao);
 
-            if (graphics.nativeLines)
+            if (webGLData.nativeLines)
             {
                 gl.drawArrays(gl.LINES, 0, webGLData.points.length / 6);
             }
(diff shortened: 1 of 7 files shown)


bigtimebuddy:
Could you explain more about why this is important and what it fixes? Also there seems to be a listing/build error. Could you look into that?

begininend:
```
const graphics = new PIXI.Graphics(true);
graphics.lineStyle(1, 0x000000)
              .beginFill(0xFFFFFF)
              .drawRect(0, 0, 10, 10)
              .endFill();
```
before change, it use gl.LINES render the fill and lines, so it display incorrect:
```
if (graphics.nativeLines)
{
    gl.drawArrays(gl.LINES, 0, webGLData.points.length / 6);
}
else
{
    webGLData.vao.draw(gl.TRIANGLE_STRIP, webGLData.indices.length);
}
```
I split the triangles and lines to different webGLData: 
```
if (webGLData.nativeLines)
{
    gl.drawArrays(gl.LINES, 0, webGLData.points.length / 6);
}
else
{
    webGLData.vao.draw(gl.TRIANGLE_STRIP, webGLData.indices.length);
}
```

englercj:
Is the Graphics `nativeLines` property ([this one](https://github.com/pixijs/pixi.js/blob/2af62817cf7a7b100322277a9dbdd55756057c44/src/core/graphics/Graphics.js#L58)) needed anymore after this change?

bigtimebuddy:
Here's a fiddle of the current bug, helpful to validate: 

https://jsfiddle.net/fcq2msqt/

bigtimebuddy:
Thanks @begininend for this addition!

PR #3956, opened 2017-04-19
Stops TickListener 'this.fn is not a function' error

themoonrat:
Fixes: https://github.com/pixijs/pixi.js/issues/3936

I believe the edge case may be because I have a ticker that does not have autoStart turned on.
Instead I manually call .update() on it.


bigtimebuddy:
Could you make this PR to master?

bigtimebuddy:
Is there a unit test for this edge case? I want to understand the problemmkre clearly. Generally, ticker listeners should always have a function.

PR #3959, opened 2017-04-19
Fixes #3954 support excluding deprecations

### LINKED ISSUE #3954: [BUG] The last PIXI dev can't support exclude some modules.
when I exclude  `interaction  deprecation loaders` , there are some errors.

In  the version built at  Apr 13, 2017 ,  #35a878 , That's OK.

I've reported  a similar issue https://github.com/pixijs/pixi.js/issues/3169  ,

Now the similar issue happens again , So I think PIXI need a test case for `exclude the  modules which not required`.

bigtimebuddy:
### Fixed

* Supports excluding deprecations from the building: `npm run dist — -e deprecation`

Addresses #3954

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,7 @@ export {
 };
 
 // Apply the deprecations
-if (typeof deprecation === "function")
+if (typeof deprecation === 'function')
 {
     deprecation(exports);
 }
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,10 @@ export {
 };
 
 // Apply the deprecations
-deprecation(exports);
+if (typeof deprecation === "function")
+{
+    deprecation(exports);
+}
 
 // Always export pixi globally.
 global.PIXI = exports; // eslint-disable-line


finscn:
Do you test it ? 

In my test case , this PR is wrong.

finscn:
In build file pixi.js , the code like this : 
```
if (typeof _deprecation2.default === 'function')
{
    _deprecation2.default(exports);
}
```

But when I exclude  `deprecation`,  there is no _deprecation2.

bigtimebuddy:
@finscn yes, I built with `npm run dist -- -e deprecation` and it ran fine. Is that how you are excluding deprecation?

finscn:
I use 
```
    npm run build -- -- --e interaction --e accessibility --e deprecation  --e polyfill
```

finscn:
Your test is wrong .
you miss a `--`
there need two `--`

finscn:
@bigtimebuddy , please see https://github.com/pixijs/pixi.js/issues/3663

bigtimebuddy:
Sorry, @finscn I ran that command and it ran correctly. 

`npm run dist -- -e deprecation`
is the same as
`npm run build -- -- -e deprecation`

finscn:
Do you test `npm run build -- -- -e deprecation` ?

bigtimebuddy:
Yes. Could you also reinstall your node_modules, then rebuild to see if that helps?

finscn:
@bigtimebuddy  , everything is OK  .
 I re-clone the code and reinstall node_modules. This PR could work correctly now.

Thanks.

bigtimebuddy:
No problem. Thanks for digging into this @finscn.

PR #3963, opened 2017-04-21
Make Interaction events closer to Browser behavior

### LINKED ISSUE #3867: PixiJS 4.5.0 mouse events wont trigger in Google Chrome using Wacom tablet
My Wacom Bamboo & Wacom Intuos Pro (up to date drivers) won't trigger mouseover / mouseout events in Chrome (v57) in the following example:

https://pixijs.github.io/examples/#/demos/interactivity.js

This interaction is working in Firefox and Safari.

The weird part is, when I use a mouse or touchpad on laptop the mouseover / mouseout the events do trigger.

Is there a way to fix this?

PixiJS 4.3.0 is working.

### LINKED ISSUE #3886: Bug with Touch in Chrome (android)

andrewstart:
Makes the event behavior of InteractionManager closer to that of standard browser behavior with:

* Expose event properties on `InteractionData` (originally from #3890)
* If TouchEvents are supported, use those instead of PointerEvents for touch based events - Fixes #3886
* Emit mouse events on pointer events from a pen device. - Fixes #3867
    * A browser would emit a fallback mouse event on those pointer events, which we were not doing.

Also does a little simplification/cleanup of how we were detecting pointer types and right clicks to determine what events to emit.

### DIFF
--- a/src/interaction/InteractionData.js
+++ b/src/interaction/InteractionData.js
@@ -43,6 +43,104 @@ export default class InteractionData
          * @member {number}
          */
         this.identifier = null;
+
+        /**
+         * Indicates whether or not the pointer device that created the event is the primary pointer.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/isPrimary
+         * @type {Boolean}
+         */
+        this.isPrimary = false;
+
+        /**
+         * Indicates which button was pressed on the mouse or pointer device to trigger the event.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button
+         * @type {number}
+         */
+        this.button = 0;
+
+        /**
+         * Indicates which buttons are pressed on the mouse or pointer device when the event is triggered.
+         * @see https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
It's not obvious to me, but are there any breaking changes here?

andrewstart:
I don't believe any change would be breaking - bullet points 1 & 3 are new additions. 2 changes what event is used for touch screens, but would match how it already is on iOS devices, so it also shouldn't be breaking.

stephenweaver:
Pretty sure this breaks a ton of stuff....like everything to do with touch and definitely everything to do with multi touch. I'll make a fiddle with before and after if you want. But I don't have time today.

englercj:
@stephenweaver Please open a new issue and link to this one if you are having an issue. Failing tests are always the best way to prove something is broken, though a fiddle displaying a bug is also very helpful.

stephenweaver:
https://github.com/pixijs/pixi.js/issues/4040

PR #3976, opened 2017-04-27
fix(Graphics): exclude holes in Graphics#containsPoint

begininend:


bigtimebuddy:
It might be nice to include a unit-test for this.

Issue #3939, opened 2017-04-16
modularizing removeItems?


mreinstein:
hey folks! Long time no see. :)

a while back we worked on an implementation of array splice that doesnt generate memory garbage. 2 years on and I have yet to find as high a quality version as the one we worked on. I want to use this in several projects, so I've modularized it here, in this npm module:

https://github.com/mreinstein/remove-array-items

Are y'all open to a PR to include this module? Happy to send one along. Just let me know what branch you'd like it in, etc.

thanks!

-Mike

englercj:
I don't have many feelings about this, looks like it is the same code we are currently using. Curious what we buy by moving to use the module rather than the one we have in the lib already?

mreinstein:
> Curious what we buy by moving to use the 
> module rather than the one we have in the 
> lib already

Not much of any substance, although I've used this snippet in at least 5 different projects/modules. Given there is little to no cost in modularizing this it seems like a good opportunity for re-use.

englercj:
As long as we can maintain the current level of documentation we have for that method I'm all for moving to a module.

Did you move our tests there too when you made the repo?

mreinstein:
> Did you move our tests there too 
> when you made the repo?

Yep its in there and wired up to Travis

PR #3977, opened 2017-04-27
modularized removeItems #3939

mreinstein:


bigtimebuddy:
@mreinstein thanks for the PR.

Looks like some tests failed. Could you please [check those out](https://travis-ci.org/pixijs/pixi.js/builds/226548790)? You can run `npm run test:debug` locally to debug.

mreinstein:
@bigtimebuddy I tried running both `npm test` and `npm run test:debug` locally. They are both complaining:

```bash
> pixi.js@4.5.1 test /Users/michaelreinstein/wwwroot/pixi.js
> floss --path test/index.js

[floss] caught outer exception: Cannot find module '../dist/pixi' Error: Cannot find module '../dist/pixi'
    at Module._resolveFilename (module.js:470:15)
    at Function.Module._resolveFilename (/Users/michaelreinstein/wwwroot/pixi.js/node_modules/electron/dist/Electron.app/Contents/Resources/electron.asar/common/reset-search-paths.js:35:12)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/michaelreinstein/wwwroot/pixi.js/test/index.js:4:1)
    at Object.<anonymous> (/Users/michaelreinstein/wwwroot/pixi.js/test/index.js:21:3)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
```

looks unrelated, like some kind of packaging step has failed/not run.

mreinstein:
opened a separate issue for `npm test` failure  https://github.com/pixijs/pixi.js/issues/3978

Given that the travis build succeeded, I think this PR should be ok (at least test wise.)

package.json
```diff
@@ -62,6 +62,7 @@
     "ismobilejs": "^0.4.0",
     "object-assign": "^4.0.1",
     "pixi-gl-core": "^1.0.3",
+    "remove-array-items": "^0.0.1",
```
bigtimebuddy: Would you mind bumping **remove-array-items** to version 1.0.0 and publishing? This would be important for a couple of reasons:
* this utility is likely to change in the near future and is stable enough
* using full semver gives us more information about what changes into the future
* we're not a fan of pre-v1 versioning
* the semantic constraint `^0.0.1` doesn't resolve to anything other than `0.0.1` (for instance, 0.0.2 does not satisfy constraint ^0.0.1, [check here](http://jubianchi.github.io/semver-check/)). So we would never be able to pick up future changes

englercj:
Like I mentioned in #3939 I don't have strong feelings about this. I don't think this particular instance is important, but the more general precedent it sets might be worth discussing. Generally moving out stuff we don't want to maintain, or doesn't make sense to version with pixi is a good idea (imo), but this is so small I don't think either way matters that much.

mreinstein:
> I don't think this particular instance is important, but the more general precedent 
> it sets might be worth discussing

I agree this particular instance is not that important. I only pulled this out because every project I work on which involves `real time stuff` + `javascript` +` array item removal` ends up needing this particular capability. Many javascript operations spew garbage, and implementations have to be chosen carefully to avoid the wrath of the garbage collector. I'm already using this in 3 other modules. It would be extra niceness if pixi were able to take advantage of this re-use.


> the more general precedent it sets might be worth discussing

That sounds interesting. Care to elaborate?


> What exactly does Pixi gain other than an additional dependency

Assuming modules are chosen that are high quality, you stand to gain in all the typical areas that re-use brings.

Consider taking off the pixi maintainer's hat and look at this from an implementor's perspective. pixi is a super awesome and important module. It's part of an ecosystem. Any real world usage is going to involve several (maybe even dozens or hundreds) of modules. 

> Unless that functionality falls outside of the mission of the library or it's a pain to maintain

Yes I agree with this completely. Great modules tend to be ones with specific purposes. if I had to summarize the mission of pixi concisely I'd say "a cross platform rendering and input handling library for javascript". Anything that doesn't fit this mission could arguably (I guess that's what we're doing) be externalized.

Issue #3972, opened 2017-04-25
Canvas Renderer does not respect texture's frame `x` and `y` when tinted


nathantalewis:
Description
---
When using the canvas renderer with a tinted `Sprite`, if you assign a new `Rectangle` to its texture's frame then the new values for `width` and `height` are drawn as expected but the new values for `x` and `y` seem to be ignored.

To demonstrate the problem, I merged the spritesheet and tinting demos together.
**Minimal example of problem:** https://jsfiddle.net/5u8sqzwc/1/
**Working behavior in canvas without tinting:** https://jsfiddle.net/0wv49ack/
**Working behavior in WebGL with tinting:** https://jsfiddle.net/3a6rotk4/

PIXI.js version: 4.5.1 (also seen on 4.4.1)
Seen on both Chrome and Firefox, on Window 10 and on Mac.

Background
---
What I really want is to emulate mipmap for the canvas renderer. To do this, I took an existing spritesheet and appended copies of it scaled down just like the GPU does for the WebGL renderer. I used the same `.json` file for the spritesheet so that the loader knows nothing about the additional pixels off to the right.

When I change the scale of my stage to zoom out, I calculate the new `x`, `y`, `width` and `height` for each texture's frame based on where it would be in the mipmap area. This works well for any textures that aren't used by tinted sprites. However, if the sprites have a non-white tint, the canvas renderer seems to only take into account the new `width` and `height`.

GoodBoyDigital:
thanks for sharing @nathantalewis , will take a look and get back to ya 👍

bigtimebuddy:
Merged fix! Closing.

nathantalewis:
Thank you guys!

PR #3985, opened 2017-05-02
fix for #3972

GoodBoyDigital:
Fix for #3972. Modifies the cached tinted frame of a texture in canvas.

### DIFF
--- a/src/core/sprites/canvas/CanvasTinter.js
+++ b/src/core/sprites/canvas/CanvasTinter.js
@@ -120,8 +120,6 @@ const CanvasTinter = {
             crop.width,
             crop.height
         );
-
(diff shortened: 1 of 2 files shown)


src/core/sprites/canvas/CanvasTinter.js
```diff
@@ -109,6 +120,8 @@ const CanvasTinter = {
             crop.width,
             crop.height
         );
+
+        document.body.appendChild(canvas);
```
bigtimebuddy: Is this correct? The canvas is added to the body?

GoodBoyDigital: er.. yes?

PR #3984, opened 2017-05-01
Test script

bigtimebuddy:
Addresses #3978

### Changed

* `npm test` now including building so it runs out of the box
* `npm run unit-test` just runs the tests (previously `npm test`)
* `npm run unit-test:debug` runs the tests in the debug mode (previously `npm run test:debug`)

PR #3993, opened 2017-05-03
Take into account baseTexture resolution when registering a BitmapText.

artsjedi:
When the base texture has a resolution scale other than 1, then the result of the rendered bitmap text becomes a mess. This change allows you to use bitmap fonts with different screen resolutions.

### DIFF
--- a/src/extras/BitmapText.js
+++ b/src/extras/BitmapText.js
@@ -469,10 +469,11 @@ export default class BitmapText extends core.Container
         const data = {};
         const info = xml.getElementsByTagName('info')[0];
         const common = xml.getElementsByTagName('common')[0];
+		const res = texture.baseTexture.resolution || 1;
 
         data.font = info.getAttribute('face');
         data.size = parseInt(info.getAttribute('size'), 10);
-        data.lineHeight = parseInt(common.getAttribute('lineHeight'), 10);
+        data.lineHeight = parseInt(common.getAttribute('lineHeight'), 10) / res;
         data.chars = {};
 
         // parse letters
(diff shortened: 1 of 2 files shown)


src/extras/BitmapText.js
```diff
@@ -469,10 +469,11 @@ export default class BitmapText extends core.Container
         const data = {};
         const info = xml.getElementsByTagName('info')[0];
         const common = xml.getElementsByTagName('common')[0];
+		const res = texture.baseTexture.resolution || 1;
```
bigtimebuddy: Small thing, could you change tabs to spaces here. 

Also, maybe this should default to `PIXI.settings.RESOLUTION` instead of `1`.... Like this: https://github.com/pixijs/pixi.js/blob/dev/src/core/textures/BaseTexture.js#L38

artsjedi:
Thanks, I hope it's correct.

src/extras/BitmapText.js
```diff
@@ -484,16 +486,16 @@ export default class BitmapText extends core.Container
             const charCode = parseInt(letter.getAttribute('id'), 10);
 
             const textureRect = new core.Rectangle(
-                parseInt(letter.getAttribute('x'), 10) + texture.frame.x,
-                parseInt(letter.getAttribute('y'), 10) + texture.frame.y,
-                parseInt(letter.getAttribute('width'), 10),
-                parseInt(letter.getAttribute('height'), 10)
+                parseInt(letter.getAttribute('x'), 10) / res + texture.frame.x / res,
```
themoonrat: Could we get some brackets around the last division on this line and the next to make the bodmas ordering explicit. I thought we had an eslint rule that enforced this (http://eslint.org/docs/rules/no-mixed-operators)

englercj: We do (https://github.com/pixijs/pixi.js/blob/dev/.eslintrc.json#L184) not sure why it didn't error here...

bigtimebuddy: This is my fault, Travis wasn't running linting since #3984. PR fixes here #3998

src/extras/BitmapText.js
```diff
@@ -484,16 +486,16 @@ export default class BitmapText extends core.Container
             const charCode = parseInt(letter.getAttribute('id'), 10);
 
             const textureRect = new core.Rectangle(
-                parseInt(letter.getAttribute('x'), 10) + texture.frame.x,
-                parseInt(letter.getAttribute('y'), 10) + texture.frame.y,
-                parseInt(letter.getAttribute('width'), 10),
-                parseInt(letter.getAttribute('height'), 10)
+                parseInt(letter.getAttribute('x'), 10) / res + (texture.frame.x / res),
+                parseInt(letter.getAttribute('y'), 10) / res + (texture.frame.y / res),
```
englercj: This is still going to cause a lint error now that @bigtimebuddy fixed the linter not running on travis. It should be `(x / res) + (frame.x / res)`. No mixed operators.

bigtimebuddy:
@themoonrat, look good?

PR #4000, opened 2017-05-06
Add `updateTexture` & some shortcuts in Sprite; Cosmetic

finscn:
* Add `updateText`.  If a sprite is created by a canvas ,  when the canvas's changed , user could call `updateText` for updating sprite.
* Add some shortcuts for accessing  some properties of sprite._texture.
* Cosmetic Text. Just define some local vars for accessing properties of `this` , for DRY.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -404,6 +404,19 @@ export default class Sprite extends Container
     }
 
     /**
+     * Update texture if the content of texture has changed.
+     * e.g. If texture is based on a canvas , and after the canvas changed, please call this method.
+     */
+    updateTexture()
+    {
+        const baseTexture = this._texture.baseTexture;
+
+        this._onTextureUpdate();
+
+        baseTexture.emit('update', baseTexture);
+    }
+
+    /**
      * Destroys this sprite and optionally its texture and children
      *
      * @param {object|boolean} [options] - Options parameter. A boolean will act as if all options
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Agree on "this" in Text, but updateTexture.. hm. Its handy for single sprites that are created from canvases, but I dont know if its good to put it there.

@GoodBoyDigital @bigtimebuddy what about "SpriteCanvas" class, that is created from canvas and not from texture? Or with canvas. Move part of Text there, add this updateTexture method.

finscn:
How about these `shortcuts` ?

finscn:
And If use SpriteCanvas ,  do we need  TextureCanvas & BaseTextureCanvas too ???

ivanpopelyshev:
@finscn definitely no. Only for sprite.

bigtimebuddy:
I like the local variable caching, that should help minifying output. Small, but good change.

I disagree with adding `baseTexture`, `orig`, `frame`, and `trim` getters for Sprite. It's bloats the API and doesn't provide a lot of convenience. These things are already publicly available `sprite.texture.baseTexture`, `sprite.texture.orig`, etc.

Also, I don't understand the need to add `updateTexture`. This functionality is also already available: 
`sprite.texture.baseTexture.update()` (example: https://jsfiddle.net/bedrdc1b/).

So in summary, I like the local variable caching, but disagree with the API changes.

bigtimebuddy:
ping @ivanpopelyshev

PR #4003, opened 2017-05-08
Minor fixes

Makio64:


englercj:
I appreciate the effort to fix up this branch, but I currently have a mass of local changes that move around all this stuff so I would hold off on going any further with this.

Thanks though!

Issue #3899, opened 2017-03-31
Apply Tint to Particle Container


ghost:
Why do particle containers not allow you to tint the entire thing, like you can do to a sprite?

How would I go about efficiently tinting an entire particle container? (It needs to be re-tinted, since it changes colour.)

englercj:
No particular reason we *cant* do it. Just a matter of adding another uniform where we upload the tint value of the `ParticleContainer`. We could also add support for uploading the tint value of each sprite in the attributes, but it does cost a bit more memory to do that.

Does having a tint on `ParticleContainer` only that is applied to all sprites seem useful?

thamaranth:
Is this still available?

englercj:
@thamaranth Sure, feel free to put in a PR for this if you are interested in contributing.

bigtimebuddy:
Closing, fixed with #4014

PR #4014, opened 2017-05-12
Apply Tint to Particle Container (#3899)

YahorPadlucki:
Applies tint to whole particle container.

### DIFF
--- a/src/particles/webgl/ParticleRenderer.js
+++ b/src/particles/webgl/ParticleRenderer.js
@@ -1,5 +1,4 @@
 import * as core from '../../core';
-import { hex2rgb } from '../../core/utils';
 import ParticleShader from './ParticleShader';
 import ParticleBuffer from './ParticleBuffer';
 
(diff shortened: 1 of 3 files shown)


src/particles/webgl/ParticleRenderer.js
```diff
@@ -153,6 +154,7 @@ export default class ParticleRenderer extends core.ObjectRenderer
 
         this.shader.uniforms.projectionMatrix = m.toArray(true);
         this.shader.uniforms.uAlpha = container.worldAlpha;
+        this.shader.uniforms.tint = hex2rgb(container.tint);
```
bigtimebuddy: `hex2rgb` creates a new Array every `render` call here, which isn't good performance-wise. Maybe consider adding a private `_tintRGB` property on the ParticleContainer when `tint` is set? 

```js
set tint(value)
{
     this._tint = value;
     hex2rgb(value, this._tintRGB);
}
```
And where this is added to the constructor before the initial `this.tint = 0xFFFFFF`.
```js
this._tintRGB = []; 
```

Then this line could be `this.shader.uniforms.int = container._tintRGB`

YahorPadlucki: Thanks for the explanation - I've added _tintRGB property

themoonrat:
A comment....
All current 'tint' properties are on classes that contain and can apply to a texture, and thus works on both canvas and webgl.
This same 'tint' property occurs on a container, and thus on is webgl only.

So I think the documentation for the property should be clear on this.

bigtimebuddy:
Thanks for the doc changes @YahorPadlucki. Look okay to you @themoonrat?

PR #4016, opened 2017-05-13
Improvements to where and how gradients can be applied to Text

themoonrat:
Allows the stroke to be a gradient as well as fill - http://stackoverflow.com/questions/43897689/how-to-apply-gradient-stroke-in-pixi-js
New gradient style, LINEAR_VERTICAL_NOREPEAT, which doesn't try to repeat the gradient on multiline to text to make each line match in style - http://stackoverflow.com/questions/43897341/how-to-apply-different-gradients-to-same-text-in-pixi-js

Demo: https://jsfiddle.net/themoonrat/d3roaLmf/ (clear I'm not artist with those chosen colours... )

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -302,12 +302,14 @@ export const TRANSFORM_MODE = {
  * @name TEXT_GRADIENT
  * @memberof PIXI
  * @type {object}
- * @property {number} LINEAR_VERTICAL Vertical gradient
- * @property {number} LINEAR_HORIZONTAL Linear gradient
+ * @property {number} LINEAR_VERTICAL Vertical gradient; maintains same gradient look over each line
+ * @property {number} LINEAR_HORIZONTAL Horizontal gradient
+ * @property {number} LINEAR_VERTICAL_NOREPEAT Vertical gradient; gradient applies over all lines; each will look different
  */
 export const TEXT_GRADIENT = {
     LINEAR_VERTICAL: 0,
     LINEAR_HORIZONTAL: 1,
+    LINEAR_VERTICAL_NOREPEAT: 2,
 };
 
 /**
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -151,8 +151,22 @@ export default class Text extends Sprite
 
         context.clearRect(0, 0, this.canvas.width, this.canvas.height);
 
+        const fillStyle = this._generateColorStyle(
+            lines,
+            style.fill,
+            style.fillGradientType,
+            style.fillGradientStops
+        );
+
+        const strokeStyle = this._generateColorStyle(
+            lines,
+            style.stroke,
+            style.strokeGradientType,
+            style.strokeGradientStops
+        );
+
         context.font = this._font;
-        context.strokeStyle = style.stroke;
+        context.strokeStyle = strokeStyle;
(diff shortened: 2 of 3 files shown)


bigtimebuddy:
Love this addition, really nice work! And thanks for the demo, that's super useful to see. As a side note, I will probably implement stroke gradients on Text in PixiAnimate after this is merged. 

I've been wanting to reduce the complexity of the Text class for some time. It's become too monolithic. I think TextMetrics helped a lot to simplify and pull out static methods. However, right now, defining gradients feels a little after-thought and might be worth thinking about a way to refactor.

There are a few fill properties (colors, stops, type), and now two additional properties for stroke. Is there a way we can think more abstractly about a Text's stroke and fill style? Potentially as a new end-user class? We could remove all six properties on TextStyle and use  `fill` and `stroke` to be either `string` or `PIXI.TextGradient` (or TextFill, TextColorStyle, whatever). This seems more scalable and easier to conveniently expose more of the CanvasGradient APIs. Something like this:

```js
const style = new PIXI.TextStyle({
    // fill could be a string, Array or Object, like this:
    fill: {
       colors: [0x0, 0x999999, 0xFF99CC]
       type: PIXI.TEXT_GRADIENT. LINEAR_VERTICAL_NOREPEAT,
       stops: [0, 0.5, 1]
    }
});
const text = new PIXI.Text('', style);
```

What do you think? Refactor doesn't need to be this PR, but I'd like to consider it as we'd be adding new APIs. At the very least, `_generateColorStyle` might be a good candidate to pull out of Text (probably suggesting continuing to keep it a private method until we agree on a better API for gradients). Since it's mostly a utility method for generating a CanvasGradient.

themoonrat:
Definitely agree on pulling out _generateColorStyle into a more generic 'generate a gradient' util function. 

Regarding the text style... when I originally did the gradient, I wanted it to be simple to use, and what could be simpler than an array of colours rather than a string. I totally get where you're coming from though. Is it necessarily awful if TextStyle has a lot of options.... when there just are so many ways people want to use and display it? Is it nicer to have a flat list of more options than some of those being broken down into more complex settings objects? Or is it better to have fewer options that developers can delve into when they need to? I'm not sure I have a strong opinion either way, annoyingly.

bigtimebuddy:
Right now I just feel like the gradient property names, like "strokeGradientType" is pretty verbose and easy to mess up, e.g. "strokeTypeGradient." I'm in favor of providing simpler ways to make more complex fills with utility methods and more complex objects setting to TextStyle. The Text class should know nothing about composing gradients. Unifying fill and stroke, from a dev point of view, I believe is a win. Having 6 properties isn't really that user-friendly. And adding more complexity, like gradient angle, starts to be a slippery slope of more TextStyle props.

themoonrat:
I'm with ya; will see what I can come up with.

bigtimebuddy:
@themoonrat Sounds like there's more work needed here. I'm going to close until you're ready to be reviewed again.

vikkilea:
@bigtimebuddy This branch has been deleted now; are there any plans for refactors to the gradient system?

Thanks

bigtimebuddy:
Potentially for v5, we are working on gutting the Text system. It needs a big overhaul. Unlikely to happen for v4.

PR #4033, opened 2017-05-16
premultiply blendModes, critical for compressed textures.

ivanpopelyshev:
We have "premultipliedAlpha" field in BaseTexture from the beginning of times. But we do not respect it. It was ok, but some time ago both me and @bigtimebuddy started to use compressed textures.

There are no compressed formats with premultiplied alpha in WebGL, DXT1,3,5 are supported, while DXT2,4 aren't.

Example: 

Current version of pixi: https://pixijs.github.io/examples/#/textures/not-premultiplied-alpha.js
This branch: https://pixijs.github.io/examples/?v=dev-blendmodes-premultiply#/textures/not-premultiplied-alpha.js

Please read 
https://www.cgdirector.com/straight-alpha-vs-premultiplied-alpha/ 
http://www.adriancourreges.com/blog/2017/05/09/beware-of-transparent-pixels/
to understand the problem.

Later, I hope we'll be able to add compressed textures as childs of BaseTexture.

### DIFF
--- a/src/core/sprites/webgl/SpriteRenderer.js
+++ b/src/core/sprites/webgl/SpriteRenderer.js
@@ -5,6 +5,7 @@ import generateMultiTextureShader from './generateMultiTextureShader';
 import checkMaxIfStatmentsInShader from '../../renderers/webgl/utils/checkMaxIfStatmentsInShader';
 import Buffer from './BatchBuffer';
 import settings from '../../settings';
+import { correctBlendMode, premultiplyTint } from '../../utils';
 import glCore from 'pixi-gl-core';
 import bitTwiddle from 'bit-twiddle';
 
@@ -226,7 +227,7 @@ export default class SpriteRenderer extends ObjectRenderer
         let currentGroup = groups[0];
         let vertexData;
         let uvs;
-        let blendMode = sprites[0].blendMode;
+        let blendMode = correctBlendMode(sprites[0].blendMode, sprites[0]._texture.baseTexture.premultipliedAlpha);
 
         currentGroup.textureCount = 0;
         currentGroup.start = 0;
@@ -251,10 +252,12 @@ export default class SpriteRenderer extends ObjectRenderer
 
             nextTexture = sprite._texture.baseTexture;
 
-            if (blendMode !== sprite.blendMode)
+            const spriteBlendMode = correctBlendMode(blendMode, nextTexture);
+
+            if (blendMode !== spriteBlendMode)
             {
                 // finish a group..
-                blendMode = sprite.blendMode;
+                blendMode = spriteBlendMode;
 
                 // force the batch to break!
                 currentTexture = null;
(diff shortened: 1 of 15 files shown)


ivanpopelyshev:
@GoodBoyDigital Yes, we need blendFuncSeparate, just blendFunc is not sufficient for NORMAL with not-premultiplied stuff. We also might need it for other tricks not related to this issue, like erasing blendmode.

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -226,7 +227,8 @@ export default class SpriteRenderer extends ObjectRenderer
         let currentGroup = groups[0];
         let vertexData;
         let uvs;
-        let blendMode = sprites[0].blendMode;
+        let blendMode = premultiplyBlendMode[
+            Number(sprites[0]._texture.baseTexture.premultipliedAlpha)][sprites[0].blendMode];
```
englercj: Rather than `Number(x)` I would do a ternary since `Number(undefined)` is `NaN`.

```js
let blendModeIndex = sprites[0]._texture.baseTexture.premultipliedAlpha ? 1 : 0;
let blendMode = premultiplyBlendMode[blendModeIndex][sprites[0].blendMode];
```

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -251,10 +253,12 @@ export default class SpriteRenderer extends ObjectRenderer
 
             nextTexture = sprite._texture.baseTexture;
 
-            if (blendMode !== sprite.blendMode)
+            const spriteBlendMode = premultiplyBlendMode[Number(nextTexture.premultipliedAlpha)][sprite.blendMode];
```
englercj: Same note as above.

src/core/utils/index.js
```diff
@@ -397,3 +398,107 @@ export function clearTextureCache()
...
+ *
+ * @memberof PIXI.utils
+ * @function correctBlendMode
+ * @param {number} blendMode supposed blend mode
+ * @param {boolean} premultiplied  whether source is premultiplied
+ * @returns {number} true blend mode for this texture
+ */
+export function correctBlendMode(blendMode, premultiplied)
+{
+    return premultiplyBlendMode[Number(premultiplied)][blendMode];
```
englercj: Same note as above about using a ternary instead of Number ctor.

src/core/sprites/webgl/SpriteRenderer.js
englercj: After I saw `correctBlendMode` I'm curious why this isn't using that function but is instead duplicating the same code?

ivanpopelyshev:
Agree about ternary. About separate blendModes - i dont want to add extra variables into the state because of only 3 extra blendmodes. Do you agree to add "setPremultiplyBlendFlag" or something like that into renderer?

englercj:
I think it is fine as is until I can refactor it in v5.

ivanpopelyshev:
Ok, so im changing it into classes and hiding NPM stuff, like you said :)

ivanpopelyshev:
@englercj Moving it into the object is a big problem. It'll be big PR, so I ask you to reconsider that and just accept this one first. I know there will be big discussion over "making BlendMode object"

englercj:
Reconsider what? I already said it was fine, then you added a bunch of changes. I'm letting you settle on an implementation before I review again.

ivanpopelyshev:
@englercj just tell me what to change in this one to get it passed ;)

englercj:
> just tell me what to change in this one to get it passed ;)

The things I mentioned above in my review should be changed.

src/core/sprites/webgl/SpriteRenderer.js
ivanpopelyshev: Because @GoodBoyDigital wants less functions calls.

ivanpopelyshev:
Ok, added ternary, found a bug in mesh. Is that enough?

PR #4044, opened 2017-05-19
Fix: Accessibility div not positioned correctly

saadataz:
**Problem**
HTML accessibility layer is not positioning the pixijs buttons correctly on a webpage with other positioned elements than just Pixijs canvas.
**Cause**
getBoundingClientRect() is used to get the coordinates of buttons relative to the viewport whereas the accessibility layer HTML elements are absolutely positioned i.e. relative to the closest positioned element. 
**Proposed Solution**
- Accessibility HTML css position should be `fixed`


saadataz:
@bigtimebuddy need any help on this?

englercj:
@GoodBoyDigital is the original author, and a user of this feature. I'd like to get his review before we merge.

That being said, it makes sense to me!

saadataz:
Closing this because of no reviews

PR #4059, opened 2017-05-24
use correct shader

finscn:


ivanpopelyshev:
Do you have a case when that does not work? Texture can be changed, trimming will be changed too. The idea is that both shaders have the same input and thus same location numbers

finscn:
"Do you have a case when that does not work?"  :  No. If no this PR, it could work too. But in logic , it's not perfect.
So , this pr is not for fixing bugs, just  cosmetic.

finscn:
By the way , the mesh in dev branch has been changed (about tint) , This branch also need to update.

ivanpopelyshev:
Lets not put this shader into glData then.

finscn:
> Lets not put this shader into glData then.

I can't understand this sentence. Could you tell me more ?

finscn:
@ivanpopelyshev , I've remove the `shader` in glData

PR #4064, opened 2017-05-28
Fix for repeated letters on wordWrapped bitmap text

themoonrat:
Record how many spaces have already been removed when resetting the index for a new line of text after hitting the maxWidth limit

Fixes https://github.com/pixijs/pixi.js/issues/4063


themoonrat:
Fiddle with the fix applied: https://jsfiddle.net/vvg5wntk/1/

PR #4067, opened 2017-05-29
deprecation.js now works with modules excluded at build time

themoonrat:
Fixes: https://github.com/pixijs/pixi.js/issues/3710

### DIFF
--- a/src/deprecation.js
+++ b/src/deprecation.js
@@ -564,24 +564,27 @@ export default function deprecation(core)
         },
     });
 
-    Object.defineProperties(extras, {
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Looking for some verification steps here.

All these should build and run without any problems, correct?

```bash
npm run dist -- -e prepare
npm run dist -- -e extras
npm run dist -- -e interaction
```

themoonrat:
I tested it removing everything apart from core. 
Build worked fine, and came down to 700kb ish :)

PR #4069, opened 2017-05-29
Text drop shadow is now the right colour, even with blur enabled

themoonrat:
Fixes: https://github.com/pixijs/pixi.js/issues/3960

Don't see why it was set to ignore the shadow colour before when blur was enabled... that behaviour goes back to v3 days. But it feels like incorrect behaviour to me which this PR fixes


themoonrat:
fiddle to demonstrate the fix: https://jsfiddle.net/themoonrat/zmt1ojph/

PR #4070, opened 2017-05-30
pixi.text Styles generator ?

djmisterjon:
hi !, is there a pixi.text generator for the style.
Because I find it very long to try different configuration styles without preview.

similar as pixi-particles but for the text style ?

http://pixijs.github.io/pixi-particles-editor/
If that does not exist, it would be a good request.
thank a lot for help


bigtimebuddy:
@djmisterjon Oops, looks like you made a pull-request and this should be an issue. I'm going to close, feel free to re-open as an issue. 

Good suggestion, by the way for an interactive Text style generator.

PR #4080, opened 2017-06-02
Adds an onLoop callback for AnimatedSprites

themoonrat:
As requested by https://github.com/pixijs/pixi.js/issues/4072


themoonrat:
Fiddle for when the continuous build has finished
https://jsfiddle.net/themoonrat/0nwg5jwh/

PR #4099, opened 2017-06-12
Remove release branch pushes upon publish

bigtimebuddy:
This PR removes the script which manually syncs the repo to the **release** branch. This is no longer needed since Travis now takes care of deploying the latest tag release and Bower has been officially deprecated.


englercj:
The only thing we lose here is that the `/release` build URLs no longer will work. Meaning there isn't a good way to have a URL that is the "latest release".

bigtimebuddy:
@englercj actually that will still work because of this: https://github.com/pixijs/pixi.js/blob/dev/.travis.yml#L112-L151

so https://pixijs.download/release/pixi.js should always be the latest release.

englercj:
Oh shit, I missed that; good call. 🚢

PR #4119, opened 2017-06-22
Add option for dual-graphics devices

ivanpopelyshev:
According to https://github.com/KhronosGroup/WebGL/issues/2343 , we can use it , because we implemented context loss / context restore. I still can't find original doc, where is that thing specified.

### DIFF
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -50,7 +50,7 @@ export default class Application
      * @param {boolean} [options.legacy=false] - `true` to ensure compatibility with older / less advanced devices.
      *  If you experience unexplained flickering try setting this to true. **webgl only**
      * @param {string} [options.powerPreference] - Parameter passed to webgl context, set to "high-performance"
-     *  for devices with dual graphics card
+     *  for devices with dual graphics card **webgl only**
      * @param {boolean} [options.sharedTicker=false] - `true` to use PIXI.ticker.shared, `false` to create new ticker.
      * @param {boolean} [options.sharedLoader=false] - `true` to use PIXI.loaders.shared, `false` to create new Loader.
      */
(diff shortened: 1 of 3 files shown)


bQvle:
Nice if it works. we have huge issues with laptops choosing integrated over nvidia (causing bad performance in our game). Ill test it asap and let you know if it works.

ivanpopelyshev:
https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/sLFmnalQHSc

> Chrome will start implementing WebGL's powerPreference context creation attribute soon, but it'll only be supported on macOS at first. There are difficulties implementing it on Windows.

> For more details I recommend emailing the webgl-dev-list Google group: https://groups.google.com/d/forum/webgl-dev-list .

ivanpopelyshev:
Done. We still dont know whether it works for anyone at all. I have no such device :(

src/core/Application.js
```diff
@@ -46,6 +46,8 @@ export default class Application
      *  FXAA is faster, but may not always look as great **webgl only**
      * @param {boolean} [options.legacy=false] - `true` to ensure compatibility with older / less advanced devices.
      *  If you experience unexplained flickering try setting this to true. **webgl only**
+     * @param {string} [options.powerPreference] - Parameter passed to webgl context, set to "high-performance"
```
bigtimebuddy: For consistency could you add `**webgl only**` like the other comments here?

ivanpopelyshev: done.

ivanpopelyshev:
@bigtimebuddy We need test on bunnymark, can you add "high-performance" and this branch in it? That way we can compare 4.5.3 and this thing on different devices.

bigtimebuddy:
@ivanpopelyshev I made a tweak to bunny-mark to support renderer options (including this new one): https://pixijs.github.io/bunny-mark/

If you check `powerPreference` and select **dev-power-preference** branch, you should be able to test.

bigtimebuddy:
@ivanpopelyshev could you resolve that small doc conflict with dev and I'll merge this.

bigtimebuddy:
Ping @ivanpopelyshev

PR #4130, opened 2017-06-29
fix ColorMatrixFilter.prototype.contrast

wxyu247:


bigtimebuddy:
Thanks for the fix @wxyu247!

Issue #4062, opened 2017-05-27
question about the src/core/display/Transform* modules


mreinstein:
I'm curious about `Transform` vs `TransformStatic`. Is it safe to say that these 2 implementations are completely identical _except_ for the fact that `TransformStatic` caches computations?

englercj:
Yup!

mreinstein:
@englercj thanks for clarification! 

I think this could be simplified substantially: We could take the `TransformStatic` class, merge it with the base class `TransformBase`, and add a parameter in the constructor (e.g., `cache`) which would default to true (the current default behavior in pixi.) Then there would be one class named `Transform`, and that could be used everywhere. Would remove the bits of code duplication across the 3 classes, reduce the number of classes, and generally be more readable.

Thoughts? Open to a PR for this?

mreinstein:
another possibility, if there is concern over changing the exposed transform APIs:  Move all of the `TransformStatic` code into `TransformBase`, and then the 2 extended classes could by very thin wrappers for the base class:

```javascript
export default class TransformStatic extends TransformBase {
    constructor() {
        super({ cache: true });
   }
}
```

```javascript
export default class Transform extends TransformBase {
    constructor() {
        super({ cache: false });
    }
}
```

This is not as clean as having 1 simple `Transform` class used everywhere, but it does preserve the existing API which obv has other benefits. :)

englercj:
> I think this could be simplified substantially

Yup! I'm redoing this paradigm completely in v5.

mreinstein:
Ah ok. I'll stay out of the way then. :) Thanks for the update!

mreinstein:
@englercj took a stab at the more conservative of the 2 above approaches (not breaking the existing API) Thought maybe this would be useful in the meantime until you do a more drastic refactor.

ivanpopelyshev:
Just a note for future me: pixi-spine will stop working after that, because it uses TransformBase a bit, so I'll have to update it. 

We need a way to stop automatic updates (by position,scale,...), suppose localTransform is changed from somewhere else. Is there an easy solution?

mreinstein:
@englercj I'm happy to close #4134 if you would prefer to take another approach. I would take no offense. :) I just felt like seeing if I could throw this together semi-sanely. If it happens to be useful that's great, but I don't want to clog up a competing solution you might have in mind.

PR #4134, opened 2017-06-30
simplify transform classes. #4062

mreinstein:


bigtimebuddy:
@englercj haven't you already worked on this convergence for v5? I think I recall a conversation awhile back about it.

ivanpopelyshev:
I'll review it a bit later. I suggest we REMOVE non-static transform completely.

mreinstein:
thanks @ivanpopelyshev!

I intentionally chose the more conservative refactor here. I was worried that breaking the API would make this PR less appealing. The current strategy manages to cut the total lines of code in the transform classes in half, without breaking any existing behavior.

englercj:
I have changed a bit of this in v5, there is only 1 transform class and it always caches by exposing properties for each value rather than observable points: https://github.com/pixijs/pixi.js/blob/455c059e8d155c1d9a05fc2ece2c02b3bdf8bf84/plugins/transform/src/Transform.js

I'm not done with it yet since I've only spent a little time with it.

englercj:
Going to close this for now, since we are unifying this a bit differently in v5.

PR #4148, opened 2017-07-06
baseRenderTexture is optional

trusktr:


trusktr:
Hmm, maybe that's not good enough doc. It needs arg1...4

bigtimebuddy:
The RenderTexture constructor signature (width, height, scaleMode, resolution) should only be accessible with `RenderTexture.create`. Therefore `baseRenderTexture` is optional, but only internally. This should not be part of the public API.

bigtimebuddy:
Going to close this since optional `baseRenderTexture` param is not a public API.

PR #4173, opened 2017-07-23
_glRenderTargets now destroyed when RenderTexture destroyed

GoodBoyDigital:


bigtimebuddy:
This is failing unit-tests locally @GoodBoyDigital . Changing to this seemed to work...

```js
if (texture._glRenderTargets && texture._glRenderTargets[this.renderer.CONTEXT_UID]) {
      texture._glRenderTargets[this.renderer.CONTEXT_UID].destroy();
      delete texture._glRenderTargets[this.renderer.CONTEXT_UID];
}
```

bigtimebuddy:
I think we need the check since `_glRenderTargets` only exists on BaseRenderTexture, not BaseTexture.

GoodBoyDigital:
Nice catch @bigtimebuddy 👍

PR #4182, opened 2017-07-26
Fix context throwing null TypeError

erku123:
When unit testing components that import PIXI the context ends up being null, 

```
TypeError: Cannot set property 'fillStyle' of null

      at createColoredCanvas (node_modules/pixi.js/lib/core/renderers/canvas/utils/canUseNewCanvasBlendModes.js:20:23)
      at canUseNewCanvasBlendModes (node_modules/pixi.js/lib/core/renderers/canvas/utils/canUseNewCanvasBlendModes.js:36:19)
      at Object.<anonymous> (node_modules/pixi.js/lib/core/sprites/canvas/CanvasTinter.js:220:61)
      at Object.<anonymous> (node_modules/pixi.js/lib/core/sprites/canvas/CanvasSpriteRenderer.js:13:21)
      at Object.<anonymous> (node_modules/pixi.js/lib/core/index.js:102:29)
      at Object.<anonymous> (node_modules/pixi.js/lib/index.js:18:13
```

### DIFF
--- a/src/core/renderers/canvas/utils/canUseNewCanvasBlendModes.js
+++ b/src/core/renderers/canvas/utils/canUseNewCanvasBlendModes.js
@@ -13,7 +13,11 @@ function createColoredCanvas(color)
     canvas.height = 1;
 
     const context = canvas.getContext('2d');
-
+    if(!context)
+    {
+     return canvas;   
+    }
+    
     context.fillStyle = color;
     context.fillRect(0, 0, 6, 1);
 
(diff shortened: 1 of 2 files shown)


englercj:
Aside from the failing linting, I'm not sure I understand why this change is necessary? I assume it is because you are trying to run a JS file that includes pixi.js in node, which is unsupported.

erku123:
Yeah I'm trying to write unit tests for a component that extends PIXI.Container,. The part that throws an TypeError is importing PiXI inside the component `import * as PIXI from 'pixi.js'`. Only fix that came to mind was to just check within the PIXI if context exists before trying to assign fillStyle to it(Maybe there's a better way?). And yeah, that's the only reason.

bigtimebuddy:
You might want to check out [floss](https://github.com/pixijs/floss) which is an Electron + Mocha tool for doing unit testing. It's what PixiJS uses! Running node unit-tests will not work, like @englercj said.

PR #4194, opened 2017-08-01
leading support

jcyuan:
add leading support to add extra space between lines before renderer measuring. it's different from lineHeight but based on lineHeight.


englercj:
Why couldn't you just increase `lineHeight`? Seems like `lineHeight=1` and `leading=1` is identical to `lineHeight=2`. Why not just use `lineHeight`?

GoodBoyDigital:
I agree, would be great to understand the difference between the two!

jcyuan:
for example.
`new PIXI.Text("fdsjkl\nfdskj\nxjhjklfd\nvncxke", { lineHeight : 1 });`
this code will show 4 lines overlapped each other as the height you set is 1 therefore it's no enough space to separate each lines.
but this code:
`new PIXI.Text("fdsjkl\nfdskj\nxjhjklfd\nvncxke", { leading : 1 });`
this will show the correct result what i want: lines have auto height + space, if use lineHeight here i need to measure my text first then set lineHeight = result + line space?
or are there some ways already to achieve?

jcyuan:
shortly i need auto-height (style.lineHeight = 0) + extra line spacing

PR #4198, opened 2017-08-02
Don't remove textures if they've been replaced in the cache

helloitsjoe:
Ran into an issue where a texture was replaced in the cache by a new texture with the same name, and was removed unintentionally, causing an error: `The frameID 'line1' does not exist in texture cache.` @andrewstart suggested this fix, making sure the texture to be removed is the same texture as the one passed into `removeFromCache()`


helloitsjoe:
Thanks guys!

PR #4199, opened 2017-08-03
Adds docs for TextStyle getters and setters

themoonrat:
https://github.com/pixijs/pixi.js/issues/4197

Example should show up at http://pixijs.download/text-style-docs/docs/PIXI.TextStyle.html once travis has finished it's thang


GoodBoyDigital:
nice one @themoonrat 👍

PR #4200, opened 2017-08-03
A display object can no longer be interactive if it has no alpha

themoonrat:
So, had a little bug on my game where I was seeing the 'hand' icon indicating it was hovering over something with interactive = true and buttonMode = true - but there was nothing there! And clicking on it made a button (which is usually in that location) callback occur. Turns out, that to make that button disappear, I'd tweened the alpha out, and the interaction manager only checks for if visible is true or not.

Considering the feel of the code felt "if it cannot be seen, it shouldn't be interactive" I've added an additional check to not check interactivity if a component has no alpha. Not sure how others feel about this, or the potential side effects it could have if people may have relied on this behaviour?

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -990,7 +990,7 @@ export default class InteractionManager extends EventEmitter
      */
     processInteractive(interactionEvent, displayObject, func, hitTest, interactive)
     {
-        if (!displayObject || !displayObject.visible)
+        if (!displayObject || !displayObject.visible || !displayObject.alpha)
         {
             return false;
         }


bigtimebuddy:
I'm not sure about this one. I think `displayobject.alpha = 0` should remain interactive. It enables devs the ability to easily create hidden interactive spots without relying on the `hitArea` property. Either way, I would consider this a breaking change.

GoodBoyDigital:
Hello! I agree with @bigtimebuddy , alpha is a visual thing only. You could run an on complete function that sets visible to false?

andrewstart:
On the one hand, people could  (probably?) use the `renderable` property to also have hidden interactive spots. 
On the other hand, `alpha === 0` is a rather fragile way of detecting interactibility - `alpha = 0.01` is realistically just as invisible but wouldn't turn off interactions. If your tween happens to stop short of 0 for whatever reason, you could still end up with an invisible interaction blocker.
For me, that second point outweighs the first, so I would tend to agree with @bigtimebuddy and @GoodBoyDigital.

themoonrat:
Cool, all in agreement, I'll close then :)

ivanpopelyshev:
That would be a breaking change. People really use that thing ;)

PR #4207, opened 2017-08-07
Support OffscreenCanvas

akihikodaki:
The target of `InteractionManager` could be an `OffscreenCanvas`, which does not have features required by the plugin.
Notify the lack of the feature to the plugin initializer and let it ignore the exception.
Now `OffscreenCanvas` is supported throughout PixiJS. Update documents accordingly.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,8 @@
     },
     "globals": {
         "global": false,
-        "__VERSION__": false
+        "__VERSION__": false,
+        "OffscreenCanvas": false
     },
     "parserOptions": {
         "ecmaVersion": 6,
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -29,7 +29,7 @@ export default class Application
      * @param {object} [options] - The optional renderer parameters
      * @param {number} [options.width=800] - the width of the renderers view
      * @param {number} [options.height=600] - the height of the renderers view
-     * @param {HTMLCanvasElement} [options.view] - the canvas to use as a view, optional
+     * @param {HTMLCanvasElement|OffscreenCanvas} [options.view] - the canvas to use as a view, optional
      * @param {boolean} [options.transparent=false] - If the render view is transparent, default false
      * @param {boolean} [options.antialias=false] - sets antialias (only applicable in chrome at the moment)
      * @param {boolean} [options.preserveDrawingBuffer=false] - enables drawing buffer preservation, enable this if you
@@ -146,7 +146,7 @@ export default class Application
 
     /**
      * Reference to the renderer's canvas element.
-     * @member {HTMLCanvasElement}
+     * @member {HTMLCanvasElement|OffscreenCanvas}
      * @readonly
      */
     get view()
--- a/src/core/autoDetectRenderer.js
+++ b/src/core/autoDetectRenderer.js
@@ -13,7 +13,7 @@ import WebGLRenderer from './renderers/webgl/WebGLRenderer';
  * @param {object} [options] - The optional renderer parameters
  * @param {number} [options.width=800] - the width of the renderers view
  * @param {number} [options.height=600] - the height of the renderers view
- * @param {HTMLCanvasElement} [options.view] - the canvas to use as a view, optional
+ * @param {HTMLCanvasElement|OffscreenCanvas} [options.view] - the canvas to use as a view, optional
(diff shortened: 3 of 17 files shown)


ivanpopelyshev:
Btw, I think we have to move InteractionManager outside of renderer in next version of pixi, add it to PIXI.Application.

bigtimebuddy:
Thank you for the pull request. Very thorough and high quality! I especially appreciate updating all the docs as well.

Despite the quality, however, I'm on the fence about this PR. In general, I think custom error types are useful and PixiJS is lacking in this area, this is attempting to solve an issue that is extremely corner-case for many developers. Until OffscreenCanvas has wider adoption (currently FF only), I personally don't believe this should be a v4 feature or we should be recommending it as an API via docs. 

Adding experimental API support to PixiJS is interesting but generally not that practical (similaraly there's another open PR for supporting a webGL new feature #4119). Not sure the ongoing maintenance cost is worth it _yet_, especially where there's no chance of fallback support.

What do others think?

ivanpopelyshev:
Make it that way interaction doesnt do anything if its not canvas. We'll think about exceptions later, also, like I said before, we might move interaction out of renderer plugins.

akihikodaki:
> Until OffscreenCanvas has wider adoption (currently FF only)

I'm actually testing on Chromium 60. Both of Firefox and Chromium support it, but require to flag manually.

> Adding experimental API support to PixiJS is interesting but generally not that practical

The reason why I use Chromium is that I'm considering to implement an offscreen renderer for Electron. I can distribute Electron configured to have the flag by default for our users. I believe that is a practical use case.

Even though those implementations are experimental, [the API is already on WHATWG HTML Standard and is no longer experimental](https://wiki.whatwg.org/wiki/OffscreenCanvas).

> Not sure the ongoing maintenance cost is worth it yet, especially where there's no chance of fallback support.

I think it is. The implementation side is making progress in the last few years, and now both of Mozilla and Google have finished the work and it can be enabled by setting flags. They are currently focusing on bug fixes and testing. Sooner or later, the feature would be publicly available. It is reasonable to have `OffscreenCanvas` support now.
Fortunately supporting `OffscreenCanvas` is easy because the differences are minor:
* `OffscreenCanvas` does not inherit `HTMLElement`.
* `OffscreenCanvas` does not have `toDataURL`.
* `OffscreenCanvas` does not have `toBlob`.
* `OffscreenCanvasRenderingContext2D` does not support text rendering.
* `OffscreenCanvasRenderingContext2D` does not support user interface features.

> I personally don't believe this should be a v4 feature or we should be recommending it as an API via docs.

Then it could be an "experimental" feature as like browsers do. Actually we could have bugs or regressions until we have sufficient use cases. I think we should have a notice about that in the case. Shall I add one?

englercj:
I tend to agree with @bigtimebuddy on this one. The InteractionManager is getting an overhaul in v5, so adding experimental features to a relatively stable area in v4 doesn't seem like a great way forward. MDN still list this feature as ["experimental"](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) and like you mentioned, both implementations are still behind a flag.

Additionally, there are multiple places that check for `instanceof OffscreenCanvas` which throws an error because `OffscreenCanvas` doesn't exist in most browsers currently.

Finally, I'm not sure I understand the point of the custom error here. It is thrown in the constructor, but ignored in the only place the constructor is used. This has the effect of now just silently failing to create the InteractionManager, but I'm not sure why this is desired?

If you don't want to use the InteractionManager you can make a custom build without it, or just unregister the plugin before creating your renderer (looks like I forgot to include that function in pluginTarget, oops, should add that).

While supporting this feature makes sense to me, I'm not sure it makes sense in v4 for the reasons mentioned above. Similarly to this, `createImageBitmap` is something we support in v5, but probably won't in v4.

bigtimebuddy:
Agreed with @englercj. We should consider this for upcoming v5, but for v4 the best solution is probably for @akihikodaki to create a fork/custom build for their Electron app.

GoodBoyDigital:
I agree with @bigtimebuddy and @englercj - not something for v4 but a potential for v5 once OffscreenCanvas slips into the mainstream 👍

bigtimebuddy:
Sounds like a consensus, so I'm closing. Thanks for the PR and the conversation. We'll revisit in the future.

PR #4211, opened 2017-08-07
Fixed nested stenicl masks not rendering correctly

ceco-fmedia:
Test case: https://jsfiddle.net/x1zq0xbd/

### DIFF
--- a/src/core/renderers/webgl/managers/StencilManager.js
+++ b/src/core/renderers/webgl/managers/StencilManager.js
@@ -51,13 +51,13 @@ export default class StencilManager extends WebGLManager
         const gl = this.renderer.gl;
         const sms = this.stencilMaskStack;
         const currentMasks = sms.length;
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
Thanks for this PR @ceco-fmedia! I will review shortly 👍

themoonrat:
Not my area of expertise, but my only comment is the maths clearly works, but it's not explained. Looking at it from my point of view, in the code at certain points you have 

Math.pow(2, currentMasks + 1) - 1
currentMasks + 1
currentMasks + 1, (2 * sFuncMask) + 1) 

For me, perhaps even just a couple of well named variables that help explain what's being done with a couple of the bits of maths might help others coming to modify the code later

ivanpopelyshev:
Please use "<<" instead of pow, we are working with binary operations here.

englercj:
+1 for more comments, explaining what is happening makes it easier for others to fix it in the future.

PR #4212, opened 2017-08-08
Support OffscreenCanvas

akihikodaki:
This is a port of #4207 for v5.

### DIFF
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -29,7 +29,7 @@ export default class Application
      * @param {object} [options] - The optional renderer parameters
      * @param {number} [options.width=800] - the width of the renderers view
      * @param {number} [options.height=600] - the height of the renderers view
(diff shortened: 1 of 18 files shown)


bigtimebuddy:
I had a chat with @GoodBoyDigital about this PR and we decided we will look into proper OffscreenCanvas support once v5 is more stable. We want to address this but it's a lower priority above refactoring some other features, namely re-doing the InteractionManager (which may impact this). Also, considering OffscreenCanvas support is so [low](http://caniuse.com/#feat=offscreencanvas), doesn't make sense to consider before other higher value features like WebGL2.

Thank you for all your work on this, we really appreciate it. When we are ready to consider OffscreenCanvas, is it okay if we tap you for help? We appreciate your expertise in this area since many of us have little experience with it.

Issue #4213, opened 2017-08-08
Spritesheet inside @0.5 spritesheet, Frame does not fit inside the base Texture dimensions


axldns:
Prior to PIXI **v4.1.1** we were able to put one spritesheet inside another (e.g. bitmap font inside bigger atlas) and use both BitmapText fonts along with other images from the atlas.
 
This applies also for the case where that bigger atlas has "@0.5" in name, used advisedly for some mobile platforms. PNG twice smaller, JSON the same, all contents correctly upscaled by PIXI.

From PIXI **v4.2.1** on, using combined atlases while parent atlas is @0.5 causes either 

- textures misplacement 

- Error: `Texture Error: frame does not fit inside the base Texture dimensions`

I'm attaching minimal source sample exposing the problem for your investigation.

Please advise. 
[pixi4v5bitmapfont.zip](https://github.com/pixijs/pixi.js/files/1208638/pixi4v5bitmapfont.zip)

ivanpopelyshev:
Ok, its a problem because different packers make different output formats. I dont know what to do about that, it really depends on user case.

themoonrat:
Closing as PR submitted fix this

PR #4214, opened 2017-08-09
[Please close] and navigate to https://github.com/pixijs/pixi.js/pull/4246

axldns:
… frames already scaled #4213

minimal source sample exposing the problem:
https://github.com/axldns/axldns.github.io/tree/master/pixi4v5bitmapfont
live:
https://axldns.github.io/pixi4v5bitmapfont/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,9 @@
-PixiJS — A 2D JavaScript Renderer
+PixiJS — The HTML5 Creation Engine
 =============
 
 ![pixi.js logo](http://www.goodboydigital.com/pixijs/pixiV4_wide_full.jpg)
 
-
-
-
-[![Inline docs](http://inch-ci.org/github/GoodBoyDigital/pixi.js.svg?branch=dev)](http://inch-ci.org/github/GoodBoyDigital/pixi.js)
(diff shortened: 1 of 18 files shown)


bigtimebuddy:
Thanks @axldns. Would you mind fixing linting error here: https://travis-ci.org/pixijs/pixi.js/builds/262621654#L308

axldns:
sure, finally passed

englercj:
Looks like this is mostly just a revert of #3993, is the issue that PR fixed active again after this change?

axldns:
@englercj indeed, looks like revert. 
Not sure what do you mean with the other part. 

- If you ask if our PR resolves our issue, then yes, it does (in number of games) 

- If you ask if our PR breaks anything up with issue #3993 - we're not able to tell since no source files, no live examples attached in that issue

axldns:
Thanks for the review @bigtimebuddy . 
The easy workaround proposed isn't that easy in fact. We have lot of games on production servers. They all benefit from bitmap fonts nested into bigger atlases @ various resolutions and PIXI supporting this until **v4.1.1**.
Understanding both scenarios analysed in your review, please find updated PR. I have also changed target branch from dev to master, since this could be considered as backward capability fix.
Best

axldns:
Moved to https://github.com/pixijs/pixi.js/pull/4246 since target branch has changed

PR #4215, opened 2017-08-09
Add Google Analytics to Documentation

bigtimebuddy:
## Overview

Implement Google Analytics for documentation as a way to understand which APIs are most used, browsed for, looked up. Also, as a way to understand the traffic relationships between Github/Wiki and Docs.


bigtimebuddy:
@englercj good catch, fixed

GoodBoyDigital:
nice work @bigtimebuddy 👍

PR #4231, opened 2017-08-16
Add `updateTexture` to Sprite

finscn:
In My game. I need this feature often. 
 I don't know what about other users.
If you think it's useless , could close it . 

Thanks.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -404,6 +404,19 @@ export default class Sprite extends Container
     }
 
     /**
+     * Update texture if the content of texture has changed.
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
We can add it in v5 when we have special CanvasResource and may be CanvasSprite on top of it.

bigtimebuddy:
I'm going to close this for reasons that @ivanpopelyshev mentioned. But mainly, we are trying to limiting new APIs going into v4, especially if the are trivial. I would suggest mixing this into your project.

If there's a lot of dev enthusiasm for this, we can re-open.

finscn:
@bigtimebuddy I hope v5 could include this API.
use canvas as baseTexture is a common request in html5 game

PR #4246, opened 2017-08-22
fixes issue #4213 by applying resolution to self-contained bitmap-fon…

axldns:
…ts only while nested ones have resolution 1:1 (since parent already scaled)

### DIFF
--- a/src/extras/BitmapText.js
+++ b/src/extras/BitmapText.js
@@ -510,7 +510,8 @@ export default class BitmapText extends core.Container
         const data = {};
         const info = xml.getElementsByTagName('info')[0];
         const common = xml.getElementsByTagName('common')[0];
-        const res = texture.baseTexture.resolution || settings.RESOLUTION;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Hey @axldns, this is a better approach and it seems like it would maintain the existing functionality. 

Couple of things:

1. If I understood your other [PR comment](https://github.com/pixijs/pixi.js/pull/4214#issuecomment-324003110) correctly, you need this PR to patch v4.1.1? Unfortunately, we cannot support a patch release for this version so the PR should be made to **dev** and not **master**.  We only use **master** for critical or blocking bugs on the current release. If you are not prepared to use the latest version of PixiJS, you should create a patch/fork for the current version you need to support.

2. Could you add an inline comment explaining why you are checking for this? This helps to create some context when people are looking at the source code. 

3. This is a good use-case for adding a unit-test. This will make sure that things aren't broken in future releases or when we refactor.

bigtimebuddy:
Please make the PR against **dev**. Thanks.

PR #4257, opened 2017-08-25
refactor, remove unnecessary operations

ivanpopelyshev:
Don't use ratio and there wont be such problem. Now, where is my beer?

I tested it. Also, improved version for projections is here: https://github.com/pixijs/pixi-projection/blob/master/src/proj2d/z_masks/SpriteMaskFilter.ts#L70


bigtimebuddy:
@ivanpopelyshev Looks good. Some questions:

1. Does this resolve any known bugs or is it simply a refactor of the filter transform math?
2. When you say you "tested it", how did you test? Have you checked yet against **pixi-filters**?

ivanpopelyshev:
Its only a refactor, I wrote down formulas, looked if they are equal what we had before.

I used it with SpriteMaskFilter: http://pixijs.github.io/examples/#/demos/alpha-mask.js . 

Also same operations work for projective matrices in my plugin, and its very sophisticated case.

I will improve it later to work with texture from atlas.

PR #4271, opened 2017-08-30
Improving SVG handling in Spritesheets with sourceScale support

davearel:
Spritesheets no longer work properly if your using the `sourceScale` feature introduced in #2614.

This fix multiplies the size by the BaseTexture's sourceScale.

Note: My only concern is Math.floor. It works, however, it's easy to see why it could become a problem. Not sure how else to fix this. sourceScale uses [Math.round](https://github.com/pixijs/pixi.js/commit/4bcacc57bb764782cb307f10ee2048f7d4a4fddd#diff-2820734d8d9cebcd6dd6120421612e81R512) to ensure a clean width/height. I can't use Math.round here because multiple frames could end up being wider or higher than it's baseTexture.

Suggestions?

### DIFF
--- a/src/core/textures/Spritesheet.js
+++ b/src/core/textures/Spritesheet.js
@@ -151,6 +151,7 @@ export default class Spritesheet
     {
         let frameIndex = initialFrameIndex;
         const maxFrames = Spritesheet.BATCH_SIZE;
+        const sourceScale = this.baseTexture.sourceScale;
 
         while (frameIndex - initialFrameIndex < maxFrames && frameIndex < this._frameKeys.length)
         {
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Different versions of texturePacker/shoebox/whateverToolYouUse => differences inside spritesheet json. I dont know what to do about it. Guys?

englercj:
@ivanpopelyshev Not sure this is a problem with tool's output json so much as an SVG scaling issue where we added sourceScale but didn't support it in the Spritesheet parser.

bigtimebuddy:
Thanks @davearel for the addition!

Issue #3926, opened 2017-04-11
FilterManager. Code style may be a potential error.


mmykola:
Hi, I used custom filter with settings: 
```
uniforms: {
        uMyTexture: {
          type: 'sampler2d',
          value: ''
        }
}
```
but it doesn't work so I changed type to 'sampler2D'. 
There was code 
```
if (uniformData[i].type === 'sampler2D' && uniforms[i] !== 0) {
```
in FilterManager.syncUniforms class.

Perhaps you need to check type for 'sampler2D' in a register lowercase?

englercj:
That's a good idea. We used `sampler2D` because that is the actual type, however you are right that doing a `.toLowerCase()` on the type could catch issues like this and probably ones with the other type too.

Do you have time to put a PR in that does a `.toLowerCase()` before checking types?

mmykola:
Yes, I'll create a pull request with fixed.

PR #4275, opened 2017-09-01
Lowercase on FilterManager type checks

staff0rd:
Fix for #3926


ivanpopelyshev:
Objection! Do it in constructor of filter or something like that, its bad to have extra object allocations in the render cycle.

englercj:
Not sure it will actually make a measurable impact, but feel free to open a diff PR.

PR #4277, opened 2017-09-02
Use local variable for `uniformData[i].type.toLowerCase()`

finscn:


ivanpopelyshev:
That's a bit better. Can we do "lowercase" in filter constructor?

PR #4281, opened 2017-09-04
Fix texture's frame dimension error report

EcutDavid:
What's the error: 
The error will tell you both X and Y are wrong if you only one of them has incorrect value:
For example, set the height to `Number.MAX_VALUE`, then:
image

Why change `frame && frame.width ...` to `frame.width ...`?
If `frame == undefined`, it already breaks during  `if (frame.x + frame.width > this.baseTexture.width || frame.y + frame.height > this.baseTexture.height)` :)

### DIFF
--- a/src/core/textures/Texture.js
+++ b/src/core/textures/Texture.js
@@ -562,7 +562,8 @@ export default class Texture extends EventEmitter
             const errorX = `X: ${x} + ${width} = ${x + width} > ${this.baseTexture.width}`;
             const errorY = `Y: ${y} + ${height} = ${y + height} > ${this.baseTexture.height}`;
 
-            throw new Error(`Texture Error: frame does not fit inside the base Texture dimensions: ${errorX} ${relationship} ${errorY}`);
+            throw new Error('Texture Error: frame does not fit inside the base Texture dimensions: '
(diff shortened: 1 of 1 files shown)


englercj:
The error message isn't saying both are wrong. It is saying something is wrong and dumping the information to let you find out the issue. I like the idea of just getting all the information when the error happens.

EcutDavid:
Thanks, I'll come up with changes that fix the wrong `>` :)
image

englercj:
If you want to be pedantic, you could just put an "OR" between them.

EcutDavid:
Hey @englercj @ivanpopelyshev thanks, changes in
image

EcutDavid:
image
🤔

ivanpopelyshev:
Guys, do we have a problem with travis?

bigtimebuddy:
Thanks for the PR @EcutDavid!

Hoangminhtrang0109:
HVT: Nguyễn Văn Công NS: 17/07/1990 CMT:031090000662 GMAIL: minhtrang.hp010915@gmail.com STK: 2108205048930 NH: Agribank Cát Bà

PR #4284, opened 2017-09-05
move toLowerCase into Filter constructor

ivanpopelyshev:
subj

### DIFF
--- a/src/core/renderers/webgl/managers/FilterManager.js
+++ b/src/core/renderers/webgl/managers/FilterManager.js
@@ -327,7 +327,7 @@ export default class FilterManager extends WebGLManager
         // TODO Cacheing layer..
         for (const i in uniformData)
         {
-            const type = uniformData[i].type.toLowerCase();
+            const type = uniformData[i].type;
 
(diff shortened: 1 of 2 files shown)


src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -327,7 +327,7 @@ export default class FilterManager extends WebGLManager
         // TODO Cacheing layer..
         for (const i in uniformData)
         {
-            const type = uniformData[i].type.toLowerCase();
+            const type = uniformData[i].type;
```
bigtimebuddy: `const {type} = uniformData[i];`?

ivanpopelyshev: Seems like pointless change to me.

bigtimebuddy: Only a suggestion. less code

ivanpopelyshev: ok, next time when i see multiple of these :)

PR #4285, opened 2017-09-05
filters can survive context loss

ivanpopelyshev:
I believe that PIXI v4 is LTS, so I backported solution from my fork.

Its not fixed in v5: https://github.com/pixijs/pixi.js/blob/next/src/core/renderers/webgl/systems/FilterSystem.js

It is fixed in my v5 fork: https://github.com/gameofbombs/gobi/blob/master/src/pixi/renderers/systems/ShaderSystem.ts

To test the feature put this code in any of pixi examples that has filters, and see if it survives after context is lost. With my patch there are no errors/warnings in console at all.

```js
function loseContext() {
	var gl = app.renderer.gl;
	var ext = gl;
	if (!ext.loseContext) {
		ext = gl.getExtension('WEBGL_lose_context');
	}
	ext.loseContext();
	setTimeout(function () {
		ext.restoreContext();
	}, 1000);
}

setTimeout(loseContext, 2000);
```

### DIFF
--- a/src/core/renderers/webgl/managers/FilterManager.js
+++ b/src/core/renderers/webgl/managers/FilterManager.js
@@ -49,6 +49,8 @@ export default class FilterManager extends WebGLManager
         this.pool = {};
 
         this.filterData = null;
+
+        this.managedFilters = [];
     }
 
     /**
@@ -231,6 +233,8 @@ export default class FilterManager extends WebGLManager
             {
                 shader = filter.glShaders[renderer.CONTEXT_UID] = new Shader(this.gl, filter.vertexSrc, filter.fragmentSrc);
             }
+
+            this.managedFilters.push(filter);
 
             // TODO - this only needs to be done once?
             renderer.bindVao(null);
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
One more issue that is possible to fix in this commit is that pixi webgl manager types are exported in docs but they're not exported in runtime. I cant just hack "PIXI.FilterManager" in a plugin :( Do you think I should add managers export in this PR?

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -488,11 +492,32 @@ export default class FilterManager extends WebGLManager
     /**
      * Destroys this Filter Manager.
      *
+     * @param {boolean} contextLost context was lost, do not free shaders
+     *
      */
-    destroy()
+    destroy(contextLost)
```
bigtimebuddy: Could you set `destroy(contextLost = false)` and update the doc `@param {boolean} [contextLost=false] context was lost, do not free shaders` to not be implicit about falsifying?

PR #4289, opened 2017-09-07
tint in ParticleContainer

### LINKED ISSUE #3777: ParticleContainer.js should support tint optionality, as many effects require this
Doesn't seem like it would effect performance all that much, unless i'm missing something. The tint is applied in shader correct, and a CPU calculation done in client to invert the hex.

... Specifically, it becomes possible to animate the tint, or re-use the same "glow" texture in many different situations, especially if your animation data is pre-rendered.

ivanpopelyshev:
Trying to fix #3777

### DIFF
--- a/src/particles/ParticleContainer.js
+++ b/src/particles/ParticleContainer.js
@@ -149,11 +149,8 @@ export default class ParticleContainer extends core.Container
             this._properties[1] = 'position' in properties ? !!properties.position : this._properties[1];
             this._properties[2] = 'rotation' in properties ? !!properties.rotation : this._properties[2];
             this._properties[3] = 'uvs' in properties ? !!properties.uvs : this._properties[3];
-            this._properties[4] = ('alpha' in properties) ? !!properties.alpha : this._properties[4];
-            if ('tint' in properties)
-            {
-                this._properties[4] = !!properties.tint;
-            }
+            this._properties[4] = 'alpha' in properties || 'tint' in properties
+                ? !!properties.alpha || !!properties.tint : this._properties[4];
         }
     }
 
--- a/src/particles/ParticleContainer.js
+++ b/src/particles/ParticleContainer.js
@@ -149,7 +149,11 @@ export default class ParticleContainer extends core.Container
             this._properties[1] = 'position' in properties ? !!properties.position : this._properties[1];
             this._properties[2] = 'rotation' in properties ? !!properties.rotation : this._properties[2];
             this._properties[3] = 'uvs' in properties ? !!properties.uvs : this._properties[3];
-            this._properties[4] = 'alpha' in properties ? !!properties.alpha : this._properties[4];
+            this._properties[4] = ('alpha' in properties) ? !!properties.alpha : this._properties[4];
+            if ('tint' in properties)
+            {
+                this._properties[4] = !!properties.tint;
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/particles/ParticleContainer.js
```diff
@@ -149,7 +149,11 @@ export default class ParticleContainer extends core.Container
             this._properties[1] = 'position' in properties ? !!properties.position : this._properties[1];
             this._properties[2] = 'rotation' in properties ? !!properties.rotation : this._properties[2];
             this._properties[3] = 'uvs' in properties ? !!properties.uvs : this._properties[3];
-            this._properties[4] = 'alpha' in properties ? !!properties.alpha : this._properties[4];
+            this._properties[4] = ('alpha' in properties) ? !!properties.alpha : this._properties[4];
```
englercj: unnecessary change?

src/particles/ParticleContainer.js
```diff
@@ -149,7 +149,11 @@ export default class ParticleContainer extends core.Container
...
+            if ('tint' in properties)
+            {
+                this._properties[4] = !!properties.tint;
+            }
```
englercj: I don't think this logic is quite right, if I set `alpha` to true, but `tint` to false then alpha doesn't get uploaded. Probably need to change this to something like:

```js
this._properties[4] = 'alpha' in properties || 'tint' in properties ? !!properties.alpha || !!properties.tint : this._properties[4];
```

That way either of them being set will use the value, and either of them being set to true will cause the upload.

src/particles/webgl/ParticleRenderer.js
```diff
@@ -387,16 +389,21 @@ export default class ParticleRenderer extends core.ObjectRenderer
...
+            const premultiplied = sprite._texture.baseTexture.premultipliedAlpha;
+            const alpha = sprite.alpha;
+            // we dont call extra function if alpha is 1.0, that's faster
+            const argb = alpha < 1.0 && premultiplied ? premultiplyTint(sprite._tintRGB, alpha)
+                : sprite._tintRGB + (alpha * 255 << 24);
+
+            array[offset] = argb;
+            array[offset + stride] = argb;
+            array[offset + (stride * 2)] = argb;
+            array[offset + (stride * 3)] = argb;
```
englercj: Looking this and the other functions, I'm not clear why there are 4 assignments in them? If the stride is 1, and you are uploading UNSIGNED_BYTE to a vec4, wouldn't you only assign `array[offset]`?

Why do you have to fill 4 Uint32s for an UNSIGNED_BYTE vec4? This pattern is used in all the upload funcs, just not sure why each of them copy the data 4 times?

ivanpopelyshev: That's how our SpriteRenderer works. 4 assignments are for different vertices, and its not clear what stride is, its always more than zero. This abstraction exists because we dont know which fields are static and which are dynamic.

src/particles/ParticleContainer.js
ivanpopelyshev: Agree.

src/particles/webgl/ParticleRenderer.js
englercj: Ah, that's right we upload it per vert (one for each uv coord). Forgot about that! Thanks!

ivanpopelyshev:
I didnt think it would be approved that fast :)

bigtimebuddy:
@ivanpopelyshev is it possible to see a jsfiddle example of this functionality?

ivanpopelyshev:
@bigtimebuddy it is possible in pixi-examples, just run https://pixijs.github.io/examples/#/basics/basic.js with current version and with new one. Suddenly, tint starts working!

bigtimebuddy:
Verified! Great work @ivanpopelyshev, you're a machine

PR #4291, opened 2017-09-08
 Add `tintScale` for improving `tint`.

finscn:
the tint is a powerful feature for Sprite . But it has a limitation : 
`tint` could make the original RGB of Sprite  become smaller , can't be bigger. e.g. :

> the original RGB is  ( 150, 160, 120) , `tint` can't make it become  (200 , 80 , 180).  R & B are bigger , G is smaller.
 
But with `tintScale`,  the `tint` could be more powerful.

When the original RGB is ( 150, 160, 120) ,  we could make  `tint = E254FF` &  `tintScale = 1.5` , then the result will be approximately equal to ( 200 , 80 , 180).

This PR  just takes  a little change , but could make `tint` be more powerful & useful .

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -75,6 +75,14 @@ export default class Sprite extends Container
         this.tint = 0xFFFFFF;
 
         /**
+         * The tintScale will scale the value of tint in shader. `new tint color = tint * tintScale` .
+         *
+         * @member {number}
+         * @default 1.0
+         */
+        this.tintScale = 1.0;
+
+        /**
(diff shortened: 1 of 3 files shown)


finscn:
This feature can't support `CanvasSpriteRenderer` now.
Because I don't understand the tint algorithm in Canvas2D. I don'e know how to  implement `tintScale` in CanvasSpriteRenderer

ivanpopelyshev:
Dont do that, I'll make two-tint scale in a plugin, and it works differently: its a linear equation. Please look at EsotericSoftware's spine dark-light shader.

finscn:
I see. waiting for you

Issue #4294, opened 2017-09-13
Incorrect documentation for ParticleContainer maxSize default value


griest024:
The [documentation](https://github.com/pixijs/pixi.js/blob/dev/src/particles/ParticleContainer.js#L31) says that the max size for `ParticleContainer` is 15000 but [the default value](https://github.com/pixijs/pixi.js/blob/dev/src/particles/ParticleContainer.js#L40) is only 1500. Is this a typo?

ivanpopelyshev:
Yes, it should be 1500

bigtimebuddy:
Welcome a pull request to fix this.

ivanpopelyshev:
Actually, when I read the code, I see that it should be 15000 by default.

ivanpopelyshev:
I dont know, I think I just change ParticleContainer that way it works better :)

PR #4295, opened 2017-09-13
particle container docs and resize mode

ivanpopelyshev:
We will need to improve ParticleContainer later, but that will do for now.

Added autoResize, changed docs.

#4294

### DIFF
--- a/src/particles/ParticleContainer.js
+++ b/src/particles/ParticleContainer.js
@@ -28,7 +28,7 @@ import { hex2rgb } from '../core/utils';
 export default class ParticleContainer extends core.Container
 {
     /**
-     * @param {number} [maxSize=1500] - The maximum number of particles that can be renderer by the container.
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Feel free to make better standart for it. As an example, I think that we should specify batchSize, initial number of batches and maximum number of batches.

src/particles/ParticleContainer.js
```diff
@@ -28,16 +28,19 @@ import { hex2rgb } from '../core/utils';
 export default class ParticleContainer extends core.Container
 {
     /**
-     * @param {number} [maxSize=15000] - The maximum number of particles that can be renderer by the container.
+     * @param {number} [maxSize=1500] - The maximum number of particles that can be renderer by the container.
```
griest024: while you're at it, it looks like this comment should read

> The maximum number of particles that can be rendered by the container.

renderer -> rendered

PR #4299, opened 2017-09-13
fixing dirty problem in plane/nineslice

### LINKED ISSUE #4296: Changing texture of NineSlicePlane does not change the image.
We have a NineSlicePlane which we use as a XP Bar. When XP reaches it max, it changes from a Red texture to a Blue Texture.

What seems to be happening in WebGl, is that when the texture is changed, I can see on the main NineSlicePlane, the uv's have changes (image is in a sprite sheet). But the _glDats for the uvBuffer, keep the old uv's of the original texture.

gldata

ivanpopelyshev:
Fixes #4296

We forgot that some instances aren't booleans.

### DIFF
--- a/src/mesh/NineSlicePlane.js
+++ b/src/mesh/NineSlicePlane.js
@@ -363,7 +363,7 @@ export default class NineSlicePlane extends Plane
         this.updateHorizontalVertices();
         this.updateVerticalVertices();
 
-        this.dirty = true;
+        this.dirty++;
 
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
https://jsfiddle.net/n47g1h7t/ - that was a problem

bigtimebuddy:
Going to merge, because this issue was validated already in #4296

PR #4309, opened 2017-09-17
fix NineSlice with resolution

ivanpopelyshev:
Somehow, we didn't use resolution for NineSlicePlane canvas version, and fix is easy. However I dont have fiddle for it, people just reported this problem and solution to me. 

Theoretically, that's how it suppose to work.


bigtimebuddy:
Thanks @ivanpopelyshev

PR #4316, opened 2017-09-20
Force base render textures to be whole pixels to fix error in IE

themoonrat:
I have a bit of code that takes a texture rendered by PIXI.Text, then creates a rope from it to create curved text. I then put that to a renderTexture to be used by sprites.
However, on IE 11 only, (WebGL in IE in fact), the texture just looks bizarre. I've created a fiddle that shows normal text (top), rope (left) then the sprite using a generated texture (right). I've forced some crazy resolution values to help replicate the issue.
By using Math.ceil on width and height of created render textures, the issue goes away. I think this is safe, since regular textures you upload with sprites couldn't be a fraction of a pixel wide, so why should created textures?

https://jsfiddle.net/themoonrat/71hadj49/

In case the fiddle doesn't work for you (remember, IE 11 WebGL only), here's how it looks to me in Chrome
chrome

And then in IE
ie11jpg

### DIFF
--- a/src/core/textures/BaseRenderTexture.js
+++ b/src/core/textures/BaseRenderTexture.js
@@ -54,8 +54,8 @@ export default class BaseRenderTexture extends BaseTexture
 
         this.resolution = resolution || settings.RESOLUTION;
 
-        this.width = width;
-        this.height = height;
+        this.width = Math.ceil(width);
+        this.height = Math.ceil(height);
 
         this.realWidth = this.width * this.resolution;
         this.realHeight = this.height * this.resolution;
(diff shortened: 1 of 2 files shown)


themoonrat:
@ivanpopelyshev If I just Math.ceil realWidth and realHeight, it does not fix the issue. I could ALSO Math.ceil realWidth and realHeight just in case you think other issues could crop up by not doing so, but it's the base width and height that require the Math.ceil in this instance

bigtimebuddy:
@ivanpopelyshev what do you want to do about this?

ivanpopelyshev:
I think we accept it in this form, I dont have good tests for other cases now.

PR #4331, opened 2017-09-27
added diagonal gradient feature for TextStyle

okajimad:
I added diagonal gradient feature for TextStyle. Previously only vertical and horizontal gradient are supported.
The demonstration page is here:
https://s3-ap-northeast-1.amazonaws.com/poderosa/gradation/angled_gradient.html

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -307,10 +307,12 @@ export const TRANSFORM_MODE = {
  * @type {object}
  * @property {number} LINEAR_VERTICAL Vertical gradient
  * @property {number} LINEAR_HORIZONTAL Linear gradient
+ * @property {number} LINEAR_DIAGONAL diagonal gradient. the angle is specified by TextStyle.gradientAngle.
  */
 export const TEXT_GRADIENT = {
     LINEAR_VERTICAL: 0,
     LINEAR_HORIZONTAL: 1,
+    LINEAR_DIAGONAL: 2,
 };
 
 /**
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -506,7 +506,18 @@ export default class Text extends Sprite
         else
         {
             // start the gradient at the center left of the canvas, and end at the center right of the canvas
-            gradient = this.context.createLinearGradient(0, height / 2, width, height / 2);
+            if (style.fillGradientType === TEXT_GRADIENT.LINEAR_DIAGONAL)
+            {
+                // if gradientAngle is positive, bottom-left corner is start point of gradient.
+                // if it is negative, top-left corner is start point of gradient.
+                // we have to calculate the parameter of createLinearGradient to let the gradient include canvas entirely.
+                const abs_angle = Math.abs(style.gradientAngle);
+                const tan = Math.tan(abs_angle);
+                const p = (width + height * tan)/(1+tan*tan);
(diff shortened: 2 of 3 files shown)


ivanpopelyshev:
@bigtimebuddy @GoodBoyDigital @themoonrat 

Guys, can we at least have degrees here, in text style?

themoonrat:
It is my (personal) belief that a style of text on one line, should be the same on each individual line on multiline text. For example, look at this fiddle: https://jsfiddle.net/themoonrat/d3roaLmf/
Note how for the vertical gradient, each individual line has the same gradient applied to it.

What your PR adds is a diagonal gradient that applies to the entire area of the text.  Which might be fine. There's a PR I need to get back to (https://github.com/pixijs/pixi.js/commit/90a4e6ac48c1903a00081552f56d43932fe60cab) that added, amongst other things, LINEAR_VERTICAL_NOREPEAT, which made it clear that this style was not going to repeat on each line

okajimad:
As for multiple lines, I think gradient should be applied to the entire text area. The meaning of 'repeat' is not clear especially if long lines and short lines are mixed.

themoonrat:
I have never wanted to use a gradient that doesn't have the gradient applied equally to all lines of text. If I want the top 50% red, the bottom 50% blue... I want every line of text to have the top 50% red, bottom 50% blue, not the top 50% lines of text being red, and the bottom 50% lines of text being blue.

For people who only ever use single lines of text, I know my usage is different. I'm just letting you know why things are as they are now :)

okajimad:
It is not realistic to apply diagonal gradation for each line by single gradient object of HTML canvas.  Initially the users set a value of angle to the TextStyle.
I think that it can not be helped if there are unnatural cases.

On the other hand, diagonal gradation should be convenient for single line text.

okajimad:
How can I proceed with the review process in the future?

ivanpopelyshev:
Make it into a plugin! I can help :)

themoonrat:
@bigtimebuddy Yeah, you rejected my PR on the same grounds, and it makes sense. I can't really think of a way of fixing things without making the v4 API as it is even more confusing, with yet another type you could pass in as a fill style. 

What I will say is that the tricky part with external utils to create a gradient to then pass to text is that for some gradients you will need to know how how many lines of text there are. And you can't know that until you've finished all of the text metric word wrap / break words functionality. In that case, the gradient can't be passed through to .fill already finished; it has to be generated at run time based on the lines of text and the request gradient style

okajimad:
@bigtimebuddy Does the overhaul or v5 development proceed at the present? If so, I would like to participate the development around the Text and gradation features. If not, please take in the PR as a provisional feature.

bigtimebuddy:
@okajimad Over the next few months we are going to start pushing out prereleases for v5. We have a few proposals on the wiki for major changes and we have a branch **next** which is where v5 will live for the time being. So far we have one _very_ provisional prerelease (v5.0.0-alpha), which mostly changes many of the low-level WebGL features to provide more flexibility.

If you want to propose changes for Text, that has been on our agenda, but no one has really dove in yet to proposal specifically what we might change in terms of API. There has been a lot of talk about trying to speed-up Text rendering by caching text characters to a BaseTexture instead of re-rendering to canvas every-time. 

Check out: https://github.com/pixijs/pixi.js/wiki/v5-How-to-Write-a-Proposal

We should limit feature development of PIXI.Text for v4, so I'm going to close this PR until we have a better strategy that also includes radial gradients as well as angled gradients.

PR #4322, opened 2017-09-22
Consistent Canvas State after render()

cawoodm:
Save and Restore canvas state so transform, style changes etc aren't propagated outside of render()

PR #4332, opened 2017-09-27
Consistent Canvas State after render() - 2

ivanpopelyshev:
Same as #4322 , just works.


ivanpopelyshev:
Dont forget to mention @cawoodm in release notes

PR #4343, opened 2017-10-02
Add unbind call for Geometry

yiwenl:
Add an unbind call to the geometry to bind the VAO to null. This is needed when working with other WebGL lib/tools which runs their own raw gl calls ( e.g. WebVR Polyfill : https://github.com/googlevr/webvr-polyfill). Need to unbind the vao before third party lib start binding their own buffers.


ivanpopelyshev:
Btw, I have my own v5 fork, it has same Geometry.js but it can even survive context loss.

yiwenl:
Hi thanks,

Yes we are aware v5 is under development. We are using it to create a WebVR experiment and it works great !

PR #4344, opened 2017-10-03
Add textures in reset method

ivanpopelyshev:
That's for pixi integration with threejs.

Another old thread for threejs is here: https://github.com/pixijs/pixi.js/issues/3230

Basically, if we dont have proper "reset", we can switch between renderers.

WORK IN PROGRESS.

Build will be here: https://pixijs.download/dev-fix-reset/pixi.js


gle6da:
Confirmed. In production build reset method doesn't work at all:
https://jsfiddle.net/gle6/667tkg4t/1/

Using this fix all just fine:
https://jsfiddle.net/gle6/667tkg4t/

ivanpopelyshev:
Works together with threejs: https://jsfiddle.net/xyL1sv02/1/

PR #4348, opened 2017-10-04
Adds an optional custom animation order for AnimatedSprite

### LINKED ISSUE #4346: Feature Request: Set a custom frame order or loop for AnimatedSprite
It would be cool if you could add a field to `AnimatedSprite.FrameObject` that designates the next frame to play after the current one. It would save me from having a bunch of `AnimatedSprites` just to implement a single character.

I would be willing to submit a PR if you are open to this feature.

griest024:
fixes #4346 

- Adds the ability to specify a next frame in `FrameObject`
- When a next frame has been defined for the current frame, the `AnimatedSprite` will play that frame, instead of the subsequent frame
- Adds a callback for when the `AnimatedSprite` detects and navigates to a 'next' frame
- Adds unit tests for `AnimatedSprite`, including the added 'next-frame' feature
- Added 5 32x32 monochromatic images to use for testing purposes


src/extras/AnimatedSprite.js
```diff
@@ -5,6 +5,7 @@ import * as core from '../core';
  * @type {object}
  * @property {PIXI.Texture} texture - The {@link PIXI.Texture} of the frame
  * @property {number} time - the duration of the frame in ms
+ * @property {number} next - the next frame to play
```
griest024: How would I mark this as optional?

griest024:
Actually don't merge this yet. I think there's a bug when it will loop instead of going to the next frame.

src/extras/AnimatedSprite.js
```diff
@@ -246,19 +261,32 @@ export default class AnimatedSprite extends core.Sprite
         }
         else if (previousFrame !== this.currentFrame)
         {
-            if (this.loop && this.onLoop)
+            const next = this._nextFrames[previousFrame];
+
+            if (next || next === 0) // 0 is valid
```
bigtimebuddy: Maybe use "next !== undefined" to avoid both conditions.

src/extras/AnimatedSprite.js
bigtimebuddy: Try "[next]"?

bigtimebuddy:
@griest024: thanks for this feature -- great idea, AnimatedSprite could definitely use some love when playing subsets of textures and having one DisplayObject is useful. The unit-tests are very well thought-out and make validating of PRs much faster, so thank you 👏 

I feel like we can improve this API approach though. Creating a pseudo linked-list (i.e. `next`) with the textures that are passed to the constructor is hard to read and confuses the declarative intention with the `FrameObject` being about the frame instead of the sequence or playback.

PixiSound has an API which creates sound sprites, which is similar in some ways to this idea, for [example](https://pixijs.github.io/pixi-sound/examples/sprites.html). You define sub-sequences of the timeline using from-to markers and reference the sound sprite by alias.

```js
const anim = new PIXI.extras.AnimatedSprite(textures);
anim.addClip('run', 0, 12); // (alias, startFrame, endFrame)
anim.addClip('walk', 13, 20);
anim.play('walk'); // also, maybe anim.gotoAndPlay('walk')?
```

griest024:
@bigtimebuddy actually my first instinct was to have named animations like [EaselJS](http://www.createjs.com/docs/easeljs/classes/SpriteSheet.html) but I thought that a breaking change might throw you off the idea.

If you like how EaselJS does it, we can completely rip off their idea and implement that.

Basically we would create a `Animation` typedef that includes the following fields:
- `frames`: An array of frames to play in order. I think this is preferable to having just a start and end frame since an array wouldn't force the user to have all the textures in the animation passed to the `AnimatedSprite` sequentially.
- `next`: A string or number representing the animation or frame to play after the animation completes. I'm not sure this can be optional since it might be strange to decide what to play after a named animation. 
- `speed`: A number representing how fast each frame is played. 1 is no change.

bigtimebuddy:
The EaselJS approach suffers some of the same issues as this PR. It tries to confuse defining clips with choreography of those clips. IMHO, these should be separate methods/systems. There are many cases where you'd want to be able to have more fine control over what animation plays after which, if they loop or play once through, how fast the playback is, if the animation completes or if it's cancelled. These approaches that use "next" field don't scale well or allow for more complexity with regards to playback.

I think I'm still in favor of adding an additional method or methods to AnimatedSprite for defining clips + some alteration to the playback methods. None of these should be breaking changes.

Also, we will not consider breaking changes or minor refactors in v4, since we are in the process of working towards v5 right now. High-value features and bug changes are okay.

ivanpopelyshev:
Make it into a plugin!

bigtimebuddy:
Yeah, plugin might be a better approach here. In general, I think AnimatedSprite is one of the more neglected features of PixiJS and would love to see it overhauled for v5.

griest024:
Are there any specifications for writing a PIXI plugin? Do I need to call `PIXI.use(plugin)` or something like that?

@bigtimebuddy You're right, I think just adding a method would be better. Should I write up a proposal for v5 or make changes now for v4?

bigtimebuddy:
I'm okay with either approach. 

If you want to make some changes to this PR or a new PR for v4 that would be fine although we move slowly here and I'd like some other folks to weigh-in before we merge--so that would take some time and back-and-forth. 

If you want to add some crazy new features and functionality, break changes, etc, the plugin option might be the more flexible way forward. If the community digs the plugin, we would definitely consider incorporating ideas for v5, and maybe could use your help with a proposal. 

I guess this depends on how much you want to be involved. Long-term make a plugin, short-term make non-breaking changes to AnimatedSprite via PR. Either way, we're grateful for your willingness to help!

To answer your plugin question: renderer, loader and prepare plugins have specific hooks that are needed, but in general a plugin can just be extending a class or adding something to the `PIXI` global window object. @ivanpopelyshev has created many plugins and would be a good resource for how to set one up from scratch. Also, new sub-systems & classes that extend core PixiJS functionality make for better, more stable plugins than adding prototype methods to existing classes, overriding existing methods, or monkey-patching classes.

griest024:
@bigtimebuddy I think I'm just going to make a plugin. You can keep up with the development [here](https://gitlab.com/griest/pixi-actor). Let me know what you want to do about the v5 proposal, I'm interested in helping with development.

bigtimebuddy:
@griest024 Sounds good! I will definitely check-out the plugin as you are developing it. You should share your progress on the PixiJS Slack channel to let other people in the PixiJS community know about it. If you aren't already a member of the Slack, email me. Email is on GitHub my profile. 

Will be closing this PR. If you want to start on drafting a proposal for v5 related to AnimatedSprite, please just add a new page to the wiki, it's open to edit for all. Otherwise, keep an eye out for info on v5. Usually post on Slack, Twitter, and for specific issues on GitHub.

Issue #4334, opened 2017-09-28
Alpha on container affects children in unexpected way


nikkop:
When two sibling Graphics objects (or Sprites) are overlapping and their parent's (in this case a container) alpha is being manipulated, the results are not what I'm expecting. When setting the container alpha to 0.5, both the children will behave as if themselves had an alpha of 0.5 (even though they have an alpha of 1)

What's actually happening:
![alphaissue1](https://user-images.githubusercontent.com/6517126/30970201-cbeb4610-a464-11e7-9ad3-f293ab75351b.jpg)
[jsfiddle](https://jsfiddle.net/6jau9e29/1/)

What I was expecting:
![alphaissue](https://user-images.githubusercontent.com/6517126/30970075-8851feee-a464-11e7-8a6b-8c37d4ab6468.jpg)

Is this a bug or am I thinking wrong about the behavior? If so, how do I get the behavior I am looking for? I'm used to change opacity of groups/layers and I was hoping it would work the same in Pixi. I tried to search for it but no success.

Thanks in advance

ivanpopelyshev:
Its not a bug. You need to do it in a separate layer, either using filter either renderTexture:

### Filter

```js
var voidFilter = new PIXI.filters.VoidFilter();
container.filters = [voidFilter];
```

but in current pixi version it doesnt work with alpha, so for now you need a filter that  treats alpha correctly:

```js
var colorMatrix= new PIXI.filters.ColorMatrixFilter();
colorMatrix.alpha = 0.5;
container.filters = [colorMatrix];
```

@GoodBoyDigital @bigtimebuddy  we really need to add alpha or tint to VoidFilter!

### RenderTexture

This one doesnt work in realtime :( Its just snapshotting the container and then applying alpha on that texture. Im posting clear solution, without "cacheAsBitmap" and "generateTexture", those things have not enough info to do the alpha fix.

```js
var bounds = container.getBounds();
var x2 = Math.ceil(bounds.x + bounds.width);
var y2 = Math.ceil(bounds.y + bounds.height);
bounds.x = Math.floor(bounds.x);
bounds.y = Math.floor(bounds.y);

var trans = new PIXI.Matrix();
trans.tx = -bounds.x;
trans.ty = -bounds.y;

var renderTexture = PIXI.RenderTexture.create(bounds.width, bounds.height);
renderer.render(container, renderTexture, null, trans);

var sprite = new PIXI.Sprite(renderTexture);
sprite.position.x = bounds.x;
[...]

PR #4349, opened 2017-10-05
Add AlphaFilter, deprecate VoidFilter

ivanpopelyshev:
subj

#4334


src/filters/alpha/AlphaFilter.js
```diff
@@ -0,0 +1,45 @@
+import * as core from '../../core';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+/**
+ * Simplest filter - applies alpha. Very handy.
```
bigtimebuddy: This description is missing two pieces of information:

1). Should have a note about how this differs from Container's `alpha`. e.g., You want to use this instead of Container's alpha property to avoid visual layering of individual elements. AlphaFilter applies alpha evenly across the entire display object and any elements it contains. 

2). VoidFilter had a purpose which was sort of a hack to apply a filter area to something. The "very handy" part... need to explain why.

src/filters/alpha/AlphaFilter.js
```diff
@@ -0,0 +1,45 @@
...
+    constructor()
+    {
+        super(
+            // vertex shader
+            readFileSync(join(__dirname, '../fragments/default.vert'), 'utf8'),
+            // fragment shader
+            readFileSync(join(__dirname, './alpha.frag'), 'utf8')
+        );
+
+        this.uniforms.uAlpha = 1.0;
```
bigtimebuddy: why not do `this.alpha = 1.0` here instead? It's shorter and doesn't duplicate `uAlpha`

src/filters/alpha/AlphaFilter.js
ivanpopelyshev: is this better?

finscn:
I need this PR . I hope it could be merged as soon as possible . Thanks.

Cristy94:
There's an issue with TypeScript currently, the master of @types/pixi.js includes AlphaFilter and not VoidFilter, but the master of npm pixi.js doesn't include AlphaFilter.

ivanpopelyshev:
oh, right, its 4.6.0 soon to be released :)

ivanpopelyshev:
4.6.0 is here: https://github.com/pixijs/pixi.js/releases/tag/v4.6.0 , now ts is up-to-date

PR #4352, opened 2017-10-06
Allow texture to be null when create a new Mesh.

finscn:
Sometimes, user would pass `null` as  argument `texture` when call `new Mesh()`

### DIFF
--- a/src/mesh/Mesh.js
+++ b/src/mesh/Mesh.js
@@ -28,6 +28,7 @@ export default class Mesh extends core.Container
          * The texture of the Mesh
          *
          * @member {PIXI.Texture}
+         * @default PIXI.Texture.EMPTY
          * @private
          */
         this._texture = texture || Texture.EMPTY;
--- a/src/mesh/Mesh.js
+++ b/src/mesh/Mesh.js
@@ -29,11 +29,11 @@ export default class Mesh extends core.Container
          * @member {PIXI.Texture}
          * @private
          */
-        this._texture = texture;
+        this._texture = texture || PIXI.texture.EMPTY;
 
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Please, No!

Just make default value of it PIXI.Texture.EMPTY and pass undefined. Null is an error there.

bigtimebuddy:
Overall, not sure if `null` should be a valid constructor argument for Mesh. Texture is not optional for Sprites, for instance. Why would have a default here? If you need to use `null` then why not set it to `Texture.EMPTY` in your use-case?

finscn:
@bigtimebuddy  , there are some codes like this : 
```
        this.uvs = uvs || new Float32Array([
            0, 0,
            1, 0,
            1, 1,
            0, 1]);

        this.vertices = vertices || new Float32Array([
            0, 0,
            100, 0,
            100, 100,
            0, 100]);

        this.indices = indices || new Uint16Array([0, 1, 3, 2]);

        this.drawMode = drawMode || Mesh.DRAW_MODES.TRIANGLE_MESH;

```

If you allow `uvs , vertices , indices ,  drawMode` to be null or undefined , why don't allow `texture` to be null ?

Or the same question :  `If user need to use null (for drawMode) then why not set it to DRAW_MODES.TRIANGLE_MESH in user use-case?`

bigtimebuddy:
@finscn would you make this PR to **dev** instead of **dev-mesh-update**? Thanks.

finscn:
OK. but dev-mesh-updat need this PR too.

bigtimebuddy:
@finscn we wanted this change on **dev**, Ivan can merge the upstream changes back into dev-mesh-update.

PR #4357, opened 2017-10-09
Feature to support trimmed Textures as masks

ivanpopelyshev:
The best test is modified https://pixijs.github.io/examples/#/basics/spritesheet.js

The only bad thing about this PR is usage of `PIXI.extras.TextureTransform`.

```js
var app = new PIXI.Application();

PIXI.loader
    .add('required/assets/basics/fighter.json')
    .load(onAssetsLoaded);
var back = new PIXI.Sprite(PIXI.Texture.WHITE);
back.width = app.screen.width;
back.height = app.screen.height;

function onAssetsLoaded()
{
    // create an array of textures from an image path
    var frames = [];

    for (var i = 0; i < 30; i++) {
        var val = i < 10 ? '0' + i : i;

        // magically works since the spritesheet was loaded with the pixi loader
        frames.push(PIXI.Texture.fromFrame('rollSequence00' + val + '.png'));
    }

    // create an AnimatedSprite (brings back memories from the days of Flash, right ?)
    var anim = new PIXI.extras.AnimatedSprite(frames);

    /*
     * An AnimatedSprite inherits all the properties of a PIXI sprite
     * so you can change its position, its anchor, mask it, etc
     */
    anim.x = app.renderer.width / 2;
    anim.y = app.renderer.height / 2;
    anim.anchor.set(0.5);
    anim.animationSpeed = 0.5;
    anim.play();

    app.stage.addChild(anim);
    app.stage.addChild(back);
    back.mask = anim;

    // Animate the rotation
    app.ticker.add(function() {
        anim.rotation += 0.01;
    });
}

```


ivanpopelyshev:
In next PR I'll address name issues, possibly move it to `PIXI.TextureUvMatrix`, rename `texture.transform` to `texture.uvMatrix`, and `texture._updateUVs` into `texture.update()` and make it public, cause, you know, people actually use it when they manually update fields of `texture.frame`.

bigtimebuddy:
Preview: https://jsfiddle.net/bigtimebuddy/ubh1t7ft/

bigtimebuddy:
Okay, just to summarize @ivanpopelyshev and my conversation on PixiJS Slack:

- The benefit of this change is that it supports _any_ Texture as a mask, for instance, trimmed and rotated textures. Not just standalone, untrimmed Textures. The benefit is, for instance, AnimatedSprite masks (as seen in example above).
- This change would also require a follow-up PR to move TextureTransform into core and clean-up some organizational issues with TextureTransform being in "extras".

@GoodBoyDigital and I decided that we were going to push refactors and major feature PRs to *next* (v5) in order to stabilize v4 and make it easier to merge upstream changes back into *next* (currently in progress). I want @GoodBoyDigital to decide whether this is a v4 or v5.

If it doesn't mess up the work we're doing on *next* and can be incorporated relatively easily, I think this is a useful feature.

ivanpopelyshev:
Btw, do we know of any engines that support masking with atlas textures?

ivanpopelyshev:
I'm going to make better clipping, right now it can cut the edge of mask a bit. Dont merge it yet :)

ivanpopelyshev:
Ok, now it can be merged.

ivanpopelyshev:
Ok, so there's one serious change in that refactoring - _updateUvs() renamed to updateUvs() and made public.

People often change "frame" or "trim" of the texture manually and they dont know what to call in that case, that case was repeated in html5gamedevs forum multiple times. We have to expose `uvMatrix` and `updateUvs` for users.

ivanpopelyshev:
I did a mistake in previous PR, and took width/height of entire baseTexture. Of course it can be "fixed" on SpriteMaskFilter side but it looks much better if we fix it in filterTransforms.

PR #4358, opened 2017-10-11
No longer skip the update transform pass on generating a texture.

themoonrat:
Fixes this function not working on the canvas renderer.
Before: https://jsfiddle.net/themoonrat/Lnkr5bkg/
After: https://jsfiddle.net/themoonrat/12u1nhnq/1/


themoonrat:
Fixes: https://github.com/pixijs/pixi.js/issues/4317

themoonrat:
I think I'm going down the same rabbit hole you're doing down Ivan; with this PR, if I look to extract pixels from the same container I get a different result every time. Doing an element.getBounds()  on the element before trying to go through the texture generation process fixes THAT issue. But maybe that's why skipTransforms was set to true in the first place.... to not effect the parent container.

Chicken. Meet egg.

ivanpopelyshev:
One of my ideas was to export sprite that has special pivot or anchor to make the positioning much easier :)

bigtimebuddy:
@themoonrat when I set `forceCanvas` to be `false`, the fiddle strobes. Is this just me?
 
https://jsfiddle.net/bigtimebuddy/tpmhe9tw/


generatetexture

ivanpopelyshev:
It works for me. Whats going on?

bigtimebuddy:
Update: MacOS on Chrome (Firefox and Safari are fine)

ivanpopelyshev:
ok, which version of chrome it is? https://github.com/pixijs/pixi.js/issues/4342

bigtimebuddy:
Version 61.0.3163.100 (Official Build) (64-bit)

ivanpopelyshev:
Exactly like in that ticket.

bigtimebuddy:
Chatted with @GoodBoyDigital about this. He's going to look at this out after we get **next** in a better shape.

bigtimebuddy:
Added this to "Future Work" so we'll remember to address. Closing for now.

themoonrat:
@bigtimebuddy & @ivanpopelyshev - For what it's worth, I think I've solved this via changing  the generateTexture function to be from
`this.render(displayObject, renderTexture, false, tempMatrix, true);`
to
`this.render(displayObject, renderTexture, false, tempMatrix, !!displayObject.parent);`

This keeps current behaviour when containers have a parent, but solves the situation I had in canvas mode when I was trying to create a texture from a container that didn't have a parent.

Might be useful info!

ivanpopelyshev:
Yes, its useful. 

I'll rewrite everything in it for v5, I already had a PR for that.

PR #4362, opened 2017-10-12
bitmapfont: backward compatibility + tests

axldns:
_08 aug 2017_ - https://github.com/pixijs/pixi.js/issues/4213
Issue raised. Problem explanation with example attached.

_09 aug 2017_ - https://github.com/pixijs/pixi.js/pull/4214
PR to dev. Identified by @englercj as a revert of https://github.com/pixijs/pixi.js/pull/3993
Advised to modify PR maintaining current functionality.

_22 aug 2017_ - https://github.com/pixijs/pixi.js/pull/4246
PR to master (due to misinterpretation).
@bigtimebuddy kindly advised to write unit tests.

_NOW_
Unit tests written, new approach taken,  visual test attached.

Four scenarios have been identified.
1. Bitmap font on its own (bare)
2. Bitmap font on its own with @X.x filename (bare-scaled)
3. Bitmap font that is part of another spritesheet (nested)
4. Bitmap font that is part of another spritesheet with @X.x filename (nested-scaled)

During writing unit tests, it came out that, approach from https://github.com/pixijs/pixi.js/pull/4246 would break up existing functionality in scenario nested-scaled.

**New approach**:
Fetch resolution for bitmap font from filename that fnt file defines, rather than from texture.baseTexture.resolution, alternatively use settings.RESOLUTION.

- maintains current functionality in all scenarios

- restores backward compatibility

- allows us what we need (one fnt file for bitmap font nested into  spritesheets @ various resolutions)

P.S.
We're migrating (3.0.6 to 4.5.6) quite a big platform with number of games on production. This is the only last issue holding us back.


bigtimebuddy:
@axldns Great work, thank you for finding a better solution. 

Could you please remove **test/loaders/index.html** from the tests and maybe make this into a jsfiddle that we can take a look at? You might need to upload your local build and resources somewhere, but would be useful to see.

ivanpopelyshev:
Oh wow, I'll approve it when @bigtimebuddy  does :)

axldns:
Thanks for your quick response.
https://jsfiddle.net/yx8d9bh8/  ||  https://axldns.github.io/pixi4bitmapfontVisual/

ivanpopelyshev:
Fonts are pain for me. Its good that some people actually making those things better :)

PR #4363, opened 2017-10-12
rename GroupD8.isVertical

ivanpopelyshev:
Better explanation of the method, part of dev-mesh-update. 

I almost convinced Clark that our typings were wrong, then I found that I forgot to make that PR in pixi itself several months ago. https://github.com/pixijs/pixi-typescript/pull/186


ivanpopelyshev:
That's where that method is used: http://pixijs.github.io/examples/#/demos/texture-rotate.js

src/deprecation.js
```diff
@@ -601,6 +601,14 @@ export default function deprecation(core)
         return this.generateCanvasTexture(scaleMode, resolution);
     };
 
+    /**
+     * @method
+     * @name PIXI.GroupD8.isSwapWidthHeight
+     * @see PIXI.GroupD8.isVertical
+     * @param {number} rotation - The number to check.
```
bigtimebuddy: On deprecated methods we typically remove the params and return values so people won't use it.

ivanpopelyshev: Lint forces me to do that.

ivanpopelyshev: also, no, you dont. Look further after my method there are many others

ivanpopelyshev: Actually, that's not important ;) I suspect that the only ones who are using this function are I and @finscn :)

Issue #4365, opened 2017-10-15
Bug in test 'should add and remove pointer events to element'


jojuntune:
I tried running `npm test` on my laptop which has touch screen. I am checked to the commit `14246a8044ea6f819bec8197501e2790f4d1bbdc` on `dev` branch. I am receiving the following error:

```
1) PIXI PIXI.interaction.InteractionManager add/remove events should add and remove pointer events to element:
     AssertionError: expected addEventListener to have been called exactly thrice, but it was called 7 times
    addEventListener(pointerdown, function bound onPointerDown() {}, true) at InteractionManager.addEvents (\pixi.js\dist\pixi.js:9:1374515)
    addEventListener(pointerleave, function bound onPointerOut() {}, true) at InteractionManager.addEvents (\pixi.js\dist\pixi.js:9:1374640)
    addEventListener(pointerover, function bound onPointerOver() {}, true) at InteractionManager.addEvents (\pixi.js\dist\pixi.js:9:1374765)
    addEventListener(touchstart, function bound onPointerDown() {}, true) at InteractionManager.addEvents (\pixi.js\dist\pixi.js:9:1375844)
    addEventListener(touchcancel, function bound onPointerCancel() {}, true) at InteractionManager.addEvents (\pixi.js\dist\pixi.js:9:1375968)
    addEventListener(touchend, function bound onPointerUp() {}, true) at InteractionManager.addEvents (\pixi.js\dist\pixi.js:9:1376095)
    addEventListener(touchmove, function bound onPointerMove() {}, true) at InteractionManager.addEvents (\pixi.js\dist\pixi.js:9:1376215)
```

## Reproducing the issue
There are two ways:
Firstly, checkout to the mentioned commit on dev branch, then either,
- run `npm test` on a device which has touch screen
OR
- if your device does not have touch screen, go to file src/interaction/InteractionManager.js and change line 155
   - this.supportsTouchEvents = 'ontouchstart' in window; -> this.supportsTouchEvents = true;
- then run `npm test`

## Hypothesis for why the error is encountered

The failing test is in test/interaction/InteractionManager.js specifically on line 295 
`expect(element.addEventListener).to.have.been.calledThrice;`

I think the test does not take into account the case when touch screen is used. Namely, when
`this.supportsTouchEvents` evaluates to `true`, four additional touch events are added to interactionDOMElement, as can be seen in file dist/pixi.js on line 34679:
```
        if (this.supportsTouchEvents) {
[...]

PR #4366, opened 2017-10-15
Fixes bug in InteractionManager- test mentioned in issue #4365

jojuntune:
Addresses #4365


andrewstart:
Could you actually split this out into two separate tests, one with `supportsTouchEvents` manually set to true, and one with it set to false? I feel like that would be better for testing expectations instead of introducing variability into the test itself.

jojuntune:
Hi @andrewstart, 

yes I agree, it is better to do this the way you suggested.
I have updated the PR.

Issue #2271, opened 2016-01-04
I wish pixi.js could run on node-webgl


vitalije:
I would like to use pixi.js not inside browser, but on WebGL canvas obtained through node-webgl library. I understand that pixi.js itself uses also Canvas2D context, and that it won't be possible to use all features of pixi.js without 2D canvas. Yet, I would like to use those functions that do not need to depend on 2D canvas. I have already made some experiments and found out that it is possible to render sprites on WebGL obtained through node-webgl. It seems however that WebGL context objects are not 100% compatible. I am willing to try to port pixi.js myself (at least those features that are used by RPGmaker MV) to use webgl context provided by node-webl module. However, I need some help. If somebody can point me to the right direction it would be really a big help. What pixi.js requires/uses from WebGL context? How to find what is missing from context obtained through node-webgl comparing Chrome's WebGL context? Maybe it would be possible to make a thin layer - adapter that would transform a context obtained through node-webgl to something that pixi.js expects.

Another library I have found is node-canvas that provides 2d context for drawing. It would be great if pixi.js could run on some combination of those two libraries.

GoodBoyDigital:
Hi there! If there was some kind of feature comparison for the webGL vs node-webGL I could help and advise.

vitalije:
I am not aware of such comparison. As I stated before I would like to know what pixi.js requires/expects from WebGL context (a minimal set of functions, properties). I tried to find the answer myself by reading pixi.js code but without much luck. I have found that in some cases node-webgl is more demanding on type checking arguments used to call some of its functions than a normal browser context. It could be tweaked to accept arguments from pixi.js.

Anyway I didn't achieve much :disappointed:. If you @GoodBoyDigital  or anyone else have some advice I would appreciate it very much.

englercj:
I assume it is the DOM features that make pixi unable to run on just node-webgl?

vitalije:
Yes, that is one part of the problem. OTOH I could live without certain functionalities of pixi.js that rely on DOM. IIRC functions for displaying styled text depend on DOM. Maybe some others too.

The problem is that even if client application never calls any of those functions, pixi.js still refuses to run functions that don't depend on DOM.

GoodBoyDigital:
any luck on this?

vitalije:
not really :disappointed:

englercj:
[...]

PR #4367, opened 2017-10-16
[WIP] Node.JS support

saschagehlich:
I'm currently working on a server-side video rendering solution that is PIXI-backed. I was able to get PIXI v3 working by doing some monkeypatching and browser object mocking, but for v4 I'd like to build the solution into PIXI itself.

For CanvasRenderer I'm using [`node-canvas`](https://github.com/Automattic/node-canvas), for WebGL rendering I'm using [`headless-gl`](https://github.com/stackgl/headless-gl). These two packages are "drop-in" replacements for the browser features, but of course there are some difficulties - e.g. the fact that they're not working together out of the box.

Tasks:
- [X] Avoid usage of `window`, use `global` if present / possible
- [X] Avoid usage of `document.createElement('canvas')` - I created a wrapper called `createCanvas`. Uses `document.createElement` in the browser environment and `new Canvas()` in Node.JS environment
- [X] If `document` is not available, disable interaction handlers / plugins that require the DOM
- [X] Rename `sample` in WebGL shaders, since this seems to be a reserved word on OpenGL ES?
- [ ] Fix resource loading (e.g. `Texture.fromImage`)
- [ ] Modify CanvasRenderer so that it can render using `node-canvas`

Current state: I am now able to perform server-side rendering using WebGLRenderer, [like this](https://gist.github.com/saschagehlich/ada18ce4d7cb73380345e1ebed953345).

This PR requires some more work in `pixi-gl-core`. I'll create a PR for these changes soon and add a link to the PR.

### DIFF
--- a/src/accessibility/AccessibilityManager.js
+++ b/src/accessibility/AccessibilityManager.js
@@ -480,7 +480,15 @@ export default class AccessibilityManager
         this.children = null;
         this.renderer = null;
     }
+
+    static isSupported () {
+        return typeof document !== 'undefined'
+    }
 }
 
-core.WebGLRenderer.registerPlugin('accessibility', AccessibilityManager);
-core.CanvasRenderer.registerPlugin('accessibility', AccessibilityManager);
+if (AccessibilityManager.isSupported())
+{
+    core.WebGLRenderer.registerPlugin('accessibility', AccessibilityManager);
+    core.CanvasRenderer.registerPlugin('accessibility', AccessibilityManager);
+}
+
--- a/src/core/renderers/SystemRenderer.js
+++ b/src/core/renderers/SystemRenderer.js
@@ -5,6 +5,7 @@ import settings from '../settings';
 import Container from '../display/Container';
 import RenderTexture from '../textures/RenderTexture';
 import EventEmitter from 'eventemitter3';
+import createCanvas from '../utils/createCanvas';
 
 const tempMatrix = new Matrix();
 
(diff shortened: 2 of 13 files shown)


bigtimebuddy:
Related issue #2271

@saschagehlich, thanks for doing this work. Great start and good idea.

We sometimes get requests for this feature and others have made an attempt a patch to achieve the same ends. Currently, however, we are working towards v5 and not taking any major features to v4. So the timing, I'm afraid, is not good for this PR.

There are few ways forward:

1. Wait until v5 is more stable (probably several weeks to a month) and think about adding a more thoughtful, intentional approach to support node/headless-gl/etc. One of the features we are adding to v5 is to create more granular plugin modules for npm/yarn, this way it would be much easier to exclude things like InteractionManager & AccessiblityManager, when you don't need it. Instead of **pixi.js** I could see a new **pixi-headless** module that pulls only the necessary PixiJS modules. 

2. Create your own fork for v4 and maintain it temporarily for your needs and projects. If you need this in fast, this is going to be the best approach.

3. You _could_ create a more aggressive shim and monkey patch to achieve this and turn it into a plugin. In a previous project, for instance, my team was able to patch to Node by creating `global.document.createElement` and a few other high-level DOM APIs being called by PixiJS. It's not pretty, but it mostly worked! Fonts and images created some headaches.

I'm going to close because this will not be considered for v4. Sorry, but happy to discuss ways to get this into v5.

ivanpopelyshev:
I dont like "_sample", better to name it "texSample" or something like that. And yeah, maintain a plugin or fork, and we'll add it to v5, that's how we work with such big changes. There wont be much changes in v4, so it wont be difficult to take changes from upstream for your fork :)

Prozi:
bump: http://npmjs.com/package/pixi-shim

PR #4370, opened 2017-10-17
Add standalone methods about `render children`

finscn:
Extract standalone `renderChildren` method from `render`
more detail see : https://github.com/pixijs/pixi.js/issues/4369


bigtimebuddy:
Chatted about this with @ivanpopelyshev and we're in agreement that this is insufficient to allow more flexibility in modifying renderer behavior. This creates two methods _only_ so that they can be overridden. It doesn't provide any useful functionality aside from that.

I'm closing this because we will not consider for v4 as we are trying to limit the PR to critical bugs and very high-value API changes. While this does not quality for either of those, we would welcome a larger proposal for v5 to provide more flexibility to rendering children that provides more utility.

finscn:
@bigtimebuddy  ， I'm sorry for my bad english. I can't understand 

>  this is insufficient to allow more flexibility in modifying renderer behavior. This creates two methods only so that they can be overridden. It doesn't provide any useful functionality aside from that.

I can know every word with google translate , but I can understand the whole paragraph.

Could you explain them for me please ? Thanks.

bigtimebuddy:
Sure, no problem @finscn. Let me try to explain more simply. 

- This PR allows you to override two methods (`renderChildrenCanvas` and `renderChildrenWebGL`)
- This PR doesn't provide enough value for all users
- We'd welcome a larger PR to support more flexibility rendering children in v5

PR #4380, opened 2017-10-23
Adding "name" as optionnal parameter for all display objects

julien-moreau:
Not sure you'll like this pull request or if someone already asked you for this
I'm a fan of smart one liners and having the possibility (optional) to set a name directly when creating a container may simplify writing. It also leads the writers to use names and then potential editors may display interesting names instead of just a Javascript constructor name.

Also gives more sense to "getContainerByName" in extras

You ok with that?

Thanks for your answer!

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -17,11 +17,11 @@ import DisplayObject from './DisplayObject';
 export default class Container extends DisplayObject
 {
     /**
-     *
+     * @param {string} name - The name of the display object.
      */
-    constructor()
+    constructor(name = null)
     {
-        super();
+        super(name);
 
         /**
          * The array of children of this container.
--- a/src/core/display/DisplayObject.js
+++ b/src/core/display/DisplayObject.js
@@ -18,9 +18,9 @@ import { Rectangle } from '../math';
 export default class DisplayObject extends EventEmitter
 {
     /**
-     *
+     * @param {string} name - The name of the display object.
      */
-    constructor()
+    constructor(name = null)
     {
         super();
 
@@ -121,6 +121,13 @@ export default class DisplayObject extends EventEmitter
          * @readonly
          */
         this._destroyed = false;
+
+        /**
+         * The name of the object, in order to be easily identified when searching for a container or sprite
+         *
+         * @member {string}
+         */
+        this.name = name;
 
         /**
          * Fired when this DisplayObject is added to a Container.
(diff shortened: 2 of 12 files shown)


ivanpopelyshev:
There are many things that PIXI object lacks that were present in flash.

1. Name
2. Symbol
3. colorTransform
4. ...

The problem is that any of that increases number of functions in Container/DisplayObject and its already complicated for open-source project where everyone can contribute.

Also some of names functionality cant be done without Stage, because we cant add special cache for them :(

The only way to add names in pixi is to make it like mixin (https://github.com/pixijs/pixi.js/blob/dev/src/interaction/interactiveTarget.js). Also, its better to separate it into a plugin.

ivanpopelyshev:
SUMMARY: The idea of names by itself is good, but we have to make it maintainable for open-source => need mixin + plugin.

ivanpopelyshev:
Also, you can make "ts.d" for this plugin just fine: `setName` must be `this`, and its easy to add extra functions in `declare module PIXI { export interface Container { ... } }`

julien-moreau:
I'm not sure to understand your latest message:
- Do you prefer that I add "setName" to DisplayObject ?
- Do you prefer that I add a plugin which "augmentifies" the DisplayObject class using the Object.assign like InteractiveTarget ?

Sorry my misunderstanding :)

ivanpopelyshev:
Yeah, one more thing - it should be in another repository. Currently we are working on pixi-v5 in "next" branch and Im sure that "name" wont be integrated before it'll be used by many people as a plugin.

I'll create repo and basic structure, then I'll give you write rights for it :)

julien-moreau:
Thanks!
I now understand the project’s philosophy and that’s great!
I’ll just use the property .name for instance and will update our code when the plugins will be available :)

I keep this pull request opened until you create the repository, or you can close it, as you want :)

bigtimebuddy:
Thanks @julien-moreau for this, but unfortunately, like @ivanpopelyshev said, we are in the process of v5 and are freezing *dev* except for bugs or very high-value features. I'm going to close this now and suggest you work with Ivan on a plugin/mixin approach. If there's a lot of community enthusiasm for this feature, we'd definitely consider for v5.

julien-moreau:
Thanks! That sounds great!
Just created a plugin on my side until I work with @ivanpopelyshev :)

PR #4385, opened 2017-10-27
Ensures that text style properties are copied by value, not by reference

themoonrat:
Fixes an issue where
```js
const style1= new PIXI.TextStyle();
const style2 = new PIXI.TextStyle();
style1.fillGradientStops.push(0);
```
Because the array is copied by reference (as an `Object.assign` is currently used), 
`style2.fillGradientStops === style1.fillGradientStops`

Which is incorrect. The arrays should be separate instances

### DIFF
--- a/src/core/text/TextStyle.js
+++ b/src/core/text/TextStyle.js
@@ -95,10 +95,7 @@ export default class TextStyle
 
         this.reset();
 
-        for (const key in style)
-        {
-            deepCopyProperty(this, style, key);
-        }
+        deepCopyProperties(this, style, style);
     }
 
     /**
@@ -111,10 +108,7 @@ export default class TextStyle
     {
         const clonedProperties = {};
 
-        for (const key in defaultStyle)
-        {
-            deepCopyProperty(clonedProperties, this, key);
-        }
+        deepCopyProperties(clonedProperties, this, defaultStyle);
 
         return new TextStyle(clonedProperties);
     }
@@ -124,10 +118,7 @@ export default class TextStyle
      */
     reset()
     {
-        for (const key in defaultStyle)
-        {
-            deepCopyProperty(this, defaultStyle, key);
-        }
+        deepCopyProperties(this, defaultStyle, defaultStyle);
     }
 
     /**
@@ -767,16 +758,16 @@ function areArraysEqual(array1, array2)
 /**
  * Utility function to ensure that object properties are copied by value, and not by reference
  *
- * @param {Object} target Target object to copy property into
- * @param {Object} source Source object for the proporty to copy
- * @param {string} prop Property name to copy
+ * @param {Object} target Target object to copy properties into
+ * @param {Object} source Source object for the proporties to copy
+ * @param {string} propertyObj Object containing properties names we want to loop over
  */
-function deepCopyProperty(target, source, prop) {
-    if (Array.isArray(source[prop])) {
-        target[prop] = source[prop].slice();
-    } else if (source[prop] && typeof source[prop] === 'object') {
-        Object.assign(target[prop], source[prop]);
-    } else {
-        target[prop] = source[prop];
+function deepCopyProperties(target, source, propertyObj) {
+    for (const prop in propertyObj) {
+        if (Array.isArray(source[prop])) {
+            target[prop] = source[prop].slice();
+        } else {
+            target[prop] = source[prop];
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/core/text/TextStyle.js
```diff
@@ -93,7 +93,15 @@ export default class TextStyle
     {
         this.styleID = 0;
 
-        Object.assign(this, defaultStyle, style);
+        for (const key in defaultStyle)
+        {
+            deepCopyProperty(this, defaultStyle, key);
+        }
```
bigtimebuddy: Why not just call `this.reset()` here?

src/core/text/TextStyle.js
```diff
@@ -119,7 +124,10 @@ export default class TextStyle
      */
     reset()
     {
-        Object.assign(this, defaultStyle);
+        for (const key in defaultStyle)
+        {
+            deepCopyProperty(this, defaultStyle, key);
+        }
```
bigtimebuddy: Why not include the loop inside the deep copy function? Would reduce the couple of loops.

```js
function deepCopy(source, target) {
  for (const prop in defaultStyle) {
    if (Array.isArray(source[prop]) {
       target[prop] = source[prop].slice();
    }
   else {
      target[prop] = source[prop];
   }
  }
}
```

src/core/text/TextStyle.js
```diff
@@ -755,3 +763,20 @@ function areArraysEqual(array1, array2)
...
+ *
+ * @param {Object} target Target object to copy property into
+ * @param {Object} source Source object for the proporty to copy
+ * @param {string} prop Property name to copy
+ */
+function deepCopyProperty(target, source, prop) {
+    if (Array.isArray(source[prop])) {
+        target[prop] = source[prop].slice();
+    } else if (source[prop] && typeof source[prop] === 'object') {
+        Object.assign(target[prop], source[prop]);
```
bigtimebuddy: I don't think we need this `Object.assign` because no style properties are plain objects.

src/core/text/TextStyle.js
themoonrat: Because in the constructor, you want to loop through just the properties passed through in the style object, and in this scenario you want to loop through all properties that are in default style. 

I could add a third parameter which is the object whose properties you want to loop through?

src/core/text/TextStyle.js
themoonrat: Right now we don't, but in the past there have been PR's that have suggested having one, so that's why I kept it in just in case.

Happy to remove if we're sure 4.x will never have style properties has objects, and if we're changing how it works in 5.x

src/core/text/TextStyle.js
bigtimebuddy: That works, and default arg to defaultStyle.

src/core/text/TextStyle.js
bigtimebuddy: Positive v4 will never have object style properties.

src/core/text/TextStyle.js
themoonrat: Changed made. I didn't make default arg to defaultStyle, because that then causes issues when you pass in nonsense into TextStyle constructor.

For example, give nothing with
```new PIXI.TextStyle()```
And your style object is null, which will kick in the es6 default parameter.

So just explicitly passing through the object to check properties for on the 3 occasions the function is used solved that in the easier way

File: src/core/text/TextStyle.js [hunk 4]
```diff
@@ -767,16 +758,16 @@ function areArraysEqual(array1, array2)
...
+ * @param {string} propertyObj Object containing properties names we want to loop over
  */
-function deepCopyProperty(target, source, prop) {
-    if (Array.isArray(source[prop])) {
-        target[prop] = source[prop].slice();
-    } else if (source[prop] && typeof source[prop] === 'object') {
-        Object.assign(target[prop], source[prop]);
-    } else {
-        target[prop] = source[prop];
+function deepCopyProperties(target, source, propertyObj) {
```
bigtimebuddy: What about `propertyObj = defaultStyle`? Then you don't need to define it above in those two places.

themoonrat: If you do that, then when you create a new TextStyle with nothing passed in, that null comes through to this function and the es6 default parameter kicks in, which you don't want.

bigtimebuddy: 👍

PR #4386, opened 2017-10-27
Adds alpha param to constructor of AlphaFilter

bigtimebuddy:
Really don't know how we missed this... 🤔 

http://pixijs.download/alpha-filter/docs/PIXI.filters.AlphaFilter.html


ivanpopelyshev:
WTF OMG

PR #4387, opened 2017-10-27
Next Restructure for v5

bigtimebuddy:
## Overview

This is a very large reorganization of the **next** branch. The goal of this reorganization was to split the **pixi.js** package into multiple packages which could be included by developers creating custom builds of PixiJS. 

In terms of build process, this PR removes the use of **pixify** (Browserify + Babel) and replaces it with [Rollup](https://rollupjs.org/) + [Buble](https://buble.surge.sh/guide/). Both of these tools have distinct advantages over Browserify and Babel. Namely, Rollup produces an ES6 build which can be used via the `module` field in the **package.json** by tools like Webpack for better tree-shaking.

In addition, [Lerna](https://lernajs.io/) has been integrated into this project to help manage a multiple package publishing structure. This was first test-driven on [pixi-filters](https://github.com/pixijs/pixi-filters) and has worked very successfully for the past handful of releases. This helps make running scripts and publish much easier.

Lastly, the goal of this PR was not to fix any code. Some basic unit-tests were added to packages without coverage. Mostly, however, code changes involved how packages were imported and exported as well as decoupling dependencies. The one exception are deprecations from v4 and v3--these were removed because they would have been a pain and unnecessary to port to each package.

### Folder Structure

* **./bundles** This is where **pixi.js** package now lives. This might also include feature bundles of PixiJS, for instance "pixi-minimal", which might just include the essential, "pixi-3d", etc.
* **./packages** This is the collection of PixiJS packages. All these new packages will be registered with `@pixi` NPM namespace.
* **./filter** This contains all the current core filters. This will eventually include **pixi-filters** filters.
* ~**./integrations** This is the collection of unit tests which require integration between multiple packages, for instance, rendering tests~
* **./tools** This contains tools for building and testing all the packages, to reduce the overhead of each package contain it's own Rollup configuration. These tools are designed to be private-only.

### Packages

* **packages/accessibility** - AccessibilityManager
* **packages/app** - Application
* **packages/constants** - constants like PIXI.BLEND_MODES
* **packages/core** - the core Shader, Geometry, Buffer and Renderers
* **packages/display** - DisplayObject and Container
* **packages/extract** - extract plugin for renderers
* **packages/fragments** - common filter vertex shader
* **packages/graphics** - Graphics
* **packages/interaction** - Interaction plugin for renderers
* **packages/loaders** - loaders that extend resource-loader
* **packages/math** - shapes like rectangle, polygon and point
* **packages/mesh** - contains Plane, Rope, etc
* **packages/mixin-cache-as-bitmap** - Functionality for cacheAsBitmap
* **packages/mixin-get-child-by-name** -  functionality for getChildByName
* **packages/mixin-get-global-position** - functionality for getGlobalPosition
* **packages/particles** - ParticleContainer
* **packages/polyfill** - legacy JS support
* **packages/prepare** - prepare plugin for renderers
* **packages/settings** - PIXI.settings currently, like PIXI.settings.RESOLUTION
* **packages/sprite** - The sprite and sprite renderer
* **packages/sprite-animated** - AnimatedSprite
* **packages/sprite-tiling** - TilingSprite
* **packages/spritesheet** - Spritesheet
* **packages/text** - Text
* **packages/text-bitmap** - BitmapText
* **packages/ticker** - Ticker
* **packages/utils** - Random utilities, like color utils

### Filters

* **filters/alpha** - AlphaFilter
* **filters/blur** - BlurFilter
* **filters/color-matrix** -  ColorMatrixFilter
* **filters/displacement** - DisplacementFilter
* **filters/fxaa** - FXAAFilter
* **filters/noise** - NoiseFilter

### Bundles

* **bundles/pixi.js** - Equivalent to current pixi.js package

### Outstanding Issues

* Currently, the CanvasRenderer is broken as well as particles and mesh classes, same as on **next**, will be fixed _after_ this merge. 
* The ability to publish **coverage** has been disabled from Travis. Because of the new dependency structure being more complicated, will have to re-tool how coverage is figured out.
* ~Travis is still failing, investigating~ _wrong shebang in tester/index.js_

### Future Suggestions

* Mixins are a bit of a mess across the project and handled inconsistently. For instance, `PIXI.utils.mixins.performMixins` happens after static plugins are registered with renderers (e.g. accessibility and interaction targets to DisplayObject). And there are also mixins like `cacheAsBitmap` and deprecations which are also added. Without moving to a different component-model for composing elements, these could be cleaned up.
* Would strongly suggest moving DisplacementFilter and NoiseFilter to **pixi-filters**, these do not seem like core filters and could be added to the rest of the filters. Also, there's a case to be made that DropShadowFilter and GlowFilter should be moved into the core packages, since they have much more general utility.
* Find a way to decouple Container from SystemRenderer. This would remove the dependency of core of DisplayObject/Container and Transforms. 
* Potentially make DisplayObject transforms and bounds new packages. Might be nice to exclude if your project doesn't call for them.
* Would like to remove unnecessary namespaces (such as `extras`, `ticker`, `loader`). In general, the PIXI top namespace could be flatter and would make for easier import for custom bundles, e.g.,`export * from '@pixi/whatever'`. 
* Need to figure out if documentation should be re-tooled by module, or still documentation represents docs of pixi.js (bundle).
* Pull-out all the canvas plugins/functionality into separate plugins. Would make it much easier for developers to exclude ALL canvas, if they are not building for those environments. For instance, extract, prepare could have canvas plugins externalized


ivanpopelyshev:
Im still looking through it.

bigtimebuddy:
@englercj Thanks for reviewing this. To your points:

- I talked with @GoodBoyDigital about this yesterday and we both strongly agreed that all filters to live in here and will probably deprecate **pixi-filters**. Maintenance-wise it makes sense to have those be together. I'll probably create a folder for all **filters**.
- `@internal` is my convention, but admittedly an imperfect one. It is an attempt to create internal packages that we consider separate from the `@pixi` namespace. Doing this provides some convenience for using Lerna commands, for instance `lerna run build --ignore @internal/*`.

ivanpopelyshev:
Please don't move pixi-filters here yet.

bigtimebuddy:
Np @ivanpopelyshev. I will await your feedback before moving ahead with this merge.

bigtimebuddy:
@ivanpopelyshev, I am not planning on moving pixi-filters until we get much closer to an official release, that will take some time to do anyways and outside the scope of this PR. I also would like to merge git histories too, which will take some time to do.

ivanpopelyshev:
Just imagine what'll happen if we merge all issues in filters and main pixi repo.

bigtimebuddy:
To @ivanpopelyshev's point, I think filters are a rather benign issue overall. While there's lots of activity now on **pixi-filters**, development on those things will stabilize mostly because there's not a lot of functionality in filters. They mostly just do one thing and changes are superficial after being created. However, we should definitely consider what should be "official" filters and what might be left behind. There are some filters that look demo-y but just don't have much utility in practice (e.g., Ascii, CrossHatch, Emboss) and we should consider not making those "official". A smaller set of high-value filters (Glow, Blur, DropShadow) is better than a library of gimmicky ones.

If the choice is a few more issues and PRs vs lower maintenance cost, ability easily refactor and publish all PixiJS code at once, I'll take the latter any day.

ivanpopelyshev:
Outline should be here too, its one of the most used. Btw, I really want outline without a filter to be added in v5. That'll be a killer feature.

ivanpopelyshev:
Ok, merge premultiplied utils into one file. And we'll reorganize filter after that PR gets merged.

PR #4397, opened 2017-11-01
Removes URL_FILE_EXTENSION and getUrlFileExtension

bigtimebuddy:
_I’m working on a series of incremental changes which remove, reorganize and realign the current utils (aka the “junk drawer” of PixiJS)._

### Changed

* Removes `URL_FILE_EXTENSION` and `getUrlFileExtension`, these are no longer used by BaseTexture


ivanpopelyshev:
What is used instead?

bigtimebuddy:
This is where it's being used currently in **dev** https://github.com/pixijs/pixi.js/blob/dev/src/core/textures/BaseTexture.js#L430-L463

It doesn't seem like decomposing data uri is supported anywhere in **next** for images (other than SVG).

ivanpopelyshev:
How exactly is fromImage detecting resource type in `next`?

bigtimebuddy:
I dunno? @GoodboyDigital?

GoodBoyDigital:
pow:
https://github.com/pixijs/pixi.js/blob/next/packages/core/src/textures/resources/createResource.js

bigtimebuddy:
Okay makes sense. I actually think the regexp that Chad has for the extension is more generalized for things like myfile.jpg?v=1, but we can fix that moving forward. Also, creatResource should handle data uri strings for SVG, looks like it doesn’t right now.

ivanpopelyshev:
Leave it in utils then. Leave all file-related stuff in utils.

bigtimebuddy:
The point I’m trying to make here is we should be shrinking our API by making as many of these utilities private or inline. This is a case were this functionality should just be in createResource. I’ll rework this a bit.

ivanpopelyshev:
This particular regexp and function can be used by plugins. I'll leave decision to you. Private function also can be referenced as `look how this module does the stuff`.

packages/core/src/textures/resources/createResource.js
```diff
@@ -3,22 +3,28 @@ import SVGResource from './SVGResource';
...
-        if (source.match(/\.(mp4|webm|ogg|h264|avi|mov)$/))
-        {
-            return new VideoResource.fromUrl(source);
-            // video!
-            // return Texture.fromVideoUrl(source);
-            // return SVGResource.from(url);
-        }
-        else if (source.match(/\.(svg)$/))
+        // search for file extension: period, 3-4 chars, then ?, # or EOL
+        const result = (/\.(\w{3,4})(?:$|\?|#)/i).exec(source);
```
bigtimebuddy: Using the same RegExp from v4's BaseTexture for resolving the file extension. This solves issues with files ending in "?...", "#...", as well as non-lowercasing.

ivanpopelyshev: nice. merge it.

PR #4407, opened 2017-11-06
Clear some properties in `ParticleBuffer.destroy()`.

finscn:
* Clear some properties in `ParticleBuffer.destroy()`.
* improve the shader.

### DIFF
--- a/src/particles/webgl/ParticleShader.js
+++ b/src/particles/webgl/ParticleShader.js
@@ -31,10 +31,10 @@ export default class ParticleShader extends Shader
                 'varying vec4 vColor;',
 
                 'void main(void){',
-                '   vec2 v = vec2(0.0, 0.0);',
+                '   float x = (aVertexPosition.x) * cos(aRotation) - (aVertexPosition.y) * sin(aRotation);',
+                '   float y = (aVertexPosition.x) * sin(aRotation) + (aVertexPosition.y) * cos(aRotation);',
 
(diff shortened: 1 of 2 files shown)


src/particles/webgl/ParticleShader.js
```diff
@@ -31,7 +31,7 @@ export default class ParticleShader extends Shader
                 'varying vec4 vColor;',
 
                 'void main(void){',
-                '   vec2 v = aVertexPosition;',
+                '   vec2 v = vec2(0.0, 0.0);',
```
bigtimebuddy: Could you explain this change @finscn?

finscn: The results of  `vec2 v = aVertexPosition;` & `vec2 v = vec2(0.0, 0.0);` are both OK.

But in the semantic  , the ` vec2 v = aVertexPosition;` is not accuracy  , and needless.

In this shader , the  `v` is a new vec2 , we don't need  copy or clone `aVertexPosition` to it. 

Use `v = aVertexPosition` will puzzle users (the people that read the code).

bigtimebuddy: Thanks for the explanation.

englercj: I prefer to just initialize it when I do shaders:

```glsl
 vec2 v(
    (aVertexPosition.x) * cos(aRotation) - (aVertexPosition.y) * sin(aRotation),
    (aVertexPosition.x) * sin(aRotation) + (aVertexPosition.y) * cos(aRotation)
);
```

PR #4408, opened 2017-11-06
Fixes Deprecated Strategy

bigtimebuddy:
### Changed

Rollup has a limitation (https://github.com/rollup/rollup/issues/1693) where namespace exports are frozen using `Object.freeze`. This PR works around this to allow changing PIXI namespaces (e.g. `PIXI.utils`)

Also, this PR resolves the proper way to inject deprecations into the build file as well as is overall a better strategy for deprecated method on the bundle only.


ivanpopelyshev:
+380 −280 

I dont like the balance. Are you sure? Please count how many times you export/import the same thing here, just for us to know how many times we need to change something if we change the class name or add a new one.

ivanpopelyshev:
Like, user with notepad needs to add X lines of code per new object in some of inner modules. At the same time, user with IDE has special operations.. ok, i dont think IDE's will help us even in renaming this.

We need clear instructions. If `Object.freeze` is the only thing that adds 100 more lines, then lets remove it or make it only in the bundle after all things are ready to be frozen.

bigtimebuddy:
I have a pull-request into Rollup (https://github.com/rollup/rollup/pull/1696) to allow opt-out of `Object.freeze`. Who knows when that'll happen.

I don't like importing and export all things explicitly. But right now we have no way to add proper deprecations. For instance `Object.defineProperty(PIXI.utils, 'whatever')` will not work. This throws `Cannot define property baz, object is not extensible`.

I've been grinding on this issue for awhile without a good option. I don't want to fork Rollup to accomplish what I want. Any other ideas?

ivanpopelyshev:
Fork it or remove deprecations for some time. I dont think incremental deprecations is a good thing here, we are changing lots of API's, its better to look at the result and then add necessary deprs.

ivanpopelyshev:
Also, you dont have to fork rollup, just use "bigtimebuddy:master" in package.json for now

themoonrat:
Couldn't deprecations be it's own package? And then within the src folders are a way for us to fine tune version deprecations.
For example, when v5 is released, there is a 'v4 -> v5.0' folder.
But if v5.1 is released with deprectations, there is a new 'v5.0 -> v5.1' folder.

Maybe for developer builds this is bundled in automatically, but for release builds not?

bigtimebuddy:
I had thought about the deprecations being another package like you suggested @themoonrat. I think there’s something user friendly about it just working. Particularly the pixi.js bundle. Some JS projects also build a compatibility mode which is designed to be more backward compatible. I want to try to find the right balance. Including another package seems a little annoying, but I’m not convinced either way.

ivanpopelyshev:
That is what used for our examples: https://github.com/ivanpopelyshev/pixi-legacy . It works with v3->v4 and back.

Anyway, temporary roolup fork in our case is a right move.

bigtimebuddy:
Okay, I'm going to close and try again with a less intense workaround.

PR #4411, opened 2017-11-07
Optimization for Build Times with Rollup

bigtimebuddy:
Simple change to optimize the build times overall with Rollup this removes double command line call for each build and takes advantage of any internal Rollup caching and uses less system memory.

Also, removes **rollup-watch** which is no longer needed in Rollup 0.46+

### `npm run build:dev`

Current: 35s
After PR: 27s
**22% faster**

### `npm run build`

Current: 54s
After PR: 44s
**18% faster**

### DIFF
--- a/bundles/pixi.js/package.json
+++ b/bundles/pixi.js/package.json
@@ -19,9 +19,9 @@
     "url": "https://github.com/pixijs/pixi.js.git"
   },
   "scripts": {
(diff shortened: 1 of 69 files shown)


cursedcoder:
+1, does it affect bundle size btw?

bigtimebuddy:
It does not affect bundle size, but in general, Rollup produces much smaller builds than the current Browserify approach.

PR #4424, opened 2017-11-10
Deprecate TransformStatic, TRANSFORM_MODE

bigtimebuddy:
## Overview

Refactor to get rid of two different Transforms which are basically the same. TransformStatic was introduced as a performance optimization to Transform by using ObservablePoint and only updating when it needs to. TransformStatic is a superior Transform and opting for the slower Transform should not be a user-option.

### Deprecated

* Removes `PIXI.TransformStatic`
* Removes `PIXI.TransformBase`
* Removes `PIXI.TRANSFORM_MODE`
* Removes `PIXI.settings.TRANSFORM_MODE`

### Changed

* Moves Transform class to `@pixi/math` package instead of `@pixi/display`.


ivanpopelyshev:
I just leave the note here: I have to add "autoUpdate" in transform in case someone calculates matrix differently and sets it with "setFromMatrix". Its required for spine, that's why TransformBase existed.

ivanpopelyshev:
https://github.com/pixijs/pixi.js/issues/4425

PR #4429, opened 2017-11-10
Decouple Canvas Functionality (v5)

bigtimebuddy:
## Overview

The goal of the major v5 refactor PR is to take all the Canvas-related code and make WebGL the default Renderer.

### Added

* Creates the following Canvas packages:
  - canvas-display
  - canvas-extract
  - canvas-graphics
  - canvas-mesh
  - canvas-prepare
  - canvas-renderer
  - canvas-sprite
  - canvas-sprite-tiling

### Changed

* Moved **filters/** to **packages/filters/**
* The **pixi.js** bundle contains _no_ canvas functionality, use **pixi.js-legacy** instead
  - http://pixijs.download/next-canvas/pixi-legacy.js
  - http://pixijs.download/next-canvas/pixi.js
### Fixed

* Travis now deploys minified builds

### Deprecated

* Renames `PIXI.WebGLRenderer` to `PIXI.Renderer`
* Renames `PIXI.Container#renderWebGL` to `PIXI.Container#render`
* Renames `PIXI.Container#_renderWebGL` to `PIXI.Container#_render`
* Renames `PIXI.Container#renderAdvancedWebGL` to `PIXI.Container#renderAdvanced`


ivanpopelyshev:
Wait while I check it

ivanpopelyshev:
It looks good. Something must be done to amoun of modules that we have, I'll talk about it in slack.

bundles/pixi-modern/README.md
```diff
@@ -0,0 +1,14 @@
+PixiJS Modern
```
cursedcoder: Not sure as of `modern` name here. I would think of it as something "new and unrelated to old pixi" .

Do you think there is no better options?

GoodBoyDigital: I agree with @cursedcoder, although in true critic fashion I can't think of anything better! How about pixi-webgl or something like that?

cursedcoder: @GoodBoyDigital `pixi-canary`?

bigtimebuddy: I chatted offline with @GoodBoyDigital about this, we we discussed that **pixi.js** should be the package _without_ canvas support. Much of what we're doing is making WebGL first-class and Canvas 2nd-class for v5. This change should also reflect that. The current **pixi.js** in next would be renamed to  **pixi.js-legacy** (or something similar) to have fallback-support for CanvasRenderer and compatibility with older browsers.

cursedcoder: I had this in my mind but didn't think you want to drop canvas that hard.

Big +1 :D

ivanpopelyshev:
I like it!

ivanpopelyshev:
Only one question, traditional: why +1900 -901 ?

ivanpopelyshev:
Yeah, that's what I talked about! I like those -100 !

PR #4432, opened 2017-11-12
PIXI.extras.MultiAnimatedSprite

sombriks:
It inherits AnimatedSprite and intends to make the texture array change an internal behavior.

This is a simple syntax sugar to allow us to do things like this:

```js
let walk = ["image_sequence_01.png","image_sequence_02.png","image_sequence_03.png","image_sequence_04.png"];

for (let i=0; i < 4; i++)
  walk[i] = PIXI.Texture.fromImage(walk[i]);
 
let stop = [walk[1]];
 
let mc = new PIXI.MultiAnimatedSprite({walk, stop},"stop");
 
mc.play("walk")
```

### DIFF
--- a/src/extras/MultiAnimatedSprite.js
+++ b/src/extras/MultiAnimatedSprite.js
@@ -1,5 +1,7 @@
 import * as core from '../core';
 
+import AnimatedSprite from "./AnimatedSprite";
+
 /**
  * @typedef FrameObject
  * @type {object}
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Good idea for a plugin. Whether it ends in v5 I dont know , but I like simplicity of your solution!

sombriks:
@ivanpopelyshev thank you, i am open to do any desired change into it. Also a better name to this class, MultiAnimatedSprite is too much upbeat for such small sugar.

cursedcoder:
I would say it doesn't make much sense to have another animated sprite, more vital would be to merge the functionality into existing `AnimatedSprite`.

Otherwise you can easily publish standalone package.

bigtimebuddy:
This PR seems related conceptually to #4348 and the author's current plugin [pixi-actor](https://gitlab.com/griest/pixi-actor).

@sombriks, like @ivanpopelyshev said, we are making a big push to v5 and we are saying "no" to new APIs or major changes to v4. For that reason, I'm going to close, but encourage you all to discuss further improvements to this approach.

We would definitely consider a change to v5 to support sub-clips or sequences to AnimatedSprite's main timeline. I agreed with @cursedcoder, this should be an additional feature to AnimatedSprite rather than a new class. In general, wouldn't be opposed to refactoring AnimatedSprite to be more robust.

PR #4433, opened 2017-11-12
making AnimatedSprite comfort the use of a map containing "named animations"

sombriks:
As suggested in the previous pull request, i've baked the proposed functionality into AnimatedSprite instead of inherit it.

it will allow to define AnimatedSprites in such way:
```javascript
let walk = ["p0.png", "p1.png", "p0.png", "p2.png"]
walk = walk.map(e => PIXI.Texture.fromImage(e))
let stop = [walk[0]]

let mc = new AnimatedSprite({stop, walk},"stop")
mc.play("walk")
```


src/extras/AnimatedSprite.js
```diff
@@ -30,13 +30,61 @@ import * as core from '../core';
...
-     *  objects that make up the animation
+     * @param {PIXI.Texture[]|FrameObject[]|object} textures - an array of {@link PIXI.Texture} or frame
+     * objects that make up the animation. it also can be a series of 'named animations', an object containing
+     * arrays of textures or frames.
+     * @param {string} initialAnimation - name of the initial animation. must exists inside the textures if textures
+     * is an object containing named animations. this param must be ommited if textures is an Array.
      * @param {boolean} [autoUpdate=true] - Whether to use PIXI.ticker.shared to auto update animation time.
      */
-    constructor(textures, autoUpdate)
+    constructor(textures, initialAnimation, autoUpdate)
```
cursedcoder: -1 for this BC

overall code seems bloated with all these statements

I would suggest using some defined structure for `textures`

here is some example for button with skins:

image

sombriks: Hi, @cursedcoder!

I am trying to not break backwards compatibility, indeed it would be much simpler, but i am not there to break someone's else code. minor releases are not intended to to such thing.

Also your image suggestion uses TypeScript idioms which are simply not found on pixi.js codebase, but i am still interested to be able to offer to other people the named animation idiom. 

Have you anoy other suggestion regarding how could i introduce this new idiom? 
- it must be backwards compatible (the inheritance strategy did that, but i was suggested to merge functionality on AnimatedSprite)
- it must pass happily from "npm run lint"

I'm willing to send another patch, but it only will pass if i got someone to support and guide me on this.

sombriks:
@ivanpopelyshev internally i check if we got a list or a map at the first argument therefore the code behaves as it should.

The first PR defined an entire new and simpler class (which had a very poor name) and then i was suggested to merge functionality into AnimatedSprite.

ivanpopelyshev:
Oh sorry, I read the code but didnt notice it. I like the idea, lets make it into v5 :) 

We can break stuff in it, so we can agree on better parameters. What if we make it all into first parameter? Either textures, either map of animations. I'll think about that.

PIXI vanilla AnimatedSprite is too poor, may be with just a few lines of code we can make it better.

sombriks:
@ivanpopelyshev sounds great!

I am particularly fond on destructuring assignments as default idiom for this, but i'll play a bit with more freedom on a new PR. willing for comments already!

bigtimebuddy:
@sombriks, One thing that is *very* important for any new feature PR is adding unit-tests. This feature definitely would require some good unit-testing to make sure the feature stays stable into the future. Also, creating a visual example of how it's used (e.g. jsfiddle), is a great way to get people excited about that feature.

llorenspujol:
+1 on this PR

PR #4434, opened 2017-11-12
Fixes global namespace in ES module

bigtimebuddy:
Rollup was exporting `PIXI` as a window object automatically using UMD format. However, with the ES version (which is used by Webpack), it had to be explicitly required, which was breaking other plugins that required a global PIXI object.

cc @GoodBoyDigital


GoodBoyDigital:
nice one @bigtimebuddy !

bigtimebuddy:
@GoodBoyDigital could you test this with your Webpack build?

PR #4435, opened 2017-11-13
Accessibility Updates: Shift-Tab, div order, and button names

pbblair:
Fixes:
1. Previously Safari couldn't shift-tab, and now it can.  Added a check so Pixi won't be tricked by false positives for mouse movements.
2. Previously the fallback logic was broken for buttons without titles or hints.  A correction and a few extra checks mean it should now work as intended on all major browsers.
3. Previously buttons were in reverse order in the DOM.  Braille displays make use of that order and would show buttons in backwards-order.  Flipped this around (also makes debugging a little more intuitive).

Enhancement:
3. One of the first things I tried to do with Pixi accessibility was to update a button name to reflect a button's state change, but it had no effect when I tried.  The official Pixi demo is a great example of this need!  Buttons are toggled to light up the signs, but there's no indication of the button's new state.  I've added this functionality.

Live demo at:
http://superkeen.com/web/pixi

To get as many browser/screenreader combinations as possible to announce the change as soon as its made, I've added some ARIA attributes.  I've done quite a bit of testing to reveal how well this is supported.  See below for details:

Full support:
Windows Chrome (Narrator and NVDA)
Windows Firefox (Narrator and NVDA)
Windows IE (NVDA)

Partial support:
Mac Safari (Voiceover): fine until the focus leaves and returns to the page, then button name changes not announced right away, only on focus change; recoverable by restarting Voiceover
Win/Mac/ChromeOS Chrome (ChromeVox): when toggling, sometimes announces the old button name instead of the new one

No button name change support (just like legacy behavior):
Mac Chrome (Voiceover)
Mac Firefox (Voiceover)
Windows Chrome (JAWS)
Windows Firefox (JAWS)
Windows IE (Narrator and JAWS)
Windows Edge (Narrator, NVDA, JAWS)


pbblair:
@ivanpopelyshev thanks for demonstrating where the null value was coming from.  You're right, no need to check for undefined.  Should be addressed in newest commit.

bigtimebuddy:
@pbblair Thank you so much for this high-quality PR and sorry we haven't been more responsive. The team is spread a little thin right now trying to put a big effort into v5. @GoodBoyDigital is going to look at this when he's free. I'm for getting this in v4 since it does a better job at provide more device/browser coverage for accessibility.

PR #4437, opened 2017-11-13
Extract fix (v5)

GoodBoyDigital:
It was busted, but not any more!


englercj:
I feel like some rebasing may need to happen here...

GoodBoyDigital:
eek! yes, that does not look right at all!

bigtimebuddy:
Bam! Just the wrong branch.

bigtimebuddy:
Needs linking, can you run “npm run lintfix”

packages/extract/src/webgl/WebGLExtract.js
```diff
@@ -180,9 +174,12 @@ export default class WebGLExtract
 
         if (renderTexture)
         {
-            textureBuffer = renderTexture.baseTexture._glRenderTargets[this.renderer.CONTEXT_UID];
+            // textureBuffer = renderTexture.baseTexture;//._glRenderTargets[this.renderer.CONTEXT_UID];
```
bigtimebuddy: Is this line okay to remove?

englercj: Yeah with this commented out I don't see where `textureBuffer` is assigned anymore?

bigtimebuddy: @GoodBoyDigital could you take a look at this. Is the comment-out a mistake?

GoodBoyDigital: tweaked! good catch chaps

PR #4442, opened 2017-11-14
same textures bound in same places

ivanpopelyshev:
A long time I knew the fact that SpriteRenderer wont like same texture bound in two locations.

But every time I thought that none of our plugins use the same textures as SpriteRenderer.

Now, https://github.com/pixijs/pixi-ui https://github.com/gameofbombs/pixi-heaven, https://github.com/pixijs/pixi-projection and possible others (i havent check sdf and neutrino) are forcing textures to be binded in specific locations, and SpriteRenderer wants to use same textures.

Thankfully, v5 Sprite Renderer doesn't have that issue because @GoodBoyDigital dropped the optimization that caused it.

Now, for the sake of plugins, for the sake of users who have WebGL: INVALID_ENUM: activeTexture: texture unit out of range that can be explained only by virtalBoundId=-1 , I want to merge this fix into 4.6.1 .

I dont want to make better algorithm, I just made a fix for algorithm that we use in v4. Better algorithms can be introduced in v5.

I and @cursedcoder tested the fix on a real project.

### DIFF
--- a/src/core/sprites/webgl/SpriteRenderer.js
+++ b/src/core/sprites/webgl/SpriteRenderer.js
@@ -241,9 +241,19 @@ export default class SpriteRenderer extends ObjectRenderer
         // copy textures..
         for (i = 0; i < MAX_TEXTURES; ++i)
         {
-            boundTextures[i] = rendererBoundTextures[i];
-            boundTextures[i]._virtalBoundId = i;
+            const bt = rendererBoundTextures[i];
+
+            if (bt._enabled === TICK)
+            {
+                boundTextures[i] = this.renderer.emptyTextures[i];
+                continue;
+            }
+
+            boundTextures[i] = bt;
+            bt._virtalBoundId = i;
+            bt._enabled = TICK;
         }
+        TICK++;
 
         for (i = 0; i < this.currentIndex; ++i)
         {


ivanpopelyshev:
I emulate third-party plugin by pushing same texture in two locations: 0 and 1. It's ok because people often need two samplers, even if they have the same texture inside.

Paste this in our examples (http://pixijs.io/examples/#/basics/basic.js):

var app = new PIXI.Application(800, 600, {backgroundColor : 0x1099bb});

// create a new Sprite from an image path
var bunny = PIXI.Sprite.fromImage('required/assets/basics/bunny.png')

// center the sprite's anchor point
bunny.anchor.set(0.5);

// move the sprite to the center of the screen
bunny.x = app.renderer.width / 2;
bunny.y = app.renderer.height / 2;

app.stage.addChild(bunny);
for (let i = 0; i<80;i++) {
   var txt = new PIXI.Text("text number "+i);
   txt.position.y = i%20 * 20;
   txt.position.x = Math.floor(i/20) * 200;
   app.stage.addChild(txt);
}


// Listen for animate update
app.ticker.add(function(delta) {
    // just for fun, let's rotate mr rabbit a little
    // delta is 1 if running at 100% performance
    // creates frame-independent tranformation
    bunny.rotation += 0.1 * delta;
    app.renderer.bindTexture(bunny.texture, 0, true);
    app.renderer.bindTexture(bunny.texture, 1, true);
});

src/core/sprites/webgl/SpriteRenderer.js
```diff
@@ -241,9 +241,19 @@ export default class SpriteRenderer extends ObjectRenderer
...
+            {
+                boundTextures[i] = this.renderer.emptyTextures[i];
+                continue;
+            }
+
+            boundTextures[i] = bt;
+            bt._virtalBoundId = i;
+            bt._enabled = TICK;
         }
+        TICK++;
```
finscn: Line 237  :  TICK++;
Line 256 :  TICK++;  

Is it Duplicated ?

ivanpopelyshev: No, its not.

finscn: I see. Thanks.

PR #4447, opened 2017-11-17
CJK Word Wrap

branch-henderson:
Added CJK Word Wrap functionality in TextMetrics along with tests for some helper functions.

ReadMe and package.json changes can be ignored. I just changed them so I can use this fork with my own project for now, until it hopefully gets merged with the official pixi.js release.


branch-henderson:
I've reverted package.json, README, and resolved the conflicts and committed as requested. It's still telling me I need to resolve conflicts, do I need to do something else to make the pull request aware of my recent changes?

themoonrat:
Hey @branch-henderson - I also had a CJK word wrap solution in my pixi version; a much hackier version than what you have, however, with no unit tests! https://github.com/themoonrat/pixi.js/tree/cjk-word-wrap
When we get time, might be worth looking at if we had any implementation differences. A plugin would be cool for v4, and then maybe we'll see about fitting it in to v5

branch-henderson:
Hi @themoonrat. Oh cool! Yeah, let's try to merge our ideas. I can look at that tomorrow. I didn't know there was a language setting LOL. Do you know how I'd go about turning this into a plugin? Is there any documentation on that?

themoonrat:
There isn't a language setting..... I added it as a way for my game to tell PIXI what language mode to be in, so that the code could then just work. Probably not the best way, but I wasn't expecting me to let people look at that horrible fudge of code ;)

The plugin can be a simple case of adding your own project with code that injects itself to within the PIXI namespace, or overriding existing prototype functions
https://github.com/pixijs/pixi-ui adds itself to the pixi namespace in index.js
https://github.com/pixijs/pixi-sound with typescript
https://github.com/themoonrat/pixi-tween a branch of a plugin I've updated; and you see their in it's index.js it adds a function to the graphics prototype

I think what we'd need here, is the CJK checking code being it's own util file(s), and then overriding the existing "wordWrap" prototype function within TextMetrics, with one that is nearly identical aside from the points required to check if CJK word wrapping is required

branch-henderson:
@themoonrat 

Thanks for the help. I'll try to follow those examples.

RE: "with one that is nearly identical aside from the points required to check if CJK word wrapping is required"

From my understanding the algorithm itself would have to be different from the original too since it would need to travel backward from the usual line break point to find the latest point which it can break cleanly, if any. If that's what you meant then that makes sense to me.

themoonrat:
Yeah... the whole function would be different, but we'd want to keep as much of it like pixi v4 behaviour as before, and since it's a decently size function, a fair bit will be copy pasted!

If you'd like me to help set up the repo then I'm more than happy to (I'd do it in the style of my pixi-tween as I got everything working as I liked it there!). But if you want to have a go, then I'll leave you to it :)

branch-henderson:
help would be appreciated, it would help me to learn by example. Thank you!

branch-henderson:
Do I have to give you some permissions to add to modify this pull request, or are you going to make your own? I'm not sure how collaboration works here.

bigtimebuddy:
Hey all, I thought about this some more and we need to close this PR since we are not going to merge any new major features into v4. Definitely please continue to collaborate on a plugin/mixin for v4 (I'm happy to setup a repo on **pixijs** for it if you need!).

themoonrat:
@branch-henderson when I get time (soon, I promise!) I'll set up a new repo with us having permission on it,  and we can work on the functionality as a plugin :)

branch-henderson:
Any update on this? @themoonrat

branch-henderson:
Sorry to bring this up again, have you had any time to look at this again @themoonrat ?

branch-henderson:
hi @themoonrat, the problem I see with #2 is: if the width is about to be exceeded, and the breakword does not obey CJK rules, it should still try to find an earlier point that will be acceptable and break there instead. From what I remember, that's why I had to rewrite some of the logic.

themoonrat:
I believe we've now got that scenario covered :)

branch-henderson:
In that case everything you said should be fine!

PR #4462, opened 2017-11-20
renderer resets vao correctly.

ivanpopelyshev:
Dont mind the branch name.


finscn:
There are many `renderer.bindVao(null)` in PIXI.
Shall we change all of them to `renderer._activeVao = null` ?

PR #4477, opened 2017-11-23
Region parameter for Renderer.generateTexture()

bma73:
Added an optional "region" parameter to Renderer.generateTexture() to only render the specified area.
https://jsfiddle.net/bma73/92os3a0e/17/

### DIFF
--- a/src/core/renderers/SystemRenderer.js
+++ b/src/core/renderers/SystemRenderer.js
@@ -244,11 +244,17 @@ export default class SystemRenderer extends EventEmitter
      * @param {PIXI.DisplayObject} displayObject - The displayObject the object will be generated from
      * @param {number} scaleMode - Should be one of the scaleMode consts
      * @param {number} resolution - The resolution / device pixel ratio of the texture being generated
+	 * @param {PIXI.Rectangle} region - The region of the displayObject, that shall be rendered
      * @return {PIXI.Texture} a texture of the graphics object
      */
-    generateTexture(displayObject, scaleMode, resolution)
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
I agree that we need such function. Can you change it that way "getLocalBounds()" are not called if region is specified?

I'd like to add that option in pixi-v5 ("next" branch). Sorry that we're very busy on that branch/ `dev4` wont be updated with new features

bma73:
I've updated it accordingly...

src/core/renderers/SystemRenderer.js
```diff
@@ -244,16 +244,17 @@ export default class SystemRenderer extends EventEmitter
      * @param {PIXI.DisplayObject} displayObject - The displayObject the object will be generated from
      * @param {number} scaleMode - Should be one of the scaleMode consts
      * @param {number} resolution - The resolution / device pixel ratio of the texture being generated
+	 * @param {PIXI.Rectangle} region - The region of the displayObject, that shall be rendered
```
bigtimebuddy: Documentation needs a little more love here. Should be `[region]` to denote that it's optional. Also, I would add a note that it default to the bounds of the displayObject. Otherwise 👍

ivanpopelyshev:
Suddenly, a feature got merged in `dev` :)

bigtimebuddy:
@ivanpopelyshev Yes, but this was pretty low-impact, high-value and didn't breaking anything on **next** so it was easy to merge in both places.

PR #4487, opened 2017-11-27
obserbvable_point: public subscribe

axldns:
More a feature request than issue. Resolves https://github.com/pixijs/pixi.js/issues/4486
Back in the times of PIXI 3.x, it was easy to follow position / scale / anchor / pivot of particular display object in dispersed solutions. Simple assignment `b.position = a.position`, regardless of good practices conduct, allowed to stick and unstick display objects easily.

Re-designed transformation system in PIXI 4.x put kibosh on it, and among other benefits, made tracking/following changes in position, scale, anchor, pivot as easy and faulty as _per frame check_ or _hacky as duck_.

Can we make this transparent and more friendly? 
If pub-sub + link methods are too much, coud we at least give pub sub to ObservablePoint? The code and tests from PR are already split to two sections.

### DIFF
--- a/src/core/math/ObservablePoint.js
+++ b/src/core/math/ObservablePoint.js
@@ -21,6 +21,10 @@ export default class ObservablePoint
 
         this.cb = cb;
         this.scope = scope;
+        this.set = this.set.bind(this);
+        this.runCallbacks = this.runCallbacks.bind(this);
+        this._listeners = [];
+        this._broadcasters = [];
     }
 
     /**
@@ -39,7 +43,7 @@ export default class ObservablePoint
         {
             this._x = _x;
             this._y = _y;
-            this.cb.call(this.scope);
+            this.runCallbacks();
         }
     }
 
@@ -54,7 +58,7 @@ export default class ObservablePoint
         {
             this._x = point.x;
             this._y = point.y;
-            this.cb.call(this.scope);
+            this.runCallbacks();
         }
     }
 
@@ -73,7 +77,7 @@ export default class ObservablePoint
         if (this._x !== value)
         {
             this._x = value;
-            this.cb.call(this.scope);
+            this.runCallbacks();
         }
     }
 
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
@axldns, Thanks for this pull-request. The unit-tests are ace!

I see how this could be useful for handling point changes. My problem with this is it's a lot of code. Why not use an EventEmitter? It's already in PixiJS and  you don't have to build all this plumbing to call and subscribe listeners.

```js
var obj = new DisplayObject();
obj.position.on('change', (pt) => {
   console.log(pt.x, pt.y);
});
var obj2 = new DisplayObject();
// obj2 follows obj
obj.position.on('change', (pt) => obj2.position.set(pt.x, pt.y));
```
You could create a mixin plugin to do your link/unlink
```js
ObservablePoint.prototype._onChange = null;
ObservablePoint.prototype.link = function link(pt) {
  pt._onChange = function(pt) {
    this.set(pt.x, pt.y);
  };
  this.on('change', pt._onChange, pt);
};
ObservablePoint.prototype.unlink = function unlink(pt) {
  this.off('change', pt._onChange, pt);
  pt._onChange = null;
};
```
The advantage is we can also treat `cb` like all the other listeners, because EventEmitter supports scope. So the constructor might look like this:

```js
constructor(cb, scope, x = 0, y = 0)
{
    this._x = x;
    this._y = y;

    this.on('change', cb, scope);
}
```

Alternatively, we could use **mini-runner** instead of EventEmitter. There's a growing consensus in the core PixiJS team that we should move away from events and move towards a signal-like system. I digress...

The timing isn't awesome on this, I'm afraid, because we are mostly closing v4 except to critical features and bugs. A pub-sub point doesn't fit into those categories, so this most likely has a better shot at v5, currently being worked on in the **next** branch. It'd be better to focus your efforts there.

axldns:
Many thanks for your interest and suggestions @ivanpopelyshev and @bigtimebuddy 
We're more than happy with just exposing an option to listen changes in OP.
I can see your point with number of new lines to maintain, therefore I really appreciate the solution you proposed. I followed it in the [latest commit](https://github.com/axldns/pixi.js/commit/60a3de61ce97bba904c162bf1a44e6154f0ee3d5), but as to the constructor... 
Despite the fact
```
constructor(cb, scope, x = 0, y = 0)
{
    super()
    this._x = x;
    this._y = y;

    this.on('change', cb, scope);
}
```
 looks sweet and is ultra-consistent imho, unfortunately calling **`pt.removeAllListeners()`** would left instance detached from original context (TransformStatic). Without keeping callback and scope's reference (as it was) it'd be pretty much broken / unusable. This is the reason for which I decided to keep constructor passed callback separated from the pub-sub callbacks.

Once again thanks for your support.
For now we're still migrating games platform from 3.x to 4.x and trying to straighten up nifty hacks arisen on our side, but I promise to have a go with pixi-next soon :)

bigtimebuddy:
I chatted about this with the core team and we decided that this isn't appropriate for v4 and I'll be closing. We are trying to make sure v4 stays as stable as possible while we are on v5. Because ObservablePoint is a hot code area (performance critical), we think it's wise to be extra cautious with new features that impact these areas. However, I think the solution we came up with is a good idea. I'm going to tag this as future work we hope to incorporate into v5. Thanks for your help @axldns.

PR #4489, opened 2017-11-27
 Optimize about `gradient of Text`

finscn:
* Cache `gradient`
* check whether canvas's size changed
* If `canvas's size` or `style.fill` changed , re-create gradient, or use the cached one.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -144,15 +144,15 @@ export default class Text extends Sprite
         const maxLineWidth = measured.maxLineWidth;
         const fontProperties = measured.fontProperties;
 
-        const prevTextWidth = this.textWidth;
-        const prevTextHeight = this.textHeight;
+        const prevCanvasWidth = this.canvasWidth;
+        const prevCanvasHeight = this.canvasHeight;
 
-        this.textWidth = Math.ceil(width + (style.padding * 2));
-        this.textHeight = Math.ceil(height + (style.padding * 2));
-        this._sizeChanged = this.textWidth !== prevTextWidth || this.textHeight !== prevTextHeight;
+        this.canvasWidth = Math.ceil(width + (style.padding * 2));
+        this.canvasHeight = Math.ceil(height + (style.padding * 2));
+        this._sizeChanged = this.canvasWidth !== prevCanvasWidth || this.canvasHeight !== prevCanvasHeight;
 
-        this.canvas.width = this.textWidth * this.resolution;
-        this.canvas.height = this.textHeight * this.resolution;
+        this.canvas.width = this.canvasWidth * this.resolution;
+        this.canvas.height = this.canvasHeight * this.resolution;
 
         context.scale(this.resolution, this.resolution);
 
@@ -462,8 +462,8 @@ export default class Text extends Sprite
         let currentIteration;
         let stop;
 
-        const width = this.textWidth;
-        const height = this.textHeight;
+        const width = this.canvasWidth;
+        const height = this.canvasHeight;
 
         // make a copy of the style settings, so we can manipulate them later
         const fill = style.fill.slice();
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -147,8 +147,8 @@ export default class Text extends Sprite
         const prevTextWidth = this.textWidth;
         const prevTextHeight = this.textHeight;
 
-        this.textWidth = Math.ceil((width + (style.padding * 2));
-        this.textHeight = Math.ceil((height + (style.padding * 2));
+        this.textWidth = Math.ceil(width + (style.padding * 2));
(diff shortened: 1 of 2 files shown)


src/core/text/Text.js
```diff
@@ -144,8 +144,15 @@ export default class Text extends Sprite
         const maxLineWidth = measured.maxLineWidth;
         const fontProperties = measured.fontProperties;
 
-        this.canvas.width = Math.ceil((width + (style.padding * 2)) * this.resolution);
-        this.canvas.height = Math.ceil((height + (style.padding * 2)) * this.resolution);
+        const prevTextWidth = this.textWidth;
```
themoonrat: I'd call these 'previousCanvasWidth' and 'previousCanvasHeight' as that is what we are actually measuring, not the width / height of the text itself

src/core/text/Text.js
```diff
@@ -144,8 +144,15 @@ export default class Text extends Sprite
...
+        const prevTextHeight = this.textHeight;
+
+        this.textWidth = Math.ceil(width + (style.padding * 2));
```
themoonrat: there's no need to create 2 new member variables here, when you could just check against this.canvas.width and this.canvas.height

src/core/text/Text.js
```diff
@@ -144,8 +144,15 @@ export default class Text extends Sprite
...
+        this.textHeight = Math.ceil(height + (style.padding * 2));
+        this.sizeChanged = this.textWidth !== prevTextWidth || this.textHeight !== prevTextHeight;
```
themoonrat: this._canvasSizeChanged for clarity on what size we are actually tracking

src/core/text/TextStyle.js
```diff
@@ -284,6 +284,7 @@ export default class TextStyle
         if (this._fill !== outputColor)
         {
             this._fill = outputColor;
+            this._gradient = null;
```
themoonrat: First, there's the issue that the gradient could also be effected by the fillGradientStops and also fillGradientType changing

Second, it doesn't feel right that the text style class, which describes how a style should look, is suddenly holding information that the Text class has generated (the gradient). I feel that the Text class itself should store any previously generated fillGradient. 

Could it work where we have a this.fillStyleID that sits along a styleID? The text class could check against that fillStyleID having changed, and know if to re-generate a gradient that way?

src/core/text/Text.js
finscn: I create 2 new member variables , because they could reduce  two `/ this.resolution` operations in `_generateFillStyle`.

src/core/text/TextStyle.js
finscn: use fillStyleID/styleID will make the code become complex.
Now juse 3 places would call `this._gradient = null`.

I think in the future , if there are more places will call it , 
We could try to refactor and use   `fillStyleID/styleID`.

themoonrat:
@finscn I think where we got with this in the Slack channel, was, that it's my opinion, compared to the cost of regenerating a texture, this change will make a negligible difference, and at the expense of more confusing code (Text now saves stuff to TextStyle whereas previously Text would only use TextStyle as a source of information).

We really need our own jspef for PIXI, so that we can benchmark changes like this to see if they're really worth it.
I created a test: https://jsbench.me/bojc6f56g4/1
And ran it on my aweful work phone (Galaxy J5) on Chrome and Firefox

chrome
firefox

There is a big difference on Firefox (but so fast, it's not worth caring about) and only a smaller difference on Chrome.

My personal preference is that we keep things as they are for v4. It's in maintenance mode, and this isn't a bug fix; it's a small fraction of a ms saving when changing text.

For v5 though, it might be worth taking a look; and finding a better place to save off stored gradients that isn't back in TextStyle

finscn:
In my game ( when open  ui , there are many texts ) , the performance is from 30FPS to 40FPS (on my android) .
I think it's not ` a small fraction of a ms saving`.

ivanpopelyshev:
I think its ok to merge this thing.

themoonrat:
If others wanna go for it, then I'd accept it if gradient didn't have this undeclared _gradient property. For me, underscored means don't access externally. So declare it in the constructor of TextStyle as gradient, document why it's there, and stick a @private js doc over it to not appear as part of the public API

themoonrat:
PS. In what scenarios were you saving generating this gradient? Because if the size of the canvas changes, then the gradient has to be re-created. Changing text from one string to another will pretty much always change the size requirements of the canvas, won't it?

bigtimebuddy:
Going to close this because we need to focus on larger performance issue with Text for v5. This is a minor performance improvement. If users are hitting this as a performance issue on low end devices, there's 2 workarounds: use BitmapText or disable the use of gradients on these devices.

PR #4490, opened 2017-11-28
Spritesheet: Use frame height and width if not trimmed

nthapaliya:
Some [tools](https://github.com/urraka/texpack) will omit `sourceSize` if `trimmed = false`. This, I guess, is to eliminate superfluous information and make the outputted json file smaller.

If that is the case, we can simply use the `frame.h` and `frame.w` as sourceSize values.


ivanpopelyshev:
Some tools also omit "trimmed=true". I dont know what to do about it :)

bigtimebuddy:
@nthapaliya could you resolve the linting errors https://travis-ci.org/pixijs/pixi.js/builds/308236020#L654-L655

You can run the lint locally by running `npm run lint`.

Also, to @ivanpopelyshev's comment, you might tweak the ternary operator a bit to include instances where `trimmed` might be `undefined`:

```js
const sourceSize = this._frames[i].trimmed === false ?
    this._frames[i].frame :
    this._frames[i].sourceSize;
```

nthapaliya:
@ivanpopelyshev I'm not sure I understand about the app you posted. Its a photo backup tool. I installed (and created an account) and uploaded a few photos. Now what?

@bigtimebuddy I can fix the linting issues. The reason for opening the PR is this: if `this._frames[i].sourceSize` is undefined, then pixi throws an error, everything breaks. So I think i'd rather change the ternary to:
```
const sourceSize = this._frames[i].sourceSize
  ? this._frames[i].sourceSize
   : this._frames[i].frame;
```

We're ignoring 'trimmed'. It might be true, false, or undefined. We only care if `sourceSize` exists. If it exists, use it. If not, use `frame`, because you might as well use something.

Does that sound reasonable?

ivanpopelyshev:
I agree to @bigtimebuddy code. That will satisfy everyone. 

```
const sourceSize = this._frames[i].trimmed === false ?
    this._frames[i].frame :
    this._frames[i].sourceSize;
```

nthapaliya:
Sample json files generated by texpack (I'm trying to accommodate both).

case: trimmed = true
```
{
	"frames": {
		"utility_7": {
			"frame": {
				"x": 4,
				"y": 4,
				"w": 68,
				"h": 56
			},
			"trimmed": true,
			"spriteSourceSize": {
				"x": 35,
				"y": 24,
				"w": 68,
				"h": 56
			},
			"sourceSize": {
				"w": 136,
				"h": 104
			}
		},
}
```

case: trimmed = false, i.e trimmed and sourceSize completely omitted
```
{
	"frames": {
		"kiosk_marker": {
			"frame": {
				"x": 0,
				"y": 0,
				"w": 348,
				"h": 256
			}
		},
}
```

bigtimebuddy:
Awesome, great work @nthapaliya. Would it be possible to create unit-test for these different data output? I want to make sure that we are clear about what types of output we support and that we don't break it in the future.

nthapaliya:
@bigtimebuddy I've tweaked the existing test data to include all the test cases we want (in this PR specifically).

Let me know if I need to go about this a different way.

bigtimebuddy:
Hmm... I actually think we need a separate test for this @nthapaliya. We should make sure the Textures (width/height particularly) are representative of the data being parsed for the Spritesheet. No errors being thrown doesn't mean the Texture was created correctly. My preference would be to write a separate test with that data you specified.

ivanpopelyshev:
Make a test where `trimmed` is undefined, in  that case we need to check if sourceSize exists and if its differ from original

bigtimebuddy:
@ivanpopelyshev so only `frame` and `sourceSize` fields?

ivanpopelyshev:
Shoebox: https://renderhjs.net/shoebox/

ivanpopelyshev:
It would make me happy if you check it and confirm. I had that problem this summer, that's why I asked to modify this PR back in december.

bigtimebuddy:
@ivanpopelyshev okay, I added a test for shoebox's output. Could you review

bigtimebuddy:
Thank you @nthapaliya

Issue #4448, opened 2017-11-17
wrong canvas width calcuated when some emojis in the text content.


jcyuan:
I think there should be ways to tell emoji characters apart from common characters in TextMetrics.wordWrap method as each emoji character actually has 2x chars so it caused the final width to break lines increased wrongly.
here is my demo screenshot:
![code](https://user-images.githubusercontent.com/850854/32941205-f826daa0-cbbf-11e7-8ecf-6023409a58af.jpg)
an user's name with emojis : "変な鬼"☘️🍀🍃🌸🌼 
unicode ("変な鬼"\xE298\x98EF\xB88F\xF09F\x8D80\xF09F\x8D83\xF09F\x8CB8\xF09F\x8CBC)

![code2](https://user-images.githubusercontent.com/850854/32941222-042b6668-cbc0-11e7-9bfe-a154daf97e88.png)
the wrapper chops its content off as the content(textfied) has a wrong width.

jcyuan:
I can make a PR but I don't know if there is a best way to recognize emojis from a string?

jcyuan:
PR submitted.

PR #4421, opened 2017-11-10
fix bug for adding useless extra blank ahead of the first text line [#4251]

jcyuan:
original code is in a wrong way to add blank character when the style.breakWords is enabled.
see also issue: [#4251](https://github.com/pixijs/pixi.js/issues/4251)

PR #4493, opened 2017-11-28
fix bug: #4448

jcyuan:
if emojis in the text the width to wrap words will be a wrong value.
see fiddle based on which @bigtimebuddy created for #4421:  [https://jsfiddle.net/gpxow75d/2/](https://jsfiddle.net/gpxow75d/2/)

besides, I think the method isEmojiChar should be placed somewhere else but not in TextMetrics.js.

### DIFF
--- a/src/core/text/TextMetrics.js
+++ b/src/core/text/TextMetrics.js
@@ -88,6 +88,42 @@ export default class TextMetrics
             fontProperties
         );
     }
+    
+    /**
+     * check whether a string is an emoji character or not.
+     * @param charCode the byte to test
+     * @param nextCharCode the possible second byte of the emoji
+     * @returns 0 means not a emoji, 1 means single byte, 2 means double bytes
+     */
+    static isEmojiChar(charCode:number, nextCharCode:number):number {
+        const hs = charCode;
+        const nextCharValid = typeof nextCharCode == "number" && !isNaN(nextCharCode) && nextCharCode > 0;
+        if (0xd800 <= hs && hs <= 0xdbff) {
+            if (nextCharValid) {
+                const uc = ((hs - 0xd800) * 0x400) + (nextCharCode - 0xdc00) + 0x10000;
+                if (0x1d000 <= uc && uc <= 0x1f77f)
+                    return 2;
+            }
+        } else if (nextCharValid) {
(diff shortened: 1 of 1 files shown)


jcyuan:
ooops..... 
what does  "warning  Expected literal to be on the right side of <=        yoda" mean?
line 104: if (0xd800 <= hs && 0xdbff >= hs)

englercj:
`if (0x1d000 <= uc && 0x1f77f >= uc)` is what is known as a "yoda condition" in that it is backwards from what most people would expect (like how yoda talks). In English what you are testing is that "uc is greater than 0x1d000 and uc is less than 0x1f77f", so write the condition the same way you would say it outloud:

`if (uc >= 0x1d000  && uc <= 0x1f77f)`

jcyuan:
@englercj got it, thank you.

cursedcoder:
Good enough.

TextMetrics can be refactored later as it is only used under the hood via Text entity.

src/core/text/TextMetrics.js
```diff
@@ -89,6 +89,50 @@ export default class TextMetrics
...
     }
 
+    /**
+     * Check whether a byte is an emoji character or not.
+     *
+     * @param {number} charCode - the byte to test.
+     * @param {number} nextCharCode - the possible second byte of the emoji.
+     * @return {number} 0 means not a emoji, 1 means single byte, 2 means double bytes.
+     */
+    static isEmojiChar(charCode, nextCharCode)
```
themoonrat: Overall, lgtm.
1. I'd wonder, as an API that could be used externally, whether "charCode" and "nextCharCode" could be "firstCharCode" and "secondCharCode". I know internally, where it's used, it's the current one and the next one, but I could see it being useful to fire 2 characters at this static function in my games.
2. For other pixi peeps - I'd want https://github.com/pixijs/pixi.js/pull/4545 merged in before this please

bigtimebuddy:
@jcyuan Thanks for your patience here while we've been head-down working on v5. Would like to see this in but like @themoonrat said, will need to merge #4545 first, but we'll consider merging this for our next minor release.

jcyuan:
@bigtimebuddy sure thing sir, no problem.. sorry missed your notice.

PR #4503, opened 2017-11-30
Next texture resource cleanup (v5)

bigtimebuddy:
### Changed

* Create `PIXI.resources` namespace for all resources
* Removes IResource & TextureResource replaces with Resource and moves functionality up
* Adds lots of documentation for everything
* Property deprecated `fromCanvas` `fromImage` and `fromSVG` on BaseTexture
* Removes all the `from` static methods on Resources, to make more flexible constructors


ivanpopelyshev:
setRealSize should accept resolution because its useful for Text and in some other cases I forgot about.

ivanpopelyshev:
> Removes IResource & TextureResource replaces with Resource and moves functionality up

This division exists because there will be multi-texture resources. TextureResource is a single texture resource.

bigtimebuddy:
Closing this for now so I can retool based on our conversation this morning.

PR #4502, opened 2017-11-30
Texture resource update (v5)

ivanpopelyshev:
Huge update on texture APIs.

Work in progress.

Problems:
- Decide whether we should use 'loaded'/'hasLoaded' or 'valid'/'validate' in BaseTexture. Its the same thing.
- Framebuffers do not have resolution? wtf?
- Fix Rope example. It broke in `next` too
- Texture becomes validated before createImageBitmap() runs, we can see frame or two of empty texture. What is expected behaviour?
- Check CubeTexture

TODO:
- Example of createBitmap feature
- Example for compressed textures

PR #4507, opened 2017-12-01
Texture and Resources Refactor (v5)

bigtimebuddy:
## Overview 

This PR Builds on the work of #4502 & #4503.

Essentially, this pull-request refactors how BaseTexture handles different resource types, such as Images, Videos, SVG images, etc. It cleans up previous work related to resources and makes the resource system more extensible. 

### Changed

Here are the list of changes that diverge from previous iterations of this refactor.

* This decouples BaseTexture from the resources by using **mini-runner** instead of sending along a references, this allow resources to be shared by multiple BaseTexture objects. Uses a ref-counter and `bind` `unbind` methods in resource to attach to BaseTextures
* Adds an `autoDetectResource` method for creating resources. Custom resources types can be added to auto-detection of other sources.
* Conforms the function signature of all Resources (source, options)
* Renames the `from` methods on BaseTexture for creating buffer sources
* More documentation and unit-tests

### Documentation 

https://pixijs.download/next-texture-resource2/docs/PIXI.resources.html

### Build 

* https://pixijs.download/next-texture-resource2/pixi.js
* https://pixijs.download/next-texture-resource2/pixi-legacy.js


ivanpopelyshev:
BaseTexture changes are interesting, but they are not simple. It looks like we move all our problems inside an object and think that `Object.assign` will solve the order of control for us.

With this code we would not say to user `look at the source` if he stumbles across a problem in a custom resource or texture. I could say more, but I want to sleep :)

bigtimebuddy:
@ivanpopelyshev yes, please explain why you don't want options for the BaseTexture. In v4, one can set scaleMode and resolution from the constructor. This isn't really any different than overriding other things that developers can already set, like wrapMode, etc.

ivanpopelyshev:
Just a reminder of wrong things from my perspective:

- Usage of `Object.assign` and options. We know what those fields are doing, we can force user to use `fluent interface` instead. API has to reflect internal mechanisms.

Resource also has options, and that's sad.

- onRealSize, onUpdate, onLoaded. I really dont see the difference for those in our use-cases. `update` is enough.

My invariant for BaseTexture: "its valid only if width and height are greater than zero.". Code reflects that, `update` function.

If you want separate those events, please provide use-cases, imaginary or real, we can talk about it. Yes, we can add another invariant, " hasLoaded becomes true when resource bitmap is ready for upload". Current implementation of `loaded` function is very strange. Not every resource has to be loaded.

We really need invariants or pre/post conditions there, tests only reflect those things. If you dont have that concept in mind or in paper, tests wont help, there'll be a fight over intended behaviour.

- setRealSize has no "resolution" parameter.

It has use-case: create a canvas of particular size and then create a texture of (canvas.width, canvas.height, resolution). Sometimes users deal with real pixels, sometimes with CSS, resolution can be set anyway. Why trigger update event twice?

ivanpopelyshev:
Regarding options, I understand that 

```js
const resource = new BufferResource(data, { width, height });
``` 

really looks cool. I have nothing against it on app-level usage. Can we please be more ascetic in the core?

bigtimebuddy:
@ivanpopelyshev I'm not finished yet. When I am, I'll ping you for a review.

ivanpopelyshev:
Its ok, I'm just making sure that all my points are somehow documented :)

bigtimebuddy:
Just a note to myself:

* I've removed the ability to set custom params/tag for resources from Ivan's original PR
* I've removed support for sharing resources between BaseTextures (ref-counter)

These are edge-cases to support more complex resource types and should be added later to minimize what's needed in this PR.

ivanpopelyshev:
Exactly.

I liked the idea of ref-counter but we really need more time to investigate it, something that we dont have right now :( 

The idea is to make those things possible to investigate through plugins for v5.

bigtimebuddy:
Okay, couple of changes:
* Removes the usage of Object.assign based on conversation and compromise with @ivanpopelyshev , but kept the options
* Changes `setStyle(options)` to `setStyle(scaleMode, mipmap)`
* Removes initial constructor `this.setSize` call

bigtimebuddy:
@ivanpopelyshev this is ready for review.

bigtimebuddy:
Just FYI, about function signatures:

### Auto-Detectable

These have compatible constructor signatures and can be created via `BaseTexture.from` or `autoDetectResource`.

* `CanvasResource(source:HTMLCanvasElement)`
* `ImageResource(source:HTMLImageElement|string, options?)`
* `SVGResource(source:string, options?)`
* `VideoResource(source:HTMLVideoElement|string, options?)`
* `BufferResource(source:Float32Array|Uint8Array, options?)`

### Non Auto-Detectable

These resources have incompatible constructor signatures and cannot be auto-detected based on the source.

* `CubeResource(source:any[], width:number, height:number, options?)`
* `ArrayResource(source:any[], width:number, height:number, options?)`

ivanpopelyshev:
I cant believe it!

PR #4512, opened 2017-12-04
Let PIXI.Shader support `precision & attributeLocations`

finscn:
Now PIXI has updated the minimal version of pixi-gl-core.
So it could do something for  custom precision & attribute-locations.
and this PR does NOT break any existing case.


finscn:
The document is copy from pixi-gl-core's GLShader .
Maybe there are some typo errors.

Issue #4534, opened 2017-12-13
CanvasMeshRenderer: display list transparency bug


kvint:
Hello there!

Looks like mesh on canvas inherits transparency from previously drawn sprite and can't be changed separately (works fine on WebGL though).
 
Can be reproduced on http://pixijs.io/examples/#/basics/textured-mesh.js
1) add `forceCanvas: true` into application
2) put some transparent object under the mesh:
```
var alphSprite = PIXI.Sprite.from(PIXI.Texture.fromImage('required/assets/snake.png'))
alphSprite.alpha = 0.5;
app.stage.addChild(alphSprite);
app.stage.addChild(strip); // mesh object (Plane or Rope)
```
As a result: the mesh becomes transparent as a sprite underneath.
![preview](https://user-images.githubusercontent.com/621317/33930107-e55374b4-dffc-11e7-9db5-4636974ea167.png)

kvint:
Have any workarounds on this? Sorry if it's wrong place to ask.

ivanpopelyshev:
Seems like a bug.

https://github.com/pixijs/pixi.js/blob/dev/src/mesh/canvas/CanvasMeshRenderer.js#L28 It doesnt set alpha.

We need to add `context.globalAlpha = (mesh.worldAlpha` somewhere.

Workaround

```js
PIXI.mesh.CanvasMeshRenderer.prototype.render = function(mesh)
    {
        var renderer = this.renderer;
        var context = renderer.context;

        var transform = mesh.worldTransform;
        var res = renderer.resolution;

        if (renderer.roundPixels)
        {
            context.setTransform(
                transform.a * res,
                transform.b * res,
                transform.c * res,
                transform.d * res,
                (transform.tx * res) | 0,
                (transform.ty * res) | 0
            );
        }
        else
        {
            context.setTransform(
                transform.a * res,
                transform.b * res,
                transform.c * res,
                transform.d * res,
                transform.tx * res,
                transform.ty * res
            );
        }

        renderer.setBlendMode(mesh.blendMode);
        context.globalAlpha = mesh.worldAlpha;

        if (mesh.drawMode === Mesh.DRAW_MODES.TRIANGLE_MESH)
        {
            this._renderTriangleMesh(mesh);
        }
        else
        {
            this._renderTriangles(mesh);
        }
    }
```

kvint:
Wow, thanks a lot. I'll give it a try

bigtimebuddy:
[...]

PR #4535, opened 2017-12-13
Canvas mesh alpha. Fix for #4534

ivanpopelyshev:
#4534


themoonrat:
I'm taking it the context.globalAlpha will get reset elsewhere in the next canvas render function it hits?

ivanpopelyshev:
Yep. It will also be reset after whole stage is rendered.

Issue #4544, opened 2017-12-19
TextMetrics.wordWrap() ignores TextStyle.letterSpacing


brenwell:
Hey everyone, I found an issue with `TextMetrics.wordWrap`, which also effects `TextMetrics.measureText` & `TextMetrics.measureFont`. The wordwrap function does not account for letterSpacing values larger than 0.

**jsfiddle**

https://jsfiddle.net/brenwell/59660ayb/

I will make a pull request with the fix

brenwell:
I found that my pull request has improved the calculation but not completely fixed it. I am working on a slightly improved version. 

Also want to confirm that this is a bug that needs fixing and not an issue with how I was using TextMetrics??

HansKaufmann:
+1

Esshahn:
+1

brenwell:
I have also noticed that the dropshadow is not factored into the calculation of TextMetrics. Am I correct in assuming that this is intentional?

brenwell:
I have updated my pull request. It’s calculation is more accurate now. 

Would be great if I can get some feedback.

brenwell:
Link with changes https://jsfiddle.net/brenwell/59660ayb/1/

themoonrat:
Closing as I think you're PR fixed this if I remember correctly :)

PR #4545, opened 2017-12-19
TextMetrics.wordWrap() ignores TextStyle.letterSpacing Issue #4544

brenwell:
TextMetrics.wordWarp() ignores TextStyle.letterSpacing 

This is to *hopefully* resolve issue #4544

I found an issue with TextMetrics.wordWarp, which also effects TextMetrics.measureText & TextMetrics.measureFont. The wordwrap function does not account for letterSpacing values larger than 0.

*The problem* - jsfiddle

https://jsfiddle.net/brenwell/59660ayb/


brenwell:
Just a quick ping to see if someone can check this issue. @bigtimebuddy @themoonrat @ivanpopelyshev

src/core/text/TextMetrics.js
```diff
@@ -103,86 +108,139 @@ export default class TextMetrics
...
+
+            return width;
+        }
+
+        // convienience function for logging each line
+        function addLine(l, newLine = true)
+        {
+            l = (newLine) ? `${l}\n` : l;
+
+            // console.log(l)
```
themoonrat: remove this commented out console.log

src/core/text/TextMetrics.js
```diff
@@ -103,86 +108,139 @@ export default class TextMetrics
...
+        // ideally there is letterSpacing after every char except the last one
+        // t_h_i_s_' '_i_s_' '_a_n_' '_e_x_a_m_p_l_e_' '_!
+        // so for convenience the above needs to be compared to width + 1 extra space
+        // t_h_i_s_' '_i_s_' '_a_n_' '_e_x_a_m_p_l_e_' '_!_
+        // ________________________________________________
+        // And then the final space is simply no appended to each line
+        const wordWrapWidth = style.wordWrapWidth + style.letterSpacing;
+
+        // store the width of calculated character(s)
+        function getFromCache(key)
```
themoonrat: I have a personal preference for having these functions stored outside of this static wordWrap function, so that they are not re-created every time the wordWrap function is called

brenwell:
Alright, thanks for the reply & no sweat about the delay. I have made all the changes you suggested by whilst doing that I noticed something else. Pixi word-wrap doesn't behave like css does, is that intentional? Notice "Supercal..." starts on a new line with css

***Pixi**
screen shot 2018-01-22 at 06 57 16


**Css**
screen shot 2018-01-22 at 06 57 12


```html
<div style="
    width: 200px;
    font-size: 20px;
    font-style: italic;
    font-family: arial;
    font-weight: bold;
    word-wrap: break-word;
    letter-spacing: 4px;
">Pixi.js - The HTML5 Creation Engine. Create beautiful digital content with the supercalifragilisticexpialidociously fastest, most flexible 2D WebGL renderer.</div>
```

themoonrat:
@brenwell Gave it another hammering today and it all looks as I'd expect :+1: 

As for the pixi and css difference... I don't believe it's intentional, no, but I wouldn't let it hold back this PR, as that's just how pixi has always worked. But it might be something to consider changing for v5 if others agree.
It appears as if the css strategy is "if breakwords is on, and a long word will be broken, it will start that string on a new line". Whereas pixi seems to act as if there is no concept of words at all when breakwords is on, and just break where it needs to.

brenwell:
Ok great. Hopefully we can get it merged soon. I can make another pull request for the CSS behaviour if you like. 

Thanks @themoonrat

themoonrat:
I think the CSS behaviour would be a good PR after this is in, yeah, thanks :)

brenwell:
Ok will do. This hasn't been merged yet I believe, what is the process for merging it, I would really love to have this change available. Is it just a waiting game??

brenwell:
Yeah!

bigtimebuddy:
Sorry for the delay @brenwell. Your PR is great, we've just been focused on v5 and super busy. Also, there are also a handful of other Text PRs that I know would be stomped on by this one. I think this is the better PR to merge first because it fixes a bug, the others add new features and APIs (emoji support, change metrics) which should probably be lower priority.

brenwell:
Understandable

One of the other things this issue highlighted was that there is a difference in behaviour between pixi js and css regarding word-wrapping. Would you guys want the behaviour to be like css and if so should I make a pull-request. The details can be seen in previous comments

bigtimebuddy:
We are saying no to any to major refactors in v4. Not exactly sure what you had in mind for behavior to be like CSS, but this effort sounds like it would be better focused on v5 (**next** branch). Text is v5 is going to get overhauled anyways. Two of the biggest issues with Text are performance (caching of characters to dynamic atlas) and proper kerning. We've been talking with a few folks on PixiJS Slack about this, would love your input.

themoonrat:
@bigtimebuddy fyi: the behaviour change here would just be 'when does text wrap if it's too long on a line'
With PIXI, it goes
```
here's some text andthispartistoo
long so will wordwrap. thank you
have fun
```

css style it would go
```
here's some text
andthispartistoolong so will
wordwrap. thank you have fun
```

So if piece of text is going to have to be wrapped, css/html style is to start that wrapped word on a new line. This has the advantage of potentially stopping the word requiring being broken up in the first place, as it gets a fresh new line to take up space on.

bigtimebuddy:
Ahh, thanks for clarifying @themoonrat. I'm okay with that approach. What do you think?

themoonrat:
@bigtimebuddy You're ok with keeping the current approach or ok with changing to the proposed approach? 
I'd like a future pr for the proposed approach because, not only does it match the expectations of how word wrapping works by matching how html text works, but it's just a plain better looking solution.

An argument could be made it should go onto v5 not v4 as it _will_ change the visuals of existing content.... but then we're not even sure what Text we're having for v5 atm!

bigtimebuddy:
I’m okay with that change for v4, I agree it looks better and is more consistent with expectations.

brenwell:
Ok no problem, when I get a moment I will make a new pull request for that particular behaviour.

Regarding the text overhaul, its my understanding the kerning is really quite hard to do right. I too would be interested in what others have as ideas for that

PR #4547, opened 2017-12-19
Round one of removal of GL core, general tidy and refactor of GLTexture

GoodBoyDigital:
- removed filters from build (until fixed)
- removed particles from build (until fixed)
- deleted RenderTarget (no longer required)
- deleted GLTexture (reduced to a small class in Texture System,
similar to Buffers)
- removed pixi-gl-core dependency.
- moved upload code from GLTexture to the Resources. This is a better
home for it


ivanpopelyshev:
Gonna look at it !

ivanpopelyshev:
One of problems that i see from afar - no "width" and "height" in GLTexture. We had code in GLTexture.upload that allowed to use subdata instead of data if width and height are the same. Its not dirty id, its like id that changes only on width/height change.

We can move it to BaseImageResource.

packages/core/src/renderers/systems/textures/TextureSystem.js
```diff
@@ -1,7 +1,34 @@
 import WebGLSystem from '../WebGLSystem';
-import GLTexture from './GLTexture';
 import { removeItems } from '@pixi/utils';
 
+class GLTexture
```
bigtimebuddy: Preference would be to externalize GLTexture to it's only file since it's not just an internal class and is shared. Also, need at least a class definition here. What about a `destroy` method which would take care of `texture._glTextures[this.renderer.CONTEXT_UID].texture.destroy();` below, but you could also nullify the `texture` property.

bigtimebuddy:
One organizational comment about GLTexture, otherwise looks good

Issue #4046, opened 2017-05-20
lineTo with a duplicate point renders a gap in a line drawing.


XerxesNoble:
Hey!

I was working on some line drawing and came across this potential issue:

If you create a graphics object, and then start drawing a line.. when you add a duplicate point into to `lineTo` method, the renderer somehow seems to add a gap in the line. I am unintentionally adding the duplicate point, so I'm not sure how much of a 'bug' this is.. However, this is certainly "unexpected behaviour" in my eyes.

Here is a working example using the latest PIXI:

```javascript

var app = new PIXI.Application(800, 600, { antialias: true });
document.body.appendChild(app.view);

var lines = [
  [
    [-8.154,558.886],
    [121.097,294.348],
    [133.744,282.209],
    [134.591,281.512],
    [166.995,250.542], // < --- Duplicate set of points
    [166.995,250.542], // NOTE: comment this out to close the line
    [167.029,250.518],
    [170.649,248.128],
    [172.666,246.822],
    [176.68,244.274],
    [178.455,243.17],
    [292.488,-14.857]
  ]
]

lines.forEach(function(line){
   var graphics = new PIXI.Graphics();
   graphics.lineStyle(4, 0xffd900, 1);
   line.forEach(function(point, i) {
      if (i === 0) graphics.moveTo(point[0], point[1]);
      else graphics.lineTo(point[0], point[1]);
   })
   graphics.endFill();
   app.stage.addChild(graphics);
})
```

finscn:
Yes, I think it's a bug really.
similar to https://github.com/pixijs/pixi.js/issues/3592

j-funk:
@XerxesNoble this has been affecting me, thanks for figuring it out!

G-Rath:
This bug is affected me as well. I've been doing pathfinding with A*, and have a function for drawing the paths:
```
    /**
     *
     * @param {Array<Array<number>>} path
     *
     * @return {PIXI.Graphics}
     */
    drawPath(path) {
        const graphic = new PIXI.Graphics();

        graphic.lineStyle(1, Utils.hexColor('#ff999f'));

        const startingPoint = path[0];
        graphic.moveTo(startingPoint[0] + 1, startingPoint[1]);

        path.forEach(point => graphic.lineTo(...point));

        gameApp.stage.addChild(graphic);

        return graphic;
    }
```

[...]

Issue #3891, opened 2017-03-29
NineSlicePlane in canvas renderer has little gap on it


DerivcoHK:
As you can see here, NineSlicePlane has some little gap on it when using CanvasRenderer. Any work around can fix this issue?
![nine-slice-plane](https://cloud.githubusercontent.com/assets/16910927/24436112/332a7d3e-146c-11e7-919d-64d7fa1114ce.png)

themoonrat:
does 
roundPixels: true
help when creating the renderer?

DerivcoHK:
It does help when the scale of the mesh is 1.
Problem still exists when scale > 1

themoonrat:
Closing as was fixed in 4549

PR #4549, opened 2017-12-20
fix Canvas Mesh Render bugs 

qingyangmoke:
1.  fix line gap bug when you set mesh canvasPadding>0 and sprite set scale=-1 
2.  add mesh defaults config object, so you can set change the default canvasPadding value ,specially you use dragonbones with gap line problems you can set `PIXI.mesh.Mesh.defaults.canvasPadding=1`
to fixed gap line problems  

like issue #4046 and #3891


ivanpopelyshev:
Seems like a valid fix, but we need fiddle for it anyway.

Can you please move defaults to settings? https://github.com/pixijs/pixi.js/blob/dev/src/core/settings.js

@bigtimebuddy @GoodBoyDigital Regardless of deprecation of canvas renderer, this fix is important for both v4 and v5

qingyangmoke:
@ivanpopelyshev done ! move defaults to settings, ` PIXI.settings.MESH_CANVAS_PADDING = 1;`,please review it, i fix the bug because my team member use dragonbones have this gap line problems yesterday ,i fixed it today, and we  test in my project already, Performing well, please test again ,thx, hope help someone who real need it

qingyangmoke:
when i  init 
 `new PIXI.CanvasRenderer({ 
      resolution: 2, // 
    });`
 and set `PIXI.settings.MESH_CANVAS_PADDING = 2; `
i find perform different，so i check the code, find it is wrong, then i change the `const canvasPadding = mesh.canvasPadding * this.renderer.resolution;` to `const canvasPadding = mesh.canvasPadding / this.renderer.resolution;`    
then test  again
set `PIXI.settings.MESH_CANVAS_PADDING = 2` whether resolution=2 or resolution=1
it perform same  @ivanpopelyshev

bigtimebuddy:
Thanks @qingyangmoke!

PR #4550, opened 2017-12-20
Add ability to customize strings '|MÃ‰q' and 'M' in TextMetrics.measureFont()

ilya-klindukhov:
Pixi calculate font metrics by string "|MÃ‰q". But in some cases it can work incorrect e.g. if this symbols is missing in font or special 'locale' font where cut all symbols except glyps included in this language.

Some developers have the same problem:

1. https://github.com/pixijs/pixi.js/issues/4500
2. https://github.com/pixijs/pixi.js/issues/4228
3. some closed issues

I have this problem too

You are recomend 2 solutions in this case:
1) use 'padding' property
2) override TextMetrics.measureFont() for set custom string instead "|MÃ‰q".

First solution is help only for render text but not fixing calulate text metrics. So we have problem with multiline text (lines of text placed into each other) and 'y' position of text need to correct by 'padding' value.
Second solution work perfect but override all function only for change 1 line - this is bad practice for supporting this code ( e.g. need to check diff after every pixi update ).
Will be good to add ability to customize strings "|MÃ‰q" and "M" without override functions on TextMetrics


ivanpopelyshev:
Add "1.4" to the params too!

ilya-klindukhov:
I try to understand what means of value '1.4' but it left magic for me so I did not move it to params. Could you tell what means in multiply by '1.4'(to know how to named property)?

ivanpopelyshev:
Actually, I dont understand it too. I also dont like 'height = 2 * baseline` , because that's not true for https://fonts.google.com/specimen/Amatic+SC

ivanpopelyshev:
Btw, if you're from ex-USSR, you should join this telegram channel https://t.me/html5dev

ivanpopelyshev:
@themoonrat @bigtimebuddy can we store those params in `TextMetrics`, or should we move it to `PIXI.settings` and `PIXI.const` ?

bigtimebuddy:
We should add any setting to TextMetrics, like setting the validation string. Not to PIXI.settings.

src/core/text/TextMetrics.js
```diff
@@ -288,6 +289,36 @@ export default class TextMetrics
...
         return properties;
     }
+
+    /**
+     * Set custom params for calculate font metrics
+     * If validateString or baselineSymbol is not the same as a previous values than font metrics cache will reset
+     *
+     * @static
+     * @param {string} validateString - Symbols with different metrics: the highest symbol, the lowest, the widest,...
+     * @param {string} baselineSymbol - The widest symbol in fonts
```
themoonrat: improve doc a little to state that this is used to help measure the height of a line of text

bigtimebuddy:
My only issue with this is that it can be done more simply and with more flexibility. I want to propose an alternative API.

```js
TextMetrics.VALIDATE_STRING = '|Éq';
TextMetrics.BASELINE_SYMBOL = 'M';
TextMetrics.clear = function clear(font = null) {
  if (font) {
    delete TextMetrics._fonts[font];
  }
  else {
    TextMetrics._fonts = {};
  }
}
```

I think the reason you were adding `setValidateParams` was so that you could clear the cache. This convenience comes at a cost of blowing away other measured fonts. I'd rather give the user control over flushing the save font metrics either all or on a per-font basis.

Usage to change the validation string would look something like this:

```js
TextMetrics.VALIDATE_STRING = '|Ãq';
TextMetrics.clear(); // or clear only the offending font
```

themoonrat:
I'm also with Ivan that this magic "1.4" number could be either changed by default, or make it configurable, or maybe even both.

@ilya-klindukhov What that number is for is.....
We are trying to work how tall text can possibly be. Unfortunately for us, the 'measureText' function can only return a width. So what we do, is take what we think is the _widest_ character (which is often 'M'), and then say 'well, because we're not really sure if the width of M is really going to cover how tall it also is, we will multiply that by a number to have 'a bit more'. Then we'll use that value to create a canvas, draw the letter to it, and then measure how tall it _really_ is by trying to find when we see blank space when analysing the pixels.

Where things can do wrong, is in fonts that are very tall. Text is already taller than the width of M * 1.4. So the height it gives per line is that value, even though it actually isn't enough.

Now, if that number is 1.4 or 2, doesn't _really_ matter... a value of 2 would theoretically take a tiny amount longer to process the first time the font is used, but after that, the value is cached, but you now have more chance to accurately report taller fonts.

We'd want a little testing to confirm that this theory is correct before just changing that number, tho :) Maybe we leave it at 1.4 as that is what everyone is expecting who is on v4, but allow that value to be updated. Maybe we only increase is to 2 by default for v5?

ilya-klindukhov:
PR updated

@themoonrat:

1. 'validateString' was renamed to 'metricsString', 'setValidationParams' to 'setMetricsParams'. Documentation was updated too.
2.  I combine 'metricsString' and 'baselineSymbol' becouse baselineSymbol not include in metricsString now. It's need if user don't know how textMetrics work in pixi and set custom metrics string ( and baselineSymbol not include in metricsString) for guessing metrics for custom font. Combine  'metricsString' and 'baselineSymbol' guarantee that baselineSymbol will include in metricsString in measureFont().
3. added 'baselineMultiplier' and default value for it. Also this param can change by setMetricsParams()

@bigtimebuddy
If user change 'baselineSymbol' or 'metricsString' than cached metrics for some fonts can be incorrect. If user forgot call clear() after set public property 'baselineSymbol' or 'metricsString' than some text can show incorrect. Use 'setMetricsParams' guarantee than cache will clear after change params.
But use manual clear(font) has sence too. I.e when user change font texture than cached metrics will be incorrect. Call clear(font) can help for update metrics for this font.
So, I added clearMetrics() too.

src/core/text/TextMetrics.js
```diff
@@ -288,6 +289,63 @@ export default class TextMetrics
...
+        {
+            TextMetrics.clearMetrics();
+        }
+    }
+
+    /**
+     * Clear font metrics in metrics cache.
+     *
+     * @static
+     * @param {string} font - font name. If font name not set than clear cache for all fonts.
```
cursedcoder: `than` -> `then`

tywang2006:
I found the code `var lineWidth = context.measureText(lines[i]).width + (lines[i].length - 1) * style.letterSpacing; in TextMetrics.measureText`. I suggest to change it to 
```
for(var j = 0; j < lines[i].length; j++) {
       lineWidth += context.measureText(lines[i].charAt(j)).width;
 }
 lineWidth += (lines[i].length - 1) * style.letterSpacing;
```
It will make the text display more accurate. measureText is decided by the renderer of the browser. for example, measureText is perfect to work in IE edge but not really nice in chrome.

ivanpopelyshev:
@tywang2006 Do you know how it affects performacne? It looks like extra memory allocation for every symbol, so I suggest we add it as a optional method. We can make it main if people prefer it.

tywang2006:
I do know it will hit performance but it will be nice to be optional for some cases which they really need better accuracy of text size.

PR #4565, opened 2017-12-28
Ensure argument type is string for getChildByName

dolow:
Added the argument validation for extras/getChildByName.

If user accidentally pass `undefined` as argument, the first unnamed child will be returned unexpectedly since the property `name` is not a default property for `Container` or its ancestors.

If the argument validation is not suit for design policy of pixi.js, I will follow the policy and this pull request should be closed.

Thanks in advance.

### DIFF
--- a/src/extras/getChildByName.js
+++ b/src/extras/getChildByName.js
@@ -17,6 +17,11 @@ core.DisplayObject.prototype.name = null;
  */
 core.Container.prototype.getChildByName = function getChildByName(name)
 {
+    if ((typeof name) !== 'string')
+    {
+        return null;
+    }
+
     for (let i = 0; i < this.children.length; i++)
     {
         if (this.children[i].name === name)


dolow:
> If user accidentally pass `undefined` as argument, the first unnamed child will be returned unexpectedly since the property `name` is not a default property for `Container` or its ancestors.

Sorry, this part was bit wrong.
The property `name` is defined in EventEmitter as a prototype property of `Function`.
so, default value for `name` would be `null`.

I think it means using `name` property is quite unsafe.
This is not a topic for this pull request though.
I'm gonna leave issue instead.

bigtimebuddy:
Thanks for the PR @dolow. However, it is not common design in PixiJS to check for incorrect types. Since `name` is not optional and typed as a `string` in the documentation, we feel it's unnecessary to check for non-string values.

Closing this.

I did noticed that the `getChildByName` method is not showing up in the documentation for [Container](http://pixijs.download/release/docs/PIXI.Container.html) and `name` property isn't showing up in documentation for [DisplayObject](http://pixijs.download/release/docs/PIXI.DisplayObject.html). Would welcome a PR to fix that anyone out there!

dolow:
@bigtimebuddy 

Thanks for letting me know about PixiJS design, I understood that closing this PR is appropriate.
Anyway, I made another PR to fix jsdoc ( https://github.com/pixijs/pixi.js/pull/4577 ).
I hope this would help you.

PR #4573, opened 2018-01-02
Removing loadType from the loadOptions

mbittarelli-jibo:
Assuming a loadType of Resource.LOAD_TYPE.IMAGE conflicts with the ability to load an image which can't be loaded via the <image> tag.  This happens in the case of loading a compressed image which should be handled by the XmlHttpRequest file loader as an array buffer type.  In the resource loader, there is already a lookup to determine the load type: https://github.com/englercj/resource-loader/blob/master/src/Resource.js#L126
In the case of a binary image asset, the spritesheetParser's loadType stomps on a dynamic lookup based on filetype.


ivanpopelyshev:
does it work with pixi-compressed-textures now?

mbittarelli-jibo:
The setup I had tested this with makes use of pixi-compressed-textures.  Essentially this code snippet:
```let basePath = "file:/path/to/assets";
let renderer = new PIXI.autoDetectRenderer(1280, 720);
 
let extensions = PIXI.compressedTextures.detectExtensions(renderer);
let loader = new PIXI.loaders.Loader();
loader.pre(PIXI.compressedTextures.extensionChooser(extensions));

let metadata = { default: { metadata: { imageMetadata: { choice: [".crn"] } } } };
 
loader.add(assetId, basePath + assetPath, metadata);
loader.once('complete', doneCallback).load();
```

bigtimebuddy:
@mbittarelli-jibo hey Matt! Good to see you here. Could you add a unit test for this case? Want to make sure we protect this functionality long-term.

mbittarelli-jibo:
@bigtimebuddy sure, I can add a test, but I'm not sure what you're looking for to be tested...
I haven't changed the interface to this method, so adding something to "test that all the inputs give you valid outputs" doesn't really apply here.

This PR really changes essentially the parameters that are being passed to a loader which is in another repo entirely, so really the resource loader should be the thing testing that not receiving a `loadType` falls back to determining a valid loadType.

I could add a test that verifies that the `this` object (which would need an `add` function) doesn't receive a `loadType` property in the `loadOptions`.  Though, I'm not sure how useful this would be as a test...

Thoughts?

bigtimebuddy:
My suggestion is to create a unit-test which loads a spritesheet that does not have [one of these](https://github.com/englercj/resource-loader/blob/master/src/Resource.js#L1002-L1012) extensions and it should load as the XHR default. The resource 's`data` field should _not_ be an HTMLImageElement.

```js
const loader = new PIXI.loaders.Loader();
loader.add('atlas', 'atlas.json'); // <-- image field should be not image extension
loader.load((loader, resources) => {
	expect(resources.atlas.data).to.not.be.instanceof(HTMLImageElement);
});
```

mbittarelli-jibo:
@bigtimebuddy thanks!  Do you know when this might be merged / a release made?

ivanpopelyshev:
@mbittarelli-jibo week or two. We need more fixes for new release. Maintain your fork for now.

bigtimebuddy:
Yes what @ivanpopelyshev said. Release will probably be in a week or two max.

PR #4577, opened 2018-01-04
fixed jsdoc annotations

dolow:
As @bigtimebuddy mentioned in  https://github.com/pixijs/pixi.js/pull/4565 , there are properties those are not documented as they are defined asynchronously.
This PR fixes it.


GoodBoyDigital:
Thanks for this @dolow ! 👍

PR #4578, opened 2018-01-04
Bugfix/frame buffer textures

GoodBoyDigital:
Easy fix, we forgot to update the texture constructors to use the new
resource / object format.


GoodBoyDigital:
bish bash bosh!

Issue #4133, opened 2017-06-30
Rework APIs that use radians to use degrees instead


englercj:
Stated benefits:

- Easier to use end-user API
- Easier to reason about and read user-code
- Easier to persist degree values rather than radian values
- More consistent with other web-based, and other 2D engines

mreinstein:
A few const that come to mind:

* built-in javascript math functions standardize on radians
* built-in glsl functions standardize on radians
* radians are a more convenient format for dealing with some types of computations (e.g., I can lerp from 0-> Math.pi * 2 and get a full rotation

ivanpopelyshev:
I'm math guy, and I prefer to use either degrees either 0..1 either 0..512 because of persistance and shader format reasons. If you asked me same question two or more years ago I could say that radians are better because they are "pure" in math meaning.

However, storing 2*Math.PI in files is a problem, and "360" is just faster to write than 2*PI, it cant be represented clearly in binary form. 

Flash-based engines are working moslty with degrees:

https://libgdx.badlogicgames.com/nightlies/docs/api/com/badlogic/gdx/graphics/g2d/Sprite.html#getRotation--

http://cocos2d-x.org/docs/programmers-guide/sprites/index.html

http://help.adobe.com/en_US/ActionScript/3.0_ProgrammingAS3/WS5b3ccc516d4fbf351e63e3d118a9b90204-7def.html

https://www.youtube.com/watch?v=zAsDbHXlFWI

ivanpopelyshev:
@englercj I have to work at that code too, I already have something like that in my fork: https://github.com/gameofbombs/gobi/tree/master/src/core/transform
https://github.com/gameofbombs/gobi/blob/master/src/core/math/FlatTransform2d.ts

@mreinstein I have to ask to look at that code too since you know something about pixi transforms ;)

englercj:
Suggestions from @bigtimebuddy & @GoodBoyDigital, make the API configurable:

```js
// useDegrees
CONVERSION = Math.PI / 180;

// useRadiens
CONVERSION = 1;

set rotation(rot)
{
    this._rotation = rot * CONVERSION;
}
```

OSUblake:
Maybe like Phaser does it? Angle is for degrees, rotation is for radians.

ivanpopelyshev:
@englercj Agree. That's the simplest way there. Engines based on pixi can just fix their "CONVERSION" constant.

1j01:
[...]

PR #4579, opened 2018-01-04
Next degrees

ivanpopelyshev:
According to #4133

Its a small API change, but seriously breaking.

For Pixi-v4 we can set `PIXI.settings.DEGREES_CONVERSION = 1`


bigtimebuddy:
I'm not convinced this is the best API solution. What if some plugins use radians and some use degrees? You don't want them fighting over this setting.

What do you think about this https://github.com/pixijs/pixi.js/issues/4133#issuecomment-312403814 by having `angle` (degrees) and `rotation` (radians). it doesn't solve all issues where radians are used, like skew, but I think it's better than this approach.

ivanpopelyshev:
@bigtimebuddy All plugins will have to move to v5 anyway, there are critical changes. Most of plugins are mine, and you can change stuff in `pixi-animate`. Also please notice that it works per-transform.

Skew also goes to degrees.

ivanpopelyshev:
Most of tools also use degrees and not rads. For spine I'll make entirely new runtime and deprecate v3/v4 support in it, there'll be spine for v3-v4 and for v5.

ivanpopelyshev:
WIP: Graphics.

bigtimebuddy:
Make setting the degrees or radians a boolean, not setting a numerical value, that can happen behind the scenes. `Transform.prototype.useRadians` `PIXI.settings.USE_RADIANS`, or `useDegrees`

packages/math/src/Matrix.js
```diff
@@ -361,6 +361,7 @@ export default class Matrix
         const skewX = -Math.atan2(-c, d);
         const skewY = Math.atan2(b, a);
 
+        const rad2deg = transform.useRadians ? 1.0 : (180.0 / Math.PI);
```
bigtimebuddy: Consider `import { RAD_TO_DEG } from './const'` and using `RAD_TO_DEG` here.

packages/math/src/Transform.js
```diff
@@ -87,10 +95,12 @@ export default class Transform
      */
     updateSkew()
     {
-        this._cx = Math.cos(this._rotation + this.skew._y);
-        this._sx = Math.sin(this._rotation + this.skew._y);
-        this._cy = -Math.sin(this._rotation - this.skew._x); // cos, added PI/2
-        this._sy = Math.cos(this._rotation - this.skew._x); // sin, added PI/2
+        const deg2rad = this.useRadians ? 1.0 : (Math.PI / 180.0);
```
bigtimebuddy: Same comment here as above: `import { RAD_TO_DEG } from './const'`

packages/math/src/Matrix.js
ivanpopelyshev: That'll be extra import line, I think my version is more verbose.

packages/math/src/Transform.js
ivanpopelyshev: ^^^

packages/math/src/Matrix.js
bigtimebuddy: With rollup it’s all flattened anyways so the extra import doesn’t matter. I’m fact, it’s better to import because this uses a single instance variable and compressed smaller. Verbosity is not useful here and constants  should be declared, imo.

packages/graphics/src/Graphics.js
```diff
@@ -1,5 +1,5 @@
 import { Container, Bounds } from '@pixi/display';
-import { BLEND_MODES } from '@pixi/constants';
+import { BLEND_MODES, DEG_TO_RAD } from '@pixi/constants';
```
bigtimebuddy: `DEG_TO_RAD` is in `@pixi/math`

packages/settings/src/settings.js
```diff
@@ -200,4 +200,19 @@ export default {
...
+    /**
+     * Since v5 pixi uses degrees.
+     *
+     * This legacy option allows to use radians to store rotation in Transform.
+     * Also affects TilingSprite and Graphics.
+     * Please set it before creation of instances of those classes.
+     *
+     * @static
+     * @constant
+     * @memberof PIXI
```
bigtimebuddy: Should be `@memberof PIXI.settings`

fritzy:
At first I wondered why, but after reading the issue, better compatibility and being more in line with other tools makes it worth it, IMO.

tobireif:
I'm against this because it's a breaking change. Perhaps it could be offered as option? (but the default would be the current state so that nothing gets broken)

ivanpopelyshev:
@bigtimebuddy need your feedback, whether we make `use_radians` a property of `Transform`, and add `rad2deg` and `deg2rad` inside.

brenwell:
I’m with @tobireif. Add util functions and work on the something more fun instead

themoonrat:
I thumbed up, but ultimately for me, internally, I don't really mind. Most of my annoyance is with DisplayObjects having rotation as radians, and it's not friendly using numeric data via json to set a display objects rotation. I'd be happy alone with @bigtimebuddy's comment of an additional getter / setter of 'angle' in degrees to compliment the 'rotation' in radians. 2 ways of accessing/changing the same information, just like setting width/height and setting scale do. Both useful in different contexts.

markknol:
I vote for keeping radians, I liked that tbh. Degrees are only nice for readability, not for math, There aren't much places you actually need degrees (?). Radians is used in trig functions (sin/cos), so that would be awkward to work with. I think for support libraries like spline etc is easier to put a  `* DEG_TO_RAD` once in the code instead of forcing pixi users to use `* RAD_TO_DEG` in their code.

As alternative you could also add some convenience `rotationInDegrees` getter/setter function which works with degrees and alters `rotation` property, but I can understand this might make the api look less nice.

cursedcoder:
`It is a breaking change` – want to remind it is `5.0.0` which states for BREAK in https://semver.org/

`Other tools are dependent on us` – anyway they will have to migrate with a bunch of breaking changes, this one should not take more than 5 LoC's

`Have different setters for everyone` – big thumbs up

ivanpopelyshev:
Radians are now by default, like in v4. I put link to USE_DEGREES in relevant places in docs.

djmisterjon:
now this extend will helping
```javascript
(function (R) {
  Math.cosd = function(d) { return Math.cos(d * R); };
  Math.sind = function(d) { return Math.sin(d * R); };
  Math.tand = function(d) { return Math.tan(d * R); };
})(Math.PI / 180);
```
Able to get the cosine at 45° by doing Math.cosd(45) 
or the sine at 75° by doing Math.sind(75) 
or the tangent at 135° by doing Math.tand(135) ....

bigtimebuddy:
@ivanpopelyshev I think you dismissed my point about plugins. I'd like to just put a finer point on it. Here's an example:

Here's how we'd implement a simple plugin now that handles rotation:

```js
PIXI.DisplayObject.prototype.flip = function flip() {
  this.rotation = Math.PI;
};
```

If I wanted to implement this _same_ plugin using this PR, I'd probably have to do this:

```js
PIXI.DisplayObject.prototype.flip = function flip() {
  this.rotation = Math.PI * this.transform.rad2deg;
};
```

So a couple of points about this:
* The name `rad2deg` is pretty deceiving here because we are actually normalizing the units not converting them. More appropriate names might be `normalizeRadians` and `normalizeDegrees` instead of `rad2deg` and `deg2rad`
* To me this approach is confusing for 3rd-party plugins that which to utilize rotation and skew. All units need be converted so it's a big gotcha. 
* Finally, and this is probably very obvious, but this makes the `rotation` return dynamic. It can return different values in different contexts. This approach feels like an anti-pattern in Pixi, I don't know of any other API who's return value units changes because of an external switch. I feel like this is an inconsistent and bad API design.

ivanpopelyshev:
The PR wont be accepted if it will have too many lines of code.

My concern is to enable storage in degrees, and ability to assign degrees to rotation.

I see two ways:
1. rotation is dynamic , storage is dynamic
2. storage is always degrees, add rotationDeg field

ivanpopelyshev:
Two storage fields is a bad idea, both for core and plugins. Please confirm that you agree to store rotation in degrees.

bigtimebuddy:
It’s not creating two storage fields, it’s one storage and one convenience method on DisplayObject. We should continue to store radians in Transform.

ivanpopelyshev:
> We should continue to store radians in Transform.

```js
element.rotationDeg = 30;
console.log(element.rotationDeg === 30);
```

```js
var deg2rad = Math.PI / 180.0;
var rad2deg = 180.0 / Math.PI;
console.log(30 * deg2rad * rad2deg);
```

ivanpopelyshev:
That's why I want to store degrees. Extra multiplication before using trigonometric functions is not a problem, especially when its always there.

PR #4583, opened 2018-01-05
Create TilesetLoader plugin to load images of TSX (Tile Set XML) files

winniehell:
see also http://docs.mapeditor.org/en/latest/reference/tmx-map-format/?highlight=tsx#image


ivanpopelyshev:
Yep, good enough. Lets add it in v5 though, it'll be released soon.

winniehell:
> Lets add it in v5 though, it'll be released soon.

@ivanpopelyshev Is there anything I need to do for this? Should I target `next` instead of `dev`?

ivanpopelyshev:
yeah, but first look at `next` and try to setup it in different folder, because its dependencies really differ

bigtimebuddy:
@winniehell would like to see this loader middleware in v5, like @ivanpopelyshev suggested.

The loader plugins are organized a little different in v5. See this example: https://github.com/pixijs/pixi.js/blob/next/packages/spritesheet/src/SpritesheetLoader.js

Loader plugins are also registered here: https://github.com/pixijs/pixi.js/blob/next/bundles/pixi.js/src/index.js#L46-L47

For this code, it might make sense to create a new **packages/tileset** module. Very similar to **packages/spritesheet**. 

If you have any questions about the new organization of **next** let me know.

englercj:
Recently I was playing around with a webgl renderer for tiled maps ([here](https://github.com/englercj/gl-tiled)) which included a resource-loader middleware for Tiled json maps ([here](https://github.com/englercj/gl-tiled/blob/6d5119933863281c1328dad912deaa7f77673122/bundles/resource-loader/index.ts)). Should be pretty easy to buff that one up to support tsx as well.

winniehell:
> would like to see this loader middleware in v5

Ok, I will update this branch and target `next`.

> If you have any questions about the new organization of next let me know.

Thank you—also for the pointing me to the right places! :smiley:

> create a new packages/tileset module

@bigtimebuddy I'm planning to also write a loader for TMX map format and the corresponding JSON format (to automatically load tilesets for maps). Should we name the package `tiles` or `tiled` rather than `tilesets`?

> Recently I was playing around with a webgl renderer for tiled maps 

@englercj I have recently created a canvas renderer for tiled maps in KotlinJS ([code](https://gitlab.com/winniehell/kotmx), [demo](https://winniehell.gitlab.io/kotmx/)) which also supports [compressed layers](http://docs.mapeditor.org/en/latest/reference/tmx-map-format/?highlight=zlib#data) via [pako.inflate](http://nodeca.github.io/pako/#inflate). My struggle with WebGL is how I ended up using PixiJS. :smiley: I'm happy to join efforts if you like.

bigtimebuddy:
Totally open to package name. “tiled” is good and aligned with the editor name.

winniehell:
@bigtimebuddy I have ported this to `next` now. Could you please have a look?

packages/tiled/src/TilesetLoader.js
```diff
@@ -0,0 +1,53 @@
...
+
+            return;
+        }
+
+        const loadOptions = {
+            crossOrigin: resource.crossOrigin,
+            parentResource: resource,
+        };
+
+        const imageElement = tileset.getElementsByTagName('image')[0];
```
bigtimebuddy: In @englercj's loader middleware example for loading tiled JSON files, he is handling many more image instances. In this case, the loader plugin only loads the first image element within the tileset. I'm not super familiar with the format, but shouldn't this handle more images?

bigtimebuddy:
The package is well formatted, good job there. I have some questions about usefulness in general about this. The plugin just loads the Texture image from the tmx file. But there isn't any other support for creating sub-textures, formatting the data, etc. @winniehell, can you talk more about the use-case and why just loading the image is helpful like this?

packages/tiled/src/TilesetLoader.js
winniehell: > he is handling many more image instances

@bigtimebuddy I think that is due to the fact that @englercj is loading the [JSON Map Format] which represents a complete map. It is the JSON counterpart of the [TMX Map Format].

The plugin that I have written here is only for [TSX (Tile Set XML) files] which are external tilesets and may be referenced by either [JSON Map Format] or [TMX Map Format].

I'm planning to also write plugins for the map formats (both TMX/XML and JSON) but thought that I'd start with a smaller contribution that can be finished on a weekend. :smiley: 

>  shouldn't this handle more images?

Yes, I was misreading the format. I only thought of

```xml
<?xml version="1.0" encoding="UTF-8"?>
<tileset name="building1" tilewidth="16" tileheight="16" tilecount="64" columns="8">
 <image source="building1.png" width="128" height="128"/>
</tileset>
```

but this is also possible:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<tileset name="building1" tilewidth="209" tileheight="144" tilecount="2" columns="13">
 <grid orientation="orthogonal" width="16" height="16"/>
 <tile id="0">
  <image width="128" height="128" source="building1.png"/>
 </tile>
 <tile id="1">
  <image width="209" height="144" source="paint-tanuki.png"/>
 </tile>
</tileset>
```

Thank you for making me check this! :+1: I will adjust the loader plugin and make sure it is covered by the tests.

[JSON Map Format]: http://doc.mapeditor.org/en/latest/reference/json-map-format/
[TMX Map Format]: http://doc.mapeditor.org/en/latest/reference/tmx-map-format/
[TSX (Tile Set XML) files]: http://doc.mapeditor.org/en/latest/reference/tmx-map-format/#tileset

winniehell:
> can you talk more about the use-case and why just loading the image is helpful like this?

@bigtimebuddy This loader plugin reduces this (simplified) code by one `loader.load`:

```javascript
const loader = new PIXI.Loader();
loader.add('map', 'my-map.tmx');
loader.load((loader, resources) => {
  const externalTilesets = resources.data.querySelectorAll('tileset[source]');

  for (let tilesetIndex = 0; tilesetIndex < externalTilesets.length; ++tilesetIndex) {
    loader.add(`map_tileset${tilesetIndex}`, externalTilesets[tilesetIndex].getAttribute('source'))
  }

  loader.load((loader, resources) => {
    for (let tilesetIndex = 0; tilesetIndex < externalTilesets.length; ++tilesetIndex) {
      const externalImages = resources[`map_tileset${tilesetIndex}`].querySelectorAll('image[source]');

      for (let imageIndex = 0; imageIndex < externalImages.length; ++imageIndex) {
        loader.add(`map_tileset${tilesetIndex}_image${imageIndex}`, externalImages[imageIndex].getAttribute('source'));
      }

      loader.load((loader, resources) => renderMap(resources));
  });
});
```

I would really like to have 

```javascript
const loader = new PIXI.Loader();
loader.add('map', 'my-map.tmx');
loader.load((loader, resources) => renderMap(resources));
```

In the examples above `renderMap` would be provided by application code but use PixiJs features.

> there isn't any other support for creating sub-textures, formatting the data, etc.

After I had read the following comment by @englercj:

> Nope, tilemaps are not something we will put into core pixi.js.
> doesn't make sense to be in our thin rendering layer

(https://github.com/pixijs/pixi.js/issues/1714#issuecomment-95947955), I thought that in-depth parsing or rendering in PixiJS is not wanted. If you have changed your mind since then, I'm happy to contribute more advanced logic.

Please note however that there are already the following projects targeting this:
- https://github.com/pixijs/pixi-tilemap
- https://github.com/englercj/gl-tiled
- https://github.com/finscn/tiled-kit.js (https://github.com/pixijs/pixi.js/issues/4308)

bigtimebuddy:
First of all, thank you for submitting this and retooling it for v5. You did a killer job @winniehell and it's a solid pull-request. 

However, after having discussed with @GoodBoyDigital and seeing the description of use-case, I don't believe this is right for PixiJS v5. We want to support a variety of media types, but TMX files aren't really a media format (i.e., a TexturePacker atlas) as much as it's a layout file. Without supporting the rendering of TMX files, adding this middleware is only a half-solution.

While PixiJS is used for a variety of types of projects (games, tools, apps, website, visualizations), supporting Tiled as default feels too bias towards games and therefore should not be a default package. I would strongly suggest adding this to one of the other projects described above or creating a small standalone loader middleware project you can import in your projects.

Sorry for the about-face on this, but we needed to chew on it a little more.  I will be closing this PR.

winniehell:
@bigtimebuddy Thank you for your feedback and for giving me the opportunity to have a look into the upcoming v5 release! :smiley: 

Given that there is a big refactoring pull request pending for `pixi-tilemap` (https://github.com/pixijs/pixi-tilemap/pull/47) and that the other two projects are self-contained and non-official, I will likely reimplement this in a new project using Kotlin (in the hope of a desktop version in the future—https://github.com/pixijs/pixi-native :wink: ).

I will post a link to the project here for reference once it is setup.

bigtimebuddy:
Thanks, keep us posted!

packages/tiled/src/TilesetLoader.js
englercj: Just FYI, there are two types of tilesets for tilelayers. One that has a single image of tiles (you're first xml block), and one that is a "multi-image tileset" where each individual tile is a single image (you're second xml block).

englercj:
@winniehell I'd be happy to take a PR that adds XML support to `gl-tiled` and I haven't started writing the pixi bundle yet if you were interested in that.

winniehell:
> I'd be happy to take a PR that adds XML support to gl-tiled

@englercj I'm sorry, I don't know TypeScript (yet). :disappointed: As soon as I have started, I can give the pull request a try.

englercj:
It is just ES6 with `: type` on variable/param declarations. It really isn't as different as you might think.

PR #4586, opened 2018-01-06
Next stage

ivanpopelyshev:
### Work in progress - more tests, examples.

# Return of the stage

This is colossal PR that was waited too long, people really want correct `added` and `removed` events and a way to track all animations that exist in main stage tree. This is critical for projects that need to free resources and count references to them.

`InnerStage` tracks sets of DisplayObjects and Containers that exist in the stage, some in detached stage. 170 lines of serious algorithm code were always on used-side.

`parentStage` property always points to the stage where the object is registered.

As a default implementation that can be easily changed in plugins and in user-side code, any object that has "animate" method will be called when the time comes. It can be changed both by overriding Stage methods and handling `added` and `removed` events.

There can be nested stages.

`added` and `removed` are fired for whole subtree.

```
var stage = new PIXI.Stage();
var child = new PIXI.DisplayObject();

container.addChild(child); // nothing
stage.addChild(container); // child is ADDED!
```

Additional way of moving objects in the same stage without firing events: `detachChild`. It exists in cocos2d-x as a flag to `removeChild`, which is not verbose. 

```js
var container1 = new PIXI.Container();
var container2 = new PIXI.Container();
var stage = new PIXI.Stage();

stage.addChild(container1);
stage.addChild(container2);

var child = new PIXI.Container();
container1.addChild(child); //ADDED
container1.detachChild(child);// event doesnt fire
//some calculations
container2.addChild(child); // event doesnt fire because its in the same stage
//some calculations
container2.detachChild(child);

stage.innerStage.flushDetached(); // actual REMOVED
```

### DIFF
--- a/packages/app/src/Application.js
+++ b/packages/app/src/Application.js
@@ -1,5 +1,5 @@
 import { settings } from '@pixi/settings';
-import { Container } from '@pixi/display';
+import { Stage } from '@pixi/display';
 import { Renderer } from '@pixi/core';
 import { Ticker, UPDATE_PRIORITY } from '@pixi/ticker';
 
@@ -52,6 +52,7 @@ export default class Application
      *  for devices with dual graphics card **webgl only**
      * @param {boolean} [options.sharedTicker=false] - `true` to use PIXI.Ticker.shared, `false` to create new ticker.
      * @param {boolean} [options.sharedLoader=false] - `true` to use PIXI.Loaders.shared, `false` to create new Loader.
+     * @param {PIXI.Container} {options.stage} - Pass existing stage or container
      */
     constructor(options, arg2, arg3, arg4, arg5)
     {
@@ -88,7 +89,7 @@ export default class Application
          * The root display container that's rendered.
          * @member {PIXI.Container}
          */
-        this.stage = new Container();
+        this.stage = options.stage || new Stage();
 
         /**
          * Internal reference to the ticker
@@ -142,8 +143,9 @@ export default class Application
     /**
      * Render the current stage.
      */
-    render()
+    render(delta)
     {
+        this.stage.animate(delta);
         this.renderer.render(this.stage);
     }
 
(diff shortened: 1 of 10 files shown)


packages/app/src/Application.js
```diff
@@ -52,6 +52,8 @@ export default class Application
      *  for devices with dual graphics card **webgl only**
      * @param {boolean} [options.sharedTicker=false] - `true` to use PIXI.Ticker.shared, `false` to create new ticker.
      * @param {boolean} [options.sharedLoader=false] - `true` to use PIXI.Loaders.shared, `false` to create new Loader.
+     * @param {PIXI.Container} {options.stage} - Pass existing stage or container
+     * @param {number} {options.animationDeltaMax} - How many frames will be processed by animation if user switches the tab
```
cursedcoder: 55-56 missing dots :)

cursedcoder: ok, I think this can be addressed as separate PR, we can add this to CI or somewhere

packages/text/src/Text.js
```diff
@@ -645,4 +645,10 @@ export default class Text extends Sprite
         this._text = text;
         this.dirty = true;
     }
+
+    onRemoved()
+    {
+        // Text frees videomemory when it leaves the stage
```
cursedcoder: let's move to the docblock?

packages/display/src/Stage.js
```diff
@@ -0,0 +1,63 @@
+import Container from './Container';
+import InnerStage from './InnerStage';
+import Runner from 'mini-runner';
+
+/**
+ * Stage is a Container that takes care of add/remove events and animations
```
cursedcoder: As for me, I don't quite get a reason to have multiple stages.

Would be cool to add short example/use-cases in docs?

cursedcoder: I think something like your description is a great thing to put here.

Just maybe make it more compact.

PR #4588, opened 2018-01-07
[next] Fix CanvasRenderer

ivanpopelyshev:
Obvious fix. Only CanvasRenderTexture contains source here, we need extra function for it.


bigtimebuddy:
Nice work! There are a few unit-tests that were disabled (e.g. search for `it.skip(`) because canvas was broken. Could you re-enable these tests?

bigtimebuddy:
Verified, graphics and textures work in canvas: https://jsfiddle.net/68t4bsva/3/

themoonrat:
for me "canvasSource" is a little ambiguous. When I read that, it could mean
1. A source, which is of type canvas (as opposed to html img)
2. The source in a format required by the canvas renderer

Maybe there could be a name change to clarify this? (not that I have any good ideas right now)

bigtimebuddy:
Maybe `canvasResource`?

themoonrat:
Sounds better to me, yeah

ivanpopelyshev:
I think `getDrawableSource` is better. I dont want to add extra properties for things, its not easy to override them if something goes wrong. 

In that kind of cases "getXXX" shows that there's actually some non-trivial logic inside.

bigtimebuddy:
I'm okay with that @ivanpopelyshev

bigtimebuddy:
Thanks @ivanpopelyshev!

PR #4596, opened 2018-01-10
[next] updateTransform temp parent

ivanpopelyshev:
Its not every day we have that kind of balance:  +14-51 ,3x times deleted!

Those methods have good test coverage. It allows me to modify those parts I and @mat worked on two years ago. For example, if I forget `updateTransform` at line 192, several tests will break.

Without this change we cant bring more heavy things and optimizations in updateTransform, we cant add new stuff if old is obfuscated.

Extra credit goes to @finscn for pushing me.

### DIFF
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -318,12 +318,13 @@ export default class Container extends DisplayObject
      */
     updateTransform()
     {
+        const parent = this.parent || this._tempDisplayObjectParent;
+
         this._boundsID++;
-
-        this.transform.updateTransform(this.parent.transform);
-
-        // TODO: check render flags, how to process stuff here
-        this.worldAlpha = this.alpha * this.parent.worldAlpha;
+        // coords transform
+        this.transform.updateTransform(parent.transform);
+        // color/alpha transform
+        this.worldAlpha = this.alpha * parent.worldAlpha;
 
         for (let i = 0, j = this.children.length; i < j; ++i)
         {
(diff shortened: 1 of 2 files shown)


packages/display/src/Container.js
```diff
@@ -318,12 +318,13 @@ export default class Container extends DisplayObject
      */
     updateTransform()
     {
-        this._boundsID++;
-
-        this.transform.updateTransform(this.parent.transform);
+        const parent = this.parent || this._tempDisplayObjectParent;
```
englercj: Note that the original reason we did it the way we did was that this branch here caused significant performance issues. That may have changed recently, but I wanted to call out the context for why we did it a more convoluted way.

ivanpopelyshev: Actually, that's a good idea, we can make parent a property :)

ivanpopelyshev:
What if we add `safeParent` and make it private?

ivanpopelyshev:
OK, as @englercj said, I need to think about it. Dont merge it yet.

finscn:
I think the code like this maybe better (I like this style) : 
```
if (parent) {
   this.transform.updateTransform(this.parent.transform);
} else {
    this.transform.anotherUpdateTransform(); // don'e need parentTransform
}
```

OR 

refactor `transform.updateTransform()` like this :  https://github.com/pixijs/pixi.js/pull/3415#issuecomment-265204942

then  use this
```
 this.transform.updateTransform(this.parent ? this.parent.transform : null);
```

ivanpopelyshev:
updateTransformLocal + copyTo could work.

But I also have an idea of `safeUpdateTransform`, just call it instead of updateTransform is you're not sure what can be there

PR #4598, opened 2018-01-11
[next] temp parent refactor

ivanpopelyshev:
Another version of #4596

I agree with @englercj that we should not add this branching in critical place. Although transform itself has two branches inside, I just dont like that temporary parent hack in a critical place.

I can rename `safeUpdateTransform` if anyone has ideas about new name.

### DIFF
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -315,10 +315,11 @@ export default class Container extends DisplayObject
 
     /**
      * Updates the transform on all children of this container for rendering
+     * parent must exist
      */
     updateTransform()
     {
-        const parent = this.parent || this._tempDisplayObjectParent;
+        const parent = this.parent;
 
         this._boundsID++;
         // coords transform
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -318,12 +318,13 @@ export default class Container extends DisplayObject
      */
     updateTransform()
     {
+        const parent = this.parent || this._tempDisplayObjectParent;
+
         this._boundsID++;
-
-        this.transform.updateTransform(this.parent.transform);
-
-        // TODO: check render flags, how to process stuff here
-        this.worldAlpha = this.alpha * this.parent.worldAlpha;
+        // coords transform
+        this.transform.updateTransform(parent.transform);
+        // color/alpha transform
+        this.worldAlpha = this.alpha * parent.worldAlpha;
 
         for (let i = 0, j = this.children.length; i < j; ++i)
         {
--- a/packages/display/src/DisplayObject.js
+++ b/packages/display/src/DisplayObject.js
@@ -1,7 +1,6 @@
 import EventEmitter from 'eventemitter3';
 import { Rectangle, Transform } from '@pixi/math';
 import Bounds from './Bounds';
-// _tempDisplayObjectParent = new DisplayObject();
 
 /**
  * The base class for all objects that are rendered on the screen.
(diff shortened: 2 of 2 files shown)


packages/display/src/DisplayObject.js
```diff
@@ -146,17 +145,38 @@ export default class DisplayObject extends EventEmitter
...
+        this.transform.updateTransform(parent.transform);
+        // color/alpha transform
+        this.worldAlpha = this.alpha * parent.worldAlpha;
+    }
+
+    /**
+     * Updates the object transform (coordinates and alpha) for rendering
+     * even if parent is null
+     */
+    safeUpdateTransform()
```
bigtimebuddy: Couldn't this all be simplified to and remove the if/else:
```js
const parent = this.parent;
this.parent = parent || this._tempDisplayObjectParent;
this.updateTransform();
this.parent = parent;
```

packages/display/src/DisplayObject.js
```diff
@@ -266,20 +277,6 @@ export default class DisplayObject extends EventEmitter
         if (!skipUpdate)
         {
             this._recursivePostUpdateTransform();
```
bigtimebuddy: Don't you need to call `this.safeUpdateTransform();` here?

packages/display/src/DisplayObject.js
```diff
@@ -306,20 +303,6 @@ export default class DisplayObject extends EventEmitter
...
```
bigtimebuddy: What about here too?

packages/display/src/DisplayObject.js
ivanpopelyshev: Nope, its not required for toLocal/toGlobal. there was single displayObjectUpdateTransform, without children, but its called from _recursivePostUpdateTransform anyway!

packages/display/src/DisplayObject.js
ivanpopelyshev: ^^^ same.

packages/display/src/DisplayObject.js
bigtimebuddy: Cool, just making sure

bigtimebuddy:
I'm not the biggest fan of the name `safeUpdateTransform`.  I think we need to be more explicit than just "safe" about what this is doing. First question, does this need to be a public method?

packages/display/src/DisplayObject.js
ivanpopelyshev: I think its making code less verbose.

ivanpopelyshev:
Its one of those plugin-level methods. I agree that it has to be renamed.

packages/display/src/DisplayObject.js
cursedcoder: @bigtimebuddy you can't modify `const` value :)

I would give +1 for `if-else` statement, it improves readability at least twice

packages/display/src/Container.js
```diff
@@ -315,15 +315,17 @@ export default class Container extends DisplayObject
 
     /**
      * Updates the transform on all children of this container for rendering
+     * parent must exist
```
cursedcoder: missing dot here

packages/display/src/Container.js
```diff
@@ -315,15 +315,17 @@ export default class Container extends DisplayObject
...
-
-        this.transform.updateTransform(this.parent.transform);
+        const parent = this.parent;
 
-        // TODO: check render flags, how to process stuff here
-        this.worldAlpha = this.alpha * this.parent.worldAlpha;
+        this._boundsID++;
+        // coords transform
+        this.transform.updateTransform(parent.transform);
+        // color/alpha transform
```
cursedcoder: doesn't 325 and 327 comments seem obvious?

bigtimebuddy:
I'm going to close this because of inactivity, no consensus around naming the new API, and the lack of a clearly-defined performance to understand benefits. If someone can describe the performance benefit, will consider avoiding the extra branch.

PR #4600, opened 2018-01-11
[next] Particles fix

ivanpopelyshev:
What did I know about GeomtrySystem:
1. It does not take into account that other systems can mess with index buffer. ThreeJS can do that.

It is fixed in this PR.

What do I know after spending several hours on it at 2:30AM :
1. If you assign uniforms not through groups, but old `shader.uniforms.xxx` way, do it before first bind, oterwise sync code wont be generated
2. There's no good way to debug WTF is uploaded into uniforms. I know its a sum of Program and Shader, but really, no way to see it.
3. GeometrySystem does `for (attributeName in attributes)` cycle that assigns offsets, order of attributes is not defined, thus automatic interleaving causes undefined behaviour.

Those things have to be fixed or documented later.

Particles code will be simplified later: we dont need that many variables there and in renderer.

I removed random `new` from TilingSprite, it works fine.


bigtimebuddy:
Do we have something we can easily test particles with using this build https://pixijs.download/next-particles/pixi.js ?

ivanpopelyshev:
Good old http://pixijs.io/examples/#/demos/batch.js . I dont create fiddles because I have it all setup locally ;)

bigtimebuddy:
@ivanpopelyshev Could you support examples to use **next**, at least temporary, as we make our way thru v5? Will be helpful to test PRs

ivanpopelyshev:
http://pixijs.io/examples/?v=next-particles#/demos/batch.js

Dont forget there's the bug - if you press on other example, you have to refresh the page to use "?v=".

ivanpopelyshev:
To test canvas particles, we need to modify examples again, to use `pixi.js-legacy`. I checked canvas particles locally and it worked.

packages/core/src/renderers/systems/geometry/GeometrySystem.js
```diff
@@ -137,6 +137,23 @@ export default class GeometrySystem extends WebGLSystem
...
+
+            if (geometry.indexBuffer)
+            {
+                const glIndex = geometry.indexBuffer._glBuffers[this.CONTEXT_UID];
+
+                if (glIndex)
+                {
+                    gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, glIndex.buffer);
+                }
+            }
```
GoodBoyDigital: Does a native VAO not save the bound index buffer as well as any vertex buffers? I though it captured both :)

ivanpopelyshev: Index buffer is not bound to an attribute, it changes whenever you want to do any operations with that slot.

Other renderers dont respect that. We also didnt respect it in v4.

ivanpopelyshev:
The problem is that other renderers have different order of operations. They create and fill index buffer, then they create vao.. V4 also did that. 

Our GeometrySystem is fine.

packages/particles/src/ParticleBuffer.js
```diff
@@ -182,22 +184,24 @@ export default class ParticleBuffer
...
-                    gl.UNSIGNED_BYTE,
+                    0,
                     true,
+                    TYPES.UNSIGNED_BYTE,
                     this.staticStride * 4,
                     property.offset * 4
                 );
             }
             else
             {
```
GoodBoyDigital: Maybe for later, but we could replace ```property.unsignedByte``` and add ```normalise``` and ```type```  to the properties instead. This is possible as we now have access to webGL types all the time. Then we could remove the if statement here

```
if (property.unsignedByte)
            {
                geometry.addAttribute(
                    property.attributeName,
                    this.staticBuffer,
                    0,
                    true,
                    TYPES.UNSIGNED_BYTE,
                    this.staticStride * 4,
                    property.offset * 4
                );
            }
            else
            {
                geometry.addAttribute(
                    property.attributeName,
                    this.staticBuffer,
                    0,
                    false,
                    TYPES.FLOAT,
                    this.staticStride * 4,
                    property.offset * 4
                );
            }
```
could become
```
                geometry.addAttribute(
                    property.attributeName,
                    this.staticBuffer,
                    0,
                    property.normalised,
                    property.type,
                    this.staticStride * 4,
                    property.offset * 4
                );
```

Same applies to the dynamic attribute adding too.

packages/particles/src/ParticleRenderer.js
```diff
@@ -153,15 +132,16 @@ export default class ParticleRenderer extends ObjectRenderer
...
 
-        this.shader.uniforms.projectionMatrix = m.toArray(true);
+        this.shader.uniforms.translationMatrix = m.toArray(true);
 
         this.shader.uniforms.uColor = premultiplyRgba(container.tintRgb,
             container.worldAlpha, this.shader.uniforms.uColor, baseTexture.premultiplyAlpha);
 
-        // make sure the texture is bound..
-        this.shader.uniforms.uSampler = renderer.bindTexture(baseTexture);
+        renderer.texture.bind(baseTexture, 0);
```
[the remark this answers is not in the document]
ivanpopelyshev: I know, I just want it to be different from TilingSprite.

bigtimebuddy:
Good work @ivanpopelyshev

PR #4603, opened 2018-01-12
Support blendModes with NineSlicePlane and CanvasRenderer

ivanpopelyshev:
That's from Mesh:

```js
renderer.context.globalAlpha = mesh.worldAlpha;
renderer.setBlendMode(mesh.blendMode);
```

blendMode line is missing in NineSlicePlane. I can be a problem if something sets not-normal blendmode before NineSlicePlane is rendered.

Previously in Canvas Mesh: #4535


ivanpopelyshev:
Good, I just want to mention that usually people dont use anything except `Normal` on NineSlice, but they can use different blendMode in the element that appears before Plane

PR #4606, opened 2018-01-12
Follow W3 specs for `pointerId` mouse value

mudcube:
Pixi currently shims the mouse `event.pointerId` to be equals to `MOUSE`. The W3 specs for `pointerId` state:

> If the device producing the event is a mouse, then the pointerId must be 1. 

https://www.w3.org/Submission/pointer-events/#widl-PointerEvent-pointerId


bigtimebuddy:
How breaking is this? I'd like a fuller sense and what the impact will be. I would guess that if someone is use some logic `event.pointerId === 'MOUSE'` then they would have a problem with this change. How common do we think that is given that normalizing to pointer events is somewhat recent?

themoonrat:
I would have thought that if anyone would come to use pointerId on a mouse before; it not being a number would be annoying and they'd have come on here to complain! If you wanted to detect what device had emitted the event, you'd use pointerType.

I'm not pro breaking changes on v4, but this _is_ a fix for something that is incorrect.

bigtimebuddy:
I agree with @themoonrat. I think this is more a fix for an existing misconfiguration, then a breaking feature change. Would you reconsider @ivanpopelyshev?

bigtimebuddy:
@themoonrat I [checked](https://jsfiddle.net/bigtimebuddy/wx4jpLjx/) the `pointerId` and `pointerType` on mobile (iOS) and desktop (macOS) for touch, pointer and click events. In all cases, these values were what I expected. What else did you have in mind to check?

themoonrat:
I'm happy with that. Just didn't want other event types reporting 1 and throwing off the code :+1:

bigtimebuddy:
Thanks @mudcube!

PR #4608, opened 2018-01-14
Children outside of hitArea cannot be hit

themoonrat:
Fixes https://github.com/pixijs/pixi.js/issues/4564

Unit tests to cover this new case added (some interaction unit tests were within graphics; I've moved these into the interaction manager)

Example on 4.6.2: http://jsfiddle.net/themoonrat/fjo9aeew/1/
Example with fix: http://jsfiddle.net/themoonrat/fjo9aeew/2/

Note on 4.6.2, you see the hand appear if you follow the circles outside of the box on the bottom right corner. A console log is also output when a circle is hovered over. This no longer occurs with the fix

### DIFF
--- a/src/interaction/InteractionManager.js
+++ b/src/interaction/InteractionManager.js
@@ -1014,24 +1014,45 @@ export default class InteractionManager extends EventEmitter
         let hit = false;
         let interactiveParent = interactive;
 
-        // if the displayobject has a hitArea, then it does not need to hitTest children.
+        // Flag here can set to false if the event is outside the parents hitArea or mask
+        let hitTestChildren = true;
+
+        // If there is a hitArea, no need to test against anything else if the pointer is not within the hitArea
+        // There is also no longer a need to hitTest children.
         if (displayObject.hitArea)
         {
+            if (hitTest)
+            {
+                displayObject.worldTransform.applyInverse(point, this._tempPoint);
+                if (!displayObject.hitArea.contains(this._tempPoint.x, this._tempPoint.y))
+                {
+                    hitTest = false;
+                    hitTestChildren = false;
+                }
+                else
+                {
+                    hit = true;
+                }
+            }
             interactiveParent = false;
         }
-        // it has a mask! Then lets hit test that before continuing
-        else if (hitTest && displayObject._mask)
+        // If there is a mask, no need to test against anything else if the pointer is not within the mask
+        else if (displayObject._mask)
         {
-            if (!displayObject._mask.containsPoint(point))
+            if (hitTest)
             {
-                hitTest = false;
+                if (!displayObject._mask.containsPoint(point))
+                {
+                    hitTest = false;
+                    hitTestChildren = false;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
I'm looking at it.

ivanpopelyshev:
I thought that mask can work without hitArea.

themoonrat:
http://jsfiddle.net/9tm62nx5/

st3v0:
@themoonrat This change has broken all my buttons which use hitArea. 

If i rolled over a button and then moved the mouse off to the left the out events would not fire. However If i move off to the right then they would work fine.

I've rolled back to the 4.6.2 release and everything is working fine again.

st3v0:
The Example below will turn red when you roll over the box and white when not.
https://codepen.io/st3v0/pen/WdqrXX

I did notice a couple things:
- If i use Graphics instead of Sprite or Container it seems to work fine.
- The hit area is also affected by the anchor position. If you uncomment the `test.anchor.set` line you'll see that the hit area now starts halfway through the box, even though the hit area x,y to 0,0

themoonrat:
Cheers for the example, I'll take a look today. Our unit tests only cover checking hitArea with graphics, which is an oversight on our part

ivanpopelyshev:
hitArea is affected only by transform (position, scale, pivot, e.t.c.). If you use anchor to move the image, you have to adjust its hitArea too.

ivanpopelyshev:
In the fiddle, you change width/height of sprite, thus you change its `scale`, and it affects hitArea. Local coords (0,0,100, 100) become global (100, 0, 1000, 1000)

themoonrat:
Also, I'm trying this in 4.6.2 and I'm getting the same behaviour

qtiki:
This has introduced a breaking change in 4.7. In version 4.6.2 it was possible to create complex hierarchies of hit-areas. Now the hit-check early-exits if the point is not inside the parent. This is not consistent with what the user sees. In my opinion if a button is visible, it should also be clickable. I modified the example to demonstrate this: https://codepen.io/anon/pen/vRyMON

In 4.6.2 the whole white area is hittable, whereas in 4.7 only the top-left rectangle is.

ivanpopelyshev:
override containsPoint and not hitArea, then it'll work.

qtiki:
@ivanpopelyshev Do you mean I should override containsPoint in my example or that the pull request code should be modified to use containsPoint instead of hitArea? I'm just wondering whether this is considered a bug in PIXI or "as designed" - which is why I commented here instead of just opening a new issue.

ivanpopelyshev:
@qtiki you should use containsPoint . if you have a shape inside, you can just do

```js
element.containsPoint = function(p) {
return shape.contains(p.x, p.y)`
}
```

Hyperkettu:
@ivanpopelyshev How do you put that shape inside an element so that it maintains its transform hierarchy? Moreover, if I have a container hierarchy that has also sprites as children (buttons), how does that containsPoint method work since containers don't have containsPoint function? Is there any way?

ivanpopelyshev:
`p` here is a point in local coordinates, already reverse transformed.

PR #4612, opened 2018-01-16
Adds `drawStar` method to Graphics

bigtimebuddy:
## Overview

I was looking at old un-submitted additions to Graphics that I created. This one is a very simple function that allows a user to draw a star shape with an arbitrary number of points with Graphics. This change is pretty light-weight because no new internal Shapes are added; it's rendered as a polygon shape. 

I think the functionality here is pretty small and generalized enough to be useful in Graphics and not as a plugin. Also, I'm making this PR to v4 because it's also an easy merge to v5 as well.

### :gift: Added

* Adds `drawStar` method to Graphics

### ✏️ Changed

* Replaced instances in Graphics of `Math.PI * 2` with the `PI_2` PIXI const

### 🔎 Example

https://jsfiddle.net/bigtimebuddy/08kcovyq/

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -5,7 +5,7 @@ import GraphicsData from './GraphicsData';
 import Sprite from '../sprites/Sprite';
 import { Matrix, Point, Rectangle, RoundedRectangle, Ellipse, Polygon, Circle } from '../math';
 import { hex2rgb, rgb2hex } from '../utils';
-import { SHAPES, BLEND_MODES } from '../const';
+import { SHAPES, BLEND_MODES, PI_2 } from '../const';
 import Bounds from '../display/Bounds';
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Agree on v4. For v5 wel'll have problems, i think we can add PI to transform constants there too.

Busyrev:
What about drawMandelbrotFractal()?

Busyrev:
About PI_2, I think it's little bit confusing, in glib PI_2 is PI divided by 2 [glib docs],(https://lazka.github.io/pgi-docs/GLib-2.0/constants.html#GLib.PI_2) but Rust defines PI_2 as PI mutiplied by 2 [rust docs](https://doc.rust-lang.org/1.1.0/std/f32/consts/constant.PI_2.html)

PR #4622, opened 2018-01-19
Update mouse position in onPointerMove, when using a pen tablet

rich-earth:
Update the mouse position in onPointerMove when using a pen tablet (like a wacom), the same as with a mouse. (Browsers that support Pointer events)


bigtimebuddy:
Great work, thanks @rich-earth!

PR #4627, opened 2018-01-22
added min size restriction for Text canvas

dolow:
Fix for this issue,  details described in there.
https://github.com/pixijs/pixi.js/issues/4618

This pull request also contains unit test to ensure minimum canvas size.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -144,8 +144,8 @@ export default class Text extends Sprite
         const maxLineWidth = measured.maxLineWidth;
         const fontProperties = measured.fontProperties;
 
-        this.canvas.width = Math.ceil((width + (style.padding * 2)) * this.resolution);
-        this.canvas.height = Math.ceil((height + (style.padding * 2)) * this.resolution);
+        this.canvas.width = Math.ceil((Math.max(1, width) + (style.padding * 2)) * this.resolution);
+        this.canvas.height = Math.ceil((Math.max(1, height) + (style.padding * 2)) * this.resolution);
 
         context.scale(this.resolution, this.resolution);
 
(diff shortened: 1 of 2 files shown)


cursedcoder:
Straight copy-paste from @bigtimebuddy code, seems legit

src/core/text/Text.js
```diff
@@ -144,8 +144,8 @@ export default class Text extends Sprite
         const maxLineWidth = measured.maxLineWidth;
         const fontProperties = measured.fontProperties;
 
-        this.canvas.width = Math.ceil((width + (style.padding * 2)) * this.resolution);
-        this.canvas.height = Math.ceil((height + (style.padding * 2)) * this.resolution);
+        this.canvas.width = Math.ceil((Math.max(1, width) + (style.padding * 2)) * this.resolution);
```
themoonrat: wouldn't you want the Math.max to apply over the _entire_ calculation? Because as it stands right now, if I had a resolution multiplier of less than 1, I could get a canvas width of less than 1 again

cursedcoder: don't think you ever want to pass negative padding, and here `width` is copied from `measuredWidth` which is `0` because of valid symbol `\n`

bigtimebuddy: But the `ceil` would round that fraction up to 1... `width` = 1.

```js
Math.ceil ((1 + 0) * 0.5);
```
That would equal 1.

themoonrat: Yeah, true.
Just seems odd, that when the goal is to make sure that the width is at least 1, it's done on part of a calculation rather than the whole thing, and relying on a ceil of a max  *shrugs*. Just could read a lot simpler.
I'll remove my request to change since others are happy tho

bigtimebuddy:
Thanks @dolow!

Issue #4624, opened 2018-01-20
Proposal: Add PIXI.extract.pixelSampler(target, x, y)


soimy:
In the current implementation of PIXI.extract, if we want to sample a pixel of `DisplayObject` or `RenderTexture`, we have to extract the entire canvas first. And this is not good for real-time (not static and cached) usage.
I was able to extend the `WebGLExtract.js` to add a `pixelSampler(target: DisplayObject | RenderTexture, x: number, y: number)` method with only simple changes:
``` javascript
gl.readPixels(x * resolution, y * resolution, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, webglPixels);
```
and the latency is around 3-5ms in my late 2014 13inch rmbp. I don't think it's good enough, maybe some lower-level approach is possible but far beyond my abilities.

ivanpopelyshev:
readPixels is the only synchronous function in webgl context, it waits info from GPU. Your program doesn't do anything until it gets that info. Its ok to use it once per frame, after its end (same as preserveDrawingBuffer).

soimy:
OK, If your think this `pixelSampler` API is necissary, I can create a pull request.

themoonrat:
@soimy Is this something you think you could get around to submitting? Would be appreciated! Thanks

Issue #4530, opened 2017-12-11
canvas() on extract double and flipping the drawings


orklann:
Bug demo code: https://jsfiddle.net/kkjwfk1t/

Steps to reproduce:

* Open above link
* Click the `start test` button
* An HTML Canvas element will be insert into the page which double the drawings and flipping it.

But if change line 6:

```javascript
 var app = new PIXI.Application(800, 600, { antialias: true, transparent: true, backgroundColor: 0xffffff});
```

to

```javascript
 var app = new PIXI.Application(800, 600, { antialias: true, transparent: false, backgroundColor: 0xffffff});
```

the bug is gone.

I think this is a bug on WebGL renderer and transparent.

Need help

ivanpopelyshev:
It depends on browser. We think that its always flipped, and flip it back: https://github.com/pixijs/pixi.js/blob/dev/src/extract/webgl/WebGLExtract.js#L98 , however you shown that its not the case.

orklann:
I created another demo which let you download an image presenting the whole renderer without talking to canvas(), and it works normally.

https://jsfiddle.net/4yLu3g3w/ 

And the code is almost the same as the first one, beside last part.

ivanpopelyshev:
```var data = renderer.view.toDataURL();```

That works only if you use "preserveDrawingBuffer". Otherwise, no guarantees.

Issue #3850, opened 2017-03-20
renderer.extract.base64(..) makes wrong size image with devicePixelRatio > 1


bQvle:
Here is a modified version of the RenderTexture example (it will make a popup with the rendered image, so **make sure you allow popups**)

with devicePixelRatio 2, the image will be 4x to big, and only 1/4 of the image area is rendered (top left).

set browser zoom to 200% and refresh the page to simulate PixelRatio 2.

```js
PIXI.settings.RESOLUTION = devicePixelRatio;

var app = new PIXI.Application(800, 600, {
    backgroundColor : 0x1099bb,
    resolution: devicePixelRatio
});
document.body.appendChild(app.view);

var container = new PIXI.Container();
app.stage.addChild(container);

var texture = PIXI.Texture.fromImage('required/assets/basics/bunny.png');

for (var i = 0; i < 25; i++) {
    var bunny = new PIXI.Sprite(texture);
    bunny.x = (i % 5) * 30;
    bunny.y = Math.floor(i / 5) * 30;
    bunny.rotation = Math.random() * (Math.PI * 2)
    container.addChild(bunny);
}

var brt = new PIXI.BaseRenderTexture(300, 300, PIXI.SCALE_MODES.LINEAR, 1);
var rt = new PIXI.RenderTexture(brt);

var sprite = new PIXI.Sprite(rt);

sprite.x = 450;
sprite.y = 60;
app.stage.addChild(sprite);

/*
 * All the bunnies are added to the container with the addChild method
 * when you do this, all the bunnies become children of the container, and when a container moves,
 * so do all its children.
 * This gives you a lot of flexibility and makes it easier to position elements on the screen
 */
container.x = 100;
container.y = 60;

app.ticker.add(function() {
    app.renderer.render(container, rt);
});


setTimeout(function() {window.open(app.renderer.extract.base64(container)); }, 1000);
```

bigtimebuddy:
Is this related to #5016?

themoonrat:
Looks like it could be

Issue #2695, opened 2016-07-01
getCanvas() on a RenderTexture handles alpha incorrectly.


jackie-t:
Similar to https://github.com/pixijs/pixi.js/issues/2302, the getCanvas on a RenderTexture does not appear to handle the alpha calculations correctly.

Step 1: Load an image that relies on an alpha channel. 
Step 2: Call hue(1) on the ColorMatrixFilter and apply that ColorMatrixFilter.
Step 3: Render the texture, using generateTexture.
Step 4: Call getCanvas() on the generated texture.

This should result in a slight change to the hue, barely noticable.
Instead, the color is heavily distorted.

http://codepen.io/bringerofbabies/pen/yJbGdY

(this uses the newest PixiJS as of today)

andrewstart:
This seems to actually be a problem with getCanvas() on a RenderTexture: http://codepen.io/bringerofbabies/pen/yJbGdY
I would guess due to some sort of premultiplied alpha/non-premultiplied alpha mismatch.

jackie-t:
Alright - nice catch. I'll adjust the title, so it's more easy to find.

bastienatome:
I've got a similar problem with the get canvas on a sprite with a custom texture. It seems that came from the getCanvas of the RenderTexture.
http://codepen.io/anon/pen/xEkWRK

mnori:
Here's a pen I did to replicate this problem, using Graphics instead of Sprite

https://codepen.io/wildtype/pen/NjeRMN

setpixel:
+1

ivanpopelyshev:
Oh, that old thing. Thanks to @setpixel we have an answer:

Everything that is rendered in pixi has premultiplied alpha. There's no way to get non-premultiplied info, unless you patch all the object renderers.

To compare the effort, look at #4033, the PR that introduced non-premultiplied alpha in source textures. For NPM targets you'll need even more code.

ivanpopelyshev:
Thus, `canvas()` result is correct only if alpha=1 everywhere. Otherwise, you have to take pixels() and divide result by alpha. Should we add special flag in `canvas()` that allows to do it automatically?

mnori:
Nice, I applied #4632 and passed the premultiply param as true to solve my old codepen.

```
var canvas = renderer.extract.canvas(stagingContainer, null, true);
```

Proof that it works:
https://codepen.io/anon/pen/GQWYZK

Can resurrect my old project now :D

PR #4632, opened 2018-01-23
Visual test that covers DisplayObject, SystemRenderer, Extract.

ivanpopelyshev:
Update 20/02/2018: It is WIP again, I've just found a way to make `generateTexture` and `extract` results more consistent, and clean up the code.

## Overview

I was going to add a few params in Extract API, and made a test. I dont like to make tests, but when I do, my test breaks several subsystems and leads to refactoring.

I can split to several individual tests and rework API's in v5.

That PR was completed in friday night.

### :gift: Added

* `renderer.extract.data` accepts `region` and `resolution` parameters, returns `CanvasData` instant that containss pixel data, their format, region and resolution it was taken from
* Adds `CanvasEquals` tests module that takes into account flipY and premultiplyAlpha problems
* Adds megatest that covers several subsystems

### ✏️ Changed

* `generateTexture` API, new `region` parameter replaces `scaleMode`, which is now the final parameter.
* All temporary parent/transform operations are moved in `UpdateTransformHelper` class.

### 🐛 Fixed

* All temporary changes of parent and transform in `DisplayObject`, now take into account several cases of lazy displayObject evaluations.
* `CanvasRenderer` and `WebGLRenderer` use `UpdateTransformHelper` API and don't participate in those hacks.

### 🔎 Solved issues

#2695
#4624
#4530
#4670, [fiddle](https://jsfiddle.net/ASCII_HUSKY/7dt7fpsv/27/)
#3850, resolution can be passed through new function


ivanpopelyshev:
@setpixel try this build: pixijs.download/dev-extract-params/pixi.js , submit "true" as third parameter to canvas/pixels.

bigtimebuddy:
I think we need some unit-tests here, especially for `flipY`, `postDivide` and support for the `region` param.

ivanpopelyshev:
Yep, lets wait when someone has free time for this thing. 

I made it to address all those issues, its nor critical nor high priority ;)

ivanpopelyshev:
Now we have a class that combines data, dimensions and flags (flip, premultiply).

In v5 its better to just add `extract` function to renderer as a mixin that returns it.

I've added a test but it is failing :(

test/extract/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+const EqualsData = require('./equals');
+
+describe('PIXI.extract', function ()
+{
+    it('should use generateTexture, check the region, premultiplyAlpha and flipY', function (done)
+    {
+        const renderers = [new PIXI.CanvasRenderer(36, 47, { transparent: true })];
+
+        // renderers.pop();
```
cursedcoder: remove?

test/extract/equals.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * equality that takes into account flips and multiplied alpha
+ * @param {PIXI.extract.CanvasData} arg1 the other canvas data
+ * @param {PIXI.extract.CanvasData} arg2 the other canvas data
+ * @returns {boolean} result
+ */
+
+/* eslint-disable no-console */
+module.exports = function equals(arg1, arg2)
```
cursedcoder: `export default function`

test/extract/equals.js
```diff
@@ -0,0 +1,66 @@
...
+            alpha2 = arg2.premultiplyAlpha ? 1.0 : (alpha2 / 255.0);
+
+            for (let t = 0; t < 3; t++)
+            {
+                const R1 = Math.round(pixels1[ind1] * alpha1);
+                const R2 = Math.round(pixels2[ind2] * alpha2);
+
+                if (Math.abs(R1 - R2) > 1)
+                {
+                    console.log(`Image comparsion: color fail x=${x} y=${y}, a1=${alpha1}, a2=${alpha2},
```
cursedcoder: Image compa*_riso_*n

test/extract/equals.js
```diff
@@ -0,0 +1,66 @@
...
+
+                return false;
+            }
+
+            alpha1 = arg1.premultiplyAlpha ? 1.0 : (alpha1 / 255.0);
+            alpha2 = arg2.premultiplyAlpha ? 1.0 : (alpha2 / 255.0);
+
+            for (let t = 0; t < 3; t++)
+            {
+                const R1 = Math.round(pixels1[ind1] * alpha1);
```
cursedcoder: What does `R` mean here?

src/extract/webgl/WebGLExtract.js
```diff
@@ -8,10 +9,12 @@ const BYTES_PER_PIXEL = 4;
  *
  * An instance of this class is automatically created by default, and can be found at renderer.plugins.extract
  *
+ * All methods use global coordinates of the target. Please assign temporary parent and call updateTransform to use local.
+ *
  * @class
  * @memberof PIXI.extract
  */
-export default class WebGLExtract
```
cursedcoder: Why remove?

src/extract/CanvasData.js
```diff
@@ -0,0 +1,205 @@
+import * as core from '../core';
+
+/**
+ * Contains extracted data
+ * @class
+ * @memberof PIXI.extract
+ */
+export default class CanvasData
```
cursedcoder: I like it, +1

test/extract/equals.js
ivanpopelyshev: Can't, tests are old node modules.

test/extract/equals.js
ivanpopelyshev: Premultiplied color.

ivanpopelyshev:
I can make temporary displayObject and Transform global. If something taints them then whole pixi will be unusable, not only one object.

`copy` fields store real values of `parent` and `transform`, one is temporary transform for `getLocalBounds` made the same way temporary parent. One flag to store the type of operation performed on the element. 

I guess we can move it to one big temporary object to reduce number of fields.

The problem with those functions is that I cant reduce them - everything less will fail my megatest in strange ways, for example when we call second `renderer.render` it can break because of `wrongID` weren't updated properly when doing those operations.

bigtimebuddy:
Hey @ivanpopelyshev: no refactors for v4, sorry. Only new features for v4 if they are extremely high value. This is not, IMO, and would be "nice to have". Will _only_ consider the API additions for extract in v5.

Your bug fixes are useful, but please do it one PR at a time. It's too difficult to verify things are fixed when you do it all together and it's more likely to cause instability issues.

PR #4636, opened 2018-01-24
Fix ColorMatrixFilter negative effect

inukshuk:
First off, thanks for PixiJS!

I have been experimenting with `ColorMatrixFilter` and was confused by the `negative` effect: my understanding was that it would invert the RGB colors, but that did not seem to be the case (as simple case try to invert a white shape on a black background); either I'm doing something wrong or the inversion matrix is simply wrong?

Using the matrix:

```
-1  0  0  0  1
 0 -1  0  0  1
 0  0 -1  0  1
 0  0  0  1  0
```

produces negatives for me.

On a related note, I noticed that multiplying effects (using the `multiply` parameter) produces strange results (using both built-in matrices and my own), because of the adjustments done by `_colorMatrix()`. If I multiply effects using `_multiply` and set the resulting matrix myself the result looks good to me (both the matrix and the visual effect).

### DIFF
--- a/src/filters/colormatrix/ColorMatrixFilter.js
+++ b/src/filters/colormatrix/ColorMatrixFilter.js
@@ -299,10 +299,10 @@ export default class ColorMatrixFilter extends core.Filter
     negative(multiply)
     {
         const matrix = [
-            -1,  0,  0,  0,  1,
-             0, -1,  0,  0,  1,
-             0,  0, -1,  0,  1,
-             0,  0,  0,  1,  0,
+            -1, 0, 0, 0, 1,
+            0, -1, 0, 0, 1,
+            0, 0, -1, 0, 1,
+            0, 0, 0, 1, 0,
         ];
 
         this._loadMatrix(matrix, multiply);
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks @inukshuk. 

Would you mind fixing those [linting errors](https://travis-ci.org/pixijs/pixi.js/builds/332956759#L569-L585)? You can run `npm run lint` locally to verify.

inukshuk:
Sure, hope it's fixed now!

ivanpopelyshev:
Btw, here is sprite renderer that can produce negative effect, less cost than ColorMatrixFilter: https://github.com/gameofbombs/pixi-heaven/

inukshuk:
So should that be:

```
-1  0  0  1  0
 0 -1  0  1  0
 0  0 -1  1  0
 0  0  0  1  0
```

Or rather:

```
-1  0  0  1  1
 0 -1  0  1  1
 0  0 -1  1  1
 0  0  0  1  0
```

Thanks!

ivanpopelyshev:
First one.

Imagine that we have (Ra,Ga,Ba,a,1) and the result has to be (a - Ra, a - Ga , a - Ba, a)

bigtimebuddy:
@ivanpopelyshev could you approve. I believe @inukshuk made that change as you described

bigtimebuddy:
Thanks @inukshuk 🎉

PR #4644, opened 2018-01-28
Improve setChildIndex error message

artvinn:
Add supplied index to setChildIndex error message. Test updated.


bigtimebuddy:
Thanks @artvinn!

PR #4648, opened 2018-01-29
add option: forceNative

finscn:
details at https://github.com/pixijs/pixi.js/issues/4646

### DIFF
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -53,6 +53,7 @@ export default class Application
      *  for devices with dual graphics card **webgl only**
      * @param {boolean} [options.sharedTicker=false] - `true` to use PIXI.ticker.shared, `false` to create new ticker.
      * @param {boolean} [options.sharedLoader=false] - `true` to use PIXI.loaders.shared, `false` to create new Loader.
+     * @param {boolean} [options.forceNative=false] - Whether use the native methods or the WebGL extension.
      */
     constructor(options, arg2, arg3, arg4, arg5)
     {
--- a/src/core/Application.js
+++ b/src/core/Application.js
@@ -53,7 +53,7 @@ export default class Application
      *  for devices with dual graphics card **webgl only**
      * @param {boolean} [options.sharedTicker=false] - `true` to use PIXI.ticker.shared, `false` to create new ticker.
      * @param {boolean} [options.sharedLoader=false] - `true` to use PIXI.loaders.shared, `false` to create new Loader.
-     * @param {boolean} [options.forceNative=false] - Whether use the native methods or the WebGL extension.
+     * @param {boolean} [options.forceNative=false] - Whether use the native methods or the WebGL extension. **webgl only**
      */
     constructor(options, arg2, arg3, arg4, arg5)
     {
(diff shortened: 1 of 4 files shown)


src/core/renderers/webgl/WebGLRenderer.js
```diff
@@ -56,14 +56,16 @@ export default class WebGLRenderer extends SystemRenderer
      *  with older / less advanced devices. If you experiance unexplained flickering try setting this to true.
      * @param {string} [options.powerPreference] - Parameter passed to webgl context, set to "high-performance"
      *  for devices with dual graphics card
+     * @param {Boolean} [options.forceNative=false] - Whether use the native methods or the WebGL extension.
```
bigtimebuddy: Documentation should also be duplicated in Application and autoDetectRenderer

themoonrat: and it would be boolean, not Boolean

finscn: done.

src/core/autoDetectRenderer.js
```diff
@@ -32,6 +32,7 @@ import WebGLRenderer from './renderers/webgl/WebGLRenderer';
  *  If you experience unexplained flickering try setting this to true. **webgl only**
  * @param {string} [options.powerPreference] - Parameter passed to webgl context, set to "high-performance"
  *  for devices with dual graphics card **webgl only**
+ * @param {boolean} [options.forceNative=false] - Whether use the native methods or the WebGL extension.
```
bigtimebuddy: Could you add `**webgl only**` here

src/core/Application.js
```diff
@@ -53,6 +53,7 @@ export default class Application
      *  for devices with dual graphics card **webgl only**
      * @param {boolean} [options.sharedTicker=false] - `true` to use PIXI.ticker.shared, `false` to create new ticker.
      * @param {boolean} [options.sharedLoader=false] - `true` to use PIXI.loaders.shared, `false` to create new Loader.
+     * @param {boolean} [options.forceNative=false] - Whether use the native methods or the WebGL extension.
```
bigtimebuddy: Same here, `**webgl only**` comment to this option

finscn: done

src/core/autoDetectRenderer.js
finscn: done

ivanpopelyshev:
By the way, state is not re-created after context loss. We have to do something about it.

bigtimebuddy:
Why not just do this below? These are all public properties. Would like to not add a convenience setting for a very specific platform bug (old Android).

```js
app.renderer.state.nativeVaoExtension  = null;
PIXI.glCore.VertexArrayObject.FORCE_NATIVE;
```
I'm in favor of closing this.

finscn:
I don't understand @ivanpopelyshev 's words.
And I agree with "not add a convenience setting for a very specific platform bug"

bigtimebuddy:
Thanks for the PR @finscn. Sorry it took so long to resolve.

PR #4655, opened 2018-02-01
fix comment mistake for matrix.js

### LINKED ISSUE #4654: comment mistake in Matrix.js
the original comment for Matrix class
```javascript
/**
 * The PixiJS Matrix class as an object, which makes it a lot faster,
 * here is a representation of it :
 * | a | b | tx|
 * | c | d | ty|
 * | 0 | 0 | 1 |
 *
 * @class
 * @memberof PIXI
 */
```

according to the source code , I think the matrix should be like this(also same as canvas [transform](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/transform) api).
```javascript
/**
 * | a | c | tx|
 * | b | d | ty|
 * | 0 | 0 | 1 |
 */
```
and comment for constructor should be like this(b is x skew,c is y skew).
```javascript
/**
  * @param {number} [a=1] - x scale
  * @param {number} [b=0] - x skew
  * @param {number} [c=0] - y skew
  * @param {number} [d=1] - y scale
  * @param {number} [tx=0] - x translation
  * @param {number} [ty=0] - y translation
 */
```

koalaylj:
fix #4654  comment mistake in Matrix.js

### DIFF
--- a/src/core/math/Matrix.js
+++ b/src/core/math/Matrix.js
@@ -3,8 +3,8 @@ import Point from './Point';
 /**
  * The PixiJS Matrix class as an object, which makes it a lot faster,
  * here is a representation of it :
- * | a | b | tx|
- * | c | d | ty|
+ * | a | c | tx|
+ * | b | d | ty|
  * | 0 | 0 | 1 |
  *
  * @class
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
We have a problem with Travis, I'll approve it as soon as we solve it.

bigtimebuddy:
Looks like this is related to: https://github.com/SitePen/remap-istanbul/issues/162, I'll fix the depencencies on **floss** which should fix the problem.

PR #4675, opened 2018-02-15
Matrix fix

ceco-fmedia:
Fixed Matrix.setTransform gives different result from Transform.updateLocalTransform.

`Return rotation instead of skew if skewX+skewY is slightly below 360 degrees.`


ivanpopelyshev:
I dont know why do we need this method anyway :)

cursedcoder:
Adding a test case for that fix is highly appreciated.

ivanpopelyshev:
I can't double-approve :(

src/core/math/Matrix.js
```diff
@@ -363,7 +351,7 @@ export default class Matrix
 
         const delta = Math.abs(skewX + skewY);
 
-        if (delta < 0.00001)
+        if (delta < 0.00001 || Math.abs((2 * Math.PI) - delta) < 0.00001)
```
bigtimebuddy: Minor thing, could you use import `PI_2` from **src/core/const.js** instead of using `2 * Math.PI`?

ivanpopelyshev:
We need to fix the random seed. Is there any way to do it in JS? If scale is near zero, transform will go fubar.

ivanpopelyshev:
I think we have to add [seedrandom](https://github.com/davidbau/seedrandom) to dev-dependencies and use it there.

GoodBoyDigital:
you da man @ceco-fmedia 👍

PR #4682, opened 2018-02-20
Filter fullScreen mode

ivanpopelyshev:
Finally solution for all the problems of [V4 filters](https://github.com/pixijs/pixi.js/wiki/v4-Creating-Filters)

It is emulation of PixiJS v3.

I'm feeling stupid.

### ✏️ Changed

* FilterManager pool has a bucket for screen-sized renderTargets.
* FilterManager detects if `filterArea` is equal to `renderer.screen` and turns on fullScreen mode

### 🐛 Example

DisplacementFilter old style: https://jsfiddle.net/Hackerham/3kvazjw1/7/ , move the glass to the border and see that there are no black artifacts. 

Those black artifacts can be seen with current version of pixi: https://jsfiddle.net/Hackerham/3kvazjw1/11/ .

This fiddle is a big different than in `pixi-examples`, it has fixed `filterArea = app.screen` and `padding` is not zero, that's why it has black edges.

### DIFF
--- a/src/core/renderers/webgl/filters/Filter.js
+++ b/src/core/renderers/webgl/filters/Filter.js
@@ -99,6 +99,12 @@ export default class Filter
          * @member {boolean}
          */
         this.autoFit = true;
+
+        /**
+         * If enabled, filterArea for element will be same size of the screen. PixiJS v3 mode.
+         * @member {boolean}
+         */
+        this.fullScreen = false;
     }
 
     /**
--- a/src/core/renderers/webgl/filters/Filter.js
+++ b/src/core/renderers/webgl/filters/Filter.js
@@ -99,12 +99,6 @@ export default class Filter
          * @member {boolean}
          */
         this.autoFit = true;
-
-        /**
-         * If enabled, filterArea for element will be same size of the screen. PixiJS v3 mode.
-         * @member {boolean}
-         */
-        this.fullScreen = false;
     }
 
     /**
--- a/src/core/renderers/webgl/managers/FilterManager.js
+++ b/src/core/renderers/webgl/managers/FilterManager.js
@@ -102,7 +102,7 @@ export default class FilterManager extends WebGLManager
         const fullScreen = target.filterArea
             && target.filterArea.x === 0
             && target.filterArea.y === 0
-            %% target.filterArea.width === renderer.screen.width
+            && target.filterArea.width === renderer.screen.width
             && target.filterArea.height === renderer.screen.height;
 
         // for now we go off the filter of the first resolution..
--- a/src/core/renderers/webgl/managers/FilterManager.js
+++ b/src/core/renderers/webgl/managers/FilterManager.js
@@ -26,6 +26,8 @@ class FilterState
     }
 }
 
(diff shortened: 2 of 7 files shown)


ivanpopelyshev:
Work in progress: we have to add `All fullscreen` setting, to emulate v3 for all filters.

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -603,8 +644,40 @@ export default class FilterManager extends WebGLManager
...
+            if (textures)
+            {
+                for (let j = 0; j < textures.length; j++)
+                {
+                    textures[j].destroy(true);
+                }
+            }
+            this.pool[screenKey] = [];
+        }
+    }
```
GoodBoyDigital: Hey, instead of destroying here, we could resize render targets instead?

src/core/renderers/webgl/filters/Filter.js
```diff
@@ -99,6 +99,12 @@ export default class Filter
          * @member {boolean}
          */
         this.autoFit = true;
+
+        /**
+         * If enabled, filterArea for element will be same size of the screen. PixiJS v3 mode.
+         * @member {boolean}
+         */
+        this.fullScreen = false;
```
GoodBoyDigital: do we need this? as you have the check further down that compares sizes?

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -85,16 +99,32 @@ export default class FilterManager extends WebGLManager
...
         }
 
+        let fullScreen = this.fullScreen;
+
+        if (renderTargetFrame.width === renderer.screen.width
+            && renderTargetFrame.height === renderer.screen.height)
+        {
+            for (let i = 0; i < filters.length; i++)
+            {
+                if (filters[i].fullScreen)
```
GoodBoyDigital: is this required? If the bounds match, we should always be full screen?

ivanpopelyshev: That's our current renderTarget size. If a non-fullscreen filter is above us.

src/core/renderers/webgl/filters/Filter.js
ivanpopelyshev: It forces use fullScreen for this filter.

src/core/renderers/webgl/managers/FilterManager.js
ivanpopelyshev: Extra destroy is fine, I dont want to get a sudden memory increase from 640x480 to 2560x1280.

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -51,6 +53,11 @@ export default class FilterManager extends WebGLManager
         this.filterData = null;
 
         this.managedFilters = [];
+
+        this.renderer.on('prerender', this.onPrerender, this);
```
bigtimebuddy: Can you make sure to remove this event listener on destroying FilterManager?

PR #4686, opened 2018-02-21
Filter backdrop

ivanpopelyshev:
Hello, new blendmodes!

Big problem: We cant merge this thing before we get rid of "non-compatible format" error in console. It existed in `pixi-picture` and it exists here, I dont know what's wrong with my implementation :(

### :gift: Added

* `filter.backdropUniformName` if set, filterManager copies backdrop buffer to the filter input, only if current renderTarget is a Framebuffer and not root.
* `filter.clearColor` specifies color to clear filter input.

### 🐛 Example

[JsFiddle](https://jsfiddle.net/Hackerham/xgnq4fc9/5/)

One displacementFilter takes care of two sprites at the same time. Try remove clearColor attribute and see a disaster in the center and around the sprites (filter padding)

### DIFF
--- a/src/core/renderers/webgl/filters/Filter.js
+++ b/src/core/renderers/webgl/filters/Filter.js
@@ -99,6 +99,19 @@ export default class Filter
          * @member {boolean}
          */
         this.autoFit = true;
+
+        /**
+         * If set, filterManager calculates background image and stores location in to specified uniform
+         * @member {string|null}
+         */
+        this.backdropUniformName = null;
+
+        /**
+         * temporary storage for backdrop render texture
+         * @member {PIXI.RenderTexture}
+         * @private
+         */
+        this._backdropRenderTexture = null;
     }
 
     /**
--- a/src/core/renderers/webgl/filters/Filter.js
+++ b/src/core/renderers/webgl/filters/Filter.js
@@ -102,12 +102,14 @@ export default class Filter
 
         /**
          * If set, filterManager calculates background image and stores location in to specified uniform
+         * Backdrop is explained here: https://www.w3.org/TR/compositing-1/#backdropCalc
+         * All those blending modes are possible to implement through this feature
          * @member {string|null}
          */
         this.backdropUniformName = null;
 
         /**
-         * temporary storage for backdrop render texture
+         * Temporary storage for backdrop render texture
          * @member {PIXI.RenderTexture}
          * @private
          */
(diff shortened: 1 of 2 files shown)


cursedcoder:
Would be nice to add link for backdrop https://www.w3.org/TR/compositing-1/#blendingsoftlight

GoodBoyDigital:
Yo! thanks for this dude. This is clever stuff, although I think that in its current state this feature may be better placed inside a filter itself that requires it. Mainly as there are no common use cases for this currently. 

Another reason is that this feature only works when the user has bound a render texture, it may lead to misleading results. 

I would be keen to explore ways of making this work even without a render texture bound (render scene to frame buffer and then use a copy pixels to add it to the main screen?)

We could also caveat it as an expensive feature that may affect performance. But the plus side would be that we could create advanced blend modes using filters :)

ivanpopelyshev:
> Mainly as there are no common use cases for this currently.

I just gave you the case in the fiddle - the DisplacementFilter. It can be reproduced in vanilla pixi but its too hard for people to use it that way. I'm decreasing the complexity of usage of existing feature that people want, not inventing something new. Ask @bqvle or @nuthinking in the slack.

> Another reason is that this feature only works when the user has bound a render texture, it may lead to misleading results.

There's no way out of it, readPixels will give misleading results in preformance. We can bind temporary texture there to show that result is incorrect.

> I would be keen to explore ways of making this work even without a render texture bound (render scene to frame buffer and then use a copy pixels to add it to the main screen?)

Its about FrameBuffer, not the render texture.

> We could also caveat it as an expensive feature that may affect performance. But the plus side would be that we could create advanced blend modes using filters :)
Same performance as filters. And one advanced displacement filter that manages N sprites is definitely faster than N old displacements.

ivanpopelyshev:
Ok, after conversation with @GoodBoyDigital we've decided to add this after Stage PR.

rvion:
Just found this thread while looking for a solution to displacement-map borders:

I use pixi v4+ pixi-display (so practical, makes everything easier)
water layer contains water tiles + reflections
Displacment map produces visual glitch near borders.

jul-23-2018 19-35-36

is there a recommended solution until this is merged ?

ivanpopelyshev:
@rvion join pixi slack and we'll help you. there are number of people who know solutions for that problem.

ivanpopelyshev:
ok, you can remove it now. join #shaders and #rpgmaker , people will help you :)

ivanpopelyshev:
I gave it afterlife: https://gameofbombs.github.io/examples-heaven/#/picture/displacement-backdrop.js

PR #4687, opened 2018-02-21
Feature/filters

GoodBoyDigital:
Filter manager re-write complete
- less code
- less draw commands per filter
- any filter can have any resolution
- happy days
- bundled filters converted and tested and working


packages/core/src/renderers/filters/Filter.js
```diff
@@ -92,6 +56,8 @@ export default class Filter extends Shader
          * @member {boolean}
          */
         this.autoFit = true;
+
+        this.state = new State();
```
bigtimebuddy: This could use a jsdoc

packages/filters/filter-fxaa/src/fxaa.frag
```diff
@@ -113,11 +114,11 @@ vec4 fxaa(sampler2D tex, vec2 fragCoord, vec2 resolution,
 
 void main() {
 
-      vec2 fragCoord = vTextureCoord * filterArea.xy;
+      vec2 fragCoord = vTextureCoord * destinationFrame.ba;
```
ivanpopelyshev: Please dont use rgba fields for non-colors.

GoodBoyDigital: fair point :)

packages/core/src/renderers/filters/Filter.js
GoodBoyDigital: done!

packages/filters/filter-fxaa/src/fxaa.frag
GoodBoyDigital: done!

ivanpopelyshev:
Proposal.

All global uniforms should be "filterXXXX".

1. rename `sourceFrame` to `filterFrame`
2. rename `destinationFrame` to `filterSize` or something like that, make it vec2.
3. support old `filterArea` and `filterClamp` for the sake of compatibility
4. add `filterPixel`, its value is `1.0 / filterSize / resolution`

ivanpopelyshev:
After those changes we'll check compatibility with `pixi-filters`. @finscn, @bigtimebuddy and I made such a great job with it, and users have special filters that are based on them. 

It would be a shame if the implementation wont be compatible.

ivanpopelyshev:
Legacy mode for filter attributes: aVertexPosition and aTextureCoord should be supported too.

ivanpopelyshev:
filterArea shouldn't be transformed, its up to user to do that, he can override object `getBounds()` and it'll be transformed properly.

ivanpopelyshev:
Ok, you can merge it but I fix all those problems tomorrow. We need some filters implementation, this one is fine but we cant put it in release candidate.

GoodBoyDigital:
Auto fit added along with clamping! @ivanpopelyshev makes sense! Agree that there is a more to be done for RC. But this gets us way closer as we now have filters in v5 👍 

- The next step is to figure out the best way to get all those lovely existing filters working in v5.
- Once we have that nailed, lets PR any new features separately.

PR #4688, opened 2018-02-21
First pass at tidying up shader

GoodBoyDigital:
- reduced GLShader to be a storage object
- also GLShader renamed to GLProgram as it is attached to programs
- This above two points bring it in line with GLTexture and GLBuffer
- Now shader inspection only happens once instead of twice!
- Removed unused classes
- moved precision check to program

This paves the way for a some flattening and better grouping of the core pix package


GoodBoyDigital:
@bigtimebuddy sorted, still getting that weird lint error in Graphics renderer, which is clearly fixed!

PR #4697, opened 2018-02-24
Bugfix/uniform global bug

GoodBoyDigital:
oops.. pretty silly mistake by me here.

Global uniforms are now added in he correct place. This would have led to some bizarre shader issues!

At least its an easy fix :)


bigtimebuddy:
@GoodBoyDigital something's wrong with your PR here. Could you rebase?

PR #4698, opened 2018-02-24
Update README.md

jakubrpawlowski:
Helps npm users understand the right way to import pixi without digging into node modules and looking at the source code.
https://github.com/pixijs/pixi.js/issues/3204#issuecomment-256751902

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,11 @@ _Note: After v4.5.0, support for the [Bower](https://bower.io) package manager h
 ```sh
 npm install pixi.js
 ```
+There is no default export. The correct way to import PixiJS is:
+
+```js
+import * as Pixi from 'pixi.js'
+```
 
 #### CDN Install (via cdnjs)
 
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ npm install pixi.js
 There is no default export. The correct way to import PixiJS is:
 
 ```js
-import * as Pixi from 'pixi.js'
+import * as PIXI from 'pixi.js'
 ```
 
 #### CDN Install (via cdnjs)


README.md
```diff
@@ -51,6 +51,11 @@ _Note: After v4.5.0, support for the [Bower](https://bower.io) package manager h
 ```sh
 npm install pixi.js
 ```
+There is no default export. The correct way to import PixiJS is:
+
+```js
+import * as Pixi from 'pixi.js'
```
bigtimebuddy: Could you do PIXI instead of Pixi, to be consistent with documentation?

dekdevy:
Random question, but in which use-cases would you actually want to import the /lib/ based Pixi.js as ES6 in its current state? It wont work out of the box, or am I wrong?

For example: Right now it would be incredibly useful to be able to use Pixi.js with rollup. However, when using 
```js
```
to import it, it breaks in a lot of places - obviously, because `` node_modules\pixi.js\lib\ `` is actually CJS styled code, so you'll get tons of errors: 

```
(!) 'autoDetectRenderer' is not exported by 'node_modules\pixi.js\lib\index.js'
(!) 'Container' is not exported by 'node_modules\pixi.js\lib\index.js'
(!) 'Graphics' is not exported by 'node_modules\pixi.js\lib\index.js'
```

The list of issues goes on, such as having to resolve / browser shim a lot of little modules that Pixi uses.
My point is: How does anyone make effective use of `` import * as PIXI from 'pixi.js' `` without lots and lots of manual patching? Is this not a bit confusing to be in the main Readme?

ivanpopelyshev:
```js
import 'pixi.js';
```

Now you can use globals.

bigtimebuddy:
We do not generate an ES build for rollup (we will in v5), you need to use rollup-plugin-commonjs to define exports if you’re going to import that way with rollup.

dekdevy:
I appreciate you two taking the time to answer.

I've tried doing all that a few weeks ago (any form of import syntax imaginable) - also using ``rollup-plugin-commonjs``, however, the problems didn't stop there. I also needed ``rollup-plugin-node-resolve`` since there are quite a few require()s in the code still, and then I also needed to provide shims for the node globals that are used in the code, such as ``url`` which can be done via ``rollup-plugin-node-builtins``. (I had to manually patch some other things that rollup plugins couldnt handle). Ultimately though I could not get it to work because some parts of the compilation process that is done by ``pixify`` could not be emulated in my build process.

For laughs I also decided to try to import the actual PIXI.js source from the git repo, which worked much smoother (since most of the code works as rollup would expect it, except some parts of the import synthax). Finally though I think I gave up because I did not manage to get rollup to resolve the shader files if I remember correctly. It was a bit of a shame because I could already see rollup doing its work - treeshaking everything that my project didnt need - I think I got it to shake off several hundred kilobytes of code.
@bigtimebuddy Did you ever manage to get a rollup based project running with an imported PIXI.js?

ivanpopelyshev:
We dont have time to do it for v5, and tree-shaking is useless for us because there's bundle part: 
https://github.com/pixijs/pixi.js/blob/next/bundles/pixi.js/src/index.js

You can take those modules and build your own pixi but you need bundle anyway to register available renderers and for other things. Basically, you have to do manual tree-shaking anyway.

ivanpopelyshev:
For v4 you can just remove useless modules from "index.js" file, this operation corresponds to bundle editing in v5. 

I doubt that there's a way to get rid of bundles for 2d/3d renderers and webgl projects in general.

dekdevy:
I see, thank you for the link to the v5 bundle. I will try to work with that soon and im excited to see v5 come along.  
I'm not involved with Pixi development so I have no doubts that the architectural changes for a process like this are really complex. I recently saw Rich Harris' commit from when [THREE.js switched to modules and rollup](https://github.com/mrdoob/three.js/pull/9310) and it sure looks absolutely massive.

bigtimebuddy:
Yup, we have made a similar massive change in v5 too to be individual packages + rollup. It's working out pretty well so far and will be awesome once we release.

As for v4, I _know_ there's a solution for Rollup, and I'll see if I can dig into it. It's a magic combo of the plugins mentioned above with some config settings. Rollup certainly wasn't a use-case that we designed for in v4. Webpack has been the packager of choice for many users here so most of the issues are related to that.

bigtimebuddy:
@dekdevy Here's a simple example of using Rollup + PixiJS:

https://github.com/bigtimebuddy/test-pixijs-rollup

My recommendation, however, is to treat PixiJS as a global resource. Because it's CommonJS, you don't get any of the tree-shaking _AND_ it will add serious extra time to your building/watching with Rollup.  For instance, I'm doing this with: https://github.com/pixijs/pixi-text-style

dekdevy:
@bigtimebuddy a massive thank you for taking your time with this, ill definitely give it a try! <3

Regarding your second statement: Does this change with v5, meaning tree-shaking will become more effective? I remember, when I attempted to import the source code directly ( not the compiled code in /lib/), that the tree-shaking was definitely having an impact. I saw your other comment on that one Issue you closed, and ill try to see what I get with v5 as well, Ill report back here.

bigtimebuddy:
Yes, that makes sense. Tree-shaking will improve with v5 because we are generating a CommonJS file and an ES module with Rollup, so you don't have to use the src directly.

dekdevy:
Ok, just as a little follow up in case you are interested, @bigtimebuddy.
Your example has helped me a lot and it is fully functioning, thanks again! The little gotchas are easy to implement.
The savings on file-size are minimal, like you already suggested, but I think I will still use it in production. There are other parts in the equation that make it interesting for me, which I understand that not most people would need. (For example, I can now property-mangle the entire codebase including pixi, which might be useless to most people but very helpful in my case).
The only drawback so far with this method is that I cannot have a custom compile (exclusion of modules that I do not need), but I can theoretically fix this by forking PIXI.

I also tried my luck with v5.  And I would be happy to use a beta version in production, but I dont think there is any way for me to include v5 through NPM right now because the repository lacks a functioning package.json file. (`npm install git://github.com/pixijs/pixi.js#next`)

bigtimebuddy:
FYI, we should have v5 on npm in a few weeks.

PR #4702, opened 2018-02-24
Blur Filter tweak

GoodBoyDigital:
- Merged ```BlurXFilter``` and ```BlurYFilter``` to
```BlurFilterPass```. (They shared 90%^ same code)
- X or Y Blur is skipped if strength of that axis is 0.
- Added ```repeatEdgePixels``` #4690

### CODE REVIEW

**Comment by bigtimebuddy** [comment]
@GoodBoyDigital, awesome work, I love when we delete code.

I think we should add BlurXFilter and BlurYFilter to deprecations in the **pixi.js** bundle. If we export BlurFilterPass, it would be easy to provide 1-to-1 deprecations without breaking anything. For example:

```js
PIXI.filters.BlurXFilter = class BlurXFilter extends PIXI.filters.BlurFilterPass {
    constructor(strength, quality, resolution, kernelSize) {
       super(true, strength, quality, resolution, kernelSize);
       warn('BlurXFilter has been deprecated use BlurFilterPass');
    }
}
```

PR #4705, opened 2018-02-26
Fix issue with nested filters

GoodBoyDigital:
Added legacy check to update


bigtimebuddy:
**Broken:** https://jsfiddle.net/zd9o0pxj/46/
**Fixed:** https://jsfiddle.net/bigtimebuddy/zd9o0pxj/47/

PR #4707, opened 2018-02-26
Feature/core flatten

GoodBoyDigital:
Heres the first pass at flattening the structure of core. 

Definitely feels better! but not 100% happy with it yet!

Some thoughts:
- WebGLSystem <- where should that go?
- AbstractRenderer <- as we are WebGL first now, we should remove this

Feedback welcome!


ivanpopelyshev:
Cannot read property 'FilterSystem' of undefined

bigtimebuddy:
Overall this is really awesome and much easier to find things. I like the systems grouped this way and not all tucked into nested files.

I have a few suggestions:

- We should export all system classes in an namespace (like "resources") Actually, I believe that Travis failing because I added a deprecation for `PIXI.systems.FilterSystem`, so you'll need to add that back. I like exposing these classes because they are so integral to v5. It should be easy for users to extend or mixin methods to these classes.
- I think WebGLSystem should be renamed to `RendererSystem` or simply `System`. We've removed "WebGL" from most everything in v5, makes sense to do here too.

GoodBoyDigital:
done! although not 100% sure i did the namespace correctly!

packages/core/src/System.js
```diff
@@ -1,9 +1,9 @@
 /**
  * @class
- * @extends PIXI.systems.WebGLSystem
+ * @extends PIXI.systems.System
```
ivanpopelyshev: oops, that was my mistake

packages/core/src/batch/BatchSystem.js
```diff
@@ -1,13 +1,13 @@
-import WebGLSystem from './WebGLSystem';
-import ObjectRenderer from '../utils/ObjectRenderer';
+import System from '../System';
+import ObjectRenderer from './ObjectRenderer';
 
 /**
  * @class
- * @extends PIXI.WebGLSystem
+ * @extends PIXI.System
```
ivanpopelyshev: no namespace

packages/core/src/batch/ObjectRenderer.js
```diff
@@ -1,13 +1,13 @@
-import WebGLSystem from '../systems/WebGLSystem';
+import System from '../System';
 
 /**
  * Base for a common object renderer that can be used as a system renderer plugin.
  *
  * @class
- * @extends PIXI.WebGLSystem
+ * @extends PIXI.System
```
ivanpopelyshev: namespace missing

bigtimebuddy: `PIXI.System` is the main class and `PIXI.systems.*`  are where the actual system classes will live like TextureSystem, FilterSystem, etc.

PR #4711, opened 2018-02-27
WebGL 2 Lets do it :)

GoodBoyDigital:
Lets enable webGL 2, petty stable!


ivanpopelyshev:
Do you know that we can remove "pow2" obstruction for webgl2?

PR #4713, opened 2018-02-27
Bugfix/fromBuffer

GoodBoyDigital:
Fixed a from buffer issue


ivanpopelyshev:
Actually, there's matter of premultiplyAlpha there. BufferResource can upload premultiplied stuff.

ivanpopelyshev:
I found it: premultipliedAlpha is taken into account in BufferResource

https://github.com/pixijs/pixi.js/blob/next/packages/core/src/textures/resources/BufferResource.js#L46

I think its better to leave it false by default, because BufferResource will be mostly used for data for Float32 textures.

PR #4716, opened 2018-02-27
fixed optimisation for buffer upload!

GoodBoyDigital:


ivanpopelyshev:
oh , right, i forgot you had the bug there. I remember when I ported it to typescript.

PR #4721, opened 2018-02-28
[next] Mesh and Texture fixes

ivanpopelyshev:
This PR contains multiple serious changes.

1. TextureMatrix is stored in `uvMatrix` field
2. Texture's `_updateUvs` is not a public method, because devs change the frame all the time, for example for  HPBar's.
3. Mesh is up-to-date with v4
4. Mesh `opts` parameter in constructor marked as temporary
5. Removed duplicate RenderTexture file
6. Changed several instances of old `hasLoaded` flag to `valid`

Hamster NineSlicePlane - https://jsfiddle.net/Hackerham/cpL5n1u9/9/
Hamster NineSlicePlane Canvas - https://jsfiddle.net/Hackerham/cpL5n1u9/10/
Animated Plane (no trim) -  https://jsfiddle.net/Hackerham/s56wv7aa/6/
Animated Plane Canvas -  https://jsfiddle.net/Hackerham/s56wv7aa/11/
Animated Plane uploadUvMatrix - https://jsfiddle.net/Hackerham/s56wv7aa/10/


packages/mesh/src/Mesh.js
```diff
@@ -85,6 +87,31 @@ export default class Mesh extends RawMesh
...
 
         this.blendMode = BLEND_MODES.NORMAL;
+
+        /**
+         * TextureMatrix instance for this Mesh, used to track Texture changes
+         *
+         * @member {PIXI.TextureMatrix}
+         * @private
+         */
+        this.uvMatrix = new TextureMatrix(this._texture);
```
bigtimebuddy: Make sure to destroy uvMatrix (and remove texture reference) on destroying Mesh

bigtimebuddy: Also can `uvMatrix` be private?

ivanpopelyshev: It has clamping changes inside. We might make it readOnly, but its not private.

ivanpopelyshev: There are no destoy method in mesh since v4. Lets fix it in different PR

packages/mesh/src/Mesh.js
```diff
@@ -27,6 +27,8 @@ export default class Mesh extends RawMesh
     {
         const geometry = new Geometry();
 
+        texture = texture || Texture.EMPTY;
```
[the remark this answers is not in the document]
ivanpopelyshev: I'll change it, I dont like @finscn suggestions about nulls there

packages/mesh/src/Mesh.js
ivanpopelyshev: uploadUvMatrix is not readonly.

ivanpopelyshev:
As for public vs private, the problem is that when people use Mesh in projects, they need those parameters. It would be good to have custom scopes like in Flash. "very public", "public"," for tweaks" ,"for plugins", "super-private"

ivanpopelyshev:
In pixi v3 it was difficult to use Texture regions, and in v4 we've really improved the situation. Now Texture Region API is stable enough to be used by users who know what Atlas is.

packages/core/src/textures/Texture.js
```diff
@@ -102,13 +102,21 @@ export default class Texture extends EventEmitter
...
          */
         this._uvs = null;
 
+        /**
+         * Default TextureMatrix instance for this texture
+         * By default that object is not created because its heavy
+         *
+         * @member {PIXI.TextureMatrix}
+         */
+        this.uvMatrix = null;
```
bigtimebuddy: What about renaming this to `textureTransform` instead of `uvMatrix`?

packages/mesh/src/Mesh.js
```diff
@@ -85,6 +85,31 @@ export default class Mesh extends RawMesh
...
+        this.uvMatrix = new TextureMatrix(this._texture);
+
+        /**
+         * whether or not upload uvMatrix to shader
+         * if its false, then uvs should be pre-multiplied
+         * if you change it for generated mesh, please call 'refresh(true)'
+         * @member {boolean}
+         * @default false
+         */
+        this.uploadUvMatrix = false;
```
bigtimebuddy: `uvMatrix` => `textureTransform`, `uploadUvMatrix` => `uploadTransform`?

packages/core/src/textures/Texture.js
ivanpopelyshev: Because its not Transform, it cant be combined with other Transforms, its just single Matrix. I want it to be as verbose as possible.

ivanpopelyshev: Flash has Transform and ColorTransform, and both have parent and composition rules. This thing cant be composed with anything, its just a Matrix

packages/mesh/src/Plane.js
```diff
@@ -21,61 +21,51 @@ export default class Plane extends Mesh
...
         this.refresh();
     }
 
     /**
-     * Refreshes
+     * Refreshes plane coordinates
      *
      */
-    refresh()
+    _refresh()
```
bigtimebuddy: Please mark `_refresh` as private

packages/mesh/src/Rope.js
```diff
@@ -35,11 +34,11 @@ export default class Rope extends Mesh
         this.points = points;
         this.refresh();
     }
-
     /**
      * Refreshes
+     *
      */
-    refresh()
+    _refresh()
```
bigtimebuddy: Please mark this as @private

ivanpopelyshev: its supposed to be overriden in generated meshes. There's no "protected" or "default" java-style visibility here, so im sticking with public.

PR #4722, opened 2018-03-01
[next] Filter fullScreen mode 

ivanpopelyshev:
This is difficult commit, because I found the problem with resolution in `FilterSystem`: if I use resolution 2.75 for 128x128, it will create 352x352 texture in real pixels, its not pow2.

I've tried to fix it and, like in v4, store textures with same dimensions in real pixels in each bucket, but there's no way to change resolution of existing renderTexture.

https://jsfiddle.net/Hackerham/p2p7ssnh/12/ - it should give 256x256x2. 

Lets try to hack it and put 512x512x1 in cache before it.

https://jsfiddle.net/Hackerham/p2p7ssnh/18/ - It gives 512x512x1 but i dont know what are consequences of that. It looks fine, but I think parts like `| rt.baseTexture.resolution` can affect it. What if we store filterResolution in `renderTexture` too?

https://jsfiddle.net/Hackerham/p2p7ssnh/19/ - fullScreen version works fine.

https://jsfiddle.net/Hackerham/p2p7ssnh/22/ - fullScreen + resize in setTimeout


ivanpopelyshev:
May be this implementation is fine, just use `resolution=1` for all pooled render textures.

ivanpopelyshev:
Its not the same thing. If someone creates a filter that can only work in fullScreen mode, he can specify it, and people will use it. 

Filter has no connection to application or renderer. Lets wait what @GoodBoyDigital says about this API.

packages/math/src/shapes/Rectangle.js
```diff
@@ -235,14 +235,22 @@ export default class Rectangle
...
+    /**
+     * Enlarges rectangle that way its corners lie on grid
+     *
+     * @param {number} [resolution=1] resolution
+     * @param {number} [eps=0.001] precision
+     */
+    ceil(resolution = 1, eps = 0.001)
     {
-        value = value || 1;
+        const x2 = Math.ceil((this.x + this.width - eps) * resolution) / resolution;
```
bigtimebuddy: Based on our findings from #4745, I think you should take out this code for ceil. Unnecessary

ivanpopelyshev: in this particular place its actually needed. This rectangle can be anything and we dont know order of operations for getBounds() , it could be something really devastating.

PR #4724, opened 2018-03-01
Polyfill interactionData buttons property via event.which for Safari browser

themoonrat:
event.buttons is not supported on Safari: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons
However, it does have a non-standard event.which property, which we can use as a backup to convey the same information.
Also, whilst around here, changed private functions to public in InteractionData; they were clearly not private as they are used externally in InteractionManager.js

### DIFF
--- a/src/interaction/InteractionData.js
+++ b/src/interaction/InteractionData.js
@@ -164,9 +164,8 @@ export default class InteractionData
      * Copies properties from normalized event data.
      *
      * @param {Touch|MouseEvent|PointerEvent} event The normalized event data
-     * @private
      */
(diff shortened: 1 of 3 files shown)


alexedwardjones:
Looks like Number.isInteger isn't supported in IE11, so this is causing issues.

bigtimebuddy:
@themoonrat could you take a look at @alexedwardjones’ note?

themoonrat:
Done. Sorry about that, was testing in my content which must have that polyfilled already via babel

PR #4725, opened 2018-03-01
[next] mask alpha fix

ivanpopelyshev:
First, MaskSystem was broken.
Second, I found the problem in SpriteMaskFilter shader that existed in v4.

This line in SpriteMaskFilter shader is wrong for masks that have alpha component, because R is premultiplied.

```glsl
original *= (masky.a * masky.r * alpha * clip);
```

Lets fix it and allow to use NPM textures too.

Current: http://pixijs.io/examples/?v=next#/demos/alpha-mask.js
Fix: http://pixijs.io/examples/?v=next-mask-alpha#/demos/alpha-mask.js
NPM fix: https://jsfiddle.net/Hackerham/6os1us0o/6/

### DIFF
--- a/packages/core/src/mask/MaskSystem.js
+++ b/packages/core/src/mask/MaskSystem.js
@@ -113,7 +113,7 @@ export default class MaskSystem extends System
         // TODO - may cause issues!
         target.filterArea = maskData.getBounds(true);
 
-        this.renderer.filterSystem.pushFilter(target, alphaMaskFilter);
+        this.renderer.filter.push(target, alphaMaskFilter);
 
         this.alphaMaskIndex++;
     }
@@ -124,7 +124,7 @@ export default class MaskSystem extends System
      */
     popSpriteMask()
     {
-        this.renderer.filterSystem.popFilter();
+        this.renderer.filter.pop();
         this.alphaMaskIndex--;
     }
 
--- a/packages/core/src/filters/spriteMask/SpriteMaskFilter.js
+++ b/packages/core/src/filters/spriteMask/SpriteMaskFilter.js
@@ -52,6 +52,7 @@ export default class SpriteMaskFilter extends Filter
         }
         tex.transform.update();
 
+        this.uniforms.npmAlpha = tex.baseTexture.premultiplyAlpha ? 0.0 : 1.0;
         this.uniforms.mask = tex;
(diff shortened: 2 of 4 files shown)


ivanpopelyshev:
Im gonna make better examples.

packages/core/src/mask/MaskSystem.js
```diff
@@ -113,7 +113,7 @@ export default class MaskSystem extends System
         // TODO - may cause issues!
         target.filterArea = maskData.getBounds(true);
 
-        this.renderer.filterSystem.pushFilter(target, alphaMaskFilter);
+        this.renderer.filter.push(target, alphaMaskFilter);
```
bigtimebuddy: TypeScript would've caught this! Just saying...

ivanpopelyshev: IntellIJ idea caught this when i opened the file :)

ivanpopelyshev: Also, you know I'm all pro-typescript, but our users aren't . I'll make separate TS repository with pixi plugins and typings for v5.

bigtimebuddy: Can we just add the typing into this repo directly? It’ll be easier to manage API changes. Kind of what I did with Pixi-filters. Not much extra effort.

ivanpopelyshev:
Ok, i've made it and its wrong: https://jsfiddle.net/Hackerham/6os1us0o/17/

packages/core/src/mask/MaskSystem.js
ivanpopelyshev: (remembers nobel laureat) Yes, we can!

ivanpopelyshev:
Ok, what the hell? Both v4 and this version have wrong tinting for NPM version.

ivanpopelyshev:
Edge shows different results: this branch is correct, V4 is correct only in left-top.

bigtimebuddy:
Ready to merge @ivanpopelyshev?

ivanpopelyshev:
How to force WebGL1 after lastest changes?

ivanpopelyshev:
Ok, lets merge this one, but investigate this fiddle later.

ivanpopelyshev:
Mozilla is also correct.

PR #4726, opened 2018-03-01
Mesh constructor texture fix

ivanpopelyshev:
This is fix for problem that was discovered in v5: #4721 . it can be hardly reproduced at https://jsfiddle.net/Hackerham/cpL5n1u9/13/ . 

I'm trying to get the correct order of operations and throttling to show the bug

### DIFF
--- a/src/mesh/Plane.js
+++ b/src/mesh/Plane.js
@@ -108,13 +108,14 @@ export default class Plane extends Mesh
      */
     _onTextureUpdate()
     {
-        Mesh.prototype._onTextureUpdate.call(this);
-
-        // wait for the Plane ctor to finish before calling refresh
+        // constructor texture update - stop 1
         if (this._ready)
         {
-            this.refresh();
+            return;
         }
+
+        Mesh.prototype._onTextureUpdate.call(this);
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
I don’t see what’s wrong with the fiddle. What happens when the big shows up?

src/mesh/Plane.js
```diff
@@ -108,13 +108,14 @@ export default class Plane extends Mesh
...
-        // wait for the Plane ctor to finish before calling refresh
-        if (this._ready)
+        // constructor texture update - stop 1
+        if (!this._ready)
         {
-            this.refresh();
+            return;
         }
+
+        Mesh.prototype._onTextureUpdate.call(this);
```
bigtimebuddy: Why not `super._onTextureUpdate()`?

ivanpopelyshev:
That's better fix, if its ok im gonna put it in v5 too.

ivanpopelyshev:
The fix isn't necessary, it looks like `TextureMatrix` tracks initial load as good as `onTextureUpdate`

PR #4729, opened 2018-03-02
[next] image bitmap bug

ivanpopelyshev:
Chrome ignores webgl texture upload params in favor of createImageBitmap params.

Pixi: https://jsfiddle.net/Hackerham/6os1us0o/113/
WebGL: https://fiddle.jshell.net/sosw339h/36/

Coordinates were added as params because Firefox doesnt like the fact I pass only `options`

Reported: https://bugzilla.mozilla.org/show_bug.cgi?id=1442632

### DIFF
--- a/packages/core/src/textures/resources/ImageResource.js
+++ b/packages/core/src/textures/resources/ImageResource.js
@@ -66,7 +66,7 @@ export default class ImageResource extends BaseImageResource
          * @member {boolean|null}
          * @readonly
          */
-        this.premultiplyAlpha = options.premultiplyAlpha !== 'undefined' ? options.premultiplyAlpha : true;
+        this.premultiplyAlpha = options.premultiplyAlpha !== false;
 
         /**
          * The ImageBitmap element created for HTMLImageElement
--- a/packages/core/src/textures/resources/ImageResource.js
+++ b/packages/core/src/textures/resources/ImageResource.js
@@ -15,6 +15,7 @@ export default class ImageResource extends BaseImageResource
      * @param {boolean} [options.createBitmap=true] whether its required to create
      *        a bitmap before upload defaults true
      * @param {boolean} [options.crossorigin=true] - Load image using cross origin
+     * @param {boolean} [options.premultiplyAlpha=true] - Premultiply image alpha in bitmap
      */
     constructor(source, options)
     {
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
According to Khronos specification, Chrome behaviour is correct.

> The WebGL API supports the following additional parameters to pixelStorei.

> UNPACK_FLIP_Y_WEBGL
> UNPACK_PREMULTIPLY_ALPHA_WEBGL
> UNPACK_COLORSPACE_CONVERSION_WEBGL

> If the TexImageSource is an ImageBitmap, then these three parameters will be ignored. Instead the equivalent ImageBitmapOptions should be used to create an ImageBitmap with the desired format.

packages/core/src/textures/resources/ImageResource.js
```diff
@@ -142,9 +142,10 @@ export default class ImageResource extends BaseImageResource
      * Called when we need to convert image into BitmapImage.
      * Can be called multiple times, real promise is cached inside.
      *
+     * @param {PIXI.BaseTexture} baseTexture - BaseTexture for this resource
      * @returns {Promise} cached promise to fill that bitmap
      */
-    process()
+    process(baseTexture)
```
bigtimebuddy: Why did you add this argument? Doesn't look like it's used.

ivanpopelyshev: baseTexture.premultiplyAlpha

bigtimebuddy: Missed that okay

packages/core/src/textures/resources/ImageResource.js
```diff
@@ -59,6 +60,14 @@ export default class ImageResource extends BaseImageResource
          */
         this.createBitmap = options.createBitmap !== false && settings.CREATE_IMAGE_BITMAP && !!window.createImageBitmap;
 
+        /**
+         * Controls texture premultiplyAlpha field
+         * Copies from options
+         * @member {boolean|null}
+         * @readonly
+         */
+        this.premultiplyAlpha = options.premultiplyAlpha !== 'undefined' ? options.premultiplyAlpha : true;
```
bigtimebuddy: What about?
```js
this. premultiplyAlpha = options.premultiplyAlpha !== false;
```

bigtimebuddy: Maybe this doesn't work if you pass `null`, defaults to true?

ivanpopelyshev:
Here it is: https://jsfiddle.net/6os1us0o/121/ . I override premultiplyAlpha every tick, so if you put "true" in the ImageResource, it'll show dark sprite.

PR #4736, opened 2018-03-05
Double resolution multiplication bug

ivanpopelyshev:
@themoonrat found a problem in https://github.com/pixijs/pixi.js/pull/4682


themoonrat:
I haven't got a independent fiddle for this, right now, but within my libs and game I can confirm that this change fixes filter display issues introduced in 4.7.0

Issue #4443, opened 2017-11-14
Repeating deprecation warnings


bash-tp:
If you use a deprecated method, you'll get a console warning every time you call it. This can cause performance problems and can also make it harder to spot issues because the console log gets flooded with repeating deprecation warnings.

Would it be possible to have each deprecation warning be emitted only once? Looking at deprecation.js, this should be fairly easy to do, by having warn() ignore duplicate calls with the same msg+stack. If you agree, I can make this change and send a PR.

ivanpopelyshev:
Agree, I'm using that approach in pixi-spine. I agree that we have to move it to v5 too.

bigtimebuddy:
Not a bad idea! Would welcome a PR for that. Ideally it would be added to the warn function instead of to each deprecation to make the change minimal.

ivanpopelyshev:
add a key to warn function, so it stores whether particular key was used

wildaces215:
So is it a lot deleting in the depreactionjs file?

bigtimebuddy:
@wildaces215 Deprecation warnings were not designed to be hard breaks, that's why we aren't throwing Errors. It gives people time to update their code between major releases. However, we should fix this and remove repeated calls to make it more performant.

YahorPadlucki:
Hi, @bigtimebuddy , I've added a PR for this issue

PR #4738, opened 2018-03-06
Repeating deprecation warnings #4443

YahorPadlucki:
Added message-stack map to prevent showing multiple warning messages from same place. 
Didn't change anything for case with "IE < 10 and Safari < 6" - should I add separate array to store shown messages for this case?

Fix for issue #4443

### DIFF
--- a/src/deprecation.js
+++ b/src/deprecation.js
@@ -1,6 +1,10 @@
 // provide method to give a stack track for warnings
 // useful for tracking-down where deprecated methods/properties/classes
 // are being used within the code
+
+// message/stack map - helps to prevent showing deprecation warning if it was already shown
+const shownWarningStacks = {};
+
 function warn(msg)
 {
     // @if DEBUG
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@YahorPadlucki thanks for this! I had just added this feature to v5, but went for a slightly simpler approach. You can check it out here: https://github.com/pixijs/pixi.js/blob/next/bundles/pixi.js/src/deprecated.js#L2-L48

I would argue slightly against this approach only because if there are the same warnings called each frame from different places, it would still spam both messages and kill performance. I tend to lean towards being a little aggressive about silencing repeat warnings to not hurt performance overall.

What do you think?

YahorPadlucki:
@bigtimebuddy, I thought that it is useful to show exact place were the warning is called from - to know where to fix it.
With your approach - if you want to fix all deprecations, it can be a bit confusing - when you fix it in one place, refresh, and get it in a new  place - which wasn't logged before.  So you need to know that you better to search same places in code. 
Thats the only concern I have.

bigtimebuddy:
Yeah, I think the trade-off for fewer warnings for better performance is better. So I'm okay being conservative with warning and the dev having to find-and-replace, refresh than having to spit out every possible warning. I'm considering the worst case scenario, which I saw when we deprecated some filters APIs that get called every frame.

bigtimebuddy:
@YahorPadlucki, would you mind changing this to block repeat deprecations based on the message and not the stack trace? Then, I'd like to merge for 4.8.0

YahorPadlucki:
Changed. Sorry - didn't get that I need to update my branch - thought you will merge your solution.

bigtimebuddy:
Thank you @YahorPadlucki!

PR #4739, opened 2018-03-06
Bugfix/masking whilst rendering to a render texture

GoodBoyDigital:
Fixes issue where masking is not currently working on render textures. Enjoy!


ivanpopelyshev:
I'm afraid thats also a mask problem. You sure?

ivanpopelyshev:
Also, make sure that filterArea is set to null after the fact. Make sure that if user specifies his own filterArea, its used instead of the substitute.

GoodBoyDigital:
Hey @ivan this is fixing masking whilst rendering to a render texture, not masking with a render texture. Different problems init :)

bigtimebuddy:
@GoodBoyDigital could you fix up the linting on this?

PR #4749, opened 2018-03-09
Adds package-lock.json, upgrades Travis to Node.js v8

bigtimebuddy:
It's time that we upgrade to Node.js v8 for Travis. This include committing **package-lock.json** to our repo. Information about **package-lock.json**: https://docs.npmjs.com/files/package-lock.json

Mainly, we need to lock down dependencies so we don't have dependencies breaking our builds all the time.  Related issue that just happened when rollup-plugin-commonjs broke our build: #4737

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-    - "6"
+    - "8"
 
 env:
   - CXX=g++-4.8
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5928 @@
+{
+	"requires": true,
+	"lockfileVersion": 1,
+	"dependencies": {
+		"@pixi/jsdoc-template": {
+			"version": "2.4.0",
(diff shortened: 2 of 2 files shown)


cursedcoder:
Honestly 👎 for that.

Lock files are for private projects IMO.

Just look at out package.json, most of versions aren't strictly locked eg `^`.

Would strongly recommend lock for strict versions like `1.0.2`,`~1.0.2`, `>1.0.2, <1.0.5` instead of `^1.0.2` and this way it should not break.

https://stackoverflow.com/a/22345808

bigtimebuddy:
Many large public projects commit their yarn.lock or package-lock.json file. It's actually not really about locking the immediate dependencies in the package.json file, but making sure the whole dependency tree is locked down. It's really the child dependencies that we should worry about most. Upgrading _any_ dependency would be opt-in.

ivanpopelyshev:
We have regular problems with travis. I think that's good enough solution, we can think later about other ways to solve it.

PR #4760, opened 2018-03-16
[next] Adds TypeScript typings to bundles

ivanpopelyshev:
Typings can be checked out from https://github.com/pixijs/pixi-typescript and tested on a few tests and pixi-filters typings, just type `tsc` and see if there are any errors. For example, if you remove canvas-related part of typings, CanvasRenderer wont be found by tests, but everything else will be fine..

Due to extreme popularity of typescript among our users, I propose we add typings in v5 alpha. 

At the same time, I dont want it to affect the speed of development, so lets not fight over small things, I will accept PR's and edit typings before every new release until we have better solution.

I believe that with help of @cursedcoder @bigtimebuddy and @xerver we can finish https://github.com/englercj/tsd-jsdoc/ , but it will take time.

This PR is required to start development of v5 plugins.


bundles/pixi.js-legacy/package.json
```diff
@@ -12,6 +12,7 @@
   "main": "lib/pixi-legacy.js",
   "module": "lib/pixi-legacy.es.js",
   "bundle": "dist/pixi-legacy.js",
+  "types": "pixi-legacy.js.d.ts",
```
bigtimebuddy: Need to add to "files" field

bundles/pixi.js/package.json
```diff
@@ -12,6 +12,7 @@
   "main": "lib/pixi.js",
   "module": "lib/pixi.es.js",
   "bundle": "dist/pixi.js",
+  "types": "pixi.js.d.ts",
```
bigtimebuddy: Also need to add to files field for publishing

clark-stevenson:
Hey guys!

I am sorry for the long delay, I have been swamped this past 2 weeks. 

Amazing that pixi is being shipped with types! One thing though is that I would possibly recommend using the `@types/` version of pixi rather than `pixi-typescript` version.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pixi.js/index.d.ts

This is the same state as our `pixi-typescript` but it passes all of Microsoft Linting and is generally a cleaner and better formatted file.  👍

ivanpopelyshev:
Thanks I'll look into it :)

mmermerkaya:
Why don't you use [Clutz](https://github.com/angular/clutz) instead of creating your own JSDoc to TypeScript tool ([tsd-jsdoc](https://github.com/englercj/tsd-jsdoc/))?

PR #4765, opened 2018-03-17
Suggest npm v5 and lock dependencies

allanbreyes:
Feel free to close if unneeded... couldn't find a corresponding issue that this addresses. Just noticed that there wasn't a dependency lock file. I figured it's a better thing to have for build determinism.

Docs on [`engines`](https://docs.npmjs.com/files/package.json#engines): basically, it's advisory on `npm`, and IIRC _strict_ on `yarn`.

Docs on [`package-lock.json`](https://docs.npmjs.com/files/package-lock.json): basically, npm recommends committing it to source. Caveat: the diff I introduced is _huge_. _I swear, I'm not trying to inflate my LOC!_ Anyway, I built this in a Linux VM, so it _should_ match CI. (e.g. sometimes the lockfiles look different when built on macOS, which can introduce [`fsevents`](https://www.npmjs.com/package/fsevents)).

Recommendations for confirming:
- Pull the branch down
- Use `npm` v5 (default for `node` v8+)
- Delete `package-lock.json`
- Run `npm install`
- `git diff` (ideally, should be none)
- If it's not, feel free to push onto this branch (edit access enabled). Or, you know, just close this PR.

Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7563 @@
+{
+  "name": "pixi.js",
+  "version": "4.7.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@gulp-sourcemaps/map-sources": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@gulp-sourcemaps/map-sources/-/map-sources-1.0.0.tgz",
+      "integrity": "sha1-iQrnxdjId/bThIYCFazp1+yUW9o=",
+      "dev": true,
+      "requires": {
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
@allanbreyes thank you for this. We've already made this similar change for **next** branch (v5) so great minds think alike! I, however, would like to pass on this change for PixiJS v4 since it does not fix any bug and we are trying to limit PRs to this branch right now. If you have any housekeeping improvements/recommendations please help us focus them on v5!

allanbreyes:
Sounds good to me. Thanks!

PR #4779, opened 2018-03-20
Adds Number.isInteger polyfill

themoonrat:
Fixes IE11, which doesn't have this functionality (as reported in https://github.com/pixijs/pixi.js/pull/4724)

Before: https://jsfiddle.net/themoonrat/zmmndr00/
After: https://jsfiddle.net/24cjpr2w/2/ (will work when build has finished for this branch)


bigtimebuddy:
Thanks for this

Issue #3973, opened 2017-04-26
FilterState retains destroyed sprites


lucap86:
Steps

- Take heap snapshot
- add some bunny
- take heap snapshot
- destroy bunnies container
- collect garbage
- take an heap snapshot
- filter by "Objects allocated between Snapshow 1 and Snapshot 2"

Some destroyed sprites (and related resources like _boundsRect) are retained by FilterState
![test](https://cloud.githubusercontent.com/assets/15343100/25430727/0da201b8-2a7e-11e7-8be9-370dfc4d3692.png)

[test.zip](https://github.com/pixijs/pixi.js/files/958086/test.zip)

englercj:
Good catch, thanks for the steps and test case. Hopefully we can take care of this soon.

YahorPadlucki:
Hi @englercj, @lucap86  - can you check my PR for this issue?

PR #4783, opened 2018-03-21
Destroy FilterState when object destroyed #3973

YahorPadlucki:
Examples
**pixijs 4.7.1:**
https://yahorpadlucki.github.io/pixiExample/default.html
image

**this PR:**
https://yahorpadlucki.github.io/pixiExample/custom.html
image

Seems to fix this issue - but not sure if I am doing it right?


YahorPadlucki:
Hi, @bigtimebuddy @themoonrat @ivanpopelyshev 
can somebody look at it? Some how it is the only PR without comments ( Did I prepare it wrong?

englercj:
I don't think this is the right way to fix this issue. Having every container hold a ref to the last renderer to operate on it seems like a code smell.

Instead I would have objects notify that they are being destroyed and have the filter manager listening for those events from all the targets it works on. It can then cleanup any state needed.

YahorPadlucki:
Thanks a lot for explanation, Chad ! 
Will change it.

YahorPadlucki:
Updated with 'destroyed' event listener

englercj:
I think this solution a lot more. Thanks for iterating on that feedback!

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -53,6 +65,7 @@ export default class FilterManager extends WebGLManager
         this.filterData = null;
 
         this.managedFilters = [];
+        this.managedDisplayObjects = [];
```
bigtimebuddy: Would we maybe use something a little less verbose like `managedTargets`?

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -537,6 +556,41 @@ export default class FilterManager extends WebGLManager
...
+        });
+        this.managedDisplayObjects = [];
+    }
+
+    /**
+     * Destroys FilterState of specified target
+     *
+     * @param {PIXI.DisplayObject} target, which FilterState should be destroyed
+     */
+    destroyFilterStateByTarget(target)
```
bigtimebuddy: Why create a separate function? Wouldn't it just be easier to inline in the destroy handler?

src/core/renderers/webgl/managers/FilterManager.js
```diff
@@ -666,4 +720,20 @@ export default class FilterManager extends WebGLManager
             this.pool[screenKey] = [];
         }
     }
+
+    /**
+     * Called when managed DisplayObject is destroyed
+     *
+     * @param {PIXI.DisplayObject} destroyedDisplayObject, object which was destroyed
+     */
+    onManagedDisplayObjectDestroyed(destroyedDisplayObject)
```
bigtimebuddy: Less verbose name suggestion: `onTargetDestroyed`

YahorPadlucki:
@bigtimebuddy , Thanks a lot for review. 
Made changes according to review, added unit-test for destroyed event

bigtimebuddy:
Closing in favor of #4897

PR #4787, opened 2018-03-22
Fixed: wordWrap doesn't honor anything except standard space.

ceco-fmedia:


bigtimebuddy:
Can you please provide more info about this change? Specifically would be very helpful to get a jsfiddle so we can verify the fix and a unit test to make sure this change doesn’t break in the future.

ceco-fmedia:
@bigtimebuddy 
When calculating where the line should brake the function didn't took into consideration new lines and other Unicode white space characters.
https://en.wikipedia.org/wiki/Whitespace_character

brenwell:
It looks like it would certainly overlap. However if the only issue is to manage the following I can certainly incorporate that into my current work

 '\u00A0'  '\u202F' '\u2060' '\uFEFF'

Then that should not be a problem

ceco-fmedia:
'\u00A0' '\u202F' '\u2060' '\uFEFF' are non-breaking spaces and you handle them correctly. The other unicode whitespace characters are the problem.

brenwell:
As a side note. Is it ok to use the trimRight function 

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/TrimRight

brenwell:
Oh sorry I see. I thought the nbsp would have been handled. This makes sense.  

Anything else?

bigtimebuddy:
Duplicate, closing this in favor of #4794

PR #4792, opened 2018-03-23
Added sync cache

GoodBoyDigital:
Added a caching layer to generateSyncUniforms. 

Basically I get a signature of the uniform object and store the generated function on that. 

The cached function is returned if an objects shares a signature with another.

Nice little under the hood boost!


packages/core/src/shader/ShaderSystem.js
```diff
@@ -99,11 +100,39 @@ export default class ShaderSystem extends System
...
+        {
+            this.cache[id] = generateUniformsSync(group, this.shader.program.uniformData);
+        }
+
+        group.syncUniforms[this.shader.program.id] = this.cache[id];
 
         return group.syncUniforms[this.shader.program.id];
     }
 
+    getSigniture(group, uniformData)
```
ivanpopelyshev: typo.

GoodBoyDigital:
Cool! all done :)

bigtimebuddy:
Thanks @GoodBoyDigital

PR #4793, opened 2018-03-23
Bug/sprite sheet image

GoodBoyDigital:
imageResourceName is now the actual path of the loaded image
This means it site in the pixi cache correctly now and will not get loaded /prepared twice.


bigtimebuddy:
@GoodBoyDigital I'm not sure this is a good idea after working on it for awhile. Basically, if there's an error with the image loading, it's easy to check `*_image.error` in the resources. In your implementation here, you only get the resource by full path. We could add both or leave it how it is. Either way, you should add a test which demonstrates the use-case you described.

Issue #4791, opened 2018-03-23
TextMetrics.MeasureText () handles '\n' incorrectly (4.7.1)


robertswan:
4.7.1 has introduced a new bug: strings using wordwrap that also include newline character are calculated incorrectly.

Looking through the code it appears the string is broken into words and measured per word (correct) but when checking for wordwrap it does not respond to a newline in the word (incorrect) by starting a new line measurement at 0 width. This leaves the the line and subsequent lines mis-measured and wrapping incorrectly. Rendering confirms this.

Repro on: 
* machine: macbook pro
* browser: chrome 65.0.3325.162
* I'd expect the bug on all platforms

Expected Result (4.7.0):
* http://pixijs.io/examples/?v=v4.7.0#/basics/text.js
* change ```Rich text with a lot of options and across multiple lines``` -> ```Rich text with a lot of options\nand across multiple lines```
* rerun
* see:
```
Rich text with a lot of
options
and across multiple lines
```

Actual Result (4.7.1):
* http://pixijs.io/examples/?v=v4.7.1#/basics/text.js
* change ```Rich text with a lot of options and across multiple lines``` -> ```Rich text with a lot of options\nand across multiple lines```
* rerun
* see:
```
Rich text with a lot of
options
and across
multiple lines
```

bigtimebuddy:
Probably related to #4545 
@brenwell could you please take a look at this?

brenwell:
Yes I see, I will take a look

brenwell:
@bigtimebuddy 
Ok I am looking into the issue, and it has brought about a question. How to handle spaces, should they be preserved on new lines or not, I think previously it was removing the first space and keeping the rest on a new line unless its the first line.

### Multiple spaces

If the text is `this-is-my-text---I-love-it` (*dashes represent spaces*)

Should it be like this
```
|this-is-my-text-|
|--I-love-it     |
```
Or like this
```
|this-is-my-text  |
|I-love-it        |
```

### Single spaces

If the text is `this-is-my-text-I-love-it`  (*dashes represent spaces*)

Should it be like this

```
|this-is-my-text|
|-I-love-it     |
```
Or like this
```
|this-is-my-text|
|I-love-it      |
```

bigtimebuddy:
These are good questions. In general if we are auto-wrapping I think it makes sense to trim. However, we should probably maintain spaces after explicit newline characters:

this-is-my-text--\n-I-love-it

this-is-my-text
-I-love-it

[...]

PR #4794, opened 2018-03-24
Fixes for TextMetrics WordWrap newline issue

brenwell:
fixes for #4545  #4791 

I hope this covers everything

Sorry about the ugly tests


brenwell:
In response to #4787  I have attempted to add wrapping for all unicode wrapping whitespaces. I took the ones from here that seemed important https://en.wikipedia.org/wiki/Whitespace_character#Unicode.

Perhaps it was overkill

brenwell:
Probably some ways to improve this, had to do it very quickly, sorry

ivanpopelyshev:
Wow, ideographic space is there too! Nice

src/core/text/TextMetrics.js
```diff
@@ -255,6 +284,109 @@ export default class TextMetrics
...
+        return (breakingSpaces.indexOf(char.charCodeAt(0)) >= 0);
+    }
+
+    /**
+     * Splits a string into words, spaces and newLine characters
+     *
+     * @param  {string}  text       The text
+     * @return {array}  A tokenized array
+     */
+    static explodeIntoWordsSpacesAndNewlineChars(text)
```
bigtimebuddy: This name is pretty cumbersome, what about something simpler like "explode"

src/core/text/TextMetrics.js
```diff
@@ -255,6 +284,109 @@ export default class TextMetrics
         return width;
     }
 
+    /**
+     * trims breaking whitespaces from string
+     *
+     * @param  {string}  text  The text
+     * @return {string}  trimmed string
+     */
+    static trimRight(text)
```
bigtimebuddy: Unit tests for this method would be ace

src/core/text/TextMetrics.js
```diff
@@ -255,6 +284,109 @@ export default class TextMetrics
...
+        return text;
+    }
+
+    /**
+     * Determines if char is a newline.
+     *
+     * @param  {string}  char  The character
+     * @return {boolean}  True if newline, False otherwise.
+     */
+    static isNewline(char)
```
bigtimebuddy: Also, some simple unit test coverage for this, however, it can be pretty basic!

src/core/text/TextMetrics.js
```diff
@@ -255,6 +284,109 @@ export default class TextMetrics
...
+        return (newlines.indexOf(char.charCodeAt(0)) >= 0);
+    }
+
+    /**
+     * Determines if char is a breaking whitespace.
+     *
+     * @param  {string}  char  The character
+     * @return {boolean}  True if whitespace, False otherwise.
+     */
+    static isBreakingSpace(char)
```
bigtimebuddy: And a very basic unit test for this. Make it return true/false, check for return type

cursedcoder:
This looks solid!
Would be glad to review again after comment-fixing.

src/core/text/TextMetrics.js
brenwell: ok shortened

src/core/text/TextMetrics.js
brenwell: will do

src/core/text/TextMetrics.js
brenwell: will do

src/core/text/TextMetrics.js
brenwell: will do

brenwell:
Have added `@private` and fixed the name, still need to write the tests

brenwell:
I was thinking that because html actually reduces multiple white spaces in to 1 space unless the css whiteSpace attribute is changed, perhaps we should have the same


New lines | Spaces and tabs | Text wrapping
-- | -- | --
Collapse | Collapse | Wrap
Collapse | Collapse | No wrap
Preserve | Preserve | No wrap
Preserve | Preserve | Wrap
Preserve | Collapse | Wrap

https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

themoonrat:
Oh man, this is awesome work so far. It's also a rabbit hole of the further you go, the more things need to be done to be more like html/css text. If you think you can match it via whiteSpace, then it'd be great, but if it matches 'normal' behaviour for now and perhaps a patch later to add support for the others, I'd be happy with that too.

@bigtimebuddy - You know you showed me your potential text via svg implementation? I wonder if that class could be used to validate changes made in this PR? Something a bit like http://pixijs.io/pixi-text-style/# - but with 2 canvases, one using the current PIXI Text, and one using your SVG Text. I had my doubts about the SVG Text due to it relying on browser support for things like gradients; but it could prove to be vital in allowing us to easily compare side by side visual differences regarding word wrapping, word breaking, spacing etc to make sure we match html/css behaviour where possible

brenwell:
I am keen to add `white-space` and this is what I have learnt

What pixi has at the moment is the equivilant of css `pre-wrap` (Well thats what my code does)

Css `white-space` also effects word wrapping, and assuming we don't want to change `PIXI.TextStyle` api, this means we can keep that property and simplify the `white-space` requirements whilst keeping all the options

So the standard css `white-space` can be simplified from

Value | New lines | Spaces and tabs | Text wrapping
--| -- | -- | --
**normal** | Collapse | Collapse | Wrap
**no-wrap** | Collapse | Collapse | No wrap
**pre** | Preserve | Preserve | No wrap
**pre-wrap** | Preserve | Preserve | Wrap
**pre-line** | Preserve | Collapse | Wrap

to

Value | New lines | Spaces and tabs
--| -- | --
**normal** | Collapse | Collapse
**no-wrap** | Collapse | Collapse
**pre** | Preserve | Preserve
**pre-wrap** | Preserve | Preserve
**pre-line** | Preserve | Collapse

Which equals

Value | New lines | Spaces and tabs
--| -- | --
**normal** | Collapse | Collapse
**pre** | Preserve | Preserve
**pre-line** | Preserve | Collapse

The question is what should default, and I would suggest 'pre' and not normal. I think most people what newlines to be honoured if they have added them.

Does that all make sense

brenwell:
https://jsfiddle.net/brenwell/0w7z2buo/28/ if you want a visual

brenwell:
Ok its in.

There is an overlap between `wordWrap` and `whiteSpace` theoretically one could drop `wordWrap` and complete all the `whiteSpace` options.

I kinda rushed through this, so the tests are ugly but it seems to behaving as per the mozilla docs.

I made a little fiddle to test it out, its quite a tricky thing to build and also observe so I recommend using it to verify its accurate

https://jsfiddle.net/brenwell/9c9yL1j3/3/

brenwell:
Additonally `whiteSpace` only works with `wordWrap =  true`

src/core/text/TextMetrics.js
brenwell: Done

src/core/text/TextMetrics.js
brenwell: Done

src/core/text/TextMetrics.js
brenwell: Done

src/core/text/TextMetrics.js
brenwell: Done

themoonrat:
@brenwell so I've had a good play by importing those changes into my game lib. And it all looks good... apart from when I come to CJK text

Take a look at https://jsfiddle.net/b9aLpoyg/1/ - It's got breakwords on, and it's using some Chinese text. Notice how there is an empty line at the very top when there shouldn't be

src/core/text/TextMetrics.js
```diff
@@ -108,54 +108,95 @@ export default class TextMetrics
...
-            const wordWidth = TextMetrics.getFromCache(word, ls, cache, context);
+            const tokenWidth = TextMetrics.getFromCache(token, letterSpacing, cache, context);
 
             // word is longer than desired bounds
-            if (wordWidth > wordWrapWidth)
+            if (tokenWidth > wordWrapWidth)
             {
+                // start newlines for overflow words
+                lines += TextMetrics.addLine(line);
+                // canPrependSpaces = !collapseSpaces; Seems wrong
```
themoonrat: can this line be removed now?

src/core/text/TextMetrics.js
```diff
@@ -217,23 +262,29 @@ export default class TextMetrics
...
+     * @param  {boolean}  newLine     - Add new line character to end
      * @return {string}   A formatted line
      */
     static addLine(line, newLine = true)
     {
+        line = TextMetrics.trimRight(line);
+
         line = (newLine) ? `${line}\n` : line;
 
+        // line = line.replace(new RegExp(' ', 'g'), '*');
```
themoonrat: Can this line be removed now?

brenwell:
Ok I see the issue. The chinese text is using the `fullwidth comma` U+FF0C instead of normal commas and therefore my wordwrapper thinks its all one word. So there is 1, but possibly 2 bugs.

1. is that the first line shouldn't have a new line if the first word is an overflow word, see https://jsfiddle.net/b9aLpoyg/6/

2. should the chinese sentences have broken at these `fullwidth commas` see
https://jsfiddle.net/b9aLpoyg/9/

src/core/text/TextMetrics.js
brenwell: yes

src/core/text/TextMetrics.js
brenwell: yes

themoonrat:
1. to definitely be fixed, yeah.
2. support for cjk text is a whole other issue that I hackily do in my code, and I know someone else who had made a nicer version and wanted to make a plugin out of it. In my hack, I detect if I'm using a CJK string, and if so force breakwords to be on, even if the style for breakwords is false. This is because you can get long unbroken characters so normal wordWrap just can't handle it.

src/core/text/TextMetrics.js
```diff
@@ -255,6 +307,158 @@ export default class TextMetrics
         return width;
     }
 
+    /**
+     * Determines whether we should collapse breaking spaces
+     *
+     * @param  {string}   whiteSpace  The TextStyle property whiteSpace
+     * @return {boolean}  should collapse
+     */
+    static collapseSpaces(whiteSpace)
```
bigtimebuddy: Can you make this `@private`?

src/core/text/TextMetrics.js
```diff
@@ -255,6 +307,158 @@ export default class TextMetrics
...
+        return (whiteSpace === 'normal' || whiteSpace === 'pre-line');
+    }
+
+    /**
+     * Determines whether we should collapse newLine chars
+     *
+     * @param  {string}   whiteSpace  The white space
+     * @return {boolean}  should collapse
+     */
+    static collapseNewlines(whiteSpace)
```
bigtimebuddy: Also, Can you make this `@private`?

bigtimebuddy:
Really great work @brenwell, thanks for digging into all the details on this. I should also add `whiteSpace` to https://github.com/pixijs/pixi-text-style once we merge it.

brenwell:
@bigtimebuddy Ok made them private. Also I didn't know that tool, I probably should have used it.

So I was thinking about the CJK issue yesterday and I realised that though supporting it is too big to add to pixi, having a bit of flexibility might be nice. 

**Add soft-hyphen supper**
And supporting the `&shy;` soft hyphen would make line wrapping a bit more customisable. 

**Modifying character sets**
Being able to modify the breakingSpace and newLine chars found in the methods `isNewline` and `isBreakingSpace` would offer some flexibility. Perhaps these should be class variables.

so one could do something like
```js
PIXI.TextMetrics.breakingSpaces = [
            0x0009, // character tabulation,
            ...
            //  whatever characters you want
        ];

new PIXI.Text('text with custom breaking spaces')
```

**Sets rather than arrays**

Also perhaps these arrays should be `Sets` I noticed that the `Set.has` method is considerable faster than `Array.indexOf`. But I didn't see Sets being used anywhere in the codebase.

**CJK helpers rather than support**

After reading there appears to be many CJK "standards" so If you really want to support CJK I would recommend having to new methods similar to `isNewline` and `isBreakingSpace`. 

`canBeLastCharOfLine` and `canBeFirstCharOfLine` and then users can add there own character sets for them depending on the standard they want, rather than picking one

---

I think all these can be new tasks, but the only one I think is really important is the soft-hyphen support.

themoonrat:
I've built nightly versions of our games against this change, and had a tester go through them all, and everything looks good :+1: 
@brenwell and I have had a chat on slack, and he will add a couple of functions that will allow hooks into managing breakwords... one to determine whether breakwords is allowed to be used or not, and one to determine if a break can be inserted in between 2 characters or not.

bigtimebuddy:
I’m cool with that

cursedcoder:
Like it +1.

Very solid @brenwell !

brenwell:
Ok the hooks are in for plugins like the one discussed in #4447 

You can see them in action here

https://jsfiddle.net/brenwell/ytu8gx65/25/

Additionally there is a small optimization, making the breakSpaces and newLines arrays static

----

I think (hope) it's done

src/core/text/TextMetrics.js
```diff
@@ -602,3 +630,37 @@ TextMetrics._context = canvas.getContext('2d');
  * @private
  */
 TextMetrics._fonts = {};
+
+/**
+ * Cache of new line chars.
+ * @memberof PIXI.TextMetrics
+ * @type {Object}
```
themoonrat: @type {number[]}

themoonrat: @type {number[]}

src/core/text/TextMetrics.js
```diff
@@ -602,3 +630,37 @@ TextMetrics._context = canvas.getContext('2d');
...
+ */
+TextMetrics._newlines = [
+    0x000A, // line feed
+    0x000D, // carriage return
+];
+
+/**
+ * Cache of breaking spaces.
+ * @memberof PIXI.TextMetrics
+ * @type {Object}
```
themoonrat: @type {number[]}

brenwell:
I knew when I wrote "done" it was a mistake ;)

bigtimebuddy:
Could you fix the broken test?

themoonrat:
Aside from unit test fail, I'm happy with this with the added hooks that allow for CJK word wrapping

bigtimebuddy:
@themoonrat If you haven't already, could you please double check your projects again to make sure we haven't broken something with this large PR?

brenwell:
i am not sure why that test is failing on travis. Passes locally with node versions 4 & 6

brenwell:
Anyone have any ideas what might cause a test to fail on Travis and not locally

themoonrat:
@bigtimebuddy I'll do a rebuild of all the games today and take another look

brenwell:
So I wrote a new test, the previous one with the Chinese chars never passed on travis but always passed locally. Perhaps it was a OSX Ubuntu issue

themoonrat:
Brenwell is just investigating my latest annoying scenario: https://jsfiddle.net/uhk943vx/ - where it breaks at the space between hello and world, when it should't, it should keep them on the same line as the later cjk characters can break at the end of a line

这hello
world这这这这这这这这这这
这这这这这这这这这这这这
should imo look like
这hello world这这这这这这
这这这这这这这这这这这这
这这这这

brenwell:
So guys I have looked into this, and I think perhaps its better to allow users to override the tokenisation method than the layout method. https://jsfiddle.net/uhk943vx/10/

I need to test it all but I think its simpler. I will check with @themoonrat to make sure it still meets his requirements

bigtimebuddy:
@brenwell and @themoonrat where are we on this? Would love to get this in.

themoonrat:
Brenwell, if you could add sending style.breakwords to canBreakChars, I'd be happy at getting this pr in now as is, and then further tweak in a future pr. My edge issues with CJK text shouldn't stop the other good fixes for line breaking this pr contains from being merge in, imo

brenwell:
Ok no problem.

brenwell:
Ok so I rolled back the last change and add breakWords to canBreakChars. 

@themoonrat You are right we should ship this and then focus on the plugin support as a separate feature.

brenwell:
So guys. I am hoping that this is ready to ship.

bigtimebuddy:
As soon as @themoonrat approves...

cursedcoder:
Amazing!

brenwell:
Yeah!

themoonrat:
Thanks for the huge amount of time and effort on this @brenwell :1st_place_medal:

PR #4797, opened 2018-03-25
Pedantic spelling fixes

MercifulCode:
Mostly touching up various squiggles that showed up in my IDE for code comments 😛 

However, there were a few non-comment typos, such as class/function name inconsistencies.

Cheers!


packages/accessibility/src/AccessibilityManager.js
```diff
@@ -117,7 +117,7 @@ export default class AccessibilityManager
          * @private
          */
         this.isActive = false;
-        this.isMobileAccessabillity = false;
+        this.isMobileAccessability = false;
```
allanbreyes: Pedantic suggestion for a pedantic fix: `isMobileAccessibility`

package-lock.json
```diff
@@ -568,7 +568,7 @@
 		"circular-json": {
 			"version": "0.3.3",
 			"resolved": "https://registry.npmjs.org/circular-json/-/circular-json-0.3.3.tgz",
-			"integrity": "sha1-gVyZ6oT2gJUp0vRXkb34JxE1LWY=",
+			"integrity": "sha512-UZK3NBx2Mca+b5LsG7bY183pHWt5Y1xts4P3Pz7ENTwGVnJOUWbRb3ocjvX7hx9tq/yTAdclXm9sZ38gNuem4A==",
```
allanbreyes: @bigtimebuddy: I suspect there might be some continued thrash on the lock file (see: https://github.com/npm/npm/issues/17749), as it appears the hashing algorithm is chosen based on OS, npm version, package source, etc.

It sounds like two options may help:
- The classic `rm -rf node_modules`........
- Standardizing Node.js and npm versions (maybe)

bigtimebuddy: Thanks for bringing this up @allanbreyes, I want to dig into this issue more. But I think reverting for the sake of this PR is a good option for now.

PR #4803, opened 2018-03-27
Edit to RegEx to allow parsing "minified" SVG width + height 

rich-earth:
Relaxed this regex, so that SVG xml produced by using [this]( https://www.npmjs.com/package/mini-svg-data-uri) works with BaseTexture.fromImage(datauri.., )


englercj:
Looks like this fails a unit test that asserts that if width & height aren't both present an empty object should be returned.

PR #4806, opened 2018-03-27
Edit to RegEx to allow parsing SVG width / height, when using "minified" svg

rich-earth:
Relaxed this regex (should pass unit test now),  so that SVG xml produced by using [this]( https://www.npmjs.com/package/mini-svg-data-uri) works with BaseTexture.fromImage(datauri.., )


bigtimebuddy:
Hey @rich-earth, thanks for this PR, however, before you go any further down this path I'd like to let you know we are not considering refactors or new features for v4, especially ones that don't solve existing major or critical bugs. This definitely feels like it would fall under nice-to-have. If you would like to provide more support minified data-uri strings for SVG, I would encourage you to help us roll this into feature into v5. Going to close this.

PR #4829, opened 2018-04-04
Revert "Fixed a difference between canvas and webgl sprite renderer on scaled textures."

bigtimebuddy:
Reverts pixijs/pixi.js#4800


bigtimebuddy:
Nope, need to manually fix this...

PR #4800, opened 2018-03-26
Fixed a difference between canvas and webgl sprite renderer on scaled textures.

ceco-fmedia:
You can see it here:
https://jsfiddle.net/1cj8w0mv/24/

PR #4830, opened 2018-04-04
🔥 Reverts #4800

bigtimebuddy:
Need to revert #4800 and publish a hotfix. It broke existing functionality. @themoonrat can follow-up with details.


themoonrat:
Caused our games that use texture atlases at 0.5 resolution to display images larger and in the wrong aspect ratio

PR #4833, opened 2018-04-06
Ability to set certain display components to be drawn to rounded pixels

themoonrat:
Ability to set certain display components to be drawn to rounded pixels, independent of the global roundPixels renderer setting.

We get constant reports of Text looking blurry in Pixi
http://www.html5gamedevs.com/topic/33309-can-pixi-create-pixel-perfect-text-in-a-scaled-container/
http://www.html5gamedevs.com/topic/30018-blurry-text-after-scaling/
http://www.html5gamedevs.com/topic/29576-scalable-text-for-pixijs/
http://www.html5gamedevs.com/topic/27810-text-objects-are-blurry/
http://www.html5gamedevs.com/topic/26880-blurry-text-in-pixijs/
http://www.html5gamedevs.com/topic/23136-how-to-disable-font-antialiasing/
https://github.com/pixijs/pixi.js/issues/4242
https://github.com/pixijs/pixi.js/issues/3263

etc. etc.

And the solution is always 'draw to whole pixel positions and it looks a _lot_ sharper'. Which is true.

So this PR adds a 'roundPixels' property to Sprite and some components that inherit from Display Objects. roundPixels is a property that was already found in some Display Objects, so it's not a new usage, just extending it.

And then for the render plugins, it checks if either the roundPixels property is set on the renderer OR on the display object. I've set only set roundPixels to true by default for Text and Bitmap text, so they will always look as sharp as possible.

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -132,6 +132,14 @@ export default class Sprite extends Container
          * @default 'sprite'
          */
         this.pluginName = 'sprite';
+
+        /**
+         * Do we want to force this display component to be drawn at whole pixels only.
+         *
+         * @member {boolean}
+         * @default false
+         */
+        this.roundPixels = false;
     }
 
     /**
--- a/src/core/sprites/canvas/CanvasSpriteRenderer.js
+++ b/src/core/sprites/canvas/CanvasSpriteRenderer.js
@@ -96,7 +96,7 @@ export default class CanvasSpriteRenderer
             dy -= height / 2;
 
             // Allow for pixel rounding
-            if (renderer.roundPixels)
+            if (renderer.roundPixels || sprite.roundPixels)
             {
                 renderer.context.setTransform(
                     wt.a,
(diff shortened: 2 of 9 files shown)


themoonrat:
(happy to turn off by default for text if we're worried about backwards compatibility and the expectation for text to look the same as it already is, and have this as an idea way to opt in on certain pieces of text that have issues)

Lobsterovich:
Cool app

themoonrat:
@ivanpopelyshev This PR definitely improves the visuals of Text whilst not touching the visuals of regular sprites.

@bigtimebuddy I'm happy to do that. I'll be staying on v4 for a while after v5 is released, so I need to get into the habit of submitting any of my pixi v4 enhancements to official pixi v5 :)
Could close this PR and then point to it in the FAQ to let others know how to solve this issue for v4?

ivanpopelyshev:
I'm against changing behaviour for v4 because people solve this problem differently, and my solution includes extra arithmetical operations, i dont see them in this commit.

Please add this workaround to the wiki (GOTCHAS)

PR #4838, opened 2018-04-07
Set  bound texture size on context change

GoodBoyDigital:


packages/core/src/textures/TextureSystem.js
```diff
@@ -68,6 +49,17 @@ export default class TextureSystem extends System
 
         this.CONTEXT_UID = this.renderer.CONTEXT_UID;
 
+        let i;
+
+        const maxTextures = gl.getParameter(gl.MAX_TEXTURE_IMAGE_UNITS);
+
+        this.boundTextures.length = 0;
```
englercj: You should set this to `boundTextures` so the engine doesn't create a sparse array later when you assign to out-of-bounds indices.

packages/core/src/textures/TextureSystem.js
```diff
@@ -81,8 +73,6 @@ export default class TextureSystem extends System
 
         gl.bindTexture(gl.TEXTURE_CUBE_MAP, this.emptyTextures[gl.TEXTURE_CUBE_MAP].texture);
 
-        let i;
```
englercj: Rather than removing this, you should move it into the intialization portion of the for loop.

bigtimebuddy:
Thanks @englercj

PR #4842, opened 2018-04-11
Compressed textures initial commit

ivanpopelyshev:
Support for compressed textures. `cnr` and `ktx` are not included.

Work is in progress.

Example: http://pixijs.io/examples/?v=next-compressed#/textures/not-premultiplied-alpha.js

### DIFF
--- a/packages/core/src/textures/TextureSystem.js
+++ b/packages/core/src/textures/TextureSystem.js
@@ -55,6 +55,13 @@ export default class TextureSystem extends System
          * @readonly
          */
         this.managedTextures = [];
+
+        /**
+         * List of compressed textures extensions that are available
+         * @member {{}}
+         * @readonly
+         */
+        this.compressedExtensions = {};
     }
 
     /**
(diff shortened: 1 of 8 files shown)


ivanpopelyshev:
Can someone help me with crossOrigin ?

ivanpopelyshev:
Also, please look up old pixi-compressed-textures example with extension chooser: http://pixijs.io/examples/#/textures/dds.js , code is here: https://github.com/pixijs/pixi-compressed-textures/blob/master/src/extensionChooser.js .

I think that's enough information for you guys to decide the future of compressed textures in pixi: whether its a package with one file or multiple files. 

Everything critical is already inside v5, so I leave all the architectural decisions up to you, @mat and @bigtimebuddy !

bigtimebuddy:
Chatted about this with @GoodBoyDigital this week. I think there are a couple things to consider about this PR: Because of the ecosystem is so fragmented with different filetypes and formats, in order to have this be default functionality in PixiJS, we have a responsibility to simply the problem and not push all the burden to devs. Part of that is focusing on the highest values filetypes, providing some convenience for loading/choosing those types and for making recommendations (or create) tooling to support the export of these formats. The benefits seems pretty huge for compressed textures in terms of managing GPU memory, but the bar is also much higher than simply using a PNG. 

I think this is a good start, but not there yet. We should kick this over to Mat and have him iterate a little more on it.

ivanpopelyshev:
Its fine, I'm just happy its not my problem anymore :)

ivanpopelyshev:
Please dont leave this thing hanging.

We need to do something about it:
1. I can create a plugin for v5.
2. We can move this thing to a module and create "pixi-extra" bundle that has even more modules than legacy.

bigtimebuddy:
I’m closing this until @GoodBoyDigital has time to work on it, but will leave the branch for future development.

Issue #3866, opened 2017-03-23
Uncaught TypeError: Cannot read property 'push' of undefined


danipaleu:
Uncaught TypeError: Cannot read property 'push' of undefined
    at FilterManager.freePotRenderTarget (pixi.js:114)
    at FilterManager.popFilter (pixi.js:114)
    at MaskManager.popSpriteMask (pixi.js:114)
    at MaskManager.popMask (pixi.js:114)

Hello , I am receiving this error after a while playing the game, and I need to restart the PC for the game to work second time, can somebody help me with this ?

GoodBoyDigital:
Hi there! Please can you provide more information. Thanks!

englercj:
Hey @danipaleu, do you have a code sample that reproduces this issue that we can investigate?

Thanks!

nunocf:
We've come across the same problem over the past few weeks and we've narrowed it down to a key generation not matching it's key on the filter's map.

in freePotRenderTarget() function in FilterManager.js
```
freePotRenderTarget(renderTarget)
    {
        const minWidth = renderTarget.size.width * renderTarget.resolution;
        const minHeight = renderTarget.size.height * renderTarget.resolution;
        const key = ((minWidth & 0xFFFF) << 16) | (minHeight & 0xFFFF);

        this.pool[key].push(renderTarget);
    }
```
For us, it doesn't break on all of the devices.
When it does happen, the renderTarget used in this function has a different width and height to the one used when generating the key for the pool.
At the moment it's unclear wether this is because we're handling a different renderTarget or if the width and height have been recalculated at some point (this issue only happens for a few of our users, and even the same copy of the code hosted on another domain works fine).

We've come up with a way of avoiding the error, but we're causing a memory leak in it's place.
the solution is as follows:
```
freePotRenderTarget(renderTarget) {
        const minWidth = renderTarget.size.width * renderTarget.resolution;
        const minHeight = renderTarget.size.height * renderTarget.resolution;
        let key = ((minWidth & 0xFFFF) << 16) | (minHeight & 0xFFFF);

        if (this.pool[key] === undefined) {
            for (var current in this.pool) {
                key = current;
            }
        }
        this.pool[key].push(renderTarget);

    }
```
bear in mind this only works in our case because we never use more than one key in our pool.

I hope this information helps you guys coming up with a possible fix for this. 
Thanks!
nunocf

ivanpopelyshev:
anyway, that thing is fixed in pixi-v5. 

[...]

PR #4858, opened 2018-04-23
backport filter pool key

ivanpopelyshev:
#3866 and other issues.

@Patrick-Hammond confirmed that we still have the problem, and I want to solve it radical way.

### DIFF
--- a/src/core/renderers/webgl/managers/FilterManager.js
+++ b/src/core/renderers/webgl/managers/FilterManager.js
@@ -595,6 +595,7 @@ export default class FilterManager extends WebGLManager
         renderTarget.resolution = resolution;
         renderTarget.defaultFrame.width = renderTarget.size.width = minWidth / resolution;
         renderTarget.defaultFrame.height = renderTarget.size.height = minHeight / resolution;
+        renderTarget.filterPoolKey = key;
 
         return renderTarget;
     }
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Works like before:

https://jsfiddle.net/3kvazjw1/17/ - key = 'screen'
https://jsfiddle.net/3kvazjw1/20/ - key = '33554944' 

I dont have examples with a bug, but I'm sure that solution is correct - we store the result instead of recalculation.

bigtimebuddy:
“I’m sure the solution is correct”, I don’t like any uncertainty here. Can you please try to dig up an erroring example so we can verify?

ivanpopelyshev:
Its the same case as in https://github.com/pixijs/pixi.js/issues/4745 and in https://github.com/pixijs/pixi.js/pull/4579#issuecomment-378627432

Are you sure you want a third demo for it?

ivanpopelyshev:
You are right. it was fun to reproduce it, we have one more case. Here it is: https://jsfiddle.net/3kvazjw1/21/

Chrome at 90% zoom gives us `0.8999999761581421`

```
var res = 0.8999999761581421;
console.log(1 / res * res);
```

shows `0.9999999999999999`

That's why minWidth and minHeight are less than expected, and the key is wrong.

ivanpopelyshev:
@cursedcoder all according to [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754)

PR #4861, opened 2018-04-25
Interaction and Accessibility as Application plugins

ivanpopelyshev:
##  Overview

**💥Breaking Change** This PR removes `app.renderer.plugins.interaction` and `app.renderer.plugins.accessibility` API and moves the InteractionManager and AccessibilityManager to be plugin of Application, not Renderer (e.g. `app.interaction`, `app.accessibility`). This is more aligned with functionality like Loader and Ticker, which are not strictly systems for rendering (unlike extract and prepare).

### Application Usage

Interaction is automatically created when creating a new Application instance.

```js
app.interaction.on('click', () => {});
```
### Non-Application Usage

I've made an article: https://github.com/pixijs/pixi.js/wiki/v5-Custom-Application-GameLoop


ivanpopelyshev:
@bigtimebuddy please fix the build system. Both on my computer and in travis, it somehow uses older version of legacy bundle.

bigtimebuddy:
@ivanpopelyshev can you take a look at this, created `registerPlugin` for Application ALSO slightly changed the InteractionManager constructor arguments.

This feels better to me.

ivanpopelyshev:
I really like what I see there.

We have to add custom Application class that creates plugins directly in wiki and reference it from Application docs.

bigtimebuddy:
Here's an example extending PIXI.Application, pretty straightforward!

```js
class MyGame extends PIXI.Application {
  constructor(options) {
    super(options);
    this.physics = new PhysicsSubsystem();
  }
  destroy(removeView) {
    this.physics.destroy();
    this.physics = null;
    super.destroy(removeView);
  }
}

const app = new MyGame();
```

Here's an example implemented as a plugin, this is applied to ANY instance of PIXI.Application.

```js
PIXI.Application.registerPlugin({
  init() {
    this.physics = new PhysicsSubsystem();
  }
  destroy() {
    this.physics.destroy();
    this.physics = null;
  }
});

```

ivanpopelyshev:
Ok, I'll make my own variant of that straightforward app

bigtimebuddy:
I had a conversation with @GoodBoyDigital about this and wanted to echo some thoughts:

* I suggested we remove all non-rendering plugins from Renderer. Additionally, this would probably include accessibility API, should have a plugin middleware.
* Documentation and example should reflect both basic usage (e.g. new Application) and advanced/custom usage (e.g. new InteractionManager example above)

themoonrat:
I really like this as a way to keep the renderer purer. I assume the same process could take place for a number of other plugins, like Accessibility?

Couple of points....
1. The only downside to this specific change to interaction manager is that in my version, rather than passing through a resolution to the interaction manager, I have it look directly to the renderer.resolution property. This allows me to dynamically change the resolution of a game during runtime and still have interaction work. This separation no longer gives me that natural hook to be able to do that.
2. JSDoc wise, I wonder if it might be best to create a new settings object in the InteractionManager itself, and then reference that in places like the InteractionPlugin, and Application - otherwise you're repeating
```
* @param {object|boolean} [options.interaction] - Options for interaction manager.
*  Set it to `false` to remove interaction completely.
* @param {boolean} [options.interaction.autoPreventDefault=true] - Should the manager automatically prevent
*  default browser actions.
* @param {number} [options.interaction.interactionFrequency=10] - Frequency increases the interaction events
*  will be checked. 
```
in multiple places. You'd rather Application just referring to docs located within the relevant class
```
* @param {PIXI.interaction.InteractionManager#constructorConfig} [options.interaction] - Options for interaction manager.
```
of #constructorSettings or #constructorOptions. But you get the point!

bigtimebuddy:
@themoonrat, thanks for the feedback.

In regards to 2): we have the same problem with Application, autoDetectRenderer and Renderer options, lots of duplicate JSDocs. In this case, I actually think I want to remove the interaction object all together so the options would be simply:

```
@param {boolean} [options.interaction=true]
@param {boolean} [options.autoPreventDefault=true]
@param {number} [options.interactionFrequency=10]
```

This is more consistent with Ticker options. There's not an `options.ticker` object. A future PR could clean up the options and maybe make typedefs for them or something with less duplication.

As far as 1): how common is changing resolution after a game starts? I've never had to do this for any games I've worked on. Could you provide some more context why you're doing this and how important it is? If this use isn't generalized, it would be easy enough to add another `interaction.resolution = 2` line after you set the renderer's resolution. If it appears to be a generalized use-case, I suggest we make an event on renderer that listens for resolution changes and have app/interaction manager subscribe to that.

themoonrat:
The use case for 1. is dynamic resolution based on performance. If the measure fps consistently goes below a certain threshold, lower the resolution help get the fps back up. 
To be fair, as long as there is a interaction.resolution public property, this is something I could manually update on both the renderer and on the interaction plugin at the same time.

ivanpopelyshev:
There are problems with dynamic resolution anyway: you have to change FILTER_RESOLUTION, and some other fields. If we want dynamic resolution to be a thing, we need more changes.

bigtimebuddy:
Okay, thanks for clarifying @themoonrat. Based on your description, I'd qualify your usage as advanced. It's probably not something we should build convenience around and we'll make sure `resolution` is documented properly on InteractionManager so you can set it directly.

bigtimebuddy:
I also converted accessibility to an Application plugin as well.

GoodBoyDigital:
Hey Peeps!

Had a good chat with @bigtimebuddy around this one. I have had a good time to digest this and here are my thoughts:

Firstly, im all up for a plugin system for pixi application! I thinks its great :D 

My only concern is that moving interaction to an application level may have negative effects. Even though interaction and loading are not technically part of the rendering process. The renderer itself becomes practically useless for day to day use without these features. Looking at other rendering engines out there (both 2D and 3D) Interaction lives in the renderer part. I think this is correct. 

Heres my understanding of what the two main class are:

**Renderer** 
All basic functionality to deal with all aspects of rendering without being too opinionated about how you structure the rest of your application. 

**Application** 
Purpose is convenience. Make life easy for users by providing a high level framework to build awesome Pixi apps.

**My suggestion to this PR:**
- New plugin architecture for Application is _sweet_. This is where all highlevel plugins would and should live. I approve 👍 
- I believe Interaction is a core feature of the renderer and belongs in the renderer itself (as it currently does) This is because it is not about convenience, but essential functionality.

**Benefits:**
- The renderer maintains all low level functionality required to remain practical and be used as a component to a users application.
- We still have the benefit of Application plugins and convenience features that so many user will find super useful.

Hope that makes sense!

Cheers!

ivanpopelyshev:
@GoodBoyDigital I will make an example of how to make custom application class or just roll with it like in pixi-v3. Its important that people can introduce a few variables and make their own gameloop without using our mashup.

If we satisfy this condition, pixi remains a collection of libraries and not a overgrown framework.

bigtimebuddy:
@GoodBoyDigital Could we make interaction is a "system" instead of a "plugin"? `renderer.interaction` instead of `renderer.plugins.interaction`? That feels more in-line with what you are suggesting. It doesn't feel essential right now being at the same level as extract, prepare.

Also, what about accessibility? App plugin or renderer plugin?

GoodBoyDigital:
@bigtimebuddy definitely! Anything that goes in the renderer should be a system for sure.

Then 'plugins' will be an Application level thing. Seems clean 👍

ivanpopelyshev:
> definitely! Anything that goes in the renderer should be a system for sure.

What about custom renderer plugins? There is a different between app-level plugins and renderer-level plugins.

bigtimebuddy:
@ivanpopelyshev can you give examples of a custom renderer plugins that use `Renderer.registerPlugin`

ivanpopelyshev:
Its possible that in v5 I wont need it, because `geometry` and `shader` systems are very powerful in v5. 

At the current moment we cant just remove it, because there are no new mechanisms that correspond to old `pluginName` field

ivanpopelyshev:
When we will find a way to move ObjectRenderer somewhere else, then we will be able to remove renderer plugins.

ivanpopelyshev:
Ok, I've done it!

https://github.com/pixijs/pixi.js/wiki/v5-Custom-Application-GameLoop

ivanpopelyshev:
@bigtimebuddy please add a link to this article in Application docs.

bigtimebuddy:
I'm going to close this PR and make a new one with only the Application plugins, not changing interaction or accessibility's relationship to the Renderer.

PR #4879, opened 2018-04-30
Add vertexDirty boolean to Mesh, with backwards compatibility flag.

### LINKED ISSUE #4037: MeshRenderer#render: Question
i've noticed that for uvBuffer upload
```
 if (mesh.dirty !== glData.dirty) {
      glData.dirty = mesh.dirty;
      glData.uvBuffer.upload(mesh.uvs);
}
```

and indexBuffer upload

```
if (mesh.indexDirty !== glData.indexDirty) {
     glData.indexDirty = mesh.indexDirty;
     glData.indexBuffer.upload(mesh.indices);
}
```

a dirty check is defined

Why not for vertexBuffer upload?

```
 glData.vertexBuffer.upload(mesh.vertices);
```

Cristy94:
This should fix #4037, should not introduce any breaking changes and can be used as an optimization for static meshes.

### DIFF
--- a/src/mesh/webgl/MeshRenderer.js
+++ b/src/mesh/webgl/MeshRenderer.js
@@ -99,12 +99,6 @@ export default class MeshRenderer extends core.ObjectRenderer
             glData.indexBuffer.upload(mesh.indices);
         }
 
-        // For backwards compatibility
-        if (mesh.autoVertexUpdate)
-        {
-            mesh.vertexDirty++;
-        }
-
         if (mesh.vertexDirty !== glData.vertexDirty)
         {
             glData.vertexDirty = mesh.vertexDirty;
--- a/src/mesh/webgl/MeshRenderer.js
+++ b/src/mesh/webgl/MeshRenderer.js
@@ -73,6 +73,7 @@ export default class MeshRenderer extends core.ObjectRenderer
                 vao: null,
                 dirty: mesh.dirty,
                 indexDirty: mesh.indexDirty,
+                vertexDirty: mesh.vertexDirty,
             };
 
             // build the vao object that will render..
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
What if we call it "autoVertexUpdate" and make it true by default? Pixi doesnt have "manualXXX" fields

Cristy94:
That's what I had initially, but it was strange setting something to false to enable it, I can change it though.

bigtimebuddy:
I'm also fine with `autoVertexUpdate = true`, it's more that the user is opting out of default behavior than enabling something else.

src/mesh/webgl/MeshRenderer.js
```diff
@@ -98,7 +99,17 @@ export default class MeshRenderer extends core.ObjectRenderer
             glData.indexBuffer.upload(mesh.indices);
         }
 
-        glData.vertexBuffer.upload(mesh.vertices);
+        // For backwards compatibility
+        if (mesh.autoVertexUpdate)
```
ivanpopelyshev: This increment has to be moved to refresh(): https://github.com/pixijs/pixi.js/blob/dev/src/mesh/Mesh.js#L211

Cristy94: Isn't it the same as the one for `indicies` and `uvs`? I see that the refresh function is a noop.

ivanpopelyshev: refresh function is called from renderWebGL. I know that your code works, i just wantto move that particular part to Mesh itself, because its his updateID.

ivanpopelyshev: At this moment, refresh takes care of texture changes, and forces recaltulation of uvTransform if needed. ITs not fully no-op and its the only function there that does something with updates.

bigtimebuddy: Thoughts @Cristy94?

Cristy94: If that refresh function is called each tick it should be fine. I didn't have the time to go back to it and make the changes yes.

ivanpopelyshev:
I tested it locally on our strip example, it works.

PR #4880, opened 2018-04-30
fix data_uri regex to allow charset=utf8;base64 for the encoding part…

guigzz:
This PR initially solves IE11 issue where inline SVG cannot load in a BaseTexture. By digging into the code, I just enabled the DATA_URI regex to catch the whole "charset=utf8;base64" string as the encoding, and then I adapted the check on the encoding in BaseTexture to pass when this new type of encoding is found.

So this PR fixes inline SVG loading on IE11, please let me know if it is not enough generic or could be done in a better way, or if it actually breaks things (I tested it though).

Cheers.

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -227,7 +227,7 @@ export const URL_FILE_EXTENSION = /\.(\w{3,4})(?:$|\?|#)/i;
  * @type {RegExp|string}
  * @example data:image/png;base64
  */
-export const DATA_URI = /^\s*data:(?:([\w-]+)\/([\w+.-]+))?(?:;(charset=[\w-]+|base64|charset=[\w-]+;base64))?,(.*)/i;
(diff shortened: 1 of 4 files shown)


englercj:
I would prefer this regex:

```js
/^\s*data:(?:([\w-]+)\/([\w+.-]+))?(?:;charset=([\w-]+))?(?:;(base64))*?,(.*)/
```

So it captures the charset and encoding separately. For exmaple for your test case of `data:image/svg+xml;charset=utf8;base64,PGRpdiB4bWxucz0Pg==` you get:

| Group Index | Value |
| -- | -- |
| 0 | data:image/svg+xml;charset=utf8;base64,PGRpdiB4bWxucz0Pg== |
| 1 | image |
| 2 | svg+xml |
| 3 | utf8 |
| 4 | base64 |
| 5 | PGRpdiB4bWxucz0Pg== |

Issue #4090, opened 2017-06-07
SpriteRenderer#sprites retains destroyed sprites


lucap86:
Steps

- Take heap snapshot
- add some bunny
- take heap snapshot
- destroy bunnies container
- collect garbage
- take an heap snapshot
- filter by "Objects allocated between Snapshot 1 and Snapshot 2"

Try also to look at  ```app.renderer.plugins.sprite.sprites``` before and after destroy them all

![screen shot 2017-06-07 at 16 27 56](https://user-images.githubusercontent.com/15343100/26883878-a8116fcc-4b9e-11e7-949a-3a4580f993f9.png)

[test.zip](https://github.com/pixijs/pixi.js/files/1058342/test.zip)

englercj:
Which version of the lib is this?

lucap86:
sorry in the example is 4.4.4 bu it also happen with 4.5.2
here it is the test case with last pixi
[test-4.5.2.zip](https://github.com/pixijs/pixi.js/files/1060170/test-4.5.2.zip)
 version

englercj:
Thanks @lucap86 for the details, we'll look into this as soon as we find time.

nunocf:
Any update on this? We're facing the same problem of having lots of tiny memory leaks that would be good to address.

PR #4899, opened 2018-05-08
Resetting sprites array on flush (#4090)

YahorPadlucki:
Fix for #4090

### DIFF
--- a/src/core/sprites/webgl/SpriteRenderer.js
+++ b/src/core/sprites/webgl/SpriteRenderer.js
@@ -466,6 +466,7 @@ export default class SpriteRenderer extends ObjectRenderer
 
         // reset elements for the next flush
         this.currentIndex = 0;
+        this.sprites.length = 0;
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
I dont think that `.length=0` removes that optimization, array does stilll has the same capacity and `push` has to work fast.

GoodBoyDigital:
Cool! If theres no difference in perf, i'm happy! But lets test first 👍

ivanpopelyshev:
https://jsperf.com/js-empty-array/1

30-50% slower depends on browser and platform.

I agree that `length=0` is fast enough for rendering loop, but for SpriteRenderer we need as much speed as we can.

Line 262, `sprites[i] = null` will handle this better.

YahorPadlucki:
@GoodBoyDigital  @ivanpopelyshev thanks for review and explanations. It is valuable experience for me.

Issue #4279, opened 2017-09-04
PIXI.mesh.NineSlicePlane display error when its width less than its center block width.


jcyuan:
see online test: 
    http://jc-space.com/testnsp/

when the image's width / height is set to less than the width of it's nine slice center, it shows a weird display.  i think in this siutation the image should be scaled like a normal image which does not have a nineslice setting.

ivanpopelyshev:
Feel free to change it: https://github.com/pixijs/pixi.js/blob/dev/src/mesh/NineSlicePlane.js

If you add less than 50 lines of code for that, I suppose we can merge it. Otherwise, its a good idea for a plugin.

jcyuan:
i'll try it ,man

ivanpopelyshev:
I'll help if you have any questions about that code.

ivanpopelyshev:
Also, you may just make a simple patch or make your own class that extends PIXI.mesh.Plane, I can help with converting it onto our code style.

jcyuan:
fixed, so close

PR #4916, opened 2018-05-18
fixed NineSlicePlane resize problem

jcyuan:
issue: #4279
test:
https://jsfiddle.net/ygq38tgs/2/

This PR I mainly fixed the resize problem.
but this NineSlicePlane class still has problem which I cannot or hard to fix, probably you guys will re-construct the class so I don't handle on it.

The problem:

1. `super(texture, 4, 4)`, in Plane.constructor, it will call refresh method which casuses useless calucations, because now `this._leftWidth` / `this._topHeight` etc are `undefined`, so all things in `_refresh`, `updateHorizontalVertices`, `updateVerticalVertices` methods are calculated as `NaN`!

2. as you can review on what I modified, in the constructor I modified all the `_this.leftWidth` etc. as `_this._leftWidth`, because we don't need to call attribute here but only set value for members instead, the attribute will also call refresh which is unnecessary, at the last line of this constructor it will finally do force `refresh`.

### DIFF
--- a/src/mesh/NineSlicePlane.js
+++ b/src/mesh/NineSlicePlane.js
@@ -74,7 +74,7 @@ export default class NineSlicePlane extends Plane
          * @memberof PIXI.NineSlicePlane#
          * @override
          */
-        this.leftWidth = typeof leftWidth !== 'undefined' ? leftWidth : DEFAULT_BORDER_SIZE;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Great work @jcyuan, and thanks for the brilliant demo example!

PR #4934, opened 2018-05-25
Fixes Sprite's texture listener not getting removed

### LINKED ISSUE #4930: Sprite listens to texture update event after being detroyed
Here is where the listener in question is registered https://github.com/pixijs/pixi.js/blob/dev/src/core/sprites/Sprite.js#L591
I'm guessing it should be removed in the destroy function, otherwise if the sprite is destroyed before the texture is loaded, it will throw an exception when the event triggers.
Probably not a common use case, lazy loading sprites, but should be an easy fix.

bigtimebuddy:
Fixes #4930

Adds a unit-test that fails on **dev**

### DIFF
--- a/src/core/sprites/Sprite.js
+++ b/src/core/sprites/Sprite.js
@@ -418,6 +418,8 @@ export default class Sprite extends Container
     {
         super.destroy(options);
 
+        this._texture.off('update', this._onTextureUpdate, this);
+
         this._anchor = null;
 
         const destroyTexture = typeof options === 'boolean' ? options : options && options.texture;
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Maybe call it only if texture wasn't loaded, we dont want to add extra iterations there.

```js
if (!value.baseTexture.hasLoaded) ...
```

Its up to you.

bigtimebuddy:
Seems kinda silly to optimize with that if-check in the destroy function. Not a performance critical place in the code.

Adam13531:
I just upgraded to a version of Pixi that includes this change, and my code ended up breaking because I was relying on being able to call `destroy` multiple times. Now, calling it more than once will result in an undefined error for `this._texture.off(...)`.

I don't think that what I'm doing should be considered a Pixi bug, but I just wanted to add my two cents here. I'll likely end up wrapping my `destroy` calls with a check to see if `texture` exists.

PR #4943, opened 2018-05-30
feature/rectangle-update

YopSolo:
added some getter (topLeft, bottomRight and size) that return PIXI.Point
added some useful method when dealing with Rect (intersects, intersection, Union)

### DIFF
--- a/src/core/math/shapes/Rectangle.js
+++ b/src/core/math/shapes/Rectangle.js
@@ -102,18 +102,18 @@ export default class Rectangle
     get bottomRight()
     {
         return new Point(this.right, this.bottom);
-    }    
+    }
 
     /**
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@YopSolo Thank you for this PR and we definitely appreciate the illustration! I think the only thing that's missing from this PR is some unit-tests for all these new members and methods. These are fairly easy and straightforward things to test, so that shouldn't be hard.

More importantly, PixiJS v4 (the current **dev** branch) is not accepting new features, we are focusing our effort on v5 (the **next** branch). I'm going to close this, but encourage you to make this PR for v5 (to **next**). If this is something you need right away, I suggest you make a plugin this looks like easy functionality to mixin.

YopSolo:
oh thx, i'll do it soon

PR #4951, opened 2018-06-03
Creates unit-test and improves single and multiple image font loading

### LINKED ISSUE #4950: Single Page BitmapFont parsing issue
### Expected Behavior
version 4.7.3 I load a bitmap font (.xml) and BitmapText was working
### Current Behavior
version 4.8.0 I load the a bitmap font (.xml) and BitmapText is not working, because the frame size of the children of BitmapText is x: 0, y:0 , w:0, h: 0 but the texture is the proper BitmapFont texture
### Possible Solution
If there is one page for bitmapFont in the xml there is no page attribute for characters, which means 
the page is NaN and the frame rectangle will be 0 0 0 0 for it. Some more additional conditions are needed. 
//BitmapText.js
...
page = parseInt(letter.getAttribute('page'), 10);
if (isNaN(page)) {
    textureRect = new core.Rectangle(0, 0, 0, 0);
    page = 0;
} else {
    textureRect = new core.Rectangle(parseInt(letter.getAttribute('x'), 10) / res + textures[page].frame.x / res, parseInt(letter.getAttribute('y'), 10) / res + textures[page].frame.y / res, parseInt(letter.getAttribute('width'), 10) / res, parseInt(letter.getAttribute('height'), 10) / res);
}
...

### Steps to Reproduce
make a bitmapFont in xml format which has only one texture , where char tags doesn't have a page attribute and create a BitmapText.  Use PIXI 4.8.0 version - with 4.7.3 it works
### Environment

- `pixi.js` version:  4.8.0
- Browser: Chrome 66
- Device: Mac

bigtimebuddy:
Fixes #4950

This is a clean up of #4641 which had some issues. Namely:
* Did not gracefully support manually side-loading images with the loader
* Introduced potential race-conditions when loading split image fonts
* Fails to work with single page fonts which do not contain a `page` attribute on `<char>` elements

### Fixed

* Without `page` attributes for the `<char>` element, falls back to a default of `0` instead of creating a 0 width and 0 height Texture frame.
* Fixes and adds tests for side-loading images inside fnt file and prefer an explicit object map of page file names to textures as the first parameter for `BitmapText.registerFont`. This will make sure that textures line up, even if the pages are out of order or have non-numerical IDs.

```js
// If registering fonts with split images, use a map of the "file" attributes
// referenced in the <page> object within the fnt file
PIXI.extras.BitmapFont.registerFont(data, {
   'font-0.png': texture0,
   'font-1.png': texture1,
   'font-2.png': texture2
});

// Using an array will just map the textures to the page in the order
// which they appear in the fnt file <pages> section
PIXI.extras.BitmapFont.registerFont(data, [
   texture0,
   texture1,
   texture2
]);
```
cc @ceco-fmedia

### DIFF
--- a/src/extras/BitmapText.js
+++ b/src/extras/BitmapText.js
@@ -544,7 +544,7 @@ export default class BitmapText extends core.Container
      * @static
      * @param {XMLDocument} xml - The XML document data.
      * @param {Object.<string, PIXI.Texture>|PIXI.Texture|PIXI.Texture[]} textures - List of textures for each page.
-     *  If providing an object, the key is the `<page />` element `id` attribute in the FNT file.
+     *  If providing an object, the key is the `<page>` element's `file` attribute in the FNT file.
      * @return {Object} Result font object with font, size, lineHeight and char fields.
      */
     static registerFont(xml, textures)
@@ -554,6 +554,7 @@ export default class BitmapText extends core.Container
         const common = xml.getElementsByTagName('common')[0];
         const pages = xml.getElementsByTagName('page');
         const res = getResolutionOfUrl(pages[0].getAttribute('file'), settings.RESOLUTION);
+        const pagesTextures = {};
 
         data.font = info.getAttribute('face');
         data.size = parseInt(info.getAttribute('size'), 10);
@@ -566,17 +567,14 @@ export default class BitmapText extends core.Container
             textures = [textures];
         }
 
-        // Convert the texture array to a map of the textures
-        // where the key is the page ID and value is the texture
-        if (textures instanceof Array)
+        // Convert the input Texture, Textures or object
+        // into a page Texture lookup by "id"
+        for (let i = 0; i < pages.length; i++)
         {
-            const map = {};
+            const id = pages[i].getAttribute('id');
+            const file = pages[i].getAttribute('file');
 
-            for (let i = 0; i < pages.length; i++)
-            {
-                map[pages[i].getAttribute('id')] = textures[i];
-            }
-            textures = map;
+            pagesTextures[id] = textures[file] || textures[i];
         }
 
         // parse letters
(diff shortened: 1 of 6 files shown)


bigtimebuddy:
I think I’ve noticed another major flaw in the BitmapText loader, it assumes that the images are loaded in the correct order of the ids of the page elements. The registerFont method should probably take a map of page ids or names to Textures in addition to an Array.

src/extras/BitmapText.js
```diff
@@ -543,58 +543,61 @@ export default class BitmapText extends core.Container
...
         }
 
+        // Convert the input Texture, Textures or object
+        // into a page Texture lookup by "id"
+        for (let i = 0; i < pages.length; i++)
+        {
+            const id = pages[i].getAttribute('id');
+            const file = pages[i].getAttribute('file');
+
+            pagesTextures[id] = textures[file] || textures[i];
```
andrewstart: Should this be `file ? textures[file] : textures[i]` or something similar in the off chance that someone has a file with the exact same name as a property or method of arrays?

test/loaders/bitmapFontParser.js
```diff
@@ -370,6 +370,50 @@ describe('PIXI.loaders.bitmapFontParser', function ()
...
+        loader.load(() =>
+        {
+            const page0 = path.join(this.resources, 'split_font_ab.png');
+            const page1 = path.join(this.resources, 'split_font_cd.png');
+
+            expect(loader.resources[page0].metadata.pageFile).to.equal('split_font_ab.png');
+            expect(loader.resources[page1].metadata.pageFile).to.equal('split_font_cd.png');
+
+            const font = PIXI.extras.BitmapText.fonts.split_font2;
+            const charA = font.chars['A'.charCodeAt(0) || 65];
```
andrewstart: Any particular reason to have the `|| 65` here? If `charCodeAt()` stopped working, then that would be a problem for BitmapText.

bigtimebuddy: Copy and pasted from another test, but I think it’s silly in this context. Will fix

PR #4952, opened 2018-06-04
Added support for drawing complex rounded rectangles

Mwni:
Next to Graphics.drawRoundedRect() there now is Graphics.drawRoundedRectComplex() which expects 4 radii instead of 1. Going clockwise starting from top left.
core.math.shapes.RoundedRectangle has been changed in order to hold the 4 different radii. The contains(x,y) function has been adjusted, as well as the associated drawing methods for Canvas & WebGL.

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -751,7 +751,24 @@ export default class Graphics extends Container
      */
     drawRoundedRect(x, y, width, height, radius)
     {
-        this.drawShape(new RoundedRectangle(x, y, width, height, radius));
+        return this.drawRoundedRectComplex(x, y, width, height, radius, radius, radius, radius);
+    }
+
+    /**
+     *
+     * @param {number} x - The X coord of the top-left of the rectangle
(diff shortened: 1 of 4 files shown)


Mwni:
Ah, I see. Since I changed the RoundedRectangle class, the tests for it fail.

ivanpopelyshev:
We dont accept those things in `dev` anymore, only in `next`. Unless @bigtimebuddy wants to make another `drawStar` :)

ivanpopelyshev:
Also you didnt have to close it, just push more commits in your branch (Mwni:dev). However, its better to make a PR against `next` (pixi-v5)

Mwni:
@ivanpopelyshev Alright, thanks for the info. Should I modify the RoundedRectangle class in a way, so it accepts both 1 or 4 radii, or create a separate class ComplexRoundedRectangle for that?

ivanpopelyshev:
The problem is that @GoodBoyDigital is doing Graphics refactoring, so I really cant advice anything. We have to wait for both @GoodBoyDigital and @bigtimebuddy

PR #4953, opened 2018-06-04
updating PIXI Rectangle with some props and methods

YopSolo:
added some getter (topLeft, bottomRight and size) that return PIXI.Point
added some useful method when dealing with Rect (intersects, intersection, Union)
added test for methods
40722763-7ed58ee8-641d-11e8-9415-2f309ef84a5c
40722764-7ef2dc00-641d-11e8-9aad-c2fef5a245cc

### DIFF
--- a/packages/math/src/shapes/Rectangle.js
+++ b/packages/math/src/shapes/Rectangle.js
@@ -1,3 +1,4 @@
+import { Point } from '../Point';
 import { SHAPES } from '../const';
 
 /**
@@ -90,6 +91,38 @@ export default class Rectangle
     get bottom()
     {
         return this.y + this.height;
+    }
+
+    /**
+     * The location of the Rectangle object's bottom-right corner, determined by the
+     * values of the right and bottom properties.
+     *
+     * @member {PIXI.Point}
+     */
+    get bottomRight()
+    {
+        return new Point(this.right, this.bottom);
+    }
+
+    /**
+     * The location of the Rectangle object's top-left corner, determined by the x
+     * and y coordinates of the point.
+     *
+     * @member {PIXI.Point}
+     */
+    get topLeft()
+    {
+        return new Point(this.left, this.top);
+    }
+
+    /**
+     * The size of the Rectangle object, expressed as a Point object with the values of
+     * the width and height properties
+     * @member {PIXI.Point}
+     */
+    get size()
+    {
+        return new Point(this.width, this.height);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
First off, great PR, well organized, easy to read, good unit tests. Thank you so much.

For some context, in the past we've said "no" to more Point methods for things like calculating distance, and other more advanced calculations. In general, we have left math methods that aren't useful for PixiJS directly (that is, not used internally) up to devs to add for themselves and consider outside the scope of PixiJS which is not a framework, but a renderer. I worry about accepting these methods because they add bloat when they could easily be a plugin/mixins to existing `PIXI.Rectangle.prototype`. It might be nice to create a standalone library that contains all the Point, Rectangle advanced stuff.

A good example of why we'd want to have developer implement these types of things themselves is for performance. For instance, in the `bottomRight`, etc getters you are creating a new Point every time it's called. In some circumstances, this isn't acceptable for high performing games like being called every tick.

That said, if there's a lot of enthusiasm for adding `union` and `intersection` methods to Rectangle, we should consider adding. 

@YopSolo could you provide some use-cases of why these methods would improve rendering for you?

What do other people thing? Is this a must-have? I think if there's not a decent amount of enthusiasm then a plugin seems like the best approach to move this forward.

cursedcoder:
@bigtimebuddy great point about "not a framework, but a renderer".

In my projects we've implemented such methods as external helpers.

However lot of people could find this useful, but the problem is – for such small/easy thing they won't go to github and investigate if there's special plugin/package. Much easier to copy-paste from SO or somewhere else.

**Maybe** this PR is not much of bloat and probably won't required further bugfixes/support expenses, since formulas are always constant.

Or this can be a good addition via separate pixi package i.e. `@pixi/rectangle-diff`.

bigtimebuddy:
@ivanpopelyshev what do you think about this?

ivanpopelyshev:
@bigtimebuddy Methods are general enough, and it makes sence to add them. Im sure we have something like that inlined, we have to search it and replace.

I think code style for union and intersects is different. one is using props `bottom` `right` while other doesn't.

themoonrat:
For me, the functions are useful, but the getters are inefficient (due to returning new points each time) bloat (because it's information that's already very easily available). They're the kinda thing I'd expect a dev to add some util functions for their side if they feel they need them. Since no-one has asked for them up until this point, I'd assume that's what devs have done and been happy with. Include them here, and you've increase the size of the API, and next would be 'well, why do rectangles have a top left, bottom right and a size, but not sprites, text, containers etc.' Then would we include those getters there too for consistency? It's a potential slippery slope

bigtimebuddy:
Had a chat with @GoodBoyDigital about this PR and we discussed creating a new `@pixi/math-extras` optional package which could contain a bunch of mixins for Rectangle, Point, etc. This could be the start of that.

cursedcoder:
So @YopSolo would you like to extract this PR into `@pixi/math-extras`?

bigtimebuddy:
@miltoncandelero math-extras ☝️

PR #4968, opened 2018-06-11
Add options to VideoBaseTexture constructor

cobywhite:
Since actions are taken in the constructor if the video is already loaded and ready to play that users are unable to stop with setting properties after the constructor runs this adds those options to the VideoBaseTexture constructor.


cursedcoder:
Would be nice to add simple tests just to ensure options are properly set.

cobywhite:
Abandoning this PR will open another.

cobywhite:
In next this issue is already addressed.

PR #4969, opened 2018-06-12
Do not quote escape generic font families

cobywhite:
The CanvasRenderingContext2D.font property when taking in CSS generic font families doesn't work correctly when those families are quoted. This makes a whitelist to not quote those families.


cursedcoder:
+1

Issue #4970, opened 2018-06-17
PIXI.Text  with anchor.x = 0.5 is blurred if resulting textures's width is odd.


Megabyteceer:
to centralize single-line text i use this.anchor.set(0.5, 0);
but in 50% cases this text become blurred if  text.texture.widht is not even.

### Possible Solution
temporary i added this fix for me
```
PIXI.Text.prototype._onTextureUpdate = function() { // centred text with odd width is blurred bug fix
	if((this.style.align === 'center') && (this.texture.width & 1)) {
		this.x = Math.floor(this.x) + 0.5;
	} else {
		this.x = Math.floor(this.x);
	}
	PIXI.Text.__proto__.prototype._onTextureUpdate.call(this);
}
```
but one of better solutions may be is expand text's texture by one pixel if widht is not even.

### Environment

- `pixi.js` version: _e.g. 4.8.0
- Browser: Chrome 67
- Device:  Desktop

Megabyteceer:
Additionally, any textures with 0.5 0.5 anchor and  odd sized texture will be blured. Mabe wiil  be  great to fix anchor to set pivot to rounded x,y values.

Megabyteceer:
Additionally, any textures with 0.5 0.5 anchor and  odd sized texture will be blured. Mabe wiil  be  great to fix anchor to set pivot to rounded x,y values.

themoonrat:
Also, potentially not needed if in v5 I add the PR for https://github.com/pixijs/pixi.js/pull/4833 . This PR allows each display component to draw to rounded pixels only, and turns that on for Text by default to stop blurriness

Megabyteceer:
Very often case of using text in games is fly up text "+10"  when hero pick up coin of "-4" when hero gets damage. So with rounded x y text will fly not smoothly. Another case - UI panel with text which scrolls or meves sometime with easing. Any text with rounded x/y on UI element will move sharply. So i think enforce one object to be rounded and another to not is good for static objects only. In motion it will look bad.

PR #4992, opened 2018-06-30
Fix for odd sized text and svg textures. #4970

Megabyteceer:
When Text generates odd sized texture, its become blurred if set anchor to (0.5, 0.5).
PR expands canvas to even sizes. Trimmed canvas and SVG generated canvas included also by the way.

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -2,7 +2,7 @@
 import Sprite from '../sprites/Sprite';
 import Texture from '../textures/Texture';
 import { Rectangle } from '../math';
-import { sign } from '../utils';
+import { sign, expandToEven } from '../utils';
 import { TEXT_GRADIENT } from '../const';
 import settings from '../settings';
 import TextStyle from './TextStyle';
@@ -145,8 +145,8 @@ export default class Text extends Sprite
         const maxLineWidth = measured.maxLineWidth;
         const fontProperties = measured.fontProperties;
 
-        this.canvas.width = Math.ceil((Math.max(1, width) + (style.padding * 2)) * this.resolution);
-        this.canvas.height = Math.ceil((Math.max(1, height) + (style.padding * 2)) * this.resolution);
+        this.canvas.width = expandToEven(Math.ceil((Math.max(1, width) + (style.padding * 2)) * this.resolution));
+        this.canvas.height = expandToEven(Math.ceil((Math.max(1, height) + (style.padding * 2)) * this.resolution));
 
         context.scale(this.resolution, this.resolution);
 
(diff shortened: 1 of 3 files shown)


src/core/utils/index.js
```diff
@@ -123,6 +123,24 @@ export function hex2string(hex)
...
+ * @memberof PIXI.utils
+ * @function expandToEven
+ * @param {number} input - the integer to convert
+ * @return {number} even integer.
+ */
+export function expandToEven(input)
+{
+    if (input & 1)
+    {
+        input++;
```
themoonrat: would we want to put a Math.ceil around the input to ensure that any decimal places are also rounded up to whole integers?

Megabyteceer: In all cases function used, values is rounded already. But ok, i added Math.ceil.

src/core/text/Text.js
```diff
@@ -319,8 +319,8 @@ export default class Text extends Sprite
         {
             const trimmed = trimCanvas(canvas);
 
-            canvas.width = trimmed.width;
-            canvas.height = trimmed.height;
+            canvas.width = trimmed.width(trimmed.width);
+            canvas.height = trimmed.width(trimmed.height);
```
bigtimebuddy: Can you explain and give context for this change? It doesn't make sense to me. Should it be `expandToEvent(trimmed.height)`?

Megabyteceer: You are right, it is my mistake. Sorry. expandToEven is correct.

I think about texture extending, it is will add one pixel at right side, and all right aligned text may look not axactly aligned.

Maybe better will be to fix anchor applying? Round anchor's shifting:
Sprite.js
```
            w1 = trim.x - Math.round(anchor._x * orig.width);
            w0 = w1 + trim.width;

            h1 = trim.y - Math.round(anchor._y * orig.height);
            h0 = h1 + trim.height;
```
I tried this aproach, fixes blureness perfectly too.

themoonrat: Isn't that effectively enabling 'roundToPixels' on the display object?

ivanpopelyshev: No, its not. RoundPixels is implemented wrong way in pixi and its very hard to fix it. 

@Megabyteceer idea looks better because its very stupid and it works.

Megabyteceer: Math.round for anchor only is better, because you can still move object smoothly by x+=0.1 on subpixes speed, and object is sharp when it is stops at rounded x y.

bigtimebuddy:
Thank you very much for this PR @Megabyteceer, but I chatted with @GoodBoyDigital and we both feel like we should close it.

This doesn't feel this is an appropriate change for v4 right now since it introduces changes in the visual display of items in a way that is not backward compatible. This change could potentially introduce a bunch of 1px issues that dev are currently compensating for.

Overall, we should fix or re-implement `roundPixels` to accomplish this same thing, expanding to an even pixel amount seems not that elegant. The `roundPixels` option is the appropriate API to address this underlying issue. One idea that @GoodBoyDigital and I discussed was adding per-DisplayObject roundPixel snapping.

PR #4994, opened 2018-07-04
Fixes JSdoc Events in 3.5.5

bigtimebuddy:
Follow-up to #4232

This is a workaround for https://github.com/jsdoc3/jsdoc/issues/1425 which doesn't document `@event` correctly if the events are at the bottom of the constructor. Since this is a convention that PixiJS uses in a few places, I moved the events just above the last line. Event now show up correctly and works with JSdoc 3.5.5+.

### DIFF
--- a/packages/interaction/src/InteractionManager.js
+++ b/packages/interaction/src/InteractionManager.js
@@ -645,6 +645,7 @@ export default class InteractionManager extends EventEmitter
          * @event PIXI.DisplayObject#touchmove
          * @param {PIXI.interaction.InteractionEvent} event - Interaction event
          */
+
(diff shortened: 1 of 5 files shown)


packages/interaction/src/InteractionManager.js
```diff
@@ -647,6 +645,7 @@ export default class InteractionManager extends EventEmitter
          * @event PIXI.DisplayObject#touchmove
          * @param {PIXI.interaction.InteractionEvent} event - Interaction event
          */
+        this.setTargetElement(this.renderer.view, this.renderer.resolution);
```
themoonrat: style wise i'd move this just 1 line lower with a blank space in between, otherwise it looks like you're documenting setTargetElement

Issue #4154, opened 2017-07-10
Paths in PIXI.Graphics get auto-closed when using PIXI.CanvasRenderer and beginFill


fmhdp:
Hi,

first of all, thanks for this really nice framework. PIXI basically spared me writing my own, bug-ridden abstraction from WebGL with a canvas fallback.

One thing, that I believe to be a bug, I came across is a difference in render behavior between the WebGLRenderer and the CanvasRenderer. 
Assume a PIXI.Graphics object that draws an arc. If there is a beginFill command, even one with an alpha value of 0, the paths gets closed automatically in when using the PIXI.CanvasRenderer, while that is not the case with the WebGLRenderer.

Example:
https://jsfiddle.net/3snvu0dy/5/
PIXI version used in the fiddle should be 4.5.3.
I see no differences between Chrome, Firefox and Safari on OSX.

Expected behavior:
Same rendering behavior for canvas and webGL. The arc is not closed.

Experienced behavior:
The arc is closed only when using PIXI.CanvasRenderer.

I unfortunately did not have the time yet to dig into the PIXI code to see what causes the difference.

Thanks in advance for considering this :)

bigtimebuddy:
Looks like a bug. Thanks for reporting this!

PR #4997, opened 2018-07-05
Bugfix: Don't autoclose paths when using canvas (#4154)

fmhdp:
When defining a path, any call to beginFill results in the path being closed,
but only for the Canvas renderer, not for the WebGL renderer.
This commit removes the check that sets the shape to be closed when a fill is
active. This unifies the rendering behavior between Canvas and WebGL.

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -1223,7 +1223,7 @@ export default class Graphics extends Container
 
         if (data.type === SHAPES.POLY)
         {
-            data.shape.closed = data.shape.closed || this.filling;
+            data.shape.closed = data.shape.closed;
             this.currentPath = data;
         }
 


ivanpopelyshev:
Good catch! But I certainly know that it wont be accepted by other members of team if you dont make a fiddle for it :)

fmhdp:
@ivanpopelyshev 
I'll happily do that, but how do I get a build of pixi with my bugfix in a fiddle?

For the original misbehavior there was a fiddle in the referenced bug report.
This is the same fiddle, updated to use version 4.8.1 and show that the behavior is still the same:
https://jsfiddle.net/3snvu0dy/16/

ivanpopelyshev:
Its possible to use pixi builds from PR's if they are made in this repo (if I make a branch based on your PR). 

I believe that fiddle is enough. Lets wait when travis CI problem will be resolved.

bigtimebuddy:
Thanks @fmhdp!

PR #5010, opened 2018-07-09
Fixes #5006 for the arc method by not adding an additional point if i…

### LINKED ISSUE #5006: Path rendering in WebGL differs from Canvas, WebGL has "chips" in paths
Expected behavior:
Paths look the same when rendered with CanvasRenderer and WebGL renderer.

Observed behavior:
Paths look different when rendered with WebGL. 
An initial moveTo command adds an arrow like shape.
An arc added after a line chips off a bit of the line.
Example: https://jsfiddle.net/3snvu0dy/63/

This certainly is looking as obvious as it looks due to the fact that the line width in this example is really high. The example isn't artificial though, as I actually want to render that shape.

Relationship with other bugs:
This bug is very likely related to #3592 and #4046. 

I haven't verified it, but my best guess is that adding the second arc causes a duplicate point to be added to the points array for the shape, reducing the chip damage to what's described in #4046. Except that it's probably even less expected when adding a new shape to the graphics.

This is also not verified, but my best guess for the arrow-like shape at the start is that there's probably a point added to the points array for the shape when calling a moveTo command.

Environment:
PIXI 4.8.1, Fiddle looked the same in current Chrome (67) and Safari (11.1.1) on an iMac.
A random Windows 10 box renders it the same with current Chrome (67).

fmhdp:
Fixes #5006 for the arc method by not adding an additional point if it's too close to the last existing point.

The same problem might exist in other parts of the graphics code where I haven't encountered it yet. This changes only the arc method for the PIXI.Graphics.


fmhdp:
@ivanpopelyshev I think this one might warrant a fiddle showing the difference. If I understood you right, I'll have to wait until you've added the branch to the pixi.js repo.

bigtimebuddy:
Verified:

**Broken:** https://jsfiddle.net/3snvu0dy/63/
**Fixed:** https://jsfiddle.net/2fLue3g6/

ivanpopelyshev:
@fmhdp you are my hero. "0.001" is good enough temporary fix.

fmhdp:
@ivanpopelyshev You're welcome :)
I appreciate the fix going upstream.

PR #5011, opened 2018-07-10
Extended spritesheet

CodeAndWeb:
New features:

1) Takes default pivot points from sprite sheet if present in the data
2) Sprite sheet has function to receive animations defined in a sprite sheet

Tests and updated documentation included.


ivanpopelyshev:
Btw, I've added polygon-packing feature in pixi-heaven plugin, but changes are too big, need to consult with @bigtimebuddy @GoodBoyDigital if we add it in v4 or v5.

ivanpopelyshev:
I assume you've made changes in pixi template for texturePacker?

CodeAndWeb:
We've added a property called "anchorPoint" that holds the anchor point if enabled in TexturePacker. 
We've also added a map of animations.
Both are optional in the loader and only used if present.

Do you have an ETA for the next Pixi release with these features? We'll have to wait with the next TexturePacker release until the feature is available in Pixi.

About the Polygon-Packing: This is great news. We are happy to enable the feature with one of the next updates. Let's coordinate this after the current release.

djmisterjon:
it would also be great to see the option "PACK WITH SAME LAYOUT" .
I made my own hack to use it, but it not very operational.
https://github.com/pixijs/pixi-display/issues/30

image

CodeAndWeb:
Is the release going to be 4.9.0 or 4.8.2?  
We have to know this for the TexturePacker release notes... people should know which version of Pixi they have to use.

ivanpopelyshev:
looks like 4.8.2

themoonrat:
Might be just me, but a Texture having a property called anchor, and a Sprite having a property called anchor... isn't this ripe for some confusion? Won't people think about changing the texture anchor and expecting changes for existing sprites?

Could it be called defaultAnchor or something like that?

src/core/textures/Texture.js
```diff
@@ -40,8 +40,9 @@ export default class Texture extends EventEmitter
      * @param {PIXI.Rectangle} [orig] - The area of original texture
      * @param {PIXI.Rectangle} [trim] - Trimmed rectangle of original texture
      * @param {number} [rotate] - indicates how the texture was rotated by texture packer. See {@link PIXI.GroupD8}
+     * @param {PIXI.Point} [anchor] - Default anchor point used for sprite placement / rotation
      */
-    constructor(baseTexture, frame, orig, trim, rotate)
+    constructor(baseTexture, frame, orig, trim, rotate, anchor)
```
bigtimebuddy: If a default argument value is set here: `anchor = new Point(0, 0)`, you don't need to check for it later. Little cleaner.

src/core/textures/Texture.js
```diff
@@ -143,6 +144,13 @@ export default class Texture extends EventEmitter
             baseTexture.once('loaded', this.onBaseTextureLoaded, this);
         }
 
+        /**
+         * Anchor point that is used as default if sprite is created with this texture
+         * @type {PIXI.Point}
+         * @default {0,0}
+         */
+        this.anchor = anchor ? new Point(anchor.x, anchor.y) : new Point(0, 0);
```
bigtimebuddy: I think this `anchor` property name is problematic for reasons that @themoonrat mentioned. Another idea might be `inheritedAnchor`, `defaultAnchor` or something else? Another approach is that we mark this as `@private` doc since it's only used now internal to support TexturePacker's new feature and don't want to create confusion for users or expose new APIs.

src/core/textures/Texture.js
```diff
@@ -143,6 +144,13 @@ export default class Texture extends EventEmitter
             baseTexture.once('loaded', this.onBaseTextureLoaded, this);
         }
 
+        /**
+         * Anchor point that is used as default if sprite is created with this texture
```
bigtimebuddy: Since this is not an observable point, it should be noted that changing values to this point do not affect Sprites that reference this texture. That could confuse people.

src/core/textures/Texture.js
ivanpopelyshev: that's extra object allocation because it assigns "anchor" later too.

src/core/sprites/Sprite.js
```diff
@@ -30,14 +42,18 @@ export default class Sprite extends Container
...
-         * Setting the anchor to 0.5,0.5 means the texture's origin is centered
-         * Setting the anchor to 1,1 would mean the texture's origin point will be the bottom right corner
+         * The default is 0,0 or taken from the {@link PIXI.Spritesheet|Spritesheet}.
+         * 0,0 means the texture's origin is the top left, 0.5,0.5 is the center, 1,1 the bottom right corner.
          *
          * @member {PIXI.ObservablePoint}
          * @private
          */
-        this._anchor = new ObservablePoint(this._onAnchorUpdate, this);
+        this._anchor = new ObservablePoint(
```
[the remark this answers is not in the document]
ivanpopelyshev: I dont think that's a good idea. Animation are supposed to have same anchor every time, and user can change anchor somewhere in the middle of animation.

src/core/textures/Texture.js
ivanpopelyshev: The feature is good even without TexturePacker. Setting texture default anchor instead of setting sprite anchor every time. As i said earlier, its better to assign in constructor, so its not important whether its null or `new Point` here.

ivanpopelyshev: The feature is good even without TexturePacker. Setting texture default anchor instead of setting sprite anchor every time. As i said earlier, its better to assign in constructor, so its not important whether its null or new Point here.

bigtimebuddy: We have told many devs here that there are no new API features in v4. Supporting external tools like TexturePacker seems 🆗 though, however, Texture anchor should probably be v5 as a public API.

src/core/textures/Texture.js
bigtimebuddy: I don't feel strongly about this.

src/core/sprites/Sprite.js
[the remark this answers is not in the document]
bigtimebuddy: That fine, but documentation should notes this. Reseting texture does not update the anchor.

djmisterjon:
@CodeAndWeb 
First , thank a lot for this new update it just very nice for me and my workFlow.

I have a little problem with auto-animations.
It possible to add a special suffixe ? or regex ? for autoDetect from the end if we add a suffix ?
Because it not work in context  of my animation sequence have a special suffixe at the end.
How i can manage with this ?
I don't use Pack with same layout for now, because it not yet implement in PIXI loader.
image

thanks a ;lot

GoodBoyDigital:
Hey @CodeAndWeb quick question, is the anchor optional? As in if i don't specify it in Texturepacker it will not export an anchor point?

CodeAndWeb:
@GoodBoyDigital You only get anchor points in the file if you specify them in TexturePacker.

The pixi loader applies them when found while creating a new sprite. They can be overwritten from code at any time.

bigtimebuddy:
@CodeAndWeb thanks, just made a few doc changes and resolved conflicts. 

I want to double-check that TexturePacker generates anchor values (float from 0-1) and not pivot (real pixel values), is this correct? There were some places in the your docs and in your description in this PR where you refer to it as a "pivot". Sprite has both anchor and pivot.

CodeAndWeb:
@bigtimebuddy TexturePacker exports anchor values between 0..1 (or bigger, if anchor is outside of the sprite), the key in the JSON format is also called "anchor"

pietrovismara:
I'd like to notify that in my case, upgrading from 4.8.1 to 4.8.2 broke loading textures with `PIXI.loader.add(path)`, as they will miss the `defaultAnchor` property, throwing an exception (`Cannot read property 'x' of undefined`) when creating a sprite via `new Sprite(texture)`. When created with `PIXI.Texture.fromImage` they instead have `defaultAnchor` and sprite creation works. 

This might be or not an issue, perhaps my mistake but unfortunately i don't have more time to investigate/debug/open a real issue, so i thought i'd just leave a comment here.

bigtimebuddy:
@djmisterjon or @pietrovismara could either of you create a jsfiddle or pixiplayground.com that reproduces this? I've been unable to reproduce and I'm not sure if I'm missing something.

PR #5016, opened 2018-07-11
[BUG] Canvas Extract resolution

Bokos-:
The height and the width of the canvas were 2 times multiplied by the resolution.

http://jsfiddle.net/54zu1y7n/60/


ivanpopelyshev:
wow.

PR #5017, opened 2018-07-11
FilterSystem bugfix

ivanpopelyshev:
FilterArea is padded every frame! Lets fix it.


ivanpopelyshev:
Here's a fiddle: https://codepen.io/clf/pen/RJXXqM
Fixed: https://codepen.io/ivanpopelyshev/pen/vaOmeZ

Original thread: http://www.html5gamedevs.com/topic/38711-pixijs-sprite-and-mask/

The problem is that I know that this thing wont work properly with `resolution!=1` and there are bugs across `FilterSystem` and `RenderTextureSystem`. 

I will make some tests with mocks later, as a very big PR, to address that filter rectangles issues properly. I dont want people to stop just because I need time to study it, that's why I submit temporary fix.

PR #5021, opened 2018-07-13
Added removeFromParent method to DisplayObject

arnovanoordt:
Simplifies removing a child from a parent.
No need anymore for checking if the child is actually attached to a parent in the first place.


bigtimebuddy:
Thank you for this @arnovanoordt. Like @ivanpopelyshev said, no new API changes for v4. Feel free to make this PR for **next**, will be closing. 

Also, we would like need to see some simple unit-tests to validate this new method when you resubmit. Thanks.

PR #5024, opened 2018-07-16
Fixed tint of graphics object doesn't change on canvas renderer.

ceco-fmedia:


bigtimebuddy:
Thanks @ceco-fmedia made a jsfiddle that proves this fixed:
* Fixed: http://jsfiddle.net/shqmgf9w/8/
* Broken: http://jsfiddle.net/Lz654jre/2/

PR #5026, opened 2018-07-16
Fix typo in Graphics.lineStyle's jsdoc

tleb:
The default alignment value is 0.5, not 1.


bigtimebuddy:
Good catch @tleb! Thank you and very helpful first PR.

PR #5027, opened 2018-07-16
BitmapFont loader fix

### LINKED ISSUE #5025: pixi 4.8.1 bitmap font loading error. Cannot read property 'frame' of undefined
When loading bitmapfont I get a runtime error.

I bumped my pixi.js from v3.7.4 to v4.8.1 and then this error was introduced.

### Expected Behavior
Bitmap font should just be loaded and registered without errors. 

### Current Behavior
I get this runtime error
```
Uncaught TypeError: Cannot read property 'frame' of undefined
    at Function.registerFont (BitmapText.js:581)
    at parse (bitmapFontParser.js:5)
    at completed (bitmapFontParser.js:72)
    at Loader.bitmapFontParser (bitmapFontParser.js:89)
    at Loader.js:614
    at async.js:35
```

### Steps to Reproduce
* It doesn't _always_ fail, might be related to browser cache. Please test with empty cache.
* It doesn't seem to matter if I first try to load the image then fnt file.

0. Open <https://codepen.io/markknol/pen/WKweKd>
0. Open devtools
0. Observe error
0. If there is no error, add a space in the js, to re-run the pen. This triggers the error.

In general this is enough to reproduce
```js
let options = { crossOrigin: true };
let loader = new PIXI.loaders.Loader();
loader.add("fnt", "font.fnt", options); 
loader.add("png", "font_0.png", options);
loader.load();
```

### Environment

- `pixi.js` version: 4.8.1
- Browser: Chrome 67
- Device: _e.g. Desktop

ivanpopelyshev:
#5025

### DIFF
--- a/src/loaders/bitmapFontParser.js
+++ b/src/loaders/bitmapFontParser.js
@@ -92,10 +92,19 @@ export default function ()
             // using the same loader, resource will be available
             for (const name in this.resources)
             {
-                if (this.resources[name].url === url)
+                const bitmapResource = this.resources[name];
+
+                if (bitmapResource.url === url)
                 {
-                    this.resources[name].metadata.pageFile = pageFile;
-                    completed(this.resources[name]);
+                    bitmapResource.metadata.pageFile = pageFile;
+                    if (bitmapResource.isComplete)
+                    {
+                        completed(bitmapResource);
+                    }
+                    else
+                    {
+                        bitmapResource.onComplete.add(completed);
+                    }
                     exists = true;
                     break;
                 }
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
How to test it:

Add setTimeout to textureParser:

```js
export default function ()
{
    return function textureParser(resource, next)
    {
        // create a new texture if the data is an Image object
        if (resource.data && resource.type === Resource.TYPE.IMAGE)
        {
            setTimeout(function test()
            {
                resource.texture = Texture.fromLoader(
                    resource.data,
                    resource.url,
                    resource.name
                );
                next();
            }, 3000);
        }
        else
        {
            next();
        }
    };
}
```

Then run https://codepen.io/ivanpopelyshev/pen/LBNERQ on it.

englercj:
Always remember, all middleware in resource-loader are async. ;)

Good catch @ivanpopelyshev!

PR #5028, opened 2018-07-17
canvasRenderer could affect bounds calculation

ivanpopelyshev:
Related to #5024

"dirty" is a number, and we cant just set boolean there, because it affects other calculations based on it.


bigtimebuddy:
**Broken:** https://jsfiddle.net/9jvt1yhb/9/
**Fixed:** https://jsfiddle.net/ekrzbydv/3/

PR #5030, opened 2018-07-17
Fixes generation of textures on the canvas renderer when it has no parent

themoonrat:
Don't know why I closed https://github.com/pixijs/pixi.js/issues/4317 without submitting a PR
Essentially the fix is to force updating of transforms if the display component you are trying to generate a texture from has no parent itself. 

Before:  https://jsfiddle.net/themoonrat/Lnkr5bkg/
After: https://jsfiddle.net/tx4g1jma/


themoonrat:
It was true to avoid having to re-calculate of all transforms along the render tree down to your display object.

But if your object hasn't got a parent, then you need to set it to false to allow it to use the _tempDisplayObjectParent as it's parent

a-rasinski:
After this change was released in 4.8.2, I have observed strange issue happen in a project I'm working on.
We're using generated textures there and trying to support both canvas and webgl renderer.

The case is that after running code like:
```
const sprite = Sprite.fromFrame(textureName)
const image = renderer.extract.image(sprite)
```
where sprite isn't attached to any parent, we stop getting some transform updates in the first container in a stage after canvas resize. I tried to isolate this into a codepen, but it's hard to reproduce in sterile env somehow...

If mentioned texture generation on parentless target is never run, everything works fine.

So I did a bisect and got into this pull request. After some investigation I've found the probable cause here:
https://github.com/pixijs/pixi.js/blob/037c65068607f3f01bacc5159aa6b5263ff7727b/src/core/renderers/canvas/CanvasRenderer.js#L188

I've searched the codebase for occurencies of `_worldID` and `_parentID` and I noticed a practice of flagging parent transform as dirty by assigning `_parentID = -1`, but never to `_worldID`

After changing the code line mentioned above in CanvasRenderer to:
`this._tempDisplayObjectParent.transform._parentID = -1;`
issue with lack of transform on canvas resize disappears and everything is working fine again.

I don't fully know how transforms are built in pixi yet, so it's just a hunch from my side.

a-rasinski:
@themoonrat Hi, did you have a chance to look into the issue I mentioned above? Or should I try to isolate this and create new ticket?

themoonrat:
Sorry @a-rasinski - I added this as a bookmark as a 'come back to later' but just didn't :/

If you could have a look regarding this that'd be great. Would be good to start with the latest v5 release candidate, and see if it's something that's all sorted in v5 that perhaps needs backporting to v4, or if it's an issue that's carried over.

Thanks!

themoonrat:
@a-rasinski I can't see this as an issue in v5
Using https://www.pixiplayground.com/#/edit/8n9Jy3acWWJyv9XZRlKm1 and commenting out the addChild(bunny) so it has no parent... that second bunny is being created via renderer.extract.image and works fine

Dawson1999-stack:
Earth Taco's

PR #5031, opened 2018-07-17
[Next] Basic zIndex support

themoonrat:
Taps into onChildrenChange function to re-sort children automatically.
May or may not be wanted.... but thought I'd throw this out there as a basic solution

Example: https://jsfiddle.net/r4k86tn3/

### DIFF
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -1,6 +1,16 @@
 import removeItems from 'remove-array-items';
 import DisplayObject from './DisplayObject';
 
+function sortChildren(a, b)
+{
+    if (a.zIndex === b.zIndex)
+    {
+        return a.updateOrder - b.updateOrder;
+    }
+
+    return a.zIndex - b.zIndex;
+}
+
 /**
  * A Container represents a collection of display objects.
  * It is the base class of all display objects that act as a container for other objects.
@@ -33,13 +43,17 @@ export default class Container extends DisplayObject
     }
 
     /**
-     * Overridable method that can be used by Container subclasses whenever the children array is modified
-     *
-     * @private
+     * Overridable method that can be used by Container subclasses whenever the children array is modified.
+     * By default it will ensure that children are ordered by zIndex and updateOrder.
      */
     onChildrenChange()
     {
-        /* empty */
+        for (let i = 0, j = this.children.length; i < j; ++i)
+        {
+            this.children[i].updateOrder = i;
+        }
+
+        this.children.sort(sortChildren);
     }
 
     /**
--- a/packages/display/src/DisplayObject.js
+++ b/packages/display/src/DisplayObject.js
@@ -1,7 +1,8 @@
 import EventEmitter from 'eventemitter3';
 import { Rectangle, Transform } from '@pixi/math';
 import Bounds from './Bounds';
-// _tempDisplayObjectParent = new DisplayObject();
+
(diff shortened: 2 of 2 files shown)


ivanpopelyshev:
Calling `sort` on every onChildrenChange is a serious performance problem.

I recommend special sorting container, with basic but fast algorithm like this one: https://github.com/pixijs/pixi-display/wiki#fast-container-sort

themoonrat:
As ivan points out, the sorting algorithm sucks and is inefficient; but I think a basic premise for an API called `zIndex` that effectively sorts the rendering order for you is sound. It sucks to deal with them manually. 

I'm happy to instead move this to a plugin for v4 that is 'basic sorting only'

But zIndex for v5. Think about it ;)

bigtimebuddy:
@themoonrat, thank you for this. I think it's a terrific idea for v5. Couple of questions:

- Can you elaborate more on the rationale for `updateOrder`? Falling-back to sort based on the order in which an object was created seems a little too magical. Why not just have the sort method return `0`?

- If `sort` is expensive, we may consider making it opt-in like `interactive` for performance reasons. Maybe Container has a `sortableChildren` boolean which default to `false`. I would hate for someone's scene with many sprites to get hosed because of this addition. Thoughts?

themoonrat:
`updateOrder` is required to make the sort behaviour act normal, as the basic sort just caused unexpected issues for me. It also covers the following scenario:
object a and object have the same zIndex of 0. object a is earlier in the array. I set object a zIndex to 1. object a now appears later in the array than object b. If I set the zIndex of object a back to 0 again... I still expect object a to appear over the top of object b because the last render it had a higher priority.
I see with Ivans examples he took a different stance and had an 'arrival order' which stays static from when a child is first added to the array.
But in either case, you need to a backup comparison to maintain consistency

regarding the sort - it could just automatically enabled itself if it detects it is needed. So, by default, it never sorts anything. But as soon as a child is added that has a zIndex, the sorting turns on. Or, if a child is added without a zIndex, but then later the setter is used.... then that setter notifies the parent to enable sorting.

So if no-one uses it, nothing changes.
If someone starts to use it, the sorting starts.

bigtimebuddy:
@themoonrat:

* If it's possible to auto-detect without too much overhead, then that's a better approach.
* I think the arrival order has a lot of design advantages over the DisplayObject creation order, as you've implemented it. Basically, reverts back to the original children's order. If any child has non-zero `zIndex`, will sort. 
* We'll need some very good test coverage on this feature, btw. There are a lot of test cases to cover like adding children before and after changing zIndex, reverting all zIndex to 0, etc.

bigtimebuddy:
@themoonrat, how do you want to move forward? Do you want to close, rework and resubmit the PR when it's ready?

themoonrat:
@bigtimebuddy Sure, can close down for now.
I think the one thing I'm concerned about, having thought about it, is putting in and using zIndex as a property when @ivanpopelyshev 's plugin is so popular as an advanced solution, and already uses the zIndex property. 
So either I'll create this as a 'simple zIndex solution' plugin, or, @ivanpopelyshev, is there a way this could be implemented without effecting your work?

ivanpopelyshev:
Make it custom container, and maybe one small change in Container/DisplayObject to store something temporarily on addChild removeChild and such. I'll help to optimize it.

PR #5033, opened 2018-07-23
check if tint & fillColor are changed

finscn:
if tint & fillColor are not changed, won't recompute `sprite.tint`


bigtimebuddy:
@finscn could you provide an example?

finscn:
An example for what ? for performance comparing ?

bigtimebuddy:
Sorry, I should have been more specific.

In the absence of any unit-testing, we need to see an example of this functionality to ensure that nothing broke.

Issue #5039, opened 2018-07-25
Getting error when set trim=true to the empty text


corporalik:
![image](https://user-images.githubusercontent.com/1940989/43205934-30c1378e-902d-11e8-8626-c84d9412065b.png)

Hi there. Getting an error when trying to create any text with empty value. In my case the value will be assigned dynamically after some calculations.

### Steps to Reproduce
`let text = new PIXI.Text('',{trim:true});`

### Steps to Reproduce
`let text = new PIXI.Text('Initial text',{trim:true});`
`text.text = '';`

### Environment

- `pixi.js` version: 4.8.1

johnSjo:
PR with a suggested fix: https://github.com/pixijs/pixi.js/pull/5045

bigtimebuddy:
Merged into dev. Will be released with 4.8.2

PR #5045, opened 2018-07-29
Fixes trimming a Text object with an empty string (#5039)

johnSjo:
If you give this trim function an empty canvas the ```bound``` objects properties will remain ```null```, causing an error when trying to get the image data.

So if this happens we can simply set the bounds manually to give us one empty pixel instead.


bigtimebuddy:
@johnSjo thanks for submitting this PR 👍 

@themoonrat what do you think of this? I think you originally created `trimCanvas`.

My only concern/issue is that I don't think the underlying issue (trimming a PIXI.Text with an empty string) is a bug with trimCanvas and more of a Text issue with how it handles a `null` result. But I could be convinced that trimCanvas should always return a min 1x1 canvas. Thoughts?

themoonrat:
Hey @bigtimebuddy - it wasn't me who originally added this, but I share your concerns about the 'fudge'. 
The trimCanvas util is not exported anywhere, so the upside is what we decide won't have a knock-on effect.
I wouldn't have it as 1 pixel... because if you're using a resolution of less than 1, then that 1 pixel texture is going to be an issue. Our 'starting' canvas for each Text item is at least 3x3!

Options are
1. Able to set dimensions to return should canvas be totally transparent as a parameter, with a sensible default set
2. Modify the util function to stop it erroring, and to allow it to return 0 for width and height, and null for data. Then do something in the Text class. to handle this

I kinda like 2... if I'm saying 'find the trimmed canvas', and it's all trimmed, that's a valid response. 
Then in the text class, rather than
```
if (this._style.trim)
{
    const trimmed = trimCanvas(canvas);

    canvas.width = trimmed.width;
    canvas.height = trimmed.height;
    this.context.putImageData(trimmed.data, 0, 0);
}
```
it could be
```
if (this._style.trim)
{
    const trimmed = trimCanvas(canvas);

    if ( trimmed.data )
    {
        canvas.width = trimmed.width;
        canvas.height = trimmed.height;
        this.context.putImageData(trimmed.data, 0, 0);
    }
}
```

If the image is transparent... then it doesn't matter how big the texture is going to be, surely? So save a slow putImageData call altogether

bigtimebuddy:
Like this! Thanks for the input @themoonrat I like your suggestion (#2)

@johnSjo, is this something you’d like to rework?

PR #5049, opened 2018-07-31
Next speedup

GoodBoyDigital:
Added a couple of caching layers for binding textures and binding buffers.

### DIFF
--- a/packages/core/src/geometry/GeometrySystem.js
+++ b/packages/core/src/geometry/GeometrySystem.js
@@ -42,6 +42,11 @@ export default class GeometrySystem extends System
          * @private
          */
         this.cache = {};
+
+        /**
+         * A cache of currently bound buffer..
+         */
+        this.boundBuffers = {};
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/core/src/geometry/GeometrySystem.js
```diff
@@ -42,6 +42,11 @@ export default class GeometrySystem extends System
          * @private
          */
         this.cache = {};
+
+        /**
+         * A cache of currently bound buffer..
```
bigtimebuddy: Could you add at `@member {object}` to this doc? Also would be nice to elaborate on the description here if this is, in fact, public.

bigtimebuddy:
@GoodBoyDigital I noticed that many of the Systems don't have destroy methods (although it looks like they are supported via Renderer's runners). Maybe another PR to do some cleanup?

englercj:
Note that this will only work if these are the only systems that make those calls. Otherwise what they have cached as bound and what is actually bound can desync.

ivanpopelyshev:
`reset()` is needed, yes, i'll check that it works with threeJS, I have a number of demos.

GoodBoyDigital:
yes, another PR would be best

PR #5057, opened 2018-08-05
Ability to set individual display components to be drawn to rounded pixels

themoonrat:
v5 version of https://github.com/pixijs/pixi.js/pull/4833


themoonrat:
The only effect is at the point of the render... so only a visual test would show anything

bigtimebuddy:
@themoonrat thoughts about @GoodBoyDigital's feedback?

tools/renders/src/Renderer.js
```diff
@@ -75,7 +75,6 @@ class Renderer
             backgroundColor: 0xffffff,
             antialias: false,
             preserveDrawingBuffer: true,
-            roundPixels: true,
```
themoonrat: This is the one place I couldn't think of how to replicate existing behaviour... since this test renderer renders both canvas and webgl at the same time...

bigtimebuddy: I’m not sure this is even critical for these tests. I’m thinking about removing these renders anyways in favor of the v5 render test suite that @Goodboydigital and I’ve have been building. Basically, don’t worry about it.

packages/mesh/src/Mesh.js
```diff
@@ -2,6 +2,7 @@ import { State } from '@pixi/core';
 import { DRAW_MODES } from '@pixi/constants';
 import { Point, Polygon } from '@pixi/math';
 import { Container } from '@pixi/display';
+import { settings } from '@pixi/settings';
```
bigtimebuddy: You need to add this package to the dependencies within mesh.

bigtimebuddy:
I like this approach much better. Do you think it’s worth a deprecated warn if someone sets this renderer option? I don’t think it’s critical but these are nice things to do for developers.

packages/core/src/AbstractRenderer.js
```diff
@@ -57,6 +55,14 @@ export default class AbstractRenderer extends EventEmitter
         // Add the default render options
         options = Object.assign({}, settings.RENDER_OPTIONS, options);
 
+        if (options.roundPixels)
+        {
+            // eslint-disable-next-line no-console
+            deprecation('5.0.0', 'roundPixels is no longer a valid configuration setting passed to the renderer. '
+            + 'Instead, each sprite/mesh has its own roundPixels property. '
+            + 'Change PIXI.settings.ROUND_PIXELS to set the default roundPixels value for newly created sprites/meshes.');
+        }
```
bigtimebuddy: We have not found a good way to deprecate things like options, which are inline. However, I like where you're going with this. The v4 deprecation function is here:

https://github.com/pixijs/pixi.js/blob/next/bundles/pixi.js/src/deprecated.js#L1-L48

I'm not a huge fan of having two functions that do the same thing. For the sake of getting this PR through, why don't you add a TODO inline comment to add a deprecation notice when detecting `options.roundPixels`. In another PR, we can generalize the deprecation function, because I think that's a little more complicated because of the `slice` and the differences in function signatures.

bigtimebuddy:
@GoodBoyDigital, could you give this another look?

bigtimebuddy:
@themoonrat could you resolve these conflicts?

themoonrat:
@bigtimebuddy Done. However, SpriteRenderer no longer exists. So where I'd done
```
if (sprite.roundPixels)
            {
                const resolution = this.renderer.resolution;
                  // xy
                 float32View[index] = ((vertexData[0] * resolution) | 0) / resolution;
                 float32View[index + 1] = ((vertexData[1] * resolution) | 0) / resolution;
 
                 // xy
                 float32View[index + 5] = ((vertexData[2] * resolution) | 0) / resolution;
                 float32View[index + 6] = ((vertexData[3] * resolution) | 0) / resolution;
 
                 // xy
                 float32View[index + 10] = ((vertexData[4] * resolution) | 0) / resolution;
                 float32View[index + 11] = ((vertexData[5] * resolution) | 0) / resolution;
 
                 // xy
                 float32View[index + 15] = ((vertexData[6] * resolution) | 0) / resolution;
                 float32View[index + 16] = ((vertexData[7] * resolution) | 0) / resolution;
             }
```
I couldn't replicate this post merge. It's replacement, BatchRenderer, has no reference to resolution, and no reference to roundPixels. Dunno what to do here?

themoonrat:
Also, there are 2 Mesh.js files now, and I've no idea which one is the correct one to put roundPixels into. So i've put into both... for now......

bigtimebuddy:
We want to resolve mega-batch PR first before merging this in. Will probably need your help to resolve conflicts when we're ready.

themoonrat:
okies

bigtimebuddy:
@themoonrat could you work to resolve conflicts now? Thanks.

GoodBoyDigital:
hey @themoonrat , we could does ```element.roundPixels``` give you the info?
We could also access resolution by passing it through as a function argument?

themoonrat:
@GoodBoyDigital I think from this point, it's best if someone with ninja WebGL skills takes over the branch from me to re-add in the roundPixels in the right place in the right way!

ivanpopelyshev:
Explanation: https://www.pixiplayground.com/#/edit/EjwZocvU~CuhEDTTHKoTB

themoonrat:
@ivanpopelyshev isn't the current v4 behaviour and v5 canvas behaviour to just floor any values rather than round?

ivanpopelyshev:
Oh, wait, we didnt think about `resolution` there... Well, I think that round is better

ivanpopelyshev:
Lets ask @GoodBoyDigital if he wants rounding done inside the batcher or inside the element. I vote we leave it in Sprite/Mesh and don't mess with renderer internals.

themoonrat:
I agree

themoonrat:
I'm happy with it in terms of the roundPixels addition Ivan has made. 
Regarding resolution; it may already be handled, it's just something I noticed different from the area of code roundPixels was in for a v4 implimentation compared to a v5 implimentation. I think it's status needs checking independently of this PR.

bigtimebuddy:
Great work @themoonrat and @ivanpopelyshev for helping to land this one.

PR #5059, opened 2018-08-05
Setting text to be "" means it is actually "", and not transformed to " "

themoonrat:
As requested by https://github.com/pixijs/pixi.js/issues/4942
It is annoying to have to check against " " for an empty string when you've set that piece of text to "".
Internally, it'll use " " as it's just easier to deal with for measuring the size of text

### DIFF
--- a/packages/text-bitmap/src/BitmapText.js
+++ b/packages/text-bitmap/src/BitmapText.js
@@ -150,7 +150,7 @@ export default class BitmapText extends Container
         const pos = new Point();
         const chars = [];
         const lineWidths = [];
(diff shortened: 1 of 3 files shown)


themoonrat:
@bigtimebuddy Might be worth starting a 'v4 to v5' difference wiki page? Not sure what format it would take, but this is definitely a workaround a lot of devs may have had to implement previously that wouldn't be needed here

bigtimebuddy:
I don’t think we have one yet, but a great idea to keep a running list. The only serious breakage was BaseTextures not have resources. I’ll add something this week.

PR #5060, opened 2018-08-05
Able to set resolution of text separately to the renderer

themoonrat:
https://github.com/pixijs/pixi.js/issues/4915

By default, it auto adjusts to the render resolution, as per current behaviour
But if you use the resolution setter on the text, it becomes manually controlled.


bigtimebuddy:
Could add some basic unit-tests for the different scenarios in which resolution can be set (renderer, settings, manually)?

themoonrat:
@bigtimebuddy done

PR #5076, opened 2018-08-17
Add missing GraphicsData default param.

SonyStone:
Little fix to make GraphicsData types work like in js source file.


SonyStone:
@bigtimebuddy 

`lineAlignment` in js don't have default parameter.
image
What better to do?
Make `lineAlignment: number = 0` -> `lineAlignment?: number`
and `this.lineAlignment = lineAlignment || 0`;
or 
`lineAlignment: number = 0` -> `lineAlignment: number` ?

Issue #4882, opened 2018-05-01
Video type wrongly set


dtemes:
Trying to use a video from cloudfront, with a url like:

http://mycd.com/media/myvideo.mp4?Expires=12345678...

In firefox the video will not load and firefox will complains about the wrong type of video

### Expected Behavior
The media type is set to video/mp4

### Current Behavior
type is set to 'video/mp4?Expires=12345678......'

note: my cdn config is not setting a proper mime type
### Possible Solution

change the createSource function in videoBaseTexture.js which parses the video url badly:

if (!type)
    {
        type = `video/${path.substr(path.lastIndexOf('.') + 1)}`;
    }

### Steps to Reproduce
var texture = PIXI.Texture.fromVideoUrl('http://mycd.com/media/myvideo.mp4?Expires=12345678...');
var videoSprite = new PIXI.Sprite(texture);

### Environment

- `pixi.js` version: _e.g. 4.7.3_
- Browser: Firefox

gooderist:
Hello, can I work on this? If so, is there anything else I should be aware of?

gooderist:
Hello, can I work on this? Is it up for grabs?

bigtimebuddy:
Fixed in **dev**. Will be released in v4.8.2

PR #5077, opened 2018-08-18
fix Video type wrongly set

gooderist:
Addresses issue #4882 by throwing away the query string (everything after and including the '?')

### DIFF
--- a/src/core/textures/VideoBaseTexture.js
+++ b/src/core/textures/VideoBaseTexture.js
@@ -317,6 +317,7 @@ function createSource(path, type)
 {
     if (!type)
     {
+        path = path.split('?').shift().toLowerCase();
         type = `video/${path.substr(path.lastIndexOf('.') + 1)}`;
     }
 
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Thanks for this, would you mind adding a simple unit test to cover the query string URL?

gooderist:
Of course! However, I don't see an existing test file for VideoBaseTextures. How should I proceed? Add a new file or add the test to Texture.js or somewhere else?

themoonrat:
I would create a new file. It's good to keep all tests for a single class in their own file

gooderist:
Hello, new file and unit test for VideoBaseTexture added. Let me know if it looks off.

themoonrat:
lgtm :)
I'd just add a test a 'normal' scenario alongside a test for your bugfix. What if just a normal .ogv url was loaded for example? This will check that (1) the type shows ogv, and doesn't just work for mp4, and (2) checks that your bugfix doesn't negatively effect regular urls.

gooderist:
@themoonrat thanks for the feedback! I think I got it this time.

Issue #4863, opened 2018-04-26
PIXI 4.7.3 - mouseup/pointerup bug


Kamil93:
mouseup / pointerup bug

```
    this.on("pointerdown", (e) => {
      console.log("up", e.data.buttons); // left mouse button prints '1'
    });
    
    this.on("pointerup", (e) => {
      console.log("up", e.data.buttons); // any mouse button prints always '0'
    });
```

gooderist:
Hello, I'm trying to figure out what the bug is. What's exactly is the expected behavior?

On my setup, Firefox / macOS, "pointerup" only fires after I release all the buttons. "mouseup", on the other hand, will fire even if I'm holding other buttons down (ex: right click & hold, left click & release -> "mouseup" fires but not "pointerup"). That being the case, `onPointerUp` will only get called when `buttons === 0` (all buttons are released).

Here is [fiddle](https://jsfiddle.net/oeh0c1rz/7/) to test. This seems more like a browser issue than a PIXI one. Or am I completely off?

gooderist:
Hello @themoonrat - Thanks for the illuminating reply. It does seem like mouseup/down is the way to go especially since pointerup/down behavior varies between browsers. Maybe update the docs to recommend mouse events and call it a day?

bigtimebuddy:
Doc changes to clarify this welcomed!

themoonrat:
@gooderist Updated docs, yeah. I'll close this issue, because the behaviour is as intended, but updated docs to clarify the situation would be awesome.

I'd say that, in general, for the vast majority of use cases, using the 'pointer' events is much better. Because it covers mouse/touch/pen/whatever you want, and it just works. But in this specific case, where you're wanting to know the events of different mouse buttons as they are pressed; well then use the 'mouse' events that are there for just that reason :)

PR #5079, opened 2018-08-19
Update pointerup docs

gooderist:
Leave note in pointerup docs suggesting mousedown / mouseup instead when multiple button events are desired. Clarification added for #4863


bigtimebuddy:
Thanks @gooderist, nice one 👍

Issue #4890, opened 2018-05-04
bug: lineTo() removes previous line


Friksel:
I was just searching for some time to find out why my lines weren't showing up on screen during development of a path creation tool. I didn't expect pixi to erase my lines just by drawing them, but it does it for some reason;

When drawing a line with the Graphics object with `lineTo(x, y) ` and after that draw another line with `lineTo(x, y)` and the same x and y values, none of both lines will be drawn!!!

I understand drawing a line to to the same spot is not very useful, but during debug process this can happen, for example for quick testing, like happened to me today. 
It's confusing if, like in my case, no line is drawn at all if this happens. I would never expect drawing two lines would result in no line at all.

So this is drawing a line:
```
g.moveTo(0, 0);
g.lineTo(200, 200);
```

But this is not drawing anything:
```
g.moveTo(0, 0);
g.lineTo(200, 200);
g.lineTo(200, 200);
```

### Expected Behavior
The second lineTo() should be ignored by pixi, because it's having no length. But the first line should not be removed.

### Current Behavior
None of the lines are drawn when there are more following `lineTo`s with same values

### Possible Solution
Ignore lineTo's if they have zero length but don't change previous lines.

### Steps to Reproduce
I've just created this pen with an example: https://codepen.io/Friksel/pen/OZjpBg
The first line is removed by the second `lineTo`.
Just remove the second `lineTo` in the example and the first line is visible again

### Environment
- `pixi.js` version: 4.7.3
- Browser: Chrome, Firefox, Edge etc.
- Device: Tested on Windows 10

ivanpopelyshev:
> I understand drawing a line to to the same spot is not very useful, but during debug process this can happen, for example for quick testing, like happened to me today.

There are many operatsions that cause `PIXI.Graphics` to lose screws. I want to port something better from one of existing vector libraries :)

Here is what i did with repeated lineTo and moveTo's one of those libs, that one is used by Google Chrome canvas2d context: https://bugs.chromium.org/p/skia/issues/detail?id=7792

Friksel:
Thanks @ivanpopelyshev . I see you have been busy up there.

Issue #3592, opened 2017-01-16
[Bug] Don't support lineTo the last position.


finscn:
The test code 
```
graphics.clear();
graphics.beginFill(0xFF3300, 0.2);
graphics.lineStyle(4, 0xffd900, 1);

graphics.moveTo(50,50);

// **** lineTo the last position  ( equals moveTo)
graphics.lineTo(50,50);

graphics.lineTo(250, 50);
graphics.lineTo(250, 250);
graphics.lineTo(50, 250);

// **** lineTo the last position  ( equals the last lineTo)
graphics.lineTo(50, 250);

graphics.lineTo(50, 50);
graphics.endFill();

```

In other game/render engine, the code could run correct, 
But in PIXI.js it will take an unexpect result.

staff0rd:
Hey @finscn - can you post an image of your expected and actual result?

finscn:
@staff0rd  ， 
```
var app = new PIXI.Application(800, 600, { antialias: true });
document.body.appendChild(app.view);

var graphics = new PIXI.Graphics();

graphics.clear();
graphics.beginFill(0xFF3300, 0.2);
graphics.lineStyle(4, 0xffd900, 1);

graphics.moveTo(50,50);

// **** lineTo the last position  ( equals moveTo)
graphics.lineTo(50,50);

graphics.lineTo(250, 50);
graphics.lineTo(250, 250);
graphics.lineTo(50, 250);

// **** lineTo the last position  ( equals the last lineTo)
graphics.lineTo(50, 250);

graphics.lineTo(50, 50);
graphics.endFill();

app.stage.addChild(graphics);
```

paste the code to  http://pixijs.github.io/examples/#/basics/graphics.js , you will see .

englercj:
Hey @finscn is this issue still active in v4.5.0?

Thanks!

finscn:
Yes , there is still this bug.

PR #5084, opened 2018-08-22
prevent gaps in duplicate lineTo calls

gooderist:
Ignore duplicate calls to lineTo (same x, y inputs) to prevent gaps in WebGL render. Addresses #4046 and #3592. edit: also #4890

Notes: it is not yet understood why duplicate lineTo calls caused gaps to begin with. Also, I'm not sure of a good way to unit test this without comparing render output. If anyone has anything to say about either point, I'm all ears. Thanks!

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -423,8 +423,16 @@ export default class Graphics extends Container
      */
     lineTo(x, y)
     {
-        this.currentPath.shape.points.push(x, y);
-        this.dirty++;
+        const points = this.currentPath.shape.points;
+
+        const fromX = points[points.length - 2];
+        const fromY = points[points.length - 1];
+
+        if (fromX !== x || fromY !== y)
+        {
+            points.push(x, y);
+            this.dirty++;
+        }
 
         return this;
     }
--- a/test/core/Graphics.js
+++ b/test/core/Graphics.js
@@ -104,6 +104,18 @@ describe('PIXI.Graphics', function ()
             expect(graphics.width).to.be.equals(70);
             expect(graphics.height).to.be.equals(70);
         });
+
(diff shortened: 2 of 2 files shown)


themoonrat:
Unit test wise you could count how many points in the path there are?

gooderist:
Okay, sounds good to me! Thanks!

englercj:
I think more importantly for the test is that you have the right points. Can you assert that the path is `[0, 0, 10, 0]`?

Otherwise :shipit:

gooderist:
Good idea, will do!

Btw, would you happen to know how to run the WebGLRenderer unit tests? They turn up as "pending" when I do npm test. Clearly, I need to run them on a browser or something but I'm not sure how.

gooderist:
Thanks @bigtimebuddy. So, I'm trying to run WebGLRenderer related tests. When I `npm run` they show up as `pending` like this...





This is to be expected since there is no WebGL renderer in node, but I did notice that electron is in the project and so I assumed there was some automated script to do render tests. Am I assuming correctly or should I check my settings or... ?

gooderist:
In case anyone else is unable to run WebGLRenderer tests (people with older graphics cards)

Your GPU is likely [blacklisted by Chromium](https://github.com/electron/electron/issues/8217) like mine (Intel HD Graphics 3000). The link contains instructions on how to ignore the gpu blacklist which will let your tests run like the wild stallions they yearn to be.

Or you can buy a new(er) Macbook. Up to you!

distinctdan:
Just ran into this myself when calling a `moveTo(100, 100); lineTo(100, 100)`. I'm using font path data exported by opentype.js, and the glyphs have some duplicate calls like that. Thanks for fixing it!

PR #5086, opened 2018-08-22
Fix gaps in circle borders during WebGL render

### LINKED ISSUE #4701: drawCircle() produces circles with a small wedge cut out of them if the radius is small
So I'm having an issue with drawCircle where it looks like the circle's outer border isn't closed. It's only visible at small radii. This screenshot was taken in Chrome, using pixi version 4.7.0.

circlegaperror

Here's the code for drawing the circle. I'm drawing it once, then updating the position repeatedly using position.set():

```
// In constructor
this.pixiGraphics = new PIXI.Graphics();

// In initial drawing function
this.pixiGraphics.clear();
this.pixiGraphics.beginFill(0x3DB3E6);
this.pixiGraphics.lineStyle(5, 0x8FC7E6);
this.pixiGraphics.drawCircle(0, 0, 3);
this.pixiGraphics.endFill();

// In move function - gets called repeatedly
this.pixiGraphics.position.set(this.loc[0], this.loc[1]);
```

gooderist:
Fixes #4701 

Issue caused by precision errors while calculating the line border's last point coordinates. Instead of calculating the final point, just set to the same coordinates as the first point (since circles generally end where they start).

Note: I wrote a test but when I run `npm test` it shows up as `pending`. How do I run my test? I will fix any errors in it once I get it to run. Thanks!

### DIFF
--- a/src/core/graphics/webgl/utils/buildCircle.js
+++ b/src/core/graphics/webgl/utils/buildCircle.js
@@ -82,13 +82,18 @@ export default function buildCircle(graphicsData, webGLData, webGLDataNativeLine
 
         graphicsData.points = [];
 
-        for (let i = 0; i < totalSegs + 1; i++)
+        for (let i = 0; i < totalSegs; i++)
         {
             graphicsData.points.push(
                 x + (Math.sin(seg * -i) * width),
                 y + (Math.cos(seg * -i) * height)
             );
         }
+
+        graphicsData.points.push(
+            graphicsData.points[0],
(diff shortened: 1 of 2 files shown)


themoonrat:
It might be because you're enforcing using a WebGLRenderer? I know other unit tests locations that uses the WebGLRenderer has a little guard there so that they are skipped in scenarios where WebGL isn't available.

Take a look at `WebGLRenderer.js` unit tests, and it's `const withGL = require('../withGL');` and how that is used

PR #5093, opened 2018-08-29
Fixes ObservablePoint compatibility with Point

st3v0:
Observable point was missing clone and equals methods. Since the ObservablePoint is highly coupled to Point, and Point already had those methods, I've updated it to extend.

### DIFF
--- a/src/core/math/Point.js
+++ b/src/core/math/Point.js
@@ -14,18 +14,16 @@ export default class Point
     constructor(x = 0, y = 0)
     {
         /**
-         * @protected
          * @member {number}
          * @default 0
          */
-        this._x = x;
+        this.x = x;
 
         /**
-         * @protected
          * @member {number}
          * @default 0
          */
-        this._y = y;
+        this.y = y;
     }
 
     /**
@@ -35,7 +33,7 @@ export default class Point
      */
     clone()
     {
-        return new Point(this._x, this._y);
+        return new Point(this.x, this.y);
     }
 
     /**
@@ -45,10 +43,7 @@ export default class Point
      */
     copy(p)
     {
-        if (!this.equals(p))
-        {
-            this.set(p.x, p.y);
-        }
+        this.set(p.x, p.y);
     }
 
     /**
@@ -59,7 +54,7 @@ export default class Point
      */
     equals(p)
     {
-        return (p.x === this._x) && (p.y === this._y);
+        return (p.x === this.x) && (p.y === this.y);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/core/math/Point.js
```diff
@@ -66,8 +71,38 @@ export default class Point
...
+    }
+
+    /**
+     * The position of the point on the x axis
+     *
+     * @member {number}
+     */
+    get x()
+    {
+        return this._x;
```
englercj: This is unfortunate, do we have to change Point to use getters/setters for x/y?

bigtimebuddy:
I don't believe this is right for v4, since we're only accepting PRs for high-value and critical bug fixes. This PR is more of a housekeeping, code maintenance, consistency issue and doesn't fall into one of those two buckets.

Also, overall I'm not sure about this change. On the balance, I'm not sure it provides a ton of value. Since many of the functions are overridden by the ObservablePoint, it probably is easier and less complicated to just add `clone` and `equal` methods to that class if that's the issue. Also, to @englercj's comment, it adds getters/setters to what is otherwise a very simple data structure (Point).

Curious to hear what other people think. Could be convinced that this is for v5, but probably not v4.

themoonrat:
If there is a need to keep the 2 APIs compatible (not something I've required, but I can sorta see someone switching between the 2), I'd vote for adding clone and equal methods to the Point class, without inheritance, for v5 only.

Faster performance, less code (only minor duplication).

st3v0:
Okay I see where you all are coming from. To keep things simple, I've reverted the changes an added only the missing methods. 

There would have been no need for this change had someone not changed a `Point` to an `ObservablePoint` which broke our equal checks.

On a side note; I'm a bit worried that you're locking v4 and focusing on v5, while I would love nothing more then to use the latest and greatest v5 changes, it's not piratical at this time to move to a WebGL only platform. Our analytics show more then 5% of our users are still running with no WebGL support at all. I can't divulge specifics however I can tell you that's from a sample size of 200,000+ users. And 50% of those users only support WebGL1.0. 

From our point of view we won't be moving to v5 until those 10's of thousands of no WebGL users becomes more like hundreds. This may be another year or two. All I'm saying is perhaps v4 should remain a little more open to changes until v5 is ready and commercially viable.

bigtimebuddy:
Part of the problem @st3v0 is that we have extremely limited resources on the team that manages PixiJS. We're trying to spend those limited resources wisely, which is pushing forward to v5 and trying to slow non-critical PRs to v4.

Also, just FYI, v5 still still continue to support canvas fallback in a new package **pixi.js-legacy**. Also, we are progressively enhancing when it comes to WebGL 1 vs 2. 

But I think I would now consider your use-case a bug to fix backward compatibility.

PR #5094, opened 2018-08-30
readme typo fix

nikolas:


bigtimebuddy:
@nikolas could you add this #5095 instead of having a separate PR? Would make it easier on me to merge and make sure changes get pushed to v5 as well.

PR #5109, opened 2018-09-09
fix: call `_onCanPlay()` on nextTick to make `autoPlay=false` working

meowtec:
`autoPlay=false` not working in my code:

```javascript
// video has been ready
const videoBaseTexture = new PIXI.VideoBaseTexture(video)
videoBaseTexture.autoPlay = false
```

Because `_onCanPlay()` was called before `autoPlay` set to `false`.


bigtimebuddy:
Closing in favor of #5128

bigtimebuddy:
Thanks for highlighting this issue @meowtec

PR #5112, opened 2018-09-10
video autoplay from source

djmisterjon:
user should allowed to setup autoplay manually

I don't know what other dev think about this.
It's very problematic on my side.

this allow user to manually setup the source if need **autoplay**
Default it **false**;
```javascript
    const texture = PIXI.Texture.fromVideo('data2/Video/intro/vidA1.webm');
    texture.baseTexture.source.autoplay = true; // working if need autoplay
    const videoSprite = new PIXI.Sprite(texture);
    const videoControler = texture.baseTexture.source;
```

### DIFF
--- a/src/core/textures/VideoBaseTexture.js
+++ b/src/core/textures/VideoBaseTexture.js
@@ -67,7 +67,7 @@ export default class VideoBaseTexture extends BaseTexture
          * @member {boolean}
          * @default true
          */
-        this.autoPlay = true;
+        this.autoPlay = source.autoplay;
 
         this.update = this.update.bind(this);
         this._onCanPlay = this._onCanPlay.bind(this);


englercj:
Can't you do this?

```
const texture = PIXI.Texture.fromVideo('data2/Video/intro/vidA1.webm');
texture.baseTexture.autoplay = false;
const videoSprite = new PIXI.Sprite(texture);
const videoControler = texture.baseTexture.source;
```

If you don't want auto play?

src/core/textures/VideoBaseTexture.js
```diff
@@ -67,7 +67,7 @@ export default class VideoBaseTexture extends BaseTexture
          * @member {boolean}
          * @default true
          */
-        this.autoPlay = true;
+        this.autoPlay = source.autoplay;
```
bigtimebuddy: Substituting `source.autoplay !== false` would be non-breaking (default to `true`).

djmisterjon:
@englercj yep i was try this.
it dont work and the **autoPlay** true will also break the custom currentTime.

```javascript
Scene_IntroVideo.prototype.create_IntroVideo = function() {
    const texture = PIXI.Texture.fromVideo('data2/Video/intro/vidA1.webm');
// dont work , the video autoplay and also break the custom currentTime and start to 0
texture.baseTexture.autoplay = false;
texture.baseTexture.source.pause();
    const videoSprite = new PIXI.Sprite(texture);
    const videoControler = texture.baseTexture.source;

    videoSprite.width = 1920;
    videoSprite.height = 1080;

    videoControler.currentTime = 12.2;
    videoControler.onended = function(){
        this.nextVideo();
    }.bind(this);

    this.addChild(videoSprite);
    this.videoControler = videoControler;
};
```

src/core/textures/VideoBaseTexture.js
djmisterjon: am not understand sorry can you elaborate?
by default the **source.autoplay** it alway false;
the problem is that pixi forces the autoplay, without take the source parameter.
this dont allow to force video to stop.
but if we use the video source, we are allow to force the autoplay if we need.

djmisterjon:
another solution are proposed here but i'm not fan to add a setTimeout for performance reason.
But the guy explain why this happen.
https://github.com/pixijs/pixi.js/pull/5109

src/core/textures/VideoBaseTexture.js
englercj: Could add a new constructor param that defaults to true, but allows you to override this value.

djmisterjon: @englercj yes we can also add a new arg in `fromVideo(video, scaleMode, sourceOptions)`
and also verify argument length in **VideoBaseTexture** to add options for the video source.
Options will be a obj `option:{autoplay:false} `parameter based on the **HTMLMediaElement**.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement

Using the `arguments[]`, it would be possible to avoid any breakChange.

djmisterjon:
yes it would be good to review this source code for v5.
Anyway it is a minor bug, i fix it on my side for my own use, I just wanted to report the anomali to your team, and understand why pixi force restrict the normal operation of the HTMLMediaElement source!

bigtimebuddy:
Closing this in favor of #5128. Thank you @djmisterjon for this contribution.

PR #5120, opened 2018-09-18
LOD Bias support for BaseTexture (fixes #2277)

### LINKED ISSUE #2277: Use LOD Bias parameter in default texture shader to improve mipmapped textures at all scales
When setting `mipmap=true` for a texture, any sprites that use it become noticeably blurry when downsized.  After some investigation into this issue I have determined that this is down to the default WebGL mipmap LOD bias setting, which appears to not always give the best size mipmap at all levels.  There is a very simple fix however.

`texture2D` takes a third parameter which modify's this LOD bias:

For an example see here:

https://www.shadertoy.com/view/4slGWn

In my case simply adding the 3rd parameter to `texture2d` in PIXI's default texture shader and setting it to `-1.0` makes all my sprites perfectly sharp at all levels of zoom/scale. e.g.

 `texture2D(uSampler, vTextureCoord, -1.0)`

The difference in my case is absolutely tremendous.  Prior to doing this I had implemented manual mip-mapping in our application code simply because our textures were so blurry when downsized.  Obviously this was not an ideal solution either performance or quality-wise.  

It would be great to see this as an option in the `webGL` renderer, something like this:

```

var renderer = PIXI.autoDetectRenderer(width, height, {
             transparent : true,
             resolution : pixelRatio,
             preserveDrawingBuffer : true,
             antialias : false,
             forceFXAA : false,
             mipmapLODbias : -1
});
```

If present it could simply be sen into the default shader as a uniform.

For clarity here is the complete modified default shader:

```
TextureShader.defaultFragmentSrc = [
    'precision lowp float;',

    'varying vec2 vTextureCoord;',
    'varying vec4 vColor;',

    'uniform sampler2D uSampler;',

    'void main(void){',
    '   gl_FragColor = texture2D(uSampler, vTextureCoord, -1.0) * vColor ;',
    '}'
].join('\n');

```

Incidentally, as shown in in the shadertoy example above, this is also a very good way to add a fake box blur to any sprite, which is infinitely faster than actually using a custom blurring filter.  Perhaps worth thinking about?

Here is a quick comparison, you will need to open up the actual images to full size in order to see the dramatic difference.  Please note the blurriness of the girl in the first example.  Which is a single sprite with a single texture.

No LOD bias set:
example1

LOD Bias set to -1.0 in default shader:
example2

Also it is worth noting that adding this parameter does not effect non mip-mapped textures in any way, so it should be a very safe change.

Teoxoy:
Fixes #2277

### DIFF
--- a/src/core/sprites/webgl/generateMultiTextureShader.js
+++ b/src/core/sprites/webgl/generateMultiTextureShader.js
@@ -7,7 +7,6 @@ const fragTemplate = [
     'varying vec4 vColor;',
     'varying float vTextureId;',
     'uniform sampler2D uSamplers[%count%];',
-    'uniform float uLODBiases[%count%];',
 
     'void main(void){',
     'vec4 color;',
@@ -36,7 +35,6 @@ export default function generateMultiTextureShader(gl, maxTextures)
 
     shader.bind();
     shader.uniforms.uSamplers = sampleValues;
-    shader.uniforms.uLODBiases = new Float32Array(maxTextures);
 
     return shader;
 }
@@ -61,7 +59,7 @@ function generateSampleSrc(maxTextures)
         }
 
         src += '\n{';
-        src += `\n\tcolor = texture2D(uSamplers[${i}], vTextureCoord, uLODBiases[${i}]);`;
+        src += `\n\tcolor = texture2D(uSamplers[${i}], vTextureCoord, ${settings.MIPMAP_LOD_BIAS});`;
         src += '\n}';
     }
 
--- a/src/core/sprites/webgl/generateMultiTextureShader.js
+++ b/src/core/sprites/webgl/generateMultiTextureShader.js
@@ -1,6 +1,7 @@
 import Shader from '../../Shader';
 import { readFileSync } from 'fs';
 import { join } from 'path';
+import settings from '../../settings';
 
 const fragTemplate = [
     'varying vec2 vTextureCoord;',
--- a/src/core/sprites/webgl/generateMultiTextureShader.js
+++ b/src/core/sprites/webgl/generateMultiTextureShader.js
@@ -36,7 +36,7 @@ export default function generateMultiTextureShader(gl, maxTextures)
 
     shader.bind();
     shader.uniforms.uSamplers = sampleValues;
-    shader.uniforms.uLODBiases = Array(maxTextures).fill(0);
+    shader.uniforms.uLODBiases = new Float32Array(maxTextures);
 
     return shader;
 }
--- a/src/core/sprites/webgl/generateMultiTextureShader.js
+++ b/src/core/sprites/webgl/generateMultiTextureShader.js
@@ -7,6 +7,7 @@ const fragTemplate = [
     'varying vec4 vColor;',
     'varying float vTextureId;',
     'uniform sampler2D uSamplers[%count%];',
+    'uniform float uLODBiases[%count%];',
 
     'void main(void){',
     'vec4 color;',
@@ -35,6 +36,7 @@ export default function generateMultiTextureShader(gl, maxTextures)
 
     shader.bind();
     shader.uniforms.uSamplers = sampleValues;
+    shader.uniforms.uLODBiases = Array(maxTextures).fill(0);
 
     return shader;
 }
@@ -59,7 +61,7 @@ function generateSampleSrc(maxTextures)
         }
 
         src += '\n{';
-        src += `\n\tcolor = texture2D(uSamplers[${i}], vTextureCoord);`;
+        src += `\n\tcolor = texture2D(uSamplers[${i}], vTextureCoord, uLODBiases[${i}]);`;
         src += '\n}';
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "pixi.js",
-  "version": "4.8.1",
+  "version": "4.8.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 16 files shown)


englercj:
Hey @Teoxoy, thanks for looking into this.

Can you please rebase onto the `dev` branch? As mentioned in our [Contributing Guide](https://github.com/pixijs/pixi.js/blob/master/CONTRIBUTING.md) we don't accept PRs to master except in extreme circumstances.

Additionally, can you look into the linting errors CI mentions?

Thanks!

englercj:
@bigtimebuddy manages the releases, but generally they are not released day-one after being merged.

If you need this now, I would recommend using a custom build until it gets released.

src/core/sprites/webgl/generateMultiTextureShader.js
```diff
@@ -35,6 +36,7 @@ export default function generateMultiTextureShader(gl, maxTextures)
 
     shader.bind();
     shader.uniforms.uSamplers = sampleValues;
+    shader.uniforms.uLODBiases = Array(maxTextures).fill(0);
```
englercj: Don't think we can use this function as it isn't supported by all our browsers. Probably easier to create a `new Float32Array(maxTextures)` which will default to zeroes.

Teoxoy: Yes, that would be better.

Teoxoy:
I modified the implementation to use `Float32Array`,
Can this get merged?

bigtimebuddy:
This looks good. The only thing that kinda feels inconsistent is the property name `LODBias`. In general, PixiJS uses capitalization only for constants and class names. But I know this is an abbreviation. Thoughts on `lodBias`? Or something else?

Teoxoy:
I don't mind changing it, but off the top of my head, I remember at least one instance in pixi.js where an abbreviation is being used and is uppercase: `forceFXAA` in the OpenGL renderer settings. To be fair `forceFXAA` has force in front of FXAA so it's not a really a 1:1 comparison.
As I said, I'm neutral on this change.

bigtimebuddy:
biasLOD, lodBias, bias, levelOfDetailBias? Anyone have a strong opinion? What feels the most consistent?

englercj:
I vote for `lodBias`.

themoonrat:
I'd be happy with lodBias too

bigtimebuddy:
I like that too.

Teoxoy:
Should I rename this one too: `this.shader.uniforms.uLODBiases` ?

themoonrat:
I personally think that a `PIXI.settings.MIPMAP_LOD_BIAS` would be great, but I'd wait on a backup opinion on that :) And it wouldn't hold up this PR going in as it is, in any case

bigtimebuddy:
I agree with with the `MIPMAP_LOD_BIAS` setting to set the default.

Teoxoy:
Should I push another commit with that then?

bigtimebuddy:
Yes, please!

Teoxoy:
Done!

bigtimebuddy:
@Teoxoy thanks for all your hard work on this PR! I appreciate how you've taken a lot of feedback and incorporated it very quickly.

Before we merge, I want to have a chat with the Pixi team. Want to make sure there are no performance implications here. Since this affects the rendering of Sprites, it's considered "hot code" and we are very careful about changes to this area.

In addition, we are in the middle of pushing v5 out the door and I want to make sure we have a solution for v5 as well. We don't like introducing new APIs in v4 for non-critical issues because it means we need to also find a solution for v5 as well. This PR fixes a minor rendering improvement and I would consider it non-critical.

Lastly, I realize that setting the default of LOD_BIAS to `-1` gives a slightly better result, but it is a breaking change, albeit a small one. Using a default of `0` would probably be better. Using mipmaps is itself a performance optimization on the GPU and using a default of `-1` seems a little overly aggressive for all users, for all Sprites.

Teoxoy:
I can help with the v5 implementation as well. I wanted to switch to v5 as soon as it became stable anyway.
As for the `MIPMAP_LOD_BIAS` default value, I will set it to 0.

bigtimebuddy:
@Teoxoy, Sorry it took so long to get back to you on this. 

I had a chat with @GoodBoyDigital about it and I think we have a good path forward. Our concern was that we felt like there was a simpler solution to this edge-case that doesn't involve additional overhead of uploading a new Float32Array to the shader. 

Our preference would be to remove the `lodBias` API from BaseTexture and from the uniforms and just leave in the global setting which is added directly to the shader string. This would be a much easier thing to port to v5 and it's a lot less complicated or likely to have performance implications. This would probably also solve the issue in an easier way.

```js
src += `\n\tcolor = texture2D(uSamplers[${i}], vTextureCoord, ${settings.MIPMAP_LOD_BIAS});`;
```

In v5, you'll be able to create your own shader and if someone wanted to create a per-BaseTexture API for adjusting the LOD bias, they could do that more easily.

Thoughts? Do you think you could simplify this? 

Thanks again for your patience.

Teoxoy:
No problem.
I also think it's better to just use `settings.MIPMAP_LOD_BIAS` in the shader.
Should I go ahead and implement it this way?

bigtimebuddy:
Please! Thanks for being flexible about this.

bigtimebuddy:
Just the linting problem: https://travis-ci.org/pixijs/pixi.js/builds/450977828#L630
Looks like you need to import settings.

bigtimebuddy:
@Teoxoy it seems that this PR is not playing well with filters. We are getting reports of crashing from folks because it's unable to compile shader. I unfortunately need to revert because it's a critical bug in **dev**. Could you dig into why this is happening and resubmit a PR?

PR #5122, opened 2018-09-20
Fix cache as bitmap filters check

NoxWings:
In our internal PIXI-wrapped engine we ended up setting an empty array on the filters property of some cached elements and I found this.

Not that I consider this is a "bug", but adding this won't harm.

Edit: I couldn't find any associated unit tests. Am I missing something?


bigtimebuddy:
Thanks @NoxWings 👍

PR #5129, opened 2018-09-23
Added 'angle' property for DisplayObject

themoonrat:
The proposed solution from https://github.com/pixijs/pixi.js/pull/4579 which got recently closed.

I've had `angle` as a property like this on my version of PIXI for a while, and it makes it so much easier to configure a display object's rotation via JSON :)


bigtimebuddy:
Might be nice to add a simple unit test or two. To validate the API

themoonrat:
@bigtimebuddy done!

PR #5144, opened 2018-10-02
[Next] fix cubemap

ivanpopelyshev:
Cubemap resource wasn't properly tested. @GoodBoyDigital  knows how to test it.


ivanpopelyshev:
One of the problem is that we have only one glTexture for cubeMap, and we dont store width and height for every side of cube, that's why we cant do that texSubImage2D optimization.

bigtimebuddy:
Thanks @ivanpopelyshev.

@GoodBoyDigital said that he need to verify but has been time-limited. If you could provide an example, that would definitely help speed-up the process of getting this in.

PR #5150, opened 2018-10-10
render children first

Bastl34:
added possibility to render all children underneath the parent element

**Example:**
```
let pixiItem = new Sprite(tex);
let pixiItem2 = new Sprite(tex);
let pixiItem3 = new Sprite(tex);
let pixiItem4 = new Sprite(tex);

pixiItem.renderChildrenFirst = true;
pixiItem2.renderChildrenFirst = true;
pixiItem3.renderChildrenFirst = true;
pixiItem4.renderChildrenFirst = true;

stage.addChild(pixiItem);
pixiItem.addChild(pixiItem2);
pixiItem2.addChild(pixiItem3);
pixiItem3.addChild(pixiItem4);
```

**Result:**

`renderChildrenFirst = false`:
from

`renderChildrenFirst = true`:
to

### DIFF
--- a/src/core/display/Container.js
+++ b/src/core/display/Container.js
@@ -387,6 +387,31 @@ export default class Container extends DisplayObject
     }
 
     /**
+     * Renders all the object based on renderChildrenFirst
+     *
+     * @private
+     * @param {PIXI.WebGLRenderer} renderer - The renderer
+     */
+    _renderObjects(renderer)
+    {
+        if (!this.renderChildrenFirst)
+        {
+            this._renderWebGL(renderer);
+        }
+
+        // now loop through the children and make sure they get rendered
+        for (let i = 0, j = this.children.length; i < j; i++)
+        {
+            this.children[i].renderWebGL(renderer);
+        }
+
+        if (this.renderChildrenFirst)
+        {
+            this._renderWebGL(renderer);
+        }
+    }
+
+    /**
      * Renders the object using the WebGL renderer
      *
      * @param {PIXI.WebGLRenderer} renderer - The renderer
@@ -406,21 +431,7 @@ export default class Container extends DisplayObject
         }
         else
         {
-            if (!this.renderChildrenFirst)
-            {
-                this._renderWebGL(renderer);
-            }
-
-            // simple render children!
-            for (let i = 0, j = this.children.length; i < j; ++i)
-            {
-                this.children[i].renderWebGL(renderer);
-            }
-
-            if (this.renderChildrenFirst)
-            {
-                this._renderWebGL(renderer);
-            }
+            this._renderObjects(renderer);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/core/display/Container.js
```diff
@@ -459,14 +467,22 @@ export default class Container extends DisplayObject
         }
 
         // add this object to the batch, only rendered if it has a texture.
-        this._renderWebGL(renderer);
```
ergoposti: Since it's exactly the same code as above, then maybe separate into private function that could be reused in both places?

Bastl34: sure. I will change this

src/core/display/Container.js
```diff
@@ -386,6 +386,31 @@ export default class Container extends DisplayObject
         // FILL IN//
     }
 
+    /**
+     * Renders all the object based on renderChildrenFirst
+     *
+     * @private
+     * @param {PIXI.WebGLRenderer} renderer - The renderer
+     */
```
ergoposti: Not sure about the name of the method. Maybe something like "_renderWebGLObjects"? Since it only deals with WebGL case.  Any ideas?

Bastl34: You are right. This name is better

englercj:
I think you might be better off doing this by having a separate container that overrides renderWebGL to render children first.

In the past, we have avoided extra branches in this function because they have impacted performance in noticeable ways.

Bastl34:
> In the past, we have avoided extra branches in this function because they have impacted performance in noticeable ways.

Hm... I do not think that this would have a impact on the performance. It just checks if this flag is set or not.

bigtimebuddy:
Hi @Bastl34. Thank you for taking the time to submit a pull-request. We love to see first-time contributors!

Unfortunately, I'm not quite sure I understand the need for this PR. Adding children to a Sprite generally feels like an anti-pattern to me. Plus, there are existing ways to workaround your depth-sorting issue: either extend Sprite to create your own custom rendering or use generic Container objects to manage your nesting. Adding a verbosely named API flag is trying to solve something very specific and also touches code that we'd consider "hot" (which is the most performant in Pixi). Adding a few `if` statements feels trivial, but not when you're rendering thousands of Sprites. 

IMO, this feels more like a plugin or project-specific solution than a core library change.  In addition, I would consider this to be a non-critical change, so only for v5 (**next** branch).

Other thoughts from the team?

englercj:
> Hm... I do not think that this would have a impact on the performance. It just checks if this flag is set or not.

The specific case I'm talking about previously was adding `if (this.parent)`. Having the check meant a ~5-10% increase in update times. Consistently. Hopefully things have changed since then, just wanted to give the context.

bigtimebuddy:
Had a chat about this with @GoodBoyDigital and we both agreed that this PR should be closed. If one is willing to make these changes to hot-code, then they should consider using Containers instead. This approach does not involve any changes to core PixiJS and is a stable, well-tested alternative. For instance:

```js
const bone1 = new Container();
const bone2 = new Container();
const bone3 = new Container();
const bone4 = new Container();

bone1.addChild(new Sprite(tex));
bone2.addChild(new Sprite(tex));
bone3.addChild(new Sprite(tex));
bone4.addChild(new Sprite(tex));

stage.addChild(bone1);
bone1.addChildAt(bone2, 0);
bone2.addChildAt(bone3, 0);
bone3.addChildAt(bone4, 0);
```

Adding new APIs is always a trade-off between adding high-value convenience versus maintainability. Because this flag only affects rendering, it's difficult to unit-test. Also, it introduces an alternative way for changing rendering order. There are other plugins like pixi-layers which also override the default rendering order. Not sure how this flag would play with those other approaches.

Lastly, while Sprite does extends Container, it's definitely not typical to nest Sprite objects. This inheritance model was inspired by Flash/AS3, which is to say, that it's not necessarily the best. We had  considered making Sprite extend DisplayObject, which is probably closer to how it's intended to be used.

I appreciate the work you put into this PR and I hope you'll consider making PRs in the future. If you aren't already, please join the PixiJS Slack (send me your email address and I'll add you), that's a great way to get feedback on new features like this.

PR #5156, opened 2018-10-13
[Next] Allows more texture units to be used by modern Android and Apple devices

themoonrat:
Also defaults precision fragment on ios to 'highp' - to stop issues of wobbling sprites by default like with https://github.com/pixijs/pixi.js/issues/3742
canUploadSameBuffer to not use it's own ios detection code, and use 'ismobilejs'


themoonrat:
@GoodBoyDigital PR as requested. Didn't do as part of 'legacy' because legacy mode limits MAX_TEXTURES to 1 in SpriteRenderer anyway.

PR #5161, opened 2018-10-15
fix a bug

wswei99:
please check this...

### DIFF
--- a/src/core/ticker/TickerListener.js
+++ b/src/core/ticker/TickerListener.js
@@ -121,8 +121,8 @@ export default class TickerListener
         if (previous.next)
         {
             previous.next.previous = this;
+            this.next = previous.next;
         }
-        this.next = previous.next;
         previous.next = this;
     }
 


themoonrat:
Could you supply a bit more detail about the issue and the fix please?

bigtimebuddy:
@wswei99: In general, the responsibility to test and verify changes is on the person submitting the pull-request because we are all volunteers working on this project. This change requires a unit-test, at least, to be added to make sure it addresses some underlying issue. Also, it's very important to give some more context for what this change accomplishes in the description of this pull-request.

wswei99:
/**
 * Connect to the list.
 * @param {TickerListener} previous - Input node, previous listener
 */
connect(previous)
{
    this.previous = previous;
    /** 
    * If the last node is not a tail node, the current operation is an insert.
    * The order of elimination is '... <-> previous <-> this <-> previous.next <-> ...'
    * so 'previous.next.previous = this;' and 'this.next = previous.next;' Is the logic code for the insertion operation, should be written in the 'if' statement.
    */
    if (previous.next)
    {
        previous.next.previous = this;
    }
    this.next = previous.next;
    previous.next = this;
}

bigtimebuddy:
Can you please provide and example or a jsfiddle that reproduces this problem?

bigtimebuddy:
You’re just explaining the code change, but what is the effect? What is the bug or crash?

bigtimebuddy:
@wswei99 Would you mind showing an example or JSFiddle of this breaking? Would help to verify this change. Thank you!

bigtimebuddy:
@wswei99 I'm going to close this PR because of insufficient evidence showing this fix resolves a real bug. If you can demonstrate a bug or make a unit-test that shows the problem, we will reconsider this.

Issue #5191, opened 2018-10-24
Bolder text is more higher


liajoy:
### Expected Behavior
Maybe bolder text should be as high as the normal text, just like in canvas do?
see: https://jsfiddle.net/1bvyrck6/2/
### Current Behavior
Bolder text is higher than the normal one.  
see: https://jsfiddle.net/1bvyrck6/3/

### Environment

- `pixi.js` version:  4.8.2
- Browser: Chrome 69

liajoy:
Seems the reason is the letter 'É'.
Bolder É is not as high as normal even in canvas. So measureFont function return a larger fontSize.
Is there any solution?

themoonrat:
Could changing the text metrics settings `TextMetrics.METRICS_STRING`, `TextMetrics.BASELINE_SYMBOL = 'M';` and `TextMetrics.BASELINE_MULTIPLIER = 1.4` help?
http://pixijs.download/dev/docs/core_text_TextMetrics.js.html#line674

liajoy:
@themoonrat 
I'll appreciate it very much if you could tell me why letter 'M' could be the baseline symbol.

themoonrat:
Unfortunately, the canvas api only lets us measure the width of text drawn, not the height.
So the baseline symbol and multiplier are basically trying to use a wide symbol, and use that as a guesstimate as to how tall text might be

liajoy:
thanks a lot!
I am also very curious about ascent calculation. Seems Pixi scans from top to bottom until find a non red pixel, then the diff of line index and baseline will be the ascent. But sometimes ascent isn't the top of some letter, maybe the ascent line is higher than the top of letter. How does Pixi's text position almost equal to text dom under the most conditions?

liajoy:
thanks a lot! 
Could you please take a quick look on the pr?

PR #5193, opened 2018-10-24
Fix fontSize larger because of fontWeight

liajoy:
#5191 
If calculate text ascent & descent with fontWeight, the ascent & descent unequal with normal one.
So the properties.fontSize will be larger.

### DIFF
--- a/src/core/const.js
+++ b/src/core/const.js
@@ -336,3 +336,15 @@ export const UPDATE_PRIORITY = {
     LOW: -25,
     UTILITY: -50,
 };
+
+/**
+ * Regexp for fontWeight.
+ *
+ * @static
+ * @constant
+ * @name FONT_WEIGHT
+ * @memberof PIXI
+ * @type {RegExp|string}
+ * @example bold;
+ */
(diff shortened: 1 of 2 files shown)


themoonrat:
So had quick look!
My immediate worry is that this could subtlety change how text is aligned for existing content in v4, so my initial thought would be for v5 only.

My second thought is that this then leads down a bit of a rabbit hole. Because what about 2 bits of text where one has italics and one doesn't? Or one has smallcaps and the other doesn't? The reality is, that the parts that make up the fontString
`${this.fontStyle} ${this.fontVariant} ${this.fontWeight} ${fontSizeString} ${fontFamilies.join(',')}`
All do actually change the size of the font that is rendered. So to then not take into account measuring just on aspect? Doesn't seem right to me.

Finally, you can achieve the look you are going for via using the `trim` property on text style: https://jsfiddle.net/5tnd7v10/

So I'd either go down the trim route, or override PIXI's TextMetrics.measureFont function to eliminate parts of the font string you personally don't want to worry about
(and if you do that, for what it's worth, I'd manipulate the font string as the very first thing you do, so that the caching here works independent of 'boldness')
```
        if (TextMetrics._fonts[font])
        {
            return TextMetrics._fonts[font];
        }
```


Second: I'm not sure it should work

PR #5196, opened 2018-10-26
Allow to re-use existing SVG images

shlajin:
The problem: `BaseTexture#_updateImageType` will refuse to do anything if `BaseTexture#imageUrl` is null. This will lead to SVG be loaded as a non-SVG source:
```
            if (this.imageType === 'svg') { // imageType has not been set, so it's not svg
                this._loadSvgSource();
            } else {
                this._sourceLoaded(); // called for svg file - bummer!
            }
```

This won't happen in normal use cases when `BaseTexture` is created with `BaseTexture.from`. However, this will happen if `BaseTexture` is created with `new`, e.g.

```
// `image` is an already loaded `Image` instance
const baseTexture = new PIXI.BaseTexture(image);
```

Solution: make sure to at least try to set `source` before guessing type.

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -324,6 +324,11 @@ export default class BaseTexture extends EventEmitter
         // Apply source if loaded. Otherwise setup appropriate loading monitors.
         if (((source.src && source.complete) || source.getContext) && source.width && source.height)
         {
+            if (!this.imageUrl)
+            {
+                this.imageUrl = source.src;
+            }
+
             this._updateImageType();
 
             if (this.imageType === 'svg')


bigtimebuddy:
Could you please provide a unit-test(s) for this code change that breaks in **dev** but works in your branch? Thank you for the PR.

shlajin:
Yeah, I'm on it, but suddenly ran short on time... I'll fix / add specs soon

bigtimebuddy:
Have you had a chance to look at this again?

shlajin:
Sorry for severe delays :(
Concerning the bug, I'm not sure if I can fix the failing test on my own... the fail may lay deeper.
I believe this happening due to this:
Spritesheet's constructor has these lines:
```
        this.resolution = this._updateResolution(
            resolutionFilename || this.baseTexture.imageUrl // <--- this
        );
```

with the provided fix, the `baseTexture.imageUrl` gets set and without it it results to `null`. This affects `this.resolution` in this way:
- Without the fix it is `1`.
- With the fix, it is `0.5`

I don't think the correct value is `0.5` (because why would it be), but I'm not sure if I should change the Spritesheet for this fix... any thoughts?

bigtimebuddy:
Could you please clarify/show that problem you're trying to solve? I'm not sure I understand what you're trying to fix with this PR.

https://pixiplayground.com/#/edit/QdnV3MkJbqW9I7Zyb_EKI

bigtimebuddy:
Thanks @shlajin for the PR but I don’t think this is ready quite yet. If and when you have the time to look at it again and address the unit test issues as well as provide a functional example of the underlying issue, then feel free to resubmit a new PR. Closing this for now.

PR #5200, opened 2018-10-30
Add support for absolute url on bitmap text file path

EloB:
Right now it always concatenate xml url with file url even if the file url is an absolute url.

More info why this could be good to have :)
https://github.com/pixijs/pixi.js/issues/5199

### DIFF
--- a/src/loaders/bitmapFontParser.js
+++ b/src/loaders/bitmapFontParser.js
@@ -2,6 +2,11 @@ import * as path from 'path';
 import { Resource } from 'resource-loader';
 import { BitmapText } from '../extras';
 
+function isAbsoluteUrl(url)
+{
+    return (/^[a-z][a-z0-9+.-]*:/).test(url);
+}
+
 /**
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@EloB, thanks for the contribution! Love to see first-time contributors.

This change definitely requires unit-tests and should consider the following protocols: `http://`, `https://`, `file://`, as well as relative paths (e.g. `../foo/bar`) and root-relative `/foo/bar`) paths. I don't think your RegExp is going to be adequate enough to handle these cases. You might want to double check to make sure that we aren't already include a dependency which can check for this or if PixiJS internally has a similar function (I'm not sure off the top of my head).

bigtimebuddy:
Do you have time to update this PR with tests and a more robust regexp?

bigtimebuddy:
After no response for 4 weeks, I'm going to close this PR. Insufficient tests and consideration of all edge case to be merged. If someone wants to take another stab at this PR, welcomed.

PR #5204, opened 2018-11-01
[Next] zIndex support take 2

themoonrat:
A follow up PR to https://github.com/pixijs/pixi.js/pull/5031

Of the 3 main comments before

- If it's possible to auto-detect without too much overhead, then that's a better approach.

There is both a global PIXI.settings.ZINDEX_AUTO_SORT that can disable the behaviour across all containers, and also a container by container zIndexAutoSort setting (that defaults to the above).

I've also fixed the sorting so that it only occurs before rendering, the flag to have it occur before rendering only gets set on children being added or a child zIndex being changed, and to not perform any sort if all children have a zIndex of 0.

- I think the arrival order has a lot of design advantages over the DisplayObject creation order, as you've implemented it. Basically, reverts back to the original children's order. If any child has non-zero zIndex, will sort.

I've had a deep look into this, and arrival order has all sorts of problems when you think about `addChildAt` or `swapChildren`. Use arrival order, then use the above functions, and any sorting the children would change their order even if they had a matching zIndex. Using the previous array order as the decider keeps the above functions working as intended. 

- We'll need some very good test coverage on this feature, btw. There are a lot of test cases to cover like adding children before and after changing zIndex, reverting all zIndex to 0, etc.

I've added a bunch of unit tests to cover all the above scenarios

Finally - regarding what sort algorithm to use... I did a bunch of benchmarking comparing the in-built `sort` compared to a couple of algorithms Ivan posted at https://github.com/pixijs/pixi-display/wiki 
I completed on all of the latest browsers, comparing 100, 1k, 10k and 100k children.
(this laptop is a dual core i7-5600U)

The first table is comparing the sort times when changing the middle value zIndex
image

The second table compares the sort times when the zIndex is changed on all children in a random (but consistent across runs) way.
image

I feel that it's well worth just sticking to using the native sort, unless there an any objections.

### DIFF
--- a/packages/canvas/canvas-display/src/Container.js
+++ b/packages/canvas/canvas-display/src/Container.js
@@ -18,6 +18,12 @@ Container.prototype._renderCanvas = function _renderCanvas(renderer) // eslint-d
  */
 Container.prototype.renderCanvas = function renderCanvas(renderer)
 {
+    if (this.sortChildrenNextRender)
+    {
+        this.sortChildren();
+        this.sortChildrenNextRender = false;
+    }
+
     // if not visible or the alpha is 0 then no need to render this
     if (!this.visible || this.worldAlpha <= 0 || !this.renderable)
     {
--- a/packages/canvas/canvas-display/src/Container.js
+++ b/packages/canvas/canvas-display/src/Container.js
@@ -18,11 +18,6 @@ Container.prototype._renderCanvas = function _renderCanvas(renderer) // eslint-d
  */
 Container.prototype.renderCanvas = function renderCanvas(renderer)
 {
-    if (this.sortDirty && this.zIndexAutoSort)
-    {
-        this.sortChildren();
-    }
-
     // if not visible or the alpha is 0 then no need to render this
     if (!this.visible || this.worldAlpha <= 0 || !this.renderable)
     {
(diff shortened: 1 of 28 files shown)


bigtimebuddy:
Great work. This is looking promising.

What do you think about “sortDirty” instead of “sortChildrenNextRender”. It’s a little more consistent with other dirty-like flags internally and more concise.

themoonrat:
yeps, makes sense

ivanpopelyshev:
Good job! 
Give me the benchmark please :)

ivanpopelyshev:
addChild doesn't set "sortDirty" flag even if added child has a zIndex.

themoonrat:
>           addChild doesn't set "sortDirty" flag even if added child has a zIndex.

It does! Line 103 :)

ivanpopelyshev:
Please move it to `updateTransform()`, I'll approve it

bigtimebuddy:
Awesome work @themoonrat.

bigtimebuddy:
This feels like the right API choices to me:
Enable automatic sorting:
```js
container.sortableChildren = true;
```
Do a manual sorting of children:
```js
container.sortChildren();
```
Global automatic sorting (apply to all Containers):
```js
PIXI.settings.SORTABLE_CHILDREN = true;
```

themoonrat:
And since the names are generic, easy for devs to come in, override the sortChildren function, then set the sortDirty flag on containers in their own way to sort children however they want. It's much better than where the review started, that's for sure. Cheers for the help

PR #5206, opened 2018-11-06
[Next] Graphics does not depend on Mesh

ivanpopelyshev:
The graphics-mesh compatibility exists on Geometry level. 

`state` and `shader` are the only things Graphics took from Mesh.


packages/graphics/src/Graphics.js
```diff
@@ -36,27 +39,44 @@ let defaultShader = null;
  * @extends PIXI.Mesh
```
bigtimebuddy: Could you update this doc?

bigtimebuddy: Fixed.

PR #5207, opened 2018-11-06
[next] Graphics LineStyle refactor

ivanpopelyshev:
I don't like to spawn many objects without prototype.

What if we reduce the number a bit?


ivanpopelyshev:
Are you sure that your method `toJSON` has to go?

bigtimebuddy:
The toJSON method isn’t needed. Clone is better for the reasons you point out in maintaining the prototype. Makes it easier to debug, mixin, profile

englercj:
`Rectangle` uses `copyTo` and `copyFrom`. I think `copyTo` was the correct API choice.

ivanpopelyshev:
Oh, then we need to refactor MAtrix too :) ok, moving it back.

bigtimebuddy:
@ivanpopelyshev, could you please remove changes to package-lock.json

PR #5209, opened 2018-11-07
[next] webgl1 and wrapMode

ivanpopelyshev:
Setting incorrect wrapMode of texture in webgl1 should not lead to error.

Example:
https://www.pixiplayground.com/#/edit/X7GwGFqITM40hRjBd2cID

Set "PREFER_ENV=2" in first line , see many bunnies. 

Set the build `https://pixijs.download/next-fix-webgl1-wrap/pixi.js`, see one bunny.

### DIFF
--- a/packages/core/src/textures/GLTexture.js
+++ b/packages/core/src/textures/GLTexture.js
@@ -20,14 +20,20 @@ export default class GLTexture
 
         /**
          * Texture style dirty flag
-         * @type {number}
+         * @member {number}
          */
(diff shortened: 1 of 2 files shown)


englercj:
One minor trailing space error, but otherwise lg2m

ivanpopelyshev:
Something strange is going on. Trailing space was fixed automatically but only after the commit. I guess our commit hooks are wrong.

bigtimebuddy:
Try running `npm run lint -- --fix`

PR #5213, opened 2018-11-10
[Next] default sprite anchor

themoonrat:
It's such a common requirement to easily have all Sprites be anchored from the center point... here's a setting to make that possible

### DIFF
--- a/packages/core/src/textures/Texture.js
+++ b/packages/core/src/textures/Texture.js
@@ -159,10 +159,10 @@ export default class Texture extends EventEmitter
         /**
          * Anchor point that is used as default if sprite is created with this texture.
          * Changing the `defaultAnchor` at a later point of time will not update Sprite's anchor point.
-         * @member {PIXI.Point}
-         * @default {0,0}
+         * @member {?PIXI.Point}
+         * @default null
          */
-        this.defaultAnchor = anchor ? new Point(anchor.x, anchor.y) : new Point(0, 0);
+        this.defaultAnchor = anchor ? new Point(anchor.x, anchor.y) : null;
 
         /**
          * Update ID is observed by sprites and TextureMatrix instances.
--- a/packages/sprite/src/Sprite.js
+++ b/packages/sprite/src/Sprite.js
@@ -45,7 +45,7 @@ export default class Sprite extends Container
         /**
          * The anchor sets the origin point of the texture.
          * The default is either taken from the {@link PIXI.Texture#defaultAnchor|Texture}
-         * passed to the constructor, or from {@link PIXI.settings.DEFAULT_SPRITE_ANCHOR},
+         * passed to the constructor, or from {@link PIXI.settings.SPRITE_DEFAULT_ANCHOR},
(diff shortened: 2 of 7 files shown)


themoonrat:
https://www.pixiplayground.com/#/edit/LtHNHRgXmyrTnxqZvwA2y

packages/core/src/textures/Texture.js
```diff
@@ -159,10 +159,10 @@ export default class Texture extends EventEmitter
         /**
          * Anchor point that is used as default if sprite is created with this texture.
          * Changing the `defaultAnchor` at a later point of time will not update Sprite's anchor point.
-         * @member {PIXI.Point}
-         * @default {0,0}
+         * @member {?PIXI.Point}
+         * @default null
          */
-        this.defaultAnchor = anchor ? new Point(anchor.x, anchor.y) : new Point(0, 0);
+        this.defaultAnchor = anchor ? new Point(anchor.x, anchor.y) : null;
```
bigtimebuddy: Why not just set the second part of the ternary to `setting.SPRITE_DEFAULT_ANCHOR` and call it a day? (maybe `TEXTURE_DEFAULT_ANCHOR` is more appropriate a name too) In general, I think it's better practice to always set defaults than to set to `null` and have to check later.

themoonrat:
I just felt it's weird how the Texture defines a Sprite property. I know it was done for TexturePacker purposes, and the only data they could influence is on loading the textures.

I'm happy to go down the TEXTURE_DEFAULT_ANCHOR route tho, since we have something already in place. Will make the change in a bit.

bigtimebuddy:
Wanna make this change?

themoonrat:
So, I've got it pretty much done, but it really doesn't feel right to me. You see, if I create a Sprite with no texture but then assign one, it would end up with a different visual than if I created a Sprite with a texture, because of how the texture defaultAnchor works and when it is picked up. That's not right surely?

And reusable sprites with different textures become an odd situation too.

But we can't make the texture setter also adjust the anchor.

So that's why I'm in limbo on this pr, and I've not thought of a good way out of it yet.

bigtimebuddy:
The `defaultAnchor` was a property added to support a TexturePacker feature. And I think you're wanting to use it more generally is creating some conflicts. It was really only meant to be added to the constructor upon loading Textures. So it wasn't applied in situations where you reassign Textures to Sprites. I think we specifically avoided that use for keeping the PR simple and limited to the TexturePacker feature.

The idea that a Sprite would get its anchor point from a Texture is weird. If you want to set the default anchor for Sprites to be used globally I don't think you should use `defaultAnchor` at all.

themoonrat:
Which, aside from renaming the texture `defaultAnchor` property, leads us back to what I have right now, does it not?

themoonrat:
Sure. I think the difference with a default anchor is that there is a strong use case in projects for all sprites to be anchored centrally, whereas you couldn't say that about any other display property.

But I'll close away, it's fine :)

PR #5215, opened 2018-11-12
keep a origin values.

djmisterjon:
I add a simple cache values to **observablePoint** in the v4 to manage more effectively my Tweens and my TimeLines, and give opportunity to trace the original values in time.
I do not know if the subject has already been addressed, so I want to propose the idea here.

The idea is to be able to store and keep a origin value.
Maybe also need to add it to constructor?

### DIFF
--- a/packages/math/src/ObservablePoint.js
+++ b/packages/math/src/ObservablePoint.js
@@ -79,6 +79,20 @@ export default class ObservablePoint
     }
 
     /**
+     * retore to default cache values
+     *
+     */
+    restore()
+    {
+        if (this._x !== this._xx || this._y !== this._yy)
+        {
+            this._x = this._xx;
+            this._y = this._yy;
+            this.cb.call(this.scope);
+        }
+    }
+
+    /**
      * Copies x and y from the given point
      *
      * @param {PIXI.Point} p - The point to copy from.
--- a/packages/math/src/ObservablePoint.js
+++ b/packages/math/src/ObservablePoint.js
@@ -18,6 +18,8 @@ export default class ObservablePoint
     {
         this._x = x;
         this._y = y;
+        this._xx = x;
+        this._yy = y;
 
         this.cb = cb;
         this.scope = scope;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks for the PR.

I strongly think this doesn't belong in core PixiJS. It seems more like application-specific logic. You should mixin this functionality or replace your project's ObservablePoint--which is sounds like what you are already doing. In addition, it diverges from Point by introducing new APIs, which is problematic. We've tried to keep these classes pretty much consistent (except for the callback).
 
Interested to see what other say, but I'm convinced this is generalized enough.

Also, any PR like this requires unit-tests (there are none here), good documentation (should explain _why_ you'd want to use an API and not just _what_ that API does) and it should pass Travis (there are a bunch of linting errors: https://travis-ci.org/pixijs/pixi.js/builds/453897335#L658-L665).

englercj:
I think a better strategy than this would be allow our math library to be plugable.

If we had a set of interfaces we expected for a math library, and let the user specify their own to use instead of ours it would be easy for people to enable these kinds of things when it makes sense for them.

bigtimebuddy:
That's not a bad idea. It would also solve issues with people wanting to add a bunch of math functionality to Point or Rectangle, which pops up occasionally. I'm definitely in favor of keeping these classes as small as possible, but having a paved road for swapping in your own functionality might help enable some good 3rd party plugins.

@englercj do you have a suggestion of what that API might look like?

RichardGale:
I wouldn't recommend going down the pluggable math route, typically you'll spend a lot of resources for little gain. Math libraries need to be small, focussed and fast.

On degrees and radians, pick one, use it everywhere. So far I've only seen radians, that's all I should need to see.

bigtimebuddy:
After considering this PR, I’m going to close. This is not generalized enough to add to core and there are viable plugin alternatives to getting it in your app, either by mixin props/methods.

djmisterjon:
sans-titre-1

djmisterjon:
a lot more context where need this, sorry if dont express myself good  :)

PR #5225, opened 2018-11-19
UpdateTexture should not switch FrameBuffer

ivanpopelyshev:
Bug was reported in RpkMaker MV year ago , fix appeared in [corescript](https://github.com/rpgtkoolmv/corescript/pull/194), then I found actual cause in [pixi-tilemap](https://github.com/pixijs/pixi-tilemap/commit/1053aab63a7a30004ca7ab249fafb31b8cc0dfe3)

https://www.pixiplayground.com/#/edit/j38lu2giZW7uyEBkI0Eab

### DIFF
--- a/src/core/renderers/webgl/utils/RenderTarget.js
+++ b/src/core/renderers/webgl/utils/RenderTarget.js
@@ -138,7 +138,7 @@ export default class RenderTarget
          *
          * @member {boolean}
          */
-        this.root = root;
+        this.root = root || false;
 
         if (!this.root)
         {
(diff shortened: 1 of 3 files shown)


src/core/renderers/webgl/utils/RenderTarget.js
```diff
@@ -138,7 +138,7 @@ export default class RenderTarget
          *
          * @member {boolean}
```
bigtimebuddy: Is it possible to add `@default false` to the `root` doc? Not critical, but helpful to be more explicit.

ivanpopelyshev: Its already `false` in constructor docs, but i added it anyway.

PR #5227, opened 2018-11-20
If canvas has been resized , call `baseTexture.update`

finscn:
I use 
```
var baseTexture = PIXI.BaseTexture.from(canvas);
var texture = new PIXI.Texture(baseTexture);
...
```

But the canvas would be changed , content or size.
If size changed , need to call `baseTexture.update`, otherwise will take some errors like this:
```
Texture.js:572 Uncaught Error: Texture Error: frame does not fit inside the base Texture dimensions: X: 0 + 700 = 700 > 600 and Y: 0 + 880 = 880 > 350
    at Texture.set (Texture.js:572)
    at new Texture (Texture.js:135)
```

### DIFF
--- a/src/core/textures/BaseTexture.js
+++ b/src/core/textures/BaseTexture.js
@@ -730,7 +730,7 @@ export default class BaseTexture extends EventEmitter
             baseTexture = new BaseTexture(canvas, scaleMode);
             BaseTexture.addToCache(baseTexture, canvas._pixiId);
         }
-        else if (canvas.width !== baseTexture.realWidth || canvas.height !== baseTexture.realHeight)
+        else if (baseTexture.realWidth !== canvas.width || baseTexture.realHeight !== canvas.height)
         {
             baseTexture.update();
         }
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
@bigtimebuddy what if we add one more flag here, to force update?

bigtimebuddy:
@ivanpopelyshev I disagree with adding a new argument that adds convenience for forcing update `update()`. My suggestion is that we return `this` in `update()` so you can do something like this:

```js
const baseTexture = PIXI.BaseTexture.fromCanvas(canvas).update();
```

ivanpopelyshev:
That'll be double update in case size was changed.

bigtimebuddy:
This catches the edge-case where you use a canvas to create a BaseTexture after it's already been created at another dimension. 

After consideration, I think I have to disagree with the premise of this PR.  It feels a little like protecting a developer from themselves. It's not Pixi's responsibility to handle external changes made to canvas (or image) elements and then update internal Textures or BaseTextures. I believe if you change a canvas externally its your responsibility to tell PixiJS to update that resource. Also, this PR does not provide much convenience since it's easily remedied by this one-liner: `baseTexture.update()`.

In summary, minimal convenience with a small edge-case and promotes bad-behavior by partially handling changes in external resources.

Thoughts? @themoonrat @andrewstart @GoodBoyDigital?

bigtimebuddy:
I had a talked with @GoodBoyDigital today about this. I think we are going to pass on this PR. Mat had a good point which is that the `from*` methods are for creating, caching and retrieving objects, not for updating.

PR #5242, opened 2018-11-23
[next] Rounded rect should be drawn clockwise, like other shapes.

ivanpopelyshev:
#5241


bigtimebuddy:
Going to close this as the change seems unnecessary. Unable to reproduce this bug in v5 (next) since our big Graphics overhaul. https://jsfiddle.net/fb2gk9ad/

PR #5243, opened 2018-11-23
Fix TilingSprite:_renderCanvas to account rotated textures from texture atlas

a-rasinski:
Hello,
I've found a bug in TilingSprite implementation for canvas renderer. 
In current implementation it doesn't include the necessary transformation for textures rotated in texture atlases, thus rendering them incorrectly. I've tested this with atlases generated by TexturePacker for pixijs.

I hope the description is enough, if you have any questions feel free to ask.

This is my proposal of a fix for this issue.

PS This is my first contribution to pixi.js and I hope that I did everything according to contribution guide.

### DIFF
--- a/src/extras/TilingSprite.js
+++ b/src/extras/TilingSprite.js
@@ -179,18 +179,21 @@ export default class TilingSprite extends core.Sprite
         const context = renderer.context;
         const transform = this.worldTransform;
         const resolution = renderer.resolution;
+        const isTextureRotated = texture.rotate === 2;
         const baseTexture = texture.baseTexture;
         const baseTextureResolution = baseTexture.resolution;
-        const modX = ((this.tilePosition.x / this.tileScale.x) % texture._frame.width) * baseTextureResolution;
-        const modY = ((this.tilePosition.y / this.tileScale.y) % texture._frame.height) * baseTextureResolution;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks for the PR! Great first submission.

If you don't mind, would you mind throwing something together so we can see this working in action. It's nice to see what it does in the latest release (https://pixijs.download/release/pixi.min.js) and with your fix applied.  It'll be easier to approve.

Also, @ivanpopelyshev originally worked on this code. Would be good to get a review from him. 

Thanks!

a-rasinski:
I've prepared a pen with comparison of canvas and webgl renderer:
https://codepen.io/a-rasinski/pen/vQjKXw

bigtimebuddy:
Nice work, thanks for the verification!

a-rasinski:
I just started more digging in TilingSprite and it seems that it also doesn't include trimming transformation from atlas, which results in texture bleeding. 

So I'll be posting some updates to this during the weekend to have a full solution.

bigtimebuddy:
This looks better, thanks for fixing it.

ivanpopelyshev:
Good one! I can enchant it later that way it allows any rotation, not only texturePacker 90-degree

bigtimebuddy:
Thanks so much @a-rasinski for your contribution. Much appreciated.

PR #5245, opened 2018-11-25
Fix Graphics _localBounds calculation

kib357:
Graphics bounds calc was broken after adding `lineAlignment` option for non-default values of `lineAlignment`. For example, this is what we getting when value is `1` (outer lines):
image
And this is for `0` (inner lines):\
image
Code in PR resolves this issue.

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -1114,16 +1114,7 @@ export default class Graphics extends Container
                 const lineWidth = data.lineWidth;
                 const lineAlignment = data.lineAlignment;
 
-                let lineOffset = 0;
-
-                if (lineAlignment === 0.5)
-                {
-                    lineOffset = lineWidth / 2;
-                }
-                else if (lineAlignment === 1)
-                {
-                    lineOffset = lineWidth;
-                }
+                const lineOffset = lineWidth * lineAlignment;
 
                 shape = data.shape;
 
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -1112,15 +1112,27 @@ export default class Graphics extends Container
                 const data = this.graphicsData[i];
                 const type = data.type;
                 const lineWidth = data.lineWidth;
+                const lineAlignment = data.lineAlignment;
+
+                let lineOffset = 0;
+
+                if (lineAlignment === 1)
+                {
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
I'm gonna make fix for polygon, because there's no guarantee about its orientation.

src/core/graphics/Graphics.js
```diff
@@ -1112,15 +1112,27 @@ export default class Graphics extends Container
...
+
+                let lineOffset = 0;
+
+                if (lineAlignment === 0.5)
+                {
+                    lineOffset = lineWidth / 2;
+                }
+                else if (lineAlignment === 1)
+                {
+                    lineOffset = lineWidth;
```
bigtimebuddy: lineAlignment should be treated as a float from 0-1. So probably lineOffset = lineWidth * lineAlignment instead.

src/core/graphics/Graphics.js
```diff
@@ -1112,15 +1112,27 @@ export default class Graphics extends Container
                 const data = this.graphicsData[i];
                 const type = data.type;
                 const lineWidth = data.lineWidth;
+                const lineAlignment = data.lineAlignment;
+
+                let lineOffset = 0;
```
bigtimebuddy:
```suggestion
                const lineOffset = lineWidth * lineAlignment;
```

kib357: Thank you, good point!

kib357:
> I'm gonna make fix for polygon, because there's no guarantee about its orientation. Alignment 0 and 1 is the same there.

@ivanpopelyshev Bounds calculation is correct for non-clockwise oriented polygons, but rendering is consider this polygons as turned inside out. So i don't now what should be the right behaviour...

ivanpopelyshev:
> @ivanpopelyshev Bounds calculation is correct for non-clockwise oriented polygons, but rendering is consider this polygons as turned inside out. So i don't now what should be the right behaviour...

It's correct for clockwise, its incorrect for non-clockwise, i will prove it tomorrow :)

bigtimebuddy:
Thank you for this change @kib357! @ivanpopelyshev, if you want to do a follow-up PR to consider the direction of the polygon, that's welcomed, but I think this is a good improvement over the existing local bounds.

PR #5257, opened 2018-12-01
Problem with destroying active framebuffer

ivanpopelyshev:
Follow-up for #5225

https://www.pixiplayground.com/#/edit/RnH1N_j5H3GaMxiyFxykr


bigtimebuddy:
Thanks @ivanpopelyshev 👍

PR #5261, opened 2018-12-03
Fixes defaultAnchor cloning with Texture.clone

bigtimebuddy:
Follow up to #5011, where `Texture.clone` does not consider  copying `defaultAnchor`.

### DIFF
--- a/src/core/textures/Texture.js
+++ b/src/core/textures/Texture.js
@@ -274,7 +274,7 @@ export default class Texture extends EventEmitter
      */
     clone()
     {
-        return new Texture(this.baseTexture, this.frame, this.orig, this.trim, this.rotate);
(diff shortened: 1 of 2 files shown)


djmisterjon:
@bigtimebuddy 
here you go
https://www.pixiplayground.com/#/edit/LfszOC7chpYp_2oHya9dr

So because i was use this way, but now it not possible, i found alternative, but before change , it was possible, so maybe something are break somewhere!
good luck

bigtimebuddy:
PIXI.Sprite does not accept BaseTexture objects in the constructor. I'm not sure when this ever worked. But this works fine:

https://www.pixiplayground.com/#/edit/m3qS2eBvf3VRRk0gCNXsX

djmisterjon:
well ok nice , sorry for the bas issue !

PR #5267, opened 2018-12-06
Fixed text overflow when using letterSpacing.

ceco-fmedia:
Fixed problem with text metrics when having letterSpacing, which could cause the text to overflow the canvas:
https://jsfiddle.net/sqzbjonm/15/

### DIFF
--- a/src/core/text/Text.js
+++ b/src/core/text/Text.js
@@ -293,7 +293,7 @@ export default class Text extends Sprite
 
         for (let index = 0, textLen = text.length; index < textLen; ++index)
         {
-            const current = text.charAt(index);
+            const current = text[index];
 
             if (isStroke)
             {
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Thanks for submitting the PR.

It's not super clear to me from your fiddle what you're trying to fix. Are you trying to solve the clipping created by using letter-spacing? Could you please be clearer, more explicit and verbose about your issue before you submit a solution.

I have a few problems with this PR but overall I feel like you need to find a simpler solution. Since you are changing tests that affect wrapping, this is a breaking change and will not be considered for v4.

ceco-fmedia:
@bigtimebuddy 
The PR is aimed to fix the clipping of the text, caused by the different rendering of the text when using letterSpacing.
The width was determined by measuring whole lines, but the rendering when using letterSpacing is done letter by letter. Because of the font kerning, measureText("abc") is not equal to measureText("a") + measureText("b") + measureText("c").

The change in the unit test was required, as currently one of the words in longText(the first line rendered in fiddle) sample didn't actually fit inside the wordWrapWidth, and 2 other have different result depending on the OS(probably caused by different font implementation)

PR #5268, opened 2018-12-06
Update resource-loader version

### LINKED ISSUE #5265: Cross Origin errors in Sandboxed iFrame
### Expected Behavior
When executing Pixi.js code inside a sandboxed iframe that does not have the `allow-same-origin` attribute, it should automatically make cross origin requests for any loaded resources.

### Current Behavior
Currently, this does not happen, due to the default behaviour of https://github.com/englercj/resource-loader which fails to check `window.origin` to see if it is set to `null` which it will be inside a sandboxed iframe.

### Possible Solution
This issue has been reported upstream at https://github.com/englercj/resource-loader/issues/119 and fixed at https://github.com/englercj/resource-loader/issues/120, once this fix has been merged and a new version of `resource-loader` has been released, upgrading the package should fix this issue in Pixi.js.

### Steps to Reproduce
1. Run any Pixi.js app that uses resources (particularly WebGL resources) from inside a sandboxed iframe:
```
<iframe sandbox='allow-scripts' src='index.html'></iframe>
```
Where the index.html is a bootstrap for the Pixi.js app.

2. Observe the CORS errors in the console, as WebGL complains about reading out data from a 'tainted' canvas (i.e. one that has improperly loaded cross origin images).

### Environment

- `pixi.js` version: _4.5.2_ (but should still be extant)
- Browser: _Chrome 70_
- Device: _Ubuntu 16.04 Desktop_

rtibbles:
Updates resource-loader to 2.2.0 in order to fix #5265

Wasn't sure if I should include the updates to the package-lock.json also, so defaulted to yes. Happy to revert if needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,14 +38,14 @@
       "integrity": "sha512-p77r45N+RrA2QyGpYnnEIGls6igqTRzCMxDPJQhk09Y0GJPF7i0m7PtQuzXMhONz/J4B/6slMmX0lhh6mbDjzw==",
       "dev": true,
       "requires": {
-        "taffydb": "2.7.3",
-        "underscore": "1.8.3"
+        "taffydb": "^2.7.2",
+        "underscore": "^1.7.0"
       }
     },
     "@sinonjs/formatio": {
       "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Thanks @rtibbles

PR #5271, opened 2018-12-07
Dependency updates

mrmajik45:
Dududu-updates-dudududu

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6266,9 +6266,9 @@
       }
     },
     "remove-array-items": {
-      "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Thanks, but we will not be upgrading all these dependencies at once for v4. Also, we have kept some old dependencies intentionally to avoid subtle breakages. For instance moving to jsdoc 3.5+ breaks some of our event docs. We’d welcome a PR like this to v5 (next branch) but upgrading all these dependencies introduces too much instability to v4. Unless someone needs a specific feature in a dependency, my preference is to keep things locked down.

PR #5272, opened 2018-12-07
Lossless compresso

mrmajik45:
*Total -- 108.75kb -> 99.96kb (8.08%)

/test/core/resources/src/star2.png -- 1.77kb -> 1.01kb (42.94%)
/test/core/resources/src/star3.png -- 1.80kb -> 1.06kb (40.89%)
/test/core/resources/src/star1.png -- 1.78kb -> 1.05kb (40.82%)
/test/core/resources/src/star4.png -- 1.70kb -> 1.07kb (37.16%)
/test/core/resources/src/goldmine_10_5.png -- 73.02kb -> 67.57kb (7.46%)
/test/loaders/resources/atlas.png -- 14.34kb -> 14.10kb (1.66%)
/test/loaders/resources/atlas@0.5x.png -- 14.34kb -> 14.10kb (1.66%)


englercj:
@mrmajik45 Thanks for the PR.

What is this trying to solve?

mrmajik45:
Just reduced file size.

englercj:
> Just reduced file size.

Sure, but for what purpose? Changing them in the repo will increase repo size (the olds ones don't go away from the history), and they are used in tests only which are served locally so there isn't any download time.

bigtimebuddy:
I agree with @englercj. These are not critical to compress since they are only used for tests.

However @mrmajik45, if you want to compress images, consider doing this for the examples: https://github.com/pixijs/examples

These images are actually served to people and would be helpful to see them optimized (if they aren't already).

PR #5274, opened 2018-12-09
Rename parameters for user friendly

eXponenta:
Constructor with (x, y, width, height) parameters confused, because rectangle-based ellipse supposed, but it's created from two radiuses and middle point.


bigtimebuddy:
Thank you @eXponenta for the PR!

PR #5277, opened 2018-12-09
[next] House-keeping for dependencies and tests

bigtimebuddy:
## Overview

As we get ready for our release candidate of v5, taking the opportunity to do some house-keeping on our dependencies and tests. Details below.

### Changed

* Updates devDependencies:
  * Updates **floss** (latest)
  * Updates **electron** (latest @2)
  * Updates **rollup** and plugins  (latest)
  * Updates **lerna** (latest @2)
  * Updates **eslint** (latest)
* Removes **bit-twiddle**
* Removes render tests and app for generating tests, this is formalized in the [v5-render-tests](https://pixijs.io/v5-render-tests) page. This is much better for comparing relative changes since Travis doesn't support WebGL anyways. 

### Fixed

* Fixes circular export/import within the core resources package


englercj:
lg2m, as soon as the build failures are cleaned up we can merge.

bigtimebuddy:
Something funny is happening here with Travis & Electron. The WebGL unit-tests _should_ be ignored because we're running on Travis however these not getting skipped like they did in Electron@1. I think this is because Electron@2 might be returning a false-negative in `isWebGLSupported` function.

bigtimebuddy:
# :poop: 
Yes, I did a little test and Electron is returning `true` for `isWebGLSupported` when running in Travis. 

https://travis-ci.org/bigtimebuddy/test-electron-webgl-travis/builds/465800657#L649

bigtimebuddy:
I replaced the `isWebGLSupported` check with an environmental variable. I think this a better approach because now I can simulate Travis locally (which I couldn't do before). It _may_ be nice to fix the feature detection to consider a headless environment but it seems like a super tiny use-case. I couldn't find an easy solution after searching for an hour. I'm good on this PR now.

PR #5280, opened 2018-12-10
[next] Batch renderer fix O(N^2) memory problem

ivanpopelyshev:
Better test it on bunnymark.

### DIFF
--- a/packages/core/src/batch/BatchRenderer.js
+++ b/packages/core/src/batch/BatchRenderer.js
@@ -4,7 +4,7 @@ import ObjectRenderer from './ObjectRenderer';
 import checkMaxIfStatementsInShader from '../shader/utils/checkMaxIfStatementsInShader';
 
 import { settings } from '@pixi/settings';
-import { premultiplyBlendMode, premultiplyTint } from '@pixi/utils';
+import { premultiplyBlendMode, premultiplyTint, nextPow2, log2 } from '@pixi/utils';
 
 import BatchBuffer from './BatchBuffer';
 import generateMultiTextureShader from './generateMultiTextureShader';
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
I made minimal size of buffer, 2 sprites (8 vertices, 12 indices). Is that fine, @GoodBoyDigital

packages/core/src/batch/BatchRenderer.js
```diff
@@ -169,21 +169,37 @@ export default class BatchRenderer extends ObjectRenderer
...
+        // 12 indices is enough for 2 quads
+        const roundedP2 = nextPow2(Math.ceil(size / 12));
+        const roundedSizeIndex = log2(roundedP2);
+        const roundedSize = roundedP2 * 12;
 
-        let buffer = this.iBuffers[roundedSize];
+        while (this.iBuffers.length < roundedSizeIndex)
+        {
+            this.iBuffers.push(null);
+        }
```
englercj: I think this is off-by-one, should be `<=` or it will not have the index `roundedSizeIndex`.

Instead can we just set the length?

```js
if (this.iBuffers.length <= roundedSizeIndex)
    this.iBuffers.length = roundedSizeIndex + 1;
```

packages/core/src/batch/BatchRenderer.js
```diff
@@ -169,21 +169,37 @@ export default class BatchRenderer extends ObjectRenderer
...
-        const roundedSize = Math.ceil(size / 100.0) * 100;
+        // 8 vertices is enough for 2 quads
+        const roundedP2 = nextPow2(Math.ceil(size / 8));
+        const roundedSizeIndex = log2(roundedP2);
+        const roundedSize = roundedP2 * 8;
+
+        while (this.aBuffers.length < roundedSizeIndex)
+        {
+            this.aBuffers.push(null);
+        }
```
englercj: Same feedback as above.

packages/utils/src/data/pow2.js
```diff
@@ -26,10 +26,35 @@ export function nextPow2(v)
  *
  * @function isPow2
  * @memberof PIXI.utils
- * @param {number} value
- * @param {boolean} `true` if value is power of two
+ * @param {number} v input value
+ * @return {number} `true` if value is power of two
```
englercj: Typo? This is a boolean return type.

packages/core/src/batch/BatchRenderer.js
ivanpopelyshev: You're right but it actually doesn't matter, I read somewhere that setting a[a.length] is same as push. I'll fix it

englercj: For example, using push:

image

And using `.length =`:

image

There is a difference in the engine, even if the JS user-experience is different only in that you get `null` rather than `undefined`. I'd prefer `.length=` because it is easier to read, smaller, and can be faster.

bigtimebuddy:
@englercj could you take another look?

packages/core/src/batch/BatchRenderer.js
ivanpopelyshev: i added `=` back then.

Ok, its `length=` now :)

PR #5287, opened 2018-12-14
Refactor uploadUvs for ParticleRenderer

michaeljh619:
First time contribution here, I saw a "TODO" and just went for it.
Updated some of the docs as well, as I saw they were copy/pasted but weren't updated to reflect the function they were documenting.


ivanpopelyshev:
Of course UV's are uploaded rarely (each time new element is added in middle of array) but that idea also contradicts the nature of ParticleBuffer upload functions - each functions supposed to be as effective as possible.

The `if (textureUVs)` is also a precaution here, all elements will have same texture => its valid, all of its subtextures are supposed to be evaluated.

I can tell many things about our ParticleRenderer impementation , I just don't want ot focus on it right now. Thank you for a try!

ivanpopelyshev:
maybe @englercj knows if this thing can possibly be unfolded in runtime and produce the same code?

PR #5295, opened 2018-12-19
[next] texturefill matrix

ivanpopelyshev:
If we start to draw graphics from (0, 40) , we have to pass matrix with ty=40, after inversion it becomes "-40", and result UV is zero.

Also, we have to clone matrix because someone can change this thing after `beginTextureFill`.

```js
beginTextureFill(... , matrix);
matrix.tx = 40; //BWAHAHAHAHA
```


bigtimebuddy:
@ivanpopelyshev does this also apply to `lineTextureStyle`? that also has a texture transform matrix.

PR #5301, opened 2018-12-20
Intorduction of miter, bevel and round lineJoin

zawor:
Adresses pixijs/pixi.js#1637

- introduces lineJoin param for Graphics with possible values: miter, bevel, round
- defaults are backward compatible, falling back to miter
- among many its inspiration heavily comes from https://hypertolosana.wordpress.com/2015/03/10/efficient-webgl-stroking/
- includes bunch of unit tests which cover most of the introduced branches in code
- changes look of bevel line join in case of `pdist > 196 * width * width` see image in the [link](https://www.dropbox.com/s/f83spsboyoaixvo/pixi-bevel-line-join-before-n-after.png?dl=0)

PS: round lineJoin implementation is more or less prepared for reuse in round lineCap

### DIFF
--- a/src/core/graphics/Graphics.js
+++ b/src/core/graphics/Graphics.js
@@ -74,6 +74,14 @@ export default class Graphics extends Container
         this.lineAlignment = 0.5;
 
         /**
+         * The style of line join ('miter', 'round', 'bevel')
+         *
+         * @member {string}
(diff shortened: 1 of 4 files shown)


zawor:
argh... forgot about linting... will be back with new one in a sec

PR #5302, opened 2018-12-20
 Intorduction of miter, bevel and round lineJoin

zawor:
Adresses pixijs/pixi.js#1637

- introduces lineJoin param for Graphics with possible values: miter, bevel, round
- defaults are backward compatible, falling back to miter
- among many it is heavy inspiration comes from https://hypertolosana.wordpress.com/2015/03/10/efficient-webgl-stroking/
- includes bunch of unit tests which cover most of the introduced branches in code
- changes look of bevel line join in case of `pdist > 196 * width * width` see image in the [link](https://www.dropbox.com/s/f83spsboyoaixvo/pixi-bevel-line-join-before-n-after.png?dl=0)

PS: round lineJoin implementation is more or less prepared for reuse in round lineCap

### DIFF
--- a/test/core/Graphics.js
+++ b/test/core/Graphics.js
@@ -116,6 +116,845 @@ describe('PIXI.Graphics', function ()
 
             expect(graphics.currentPath.shape.points).to.deep.equal([0, 0, 10, 0]);
         });
+
+        describe('lineJoin', function ()
+        {
+            const renderer = new PIXI.WebGLRenderer(200, 200, {});
+
+            describe('miter', function ()
+            {
+                it('is miter by default (backwards compatible)', function ()
+                {
+                    // given
+                    const graphics = new PIXI.Graphics();
+
+                    // then
+                    expect(graphics.lineJoin).to.be.equal('miter');
+                });
+
+                it('clockwise miter', function ()
+                {
+                    // given
+                    const graphics = new PIXI.Graphics();
+
+                    graphics.lineStyle(2, 0, 1, 0.5);
+                    graphics.lineJoin = 'miter';
+
+                    graphics.moveTo(0, 0);
+                    graphics.lineTo(50, 0);
+                    graphics.lineTo(50, 50);
+
+                    // when
+                    renderer.render(graphics);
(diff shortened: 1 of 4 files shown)


zawor:
`[floss]: Error: This browser does not support webGL. Try using the canvas renderer` ok this one i see for the 1st time... can someone have a look?

test/core/Graphics.js
```diff
@@ -116,6 +116,845 @@ describe('PIXI.Graphics', function ()
 
             expect(graphics.currentPath.shape.points).to.deep.equal([0, 0, 10, 0]);
         });
+
+        describe('lineJoin', function ()
```
englercj: This function needs to be wrapped in withGL( like the other functions that use webgl, so that it is ignored by platforms that don't support it.

zawor: oh snap... i totally overlooked that... will fix asap... TY

zawor:
yay ✅ build! 🎉

now only beating in code review and hopefully it will be ready for merge ;)

ivanpopelyshev:
You are a hero.

ivanpopelyshev:
Oh, you referenced Ibon's blog, nice

zawor:
no problem... will rebase this on `next` tomorrow and make yet another PR... for us this means that for short time (until v5 is out) we'll have to maintain our own custom pixi dist on the side but this way it will be cleaner

bigtimebuddy:
Yes, that's correct.

This will, however, be a *really* great features to get more devs on v5. Super appreciate your flexibility. Also, rebasing on `next` might be difficult since we moved stuff around a lot. Let me know if you have any question about that.

englercj:
@zawor `next` has been renamed to `dev`. So you will need to rebase on the the new `dev` from the remote.

zawor:
already on it... but now fighting a bit with tests as they look a bit different and on top of it i was robbed from time today so hopefully i'll scrape a bit of time during xmas... anyway most of the changes i have in place and have to make unit tests pass again ;)

PR #5310, opened 2018-12-23
Make default filter work like in v4

ivanpopelyshev:
Also, remove unused uniform from BlurFilter

### DIFF
--- a/packages/core/src/filters/defaultFilter.frag
+++ b/packages/core/src/filters/defaultFilter.frag
@@ -1,22 +1,7 @@
 varying vec2 vTextureCoord;
-varying vec2 vFilterCoord;
 
 uniform sampler2D uSampler;
-uniform sampler2D filterSampler;
 
 void main(void){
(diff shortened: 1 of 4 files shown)


packages/core/src/filters/defaultFilter.frag
```diff
@@ -1,22 +1,7 @@
...
-   {
-     color = vec4(1.0, 0.0, 0.0, 1.0);
-   }
-   else
-   {
-     color = vec4(0.0, 1.0, 0.0, 1.0);
-   }
-   // gl_FragColor = vec4(mod(vFilterCoord.x, 1.5), vFilterCoord.y,0.0,1.0);
-   gl_FragColor = mix(sample, masky, 0.5);
-   gl_FragColor *= sample.a;
```
GoodBoyDigital: haha! what on earth is that code doing there! Good catch @ivanpopelyshev !

PR #5315, opened 2018-12-30
v4 - interaction assignment fix

themoonrat:
##### Description of change
For some reason I can't understand, but probably due to polyfills, an actual pointer event can come through to the 'normalising' function; and you cannot overwrite existing properties on this object, causing the error seen at https://github.com/pixijs/pixi.js/issues/5283

##### Pre-Merge Checklist


themoonrat:
@bigtimebuddy Dunno if you'd like me to submit a separate PR for this for v5?

bigtimebuddy:
I can cherry pick this commit into v5, should be okay without another PR.

PR #5317, opened 2019-01-01
Auto generation for typescript definitions

ivanpopelyshev:
Work in progress.

```bash
npm install
npm run types
```

The result is inside `dist/types` folder.

Current problems:

1. resource-loader dependency
2. we have to separate canvas definitions from the rest
3. `private:true` for jsdocs
4. strange messages about `module.exports`
5. `module` for static members of classes


ivanpopelyshev:
OK, now we have only one problem before it generates correct typescript definition - static overrides, `from` and `fromImage`.

I'll add it to manual steps list. 

Please review and merge this.

bigtimebuddy:
Please don’t merge this yet, I’m testing it today.

ivanpopelyshev:
Manual changes list:
1. add typings for resource-loader
2. event-emitter
3. change `from` and `fromImage`, add extra declarations in `TilingSprite` and `CubeTexture`.

I'm gonna commit typings based on that thing and manual changes in separate PR soon

bigtimebuddy:
Currently, Travis is blocked on getting the latest tsd-jsdoc published. There's critical fix in master not yet on NPM. Also @ivanpopelyshev created 2 PRs that would be helpful here (we could remove some hacks).

We have been testing typings in this branch of [pixi-typescript](https://github.com/pixijs/pixi-typescript/tree/v5.0-rc). Seems to work well and support typings for legacy and default bundles.

englercj:
Just published [tsd-jsdoc@2.1.1](https://github.com/englercj/tsd-jsdoc/releases/tag/v2.1.1)

package.json
```diff
@@ -10,12 +10,16 @@
     "pretest": "npm run lint && npm run build",
     "test": "floss --path test",
     "unit-test": "floss --path test",
-    "docs": "jsdoc -c jsdoc.conf.json -R README.md",
+    "docs": "mkdirp dist && jsdoc -c jsdoc.conf.json -R README.md",
+    "types": "npm run types:legacy && npm run types:default",
+    "types:default": "mkdirp dist/types/pixi.js && jsdoc -c types/jsdoc.conf.json && node types/assemble pixi.js",
+    "types:legacy": "mkdirp dist/types/pixi.js-legacy && jsdoc -c types/jsdoc-legacy.conf.json && node types/assemble pixi.js-legacy",
```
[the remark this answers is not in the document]
ivanpopelyshev: It makes sense, however we have no userbase that use mini-modules yet, and there's no tutorial on how to make your own analogue of `bundle.js` that is required for that case.

ivanpopelyshev:
I can separate it, if it doesn't makes a problem for you to use them both at the same time to test it.

package.json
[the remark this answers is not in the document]
englercj: We have no userbase that uses individual packages yet, because they are not released.

Without doing it the way I'm describing it isn't even optional, it is just unsupported. If we are going to publish separate packages to npm (we are) then they should work.

englercj:
>I can separate it, if it doesn't makes a problem for you to use them both at the same time to test it.

I'm saying we should PR the jsdoc comment changes and merge them, then look at (and test) the type changes separately. Should be easy to test the type changes because we should've already merged the comment changes if we are doing that.

englercj:
The jsdoc comments have been merged into dev, lets rebase this to include only the type generation changes.

types/loader.d.ts
```diff
@@ -0,0 +1,138 @@
+declare namespace PIXI {
```
englercj: Is this needed? Couldn't you just have the script load the typings in `node_modules/resource-loader/typings/resource-loader.d.ts`?

types/events.d.ts
```diff
@@ -0,0 +1,129 @@
+declare namespace PIXI {
+    namespace utils {
+// https://github.com/primus/eventemitter3
+        export interface EventEmitter {
```
englercj: Is this needed? Couldn't you just have the script load the typings in `node_modules/eventemitter3/index.d.ts`?

types/loader.d.ts
ivanpopelyshev: resource-loader typings depend on mini-signals which are not exposed in pixi. @mat's Runner is not exposed too. When I tried to expose signals my PR was rejected.

types/events.d.ts
ivanpopelyshev: Pixi classes extend `PIXI.utils.EventEmitter` and that class is defined in `utils` through jsdoc. The other thing is that interface. 

I can make script that transforms EventEmitter3 typings to what we need, but I think that its over-complicating things.

types/loader.d.ts
englercj: Pixi's renderer is just aliased resource-loader, signals are actually exposed.

You can do this:

```
PIXI.loader.onComplete.add()
```

types/events.d.ts
ivanpopelyshev: autogen based on someone else definitions looks like a leftpad for me.

englercj: > The typings from EventEmitter have an export as namespace EventEmitter; line that needs to be removed (not a big deal tho). 

Good point, this is why I feel we should be using import rather than generating single-files, but we can do that in the future.

package.json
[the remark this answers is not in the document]
qtiki: Hey guys!

I just ran into some typing problems with the new v5 release. We use TypeScript internally for everything so naturally having working types for Pixi is a must. One of the biggest cool new features for v5 is the individual packages which we can leverage to reduce our bundle size.

Unfortunately as discussed here there are no typings for them. It seems pixi-sound has ran into this issue as well since there is a "hackyish" workaround for it: https://github.com/pixijs/pixi-sound/tree/master/typings

On top of that I ran into some problems with the types in pixi-sound that I reported: https://github.com/pixijs/pixi-sound/issues/107

So I guess my main question is that since pixi-sound and pixi-spine are already written in TypeScript, what is your general feeling towards it? I am playing around with the idea of doing a fork of Pixi to convert it to TypeScript and use the TypeScript compiler to auto-generate the declarations (with comments included so you get nice tooltips inside IDEs like Visual Studio Code). This is essentially how all of our internal packages work as well.

PR #5325, opened 2019-01-04
Introduction of line join

zawor:
Continuation/port of #5302 on current dev branch

##### Description of change
See #5302. Please review changes in Graphics object api, as I'm not quite sure whether they follow current vibe.

##### Pre-Merge Checklist

- [x] Documentation is changed or added (somewhat, please review whether nothing is missing or is sloppy)


zawor:
`drawCircle->should have no gaps in line border` got this change as side effect as well (see [here](https://github.com/pixijs/pixi.js/pull/5325/commits/36fd059576d4faf804fb532e67779dd0abe2e16d#diff-d8a1632d7c917fd816478deb0f6c7a4fL1148))

bigtimebuddy:
It should be easy to support [lineJoin](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineJoin) in CanvasRenderer by updating **packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.js**

zawor:
i would go with one-liner... 

although on 2nd thought i think that round line join is drawn a little bit different by 2d context where p2 is taken as a center of circle and i took point of intersection as a center... which results in a bit "flatter" round line join in current impl

bigtimebuddy:
@zawor I'm testing out this now and having issues with round. Could you please check this out.

https://jsfiddle.net/bigtimebuddy/p61ty9sL/

screen shot 2019-01-08 at 9 51 20 am

packages/graphics/src/Graphics.js
```diff
@@ -262,10 +263,11 @@ export default class Graphics extends Container
      * @param {PIXI.Matrix} [matrix=null] Texture matrix to transform texture
      * @param {number} [alignment=0.5] - alignment of the line to draw, (0 = inner, 0.5 = middle, 1 = outter)
      * @param {boolean} [native=false] - If true the lines will be draw using LINES instead of TRIANGLE_STRIP
+     * @param {string} [lineJoin='miter'] - shape used to join two line segments where they meet
      * @return {PIXI.Graphics} This Graphics object. Good for chaining method calls
      */
     lineTextureStyle(width = 0, texture = Texture.WHITE, color = 0xFFFFFF, alpha = 1,
-        matrix = null, alignment = 0.5, native = false)
+        matrix = null, alignment = 0.5, native = false, lineJoin = LINE_JOIN.MITER)
```
bigtimebuddy: Argument and doc should probably be added to `lineStyle` too. I think I will refactor these argument into an object in a future PR. This method signature is getting too crazy.

bigtimebuddy:
Yeah, don't worry about the Ticker tests, those are intermittently failing.

zawor:
it seems like way how indices are created changed and it is not quite compatible with v4 used to roll... if yes you would have point me to discussion as i simply hacked my way through based on what i could understand from code ;)

here is better jsfiddle:
https://jsfiddle.net/rowagnairda/qak0w6fd/

homebrew build can be found here:
https://beta.tractalis.com/pixi-hack-and-slash/pixi.min.js

bigtimebuddy:
@GoodBoyDigital could you help out @zawor here with some more context about the indices change in v5.

GoodBoyDigital:
nice work @zawor , the main change is that v4 used `triangle_strip` draw method where as now we use `triangles`

one of the things we changed to make everything batch-able :D

zawor:
ok that explains everything...

but now i started to wonder whether it is really needed to have all of the points? I believe some were imposed by the way  how triangle strip used to work... hmm...

bigtimebuddy:
This is looking much better. I'm comparing Canvas and WebGL. There's a little difference with the round line joins. Personally, I think Canvas looks better and more circular. How difficult would it be to get WebGL a little closer to how canvas renders round line join?

### Canvas
screen shot 2019-01-09 at 9 51 36 pm

### WebGL
screen shot 2019-01-09 at 9 51 19 pm

bigtimebuddy:
@zawor thanks for the explanation. If you can take a stab at it, I would really love to see the round improved. Of all the line types, I think round is one could be used the most and visually I expect it to be more like canvas.

zawor:
sure... will do. I wouldn't expect it today, but rather over the weekend as one of my younglings got sick and sponsors me "holidays" :P

wangrenyn:
If you can take a stab at it, I would really love to see the round improved. Of all the line types, I think,,

bigtimebuddy:
@zawor I feel like we're so close on this, have you had a chance to cleanup your code and commit?

zawor:
Hi i'm really sory for such delay... i've got dragged by emergency from the side... hopefully i'll be able to finish it this week as only thing which is broken now is few unit tests

zawor:
hmm... tests and transparency revealed that i'm further than i thought :P

obraz

ivanpopelyshev:
That's fun. Vector is a difficult thing, dont be sorry for delays.

zawor:
i'm rather getting angry... as further i am, more i get dragged away but other stuff which falls on me... 

Anyway, what is planned release date for 5.0.0?

bigtimebuddy:
Not sure but v5 is around the corner. I’m also okay if we do this for 5.1.0 release. No rush. There’s not many people who have the patience for this work, so take your time. I would ask that you close this PR until you’re ready to resubmit.

zawor:
Ack... so closing so that it won't hang there waiting for salvation.

PR #5326, opened 2019-01-04
Assisted typings for pixijs

ivanpopelyshev:
Built with https://github.com/pixijs/pixi.js/pull/5317


ivanpopelyshev:
Here are modified tests: https://github.com/pixijs/pixi-typescript/tree/v5.0-rc

PR #5331, opened 2019-01-06
Update Program with shader language version

ivanpopelyshev:
Based on #5530 by @Dairnon, I moved it to our repo.

@bigtimebuddy is there a way to change branch in original PR?

> adds the ability to define the shader language version.

> Very useful if trying to write shaders in different versions like OpenGL ES 3

> So we can access stuff like MRT

build: https://pixijs.download/dev-shader-version/pixi.js


Dairnon:
Here is an example of the new arg in action.

GLSL 3.0.0 es shader code conversion of mesh triangle. 

https://www.pixiplayground.com/#/edit/u~ylQ25fgguC~Eoqy91Qt

ivanpopelyshev:
OK, I removed parameter and modified our pixi-playground example. Now the patch is easy and helps us to go forward without waiting for big refactoring.

Dairnon:
Demo

https://www.pixiplayground.com/#/edit/K8If4VxXDfFlLCUqQBxvZ

PR #5340, opened 2019-01-08
added a warning for auto detection of stride

GoodBoyDigital:
Created a slightly better naming for shaders.
Also added a warning for when the auto generation of attributes is not possible

### DIFF
--- a/packages/core/src/shader/Program.js
+++ b/packages/core/src/shader/Program.js
@@ -12,6 +12,8 @@ import { settings } from '@pixi/settings';
 
 let UID = 0;
 
+const nameCache = {};
+
 /**
  * @class
  * @memberof PIXI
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Can you please integrate https://github.com/pixijs/pixi.js/pull/5331 in it?

packages/core/src/shader/Program.js
```diff
@@ -43,8 +45,20 @@ class Program
...
+        }
+        else
+        {
+            nameCache[name] = 1;
+        }
+
+        this.vertexSrc = `#define SHADER_NAME ${name}\n${this.vertexSrc}`;
+        this.fragmentSrc = `#define SHADER_NAME ${name}\n${this.fragmentSrc}`;
+
+        // console.log('settings: ', settings);
```
englercj: Remove dead code.

bigtimebuddy:
```suggestion
```

PR #5342, opened 2019-01-09
Fix text displaying bug for browsers that block canvas data extraction

ProtonGustave:
##### Description of change

There is some privacy-focused browsers(Firefox 58+, Tor) that disallow canvas data extraction(_toDataURL_, _getImageData_) by default(not really disallow, just return white picture), i.e. they ask for permission to do so(although i don't have a permission request on localhost). **TextMetrics.js** is using _getImageData_ in static function called _measureFont_ to calculates the ascent, descent and font size of a font-style, but when user decline permission request or just ignore it function return zeroes for ascent, descent and font size, then text element became 1px tall(not visible on screen). No need to use _measureFont_ if we can't, there is fallback in commit that allow to show text if we can't extract canvas data.
##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/TextMetrics.js
+++ b/packages/text/src/TextMetrics.js
@@ -49,6 +49,13 @@ export default class TextMetrics
         wordWrap = (wordWrap === undefined || wordWrap === null) ? style.wordWrap : wordWrap;
         const font = style.toFontString();
         const fontProperties = TextMetrics.measureFont(font);
+
+        if (fontProperties.fontSize === 0)
+        {
+            fontProperties.fontSize = style.fontSize;
+            fontProperties.ascent = style.fontSize;
+        }
+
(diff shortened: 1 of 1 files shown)


packages/text/src/TextMetrics.js
```diff
@@ -49,6 +49,13 @@ export default class TextMetrics
         wordWrap = (wordWrap === undefined || wordWrap === null) ? style.wordWrap : wordWrap;
         const font = style.toFontString();
         const fontProperties = TextMetrics.measureFont(font);
+
+        if (fontProperties.fontSize === 0)
```
bigtimebuddy: Great explanation in your PR description. Could you add a comment inline about why you're doing that here? I think it's very important for context whenever there are weird edge-cases like this.

PR #5348, opened 2019-01-11
Remove accessibility manager hook div from dom on destruction

Izumi-kun:
##### Description of change

When running on phone or tablet accessibility manager creates hook div that doesn't removes when manager dies before the hook has been focused.

##### Pre-Merge Checklist

### DIFF
--- a/packages/accessibility/src/AccessibilityManager.js
+++ b/packages/accessibility/src/AccessibilityManager.js
@@ -144,6 +144,7 @@ export default class AccessibilityManager
     /**
      * Creates the touch hooks.
      *
+     * @private
      */
     createTouchHook()
     {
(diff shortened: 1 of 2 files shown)


packages/accessibility/src/AccessibilityManager.js
```diff
@@ -157,10 +162,25 @@ export default class AccessibilityManager
...
         });
 
         document.body.appendChild(hookDiv);
+        this._hookDiv = hookDiv;
+    }
+
+    /**
+     * Destroys the touch hooks.
+     *
+     */
```
bigtimebuddy: Could you mark this as `@private`?

Izumi-kun: Also for `createTouchHook` method?

bigtimebuddy: Yes

PR #5349, opened 2019-01-12
Proper disposal for mesh and graphics geometries

ivanpopelyshev:
We have to discuss API, I've added `shared` field to Buffer and Geometry, and dev can mark them with `markShared()` method.

Demo:
https://www.pixiplayground.com/#/edit/YqUB4vUo8_Qcyqh6Z~PPT

You can see in the console that each second either strip either all geometries are disposed and it works. Extra VAO that is destroyed in apocalyptic event belongs to BatchRenderer.

Next step after this PR is fix of texture and shader disposal, and `contextLost` test.

### DIFF
--- a/packages/core/src/geometry/Buffer.js
+++ b/packages/core/src/geometry/Buffer.js
@@ -80,17 +80,6 @@ export default class Buffer
     }
 
     /**
-     * Marks this buffer as shared, it wont be disposed at the same time as geometry
-     * @returns {PIXI.Buffer}
-     */
-    markShared()
-    {
-        this.shared = true;
-
-        return this;
-    }
-
-    /**
      * Helper function that creates a buffer based on an array or TypedArray
      *
      * @static
--- a/packages/core/src/geometry/Buffer.js
+++ b/packages/core/src/geometry/Buffer.js
@@ -39,12 +39,6 @@ export default class Buffer
 
         this.static = _static;
 
-        /**
-         * How many geometries with existing VAO's referenced this Buffer
-         * @member {boolean}
-         */
-        this.refCount = 0;
-
         this.id = UID++;
 
         this.disposeRunner = new Runner('disposeBuffer', 2);
--- a/packages/core/src/geometry/Buffer.js
+++ b/packages/core/src/geometry/Buffer.js
@@ -1,3 +1,5 @@
+import Runner from 'mini-runner';
+
 let UID = 0;
 /* eslint-disable max-len */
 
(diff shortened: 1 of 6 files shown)


bigtimebuddy:
The API `markShared` seems silly and not useful. Why not just set `shared` directly?

ivanpopelyshev:
That's a fluent interface.

```
var geom = new GraphicsGeometry().markShared();
var graphics1 = new Graphics(geom);
var graphics2 = new Graphics(geom);
```

Its optional, I can removei t.

packages/core/src/geometry/Buffer.js
```diff
@@ -11,8 +13,11 @@ export default class Buffer
 {
     /**
      * @param {ArrayBuffer| SharedArrayBuffer|ArrayBufferView} data the data to store in the buffer.
+     * @param {boolean} [_static=true] `true` for static buffer
+     * @param {boolean} [index=false] `true` for index buffer
+     * @param {boolean} [shared=false] if `true`, buffer will stay alive when geometry is disposed
      */
-    constructor(data, _static = true, index = false)
```
GoodBoyDigital: Hello! There must be a way of not having to pass in shared as a property.
Could we do object counting instead?

Dairnon: Wouldn't that take a serious effort to add?

ivanpopelyshev: I don't want to open this can of worms.

We can remove it from constructor, its fine as a property. If user makes shared buffer, he needs extra operations with it anyway and one more line is not a problem

GoodBoyDigital: @Dairnon agreed, It definitely would take more effort. I see a buffer the same way as I see a texture. You would not create a texture and declare it to be shareable iyswim?

packages/core/src/geometry/GeometrySystem.js
```diff
@@ -257,7 +275,7 @@ export default class GeometrySystem extends System
         const attribs = geometry.attributes;
         const shaderAttributes = program.attributeData;
 
-        const strings = [geometry.id];
+        const strings = ['g', geometry.id];
```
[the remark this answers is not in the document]
ivanpopelyshev: to distinguish original and copies.

ivanpopelyshev:
I don't think changing `geometry` of mesh is intended at all. Are you sure we need those setters and not just "@readonly" modifier for geometry?

PR #5355, opened 2019-01-13
Allows developer to limit the texture update rate of videos

themoonrat:
Can really help the performance; no need to be updating the texture when no change to the video has occured
https://github.com/pixijs/pixi.js/issues/4789


themoonrat:
https://www.pixiplayground.com/#/edit/aVfcZ4_pg4nzHlhYtROB~ for a demonstration.

PR #5357, opened 2019-01-14
V5/ticker enhancements

themoonrat:
There is now a `deltaMS` property, which scales alongside `deltaTime`
There is also now the ability to set a `maxFPS` which compliments the `minFPS`, and lets you throttle the update emitter


themoonrat:
[demo](https://www.pixiplayground.com/#/edit/TPtbM9x3gtMWIEHMidWbL)

PR #5364, opened 2019-01-15
BasePrepare was imbedding settings rather than having a settings.js file like other classes

themoonrat:
Also moved import of settings higher in other index files. Might have been ok where it was, but logically makes sense to import them as early as possible

### DIFF
--- a/packages/utils/src/index.js
+++ b/packages/utils/src/index.js
@@ -63,6 +63,8 @@ export { mixins };
  * @param {number} [dimensions=2] The number of coordinates per vertex in the input array
  * @return {number[]} Triangulated polygon
  */
+import './settings';
+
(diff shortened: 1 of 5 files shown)


packages/utils/src/index.js
```diff
@@ -63,6 +63,8 @@ export { mixins };
  * @param {number} [dimensions=2] The number of coordinates per vertex in the input array
  * @return {number[]} Triangulated polygon
  */
+import './settings';
```
bigtimebuddy: Could you maybe move this to before the `./browser` export. This is weird because the earcut doc is for the earcut export.

bigtimebuddy:
Good cleanup, conforms with how other packages are adding settings.

PR #5369, opened 2019-01-15
A system ticker to sit alongside the shared ticker.

themoonrat:
The shared ticker is tricky atm because you may want to pause the visuals on screen, but you don't want to effect the interactions or core system behaviour.
So having this system ticker separate means that you can now safely pause the shared ticker without worrying about it effecting that system behaviour

### DIFF
--- a/packages/interaction/src/InteractionManager.js
+++ b/packages/interaction/src/InteractionManager.js
@@ -711,7 +711,7 @@ export default class InteractionManager extends EventEmitter
             return;
         }
 
-        Ticker.shared.add(this.update, this, UPDATE_PRIORITY.INTERACTION);
+        Ticker.system.add(this.update, this, UPDATE_PRIORITY.INTERACTION);
 
         if (window.navigator.msPointerEnabled)
         {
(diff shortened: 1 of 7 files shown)


bigtimebuddy:
Do you think it's worth re-thinking the names of `UPDATE_PRIORITY` now that you've separated out the systems stuff?

ivanpopelyshev:
```
deprecation(v5, 'PIXI.ticker.shared has moved to PIXI.Ticker.shared');
```

fix it please.

themoonrat:
@bigtimebuddy I think it's worth keeping as it is. The `system` ticker is still exposed, and devs can use it for their own system stuff that they wouldn't want to pause.

ivanpopelyshev:
Its deprecation message for v4, in case somebody uses that ticker. We should change that statement too :)

bigtimebuddy:
No it's a deprecation message for v5, we removed `PIXI.ticker` the namespace and moved the shared ticker to a static property on `PIXI.Ticker`

ivanpopelyshev:
Oh, I thought we renamed it to `system` but its different. ok, never mind.

PR #5371, opened 2019-01-17
Add canvas tint support to NineSlicePlane. Fixes #3862

### LINKED ISSUE #3862: NineSlicePlane tint not working at CanvasRenderer
NineSlicePlane cannot apply tint for canvas renderer but work fine on WebGL renderer

---------------

Found that #2656 does not applied to CanvasMeshRenderer. It is by design?

TheWyo:
##### Description of change

Adds tint support to NineSlicePlane using CanvasTinter & caching the result, like CanvasSpriteRenderer does. Not sure how v5-legacy is setup for handling canvas but should be easily portable to v5 assuming similar codebase?

##### Pre-Merge Checklist

Visual stuff, so N/A?

### DIFF
--- a/src/mesh/NineSlicePlane.js
+++ b/src/mesh/NineSlicePlane.js
@@ -1,5 +1,5 @@
 import Plane from './Plane';
-import CanvasTinter from '../core';
+import CanvasTinter from '../core/sprites/canvas/CanvasTinter';
 
 const DEFAULT_BORDER_SIZE = 10;
 
@@ -105,14 +105,6 @@ export default class NineSlicePlane extends Plane
         this._bottomHeight = typeof bottomHeight !== 'undefined' ? bottomHeight : DEFAULT_BORDER_SIZE;
 
         /**
-         * Tint to be used on the nine slice.
-         *
-         * @member {number}
-         * @memberof PIXI.NineSlicePlane#
-         */
-        this._tint = 0xFFFFFF;
-
-        /**
          * Cached tint value so we can tell when the tint is changed.
          *
          * @member {number}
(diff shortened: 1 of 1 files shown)


TheWyo:
Oh wait, completely broke it on WebGL, whoops, will fix.

ivanpopelyshev:
I don't like the fact that we added canvas hack in `renderWebGL` method. Its better if we put more if's in `renderCanvas` or `drawSegment`. By the way, current code in `drawSegment` wont work if UV's are not multiplied (`mesh.uploadUvTransform = true`).

Of course we can merge this thing to v4 for a hotfix, but we have to do the same kind of job for v5. You can try it by yourself or wait for me, I'll try to dedicate some time to this PR in weekend.

TheWyo:
Yeah the canvas hack isn't ideal, quick bodge from the fact that objects aren't necessarily tied to any renderer instance, so we don't have any sort of "is this canvas/webgl" aside from the render methods. If managing the UVs isn't too expensive of an operation then yeah maybe doing that inside `_renderCanvas` etc might be a good method.

Not sure what you mean about `drawSegment` and `uploadUvTransform` though? Not noticing any issues locally with the UVs for either true or false, unless there's a scenario I've haven't thought of while testing?

TheWyo:
Oh wait, I see it now. On non-tinted renders. Thought I'd already checked that.

ivanpopelyshev:
There are multiple situations for CanvasRenderer
1. UVs are premultiplied, we need premultiplied
2. UV are not premultiplied, we need premultiplied
3. UV are premultiplied but we need non-premultiplied

Pixi supports 1, you fixed 3, but I also see that 2 is a problem. We can treat it differently - add a flag to _drawSegment whether we use premultiplied uv's or we have to calculate non-premultiplied pixel coords right now.

TheWyo:
Had a go at another idea based on your comment pointing out which types of UVs are needed when.  
When refresh is called, we now store a copy of both pre-mult and non-pre-mult uvs, and pick which of them to use within `drawSegment`. Seems to work much better.

ivanpopelyshev:
That's better.

I'm still sick so, cant test it properly :(

ivanpopelyshev:
I've made new PR based on this one.

PR #5375, opened 2019-01-20
v5 - Fixes Text drop shadows

themoonrat:
Uses the canvas api for applying the drop shadow functionality, rather than adding it via extra draw calls.
This should be quicker to draw, as well as giving more accurate results; fixing issues such as https://github.com/pixijs/pixi.js/issues/4772

### DIFF
--- a/packages/text/src/Text.js
+++ b/packages/text/src/Text.js
@@ -3,7 +3,7 @@ import { Sprite } from '@pixi/sprite';
 import { Texture } from '@pixi/core';
 import { settings } from '@pixi/settings';
 import { Rectangle } from '@pixi/math';
-import { sign, trimCanvas } from '@pixi/utils';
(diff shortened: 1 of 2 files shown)


themoonrat:
It will change visuals, yes, especially if using dropShadowBlur, as that's been broken since it's inception.

I'll get some examples up to show the differences :)

themoonrat:
Before: https://www.pixiplayground.com/#/edit/F~I3c8uur81YX6s7UwUd_
After: https://www.pixiplayground.com/#/edit/9ogclE1WNDeR2W5eBOZsR

englercj:
Would you mind rebasing this onto dev to see if that fixes the travis failure?

bigtimebuddy:
I need to disable that other Ticker test too. It’s not this PR.

PR #5377, opened 2019-01-21
there is no BLEND_MODES.NONE

staff0rd:
As seen [here](https://github.com/pixijs/pixi.js/blob/dev/packages/constants/src/index.js#L63-L83), there is no `BLEND_MODES.NONE`

### DIFF
--- a/packages/constants/src/index.js
+++ b/packages/constants/src/index.js
@@ -59,6 +59,10 @@ export const RENDERER_TYPE = {
  * @property {number} SATURATION
  * @property {number} COLOR
  * @property {number} LUMINOSITY
+ * @property {number} NORMAL_NPM
+ * @property {number} ADD_NPM
+ * @property {number} SCREEN_NPM
+ * @property {number} NONE
  */
 export const BLEND_MODES = {
     NORMAL:         0,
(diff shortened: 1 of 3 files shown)


Dairnon:
Whats the use case for NONE? Isn't NORMAL basically no blendMode?

bigtimebuddy:
Looks like @GoodBoyDigital had originally added that here e416a1ed318904b57731c0e9de5e98f59d3f6d67 and @ivanpopelyshev removed NONE here db0caa656e35b906b6bc8ca93aca07b1adc92b40

I think this is the result of a bad merge from dev -> next and we should re-add `NONE`, assuming it has some original intention. Maybe @GoodBoyDigital could elaborate what was the purpose?

englercj:
> Whats the use case for NONE? Isn't NORMAL basically no blendMode?

None was no blending, normal is actual an Alpha blend.

Dairnon:
@englercj So what would the standard use case of NONE be? Just rendering the entire image with no alpha?

ivanpopelyshev:
Copy pixels instead of blending with blackground.

bigtimebuddy:
Sounds like there's still some utility here. @staff0rd based on this conversation, could you re-add `NONE` to the list of blend modes?

bigtimebuddy:
Thanks @staff0rd

PR #5391, opened 2019-01-27
More blendmodes

ivanpopelyshev:
Ready for review!

Example: https://www.pixiplayground.com/#/edit/Y3HeuBbbllstbzsfp5g8t

There's no `SUBTRACT` blendMode for canvas2d.

Some blendModes require extra `clip` on canvas2d.

Code of CanvasSpriteRenderer is improved, slightly.

Aliases: 

1.  `DST_OUT` aka `ERASE` is the best eraser, it also does not require `clip` on canvas2d!
2. `SRC_OVER` is our default normal blendMode.

PR #5392, opened 2019-01-28
Fix for BLEND_MODES.NONE in StateSystem

bigtimebuddy:
This is an update to #5377

I pulled this out of @ivanpopelyshev's PR #5391 because it's a bug and shouldn't be folded into that new feature PR.

### DIFF
--- a/packages/core/src/state/StateSystem.js
+++ b/packages/core/src/state/StateSystem.js
@@ -1,6 +1,7 @@
 import mapWebGLBlendModesToPixi from './utils/mapWebGLBlendModesToPixi';
 import System from '../System';
 import State from './State';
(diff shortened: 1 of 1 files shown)


packages/core/src/state/StateSystem.js
```diff
@@ -71,10 +72,10 @@ export default class StateSystem extends System
         /**
          * Blend mode
          * @member {number}
-         * @default 17
+         * @default PIXI.BLEND_MODES.NONE
```
rvion: @bigtimebuddy @ivanpopelyshev should be `BLEND_MODES.NONE` instead of `PIXI.` as below ?

PR #5393, opened 2019-01-28
V4.x nineslice tint

### LINKED ISSUE #3862: NineSlicePlane tint not working at CanvasRenderer

ivanpopelyshev:
My version of #5371 by @TheWyo , that fixes #3862.

`drawSegment` removed, and `_renderCanvas` has its own temporary uv's. That calls for more optimizations of this function in the future, like custom logic for roundPixels.

I've made an example that uses both resolution and frame cut, based on another old NineSlicePlane example: [https://www.pixiplayground.com/#/edit/HrETbRGJmES~XExxtdFy~](https://www.pixiplayground.com/#/edit/HrETbRGJmES~XExxtdFy~)

There are two renderers (WebGL and canvas2d, left and right), and I cut nineSlice texture exactly on the middle of the border.

### DIFF
--- a/src/mesh/NineSlicePlane.js
+++ b/src/mesh/NineSlicePlane.js
@@ -108,15 +108,15 @@ export default class NineSlicePlane extends Plane
          * Cached tint value so we can tell when the tint is changed.
          *
          * @member {number}
-         * @memberof PIXI.NineSlicePlane#
+         * @protected
          */
         this._cachedTint = 0xFFFFFF;
 
         /**
          * Cached tinted texture.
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
@TheWyo please confirm that it works for you. Don't worry, your name will appear in credits with a reference on #5371 anyway ;)

PR #5395, opened 2019-02-01
Fixes depthTexture to work properly

Dairnon:
This is the start of  enabling depthTexture support.

This fixes the depthTexture from not being written to at all, and adds a depthResource that is used to fix the internalType needed for webgl2.0

This is a first pass at it, to get it working fully in webgl2.0. Then Ill move to 1.0 + extensions.

The depth resource is also something I'd like to instead be a option we can pass into either Resource itself as a property (internalFormat) or as a property for baseTexture. For now tho it'll have to do.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/framebuffer/Framebuffer.js
+++ b/packages/core/src/framebuffer/Framebuffer.js
@@ -1,5 +1,6 @@
 import Runner from 'mini-runner';
 import Texture from '../textures/BaseTexture';
+import DepthResource from '../textures/resources/DepthResource';
 import { FORMATS, TYPES } from '@pixi/constants';
 
 /**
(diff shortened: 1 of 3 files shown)


englercj:
You can rename the branch with `git branch -m depthResource dev-depth-resource` but pretty sure you will then need to open a new PR with the new name.

englercj:
Tests would be awesome as well.

Dairnon:
@englercj Ah okay, should I just delete this pr and redo it to fix the name? 

As for the tests, I'm familar with providing example demos but am unsure about the standard of how we do tests here. 

I extended BufferResource which appeared to have a test already, so I wasn't sure if I needed to create another for that DepthResource specifically.

bigtimebuddy:
I wouldn't sweat the branch renaming. Not a big deal but in, general, using kabab lower case is preferred.

Dairnon:
Ill work out a webgl1.0 extension version soon too

I just wanted to make sure we got this in 

I am curious, anyone have any ideas on how we should manage the different shader formats between 1 and 2.0? 

How should we go back backwards compatibility between 1.0 + extensions and 2.0 for shaders

ivanpopelyshev:
no need to rename it

Issue #5397, opened 2019-02-01
Add types for the packages


Dairnon:
Currently to import stuff with tree shaking working, we need to do stuff like 

`import { Sprite } from '@pixi/sprite'`

This is fine except that we don't have type support yet, so its hard to figure out which module we are trying to grab. 

It'd be even better if we could get the types to start giving us hints as soon as we write import {
so for example when we type `import { Sp` it auto show us anything with Sp in the name 

If that's not possible at the very least getting types for the `@pixi` part would help a bunch.

- **`pixi.js` version**: _e.g. 5.0.0-rc

englercj:
> It'd be even better if we could get the types to start giving us hints as soon as we write import {
so for example when we type import { Sp it auto show us anything with Sp in the name

That part is completely up to the editor, vscode does this for example.

The idea to solve this would be to generate a module definition for each package, then combine those into typings for the bundles.

bigtimebuddy:
Some context:

There are three build flavors for bundles: ESM, CJS, UMD. People that consume with their browsers will use UMD bundles, and people using bundlers like Webpack and Rollup will consume ESM for the most part, with CJS as a fallback. All the individual sub-packages are only built for ESM and CJS (no browser-supported builds).

Currently, we define the global `PIXI` object in all three flavor for bundles (but not in the sub-packages). I think we should change this and only define `PIXI` for UMD bundles. This would make things a little cleaner, would better support tree-shaking in Webpack/Rollup, but would drop support for importing using `import "pixi.js"` and using APIs via the global `PIXI` object.

All that is to say, I think that we should generate individual package typings, but we might need to think a little differently about the tsd-jsdoc implementation, which relies mostly on namespacing. @englercj , do you have thoughts about supporting module definitions?

englercj:
I definitely want to support module definitions in tsd-jsdoc. The challenge lies in jsdoc having poor support for actually defining them well.

I have some experiments locally, but I need to iterate more to get somewhere I am happy with. If anyone has strong feelings about how it should work, or just knows a syntax they want to use in jsdoc that is very helpful. Please post such things on the [tsd-jsdoc](https://github.com/englercj/tsd-jsdoc/issues) repo.

englercj:
I'm looking into it this weekend.

Problem I am encountering is that jsdoc is bad at parsing es2015. Our hack to fix es6 names also breaks my ability to know if something was export named or default.

[...]

PR #5399, opened 2019-02-02
Removes PIXI global object in ES and CJS bundles to enable tree-shaking

bigtimebuddy:
### Overview

Partially addresses #5397.

This change removes `PIXI` from window object when defined in **es** and **cjs**  bundle builds. It's now _only_ enabled for the browser build. This will help with tree-shaking using bundlers like Webpack and Rollup since everything doesn't hang off a single global object.

#### Changes

* Changes browser build from **umd** format to **iife** format
* Removes `PIXI` global being defined in **es** or **cjs** formats for bundles
* Fixes tests that were still relying on `PIXI`
* Implements an new API to use deprecations for **es** and **cjs** builds

#### Pre-Merge Checklist


bundles/pixi.js-legacy/test/index.js
```diff
@@ -1,8 +1,23 @@
...
-
 describe('PIXI', function ()
 {
+    before(function (done)
+    {
+        this.head = document.querySelector('head');
+        this.script = document.createElement('script');
+        this.script.onload = () => done();
+        this.script.src = require.resolve('../dist/pixi-legacy');
+        this.head.appendChild(this.script);
```
englercj: What is the benefit of putting these on `this` rather than just having them be vars in the describe scope?

bigtimebuddy: No benefit, mostly best practice. In general I like to be able to setup and tear down in mochas conventions

bigtimebuddy:
The PIXI global is set via Rollups “output.name” field.

bundles/pixi.js-legacy/test/index.js
englercj: Sorry I didn't meant to not set them here, just that the vars would be scoped to the describe:

```js
describe(function ()
{
    let head, script;

    before(function () { head = ... });
    after(function () { ... });
});
```

Doesn't matter, just curious.

englercj:
Thanks @bigtimebuddy, I think this is good to go.

PR #5401, opened 2019-02-03
Fix non-batched graphics.

### LINKED ISSUE #5400: Error: WebGL warning: drawElements: Index buffer too small.
### Expected Behavior
Drawing more than ~24 lines with different styles and positions without errors/warnings.

### Current Behavior
No lines are drawn and I get the following console output:
```
00:02:23.977 PixiJS 5.0.0-rc - WebGL 2 - http://www.pixijs.com/ hello.js:51:8
```
then 32 times this line:
```
00:02:24.058 Error: WebGL warning: drawElements: Index buffer too small. GeometrySystem.js:596:16
```
and finally:
```
00:02:24.589 Error: WebGL: No further warnings will be reported for this WebGL context. (already reported 32 warnings) GeometrySystem.js:596:16 
```

### Possible Solution
I wish I had one. :/

### Steps to Reproduce
Draw more than about ~24 lines on a single PIXI.Graphics instance and set a new lineStyle every time. 
Check my fiddle link. When you set the for loop to run only 20 iterations everything will work as expected. When you set it to run 30 times, it won't work.

### Environment

- **`pixi.js` version**: _v5.0.0-rc_
- **Browser & Version**: _Firefox 65.0_
- **OS & Version**: _Fedora 28_
- **Running Example**: https://jsfiddle.net/s3w2L4gj/2/

Edit:

On `PixiJS 4.8.5 ` it seems to work, but I still get an error:
```
01:29:02.849 PixiJS 4.8.5 - WebGL - http://www.pixijs.com/ pixi.js:29115:9
01:29:02.831 Error: WebGL warning: <SetDimensions>: Exceeded 16 live WebGL contexts for this principal, losing the least recently used one. index.js:313:4 
```

ivanpopelyshev:
Fixes #5400

We have a bug in batch group pool.

Fixed fiddle: https://jsfiddle.net/Hackerham/6xvg4L18/

### DIFF
--- a/packages/graphics/src/GraphicsGeometry.js
+++ b/packages/graphics/src/GraphicsGeometry.js
@@ -1,6 +1,7 @@
 import { SHAPES } from '@pixi/math';
 import { Bounds } from '@pixi/display';
 import { BatchGeometry, BatchDrawCall } from '@pixi/core';
+import { DRAW_MODES } from '@pixi/constants';
 
 import GraphicsData from './GraphicsData';
 import buildCircle from './utils/buildCircle';
@@ -600,13 +601,13 @@ export default class GraphicsGeometry extends BatchGeometry
         currentGroup.textureCount = 0;
         currentGroup.start = 0;
         currentGroup.size = 0;
-        currentGroup.type = 4;
+        currentGroup.type = DRAW_MODES.TRIANGLES;
 
         let textureCount = 0;
         let currentTexture = null;
         let textureId = 0;
         let native = false;
-        let drawMode = 4;
+        let drawMode = DRAW_MODES.TRIANGLES;
 
         let index = 0;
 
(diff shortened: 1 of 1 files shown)


packages/graphics/src/GraphicsGeometry.js
```diff
@@ -599,6 +599,8 @@ export default class GraphicsGeometry extends BatchGeometry
 
         currentGroup.textureCount = 0;
         currentGroup.start = 0;
+        currentGroup.size = 0;
+        currentGroup.type = 4;
```
englercj: It took me a while to discover this is actually `WebGLRenderingContext.TRIANGLES`. Can we use the real constants, or at least named constants that proxy the values? I see that the values `1` and `4` are used in a few places without names too. The type in BatchDrawCall isn't documented either which made this very difficult to discover.

bigtimebuddy: Good point 👍

ivanpopelyshev: It was like that before I edited it. OK, fixed :)

PR #5403, opened 2019-02-03
NineSlice tint and protected field renames

ivanpopelyshev:
#5393 for V5.

https://www.pixiplayground.com/#/edit/NZnBmx42BZkvi1ufdK6sT

I also renamed sprite tint, so example also contains tinted Sprite and tinted TilingSprite.

### DIFF
--- a/packages/canvas/canvas-mesh/src/NineSlicePlane.js
+++ b/packages/canvas/canvas-mesh/src/NineSlicePlane.js
@@ -1,4 +1,29 @@
+import { CanvasTinter } from '@pixi/canvas-renderer';
 import { NineSlicePlane } from '@pixi/mesh-extras';
+
+/**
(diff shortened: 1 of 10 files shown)


packages/canvas/canvas-mesh/src/NineSlicePlane.js
```diff
@@ -11,14 +36,56 @@ import { NineSlicePlane } from '@pixi/mesh-extras';
...
+    }
 
     context.globalAlpha = this.worldAlpha;
     renderer.setBlendMode(this.blendMode);
 
-    const transform = this.worldTransform;
-    const res = renderer.resolution;
-
-    if (this.roundPixels)
+    if (renderer.roundPixels)
```
themoonrat: v5 has a 'roundPixels' flag per display component, not on the renderer

PR #5419, opened 2019-02-09
Added option to update anchor with defaultAnchor when frame changes.

jramstedt:
##### Description of change

Each texture in animation can have different defaultAnchor.
This change allows AnimatedSprite's anchor to be updated on each frame.

##### Pre-Merge Checklist

### DIFF
--- a/src/extras/AnimatedSprite.js
+++ b/src/extras/AnimatedSprite.js
@@ -90,10 +90,15 @@ export default class AnimatedSprite extends core.Sprite
         /**
          * Update anchor to [Texture's defaultAnchor]{@link PIXI.Texture#defaultAnchor} when frame changes.
          *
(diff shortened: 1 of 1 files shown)


themoonrat:
lgtm - only query is whether `animateAnchor` is the right name for this. I'm 60% happy with it, but it's not exactly animating the anchor, just updating it. I'd definitely add to the docs that if you enable this, whatever it's called, that it effects the ability to use the normal `.anchor` property

bigtimebuddy:
I think the piece for me that's missing from your documentation description for the property is _why_ you'd want to use this opt-in property, what are the use-cases for adding this vs why would you want to leave it off.

Also, similar to @themoonrat, I don't think `animateAnchor` is the best name. I'd be more inclined to use `updateAnchor` since it happens on `updateTexture`. Regardless whether it's actually "animating" or not.

jramstedt:
Did some changes based on your comments.

PR #5422, opened 2019-02-09
Mipmap pow2 compatibility

### LINKED ISSUE #5407: Different results of scaling between 5.0.0rc and 4.8.5
Hello again ! I updated to the pre-release 5.0.0rc and I figured out that it was more blurry. I already had problem with blur before but it was not a bug (just something i did wrong).

But now, I just updated to 5.0.0rc (and replace the name of methods) and some part of the game is now more blur.

Theses objects become blurry when I resize an object by 50% (scale.x = 0.5 / scale.y = 0.5).

=> I use the SCALE_MODES.LINEAR

### Expected Behavior
I scale down to 50% my object and it looks "nice"

### Current Behavior
I scale down to 50% my object and it looks "blurry"

### Possible Solution
I supposed it was a problem with PIXI.settings.SCALE_MODE, but even by forcing the LINEAR, it gets more blurry on the 5.0.0 version.

### Steps to Reproduce
My game was in Flash and I remade it in JS using Pixi.JS, so to have good effects when I need to scale up ( doing scale x 2) on my texture. I decided to have all my texture in scale x2 (so my raw .PNG are two times bigger than the size of the object on Flash).

Finally to have the correct size, I scaled down x 0.5. It was perfect on 4.8.5 but now it's blurry on 5.0.0rc.

I tried to reproduce it by using a x2 image of the bunny.png

http://93.113.206.174/pixi485_2.html
http://93.113.206.174/pixi500_2.html

We can see that the bunny from 5.0.0rc is one pixel bigger but it's only in the renderer, when I do the localBounds it has the same size on both.

```
// 4.8.5
	PIXI.settings.ROUND_PIXELS = true;
	PIXI.settings.SCALE_MODE = PIXI.SCALE_MODES.LINEAR;

	var renderer = new PIXI.autoDetectRenderer(160, 190,  { transparent: true, resolution: 1 });
    var bunny = new PIXI.Container();

    document.getElementById("renderer1").appendChild(renderer.view);
	let tmpTexture = new PIXI.Sprite(PIXI.Texture.fromImage("bunny2.png"));
	bunny.addChild(tmpTexture);
	
	  var inn = setInterval(function (r) { renderer.render(bunny); }, 1000 / 24);

	bunny.scale.x = 0.5;
	bunny.scale.y = 0.5;

// 5.0.0
	PIXI.settings.ROUND_PIXELS = true;
	PIXI.settings.SCALE_MODE = PIXI.SCALE_MODES.LINEAR;
	
	var renderer = new PIXI.Renderer(160, 190,  { transparent: true, resolution: 1 });
    var bunny = new PIXI.Container();

    document.getElementById("renderer1").appendChild(renderer.view);
	let tmpTexture = new PIXI.Sprite(PIXI.Texture.from("bunny2.png"));
	bunny.addChild(tmpTexture);
	
	  var inn = setInterval(function (r) { renderer.render(bunny); }, 1000 / 24);

	bunny.scale.x = 0.5;
	bunny.scale.y = 0.5;

```

bunny2

You can see how it is more blurry on the 5.0.0 and it's taller. 

### Environment
Firefox & Chrome last version on Windows

On Edge it's the contrary, it's better (less blurry) on the 5.0.0rc than the 4.8.5

ivanpopelyshev:
Fixes #5407

This is compatibility PR, that way webgl1 and webgl2 produce the same result. If someone want non-pow2 texture to have mipmaps, they'll have to specify it.

```js
baseTexture.mipmap = PIXI.MIPMAP_MODES.ON;
```

The values are compatible with booleans in v4. `false` is `OFF=0`, `true` is `POW2=1`.

Demo: https://www.pixiplayground.com/#/edit/jk06IGwLU1Q9yT41pv7_j

V4: https://www.pixiplayground.com/#/edit/BuGK_ogiBOqaXCsSHyCe7

First and third colums are the same in V4 and V5, second is `mipmap=ON` for non-pow2 texture. As you see, webgl2 mipmaps are almost the same as webgl1 for enlarged textures.

I propose we use `POW2` mode by default for compatibility, right column for pow2 textures and left for everything else, like in v4.

image

### DIFF
--- a/packages/constants/src/index.js
+++ b/packages/constants/src/index.js
@@ -257,9 +257,9 @@ export const WRAP_MODES = {
 };
 
 /**
- * The wrap modes that are supported by pixi.
+ * Mipmap filtering modes that are supported by pixi.
  *
- * The {@link PIXI.settings.MIPMAP_TEXTURES} affects the texture filtering.
+ * The {@link PIXI.settings.MIPMAP_TEXTURES} affects default texture filtering.
  * Mipmaps are generated for a baseTexture if its `mipmap` field is `ON`,
  * or its `POW2` and texture dimensions are powers of 2.
  * Due to platform restriction, `ON` option will work like `POW2` for webgl-1.
--- a/packages/constants/src/index.js
+++ b/packages/constants/src/index.js
@@ -270,9 +270,9 @@ export const WRAP_MODES = {
  * @memberof PIXI
  * @static
  * @enum {number}
- * @property {number} OFF - The textures uvs are clamped
- * @property {number} POW2 - Only pow2
- * @property {number} ON - Texture has mipmaps
+ * @property {number} OFF - No mipmaps
+ * @property {number} POW2 - Generate mipmaps if texture dimensions are pow2
+ * @property {number} ON - Always generate mipmaps
  */
 export const MIPMAP_MODES = {
     OFF: 0,
--- a/packages/constants/src/index.js
+++ b/packages/constants/src/index.js
@@ -257,6 +257,30 @@ export const WRAP_MODES = {
 };
 
 /**
+ * The wrap modes that are supported by pixi.
+ *
+ * The {@link PIXI.settings.MIPMAP_TEXTURES} affects the texture filtering.
+ * Mipmaps are generated for a baseTexture if its `mipmap` field is `ON`,
+ * or its `POW2` and texture dimensions are powers of 2.
+ * Due to platform restriction, `ON` option will work like `POW2` for webgl-1.
+ *
+ * This property only affects WebGL.
+ *
+ * @name MIPMAP_MODES
+ * @memberof PIXI
+ * @static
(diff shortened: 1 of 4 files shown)


ivanpopelyshev:
Updated example. WTH is going on here?

ivanpopelyshev:
Now, I need @GoodBoyDigital decision, whether we keep old behaviour as default, that's the same on webgl1 and webgl2, or we actually go for `webgl2 is different even on small examples` route.

packages/constants/src/index.js
```diff
@@ -256,6 +256,30 @@ export const WRAP_MODES = {
     MIRRORED_REPEAT: 33648,
 };
 
+/**
+ * The wrap modes that are supported by pixi.
```
englercj: Description needs an update here.

ivanpopelyshev: thanks!

PR #5298, opened 2018-12-19
cacheAsBitmap fix bounds calculation

ivanpopelyshev:
I hate cacheAsBitmap.

https://www.pixiplayground.com/#/edit/T6MYvWjSO9EXakjVaTHSF

Try add more velocity and see how circle is cropped even more.

PR #5423, opened 2019-02-09
[v4] cacheAsBitmap fix calculateBounds problem

### LINKED ISSUE #5416: cacheAsBitmap makes container 'jump'
This issue might intersect with some other cacheAsBitmap issues, but I've got a workaround and working example of it, just in case this will help someone.

If we have a PIXI.Container with some stuff in it, and it's bounds are not integers, and all of it's content positioned somewhere far from 0:0 coordinates, toggling cacheAsBitmap will make this container 'jump': move some pixels up and left. The farther from 0:0 is content, the farther it will jump.
Bug can be seen through comparing bounds of the container before and after cacheAsBitmap toggling. Only in Canvas mode, though: in WebGL mode bounds will be the same, even though the jump is obvious.

Problem seen in pixi.js PixiJS 4.8.1 on both WebGL and Canvas renderers and in Chrome and Firefox browsers for Windows 7.

There is a workaround for this bug: before setting cacheAsBitmap add something to 0:0 of the container. 1x1 pixel transparent Graphics rectangle would do.

Full exmaple with a workaround and explainations:
https://codepen.io/takopus/pen/NoXGMv

ivanpopelyshev:
Fix #5416.

Example by @takopus.

Crash Bug, 4.x: https://www.pixiplayground.com/#/edit/UFD711FVtXRvwA~8mIQie 

Press button 4 times, and it'll crash.

Shift Bug, 4.8.1: https://www.pixiplayground.com/#/edit/rIVogz0N~JfQQKR3ItYUc

Press button, sprite shifts.

Fixed: https://www.pixiplayground.com/#/edit/8~rAjGg3Bgicd9o050tKs

Old tests from #5298: 

https://www.pixiplayground.com/#/edit/jg7WEsoNFL1Cy55HB9f69 - shift in 4.8.1

https://www.pixiplayground.com/#/edit/T6MYvWjSO9EXakjVaTHSF - was cut in 4.8.1

https://www.pixiplayground.com/#/edit/~vUkTrNV9CYlxgR_lS_Ue - wrong bounds in 4.8.1

Everything passes on my comp, both Canvas and WebGL.

Also need to cherry-pick it to v5, file looks the same.

I marked it as High, because cacheAsBitmap crashes on CanvasRenderer. 

Lines 84 and 110 are real changes, the rest are refactoring so canvas behaves the same way as webgl.

### DIFF
--- a/src/extras/cacheAsBitmap.js
+++ b/src/extras/cacheAsBitmap.js
@@ -28,6 +28,7 @@ class CacheData
         this.originalRenderWebGL = null;
         this.originalRenderCanvas = null;
         this.originalCalculateBounds = null;
+        this.originalCalculateBounds2 = null;
         this.originalGetLocalBounds = null;
 
         this.originalUpdateTransform = null;
@@ -81,7 +82,8 @@ Object.defineProperties(DisplayObject.prototype, {
                 data.originalRenderCanvas = this.renderCanvas;
 
                 data.originalUpdateTransform = this.updateTransform;
-                data.originalCalculateBounds = this._calculateBounds;
+                data.originalCalculateBounds = this.calculateBounds;
+                data.originalCalculateBounds2 = this._calculateBounds;
                 data.originalGetLocalBounds = this.getLocalBounds;
 
                 data.originalDestroy = this.destroy;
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
I've made small refactor - Canvas behaviour the same way as WebGL. I'll make better test for it.

PR #5425, opened 2019-02-09
Rename `renderTexture` field to `current`, like in FramebufferSystem

ivanpopelyshev:
subj.

Proof that filters work: https://www.pixiplayground.com/#/edit/ZLTblGp99em2n1fqT~NGF

### DIFF
--- a/packages/core/src/filters/FilterSystem.js
+++ b/packages/core/src/filters/FilterSystem.js
@@ -203,7 +203,7 @@ export default class FilterSystem extends System
 
         if (filterStack.length === 1)
         {
-            this.defaultFilterStack[0].renderTexture = renderer.renderTexture.renderTexture;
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
I'm a little torn about merging this because I'm not sure if we should provide a deprecation for this property. Technically we don't _need_ to because this is a pre-v5 release, but it can't hurt. @themoonrat, what do you think? I generally assume @ivanpopelyshev is a little more bullish on these types of changes.

ivanpopelyshev:
I thought its safe because no one in community actually used it except @Dairnon and me, its renderer-level field. Sorry, I forgot to write it here :)

themoonrat:
My thoughts on this are: if a property is new to v5, we should feel free to rename without consideration to depreciation or compatibility until we hit the first official release.

It's hard enough maintaining that through official versions, let alone trying to do it in alphas too

PR #5430, opened 2019-02-11
V4.x reset textures

### LINKED ISSUE #5427: Renderer.reset is breaking things
**Is it only supposed to reset the webGL state only ?** 
I suspect this method omits to reset some states ? and is breaking things.

Running renderer.reset in a loop before getting things drawn on screen is breaking things

PixiJS 4.8.5 :
https://jsfiddle.net/johndordman/e7ajcfgy/4/
Returns an error " Source and destination textures of the draw are the same "
The example here is trying to render thing in a framebuffer and on screen, but there might be a confusion in the framebuffer bindings using renderer.reset ?

PixiJS 5.0 :
https://jsfiddle.net/johndordman/x4nvwfaz/17/
Returns no error, but don't show anything on screen as well, just the blue background but not the sprite.

Removing the renderer.reset line makes each of them working.

**Expected render :** 

capture d ecran 2019-02-10 a 13 35 04

ivanpopelyshev:
Fixes #5427.

### DIFF
--- a/src/core/renderers/webgl/WebGLRenderer.js
+++ b/src/core/renderers/webgl/WebGLRenderer.js
@@ -241,6 +241,13 @@ export default class WebGLRenderer extends SystemRenderer
         this.boundTextures = new Array(maxTextures);
         this.emptyTextures = new Array(maxTextures);
 
+        /**
+         * Did someone temper with textures state? We'll overwrite them when we need to unbind something.
+         * @member {boolean}
+         * @private
+         */
+        this._unknownBoundTextures = false;
+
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
https://www.pixiplayground.com/#/edit/JZaMtwXD03kXhoCvOx_55

ivanpopelyshev:
Here's the old PixiJS and ThreeJS demo that works: https://www.pixiplayground.com/#/edit/AGe~Bx697qK5yiXDWHkWT

PR #5431, opened 2019-02-11
Make `reset` work like in v4.

### LINKED ISSUE #5427: Renderer.reset is breaking things

ivanpopelyshev:
It started in #5427, then I found out that we dont have enough `reset` methods.

IT WORKS! https://www.pixiplayground.com/#/edit/h9BQpJ3scsslkT~RO2JkM

And now, the old pixiJS-threeJS demo, but for V5 and ThreeJS 101. https://www.pixiplayground.com/#/edit/sPigqH5G1ogJ_5pocOs42

### DIFF
--- a/packages/core/src/textures/TextureSystem.js
+++ b/packages/core/src/textures/TextureSystem.js
@@ -149,6 +149,11 @@ export default class TextureSystem extends System
         }
     }
 
+    /**
+     * Resets texture location and bound textures
+     *
+     * Actual `bind(null, i)` calls will be performed at next `unbind()` call
+     */
     reset()
     {
         this._unknownBoundTextures = true;
--- a/packages/core/src/textures/TextureSystem.js
+++ b/packages/core/src/textures/TextureSystem.js
@@ -1,4 +1,5 @@
 import System from '../System';
+import BaseTexture from './BaseTexture';
 import GLTexture from './GLTexture';
 import { removeItems } from '@pixi/utils';
 import { WRAP_MODES } from '@pixi/constants';
@@ -45,7 +46,14 @@ export default class TextureSystem extends System
          * @member {boolean}
          * @private
          */
-        this._unknownBoundTextures = false;
+        this._undefinedBoundTextures = false;
+
+        /**
+         * BaseTexture value that shows that we don't know what is bound
+         * @member {PIXI.BaseTexture}
+         * @readonly
+         */
+        this.undefinedTexture = new BaseTexture();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


ivanpopelyshev:
@GoodBoyDigital our renderer calls `renderTexture.bind(undefined)` while resize calls `renderTexture.bind(null)`, that saved us from resize bug, but I better guard `bind` method.

ivanpopelyshev:
I think it works

ivanpopelyshev:
I think we need a demo of usage of both threejs and pixijs here. I'll search it in my stash.

ivanpopelyshev:
Shoutout for @izar89 for that old demo.

ivanpopelyshev:
The meaning is "we dont know what is bound there". I propose `unknown` or `undefined`, but lets wait for @GoodBoyDigital

englercj:
Is there a reason we can't use `null` as the "unknown" value? If not, I'd definitely want it to be called "empty" or "unknown", not "undefined".

ivanpopelyshev:
@englercj That's right question. We already have null's for "empty texture is bound" and "main framebuffer is bound". I'll change it to `unknown`.

packages/core/src/textures/TextureSystem.js
```diff
@@ -39,6 +40,20 @@ export default class TextureSystem extends System
...
+         * @private
+         */
+        this._undefinedBoundTextures = false;
+
+        /**
+         * BaseTexture value that shows that we don't know what is bound
+         * @member {PIXI.BaseTexture}
+         * @readonly
+         */
+        this.undefinedTexture = new BaseTexture();
```
GoodBoyDigital: can we just use `emptyTextures` here?

ivanpopelyshev: We can do it like in v4, but it wont be clear that we use assignment in `reset()` to fake unknown bound textures and not to rebinding everything.

PR #5432, opened 2019-02-12
RenderTextureSystem invalidation problem

ivanpopelyshev:
We should perform all renderTexture checks even if its the same as we bound before. Our `renderer.resize` worked only because `renderer.render` passed `undefined` in this parameter, while `resize` was using `null`.

Try render null's: https://www.pixiplayground.com/#/edit/UomPY6pRxm6qncEEWxl2m
Rendertexture resize after render bug: https://www.pixiplayground.com/#/edit/dCYHcHa~UaegyTf9jDlr2

Those cases are rare, but its better if we cleanup the code.

`current` check is moved to framebuffer.

If we bind the renderTexture second time in a row, there's only one downside of this PR: pixi will check global uniforms projectionMatrix and reupload it. It wont rebind the framebuffer, and it wont call `gl.viewport`.

### DIFF
--- a/packages/core/src/framebuffer/FramebufferSystem.js
+++ b/packages/core/src/framebuffer/FramebufferSystem.js
@@ -78,15 +78,17 @@ export default class FramebufferSystem extends System
     {
         const { gl } = this;
 
-        this.current = framebuffer;
-
         if (framebuffer)
         {
             // TODO caching layer!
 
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Another case, now its a renderTexture: https://www.pixiplayground.com/#/edit/dCYHcHa~UaegyTf9jDlr2 

Uncomment any rt reset operation and it'll work.

I don't know how to fix it yet.

PR #5441, opened 2019-02-14
Remove ternary operator

Ewan-Roberts:
##### Description of change

Removed a slight complexity from the SetTransform function. Maybe a little cleaner 🤷‍♂️ ?

(Love your work btw guys!) 

##### Pre-Merge Checklist

### DIFF
--- a/packages/display/src/DisplayObject.js
+++ b/packages/display/src/DisplayObject.js
@@ -385,28 +385,28 @@ export default class DisplayObject extends EventEmitter
     /**
      * Convenience function to set the position, scale, skew and pivot at once.
      *
-     * @param {number} [x=0] - The X position
-     * @param {number} [y=0] - The Y position
-     * @param {number} [scaleX=1] - The X scale value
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@Ewan-Roberts, thanks for contributing. We appreciate all PRs big and small. 

For us, it's important that contributors are explicit about _why_ they are changing something. It's not clear to me what problem you're trying to solve by your description. There's no significant performance implications by removing the ternary operator so I'd assume most of what you're doing is just stylistic.

Also, as far as your change goes, there are some non-starting issues:

* Tests to not pass on Travis
* Did you mean to use `||` (or) not `|` (bitwise)? Because you changed the meaning here.
* We'd like to maintain default values in documentation and in the argument, that's used throughout PixiJS and actually add _less_ clarity for those looking at the code or documentation.

Unfortunately, for these reasons, I'm going to close this PR. Thank you for your suggestion.

PR #5442, opened 2019-02-14
Adaptive curve formulas fix

ivanpopelyshev:
Because of bug in formula, curves that had big coords (>100) were divided by 2048 small segments.

Because of another bug in `pixi-const` v5 had adaptive curves enabled by default. Gladly, adaptive mode wasn't enabled by default in pixi-v4.

When we reviewed #4658 everyone, including the math guy (me), somehow didnt see obviously wrong equation. Obviously wrong because of `(cpX - 2.0)` , a constant subtracted from coordinate.

This PR fixes parenthesis order, that @icosaeder put there to fit it in our lint "no-mixed-operators" rule.

Demo has more than 32 arcs, that's more than 65536 vertices => 16-bit index buffer fails. We can see how its failing exactly on 32-th arc.

Demo 5.0.0-rc: https://www.pixiplayground.com/#/edit/9KVY7oZCiJ7lUG4Zx3bID

Demo this branch: https://www.pixiplayground.com/#/edit/5o26V7Jn9yCl28wub4mYI

Good:

image

Broken:

image

Low-poly:

image

### DIFF
--- a/packages/graphics/src/const.js
+++ b/packages/graphics/src/const.js
@@ -14,13 +14,13 @@
  * @property {number} maxSegments=2048 - maximal number of segments in the curve (if adaptive = false, ignored)
  */
 export const GRAPHICS_CURVES = {
-    adaptive: false,
+    adaptive: true,
     maxLength: 10,
     minSegments: 8,
     maxSegments: 2048,
     _segmentsCount(length, defaultSegments = 20)
     {
-        if (this.adaptive)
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
I feel like we need to move those settings inside Graphics or even in separate curves settings class , that way multiple graphics elements can have one setting. Of course we have to do it in separate PR.

PR #5444, opened 2019-02-14
SpriteMask filterArea 1-frame lag

ivanpopelyshev:
Thanks to @bigtimebuddy for the demo! It works on both v4 and v5.

https://pixiplayground.com/#/edit/0FMtqLQIUyNB91SKORFeL

PR #5445, opened 2019-02-14
[v4.x] mask filterArea 1-frame lag

ivanpopelyshev:
#5444 copy for v4.

https://pixiplayground.com/#/edit/0FMtqLQIUyNB91SKORFeL


bigtimebuddy:
Merging because #5444 was approved, this is a back-port of that fix.

PR #5449, opened 2019-02-16
Best Performance in AnimatedSprite loops

rtalon83:
##### Description of change

The best way of work with 'for' is, first, save the array length then use it. 
Here a test: http://jsben.ch/F9kXl

To me, sometime I use large texture atlas with hundreds to small texture (100, 200, 1000).

##### Pre-Merge Checklist


ivanpopelyshev:
Remove packagelock and lets do merge it.

Never mind that `Texture.from` gets a value from map which is supposed to be much heavier operation ;)

rtalon83:
Solved! I did not see it myself

themoonrat:
Style wise, doesn't it keep it neater doing something like
`for (let i = 0, length = frames.length; i < length; ++i)`
And then the only diffs each time is a change to the for loop, rather than keeping a variable outside that scope each time?

In practical usage, I very much doubt that this change is going anything? Yes it shows as 5% quicker on chrome for me, but we're already in the realm of being able to do many millions of operations per second. So 5% off that has no practical benefit?

If you go down this route here, that the benefits are worth it, then there are much more 'critical' places to do this throughout the libs, surely?

rtalon83:
** Style wise, doesn't it keep it neater doing something like
`for (let i = 0, length = frames.length; i < length; ++i)`
And then the only diffs each time is a change to the for loop, rather than keeping a variable outside that scope each time?**

- I think that is expensive to access each time to **array.length**
- Yes, your style is better 

**If you go down this route here, that the benefits are worth it, then there are much more 'critical' places to do this throughout the libs, surely?**

Yes, but in this moment I using AnimatedSprite in old devices and my goal is the best performance from wherever.

ivanpopelyshev:
> Yes, but in this moment I using AnimatedSprite in old devices and my goal is the best performance from wherever.

That didnt help me on old devices. Using polygon packing may help, because of less calls of fragment shader, but its available only in v4 with pixi-heaven. I dont believe for's are bottleneck there.

In general, yeah, I like style with `len` better. I just dont participate in those kind of PR's , I lurk for bigger stuff.

djmisterjon:
it called micro-optimisations, and for a big game it important.
image
i alrealy hack all loops like thats in animated sprite, spine, pixi-display..
Am in favor of this.

A good tricks it to add it to your IDE, vscode ex:
image
so it write it automatically when you need loop.

the user code for vscode
```javascript
	 "for loop l<-": {
		"prefix": "for i<l",
		"body": [
		"for (let i=0, l=${1:array}.length; i<l; i++) {",
			"\t$0const e = ${1:array}[i]",
		"};",
		"\t$0"
		],
		"description": "For Loop"
	},
```

ivanpopelyshev:
please return package.lock to the state it was before commit :) I didn't mean to actually remove the file.

englercj:
Micro-benchmarks are so dangerous. I'm willing to bet that this has never shown up on a profiler as needing this kind of change. I don't like theoretical "optimizations".

The difference here is 40ms when iterating 1k frames 1k times. Which means this improvement is 0.00004ms per iteration, or for your case of having 1k frames 0.04ms. Also, the tests use eval, which does an awesome job of messing with the optimizer, preventing it from doing its job well. So who knows if this is even a benefit at all in optimized code?

So *best case* scenario you drop 0.04ms of a process that surely takes hundreds of milliseconds or even seconds (downloading, decoding, and parsing the texture). Though I'm willing to bet you never get that either because in the wild it won't actually be any different to the optimizer. And to get there we make the code less readable, and more prone to error.

If you can show me on a flame graph that this optimization was your bottleneck and it actually did increase your performance in a meaningful way, then I'd be happy. But as it stands this looks like a simple case of pre-optimization that doesn't bring any real-world benefit along with it.

bigtimebuddy:
@rtalon83, thank you for your contribution and being willing to help make PixiJS better. It's great to see first-time contributors.

That said, I agree with @englercj's thorough explanation above. This is not a significantly meaningful performance optimization and unfortunately we need to decline and close this PR. 

If you are passionate about performance and optimization, it would be incredibly useful to have folks like yourself profile real-world rendering conditions. We are always on the hunt for pushing Pixi to be faster, but we want to make sure that we are looking broadly at performance and JavaScript optimizations across all browsers. Here are some ideas of projects we'd love to see the community tackle:
* The core Pixi team has talked about finding reliable ways to measure performance regressions in PixiJS in an automated way (e.g. this version or change showed a X% drop or gain in performance). If someone can crack that nut, it would really help the library as a whole. 
* Text rendering is incredibly slow, finding ways to improve performance here would also be tremendous. 
* Lastly,  since we moved from ES5 to ES6 code-base, we started relying on transpiling via Babel. We know this came at a performance cost, but maybe there are style improvements, linting changes, or tool settings we could tinker with to improve the codebase broadly.
* Also, consider opening an issue and ask the other core members like @themoonrat, @GoodBoyDigital, @ivanpopelyshev, @englercj what types of performance projects they think would benefit PixiJS.

rtalon83:
I agree with you @bigtimebuddy. I like the optimizations and performance in JavaScript app. I Will think how I can help you. Maybe, the render Text could be my first way.
Thanks!

PR #5450, opened 2019-02-16
Fix typings

ivanpopelyshev:
Yes, I know that VideoResource Promise passes VideoResource and not void, we can deal with that later, that's a problem with overrides.

Yes, we can move `IFontMetrics` it to namespace TextMetrics if you guys want it.

`update(deltaTime)` has default param value because of override issue.

### DIFF
--- a/packages/ticker/src/TickerListener.js
+++ b/packages/ticker/src/TickerListener.js
@@ -11,7 +11,7 @@ export default class TickerListener
      * Constructor
      * @private
      * @param {Function} fn - The listener function to be added for one update
-     * @param {Function} [context=null] - The listener context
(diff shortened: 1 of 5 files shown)


packages/ticker/src/TickerListener.js
```diff
@@ -27,7 +27,7 @@ export default class TickerListener
         /**
          * The calling to execute.
          * @private
-         * @member {Function}
+         * @member {object}
```
englercj: These should be `any` or `*`. When you use `object` it means `{}` specifically.

ivanpopelyshev: OK

PR #5457, opened 2019-02-19
[v4.x] Fix Matrix decomposition for a rare case.

ivanpopelyshev:
Maybe long time ago there was something about scale. Now this branch is triggered only in one case and returns wrong answer.

Can be cherry-picked.

https://www.pixiplayground.com/#/edit/tbHFSNFRW4KCULGpdbygB


bigtimebuddy:
Looks like that was added here in 4.0.0-rc3
https://github.com/pixijs/pixi.js/commit/4bf3f687b0a00d6d00b1186e3da58507e1f0273a

ivanpopelyshev:
That's the only way to hit that if.

PR #5462, opened 2019-02-20
Fixes incomplete path

rtalon83:
#### Description of change
When you want to draw a star using PIXI.Graphics.drawStar + LineStyle the line-path is incomplete because there is a point missing.

`closedPath = false`

Bug here:
https://www.pixiplayground.com/#/edit/Ytz80nX7aREstArPfAkV3

#### Pre-Merge Checklist


ivanpopelyshev:
Here is where I fixed it for all other shapes: https://github.com/pixijs/pixi.js/pull/5453/files#diff-e646f77e0abdeecb59f2e3c36767ef35R60

For consistency, you can just modify the same line. 

I'll rewrite all those checks later, its better to handle it in the same place.

rtalon83:
perfect @ivanpopelyshev

rtalon83:
I don't understand your solution @ivanpopelyshev , same line? I just saw that the last point for "Star" is missing. Do you want manage this case in `buildLine.js`?
Then, I think that is better I waiting your changes.
Thanks

bigtimebuddy:
I asked @GoodBoyDigital to take a quick look at this. I want to make sure we are fixing the right issue. It’s my expectation that drawing any arbitrary polygon should close the shape automatically with an extra point. Will hold of merging until we can look more closely.

ivanpopelyshev:
No, its true only for non-polygon shapes. We dont need autoClose on general polygons that user specified for us.

bigtimebuddy:
Seems like `closePath` doesn't work, which looks like a bigger issue:
https://www.pixiplayground.com/#/edit/BDm_4pTY_4DUR3X1XvzSV

ivanpopelyshev:
again? oh.

bigtimebuddy:
Closing this in favor of #5516.

Thank you @rtalon83 for making the PR and for helping us tidy up v5.

PR #5463, opened 2019-02-20
Fallback for `new function` eval

### LINKED ISSUE #5459: Can you support WeChat?
WeChat is one of the most popular h5 game platforms in China, but pixi.js v5 does not support WeChat games, new Function cannot be successfully executed.

ivanpopelyshev:
Fixes #5459

The only question I have is should I move it in a package separate from `core`?

Test: https://www.pixiplayground.com/#/edit/C0Ml3PATSbhq1NNmadFA~

How did I make it: I took old version from `pixi-gl-core`, added `cv` parameter in all those small functions, copied `generateUniformsSync` and removed all template usage from it through several replaces.


packages/core/src/settings.js
```diff
@@ -16,4 +16,6 @@ import { ENV } from '@pixi/constants';
  */
 settings.PREFER_ENV = ENV.WEBGL2;
 
+settings.JS_NEW_FUNCTION_AVAILABLE = true;
```
bigtimebuddy: Remove?

packages/settings/src/settings.js
```diff
@@ -222,4 +223,16 @@ export default {
...
+     * PixiJS generates some efficient webgl-related functions, but not all platforms support it.
+     * Default value is auto-detected.
+     *
+     * @static
+     * @constant
+     * @memberof PIXI.settings
+     * @type {boolean}
+     * @default true
+     */
+    CAN_GENERATE_NEW_FUNCTION: canGenerateNewFunction(),
```
bigtimebuddy: I don’t quite see the use case for another setting here. Just use the feature detection. The use case where someone needs to set this explicitly seem so remote. And in that case they could override canGenerateNewFunction.

packages/core/src/settings.js
ivanpopelyshev: oops

packages/settings/src/settings.js
ivanpopelyshev: > And in that case they could override canGenerateNewFunction.

We already have mobile detection and other stuff detection.

Usecase - debug of the generator. Also I have several ideas for other generators, so better use it in one place.

ivanpopelyshev: I can put it in `utils` :)

bigtimebuddy: I disagree with your rationale here. I feel like you often add more than is necessary. Just solve the original problem as succinctly as possible. When you introduce new generators in the future, then we can discuss on adding a debug flag then. Just cache this as a private variable. No new unnecessary settings or APIs for v5.

bigtimebuddy:
Can we talk about if this is the right solution to this problem?

The issue this PR is trying to solve is support for WeChat which disallows the use of Function constructor, for whatever reason. There are cases in PixiJS where have decided to not support all browser, environments or platforms. In some cases this can be achieve with a polyfill (requestAnimationFrame), in some cases it opt-in (pixi-shim for node.js),  in some cases we built it into the library (mobile/touch), and other times we just decide to not include it (most recently we dropped support for CocoonJS).

Because `new Function` is widely supported on all browsers, I'm not sure the best approach is to provide built-in, out-of-the-box support. Especially when this PR creates a decent amount of extra code that needs to be maintained and which will never run in the majority of browsers.

I'm guess I'm leaning towards providing a workaround to this problem by mixing-in an external self-contained solution. Possibly a plugin that overrides `PIXI.systems. ShaderSystem.prototype.syncUniformGroup`. This solution is more brittle, but feels like the best solution for this edge-case platform.

Thoughts?

ivanpopelyshev:
That's exactly what I thought. I just wanted to provide a workaround ASAP :) Now we can discuss where to put this thing.

englercj:
I agree with @bigtimebuddy. Especially with v5 having good plugin support I like the idea of having a `@pixijs/wechat`-esque plugin to support deficiencies in those particular environments.

bigtimebuddy:
My suggestion is that we replace the bits in `syncUniformGroup` with with single public function that can be easily overridden. Then, we should take the `staticUniformsSync` and create a new external plugin that is opt-in (e.g., `@pixi/mixin-wechat`), importing this package should apply the feature-detecting and override the prototype method on `PIXI.systems.ShaderSystem`.

```js
syncUniforms(group, glProgram, renderer) {
    const syncFunc = group.syncUniforms[this.shader.program.id] || this.createSyncGroups(group);

    syncFunc(glProgram.uniformData, group.uniforms, this.renderer);
}
```

If there are no concerns about this direction, I believe we should close this PR.

ivanpopelyshev:
@bigtimebuddy Lets at least leave feature detection in core and notify the user that he needs different build of pixi.

englercj:
But you can't make the decision at runtime which build you need, so the feature detection isn't actually useful unless you already need the other bundle.

ivanpopelyshev:
Then we have to add third bundle. It takes extra time and effort to explain Chinese devs all those nuances. We need that warning to be as noticeable as possible.

englercj:
> Then we have to add third bundle.

No, if people need the mixin, then they include the mixin. We don't need a manual bundle for every permutation of plugins you want. That is what building your own custom plugin is for. All all devs in China are making WeChat apps.

bigtimebuddy:
I had a chat about this with @GoodBoyDigital tonight. His feeling was that WeChat is a pretty big audience and we should find an easier way to make this happen than an external patch. He feels like we should do something proactive in the runtime to let people know if their environment doesn't support this. I'm on board with this approach.

Basically, we should throw an error in the Renderer class if the environment doesn't support `new Function` and let people know they should use a special package to support this. Maybe something non-WeChat branded since there are other environments that disallow eval and new Function (e.g. something specific like `@pixi/unsafe-eval`). 

@ivanpopelyshev I'm happy to take a stab at this and massage this PR, if that's okay with you.

ivanpopelyshev:
That's the result I was hoping for. I'm glad things are going the way I predicted :)

bigtimebuddy:
I created a `@pixi/unsafe-eval` package that is _not_ bundled with pixi.js or pixi.js-legacy. It must be included separately. Here's how it would be used.

If you are in an environment that doesn't support eval or new Function,  you get this Error thrown when you create a new Renderer:

```
Current environment does not allow unsafe-eval,
please use @pixi/unsafe-eval module to enable support.
```

If a developer is using Webpack, Parcel or another bundler, would be install package and hook up like this:

```js
import * as PIXI from 'pixi.js'; // or '@pixi/core'
import { install } from '@pixi/unsafe-eval';

// Apply the patch to PIXI
install(PIXI);

// Create the renderer with patch applied, no error thrown
const renderer = new PIXI.Renderer();
```

If you are including the **unsafe-eval** bundle directly, you do not need to do anything else:

```html
<script src="pixi.min.js"></script>
<script src="unsafe-eval.min.js"></script>
```

ivanpopelyshev:
You forgot to remove "canGenerateNewFunction.js"

packages/utils/src/browser/unsafeEvalSupported.js
```diff
@@ -0,0 +1,33 @@
+// Cache the result to prevent running this over and over
```
englercj: Looks like a dupe of the same file above, should we be using the one in the utils package or the one in the core package?

bigtimebuddy:
Here are the builds. Would love it if someone could help verify:

https://pixijs.download/dev-new-function-fallback/pixi.min.js
https://pixijs.download/dev-new-function-fallback/unsafe-eval.min.js

ivanpopelyshev:
It works with a trigger script: https://www.pixiplayground.com/#/edit/XAYl0l1CglmThBl9N5NFO

I put breakpoint in syncUniform and it was called.

If you remove unsafe-eval it'll crash.

bigtimebuddy:
I disallowed unsafe-eval using Content Security Policy: https://mattkarl.com/unsafe-eval-test/ and I'm not getting this error:

```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' https://pixijs.download".

    at new Function (<anonymous>)
    at Function.MiniRunner.generateRun (mini-runner.min.js:1)
    at new MiniRunner (mini-runner.min.js:1)
    at CanvasResource.Resource (Resource.js:57)
    at CanvasResource.BaseImageResource [as constructor] (BaseImageResource.js:17)
    at new CanvasResource (CanvasResource.js:10)
    at createWhiteTexture (Texture.js:564)
    at Texture.js:595
    at index.js:97
```
There's another issue with mini-runner using `new Function`, so I'll need to address that too.

ivanpopelyshev:
Wow, another place with `new Function`. Sorry, I forgot about that.

bigtimebuddy:
I talked with @GoodBoyDigital and he agreed we could import mini-runner into Pixi. It's a small class and seems like it's a significant optimization enough to keep as a standalone package. I have named this package `@pixi/runner` which exports a single class called Runner.

bigtimebuddy:
# unsafe-eval-test now passes!

ivanpopelyshev:
Its a rare display, `+1,030 −26` PR with particular author name gets approved ;)

RobertAKARobin:
@ivanpopelyshev @bigtimebuddy It looks like `Runner` is an alternative to `EventEmitter`. Is there a particular use case for one over the other? I get the impression that `EventEmitter` is more "core" to Pixi.

bigtimebuddy:
Runner is primarily used internally for performance-critical code. EventEmitter is slower with some of our hot code spots. For example, Runner is used a bunch in the Renderer. For things like interaction events, and DisplayObject events, EventEmitter is fine.

KayacChang:
Hello, I got this issue recently.
Could you please put this subject into official document or wiki.

PR #5466, opened 2019-02-21
Avoid equal compare between float number

uinz:
##### Description of change

Avoid equal compare between float number.

In extreme conditions, precision conversion can cause some strange problems.

```cpp
// vTextureId = 2.0
float textureId = floor(vTextureId + 0.5);
bool isEqual = textureId == 2.0; // isEqual = false
```
##### solution
```js
if (i < maxTextures - 1) {
  src += `if (vTextureId < ${i}.5)`;
}
```

##### Pre-Merge Checklist

### DIFF
--- a/src/core/sprites/webgl/generateMultiTextureShader.js
+++ b/src/core/sprites/webgl/generateMultiTextureShader.js
@@ -10,7 +10,6 @@ const fragTemplate = [
 
     'void main(void){',
     'vec4 color;',
-    'float textureId = floor(vTextureId+0.5);',
     '%forloop%',
     'gl_FragColor = color * vColor;',
     '}',
@@ -55,7 +54,7 @@ function generateSampleSrc(maxTextures)
 
         if (i < maxTextures - 1)
         {
-            src += `if(textureId == ${i}.0)`;
+            src += `if(vTextureId < ${i}.5)`;
         }
 
         src += '\n{';


ivanpopelyshev:
Its true, I had some bad cases about it.

We have this line in v4: https://github.com/pixijs/pixi.js/blob/v4.x/src/core/sprites/webgl/generateMultiTextureShader.js#L13

Somehow, we forgot to merge it in v5.
Can you please modify PR that way it just adds that line and nothing else?

`float textureId = floor(vTextureId+0.5);`

uinz:
@ivanpopelyshev done!

ivanpopelyshev:
you can remove "<" now, because its always integer number.

uinz:
I found that on some Android devices,

`floor(X.5)` not equal `X.0`.

So I suggest  use `<` instead of `floor` and `==`.

ivanpopelyshev:
I am aware of this kind of effects, but i dont have a proof for it, every time i encountered it i just fixed it in plugins. I can merge that change in my libs: pixi-heaven, pixi-project, however I cannot accept it in pixijs vanilla core because i have teammates that will ask me to find that device and show them.

Also, it should work if X is big enough. For 0.0-32.0 it should work.

Please submit the demo and name of device, better if we know which videochip does that.

ivanpopelyshev:
@GoodBoyDigital I summon you!

uinz:
Fine, Thank you for your response.
I found this problem on `Samsung S9(8.0)` `Mi8(8.0&8.1)` `vivo nex A(8.0)`running on custom engine,
but on system browser works fine.

ivanpopelyshev:
Did you test specific numbers or only small ones ?

PR #5468, opened 2019-02-22
Update docs for the forceCanvas option

nikolas:
It seems this option is only available in pixi.js-legacy, and you can't use forceCanvas with pixi.js anymore. The docs should state that.

* https://github.com/pixijs/pixi.js/issues/5417
* https://github.com/pixijs/pixi.js/issues/5389


themoonrat:
I almost wonder if for this option it should output a console error message as well.... otherwise I can see so make incoming questions about this option not working...

Issue #5471, opened 2019-02-24
Container Events


davidfig:
DisplayObject has the 'added' and 'removed' events. It would be helpful if Container included events for 'addedTo' and 'removedFrom' (or something along those lines) for when the container's children change. the Container.onChildrenChange() overridable is not ideal since it doesn't say what was added or removed.

Happy to create a PR if this is something that's interesting.

bigtimebuddy:
Good idea, PR welcome!

PR #5472, opened 2019-02-24
Adds Containers events for addedTo and removedFrom

davidfig:
##### Description of change
Pull request for #5471 

DisplayObject has the 'added' and 'removed' events. It would be helpful if Container included events for 'addedTo' and 'removedFrom' (or something along those lines) for when the container's children change. the Container.onChildrenChange() overridable is not ideal since it doesn't say what was added or removed.

##### Pre-Merge Checklist

- [didn't add tests] Tests and/or benchmarks are included
- [? (I had issue with npm test -- it wouldn't complete but didn't list any errors] Tests passed (`npm run test`)


bigtimebuddy:
Thank you @davidfig! I think it's important that we add some basic tests here so we don't break these events in the future. Do you need help with that?

davidfig:
I added the tests for these events.

bigtimebuddy:
@davidfig, could you please remove package-lock.json?

davidfig:
Removed.

bigtimebuddy:
Oops, sorry I wasn't clearer. You actually should just revert it to the one that's in the dev branch. You should actually delete. We need that file, but different versions of npm make small subtle changes to that file that we don't want as part of this PR.

davidfig:
:smile: Okay, it's back to the current one.

davidfig:
Good suggestion. I renamed the events to childAdded and childRemoved, and added the index to the events. Tests are also updated.

bigtimebuddy:
Nice one. Thanks @davidfig for reacting to our feedback so quickly.

PR #5476, opened 2019-02-25
Remove canvas-renderer from @pixi/prepare dependency

rsferreira11:
##### Description of change

Remove unused "@pixi/canvas-renderer" dependency from the package "@pixi/prepare".


bigtimebuddy:
Thank you @rsferreira11!

PR #5492, opened 2019-03-05
Fix a bug when resizing render texture with resolution other than one

GoodBoyDigital:
Howdy! Currently in v5 if you resize a renterTexture with a resolution different from 1 then the frame buffer breaks.. 

This is because the resolution  of the textures was not being taken into account when resizing the frame buffers (which are always resolution 1)

This fixes it 👍 

Here is an example of it busted:

https://www.pixiplayground.com/#/edit/gvG2JWZw7xSdCQIn8LaBB

running with this branch fixes it :)

### DIFF
--- a/packages/core/src/framebuffer/Framebuffer.js
+++ b/packages/core/src/framebuffer/Framebuffer.js
@@ -134,7 +134,7 @@ export default class Framebuffer
         this.dirtyId++;
         this.dirtySize++;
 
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Lint 👆

ivanpopelyshev:
I will challenge it with `Math.round` later, I remember very rare issue with `X-files` label.

PR #5501, opened 2019-03-08
Remove isOdd-like dependency

ivanpopelyshev:
We also removed two our local deps for v5 (gl-core, mini-runner). Sorry, @mreinstein


ivanpopelyshev:
I just don't know what went wrong.

mreinstein:
I'm not sure why you're calling it "leftpad-like dependency" since removing array items has nothing to do with padding strings but you're free to do what you wish in your own projects.

ivanpopelyshev:
It's ok, that meme isn't worth your time ;)

mreinstein:
> It's ok, that meme isn't worth your time ;)

Well you pinged me on it. Perhaps you could explain it. :)

I really hope you are not suggesting that all dependencies on "external" modules is bad, or that no one can be trusted to write and maintain software dependencies. If you're comparing the module I maintain to one that was deleted in a temper tantrum by some random developer on the internet....that's kind of mildly insulting.

ivanpopelyshev:
Nope, only as single-module of small size. Yes, I agree that its more close to another example: https://www.npmjs.com/package/is-odd .

mreinstein:
I understand that it's easy to just take one of these small modules and rip it directly into your project.

I'm not going to suggest that a simple array removal function is super critical to the function of pixi.js. Obviously this is a small utility function out of many. The more disappointing thing is that it fits into a general philosophy in this project to reject re-use. 

Video games and other real time simulations written in javascript/typescript are the bulk of projects that might consider pixi, and those types of projects tend to need to quickly add and remove things from arrays in ways that don't smash the garbage collector. It pervades the entire stack; the rendering, the physics, the game logic, etc. If every framework that gets included decides to say "I'm most important, other modules don't matter, I'm a monolith. I'll just pull everything into one massive package" well then now I have to have 15 copies of every utility function, which bloats the size of my code. Especially important if you're trying to ship things on the web.

bigtimebuddy:
@mreinstein First, thank you for maintaining this utility. We appreciate the effort to manage and publish helpful utilities like this for v4. We have no plans to remove this dependency from v4, this is for v5 and moving forward.

@ivanpopelyshev did a terrible job explain the reasoning for this. Let me provide a little more context here.

We want to simplify our dependency tree overall moving forward and be a little more intentional about what we spin off as external dependencies and what we include as packages or sub-packages within the project. The places where it makes sense for us to rely on external dependencies are eventemitter3, earcut and ismobile, for instance. These packages have a lot of complexity and require thorough testing and validating that we don't want to handle ourselves.

By contrast mini-runner and remove-array-items are simple functions/class that are actually easier to maintain internally because the maintenance cost isn't that high and they are stable utils. Also, it's likely in the future we'll be moving to more TypeScript so simplifying dependencies now will make that process much easier later.

Lastly, folks can still including this function in their projects without all the bulk of PixiJS because now we publish the internal packages as standalone modules. Especially if you're using Webpack or Rollup's tree-shaking, doing something like this will just import that function into your project without any extra overhead:
```js
```

ivanpopelyshev:
It would be cool to add such modules of slightly bigger size to separate space, with a name like "@super-game-lib/array-utils" . In #5463 runner goes to "@pixi/runner".

Also, you've got me! I have TS version of several libs (event pixi), that are used in my production projects, I update by hands to check if authors missed something.

Yeah, @bigtimebuddy explained it better while i was writing response :)

mreinstein:
> We have no plans to remove this dependency from v4, this is for v5 and moving forward.

Okay, so the shiny new upcoming pixi version will reduce even further the few remaining places where it currently does manage to re-use modules.  That's not really comforting. The values of re-using software don't go away because pixi is cutting a new release. If anything I'd be hoping for _more_ re-use in a new version, not less.



>  it's likely in the future we'll be moving to more TypeScript so simplifying dependencies
> now will make that process much easier later.

This exactly re-inforces the previous point. Re-using stuff is good! If you were to instead submit a patch to `remove-array-items` for typescript support then other people could reap the benefits of that work, in a direct immediate way. _collaborate_ with people so that we can not add typescript support to remove array items utility functions in 15 different projects. :)

Please don't take what I'm saying as an attack or anything personal. I'm hoping this comes across as constructive because there are some smart, hardworking people that maintain pixi. It's just a shame that pixi is hosted on platforms like github and npm which are designed to foster sharing and collaboration, but we're still stuck in this old school mentality of "everything inside my castle's walls".

bigtimebuddy:
Looking back on the original PR #3977 there were concerns raised by Chad and I which I still think are valid now.

With all respect, I don't agree with the characterization of this PR as "everything inside my castle's walls." PixiJS obviously _needs_ to rely on external projects in a number of cases and we contribute to them as needed.  We are, however, choosing to apply a new heuristic here about what's a production dependency and what's not.

I'm fine agreeing to disagree and for @mreinstein to continue to main an externalized **remove-items-array** package.

PR #5502, opened 2019-03-09
Don't use webgl2 on mobiles by default

ivanpopelyshev:
If anyone has better idea how to detect chromium on mobiles, feel free to alter this PR.

https://bugs.chromium.org/p/chromium/issues/detail?id=934823

### DIFF
--- a/packages/core/src/settings.js
+++ b/packages/core/src/settings.js
@@ -18,6 +18,6 @@ import { isMobile } from '@pixi/utils';
  * @type {number}
  * @default PIXI.ENV.WEBGL2
  */
-settings.PREFER_ENV = isMobile.any ? ENV.WEBGL : ENV.WEBGL2;
+settings.PREFER_ENV = !isMobile.any || isMobile.apple.device ? ENV.WEBGL2 : ENV.WEBGL;
 
 export { settings };
--- a/packages/core/src/settings.js
+++ b/packages/core/src/settings.js
@@ -10,7 +10,7 @@ import { isMobile } from '@pixi/utils';
  * baseline, prefer a lower environment.
  *
  * Due to {@link https://bugs.chromium.org/p/chromium/issues/detail?id=934823|bug in chromium}
- * we disabled webgl2 by default on all mobile devices except apples.
+ * we disable webgl2 by default for all non-apple mobile devices.
  *
  * @static
  * @name PREFER_ENV
--- a/packages/core/src/settings.js
+++ b/packages/core/src/settings.js
@@ -10,7 +10,7 @@ import { isMobile } from '@pixi/utils';
  * baseline, prefer a lower environment.
  *
  * Due to {@link https://bugs.chromium.org/p/chromium/issues/detail?id=934823|bug in chromium}
- * we disabled webgl2 by default for all mobile devices.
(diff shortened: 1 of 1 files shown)


themoonrat:
Couldn't we allow it on mobile just for ios? That's still gonna be a fair % of mobile users?

ivanpopelyshev:
@themoonrat alter the PR, I agree. We didnt have reports on iOS.

ivanpopelyshev:
@themoonrat done. can you check it?

themoonrat:
Code looks good, the `we disabled webgl2 by default for all mobile devices.` needs updating tho to `we disable webgl2 by default for all non-apple mobile devices`

goldenratio:
Seems the issue is related to Mali driver, 
Why not check for `UNMASKED_RENDERER_WEBGL`

ivanpopelyshev:
@goldenratio good idea! Now we only need someone who has affected devices to tell us what's the value of this string, then use it on test context we already have something there.

goldenratio:
@ivanpopelyshev 
checking for `Mali` string in `UNMASKED_RENDERER_WEBGL` should be good enough. From chromium defect below are the affected Mali GPU names,

```
Honor 9 Lite: Mali-T830 MP2
Alcatel A5 LED: Mali-T720MP3
LG K10 (2017): Mali-T860MP2
Samsung Galaxy Tab S2 (SM-T710): Mali-T760 MP6
```

Monday I can get you more Mali GPU names from our work test phones.

ivanpopelyshev:
Ok, so, we need advanced check , something with `getTestContext()` or even inside it. I'm calling @GoodBoyDigital

GoodBoyDigital:
Howdy! My 2 pence would be to disable WebGL2 for all mobiles as a hotfix at this stage and rely on users to make the decision until we are confident that it all works ok. Want to ensure v5 works for everyone and this is quite a large number of devices are affected. 

But medium term let's explore the blacklisting option for sure.

iOS does not support WebGL2 and I think that it may be better to ensure that we enable it in a controlled way when it does land. Essentially we should flick that switch, not the browser update as there may be teething issue with webGL2 on iOS (as we have found with android) and a bunch of Pixi content could end up breaking.

ivanpopelyshev:
OK, done, we can merge it.

packages/core/src/settings.js
```diff
@@ -8,12 +9,15 @@ import { ENV } from '@pixi/constants';
...
+ * we disable webgl2 by default for all non-apple mobile devices.
+ *
  * @static
  * @name PREFER_ENV
  * @memberof PIXI.settings
  * @type {number}
  * @default PIXI.ENV.WEBGL2
  */
-settings.PREFER_ENV = ENV.WEBGL2;
+settings.PREFER_ENV = !isMobile.any || isMobile.apple.device ? ENV.WEBGL2 : ENV.WEBGL;
```
GoodBoyDigital: should it be 
`settings.PREFER_ENV = (!isMobile.any || !isMobile.apple.device) ? ENV.WEBGL2 : ENV.WEBGL;` 
would do the trick?

ivanpopelyshev: I forgot to push. All apple devices are mobile, right?

PR #5507, opened 2019-03-12
Update displacement filter to support rotation of displacement map pixels

distinctdan:
##### Description of change

Updated displacement filter to support rotation of displacement map pixel values. Allows displacement map pixels to displace in local-space instead of always in global space. For more info, please see https://github.com/pixijs/pixi.js/issues/5491

##### Pre-Merge Checklist


bigtimebuddy:
Could you also make a public API to make it easier to set rotation? Setting a 2x2 matrix via uniforms isn't that intuitive for most folks.

distinctdan:
@ivanpopelyshev Hmm, it did look fine there, but I realized the playground I put together didn't have any displacement in the y axis, so you couldn't tell it was doing something weird. If you look at this playground and you change them to inverted, you can see spikey displacement at the top and bottom of the edges of the sprite: https://www.pixiplayground.com/#/edit/M~8Ld47o6vOWvFTxmvvOX

@bigtimebuddy Not sure I understand here. End users shouldn't have to do anything for this to work, it's automatic based on the rotation of the displacement sprite.

bigtimebuddy:
Oooh you’re right! Sorry I misinterpreted the code. Makes sense

distinctdan:
Cool cool, no worries.

PR #5515, opened 2019-03-14
Graphics closePath()

ivanpopelyshev:
Based on @bigtimebuddy comment from #5462

Original demo : https://www.pixiplayground.com/#/edit/BDm_4pTY_4DUR3X1XvzSV

PR #5516, opened 2019-03-14
Graphics holes for canvas2d

ivanpopelyshev:
Lets remove `closed` field and do everything right.
This PR relies on #5515 and #5462.

Demo: https://www.pixiplayground.com/#/edit/L3AqhuxsEI8NeX8_LEMPt


ivanpopelyshev:
OK, rebased to make it easier

ivanpopelyshev:
Need confirmation that we delete `polygon.close()` method

Issue #5520, opened 2019-03-17
PIXI.Graphics initially drawing transparent


btmorex:
### Expected Behavior

I'm trying to draw a white bar attached to the player sprite.

### Current Behavior

The bar is drawn transparent initially. You can see the area blocked off in the upper right here: https://imgur.com/5gsTd06 Oddly enough, after firing the ship's weapon a few times which essentially creates a few more sprites that quickly disappear the graphics turn white and stay that way. As in: https://imgur.com/gI9ilpF This is strange because I'm completely clearing the stage and recreating everything every frame.

### Steps to Reproduce

Here's my code:

```typescript
const createCapacitorBar = (capacitor: Capacitor) => {
  const bar = new PIXI.Graphics();
  bar.beginFill(0xffffff);
  bar.drawRect(0, 0, 100, 100);
  bar.endFill();
  return bar;
};
```

And elsewhere:

```typescript
    // add player
    const playerSprite = createSprite(player.texture, player.scale, player.static);
    playerSprite.addChild(createCapacitorBar(player.capacitor));
    this.stage.addChild(playerSprite);
    this.stage.addChild(createCollisionShape(player));
```

### Environment

- **`pixi.js` version**: v5.0.0-rc.2
- **Browser & Version**: Chrome 73
- **OS & Version**: Ubuntu 18.04.2

btmorex:
Oh, one more thing, if I use `pixi.js-legacy` and force a canvas renderer the problem goes away. It doesn't appear to be a problem with my graphics card or drivers though because my laptop has both intel and nvidia cards and it happens with both. Also, occurs with both firefox and chrome.

ivanpopelyshev:
I see why it can happen, but it requires other Graphics objects on stage. What other Graphics do you have?

btmorex:
I'm drawing my collision shapes right now, so for example:

```typescript
      const shape = new PIXI.Graphics();
      shape.lineStyle(1, color);
      shape.drawCircle(
        entity.static.position[0],
        entity.static.position[1],
        entity.shape.radius,
      );
      shape.endFill();
      return shape;
```

and

```typescript
      const shape = new PIXI.Graphics();
      shape.lineStyle(1, color);
      shape.drawCircle(entity.static.position[0], entity.static.position[1], 1);
      shape.endFill();
      return shape;
```

ivanpopelyshev:
[...]

PR #5522, opened 2019-03-17
Graphics batch fixes

ivanpopelyshev:
Attempt to solve #5520 without actual reproduction.


bigtimebuddy:
I think we need a playground or fiddle for this @ivanpopelyshev.

ivanpopelyshev:
My changes are completely wrong, I re-read the code and I see several places that can cause problems but I failed to reproduce them. We need more info from author of #5520

PR #5524, opened 2019-03-17
Better support for IE 11, by moving polyfills to default bundle

bigtimebuddy:
Many of the polyfills are required for IE 11 even though WebGL is supported like `Object.assign`, `Math.sign`, `Number.isInteger`, `Promise`. This moves the polyfills package to the default  bundle, not just the `pixi.js-legacy` bundle. In the future, we might want to revert this change. But for now we should try to broadly support all major browsers with WebGL.


themoonrat:
This fixes https://github.com/pixijs/pixi.js/issues/5258

ivanpopelyshev:
Oh no, library size!

PR #5528, opened 2019-03-18
Browser Bundles for Packages

bigtimebuddy:
## Overview

The purpose of this PR is to generate browser bundles for each of the standalone packages. This is useful because it would allow us to create some web tooling to produce custom PixiJS builds without needing a bundler (e.g., Webpack, Rollup). Also, it would allow users to test by importing only the package they needed.

**Minimal PixiJS Example**
This is the minimum packages need to create a renderer (~170KB minified, ~52KB gzipped)
https://jsfiddle.net/bigtimebuddy/Lgadeocs/

```html
<script src="https://pixijs.download/dev-bundles/packages/runner.min.js"></script>
<script src="https://pixijs.download/dev-bundles/packages/constants.min.js"></script>
<script src="https://pixijs.download/dev-bundles/packages/settings.min.js"></script>
<script src="https://pixijs.download/dev-bundles/packages/utils.min.js"></script>
<script src="https://pixijs.download/dev-bundles/packages/ticker.min.js"></script>
<script src="https://pixijs.download/dev-bundles/packages/math.min.js"></script>
<script src="https://pixijs.download/dev-bundles/packages/display.min.js"></script>
<script src="https://pixijs.download/dev-bundles/packages/core.min.js"></script>
<script>
  const renderer = PIXI.autoDetectRenderer();
</script>
```

### Changed

* Removes `@pixi/fragments` this is an unnecessary dependency, moved default vert shaders to core
* Standardizes building minified browser bundles


ivanpopelyshev:
What is `namespaceInject` and how does it help us?

bigtimebuddy:
Is a build option I created, not a convention. Basically it means append to existing namespace, don’t create it. For instance, all the classes in math, these should be added to PIXI global object and the math package should not set PIXI.

englercj:
Would it be easier if each of them included the bit that created it if it didn't exist, and added themselves to it if it did? That way any bundle would create, and any bundle would add.

For example, why not have math create it if I only include math?

bigtimebuddy:
There are a handful of cases where it doesn’t do that: like accessibility, interaction, prepare, extract, utils where it’s easier to do what Rollup does out of the box. But I could probably just normalize around this behavior rather than having two build paths.

bigtimebuddy:
Hmm, I wonder if I should roll all the core dependencies into core bundle. Would really simplify things and those math, settings, etc packages need to be included in a specific order to make core work anyways.

bigtimebuddy:
Few more clean-up changes:
* Reduce the dependencies between packages (`ismobilejs`, `eventemitter3`, `url`), this will prevent the dependency from being inlined twice if using bundle packages. Since most packages depend on utils, this doesn't affect anything or change internal package dependencies, but reduces the overall size.
* Adds proper documentation for `isMobile` object exported by utils
* Exports `url` dependency from utils to standardize external access to dependencies, just like `EventEmitter`, `earcut` and `isMobile`

PR #5537, opened 2019-03-22
[FEAT] Added anisotropic filtering

### LINKED ISSUE #3929: Enable Anisotropic filtering for textures where available
Using standard bilinear filtering for sprite textures in the manner that PIXI currently does, often leads to slightly blurry sprites when a sprite is scaled to a size that is inbetween two mipmaps.

Since part of our app is a WebGL powered photo editor, using PIXI, this is not ideal.  We have previously worked around this issue by modifying PIXI's sprite shader generator to set a negative mipmap LOD bias:

```vec4 sample = texture2D(uSampler, vTextureCoord, -0.5);```

This ensures that the sprite is always using a slightly higher resolution mipmap.

While this works fine, we have recently realised that WebGL now has good support for anisotropic filtering, which enables us to remove the negative LOD bias and always get perfectly sharp sprites.  Furthermore, performance is significantly better than our previous method since on modern GPU's, anisotropic filtering is pretty much free, or if not, the performance impact is absolutely miniscule.

Support for anisotropic filtering can be checked like so:

```
 var ext = (
                gl.getExtension('EXT_texture_filter_anisotropic') ||
                gl.getExtension('MOZ_EXT_texture_filter_anisotropic') ||
                gl.getExtension('WEBKIT_EXT_texture_filter_anisotropic')
            );

```

And can then be set on a per texture basis in the same manner as PIXI currently sets its current linear/nearest filtering in the ```Texture.minFilter``` function:

```
if(ext){
     var max = gl.getParameter(ext.MAX_TEXTURE_MAX_ANISOTROPY_EXT);
     var anisotropy_level = Math.min(this.anisotropic, max);
     gl.texParameterf(gl.TEXTURE_2D, ext.TEXTURE_MAX_ANISOTROPY_EXT, anisotropy_level );
}
```

Please note than in this example we have simply added an anisotropic property to the ```Texture``` object, which accepts values from 0-16.  We then check to see the maximum supported anisotropy level, and apply the chosen level or the highest supported value.

Also, in our hacked function, we only apply this technique if linear filtering is selected for a texture, and if mipmapping is enabled and allowed for said texture (since it wouldn't make any difference without mipmaps anyway).

It would be great to see this added into PIXI as the improvement in scaled sprites is hugely significant.

More info can be found here: [https://developer.mozilla.org/en-US/docs/Web/API/EXT_texture_filter_anisotropic](url)

Teoxoy:
##### Description of change
Adds anisotropic filtering for textures (closes #3929)

##### Pre-Merge Checklist

##### Notes:

- I tested the feature in my own pixi project and it seems to work fine
- 5 tests related to `BitmapFontLoader` and `TextMetrics` failed locally


ivanpopelyshev:
I waited this moment for years.

I need time to review it.

I dont think using the extensions by default if its available is good idea.

Teoxoy:
I could set the default value to 0 so that it's off by default. Nowadays anisotropic filtering is "free" from a performance perspective but some apps made with pixi don't need AF so I guess we should make it opt in rather than opt out. Tell me what you guys think.

ivanpopelyshev:
Yeha, we can set it to 16 by default in next versions when its tested enough :)

Teoxoy:
Sounds good. I will push the commit in a bit.

ivanpopelyshev:
Are we sure that anisotropic mip levels will be generated in `generateMipmaps` ?

Teoxoy:
I am not 100% sure how anisotropic filtering works under the hood but I think it just composes a better version of the texture using the generated mipmaps. I don't think it generates any anisotropic mip levels.

bigtimebuddy:
@Teoxoy could you potentially resolve the conflicts? I think we're finally ready to merge this in.

Teoxoy:
Done!

bigtimebuddy:
Thank you @Teoxoy!

Teoxoy:
Glad I could help!

vvashisht:
Hey! Is this planned for v5.0.5?

Teoxoy:
From what I can see, it's in the 5.1.0 milestone.

vvashisht:
Thanks @Teoxoy! Is there a way I can implement it on top of 5.0.4? This is what I've tried so far:

`    
const gl = app.renderer.gl;
const anisotropicExt =
    gl.getExtension('EXT_texture_filter_anisotropic') ||
    gl.getExtension('MOZ_EXT_texture_filter_anisotropic') ||
    gl.getExtension('WEBKIT_EXT_texture_filter_anisotropic');

    const anisotropy = 16;
    
    if (anisotropicExt && anisotropy >= 0 && texture.scaleMode === PIXI.SCALE_MODES.LINEAR) {
      gl.texParameterf(
        texture.target,
        anisotropicExt.TEXTURE_MAX_ANISOTROPY_EXT,
        Math.min(anisotropy, gl.getParameter(anisotropicExt.MAX_TEXTURE_MAX_ANISOTROPY_EXT),
        ),
      );
    }

    app.renderer.texture.updateTextureStyle(tex);
`

I'm probably not doing it right because it has no effect and additionally on Android (Chrome) I get an error saying:

GL ERROR GL_INVALID_OPERATION glGenerateMipmap Can not generate mips

All of my textures are POW2.

Teoxoy:
What you can do is: Fork my fork, checkout the branch, build it and then copy paste the built files.

vvashisht:
Thanks! Will do.

Teoxoy:
Guys, when I changed the default `ANISOTROPIC_LEVEL` to be 0 I forgot to change the jsdoc.
The [pixi docs](http://pixijs.download/release/docs/PIXI.settings.html#.ANISOTROPIC_LEVEL) now state the default is 16 which is not true.

I don't think starting another PR just to fix this is worth it. If one of the core contributors could just make this quick change that would be great!
Sorry for overlooking this.

PR #5544, opened 2019-03-27
Add note about memory leaks to Graphics; make GraphicsGeometry useabl…

btmorex:
…e by endusers

##### Description of change

- Added a note about calling destroy() on Graphics to avoid memory leaks
- Fixed typing of geometry member
- Exported FillStyle and LineStyle so that endusers can call drawShape() on GraphicsGeometry

##### Pre-Merge Checklist

### DIFF
--- a/packages/graphics/src/Graphics.js
+++ b/packages/graphics/src/Graphics.js
@@ -36,7 +36,7 @@ let defaultShader = null;
  * Note that because Graphics can share a GraphicsGeometry with other instances,
  * it is necessary to call `destroy()` to properly dereference the underlying
  * GraphicsGeometry and avoid a memory leak. Alternatively, keep using the same
- * Graphics instance and call `clear()` between frames.
+ * Graphics instance and call `clear()` between redraws.
  *
  * @class
  * @extends PIXI.Container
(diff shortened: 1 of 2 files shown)


packages/graphics/src/Graphics.js
```diff
@@ -33,6 +33,11 @@ let defaultShader = null;
  * The Graphics class contains methods used to draw primitive shapes such as lines, circles and
  * rectangles to the display, and to color and fill them.
  *
+ * Note that because Graphics can share a GraphicsGeometry with other instances,
+ * it is necessary to call `destroy()` to properly dereference the underlying
+ * GraphicsGeometry and avoid a memory leak. Alternatively, keep using the same
+ * Graphics instance and call `clear()` between frames.
+ *
```
bigtimebuddy: Looks good, the only comment here is “frames” is confusing. Might suggest “redraws” instead. Don’t want people to call clear between every tick or requestAnimationFrame if they don’t need to.

btmorex: Fixed 502312955fbff9c15a849ab51b5a93155ad8e4e7

PR #5546, opened 2019-03-28
Assure that TICK is different for multi-textured renderers

### LINKED ISSUE #5520: PIXI.Graphics initially drawing transparent

ivanpopelyshev:
Fixes #5520 

I added methods to BaseTexture because that's where `_batchEnabled` field is stored.

I also suggest that we expose those methods for plugin-makers but `_` means that it can be changed in any version without deprecation.

Broken: https://www.pixiplayground.com/#/edit/1WCb9fBbUQieUK2IdMKzR
Works: https://www.pixiplayground.com/#/edit/69HLoVsdpfl0NBwEyJws0

### DIFF
--- a/packages/core/src/textures/BaseTexture.js
+++ b/packages/core/src/textures/BaseTexture.js
@@ -653,14 +653,13 @@ export default class BaseTexture extends EventEmitter
 
     /**
      * Called from multi-texture rendered to save the last batch number.
-     * Only first 32 bits matter.
      *
      * @static
      * @param {number} savedValue last used batch number
      */
     static _batchEnd(savedValue)
     {
-        BaseTexture._globalBatch = savedValue | 0;
+        BaseTexture._globalBatch = savedValue;
     }
 }
 
--- a/packages/core/src/textures/BaseTexture.js
+++ b/packages/core/src/textures/BaseTexture.js
@@ -226,7 +226,7 @@ export default class BaseTexture extends EventEmitter
         this.resource = null;
 
         /**
-         * Global number of the texture batch, used by multi-texture renderers
+         * Number of the texture batch, used by multi-texture renderers
          *
          * @member {number}
          */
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
@btmorex 

Thanks, I stumbled across this problem when i was making plugins for v4.

packages/core/src/textures/BaseTexture.js
```diff
@@ -632,4 +639,29 @@ export default class BaseTexture extends EventEmitter
...
+    /**
+     * Called from multi-texture rendered to save the last batch number.
+     * Only first 32 bits matter.
+     *
+     * @static
+     * @param {number} savedValue last used batch number
+     */
+    static _batchEnd(savedValue)
+    {
+        BaseTexture._globalBatch = savedValue | 0;
```
GoodBoyDigital: looking good, can you explain how this magic line works pls? Why only first 32 bits matter? Just curious! :)

ivanpopelyshev: just in case we have more than 2^54 texture batches. Ugh... 100 years I think? lets remove that.

ivanpopelyshev: ok, changed.

bigtimebuddy:
@GoodBoyDigital did you have any other feedback on this from last we talked?

packages/graphics/src/GraphicsGeometry.js
```diff
@@ -580,7 +579,7 @@ export default class GraphicsGeometry extends BatchGeometry
      */
     buildDrawCalls()
     {
-        TICK++;
+        let TICK = BaseTexture._batchBegin();
```
GoodBoyDigital: how about 

```let TICK = BaseTexture._globalBatch++;```

File: packages/graphics/src/GraphicsGeometry.js [hunk 3]
```diff
@@ -678,6 +677,8 @@ export default class GraphicsGeometry extends BatchGeometry
             this.addTextureIds(textureIds, textureId, data.attribSize);
         }
 
+        BaseTexture._batchEnd(TICK);
```
GoodBoyDigital: and ```BaseTexture._globalBatch = TICK;```

PR #5589, opened 2019-04-17
Fixed parameter ordering

### LINKED ISSUE #5586: Attribute.from has wrong parameter order
Seen on PIXI 5.0.0.
Even though it's not used anywhere, the static function Attribute.from has wrong parameter order:

Attribute.from = function from (buffer, size, stride, start, normalized)
{
    return new Attribute(buffer, size, stride, start, normalized);
};

function Attribute(buffer, size, normalized, type, stride, start, instance)

normalized and stride are reversed.

abhinavsagar:
[Fixes #5586 ](https://github.com/pixijs/pixi.js/issues/5586)

### DIFF
--- a/packages/core/src/geometry/Attribute.js
+++ b/packages/core/src/geometry/Attribute.js
@@ -50,8 +50,8 @@ export default class Attribute
      *
      * @returns {PIXI.Attribute} A new {@link PIXI.Attribute} based on the information provided
      */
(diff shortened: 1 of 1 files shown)


ASWC:
I wonder now if the "start" is really meant to be "type", as "start" it is out of place as well in the constructor parameters since it's supposed to come after stride and not between normalized and stride (which is the "type" place)

PR #5592, opened 2019-04-17
Adds format and quality options to extract plugins

### LINKED ISSUE #4846: Only PNGs passed back when using pixi.renderer.extract.image
This is an FYI for anyone using pixi.renderer.extract.image under Canvas renderer. Who might run into this issue. 

Our application takes a user image, renders it to the Pixi canvas while adding some effects. This final image is then shared on Twitter and Facebook.

We were was using pixi.renderer.extract.image to create a screenshot to post on Twitter's API using the 'https://upload.twitter.com/1.1/media/upload.json' endpoint. 

However, by default, pixi.renderer.extract.image would return a high fidelity PNG image which would sometimes create a "413 Request Entity Too Large" error on a POST action to the Twitter API.

pixi.renderer.extract.image call this function.

`    CanvasExtract.prototype.base64 = function base64(target) {
        return this.canvas(target).toDataURL();
    };
`

Our fix was to modifying the lib to give back JPEGs at 0.8

`    CanvasExtract.prototype.base64 = function base64(target) {
        return this.canvas(target).toDataURL('image/jpeg', 0.8);
    };
`

### Environment

- `pixi.js` version: 4.7.3
- Browser: All
- Device: All

bigtimebuddy:
Fixes #4846

Adds `format` and `quality` parameters to extract plugins. This is a pass-through for `toDataURL`, see: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL

### DIFF
--- a/packages/canvas/canvas-extract/src/CanvasExtract.js
+++ b/packages/canvas/canvas-extract/src/CanvasExtract.js
@@ -35,8 +35,8 @@ export default class CanvasExtract
      *
      * @param {PIXI.DisplayObject|PIXI.RenderTexture} target - A displayObject or renderTexture
      *  to convert. If left empty will use the main renderer
-     * @param {string} [format] - Image format, e.g. "image/jpeg".
-     * @param {number} [quality] - JPEG compression from 0 to 1, must specify format.
+     * @param {string} [format] - Image format, e.g. "image/jpeg" or "image/webp".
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
0.92 is the browser default per MDN docs

PR #5603, opened 2019-04-21
fix for cacheAsBitmap

GoodBoyDigital:
Setting a custom transform slipped out of the render function at some point!
Adding it back in fixes cacheAsBitmap.

### DIFF
--- a/packages/core/src/Renderer.js
+++ b/packages/core/src/Renderer.js
@@ -325,6 +325,9 @@ export default class Renderer extends AbstractRenderer
         this.runners.prerender.run();
         this.emit('prerender');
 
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
working example here: 

https://www.pixiplayground.com/#/edit/r657IDxP9ysr1dK4Wtxic 

👍

ivanpopelyshev:
I tested it on another issue, it works! 

Also, I took one of older tests: https://www.pixiplayground.com/#/edit/jg7WEsoNFL1Cy55HB9f69

It doesn't work in rc3, but works with your fix

PR #5605, opened 2019-04-21
Allow pointerdown & pointerup events to handle simultaneous presses

themoonrat:
##### Description of change
https://github.com/pixijs/pixi.js/issues/4048
https://github.com/pixijs/pixi.js/issues/4863
https://github.com/pixijs/pixi.js/issues/5384

Native pointer events only send a 'pointerdown' on the initial button being pressed.
Hold one mouse button down and press another, and you get a 'pointermove' instead.
Release that second mouse button whilst still holding the first, and again you get a 'pointermove' event.
So here, we will detect any change in the 'buttons' being pressed from last event to this move event.
If the number has changed, we know it's due to multiple mouse presses.

Goes slightly against the native pointerdown spec, but is much more useful for the developer

##### Pre-Merge Checklist

### DIFF
--- a/packages/interaction/package.json
+++ b/packages/interaction/package.json
@@ -31,6 +31,7 @@
     "@pixi/core": "^5.0.0-rc.3",
     "@pixi/display": "^5.0.0-rc.3",
     "@pixi/math": "^5.0.0-rc.3",
+    "@pixi/settings": "^5.0.0-rc.3",
     "@pixi/ticker": "^5.0.0-rc.3",
     "@pixi/utils": "^5.0.0-rc.3"
   },
--- a/packages/interaction/src/InteractionManager.js
+++ b/packages/interaction/src/InteractionManager.js
@@ -1,10 +1,11 @@
 import { Ticker, UPDATE_PRIORITY } from '@pixi/ticker';
 import { Point } from '@pixi/math';
 import { DisplayObject } from '@pixi/display';
+import { EventEmitter } from '@pixi/utils';
+import { settings } from '@pixi/settings';
 import InteractionData from './InteractionData';
 import InteractionEvent from './InteractionEvent';
 import InteractionTrackingData from './InteractionTrackingData';
-import { EventEmitter } from '@pixi/utils';
 import interactiveTarget from './interactiveTarget';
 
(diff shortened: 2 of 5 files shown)


themoonrat:
Before: https://pixiplayground.com/#/edit/AL2vtebBZcxrCaHYftC4G
After: https://pixiplayground.com/#/edit/U_ybC8fu4sEIPPkNecEGI

andrewstart:
I am largely fine with this as is (wary of doing it for `pen` pointers), but if you really wanted to keep the `pointerdown` event standard, you could emit mouse/click events (add a `emitPointer` boolean to `onPointerDown/Up`) and _optionally_ add a completely non-standard `pointerbuttonchange` event.

themoonrat:
@andrewstart A good summary of the options. I'm not sure if I've gone the right way on this now!
I could imagine _some_ code by devs may be listening to pointerup, and acting upon it. With this current change, you'd want to explicitly check for `.buttons === 0` to make sure the button is totally released. So I guess that's why the spec makers went for only a pointerup when no buttons are in a down state.

themoonrat:
But then you've also got the scenario of there _not_ being native pointer event support, so when dealing with multiple mouse presses, I'd have to find a way to _not_ emit pointerdown events generated from the mouse. 
Arg!

themoonrat:
Ok, gone a different route for this. 

So now, by default, the MouseEvents are used for mouse presses, TouchEvents are used for touches, and they get emitted as a polyfilled 'pointer' event, which contains the format of a pointer event, but not it's behaviour. I thinks, as a default, will achieve what most people require and expect from our library.

If they _do_ want to use the official pointer behaviour, then they can opt into it.

I'll want to check this all works as intended with a ms surface tomorrow, as that has both a touch screen and a mouse.

themoonrat:
I may need longer on this, sorry. Interaction is a rabbit hole of different use cases, and I wanna get this right.

bigtimebuddy:
@themoonrat I'd rather you get this right than rush it in. I'd like to put this for v5.1. I know it's a biggish change, but I'd like to not hold up v5.

themoonrat:
Closing until I work out what to do

PR #5608, opened 2019-04-21
Fixes emoji characters not displaying when letterSpacing style is used

themoonrat:
##### Description of change
https://github.com/pixijs/pixi.js/issues/5571
Changed fix to use Array.from as linting didn't like the use of for each over a string

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/Text.js
+++ b/packages/text/src/Text.js
@@ -261,9 +261,12 @@ export default class Text extends Sprite
 
         let currentPosition = x;
 
-        // using Array.from correctly splits characters whilst keeping emoji together
-        // https://medium.com/@giltayar/iterating-over-emoji-characters-the-es6-way-f06e4589516
-        const stringArray = Array.from(text);
+        // Using Array.from correctly splits characters whilst keeping emoji together.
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

Array.from is not supported in IE, need a polyfill in the polyfills package.

themoonrat:
Ahhhh sorry! I forget we're only in a semi es6 world.

themoonrat:
So, I've tweaked it so that it only uses Array.from if it exists.
There is no use polyfilling it, as it will still display incorrectly in IE; it requires the iterable powers of es6 that that just can't be polyfilled from my research.
It's also not _perfect_ at splitting emoji; 👩‍❤️‍💋‍👩 will get split into 👩❤💋👩.
To get it all perfect would required a big 100kb library like https://github.com/orling/grapheme-splitter which we're not going to be adding into PixiJS.

But at least in most scenarios things are _better_ with this PR. They're never worse, anyway!

PR #5621, opened 2019-04-28
Fixes transparent graphics do not perform ContainsPoint correctly

### LINKED ISSUE #5614: Transparent Graphics cannot trigger click(v5.0.0-rc3)
But it's normal in v4.

demos:
[v4](https://jsfiddle.net/JetLu/1fuLqjmo/)
[v5](https://jsfiddle.net/JetLu/yLozj731/)

JetLua:
Fixes #5614

It may not be a bug, just to be consistent with v4.

If there are side effects, close this PR.

[before](https://pixiplayground.com/#/edit/Wp5T42lDt0W9XHW5g5Z5k)
[after](https://pixiplayground.com/#/edit/3MoUUu4MD8~ZGpY25LaK_)

### DIFF
--- a/packages/graphics/src/GraphicsGeometry.js
+++ b/packages/graphics/src/GraphicsGeometry.js
@@ -364,11 +364,6 @@ export default class GraphicsGeometry extends BatchGeometry
         {
             const data = graphicsData[i];
 
-            if (!data.fillStyle.visible)
-            {
-                continue;
-            }
-
             // only deal with fills..
             if (data.shape)
             {


ivanpopelyshev:
That's the case: https://github.com/pixijs/pixi.js/blob/dev/packages/graphics/test/index.js#L153

It can become real case if we add one `stroke` call.

Anyway, thank you! You gave enough material to discuss with the rest of the team when we fix Graphics behaviour. Maybe we'll look whether Adobe Flash has the same behaviour.

PR #5623, opened 2019-04-29
Add childrenToName [suggest]

djmisterjon:
##### Description of change
Allow name of childrens , getChildByName are too heavy to call and bad performance.

Example:
```javascript
    initialize(){
        const master = new PIXI.Container();
        ['A','B','C','D'].forEach(n => {
            const cage = new PIXI.Container();
            cage.name = `blabla${n}`; // name the children
            master.addChild(cage);
        });
        master.childrenToName(); // store and ref child with name in .child
    };

// later , stuff ....
    event(e){
        const ee = e.currentTarget.child.blablaA; //good performance instead of getChildByName
        // ..... if no more need, maybe will need add a remover key in  removeChild?
        delete e.currentTarget.child.blablaA;
        e.currentTarget.removeChild(ee);
    };
```

- [ x] Documentation is changed or added
- [ x] Tests passed (`npm run test`)

### DIFF
--- a/packages/mixin-get-child-by-name/src/index.js
+++ b/packages/mixin-get-child-by-name/src/index.js
@@ -28,3 +28,23 @@ Container.prototype.getChildByName = function getChildByName(name)
 
     return null;
 };
+
+/**
+ * Add reference by name from children in child.
+ *
+ * @method childrenToName
+ * @memberof PIXI.Container#
+ * @return {Object.keys} Keys name.
+ */
+Container.prototype.childrenToName = function childrenToName()
+{   
+    this.child = {}; // clear
+    for (let i = 0; i < this.children.length; i++){
+        const name = this.children[i].name;
+        if(name){
+            if(object1.hasOwnProperty(name)){ throw console.error(`Conflic name alrealy exist ${name}`,this) };
+            this.child[name] = this.children[i]; // ref obj
+        }
+    };
+    return Object.keys(this.child);
+};
--- a/packages/mixin-get-child-by-name/src/index.js
+++ b/packages/mixin-get-child-by-name/src/index.js
@@ -42,7 +42,7 @@ Container.prototype.childrenToName = function childrenToName()
     for (let i = 0; i < this.children.length; i++){
         const name = this.children[i].name;
         if(name){
-            if(object1.hasOwnProperty(name)){ throw console.error(`Conflic name alrealy exist ${name}`,this) };
+            if(this.child.hasOwnProperty(name)){ throw console.error(`Conflic name alrealy exist ${name}`,this) };
             this.child[name] = this.children[i]; // ref obj
         }
     };


eXponenta:
Great idea, but if you know realisaton of it, without some problems, implement it and send PR.
Some problems:

1.  Childs haven't name in default. How store it in map?
2.  Childs can has same name. 
3.  You can change name of child after adding at in Container.
4.  How resolve naming conflict after re-adding without renaming?

djmisterjon:
> Great idea, but if you know realisaton of it, without some problems, implement it and send PR.
> Some problems:
> 

thank you, I am not very good in TS and PIXI background and I think not be the best person to add such a feature in pixi js core, but I would love to have this possibility in a module.
> 1. Childs haven't name in default. How store it in map?

I think It's should be a manual process when you initialize.
Because you can need this only in some context. ex: build and animate a big spriteSheet with multi parents and childs.
Only if you wish to access to this child, at initialize, you give him a `.name='uniqueName'`.
Then you call parent.childrenToName();, the method will register automatically all childrens with name int parent.child.
> 2. Childs can has same name.

Nop unique name only.
> 3. You can change name of child after adding at in Container.
yes, but need call again the parent `childrenToName`
```
child.name = 'newName';
child.parent.childrenToName(); // reDo the registe
```
> 4. How resolve naming conflict after re-adding without renaming?

it alway clear child reference and only add reference with name `this.child = {};`
If 2 child have same name, i add a alert in console.

bigtimebuddy:
I've discussed a similar idea with @andrewstart awhile ago about pixi-animate. I think a better approach here would be to create a simple object map on Container (maybe called `instances` or `namedChildren`) and add the named keys on that. Seems a little more future proof and less potential for de-optimizations or conflicts with existing props without requiring the check/error throwing.

```js
const child = new PIXI.DisplayObject();
child.name = 'foo';

container.addChild(child);

console.log(container.instances.foo === child);
```

This is a little trickier because we'd need to do some stuff on addChild/removeChild, but the best API is no API. Calling `childrenToName` seems cumbersome.

djmisterjon:
I like this approach and seems stable.
It would be great to see it in a near future update if any of you guys have time.
No very a priority or critical here, it more a enhancement.
Also good to know that this has already been discussed internally.

bigtimebuddy:
I'm going to close for now, since this current implementation probably isn't the best approach. If someone wants to take a crack at it for v5.1.0 with tests and all, I added the "Future Work" tag to this PR.

djmisterjon:
just add a update, i change little the ways to perform this, it this can interest people.
This it what the result i expect :)
image
This will give more power control if you have very complexe parenting displays objects and need tweens each sprites.
So just need give a name to sprite needed, and call one time the builder to get references to each sprites name in the `parent.child`.
I think it can be a very interesting features to add to pixijs and allow more control.
The code can be heavy but logicly you do this a loading time or boot app.
```javascript
/**
 * Add reference by name from children in child.
 *
 * @method childrenToName
 * @memberof PIXI.Container#
 * @return {Object.keys} Keys name.
 */
PIXI.Container.prototype.childrenToName = function childrenToName()
{   
    let child = this.child = {}; // clear
    let buffer = [this.children];
    let childrens = buffer.shift();
    while (childrens) {
        for (let i=0, l=childrens.length; i<l; i++) {
            const _child = childrens[i];
            if(_child.name){
                if(child.hasOwnProperty(_child.name)){ 
                    throw console.error(`Conflic name alrealy exist ${_child.name}`,child) 
                };
                child[_child.name] = _child;
            };
            if(_child.children.length){buffer.push(_child.children)}
        };
        childrens = buffer.shift();
    };
    return Object.keys(child);
};
```

djmisterjon:
btw a simple method tool like.
`this.child = PIXI.utils.childrenToName(display) ;`
No break change here , and return a objects with all display objects with name registered.
Another props can added like

`const c = new PIXI.Container().setName('Master');`

latter
`this.child.Master.position.set(x,y)`

djmisterjon:
i think give this as a tool method will be a nice idea instead inside a display object method.

PR #5631, opened 2019-04-30
Temporary disable createImageBitmap in resources

### LINKED ISSUE #5626: PIXI 5 - Sprites disappearing in RenderTextures after sprite texture unloaded?
I have noticed some of my sprites disappearing that are in render textures.  This did not happen on PIXI 4, only PIXI 5.

It has been a frustrating bug to evaluate as it is very intirmittent.  If the rendertexture is rendered constantly I do not see the bug.  However, the rendertextures with the problem are rendered infrequently.  When the problem occurs, if I call render on the rendertexture again in the next frame the sprites are back in the rendertexture again.

I realised only sprites which used textures not in the main scene were affected.  So I found one way to fix this was to place dummy sprites in the main scene, but hidden (like with a tiny scale).  This isn't really a solution though.  I did try called render on renderTexture twice but this did not seem to work.

I presume that anything that is rendered constantly is kept "fresh".  I soon learn about there is a special PIXI garbage collector removing old resources, though changing it's settings to manual seemed to have no effect.  I also did not want to manage it manually.  However, I did wonder if the texture is being unloaded from the GPU some other way.

I then looked at the sprites basetextures used by the sprites and noticed that values suggested it was being unloaded, confirming my assumptions:

The value baseTexture._eventsCount becomes "0" when unloaded.  Once active again baseTexture.dirtyId is incremented by one.  Hopefully this will mean something to the PIXI devs.

Since this didn't occur in PIXI 4, it seems PIXI 5 does not reload the texture back onto the GPU in time for it to be available when render is called on the renderTexture?

PIXI 5 RC 3
Chrome 73
Windows 10 64bit

I have made a fiddle...

https://jsfiddle.net/fanjules/h8dpgxw1/

Click render to render the render texture.  Pay attention to "_eventsCount" in the table.  When it becomes 0, click render again - and no sprite!   A second click and normality is restored since the texture is presumably in memory.   On PIXI 4 I presume the texture is reloaded to memory again in time before the render.

bigtimebuddy:
Partially fixes #5626 
There are some issues with using `createImageBitmap` for now, we'll disable and use v4 functionality.

### DIFF
--- a/packages/core/src/textures/resources/ImageResource.js
+++ b/packages/core/src/textures/resources/ImageResource.js
@@ -12,7 +12,7 @@ export default class ImageResource extends BaseImageResource
     /**
      * @param {HTMLImageElement|string} source - image source or URL
      * @param {boolean} [options.autoLoad=true] start loading process
-     * @param {boolean} [options.createBitmap=true] whether its required to create
+     * @param {boolean} [options.createBitmap=false] whether its required to create
      *        a bitmap before upload defaults true
      * @param {boolean} [options.crossorigin=true] - Load image using cross origin
      * @param {boolean} [options.premultiplyAlpha=true] - Premultiply image alpha in bitmap
@@ -58,7 +58,7 @@ export default class ImageResource extends BaseImageResource
          * @member {boolean}
          * @default PIXI.settings.CREATE_IMAGE_BITMAP
          */
-        this.createBitmap = options.createBitmap !== false && settings.CREATE_IMAGE_BITMAP && !!window.createImageBitmap;
+        this.createBitmap = (options.createBitmap || settings.CREATE_IMAGE_BITMAP) && !!window.createImageBitmap;
 
         /**
          * Controls texture premultiplyAlpha field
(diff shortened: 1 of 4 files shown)


ivanpopelyshev:
@fanjules congrats, 0th day critical issue of major release.

packages/core/src/textures/resources/ImageResource.js
```diff
@@ -58,7 +58,7 @@ export default class ImageResource extends BaseImageResource
          * @member {boolean}
          * @default PIXI.settings.CREATE_IMAGE_BITMAP
          */
-        this.createBitmap = options.createBitmap !== false && settings.CREATE_IMAGE_BITMAP && !!window.createImageBitmap;
+        this.createBitmap = (options.createBitmap || settings.CREATE_IMAGE_BITMAP) && !!window.createImageBitmap;
```
bigtimebuddy: @ivanpopelyshev does ☝️ this look okay?

ivanpopelyshev: You are right, we have to patch this logic

PR #5685, opened 2019-05-16
Fixes SVGResource can not trigger resize

### LINKED ISSUE #5682: SVG is displayed 0.5-scaled on height
### Expected Behavior
Sprite from SVG texture should be displayed with SVG's original defined aspect ratio

### Current Behavior
Sprite from SVG texture has it's height scaled down by 0.5. E.g. a circle 80x80 is displayed as if it's an ellipse 80x40.

### Possible Solution
I assume there's something wrong with the width and height properties of the SVGResource. I noticed that SVGResource passes a new canvas element to the parent here:
https://github.com/pixijs/pixi.js/blob/a314b2fcf6f4bd42f73421af38eedd88e0638b09/packages/core/src/textures/resources/SVGResource.js#L20

I checked in Chrome and `document.createElement('canvas')` creates a canvas that by default has a 300x150 size. So the height is half the width which is exactly what's happening to SVG sprites (see 'Current Behavior' above).

I couldn't find how that canvas size could be affecting the texture though.

**UPDATE:**
Logging the width and height of the sprite to the console shows 300x150! Same as the default canvas size in Chrome (see above).
Adding another circle that is 10 times bigger than the first one results in two equally-sized ellipses. The width and height of the second circle's sprite is also 300x150.
I've updated the playground code. Take a look at the browser console to see sprite's width and height for both circles.

### Steps to Reproduce
Just create a new texture from an inline SVG string and then create a sprite from that. Create a circle or a square so that you can clearly see the difference between width and height on rendering.
Also see the running example in the Pixi Playground (link below).

### Environment

- **`pixi.js` version**: 5.0.2
- **Browser & Version**: Chrome 74.0.3729.131
- **Running Example**: https://www.pixiplayground.com/#/edit/Og4Qr9Q6w5yyRTepEaNJs

JetLua:
close #5682

Demo: [https://www.pixiplayground.com/#/edit/Og4Qr9Q6w5yyRTepEaNJs](https://www.pixiplayground.com/#/edit/Og4Qr9Q6w5yyRTepEaNJs)

Fixed: [https://www.pixiplayground.com/#/edit/nvtwQS6GOGYzqkmzLPO9T](https://www.pixiplayground.com/#/edit/nvtwQS6GOGYzqkmzLPO9T)

### DIFF
--- a/packages/core/src/textures/resources/SVGResource.js
+++ b/packages/core/src/textures/resources/SVGResource.js
@@ -186,20 +186,20 @@ export default class SVGResource extends BaseImageResource
             }
 
             // Scale realWidth and realHeight
-            this._width = Math.round(svgWidth * this.scale);
-            this._height = Math.round(svgHeight * this.scale);
+            const width = Math.round(svgWidth * this.scale);
+            const height = Math.round(svgHeight * this.scale);
 
             // Create a canvas element
             const canvas = this.source;
 
-            canvas.width = this._width;
-            canvas.height = this._height;
+            canvas.width = width;
+            canvas.height = height;
             canvas._pixiId = `canvas_${uid()}`;
 
             // Draw the Svg to the canvas
             canvas
                 .getContext('2d')
-                .drawImage(tempImage, 0, 0, svgWidth, svgHeight, 0, 0, this.width, this.height);
+                .drawImage(tempImage, 0, 0, svgWidth, svgHeight, 0, 0, width, height);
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
I'm very happy that Resource API is good enough for people to patch it :) Its easier to make an approve.

bigtimebuddy:
Good fix @JetLua, thanks!

PR #5691, opened 2019-05-17
Change resolution of temp renderTexture for filter

ivanpopelyshev:
Demo: https://www.pixiplayground.com/#/edit/xwiTwNFC3Q8A5iWPHKkOw

Fixed: https://www.pixiplayground.com/#/edit/S7Om1JRG4jqxbLz0oZ9Lu

### DIFF
--- a/packages/core/src/renderTexture/RenderTexture.js
+++ b/packages/core/src/renderTexture/RenderTexture.js
@@ -131,13 +131,13 @@ export default class RenderTexture extends Texture
     }
 
     /**
-     * changes the resolution of baseTexture, but does not change framebuffer size
+     * Changes the resolution of baseTexture, but does not change framebuffer size.
      *
-     * @param {number} resolution
+     * @param {number} resolution - The new resolution to apply to RenderTexture
      */
     setResolution(resolution)
     {
-        const baseTexture = this.baseTexture;
+        const { baseTexture } = this;
 
         if (baseTexture.resolution === resolution)
         {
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Lets discuss the API. I think that if `renderTexture` already has its own `resize`, it should also have `setResolution`.

I have second PR coming right through after that, for more esoteric case when our fullscreen-filter optimization can force weird things.

packages/core/src/renderTexture/RenderTexture.js
```diff
@@ -130,6 +130,24 @@ export default class RenderTexture extends Texture
...
     }
 
+    /**
+     * changes the resolution of baseTexture, but does not change framebuffer size
+     *
+     * @param {number} resolution
+     */
+    setResolution(resolution)
+    {
+        const baseTexture = this.baseTexture;
```
bigtimebuddy:
```suggestion
        const { baseTexture } = this;
```

packages/core/src/renderTexture/RenderTexture.js
```diff
@@ -130,6 +130,24 @@ export default class RenderTexture extends Texture
         this.updateUvs();
     }
 
+    /**
+     * changes the resolution of baseTexture, but does not change framebuffer size
```
bigtimebuddy:
```suggestion
     * Changes the resolution of baseTexture, but does not change framebuffer size.
```

packages/core/src/renderTexture/RenderTexture.js
```diff
@@ -130,6 +130,24 @@ export default class RenderTexture extends Texture
...
+     *
+     * @param {number} resolution
```
bigtimebuddy:
```suggestion
     * @param {number} resolution - The new resolution to apply to RenderTexture
```

PR #5695, opened 2019-05-19
touchmoves will now have the 'target' property set

themoonrat:
Ever since the 'moveWhenInside' property was added - https://github.com/pixijs/pixi.js/commit/24b1580124496c547280026ad65a77ec671188f9 -  for whatever reason, it's asked to not do the processInteractive _without_ doing a hitTest.
I have no idea why. Without it, all moves due to a touchevent give a 'target' property of null, even if the move IS over an object - showing bugs like https://github.com/pixijs/pixi.js/issues/4887
This doesn't change the way 'moveWhenInside' works at all.... it's just enabling the this 'target' property to be filled via allowing a hitTest.

@GoodBoyDigital You first added this over 3 years ago now.. I wonder if you can remember any good reason you didn't want a hitTest to occur for touch events only on touchmove?


GoodBoyDigital:
Hey ya! Honestly, i can't even remember what it was for :/
Im sure there was a reason, but you are much closer to this part of the codebase than I am for sure. If it makes sense to remove, please do! 👍

GoodBoyDigital:
@themoonrat You know I finally remembered what it was for! It was a performance thing. I found that most of the time mouse / touch move hit testing was not important so I decided to not perform the test. (The idea was that the user could build the functionality by combining this with mouse over / mouse out)

bigtimebuddy:
@themoonrat what do you want to do about this PR?

themoonrat:
Considering desktop does that test anyway without any performance complaints, and devices are far more powerful now than they were when the optimisation was added.... I think we're ok for this PR to go through? Don't think there's another way of solving the issue reported, otherwise

PR #5697, opened 2019-05-20
Normalizes SVGResource parsing to support inline SVG strings, base64 data URIs, and URLs

### LINKED ISSUE #5693: SVG loading breaks when file contains hash (#)
### Steps to Reproduce

Try to load a SVG file with colors specified as hashes (e.g. `#fff` instead of `rgb(255, 255, 255)`).

Example: https://www.pixiplayground.com/#/edit/lcocUvQS6jzQc148ZVWb7

### Expected Behavior

Texture should load without issues.

### Current Behavior

Loading texture fails with `Unable to load image from: ...`

### Possible Solution

Looking at the code, the data URI is not being encoded at all (either with `encodeURIComponent` or as Base64). This seems incorrect, because `#` is illegal in data URIs.

### Environment

- **`pixi.js` version**: 5.0.2
- **Browser & Version**: Firefox 66.0.5, Chrome 74.0.3729.157 
- **Running Example**: https://www.pixiplayground.com/#/edit/lcocUvQS6jzQc148ZVWb7

This is new in Pixi v5. Pixi v4 loads the same SVG file without issue (change the version in pixi-playground link to verify).

JetLua:
close #5693 and more.

Demo: [https://jsfiddle.net/JetLu/js6q2ekz/](https://jsfiddle.net/JetLu/js6q2ekz/)

Fixed: [https://jsfiddle.net/JetLu/6so9p2d1/](https://jsfiddle.net/JetLu/6so9p2d1/)

1. We can remove `SVGResource.getSize(svgString)` because we can get SVG dimensions in `onload` event.

2. From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa#Unicode_strings)

> In most browsers, calling btoa() on a Unicode string will cause an InvalidCharacterError exception.

```js
/* hack */
this.svg = `data:image/svg+xml;base64,${btoa(unescape(encodeURIComponent(this.svg)))}`;
```

Now `SVGResource` supports(don't care if there's Unicode):
 - inline: `<svg>...</svg>`
 - base64: `data:image/svg+xml;base64,...`
 - url: `//domain/*.svg`


ivanpopelyshev:
Man, you are on fire! What's wrong with the tests though?

JetLua:
Resolved. I love Pixi.js! 😄

bigtimebuddy:
I dig this normalization of the SVG inputs. Much simpler.

My main concern is that this is a breaking change as this PR currently stands. It removes `SVGResource.getSize` (in v4 this was `PIXI.utils.getSvgSize`). It's important that we provide a backward-compatible deprecation for this (see **bundles/pixi.js/src/useDeprecated.js**). I feel uncomfortable removing this if people were using is calculate sizes.

JetLua:
Yep, it was my negligence.

KTM450SXF:
Thanks, guys.  I'm hung up on this one and literally, went to report it about two hours after Jet did and then saw his report.  You guys get things fixed quickly.

I'll take Jet's edit for now and forge ahead, but I'm curious to know when this one will be released.

And if I haven't said it explicitly enough - Thanks for what you do.  PIXI is awesome.

packages/core/src/textures/resources/SVGResource.js
```diff
@@ -40,6 +40,13 @@ export default class SVGResource extends BaseImageResource
          */
         this._resolve = null;
 
+        /**
+         * Cross origin value to use
+         * @private
+         * @member {boolean|string}
+         */
+        this._crossorigin = options.crossorigin || true;
```
bigtimebuddy: This seems like a problem, `crossorigin` will always be true here even if someone sets `false`. I would just leave this as `undefined` and let `BaseImageResource.crossOrigin` handle it.
```suggestion
        this._crossorigin = options.crossorigin;
```

PR #5706, opened 2019-05-21
Adds factory for creating custom BatchRenderer plugins

ivanpopelyshev:
This feature is required by multiple plugins.

Demo will follow.

Need help with names and structure of the API.


ivanpopelyshev:
Demo: https://www.pixiplayground.com/#/edit/gyEI8Qi6~ii7hMwCTUO7k

GoodBoyDigital:
I think you are really on to something here dude! API does need a little work, but the principles are spot on!

bigtimebuddy:
This is basically currently your implementation from the demo. This feels overly burdensome. It requires that the dev knows about `BatchShaderGen`, `BatchRenderer`, `shaderGen` property and how they all fit together.

```js
const myShaderGen = new PIXI.BatchShaderGen(undefined,
`
varying vec2 vTextureCoord;
varying vec4 vColor;
varying float vTextureId;
uniform sampler2D uSamplers[%count%];

void main(void){
    vec4 color;
    %forloop%
    gl_FragColor = vColor * vec4(color.a - color.rgb, color.a);
}
`);

class MyBatch extends PIXI.BatchRenderer {
    constructor(renderer) {
        super(renderer);
        this.shaderGen = myShaderGen;
    }
}

PIXI.Renderer.registerPlugin('invert', MyBatch);
const sprite = new PIXI.Sprite();
sprite.pluginName = 'invert';
```

Maybe we could create a Factory method or class that generates the new BatchRenderer, like this. Then devs essentially only need to know about a single new API.

```js
const vertex = undefined;
const fragment = `
varying vec2 vTextureCoord;
varying vec4 vColor;
varying float vTextureId;
uniform sampler2D uSamplers[%count%];

void main(void){
    vec4 color;
    %forloop%
    gl_FragColor = vColor * vec4(color.a - color.rgb, color.a);
}
`;

PIXI.Renderer.registerPlugin('invert', PIXI.BatchRendererFactory.create(vertex, fragment));
const sprite = new PIXI.Sprite();
sprite.pluginName = 'invert';
```

ivanpopelyshev:
Yes, that'll be better for basic usage. For advanced, we still need to re-declare shaderGen in BatchRenderer constructor.

bigtimebuddy:
Also, I'd love to see the more verbose "Generator", not shorting to "Gen". PixiJS typically only abbreviates API names if it's common usage ("gl", "uvs", etc). I don't think "Gen" is a commonly understood shortening.

bigtimebuddy:
@ivanpopelyshev how do you want to proceed with this? Do you want me to take a pass?

ivanpopelyshev:
Yes, because im on vacation and my time online is limited by my fun :)

bigtimebuddy:
Here's a fork of your example with the new API. https://www.pixiplayground.com/#/edit/QYx55VopCd0rNFnc_gTYl

* Was toying with `BatchRendererFactory` vs `BatchPluginFactory`? Not bullish on either.
* Can `BatchShaderGenerator` be private now?

ivanpopelyshev:
1. I think BatchPluginFactory is better.
2. It has to be publish, because we can generate better stuff than that. For example, two textures per sprite. Texture dimensions supplied in uniforms.

GoodBoyDigital:
ability to pass in uniforms would be ace too!

bigtimebuddy:
@GoodBoyDigital You mean these uniforms?

```js
        const uniforms = {
            tint: new Float32Array([1, 1, 1, 1]),
            translationMatrix: new Matrix(),
            default: this.defaultGroupCache[maxTextures],
        };

        return new Shader(this.programCache[maxTextures], uniforms);
```

ivanpopelyshev:
I have `customUniforms` in my plugins. If several sprites have same group of uniforms, they are batched.

https://github.com/pixijs/pixi-projection/blob/master/src/curve/SpriteBilinearRenderer.ts#L124

You think its worth adding in next PR for next version? :)

GoodBoyDigital:
thats right! so you could for example set up new uniforms in the fragment source:

```
const uniforms = {extraColor:[1,0,0,1]};

const options = {fragment, uniforms};

BatchRendererFactory.create(options);

```
const uniforms = {
            tint: new Float32Array([1, 1, 1, 1]),
            translationMatrix: new Matrix(),
            default: this.defaultGroupCache[maxTextures],
            customUniforms: UniformGroup.from(uniforms);
        };
```

ivanpopelyshev:
Lets do it in next version? :) I'm not sure about uniforms API yet.

bigtimebuddy:
I tend to agree with @ivanpopelyshev, I think we could do a follow-up for uniforms. Doesn't change the overall approach here. Do you have some specific ideas how devs might use this feature @GoodBoyDigital?

PR #5710, opened 2019-05-22
added gl.UNSIGNED_INT functionality

### LINKED ISSUE #5699: [Feature Request]Support OES_element_index_uint
any plan to support gl.UNSIGNED_INT in gl.drawElements?

https://github.com/pixijs/pixi.js/blob/ba03f57001cd8bae3ae3644911cd513c68184e7c/packages/core/src/geometry/GeometrySystem.js#L594-L606

see https://developer.mozilla.org/en-US/docs/Web/API/OES_element_index_uint

abhinavsagar:
[Fixes #5699 ](https://github.com/pixijs/pixi.js/issues/5699)


ivanpopelyshev:
Is it supported everywhere?

abhinavsagar:
Yes, I think so. Did i miss something ?

Norgerman:
- should check whether webgl1 supports `OES_element_index_uint`
- should check the indexbuffer uses Uint16Array or Uint32Array, not just draw twice
- the offset is `2 * start` for Uint16Array,` 4 * start` for Uint32Array.

bigtimebuddy:
Could someone explain the benefit of this PR?

Norgerman:
Just add support for gl.UNSIGNED_INT in gl.drawElements calls, but this PR is not correct.

packages/core/src/geometry/GeometrySystem.js
```diff
@@ -588,21 +588,32 @@ export default class GeometrySystem extends System
     {
         const { gl } = this;
         const geometry = this._activeGeometry;
+        const uints_for_indices = gl.getExtension("OES_element_index_uint");
```
GoodBoyDigital: We keep the extension references in the `ContextSystem` class. There is a function called `getExtensions()` We should add it there as this check only happens once on startup. Placing it here means we check it for each draw call!

packages/core/src/geometry/GeometrySystem.js
```diff
@@ -588,21 +588,32 @@ export default class GeometrySystem extends System
...
 
         // TODO.. this should not change so maybe cache the function?
 
         if (geometry.indexBuffer)
         {
-            if (geometry.instanced)
```
GoodBoyDigital: should we be checking for the geometry.indexBuffer.data array type?

also instead of having 4 different checks we could use:
```
const byteSize = geometry.indexBuffer.BYTES_PER_ELEMENT;
const arrayType = byteSize === 2 ?  gl.UNSIGNED_SHORT :  gl.UNSIGNED_INT :

...
gl.drawElementsInstanced(type, size || geometry.indexBuffer.data.length, arrayType, (start || 0) * byteSize, instanceCount || 1);
...
```

bigtimebuddy:
Closing in favor of #5738

PR #5738, opened 2019-05-29
add support for uint32 element index

Norgerman:
##### Pre-Merge Checklist

#5699

### DIFF
--- a/packages/core/src/geometry/GeometrySystem.js
+++ b/packages/core/src/geometry/GeometrySystem.js
@@ -586,22 +586,48 @@ export default class GeometrySystem extends System
      */
     draw(type, size, start, instanceCount)
     {
-        const { gl } = this;
+        const { gl, renderer: { context } } = this;
         const geometry = this._activeGeometry;
 
         // TODO.. this should not change so maybe cache the function?
 
         if (geometry.indexBuffer)
         {
-            if (geometry.instanced)
+            const byteSize = geometry.indexBuffer.data.BYTES_PER_ELEMENT;
+
+            if (byteSize === 2)
             {
-                /* eslint-disable max-len */
-                gl.drawElementsInstanced(type, size || geometry.indexBuffer.data.length, gl.UNSIGNED_SHORT, (start || 0) * 2, instanceCount || 1);
-                /* eslint-enable max-len */
+                if (geometry.instanced)
+                {
+                    /* eslint-disable max-len */
+                    gl.drawElementsInstanced(type, size || geometry.indexBuffer.data.length, gl.UNSIGNED_SHORT, (start || 0) * byteSize, instanceCount || 1);
+                    /* eslint-enable max-len */
+                }
+                else
+                {
+                    /* eslint-disable max-len */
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Looks good. Lets see if we can test it somehow...

packages/core/src/geometry/GeometrySystem.js
```diff
@@ -586,22 +586,48 @@ export default class GeometrySystem extends System
...
+        const { gl, renderer: { context } } = this;
         const geometry = this._activeGeometry;
 
         // TODO.. this should not change so maybe cache the function?
 
         if (geometry.indexBuffer)
         {
-            if (geometry.instanced)
+            const byteSize = geometry.indexBuffer.data.BYTES_PER_ELEMENT;
+
```
GoodBoyDigital: could add:
```const glType = byteSize === 2 ? gl.UNSIGNED_SHORT : gl.UNSIGNED_INT```

then we don't need the extra  if block

File: packages/core/src/geometry/GeometrySystem.js [hunk 2]
```diff
@@ -586,22 +586,48 @@ export default class GeometrySystem extends System
...
+            if (byteSize === 2)
+            {
+                if (geometry.instanced)
+                {
+                    /* eslint-disable max-len */
+                    gl.drawElementsInstanced(type, size || geometry.indexBuffer.data.length, gl.UNSIGNED_SHORT, (start || 0) * byteSize, instanceCount || 1);
```
GoodBoyDigital: And this could become:

```gl.drawElementsInstanced(type, size || geometry.indexBuffer.data.length, glType, (start || 0) * byteSize, instanceCount || 1);```

Comment (Norgerman) [reply packages/core/src/geometry/GeometrySystem.js]: But also I need some where to check whether I can use uint32.
https://github.com/pixijs/pixi.js/pull/5738/files#diff-16f6fe0290c560188da725c2135c1eaaR613

File: packages/core/src/geometry/GeometrySystem.js (hunk 1 above)
Comment (Norgerman) [reply packages/core/src/geometry/GeometrySystem.js]: updated

File: packages/core/src/geometry/GeometrySystem.js [hunk 3]
```diff
@@ -586,22 +586,34 @@ export default class GeometrySystem extends System
...
 
         // TODO.. this should not change so maybe cache the function?
 
         if (geometry.indexBuffer)
         {
-            if (geometry.instanced)
+            const byteSize = geometry.indexBuffer.data.BYTES_PER_ELEMENT;
+            const glType = byteSize === 2 ? gl.UNSIGNED_SHORT : gl.UNSIGNED_INT;
+
+            if (byteSize === 2 || context.webGLVersion === 2 || context.extensions.uintElementIndex)
```
GoodBoyDigital: Could add this line at the top when the context is created:
```this.canUseUInt32 = context.webGLVersion === 2 || context.extensions.uintElementIndex;```

Then this line could become:
```if(byteSize === 2 || (this.canUseUInt32 && byteSize === 4))```

ivanpopelyshev:
Test is here: https://www.pixiplayground.com/#/edit/11AT1kMcq69qWKJvX67Wb

We dont have build anywhere so I just took PR changes inside playground itself.

PR #5740, opened 2019-05-29
[V4.x] Blendmode ADD again

ivanpopelyshev:
Related to #5739 and #5563

Im trying to contact @Puuca to see if BlendMode breaks his case on iPads


Puuca:
tested on iPad Pro in Safari and Chrome. Seems that #5563 is still fixed.

Issue #4538, opened 2017-12-14
Experimental: OffscreenCanvas


ruslanchek:
Hi all! How about some experiments with OffscreenCanvas API? I think it could be useful in the future. For example it can be used for generating text-sprites for Text() objects, it can highly improve performance.   

[OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas)

bigtimebuddy:
Related: #4212 and #4207

The discussion in those PRs is useful for understanding how we think about experimental API features. OffscreenCanvas is definitely not something we are planning to add support for in v4.  For v5, we are focused mostly on taking advantage of WebGL 2 support whenever possible to improve performance.

If someone can show some real-world use cases for OffscreenCanvas and the performance benefits we'd gain by implementing, then we should consider it. There are different levels of utilizing OffscreenCanvas: 1). might be using it under-the-hood to help speed up rendering operations, and 2). supporting OffscreenCanvas via the PixiJS API for things like providing a view when creating a new renderer/application.

Would like to hear some specific examples of how we might take advantage of OffscreenCanvas. The example above for using with Text wouldn't really work because Text is context2d-only and OffscreenCanvas only works with WebGL.

tekintatar:
+1 for this. I guess we can use OffscreenCanvas for extract method already build in PIXI. Offscreen canvas has a direct method of transferToImageBitmap which should be faster compared to creating buffer, readPixels and than creating a 2d canvas and getting base64. At least in theory for now.

Right here: http://pixijs.download/dev/docs/extract_webgl_WebGLExtract.js.html#line70. 

https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/transferToImageBitmap

I will do more investigation and will try to create some case and compare performance.

johnnymast:
+1 want this !!

themoonrat:
@tekintatar Did you ever manage to create some benchmarks that demonstrated the usefullness of this API within PIXI? Thanks!

eranimo:
Why was this closed?

sanbornhilland:
I'm also curious why this was closed. Was it moved somewhere else or have you decided that `OffscreenCanvas` support is not a priority? 

[...]

PR #5756, opened 2019-06-05
Allow OffscreenCanvas to work as CanvasResource

mudcube:
##### Description of change
Allows OffscreenCanvas to be used as a source for CanvasResource.

The code references OffscreenCanvas with the `global.` prefix, as some browsers do not yet support this feature. Also, `source` is checked to be defined for the same reason. There's a few ways this could be setup, but this looked the easiest to read to me. Updated JSDoc to match changes.

##### Pre-Merge Checklist
- [x] Documentation is changed or added.


bigtimebuddy:
Reference issue #4538.

Thanks for the PR, @mudcube.

There's a linting problem, if could please fix that. Also, at the very least we need an example to verify that this works. Would you mind putting something together? One better would be to create an automated test that makes sure a resource can be created from an OffscreenCanvas. Our test environment should support this.

mudcube:
Hi @bigtimebuddy, my apologies for overlooking those points, the latest revision should resolve the linting issue. Here's a test that currently fails, but works with the modifications in this pull request: https://www.pixiplayground.com/#/edit/Z5fNHO7iKnHhOsOS~Q956

ivanpopelyshev:
I've modified your example.

OK, we cant test PR's from other repos, only through hacks.

Can you make it that way it actually shows something with that texture?

rjenk88:
High gestures are most sequence to supurb nest creating a vision at posting noble Forbe based featuring motive s

mudcube:
@ivanpopelyshev I've updated the example to load the texture into a PIXI.Sprite and display that onto screen, so you can visually see the modification sorts out the issue.

packages/core/src/textures/resources/CanvasResource.js
```diff
@@ -13,11 +13,14 @@ export default class CanvasResource extends BaseImageResource
...
-     * @param {*} source - The source object
+     * @param {HTMLCanvasElement|OffscreenCanvas} source - The source object
      * @return {boolean} `true` if <canvas>
      */
     static test(source)
     {
-        return (source instanceof HTMLCanvasElement);
+        return (source && (
+            source instanceof HTMLCanvasElement
+            || source instanceof global.OffscreenCanvas
```
themoonrat: the 'global' namespace may not exist.  Shouldn't it work as just `source instanceof OffscreenCanvas`?

mudcube: > the 'global' namespace may not exist

Ah, yes, I see that `global` didn't make it into the final specs... Do you think `globalThis` or `window` would be a better option?

bigtimebuddy: `window` works, we use that in other places

bigtimebuddy:
Awesome, great work 🎉

JetLua:
#5780

PR #5769, opened 2019-06-09
Do not throw error when calling Sprite.destroy() more than once

mudcube:
##### Description of change
This update makes Sprite.destroy() perform like Container.destroy() or any other DisplayObject.destroy(). Before this update an error was thrown as `_texture` does not exist. This update simply checks for where the `_texture` exists.

Here's a test: 
https://www.pixiplayground.com/#/edit/tfnFfI01kNpXKVTuPH1c5

### DIFF
--- a/packages/sprite/src/Sprite.js
+++ b/packages/sprite/src/Sprite.js
@@ -463,20 +463,23 @@ export default class Sprite extends Container
     {
         super.destroy(options);
 
-        this._texture.off('update', this._onTextureUpdate, this);
-
-        this._anchor = null;
-
-        const destroyTexture = typeof options === 'boolean' ? options : options && options.texture;
-
-        if (destroyTexture)
+        if (this._texture)
         {
-            const destroyBaseTexture = typeof options === 'boolean' ? options : options && options.baseTexture;
+            this._texture.off('update', this._onTextureUpdate, this);
 
-            this._texture.destroy(!!destroyBaseTexture);
+            const destroyTexture = typeof options === 'boolean' ? options : options && options.texture;
+
+            if (destroyTexture)
+            {
+                const destroyBaseTexture = typeof options === 'boolean' ? options : options && options.baseTexture;
+
+                this._texture.destroy(!!destroyBaseTexture);
+            }
+
+            this._texture = null;
         }
 
-        this._texture = null;
+        this._anchor = null;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Could you add a test for this? Want to make sure we dong regress that you can destroy after manually nulling texture.

themoonrat:
FYI, I did a PR to handle something like this ages back for v4, and it got rejected as "destroying a sprite that is already destroyed _should_ be an error"

https://github.com/pixijs/pixi.js/pull/3039

ivanpopelyshev:
Lets leave it as error, its one of things that users can patch in their pixi themselves. We use JS, its not a problem ;)

```js
PIXI.Sprite.prototype.destroy = ...
```

bigtimebuddy:
@mudcube what do you mean make it perform like Container and DisplayObject?

ivanpopelyshev:
@bigtimebuddy they can be destroyed second time.

mudcube:
> Could you add a test for this? Want to make sure we dong regress that you can destroy after manually nulling texture.

@bigtimebuddy When you say test, do you mean an [integration test](https://github.com/pixijs/pixi.js/tree/dev/tools/integration-tests/test)? I did post a PixiPlayground test illustrating the issue (following the format of my previous Pull Request). Sorry, new to this whole process, but will get there ;)

> @mudcube what do you mean make it perform like Container and DisplayObject?
> @bigtimebuddy they can be destroyed second time.

@bigtimebuddy @ivanpopelyshev Exactly, Sprite is currently a special case, I believe it's the only type of DisplayObject that cannot have `.destroy()` called on it multiple times. This pull request is meant to smooth out this inconsistency in the API.

bigtimebuddy:
There's nothing explicit in Container or DisplayObject to _prevent_ destroying twice, it just happens to not throw an error in most cases. Therefore, I don't think this PR provides consistency.

For reasons that we discussed in #3039 (as @themoonrat) pointed out, we feel that throwing an error is a feature not a bug. If developers want to prevent their projects from this, they can extend Sprite or mixin a destroy function that Ivan suggested. I would, however, welcome a PR that made this _more_ explicit for developers via documentation or maybe throwing an error.

For these reasons, unfortunately, we need to close this PR. Thanks for making the suggestion.

PR #5770, opened 2019-06-09
Floats internal format for webgl1 and webgl2

ivanpopelyshev:
Not everyone has FLOATs and HALF_FLOATs, especially when they are used in a color buffer.

For now we assume that its pixi user job to notify users that game needs better GL version, or to make a fallback.

The problem is that HALF_FLOAT_OES from webgl1 differs from webgl2.

We assume that float textures are used through BufferResource or RenderTexture. If somehow they are used in another resource (cubemap for example), we'll have to change its texture upload code to `glTexture.internalFormat` and `glTexture.type`.

Also, I don't know how to upload half-floats yet, but we can use them in colorbuffer.


ivanpopelyshev:
new BlurFilter implementation that uses FLOATs: https://www.pixiplayground.com/#/edit/BshXsyfWqJ8I3YFoGEuGV

Try FLOAT/HALF_FLOAT, Webgl1/WebGL2

I know how to make the filter work without those extensions, but right now its a good test.

bigtimebuddy:
@ivanpopelyshev could you resolve conflicts?

ivanpopelyshev:
@bigtimebuddy done

PR #5771, opened 2019-06-09
Move renderTexture pool from FilterSystem

ivanpopelyshev:
Some filters can require custom pools, made from texture of different format.

It also allows for a plugin to override FilterSystem texturePool behaviour.

Who knows, maybe soon we'll need a pool of multi-sample framebuffers or sRGB stuff!


ivanpopelyshev:
Custom pool usage for FLOAT textures: https://www.pixiplayground.com/#/edit/4rskiJ8lb8k6437WmSwKy 

Need demo for filter-heavy scene to check for other stupid errors

packages/core/src/renderTexture/RenderTexturePool.js
```diff
@@ -0,0 +1,175 @@
+import RenderTexture from './RenderTexture';
+import BaseRenderTexture from './BaseRenderTexture';
+import { nextPow2 } from '@pixi/utils';
+
+/**
+ * Experimental!
+ *
+ * Texture pool, used by FilterSystem and plugins
+ * Stores collection of temporary pow2 or screen-sized renderTextures
```
bigtimebuddy: Do you want to add a doc `@class` etc for this?

packages/core/src/renderTexture/RenderTexturePool.js
```diff
@@ -0,0 +1,177 @@
...
+    returnFilterTexture(renderTexture)
+    {
+        this.returnTexture(renderTexture);
+    }
+
+    /**
+     * Clears the pool
+     *
+     * @member {boolean} [destroyTextures=true] destroy all stored textures
+     */
```
GoodBoyDigital: could this be:
`clear(destroyTextures = false)`

packages/core/src/renderTexture/RenderTexturePool.js
```diff
@@ -0,0 +1,177 @@
...
+                textures[j].destroy(true);
+            }
+        }
+        this.texturePool[screenKey] = [];
+
+        this._pixelsWidth = this.renderer.view.width;
+        this._pixelsHeight = this.renderer.view.height;
+    }
+}
+
```
GoodBoyDigital: what is Screen Key used for? Mind adding some docs for it?

packages/core/src/renderTexture/RenderTexturePool.js
ivanpopelyshev: no, its true by default, like in previous version.

packages/core/src/renderTexture/RenderTexturePool.js
ivanpopelyshev: ok

packages/core/src/renderTexture/RenderTexturePool.js
ivanpopelyshev: i dont know how to do that part, and its internal thing for a custom class, we can change it later.

ivanpopelyshev:
I've so added `ISize` type, that way we don't need to pass renderer to renderTexturePool.

ivanpopelyshev:
https://www.pixiplayground.com/#/edit/R9DigjpNQlmI~vrs~4j~T

output is 1, 0, 1

PR #5776, opened 2019-06-11
SVGResource constructor has absolute size options 

tobyspark:
##### Description of change
SVG assets are resolution independent. At present, there is no way to precisely set the resolution they rasterise at during initialisation. Subsequent resizing will scale this fixed resolution asset – leading to blurriness etc. 

While a `scale` multiplier is provided as a constructor option to alter the fixed render size, this cannot be used to set an exact size as the SVG size is unknown before instantiation. This PR provides alternative constructor options to set size directly.

##### Pre-Merge Checklist


tobyspark:
Fixes underlying need of #5754

ivanpopelyshev:
Need our SVG specialist here :) @JetLua yo! 

I think there might be something wrong with that `round` , give me some time to think about it.

tobyspark:
The `round` was already in the code. If I were coding this from scratch, it would have been a `ceil`, as the canvas should contain whatever is drawn without clipping?

ivanpopelyshev:
Yes, but that also means that we need to use `scale`, not just `width/height` because we adjust aspect ratio slightly. Also just `ceil` isnt that good, Need eps too, like here: https://github.com/pixijs/pixi.js/blob/dev/packages/math/src/shapes/Rectangle.js#L216

tobyspark:
On `resolution` – I did wonder about that. As a general design pattern, could passing in an optional rendering destination allow the particular `from` method to act appropriately? Here it could be to set the rasterisation to e.g. retina, but perhaps there are other optimisations. Though this sounds increasingly like a separate PR.

JetLua:
> Need our SVG specialist here :) @JetLua yo!
Ah，thanks.

packages/core/src/textures/resources/SVGResource.js
```diff
@@ -115,9 +131,27 @@ export default class SVGResource extends BaseImageResource
...
             }
 
-            // Scale realWidth and realHeight
-            const width = Math.round(svgWidth * this.scale);
-            const height = Math.round(svgHeight * this.scale);
+            // Set render size
+            let width = svgWidth * this.scale;
+            let height = svgHeight * this.scale;
+
+            if (this._overrideWidth && this._overrideHeight)
```
bigtimebuddy: What about this as a more terse version?
```suggestion
            if (this._overrideWidth || this._overrideHeight)
            {
                width = this._overrideWidth || this._overrideHeight / svgHeight * svgWidth;
                height = this_overrideHeight || this._overrideWidth / svgWidth * svgHeight;
            }
```

tobyspark: More terse

tobyspark:
@bigtimebuddy Good oversight. I've pushed two commits to address your points. As part of the documentation commit, I also changed the language from 'render' to 'rasterize'. More precise, and I'd say helpful to separate out from 'render' as in the ongoing rendering of the scene.

PR #5788, opened 2019-06-16
Make createIndicesForQuads public, add uint32

ivanpopelyshev:
Subj. Follow-up for #5738 . Lets discuss the API.

### DIFF
--- a/packages/utils/src/data/createIndicesForQuads.js
+++ b/packages/utils/src/data/createIndicesForQuads.js
@@ -1,30 +1,34 @@
 /**
  * Generic Mask Stack data structure
  *
- * @memberof PIXI
+ * @memberof PIXI.utils
  * @function createIndicesForQuads
- * @private
  * @param {number} size - Number of quads
- * @return {Uint16Array} indices
+ * @param {Uint16Array|Uint32Array} [outBuffer] - Buffer for output, length has to be `6 * size`
+ * @return {Uint16Array|Uint32Array} - Resulting index buffer
  */
-export function createIndicesForQuads(size)
+export function createIndicesForQuads(size, outBuffer = null)
 {
     // the total number of indices in our array, there are 6 points per quad.
-
     const totalIndices = size * 6;
 
-    const indices = new Uint16Array(totalIndices);
+    outBuffer = outBuffer || new Uint16Array(totalIndices);
+
+    if (outBuffer.length !== totalIndices)
+    {
+        throw new Error(`Out buffer length is incorrect, got ${outBuffer.length} and expected ${totalIndices}`);
+    }
 
     // fill the indices with the quads to draw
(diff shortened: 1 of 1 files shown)


englercj:
Rather than a `32-bit` param, what if it just took the buffer and offset as parameters?

```
function createIndicesForQuads(size, outBuffer, offset)
{
    // the total number of indices in our array, there are 6 points per quad.
    const totalIndices = size * 6;

    if ((outBuffer.length - offset) < totalIndices)
        return false;

    // fill `outBuffer`

    return true;
}
```

Another option would be to pass the object to be created as a parameter:

```
function createIndicesForQuads(size, ArrayType)
{
    // the total number of indices in our array, there are 6 points per quad.
    const totalIndices = size * 6;
    const buffer = new ArrayType(totalIndices);

    // fill `buffer`

    return buffer;
}
```

ivanpopelyshev:
> Another option would be to pass the object to be created as a parameter:

Will make our typings harder. There's only Uint16 or Uint32 , we dont need anything else.

offset is kinda good idea, but maybe it needs new function? `fillIndicesForQuads(arr, size, offset1, offset2);`

ivanpopelyshev:
Oh, right, I forgot to say that this thing is mostly for plugins.

bigtimebuddy:
This would be my suggestion.  This is backward compatible, and supports providing Uint32Array. I think it's better than using a boolean.

If `offset` is useful down the road, we can add it, but I don't think we need to pile-on this simple change.

```js
function createIndicesForQuads(size, outBuffer = null)
{
    // the total number of indices in our array, there are 6 points per quad.
    const totalIndices = size * 6;

    outBuffer = outBuffer || new Uint16Array(totalIndices);

    if (outBuffer.length !== totalIndices)
        throw new Error(`Out buffer length is incorrect, got ${outBuffer.length} and expected ${totalIndices}`);

    // fill `outBuffer`
}
```

hochuen0:
Found these in my files, anyone can help me understand it's function?
  FSCK0134.REC
<https://drive.google.com/file/d/0B2J8gZH-_miSQXRSSE8tOXhYMWxqNk1tSDVpb3prRUxtLUFv/view?usp=drivesdk>

  FSCK0127.REC
<https://drive.google.com/file/d/0B2J8gZH-_miSMmptajdsNXJTdzB5VmJfdmxXQVlrS19ZQndr/view?usp=drivesdk>

  FSCK0129.REC
<https://drive.google.com/file/d/0B2J8gZH-_miSLUZEcDB6d0lUNjdURFNvXzlJcTd3T0Jsa0hz/view?usp=drivesdk>
Another suspicious file.

2019-06-05 20:41:01,221 INFO  [ChangeOverTAG ]-[127] VersionUtilssPairedVersion == null, not init

ivanpopelyshev:
@bigtimebuddy OK, done, rebased. We can make a better function with both kind of offsets later.

PR #5791, opened 2019-06-17
Feature/set state rename

GoodBoyDigital:
Small tweak, but feel set looks better and feels better to write:

`renderer.state.set(myState);`

vs

`renderer.state.setState(myState);`

small tweak!


bigtimebuddy:
I agree with you @themoonrat, fixed.

PR #5792, opened 2019-06-17
Bump floss for 3.0 & codecov integration

cursedcoder:
##### Description of change
Just bumping floss version. Originally I wanted to check PRs code coverage. And using floss 3.0.0 it's much easier to achieve. ~~I've also created another branch with nyc/codecov.io. You might want to check it there: [coverage](https://codecov.io/gh/cursedcoder/pixi.js/branch/coverage) / [code](https://github.com/pixijs/pixi.js/compare/dev...cursedcoder:coverage?expand=1).~~

##### Pre-Merge Checklist

- [x] Since test packages such as chai now propagated on monorepo level, it's not possible to run `npm test` inside single package. I can fix that with some wrapper or anything else. What do you think?


bigtimebuddy:
Thank you. You're my hero for doing this. 

What do you think about deploying coverage with Travis? We did in this in v4 for awhile and it made it easier to check testing coverage over time. You've done the hard work just adding `nyc floss` script and put coverage into `dist/coverage`.

bigtimebuddy:
@cursedcoder If there's no way to run floss in individual packages (because of missing sinon, chai etc) why not just remove the floss dependency and `test` script in all the packages? I'm okay with simplifying and running at the root level only. If a developer is working on an isolated test, they can always use `only` in Mocha. Or we could at the ability to pass scope argument at the root level.

cursedcoder:
@bigtimebuddy removed floss/test from packages. I tried few ways before, but couldn't succeed:
1. Adding test runner package with floss proxy – not possible to inject globals this way
2. Adding filter param to root script – same, floss doesn't provide filter param (we can address this in next version of floss)

So only way to run single test is to add `.only` – IMO this works best in current state.

bigtimebuddy:
I think this is failing because https://github.com/pixijs/pixi.js/blob/dev/test/packages.js#L17 only packages with a test script are run.

cursedcoder:
@bigtimebuddy build is OK now.

LiLrere757:
I'm  cramming  all info.  I possibly can.    I need  more time  & lots help ✔  I'm hooked  line & sinker.  💯

PR #5804, opened 2019-06-21
Use OffscreenCanvas for measureText if available

fserb:
##### Description of change
OffscreenCanvas setFont + measureText is 15-40% faster than the Canvas element version, when available. This detects if OffscreenCanvas 2D exists, and if so, uses it instead of canvas. There should be no visible behavior change.

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/TextMetrics.js
+++ b/packages/text/src/TextMetrics.js
@@ -699,7 +699,22 @@ export default class TextMetrics
  * @private
  */
 
-const canvas = document.createElement('canvas');
+const canvas = (() =>
+{
+    try
+    {
+        // OffscreenCanvas2D measureText can be up to 40% faster.
+        const c = new OffscreenCanvas(0, 0);
+
+        c.getContext('2d');
+
+        return c;
+    }
+    catch (ex)
+    {
+        return document.createElement('canvas');
+    }
+})();
 
 canvas.width = canvas.height = 10;
 


bigtimebuddy:
Thanks for this. Could you add an inline comment about the performance consideration? Also why a try/catch and not just check for the OffscreenCanvas API?

packages/text/src/TextMetrics.js
```diff
@@ -699,7 +699,21 @@ export default class TextMetrics
  * @private
  */
 
-const canvas = document.createElement('canvas');
+const canvas = (() =>
```
bigtimebuddy: Why not something more simple? Is there something more subtle here about OffscreenCanvas that I'm not understanding that requires a try/catch?
```suggestion
const canvas = window.OffscreenCanvas ?
    new OffscreenCanvas(0, 0):
    document.createElement('canvas');
```

fserb: Yes. OffscreenCanvas can be implemented with only WebGL support (currently on Firefox). So we need to `getContext("2d")` and try/catch to see if 2D OffscreenCanvas is supported.

fserb:
Added comment.

ivanpopelyshev:
I have a big app. If I turn on OffscreenCanvas in TextMetrics and Text, it gives weird result: my custom fonts are ignored. I'm trying to reduce the example. I hope to file a bug in chromium.

ivanpopelyshev:
OK, here it is , font problem. Try `dom` versus `offscreen`, font is not loaded even if I re-create OffscreenCanvas every frame.
[measure-offscreen.zip](https://github.com/pixijs/pixi.js/files/3318363/measure-offscreen.zip)

ivanpopelyshev:
OK guys, its not font loading, its the size: `1.6em` + offscreenCanvas = problem.

ivanpopelyshev:
People who use `em` units in font style will have problems. https://bugs.chromium.org/p/chromium/issues/detail?id=977876&can=2&q=OffscreenCanvas 

Does it matter for us?

fserb:
`em` units are not properly defined for detached canvas, like the one used here, as there is no `computed style` for detached canvas.

themoonrat:
So maybe there always had to be a standard canvas for computed font sizes, but for absolute font sizes it could use an offscreen canvas of available?

fserb:
I've just sent a change on Chrome to make sure that the relative font sizes behave the same on both canvas and OffscreenCanvas (thanks @ivanpopelyshev for the bug).

ivanpopelyshev:
@fserb I have changed all canvases to OffscreenCanvas in my work project, Its a very big thing that uses 2d/webgl to show SWF files. Very sophisticated SWF files. 7 years work of 10 artists , 500 megapixels of content. Aaaand I found the difference in vector graphics! I'm gonna report it to you tomorrow, I hope to make a minimal case for it.

ivanpopelyshev:
I'm very excited to have one more googler who works on chromium in my contact list ^_^ Do you know Cary Clark? https://bugs.chromium.org/p/skia/issues/detail?id=7792

ivanpopelyshev:
@fserb I've sent you an email regarding another OffscreenCanvas bug :)

bigtimebuddy:
I think I’m still okay with this change as is since Text favors the use of fontSize in px not em. So it only really affects current Chrome users who define Text in em units. And hopefully will get resolved soon in Chromium.

fserb:
Someone should press the button :)

PR #5813, opened 2019-06-22
OffscreenCanvas types fix

ivanpopelyshev:
Older typescript version doesn't have offscreenCanvas in it. Lets do the same as in `BufferResource` and `SharedArrayBuffer`, declare an empty interface!

I've tested it on old and new ts.


ivanpopelyshev:
What's in the output:

```ts
/**
 * @interface SharedArrayBuffer
 */
declare interface SharedArrayBuffer {
}

/**
 * @interface OffscreenCanvas
 */
declare interface OffscreenCanvas {
}
```

Issue #5816, opened 2019-06-24
Textures from base64-encoded data URIs reports wrong validity


mofnire:
### Expected Behavior

`PIXI.Texture.from('data:image/svg+xml;base64,...').valid` is expected to be `false` If it is not cahched.

### Current Behavior

It returns true regardless of the loading state.

### Possible Solution

As far as I known, this issue is caused by the SVGResource L20. Canvas elements created at that line has the non-zero width / height, thus the method `#valid` reports true even if a SVG is not loaded yet.

https://github.com/pixijs/pixi.js/blob/c4968736f2c6e928097ef7e903e14e86e6f1de01/packages/core/src/textures/resources/SVGResource.js#L20

https://github.com/pixijs/pixi.js/blob/c4968736f2c6e928097ef7e903e14e86e6f1de01/packages/core/src/textures/resources/Resource.js#L115-L118

Thus, resetting the canvas width/height explicitly solves this issue.
At least, it worked for me.
```javascript
export default class SVGResource extends BaseImageResource
{
    constructor(source, options)
    {
        options = options || {};
+       const canvas = document.createElement('canvas');
+       canvas.width = canvas.height = 0;
+       super(canvas);
-       super(document.createElement('canvas'));
```

### Steps to Reproduce

Please see the following playground:
https://www.pixiplayground.com/#/edit/_2bneblE7~49HhtvsviJZ

### Environment

- **`pixi.js` version**:  _5.0.4_
- **Browser & Version**: _Chrome 75.0.3770.100_
- **OS & Version**: _Windows 10_
- **Running Example**: _https://www.pixiplayground.com/#/edit/_2bneblE7~49HhtvsviJZ_

ivanpopelyshev:
Yes, it shouldn't be valid before we actually know its size. `super(canvas)` isn't good idea, I'll think how to modify your PR, it should look better.

ivanpopelyshev:
Btw, you are on fire, 4 PR's in short time 👍

mofnire:
> Yes, it shouldn't be valid before we actually know its size. `super(canvas)` isn't good idea, I'll think how to modify your PR, it should look better.

OK, I got it.

mofnire:
> Btw, you are on fire, 4 PR's in short time 👍

Thank you for making PixiJS 👍 
Saved me a lot.

mofnire:
The PR is merged 🎉

PR #5819, opened 2019-06-24
Fix for the SVGResource#valid issue (#5816)

mofnire:
##### Description of change

This PR is for fixing the issue #5816. As described at the issue, a canvas created by `document.createElement('canvas')` may have non-zero width and height. Thus, the method `SVGResource#valid` may return true even when loading is in process. This PR changes the `SVGResource` to set the width / height of a created canvas to zero.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/resources/SVGResource.js
+++ b/packages/core/src/textures/resources/SVGResource.js
@@ -17,12 +17,9 @@ export default class SVGResource extends BaseImageResource
     {
         options = options || {};
 
-        const canvas = document.createElement('canvas');
-
-        canvas.width = 0;
-        canvas.height = 0;
-
-        super(canvas);
+        super(document.createElement('canvas'));
+        this._width = 0;
+        this._height = 0;
 
         /**
          * Base64 encoded SVG element or URL for SVG file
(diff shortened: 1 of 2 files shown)


mofnire:
> Good idea, but I need to think about better implementation. I'm not comfortable with settings canvas size to 0.

If DOM operations are not preferable, this might be better?
```javascript
options = options || {};

super(document.createElement('canvas'));
this._width = 0;
this._height = 0;
```

Or should I introduce a new property `_hasLoaded`...?

bigtimebuddy:
I like that suggestion better: setting internal width and height to zero, instead of setting canvas.

mofnire:
> I like that suggestion better: setting internal width and height to zero, instead of setting canvas.

I changed the way of solving the issue.

bigtimebuddy:
@ivanpopleyshev, you good with this?

ivanpopelyshev:
Yes, I think so. But I'll check this PR anyway, cant just approve it, its my area of expertise.

bigtimebuddy:
MVP @mofnire! It’s easy to approve PRs with very clear tests, like this. Keep up the good work.

mofnire:
> MVP @mofnire! It’s easy to approve PRs with very clear tests, like this. Keep up the good work.

Thank you ! I will !

PR #5820, opened 2019-06-24
fix: TextStyle.whiteSpace param type

SenHeng:
##### Description of change

Type for `whiteSpace` is listed as `boolean` but should be a `string`.

##### Pre-Merge Checklist

There were 3 failing tests on the master branch but those were already failing before my change.

```zsh
3 failing

  1) PIXI
       "before all" hook:
     Uncaught ReferenceError: _url is not defined
      at file:///Users/emp-mac-heng/projects/others/pixi/bundles/pixi.js/dist/pixi.js:42340:7

  2) PIXI
       should exist as a global object:
     TypeError: Cannot read property 'call' of undefined
      at runCallback (timers.js:694:18)
      at tryOnImmediate (timers.js:665:5)

  3) PIXI
       should contain deprecations on window, not import:
     TypeError: Cannot read property 'call' of undefined
      at runCallback (timers.js:694:18)
      at tryOnImmediate (timers.js:665:5)
```


_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-cayman
```
bigtimebuddy: Please remove _config.yml

SenHeng: Removed.

bigtimebuddy:
Thank you!

Issue #5817, opened 2019-06-24
Empty graphics (graphicsData.length === 0) are not batched.


mofnire:
### Expected Behavior

PIXI.Graphics instances who have no GraphicsData (i.e., `graphics.geometry.graphicsData.length === 0`) expected be batchable (`graphics.geometory.batchable === true`).

### Current Behavior

They are not batchable (i.e., `graphics.geometory.batchable` may not be `true`) in some situations.

### Possible Solution

GraphicsGeomery#updateBatches exits without updating the property `batchable` if the `graphicsData.length` is zero.

https://github.com/pixijs/pixi.js/blob/c4968736f2c6e928097ef7e903e14e86e6f1de01/packages/graphics/src/GraphicsGeometry.js#L403-L406

Updating the property `batchable` when the `graphicsData.length` is zero may solve the issue.
```javascript
    updateBatches()
    {
        if (this.dirty === this.cacheDirty) return;
+       if (this.graphicsData.length === 0) {
+           this.batchable = true.
+           return;
+       }
-       if (this.graphicsData.length === 0) return;
```

### Steps to Reproduce

Please see the following playground:
https://www.pixiplayground.com/#/edit/C~ch4VH_tQDRG0bUrDVef

### Environment

- **`pixi.js` version**: _5.0.4_
- **Browser & Version**: _Chrome 75.0.3770.100_
- **OS & Version**: _Windows 10_
- **Running Example**: _https://www.pixiplayground.com/#/edit/C~ch4VH_tQDRG0bUrDVef_

PR #5821, opened 2019-06-24
Fix for the empty graphics batchability issue (#5817)

mofnire:
##### Description of change

This PR is for fixing the #5817.  If a `PIXI.Graphics` instance has no graphics data (i.e., `graphics.geometry.graphicsData.length` is zero), `BatchRenderer` may treat it as if it is not batchable. This is because `PIXI.GraphicsGeometry#updateBatches()` leaves the property `#batchable` as it is when `#graphicsData.length` is zero. This PR fixes the method `#updateBatches()` to update the property `#batchable` when `#graphicsData.length` is zero.

##### Pre-Merge Checklist


bigtimebuddy:
Thank you @mofnire!

PR #5822, opened 2019-06-24
Make DisplayObject#mask nullable

mofnire:
##### Description of change

In Pixi.js v5.0.4, the `DisplayObject#mask` is declared as `PIXI.Graphics | PIXI.Sprite`. However, to remove a mask from a display object, need to set a mask to null (as far as I know). This PR makes the `DisplayObject#mask` nullable.

https://github.com/pixijs/pixi.js/blob/c4968736f2c6e928097ef7e903e14e86e6f1de01/packages/display/src/DisplayObject.js#L682-L688

I tried the following variations of the type declaration. 
None of them works for me. The `tsd-jsdoc` may not support the question mark?

 - ?(PIXI.Graphics | PIXI.Sprite),
 - ?PIXI.Graphics | ?PIXI.Sprite, and
 - (?PIXI.Graphics) | (?PIXI.Sprite).

##### Pre-Merge Checklist

### DIFF
--- a/packages/display/src/DisplayObject.js
+++ b/packages/display/src/DisplayObject.js
@@ -152,7 +152,7 @@ export default class DisplayObject extends EventEmitter
         /**
          * The original, cached mask of the object.
          *
-         * @member {PIXI.Graphics|PIXI.Sprite}
+         * @member {PIXI.Graphics|PIXI.Sprite|null}
          * @protected
          */
         this._mask = null;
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Is that the only issue that prevents you from using PixiJS with strict mode? I'm sure there are other places like that.

mofnire:
> Is that the only issue that prevents you from using PixiJS with strict mode? I'm sure there are other places like that.

If the `strict` mode refers to the TypeScript's strict mode, yes it is. This might not be a PixiJS issue? Our issue?

mofnire:
> I'm afraid that we'll be flooded with PR's like that. I can approve one, but next time we'll need mass-fix. Its either TS with strict, or its for Kotlin conversion of typings.

I understood your concern.

PR #5826, opened 2019-06-24
Prevented System from attaching itself to the Renderer's contextChang…

### LINKED ISSUE #5825: Rendering systems are getting binded twice to contextChange signal
I'm not sure if my doubt is right, but I wanted to let you guys know.

The `System` constructor in `@pixi/core` attaches its instance to the `renderer.runners.contextChange` signal. However, the `Renderer#addSystem` attaches the system to all runners; hence, I think each system being added is attached twice to the `contextChange` signal.

```
// Renderer.js #addSystem()
for (const i in this.runners)
 {
  this.runners[i].add(system);
 }
```

```
// System.js #constructor
this.renderer.runners.contextChange.add(this);
```

Was this intended or is this a bug?

SukantPal:
This fixes #5825 - however, there was a side effect which I've fixed too.

PixiJS's `BatchPlugin`, which is also called `BatchRenderer` in its default export, was being treated as a `System`. This was because its inheritance chain is like this:
1. `BatchPlugin`
2. `BatchRenderer` (used as `BaseBatchRender` which references the `BatchRender.js` file)
3. `ObjectRenderer`
4. `System`

Since the `System` constructor added itself to the `contextChange` signal, there was no problem before. I removed that line; hence, `BatchPlugin` wasn't able to add itself to the `contextChange` signal. This broke the tests in `Graphics#_render: property program cannot be read from null (line 909)`. I've added the line that makes `BatchPlugin` attach to the `contextChange` signal.

So far, the tests are passing. However, if someone treated a plugin as a system somewhere else too and assumed that the plugin would attach to `contextChange` automatically, then that will become a bug.

### DIFF
--- a/packages/core/src/Renderer.js
+++ b/packages/core/src/Renderer.js
@@ -85,7 +85,7 @@ export default class Renderer extends AbstractRenderer
          * The type of this renderer as a standardized const
          *
          * @member {number}
-         * @see PIXI.RENDERER_TYPE
+         * PIXI.RENDERER_TYPE
          */
         this.type = RENDERER_TYPE.WEBGL;
 
@@ -98,7 +98,7 @@ export default class Renderer extends AbstractRenderer
         /**
          * Internal signal instances of **runner**, these
          * are assigned to each system created.
-         * @see PIXI.Runner
+         * PIXI.Runner
          * @name PIXI.Renderer#runners
          * @private
          * @type {object}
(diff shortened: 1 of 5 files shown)


ivanpopelyshev:
Actually, we have to remove it from `destroy` too.

SukantPal:
> Actually, we have to remove it from destroy too, and check if renderer destroy actually calls all systems destroy.

I don't get what you mean. The render runs the `destroy` signal, which means that all systems run the `destroy()` method (if they implement one).

SukantPal:
I've done a few more commits unrelated to the PR. I forgot to switch the branch (I don't know how to).

Should I make a new PR instead?

bigtimebuddy:
I think what Ivan is saying is that contextChange.remove should be called in BatchRenderer when it’s destroyed. Removing that reference will help GC.

SukantPal:
> I think what Ivan is saying is that contextChange.remove should be called in BatchRenderer when it’s destroyed. Removing that reference will help GC.

I think rather than `BatchRenderer` manually removing itself on `destroy`, we should actually make `Renderer` destroy its signal runners. This will help the GC collect not only the `BatchRenderer`, but all of the systems.

ivanpopelyshev:
@SukantPal

Oops, GD8 somehow made it to this PR. I appreciate that you actually made that , you're the first person in community who actually got in that module :) But it has to be in separate PR.

bigtimebuddy:
I agree do the docs for GD8 separately. Will make it easier to approve.

SukantPal:
@Ivan

You got to help me since I don't know how to separate it now. I can make a new PR but don't know how to move the commits.

bigtimebuddy:
Try using “git cherry-pick [commit hash]” to pull the comment commits into a new branch made from from dev.

ivanpopelyshev:
@SukantPal Yes, I've sent you a slack message a hour before: 

> may `git rebase` , `git reset --hard someBranch`, `git stash` and simple ctrl+c ctrl+V of files to help you. 

those operations and `cherry-pick` should be your base in dealing with that kind of issues. I also recommend to use tortoiseGit (https://tortoisegit.org/) if you do want something better than just console. 

Using git windows is also fine, but I wont be able to handle its quirks. If you use SourceTree, then I advice to switch to something else.

ivanpopelyshev:
I have fixes for your comments for GroupD8, it'll be very good PR.

SukantPal:
> @SukantPal Yes, I've sent you a slack message a hour before:
> 
> > may `git rebase` , `git reset --hard someBranch`, `git stash` and simple ctrl+c ctrl+V of files to help you.
> 
> those operations and `cherry-pick` should be your base in dealing with that kind of issues. I also recommend to use tortoiseGit (https://tortoisegit.org/) if you do want something better than just console.
> 
> Using git windows is also fine, but I wont be able to handle its quirks. If you use SourceTree, then I advice to switch to something else.

I'll try that out, but using Linux right now. I'll commit by tomorrow.

SukantPal:
I am going to re-open this same PR since I messed up my Git and mixed two things!

Issue #5823, opened 2019-06-24
If DisplayObject#hitArea accepts a PIXI.HitArea alternative, might be useful?


mofnire:
So far, I could not find the `PIXI.HitArea` alternative for PixiJS v5. And `DisplayObject#hitArea` accepts the following classes or their derived classes only:

- `PIXI.Rectangle`,
- `PIXI.Circle`,
- `PIXI.Ellipse`,
- `PIXI.Polygon`, and
- `PIXI.RoundedRectangle`.

Thus, in my case, I changed the super class of my custom hit-area classes to `PIXI.Rectangle`.
As far as I known, all the codes which use `DisplayObject#hitArea` must not require anything other than the methods `#contains(x, y)` and `#clone()` because these are the only methods the union type of the above 5 classes has.

Is it an acceptable idea introducing the virtual JSDoc comments(*1) that define an `PIXI.HitArea` interface and adding it to the `@param` of `DisplayObject#hitArea`? Namely, something looks like this:
```
/**
 * Interface for classes that represent a hit area.
 *
 * @interface PIXI.HitArea
 */

/**
 * Checks whether the x and y coordinates given are contained within this Rectangle
 *
 * @method
 * @name PIXI.HitArea#contains
 * @param {number} x - The X coordinate of the point to test
 * @param {number} y - The Y coordinate of the point to test
 * @return {boolean} Whether the x/y coordinates are within this Rectangle
 */


export default {
    /**
     * Interaction shape. Children will be hit first, then this shape will be checked.
     * Setting this will cause this shape to be checked in hit tests rather than the displayObject's bounds.
     *
     * @example
     * const sprite = new PIXI.Sprite(texture);
     * sprite.interactive = true;
     * sprite.hitArea = new PIXI.Rectangle(0, 0, 100, 100);
     * @member {PIXI.HitArea|PIXI.Rectangle|PIXI.Circle|PIXI.Ellipse|PIXI.Polygon|PIXI.RoundedRectangle}
     * @memberof PIXI.DisplayObject#
     */
    hitArea: null
}
```

[*1 Please refer to the virtual comment section of the JSDoc manual](https://jsdoc.app/tags-interface.html)

### Environment

- **`pixi.js` version**: _5.0.4_
- **Browser & Version**: _Chrome 75.0.3770.100_
- **OS & Version**: _Windows 10_
- **Running Example**: _None_

mofnire:
[...]

PR #5827, opened 2019-06-25
Change DisplayObject#hitArea type declaration to support more hit-area classes

mofnire:
##### Description of change

Underlying needs are described at #5823. PixiJS v5 does not have a PIXI.HitArea alternative. And we are not allowed to assign anything to `DisplayObject#hitArea` other than the instances of the following classes or their derived classses:

- `PIXI.Rectangle`,
- `PIXI.Circle`,
- `PIXI.Ellipse`,
- `PIXI.Polygon`, and
- `PIXI.RoundedRectangle`.

As a consequence, in my project, I changed a superclass of custom hit-area classes to `PIXI.Rectangle` from `PIXI.HitArea` of PixiJS v4. So I thought if `DisplayObject#hitArea` accepts `PIXI.HitArea` -like classes, it would be useful. This PR introduces a new interface `PIXI.IHitArea` and changes the `DisplayObject#hitArea` type declaration for this sake.

##### Pre-Merge Checklist


packages/interaction/src/IHitArea.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Interface for classes that represent a hit area.
+ *
+ * @interface IHitArea
+ * @memberof PIXI
+ */
+
+/**
+ * Checks whether the x and y coordinates given are contained within this Rectangle
```
themoonrat: Within this shape, rather than rectangle

packages/interaction/src/IHitArea.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Interface for classes that represent a hit area.
```
themoonrat: New line giving a couple of examples of what kind of classes. Eg PIXI.Rectangle

packages/interaction/src/IHitArea.js
mofnire: Thank you for the review. As pointed out, the `rectangle` is not appropriate. Because this is a `IHitArea` interface, I think the word `area` fits well. Do you have any thoughts?

themoonrat: yeps, area sounds good

mofnire: I updated the JSDoc comment as per the review.

packages/interaction/src/IHitArea.js
mofnire: I update the JSDoc comment as per the review comment.

ivanpopelyshev:
TS infer interfaces.

PR #5833, opened 2019-06-25
Max FPS can be disabled by setting to 0.

themoonrat:
This lets the game run as fast as it can.
Also tweaked maxFPS, so that when it is set, there is a padding to it that is relative to the target fps set. So a target of 20fps allows for more padding than a target of 60fps. Added for accuracy reasons on higher refresh rate monitors.
Should fix https://github.com/pixijs/pixi.js/issues/5741

### DIFF
--- a/packages/ticker/src/Ticker.js
+++ b/packages/ticker/src/Ticker.js
@@ -504,11 +504,11 @@ export default class Ticker
     }
 
     /**
-     * Manages the minimum amount of milliseconds allowed to
+     * Manages the minimum amount of milliseconds required to
      * elapse between invoking {@link PIXI.Ticker#update}.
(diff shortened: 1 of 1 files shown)


packages/ticker/src/Ticker.js
```diff
@@ -502,25 +507,25 @@ export default class Ticker
      * Manages the minimum amount of milliseconds allowed to
      * elapse between invoking {@link PIXI.Ticker#update}.
      * This will effect the measured value of {@link PIXI.Ticker#FPS}.
-     * When setting this property it is clamped to a value between
-     * `1` and `TARGET_FPMS * 1000`.
+     * When setting this property it is at least it could be `0` which means 'there is no max limit',
```
andrewstart: The grammar here could use a rethink.

PR #5836, opened 2019-06-25
Supports resolutions less than 1

themoonrat:
Resolutions of 0.5 / 0.75 can lead to floating point widths and heights. This PR ensures that when textures are created, `ceil` is used to ensure that everything fits in. When rendering, it uses `floor` to ensure it's not trying to render an area that doesn't exist (which IE really hates)


bigtimebuddy:
Is it possible to add tests for this? Want to make sure we harden this behavior. Otherwise awesome fix.

GoodBoyDigital:
nice! awaiting secret agent x-files tests :D

ivanpopelyshev:
For floor add `+1e-4` for ceil add `-1e-4` and it should be fine.

ivanpopelyshev:
Maybe put those floor ceil in our own functions in math. @bigtimebuddy I need your decision whether to put `1e-4` everywhere or move it to function.

bigtimebuddy:
Just inline those and we can do cleanup later since there are more places that could utilize the safe floor and ceil methods. Rather not pile on.

GoodBoyDigital:
this looks cool. Will solve the case for most users and is better than we have now. @ivanpopelyshev sounds like you have some good thoughts on how to handle the floating point precision, you cool to do a separate PR? 

Ta!

bigtimebuddy:
Thanks @themoonrat

Megabyteceer:
When scale of sprite set to 100, Math.floor causes random resize of sprite for 0 to 100 pexels related on current render resolution.

PR #5841, opened 2019-06-26
Fix the DisplayObject#cacheAsBitmap issue #5784

### LINKED ISSUE #5784: Cannot Create a texture from a Graphic if CacheAsBitmap is on
I understand this is an edge case, and may very well be 'as designed', however I wanted to raise the issue anyhow so that it is known.

I made a CodePen to prove the issue here:

https://codepen.io/anon/pen/rExjVy?editors=0011

if you remove the line `graphics.cacheAsBitmap = true;`, then things work as expected and the texture is rendered, but with that set to true, there is no texture. Or, perhaps, it's an empty texture?

### Expected Behavior

Texture to render, or at least an error to be shown.

### Current Behavior

Texture is empty. No errors are shown.

### Possible Solution

Render the texture, or throw an error

### Environment

- **`pixi.js` version**: 5.0.4
- **Browser & Version**: Chrome 75
- **Running Example**: https://codepen.io/anon/pen/rExjVy?editors=0011

mofnire:
##### Description of change

This PR is to fix #5784. There are two problems in `DisplayObject#_initCachedDisplayObject`.

1) The first one is `renderer._activeRenderTarget` does not hold a current rendering target.

https://github.com/pixijs/pixi.js/blob/4d71fa06ea787d81651fc3e624610ccfc3c3f457/packages/mixin-cache-as-bitmap/src/index.js#L185-L189

I replaced it with `const cachedRenderTarget = renderer.renderTexture.current;`.

2) The second one is `renderer.projection.transform` is not restored after the line 213.
I changed the method so that the projection transform is restored.

https://github.com/pixijs/pixi.js/blob/4d71fa06ea787d81651fc3e624610ccfc3c3f457/packages/mixin-cache-as-bitmap/src/index.js#L210-L216

Working example is [https://www.pixiplayground.com/#/edit/hk1gjmUM84Nl72GncOii~](https://www.pixiplayground.com/#/edit/hk1gjmUM84Nl72GncOii~).

*Please note I could not find the way of writing tests of features like this :( *

##### Pre-Merge Checklist

### DIFF
--- a/packages/mixin-cache-as-bitmap/src/index.js
+++ b/packages/mixin-cache-as-bitmap/src/index.js
@@ -184,7 +184,9 @@ DisplayObject.prototype._initCachedDisplayObject = function _initCachedDisplayOb
 
     // for now we cache the current renderTarget that the WebGL renderer is currently using.
     // this could be more elegant..
-    const cachedRenderTarget = renderer._activeRenderTarget;
+    const cachedRenderTarget = renderer.renderTexture.current;
+    const cachedProjectionTransform = renderer.projection.transform;
+
     // We also store the filter stack - I will definitely look to change how this works a little later down the line.
     // const stack = renderer.filterManager.filterStack;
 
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
OK, i know how to fix that, good catch! The idea is to store `renderer.renderTexture.sourceFrame` , not `projection.transform`. I'll do separate PR and use your example to test it.

stevenalanstark:
Thank you very much for this, well done! :D

PR #5843, opened 2019-06-26
Add JSDoc comments on PIXI.Transform

mofnire:
##### Description of change

This PR adds JSDoc comments on `PIXI.Transform` properties and improves the PixiJS type declaration file. I intentionally changed the visibilities of the methods `onChange` and `updateSkew` from `@private` to `@protected`. I thought it is a good thing to allow developers to override them when needed. But if it isn't, I will revert them.

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas/canvas-renderer/src/CanvasRenderer.js
+++ b/packages/canvas/canvas-renderer/src/CanvasRenderer.js
@@ -1,4 +1,4 @@
-import { AbstractRenderer } from '@pixi/core';
+import { AbstractRenderer, resources } from '@pixi/core';
 import { CanvasRenderTarget, sayHello } from '@pixi/utils';
 import CanvasMaskManager from './utils/CanvasMaskManager';
 import mapCanvasBlendModesToPixi from './utils/mapCanvasBlendModesToPixi';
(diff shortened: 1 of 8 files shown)


mofnire:
> Also I'd like to point out that (cx, sx) is normalized first column of matrix, coordinates of normalized local X axis in parent space. Same about (cy, sy)

Thank you for clarifying. I thought long about their good explanations / technical terms. "X-coordinate value of normalized local X axis" sounds good.

ivanpopelyshev:
We can put the same desc in both _cx and _cy:

> `(_cx,_sx)` is normalized local X axis, first column of local matrix without scale.

> `(_cy,_sy)` is normalized local Y axis, second column of local matrix without scale.

I leave it to you.

bigtimebuddy:
@ivanpopelyshev, why should all the members be protected?

ivanpopelyshev:
That way i can write functions like this: https://github.com/pixijs/pixi-projection/blob/master/src/base/LinearProjection.ts#L2

ivanpopelyshev:
Also, those are not global unique ID's , they are  unique only inside the instance. We can say that they are incremented. and `_worldID`  is read by others , while `_parentID` points to `_worldID` of parent that we already used for calculation. I don't know how to spell that.

mofnire:
I revised descriptions as per review comments.

PR #5848, opened 2019-06-27
Prevented System from attaching itself to the Renderer's contextChange signal

### LINKED ISSUE #5825: Rendering systems are getting binded twice to contextChange signal

SukantPal:
I've fixed #5825 and a the side effect on `BatchPlugin`. This was discussed in a failed PR https://github.com/pixijs/pixi.js/pull/5826, which was corrupted due to another change being recorded in the same branch.

### DIFF
--- a/packages/core/src/System.js
+++ b/packages/core/src/System.js
@@ -18,18 +18,6 @@ export default class System
          * @member {PIXI.Renderer}
          */
         this.renderer = renderer;
-
-        this.renderer.runners.contextChange.add(this);
-    }
-
-    /**
-     * Generic method called when there is a WebGL context change.
-     *
-     * @param {WebGLRenderingContext} gl new webgl context
-     */
-    contextChange(gl) // eslint-disable-line no-unused-vars
-    {
-        // do some codes init!
     }
 
     /**
@@ -37,7 +25,6 @@ export default class System
      */
     destroy()
     {
-        this.renderer.runners.contextChange.remove(this);
         this.renderer = null;
     }
 }
--- a/packages/core/src/Renderer.js
+++ b/packages/core/src/Renderer.js
@@ -420,6 +420,11 @@ export default class Renderer extends AbstractRenderer
     {
         this.runners.destroy.run();
 
+        for (const r in this.runners)
+        {
+            this.runners[r].destroy();
+        }
+
         // call base destroy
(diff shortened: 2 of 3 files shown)


ivanpopelyshev:
Huh. Droping the repo is certainly a radical solution! Anyway, we need @GoodBoyDigital here, its his class. 

I'll be happy to see `GroupD8` stuff, but take your time, there's no rush :)

bigtimebuddy:
What about the for loop that destroyed the runners in Renderer?

SukantPal:
@bigtimebuddy Umm, doing that failed the build.

bigtimebuddy:
I think the idea was correct but your loop was wrong. 

for (const r in this.runners) { r.destroy(); }

Should be:

for (const r in this.runners) { this.runners[r].destroy(); }

SukantPal:
Oh okay, you're totally right. I'll try again. I can't believe how stupid I am - I was thinking about the for-each loop and not the for-in loop, since I come from Java background.

bigtimebuddy:
No worries! There’s also a for..of loop, but we don’t use that because it’s not supported in older browsers.

SukantPal:
> I think the idea was correct but your loop was wrong.
> 
> for (const r in this.runners) { r.destroy(); }
> 
> Should be:
> 
> for (const r in this.runners) { this.runners[r].destroy(); }

I've added that and now the build is passing too!

packages/core/src/System.js
```diff
@@ -18,26 +18,13 @@ export default class System
          * @member {PIXI.Renderer}
          */
         this.renderer = renderer;
```
GoodBoyDigital: hello! Not sure i fully understand why we want to do this? There are many systems that hook into the runner?

PR #5849, opened 2019-06-27
GroupD8 documentation

SukantPal:
I've built proper documentation of `GroupD8`. Please check it out!

(Thanks @ivanpopelyshev for telling me about resetting my `gd8-dev` branch to `origin/dev` and then pushing.)

### DIFF
--- a/packages/math/src/GroupD8.js
+++ b/packages/math/src/GroupD8.js
@@ -33,17 +33,14 @@ const vy = [1, 1, 0, -1, -1, -1, 0, 1, -1, -1, 0, 1, 1, 1, 0, -1];
  * [Cayley Table]{@link https://en.wikipedia.org/wiki/Cayley_table}
  * for the composition of each rotation in the dihederal group D8.
  *
- * It is a 2D array as follows:
- * |    | E  | SE | S  | SW | W  | NW | N  | NE |
- * |----|----|----|----|----|----|----|----|----|
- * | E  | E  | SE | S  | SW | W  | NW | N  | NE |
- * | SE | SE | S  | SW | W  | NW | N  | NE | E  |
- * | S  | S  | SW | W  | NW | N  | NE | E  | SE |
- * | SW | SW | W  | NW | N  | NE | E  | SE | S  |
- * | W  | W  | NW | N  | NE | E  | SE | S  | SW |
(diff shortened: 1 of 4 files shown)


packages/math/src/GroupD8.js
```diff
@@ -1,46 +1,87 @@
...
+ * |    | E  | SE | S  | SW | W  | NW | N  | NE |
+ * |----|----|----|----|----|----|----|----|----|
+ * | E  | E  | SE | S  | SW | W  | NW | N  | NE |
+ * | SE | SE | S  | SW | W  | NW | N  | NE | E  |
+ * | S  | S  | SW | W  | NW | N  | NE | E  | SE |
+ * | SW | SW | W  | NW | N  | NE | E  | SE | S  |
+ * | W  | W  | NW | N  | NE | E  | SE | S  | SW |
+ * | NW | NW | N  | NE | E  | SE | S  | SW | W  |
+ * | N  | N  | NE | E  | SE | S  | SW | W  | NW |
+ * | NE | NE | E  | SE | S  | SW | W  | NW | N  |
```
ivanpopelyshev: I don't like big tables that are also trivial. We can fix that!

Either make it brife (just 4 elements) 
Either specify 4 main and 4 reflections - that table is not trivial. 

Also specify integer codes in a first row.

I have an idea how to make it even better but for now that'll be enough.

SukantPal: I could make it only show four rotations; however, `rotationCayley` (formerly `mul`) doesn't store compositions involving reflections.

ivanpopelyshev: its 16x16. It stores everything :)

ivanpopelyshev: I wouldn't make this class if we it didnt support reflections. All rotations are easy, just `(x+y)%8` or `(x+8-y)%8`.

SukantPal: I've done what you've said.

bigtimebuddy:
@ivanpopelyshev are you good on this?

ivanpopelyshev:
For future: that's why we have this class: https://github.com/EsotericSoftware/spine-runtimes/blob/3.8-beta/spine-ts/core/src/attachments/MeshAttachment.ts#L48  . This method does not exist in `pixi-spine` ;)

PR #5853, opened 2019-06-30
 'resource-loader' import and version fixes

lekzd:
##### Description of change
Usage of 'resource-loader' deafult export in `packages/loaders/src/Loader.js` replaced with named one
It happens because of actual version of 'resource-loader' is 2.3.0 where default export was removed

Also I replaced version of 'resource-loader' in package.json with strict "2.2.4" because of backwards comptibility needed instead of risky refactoring of blobMiddlewareFactory calls removed in newer version

##### Pre-Merge Checklist

### DIFF
--- a/packages/loaders/package.json
+++ b/packages/loaders/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@pixi/core": "^5.0.3",
     "@pixi/utils": "^5.0.3",
-    "resource-loader": "^2.2.4"
+    "resource-loader": "2.2.4"
   },
   "devDependencies": {
     "floss": "^2.2.0"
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Thanks for this PR. I don't think this is the correct solution tho. I asked @englercj to release a new major version of **resource-loader** and we should create a PR to upgrade to v3.0.0.

bigtimebuddy:
Closing this PR since the npm registry has now been corrected.

lekzd:
@bigtimebuddy rightnow application with blank Pixi installation falls because of two reasons above, looks like a critical

bigtimebuddy:
Could you file and issue so we can track this. Also, did you clear you npm cache?

PR #5857, opened 2019-07-01
Added generic batching + some internal refactoring

SukantPal:
I've done a lot in this PR. I'll list them below:

1. I polished `BatchRenderer` a little more and renamed it `AbstractBatchRenderer`. In addition, I added the feature of attribute-definitions, i.e. supporting any set of attributes in the shader. Some built-in attributes (which are calculated separately rather than retrieved from the object like the tint/`aColor`) are also handled. To support this, there is a `vertexSizeOf` static method in that class.

2. I fixed `BatchBuffer`'s vertices field. It was an array-buffer that held every attribute (interleaved). Hence, I renamed it to just `rawBinaryData`, since it was that.

3. I renamed `BatchPluginFactory` to `BatchRendererFactory`, since that complements `AbstractBatchRenderer`. In addition, it doens't "plug-in" the renderer. That is to do be done elsewhere, or by using the `BatchSystem`.

4. `ObjectRenderer` isn't a `System` anymore. It was never added as a system, nor was it intended to be one.

##### Pre-Merge Checklist


packages/constants/src/index.js
```diff
@@ -212,6 +212,43 @@ export const TYPES = {
...
     HALF_FLOAT: 36193,
+
+    /**
+     * Returns if any type in the `TYPES` enum is of integer
+     * nature.
+     *
+     * @param {number} glint - `TYPES` types
+     * @return if `glint` is an integral type
+     */
+    isInteger(glint)
```
bigtimebuddy: Theses are probably better in the utils package

ivanpopelyshev:
Here's why we have BatchRenderer plugins in v5. 

I have two batch renderer plugins for v4: 

pixi-projection is fixed 

https://github.com/pixijs/pixi-projection/blob/master/src/base/webgl/Sprite2dRenderer.ts 

pixi-heaven double-tint still uses my v4 hacks:

https://github.com/gameofbombs/pixi-heaven/blob/master/src/twotint/sprites/SpriteHeavenRenderer.ts#L54

The problem is that latter is using two colors, so color cant special case. This double tint appears in Adobe Flash ColorTransform, and in EsotericSoftware Spine.

There are two more in those repositories. Both need special changes in batching itself.

Bilinear renderer checks that projection is the same for sprites: 

https://github.com/pixijs/pixi-projection/blob/v4.x/src/base/webgl/SpriteRenderer.ts#L243

https://github.com/gameofbombs/pixi-heaven/blob/master/src/twotint/sprites/SpriteMaskedRenderer.ts

Masked renderer checks that sprites have same mask: 

https://github.com/gameofbombs/pixi-heaven/blob/master/src/twotint/sprites/SpriteMaskedRenderer.ts

Since we are working on this module, I'll port those 3 renderers ASAP: twotint, bilinear, masked. That way we'll be able to make better API.

@GoodBoyDigital , we need your review.

packages/constants/src/index.js
SukantPal: @bigtimebuddy Hey, I've solved that problem with #9282 (commit in `batch-dev`). I don't need `TYPES.isInteger`; hence, I've eliminated that.

In addition, I have moved `TYPES.sizeOf` to `@pixi/utils/src/data/sizeOfType.js`. It has been renamed to `sizeOfType`.

SukantPal:
Maintainers do not want the attribute definitions feature in the core library. I've filtered what was liked in this PR here: https://github.com/pixijs/pixi.js/pull/5877 (#5877).

SukantPal:
Hence, I am closing this PR. Please remove it from the 5.1.0 milestone so that it can be published :)

PR #5859, opened 2019-07-01
Replaces Bublé with Babel

bigtimebuddy:
##### Description of change

This change replaces [Bublé](https://github.com/bublejs/buble) with [Babel](https://github.com/babel/babel) for transpiling ES6 code. The main reason for this change is that, while both tools do the same thing, Babel has a much larger community and support. Also, Babel uses browserlist, which will give us much better future control about transpiling for specific browsers (possibly a follow-up PR).

This change also exposed a bug in tests where `new RenderTexture.create()` was being incorrectly called. Bonus!

##### Pre-Merge Checklist


bigtimebuddy:
Hmm... this isn't exactly working. http://pixijs.download/dev-replace-buble-with-babel/packages/core.js (`class` shouldn't be in there).

Need to dig deep to figure out what's happening.

PR #5870, opened 2019-07-02
Update to the latest version of resource-loader

happymark01:
##### Pre-Merge Checklist

### DIFF
--- a/packages/loaders/package.json
+++ b/packages/loaders/package.json
@@ -26,6 +26,6 @@
   "dependencies": {
     "@pixi/core": "^5.0.4",
     "@pixi/utils": "^5.0.4",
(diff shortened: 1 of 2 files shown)


packages/loaders/package.json
```diff
@@ -26,6 +26,6 @@
   "dependencies": {
     "@pixi/core": "^5.0.4",
     "@pixi/utils": "^5.0.4",
-    "resource-loader": "^2.2.4"
+    "resource-loader": "^3.0.0"
```
englercj: Might want this to be `v3.0.1` which includes the transpile fixes @bigtimebuddy added recently. `^3.0.0` should actually resolve to `3.0.1` but might be safer to specify that as the minimum

PR #5872, opened 2019-07-03
Fix WebGL context cannot be restored

### LINKED ISSUE #5700: WebGL context restored incorrectly (pixi v5)
Everything works fine for pixi v4, but in v5 (WebGL 2) application restored incorrectly

### Expected Behavior
For both versions of pixi, application should be restored correctly.

### Current Behavior
If you choose **v5** pixi version with webgl2, application is not restored.
If you choose **v4** pixi version with webgl, application restored fine.

### Steps to Reproduce
1. Open https://pixijs.io/examples/#/demos-basic/container.js on PC Chrome
2. Choose **v5** pixi version
3. Open chrome dev tools console with iframe context
4. Write next command to lose webgl context `document.querySelector('canvas').getContext('webgl2').getExtension('WEBGL_lose_context').loseContext()`
5. Wait for a few sec
6. Write next command to restore webgl context `document.querySelector('canvas').getContext('webgl2').getExtension('WEBGL_lose_context').restoreContext()`
### Environment

- **`pixi.js` version**: _5.0.3_
- **Browser & Version**: _Chrome 74.0.3729.157_
- **OS & Version**: _MacOS Mojave 10.14.4_

JetLua:
Fix #5700 

Wait a few seconds after opening the link.

demo: https://www.pixiplayground.com/#/edit/l8h6NrzSYO6y5g8s84PSk
fixed: https://www.pixiplayground.com/#/edit/5Uk~cUocqKhiktZ451glL

### DIFF
--- a/packages/core/src/context/ContextSystem.js
+++ b/packages/core/src/context/ContextSystem.js
@@ -65,6 +65,8 @@ export default class ContextSystem extends System
     contextChange(gl)
     {
         this.gl = gl;
+        this.renderer.gl = gl;
+        this.renderer.CONTEXT_UID = CONTEXT_UID++;
 
         // restore a context if it was previously lost
         if (gl.isContextLost() && gl.getExtension('WEBGL_lose_context'))
--- a/packages/core/src/shader/ShaderSystem.js
+++ b/packages/core/src/shader/ShaderSystem.js
@@ -64,6 +64,7 @@ export default class ShaderSystem extends System
     contextChange(gl)
     {
         this.gl = gl;
+        this.reset();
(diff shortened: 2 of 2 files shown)


JetLua:
> OK, so, some time ago I fixed GeometrySystem and TextureSystem. Now the only thing is left is ShaderSystem.
> 
> By analogy, we need better `reset` that actually goes through all `managedShaders` and frees them, but shaders are never freed, so for now just a small fix is enough.

Yes, just reset not destroyed. Wait for you to have a perfect fix :heart: .

PR #5875, opened 2019-07-04
Extract postDivide

ivanpopelyshev:
I hate to explain how to use extract to people, so lets actually rorce `postDivide` inside extract, even if it takes extra performance. `readPixels` eats more time anyway. 

We cant use it on v4 because some people already use postDivide in it, but we can put a notice in v5 migration.

Bug: https://www.pixiplayground.com/#/edit/6FShpzcn_1BjW7JvU6HWh
Fixed: https://www.pixiplayground.com/#/edit/Q_WpqSdaGP7GDh59Bdd08


packages/extract/src/Extract.js
```diff
@@ -239,4 +241,31 @@ export default class Extract
         this.renderer.extract = null;
         this.renderer = null;
     }
+
+    /**
+     * Takes premultiplied pixel data and produces regular pixel data
+     *
+     * @param pixels {number[] | Uint8Array | Uint8ClampedArray} array of pixel data
```
bigtimebuddy: One small change, could we make this private? I don’t like exposing new APIs we have to maintain backward compatibility for if it’s not absolutely necessary. If someone asks, we can make it public or move to utils or something.

ivanpopelyshev: Agree

PR #5878, opened 2019-07-04
Typo fix

danyisill:


themoonrat:
We still allowed to call people jerk faces? ;)

PR #5882, opened 2019-07-04
ViewableBuffer: Uint32View and Float32View dont need getters

ivanpopelyshev:
Subj. I just dont know cases that dont use one or both of them.

Pushing colors or something else in uint8 or int8 can happen but its rare.

Also Uint32 had wrong jsdoc type.

### DIFF
--- a/packages/core/src/geometry/ViewableBuffer.js
+++ b/packages/core/src/geometry/ViewableBuffer.js
@@ -134,8 +134,8 @@ export default class ViewableBuffer
         this._int16View = null;
         this._uint16View = null;
         this._int32View = null;
-        this._uint32View = null;
-        this._float32View = null;
+        this.uint32View = null;
+        this.float32View = null;
     }
 
     static sizeOf(type)
--- a/packages/core/src/geometry/ViewableBuffer.js
+++ b/packages/core/src/geometry/ViewableBuffer.js
@@ -27,7 +27,6 @@ export default class ViewableBuffer
          */
         this.uint32View = new Uint32Array(this.rawBinaryData);
 
-
         /**
          * View on the raw binary data as a `Float32Array`.
          *
--- a/packages/core/src/geometry/ViewableBuffer.js
+++ b/packages/core/src/geometry/ViewableBuffer.js
@@ -19,6 +19,21 @@ export default class ViewableBuffer
          * @member {ArrayBuffer}
          */
         this.rawBinaryData = new ArrayBuffer(size);
+
+        /**
+         * View on the raw binary data as a `Uint32Array`.
+         *
+         * @member {Uint32Array}
+         */
+        this.uint32View = new Uint32Array(this.rawBinaryData);
+
+
+        /**
+         * View on the raw binary data as a `Float32Array`.
+         *
+         * @member {Float32Array}
+         */
+        this.float32View = new Float32Array(this.rawBinaryData);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


SukantPal:
The getters were there so that `ViewableBuffer` could be used as a generic buffer. You might not want to use all the views. So, why initialize all of them.

You are initializing the `uint32` and `float32` in the constructor. If we use this somewhere else than in batch, it would be wasting memory and we would be allocating for no reason (unless those views were going to be used).

SukantPal:
Let us see what @bigtimebuddy says. 

I don't have a problem though. It is just that it breaks the uniformity of the code.

SukantPal:
You also have to fix `destroy()`, then.

bigtimebuddy:
Sorry Ivan, I don't understand this change. Seems superficial and there's no real benefit. In fact, I prefer the readonly getters instead of members.

SukantPal:
@bigtimebuddy, kind of agree.

ivanpopelyshev:
> Sorry Ivan, I don't understand this change. Seems superficial and there's no obvious benefit. In fact, I prefer the readonly getters instead of members.

It will make me calm that the class works like before, without extra getters. `packGeometry` didnt have any getter calls before we started the refactoring. I do not know about all that deoptimization stuff and how to detect it, but I prefer things to stay the same just in case. 

Typed arrays backed by ArrayBuffer consume only a handful of memory. I cant explain my sense of balance, but I really dont like it that we just changed stuff and now suddenly its canonical.

SukantPal:
There is no reason to the set the view. It will work like the same as before.

I don't think anyone would want to set the uint32View to a different value.

ivanpopelyshev:
I'm afraid of getters.

bigtimebuddy:
Is accessing these members "hot" code? If that's the case, let's measure the impact of getters so we understand the underlying issue. Also if you're going to add something that is seemingly anti-pattern in the code, you should make an inline comment about this being a performance optimization.

ivanpopelyshev:
> Also if you're going to add something that is seemingly anti-pattern in the code, you should make an inline comment about this being a performance optimization.

It was like that before last PR, I just noticed it too late. I want to stay it that way before someone who knows how to look at deoptimizations looks at it.

I asked @englercj to explain me those things before but i didnt get good answer from him.

ivanpopelyshev:
> This will cause a memory leak if left unchanged.

`destroy()` functions are also our paranoia, its actually the same level as "ivan doesn't like getters in critical code" :) I know js memory model enough to be sure that if we dont nullify all those small links in our destroys - nothing bad will happen. 

Event "in case pixijs will be ported to native C++ code" has very small probability.

I usually maintain `destroy()` methods just for other people to be calm, so please accept this change because it makes me calm.

Yeah, we can tag this PR with "paranoia" label, if we introduce one :) We already have X-Files.

ivanpopelyshev:
Another minor thing is that i dont want to fall into those functions while debugging flush and packGeometry step by step :) yeah, its all minor, just let me have this status quo for two fields please.

SukantPal:
I forgot I'm not writing C++. I agree. We could merge this if @bigtimebuddy agrees.

As far as "hot code" is concerned, I don't think time in JavaScript is much spent. Also, we are essentially caching the views in packInterleavedGeometry,

```
const { uint32View, float32View } = attributeBuffer;
```

SukantPal:
Be brave of getters, they're safe :P

ivanpopelyshev:
I took a big corporate holiday, I didnt just abandon this thing :)

englercj:
I know I'm late but for what it is worth I'm ambivalent about this change. I don't think the complexity of the getters justify the memory savings, and I don't think the property vs getter performance difference is big enough for us to care on modern engines.

The destroy methods are important because they notify systems to drop references they would otherwise keep. We don't need to do the "null out all members" thing anymore, but we do need to have destroy so the system can drop references we otherwise wouldn't know to drop.

PR #5887, opened 2019-07-05
Fixed parent not defined when calling updateTransform on a container with no parent

nathan-knight:
##### Description of change

I added checks for a missing parent to Container.updateTransform to prevent reference errors when calling it on a container with no parent. It can currently be worked around by wrapping the top level container you want to call updateTransform on in another container, but that should probably not be necessary and the documentation makes no mention of requiring a parent.

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas/canvas-renderer/src/CanvasRenderer.js
+++ b/packages/canvas/canvas-renderer/src/CanvasRenderer.js
@@ -200,7 +200,8 @@ export default class CanvasRenderer extends AbstractRenderer
 
             displayObject.parent = this._tempDisplayObjectParent;
 
-            displayObject.updateTransform();
+            // Use the internal version that doesn't check for the existence of a parent
+            displayObject._updateTransform();
             displayObject.parent = cacheParent;
             // displayObject.hitArea = //TODO add a temp hit area
         }
--- a/packages/core/src/batch/BatchRenderer.js
+++ b/packages/core/src/batch/BatchRenderer.js
@@ -425,7 +425,7 @@ export default class BatchRenderer extends ObjectRenderer
                 packedGeometries[this._flushId] = new (this.geometryClass)();
             }
 
-            packedGeometries[this._flushId]._buffer.update(attributeBuffer.vertices, 0);
+            packedGeometries[this._flushId]._buffer.update(attributeBuffer.rawBinaryData, 0);
             packedGeometries[this._flushId]._indexBuffer.update(indexBuffer, 0);
 
             this.renderer.geometry.bind(packedGeometries[this._flushId]);
@@ -435,7 +435,7 @@ export default class BatchRenderer extends ObjectRenderer
         else
         {
             // lets use the faster option, always use buffer number 0
-            packedGeometries[this._flushId]._buffer.update(attributeBuffer.vertices, 0);
+            packedGeometries[this._flushId]._buffer.update(attributeBuffer.rawBinaryData, 0);
             packedGeometries[this._flushId]._indexBuffer.update(indexBuffer, 0);
 
             this.renderer.geometry.updateBuffers();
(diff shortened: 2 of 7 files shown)


ivanpopelyshev:
When I tried to deal with that kind of problems i also made tests, but PR's were too big because i tried to fix all bad cases and they were rejected.

Even if we focus on finally fixing that thing, I suggest we move it to 5.2.0

nathan-knight:
> This thing doesn't update element transform at all, there's even special method `updateLocalTransform()` but its not being called here.

I can add that call for when the container has no parent. Are there any other issues I should address when I make the change?

ivanpopelyshev:
I think we can accept this change if you add `updateLocalTransform` call here.

I just worry because we had this change before: #3415 , and there were other PR's, and @englercj said something like `I dont want that if here`. I really want a confirmation from @englercj and @GoodBoyDigital because that's a very important place.

GoodBoyDigital:
Hey ya! We should performance test this code with maybe bunny mark? To ensure that there is no performance impact. I know this may sound silly, but this is the hottest part of the code base. Just some due diligence before we roll it in!

nathan-knight:
> Hey ya! We should performance test this code with maybe bunny mark? To ensure that there is no performance impact. I know this may sound silly, but this is the hottest part of the code base. Just some due diligence before we roll it in!

I gave bunny mark (https://pixijs.io/bunny-mark/ ?) a shot but it does not seem to be working. Is this the wrong place/way to be using bunny mark? Either way, I opened an issue over there (https://github.com/pixijs/bunny-mark/issues/7).

Other than that, I added the `updateLocalTransform` call and updated the test to check that it is called as expected.

englercj:
We did at one point notice a very big impact from having a branch in this code. At the time we decided to instead ensure that any object who has updateTransform called on it *always* had a parent to remove the need for the branch.

>  It can currently be worked around by wrapping the top level container you want to call updateTransform on in another container

This is exactly what we've done in previous versions.

> but that should probably not be necessary and the documentation makes no mention of requiring a parent.

We should document this behavior or have the public version of updateTransform add an empty container as a parent before calling the "real" update function which doesn't branch. Then have the library internals always call that "real" function.

nathan-knight:
> We should document this behavior or have the public version of updateTransform add an empty container as a parent before calling the "real" update function which doesn't branch. Then have the library internals always call that "real" function.

That's a good idea. I've removed the branching from the internal `updateTransform` (now called `_updateTransform`) and moved the parent check to the public `updateTransform` method in DisplayObject. I also switched everything that I could find that used `updateTransform` over to `_updateTransform` where appropriate.

ivanpopelyshev:
In terms of performance and API , yes, this is solution. 

However, `render` vs `_render`, `calculateBounds` vs `_calculateBounds`, `updateTransform` vs `_updateTransform` . In the last case its not "container" vs "displayObject", its "outer" vs "inner". 

And we already have `_recursivePostUpdateTransform` . Let the better naming for `_updateTransform` begin!

One more thing, that breaks plugin compatibility. Some plugins use `updateTransform` as a default way to update the object, like `pixi-spine`. we'll need to add `if's` in those plugins to check whether new internal function is available.

ivanpopelyshev:
OK, how about that? We separate it to two functions: 

1. `_preRenderUpdateTransform` that takes into account sorting and whatever update should happen in a plugin. 

2. `updateTransform` will have the `if` @nathan-knight suggested.

Functions are independent, they don't call each other. Pre-Render update has the same performance as before, and `updateTransform` wont have any side-effects.

bigtimebuddy:
I'm going to close this PR. I chatted about this with @GoodBoyDigital and we agreed that this can be solved when we (re-)introduce the Stage object back into Pixi v5. I don't think there's consensus about the best way to do this right now from an API standpoint and it's controversial because it's one of the hottest piece of source code. Also, it seems like there's a workaround by wrapping in a Container. Need to give this some more thought.

PR #5890, opened 2019-07-05
Moved _renderCanvas call from text package into new canvas-text package

themoonrat:
It's only required for legacy builds, not normal WebGL only builds

Also renamed render to _render, as internally in other areas of the code, when overriding a render method it is always the underscored version

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas/canvas-text/package.json
+++ b/packages/canvas/canvas-text/package.json
@@ -1,0 +1,28 @@
+{
+  "name": "@pixi/canvas-text",
+  "version": "5.0.4",
+  "main": "lib/canvas-text.js",
+  "module": "lib/canvas-text.es.js",
+  "bundle": "dist/canvas-text.js",
+  "bundleNoExports": false,
+  "description": "Canvas mixin for the text package",
+  "author": "Dave Moore",
+  "homepage": "http://pixijs.com/",
+  "bugs": "https://github.com/pixijs/pixi.js/issues",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/pixijs/pixi.js.git"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "lib",
+    "dist"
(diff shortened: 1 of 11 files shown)


packages/canvas/canvas-text/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@pixi/canvas-text",
+  "version": "5.0.4",
+  "main": "lib/canvas-text.js",
+  "module": "lib/canvas-text.es.js",
+  "bundle": "dist/canvas-text.js",
+  "bundleNoExports": false,
```
bigtimebuddy: This should be `true` since this doesn't export anything. It's like the mixin* packages.
```suggestion
  "bundleNoExports": true,
```

packages/canvas/canvas-text/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "lib",
+    "dist"
+  ],
+  "dependencies": {
+	  "@pixi/sprite": "^5.0.4",
```
bigtimebuddy: whitespace
```suggestion
	   "@pixi/sprite": "^5.0.4",
```

packages/canvas/canvas-text/src/Text.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Renders the object using the Canvas renderer
+ *
+ * @method _renderCanvas
+ * @memberof PIXI.Text#
+ * @private
+ * @param {PIXI.CanvasRenderer} renderer - The renderer
+ */
+Text.prototype._renderCanvas = function _renderCanvas(renderer) // eslint-disable-line no-unused-vars
```
bigtimebuddy: I don't think this comment is needed anymore. Render is used here.
```suggestion
Text.prototype._renderCanvas = function _renderCanvas(renderer)
```

PR #5891, opened 2019-07-05
Set parentID to -1 rather than worldID for when flagging parent transform as dirty

themoonrat:
Was recommended in a comment for a previous PR, and I've seen no negative side efffects
https://github.com/pixijs/pixi.js/pull/5030#issuecomment-441299080

### DIFF
--- a/packages/canvas/canvas-renderer/src/CanvasRenderer.js
+++ b/packages/canvas/canvas-renderer/src/CanvasRenderer.js
@@ -191,7 +191,7 @@ export default class CanvasRenderer extends AbstractRenderer
                 transform.copyTo(tempWt);
 
                 // lets not forget to flag the parent transform as dirty...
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
You guys are stepping on a very dangerous territory of #4632 . When I tried to made certain small changes, I found many edge cases that were either broken already, either broken by my changes. I had tests for it. It was too difficult for us back then. 

Maybe you guys finally ready try to fix temp parent approach, but that doesn't mean we can just merge random changes and hope that it works.

themoonrat:
@a-rasinski This PR for v5 stems from your findings in v4: https://github.com/pixijs/pixi.js/pull/5030#issuecomment-441299080

Is there any chance that you could reproduce your scenario with vanilla v5 and see if this change is required or not for v5 please?

ivanpopelyshev:
Close in favor of #5891, I made better fix.

Issue #5711, opened 2019-05-22
Using Graphics native lineStyle does not close shapes


stagr:
### Expected Behavior
The below code, with the lineStyle parameter "native" set to true, should paint a rectangle with a 1 pixel wide red border.
```
    var graphics2 = new PIXI.Graphics();
    graphics2.lineStyle(1, 0xFF0000, 1, 0.5, true);
    graphics2.drawRect(10, 120, 100, 100);
```
### Current Behavior
The rectangle is painted with only three sides, the border of the left side is not visible at all.

### Possible Solution
No idea.

### Steps to Reproduce

https://www.pixiplayground.com/#/edit/yo2dXuKUxzhfGphLmRC0k

### Environment

- **`pixi.js` version**: 5.03
- **Browser & Version**: Chrome 74.0.3729.157, Microsoft Edge 42.17134.1.0, Firefox 66.0.5 (64-bit)
- **OS & Version**: Windows 10
- **Running Example**: https://www.pixiplayground.com/#/edit/yo2dXuKUxzhfGphLmRC0k

stagr:
Pinging this to see if anyone has any news on this one?

bigtimebuddy:
Yup, this definitely looks like a bug.

JetLua:
Add notes: If you set the color of `lineStyle` to `0xffffff`, `Graphics` will not be displayed.

https://www.pixiplayground.com/#/edit/eyv91KYjD~Gz~P0N95cC_

bigtimebuddy:
That's more serious. 

Here's the same bug but showing on **dev** instead of the latest release. Will be helpful to verify.

https://www.pixiplayground.com/#/edit/EWjPwKmpr9LlGPYHugusU

cc @ivanpopelyshev

stagr:
This is now working in 5.1.0. Closing the issue. Thanks a lot!

PR #5911, opened 2019-07-14
fix & refactor updateBatches logic

ivanpopelyshev:
#5711 

Original Test: https://www.pixiplayground.com/#/edit/IVHxo3MM4vPZMZpn2VvPS

Verification: https://www.pixiplayground.com/#/edit/jh1VK3gUN~m2DusR2prgx

The problem is that it actually takes default `fillStyle` instead of our `lineStyle`. `fillStyle.visible=false`, so its not filling anything, but it doesn't take into account that `lineStyle` has `native` prop that we need later.

Of course just adding `currentNative` could fix that , but I really want to make the code readable.

I will also re-use the same approach in BatchRenderer later, to make customizable batch logic.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+open_collective: pixijs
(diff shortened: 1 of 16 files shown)


bigtimebuddy:
Do you know why the native line shape is not closed?

ivanpopelyshev:
I think it doesn't take into account `shape.closeStroke;`

ivanpopelyshev:
OK, I fixed & refactored that too. It could go in another PR, but you insisted ;)

ivanpopelyshev:
Graphics without style but with bounds: https://www.pixiplayground.com/#/edit/CZLJ66klqelKBLw7TnLoD

PR #5913, opened 2019-07-15
Minor typings fix for ViewableBuffer

ivanpopelyshev:
subj.

### DIFF
--- a/packages/core/src/geometry/ViewableBuffer.js
+++ b/packages/core/src/geometry/ViewableBuffer.js
@@ -115,7 +115,7 @@ export default class ViewableBuffer
      *
      * @param {string} type - One of `int8`, `uint8`, `int16`,
      *    `uint16`, `int32`, `uint32`, and `float32`.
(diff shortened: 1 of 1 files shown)


packages/core/src/geometry/ViewableBuffer.js
```diff
@@ -115,7 +115,7 @@ export default class ViewableBuffer
      *
      * @param {string} type - One of `int8`, `uint8`, `int16`,
      *    `uint16`, `int32`, `uint32`, and `float32`.
-     * @return typed array of given type
+     * @return {object} typed array of given type
```
bigtimebuddy: Hmm wait shouldn’t this be a list of the various types arrays instead of object?

ivanpopelyshev: Actually, it should be N separate declarations , like `('uint8') : Uint8Array` and so on. I really don't want to mention all the typed array types here.

Issue #5894, opened 2019-07-06
BatchRenderer isn't exported properly


SukantPal:
We are exporting `BatchPluginFactory.create()` rather than the actual `BatchRenderer` class. 

`index.js` in `@pixi/core` exports `BatchRenderer` from `BatchPluginFactory.js` rather than `BatchRenderer.js`. This means that end-users implementing batch rendering will be inheriting from an actually useless middle-man.

```
// BatchPluginFactory.js


// Setup the default BatchRenderer plugin, this is what
// we'll actually export at the root level
export const BatchRenderer = BatchPluginFactory.create();
```

```
// index.js

export { default as BatchPluginFactory, BatchRenderer } from './batch/BatchPluginFactory';
```

I don't think this was intended.

To maintain backward compatibility, we can rename the actual `BatchRenderer` class to `AbstractBatchRenderer` and export that as well. This will work since the actual class isn't exported right now.

bigtimebuddy:
Yes please rename to AbstractBatchRenderer. 👍

PR #5919, opened 2019-07-17
Exports AbstractBatchRenderer

SukantPal:
This should solve #5894 as described there.


themoonrat:
Some of the jsdoc descriptions have got very narrow. Are you coding in portrait mode? ;)

SukantPal:
@themoonrat I think comments should be a little less than the width of the lines below it, so they look a symmetrical.

Very long comments are hard to read and can be distracting, especially if the horizontal scroll bar pops up.

themoonrat:
But then on the other hand, you now have to scroll through tons more lines to find the code you're after.

I just thought we had an eslint rule that handled long lines; what I don't want is each time a different person gets to edit a certain jsdoc they change it from one persons preferred style to another persons preferred style!

Wouldn't block the PR over it, just noticed some changes weren't actual changes but formatting changes to your preferred style. If you'd added new code and commented on and it had been narrow I probably wouldn't have noticed!

@bigtimebuddy  ?

bigtimebuddy:
I agree with @themoonrat, I think the excessive line breaking on the comments is unnecessary, especially since @SukantPal didn't change any substance to the comments. Also, if consistent line-length is an issue that needed to be fixed (not saying it is), using something like prettier or some other tool seems like a better more universal approach across the whole codebase. Otherwise this just introduces a dev-style in one part of the code, which isn't idea.

The AbstractBatchRenderer change is a good one though.

SukantPal:
@bigtimebuddy Yeah, I was working on `DisplayObject` before; however, I forgot about that.

ivanpopelyshev:
It works: https://www.pixiplayground.com/#/edit/wDkRfBh_mws1~L3IZkKC6

OK, breaking pixi-projection, take four. I'll have to update pixi-projection again after this thing lands.

SukantPal:
I wonder why - why is pixi-projection breaking?

ivanpopelyshev:
Because that's the only plugin that uses this feature.

https://github.com/pixijs/pixi-projection/blob/master/src/base/webgl/Sprite2dRenderer.ts#L70

SukantPal:
@ivanpopleyshev Well, this refactoring is backward-compatible. I haven't removed the `BatchRenderer` export.

Think of it as a new API, not a renaming.

SukantPal:
Ivan, this PR will not break existing code. If you use `AbstractBatchRenderer`, it will make stuff a bit faster, since you aren't going to use `PIXI.BatchRenderer`. Rather, you only need to override a method from it's parent (`AbstractBatchRenderer`).

ivanpopelyshev:
oh, right, you re-exported it. Nice! I agree :)

SukantPal:
@ivanpopelyshev You don't get it. I didn't re-export it. The `BatchRenderer` class WAS NEVER EXPORTED. Only `BatchPluginFactory.create()` was, as `BatchRenderer`, which shadowed the actual class.

I just renamed the class and exported that too.

ivanpopelyshev:
I get it, you know I'm not good with words :)

bigtimebuddy:
Thanks @SukantPal!

Issue #5929, opened 2019-07-21
Cannot read property 'disposeGeometry' of undefined


mudcube:
### Possible Solution
Loop backwards through Runner items.

### Steps to Reproduce
1. Create a `PIXI.Mesh`. 
2. Render it to two `PIXI.Renderer`'s
3. Destroy results in error

### Environment

- **`pixi.js` version**: 5.1
- **Browser & Version**: Chrome 75
- **OS & Version**: MacOS
- **Running Example**: https://www.pixiplayground.com/#/edit/yjTOm9m7zJN0XMYWMVTCv

mudcube:
Resolved in #5975

PR #5930, opened 2019-07-21
Fix Cannot read property 'disposeGeometry' of undefined (#5929)

mudcube:
##### Description of change
Loop through runner items backwards, in case one of the items removes itself from the array.

https://www.pixiplayground.com/#/edit/yjTOm9m7zJN0XMYWMVTCv

##### Pre-Merge Checklist

##### Note
Just realizing that I unfortunately cannot run `npm run test` due to a [separate issue](https://github.com/pixijs/pixi.js/issues/5931) which happens from a clean install on my machine with the latest master.

### DIFF
--- a/packages/runner/src/Runner.js
+++ b/packages/runner/src/Runner.js
@@ -62,8 +62,9 @@ export default class Runner
         }
 
         const { name, items } = this;
+        let i = items.length;
 
-        for (let i = 0, len = items.length; i < len; i++)
+        while (i--)
         {
             items[i][name](a0, a1, a2, a3, a4, a5, a6, a7);
         }


ivanpopelyshev:
Please make an example for disposeGeometry, thats my method and k thought i've taken into account that thing

ivanpopelyshev:
Did you achieve that with two renderers?

ivanpopelyshev:
Oh, right, here it is: 

https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/geometry/Geometry.js#L277
https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/geometry/GeometrySystem.js#L464

Runners dont have , maybe mini-signals will be better in that case. 
@GoodBoyDigital 
@englercj

ivanpopelyshev:
Demo: https://www.pixiplayground.com/#/edit/0fUXKi8LVTJBLB5JOC~JB

mudcube:
> Did you achieve that with two renderers?

Yup, exactly.

> My idea is either we add option to reverse ticker order, either we use signal.

My only concern with reverse order is how it might affect things... that's to say, is anything expecting things to be fired in a certain way?

A third option might be to clone the `const items = [...ref.items]` before looping through it. But that might create necessary garbage?

bigtimebuddy:
We need unit test coverage to reflect this change.

GoodBoyDigital:
Hello! Figured it might actually be quite handy to be able to remove something mid run. This situation is a great example of that! 

https://github.com/pixijs/pixi.js/pull/5972

bigtimebuddy:
Closing this in favor of #5972 or #5975

PR #5934, opened 2019-07-21
Users can precisely control FPS

JetLua:
When I set `app.ticker.maxFPS = 50`:

Fixed:

before: https://www.pixiplayground.com/#/edit/hbR16awXmdxpRaCQvx71U
after: https://www.pixiplayground.com/#/edit/z06c_5wDi0co39V~kITu4

You may need to tap save or refresh to display the number of frames correctly.


packages/ticker/src/Ticker.js
```diff
@@ -137,6 +137,14 @@ export default class Ticker
          */
         this._protected = false;
 
+        /**
+         * The last time keyframe was executed.
+         * @member {number}
+         * @default -1
+         * @private
+         */
+        this._lastKey = -1;
```
bigtimebuddy: Couple of minor comments: Why a default value of -1? Also, could you indicate units in description... milliseconds I assume? And finally maybe “lastFrame” is a better name than “lastKey”

JetLua: 👌

bigtimebuddy:
@themoonrat thanks for reviewing. Do you have some code that produce those results that we can re-use to test future changes on this PR?

JetLua:
@themoonrat Thank you for your detailed test results. I made some changes to make sure that most of the time `FPS` is below `maxFPS`.

bigtimebuddy:
@themoonrat how does this look to you?

themoonrat:
The comments above
`if (this._minElapsedMS)`
Are still out of date and refer to my old implementation

themoonrat:
I can still easily exceed the maxfps: setting it as 45 on a 60fps monitor and the fps reports at 49 :/

ivanpopelyshev:
@themoonrat I think at this point you are nitpicking =)

JetLua:
Perhaps it's the interference of other things, and the performance on my machine (60FPS) is very close to the target value.

Occasionally there are more than `maxFPS` cases, basically stable in `± 2~3` FPS.

maxFPS: 40
TIM图片20190730195903
maxFPS: 45
TIM图片20190730195920
maxFPS: 50
TIM图片20190730195916

bigtimebuddy:
Thanks @JetLua for another awesome PR :tada:

PR #5938, opened 2019-07-22
Fix TextMetrics OffscreenCanvas context null

mcdaid01:
Found a bug in iOS12.3.0 webview that stops PIXI.Text from working, note it does not affect iOS safari browser. Traced it down to TextMetrics.js getContext('2d') of OffscreenCanvas returning null.

### DIFF
--- a/packages/text/src/TextMetrics.js
+++ b/packages/text/src/TextMetrics.js
@@ -705,11 +705,7 @@ const canvas = (() =>
     {
         // OffscreenCanvas2D measureText can be up to 40% faster.
         const c = new OffscreenCanvas(0, 0);
-
-        if (c.getContext('2d')===null)
-            throw('context null');
-
-        return c;
+        return c.getContext('2d') ? c : document.createElement('canvas');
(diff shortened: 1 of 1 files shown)


mcdaid01:
> I dont like exceptions. What jf we just copy 'return' statement here?

I rarely use them myself, it was as already in a try catch statement.
So instead

 if (c.getContext('2d')===null)
            return document.createElement('canvas');

themoonrat:
Yeps - but with our eslint validated formatting, which would be something like
```
if (c.getContext('2d'))
{
    return c;
}

return document.createElement('canvas');
```
or
```
return c.getContext('2d') ? c : document.createElement('canvas');
```

It always makes better reading for me if you test for truthy, return the thing you want, with the fail safe being below. Otherwise you're checking for a falsey, and then doing something you hope not to do half way through the code, and then finally doing the good thing you want at the very end.

mcdaid01:
Ok, that makes sense. I will correct and resubmit

PR #5941, opened 2019-07-22
Adds generic types to container child methods

### LINKED ISSUE #5927: Improve typings for Container.addChild and Container.addChildAt
We're switching from Pixi4 to 5, and using the included typings instead of https://github.com/pixijs/pixi-typescript.

Container.addChild and Container.addChildAt are not generic (excerpt from pixi.d.ts included in Pixi5):
```
addChild(...child: PIXI.DisplayObject[]): PIXI.DisplayObject;
addChildAt(child: PIXI.DisplayObject, index: number): PIXI.DisplayObject;
```

Therefore,  code like this doesn't work:

```
const sprite = this.stage.addChild(new PIXI.Sprite(tex));
sprite.anchor.set(0.5);
```
In this example, sprite is inferred to be a PIXI.DisplayObject, so we get the following error:
`Property 'anchor' does not exist on type 'DisplayObject'.`

If we add generics to the code, the types are properly inferred and everything works as expected:

```
addChild<TChildren extends PIXI.DisplayObject[]>(...child: TChildren): TChildren[0];
addChildAt<T extends PIXI.DisplayObject>(child: T, index: number): T;
```
Right now we're using a custom version of pixi.d.ts with these changes, and everything works great.

The pixi4 typings has support for generics, so I think this may be an issue for any typescript users moving to Pixi5.

I attempted to put together a PR for this, but I saw that the typings are automatically generated from the JSDoc comments, and I'm unsure how (or if it's even possible) to type something like this.

If someone could point me in the right direction, I'd be more than happy to help out with a PR!

jmlee2k:
##### Description of change
Adds generic types to Container.addChild, Container.removeChild and Container.addChildAt, allowing for the proper subclass of PIXI.DisplayObject to be returned.

fixes #5927.

##### Pre-Merge Checklist

### DIFF
--- a/types/assemble.js
+++ b/types/assemble.js
@@ -47,6 +47,18 @@ if (bundle === 'pixi.js') {
         .replace(/PIXI\.prepare\.CanvasPrepare \| (PIXI\.Renderer)/g, '$1');
 }
 
+// tsd-jsdoc doesn't currently support indexed generics (TChildren[0])
+buffer = buffer.replace(
+    /(addChild|removeChild)\(\.\.\.child: PIXI.DisplayObject\[\]\): PIXI\.DisplayObject;/g,
+    '$1<TChildren extends PIXI.DisplayObject[]>(...child: TChildren): TChildren[0];'
+);
+
+// tsd-jsdoc supports this case using the @template tag, but using said tag breaks documentation generation
+buffer = buffer.replace(
+    /addChildAt\(child: PIXI\.DisplayObject, index: number\): PIXI\.DisplayObject;/g,
+    'addChildAt<T extends PIXI.DisplayObject>(child: T, index: number): T;'
+);
+
 // EventEmitter3 and resource-loader are external dependencies, we'll
 // inject these into the find output
 const eventTypes = fs.readFileSync(path.join(__dirname, 'events.d.ts'), 'utf8');


jmlee2k:
I initially attempted adding generics to the jsdoc as @englercj suggested.  I was able to do it for addChildAt, but it breaks documentation generation:
```ERROR: The @template tag is not a known tag. File: Container.js, line: 149```

I assume this is due to what was mentioned here: https://github.com/pixijs/pixi.js/issues/5927#issuecomment-513473618.

I wasn't able to add it to the jdsoc for addChild/removeChild, as it doesn't like the syntax for indexed generics:
```
ERROR: Unable to parse a tag's type expression for source file C:\work\github\pixi.js\packages\display\src\Container.js in line 99 with tag title "return" and text "{TChildren[0]} The first child that was added.": Invalid type expression "TChildren[0]": Expected "!", "#", "$", "(", "-", ".", "/", "0", ":", "<", "=", "?", "@", "[]", "\\", "\u200C", "\u200D", "_", "|", "~", Unicode combining mark, Unicode decimal number, Unicode letter number, Unicode lowercase letter, Unicode modifier letter, Unicode other letter, Unicode punctuation connector, Unicode titlecase letter, Unicode uppercase letter, [1-9] or end of input but "[" found.
```
In the end, I just made all the changes in assemble.js.

ivanpopelyshev:
not related to PR, i think we need something more effective than just replace.

@englercj @bigtimebuddy lets discuss?

englercj:
MOdern jsdoc supports `@template` natively. I'd rather update and fix tsd-jsdoc to work with the newest version.

bigtimebuddy:
We've been pinned to jsdoc@3.5.5, upgrading to 3.6.x has crippled part of the docs because of the under-the-hood parser changes. I'm okay if we need to upgrade to support `@template`, but that will be some work to resolve the quirks and verify docs.

bigtimebuddy:
Thanks @jmlee2k

jmlee2k:
My pleasure!

PR #5945, opened 2019-07-23
Fixing batch breaking by empty filter array.

eXponenta:
Empty filter array in Container breaks batched render.

Bug replay:
https://www.pixiplayground.com/#/edit/pRUcpdTxAEml0qiV83HUI

Remove line 37 and bunny will nbe rendered.

##### Pre-Merge Checklist

- [ x] Tests and/or benchmarks are included

### DIFF
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -494,7 +494,7 @@ export default class Container extends DisplayObject
         }
 
         // do a quick check to see if this element has a mask or a filter.
-        if (this._mask || this.filters)
+        if (this._mask || (this.filters && this.filters.length))
         {
             this.renderAdvanced(renderer);
         }


ivanpopelyshev:
We have the same problem in v4: https://github.com/pixijs/pixi.js/blob/v4.x/src/core/display/Container.js

I'm not sure when exactly was that thing introduced, lets bisect it

ivanpopelyshev:
OMG, it was always like that! I checked several versions through the time down to beginning of 2015 and there was `flush()` even if filters length was zero.

bigtimebuddy:
I also cherry-picked this change into v4.

One feature that we added in v4 was the ability to set a filter to `enabled=false` which would skip the filter from display. This is reflected in the `renderAdvanced` call. However, if we wanted to make sure to _only_ run the advanced render if we are showing a filter, we would want to check for a valid length of _enabled_ filters.

PR #5946, opened 2019-07-24
RenderTexturePool and FilterSystem changes

ivanpopelyshev:
WIP, here'll be demos soon, explaining everything.

### DIFF
--- a/packages/core/src/renderTexture/RenderTexturePool.js
+++ b/packages/core/src/renderTexture/RenderTexturePool.js
@@ -8,6 +8,9 @@ import { nextPow2 } from '@pixi/utils';
  * Texture pool, used by FilterSystem and plugins
  * Stores collection of temporary pow2 or screen-sized renderTextures
  *
+ * If you use custom RenderTexturePool for your filters, you can use methods
+ * `getFilterTexture` and `returnFilterTexture` same as in
+ *
  * @class
  * @memberof PIXI
  */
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
Demo? Also is this done?

ivanpopelyshev:
Demo: https://www.pixiplayground.com/#/edit/4rskiJ8lb8k6437WmSwKy

OK, while making last-minute changes to #5771 I actually broke that demo :(

The problems are 
1. RenderTexturePool does not depend on renderer anymore, so it cant look at current renderTexture, and `getFilterTexture` needs RT as an argument.
2. getFilterTexture should have the same interface in FilterSystem and in RenderTexturePool to actually swap them easily in filters.

Another idea is that Texture should have `get resolution()`, to simplify `texture.baseTexture.resolution` , we use it often in places that aren't performance critical.

packages/core/src/textures/Texture.js
```diff
@@ -495,6 +495,16 @@ export default class Texture extends EventEmitter
         return null;
     }
 
+    /**
+     * Returns resolution of baseTexture
+     *
+     * @readonly
+     */
+    get resolution()
```
GoodBoyDigital: nice!

packages/core/src/renderTexture/RenderTexturePool.js
```diff
@@ -93,18 +96,20 @@ export default class RenderTexturePool
...
+     * `getFilterTexture(input, 0.5)` or `getFilterTexture(0.5, input)`
      *
-     * @param {number} [resolution] resolution, by default its the same as in current renderTexture
+     * @param {PIXI.RenderTexture} input renderTexture from which size and resolution will be copied
+     * @param {number} [resolution] override resolution of the renderTexture
+     *  It overrides, it does not multiply
      * @returns {PIXI.RenderTexture}
      */
-    getFilterTexture(resolution)
+    getFilterTexture(input, resolution)
```
GoodBoyDigital: I think we should rename input to something that makes it easy to know its a renderTexture `targetRenderTexture` ?

ivanpopelyshev: its `input` in filters. Name is the same as in https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/filters/Filter.js#L215 . Its supposed to be called from apply()

PR #5958, opened 2019-07-27
The RenderTexturePool should create renderTextures at resolution of 1

themoonrat:
Since internally it has already scaled the requested render texture size within getOptimalTexture
https://github.com/pixijs/pixi.js/issues/5954

### DIFF
--- a/packages/core/src/filters/FilterSystem.js
+++ b/packages/core/src/filters/FilterSystem.js
@@ -107,7 +107,7 @@ export default class FilterSystem extends System
          * stores a bunch of PO2 textures used for filtering
          * @member {Object}
          */
-        this.texturePool = new RenderTexturePool({ resolution: 1 });
+        this.texturePool = new RenderTexturePool();
 
         this.texturePool.setScreenSize(renderer.view);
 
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Good one. 

I suggest to add `resolution: 1` here: https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/renderTexture/RenderTexturePool.js#L46

That's real bug, I forgot that we have default resolution.

ivanpopelyshev:
I like it that you figured everything out. Now I know who to ask to test my changes in FilterSystem's,

themoonrat:
I wouldn't go that far ;) But glad to work out the issue!

The 2 options are to have it as an option passed through as in this pr, or hardcode it at the point you suggest. The one I went for had less changes, but putting it as you suggest means it becomes more of a default setting that could still be overridden by the settings. But is a better default than what it does now.

Don't mind changing it over; you've already approved @bigtimebuddy so happy for me to change over the style?

ivanpopelyshev:
Its hardcode because we need default implementation of that thing do not depend on resolution. Params are named `realWidth` and `realHeight` , that functions shouldn't create textures with res=2 by default!

ivanpopelyshev:
Can you also remove resolution from that doc? I believe that not everything that can be passed should be documented here. `resolution` will work wrong way. 

Its better if I add doc for Type.

PR #5959, opened 2019-07-27
Removed the ceil'ing of width and height in base texture

themoonrat:
As otherwise you lose precision going from one to the other.

Just trying to create some demo's that show up the issue in pixi playground

### DIFF
--- a/packages/core/src/textures/BaseTexture.js
+++ b/packages/core/src/textures/BaseTexture.js
@@ -378,8 +378,8 @@ export default class BaseTexture extends EventEmitter
     setRealSize(realWidth, realHeight, resolution)
     {
         this.resolution = resolution || this.resolution;
-        this.width = Math.ceil(realWidth / this.resolution);
-        this.height = Math.ceil(realHeight / this.resolution);
+        this.width = realWidth / this.resolution;
+        this.height = realHeight / this.resolution;
         this._refreshPOT();
         this.update();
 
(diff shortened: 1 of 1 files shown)


themoonrat:
Grrr, thought this would be an easy one to reproduce - will come back and have a fresh think about why this fixed an issue in my version of the libs and why it's not doing anything here.

ivanpopelyshev:
> As otherwise you lose precision going from one to the other.

> Just trying to create some demo's that show up the issue in pixi playground

> Grrr, thought this would be an easy one to reproduce - will come back and have a fresh think about why this fixed an issue in my version of the libs and why it's not doing anything here.

Welcome to the X-Files. I told you about extra `eps` before!

themoonrat:
v5: https://www.pixiplayground.com/#/edit/wpSKMBzOgh1zvO6q~_iBw
v5 with this PR: https://www.pixiplayground.com/#/edit/RVnw2iNv3lVmSmNf5kBcA
v4: https://www.pixiplayground.com/#/edit/ZtI5gS5beIhU~UqRTx98b

themoonrat:
Tbf, in this test, even v4 gets the height wrong.... works fine with Texture.WHITE

ivanpopelyshev:
Good examples. My `eps` doesnt help there: https://www.pixiplayground.com/#/edit/VofrhNr_0CURPAIW2LB49

PR #5964, opened 2019-07-28
Support for lineStyle and lineTextureStyle options

bigtimebuddy:
This API has been bugging me since v5...

We introduced `Graphics#lineTextureStyle(width, texture, color, alpha, matrix, alignment, native)` and this is super unwieldy to use. Because the arguments aren't _really_ in any order of importance, users often need to set defaults or ignore arguments. It's too much cognitive load to remember and makes code less readable. I constantly find myself having to refer to the docs for this function.

```js
graphics.lineTextureStyle({
  texture: PIXI.Texture.from('foo.jpg'),
  alpha: 0.5,
  alignment: 1,
  width: 10
});
// instead of:
graphics.lineTextureStyle(10, PIXI.Texture.from('foo.jpg'), 0x0, 0.5, null, 1);
```

This PR introduces an options object that can be passed to the `lineTextureStyle` and `lineStyle` methods. The old argument syntax for `lineTextureStyle` has been deprecated and `lineStyle` supports both styles.  

```js
// Now these are both supported equivalents:
graphics.lineStyle({ color: 0xffcc00, native: true });
graphics.lineStyle(1, 0xffcc00, 1, 0.5, true);
```

##### Pre-Merge Checklist


eXponenta:
I think same method arguments annotation style must be for Graphics#beginTextureFill and Graphics#beginFill for saving consistence of API also.

bigtimebuddy:
The reason I didn't do this for fill was because the arguments make a lot more logical sense and I believe are in the proper order of importance:

`beginFill(color, alpha)`
`beginTextureFill(texture, color, alpha, matrix)`

Maybe `beginTextureFill` would be nice to conform to this, but `beginFill`? Maybe there's an argument that if you're saving style externally, it's easier to apply it by reference:

```js

const style = {
  fill: { color: 0xff0000, alpha: 0.5 },
  line: { color: 0x00ff00, alpha: 0.1, alignment: 1 }
};
const graphics = new PIXI.Graphics()
  .beginFill(style.fill)
  .lineStyle(style.line);
```

ivanpopelyshev:
I didnt see any issues both here and on forums regarding lineTextureStyle. People do not use it, so deprecation + return is fine too.

ivanpopelyshev:
Yay! More coverage with obvious tests!

PR #5966, opened 2019-07-28
Better handling of webgl-less environments

### LINKED ISSUE #5664: Mesh-based classes crash if WebGL not present (pixi-legacy)
### Expected Behavior
Using autoDetectRenderer and not getting WebGL should fallback to canvas and use the canvas mesh renderers.

### Current Behavior
In a WebGL-disabled environment, creating a mesh-based object (e.g. SimplePlane, NineSlicePlane etc) will create a MeshMaterial, whose construction will eventually call getTestContext in core shader utils, which fails to get a valid WebGL context, and throws.

### Possible Solution
I've tested moving calls to getTestContext into try/catch blocks (as both places this method is called are followed by null-checks anyway), and this seems to fix the issue, but not sure if it's better to do this, or just make getTestContext not throw. Happy to make a PR for this if wanted.

### Steps to Reproduce
https://jsfiddle.net/3zshfLmy/
(Swap pixi-legacy for latest v4 and the plane initialisation lines to compare with v4 behaviour).
Firefox - about:config - webgl.disabled = true (or other webgl-disabling method)
Not reproducible by merely doing forceCanvas true, as WebGL context creation is still available.

### Environment

- **`pixi-legacy.js` version**: 5.0.1
- **Browser & Version**: Firefox Developer Edition 67.0b17
- **OS & Version**: Windows 7
- **Running Example**: https://jsfiddle.net/3zshfLmy/

bigtimebuddy:
Fixes #5664

The private function `getTestContext` is only called from these files. And in both cases, they handle if `gl` is `null`. It's not necessary to throw an exception.

* packages/core/src/shader/Program.js
* packages/core/src/shader/utils/getMaxFragmentPrecision.js

### DIFF
--- a/packages/core/src/shader/utils/getTestContext.js
+++ b/packages/core/src/shader/utils/getTestContext.js
@@ -1,7 +1,7 @@
 import { settings } from '../../settings';
 import { ENV } from '@pixi/constants';
 
-let context = null;
+let context;
 
 /**
  * returns a little WebGL context to use for program inspection.
@@ -12,7 +12,7 @@ let context = null;
  */
 export default function getTestContext()
 {
-    if (!context)
+    if (context !== undefined)
     {
         const canvas = document.createElement('canvas');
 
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Using both `null` and `undefined` ... hm..

PR #5970, opened 2019-07-29
Fix dev with getTestContext tweak

bigtimebuddy:
Broke dev with #5966

This fixes it.

Verify: https://pixijs.io/examples/?v=fix-test-context-2#/demos-basic/container.js

PR #5972, opened 2019-07-29
Feature/runner improvement

GoodBoyDigital:
##### Description of change
Removing items from a runner when its in the middle of running should no longer break, or have weird side effects

An alternative to PR #5970!

##### Pre-Merge Checklist

Once this builds will test the provided example...


bigtimebuddy:
Unit tests?

ivanpopelyshev:
you need to do `if (index !== -1)` anyway.

GoodBoyDigital:
@bigtimebuddy , unit tests are an excellent idea.. Will add them today.
@ivanpopelyshev , adding should have no effect as it get pushed onto the top and does not alter the structure of the array? Not a fan of coping each time, what to keep this as fast as possible.

ivanpopelyshev:
Its not each time, its when it was modified inside a cycle if there is more than one element.

GoodBoyDigital:
unit test added 👍

bigtimebuddy:
Closing in favor of #5975, confirmed with @GoodBoyDigital

PR #5975, opened 2019-07-30
Support mid-emit removal of Runner items

ivanpopelyshev:
Same as #5972 , but it handles many other cases, I'll prepare more tests.

This is not "I can code better" statement.

The approach was taken from mozilla shumway https://github.com/mozilla/shumway/blob/master/src/flash/events/EventDispatcher.ts#L66

I modified it a bit, we don't even spawn new array in case we had only one listener.


ivanpopelyshev:
I believe that code is more clear, and performance should be the same.

ivanpopelyshev:
@GoodBoyDigital 

We can also add method like `emitOnce` that will emit once and clear everything, ignoring all manual removals.

PR #5981, opened 2019-08-01
Fix lineStyle+fillStyle drawCalls count

ivanpopelyshev:
lineStyle never goes into one drawCall with fillStyle if batching is turned off. That means PixiJS just spams drawCalls for graphics! Bug exists even in 5.0.1

image

Demo has extra feature: it shows wireframe for fills!

Bug, 2 drawcalls: https://www.pixiplayground.com/#/edit/oXRqFpfuva9wwP6B_0V1d

Fixed, 1 drawcall: https://www.pixiplayground.com/#/edit/AjVLQ2NLsXQXM9KrUWCcp

### DIFF
--- a/packages/core/src/renderTexture/RenderTexturePool.js
+++ b/packages/core/src/renderTexture/RenderTexturePool.js
@@ -15,6 +15,7 @@ export default class RenderTexturePool
 {
     /**
      * @param {object} [textureOptions] - options that will be passed to BaseRenderTexture constructor
+     * @param {PIXI.SCALE_MODES} [textureOptions.scaleMode] - See {@link PIXI.SCALE_MODES} for possible values.
      */
     constructor(textureOptions)
     {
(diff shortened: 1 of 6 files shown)


bigtimebuddy:
This doesn’t make sense to me. Native is a line specific property. It doesn’t make sense to throw it on Fill. Are you sure you’re fixing the right thing here?

ivanpopelyshev:
Oh, right, I didnt explain it, sorry :)

One big problem: https://github.com/pixijs/pixi.js/blob/dev/packages/graphics/src/GraphicsGeometry.js#L653
Not big, but adds extra batch: https://github.com/pixijs/pixi.js/blob/dev/packages/graphics/src/GraphicsGeometry.js#L484

bigtimebuddy:
So fix that case. If native is undefined use false. Don’t change the API to something illogical.

ivanpopelyshev:
OK then, do you want me to change those to `!=` and comment "do not touch that" ?

bigtimebuddy:
Just don't change LineStyle or FillStyle. It's not fixing the problem, it just creating API confusion.

For fills, can you just do `!!style.native`

ivanpopelyshev:
ok, done

PR #5982, opened 2019-08-01
Restore stencil params after mask stack swap

### LINKED ISSUE #5974: Specific nested masks + filter breaks drawing (v5 and v4)
### Expected Behavior
Filters and masks should work together normally

### Current Behavior
Objects disappear and only certain masks are applied

### Steps to Reproduce
In Pixi5, the issue occurs when a filter is added to an object (A) which contains a masked child object, and object A's parent chain includes an EVEN number of masks >= 2.

In Pixi4, the issue occurs when object A's parent chain includes an ODD number of masks >= 3.

### Test Case

https://www.pixiplayground.com/#/edit/PmNR2aqebJB2i56E5IiBz

In the test case, 4 objects are created, each with a masked bunny and a gray bg.  Clicking on an object will move it down by 10px and add a ColorMatrixFilter which darkens the object, clicking again will restore it.

You can change the level of nesting by modifying the length prop on line 69 - a value of 0 will add the objects directly to the stage, 1 will make a single container, etc.  Each container has a different background color, and each mask's x position is offset by 10px to make it easier to visualize the nesting.

#### Pixi5
Changing the level of nesting to anything >= 2 will show the problem.  If the level of nesting is EVEN, clicking on the objects will make them disappear (since it adds a filter), and once all 4 have filters, the circle at the bottom will no longer be drawn.  If the level of nesting is odd, the objects won't disappear, but you can see that only alternating masks are being applied (changing the nesting level to 7 and clicking on the first object makes this very clear, as you can notice the "jailbars" effect at the left while the object is masked).

#### Pixi4
Running the same playground link in pixi4 will result in similar problems, except the nesting level has to be >=3, and the objects will disappear if the level of nesting is ODD, vs EVEN in pixi5.

ivanpopelyshev:
Fixes #5974

Broken: https://www.pixiplayground.com/#/edit/PmNR2aqebJB2i56E5IiBz
Fixed: https://www.pixiplayground.com/#/edit/t09LhS3_ttR6R7przuBOi

### DIFF
--- a/packages/core/src/mask/StencilSystem.js
+++ b/packages/core/src/mask/StencilSystem.js
@@ -31,8 +31,10 @@ export default class StencilSystem extends System
     setMaskStack(stencilMaskStack)
     {
         const gl = this.renderer.gl;
+        const curStackLen = this.stencilMaskStack.length;
 
-        if (stencilMaskStack.length !== this.stencilMaskStack.length)
+        this.stencilMaskStack = stencilMaskStack;
+        if (stencilMaskStack.length !== curStackLen)
         {
             if (stencilMaskStack.length === 0)
             {
@@ -44,8 +46,6 @@ export default class StencilSystem extends System
                 this._useCurrent();
             }
         }
-
-        this.stencilMaskStack = stencilMaskStack;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
@bigtimebuddy its not fully fixed - look #5974

jmlee2k:
@ivanpopelyshev, I think you made a typo, this should fix issue #5974, but you wrote that it fixes #5981.

ivanpopelyshev:
And now its truly fixed.

ivanpopelyshev:
@bigtimebuddy wanna make v4 PR for that too?

PR #5984, opened 2019-08-01
add eps in ceiling

ivanpopelyshev:
Follow-up for #5959 

Broken: https://www.pixiplayground.com/#/edit/d~jTog12dqKjAY9K1wBmg
Fixed: https://www.pixiplayground.com/#/edit/A_eMpc5bJ5C2M9KuR~wM4

consoleLog should report 15,15 , not 16,16

### DIFF
--- a/packages/core/test/BaseTexture.js
+++ b/packages/core/test/BaseTexture.js
@@ -174,4 +174,13 @@ describe('BaseTexture', function ()
         expect(resource.destroyed).to.be.true;
         expect(baseTexture.destroyed).to.be.true;
     });
+
+    it('should show correct width/height after setResolution', function ()
+    {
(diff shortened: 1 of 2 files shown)


packages/core/test/BaseTexture.js
```diff
@@ -174,4 +174,13 @@ describe('BaseTexture', function ()
         expect(resource.destroyed).to.be.true;
         expect(baseTexture.destroyed).to.be.true;
     });
+
+    it('should show correct width/height after setResolution', function ()
+    {
+        const texture = PIXI.RenderTexture.create({ width: 15, height: 15 });
```
bigtimebuddy: no `PIXI` global in tests, need you import RenderTexture from `@pixi/core`

PR #5986, opened 2019-08-02
Texture resolution type missing

ivanpopelyshev:
Follow-up for #5946 
subj


ivanpopelyshev:
dont merge yet, found another one

ivanpopelyshev:
here it is :) we've lost an array there.

PR #5988, opened 2019-08-02
Graphics close points epsilon

ivanpopelyshev:
Thanks to @eXponenta , one more obvious fix in Graphics.

Unfortunately, we are introducing new constant in GraphicsGeometry API , and without default settings (we can do it in next version).

Source: https://commons.m.wikimedia.org/wiki/File:Ghostscript_Tiger.svg

Broken: https://www.pixiplayground.com/#/edit/hA0cEqMaLra2qlieotC0F , look at the right - curve is not actually closed, and its real case from some SVG.
Fixed: https://www.pixiplayground.com/#/edit/GTnupAh_2mr6pMhvIUUfn

Why are those points different? Because `arc` in SVG cant be converted to our arc clearly, and while calculating coefficients for our arc rounding error can happen.

Alternative solution: provide a round resolution to graphicsGeometry tesselator. Make everything close to 0.001 or so.

@GoodBoyDigital are we ready to introduce an epsilon now for 5.1.1 or should we do it later? My point is - we already how strange constant inside the `arc` code.

### DIFF
--- a/packages/graphics/src/Graphics.js
+++ b/packages/graphics/src/Graphics.js
@@ -534,6 +534,7 @@ export default class Graphics extends Container
 
         const startX = cx + (Math.cos(startAngle) * radius);
         const startY = cy + (Math.sin(startAngle) * radius);
+        const eps = this.geometry.closePointEps;
 
         // If the currentPath exists, take its points. Otherwise call `moveTo` to start a path.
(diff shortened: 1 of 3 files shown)


eXponenta:
Its can't be const because there more cases with different precision.

GoodBoyDigital:
when does this const need to change? (just curious!)

PR #5989, opened 2019-08-02
InteractionManager: dispatch 'mouse/pointer-over' after 'mouse/pointer-out'

laino:
##### Description of change

Ensures that mouseover/pointerover events are always dispatched **after** mouseout/pointerout events.

Having these in essentially random ordering between components makes it unnecessarily complicated to react to hover state.

Consider this case:

```
let isMouseOver = false;

function onMouseOver() {
    isMouseOver = true;
}

function onMouseOut() {
    isMouseOver = false;
}

componentA.on('mouseover', onMouseOver);
componentB.on('mouseover', onMouseOver);

componentA.on('mouseout', onMouseOut);
componentB.on('mouseout', onMouseOut);
```

If componentA and componentB overlap, when you move from one component to the other in a certain direction with the cursor, you will receive a ``mouseover`` event **before** the other component dispatches the ``mouseout`` event, causing the above code to produce wrong results.

##### Pre-Merge Checklist


themoonrat:
Created a playground for v5 as it is: https://www.pixiplayground.com/#/edit/SX4ieoBSvUIoxAqnRjq_j
Going from red to green, you get 'true, true, false'
Going from green to red, you get 'true, false, true'

themoonrat:
I've created another playground with html divs by the side, to demonstrate that html wise, it always comes out 'true, false, true'

https://www.pixiplayground.com/#/edit/AYQVAUKAow0BC4EtqqQW3

laino:
Is there any chance this, or another fix for the same problem, could end up in the next release?

packages/interaction/src/InteractionManager.js
```diff
@@ -988,9 +1010,11 @@ export default class InteractionManager extends EventEmitter
      *  interactionEvent, displayObject and hit will be passed to the function
      * @param {boolean} [hitTest] - this indicates if the objects inside should be hit test against the point
      * @param {boolean} [interactive] - Whether the displayObject is interactive
+     * @param {boolean} [noDelayed] - Whether to process delayed events before returning. This is
+     *  used to avoid processing them too early during recursive calls.
      * @return {boolean} returns true if the displayObject hit the point
      */
-    processInteractive(interactionEvent, displayObject, func, hitTest, interactive)
+    processInteractive(interactionEvent, displayObject, func, hitTest, interactive, noDelayed)
```
bigtimebuddy: Would prefer `delayed` here instead of the opt-in negative. This creates a double-negative with `!noDelayed` which is confusing and less readable. 
```suggestion
    processInteractive(interactionEvent, displayObject, func, hitTest, interactive, delayed = true)
```

packages/interaction/src/InteractionManager.js
```diff
@@ -1065,7 +1089,7 @@ export default class InteractionManager extends EventEmitter
                 const child = children[i];
 
                 // time to get recursive.. if this function will return if something is hit..
-                const childHit = this.processInteractive(interactionEvent, child, func, hitTest, interactiveParent);
+                const childHit = this.processInteractive(interactionEvent, child, func, hitTest, interactiveParent, true);
```
bigtimebuddy: For recursive-sake, shouldn't this value be passed along here?

```suggestion
                const childHit = this.processInteractive(interactionEvent, child, func, hitTest, interactiveParent, delayed);
```

packages/interaction/src/InteractionManager.js
```diff
@@ -988,9 +1010,11 @@ export default class InteractionManager extends EventEmitter
      *  interactionEvent, displayObject and hit will be passed to the function
      * @param {boolean} [hitTest] - this indicates if the objects inside should be hit test against the point
      * @param {boolean} [interactive] - Whether the displayObject is interactive
+     * @param {boolean} [noDelayed] - Whether to process delayed events before returning. This is
```
bigtimebuddy: Documentation should include the default value when possible.
```suggestion
     * @param {boolean} [delayed=true] - Whether to process delayed events before returning. This is
```

packages/interaction/src/InteractionManager.js
```diff
@@ -1130,6 +1154,22 @@ export default class InteractionManager extends EventEmitter
             }
         }
 
+        const delayedEvents = this.delayedEvents;
```
bigtimebuddy: You can do this without creating a new array.

```suggestion
        const { delayedEvents } = this;
        const { length } = delayedEvents;
        
        if (delayed && length)
        {
                for (let i = 0; i < length / 3; i += 3)
                {
                       this.dispatchEvent(
                              delayedEvents[i],
                              delayedEvents[i+1],
                              delayedEvents[i+2]
                       );
                }
                delayedEvents.length = 0;
        }
```

packages/interaction/src/InteractionManager.js
```diff
@@ -945,6 +953,20 @@ export default class InteractionManager extends EventEmitter
...
+     * ordering of over/out events.
+     *
+     * @param {PIXI.Container|PIXI.Sprite|PIXI.TilingSprite} displayObject - the display object in question
+     * @param {string} eventString - the name of the event (e.g, mousedown)
+     * @param {object} eventData - the event data object
+     * @private
+     */
+    delayDispatchEvent(displayObject, eventString, eventData)
+    {
+        this.delayedEvents.push([displayObject, eventString, eventData]);
```
bigtimebuddy: Reduce the amount of objects/arrays that are created by pushing to a 1-dimensional list.
```suggestion
        this.delayedEvents.push(displayObject, eventString, eventData);
```

packages/interaction/src/InteractionManager.js
laino: I don't know whether something would make this safe further up the chain, but iterating over callbacks without using that new-array pattern is pretty much never a good idea.

Imagine one of the callbacks causes another event to be processed, if it adds another delayed event we'll iterate over everything twice.

packages/interaction/src/InteractionManager.js
laino: I'm personally no fan of default positives in function parameters, because if you use them
these become different calls:

```js
this.processInteractive(a, b, c, d, c, null); // delayed = null (false-ey)
this.processInteractive(a, b, c, d, c);       // delayed = true
````

You would either have to add code changing everything but ``false`` back to ``true``
or let the user deal with that.

Either would be worse than just making it a default false IMHO.

**Comment by laino** [comment]
> Would also love to see another simpler approach, if possible, to better align the pointerover and mouseover events.

Because this thing is recursive and you don't know whether you'll need to delay something in
a sibling, parent, or child, you'll have to memorize the stuff that must be dispatched later at some point. 

I *strongly* doubt there's a "simpler" approach without a fundamental rewrite. I'm open to learn something new though.

>Mainly, we need unit-tests to cover this new functionality and to handle all of the foreseeable cases.

I couldn't built this repo when I tried last, so I can't really write tests. I made these changes pretty much blind.

The only way I could test them was taking the pre-built package from npm and editing the changes in by hand, then try that with my project.

I can try again today, but I'm not gonna spend all day debugging this build process.

File: packages/interaction/src/InteractionManager.js (hunk 5 above)
Comment (laino) [reply packages/interaction/src/InteractionManager.js]: What you're suggesting would be slightly faster in V8 (which I believe can deal with that), but significantly slower in SpiderMonkey, which later likely won't be able to tell your array element's shapes as easily anymore (you're pushing stuff of at best 3 different shapes there).

The overall fastest way in all engines (if you really cared about perf) would be to just use an object, instead of confusing them with arrays containing varying stuff.

For the 3-4 elements we're pushing to that array, pretty much everything would be premature optimization though.

I played around a bit to confirm my suspicions, and it turns out that chrome is "pretty fast" in all cases, but Firefox kind of loses it if you try to put everything in 1 array:



https://jsperf.com/test-array-vs-array-in-array

**Comment by laino** [comment]
> Also, we share a single event object between everything that's dispatched from the interaction manager, so we should make sure that this isn't broken by delaying dispatching afterwards. 

None of these actually change it, so we should be good: ``processPointerDown``, ``processPointerUp``, ``processPointerCancel``, ``processPointerMove``, ``processPointerOverOut``.

The only place we change it is here: https://github.com/pixijs/pixi.js/blob/763808eccfbe61f1d4cc67053ccdd828b689f0a8/packages/interaction/src/InteractionManager.js#L1147

At the point we process the delayed (mouseout) events, that would now always contain the element that the mouse was moved to (if any).

This is (if I'm reading this correctly) a slight change from the previous behavior, where ``mouseout/mouseover`` would only contain that information if the pointer was moved to a child element.

**Comment by laino** [comment]
@bigtimebuddy Let me know if you still want any changes.

File: packages/interaction/src/InteractionManager.js (hunk 1 above)
Comment (RealDolos) [reply packages/interaction/src/InteractionManager.js]: `delayed === false` instead of `!delayed` false-y coercion  ;)

File: packages/interaction/src/InteractionManager.js (hunk 4 above)
Comment (RealDolos) [reply packages/interaction/src/InteractionManager.js]: Rule: never iterate over a collection that can mutate during the iteration. Unless you really REALLY know what you're doing.
Corollary: if you call unknown user code during an iteration, always assume that the user code will mutate your collection.

File: packages/interaction/src/InteractionManager.js (hunk 5 above)
Comment (bigtimebuddy) [reply packages/interaction/src/InteractionManager.js]: Then prefer object over array seems like a fair compromise.

File: packages/interaction/src/InteractionManager.js (hunk 1 above)
Comment (bigtimebuddy) [reply packages/interaction/src/InteractionManager.js]: What about naming the parameter “skipDelayed” to avoid the double negative?

File: packages/interaction/src/InteractionManager.js (hunk 4 above)
Comment (bigtimebuddy) [reply packages/interaction/src/InteractionManager.js]: Then slice/clone the array or iterative over it backwards to prevent issues. This is a reason why we needs unit tests.

**Comment by bigtimebuddy** [comment]
Thanks for your responses. Just FYI, we will not consider this pull requests without tests. It’s too complicated a piece of code. Let me know why you couldn’t build from source, happy to help debug that with you. If you don’t have time to finish, then let me know and we can have someone else finish it. I agree that it’s an important fix.

**Comment by laino** [comment]
@bigtimebuddy You need to pin your jsdoc dependency to a version >= 3.6.0, so the build will be fixed in node 12: https://github.com/hegemonic/requizzle/issues/5 (jsdoc depends on requizzle)

**Comment by laino** [comment]
@bigtimebuddy Also, because you're installing your own eslint to node_modules, your eslint config should include ``{"root": true, ...``}. Because otherwise it is gonna try to inherit from my global eslint config, while just looking for plugins within your node_modules folder. Then I get errors like these:

```
ESLint couldn't find the plugin "@typescript-eslint/eslint-plugin"
```

bigtimebuddy:
Ah yes, that JSDoc thing is a known issue. Best to use Node 10 with `nvm`, if you can. We have to be pinned for the moment until we can update our docs to JSDoc >= 3.6.

Could you give me some more context on the eslint issue? Is this also Node 12 related? It would be awesome if you could capture that in a new issue so we can track it separately. Whereas the JSDoc thing is a known issue, so no need for a new issue for that.

packages/interaction/src/InteractionManager.js
laino: If we slice/clone it, we still need to clear the original array. That's just twice the work.

If we really care about making a code path that's not really all that hot (it's only taken on mouseout) as fast as possible, setting ``.length = 0`` is probably not the way to go about it.

If we do care about event order, then we should probably keep it the way it is. If we *don't* care about event order and don't mind going backwards through the array, then ``.pop()`` is probably the fastest way to safely go about this.



https://jsperf.com/sum-and-clear-an-array/22

laino:
I added a test case and made sure it fails on the current dev branch but works on this branch.

I also renamed ``noDelayed`` to ``skipDelayed`` and store delayed events as objects instead of arrays...

packages/interaction/src/InteractionManager.js
bigtimebuddy: Pop is good. I’m glad you’re being performance sensitive here. Interaction can be a bottleneck for some. Want to make sure we’re doing this right!

bigtimebuddy: Can you demonstrate or make a test for where order matters? I’m assuming in this case it’s not that relevant. What’s more relevant is the registered event handlers. And the eventemitter handler queue is handled separately.

bigtimebuddy: Ping @laino ☝️

laino: > Can you demonstrate or make a test for where order matters in delayedEvents?

Normally events are dispatched in the order of "children before parent elements", which in my book is a nice promise to the user and reversing the order of events for pointerout/mouseout breaks that.

In any case this codepath isn't near hot enough. That array is maybe going to contain 5 elements at most when the user actually left some element with their mouse, which maybe happens a dozen times a second tops in a very busy scene.

bigtimebuddy:
Thank you @laino :tada:

aaclayton:
Hey folks, I just updated my application from 5.1.1 to 5.1.2 to grab some of the image caching and video browser compatibility issues. Unfortunately, after updating, interactions in my canvas are not working properly. I am not getting hover events for any of my sprites, containers, etc... 

It's possible I was doing something wrong before and this change has hardened things in a way which prevents a usage mode I am currently using. I tried a simple sandbox hover and it was working fine, but this change is a bit complex for me to debug. I just wanted to raise a yellow flag to say there may be an issue here. I'll try and see if I can narrow it down with a reproducible example.

bigtimebuddy:
@laino could you please look at this?

bigtimebuddy:
I tried a few things but couldn't reproduce. @aaclayton could you make a standalone example that demonstrates this? Maybe a [pixiplayground.com](https://pixiplayground.com)

aaclayton:
Yeah, I will try - I was unable to reproduce this in a simple playground, so it definitely seems like some sort of interaction with other logic used within my application. It's just tough to find the root cause due to the complexity of the application where the issue is surfacing.

laino:
@aaclayton @bigtimebuddy I had that exact same issue today, but in the end it was just related to me having a weird comibination of packages installed after trying to bump pixi's version.

After removing any existing ``yarn link``-ed packages (specifically ``pixi.js`` and ``@pixi/interaction``), then removing my yarn.lock and doing a clean ``yarn install`` it went away...

laino:
@aaclayton Just something for consideration: because this PR changed the order in which events are dispatched, if some code in an earlier event throws an error, now later events won't trigger a mouseout event.

laino:
@aaclayton @bigtimebuddy Ah! Also if you're calling ```.stopPropagation()``` - which never really worked as advertised, but now it doesn't work as advertised in a different way.

The short story is that ``stopPropagation()`` right now prevents any further events from being dispatched. 
This means that (and this was always the case) if you're doing a ``.stopPropagation`` in a ``pointermove`` event, you won't even get an ``mousemove`` event on *the same element*, which clearly isn't what is advertised in the docs. Additionally, because now ``mouseover`` happens AFTER ``mousemove``, if you do ``.stopPropagation()`` in a ``mousemove``, you won't get a ``mouseover``.

Basically it's broken in a different way now - and we honestly can't have that. I'll provide a fix in a bit.

aaclayton:
Hey @laino and @bigtimebuddy - just following up on this. I'm going to go ahead and refactor my code to remove the `stopPropagation` calls, but I wanted to make sure this was still something being tracked for 5.1.3. Thanks!

themoonrat:
@aaclayton https://github.com/pixijs/pixi.js/pull/6043 has been approved and merged in, so just waiting for a new release :)

aaclayton:
Aah, I missed it in the list. Thank you for pointing that out and thanks very much for the quick resolution!

PR #5990, opened 2019-08-04
Improve typing for load() function

kibertoad:
### Description of change

More accurate typing is provided for load() method of Loader.

### DIFF
--- a/types/loader.d.ts
+++ b/types/loader.d.ts
@@ -19,7 +19,7 @@ declare namespace PIXI {
         use(fn: Function): this;
         reset(): this;
         //tslint:disable-next-line:ban-types forbidden-types
-        load(cb?: (loader: Loader, resources: Record<string, LoaderResource>) => void): this;
+        load(cb?: (loader: Loader, resources: Partial<Record<string, LoaderResource>>) => void): this;
 
         destroy(): void;
 
--- a/types/loader.d.ts
+++ b/types/loader.d.ts
@@ -19,7 +19,7 @@ declare namespace PIXI {
         use(fn: Function): this;
         reset(): this;
         //tslint:disable-next-line:ban-types forbidden-types
-        load(cb?: Function): this;
+        load(cb?: (loader: Loader, resource: LoaderResource) => void): this;
 
         destroy(): void;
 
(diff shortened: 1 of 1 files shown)


types/loader.d.ts
```diff
@@ -19,7 +19,7 @@ declare namespace PIXI {
         use(fn: Function): this;
         reset(): this;
         //tslint:disable-next-line:ban-types forbidden-types
-        load(cb?: Function): this;
+        load(cb?: (loader: Loader, resources: Record<string, LoaderResource>) => void): this;
```
englercj: Should probably be `Partial<Record<string, LoaderResource>>`

kibertoad: @englercj Why? Considering that Record is just a key-value map, which properties on it do you expect to be optional? Do you perhaps mean that it should be `Record<string, Partial<LoaderResource>>`?

englercj: > Why? Considering that Record is just a key-value map, which properties on it do you expect to be optional?

All of them, not every string key a user can type will actually return a `LoaderResource`. It might return `undefined` when the key doesn't exist on the resource map.

> Do you perhaps mean that it should be Record<string, Partial<LoaderResource>>?

No, I meant `Partial<Record<K, T>>`.

`Record<string, LoaderResource` means any string key I use will give me a `LoaderResource`. No change of being undefined.

`Partial<Record<string, LoaderResource>>` means I may get `undefined` or a `LoaderResource` for any string key I use.

It is the difference between

```ts
const o1: Record<string, number> = {};
const value = o1['doesnt_exist']; // is of type `number`

const o2: Partial<Record<string, number>> = {};
const value = o2['doesnt_exist']; // is of type `number | undefined`
```

The second one is correct, the string key lookup could be undefined and typescript will make me check it for existence.

kibertoad: @englercj Why not make it `Record<string, LoaderResource | undefined>` then?

englercj: Why not make it `{ [key: string]?: LoaderResource }`? Just different ways of writing the same thing.

kibertoad: Fair point :). Changed.

PR #5991, opened 2019-08-04
Graphics CalculateBounds: take matrix into account

ivanpopelyshev:
Subj.

Fix: matrix is taken into account.

Fix: lineStyle alignment affects bounds.

Breaking change: line direction is not important, it breaks the tests :(


ivanpopelyshev:
ActionScript3 Flash test:

```js
var s:Sprite = new Sprite();
var g:Graphics = s.graphics;
g.moveTo(0,0);
g.lineStyle(1, 0xffffff, 1.0, false, LineScaleMode.NORMAL, CapsStyle.NONE);
g.lineTo(10,0);
const b = s.getBounds(s);
a.equalsWithPrecision(b.width, 11.0, 'caps none line width', 1e-4);
a.equalsWithPrecision(b.height, 1.0, 'line height', 1e-4);
```

Doesn't matter whether caps is `NONE` or `SQUARE`, bounds are the same.

I think we should follow that.

eXponenta:
Report (simple visual test)

Before:
https://www.pixiplayground.com/#/edit/SlM2FU2cjDCZ7MmM0_HfJ

After:
https://www.pixiplayground.com/#/edit/Htc4He2sOSfUaSEg3bGSL

ivanpopelyshev:
One more test by @eXponenta  : https://www.pixiplayground.com/#/edit/I4lxHQxoOoivjmVySz_Fo , linestyle alignment now affects bounds.

englercj:
These fixes look good, but you mention there is a breaking change in there? Is there anyway we can have it not be a breaking change so we can release this as a patch version?

ivanpopelyshev:
I dont know how to name that thing: breaking change or behaviour change.

If I add it back we'll have huge piece of compatibility code.

Its all about line caps.

Theoretically, bounds of single lines will be changed a bit, that can affect filterArea stuff if someone filters lines, filters or culling behaviour can change in some applications.

In practice, I doubt it matters.

Also, its a preparation for adding proper line caps.

englercj:
Fair enough, sounds like the behavior change is minor enough that it likely wont affect anyone in a meaningful way.

bigtimebuddy:
Overall, I agree with the approach of this PR, however, I think this is definitely 5.2.0 and not 5.1.2. It changes how bounds are calculated for Graphics with line style. If someone is doing a pixel-perfect layout with Graphics using bounds or width/height, this will throw things off for sure. Seems irresponsible to make this a patch release.

ivanpopelyshev:
Agree. I'll add tests.

PR #5995, opened 2019-08-06
Fix the FireFox cached image issue (#5968)

### LINKED ISSUE #5968: Firefox fails to load using Texture.from() when image is cached
### Expected Behavior

Texture.from() should succeed whether loading from cache or not

### Current Behavior

Texture.from() fails when an image is cached and and HTTP 304 is received.

### Other

Noticed that when loading from cache, textures are immediately seen as valid.

I'm currently working around this by loading image urls into Image elements, and waiting for them to load before creating the PIXI textures.

### Steps to Reproduce

```
<html>
<meta charset="UTF-8">

<body>
</body>
<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/5.1.0/pixi.js"></script>
<script>
let app = new PIXI.Application({width:640,height:480});
let tex = PIXI.Texture.from("https://i.imgur.com/vt1Bu3m.jpg");
let sprite = new PIXI.Sprite(tex);
app.stage.addChild(sprite);
</script>

</html>
```

Open the following doc in firefox. When reloading, the image won't appear. If you disable cache, the image will appear on reload

### Environment

- **`pixi.js` version**: 5.1.0
- **Browser & Version**: Firefox 68.0
- **OS & Version**: Ubuntu 18.04

mofnire:
##### Description of change

This is to fix #5968. 

FireFox 68, and possibly other versions, seems like setting the `HTMLImageElement#width` and `#height`  to non-zero values before its loading completes (namely, `HTMLImageElement#complete === false`) if images are in a cache. Because of this, need to set the `_width` and the `_height` to zero to avoid uploading incomplete images.

Please refer to the following working example:
[https://www.pixiplayground.com/#/edit/XX8a7e2XBMBpDMc0gLzz_](https://www.pixiplayground.com/#/edit/XX8a7e2XBMBpDMc0gLzz_)

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/resources/ImageResource.js
+++ b/packages/core/src/textures/resources/ImageResource.js
@@ -33,6 +33,16 @@ export default class ImageResource extends BaseImageResource
 
         super(source);
 
+        // FireFox 68, and possibly other versions, seems like setting the HTMLImageElement#width and #height
+        // to non-zero values before its loading completes if images are in a cache.
+        // Because of this, need to set the `_width` and the `_height` to zero to avoid uploading incomplete images.
+        // Please refer to the issue #5968 (https://github.com/pixijs/pixi.js/issues/5968).
+        if (!source.complete && !!this._width && !!this._height)
+        {
+            this._width = 0;
+            this._height = 0;
+        }
+
         /**
          * URL of the image source
          * @member {string}


mofnire:
I found no side effects on browsers Chrome 76.0.3809.87 and Edge 42.17134.1.0..

However, on IE 11, even without this patch, I found that PixiJS very rarely fails to render the kitty image. I have no idea why. But, on IE 11, what I saw is the loaded image very rarely had a zero width / height even when the `HTMLImageElement#complete` is `true`.

https://www.pixiplayground.com/#/edit/30igkCs2t~PyoT1ObUD_3

mofnire:
I haven't tested on mobile browsers and Safari.

bigtimebuddy:
Thank you @mofnire!

PR #6018, opened 2019-08-17
Bump floss and electron

bigtimebuddy:
This fixes Travis CI that's failing with the old version of Electron. Small tweak to Floss and bumping Electron to v5 fixes the issue. Trying to unblock v4.8.9 release.


nathanielmolinar1998:
Send aome coins

Issue #5996, opened 2019-08-06
Videos not working anymore on Chrome v76


borradaniele:
Videos are not working anymore just after updating Google Chrome to v76.
Loading the video (with the app loader) and creating a Sprite using the video as texture does not work anymore. Nothing has been changed in the site source code, I've only updated Chrome. Tested the same site from the same local development server on an older version of Chrome and everything is fine. The error in console is the following

```
WebGL: INVALID_VALUE: tex(Sub)Image2D: video visible size is empty
[.WebGL-0x7f861a496800]GL ERROR :GL_INVALID_OPERATION : glTexSubImage2D: level 0 does not exist
```
### Steps to Reproduce
I get the same error on the PIXI official site in the video demo page. And the video is not playing even there.
https://pixijs.io/examples/#/sprite/video.js

### Environment
Mac OS High Sierra 10.13.6
Chrome Version 76.0.3809.87 (Official Build) (64-bit)
PIXI v5.1.1

drpepper:
I have the same problem. I'm also on Mac Mohave 10.14.5, Chrome Version 76.0.3809.87 (Build officiel) (64 bits), PixiJS 5.0.2.

nicholas-slepchenko:
Same problem here
Windows 10
Chrome Version 76.0.3809.87 (Official Build) (64-bit)

eXponenta:
@drpepper @nicholas-slepchenko @borradaniele 
This is problem only with Chrome 76 and later.
Temporary fix:
https://www.pixiplayground.com/#/edit/F1nTXxdf9xfBQt6uTurAM

drpepper:
Thanks @eXponenta !

If understand your fix correctly, the trick is to wait for the promise from the video element's play() method resolve before creating the sprite. I just tried this and it worked for Chrome. However, now Firefox gives me an error and refuses to play the video:

`Error: WebGL warning: drawElements: TEXTURE_2D at unit 1 is incomplete: The dimensions of `level_base` are not all positive.`

eXponenta:
@drpepper Can you create fiddle for this? Because my fix works correct on Firefox.
You must memento what video can't be played before user interaction.

borradaniele:
@eXponenta Thanks for the quick reply and the temp fix, I'm going to implement it in the upcoming days.

goswamibodhgaya:
I need work

drpepper:
I haven't had the time to make a fiddle, sorry.

[...]

PR #6023, opened 2019-08-19
workaround for the issue that videos not working issue on Chrome v76

meowtec:
Ref: #5996

##### Pre-Merge Checklist

just set `preload` with some value (`'auto'`) when it be created.

I don't know the reason, but it just works.

And those also work::
```
preload='' // equal to `preload='auto'`
// or
preload='none'
```

### DIFF
--- a/packages/core/src/textures/resources/VideoResource.js
+++ b/packages/core/src/textures/resources/VideoResource.js
@@ -24,6 +24,8 @@ export default class VideoResource extends BaseImageResource
         {
             const videoElement = document.createElement('video');
 
+            // workaround for https://github.com/pixijs/pixi.js/issues/5996
+            videoElement.setAttribute('preload', 'auto');
             videoElement.setAttribute('webkit-playsinline', '');
             videoElement.setAttribute('playsinline', '');
 


meowtec:
No, it is not the solution

themoonrat:
According to the spec 

```
preload
This enumerated attribute is intended to provide a hint to the browser about what the author thinks will lead to the best user experience with regards to what content is loaded before the video is played. It may have one of the following values:

- none: Indicates that the video should not be preloaded.
- metadata: Indicates that only video metadata (e.g. length) is fetched.
- auto: Indicates that the whole video file can be downloaded, even if the user is not expected to use it.
- empty string: Synonym of the auto value.
```

Does that mean we should put 'auto' here instead of an empty string, to be more explicit? And check that it still fixes things

meowtec:
@themoonrat Yes it is.  `''` Has been changed to `'auto'`

BH-NOTHING:
@meowtec  Preload works on pc but not work on mobile device, is there a solution?

robert-gogolan:
I tried to implement the same workaround in the pixi 4.8.8 version and it does not work.
I added the preload attribute on the video in VideoBaseTexture.

meowtec:
@robert-gogolan You should add `preload` immediately after  `document.createElement('video')`.

It would be too late if your code was like this:
```
const video = Sprite.from('VIDEO_URL')
video.texture.baseTexture.media.preload = 'auto'
```

This works:
```
const video =  document.createElement('video')
video.crossOrigin = 'anonymous'
video.preload = 'auto'
video.src = 'VIDEO_URL'
const sprite = Sprite.from(video)
app.stage.addChild(sprite)
```

PR #6024, opened 2019-08-19
Made pixi 5 conversion of scissor mask implementation.

J00nz:
##### Pre-Merge Checklist


GoodBoyDigital:
You rock! thanks for this PR! Do you have an example of this working with scissor rect we I can check out? Cheers!

packages/core/test/Renderer.js
```diff
@@ -82,4 +82,57 @@ describe('PIXI.Renderer', function ()
...
+            this.renderer.destroy();
+            this.renderer = null;
+        });
+
+        it('should have scissor-masks enabled', function ()
+        {
+            expect(this.renderer.mask.enableScissor).to.equal(true);
+        });
+
+        it('should use scissor masks whith axis aligned squares', function ()
```
rin3s:
```suggestion
        it('should use scissor masks with axis aligned squares', function ()
```

bigtimebuddy:
ping @J00nz, could you please provide a live example of this feature working?

ivanpopelyshev:
I will deal with that as soon as I'm available :)

GoodBoyDigital:
scratch that, noticed that `enableScissor` was false :D

packages/core/src/mask/MaskSystem.js
```diff
@@ -80,7 +80,7 @@ export default class MaskSystem extends System
         }
         else if (this.enableScissor
             && !this.scissor
-            && this.renderer._activeRenderTarget.root
+            // && this.renderer._activeRenderTarget.root
```
GoodBoyDigital: we should replace this line with: 
```&& !this.renderer.renderTexture.current```
this makes sure we only use the stencil mask if we are on the root.

J00nz: 👍

PR #6034, opened 2019-08-22
Adds a strict mode for using texture caches

### LINKED ISSUE #6032: Slient errors when using the new Sprite.from in v5
Previously (in Pixi v4) when you tried to create a Sprite using Sprite.fromFrame('textureId'), and textureId didn't exist in the texture cache, it would throw an error. 

In the new Pixi (5.1.1), using the Sprite.from('textureId') method, it now doesn't throw any error and just displays an empty Sprite. This is a problem for me when re-skinning a game, and need to remove unused textures etc. Is there a way to turn on a warning / error in this case?

bigtimebuddy:
##### Description of change

Fixes #6032 

Introduces two new APIs:
* Adds `PIXI.settings.STRICT_TEXTURE_CACHE` (default `false`), by enabling this errors will be thrown if `Texture.from` provides a string that doesn't exist in TextureCache. Also works for BaseTexture.
* Adds an optional third argument for `BaseTexture.from` and `Texture.from` to enable this strict property on a per-use basis.

##### Pre-Merge Checklist

**Documentation**

http://pixijs.download/dev-strict-cache-setting/docs/PIXI.settings.html#.STRICT_TEXTURE_CACHE


eXponenta:
There is little mistake. It's throw `Uncaught TypeError: Cannot read property '_pixiId' of undefined` instead of warning and EMPTY Texture when source is undefined.

https://github.com/pixijs/pixi.js/blob/0c779d8fc801b60cc56863c33523f33541277945/packages/core/src/textures/Texture.js#L340

packages/core/src/textures/Texture.js
```diff
@@ -324,9 +324,10 @@ export default class Texture extends EventEmitter
      * @param {number|string|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement|PIXI.BaseTexture} source
```
eXponenta: What is number must be passed as source?

ivanpopelyshev:
Yes, `number` is processed between beginning of the function and `if (strict)`, i think we need to clarify this

packages/core/src/textures/Texture.js
```diff
@@ -346,6 +347,12 @@ export default class Texture extends EventEmitter
 
         let texture = TextureCache[cacheId];
 
+        // Strict-mode rejects invalid cacheIds
+        if (cacheId && strict && !texture)
```
eXponenta: `cacheId` getted from `source.__pixiId` but on first usage is will be undefined and assigned from `uid()` and was't exist in TextureCache

packages/core/src/textures/BaseTexture.js
```diff
@@ -552,6 +553,12 @@ export default class BaseTexture extends EventEmitter
 
         let baseTexture = BaseTextureCache[cacheId];
```
eXponenta: Same mistake as in Texture. New uid wasn't exist in BaseTextureCache.

bigtimebuddy:
@eXponenta Not sure I understand how you got that. Can you give me some code to reproduces?

packages/core/src/textures/Texture.js
bigtimebuddy: Oops, you're right. Will fix.

ivanpopelyshev:
That's better, now the problem is with `number` case, we need to clarify what it means and what its behaviour, and whether its affected by strict.

eXponenta:
So, i did mistake and use documented number ID, both throw exceptions 
https://www.pixiplayground.com/#/edit/wqZ9UBj_rQ9D1EH5hFzK4

ivanpopelyshev:
`PIXI.Texture.from(0);` produces this:

```
Texture.js:341 Uncaught TypeError: Cannot create property '_pixiId' on number '0'
```

eXponenta:
@bigtimebuddy , according to the documentation of pixi 4.8.8 it should be frame ID, but surprise!,
It doesn't work in it also.
Maybe remove them? 
**number** as key doesn't work since pixi v4.8 or less. 
I think it’s just not necessary.


http://pixijs.download/v4.8.8/docs/core_textures_Texture.js.html#line388

bigtimebuddy:
I removed it. `BaseTexture`'s `cacheId` is a string, `addToCache` only accepts a string. Numbers are not supported and should never have worked.

PR #6039, opened 2019-08-23
WIP: Implement pointerover/out for touch and pen events

laino:
##### Description of change

This adds "pointover" and "pointerout" events while you are "dragging" a touch over your scene, especially useful when there's a lot of clickable stuff and you want to give your user a "mouse" like indication of what element he is currently on.

But because a video says more than a million(?) words: https://streamable.com/c6pn9

In that video you can see me "touching and dragging" over the wolves, until I've found the right one, then releasing. I should add that I own none of the placeholder art in that video.

These changes are just an early POC and there's probably edge cases I need to take care of and also failing tests that I need to have a look at. But I wanted to get some discussion going before I invest any more time.

##### Pre-Merge Checklist


laino:
I fixed the failing tests by actually tracking whether the user touched "down" on the current element (which used a more primitive logic before that was fine as along as "moving" didn't create tracking data).

I also changed pointer "up" logic so that more specialized events run before the general "pointerup" stuff, which was reversed for touch events.

laino:
Okay. Since the old tests passed I'm gonna start changing them up a bit now.

There are some similar tests that are done once for mouse, 'pointer' and touch events, however the touch event tests usually kind of cheated and didn't test the "pointerover" event, because it didn't exist.

There are a lot of tests that are fundamentally the same and could be cleaned up by creating generic test function with slightly different input. Are you guys a fan of that or would you prefer to have those tests more copy-paste-like?

I will edit this post and document any changes to existing tests here:

### Move to other sprite, then no "tap" tests

#### with *mouse* events, should not call handler when moved to other sprite

Tests for ``over``, ``up``, then no ``tap``.

#### with *pointer* events, should not call handler when moved to other sprite

Tests for ``over``, ``up``, then no ``tap``.

This didn't actually set the ``ensurePointerEvents`` on the MockPointer.
I'll make sure all the pointer tests actually do this...

### with *touch* events, should not call handler when moved to other sprite

Tests for ``move``!, ``up``, then no ``tap``.

### Changes
The touch test will be changed to actually test the same thing as the two tests above, instead of cheating on the ``over`` part. Also ``ensurePointerEvents`` will be set.

laino:
I just noticed that the ```afterEach``` in the InteractionManager tests never worked. If it did, it would have crashed. It's ```pointer.cleanup()``` not ``pointer.cleanUp()`` ...

laino:
The ``click`` function in ``MockPointer.js`` passed the ``asPointer`` argument to ``mouseup`` as ``onCanvas``.

We're kind of "unlucky" here that the tests still worked all this time...

laino:
Turns out "pens" didn't have over/out events either, though this is probably more of a bug than a missing feature.

laino:
Question about the ``mouseOverRenderer``-thing. Why does this exist? And is it supposed to be unset for pens?

laino:
So I made sure pointerover/out is also emitted for pens. Additionally I re-ordered some events so that
specific events (``mousedown`` etc.) are always emitted *before* generic ones (``pointerdown`` etc.)

Additionally I made some tests much more in-depth, and also made them run in all possible input configurations:

packages/interaction/test/InteractionManager.js
```diff
@@ -18,6 +18,7 @@ describe('PIXI.interaction.InteractionManager', function ()
 {
     afterEach(function ()
     {
+        // TODO: This doesn't work....
```
laino: This never did work...

packages/interaction/src/InteractionManager.js
```diff
@@ -1438,34 +1453,57 @@ export default class InteractionManager extends EventEmitter
...
         {
             this.dispatchEvent(displayObject, 'pointerup', interactionEvent);
-            if (isTouch) this.dispatchEvent(displayObject, 'touchend', interactionEvent);
 
-            if (trackingData)
+            // emit pointertap if the mouse/touch block decided it was a tap
+            if (isTap)
             {
-                // emit pointertap if not a mouse, or if the mouse block decided it was a tap
-                if (!isMouse || isMouseTap)
```
andrewstart: I don't think this old code was the cleanest way to go about it, but this `!isMouse` -> emit pointertap did mean that a pointer that wasn't a mouse/pen or a touch did emit pointertap appropriately. It looks like the new code won't do that in the off chance that a pointer device that is none of those three is used. Added another `else` after the `if (isMouse) {} else if (isTouch) {}` that just checks that case and sets `isTap` appropriately should solve it.

packages/interaction/src/InteractionTrackingData.js
```diff
@@ -121,11 +121,28 @@ export default class InteractionTrackingData
...
+        this._doSet(this.constructor.FLAGS.TOUCH_DOWN, yn);
+    }
 }
 
 InteractionTrackingData.FLAGS = Object.freeze({
     NONE: 0,
     OVER: 1 << 0,
     LEFT_DOWN: 1 << 1,
     RIGHT_DOWN: 1 << 2,
+    TOUCH_DOWN: 1 << 3,
```
andrewstart: Because the instance of the IntractionTrackingData is attached to an ID for a touch, I think that we should be able to stick with `LEFT_DOWN`. If you feel that is confusing in the code, perhaps adding `primaryDown` as an alias for `leftDown` (which makes more sense for pen friendly code anyway)?

andrewstart:
> 
> 
> Question about the `mouseOverRenderer`-thing. Why does this exist? And is it supposed to be unset for pens? All it does as far as I can tell is cause hit tests for mouse events to be ignored when it is false. It is the reason we don't have `over/out` events for pens, because for pens it is always false.

It is an optimization to avoid hit testing when we know we won't hit anything. Changing it to work with pens is perfectly fine - I don't think we have enough usage that isn't just mouse and/or touch to notice these things and complain, so we miss stuff like that.

ivanpopelyshev:
@laino we need you for last push to 5.2.0!

laino:
I don't know when I can get back to this. I have quite a lot of other things on my plate right now.

ivanpopelyshev:
Don't worry, one of us will adopt it when we have some time :)

PR #6041, opened 2019-08-25
Fix Texture prepare

eXponenta:
##### Description of change

Fixing some mistakes on Texture preparation.

When you add `PIXI.Texture` in prepare plugin and then call `upload`, Texture will not be uploaded but callback will be called immediate.

Bug rep:
https://www.pixiplayground.com/#/edit/o4Wyl5rtZDVdMxfN9kglK

Fixed:
https://www.pixiplayground.com/#/edit/aIRNjeeax2Aj83HwPix5a

##### Pre-Merge Checklist


bigtimebuddy:
@eXponenta 

I still don't get this PR. Your example of the "bug" doesn't work. Can you please fix or update that?

eXponenta:
Yep, we need playground with read-only links, because someone edit example.

PR #6043, opened 2019-08-26
InteractionManager: fix event.stopPropagation()

laino:
This fixes ``event.stopPropagation()`` - which didn't ever really just stop propagation, but rather
stopped events dead in their tracks.

Basically when you called ``stopPropagation`` there would be no further events emitted for the same action. So if you did it in, say, a ``mouseup`` event, you wouldn't even get a ``click`` event. Bummer.

This becomes really apparent with the recent [mouseover/pointerover](https://github.com/pixijs/pixi.js/pull/5989) changes, because if you now call ``stopPropagation`` in a ``mousemove`` event, you wouldn't ever see any ``mouseover`` event (which *would* be dispatched after).

Previously there were no tests for event propagation, so I added three of those.

##### Pre-Merge Checklist

### DIFF
--- a/packages/interaction/src/InteractionManager.js
+++ b/packages/interaction/src/InteractionManager.js
@@ -1172,7 +1172,7 @@ export default class InteractionManager extends EventEmitter
                 const { displayObject, eventString, eventData } = delayedEvents[i];
 
                 // When we reach the object we wanted to stop propagating at,
-                // reset the propagation hint.
+                // set the propagation hint.
                 if (eventData.stopsPropagatingAt === displayObject)
                 {
                     eventData.stopPropagationHint = true;
--- a/packages/interaction/src/InteractionManager.js
+++ b/packages/interaction/src/InteractionManager.js
@@ -939,7 +939,9 @@ export default class InteractionManager extends EventEmitter
      */
     dispatchEvent(displayObject, eventString, eventData)
     {
-        if (!eventData.stopped)
+        // Even the event was stopped, at least dispatch any remaining events
+        // for the same display object.
+        if (!eventData.stopPropagationHint || displayObject === eventData.stopsPropagatingAt)
         {
             eventData.currentTarget = displayObject;
             eventData.type = eventString;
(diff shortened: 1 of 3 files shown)


laino:
I should add that this **only** does the following things:

- Make ``stopPropagation`` not kill following events for the *same* object
- Fix some additional weirdness introduced by the delayed events change

This **still** doesn't make events work the way DOM events work. That would be a lot
more work and probably is out of scope for a simple bugfix PR.

packages/interaction/src/InteractionManager.js
```diff
@@ -939,7 +939,9 @@ export default class InteractionManager extends EventEmitter
      */
     dispatchEvent(displayObject, eventString, eventData)
     {
-        if (!eventData.stopped)
+        // Even if the event was stopped, at least dispatch any remaining events
+        // for the same display object.
+        if (!eventData.stopPropagationHint || displayObject === eventData.stopsPropagatingAt)
```
laino: This is where stuff went wrong. Because we just checked ``event.stopped`` here, a following event on the same object would also get killed. So now we also check whether the current object is still the same we wanted to stop propagation at.

Additionally we now use a separate ``stopPropagationHint``. This is because with delayed events we have to start over deeper in the tree, which is why we set it to false and then back to tru once we reach ``stopPropagationAt``.
This allows us to keep ``stopped`` clean and behaving the way users are used to.

laino:
Pinging @bigtimebuddy You might want to release a hotfix for this.

bigtimebuddy:
Could you switch the branch from dev to master, since likely this will be a hot fix as 5.1.3?

laino:
@bigtimebuddy 
>     1. Is it possible to make the jsdocs for `stopPropagationHint`, and `stopsPropagatingAt` private? Would like to not expand the API unless it's absolutely necessary.
> 
>     2. Could you rebase this against master? Right now dev is 5.2.0 and would like to make this a hotfix.

Done and done.

bigtimebuddy:
Thank you @laino and for everyone else that reviewed this.

Issue #6011, opened 2019-08-13
Mask disappears when filter is enabled.


Metatron92:
### Expected Behavior
Cannot use mask + filter, this leads to disappears mask from object.

### Environment

- **`pixi.js` version**: _e.g. 5.1.1_ (Tested on all v5 PIXI)
- **Browser & Version**: _e.g. Chrome 76_
- **OS & Version**: _e.g. Windows 10_
- **Running Example**: https://www.pixiplayground.com/#/edit/C15T9BCkrhcwTiQIxRbfN

ivanpopelyshev:
That's a chromium bug in windows. I tested everything starting from chromium-74.

Android OK, Linux OK, other browsers are OK. 

Better demo: https://www.pixiplayground.com/#/edit/DUFi3RZRLH091vaFeN~xU

FAIL win chrome: 
![image](https://user-images.githubusercontent.com/695831/62951455-b81aa500-bdf2-11e9-9276-98cdfce9b30d.png)

OK firefox: https://service.crazypanda.ru/v/monosnap/2019-08-13-16-33-29-l7vl8.png
![image](https://user-images.githubusercontent.com/695831/62951463-bc46c280-bdf2-11e9-9964-67b1d5a2b25e.png)

fserb:
Seems to work fine with Chrome 76 on Mac.

fserb:
Also works fine on Linux.

eXponenta:
Bug reproduce only on WebGL2.
WebGL1 works fine on W10.

ivanpopelyshev:
Workaround:

```js
PIXI.settings.PREFER_ENV = PIXI.ENV.WEBGL;
```

djmisterjon:
work on my side :)
chromium 75 on win
![image](https://user-images.githubusercontent.com/24865815/63036634-6a9e4680-be8b-11e9-8b97-648a8959f473.png)

aaclayton:
I was struggling with this bug and had creaded the following playground fiddle to illustrate the problem (hopefully clearly): https://www.pixiplayground.com/#/edit/1K8UnXqIZ30KQlBUairvq

It seems this issue is well understood now though, and I can confirm the workaround posted by @ivanpopelyshev works on my end. Thank you for sharing that promptly.

ivanpopelyshev:
@aaclayton yep, but im so lazy i didnt report it to chromium yet. gonna make it tomorrow

PR #6044, opened 2019-08-26
Workaround chromium webgl2 renderbuffer bug

ivanpopelyshev:
I've reported the bug: https://bugs.chromium.org/p/chromium/issues/detail?id=997702

Thanks @Metatron92 for #6011


ivanpopelyshev:
Done.

PR #6061, opened 2019-08-31
Fix broken rendering of graphics by custom graphics plugin when is can't batched.

eXponenta:
Problem:
https://www.pixiplayground.com/#/edit/wDkRfBh_mws1~L3IZkKC6

Or heavy example :
https://www.pixiplayground.com/#/edit/6Q2sohQJ3Pomw_5reM_VG

Scroll down and uncomment `drawRoundedRect`.

Fixed:
https://www.pixiplayground.com/#/edit/SF_M5l0B3QB0pk7hR3Ms8

##### Pre-Merge Checklist

### DIFF
--- a/packages/graphics/src/Graphics.js
+++ b/packages/graphics/src/Graphics.js
@@ -829,145 +829,198 @@ export default class Graphics extends Container
         {
             if (this.batchDirty !== geometry.batchDirty)
             {
-                this.batches = [];
-                this.batchTint = -1;
-                this._transformID = -1;
-                this.batchDirty = geometry.batchDirty;
-
-                this.vertexData = new Float32Array(geometry.points);
-
-                const blendMode = this.blendMode;
-
-                for (let i = 0; i < geometry.batches.length; i++)
-                {
-                    const gI = geometry.batches[i];
-
-                    const color = gI.style.color;
-
-                    //        + (alpha * 255 << 24);
-
-                    const vertexData = new Float32Array(this.vertexData.buffer,
-                        gI.attribStart * 4 * 2,
-                        gI.attribSize * 2);
-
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Could you rebase to master? I think belongs in 5.1.3 instead of 5.2.0

eXponenta:
> Could you rebase to master? I think belongs in 5.1.3 instead of 5.2.0
Ok, i rebased it.

bigtimebuddy:
You need to rebase again. That didn’t work because you picked up all the changes in dev.

eXponenta:
Ups. Rebase again.
Codecov can't update coverage info after force push.

packages/graphics/src/Graphics.js
```diff
@@ -829,143 +829,198 @@ export default class Graphics extends Container
...
+            // but may be more than one plugins for graphics
+            if (!DEFAULT_SHADERS[pluginName])
             {
-                const drawCall = geometry.drawCalls[i];
-
-                const groupTextureCount = drawCall.textureCount;
+                const sampleValues = new Int32Array(16);
 
-                for (let j = 0; j < groupTextureCount; j++)
+                for (let i = 0; i < 16; i++)
```
dumganhar, line 1004: Hardcode 16 here will break some Adreno 3xx GPU devices.
Because sampler array may be less than 16, but you set a Int32Array with 16 length to the sampler array uniform  which will cause GPU driver crash.

SukantPal: Probably better to open an issue now.

bigtimebuddy: PR welcome to fix.

dumganhar: Here it is: https://github.com/pixijs/pixi.js/pull/6876

PR #6065, opened 2019-09-03
Fix sprite's uvs updating

eXponenta:
##### Description of change

Bug:
https://www.pixiplayground.com/#/edit/nygKQ4JiT675hWX35JN8k

Resume:
When `baseTexure` or `frame` was be reassigned, uvs not changes on sprites with this base texture.

Case 1:
It corrupts sprite when basetextures is valid  but dimensions isn't equals.

Case 2:
Renderer throws 'undefined' exception about uvs when you chage from Texture.EMPTY (`valid = false`) to `valid = true` texture;

Fixed:
https://www.pixiplayground.com/#/edit/gcxJ1nKj2GUp5H~AptY37

##### Pre-Merge Checklist

### DIFF
--- a/packages/sprite/src/Sprite.js
+++ b/packages/sprite/src/Sprite.js
@@ -232,6 +232,7 @@ export default class Sprite extends Container
             return;
         }
 
+        // update texture UV here, because base texture can be changed without calling `_onTextureUpdate`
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Yep, that's a critical. Lets put it in 5.1.3

PR #6077, opened 2019-09-06
Refact graphics

eXponenta:
##### Description of change

Current Graphics implementation doesn't support expandability as fact.
`GraphicsGeometry` and `Geometry` can't be extending from similar class because imports from internal utils is used, outside of PIXI namespaces.
Next problem - very heavy methods. They cannot be overridden without code duplication.

Looks this:
https://github.com/pixijs/pixi.js/blob/dev/packages/graphics/src/GraphicsGeometry.js#L549
https://github.com/pixijs/pixi.js/blob/dev/packages/graphics/src/GraphicsGeometry.js#L475
https://github.com/pixijs/pixi.js/blob/dev/packages/graphics/src/GraphicsGeometry.js#L25

Example (cannot be created without this refract) :
https://www.pixiplayground.com/#/edit/NWa_z1hMuF92Dwmyb3AEl

##### Pre-Merge Checklist


GoodBoyDigital:
hey @eXponenta, just so you know, this is on my list to review. Will get round to it as soon as I can 👍

packages/graphics/src/GraphicsGeometry.js
```diff
@@ -568,41 +492,121 @@ export class GraphicsGeometry extends BatchGeometry
...
+     */
+    validateBathching()
+    {
+        if (this.dirty === this.cacheDirty)
+        {
+            return false;
+        }
+
+        if (!this.graphicsData.length)
+        {
```
GoodBoyDigital: just wandering if this function should be modifying batchable (`this.batchable = true;`) ? feels like it should not mutate the state?

also may be . good time to remove all the extra ifs!

```
validateBathching()
{
    if (this.dirty !== this.cacheDirty)
    {
        if (!this.graphicsData.length)
        {
            this.batchable = true;

            return false;
        }

        for (let i = 0; i < this.graphicsData.length; i++)
        {
            const data = this.graphicsData[i];
            const fill = data.fillStyle;
            const line = data.lineStyle;

            if (fill && !fill.texture.baseTexture.valid) return false;
            if (line && !line.texture.baseTexture.valid) return false;
        }

        return true;
    }

    return false;
}
```

ivanpopelyshev: I placed `this.batchable=true` there in one of PR's, I think its actually fixes a bug. You think we can move it out `validateBatching`?

ivanpopelyshev: I propose we dont try to remove that line for now , address it in another PR with a test.

GoodBoyDigital: Happy to wait for another separate PR on top of this one. Then we can merge to dev together. But I think we may as well address this now. We could move

```
if (!this.graphicsData.length)
        {
            this.batchable = true;

            return false;
        }
```

into `updateBatches` just before the line where we call `validateBatching`?

eXponenta: Done

ivanpopelyshev:
Ideally, yes, but we dont have visual tests in this repo.

eXponenta:
So, VT is passed normal.
http://pixijs.io/v5-render-tests/?refact-graphics-geometry

bigtimebuddy:
I’m actually not talking just about rendering. Exercising the APIs is a good baseline. Just the basics. Part of this refactor involves creating new protected methods and making it easier to extend. Would love to see tests for that.

PR #6079, opened 2019-09-07
Typescript tools

Zyie:
##### Description of change
 
As requested by @bigtimebuddy I've added some of the basics to get the typescript conversion started.

- Added `.tsconfig`
- Updated eslint to use the recommended rules for `.ts` files
- Added typescript rollup plugin

##### Pre-Merge Checklist


.eslintrc.json
```diff
@@ -245,5 +247,11 @@
         "sort-imports": 0,
         "template-curly-spacing": [1, "never"],
         "yield-star-spacing": [1, "before"]
-    }
+    },
+    "overrides": [
+        {
+          "files": ["*.ts", "*.tsx", ".*.ts", ".*.tsx"],
```
bigtimebuddy: Looks like some duplicates here? Also, we can probably remove tsx, since we are highly unlikely to add any JSX to this codebase.

.tsconfig
```diff
@@ -0,0 +1,18 @@
+
+{
+"compilerOptions": {
```
bigtimebuddy: Leading whitespace?

.eslintrc.json
Zyie: ok ill remove it

.tsconfig
Zyie: oops. Didnt see the extra line

eXponenta:
So, we can start rewrite PIXI on TS from him branch?

bigtimebuddy:
I want to do this in phases and smaller manageable PRs which is why this only adds the toolchain.

clark-stevenson:
Oh Wow! 😃

PR #6097, opened 2019-09-15
Allows typescript files to be used with jsdocs

Zyie:
##### Description of change

This PR is following up on #6079 to get jsDocs to work with typescript files.

For creating the docs/types the project is compiled to es6 by running `lerna run compile:docs` with the output being sent to a temp folder that recreates the project hierarchy. e.g. `packages/core/src/`.

jsDocs then uses this folder to create the docs/types. It is removed once complete.

The only differences I've noticed in the docs is that the white space is not always preserved when comparing the original and compiled source files. 

If anyone knows a better approach to doing this ill happily give it another try.

##### Pre-Merge Checklist


englercj:
Is there a reason we wouldn't just use typedoc or similar?

bigtimebuddy:
I’d like to clarify the intention of this PR. I asked @Zyie to work on this. The goal here is to provide an intermediary step to support docs/types while we convert to TypeScript. Converting to another tool like Typedocs maybe where we end up, but I didn’t want to front-load a lot of trivial doc changes. Would rather do the TypeScript code first. This PR is a bridge but allows us to move forward with that without solving all of that now.

bigtimebuddy:
My only criticism about this right now is that it involves a lot of identical changes to every package. Is there a way to do the tsconfig and script stuff without maintaining so much? Even if that involves copying/removing the tsconfig or making a custom script at the root level.

ivanpopelyshev:
Good idea.

Zyie:
@bigtimebuddy I dont know how i missed this but we can just run `tsc` on the root project and get the same result.

So I've updated the PR and now the packages are not changed

bigtimebuddy:
Great work @Zyie :tada:

PR #6099, opened 2019-09-16
Optional renderTexture stencil

ivanpopelyshev:
That's for 5.1.4

First part of renderTexture stencil optimization: reduce the memory of renderTexture in case there are no masks. Comparing to v4 that was a downgrade.

Next PR's will add temporary pow2 renderbuffers for filters, and add a setting to allow run pixi in stencil-less webgl.

Open chrome or canary, single page, press SHIFT+ESC, look in GPU process memory

1000 renderTextures of size 256x256. 

In windows stencil+depth is 32bit, and STENCIL8 cant help, it always creates both. +33% tax on potential mipmaps for colorbuffer.

That's 350kb colorbuffer per one renderTexture +256kb stencil/depth. 

That means 350MB vs 600MB if we use stencil for every texture.

My measurements:

Empty playground, 140M total.
image

Stencil, 800M GPU total in chrome canary: https://www.pixiplayground.com/#/edit/8vkPJSK3IBCisZYdc6YC1
image

This branch, no stencil, 550M 
https://www.pixiplayground.com/#/edit/3Bxzpc~OW18PHxsJ3wxcc
image

I dont know where else 60M goes, but I think we can ignore that.

Proof that `forceStencil()` code works, filter on `app.stage` : https://www.pixiplayground.com/#/edit/aN_YFvmR5W6UmSl6Han6X

### DIFF
--- a/packages/core/src/framebuffer/FramebufferSystem.js
+++ b/packages/core/src/framebuffer/FramebufferSystem.js
@@ -371,7 +371,6 @@ export default class FramebufferSystem extends System
             { // you can't have both, so one should take priority if enabled
                 gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.DEPTH_STENCIL_ATTACHMENT, gl.RENDERBUFFER, fbo.stencil);
             }
-            // fbo.enableStencil();
         }
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/framebuffer/FramebufferSystem.js
```diff
@@ -427,6 +426,36 @@ export default class FramebufferSystem extends System
         }
     }
 
+    forceStencil()
```
bigtimebuddy: You could add some documentation, even if private? Please explain why and how it's used (not what it is)

ivanpopelyshev: yep

PR #6100, opened 2019-09-16
Introduce ALPHA_MODES instead of not obvious boolean values

ivanpopelyshev:
We actually have 3 behaviours about premultiplied alpha:

0. nothing
1. do pma on unpack (default)
2. its already pma, dont do it second time!

**UPD: New property name is `alphaMode`** 

Also, now user can specify premultiplyAlpha in baseTexture and ImageResource wont override it by default.

That's the first time I see its not a boolean in html5 renderers, please tell me if you saw something like that somewhere else, maybe its our invention.

Why not just two booleans? Well, I think that `premultipliedAlpha` vs `premultiplyAlpha` have not obvious distinction. If we use `before` and `after` people also wont understand that. Its better to put it all in single value.

Demo 1, compressed textures and png: https://www.pixiplayground.com/#/edit/Y27L_WR7v_dR26sk6dRcc

Demo 2, spine pma atlas:
https://www.pixiplayground.com/#/edit/aqf_N4ZXwQq_9lHhY7i4F

Left is fixed, right is not, look at the character mouth.

Reaction of Phaser3 author:

image

Reaction of spine author:

image


packages/core/src/textures/BaseTexture.js
```diff
@@ -136,12 +136,12 @@ export class BaseTexture extends EventEmitter
...
-         * Set to true to enable pre-multiplied alpha
+         * How to treat premultiplied alpha, see {@link PIXI.PMA_MODES}.
          *
-         * @member {boolean}
-         * @default true
+         * @member {PIXI.PMA_MODES}
+         * @default PIXI.PMA_MODES.DO_UNPACK
          */
-        this.premultiplyAlpha = premultiplyAlpha !== false;
+        this.premultiplyAlpha = premultiplyAlpha !== undefined ? Number(premultiplyAlpha) : PMA_MODES.DO_UNPACK;
```
bigtimebuddy: We could be doing a better job here of signaling to users to convert their code to use PMA_MODES instead of using the boolean.
```suggestion
        if (typeof premultiplyAlpha === 'boolean')
        {
            deprecation('v5.2.0', 'premultiplyAlpha is not a boolean anymore, see PIXI.PMA_MODES`);
            premultiplyAlpha = Number(premultiplyAlpha);
        }
        this.premultiplyAlpha = premultiplyAlpha !== undefined ? premultiplyAlpha : PMA_MODES.DO_UNPACK;
```

bigtimebuddy: Don't merge my suggestion, you'll probably need to import `deprecation` from `@pixi/utils` and make a better message

ivanpopelyshev: Actually we have even bigger problem. I posted solutions like `baseTexture.premultiplyAlpha=false` several times. and now it wont work because its a field, we can detect that change and all `===` statements will ignore that. Any ideas?

ivanpopelyshev: I agree on deprecations.

ivanpopelyshev: I wont merge this of course, because you have even more errors there :) like two different quotation symbols

ivanpopelyshev:
@bigtimebuddy I've added {number} as an option in two functions, and in one instance i'm using `> 0`. What do you think?

Also, still that problem: if user specified `baseTexture.premultiplyAlpha=true` he'll have a problem without deprecation warning. Of course that kind of users are supposed to know that kind of things and look at them if they see strange behaviour..

ivanpopelyshev:
One more idea is to put that `unpack` check inside a function in sandbox - maybe in Resource. But I'm against it because such function calls should be visible in `upload` directly. People have to understand the magic that is going on, and that's good learning moment.

packages/core/src/textures/BaseTexture.js
ivanpopelyshev: yeah, we have other modes. I agree, we need to rename it.

packages/core/src/textures/BaseTexture.js
```diff
@@ -141,12 +141,13 @@ export class BaseTexture extends EventEmitter
...
          */
-        this.premultiplyAlpha = premultiplyAlpha !== undefined ? premultiplyAlpha : PMA_MODES.UNPACK;
+        this.pmaMode = pmaMode !== undefined ? pmaMode : PMA_MODES.UNPACK;
 
-        if (typeof premultiplyAlpha === 'boolean')
+        if (options.premultiplyAlpha !== undefined)
         {
-            deprecation('v5.2.0', 'premultiplyAlpha is not a boolean anymore, see PIXI.PMA_MODES');
-            this.premultiplyAlpha = Number(premultiplyAlpha);
+            deprecation('5.2.0', 'PIXI.BaseTexture.premultiplyAlpha property has been changed to `pmaMode`'
```
bigtimebuddy: I don’t think you need this double deprecation. It’s already in the bundle, just set this.premultiplyAlpha and be done.

ivanpopelyshev: done

bigtimebuddy:
I like @GoodBoyDigital suggestion for the naming. Might also add:
* `pmaMode` => `alphaMode`
* `PMA_MODES` => `ALPHA_MODES`

ivanpopelyshev:
hm, If we use "ALPHA_MODES" we can add even more! ALPHA_ONLY and NO_ALPHA .. yep

ivanpopelyshev:
We are currently discussing whether to have long names `PREMULTIPLY_ON_UPLOAD` or short ones `UNPACK`.

Pros: 
1. more meaning 
2. `const { PREMULTIPLY_ON_UPLOAD } = ALPHA_MODES` can help to shorten the code

Cons:
1. `ALPHA_MODES.xxx` looks better in code if you use auto-complete feature of IDE
2. Longer name cant explain all the cases - its JSDOC and actual code in the renderer that helps with understanding.
3. Short names can be mentioned in discussion easier than long ones.

bigtimebuddy:
Use both @ivanpopleyshev! It’s not a big deal to have aliases for the same thing.

bigtimebuddy:
I'm okay with using the shorter aliases internally. @GoodBoyDigital, are you good with this?

ivanpopelyshev:
> `NPM` - `NO_PREMULTIPLIED_ALPHA`
> `UNPACK` - `PREMULTIPLY_ON_UPLOAD`
> `PMA` - `PREMULTIPLY_ALPHA`

I made PMA alias of `PREMULTIPLIED_ALPHA`. You see how its difficult to make longer names? That thing is really pre-multiplied.

bigtimebuddy:
Good work @ivanpopelyshev

PR #6105, opened 2019-09-18
Minor updates to readme

Zyie:
##### Description of change

Minor changes to the readme file to keep it up to date.

- Changed reference to docs template
- Updated cdn install to use v5 instead of 4v in example
- Added import to basic example
- User plugins link now goes to v5 instead of v3


ivanpopelyshev:
Wow, v3 user plugins.

PR #6114, opened 2019-09-22
Dependency typings

Zyie:
##### Description of change

Another request from @bigtimebuddy to add typings for the dependencies that dont have them already.


tsconfig.json
```diff
@@ -1,14 +1,64 @@
...
+        "removeComments": false,
+        "strict": true,
+        "noImplicitAny": true,
+        "noUnusedLocals": true,
+        "noUnusedParameters": true,
+        "noImplicitReturns": true,
+        "baseUrl": "./",
+        "paths": {
+            "@pixi.js": ["./bundles/pixi.js"],
+            "@pixi.js-legacy": ["./bundles/pixi.js-legacy"],
```
bigtimebuddy: `@pixi.js` should be `pixi.js` and `@pixi.js-legacy` should be `pixi.js-legacy.

tsconfig.json
```diff
@@ -1,14 +1,64 @@
...
+            "@pixi/accessibility": ["./packages/accessibility/src/"],
```
bigtimebuddy: Could you provide some more context about why paths to the source file for each package is necessary?

bigtimebuddy:
Thanks @Zyie for doing this. Great work just a couple of comments.

tsconfig.json
Zyie: Yeah so without the paths typescript doesn't know what these modules are. 

Take this line for example, `import { isMobile } from '@pixi/utils';`. We get an error saying typescript cannot find `@pixi/utils`

By using the paths we map `@pixi/utils` to `../../utils/src/`.

Here's a link to explain how it works [path mapping](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping)

tsconfig.json
Zyie: oops, all fixed

tsconfig.json
bigtimebuddy: If I remove these paths, I am still able to build (`npm run build`) fine without them. In terms of letting packages know about other packages exports, we should do this through module typings (like a very permissive `declare module "@pixi/blah";` in each package or in a global typings folder we can remove later once we emit typings). Shouldn't we be using module resolution in TS to resolve the paths to the packages?

Zyie: If you remove the paths and convert say the utils `index.js` to `index.ts` that is when you get the error.

I'll remove them and keep this pr just for the dependency types

PR #6116, opened 2019-09-23
Fixes WebGL Context Regression

### LINKED ISSUE #5997: WebGL Context Regression
Hello! I am developing a pixi app which has to create and destroy lots of PIXI canvases. I've been using PIXI 5.0.4 until recently when I tried to upgrade to 5.1.1.

When I destroy a canvas, I use `app.destroy()` to destroy the WebGL context. If I run the app without refreshing for a while, eventually I get the following message:
```
scripts.js:1 WARNING: Too many active WebGL contexts. Oldest context will be lost.
e.createContext @ scripts.js:1
```

In PIXI 5.0.4, this warning is all that happens and PIXI happily creates a new WebGL context when I ask it to.

In PIXI 5.1.0 and above, I get the same warning and then an error:

```
scripts.js:1 WARNING: Too many active WebGL contexts. Oldest context will be lost.
e.createContext @ scripts.js:1
/main-es2015.js:1 ERROR TypeError: Failed to execute 'shaderSource' on 'WebGL2RenderingContext': parameter 1 is not of type 'WebGLShader'.
```

After I get this message, I can't create any new WebGL contexts without refreshing the page. I tried reverting my project to use PIXI 5.0.4 and this error went away.

I'm not a WebGL guru, but noticed that #5872 touches on code that's somewhat related.

JetLua:
Add WebGLContext Missing Detection：

bug: https://pixiplayground.com/#/edit/m2uYajft~ZRxeTrFQbix6
fixed: https://pixiplayground.com/#/edit/nY0h0MXAStRQCnHTd1wYY

### DIFF
--- a/packages/core/src/shader/utils/getTestContext.js
+++ b/packages/core/src/shader/utils/getTestContext.js
@@ -13,7 +13,7 @@ let context = unknownContext;
  */
 export function getTestContext()
 {
-    if (context === unknownContext)
+    if (context === unknownContext || context.isContextLost())
     {
         const canvas = document.createElement('canvas');
 


ivanpopelyshev:
Uh, guys, what if context is null?

JetLua:
Yes, it seems possible to set the initial value of `unknownContext` to `null`.

PR #6127, opened 2019-09-27
Fix: unsubscribe from old Sprite texture

maximgavrilov:
##### Description of change

Unsubscribe from old texture in Sprite once the new one set.

##### Pre-Merge Checklist

### DIFF
--- a/packages/sprite/src/Sprite.js
+++ b/packages/sprite/src/Sprite.js
@@ -640,6 +640,11 @@ export class Sprite extends Container
             return;
         }
 
+        if (this._texture)
+        {
+            this._texture.off('update', this._onTextureUpdate, this);
+        }
+
         this._texture = value || Texture.EMPTY;
         this._cachedTint = 0xFFFFFF;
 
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Its only for the case when texture is not valid. I didnt want to add this thing because I didnt meet the case when it actually led to memory leaks. Whats your case?

maximgavrilov:
No cases yet. Just obvious leak on some execution paths.

ivanpopelyshev:
It also exists in `SimplePlane`. With this PR we introduce a new invariant: sprite can be subscribed only on one not valid texture update.

I leave it to other members of the team, because I dont know whether we should ask all future contributors to maintain it.

maximgavrilov:
If the intend is to provide subscriptions to multiple textures, then `Sprite::_texture` should be an array, not a single var. Because with such intend potential leak will be at the `Sprite::destroy` line: `this._texture.off('update', this._onTextureUpdate, this); ` where no multiple textures are supposed.

Issue #6112, opened 2019-09-21
Setting cacheAsBitmap=true on a Container translates positions on its children if they have a mask


apinder:
Setting `cacheAsBitmap=true` on a Container that contains a child that's a Graphic masked with a Sprite offsets the Graphic to the bottom right.

(Container->Graphic->Sprite mask)

### Expected Behavior
cacheAsBitmap should maintain the Graphic's position. No visual difference should occur after setting `cacheAsBitmap=true`

### Current Behavior
Setting `cacheAsBitmap=true` on the container translates the Graphic to the bottom right.

### Steps to Reproduce
See: https://www.pixiplayground.com/#/edit/cuRi6fT3x7QHWL7cNxRNM

This example:
- Creates a circle Graphic
- Sets the Pixi rabbit Sprite as the mask for this graphic
- Puts both in a Container
- Waits 5 seconds and sets cacheAsBitmap to true on the Container
 
On Chrome this will show the Pixi rabbit masked in the top left of the container. The example waits 5 seconds and then sets `cacheAsBitmap=true`. This shifts the Graphic to the lower right eliding parts of the rabbit. Poor rabbit.

### Environment

This was working as expected in versions <5 but seems to be consistently reproducible for versions >= 5.0.4.

- **`pixi.js` version**: 5.1.4
- **Browser & Version**:  Chrome 67
- **OS & Version**: Ubuntu 18.04
- **Running Example**: https://www.pixiplayground.com/#/edit/cuRi6fT3x7QHWL7cNxRNM

ivanpopelyshev:
Good catch! Reproduced it on dev.

ivanpopelyshev:
Actually that's a bigger problem: any filter that calculates sprite position doesn't account for projection transform stuff.

PR #6140, opened 2019-10-04
Large mask refactoring

ivanpopelyshev:
Continue of #6024.

We have to make better mask architecture before we enable scissor masks by default.

Demo, several rect masks rotating: https://www.pixiplayground.com/#/edit/FYYihntSgSm8fwuyiStWr 

You have to turn on/off masks and filters in the bottom of demo to understand whats going on.

🎁 Added

Introducing new MaskData! Its a component for displayObject that stores mask and its details, including debug: we can switch between mask types and see how many masks above in the same framebuffer were detected by renderer.

🐛 Fixed

Scissor mask is reworked, and works even better than in v4.

💩To be fixed next time

Filter+transform - doesn't work. You can see it on demo. As soon as you turn filters off, both stencil and scissor mask start working. I know how to fix that but it has to be another big PR that adds more documentation and tests both for filters and transforms. (#6112)

🚀 Future

This PR opens a road to:

1.  "stencil: true" optional params for context creation, like in v4
2. Container alpha mask.
3. Set channel (red or alpha) for sprite mask filter


ivanpopelyshev:
New demo #1 : https://www.pixiplayground.com/#/edit/FYYihntSgSm8fwuyiStWr

Current state: 
scissor doesnt work with translation
filter doesnt work with translation (possible problem with #6112)

ANDIKKY:
The source code, I wonder why other camera properties like zoom or view are not set by both methods. Normally,,

packages/core/src/mask/MaskSystem.js
```diff
@@ -166,79 +243,17 @@ export class MaskSystem extends System
...
-        this.renderer.stencil.pushStencil(maskData);
-    }
-
-    /**
-     * Removes the last filter from the filter stack and doesn't return it.
-     *
+     * Checks mask balance after frame rendering
      */
-    popStencilMask()
+    postrender()
```
cursedcoder: `postRender` ?

cursedcoder: But probably these words come together.

packages/core/src/mask/MaskData.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        this._target = null;
+    }
+
+    /**
+     * copies counters from maskData above, called from pushMask()
+     * @param {PIXI.MaskData|null} maskAbove
+     */
+    copyCounters(maskAbove)
```
cursedcoder: I see you hide implementation details for that method, so `copyCountersOrReset`?
Or if there are no maskAbove, that means the mask is the entire screen?

packages/core/src/mask/MaskSystem.js
```diff
@@ -166,79 +243,17 @@ export class MaskSystem extends System
...
-        maskData.renderable = false;
-
-        this.renderer.gl.enable(this.renderer.gl.SCISSOR_TEST);
-
-        const resolution = this.renderer.resolution;
-
-        let y;
-
-        if (this.renderer.renderingToScreen)
+        if (this.errorCounter > 0)
```
GoodBoyDigital: Yo! Do we need the error count stuff? Could it be covered in a unit test instead?

ivanpopelyshev: The thing is, I saw people are writing custom versions of container.renderAdvanced() method to add more masks. I want to cover all push-pops check later, "filter.pop()" too. I believe in that case small "if" in code can save hours of debugging.

bigtimebuddy:
@ivanpopelyshev same conflict here ☝️

PR #6141, opened 2019-10-04
Converts math package to typescript

Zyie:
##### Description of change

As requested by @bigtimebuddy I've had a go at converting the math package to typescript.

The most notable changes are (other than the new types):
- `SHAPES` is now an enum. jsdoc didnt like this so i separated out the docs from the initialization  
- `ObservablePoint` `_x` and `_y` are now `protected`, which caused small changes in `Transform` and possibly could be an issue in other places in the code base. (i guess will be fixed when converted)

##### Pre-Merge Checklist


rollup.config.js
```diff
@@ -93,7 +96,13 @@ async function main()
         // Check for bundle folder
         const external = Object.keys(pkg.dependencies || []);
         const basePath = path.relative(__dirname, pkg.location);
-        const input = path.join(basePath, 'src/index.js');
+        let input = path.join(basePath, 'src/index.js');
```
bigtimebuddy: Could you flip this and check for the Ts file first and fallback to Js? That way we can simply remove the existsSync check when we’re are converted. Also, could you add a TODO comment here to make sure we remove it later?

packages/math/src/shapes/Circle.ts
```diff
@@ -9,6 +9,11 @@ import { SHAPES } from '../const';
  */
 export class Circle
 {
+    public x: number
```
bigtimebuddy: Semicolon? We should make sure to add this to linting.

rollup.config.js
```diff
@@ -13,11 +13,14 @@ import filterPackages from '@lerna/filter-packages';
 import { getPackages } from '@lerna/project';
 import repo from './lerna.json';
 
+const fs = require('fs');
```
bigtimebuddy: Pretty sure you can just do `import fs from 'fs';` here instead

Zyie:
@bigtimebuddy all done. I've added the lint rule and applied it. Also update the rollup config and added a todo

packages/math/src/Matrix.ts
```diff
@@ -15,6 +16,15 @@ import { PI_2 } from './const';
...
 {
+    public a: number;
+    public b: number;
+    public c: number;
+    public d: number;
+    public tx: number;
+    public ty: number;
+
+    public array: Float32Array|null;
+
```
englercj: I think we need to make a ruling on how we plan to make these classes. There are a few different things we need to make a decision about and enforce for consistency across the codebase:

1. Do you assign a value to the member in the member declaration, rather than in the ctor?
2. Do you put the docs on the member declaration and not in the ctor?
3. Do you use the constructor param shorthand for members that are the same type/name as a param?
4. Do you always specify function return values, even when they are obvious or void?

My recommendations:

1. Always prefer assignment in the member declaration rather than the ctor, unless the initial value is based on a ctor param.
2. Put the docs on the member declaration, not in the ctor.
3. Do not use constructor param shorthand, it can make documentation blocks a bit messy.
4. Always specify return values, it can ensure you don't accidentally change the implicit return type.

Another note: There is no need to put `@member` or `@default` or `@type` in the docs, they are inferred from the Typescript code itself, and keeping the docs in sync with the code only becomes more difficult when you have to declare these values in two spots.

Here is how this would look with my recommendations:

```ts
class Matrix
{
    public a: number;
    public b: number;
    public c: number;
    public d: number;
    public tx: number;
    public ty: number;

    /**
     * Member Description
     * Assigned to null because the initial value is not based on a param.
     */
    public array: Float32Array | null = null;

    constructor(a = 1, b = 0, c = 0, d = 1, tx = 0, ty = 0)
    {
        this.a = a;
        this.b = b;
        this.c = c;
        this.d = d;
        this.tx = tx;
        this.ty = ty;
    }
}
```

packages/math/src/Matrix.ts
```diff
@@ -205,7 +215,7 @@ export class Matrix
      * @param {number} y How much to translate y by
      * @return {PIXI.Matrix} This matrix. Good for chaining method calls.
      */
-    translate(x, y)
+    translate(x: number, y: number): Matrix
```
englercj: These return values should be `this` and not `Matrix`.

packages/math/src/ObservablePoint.ts
```diff
@@ -9,13 +11,18 @@
...
+    protected _y: number;
+
     /**
      * @param {Function} cb - callback when changed
      * @param {object} scope - owner of callback
      * @param {number} [x=0] - position of the point on the x axis
      * @param {number} [y=0] - position of the point on the y axis
      */
-    constructor(cb, scope, x = 0, y = 0)
+    constructor(cb: Function, scope: object, x = 0, y = 0)
```
englercj: `Function` should be a type describing the actual function prototype, like `() => void`.

`object` is the equivalent of `{}`. This should be `any` instead.

If you want to get fancy, this could instead be `constructor<T>(cb: (this: T) => void, scope: T, x = 0, y = 0)` so that Typescript expresses that the callback is called with `scope` as `this`.

packages/math/src/ObservablePoint.ts
```diff
@@ -49,7 +56,7 @@ export class ObservablePoint
      * @param {number} [x=0] - position of the point on the x axis
      * @param {number} [y=0] - position of the point on the y axis
      */
-    set(x, y)
+    set(x?: number, y?: number): void
     {
         const _x = x || 0;
         const _y = y || ((y !== 0) ? _x : 0);
```
englercj: I think this could be `set(x = 0, y = x): void`

packages/math/src/ObservablePoint.ts
```diff
@@ -9,13 +11,18 @@
  */
 export class ObservablePoint
```
englercj: This should implement the `IPoint` interface.

packages/math/src/ObservablePoint.ts
```diff
@@ -148,3 +155,4 @@ export class ObservablePoint
  * @memberof PIXI
  * @typedef {(PIXI.Point|PIXI.ObservablePoint)} IPoint
  */
+export type IPoint = Point|ObservablePoint;
```
englercj: I'd rather an interface for `IPoint` that these two classes implement.

packages/math/src/Point.ts
```diff
@@ -80,7 +85,7 @@ export class Point
      * @param {number} [x=0] - position of the point on the x axis
      * @param {number} [y=0] - position of the point on the y axis
      */
-    set(x, y)
+    set(x?: number, y?: number): void
     {
         this.x = x || 0;
         this.y = y || ((y !== 0) ? this.x : 0);
```
englercj: similar feedback to above.

packages/math/src/Transform.ts
```diff
@@ -9,6 +9,32 @@ import { Matrix } from './Matrix';
...
 export class Transform
 {
+    /**
+     * A default (identity) transform
+     *
+     * @static
+     * @constant
+     * @member {PIXI.Transform}
+     */
+    public static IDENTITY = new Transform();
```
englercj: This should be `public static readonly` and the doc comment doesn't need any of those values they are all in the Typescript.

packages/math/src/const.ts
```diff
@@ -25,25 +25,25 @@ export const RAD_TO_DEG = 180 / Math.PI;
...
  * @memberof PIXI
- * @type {object}
+ * @type {enum}
  * @property {number} POLY Polygon
  * @property {number} RECT Rectangle
  * @property {number} CIRC Circle
  * @property {number} ELIP Ellipse
  * @property {number} RREC Rounded Rectangle
+ * @enum {number}
  */
```
englercj: I think this belongs above the `SHAPES` enum, and doesn't need any of these doc `@` values.

bigtimebuddy:
I agree fully will Chad’s points here. Awesome review. This will help us create some stricter standards. All the stuff about the members not being in the constructor is good.

The only exception I take is the docs. Because we made the decision not convert to TSDoc right now (and be less disruptive), I’d like the format to remain the same as JSDoc with the redundant type and defaults. I’d like to revisit this after we move the logic over.

ivanpopelyshev:
We really need IPoint, we cant just Point|ObservablePoint it. 

Also, use IPoint in as many places as possible. I already have TS version of math module, I can compare and see what else did you miss.

packages/math/src/ObservablePoint.ts
Zyie: doing the fancy approach doesnt work on the constructor. So ive done `cb: ()=> any, scope: any`

packages/math/src/const.ts
Zyie: The reason for this being below is that the jsdoc gets stripped out when compiled. If you know how to stop this from happening another way i can change it

packages/math/src/Matrix.ts
Zyie: So with jsdoc if the documentation isnt inside the constructor and above in the member declaration then it gets stripped out. 

so if we change this:
```js
public a: number
constructor(a = 1)
{
    /**
    * @member {number}
    * @default 1
    */
    this.a = a;
}
```
to this:
```js
/**
* @member {number}
* @default 1
*/
public a: number

constructor(a = 1)
{
    this.a = a;
}
```
this docs are removed.

Unless anyone knows how to stop this from happening (@bigtimebuddy?)  i dont think we move the docs.

englercj: That's really unfortunate because I'd expect that most members aren't even listen in the ctor, and only in the declaration area. I wonder if instead of using jsdoc on the typescript, we should use typedoc on the javascript.

packages/math/src/ObservablePoint.ts
englercj: Sorry it has to be on the class not the ctor, I just put it there for comment brevity:

```ts
class ObservablePoint<T>
{
    public scope: T;
    constructor(cb: (this: T) => void, scope: T, x = 0, y = 0)
    {
    }
}
```

packages/math/src/ObservablePoint.ts
```diff
@@ -56,14 +53,13 @@ export class ObservablePoint
      * @param {number} [x=0] - position of the point on the x axis
      * @param {number} [y=0] - position of the point on the y axis
      */
-    set(x?: number, y?: number): void
+    set(x = 0, y? : number): void
     {
-        const _x = x || 0;
-        const _y = y || ((y !== 0) ? _x : 0);
+        const _y = y || ((y !== 0) ? x : 0);
```
englercj: This can be removed in favor of `y = x` in the param list I'm pretty sure.

packages/math/src/Matrix.ts
Zyie: `array` was never documented. However if i put this:

```js
    /**
     * @member {PIXI.Matrix}
     */
    public array: Float32Array|null = null;
```
since its not being used in the constructor then the documentation is kept

bigtimebuddy: Ugh! That’s frustrating but I should’ve expected because I know TS removes comments before type declarations even if removeComments is false. Hmm... need to think about this. Maybe we make an initial minimal pass at all the packages to TS and do cleanup after the logic is over?

bigtimebuddy:
@Zyie could you resolve the conflict with Rectangle? I just pushed some doc changes there.

Zyie:
> @Zyie could you resolve the conflict with Rectangle? I just pushed some doc changes there.

@bigtimebuddy ok should be update to date now

ivanpopelyshev:
I can approve it if you guys wont be nitpicking on me when I change `IPoint` in follow-up PR.

ivanpopelyshev:
Is it only me , or `pixi.js.d.ts` failed , it doesnt contain `IPoint` at all?

Zyie:
@ivanpopelyshev yeah seems like that slipped through. Typescript is removing the jsdoc comment like it did with the enum in `const`.

It can be fixed by moving the jsdoc comment underneath the export for the type.

Zyie:
This has also happened for GD8Symmetry

PR #6150, opened 2019-10-10
make IPoint an interface and not union

ivanpopelyshev:
Follow-up for #6141 

After mentioned PR, `IPoint` disappeared from `pixi.js.d.ts` , completely. Type `npm run types` and see the result in `bundles/pixi.js/pixi.js.d.ts`.

I don't understand what is going on, but at least I can start moving this thing from union to normal interface, in case some classes in `pixi-projection` (Point3d) will to implement it.

`IPoint` is still missing in output, but now we have one more problem!

```js
interface ObservablePoint extends IPoint {
}
class ObservablePoint implements IPoint {
}
```

I know that it should work, but how exactly this thing is generating?

Our current typings are filled with copies of methods and classes that we dont need, we didnt have all those problems in manual v4. Please help me fix at least that one.

I also specified `"@typescript-eslint/interface-name-prefix": 0` because some of our interface dont have "I" in them, for example `SharedArrayBuffer` that exists only in case someone uses es5 ts lib and not es6.

UPD: thanks @Zyie , all fixed.

### DIFF
--- a/packages/math/src/IPoint.ts
+++ b/packages/math/src/IPoint.ts
@@ -1,0 +1,63 @@
+export interface IPoint
+{
+    x: number;
+    y: number;
+    copyFrom(p: IPoint): this;
+    copyTo(p: IPoint): IPoint;
+    equals(p: IPoint): boolean;
+    set(x?: number, y?: number): void;
+}
+/**
+ * Common interface for points. Both Point and ObservablePoint implement it
+ * @memberof PIXI
+ * @interface IPoint
+ */
+
+/**
+ * X coord
+ * @memberof PIXI.IPoint#
+ * @member {number} x
+ */
+
+/**
+ * Y coord
+ * @memberof PIXI.IPoint#
+ * @member {number} y
+ */
+
+/**
+ * Sets the point to a new x and y position.
+ * If y is omitted, both x and y will be set to x.
+ *
+ * @method set
+ * @memberof PIXI.IPoint#
+ * @param {number} [x=0] - position of the point on the x axis
+ * @param {number} [y=0] - position of the point on the y axis
+ */
+
+/**
+ * Copies x and y from the given point
+ * @method copyFrom
+ * @memberof PIXI.IPoint#
+ * @param {PIXI.IPoint} p - The point to copy from
+ * @returns {this} Returns itself.
+ */
+
+/**
+ * Copies x and y into the given point
+ * @method copyTo
+ * @memberof PIXI.IPoint#
+ * @param {PIXI.IPoint} p - The point to copy.
+ * @returns {PIXI.IPoint} Given point with values updated
+ */
+
+/**
+ * Returns true if the given point is equal to this point
+ *
+ * @method equals
+ * @memberof PIXI.IPoint#
+ * @param {PIXI.IPoint} p - The point to check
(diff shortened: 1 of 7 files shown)


packages/math/src/Point.ts
```diff
@@ -1,13 +1,29 @@
-import { IPoint } from './ObservablePoint';
+/**
+ * Common interface for points. Both Point and ObservablePoint implement it
+ * @memberof PIXI
+ * @interface IPoint
+ * @property {number} x - x coord
+ * @property {number} y - y coord
```
Zyie: @ivanpopelyshev if you move this below the initialisation then the jsdoc comment is not removed and the type appears in `pixi.js.d.ts` 

```js
export interface IPoint
{
    x: number;
    y: number;
    copyFrom(p: IPoint): this;
    copyTo(p: IPoint): IPoint;
    equals(p: IPoint): boolean;
    set(x?: number, y?: number): void;
}
/**
 * Common interface for points. Both Point and ObservablePoint implement it
 * @memberof PIXI
 * @interface IPoint
 * @property {number} x - x coord
 * @property {number} y - y coord
 */
```

ivanpopelyshev:
Result in pixi.js.d.ts:
```
    /**
     * Common interface for points. Both Point and ObservablePoint implement it
     * @memberof PIXI
     * @interface IPoint
     */
    interface IPoint {
        /**
         * X coord
         * @memberof PIXI.IPoint#
         * @member {number} x
         */
        x: number;
        /**
         * Y coord
         * @memberof PIXI.IPoint#
         * @member {number} y
         */
        y: number;
        /**
         * Sets the point to a new x and y position.
         * If y is omitted, both x and y will be set to x.
         *
         * @method set
         * @memberof PIXI.IPoint#
         * @param {number} [y=0] - position of the point on the y axis
         */
        set(x?: number, y?: number): void;
        /**
         * Copies x and y from the given point
         * @method copyFrom
         * @memberof PIXI.IPoint#
         * @param {PIXI.IPoint} p - The point to copy from
         * @returns {this} Returns itself.
         */
        copyFrom(p: PIXI.IPoint): this;
        /**
         * Copies x and y into the given point
         * @method copyTo
         * @memberof PIXI.IPoint#
         * @param {PIXI.IPoint} p - The point to copy.
         * @returns {PIXI.IPoint} Given point with values updated
         */
        copyTo(p: PIXI.IPoint): PIXI.IPoint;
        /**
         * Returns true if the given point is equal to this point
         *
         * @method equals
         * @memberof PIXI.IPoint#
         * @param {PIXI.IPoint} p - The point to check
         * @returns {boolean} Whether the given point equal to this point
         */
        equals(p: PIXI.IPoint): boolean;
    }
```

packages/math/src/IPoint.ts
```diff
@@ -0,0 +1,62 @@
+export interface IPoint
+{
+    x: number;
+    y: number;
+    copyFrom(p: IPoint): this;
+    copyTo(p: IPoint): IPoint;
```
andrewstart: An improvement for this method would be `copyTo<T extends IPoint>(p: T): T;`

ivanpopelyshev: ok, i can do it, but im after to put it in jsdocs :)

PR #6159, opened 2019-10-12
HitTest hole fix

ivanpopelyshev:
Fix for #6156 , thanks to @xuyangbill

Doesn't work: https://www.pixiplayground.com/#/edit/XSqLaGQ_Htc9T1N8WSYlj
Fixed: https://www.pixiplayground.com/#/edit/qZJGOVcZm~qTrm2orFgtX

Issue #6156, opened 2019-10-12
Graphics with hole containsPoint issue


xuyangbill:
### Current Behavior
the first graphicsData has a hole,the second graphicsData is in the hole.click the second graphicsData,hitTest return false
### Possible Solution
https://github.com/pixijs/pixi.js/blob/460992f299abd87e698d5286180e320c42482cc1/packages/graphics/src/GraphicsGeometry.js
line 385:continue when hole contains point

### Environment

- **`pixi.js` version**:4.8.8
- **Browser & Version**: Chrome 76.0.3809.100
- **OS & Version**:win10
- **Running Example**: 
```
function render(view: PIXI.Graphics) {
    view.clear();
    view.beginFill(0xff0000, 1);
    const outer = [new PIXI.Point(1000, 1000),
    new PIXI.Point(1000, 8000),
    new PIXI.Point(8000, 8000),
    new PIXI.Point(8000, 1000)];
    const holes = [[new PIXI.Point(3000, 3000),
    new PIXI.Point(3000, 6000),
    new PIXI.Point(6000, 6000),
    new PIXI.Point(6000, 3000)]];
    drawPolygonWithHoles(view, outer, holes);
    view.endFill();

    view.beginFill(0x00ff00, 1);
    const inner = [new PIXI.Point(4000, 4000),
    new PIXI.Point(4000, 5000),
    new PIXI.Point(5000, 5000),
    new PIXI.Point(5000, 4000)];
    drawPolygon(view, inner);
    view.endFill();
}

function drawPolygon(view: PIXI.Graphics, points: PIXI.Point[]) {
    points.push(points[0]);
    view.drawPolygon(points);
}

function drawPolygonWithHoles(graphics: PIXI.Graphics, outer: PIXI.Point[], holes: PIXI.Point[][]) {
    graphics.currentPath = null!;
    drawPolygon(graphics, outer);
    holes.forEach(hole => {
        graphics.currentPath = null!;
        drawPolygon(graphics, hole);
        (graphics as any).addHole();
    });
}
```

ivanpopelyshev:
Unfortunately, I can't use continue there due to jslint rule.

```
  354:9   warning  Using 'LabeledStatement' is not allowed  no-restricted-syntax
```

build will be here: https://pixijs.download/fix-hole-hittest/pixi.js

ivanpopelyshev:
Oh, right, you need pixi-v4?

ivanpopelyshev:
[...]

PR #6160, opened 2019-10-12
[v4.x] HitTest hole fix

ivanpopelyshev:
Cherry-pick #6159, thanks to @xuyangbill for noticing it and providing a fix.


ivanpopelyshev:
It was noticed in #6156

PR #6162, opened 2019-10-13
Large BatchRenderer refactoring

ivanpopelyshev:
**This PR is ready for review!**

Part #2 of custom BatchRenderer initiative.

It should also fix the problem with bunnymark - v5 is slower than v4 because of 50 full texture binds on 100.000 bunnies (max batch size = 2000)

With this PR, v5 uses same operations as v4 on GPU, and while its still 15-20% slower `flush()`, 5% overall slower than v4 by CPU, **most devices are GPU-bound and bunnymark shows same FPS**

After this PR I will add more params to bunnymark and try to improve CPU side too.

**Explanation**

I re-shuffled our operations that way each texture rebind and drawcalls correspond to a function call. There wont be 100.000 function calls on 100.000 bunnies - packGeometry also manipulates ranges.

Texture batching is now above drawCalls separation. That means it prioritizes heavy bindTexture call over light blendModes and uniforms. Yes, we can get a bit more batches in certain handmade cases, but extra drawCall << extra texture rebind.

Texture binding is smart - if batch wants to bind 16 textures, binding will check which ones already present in boundTextures for O(1) per texture. That should fix the bunnymark.

For extra kick I've added `bufferedTextures` field, that way we dont process heavy sprite array two times in `flush()`.

There are also minor things like 

* strange `_id` field in BaseTexture that wasnt specified in constructor, I renamed it to `_batchLocation`
* unnecessary active texture location change
* a few others

`drawCalls[]` and `textureArrays[]` are static pools, thus if we add more custom renderers number of JS objects wont be multiplied.

**Possible BatchSystem improvement**

As a variation of those two functions I added in `batchSystem`, there's an idea of "begin"/"end" kind of operation for multitexture batch - `batch.startMultiTexture()` returns object which has temp tickID inside, which can execute those operations, `endMultiTexture` saves tickID back. Not sure if we should obfuscate it that way.

**Names**

Yes, names are the problem.

**Tests**

Now we have a test that covers texture and blendmode batching!

**Compatibility with 5.1 custom renderers**

I've decided to leave `packInterleavedGeometry` with same interface, as I didnt find and difference if I change it to function that processes range instead of one element. `pixi-projection` and other custom renderers built by people for 5.1 should still work.


ivanpopelyshev:
Current state of benchmark on my computer

100k bunnies on nvidia 970M

4.8.8 - 17-19FPS
5.1.5 - 11-13FPS 
branch 14-16FPS

SpectorJS view:

v4

image

v5

16 texture binds, shown only last 3
image

branch

image

ivanpopelyshev:
New performance test.

20.000 bunnies
CPU i5-2400
GPU NVidia GT620

image

4.8.8 utilizes less resources than 5.1.5, and feature-branch is on top FPS. 

Scripting relative consuming is the same.

4.8.8 idle is high.

Conclusion: its just versions with different balance, but im sure that feature-branch wins against 5.1.5.

ivanpopelyshev:
Extra info:

image

`flush` aka inlined `buildTexturesAndDrawcalls` is the same for v4 and v5, but bigger by 1% (relative 5%) in feature. Thus, we trade about max 5% power of flush for architectural improvements I suggest.

`calculateVertices` in both v5 is faster.

There's still that point that feature branch doesn't spam `bindTexture` like v5 does.

GoodBoyDigital:
amazing work man! 👍

GoodBoyDigital:
looking through now

GoodBoyDigital:
does making this branch . batch size 4096 have an effect?

ivanpopelyshev:
@GoodBoyDigital yes it had on bunnymark on one of computers. I'll make more tests today-tomorrow

ivanpopelyshev:
I've tried it on my home GT970 with 40.000 bunnies. Its a CPU-bound computer, and here feature-branch and v4.8.8 lost to 5.1.5, because texture binds dont affect performance.

packages/core/src/batch/BatchDrawCall.js
```diff
@@ -9,12 +9,17 @@ export class BatchDrawCall
 {
     constructor()
     {
-        this.textures = [];
-        this.ids = [];
+        this.texArray = null;
         this.blend = 0;
-        this.textureCount = 0;
+        this.type = 4;
```
bigtimebuddy: I know this is a hold over, but what is `4`?

packages/core/src/batch/BatchSystem.js
```diff
@@ -66,4 +66,61 @@ export class BatchSystem extends System
...
+        {
+            arr[i] = boundTextures[i] || null;
+            if (arr[i])
+            {
+                arr[i]._batchLocation = i;
+            }
+        }
+    }
+
+    boundArray(texArray, boundTextures, batchId, maxTextures)
```
bigtimebuddy: Documentation for this method? These docs are very helpful fir us converting to TypeScript

packages/core/src/batch/BatchDrawCall.js
ivanpopelyshev: DRAW_MODES.TRIANGLES

bigtimebuddy: I understand not wanting to import to save space, could you inline `// DRAW_MODES.TRIANGLES` here?

packages/core/test/BatchRenderer.js
```diff
@@ -1,43 +1,136 @@
...
+        context.fillStyle = 'white';
+        context.fillRect(0, 0, w, h);
 
-        expect(() => renderer.destroy()).to.not.throw();
-    });
+        return new BaseTexture(new CanvasResource(canvas));
+    }
 
-    it('can be destroyed immediately', function ()
+    it('should pass the batching test', function ()
```
cursedcoder: Wow, this looks huge, do you plan to have more batching tests in future?

ivanpopelyshev: This test shown me a pair of stupid bugs, and it improves our coverage by more than 1%. I will make better architecture If I ever need second test.

ivanpopelyshev: With better architecture I already know a number of tests I want to add. However, I've spent too many hours on this PR to go further :)

cursedcoder: Ok, let's add more in the future.

bigtimebuddy:
@ivanpopelyshev could you fix the conflict?

PR #6164, opened 2019-10-13
Convert Runner to TypeScript

bigtimebuddy:
##### Description of change
Continuing on the TypeScript conversion, this converts `@pixi/runner`, pretty straightforward.

##### Pre-Merge Checklist


bigtimebuddy:
Topic for discussion: `strictNullChecks`

In `@pixi/math` conversion, there are no `destroy` methods but this is pretty common throughout all the other packages. In these cases, many things are set to `null` internally upon destroy. I think we should relax this compiler setting to allow for nullifying properties without introducing things more cumbersome types like `number | null` or `string | null`. Thoughts?

packages/runner/src/Runner.ts
```diff
@@ -53,9 +61,10 @@ export class Runner
 
     /**
      * Dispatch/Broadcast Runner to all listeners added to the queue.
-     * @param {...any} params - optional parameters to pass to each listener
+     * @param {...any} [params] - optional parameters to pass to each listener
+     * @return {PIXI.Runner}
      */
-    emit(a0, a1, a2, a3, a4, a5, a6, a7)
+    public emit(a0?: any, a1?: any, a2?: any, a3?: any, a4?: any, a5?: any, a6?: any, a7?: any): this
```
eXponenta: Why `...a` can't be used instead of `a0, a1,a2` ?
Then it can be used instead of `arguments`.

packages/runner/src/Runner.ts
```diff
@@ -53,9 +61,10 @@ export class Runner
 
     /**
      * Dispatch/Broadcast Runner to all listeners added to the queue.
-     * @param {...any} params - optional parameters to pass to each listener
+     * @param {...any} [params] - optional parameters to pass to each listener
```
Zyie: For me making the params optional gives an error in the types file.

`A rest parameter cannot be optional.`

packages/runner/src/Runner.ts
bigtimebuddy: It’s done this way to avoid an apply or call, which has less performance

PR #6173, opened 2019-10-15
Convert constants to TypeScript

Zyie:
##### Description of change

- Converts the constants package to typescript. Replaces objects with constant enums
- Adds tests for the couple of enums that were missing

Like with previous conversion the docs have to be placed underneath the declaration.  
##### Pre-Merge Checklist


eXponenta:
Enums of typescript generate large JS representation.

Looks this:
https://www.typescriptlang.org/play/?preserveConstEnums=true#code/KYDwDg9gTgLgBMAdgVwLZwEIBkCiA5AEQH0BZAeQJwGU4BvAWACg4W48yAlEgQSzgF44ABgA0TVnG4ECAuAEYxzViQCqWACoBJAApYAmrIBMiiVQDCHHPlkBmE6zIA1HByzcDggCz2WBbhwBpa0EAVh84LE0AcQAJdWC4ADZwszIsTiICCiicWQB2FLSMjBUOPFkADnCY-2JI2PVZAE5wqjIAMXUierjZOVFxVgJNdvaXfDNcwTkFQZYcAA0zLBUqTTJy6eM5uBiVKfk7HapudVLT9c35bx3U9I4+sJ2Vkk12NfUPeWSd9i5eIh4bQkPoFHZSYhAkHTKrHCxWPCA4F9Fq-DYHQwDRjHDhmIhOFyyLGmXFEN5GWZKFhUUlkFSNQSGbZUuA0vGnMjaIxHFkEKhdAkPRk3Xn8slXQxPUUC+lGH7SogcrmMsEslzcKgY+WmFQYdQcbhmBlwQywlkLThGVGMAC+QA

Minification will not reduce this code because 'string' literals is used.
This file will increase size to 16kb after compilation.

packages/constants/src/index.ts
```diff
@@ -1,20 +1,143 @@
+export const enum ENV {
```
englercj: `const enum` means the enum will be inlined into usages, and not emitted into the output. These should be emitted though so use just `export enum`.

See: http://www.typescriptlang.org/play/#code/MYewdgzgLgBApmArgWxgYXNAKnaMDeMAggDQwBCZaMAvgFCiSyPa6wC86mUO0AdEQDcdAPQiYAWikS6CFDABy4DE16xCpClVoNuMMMu5qYnJWBWt+QukA

Zyie: ok they have been changed to `export enum`

PR #6174, opened 2019-10-15
Pair of graphics bugs

ivanpopelyshev:
One critical bug was introduced in #5964 
You can see it at https://pixijs.io/examples/#/graphics/advanced.js - lineTextureStyle doesn't work at all.
Color should be 0xffffff by default.

As for another minor performance bug, it survived several graphics PR's, its missing `!!style.native` for case when `native` is undefined (fillStyle). 

Right now it was affected if we had several fillStyles of the same color, or lineStyle/fillStyle of the same color - it added more batches in GraphicsGeometry.

I've added one test for it.


packages/graphics/src/Graphics.js
```diff
@@ -322,7 +322,7 @@ export class Graphics extends Container
         options = Object.assign({
             width: 0,
             texture: Texture.WHITE,
-            color: 0x0,
+            color: 0xFFFFFF,
```
bigtimebuddy: What is this white? Black was intentional as a default.

eXponenta: Because it is tint for texture, and always was white. 
Black tint => black filled result, texture not showing.

bigtimebuddy: Ohhh, I didn't realized that `color` was a tint. The documentation needs to be updated.

ivanpopelyshev: oh, but in lineStyle default was black, right? eh, we might have a problem, need one more test

ivanpopelyshev:
Done.

bigtimebuddy:
Thanks @ivanpopelyshev

Issue #6181, opened 2019-10-18
PIXI.Geometry.destroy() throws error when no index provided


lokanchung:
Hi,
I found that when `PIXI.Geometry` is provided with no index information, calling `destroy()` throws an error. It tries to destroy a non-existing index buffer.

Geometry setup:
```
const geometry = new PIXI.Geometry()
        .addAttribute('aVertexPosition', vertices, 2)
        .addAttribute('aVal', val, 4)
        .addAttribute('aDx', dx, 4)
        .addAttribute('aDy', dy, 4)
        .addAttribute('aDxy', dy, 4)
        .addAttribute('aBary', baries, 3)
        .addAttribute('aCat1', cat1, 2)
        .addAttribute('aCat2', cat2, 2)
        .addAttribute('aCat3', cat3, 2)
```
when `geometry.destroy()` is called

`[TypeError: this.indexBuffer is null]`

An error happens in line 286.

https://github.com/pixijs/pixi.js/blob/5fc57749998e13dfe2f84dcf815773921496f9ea/packages/core/src/geometry/Geometry.js#L281-L289

So, a null check is needed.

Version: 5.1.5
Playground:
https://www.pixiplayground.com/#/edit/KPUx7Aomup_ywslXy48KE

ivanpopelyshev:
Good catch! Gonna fix it

PR #6182, opened 2019-10-18
Indexbuffer destroy fix

ivanpopelyshev:
Fix for #6181

The thing is, indexBuffer can be shared, and its up to `dispose` to take care of videomemory and the rest.


ivanpopelyshev:
Yeah, we didnt fix it when I made those changes regarding geometry and buffers disposal.

PR #6186, opened 2019-10-20
Convert ticker to TypeScript

Zyie:
### Description of change

Another package converted to TypeScript. This time its ticker.

Ive created a `global.d.ts` to define the settings package as without this typescript cannot find the module.

Ive also added `IApplicationOptions` temporarily as the `TickerPlugin` class takes these options in the constructor. I would have thought the best place to define it, in the end, would be in the `app` package when that gets converted. Until then i thought leaving it there would be cleaner than half converting the `app` package.

I'll happily update this if you guys disagree and have a better idea.

##### Pre-Merge Checklist


packages/ticker/src/Ticker.ts
```diff
@@ -264,7 +284,7 @@ export class Ticker
      * @param {TickerListener} listener - Current listener being added.
      * @returns {PIXI.Ticker} This instance of a ticker
      */
-    _addListener(listener)
+    _addListener(listener: TickerListener): this
```
englercj: Should be `private`.

PR #6195, opened 2019-10-26
Remove extra gl-flush and GC checks

ivanpopelyshev:
I think we have to call `gl.flush()` for certain mobile devices after our frame, but we dont have to do it for each renderTexture call.

Also GC ticker rate was increased if user draws renderTextures.

### DIFF
--- a/packages/core/src/context/ContextSystem.js
+++ b/packages/core/src/context/ContextSystem.js
@@ -229,7 +229,10 @@ export class ContextSystem extends System
      */
     postrender()
     {
-        this.gl.flush();
+        if (this.renderer.renderingToScreen)
+        {
+            this.gl.flush();
+        }
     }
 
     /**
--- a/packages/core/src/textures/TextureGCSystem.js
+++ b/packages/core/src/textures/TextureGCSystem.js
@@ -61,6 +61,11 @@ export class TextureGCSystem extends System
      */
     postrender()
     {
+        if (!this.renderer.renderingToScreen)
+        {
+            return;
+        }
+
         this.count++;
 
         if (this.mode === GC_MODES.MANUAL)


packages/core/src/textures/TextureGCSystem.js
```diff
@@ -61,6 +61,11 @@ export class TextureGCSystem extends System
      */
     postrender()
     {
+        if (!this.renderer.renderingToScreen)
+        {
+            return;
+        }
+
```
GoodBoyDigital: do you think this could cause issue where textures will never be removed if I for example am rendering my scene directly to a render texture each frame then using that as an input?

ivanpopelyshev: Input for what? You have to do readPixels() for that. readPixels()-bound app has much bigger problems than that ;)

GoodBoyDigital: ah i see! So it would get called when I render the input to the screen :D

packages/core/src/context/ContextSystem.js
```diff
@@ -229,7 +229,10 @@ export class ContextSystem extends System
      */
     postrender()
     {
-        this.gl.flush();
+        if (this.renderer.renderingToScreen)
```
GoodBoyDigital: just curious about why would _not_ want to call flush after rendering to a render texture?

ivanpopelyshev: just extra gl operation in spector logs. Its you who started it!

GoodBoyDigital: haha :D 
In you opinion, does it have any effect?

ivanpopelyshev: Gl.flush() placed after main framebuffer render helped us on some mobile devices as browser/bug workaround. We didnt have it in v4. I have no idea how it affects things if we flush it after rendertextures

GoodBoyDigital: cool!

PR #6198, opened 2019-10-27
Set method for shapes classes

pbalmasov:
##### Description of change
Set method for shapes classes for reusing classes
example
```javascript
const bounds = new Rectangle();
const temp = new Rectangle();
rectsData.forEach(({
    x,
    y,
    width,
    height
}) => {
    temp.set(x, y, width, height)
    bounds.enlarge(temp);
});
```

##### Pre-Merge Checklist


ivanpopelyshev:
I think its better to add those things when we have good use-cases for them.

bigtimebuddy:
I agree with Ivan. If you need these in your project, simply add them to the prototype. Or create a packages to distribute these helpers as a plugin.

PR #6199, opened 2019-10-27
Point return self for easier chaining

pbalmasov:
##### Description of change
Point return self for easier chaining

##### Pre-Merge Checklist


ivanpopelyshev:
Trailing space after `: this`

pbalmasov:
fixed

djmisterjon:
Am also a fan of chaining programming.
it would be great to see this kind of update at somes strategies locations.
Maybe also for all `ObservablePoint.prototype.set` for allow custom method
Example on my side `sprite.position.set(x,y).zero()` this store a zero cache values before complexe animation to restore point.

ivanpopelyshev:
I'm not fan of chaining. Even more, I suggest any new math methods and things have to be used in pixi, no sugaring stuff for users if we dont use it for ourselves.

However, in case of `set` chaining is fine :)

bigtimebuddy:
I think this is a good API change and is pretty benign. I could see it being used like this:

```js
const { x, y } = sprite.position.set(Math.random(), Math.random());
```
As opposed to the alternative:
```js
const x = Math.random() * 100;
const y = Math.random() * 100;
sprite.position.set(x, y);
```
Could also be used with `toLocal` and `toGlobal` while setting:
```js
const localPos = local.toLocal(sprite.position.set(100, 200));
```

Overall, this doesn't endorse chaining programming per-se, and could add some nicer API ergonomics.

djmisterjon:
is that can be also ok for the proto by rectangle plz?
```js
  Rectangle.prototype.pad 
```

ex:
to allow this
```js
 display.hitArea = display.getLocalBounds().pad(cpX, 0)
```

bigtimebuddy:
Sure. PR @djmisterjon?

djmisterjon:
> Sure. PR @djmisterjon?

my bad, it already done in v5.
missing in v4

PR #6226, opened 2019-11-16
Fix graphics fillStyle instead of lineStyle

ivanpopelyshev:
Another bug again, found by @eXponenta 

Bugged: https://www.pixiplayground.com/#/edit/QF2RXuO21iDTwvxGntVBr
Fixed: https://www.pixiplayground.com/#/edit/yhKcbj73S9RHy1ajD0tdX

I've added test on it. With our previous tests everything should be fine, but I'll make a pass on pixijs examples anyway.

Also it fixes our typings.


ivanpopelyshev:
OK, its not fine, we have several copies of `BatchPart` in typings. Any ideas?

ivanpopelyshev:
Should be fine now.

PR #6231, opened 2019-11-21
Fix for when '_segmentsCount' can return NaN

themoonrat:
If the moveTo y value and the quadraticCurveTo cpY and toY  values are all the same, then `QuadraticUtils.curveLength` returns NaN, causing `GRAPHICS_CURVES._segmentsCount` to return NaN.

Demo: https://www.pixiplayground.com/#/edit/8jAg09k2W~L85zPt~ZfU_
Open up the console, put a breakpoint on `const n = GRAPHICS_CURVES._segmentsCount` and refresh the playground. Notice that 'n' will resolve to NaN.

This is probably not the right solution for this, fixing the symptom rather than the cause, but just in case it's fine, here's the PR!


bigtimebuddy:
Thanks @themoonrat!

PR #6233, opened 2019-11-21
Added SimpleRope deprecation

themoonrat:
##### Description of change
Added a deprecation to PIXI.SimpleRope (which was PIXI.mesh.Rope) that I came across upgrading games from v4 to v5.

##### Pre-Merge Checklist

### DIFF
--- a/bundles/pixi.js/src/useDeprecated.js
+++ b/bundles/pixi.js/src/useDeprecated.js
@@ -1069,6 +1069,28 @@ export function useDeprecated()
         },
     });
 
+    /**
+     * @deprecated since 5.0.0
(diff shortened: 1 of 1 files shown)


themoonrat:
I've just completed an upgrade of 40+ games going from pixi v4 to pixi v5... and this is one of the changes that caused a couple of games to not work properly.

I've just PR'd deprecations over the months as I've come across upgrade issues! :)

bigtimebuddy:
I don’t think we should makes two APIs for the same thing. We should promote the use of geometry object instead of adding convenience getters. Or else we’d need to added to many other classes.

PR #6243, opened 2019-11-25
feat: add clone method to IPoint type definition

MMasey:
##### Description of change
I noticed that the IPoint interface was missing the clone method so this PR adds it in.

### DIFF
--- a/packages/math/src/IPoint.ts
+++ b/packages/math/src/IPoint.ts
@@ -40,7 +40,7 @@ export interface IPoint
  * Creates a clone of this point
  * @method clone
  * @memberof PIXI.IPoint#
- * @return {PIXI.Point} a copy of the point
+ * @return {PIXI.IPoint} a copy of the point
  */
 
 /**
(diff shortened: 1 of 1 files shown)


packages/math/src/IPoint.ts
```diff
@@ -35,6 +36,13 @@ export interface IPoint
  * @param {number} [y=x] - position of the point on the y axis
  */
 
+/**
+ * Creates a clone of this point
+ * @method clone
+ * @memberof PIXI.IPoint#
+ * @return {PIXI.Point} a copy of the point
```
bigtimebuddy: Should the return value be IPoint?

MMasey: Hi @bigtimebuddy, you're correct. I've just pushed up a fix 👍

bigtimebuddy:
@ivanpopelyshev what are you proposing?

ivanpopelyshev:
`ObservablePoint.clone()` should return `Point` and not `ObservablePoint`

bigtimebuddy:
Am closing this PR because `clone` method signatures are not the same between Point and ObservablePoint. Having ObservablePoint clone to Point is a breaking change. Would prefer to leave this as-is and have users add [type assertions](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions) for their particular IPoint objects when cloning.

Issue #6247, opened 2019-11-26
Tests are failing on a clean clone of the `dev` branch


EvidentlyCube:
I've cloned a fresh copy of the repository, `dev` branch. I first ran `npm i` then `npm run test`. There were five errors reported in the tests:

```
1) PIXI.BitmapFontLoader
       should properly register bitmap font having more than one texture:
     Uncaught Error: Texture Error: frame does not fit inside the base Texture dimensions: X: 2 + 19 = 21 > 1 and Y: 2 + 20 = 22 > 1
      at Texture.prototypeAccessors.frame.set (C:\Projects\Personal\JS\pixi.js\packages\core\lib\core.js:3671:19)
      at new Texture (C:\Projects\Personal\JS\pixi.js\packages\core\lib\core.js:3313:24)
      at Function.registerFont (C:\Projects\Personal\JS\pixi.js\packages\text-bitmap\lib\text-bitmap.js:635:26)
      at Function.parse (C:\Projects\Personal\JS\pixi.js\packages\text-bitmap\lib\text-bitmap.js:681:38)
      at completed (C:\Projects\Personal\JS\pixi.js\packages\text-bitmap\lib\text-bitmap.js:784:30)
      at MiniSignal.dispatch (C:\Projects\Personal\JS\pixi.js\node_modules\mini-signals\lib\mini-signals.js:93:18)
      at C:\Projects\Personal\JS\pixi.js\node_modules\resource-loader\dist\resource-loader.cjs.js:2262:34
      at next (C:\Projects\Personal\JS\pixi.js\node_modules\resource-loader\dist\resource-loader.cjs.js:53:9)
      at Loader.use (C:\Projects\Personal\JS\pixi.js\packages\text-bitmap\lib\text-bitmap.js:730:9)
      at C:\Projects\Personal\JS\pixi.js\node_modules\resource-loader\dist\resource-loader.cjs.js:2260:10

  2) PIXI.BitmapFontLoader
       should split fonts if page IDs are in chronological order:
     Uncaught Error: Texture Error: frame does not fit inside the base Texture dimensions: X: 2 + 19 = 21 > 1 and Y: 2 + 20 = 22 > 1
      at Texture.prototypeAccessors.frame.set (C:\Projects\Personal\JS\pixi.js\packages\core\lib\core.js:3671:19)
      at new Texture (C:\Projects\Personal\JS\pixi.js\packages\core\lib\core.js:3313:24)
[...]

PR #6248, opened 2019-11-26
Refactored some of TextMetrics tests so that they pass on Windows

EvidentlyCube:
##### Description of change
Refactored some of `TextMetric` tests so that they pass on Windows.

 * I've allowed myself to leave a comment at the top of `TextMetric` class about the fact that the text may render (and measure) differently on different platforms.
 * One failing test required changing the expected width, because `exercitationem` renders to about 217px making it impossible to fit in the requested 200px. This shouldn't invalidate the test on different platforms, because `breakWords` was set to `false` either way, so on other platforms it just rendered to a smaller dimension already. Also the test's goal is to check if the width is not exceeded by stacking >1 word in the line, not by a word too long.
 * Two other tests I refactored to be 2 tests each, one part being stuff that works the same across all platforms (trimming whitespace) the other part being stuff that depends on how things are rendered (checking exactly how things break).

##### Pre-Merge Checklist

Referenced issue: #6247

### DIFF
--- a/packages/text/test/TextMetrics.js
+++ b/packages/text/test/TextMetrics.js
@@ -1,5 +1,10 @@
 const { TextMetrics, TextStyle } = require('../');
 
+/**
+ * Fonts render slightly differently between platforms so tests that depend on a specific
+ * widths or breaking of words may not be cross-platform
+ */
+
 /* eslint-disable no-multi-str */
 const longText = 'Sed ut perspiciatis unde omnis iste natus error sit voluptatem \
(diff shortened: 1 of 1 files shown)


EvidentlyCube:
image
And the output on my machine after I run the test suite with these changes.

ivanpopelyshev:
Its something about DPI, I don't remember exactly. I agree, lets just skip it in windows, we can fix cross-platrom way later.

bigtimebuddy:
Thank you @EvidentlyCube!

PR #6249, opened 2019-11-26
Fixed BitmapFontLoader tests failing on Windows

EvidentlyCube:
##### Description of change
Fixed the two tests that failed on windows in BitmapFontLoader. I think the code itself is self-explanatory so I don't really have much more to say. I am confident the change in `dirname` method should be safe because backslashes are not allowed in a URL, so shouldn't ever break any existing code, and even though they're allowed in unix's filenames afaik no one ever does it.

##### Pre-Merge Checklist

Referenced issue: #6247


EvidentlyCube:
image
And the console output when running tests on my machine.

bigtimebuddy:
Thank you!

Issue #6201, opened 2019-10-31
TextMetrics incorrect calculations using breakWords and Emojis


Teraqz:
Using:
`var characters = token.split('');`
in TextMetrics.wordWrap will return incorrect character array and calculated width, when you have string of emoji like "🤪🤪"

### Current Behavior
`"🤪🤪".split("") => (4) ["�", "�", "�", "�"]`
### Possible Solution
Using ES6 i managed to get it right: 
`[..."🤪🤪"] => (2) ["🤪", "🤪"]`

### Environment

pixi.js version: e.g. 5.1.3
Browser & Version: e.g. ios13.1 safari

ivanpopelyshev:
Good carch!

Teraqz:
@ivanpopelyshev is there any chance you guys could integrate this lib https://github.com/orling/grapheme-splitter as a plugin or extension to TextMetrics? 
It would be very handy and it's actually make sense if you are willing to use PIXI with more complicated Text, and if not, just don't use this plugin.

bigtimebuddy:
I don't think we should integrate this dependency because it's rather heavy, but we should allow for users to plugin/replace the string-splitting functionality themselves. For users that need emojis, we should support a drop-in replacement.

Teraqz:
@bigtimebuddy yeah, no need for dependency there, just an option to allow to plugin/replace this split function

PR #6251, opened 2019-11-26
TextMetrics.wordWrap() customization of splitting method

EvidentlyCube:
##### Description of change
As requested in #6201, made it so that the method used to split a token into characters in `TextMetrics.wordWrap()` can be customized.

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/TextMetrics.js
+++ b/packages/text/src/TextMetrics.js
@@ -575,7 +575,6 @@ export class TextMetrics
      * // Correctly splits emojis, eg "🤪🤪" will result in two element array, each with one emoji.
      * TextMetrics.wordWrapSplit = (token) => [...token];
      *
-     * @private
      * @param  {string}  token The token to split
      * @return {string[]} The characters of the token
      */
@@ -735,15 +734,6 @@ const canvas = (() =>
 })();
 
 canvas.width = canvas.height = 10;
-
-/**
- * Configuration for TextMetrics
- * @type {{wordWrapTokenSplit: (function(*): (Array|string[]))}} A function that splits a token in `wordWrap` method into
- * separate characters when token is longer than the allowed width
- */
-TextMetrics.configuration = {
-    wordWrapTokenSplit: (token) => token.split(''),
-};
 
 /**
  * Cached canvas element for measuring text
--- a/packages/text/src/TextMetrics.js
+++ b/packages/text/src/TextMetrics.js
@@ -250,7 +250,7 @@ export class TextMetrics
                 if (TextMetrics.canBreakWords(token, style.breakWords))
                 {
                     // break word into characters
-                    const characters = TextMetrics.configuration.wordWrapTokenSplit(token);
+                    const characters = TextMetrics.wordWrapSplit(token);
 
                     // loop the characters
                     for (let j = 0; j < characters.length; j++)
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Thanks for pulling this together.

I don't have a problem with this functionally, but wondered if we could do something a little simpler for the API.

`PIXI.TextMetrics.configuration.wordWrapTokenSplit` is rather verbose. Also we don't have other examples in the API that use a `configuration` object. My preference would be to do something like `TextMetrics.wordWrapSplit` and specify in the docs that it can be overridden to support emojis (for example). An example of how to support emojis would also be excellent thing to document, since that's the main use-case.

themoonrat:
For reference, canBreakWords and canBreakChars are other examples of code split out into functions that can be overridden; in those cases to help with custom CJK language support.

EvidentlyCube:
@themoonrat I am happy to change things to better fit the rest of the code :).

If I understand correctly, the way you'd like this to be used would be in essence:

```
import * as PIXI from 'pixi.js`;

// If this is not done then the default implementation is used
PIXI.TextMetrics.wordWrapSplit = <put your own custom implementation here>;

```

That sounds fair. My personal preference is a configuration variable, just like the way I implemented it, but I am not here to impose my preferences on others :D. I'll update the PR soonish!

EvidentlyCube:
@themoonrat I've got one question though: you mentioned `canBreakWords` and `canBreakChars` but those are marked as `@private` and thus do not appear in the documentation (at least not when I run `npm docs`, though I don't see them in the official documentation either). Is it intended or an oversight? I feel like it'd make more sense to expose them in the docs.

And how about changing the line "This method exists to be easily overriden" to something like "Overridable helper method used internally by TextMetrics, exposed to allow customizing the class's behavior"

themoonrat:
Yeah, I'd say it's a mistake that they are marked as private! Your changes would be appreciated :)

EvidentlyCube:
Pushed the changes - `npm run lint` and `npm run test` both pass (with the exceptions of the tests that fail on Windows)

ivanpopelyshev:
Do you want to join PixiJS Slack? I can sent an invite to your public email on your site.

EvidentlyCube:
Sure! I was hoping to contribute a bit more, as long as life allows, so that might be helpful.

packages/text/src/TextMetrics.js
```diff
@@ -564,6 +564,26 @@ export class TextMetrics
...
+     *
+     * @example
+     * // Correctly splits emojis, eg "🤪🤪" will result in two element array, each with one emoji.
+     * TextMetrics.wordWrapSplit = (token) => [...token];
+     *
+     * @private
+     * @param  {string}  token The token to split
+     * @return {string[]} The characters of the token
+     */
+    static wordWrapSplit(token)
```
bigtimebuddy: Could you remove private flag?

packages/text/src/TextMetrics.js
```diff
@@ -716,6 +736,15 @@ const canvas = (() =>
 
 canvas.width = canvas.height = 10;
 
+/**
+ * Configuration for TextMetrics
+ * @type {{wordWrapTokenSplit: (function(*): (Array|string[]))}} A function that splits a token in `wordWrap` method into
+ * separate characters when token is longer than the allowed width
+ */
+TextMetrics.configuration = {
```
bigtimebuddy: No longer needed

EvidentlyCube:
<insert a generic excuse about sleep deprivation/doing fast too quickly/something here>
 Fixed those!

bigtimebuddy:
You’re on a serious roll this release @EvidentlyCube!

EvidentlyCube:
Let's see how much more I can fit before the next release 🙃. And if it's around the corner - the one after that!

PR #6252, opened 2019-11-27
Canvas mask supports container with graphics

ivanpopelyshev:
Currently assigning container as a mask causes an Exception in CanvasRenderer.

Lets fix that, add a feature, and cover with pair of tests.

Demo: https://www.pixiplayground.com/#/edit/IO~sST9K~QxzhBh9b4b3u
Two rects in two graphics instead of one.

To be fixed in next PR: Even-odd fail. https://www.pixiplayground.com/#/edit/DsFS6hI4dG0tFGrdBSbxW

### DIFF
--- a/packages/canvas/canvas-renderer/package.json
+++ b/packages/canvas/canvas-renderer/package.json
@@ -31,6 +31,8 @@
     "@pixi/utils": "^5.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


ivanpopelyshev:
OK, here comes nonzero/evenodd problem, I'm thinking how to solve it...

PR #6255, opened 2019-11-29
Extend FilterSystem, add measure APIs

SukantPal:
This merges phase's extended filter system into vanilla pixi.js (as Ivan calls it :), as discussed here: https://github.com/pixijs/pixi.js/issues/6246.

Also, this should also serve as a fix to https://github.com/pixijs/pixi.js/issues/6245.

I've added a primitive test for the new `FilterSystem`.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/filters/Filter.js
+++ b/packages/core/src/filters/Filter.js
@@ -4,6 +4,8 @@ import { State } from '../state/State';
 import { settings } from '@pixi/settings';
 import defaultVertex from './defaultFilter.vert';
 import defaultFragment from './defaultFilter.frag';
+import acsVertex from './acs.vert';
+import acsFragment from  './acs.frag';
 
 /**
  * Filter is a special type of WebGL shader that is applied to the screen.
@@ -156,10 +158,8 @@ export class Filter extends Shader
         super(program, uniforms);
 
         /**
-         * The padding of the filter. Some filters require extra space to breath such as a blur.
-         * Increasing this will add extra width and height to the bounds of the object that the
-         * filter is applied to.
-         *
+         * The extra space that is added to the target's bounds for filters to manipulate. Some
+         * filters like blurs need it to "breathe".
          * @member {number}
          */
         this.padding = 0;
@@ -167,22 +167,19 @@ export class Filter extends Shader
         /**
          * The resolution of the filter. Setting this to be lower will lower the quality but
          * increase the performance of the filter.
-         *
          * @member {number}
          */
         this.resolution = settings.FILTER_RESOLUTION;
 
         /**
          * If enabled is true the filter is applied, if false it will not.
-         *
          * @member {boolean}
          */
         this.enabled = true;
 
         /**
-         * If enabled, PixiJS will fit the filter area into boundaries for better performance.
-         * Switch it off if it does not work for specific shader.
-         *
+         * If enabled, PixiJS will apply the filter only on the visible portion of the target's
+         * filter area. You can switch it off if it does not work for specific filter.
          * @member {boolean}
          */
         this.autoFit = true;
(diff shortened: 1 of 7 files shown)


SukantPal:
I could show some examples to use the `measure` system if you deploy this branch. Additionally, we could fix the uselessly complicated calculations (with texture aspect ratio) in the radial blur filters with the ACS shaders provided in `Filter`. See, you could do the calculations in absolute coordinates, calculate the difference b/w initial & final position, and add that in texture coordinates. This saves the calculation of preserving length of a vector when rotated in texture coordinates (since the texture width & height may not be the same).

NOTE: The length of a vector on rotation is need not be preserved in texture coordinates. For example, on rotating the vector (.1, 0) by 90 degrees on a 512x256 makes it (0, .1). However, to preserve the length it should become (0, .2).

eXponenta:
Please! Don't change fields name and not change API. Backwards capability is important!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SukantPal:
> Please! Don't change fields name and not change API. Backwards capability is important!
> 
> Sent with [GitHawk](http://githawk.com)

Nice review! This change is supposed to be backward compatible, what breaking changes are you talking about. Being a bit specific would help.

ivanpopelyshev:
The scope of changes is big enough to be merged in 5.3

What do I see at first pass:
1. `getFilterTexture()` is used in sooo many demos, that we really can't deprecate it.
2. clamp on  `vTextureCoord` in vertex shader is very strange, it will stretch things because its applied before interpolation. Yes, clamp in fragment shader eats up to 1.5x fragment shader speed, because we modify UV's before we pass them to texture2D, but there are no other solutions.
3. I agree with @eXponenta that we need to separate renames from new functionality, we can discuss new names in separate PR after we finish this one.

Good things: 
Filter can affect bounds, just `padding` is not enough for sure.

I'm gonna make many tests for this thing, because last time someone changed filterSystem drastically - it was @GoodBoyDigital - he broke compatibility between v4 and v5 filters completely.

I agree that better filterSystem has to be focus of 5.3. I have many ideas from my work project.

SukantPal:
@ivanpopelyshev Which names are we talking about?

SukantPal:
@ivanpopelyshev Also, I don't have a problem if we don't clamp. It is just that it is usually not expected to have a texture coordinate that is outside the input given. If the filter's output frame fits inside its input frame, then clamping has no effect.

SukantPal:
@ivanpopelyshev Also, if a filter measures a different input than the required output, `getFilterTexture` is ambiguous. I can make a change so that message does not get printed when the filter is "legacy", i.e. it doesn't override the `onMeasure` method.

ivanpopelyshev:
> @ivanpopelyshev Also, I don't have a problem if we don't clamp. It is just that it is usually not expected to have a texture coordinate that is outside the input given. 

0.5 is usually used in fragment shader.  In your code, corners will be clamped to centers and that leads to stretch from (0 - w) to (0.5 - (w-0.5))

It can be outside if inside is a filter with autoHit=false.

packages/core/src/filters/Filter.js
```diff
@@ -201,6 +198,63 @@ export class Filter extends Shader
...
+     * @param {PIXI.Rectangle} passBounds - frame in which output is required
+     * @param {number} padding - padding applied in the target bounds
+     */
+    measure(targetBounds, passBounds, padding)
+    {
+        this.onMeasure(targetBounds, passBounds, padding);
+
+        if (this.frame === null || this.frame === undefined)
+        {
+            throw new Error(`${this.constructor.name}#onMeasure does not set Filter#_frame.`);
```
bigtimebuddy: Could you verify that `this.constructor.name` is valid in the minified version? Sometime class names get mangled with minification.

themoonrat: Even if it doesn't get mangled here, it might get mangled if pixi is minified in other build environments. 

But since it's for logging an error only so nothing relies on it... might be fine?

SukantPal:
> I'm not ready to approve this yet, need to separate into multiple PR's.
> Changes in shaders are a bad thing here.
> Measure is a good idea but I need more time to check it.

@ivanpopelyshev I agree, I'll split this into multiple pull requests.

PR #6258, opened 2019-12-03
Little shader speed up!

GoodBoyDigital:
```getShaderParameter``` is quite expensive and we only really need to call it when something has gone wrong.

A small change, but results in shaders building about 10% - 15% faster!


bigtimebuddy:
@ivanpopelyshev rebase isn't super necessary because I'm going to squash it anyways.

PR #6261, opened 2019-12-04
Converts display package to typescript

Zyie:
**UPDATE!!**

Now that #6311 has fixed the build issues this PR is ready to have its time to shine.

This PR converts the display package to typescript.

**CHANGES:**
- `DisplayObject` is now an abstract class.
- `tempDisplayObjectParent` is now a new class called `TemporaryDisplayObject`. This simply extends `DisplayObject` as you cannot instantiate an abstract class.
- Inside `Renderer.js` ive added the systems used in display to be `this.SYSTEM` instead of being dynamically added. @ivanpopelyshev  I've added a TODO as a note that this was hacked in since you are converting core. 
- Removed `this.rect` from `Bounds` as it was not being used.
- Made `_parentID` public in `Transform`

I've also seen that #6304 and ##6306 has made some changes to display. Not sure whats going to be the best way to navigate this.


ivanpopelyshev:
Yes, DisplayObject is meant to be abstract, to separate properties of this element, and properties that depend on children.
Line 469 : type is DisplayObject.or Container. MaskData was introduced by me in 5.2.0, its for people who need more control on masks.

EvidentlyCube:
I've got similar issue as the first one you mentioned but with `@pixi/constants`, so I am curious how to handle it.

ivanpopelyshev:
OK, I've started `@pixi/core` conversion, renamed files to TS, started converting `texture/resources/Resource.ts` and stumbled across `Runner` import. I don't feel like adding stuff to `global.d.ts`. It was fine in first modules, but here we'll have to import too many things, especially from `@pixi/math` module.

I think its time to figure out how to generate `d.ts` files for a few converted modules to use instead `global.d.ts`

EvidentlyCube:
@ivanpopelyshev A temporary solution might be to add the following line to `package.json` of each module that's already been converted to TypeScript:
```
  "types": "src/index.ts",
```

I'll tag you in a conversation I had on slack.

englercj:
Ideally Runner provides typescript types so the import just works.

packages/display/src/Container.ts
```diff
@@ -518,7 +531,7 @@ export class Container extends DisplayObject
      * @protected
      * @param {PIXI.Renderer} renderer - The renderer
      */
-    renderAdvanced(renderer)
+    renderAdvanced(renderer: Renderer): void
```
lekzd: protected

global.d.ts
```diff
@@ -1,5 +1,20 @@
...
+
+declare interface IHitArea {
+    contains(x: number, y: number): boolean;
+}
+
+declare interface InteractiveTarget {
+    interactive: boolean;
+    interactiveChildren: boolean;
+    hitArea: IHitArea;
+    cursor: string;
```
lekzd: maybe add something from here? https://github.com/frenic/csstype/blob/019f3524660f76764cded3fd4e429a353ef3464b/index.js.flow#L4359

laino:
> I've also seen that #6304 and ##6306 has made some changes to display. Not sure whats going to be the best way to navigate this.

You can ignore those. One is a simple bugfix (a bug that would be caught by TypeScript btw.), and the other is just something I wanted to get some input for.

Both are pretty easy to rebase on/rebase on this.

global.d.ts
bigtimebuddy: Good idea @lekzd

Zyie: Don't know how i didn't see this. Added it now.

packages/display/src/Container.ts
```diff
@@ -453,10 +461,12 @@ export class Container extends DisplayObject
...
             {
-                const maskObject = child._mask.maskObject || child._mask;
+                const maskObject = (child._mask as MaskData).maskObject || (child._mask as Container);
 
-                maskObject.calculateBounds();
-                this._bounds.addBoundsMask(child._bounds, maskObject._bounds);
+                // TODO: maskObject is getting PIXI.DisplayObject and DisplayObject confused.
+                // Once types are sorted we wont need to do this weird conversion
+                (maskObject as DisplayObject|Container).calculateBounds();
+                this._bounds.addBoundsMask(child._bounds, (maskObject._bounds as Bounds));
```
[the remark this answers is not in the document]
Zyie: I remember this happening to me. I believe i ran `npm run types` and the issue cleared up.

ivanpopelyshev:
All resolved, good job.

bigtimebuddy:
@Zyie could you resolve these conflicts? I think we're ready to merge.

ivanpopelyshev:
I see problems with _updateID, gonna fix them for you, its my area of expertise anyway. Maybe I'll add a test.

Zyie:
> @Zyie could you resolve these conflicts? I think we're ready to merge.

All sorted

PR #6262, opened 2019-12-04
Utils to typescript

EvidentlyCube:
##### Description of change
This is a draft PR for converting `utils` package to Typescript. Posting it early to make it visible that someone is already working on this package.

List of *subsections* in the package and their status:
- [x] `src/browser/*`
- [x] `src/color/*`
- [x] `src/data/*`
- [x] `src/logging/*`
- [x] `src/media/*`
- [x] `src/network/*`
- [x] `src/*`
- [x] `src/index.js`

##### Pre-Merge Checklist


ivanpopelyshev:
OK, I'm watching it!

EvidentlyCube:
I'm stuck on two errors that happen when I convert `index.js` to `index.ts`:

```
packages/utils/src/index.ts:60:41 - error TS2307: Cannot find module 'eventemitter3'.

60 export { default as EventEmitter } from 'eventemitter3';
                                           ~~~~~~~~~~~~~~~

packages/utils/src/index.ts:85:10 - error TS2305: Module '"url"' has no exported member 'default'.

85 export { default as url } from 'url';
            ~~~~~~~
```

All the other commits were pushed, if you change the file extension to `.ts` you'll reproduce the errors (didn't want to commit broken code). My thoughts on the two issues:

#### Event emitter
I've twiddled a bit with it but have no idea - it's in `node_modules`, it's in `package.json` but for some reason TS just won't see it. Maybe it's some property of `lerna` that I do not understand. Adding `@types/eventemitter3` didn't help either.

#### Url
Replacing it with:
```
import * as url from 'url';
export { url };
```

Works, but since TS won't build because of eventemitter I don't know if it truly works as intended.

EvidentlyCube:
Setting `moduleResolution` to `node` fixes this issue (got this suggestion [here](https://stackoverflow.com/questions/43964481/typescript-cannot-find-node-module-with-index-d-ts)) but then, naturally, other things fail. And when I change some things that they don't fail... it fails even more.

I think I don't know what to do here 🤔. Changing that one flag will result in a cascade of problems. Maybe there's a smart way to make it work without changing the flag, but I am out of ideas.

ivanpopelyshev:
I don't know if `Bound` is a good name here. How do we name interfaces with `top`, `left`, `bottom`, `right`? https://docs.oracle.com/javase/7/docs/api/java/awt/Insets.html - Insets!

EvidentlyCube:
@ivanpopelyshev Considering that interface is only used internally by that function I'd say it's fine to leave it there - but I'll rename it to `Inset`, it's a way better nam!

EvidentlyCube:
Alright, I undrafted before fully confirming everything works. Story time:

---

The problem I had with the module resolution was fixed by enabling `esModuleInterop` in `tsconfig`. I also got another suggestion in slack, to change the code to:
```
import EventEmitter = require('eventemitter3');
export { EventEmitter};
```

But it failed complaining of:
```
[!] Error: Export 'EventEmitter' is not defined (Note that you need plugins to import files that are not JavaScript)
packages\utils\src\index.ts (52:9)
50:
51: import EventEmitter = require('eventemitter3');
52: export { EventEmitter }
```

---

Okay, so that made the rollup pass and I pushed the changes and toggled draft status and then I realized `tsc -p tsconfig.docs.json` fails:

```
C:\Projects\Personal\JS\pixi.js>tsc -p tsconfig.docs.json
packages/utils/src/index.ts:51:26 - error TS2307: Cannot find module 'eventemitter3'.

51 import EventEmitter from 'eventemitter3';
                            ~~~~~~~~~~~~~~~
```

When I added `esModuleInterop` to `tsconfig.docs.json` it changed nothing. When I changed target to `es5` it started failing a lot with `MiniSignals` so it's likely not the way to go. 

---

At the moment I am again not sure what to do. Would be awesome if there was a simple fix but I am getting the impression that maybe `eventemitter3` can't be used the way its definition file is structured. But that's a pessimistic I-should-have-been-in-the-bed-for-an-hour thinking :). Plz help.

ivanpopelyshev:
I think we need our own definitions for eventemitter3, maybe just clone it completely. 

@cursedcoder, we summon you!

EvidentlyCube:
🥂
Thanks to everyone in the slack who helped me deal with the weirdness of module importing when CommonJS is used.

I won't repeat the details of the issue, since it's all in a comment in `utils/index.ts`. If this is merged the way it is I suppose it'll be a good idea to create an issue to ponder how to best solve it long term. @cursedcoder suggested custom typings if nothing better can be found and I'd be tempted to agree, though this has proven my understanding of modules and importing and requires is severely lacking.

---

I am now ready for the code reviews :).

ivanpopelyshev:
Do we really have to declare classes in `global.d.ts` ? If `interface` works there, it should be better.

EvidentlyCube:
> Do we really have to declare classes in global.d.ts ? I usually use interfaces when I show parts of class.

I think it's the best place at the moment. Once `core` is converted to TS that code will be redundant, but until then we need the definitions somewhere and it's better to have them in a globally accessible place rather than having copies spread around.

And from the way I understand the #6186 `global.d.ts` is (possibly) only used temporary.

ivanpopelyshev:
I mean, you can use `interface` instead of `class` there :)

packages/utils/src/data/removeItems.ts
```diff
@@ -7,7 +7,7 @@
  * @param {number} startIdx starting index
  * @param {number} removeCount how many to remove
  */
-export function removeItems(arr, startIdx, removeCount)
+export function removeItems(arr: any[], startIdx: number, removeCount: number): void
 {
     const length = arr.length;
```
lekzd: `arr.splice(startIdx, removeCount)` ?

packages/utils/src/logging/deprecation.ts
```diff
@@ -1,5 +1,5 @@
 // A map of warning messages already fired
-const warnings = {};
+const warnings: {[key: string]: boolean} = {};
```
lekzd: `const warnings = new Set<string>();`

packages/utils/src/data/removeItems.ts
bigtimebuddy: Splice creates garbage, this method does not.

EvidentlyCube:
@ivanpopelyshev Well.... You see... Hm... I mean my logic was "Those are originally classes so I should declare them as classes" but I guess it doesn't really matter. I am not sure if changing that to interface provides any benefit, but since I don't have any argument against it I am forced to concur and make the change :).

packages/utils/src/color/hex.ts
```diff
@@ -30,12 +30,11 @@ export function hex2rgb(hex, out)
  * @param {number} hex - Number in hex (e.g., `0xffffff`)
  * @return {string} The string color (e.g., `"#ffffff"`).
  */
-export function hex2string(hex)
+export function hex2string(hex: number): string
 {
-    hex = hex.toString(16);
-    hex = '000000'.substr(0, 6 - hex.length) + hex;
+    const hexString = hex.toString(16).padStart(6, '0');
```
bigtimebuddy: Please don’t use padStart, it breaks our es5 compatibility and we don’t want to polyfill this function.

EvidentlyCube:
@bigtimebuddy Done!

PR #6264, opened 2019-12-05
Feature/more accessibility updates

GoodBoyDigital:
##### Description of change
Some more tweaks to ensure **Pixi.js is the most accessible WebGL renderer in the land!** 🥇 
- added fix for android quirk (refreshing divs at 60fps messes the index order)
- added fix for resolution of renderer not being taken into account for accessibility layer
- added fix where if the div elements total became zero, the accessibility manager was being permanently deactivated (Even if new ones were added)
- made copy on hook div be a little more helpful

##### Pre-Merge Checklist


packages/accessibility/src/AccessibilityManager.js
```diff
@@ -270,6 +278,15 @@ export class AccessibilityManager
      */
     update()
     {
+        /* On Android, tab order seems to be calculated by position rather than tabIndex,
```
themoonrat: When you say Android, is this all browsers on Android? Or Chrome? Firefox? With hacky things like this, it'd be nice to know a more precise scenario to know why this exists.

packages/accessibility/src/AccessibilityManager.js
```diff
@@ -270,6 +278,15 @@ export class AccessibilityManager
...
+        ** moving buttons can cause focus to flicker between two buttons making it hard/impossible to navigate,
+        ** so I am just running update every half a second, seems to fix it
+        */
+        this.androidUpdateCount++;
+        this.androidUpdateCount %= 30;
```
themoonrat: Would it be possible to do this time based rather than tick based. If it's really it being refreshed too much, then as more 90hz / 120hz screens come out over the next year, the ticks will go up and we'll have the same issue again.

packages/accessibility/src/AccessibilityManager.js
GoodBoyDigital: As far as I know, this is the standard web browser for all android devices post (4.4) that we tested on. Will amend to reflect that!

packages/accessibility/src/AccessibilityManager.js
GoodBoyDigital: good shout, will tweak this one..

ivanpopelyshev:
For good luck, I advice to feed codecov bot with a test or multiple.

PR #6273, opened 2019-12-09
Corners supposed to be proportionally scaled

zaynyatyi:
##### Description of change
If nine slice object on scene is smaller than frame it shouldn't be distorted

https://jsfiddle.net/Heorhiy/ex35ty0m/

##### Pre-Merge Checklist


bigtimebuddy:
@zaynyatyi Thanks for this! The min scale looks much nicer. Couple of things:

* Could you please resolve the linting errors
* Could you provide a `@private` JSDoc for the `_getMinScale` method
* Could you add a unit test to test for values from `_getMinScale`

zaynyatyi:
Thanks guys :)

Issue #6288, opened 2019-12-19
Feature Support: Renderbuffer multisampling support in WebGL2 machines


SukantPal:
I think we should start supporting multisampling in our base render-textures. This would be great for the filter system: by using multisampling instead of rendering at higher resolutions, we could improve perfs.

ivanpopelyshev:
Its architectural nightmare. Support of MSAA isn't enough for real projects, no one uses it even in threejs projects.

I have a draft plugin here, but i didnt port examples from my work to it: https://github.com/gameofbombs/pixi-blit/

It has two fallbacks for webgl1, but my current implementation is slow at work. I'm trying to make it work on real project, but we need shapes cache to make it super-fast on all browsers.

When its 100% super-ready I'll put it to pixi.

SukantPal:
@ivanpopelyshev What do you mean by "shapes cache"?

SukantPal:
@ivanpopelyshev I think you got the architectural side wrong in `pixi-blit`. A renderbuffer is not supposed wrap around a texture. Note that they are supported in WebGL 1, but only that multi-sampling isn’t available. I am working on creating a slightly tweaked addition of your work in pixi.js by creating a `RenderbufferSystem`. 

Then a frame buffer could have a renderbuffer attached, rather than only color textures. This system would be more robust - we could separate the “blit” part of the process from render-textures and framebuffers.

ivanpopelyshev:
Just one ability to use MSAA is not enough because it seriously affects memory. and performance from MSAA is still not enough, well, its better than just x3 scale but bad anyway.

You have to detect all shapes that need AA in this frame, render them once in AA-ed buffer and then do one blit operation. If one buffer is not enough - do it multiple times with multiple blits, but result is collection of atlases.

I didnt put my current atlases in this because of one thing: even that is not enough. I'm working on better solution.

Your blitting API is good enough but i have an issue with it: people cant use it on apple devices and they'll never be. Without fallbacks its all just a chunk of academical code.  I don't know what to do with this one, because  yes, it can be merged, but what if it would be just dead piece of code like in threejs?

Lets ask @GoodBoyDigital.

eXponenta:
This is will be dead code for Apple platform.
Multisampling doesn't must be inside core yet, i think.
It increase package size without profits.

And PIXI force webGL 1 on any mobile devices now.

SukantPal:
[...]

PR #6290, opened 2019-12-20
Adds Renderbuffer support & multisampling for WebGL 2

SukantPal:
##### Description of change
I'm trying to add support multisampling render buffers in pixi.js; in that process, I wanted to demonstrate the change's merit using JSFiddle demos.

This is related to #6288 .

- [x] `Renderbuffer` + `RenderbufferSystem`: These manage the binding of WebGL render buffers to their contexts.

- [x] Add multisampling support

(For #6288)

##### Pre-Merge Checklist


SukantPal:
I got a (partially) working example: https://jsfiddle.net/ShukantPal/3qf6dr7b/.

The problem is some sort of y-flipping because of the blit. If you an idea, please help me figure out why.

packages/core/src/framebuffer/FramebufferSystem.js
```diff
@@ -169,6 +169,44 @@ export class FramebufferSystem extends System
...
+     * Copies the pixels in the source color buffer to the destination
+     * color buffer. This creates two temporary framebuffers to complete
+     * the operation.
+     *
+     * @param {PIXI.ColorBuffer} srcBuffer
+     * @param {PIXI.ColorBuffer} dstBuffer
+     */
+    blitColorBuffer(srcBuffer, dstBuffer)
+    {
+        const srcFb = new Framebuffer().addColorBuffer(srcBuffer);
```
lekzd: Better way is to use static methods for creating new instances with options

`const srcFb = Framebuffer.fromColorBuffer(srcBuffer);`

packages/core/src/framebuffer/FramebufferSystem.js
```diff
@@ -254,12 +305,13 @@ export class FramebufferSystem extends System
...
     /**
-     * Resize the framebuffer
+     * Resizes the bound framebuffer
      *
      * @protected
      * @param {PIXI.Framebuffer} framebuffer
+     * @param {number} [target=this.gl.FRAMEBUFFER] - binding point for `framebuffer`
      */
-    resizeFramebuffer(framebuffer)
+    resizeFramebuffer(framebuffer, target = this.gl.FRAMEBUFFE) // eslint-disable-line no-unused-vars
```
lekzd: target = this.gl.FRAMEBUFFE**R**

SukantPal:
Hi, @lekzd. Thank you for reviewing my pull request and pointing out my mistakes. I’ll follow up ASAP.

SukantPal:
@GoodBoyDigital This is an example you were asking for: https://jsfiddle.net/ShukantPal/3qf6dr7b/64/. It is a fairly low-level usage of the frame buffer API - nevertheless, it uses multisampling at 4x resolution.

I am thinking about how to integrate a `MSAARenderTexture` -  a simple replacement for `RenderTexture` when multisampling is preferred.

packages/core/src/framebuffer/FramebufferSystem.js
```diff
@@ -285,12 +364,13 @@ export class FramebufferSystem extends System
     }
 
     /**
-     * Update the framebuffer
+     * Updates a bound framebuffer
      *
      * @protected
      * @param {PIXI.Framebuffer} framebuffer
+     * @param {number} [target=this.gl.FRAMEBUFFER] - binding point for `framebuffer`
```
GoodBoyDigital: mind specifying the 3 types here for clarity?
read / write / framebuffer?

packages/settings/src/settings.js
```diff
@@ -62,6 +62,17 @@ export const settings = {
...
+    /**
+     * Override filter resolution by actually using a multisampling renderbuffer.
+     *
+     * @static
+     * @name FILTER_MULTISAMPLING
+     * @memberof PIXI.settings
+     * @type {boolean}
+     * @default true
+     */
+    FILTER_MULTISAMPLING: true,
```
GoodBoyDigital: having this set to true by default, will this have a performance impact? If so, we should default to false.

packages/core/src/renderbuffer/RenderbufferSystem.js
```diff
@@ -0,0 +1,203 @@
...
+         * @private
+         */
+        this.managedRenderbuffers = [];
+
+        /**
+         * Renderbuffer value that tells us we don't know what is bound.
+         *
+         * @member {PIXI.Renderbuffer}
+         */
+        this.unknownRenderbuffer = new Renderbuffer(10, 10);
```
GoodBoyDigital: Not a biggee, but do we need this? Could we not always have the main screen bound by default?

GoodBoyDigital:
Hey @SukantPal, What I would love to see is a few examples added to our examples page to show how this can be used in a few different ways!

SukantPal:
@GoodBoyDigital Thanks for reviewing this, I’ll try to get your requests resolved by the end of today (Eastern Time :)

packages/core/src/framebuffer/FramebufferSystem.js
SukantPal: I added these in @pixi/constants.

packages/core/src/renderbuffer/RenderbufferSystem.js
SukantPal: The unknown renderbuffer means you don’t know what Renderbuffer is attached - if it is the screen, you have to put null. When we initialize the context, I’m assuming it as unknown.

This would also be useful when accessing the renderbuffer directly in webgl and then telling the system that it should set to unknown.

packages/settings/src/settings.js
SukantPal: Oh, this was a vestigial part - since I’m not actually adding renderbuffer usage in the filter system. That’ll be a part of the filter system upgrade PR.

GoodBoyDigital:
closing in favour of #6296 

It needs to be noted that this is a really solid PR and very well structured.

Then reason for favouring the #6296 is that this feature is considered edge case / experimental and as such it is felt that there is the extra code does does not quite match the feature importance and perhaps the flexibility it brings does not yet offer much value to end users.

Whilst #6296 is less flexible it achieve the main MMSA feature in less code. 

Thanks again for this PR @SukantPal , it clearly shows you have serious knowledge about webGL and how to architect it in a solid way. Will certainly come in handy for other parts of the project! 👍

PR #6293, opened 2019-12-22
Diagnostic for provided webgl context

ivanpopelyshev:
According to @lekzd, Catalina sometimes spawn a context that is already lost. I honestly don't know what to do if `getContextAttributes()` returns null, and whether we can work like that, so lets discuss it.

For now I can try to add two more diagnostic messages.

### DIFF
--- a/packages/core/src/context/ContextSystem.js
+++ b/packages/core/src/context/ContextSystem.js
@@ -249,11 +249,18 @@ export class ContextSystem extends System
     {
         const attributes = gl.getContextAttributes();
 
-        if (gl.isContextLost() || !attributes)
+        if (gl.isContextLost())
         {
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Dont merge it yet, I found that my knowledge about our methods handling context loss and restore is wrong. I'm gonna add tests on that thing.

ivanpopelyshev:
@lekzd did you find whats wrong in your case?

lekzd:
@ivanpopelyshev something happening after Stage resize, but I have no time to investigate more before NY holydays.

bigtimebuddy:
Any update on this @ivanpopelyshev?

ivanpopelyshev:
I'm closing, its not a priority before we actually find the system and case that can be reproduced

Issue #6257, opened 2019-12-02
Pixi BlurFilter showing interference from other sprites.


gregorymark:
Hi, I'm having problems blurring images with transparent backgrounds. I'm animating multiple sprites with transparent backgrounds. The edges of each sprite seem to pick up the colour from the sprites around them as they move. The colour doesn't always correspond to the sprite itself, so I'm assuming it picks up the other sprites' colours somehow.

I'm running this on Pixi 5.1.5 on Chrome 78, Mac Mojave. Unfortunately I can't show a link as it's a client project I can't share, but I've attached a screenshot of the effect, which I'm hoping will help identify.

Some excerpts of what I'm hoping are the relevant bits of code. Please let me know if anything else would be useful!

Initialising:
```
// Create a Pixi Application
this.pixi.app = new this.pixi.Application({
    antialias: true,
    transparent: true
});

// Add the Pixi created canvas to the HTML document
this.stage.appendChild(this.pixi.app.view);

// Canvas setup
this.pixi.app.renderer.autoDensity = true;
this.pixi.app.renderer.resize(window.innerWidth, window.innerHeight);

// Add sprites
this.pixi.loader.add(this.shapeURLs)
    .load((loader, resources) => this.setupShapes(loader, resources));
```
Setting up sprites:
```
this.pixi.container = new this.pixi.Container();

for (let i = 0; i < this.requiredShapes; i++) {
    let shapeSprite;
    const texture = resources[`${this.shapesDirectory}/shape${this.shapes[i].shapeNum}-blur-0.png`].texture;
    shapeSprite = new this.pixi.Sprite(texture);
    shapeSprite.anchor.set(0.5, 0.5);
    shapeSprite.resolution = 2;
    shapeSprite.alpha = 0;

    this.shapes[i].filter = new this.pixi.filters.BlurFilter(8);
    shapeSprite.filters = [this.shapes[i].filter];

    this.shapes[i].sprite = shapeSprite;
    this.pixi.container.addChild(shapeSprite);

    shapeSprite.scale.set(this.minScale, this.minScale);
}

this.pixi.app.stage.addChild(this.pixi.container);
```
and from the render function:
```
this.shapes[i].cycleTime = timestamp - this.shapes[i].cycleStartTime;

// Fade the shape in right at the start
const alphaVal = this.shapes[i].cycleTime / 1000 > 1 ? 1 : this.shapes[i].cycleTime / 1000;
[...]

PR #6295, opened 2019-12-23
Change filter clear parameter to clear modes

ivanpopelyshev:
Previous issue #6270 #6257 and possible future changes in FilterSystem enabled me to finally clarify the `clear` variable everywhere.

BlurFilter test for this branch: https://www.pixiplayground.com/#/edit/0-VaT0fFvoiWju_9XeJO2

Modes:
`NO`/`BLEND` (false) - don't clear the output texture, it contains vital information that we need to blend upon
`YES`/`CLEAR` (true) - texture contains junk, clear it.
`AUTO`/`BLIT` (new mode!) - we know that texture partially or fully cleared and previous state should not be a problem.

By default pixi-v5 uses `renderer.filter.forceClear=false`, which corresponds to minimal number of WebGL commands.

The trick is - on some mobile videos its better to clear just after bind , because it saves the time for copying content.  In that case we need `renderer.filter.forceClear=true`.

Another case is when someone made a mistake, like it existed in BlurFilter, but we dont have time to debug all those custom filters. In that case we just clear whenever its possible.

As a joke, I thought it was possible to add one more "russian" mode, `YES_NO_MAYBE` (да нет, наверное). If filter area covers all renderTexture, we might not clear it borders and just blit it.


ivanpopelyshev:
Yep :) I'll do experiments which devices have this problem. 

Right now its just a paranoia option: v4 cleared everything, v5 doesnt and we have a switch between those modes.

bigtimebuddy:
How are you supporting backward compatibility? I’m sure there are some pixi-filters using clear Boolean? I’d suggest adding a typeof check to clear and throwing a deprecation warning.

ivanpopelyshev:
Easy.
0=false, 1=true - behaviour is the same.

2 is not used in pixi-filters

bigtimebuddy:
@ivanpopelyshev we should not use truthy conditions with constants like CLEAR_MODES. This is how we got bit with the scale modes, because someone flipped the value. Please don’t mix types. If someone uses a boolean, we should make sure that’s properly deprecated.

ivanpopelyshev:
We will get rid of booleans in pixi-filters and only numbers will remain. I just don't want the plugin to break right after we merge this thing.

We won't have problems like that in future with typescript.

packages/core/src/filters/FilterSystem.js
```diff
@@ -326,26 +332,34 @@ export class FilterSystem extends System
...
 
+    /**
+     * Binds a renderTexture with corresponding `filterFrame`, clears it if mode corresponds.
+     * @param {PIXI.RenderTexture} filterTexture renderTexture to bind, should belong to filter pool or filter stack
+     * @param {PIXI.CLEAR_MODES} [clearMode] clearMode, by default its CLEAR/YES. See {@link PIXI.CLEAR_MODES}
+     */
+    bindAndClear(filterTexture, clearMode = CLEAR_MODES.CLEAR)
+    {
+        this.renderer.renderTexture.bind(filterTexture, filterTexture ? filterTexture.filterFrame : null);
+        if (clearMode === CLEAR_MODES.CLEAR
```
bigtimebuddy: How does this work if `clearMode` is `true`?

ivanpopelyshev:
OK, I've added support for `true` and added a test. I completely forgot that I use `===` and `true!==1`

packages/core/src/filters/FilterSystem.js
```diff
@@ -326,26 +332,35 @@ export class FilterSystem extends System
...
+            || (clearMode === CLEAR_MODES.BLIT && this.forceClear)
+            || clearMode === true)
```
bigtimebuddy: This is an incorrect/incomplete way to implement backward compatibility. The API shouldn't support both booleans and CLEAR_MODES. Please move this condition out of the if statement so that we can easily remove it later. For instance:

```js
// TODO: remove in next major version
if (typeof clearMode === 'boolean') {
  clearMode = clearMode ? CLEAR_MODES.CLEAR : CLEAR_MODES.BLEND;
  // get deprecation function from utils
  deprecation('5.2.1', 'Use CLEAR_MODES when using clear applyFilter option');
}

```

ivanpopelyshev:
> Thank you for appeasing me!

You know that I can't see everything. Btw, tests show deprecation in console - can we do something about it?

bigtimebuddy:
If you see deprecation in console it means we missed a spot for upgrading. Could be a test? Probably worth tracking down.

ivanpopelyshev:
Both tests are mine: one uses BeginTextureFIll with arguments instead of object, and one here, I'm testing bool param.

PR #6296, opened 2019-12-25
Multisampling in FrameBufferSystem

ivanpopelyshev:
My minimal implementation for #6288

Demo: https://www.pixiplayground.com/#/edit/ZbmXFH7nt11L70Ux2jUy8

## 🎁 Added

- `Framebuffer.#multisample`  - number of samples that user wants. By default its 0 so shortcut `if (multisample) ` can be used. 

- Extra operation `blit` that has to be called after `renderer.render` if you think that msaa is enabled on renderTexture. Operation corresponds to low-level `gl.blitFramebuffer` and requires extra internal framebuffer object to bind the texture.

## How it works

API is secure - It won't fail in case of WegGL1 or if there's not enough samples.

Internally, `colorTextures[0]` for MSAA renderbuffer, so, theoretically, MRT still should work. I didn't test it though. However, `colorTextures[0]` still exists - that's where we blit the data for later usage.

## Why not RenderBuffer object

1. MSAA is slow and eats extra memory.
2. WebGL2 doesn't exist in Safari.
3. I didn't see ThreeJS projects that use MSAA in production.

I do not want to add new objects and abstractions in vanilla PixiJS for a case that cant be used in production.

However, it is good as a temporary solution for demos and prototyping games. Basically, its a Toy.

Two-three hundred lines of code are not much, we just mirror WebGL2 API a bit so people can experiment how it looks.

## Future:

Both PR's should be small.

### RenderTextureSystem

Add high-level `blit` operation, that can will either use framebuffer blit, either texture copy either scaling filter, that will enable trivial WebGL1 fallback. 

### FilterSystem

We can enable usage of custom texture pools per object, specify that we need particular renderTexture for dealing with a container, and that texture can have multisampling.

## What cant be in vanilla:

MSAA is slow. I don't believe anyone will use it in production even after those PR's.

I'm working on a real solution that can be used in production by heavy projects:

1. real WebGL1 antialiasing fallback
2. caching system with atlases
3. several antialiasing options for containers / graphics
4. solution for strange conflation in MSAA : 
image

If this PR will be merged, I won't need to add framebuffer hacks in the plugin.

### Another plugin

I hope that this thing covers @SukantPal case for filters, maybe not right now but after next  FilterSystem PR.


SukantPal:
Also, "4. Strange conflation with MSAA" - I think that is because you are not clearing the frame buffer with the background before rendering to it.

ivanpopelyshev:
Oh, no, macs, my bane! I hate them really, I dont have one. I'll ask for a testing device on work tomorrow.

There shoulnd't be a problem with clearMode, because its an input texture that should be cleared anyway. Maybe I'll add `clear` there in case of `forceClear` just to be compliant.

Exactly, case 4 is done when something is already drawn, and the problem is that users will do it anyway. We need a way to mitigate the damage, my solution is to apply AA to all shapes befure they start combining them.

SukantPal:
@ivanpopelyshev Why are using resolution=.125 in your demo?

eXponenta:
MacOS/ iOS WebGL2 support is worst.

SukantPal:
> MacOS/ iOS WebGL2 support is worst.

@eXponenta The other PR works on my machine (the one with Renderbuffer classes). It is actually a problem with this PR. @ivanpopelyshev What machine are you testing on, does it have WebGL 2?

eXponenta:
His machine has fully WebGL2 support.

SukantPal:
@eXponenta WebGL 2 support depends on the browser I guess, not on the machine.

eXponenta:
It isn't fully true.

SukantPal:
@eXponenta Well, MacBooks of course have the graphics ability to support WebGL 2. The browsers are supposed to bind the WebGL API to a OpenGL backend.

ivanpopelyshev:
The heck is going on on apple devices? We'll know tomorrow when I get said devices from test team! Stay tuned!

ivanpopelyshev:
Shame on me, I had a mistake in `FOR` cycle! Added a unit test for it. Now it should work on mac devices that somehow support webgl2

ivanpopelyshev:
Need to test this thing on iOS 13.2 Safari  . @SukantPal whats your device?

SukantPal:
@ivanpopelyshev It works now! (macOS Catalina, 2018 i7 MacBook Pro).

packages/core/src/framebuffer/FramebufferSystem.js
```diff
@@ -82,6 +86,44 @@ export class FramebufferSystem extends System
...
+        }
+    }
+
+    /**
+     * Detects number of samples that is not more than a param but as close to it as possible
+     *
+     * @param {number} samples number of samples
+     * @returns {number} recommended number of samples
+     */
+    detectSamples(samples)
```
GoodBoyDigital: do we need this? Can we rather, just check values are PO2?

ivanpopelyshev: I really don't know what to put there. This thing works and is covered by a test. I don't know which values are available on different systems, I didn't research it.

SukantPal:
@bigtimebuddy @GoodBoyDigital I don't think we should really delay this any longer. If you want, we could finalize this pull request.

I just want you to consider whether we should require users to do a `blit` after rendering to a multisample render-texture. I think that'll be a big source of bugs.

We might also need to consider blend modes for multisample render-textures. Renderbuffer blits don't support "blending", and this will break some expected behaviors. Instead of blitting directly onto the "texture", we might need to blit into a secondary texture and then render that onto the actual render-texture (via a shader that will blend).

ivanpopelyshev:
> Instead of blitting directly onto the "texture", we might need to blit into a secondary texture and then render that onto the actual render-texture (via a shader that will blend).

This PR doesn't actually forbid a blitting with shader, it doesnt include it on low-level. I use extra filter for blitting in my fork, and I think it has to be part of high-level object, alongside with WebGL1 fallback and some other things.

SukantPal:
@ivanpopelyshev I don't understand what you said. "This PR doesn't forbid blitting with a shader". How? First of all, blitting can't be done by a shader (by definition). It just copying the pixels from a render buffer onto a texture (no blending, direct copy).

"I use extra filter for blitting in my fork, and I think it has to be part of high-level object" - that doesn't make sense.

All my complaint is that we're going to have to add a line after rendering to use multisample in PixiJS -

```js
app.renderer.render(circle, buf);
app.renderer.framebuffer.blit();
```

What is the problem if the blitting is handled as a part of the rendering pipeline. I don't think any user would expect to a blit after calling `render`. Do you have an example of a framework that uses these semantics? You said you wanted "users to know about the blit operation". Now @bigtimebuddy & @GoodBoyDigital will have to decide whether that reason is enough to separate blitting from the rendering process.

SukantPal:
@ivanpopelyshev I still don't understand why your example uses 1/8th resolution on the textures. That just nullifies the effect of multisampling.

packages/core/src/framebuffer/Framebuffer.js
```diff
@@ -32,6 +32,14 @@ export class Framebuffer
 
         this.glFramebuffers = {};
 
+        /**
+         * Asks framebuffer to use multi-sampling if available, 0 by default.
+         * See {@link PIXI.MSAA_QUALITY}
+         * @member {number}
+         * @default PIXI.MSAA_QUALITY.LOW
```
SukantPal: Why is the "default" `multisample` property `MSAA_QUALITY.LOW` and not `NONE`? I think the latter makes more sense.

ivanpopelyshev: Oops, should be NONE

ivanpopelyshev:
> I still don't understand why your example uses 1/8th resolution on the textures. That just nullifies the effect of multisampling.

To actually see that it works.

SukantPal:
What do you see on a 4K screen? If you’re on a 4K screen, it will still just render as normal.

On my screen, it has jagged edges and doesn’t look like a circle. You’re supposed to use 1x resolution so the multisample makes the effective resolution 8. The whole point is to multisample each pixel 8 times, not multisample 8 pixels 8 times.

packages/core/src/framebuffer/FramebufferSystem.js
GoodBoyDigital: I think we should just enforce the use of the consts? 

export enum MSAA_QUALITY {
    NONE = 0,
    LOW = 0,
    MEDIUM = 4,
    HIGH = 8
}

ivanpopelyshev:
> What do you see on a 4K screen? If you’re on a 4K screen, it will still just render as normal.
> 
> On my screen, it has jagged edges and doesn’t look like a circle. You’re supposed to use 1x resolution so the multisample makes the effective resolution 8. The whole point is to multisample each pixel 8 times, not multisample 8 pixels 8 times.

Pixels are half-red, that means AA is working :)

packages/core/src/framebuffer/FramebufferSystem.js
ivanpopelyshev: image

ivanpopelyshev: it looks like there's no 16: https://webglstats.com/webgl2/parameter/SAMPLES

Maybe its my stupid time - I don't know what to write with those IF's. There will be more code in that case.

adireddy:
# [Codecov](https://codecov.io/gh/pixijs/pixi.js/pull/6296?src=pr&el=h1) Report
> Merging [#6296](https://codecov.io/gh/pixijs/pixi.js/pull/6296?src=pr&el=desc) into [dev](https://codecov.io/gh/pixijs/pixi.js/commit/ad4230b32f03cea4076c01953e97b7237d049a20&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/pixijs/pixi.js/pull/6296?src=pr&el=tree)

```diff
@@           Coverage Diff           @@
##              dev    #6296   +/-   ##
=======================================
  Coverage   74.70%   74.70%           
=======================================
  Files         105      105           
  Lines        6053     6053           
=======================================
  Hits         4522     4522           
  Misses       1531     1531           
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/pixijs/pixi.js/pull/6296?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/pixijs/pixi.js/pull/6296?src=pr&el=footer). Last update [bcd040e...dadab0c](https://codecov.io/gh/pixijs/pixi.js/pull/6296?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

packages/constants/src/index.ts
```diff
@@ -409,3 +409,22 @@ export enum MASK_TYPES {
     STENCIL = 2,
     SPRITE = 3,
 }
+
+/**
+ * Constants for multi-sampling antialiasing
```
bigtimebuddy: I would like really more documentation about this feature being _experimental_ and that there's a significant performance cost paid for using it. Also, it's not super obvious, but a code example usage setting `renderer.framebuffer.multisample`.

packages/core/src/framebuffer/Framebuffer.ts
```diff
@@ -46,6 +55,13 @@ export class Framebuffer
         this.glFramebuffers = {};
 
         this.disposeRunner = new Runner('disposeFramebuffer');
+
+        /**
+         * Desired number of samples for antialiasing. 0 means AA should not be used.
```
bigtimebuddy: Again, a warning about performance would be :+1:

packages/core/src/framebuffer/GLFramebuffer.ts
```diff
@@ -7,6 +10,9 @@ export class GLFramebuffer
 {
     public framebuffer: WebGLFramebuffer;
     public stencil: WebGLRenderbuffer;
+    public multisample: number;
```
bigtimebuddy: Why isn't this `MSAA_QUALITY` type?

packages/core/src/framebuffer/GLFramebuffer.ts
```diff
@@ -41,5 +47,24 @@ export class GLFramebuffer
          * @protected
          */
         this.dirtySize = 0;
+
+        /**
+         * Detected AA samples number
+         * @member {MSAA_QUALITY}
```
bigtimebuddy:
```suggestion
         * @member {PIXI.MSAA_QUALITY}
```

SukantPal:
@ivanpopelyshev Should we have a fallback to higher resolution if multisampling is not supported on the device? When forcing WebGL 1, the half-red pixels do not show up.

SukantPal:
@bigtimebuddy @GoodBoyDigital I approved this - but I did have concerns about requiring a `blit` operation after `render`. There is also no documentation regarding this (is there?)

packages/constants/src/index.ts
ivanpopelyshev: linked it to Framebuffer#multisample. Don't want to ctrl+C everything.

ivanpopelyshev:
@SukantPal good catch! I put it in `PIXI.Framebuffer#multisample` and linked from `PIXI.MSAA_QUALITY`

bigtimebuddy:
Thanks everyone for getting this feature through!

PR #6302, opened 2020-01-03
Feature/bmpfont txt parser

maurodetarso:
##### Description of change
Added support for bmglyph txt-based bitmap font data alongside regular XML data, using the same bitmap font API and changing existing code as little as possible.

##### Pre-Merge Checklist


bigtimebuddy:
Good feature! Thanks for doing this.

Organizationally, a place for improvement is to separate the parsers and loaders for each format instead of having different methods for xml and text. Would make it nice and clean and make it easier for us to split our into different packages if need be. Also this way each format is fully encapsulated and we should add unit test for each individually.

GoodBoyDigital:
Nice idea that! Two separate loaders would make more sense!

bigtimebuddy:
@maurodetarso do you think you'll be able to address the feedback?

bigtimebuddy:
Closing this in favor of #6347

PR #6304, opened 2020-01-03
Add empty calculateBounds to DisplayObject, remove Bounds.updateID references

laino:
##### Description of change

Since ``DisplayObject`` uses ``DisplayObject.calculateBounds``, it should at least provide an empty implementation to be overwritten in child classes like ``Container``.

Also there was two places in which ``bounds.updateID++`` was used, which is a property that doesn't actually exist on bounds (anymore?). This just causes that property to be created and set to ``NaN`` and it thus was a silent error.

##### Pre-Merge Checklist

### DIFF
--- a/packages/display/src/Bounds.js
+++ b/packages/display/src/Bounds.js
@@ -56,8 +56,6 @@ export class Bounds
      */
     clear()
     {
-        this.updateID++;
-
         this.minX = Infinity;
         this.minY = Infinity;
         this.maxX = -Infinity;
--- a/packages/display/src/DisplayObject.js
+++ b/packages/display/src/DisplayObject.js
@@ -213,6 +213,8 @@ export class DisplayObject extends EventEmitter
      */
     updateTransform()
     {
+        this._boundsID++;
+
         this.transform.updateTransform(this.parent.transform);
         // multiply the alphas..
         this.worldAlpha = this.alpha * this.parent.worldAlpha;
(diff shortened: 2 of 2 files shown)


ivanpopelyshev:
Oh, I saw that too, I forgot to make a PR!

laino:
@ivanpopelyshev The bounds IDs are tracked elsewhere. This code was trying to update a ``updateID`` on the ``Bounds`` object.

Bounds IDs are actually tracked on ``DisplayObject`` in ``DisplayObject._boundsID`` and ``DisplayObject._lastBoundsID``.

I removed two instances of ``Bounds.updateID++``, which was the only two places in the entire code base to even refer to that property afaik.

laino:
Maybe it should be ``this._boundsID++`` **instead** of ``this._bounds.updateID++`` though.

I just removed it because I assumed it wasn't needed, since it's essentially a NOOP right now and there was no apparent issues...

laino:
It should probably be ``this._boundsID++`` like in ``Container``:

https://github.com/pixijs/pixi.js/blob/1de41d78b2bb694261f7057f48f5758c863b1898/packages/display/src/Container.js#L414

I'll fix that in a second.

ivanpopelyshev:
> https://github.com/pixijs/pixi.js/blob/bca4b623a532c79a890d34158ed8228244224d89/packages/display/src/DisplayObject.js#L216
> 
> You increase boundsID, but transform not necessarily will be updated.
> 

Its like that because we dont yet have code to track bounds, but we think that they are not updated if `updateTransform` wasnt called :)

laino:
``DisplayObject._currentBounds`` appears to be another unused property.

PR #6306, opened 2020-01-04
Includes filter padding in bounds calculation and fix filter padding calculation

laino:
##### Description of change

Fixes: https://github.com/pixijs/pixi.js/issues/6303

This makes paddings be added additively to ``Container`` bounds. This means we can get rid
of that calculation in ``FilterSystem`` and ``cacheAsBitmap`` etc.

Calculating bounds additively also fixes issues where an element has multiple filters that each need some padding as well as padding of nested filters.

This is definitely a breaking change because it affects things like ``container.width = w`` that use
bounds in their calculation.

##### Pre-Merge Checklist

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,5 @@
 {
+    "root": true,
     "env": {
         "es6": true,
         "browser": true,
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -463,7 +463,7 @@ export class Container extends DisplayObject
             }
         }
 
-        if (this.filters)
+        if (this.filters && this.filters.length)
         {
             let filterPadding = 0;
 
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -411,12 +411,7 @@ export class Container extends DisplayObject
             this.sortChildren();
         }
 
-        this._boundsID++;
-
-        this.transform.updateTransform(this.parent.transform);
-
-        // TODO: check render flags, how to process stuff here
-        this.worldAlpha = this.alpha * this.parent.worldAlpha;
+        super.updateTransform();
 
         for (let i = 0, j = this.children.length; i < j; ++i)
         {
@@ -466,6 +461,18 @@ export class Container extends DisplayObject
             {
                 this._bounds.addBounds(child._bounds);
             }
+        }
+
+        if (this.filters)
+        {
+            let filterPadding = 0;
+
+            for (let i = 0; i < this.filters.length; i++)
+            {
+                filterPadding += this.filters[i].padding;
+            }
+
+            this._bounds.pad(filterPadding);
         }
 
         this._lastBoundsID = this._boundsID;
--- a/packages/display/src/DisplayObject.js
+++ b/packages/display/src/DisplayObject.js
@@ -84,7 +84,7 @@ export class DisplayObject extends EventEmitter
         this.renderable = true;
 
         /**
-         * The display object container that contains this display object.
+         * The display object container that contains this DisplayObject.
          *
          * @member {PIXI.Container}
          * @readonly
@@ -92,7 +92,9 @@ export class DisplayObject extends EventEmitter
         this.parent = null;
 
         /**
-         * The multiplied alpha of the displayObject.
+         * The multiplied alpha of the DisplayObject.
+         *
+         * Updated via updateTransform()
          *
          * @member {number}
          * @readonly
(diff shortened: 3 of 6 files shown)


.eslintrc.json
```diff
@@ -1,4 +1,5 @@
 {
+    "root": true,
```
laino: This is just a small thing I need to add to avoid conflicts with my global ``.eslintrc``

eXponenta:
Don't create PR relative another PR. 
Rebase PR onto dev branch.

packages/display/src/Container.js
```diff
@@ -411,12 +411,7 @@ export class Container extends DisplayObject
             this.sortChildren();
         }
 
-        this._boundsID++;
-
-        this.transform.updateTransform(this.parent.transform);
-
-        // TODO: check render flags, how to process stuff here
-        this.worldAlpha = this.alpha * this.parent.worldAlpha;
+        DisplayObject.prototype.updateTransform.call(this);
```
bigtimebuddy: Why not just call `this.updateTransform`?

laino: We are *in* ``this.updateTransform``. That'd be infinitely recursive.

There used to be a bunch of copy-pasted lines from ``DisplayObject.updateTransform`` here, which weren't really kept in sync thus lead to buggy code when people only updated it in one file. I fixed that here: https://github.com/pixijs/pixi.js/commit/bca4b623a532c79a890d34158ed8228244224d89

Better to not have copy-paste code....

bigtimebuddy: Just do `super.updateTransform()`

bigtimebuddy:
```suggestion
        super.updateTransform();
```

bigtimebuddy:
@laino could you check out this lint error:
```
/home/travis/build/pixijs/pixi.js/packages/display/src/DisplayObject.js
  243:5  warning  Duplicate name 'calculateBounds'  no-dupe-class-members
```

laino:
This rebasing nevers works with pull requests the way I want it to.... give me a second.

SukantPal:
@ivanpopelyshev Do you want to wait before the filter system's `measure` PR is done before this? I think that'll change the semantics of how filter paddings are calculated.

ivanpopelyshev:
Yes, I'd like to do it in one PR instead.

bigtimebuddy:
Any update on this @ivanpopelyshev?

ivanpopelyshev:
not yet

bigtimebuddy:
Okay, I'm moving this PR to v5.3.0. I think it's an awesome fix for multiple filters with padding, but it's a slight change in behavior and I want to not totally mess-up anyone who's doing some complicated filter padding over-correction. Also, I want to get 5.2.1 out soon and I think this needs more time for @ivanpopelyshev to verify.

packages/display/src/DisplayObject.js
```diff
@@ -84,15 +84,17 @@ export class DisplayObject extends EventEmitter
         this.renderable = true;
 
         /**
-         * The display object container that contains this display object.
+         * The display object container that contains this DisplayObject.
```
cursedcoder: You can make it the state of art with `{@link PIXI.DisplayObject}`

cursedcoder: Just in case https://github.com/microsoft/tsdoc/blob/master/spec/code-snippets/DeclarationReferences.ts

ivanpopelyshev:
Just a note: I'm taking a break from big pixi stuff so you guys will have to wait a month

bigtimebuddy:
Thanks for the PR @laino, I agree with Ivan about this. It’s a breaking change because it changes how bounds are calculated with filter padding. It would screw up people in a major way.

We’d consider another PR just to simply sum padding (with an option in FilterSystem to revert to max current behavior). That should fix your double filter padding issues.

laino:
@bigtimebuddy 
> We’d consider another PR just to simply sum padding (with an option in FilterSystem to revert to max current behavior).
>  That should fix your double filter padding issues.

That only fixes 50% of the issues with filter padding. Nested filters will still have broken padding.

laino:
@bigtimebuddy Here's a playground that illustrates the issue:

https://www.pixiplayground.com/#/edit/7Gyh_MFT-UQRqNDNjPRuu

Because the filter system doesn't take into account filter padding on child elements, the graphic becomes cropped when the parent element has any filter.

laino:
Basically the 2 options for fixing that are either this PR, which just adds filter padding to bounds, allowing us to forget about filter padding in code everywhere; or taking a performance hit by making all those places consider the padding of child elements too.

bigtimebuddy:
Totally fair point. I think we should address these issues separately. Sum-padding is an easy fix and has little impact for users, the bounds change is not.

Pixi currently doesn't consider filters for bounds. I think we need to introducing the concept of normal bounds (current behavior) and render/real bounds (w/ filter). Naming TBD. Overloading current bounds is going to have lots of side-effects for users positioning elements. I think pixi needs another concept to untangle this problem.

PR #6308, opened 2020-01-05
add option to disable the built-in InteractionManager updater

laino:
##### Description of change

By default ``InteractionManager`` adds ``InteractionManager.update`` to ``Ticker.system``.
Except for now adding ``tickerUpdater`` *instead* this behavior is unchanged.

``tickerUpdate`` now contains the rate limiting check based on ``deltaTime``, then calls ``update`` if it passed.

A configuration option ``InteractionManager.useSystemTicker`` has been added. If it is
set to ``true``, InteractionManager won't automatically use the system ticker.

This partially fixes the problems outlined in https://github.com/pixijs/pixi.js/issues/6307

Basically you can now use your own updating logic like this:

```js
// ...

const renderer = new pixi.Renderer();
renderer.plugins.interaction.useSystemTicker = false;

function renderOneFrame() {
    // update scene
    gameLogic();

    // update transforms once, needs a parent element
    root.updateTransform(); 

    // update hover states
    renderer.plugins.interaction.update(); 

    // the fifth argument prevents updating transforms a second time
    renderer.render(root, null, null, null, true); 
}
```

Thanks to these changes a 3 second performance capture of my application in idle state looks like 
this: 

</img>

Instead of this monstrosity:

</img>

Your battery life will thank you. On my system this saves about 500 wakeups/second in powertop.

##### Pre-Merge Checklist

### DIFF
--- a/packages/interaction/src/InteractionManager.js
+++ b/packages/interaction/src/InteractionManager.js
@@ -730,6 +730,8 @@ export class InteractionManager extends EventEmitter
      */
     setTargetElement(element, resolution = 1)
     {
+        this.removeTicker();
+
         this.removeEvents();
 
         this.interactionDOMElement = element;
@@ -852,8 +854,6 @@ export class InteractionManager extends EventEmitter
         {
             return;
         }
-
-        Ticker.system.remove(this.tickerUpdate, this);
 
         if (window.navigator.msPointerEnabled)
         {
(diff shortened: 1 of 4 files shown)


laino:
I figured I shouldn't mess with ``Ticker`` internals in a test intended for something else, so I added ``Ticker.count`` to get the number of listeners on a Ticker in a defined way.

PR #6311, opened 2020-01-06
Fixes modules/types not being found by typescript from js packages

Zyie:
##### Description of change

As discussed here #6261 and in the slack channels there have been issues with getting types and modules being recognised properly.

Before we ran into issues of either a module not being recognised or cannot use namespace as type, depending on the different solutions that have been thrown around.

With this change we no longer need to declare modules inside the `global.d.ts` file (except for settings as this one is weird). We should also get types for packages that are still in javascript.

I've tested this out on my display branch and it fixed all my build issues. 

@EvidentlyCube @ivanpopelyshev  this should help with the core/spritesheet packages as well. 

##### Pre-Merge Checklist

- [x ] Lint process passed (`npm run lint`)
- [x ] Tests passed (`npm run test`)

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,16 +1,4 @@
 declare module '@pixi/settings';
-declare module '@pixi/constants';
-declare module '@pixi/core' {
-    interface Texture {
-        destroy(): void;
-    }
-
-    interface BaseTexture {
-        destroy(): void;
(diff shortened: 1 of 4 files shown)


ivanpopelyshev:
Holy Cow!

PR #6312, opened 2020-01-07
Fixes docs for DecomposedDataUri

Zyie:
##### Description of change

This pr fixes the automatically generated types where `DecomposedDataUri` is not exported. This has been an issue before in #6150 where interfaces are stripped out along with there comments.

To document it in a way that it doesn't get removed means defining it below its declaration
##### Pre-Merge Checklist

### DIFF
--- a/packages/utils/src/network/decomposeDataUri.ts
+++ b/packages/utils/src/network/decomposeDataUri.ts
@@ -1,30 +1,46 @@
 import { DATA_URI } from '../const';
 
-/**
- * @memberof PIXI.utils
- */
 export interface DecomposedDataUri {
-    /**
-     * @property {string} Media type, eg. `image`
-     */
     mediaType: string;
-    /**
-     * @property {string} subType Sub type, eg. `png`
-     */
     subType: string;
-    /**
-     * @property {string} charset
-     */
     charset: string;
-    /**
-     * @property {string} encoding Data encoding, eg. `base64`
-     */
     encoding: string;
-    /**
-     * @property {string} data The actual data
-     */
     data: string;
 }
+
+/**
+ * @memberof PIXI.utils
+ * @interface DecomposedDataUri
+ */
+
+/**
+ * type, eg. `image`
+ * @memberof PIXI.utils.DecomposedDataUri#
+ * @member {string} mediaType
+ */
+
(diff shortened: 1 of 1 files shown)


themoonrat:
Feels odd to have the documentation separate to the code :/ Is this a temp thing until we're fully .ts and find a different solution for docs?

Zyie:
Yes i believe this is temporary until the conversion is complete. Then we will remove jsdoc for something that works better with typescript.

packages/utils/src/network/decomposeDataUri.ts
```diff
@@ -1,31 +1,47 @@
...
     encoding: string;
-    /**
-     * @property {string} data The actual data
-     */
     data: string;
 }
 
+/**
+ * @memberof PIXI.utils
+ * @interface DecomposedDataUri
```
bigtimebuddy: Why not just add the properties here?

Zyie: I'm not sure what you mean? I tried adding member like so but this didnt work

```
/**
 * @memberof PIXI.utils
 * @interface DecomposedDataUri
 * @member {string} mediaType type, eg. `image`
 * @member {string} subType Sub type, eg. `png`
 * @member {string} charset type, eg. `image`
 * @member {string} encoding Data encoding, eg. `base64`
 * @member {string} data The actual data
 */
```

I just copied the structure that IPoint implemented for its docs

bigtimebuddy: Try `@property`

Zyie: Adding property results in an empty interface

```
/**
 * @memberof PIXI.utils
 * @interface DecomposedDataUri
 * @property {string} mediaType type, eg. `image`
 * @property {string} subType Sub type, eg. `png`
 * @property {string} charset type, eg. `image`
 * @property {string} encoding Data encoding, eg. `base64`
 * @property {string} data The actual data
 */
interface DecomposedDataUri {
}
```

bigtimebuddy: Okay. Not use fighting with tsd jsdocs

PR #6315, opened 2020-01-07
Converts settings to TypeScript

Zyie:
##### Description of change

Opening another draft PR to let people know what is being worked on.

So far I have tried to use namespaces for settings however this resulted in many setting objects instead of one large one, not sure why.

For now I have gone with one interface that has optional parameters for the options defined outside the initial settings file.

Let me know if you guys have a better solution to try out. I dont think the current method is bad just not great.

##### Pre-Merge Checklist


Zyie:
> I'm not in love with the name `PixiSettings`, maybe simply `Settings`

I've updated the name

ivanpopelyshev:
@bigtimebuddy @Zyie What about `IPixiSettings` and `IRenderOptions` ? Also , `RenderOptions` in core will have more settings than that, and they will be optional. What about `IBaseRenderOptions` ?

Zyie:
> What about `IPixiSettings` and `IRenderOptions` ?'

Would it be a good idea to have a naming convention for interfaces. So all begin with I? That way we dont have a mashup of different names. For example we have IPoint, IDestoryOptions, but DecomposedDataUri, Settings, RenderOptions

> For example, it doesnt have `resolution`.

Ah yes i didnt realise that got added dynamically in `AbstractRenderer`.

ivanpopelyshev:
I think `DecomposedDataUri` is the only one without `I`

bigtimebuddy:
I like using `I` as a prefix for interfaces.

Zyie:
Ok I've updated the names to use the prefix. I'll let @ivanpopelyshev fix whatever needs to be fixed in his core pr

PR #6316, opened 2020-01-09
A fix to maintain the text gradient when modifying the drop shadow distance.

surtr-isaz:
For example, see: https://pixijs.io/examples/#/text/text.js

In the example above, when modifying the dropShadowDistance, try 100, one will observe the gradient distorting across the 3 lines of text.

Animating it will result in a clearer demonstration. For example, add the following:
```
app.ticker.add((delta) => {
  richText.style.dropShadowDistance += 1;
});
```

This is due to the drop shadow's offset increasing the dimensions of the canvas used to render the text and which the gradient's calculations rely on.

The fix corrects the dimensions by removing this offset thus maintaining the desired gradient.

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/Text.js
+++ b/packages/text/src/Text.js
@@ -437,8 +437,12 @@ export class Text extends Sprite
         let currentIteration;
         let stop;
 
-        const width = Math.ceil(this.canvas.width / this._resolution);
-        const height = Math.ceil(this.canvas.height / this._resolution);
+        // a dropshadow will enlarge the canvas and result in the gradient being
+        // generated with the incorrect dimensions
+        const dropShadowCorrection = (style.dropShadow) ? style.dropShadowDistance : 0;
+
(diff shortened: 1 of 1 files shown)


themoonrat:
Wouldn't this now cause text to get cut off with larger drop shadow distances?

surtr-isaz:
> Wouldn't this now cause text to get cut off with larger drop shadow distances?

Nope, I've tested it with very large values and and it renders correctly.

However, I did not test it with negative values. I just tested negative values the Text example and the current behaviour is that the canvas is cropped. Unfortunately, this doesn't fix that, but at least your gradients remain consistent ...

surtr-isaz:
Perhaps the distance should be absolute ...

themoonrat:
I see, was thinking of a different area of Text code. Thanks for the fix!

surtr-isaz:
My pleasure, gentlemen.

PR #6330, opened 2020-01-15
Adds measure pass to the FilterSystem.

SukantPal:
This is the pull request for the measuring pass API to PIXI.FilterSystem; it will allow filters to specify the area required by them to render into the output.

Use Cases:
When a object is partially visible, autoFit=false or true might not be the optimal solution. The filter doesn't need the whole object, neither can it work by just the visible portion (e.g. blur).

Backward Compatibility:
> measure(): A default onMeasure() is provided by the Filter class. It will cause the filter ask the input to be the whole object (if autoFit=false) or just the specified output (if autoFit=true).

> Vertex shader: inputRect & outputRect are two new uniforms. The vertex shader will automatically map the texture coords onto the input texture.

##### Description of change
* `FilterPass`: New data-class for the filter-system. It is provided to the filters using `state.filterPass`.

* `FilterSystem`: `measure()` added and the `push` method uses it. `pop()` updates `state#passIndex` and adds `inputRect`, `outputRect` to the global uniforms.

* `defaultFilter.vert`: The texture coordinates are calculated a bit differently. The old process is used to calculate `outputTextureCoords`, which is then mapped onto the input texture (output -> input frame mapping by translation). This will allow filters to specify input/output frames without changing shaders.

##### Pre-Merge Checklist


SukantPal:
@ivanpopelyshev Do you think we should provide an option to disable the measure pass?

bigtimebuddy:
Could you look at the failing mask unit test?

SukantPal:
> Could you look at the failing mask unit test?

@bigtimebuddy I'm sorry, but I can't do anything tonight. I just posted this so you could see the scope of the changes. I'm going to tackle this tomorrow afternoon 💯

ivanpopelyshev:
Calling `measure` before we start work with filters is good idea because we also can abort or ignore rendering inside. ignore means we can just render to output and not do filters.

However that PR does not address those IF's :(

SukantPal:
@ivanpopelyshev How will we decide whether we need to abort the filters?

SukantPal:
@bigtimebuddy I did push the fixes. I'll be creating an example where we require this feature for properly rendering.

@ivanpopelyshev As we discussed, measuring filter I/O rects is useful strictly only in cases when the object is not fully visible. Should we disable going through each filter's measure-method when the object is fully visible?

SukantPal:
@bigtimebuddy @GoodBoyDigital - @ivanpopelyshev was talking about some default "behaviors" for `measure` that can be applied to any filter by setting a property. He & I have decided to create a formal project proposal for this feature & any such behaviours so we could have a standardized definition. He also wants these features to conform to how Adobe Flash's filters (used to?) work.

packages/core/src/filters/defaultFilter.vert
```diff
@@ -2,25 +2,27 @@ attribute vec2 aVertexPosition;
 
 uniform mat3 projectionMatrix;
 
-varying vec2 vTextureCoord;
```
eXponenta: You can't change uniform name, because all custom filters stops work if it has custom vertex shader.
This is no case for loss of compatibility .

packages/core/src/filters/defaultFilter.vert
```diff
@@ -2,25 +2,27 @@ attribute vec2 aVertexPosition;
...
+varying vec2 vTextureCoord;// actually the input texture coord (gl_FragColor to set output)
 
 uniform vec4 inputSize;
-uniform vec4 outputFrame;
```
eXponenta: Same problem.

SukantPal: Yeah, I forgot about that. Maybe we can make them inputFrame and outputFrame?

eXponenta: We can rename it, consistently, but we must found their usage on core and pixi-* packages too.
It isn't simple, but possible.

SukantPal: If I use inputFrame, then we don’t need to rename outputFrame and it will be consistent, right?

SukantPal: I changed them.

ivanpopelyshev:
I didnt have enough time to investigate it, but here are my thoughts for this week.

However, @SukantPal had a great idea that I need to write a doc first! Here it goes.

There are **SIX** related problems that I know of:

**1. Forward pass, input becomes bigger and bigger**
Consider we have an object 30x30 and two filters, padding 5 and 6. That means our outputs after each filter should be 35 and 41. We can obtain it by a **forward** measure pass.

**2. Backwards pass, outputs are reduced**
Consider we have object 1010 and screen of 1000, and two filters of padding 40. Outputs are 1050 and 1090 , but our screen is bigger, why do we need such big outputs?
Solution is 1040->1040->1000

**3. Margin/padding**
In that case, if first filter is a blur, we don't actually need big input if we use clamping. 1000->1040->1000 is enough. We need `input=1040` in case `filter1.margin=40` and input=1000 if `filter1.padding=40`.
 
**4. Four sides**
All calculations have to be side-dependent. Top/bottom/left/right, we need `Insets`, not just one number,

**5. Measure gomomorphism**
If filter has custom logic for mapping input to output and back, we need `onMeasure` methods. Maybe two, because of two passes. 

As an example of code that can fit in onMeasure - `Graphics` and `GraphicsGeometry` bounds for rotation and scale.

**6. Partial apply **
Sometimes people want to apply filter partially - like, why do we need to run the shader on all screen if we have only small twist in the center? That means we can copy some of the input area and preserve it, and apply shader only to part of it. 

In next post I will talk about how to solve those things, whether we need to solve them, and how is this PR related.

SukantPal:
This is waiting on the PR converting to Typescript. I've created a draft of the full changes that will happen here: 
[PixiJS Feature Proposal - Filter measure() pass.pdf](https://github.com/pixijs/pixi.js/files/4113758/PixiJS.Feature.Proposal.-.Filter.measure.pass.pdf)

SukantPal:
Note that this PR is based from dev-core-typescript, and is waiting on that PR.

PR #6333, opened 2020-01-16
fix(render): make cacheAsBitmap not affect other textures with a filter for 1 frame

Loonride:
##### Description of change

The issue I was experiencing was that I had a single texture that had a custom filter. Any time I created a new Graphics object and set `cacheAsBitmap` to true, the texture with the custom filter would partially flicker for that frame while the Graphic was being cached. This change fixed the problem for me.

I saw an issue which may be the same as my problem here: https://github.com/pixijs/pixi.js/issues/2831

If you really need me to, I could make a new reproducible test demo. But preferably not, I think this change is quite straightforward.

##### Pre-Merge Checklist

### DIFF
--- a/packages/mixin-cache-as-bitmap/src/index.js
+++ b/packages/mixin-cache-as-bitmap/src/index.js
@@ -184,8 +184,8 @@ DisplayObject.prototype._initCachedDisplayObject = function _initCachedDisplayOb
 
     // for now we cache the current renderTarget that the WebGL renderer is currently using.
     // this could be more elegant..
-    const cachedRenderTexture = renderer.renderTexture.current ?
-      renderer.renderTexture.current.clone() : null;
+    const cachedRenderTexture = renderer.renderTexture.current
+        ? renderer.renderTexture.current.clone() : null;
     const cachedSourceFrame = renderer.renderTexture.sourceFrame.clone();
     const cachedProjectionTransform = renderer.projection.transform;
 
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Based on our experience with that kind of bugs:
1. I trust that you found the real issue
2. I require example anyway to ensure that issue wont appear after a regression

Flickering is easy to reproduce: turn off ticker and manually call render() one or two times.

Loonride:
> Based on our experience with that kind of bugs:
> 
> 1. I trust that you found the real issue
> 2. I require example anyway to ensure that issue wont appear after a regression
> 
> Flickering is easy to reproduce: turn off ticker and manually call render() one or two times.
> 
> If you cant reproduce it or dont have enough time - I'll deal with it in weekend.

Here is an example: https://jsfiddle.net/Loonride/vcu4n1bm/18/

You will notice that a part of the rabbit image cuts off during the frames where a new graphics object is created.

I found that these lines were part of the problem, as removing them from the example fixes the issue:

```
if (sprite) {
        if (sprite.parent) {
            var parent = sprite.parent;
            parent.removeChild(sprite);
            parent.addChild(sprite);
        }
    }
```

I use this method to bring sprites to the top of the stage in my project. Is there a better recommended way to reorder pixi sprites?

ivanpopelyshev:
OK, still need to find time to review it properly and add a unit test.

ivanpopelyshev:
I'm on a break but we need someone to add a test. @eXponenta or someone else

ivanpopelyshev:
Demo is here: https://www.pixiplayground.com/#/edit/LZli1NESJ3Vlf_8l9Fdbe

If you comment out first line, you'll see the original error.

btw, We dont need to clone renderTexture, only `sourceFrame`, please make that change in PR and I'll approve it.

ivanpopelyshev:
@Loonride in case you dont have time right now, I can make separate PR with cloning just rectangle. I think whether should I try to add tests too...

bigtimebuddy:
Closing in favor of #6466

PR #6335, opened 2020-01-16
fix typings error for 5.2.1

ivanpopelyshev:
Follow-up for #6328


ivanpopelyshev:
@bigtimebuddy that's good but we need to merge 5.2.1 to 5.3 again!

bigtimebuddy:
Done

PR #6339, opened 2020-01-17
Typescript display bounds fix

ivanpopelyshev:
I congratulate @Zyie with merge of his big PR, there are not many people that had modified that many lines in PixiJS in one commit, its certainly an achievement!

Now to fixes:
1. `_lastBoundsID` is not needed anymore.
2. my typescript still complaints about certain expression in `Container` I do not understand why, I've tried several versions of TS. Lets make it simple.

### DIFF
--- a/packages/display/src/Container.ts
+++ b/packages/display/src/Container.ts
@@ -2,7 +2,6 @@ import { MaskData, Renderer } from '@pixi/core';
 import { settings } from '@pixi/settings';
 import { removeItems } from '@pixi/utils';
 import { DisplayObject, IDestroyOptions } from './DisplayObject';
-import { Bounds } from './Bounds';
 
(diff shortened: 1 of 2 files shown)


packages/display/src/Container.ts
```diff
@@ -465,12 +464,12 @@ export class Container extends DisplayObject
             // TODO: filter+mask, need to mask both somehow
             if (child._mask)
             {
-                const maskObject = (child._mask as MaskData).maskObject || (child._mask as Container);
+                const maskObject = ((child._mask as MaskData).maskObject || child._mask) as Container;
 
                 // TODO: maskObject is getting PIXI.DisplayObject and DisplayObject confused.
                 // Once types are sorted we wont need to do this weird conversion
```
Zyie: We should be able to remove this comment now. I added it when types where being weird for me, this fixes that

PR #6340, opened 2020-01-17
Convert core to typescript

ivanpopelyshev:
Its done. Lets review it. Discussion and a few details about conversion is in the post below.


bigtimebuddy:
Good start

bigtimebuddy:
My one request is you do-no-harm with this PR. Keep the changes as straightforward as possible and no refactoring.

ivanpopelyshev:
Well, there is one small deprecation: I removed `CubeTexture` because it has `from` method that is not compatible with TS overrides.

ivanpopelyshev:
OK, it should build, but I don't know how to deal with `.glsl` files. Please help!

bigtimebuddy:
@Zyie could use your help on this one

Zyie:
I believe this should do the trick.

Add this to the the `global.d.ts` file

```
declare module '*.frag' {
    const value: string;

    export default value;
}

declare module '*.vert' {
    const value: string;

    export default value;
}
```

ivanpopelyshev:
OK, its building but now I have to fight typescript-eslint and few of those rules are just stupid.

```
   23:23  error    Identifier 'WEBGL_draw_buffers' is not in camel case              @typescript-eslint/camelcase
   24:24  error    Identifier 'OES_texture_float' is not in camel case               @typescript-eslint/camelcase
   25:23  error    Identifier 'WEBGL_lose_context' is not in camel case              @typescript-eslint/camelcase
   26:29  error    Identifier 'OES_vertex_array_object' is not in camel case         @typescript-eslint/camelcase
   27:32  error    Identifier 'EXT_texture_filter_anisotropic' is not in camel case  @typescript-eslint/camelcase
   28:30  error    Identifier 'OES_element_index_uint' is not in camel case          @typescript-eslint/camelcase
   29:24  error    Identifier 'OES_texture_float' is not in camel case               @typescript-eslint/camelcase
   30:30  error    Identifier 'OES_texture_float_linear' is not in camel case        @typescript-eslint/camelcase
   31:28  error    Identifier 'OES_texture_half_float' is not in camel case          @typescript-eslint/camelcase
   32:34  error    Identifier 'OES_texture_half_float_linear' is not in camel case   @typescript-eslint/camelcase
   33:28  error    Identifier 'WEBGL_color_buffer_float' is not in camel case
```

ivanpopelyshev:
OK, i have a few things:

I've tried to preserve everything even if I saw places where we could do better with TS. only in few cases I actually changed code a bit.

1. I tried to preserve`private`,`protected` and `readonly` modifiers but they are real pain. Maybe I'll remove a few of them to make code better
2. `castToBaseTexture` method exists to drop several occurences of `texture || baseTexture` in places that are not performance-important so one more function call surely wont be bad.
3. I prefer to write `Array<something>` for arrays with heavy usage of `push`/`pop`, and `something[]` for constant arrays. Yes, they are the same in TS, but I like that detail from other languages, like `ArrayList` from java and `List` from C#. However for experssions like `f(x:number[][] = []): number[][]` i prefer `[]` because its short.
4. Unused parameter requires 5 lines to disable eslint. What if we remove that requirement either from tsconfig? That will drop it to only one line.
5. `autoDetectRenderer()` returns Renderer. I dont know whether we want `CanvasRenderer` there too. I believe it can be addressed in canvas-PR.
6. `CubeTexture` was rarely used trash that had static `from` method with signature different from `BaseTexture`. I deprecated it.

Zyie:
> 4. Unused parameter requires 5 lines to disable eslint. What if we remove that requirement either from tsconfig? That will drop it to only one line.

I think you can add an `_` to the start of a parameter and typescript will understand that its an unused variable.
```
// eslint-disable-next-line @typescript-eslint/no-unused-vars
render(_object: any): void
{
}
```
This then makes it only one line

bigtimebuddy:
@EvidentlyCube could you review this?

EvidentlyCube:
@bigtimebuddy I'll review it tomorrow!

packages/core/src/shader/ShaderSystem.ts
```diff
@@ -168,7 +180,7 @@ export class ShaderSystem extends System
      * @returns {String} Unique signature of the uniform group
      * @private
      */
-    getSignature(group, uniformData)
+    getSignature(group: UniformGroup, uniformData: any): string
```
Zyie: This should be private

packages/core/src/shader/UniformGroup.ts
```diff
@@ -8,11 +8,18 @@ let UID = 0;
  */
 export class UniformGroup
 {
+    uniforms: {[key: string]: any};
+    group: boolean;
```
Zyie: These should be readonly.

Also is there a better type than any for this?

packages/core/src/shader/utils/defaultValue.ts
```diff
@@ -1,11 +1,23 @@
...
+
 /**
  * @method defaultValue
  * @memberof PIXI.glCore.shader
  * @param type {String} Type of value
  * @param size {Number}
  * @private
  */
-export function defaultValue(type, size)
+export function defaultValue(type: string, size: number): any
```
Zyie: This returns `number|Float32Array|Int32Array|boolean|boolean[]`

packages/core/src/shader/utils/extractUniformsFromSrc.ts
```diff
@@ -1,19 +1,11 @@
...
-}
-
-function extractUniformsFromString(string)
+function extractUniformsFromString(string: string): any
 {
     const maskRegex = new RegExp('^(projectionMatrix|uSampler|translationMatrix)$');
 
-    const uniforms = {};
-    let nameSplit;
+    const uniforms: any = {};
```
Zyie: The type here is not any.
```
{[x: string]: {
    type: string;
    dirtyId: number;
    name: string;
    value: number|Float32Array|Int32Array|boolean|boolean[];
};}

File: packages/core/src/shader/utils/extractUniformsFromSrc.ts [hunk 5]
```diff
@@ -1,19 +1,11 @@
...
-export function extractUniformsFromSrc(vertexSrc, fragmentSrc, mask)
-{
-    const vertUniforms = extractUniformsFromString(vertexSrc, mask);
-    const fragUniforms = extractUniformsFromString(fragmentSrc, mask);
-
-    return Object.assign(vertUniforms, fragUniforms);
-}
-
-function extractUniformsFromString(string)
+function extractUniformsFromString(string: string): any
```
Zyie: The type here is not any.

{[x: string]: {
    type: string;
    dirtyId: number;
    name: string;
    value: number|Float32Array|Int32Array|boolean|boolean[];
};}

packages/core/src/shader/utils/generateUniformsSync.ts
```diff
@@ -1,9 +1,11 @@
+import { UniformGroup } from '../UniformGroup';
+
 // cv = CachedValue
 // v = value
 // ud = uniformData
 // uv = uniformValue
 // l = location
-const GLSL_TO_SINGLE_SETTERS_CACHED = {
+const GLSL_TO_SINGLE_SETTERS_CACHED: any = {
```
Zyie: `{[x: string]: string}`

packages/core/src/shader/utils/generateUniformsSync.ts
```diff
@@ -51,7 +53,7 @@ const GLSL_TO_SINGLE_SETTERS_CACHED = {
     sampler2DArray: 'gl.uniform1i(location, v)',
 };
 
-const GLSL_TO_ARRAY_SETTERS = {
+const GLSL_TO_ARRAY_SETTERS: any = {
```
Zyie: `{[x: string]: string}`

packages/core/src/shader/utils/getTestContext.ts
```diff
@@ -11,13 +11,13 @@ let context = unknownContext;
  * @private
  * @returns {WebGLRenderingContext} a gl context to test with
  */
-export function getTestContext()
+export function getTestContext(): WebGL2RenderingContext
```
Zyie: I believe this function returns `WebGLRenderingContext|WebGL2RenderingContext` not just a WebGL2 context.

This also means that `gl` has the same type instead of any

packages/core/src/shader/utils/mapSize.ts
```diff
@@ -1,4 +1,4 @@
-const GLSL_TO_SIZE = {
+const GLSL_TO_SIZE: any = {
```
Zyie: `{[x: string]: number}`

packages/core/src/shader/utils/mapType.ts
```diff
@@ -1,6 +1,6 @@
-let GL_TABLE = null;
+let GL_TABLE: any = null;
```
Zyie: `{[x: string]: string}`

packages/core/src/shader/utils/mapType.ts
```diff
@@ -1,6 +1,6 @@
...
 
-const GL_TO_GLSL_TYPES = {
+const GL_TO_GLSL_TYPES: any = {
```
Zyie: `{[x: string]: string}`

packages/core/src/shader/utils/unsafeEvalSupported.ts
```diff
@@ -1,5 +1,5 @@
 // Cache the result to prevent running this over and over
-let unsafeEval;
+let unsafeEval: any;
```
Zyie: `boolean`

packages/core/src/state/StateSystem.ts
```diff
@@ -18,10 +20,19 @@ const WINDING = 4;
  */
 export class StateSystem extends System
 {
+    gl: WebGL2RenderingContext;
+    stateId: number;
+    polygonOffset: number;
+    blendMode: BLEND_MODES;
+    protected _blendEq: boolean;
+    readonly map: Array<Function>;
+    readonly checks: Array<Function>;
```
Zyie: These should have the following types

```
 readonly map: Array<(value: boolean) => void>;
 readonly checks: Array<(system: this, state: State) => void>;
```

packages/core/src/state/StateSystem.ts
```diff
@@ -289,7 +299,7 @@ export class StateSystem extends System
      * @param {Function} func  the checking function to add or remove
      * @param {boolean} value  should the check function be added or removed.
      */
-    updateCheck(func, value)
+    updateCheck(func: Function, value: boolean): void
```
Zyie: `func` should be `(system: this, state: State) => void`

packages/core/src/textures/BaseTexture.ts
```diff
@@ -41,7 +59,33 @@ const defaultBufferOptions = {
...
+    dirtyStyleId: number;
+    cacheId: string;
+    readonly valid: boolean;
+    readonly textureCacheIds: Array<string>;
+    readonly destroyed: boolean;
+    resource: Resource;
+    _batchEnabled: number;
+    _batchLocation: number;
+
+    constructor(resource: Resource | ImageSource | string | any = null, options: IBaseTextureOptions = null)
```
Zyie: I dont think the `any` is needed here

packages/core/src/textures/BaseTexture.ts
```diff
@@ -54,7 +98,7 @@ export class BaseTexture extends EventEmitter
         if (resource && !(resource instanceof Resource))
         {
             resource = autoDetectResource(resource, resourceOptions);
-            resource.internal = true;
+            (resource as any).internal = true;
```
Zyie: This conversion is not needed

packages/core/src/textures/BaseTexture.ts
```diff
@@ -397,18 +441,18 @@ export class BaseTexture extends EventEmitter
      *
      * @private
      */
-    _refreshPOT()
+    _refreshPOT(): void
```
Zyie: this should be private

packages/core/src/textures/BaseTexture.ts
```diff
@@ -493,14 +537,14 @@ export class BaseTexture extends EventEmitter
...
-    destroy()
+    destroy(): void
     {
         // remove and destroy the resource
         if (this.resource)
         {
             this.resource.unbind(this);
             // only destroy resourced created internally
-            if (this.resource.internal)
+            if ((this.resource as any).internal)
```
Zyie: I dont think this conversion is needed

packages/core/src/textures/resources/BaseImageResource.ts
```diff
@@ -10,13 +12,17 @@ import { ALPHA_MODES } from '@pixi/constants';
  */
 export class BaseImageResource extends Resource
 {
+    source: ImageSource;
+    noSubImage: boolean;
```
Zyie: This should be protected

packages/core/src/textures/resources/CubeResource.ts
```diff
@@ -55,7 +60,9 @@ export class CubeResource extends ArrayResource
      *
      * @returns {boolean} true is success
      */
-    upload(renderer, baseTexture, glTexture)
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+    // @ts-ignore
+    upload(renderer: Renderer, baseTexture: BaseTexture, glTexture: GLTexture): boolean
```
Zyie: Can use `_` here

packages/core/src/textures/resources/ImageResource.ts
```diff
@@ -245,9 +265,11 @@ export class ImageResource extends BaseImageResource
 
             let flag = true;
 
-            for (const key in baseTexture._glTextures)
+            const glTextures = (baseTexture as any)._glTextures;
```
Zyie: I dont think this conversion is needed

packages/core/src/textures/resources/Resource.ts
```diff
@@ -10,6 +12,14 @@ import { Runner } from '@pixi/runner';
  */
 export class Resource
 {
+    protected _width: number;
+    protected _height: number;
+    // readonly?
+    destroyed: boolean;
+    protected internal: boolean;
```
Zyie: This should be public

packages/core/src/textures/resources/Resource.ts
```diff
@@ -181,10 +193,15 @@ export class Resource
      * @param {PIXI.GLTexture} glTexture - texture instance for this webgl context
      * @returns {boolean} true is success
      */
-    upload(renderer, baseTexture, glTexture) // eslint-disable-line no-unused-vars
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    /* eslint-disable @typescript-eslint/ban-ts-ignore */
+    // @ts-ignore
+    upload(renderer: Renderer, baseTexture: BaseTexture, glTexture: GLTexture): boolean
```
Zyie: Can use `_` here

packages/core/src/textures/resources/Resource.ts
```diff
@@ -194,17 +211,22 @@ export class Resource
      * @param {PIXI.GLTexture} glTexture - texture instance for this webgl context
      * @returns {boolean} `true` is success
      */
-    style(renderer, baseTexture, glTexture) // eslint-disable-line no-unused-vars
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    /* eslint-disable @typescript-eslint/ban-ts-ignore */
+    // @ts-ignore
+    style(renderer: Renderer, baseTexture: BaseTexture, glTexture: GLTexture): boolean
```
Zyie: Can use _ here

packages/core/src/textures/resources/Resource.ts
```diff
@@ -194,17 +211,22 @@ export class Resource
...
+    /* eslint-enable @typescript-eslint/no-unused-vars */
+    /* eslint-enable @typescript-eslint/ban-ts-ignore */
 
     /**
      * Clean up anything, this happens when destroying is ready.
      *
      * @protected
      */
-    dispose()
+    dispose(): void
```
Zyie: This should be protected

packages/core/src/textures/resources/Resource.ts
```diff
@@ -228,4 +250,21 @@ export class Resource
...
+     * Abstract, used to auto-detect resource type
+     *
+     * @static
+     * @param {*} source - The source object
+     * @param {string} extension - The extension of source, if set
+     */
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    /* eslint-disable @typescript-eslint/ban-ts-ignore */
+    // @ts-ignore
+    static test(source: any, extension?: string): boolean
```
Zyie: Can use _ here

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -16,7 +30,15 @@ import { Ticker } from '@pixi/ticker';
...
 export class VideoResource extends BaseImageResource
 {
-    constructor(source, options)
+    protected _autoUpdate: boolean;
+    protected _isAutoUpdating: boolean;
+    protected _updateFPS: number;
+    protected _msToNextUpdate: number;
+    protected autoPlay: boolean;
+    private _load: Promise<VideoResource>;
+    private _resolve: Function;
```
Zyie: This type is `(value?: VideoResource | PromiseLike<VideoResource>) => void`

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -107,17 +131,22 @@ export class VideoResource extends BaseImageResource
      *
      * @param {number} [deltaTime=0] - time delta since last tick
      */
-    update(deltaTime = 0)
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    /* eslint-disable @typescript-eslint/ban-ts-ignore */
+    // @ts-ignore
+    update(deltaTime = 0): void
```
Zyie: Can use `_` here

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -180,10 +209,10 @@ export class VideoResource extends BaseImageResource
      *
      * @private
      */
-    _onError()
+    _onError(): void
```
Zyie: This should be private

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -192,9 +221,9 @@ export class VideoResource extends BaseImageResource
      * @private
      * @return {boolean} True if playing.
      */
-    _isSourcePlaying()
+    _isSourcePlaying(): boolean
```
Zyie: This should be private

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -205,17 +234,19 @@ export class VideoResource extends BaseImageResource
      * @private
      * @return {boolean} True if ready.
      */
-    _isSourceReady()
+    _isSourceReady(): boolean
```
Zyie: This should be private

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -205,17 +234,19 @@ export class VideoResource extends BaseImageResource
...
+        return source.readyState === 3 || source.readyState === 4;
     }
 
     /**
      * Runs the update loop when the video is ready to play
      *
      * @private
      */
-    _onPlayStart()
+    _onPlayStart(): void
```
Zyie: This should be private

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -235,7 +266,7 @@ export class VideoResource extends BaseImageResource
      *
      * @private
      */
-    _onPlayStop()
+    _onPlayStop(): void
```
Zyie: This should be private

packages/core/src/textures/resources/VideoResource.ts
```diff
@@ -249,9 +280,9 @@ export class VideoResource extends BaseImageResource
      *
      * @private
      */
-    _onCanPlay()
+    _onCanPlay(): void
```
Zyie: This should be private

packages/core/src/textures/resources/SVGResource.ts
```diff
@@ -116,20 +135,20 @@ export class SVGResource extends BaseImageResource
      *
      * @private
      */
-    _loadSvg()
+    _loadSvg(): void
```
Zyie: This should be private

packages/core/src/textures/resources/autoDetectResource.ts
```diff
@@ -57,7 +59,7 @@ export const INSTALLED = [];
  *        texture should be updated from the video. Leave at 0 to update at every render
  * @return {PIXI.resources.Resource} The created resource.
  */
-export function autoDetectResource(source, options)
+export function autoDetectResource(source: any, options: any): Resource
```
Zyie: The options here are not any

packages/core/src/textures/resources/CubeResource.ts
```diff
@@ -15,7 +18,9 @@ import { TARGETS } from '@pixi/constants';
  */
 export class CubeResource extends ArrayResource
 {
-    constructor(source, options)
+    items: Array<BaseTexture>;
+
+    constructor(source: Array<string|Resource>, options?: any)
```
Zyie: The options here are not any

packages/core/src/Renderer.ts
```diff
@@ -120,12 +140,12 @@ export class Renderer extends AbstractRenderer
...
+            contextChange: new Runner('contextChange'),
             reset: new Runner('reset'),
             update: new Runner('update'),
             postrender: new Runner('postrender'),
             prerender: new Runner('prerender'),
-            resize: new Runner('resize', 2),
+            resize: new Runner('resize'),
         };
 
         /**
```
Zyie: The comments from line 159-162 can be removed. It was added temporarily for the display object conversion

packages/core/src/Renderer.ts
```diff
@@ -401,32 +422,32 @@ export class Renderer extends AbstractRenderer
...
 
     /**
      * Clear the frame buffer
      */
-    clear()
+    clear(): void
     {
-        this.framebuffer.bind();
-        this.framebuffer.clear();
+        this.renderTexture.bind();
```
Zyie: Just wanted to check this was intentional as it changed from what the function says it will do?

ivanpopelyshev:
OK then i'll fix it and add `public` to fields that are public according to jsdoc and common sense.

Extra information.

Both I and @GoodBoyDigital have custom scene trees for pixijs. Its better if `@pixi/core` does not depend on `@pixi/display`.

Part of that is typings-related - that's why I made `IFilterTarget`, `IMaskTarget`, `ISpriteMaskTarget`.
However there's dependency from `Renderer`/`AbstractRenderer` classes that require code and API change, and that will be addressed in separate PR.

packages/core/src/batch/AbstractBatchRenderer.ts
```diff
@@ -26,13 +30,42 @@ import { ENV } from '@pixi/constants';
...
+    shaderGenerator: BatchShaderGenerator;
+    geometryClass: typeof BatchGeometry;
+    vertexSize: number;
+    state: State;
+    size: number;
+    _vertexCount: number;
+    _indexCount: number;
+    _bufferedElements: Array<any>;
+    _bufferedTextures: Array<BaseTexture>;
+    _bufferSize: number;
```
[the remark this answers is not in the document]
ivanpopelyshev: protected

packages/core/src/Renderer.ts
```diff
@@ -284,7 +304,7 @@ export class Renderer extends AbstractRenderer
      *        sure it doesn't collide with properties on Renderer.
      * @return {PIXI.Renderer} Return instance of renderer
      */
-    addSystem(ClassRef, name)
+    addSystem(ClassRef: any, name: string): this
```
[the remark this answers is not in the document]
ivanpopelyshev: good idea!

packages/core/src/Renderer.ts
ivanpopelyshev: yes, its not used inside @core, but by the meaning of the function there should be renderTexture usage, not framebuffer, because framebuffer does not containt clear color.

ivanpopelyshev:
OK, I'm applying your idea regarding public/protected/private . However I leave the fields that have mixed usage without any modifiers. I also remove `readonly` in many cases, because otherwise core code will look bad.

We still have those modifiers in jsdoc, so we can apply them if TS will allow us to have custom modifiers.

ivanpopelyshev:
Regarding statics - I left them where they were originally and our team expects them there. If we move stuff like `from` to the top of the file, it will take attention of user instead of actual constructor.

ivanpopelyshev:
> Wow this is huge, well done @ivanpopelyshev, very impressive.

Thank you! I have predisposition for large and brain-melting tasks, but I couldn't have a chance to show that without people who "prepare the field" for me.

ivanpopelyshev:
OK, access modifiers are ready.

ivanpopelyshev:
Btw, @Zyie, I had to rename all `run` to `emit` because we dont have typings for `run` in runner, and maybe we have to ask @GoodBoyDigital how to do it, its his place.

packages/core/src/batch/AbstractBatchRenderer.ts
```diff
@@ -201,7 +234,7 @@ export class AbstractBatchRenderer extends ObjectRenderer
          * @private
          * @see PIXI.AbstractBatchRenderer#getAttributeBuffer
          */
-        this._aBuffers = {};
+        this._aBuffers = {} as any;
```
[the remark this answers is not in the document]
ivanpopelyshev: We have to track whoever made it before and ask. I think I already asked that person several months ago but i dont remember the answer

packages/core/src/filters/Filter.ts
```diff
@@ -143,12 +148,18 @@ import defaultFragment from './defaultFilter.frag';
  */
 export class Filter extends Shader
 {
+    padding: number;
+    resolution: number;
+    enabled: boolean;
+    autoFit: boolean;
+    legacy: boolean;
```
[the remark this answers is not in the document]
ivanpopelyshev: No, user can change those, and `legacy` is changed in FilterSystem.

packages/core/src/textures/BaseTexture.ts
ivanpopelyshev: its needed because its gets passed to autoDetectResource, the method can accept any possible trash if user adds more resources.

packages/core/src/textures/resources/Resource.ts
ivanpopelyshev: In this particular case, user can override it in children and IDE will automatically add `_` for all three fields there. I'm not sure that its acceptable.

packages/core/src/textures/resources/Resource.ts
ivanpopelyshev: I don't know why it was protected, users should be able to call this thing,

packages/core/src/textures/resources/Resource.ts
ivanpopelyshev: Same as `style`, users might override it and bring us a bad luck with those `_`.

packages/core/src/AbstractRenderer.ts
```diff
@@ -16,8 +43,25 @@ const tempMatrix = new Matrix();
  * @extends PIXI.utils.EventEmitter
  * @memberof PIXI
  */
-export class AbstractRenderer extends EventEmitter
+export abstract class AbstractRenderer extends EventEmitter
 {
+    _backgroundColor: number;
+    _backgroundColorRgba: number[];
+    _backgroundColorString: string;
```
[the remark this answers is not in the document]
ivanpopelyshev: `_backgroundColorRgba` is used by RenderTextureSystem

packages/core/src/context/ContextSystem.ts
```diff
@@ -13,10 +15,30 @@ let CONTEXT_UID = 0;
  */
 export class ContextSystem extends System
 {
+    CONTEXT_UID: number;
```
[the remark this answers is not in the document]
ivanpopelyshev: the one above is counter, I've renamed it.

packages/core/src/geometry/GeometrySystem.ts
```diff
@@ -14,10 +17,21 @@ const byteSizeMap = { 5126: 4, 5123: 2, 5121: 1 };
  */
 export class GeometrySystem extends System
 {
+    CONTEXT_UID: number;
+    gl: WebGL2RenderingContext;
+    _activeGeometry: Geometry;
+    _activeVao: WebGLVertexArrayObject;
+    readonly hasVao: boolean;
+    readonly hasInstance: boolean;
+    readonly canUseUInt32ElementIndex: boolean;
```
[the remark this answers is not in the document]
ivanpopelyshev: you are right!

packages/core/src/geometry/GeometrySystem.ts
```diff
@@ -643,7 +643,7 @@ export class GeometrySystem extends System
      * Unbind/reset everything
      * @protected
      */
-    unbind()
+    unbind(): void
```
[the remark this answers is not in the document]
ivanpopelyshev: Agree, user can call `reset()`

packages/core/src/mask/MaskSystem.ts
```diff
@@ -12,20 +14,19 @@ import { MASK_TYPES } from '@pixi/constants';
  */
 export class MaskSystem extends System
 {
+    enableScissor: boolean;
+    alphaMaskPool: Array<SpriteMaskFilter[]>;
+    maskDataPool: Array<MaskData>;
+    maskStack: Array<MaskData>;
+    alphaMaskIndex: number;
```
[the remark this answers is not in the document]
ivanpopelyshev: I use `protected` for two of those fields because I override them in projection plugin. I'll make them private when we add another hook for it.

packages/core/src/renderTexture/BaseRenderTexture.ts
```diff
@@ -70,10 +75,10 @@ export class BaseRenderTexture extends BaseTexture
...
 
         // Set defaults
-        this.mipmap = false;
-        this.width = Math.ceil(width) || 100;
-        this.height = Math.ceil(height) || 100;
-        this.valid = true;
+        this.mipmap = 0;
+        (this as any).width = Math.ceil(width) || 100;
+        (this as any).height = Math.ceil(height) || 100;
+        (this as any).valid = true;
```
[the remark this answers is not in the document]
ivanpopelyshev: Yes, they are better with readonly

packages/core/src/shader/Program.ts
```diff
@@ -117,10 +127,10 @@ export class Program
      *
      * @returns {object} the attribute data for this program
      */
-    getAttributeData(program, gl)
+    getAttributeData(program: WebGLProgram, gl: WebGL2RenderingContext): void
```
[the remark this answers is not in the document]
ivanpopelyshev: I've added an interface

packages/core/src/shader/UniformGroup.ts
ivanpopelyshev: Its either UniformGroup either number either Float32Array or Uint16Array or a texture... we can think about it

packages/core/src/shader/utils/extractUniformsFromSrc.ts
ivanpopelyshev: old file, not being used, added an interface.

packages/core/src/shader/utils/getTestContext.ts
ivanpopelyshev: That's actually good question I need to clarify: I think its better if we use WebGL2RenderingContext everywhere. All things that we use are the same as in WebGLRenderingContext.

packages/core/src/textures/resources/BaseImageResource.ts
ivanpopelyshev: oh, both are public actually

packages/core/src/textures/resources/Resource.ts
ivanpopelyshev: I made it abstract.

packages/core/src/textures/resources/autoDetectResource.ts
ivanpopelyshev: they are being passed to autoDetectResource, we only can do something like `{[key: string: any}`

packages/core/src/textures/resources/CubeResource.ts
ivanpopelyshev: they are being passed to autoDetectResource, we only can do something like `{[key: string: any}`

ivanpopelyshev:
OK, i think I've addressed most of your questions, @Zyie , we have a few threads to discuss. Click on "show all" but dont expand resolved ones.

packages/core/src/textures/resources/autoDetectResource.ts
bigtimebuddy: Yes we do. The options are in the documentation.

packages/core/src/textures/resources/CubeResource.ts
bigtimebuddy: You need stricter types here for options. My suggestion is to make CubeResourceOptions and then combine all the options into the autoDetectResource: CubeResourceOptions|CanvasResourceOptions|VideoResourseOptions

bigtimebuddy: You need stricter types here for options. My suggestion is to make CubeResourceOptions and then combine all the options into the autoDetectResource: CubeResourceOptions|CanvasResourceOptions|VideoResourseOptions

packages/core/src/shader/utils/getTestContext.ts
bigtimebuddy: Since it’s annoying to maintain webgl|webgl2 in many places, just create an alias type like RenderingContext that combines both. And we can always drop support later for v1. I think it’s important that types be accurate. Just staying webgl2 isn’t accurate.

packages/core/src/textures/resources/Resource.ts
bigtimebuddy: destroy is the public interface. Dispose is a single method resource can override to do any special cleanup. it should be protected

packages/core/src/shader/utils/getTestContext.ts
ivanpopelyshev: We already use polyfills in a few places that webgl1 doesnt have.

packages/core/src/textures/resources/Resource.ts
ivanpopelyshev: My main concern was about ImageResource and it has ImageBitmap inside. `dispose()` is supposed to remove it, but compressed source is still there. We have `dispose()` methods not only for resources - for Geometries, Buffers and other things, and they are supposed to be recoverable.

packages/core/src/textures/resources/autoDetectResource.ts
ivanpopelyshev: We can add those fields to satisfy autocomplete but we need `{[key:string]:any} just because its possible to use in custom resources.

packages/core/src/shader/utils/getTestContext.ts
bigtimebuddy: But consider building for environments where you explicit are forbidding webgl2. Like some old device browser. We wouldn’t want to fail if webgl2 is disallowed via env types. I don’t understand the harm in being verbose.

packages/core/src/textures/resources/Resource.ts
bigtimebuddy: Okay public it is

packages/core/src/textures/resources/autoDetectResource.ts
bigtimebuddy: We should define strict types and then add a generic type for extensibility. Just using the generic type isn’t that great

packages/core/src/textures/resources/BaseImageResource.ts
Zyie: ah ok cool. 

Looking at the jsdocs should `source` be readonly?

Zyie:
@ivanpopelyshev Good job going through all the feedback! I've resolved a few more so not many left to go.

packages/core/src/textures/BaseTexture.ts
Zyie: I'm guessing this would be affected by whatever you and @bigtimebuddy decided about `autoDetectResource` below?

packages/core/src/shader/utils/getTestContext.ts
ivanpopelyshev: Our gl-level is beautiful. I do want it to be more verbose, but checks like "gl as WebGL2RenderingContext" might ruin it. 

I propose a compromise: we can make separate interface `IPixiGLContext` and add our polyfills there. However it has to be separate PR so we actually see all the places it affects. That will also help to document things - people will see what did we hack into context and extra article inside jsdoc for `IPixiContext` will prepare them for it. WebGL2 calls can be put to object inside `asGL2` or something more verbose.

packages/core/src/textures/BaseTexture.ts
ivanpopelyshev: yep

packages/core/src/textures/resources/autoDetectResource.ts
ivanpopelyshev: I dont care about this place and it can be patched later when we find a case it doesnt work. Its still experimental API :) 

You may change it and and other related places, and commit to this branch.

packages/core/src/textures/resources/CubeResource.ts
ivanpopelyshev: ^^^ its up to you. However I remind you that "|" might have side effects. I remember our struggle with `Point|ObservablePoint` ;)

packages/core/src/textures/resources/BaseImageResource.ts
ivanpopelyshev: I recommend people to use `resource.source` in several cases. Yes, it wont actually work in this version because `baseTextrure.resource` is `Resource` and not `BaseImageResource` . We might resolve it later. 

@bigtimebuddy should we make it public or protected?

packages/core/src/shader/utils/getTestContext.ts
bigtimebuddy: I’m cool with an interface to insulate ourselves from the problem, as long as  there’s not a lot to cover.

packages/core/src/textures/resources/CubeResource.ts
bigtimebuddy: Yeah that’s fine in this case because these Resource-specific options are non-functional and don’t have method signatures like Point v ObservablePoint.

packages/core/src/shader/utils/getTestContext.ts
ivanpopelyshev: I can make alias for WebGL2 for now, and we investigate what to do in separate PR, is that ok?

ivanpopelyshev:
@bigtimebuddy 
Summary of our discussion: its better to make our own interface for WebGLRenderingContext with our polyfills, we can attach documentation to it.

Currently:
1. `getTextContext` uses `WebGLRenderingContext | WebGL2RenderingContext`
2. a few shader-related functions use `WebGLRenderingContextBase` because it doesn't matter
3. all internal variables have `IRenderingContext` because its polyfilled.

Please help to choose a name for `IRenderingContext`

ivanpopelyshev:
One more note: `import type`  declarations
1. module self-reference `import { Renderer } from '@pixi/core'`
2. local path for a sub-module, like `filters` , `import {IFilterTarget } from './IFilterTarget'`
Using 1 in place of 2 is fine.

Imports that go to files in neighbour sub-module, like from filters to masks, are disorienting.

packages/core/src/batch/AbstractBatchRenderer.ts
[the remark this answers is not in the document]
ivanpopelyshev: Lets do it in separate PR maybe? Yes, the use of `length`  is not needed here, but we can separate talk to original author. I'm pro that change, but it shouldn't be blocking.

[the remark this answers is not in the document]
Zyie: Ok that's fine with me

packages/core/src/textures/resources/ArrayResource.ts
```diff
@@ -17,18 +20,27 @@ import { autoDetectResource } from './autoDetectResource';
  */
 export class ArrayResource extends Resource
 {
-    constructor(source, options)
+    readonly length: number;
+    items: Array<BaseTexture>;
+    itemDirtyIds: Array<number>;
+    private _load: Promise<ArrayResource>;
+
+    constructor(source: number|Array<any>, options?: any)
```
Zyie: I believe this is another case for the `autoDetectResource` discussion

packages/core/src/framebuffer/FramebufferSystem.ts
```diff
@@ -232,14 +243,14 @@ export class FramebufferSystem extends System
      * @protected
      * @param {PIXI.Framebuffer} framebuffer
      */
-    initFramebuffer(framebuffer)
+    initFramebuffer(framebuffer: Framebuffer): any
```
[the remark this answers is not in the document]
ivanpopelyshev: I made a class for it, finally.

packages/core/src/renderTexture/RenderTexturePool.ts
```diff
@@ -16,11 +19,16 @@ import { nextPow2 } from '@pixi/utils';
  */
 export class RenderTexturePool
 {
+    public textureOptions: IBaseTextureOptions;
+    public enableFullScreen: boolean;
+    texturePool: any;
```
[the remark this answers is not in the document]
ivanpopelyshev: That's the problem, we use it with numbers too.

packages/core/src/geometry/utils/setVertexAttribArrays.ts
```diff
@@ -6,7 +8,7 @@
  * @param state {*}
  * @private
  */
-export function setVertexAttribArrays(gl, attribs, state)
+export function setVertexAttribArrays(gl: IRenderingContext, attribs: any, state: any): void
```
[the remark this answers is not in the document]
ivanpopelyshev: Its old polyfill for the case when VAO plugin is not available. Currently pixi v5 doesnt work without it. 

Types can be fixed when someone actually makes this PR, and for that we need at least one device/browser that doesnt have VAO extension somehow.

packages/core/src/shader/ShaderSystem.ts
```diff
@@ -210,21 +218,21 @@ export class ShaderSystem extends System
...
+        const attribMap: any = {};
 
         for (const i in program.attributeData)
         {
             attribMap[i] = program.attributeData[i].location;
         }
 
         const shaderProgram = compileProgram(gl, program.vertexSrc, program.fragmentSrc, attribMap);
-        const uniformData = {};
+        const uniformData: any = {};
```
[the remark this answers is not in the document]
ivanpopelyshev: I've made an interface

packages/core/src/shader/utils/compileProgram.ts
```diff
@@ -8,7 +25,8 @@
  * @param attributeLocations {Object} An attribute location map that lets you manually set the attribute locations
  * @return {WebGLProgram} the shader program
  */
-export function compileProgram(gl, vertexSrc, fragmentSrc, attributeLocations)
+export function compileProgram(gl: WebGLRenderingContextBase, vertexSrc: string, fragmentSrc: string,
+    attributeLocations?: Array<any>): WebGLProgram
```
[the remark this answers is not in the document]
ivanpopelyshev: fixed. `{[key:string]:number}`

packages/core/src/shader/utils/generateUniformsSync.ts
```diff
@@ -78,7 +80,7 @@ const GLSL_TO_ARRAY_SETTERS = {
     sampler2DArray: 'gl.uniform1iv(location, v)',
 };
 
-export function generateUniformsSync(group, uniformData)
+export function generateUniformsSync(group: UniformGroup, uniformData: any): Function
```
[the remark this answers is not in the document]
ivanpopelyshev: Its {size, type} inside but we can add this later.

ivanpopelyshev:
I think we can merge this thing and treat `autoDetectResource` and `IRendererContext` in separate PR.

Right now the only serious thing here is `IRendererContext` name, should we rename it?

packages/core/src/renderTexture/RenderTexturePool.ts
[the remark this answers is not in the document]
Zyie: Where does this get used with numbers? I only see it being being used by `FilterSystem` as a `RenderTexture`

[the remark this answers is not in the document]
ivanpopelyshev: numbers as a key in `getOptimal`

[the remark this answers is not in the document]
Zyie: Ah so you could do this 

`texturePool: {[x in string|number]: RenderTexture[]};`

and then change `filterPoolKey` in `RenderTexture` to be `string|number`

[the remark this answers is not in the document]
ivanpopelyshev: oh, its `in`, not `is`! thank you!

bigtimebuddy:
I like IRenderContext

packages/core/src/AbstractRenderer.ts
```diff
@@ -16,8 +45,25 @@ const tempMatrix = new Matrix();
...
+    public readonly view: HTMLCanvasElement;
+    public readonly plugins: IRendererPlugins;
+    public readonly transparent: boolean | 'notMultiplied';
+    public readonly autoDensity: boolean;
+    public readonly preserveDrawingBuffer: boolean;
+
+    protected readonly _tempDisplayObjectParent: DisplayObject;
+    protected _backgroundColor: number;
+    protected _backgroundColorString: string;
+    _backgroundColorRgba: number[];
```
EvidentlyCube: In a different file you declare `private` explicitly, shouldn't that be followed here?
Also, it'd probably be a good idea to enforce explicit access modifiers in eslint if that's the way to go.

packages/core/src/AbstractRenderer.ts
```diff
@@ -37,17 +83,17 @@ export class AbstractRenderer extends EventEmitter
...
 
         // Add the default render options
         options = Object.assign({}, settings.RENDER_OPTIONS, options);
 
         // Deprecation notice for renderer roundPixels option
-        if (options.roundPixels)
+        if ((options as IRendererOptionsLegacy).roundPixels)
         {
-            settings.ROUND_PIXELS = options.roundPixels;
+            settings.ROUND_PIXELS = (options as IRendererOptionsLegacy).roundPixels;
```
EvidentlyCube: I am not very fond of this double casting but I also don't really see how to deal with this better.

packages/core/src/Renderer.ts
```diff
@@ -401,32 +417,32 @@ export class Renderer extends AbstractRenderer
...
 
     /**
      * Clear the frame buffer
      */
-    clear()
+    clear(): void
     {
-        this.framebuffer.bind();
-        this.framebuffer.clear();
+        this.renderTexture.bind();
```
EvidentlyCube: Can you change the doc to match what the function does?

packages/core/src/batch/AbstractBatchRenderer.ts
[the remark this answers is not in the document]
EvidentlyCube: Might be a good idea to create an issue for that future change so it's not lost.

packages/core/src/batch/BatchSystem.ts
```diff
@@ -74,7 +79,7 @@ export class BatchSystem extends System
      * @param arr
      * @param maxTextures
      */
-    copyBoundTextures(arr, maxTextures)
+    copyBoundTextures(arr: Array<BaseTexture>, maxTextures: number): void
```
EvidentlyCube: Why `Array<BaseTexture>` and not `BaseTexture[]` (or why the latter format used in different places)? Again may be a good idea to enforce one standard via eslint.

packages/core/src/context/ContextSystem.ts
```diff
@@ -13,10 +15,30 @@ let CONTEXT_UID = 0;
  */
 export class ContextSystem extends System
 {
+    public webGLVersion: number;
+    protected CONTEXT_UID: number;
+    protected gl: IRenderingContext;
+    /* eslint-disable @typescript-eslint/camelcase */
+    extensions:
```
EvidentlyCube: Access modifier?

packages/core/src/context/ContextSystem.ts
```diff
@@ -44,7 +66,7 @@ export class ContextSystem extends System
         this.handleContextLost = this.handleContextLost.bind(this);
         this.handleContextRestored = this.handleContextRestored.bind(this);
 
-        renderer.view.addEventListener('webglcontextlost', this.handleContextLost, false);
+        (renderer.view as any).addEventListener('webglcontextlost', this.handleContextLost, false);
```
EvidentlyCube: I presume the issue here is that the view has an explicit list of events it emits - is it maybe possible to just modify the list of events nad callbacks it supports rather than doing this cast to `any`?

packages/core/src/filters/Filter.ts
```diff
@@ -212,11 +223,13 @@ export class Filter extends Shader
...
      */
-    apply(filterManager, input, output, clearMode, currentState)
+    apply(filterManager: FilterSystem, input: RenderTexture, output: RenderTexture, clearMode: CLEAR_MODES,
+        /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+        _currentState?: FilterState): void
     {
         // do as you please!
 
-        filterManager.applyFilter(this, input, output, clearMode, currentState);
+        filterManager.applyFilter(this, input, output, clearMode);
```
EvidentlyCube: Why was this code changed?

packages/core/src/framebuffer/FramebufferSystem.ts
```diff
@@ -65,13 +77,13 @@ export class FramebufferSystem extends System
 
             if (nativeDrawBuffersExtension)
             {
-                gl.drawBuffers = (activeTextures) =>
+                gl.drawBuffers = (activeTextures: number[]): void =>
```
EvidentlyCube: Oh, TS is not able to infer the types here?

packages/core/src/geometry/Buffer.ts
```diff
@@ -11,19 +37,26 @@ let UID = 0;
  */
 export class Buffer
 {
+    public data: ITypedArray;
+    public index: boolean;
+    public static: boolean;
+    public id: number;
+     disposeRunner: Runner;
```
EvidentlyCube: Access modifiers

packages/core/src/geometry/Geometry.ts
```diff
@@ -39,11 +40,20 @@ const map = {
  */
 export class Geometry
 {
+    public buffers: Array<Buffer>;
+    public indexBuffer: Buffer;
+    public attributes: {[key: string]: Attribute};
+    public id: number;
+    public instanced: boolean;
+    public instanceCount: number;
+    glVertexArrayObjects: {[key: number]: {[key: string]: WebGLVertexArrayObject}};
```
EvidentlyCube: Access modifiers

packages/core/src/renderTexture/RenderTexturePool.ts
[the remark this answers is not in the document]
EvidentlyCube: It's marked as resolved but the code is still `any` - am I to understand this change is not being applied?

packages/core/src/geometry/Geometry.ts
eXponenta: we cannot make them explicit, becouse `glVertexArrayObjects` used in Renderer.
We can't mark it as private and as protected, becouse it hide it, but we can't mark it as public, becouse it is not for users.

packages/core/src/AbstractRenderer.ts
eXponenta: It is deprecated property, it is not existed on `IRendererOptions` by default and used only for back support.
We can't use it on public api. Casting is normal usage for this case.

packages/core/src/batch/BatchSystem.ts
eXponenta: linting rulle is good idea.
With autfix of course ;-)

packages/core/src/filters/Filter.ts
ivanpopelyshev: because that parameter is absent in filterManager.

packages/core/src/batch/BatchSystem.ts
ivanpopelyshev: yeah, maybe use [] there because MAX_TEXTURES is a constant :)

packages/core/src/context/ContextSystem.ts
eXponenta: `webglcontextlost` is not standard event, it is from extension.
There aren't serious reason for override events list for 1 event.

packages/core/src/AbstractRenderer.ts
ivanpopelyshev: That particular field is used in RenderTextureSystem. I've decided not to add modifiers that can affect code readability.

packages/core/src/renderTexture/RenderTexturePool.ts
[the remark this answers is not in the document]
Zyie: I assumed this wouldn't have been marked resolved without the change. @ivanpopelyshev what is the reason for not doing this?

packages/core/src/batch/BatchSystem.ts
EvidentlyCube: @eXponenta so I understand it's a convention used in Pixi's codebase? That sounds like something that could be written down, unless it is and I never saw it :).

packages/core/src/geometry/Geometry.ts
EvidentlyCube: @eXponenta Which means they're essentially `internal`. Too bad TS doesn't have that. 

That brings the next question - [according to the docs](https://www.typescriptlang.org/docs/handbook/classes.html#public-by-default), the default access modifier is `public`, so they'll still appear in the generated documentation, right? Wouldn't it be clearer to mark them public and add add tsdoc `@private` with a few words of explanation why?

ivanpopelyshev:
@EvidentlyCube I forgot to push it :)

packages/core/src/context/ContextSystem.ts
ivanpopelyshev: Yep, `getExtensions` is protected, but there are references to object from other systems, so - uknown for now.

packages/core/src/batch/BatchSystem.ts
ivanpopelyshev: I wrote it down in this PR but we really need a place to write those things.

ivanpopelyshev:
> And a suggestion for the future - when enormous modules like this one are converted to TS it might be a good idea to split it into smaller steps, maybe 20 files each. I have no hope of being able to go through 95 files and remain concentrated :).

I feel like I'm rushing things, but from other point of view - we cant merge other PR's related to core while this one exists. Glory to our reviewers!

packages/core/src/textures/TextureSystem.ts
```diff
@@ -108,16 +121,16 @@ export class TextureSystem extends System
...
      */
-    bind(texture, location = 0)
+    bind(texture: Texture|BaseTexture, location = 0): void
     {
         const { gl } = this;
 
         if (texture)
         {
-            texture = texture.baseTexture || texture;
+            texture = texture.castToBaseTexture();
```
ivanpopelyshev: Here lays the problem: I found in my custom pixi fork that in some shaders texture locations had numbers in them, obviously, just a sampler location number. Maybe we'll stumble across it in PixiJS apps too, if someone made filters that have default uniform with location number instead of texture type. I suggest we check if its actually texture, maybe with `if (texture && texture.castToBaseTexture) { ... `

@GoodBoyDigital

bigtimebuddy:
While not perfect, I think this is good enough to merge. Further fixes can be added to address some of the outstanding things. Are you good @ivanpopelyshev or are you waiting for any other responses?

ivanpopelyshev:
There are two possible improvements and one potential problem we can address in PR's.

1. `IRendererContext` improvements  - WebGL1 + polyfill
2. `autoDetectResource` typings
3. The problem with `castToBaseTexture` that is actually a problem with custom filters and other stuff - I need to confirm it before fix.

Summary: YES, we can merge it.

ivanpopelyshev:
Actually, we cant merge it. There are failing examples, i didnt test everything :(

bigtimebuddy:
What’s broken? Could you keep me posted.

ivanpopelyshev:
Graphics, both batched and main versions. Probably something with Geometry.. https://www.pixiplayground.com/#/edit/jPgE3ZEUhhVb7GJx9DuJb

ivanpopelyshev:
OK, I've checked all examples. The one that doesnt work - ParticleContainer - fixed in #6371

packages/core/src/textures/TextureSystem.ts
ivanpopelyshev: Resolved, Mat said "let it throw"

PR #6341, opened 2020-01-17
Fix gaps in circle geometry again

### LINKED ISSUE #6327: Sometimes there's unwanted gap when drawing circle with line style
### Expected Behavior
[See Demo](https://pixi-playground.now.sh/loading.html)
There's no gap when drawing circle with linestyle
### Current Behavior
image
Sometimes there's small gap when drawing circle with linestyle, please see the screenshot above.

### Possible Solution
I don't know

### Steps to Reproduce
```javascript
import * as PIXI from 'pixi.js';
import { hslToRgb } from '../util';
import './main.scss';

class Particle {
  constructor(container, radius, radians, screenWidth, screenHeight, index) {
    this.g = new PIXI.Graphics();
    container.addChild(this.g);
    this.cx = radius*Math.cos(radians) + screenWidth/2;
    this.cy = radius*Math.sin(radians) + screenHeight/2;
    this.cr = index*2;
  }

  update() {
    this.cr += 0.1;
    if (this.cr > 40)
      this.cr = 0;
    this.alpha = Math.max(0, 1 - this.cr/20);
  }

  render() {
    this.g.clear();
    const color = PIXI.utils.rgb2hex(hslToRgb(200, 1, this.alpha));
    this.g.lineStyle(3, color, this.alpha);
    //this.g.beginFill(0xFF0000, this.alpha);
    this.g.drawCircle(this.cx, this.cy, this.cr*2);
    //this.g.endFill();
  }
}

function makeLoading(app) {
  let isPause = false;
  graphics.interactive = true;
  graphics.buttonMode = true;
  graphics.on('pointertap', () => {
    if (isPause)
      resume();
    else
      pause();
  });
  const particles = [];

  function play() {
    app.stage.addChild(graphics);
    app.ticker.add(update);

    const radius = 100;
    const radians = 18*Math.PI/180;
    for (let i=0;i<20;++i) {
      const particle = new Particle(container, radius, radians*i, app.screen.width, app.screen.height, i);
      particles.push(particle);
    }
  }

  function pause() {
    isPause = true;
    app.ticker.remove(update);
  }

  function resume() {
    isPause = false;
    app.ticker.add(update);
  }

  function update() {
    graphics.clear();
    graphics.beginFill(0xF7F8FC);
    graphics.drawRect(0, 0, app.screen.width, app.screen.height);
    graphics.endFill();
    for (let particle of particles) {
      particle.update();
      particle.render();
    }
  }

  return {
    play
  }
}

function main() {
    width: 600, height: 600, backgroundColor: 0, resolution: window.devicePixelRatio || 1,
    view: document.getElementById('ground')
  });
  const loading = makeLoading(app);
  loading.play();
}

main();
```
### Environment

- **`pixi.js` version**: 5.2.0
- **Browser & Version**: Chrome 78
- **OS & Version**: MacOS 10.14.6
- **Running Example**: https://pixi-playground.now.sh/loading.html

eXponenta:
Fix #6327 

Sometimes (with specific radius) `totalSegs` is integer and this case generate extra point same as start point.

Ex:
```
r = 5.3;
Math.floor(30 * Math.sqrt(r)) / 2.3 => 30.00000004
```

Reproduced by:
https://www.pixiplayground.com/#/edit/Iv9DmRYIMeoPAb-TQyBpq
Look in to console. 
Start and 2 end  points are similar.

Line builder fails when more that 2 points one after another is same and generate line segments on [0,0] for them.

Fixed:
https://www.pixiplayground.com/#/edit/OzhTsdVMM0xE-7O1Tcima
##### Pre-Merge Checklist

### DIFF
--- a/packages/graphics/src/utils/buildCircle.js
+++ b/packages/graphics/src/utils/buildCircle.js
@@ -49,7 +49,7 @@ export const buildCircle = {
 
         const seg = (Math.PI * 2) / totalSegs;
 
-        for (let i = 0; i < totalSegs - 1; i++)
+        for (let i = 0; i < totalSegs - 0.5; i++)
         {
             points.push(
                 x + (Math.sin(-seg * i) * width),
--- a/packages/graphics/src/utils/buildCircle.js
+++ b/packages/graphics/src/utils/buildCircle.js
@@ -49,7 +49,7 @@ export const buildCircle = {
 
         const seg = (Math.PI * 2) / totalSegs;
 
-        for (let i = 0; i < totalSegs; i++)
+        for (let i = 0; i < totalSegs - 1; i++)
         {
             points.push(
                 x + (Math.sin(-seg * i) * width),


bigtimebuddy:
Can you reproduce this in a unit test? Always the best way to avoid a regression.

Also could you rebase to master? Would like to get this fix in the upcoming v5.2.1

ivanpopelyshev:
> Can you reproduce this in a unit test? Always the best way to avoid a regression.
> 
> Also could you rebase to master? Would like to get this fix in the upcoming v5.2.1

Its possible to make a test but it will depend on your magical constant (2.3). I think I can make a test that detects which magical constant is used at that place. Maybe its better to not make a test on that thing?

eXponenta:
> Can you reproduce this in a unit test? Always the best way to avoid a regression.
> 
> Also could you rebase to master? Would like to get this fix in the upcoming v5.2.1

Already based on master

bigtimebuddy:
Awesome, I didn’t see because I’m on my phone atm. Thanks.

bigtimebuddy:
So good that @ivanpopleyshev approved it twice! Thanks for the fix.

Issue #6337, opened 2020-01-16
Feature: Filter's program name should be the filter's constructor?


SukantPal:
Debugging PixiJS by using WebGL inspection tools like Spector.js is made hard by the fact that shader names are generic, e.g. "pixi-shader-7".

If the name given to each filter's program could be 'this.constructor.name', that would be really helpful!

ivanpopelyshev:
Only if its fourth parameter. If its first or second - you'll have to spend time on threads like https://www.html5gamedevs.com/topic/44542-need-help-of-migrating-shader-from-v4-to-v5/ explaining people why is there suddenly one more parameter :)

SukantPal:
There is no need for parameters.

SukantPal:
@ivanpopelyshev `PIXI.Program` already has a third parameter, and `PIXI.Filter` doesn't pass that. We could just do this in the filter class:

```js
super(program, uniforms, `${this.constructor.name}`);
```

ivanpopelyshev:
Agree

SukantPal:
@ivanpopelyshev Yeah, we could make it the default parameter.

PR #6343, opened 2020-01-18
Program name is default to constructor.name rather than "pixi-shader"

SukantPal:
Pretty self-descriptive. Related to #6337

### DIFF
--- a/packages/core/src/shader/Program.js
+++ b/packages/core/src/shader/Program.js
@@ -29,7 +29,7 @@ export class Program
      * @param {string} [fragmentSrc] - The source of the fragment shader.
      * @param {string} [name] - Name for shader
      */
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
Here:

https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/filters/Filter.js#L153

if program is new, assign it a name.

SukantPal:
@ivanpopelyshev @bigtimebuddy If I pass constructor name in `Filter`, then the shader's name would be `Filter`. That wouldn't be useful either.

The only way I see this happening is by an additional parameter. Is everyone fine with that?

ivanpopelyshev:
I'm not good with changing any API's of filter at the moment. If we can't do it internally - lets do it later when Filter Docs will be good enough to stop the onslaught of filter-related complaints on me.

PR #6351, opened 2020-01-23
Childrens name references

djmisterjon:
Usage. Demo
https://www.pixiplayground.com/#/edit/015KEa5JB-ZAqDtj7cp-p

##### Description of change
suggest to allow map and control children by name.

- [ x] Tests demo and/or benchmarks are included
- [ x] Documentation is changed or added
- [ x] Lint process passed (`npm run lint`)
- [ x] Tests passed (`npm run test`)

### DIFF
--- a/packages/display/src/Container.js
+++ b/packages/display/src/Container.js
@@ -85,6 +85,19 @@ export class Container extends DisplayObject
     }
 
     /**
+     * Set or change name
+     *
+     * @param {string} name - Name value
+     * @returns {this}
+     */
+    setName(name)
+    {
+        this.name = name;
+
+        return this;
+    }
+
+    /**
      * Overridable method that can be used by Container subclasses whenever the children array is modified
      *
      * @protected
--- a/packages/display/src/DisplayObject.js
+++ b/packages/display/src/DisplayObject.js
@@ -301,14 +301,14 @@ export class DisplayObject extends EventEmitter
 
         while ((childrens = pool.shift()))
         {
-            for (let i = 0, c = null, l = childrens.length; i < l; i++)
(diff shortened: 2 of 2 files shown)


djmisterjon:
ho ok fine :)
It was just to share the idea, because there are quite a few users asking how to target a child.
But sorry i don't know how made a mixin/plugin or a test coverage.
And also i don't know what you mean by zIndex sorting !

ivanpopelyshev:
You are one of most experienced people in regarding how to manage huge scene trees in pixijs. I don't know about this particular implementation but when we actually start naming objects in tree, we'll ask you to review it. 

Of course it will be related to spine and other tools that export trees.

djmisterjon:
hey thanks ivan.
The general idea is just to offer better ergonomics than what pixi already offers.
In a very basic context (website) I guess native `.getChildByName(name)` can be fine.
But for game engine, i found it userless and make impossible to accomplish basic managing for spriteSheet, gsap, spines ....

It is not uncommon to have complex situations of this kind and to need to be able to properly manipulate the behavior of sprites, containers
image

I will be happy to integrate your discussion on this subject when the time comes.

bigtimebuddy:
What I meant by zIndex was that we recently worked out a new feature to support a simple version of sorting children. There were a bunch of ways to approach it, but we went with the thing that was the simplest.

Create your ideas as a 3rd party plugin by adding them to the Container prototype. Like we do with cacheAsBitmap. It’s less code that we have to be responsible for.

And by test coverage, I mean adding unit tests to execute your code.

I’m going to close this for reasons that Ivan and I pointed out. If you feel like there’s a simple way we can improve the ergonomics of Pixi, create a new issue and let’s discuss further.

PR #6352, opened 2020-01-26
Convert Graphics to typescript

eXponenta:
##### Description of change
Convert PIXI.Graphics to TS

P.S Awaits before `core's` problems will be resolved.
 
cross-package changes:

`type IShape = Circle | Ellipse | Polygon | Rectangle | RoundedRectangle;`- moved to @pixi/math

`Transform._worldID` - is internal (without modificator)

changed signature of hex2rgb  to:

`hex2rgb(hex: number, out: Array<number> | Float32Array = []): Array<number> | Float32Array`

##### Pre-Merge Checklist


ivanpopelyshev:
It has begun.

packages/graphics/src/const.ts
```diff
@@ -13,12 +21,13 @@
...
-export const GRAPHICS_CURVES = {
+export const GRAPHICS_CURVES: IGraphicsCurvesSettings = {
     adaptive: true,
     maxLength: 10,
     minSegments: 8,
-    maxSegments: 2048,
-    _segmentsCount(length, defaultSegments = 20)
+    maxSegments:  2048,
+
+    _segmentsCount(length: number, defaultSegments = 20)
```
lekzd: `private _segmentsCount(length: number, defaultSegments = 20): number`

eXponenta: We can't do it.
It used on ArcUtils.

packages/graphics/src/GraphicsData.ts
```diff
@@ -1,3 +1,13 @@
+import { Circle, Ellipse, Rectangle, Polygon, RoundedRectangle, Matrix, SHAPES } from '@pixi/math/src';
+import { FillStyle } from './styles/FillStyle';
+import { LineStyle } from './styles/LineStyle';
+
+/**
+ * @description Complex shape type
+ * @todo Move to Math shapes
+ */
+type IShape = Circle | Ellipse | Polygon | Rectangle | RoundedRectangle;
```
Zyie: Should this be moved to math then?

packages/graphics/src/GraphicsData.ts
```diff
@@ -6,14 +16,21 @@
  */
 export class GraphicsData
 {
+    shape: IShape;
+    lineStyle: LineStyle;
+    fillStyle: FillStyle;
+    matrix: Matrix;
+    type: SHAPES;
+    points: number[];
+    holes: Array<GraphicsData>;
```
Zyie: These all seem like normal public variables so should have the `public` prefix

packages/graphics/src/GraphicsGeometry.ts
```diff
@@ -404,7 +444,7 @@ export class GraphicsGeometry extends BatchGeometry
      * Generates intermediate batch data. Either gets converted to drawCalls
      * or used to convert to batch objects directly by the Graphics object.
      */
-    updateBatches()
+    updateBatches(): void
```
Zyie: Should this have public in front of it or is it meant to be protected but cant because its used in `Graphics`?

packages/graphics/src/Graphics.ts
```diff
@@ -1158,14 +1228,16 @@ export class Graphics extends Container
      * geometry, recaculate the vertices.
      * @protected
      */
-    calculateVertices()
+    protected calculateVertices(): void
     {
-        if (this._transformID === this.transform._worldID)
+        const wtID = (this.transform as any)._worldID;
```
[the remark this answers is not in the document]
eXponenta: But i can't change core package in this PR.

packages/graphics/src/Graphics.ts
```diff
@@ -1125,13 +1195,13 @@ export class Graphics extends Container
...
      */
-    calculateTints()
+    protected calculateTints(): void
     {
         if (this.batchTint !== this.tint)
         {
             this.batchTint = this.tint;
 
-            const tintRGB = hex2rgb(this.tint, temp);
+            const tintRGB = hex2rgb(this.tint, temp as any);
```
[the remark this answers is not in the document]
eXponenta: It in math package. I can't do this in this PR.

packages/graphics/src/Graphics.ts
```diff
@@ -639,19 +711,20 @@ export class Graphics extends Container
...
+    beginTextureFill(options: IFillStyleOptions | number): Graphics
     {
         // backward compatibility with params: (texture, color, alpha, matrix)
         if (options instanceof Texture)
         {
             deprecation('v5.2.0', 'Please use object-based options for Graphics#beginTextureFill');
 
-            const [texture, color, alpha, matrix] = arguments;
+            // eslint-disable-next-line
+            const [texture, color, alpha, matrix] = (arguments as any) as Array<any>;
```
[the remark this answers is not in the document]
eXponenta: arguments isn't Array<T>, it ArrayLike, and not iterable

packages/graphics/src/GraphicsData.ts
eXponenta: Yes, i think it must be moved to math then

packages/graphics/src/GraphicsData.ts
eXponenta: Not, it is not pulicl api. It is internal.

packages/graphics/src/GraphicsGeometry.ts
eXponenta: Internal

packages/graphics/src/Graphics.ts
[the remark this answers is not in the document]
eXponenta: image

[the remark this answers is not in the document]
Zyie: Sorry maybe i didn't explain it well. 

So both `lineTextureStyle` and `beginTextureFill` both do the same thing in terms of using the arguments for options. However `beginTextureFill` has `as Array<any>;` on the end.
```
// lineTextureStyle
const [width, texture, color, alpha, matrix, alignment, native] = (arguments as any);
options = { width, texture, color, alpha, matrix, alignment, native };
```

```
// beginTextureFill
const [texture, color, alpha, matrix] = (arguments as any) as Array<any>;
options = { texture, color, alpha, matrix };
```

Why does one have it but the other doesn't?

packages/graphics/src/Graphics.ts
```diff
@@ -639,19 +711,20 @@ export class Graphics extends Container
      * @param {PIXI.Matrix} [options.matrix=null] - Transform matrix
      * @return {PIXI.Graphics} This Graphics object. Good for chaining method calls
      */
-    beginTextureFill(options)
+    beginTextureFill(options: IFillStyleOptions | number): Graphics
```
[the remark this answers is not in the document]
eXponenta: Yep!

packages/graphics/src/Graphics.ts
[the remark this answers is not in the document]
eXponenta: Thanks, i fixed it.

packages/graphics/src/GraphicsData.ts
ivanpopelyshev: Whole class is public-y internal . Hacks are allowed, but its not a public API. I dont know whether to add modifiers here.

packages/graphics/src/Graphics.ts
```diff
@@ -303,20 +374,21 @@ export class Graphics extends Container
      * @param {boolean} [options.native=false] - If true the lines will be draw using LINES instead of TRIANGLE_STRIP
      * @return {PIXI.Graphics} This Graphics object. Good for chaining method calls
      */
-    lineTextureStyle(options)
+    public lineTextureStyle(options: ILineStyleOptions | number = null): Graphics
```
bigtimebuddy: we should remove number as a type for options since it’s deprecated.

bigtimebuddy:
All the chain-able methods should return `this` Not `Graphics`, eg lineStyle method.

packages/graphics/src/Graphics.ts
```diff
@@ -303,20 +374,21 @@ export class Graphics extends Container
...
 
-            const [width, texture, color, alpha, matrix, alignment, native] = arguments;
+            // eslint-disable-next-line
+            const [width, texture, color, alpha, matrix, alignment, native] = arguments as any;
 
             options = { width, texture, color, alpha, matrix, alignment, native };
 
             // Remove undefined keys
-            Object.keys(options).forEach((key) => options[key] === undefined && delete options[key]);
+            Object.keys(options).forEach((key) => (options as any)[key] === undefined && delete (options as any)[key]);
```
bigtimebuddy: You should be able to do `keyof ILineStyleOptions` Instead of any

eXponenta:
> All the chain-able methods should return `this` Not `Graphics`, eg lineStyle method.

What you mean?

`this` is Graphics type.
All chain methods return this.

Becouse test is passed =)

packages/graphics/src/Graphics.ts
eXponenta: we can remove deprecated behavior because it is deprecated.

Zyie:
> > All the chain-able methods should return `this` Not `Graphics`, eg lineStyle method.
> 
> What you mean?

He means that a method that returns the instance of the class should use `this` instead of `Graphics`.

e.g. `public lineStyle(options: ILineStyleOptions | number = null): this`

packages/graphics/src/Graphics.ts
bigtimebuddy: Actually this isn’t important code because it’s deprecated. Ignore

packages/graphics/src/Graphics.ts
bigtimebuddy: No we should leave the if block because it still needs to work at runtime, but we should have docs or types show the old params.

eXponenta:
> > > All the chain-able methods should return `this` Not `Graphics`, eg lineStyle method.
> > 
> > 
> > What you mean?
> 
> He means that a method that returns the instance of the class should use `this` instead of `Graphics`.
> 
> e.g. `public lineStyle(options: ILineStyleOptions | number = null): this`

Thanks, done.

packages/graphics/src/Graphics.ts
ivanpopelyshev: agree with @bigtimebuddy . This place needs an object. When we make new typings people will have to fix MANY things in their TS code, it will be like major release.

packages/graphics/src/GraphicsData.ts
eXponenta: done

packages/graphics/src/Graphics.ts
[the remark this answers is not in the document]
eXponenta: Upd, ArrayLike is readonly.

packages/graphics/src/Graphics.ts
eXponenta: Done

packages/graphics/src/Graphics.ts
```diff
@@ -50,19 +79,50 @@ const DEFAULT_SHADERS = {};
...
+     * @private
+     * @member {PIXI.Point}
+     */
+
+    static _TEMP_POINT = new Point();
+
+    public shader: Shader;
+    public pluginName: string;
+
+    protected geometry: GraphicsGeometry;
```
bigtimebuddy: I don't think the access modifier for `geometry` should be `protected`. I think it should be `public readonly`

packages/graphics/src/Graphics.ts
```diff
@@ -50,19 +79,50 @@ const DEFAULT_SHADERS = {};
...
         super();
+
         /**
          * Includes vertex positions, face indices, normals, colors, UVs, and
          * custom attributes within buffers, reducing the cost of passing all
          * this data to the GPU. Can be shared between multiple Mesh or Graphics objects.
+         *
          * @member {PIXI.GraphicsGeometry}
-         * @readonly
+         * @protected
```
bigtimebuddy: Same as above, this should continue to be public

packages/graphics/src/Graphics.ts
```diff
@@ -303,20 +375,21 @@ export class Graphics extends Container
      * @param {boolean} [options.native=false] - If true the lines will be draw using LINES instead of TRIANGLE_STRIP
      * @return {PIXI.Graphics} This Graphics object. Good for chaining method calls
      */
-    lineTextureStyle(options)
+    public lineTextureStyle(options: ILineStyleOptions | number = null): this
```
bigtimebuddy:
```suggestion
    public lineTextureStyle(options: ILineStyleOptions): this
```

packages/graphics/src/Graphics.ts
```diff
@@ -639,19 +712,20 @@ export class Graphics extends Container
      * @param {PIXI.Matrix} [options.matrix=null] - Transform matrix
      * @return {PIXI.Graphics} This Graphics object. Good for chaining method calls
      */
-    beginTextureFill(options)
+    beginTextureFill(options: IFillStyleOptions | Texture): this
```
bigtimebuddy:
```suggestion
    beginTextureFill(options: IFillStyleOptions): this
```

packages/graphics/src/Graphics.ts
```diff
@@ -763,33 +837,27 @@ export class Graphics extends Container
...
         {
-            closeStroke = points.closeStroke;
-            points = points.points;
+            closeStroke = (path as Polygon).closeStroke;
+            points = (path as Polygon).points;
         }
-
-        if (!Array.isArray(points))
+        else
         {
```
bigtimebuddy: Why remove support for `...args:number[]` here? These TS conversions shouldn't remove functionality.

packages/graphics/src/styles/FillStyle.ts
```diff
@@ -0,0 +1,95 @@
+import { Texture } from '@pixi/core';
```
bigtimebuddy: Why didn't FillStyle show up as a git move?

packages/graphics/src/Graphics.ts
eXponenta: Same problem in Poligon package now. We doesn't support arrays in constructor now.
Can you write 2 annotation with ... and [], because i wasn't resolve this.

packages/graphics/src/Graphics.ts
eXponenta: we can't assign in readonly field. Blocked by TS.
`this.geometry = geometry || new Graphics Geometry();`
Or protected or public or change constructor or getter.

bigtimebuddy: I thought you could assign readonly in the constructor. Hmm. If that doesn't work, then private with getter I guess.

eXponenta:
> Why did you remove buildComplexPoly and bezierCurveTo functions?

Dead and doublicated code.
There aren't usage of it, and wasn't public because there no any exports.
Then `buildComplexPoly` is broken becouse used deaded render features, bezierCurveTo is same as BezierUtils.curveTo.

packages/graphics/src/Graphics.ts
eXponenta: Sorry, it is valid, assigment in constructor is really works. Thanks for insisting!

eXponenta: But now we can't release it =) because destrou method assign null to propery, but in readonly!
Cool!

eXponenta:
Conclusion:
- `public readonly geometry` can't be used because destroy method assign null to them, of course we always can do `(this as any).geometry = null` but what for?
- we can't create more that one method/constructor according to lint rule and some collaborators don't like this crutch ( such as me or @ivanpopelyshev )

packages/graphics/src/styles/FillStyle.ts
eXponenta: So, in github it look as move and change. Not as deleting,
image

bigtimebuddy: Weird, can’t trust GitHub UI

packages/graphics/src/Graphics.ts
eXponenta: Oke, i restore it.
Tests will come later.

packages/graphics/src/Graphics.ts
eXponenta: Ok.
Now `geometry` is getter, `_geometry` is private.
API and JSDoc wasn't change (can get, can't set), but we can assign to `_geometry` internaly.

bigtimebuddy:
Thank you @eXponenta :tada:

ivanpopelyshev:
@bigtimebuddy do you even sleep?

bigtimebuddy:
@ivanpopelyshev newborn duty 👶

PR #6354, opened 2020-01-27
workaround for firefox/mesa only allowing up to 16 samplers

laino:
##### Description of change
Fixes

``
Pixi.js Warning: gl.getProgramInfoLog() Programs with more than 16 samplers are disallowed on Mesa drivers to avoid crashing.
``

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/shader/utils/checkMaxSamplersInShader.js
+++ b/packages/core/src/shader/utils/checkMaxSamplersInShader.js
@@ -1,0 +1,20 @@
+export function checkMaxSamplersInShader(maxSamplers, gl)
+{
+    if (maxSamplers === 0)
+    {
+        throw new Error('Invalid value of `0` passed to `checkMaxSamplersInShader`');
+    }
+
+    const debugInfo = gl.getExtension('WEBGL_debug_renderer_info');
+    const renderer = gl.getParameter(debugInfo.UNMASKED_RENDERER_WEBGL);
+
+    // Mesa drivers may crash with more than 16 samplers and Firefox
+    // will actively refuse to create shaders with more than 16 samplers.
+    if (renderer.slice(0, 4).toUpperCase() === 'MESA')
+    {
+        maxSamplers = Math.min(16, maxSamplers);
+    }
+
+    return maxSamplers;
+}
+
--- a/packages/core/src/batch/AbstractBatchRenderer.js
+++ b/packages/core/src/batch/AbstractBatchRenderer.js
@@ -6,6 +6,7 @@ import { State } from '../state/State';
 import { ViewableBuffer } from '../geometry/ViewableBuffer';
 
 import { checkMaxIfStatementsInShader } from '../shader/utils/checkMaxIfStatementsInShader';
+import { checkMaxSamplersInShader } from '../shader/utils/checkMaxSamplersInShader';
 
 import { settings } from '@pixi/settings';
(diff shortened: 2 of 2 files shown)


eXponenta:
Hah, this is not clean fix, because on intel 3000 warning says about 16 textures. but really works only 8.
We have idea about blacklisting some GPU as it Chrome do.

But any `core` PR will loked before core will be converted to TS.

laino:
No it's not a clean fix. I don't think a clean fix is really possible (maybe in WebGL 2?).

Most of this diff lives in ``src/shader`` though and in a way that it could be extended to cover more special GPU cases.

ivanpopelyshev:
i think you have to rebase it on `master` branch to get it in 5.2.1

bigtimebuddy:
Is this the most future-proof approach? How much should we really be considering certain hardware/drivers, especially since devices change so quickly?

I'm more interested in an approach that uses `settings` where users can provide some global information (or hooks) for users to handle specific hardware environments themselves. An example, would be the PREFER_ENV thing we added recently. There's value in users owning some of the device problems. For example, we routinely advise users to turn on Renderer transparency to handle very old iOS/Android, but we haven't baked that logic into the library.

Let's keep this on **dev** it needs some more discussion and will not be going into 5.2.1. Would like to get @GoodBoyDigital's take.

laino:
Here's my take for what it's worth.

- This is mostly about auto-tuning a number according to the environment. Which is pretty far from peppering the codebase with if-else branches to handle specific hardware.

- The whole batch rendering shader thing is a performance optimization pixijs decided to implement, so I'd say it's pixijs' responsibility to not break random platforms with it. If there's a considerable amount of platforms (namely intel integrated graphics) for which pixijs decides to *not* implement custom workarounds, then the *default* values should be adjusted so it works everywhere (``max samplers = 8`` or something).

- Let developers opt-out by increasing that value if they prefer performance over platform support. Having broken or non-existent support for a huge chunk of Intel graphics out-of-the-box looks bad. Especially because it's not mentioned anywhere.

laino:
> Is this the most future-proof approach? How much should we really be considering certain hardware/drivers, especially since devices change so quickly?

Maybe this could be implemented in the same way pixi checks for the maximum number of ``if-else`` branches in a shader - maybe the checks could even be combined. That would be future-proof as long as the browsers take care of warning us if we added to many.

However the problem appears to be that having more than 8/16 samplers is broken in a way that doesn't result in an immediate error on some systems - and only FF decided to disallow creating such shaders on systems that may be affected.

ivanpopelyshev:
Maybe we need to put that constant to TextureSystem. Guys, what do you think?

packages/core/src/shader/utils/checkMaxSamplersInShader.js
```diff
@@ -0,0 +1,20 @@
+export function checkMaxSamplersInShader(maxSamplers, gl)
+{
+    if (maxSamplers === 0)
+    {
+        throw new Error('Invalid value of `0` passed to `checkMaxSamplersInShader`');
+    }
+
```
GoodBoyDigital: hey ya, 'debugInfo' could be null on this check. This would then throw an error.

https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_debug_renderer_info

if you could cater for that, then this PR will be good to go 👍 

Cheers!

PR #6361, opened 2020-01-29
Add eslint rule for explicit accessibility modifiers

Zyie:
##### Description of change

As mentioned by @EvidentlyCube in the core pr #6340 we should have an eslint rule to enforce some consistency about how accessibility modifiers are annotated.

Currently this is [rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md) is configured to force declaring modifiers on methods and properties.

some basic sudo code to explain current rules:
```
public x: string;
private _y: string;

constructor(){}

public test(){}

get y(){}
set y(){}
```

We dont have to use this exact configuration so lets discuss it.
##### Pre-Merge Checklist

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -255,7 +255,8 @@
           "rules": {
             "@typescript-eslint/semi": 1,
             "@typescript-eslint/no-explicit-any": 0,
-            "@typescript-eslint/interface-name-prefix": 0
+            "@typescript-eslint/interface-name-prefix": 0,
+            "@typescript-eslint/explicit-member-accessibility": [1, { "overrides": { "constructors": "no-public", "accessors": "no-public" } }]
           }
         }
     ]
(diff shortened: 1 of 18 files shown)


Zyie:
I think having a few edge cases in core is fine and we can turn the rule off for these exceptions.

Zyie:
> The problem with `@pixi/core` is that many fields are protected or private by nature but are referenced from other files. TypeScript doesn't have `default` access modifier, or custom ones like in ActionScript.
> 
> I prefer to just leave fields without modifiers in that case. "Yes, it shouldn't be public, but plugins and core itself have to access this"

ivanpopelyshev:
> I think having a few edge cases in core is fine and we can turn the rule off for these exceptions.

That's the problem! They are not few!

PR #6362, opened 2020-01-29
Add eslint rule to force I prefix for interfaces

Zyie:
Another eslint rule to enforce that interfaces start with I.

Throughout our conversion this seems to have become how interfaces are named so we might as well enforce it.

I believe we have to disable this rule in certain circumstances e.g. mixins, but its a pretty minor inconvenience.
##### Pre-Merge Checklist

### DIFF
--- a/packages/utils/src/network/decomposeDataUri.ts
+++ b/packages/utils/src/network/decomposeDataUri.ts
@@ -1,6 +1,6 @@
 import { DATA_URI } from '../const';
 
-export interface DecomposedDataUri {
+export interface IDecomposedDataUri {
     mediaType: string;
     subType: string;
     charset: string;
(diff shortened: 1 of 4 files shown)


packages/utils/src/network/decomposeDataUri.ts
```diff
@@ -51,7 +51,7 @@ export interface DecomposedDataUri {
  * @param {string} dataUri - the data URI to check
  * @return {PIXI.utils.DecomposedDataUri|undefined} The decomposed data uri or undefined
```
bigtimebuddy: What about the doc for DecomposedDataUri? Should that be updated too?

Zyie: Ah yes i didnt see that!

bigtimebuddy:
I'm willing to be overruled, but I want to throw out an alternative to provide more meaning to interface across the code base.

In general, I like `I` as a prefix for functional interfaces (e.g. some class implements interface). A good example is `IPoint`, which is implemented by `OberservablePoint` and `Point` classes. But for strict type definitions, like RendererOptions, I do not think that prefix is appropriate. I know that TS refers to both of these as interfaces, but I think the prefix is only really important for things that require `implements ISomething`, otherwise its just another type definition like `export type Blah = number | string | null`. Distinguishing between these two types of things, I think, is important. So I would vote to reject this PR and leave thing as-is, but with some agreement about how we might label things.

Issue #6355, opened 2020-01-27
getLocalBounds() causes transform to update, can cause problems in render pass


SukantPal:
I have found that calling `DisplayObject#getLocalBounds` can mess things up if called while the renderer is rendering. This is because this method will cause a transform update after removing the parent.

In many parts of `@pixi/core`, we use `DisplayObject#getBounds(false)`, which will not update the transform to _include_ the parent's transform. This is an inherent weakness in PixiJS and should be properly documented.

To solve the problem, you _must_ call update-transform after getting the local bounds.

```js
target.getLocalBounds();
target.updateTransform();
```

**However, code like `Container#width` (getter) already uses the `getLocalBounds()` method. This means that trying to access a display-object's width will result in failure of the renderer.**

Maybe we should solve this? @ivanpopelyshev thoughts?

SukantPal:
Also, using `getLocalBounds()` every time results in heavy calculations to the transform. I think it is _very inefficient_ for display-object's width to depend on this.

`getBounds()` prevents `calculateBounds()` via the `this._bounds.updateID == _boundsID` check. However, using update-transform breaks that condition and forces the `calculateBounds`.

_Even there are no real changes in the object's transform, we are causing recalculations of the bounds._

SukantPal:
**Sidenote: Display object bounds are updated only if the transform changes. There is no way to communicate that the inherent dimensions of the object have changed. This is another weakness in PixiJS.**

bigtimebuddy:
Could you please provide a working example (jsfiddle, codepen, etc) of the issue? I’m not sure what you mean by “mess things up”.

SukantPal:
@bigtimebuddy Sorry for the delay, here is a demo: https://jsfiddle.net/ShukantPal/bjc0te8x/7/.

At line 26, the `getLocalBounds()` method causes the Sprite to render at 128x128 instead of the scaled up 256x256 (due to 2x scale of stage).

ivanpopelyshev:
Congratulations, you've stumbled across bug that I cant properly fix for 5 yaers! Yes, using those in render and interaction can .

Unfortunately, fixes that i tried to make a few years ago were too big to merge.

The real fix should remove global matrices update from `getLocalBounds`. If you somehow need to update coords of elements in `render`, then even those fixes wont help, because transform inside `render()`s is a VERY bad architectural idea. In my apps I have separate `prerender()` pass where all filters and caches are measured, and elements re-positioned.

[...]

PR #6365, opened 2020-01-30
Fix local bounds problems. Swap bounds with localBounds to do the trick.

SukantPal:
Fix for #6355 

By swapping the `_bounds` object with `_localBounds`, we are preventing `calculateBounds` from editing the world-bounds.

Also, I am syncing the changes in the `_boundsID` with the world-bounds `updateID`. This will prevent the world-bounds from getting recalculated in the next `getBounds(true)` if it was already update-to-date.


SukantPal:
Hmm, this doesn't solve the JSFiddle I presented. On inspection, the `_bounds` of the object are correct now, except the sprite is still not rendering correctly.

SukantPal:
The sprite still renders wrongly because `Container` will update the transform of its children too. The rendering process does not update transforms.

SukantPal:
@ivanpopelyshev @bigtimebuddy This fixes the problem!

SukantPal:
I think instead of reversing update-transform on the children in `Container#getLocalBounds`, I will change this so that we use `getLocalBounds` when calculating bounds of Container.

SukantPal:
I think I will do one more optimization tomorrow: on a container's children, we should swap the transform with a temporary transform instance. Then we won't have to call update-transform again after `getLocalBounds`.

bigtimebuddy:
It is important if you could capture this modified bounds behavior in a unit test, preferably one that breaks on dev. That’s the best way to ensure that we don’t have a regression.

packages/display/src/Container.ts
```diff
@@ -479,6 +480,26 @@ export class Container extends DisplayObject
         this._bounds.updateID = this._boundsID;
     }
 
+    public getLocalBounds(rect?: Rectangle, noTransformSideEffects = true): Rectangle
```
eXponenta: You override public method.
JSDoc is required.

`noTransformSideEffects` what this is mean and why this is `true` by default?

I think it is must be `skipUpdateTransfrom` how in getBounds method.

ivanpopelyshev:
As I said in the issue:
1. I believe both of `getLocalBounds` in Container and DisplayObject should have `_` version of methods  because we have overrides in `Sprite` and there might be an override in spine library
2. I'm sure I can make a test case when this thing fails. I'll work on it in weekend
3. Try integrate my idea with axis-alighned fallback. However, with right architecture i can integrate it later.

packages/display/src/Container.ts
SukantPal: As I indicated before, this is just a sketch PR. That is why no JSDoc yet.

SukantPal:
@ivanpopelyshev Your idea could be added in another PR - when it becomes your priority.

SukantPal:
Overrides will not be affected at all. This should be a backward compatible change after I swap the child transforms too.

SukantPal:
@ivanpopelyshev I’m looking for your tests today that should fail with this change.

ivanpopelyshev:
OK, I wanted to merge dev into it but made something wrong. Can you please re-do it and push with force?

ivanpopelyshev:
OK, never mind, I've rebased it myself.

ivanpopelyshev:
We have to update all objects (Sprites, e.t.c.) `getLocalBounds` to pass that param. 

@bigtimebuddy we have a problem that `getBounds()` has first argument bool and second Rectangle, and for `getLocalBounds` its rectangle and then bool :) The heck we do in this case? I propose we add `skipChildrenUpdate` here as the first param. That way both methods will have the same signature and `false, null` by default.

packages/display/src/Container.ts
SukantPal: I removed that parameter.

SukantPal:
@ivanpopelyshev Please recheck the standing of the changes.

eXponenta:
> @ivanpopelyshev Please recheck the standing of the changes.

Doesn't work at all in real cases =)

DEV:
https://www.pixiplayground.com/#/edit/5uTdn6fP5w-DJrLXKuPWL

Your:
https://www.pixiplayground.com/#/edit/Qyl4KFWghAu3AwXfRLOSV

SukantPal:
@eXponenta I already know. Review the types of changes first and then I’ll fix the bugs today.

eXponenta:
> @eXponenta I already know. Review the types of changes first and then I’ll fix the bugs today.

How should we review it when it doesn't works properly?

SukantPal:
@eXponenta Sometimes people don’t like me changing something even if it works

ivanpopelyshev:
Don't worry, we'll take care of it in weekend together.

SukantPal:
@ivanpopelyshev Any updates?

ivanpopelyshev:
Cant commit, one test is failing because I added extra condition: two consecutive `getLocalBounds` calls shoould not trigger second `calculateBounds`.

SukantPal:
@ivanpopelshev Will look at it in evening

ivanpopelyshev:
Yeah, if you manage to adjust `boundsID` hacks that way tests passes - it will be awesome!

cursedcoder:
Looks good, but tests.

ivanpopelyshev:
My fault - I made extra test that I thought can be passed with this PR, but it turns out something is wrong with boundsID's. I'm waiting for @SukantPal  when he will be free to investigate whether i made a mistake or he did

SukantPal:
@ivanpopelyshev I pushed changes to make the tests to pass. I following breaking changes will occur:

* Display objects that have a change in their natural bounds will _have to declare that by updating `_boundsID`_. They will not get bounds update for free as before (because `updateTransform` will not be called if bounds are already update-to-date in `getLocalBounds`).

---

* `Graphics.clear()` updates `_boundsID`.

bigtimebuddy:
Could you help me understand the breaking change more? If you could show some code of what would change, that’d be helpful.

SukantPal:
@bigtimebuddy I'll explain the "breaking" change a bit rigorously as follows:

Let's say we have a display-object `object`;  it is (and was) expected that if its transform has not changed and `object._boundsID` is not dirty then`object.getBounds()` will calculate the same bounds.

Now, the local bounds are, by definition, defined as bounds _without any transforms applied_. That leaves us with: if `object._boundsID` is not marked dirty, then `object.getLocalBounds()` will calculate the same value.

---

Previously, `getLocalBounds` always deferred to `getBounds` - which always did an `updateTransform`. Now, `updateTransform` marks `_boundsID` dirty, well because it thinks that the transform has changed.

Hence, since `_boundsID` was made dirty internally in `getLocalBounds` - that meant `calculateBounds` was called each and every time `getLocalBounds` was called, and that authors of display-objects (like `PIXI.Graphics#clear()`) did not have to worry about marking their bounds dirty.

**This assumption is now incorrect with this PR**. If `object._boundsID` is in-sync with `this._bounds.updateID`, then the local bounds won't be updated.

---

## Why?

By _not_ updating the local bounds when they aren't marked dirty, we prevent unnecessary bounds calculations and transform updates.

## Future work

Similarly, we could prevent `getBounds()` updates if the transform itself has not changed after a `updateTransform`. That would be done in a different PR.

## Another use-case of this PR

When the scene graph is being rendered, it is expected that `this._bounds` are update-to-date. However, calling `getLocalBounds()` previously changed what `getBounds(true)` returned. This is because `getLocalBounds()` would store its result in `this._bounds` and `getBounds(true)` wouldn't update `_bounds` (no transform update). This results in many errors throughout PixiJS (because the bounds returned by the node are incorrect).

This PR swaps `this._bounds` with another `this._localBounds`. This way `getLocalBounds` will not affect the result of `getBounds(true)`. Sadly though, it is still unable to prevent `getLocalBounds` from affecting the `getBounds(true)` value of grandchildren (and nodes below 1st generation children). I will work on a performant fix for that later on.

SukantPal:
@bigtimebuddy We have a problem with parent containers whose local bounds depend on the transform of children in the parent’a reference frame.

I will push a fix for that, if possible.

bigtimebuddy:
@SukantPal, I think the work that you've done here is terrific. I commend you for optimizing these code paths related to calculating bounds, because it's subtle, difficult work.

One of the challenges of PRs like this, is that I (and probably others on the PixiJS team), don't fully understand the extent to which this may produce unwelcome side-effects for developers. It's one thing add unit-tests, it's another to do a full integration where users heavily rely on bounds.

Our goal as an open source library is, most importantly, is to make sure we do no harm. I think it's important that you can work with developers (@themoonrat, @GoodBoyDigital, and others) to test this build into real-world, complex apps and games. As much as you're able to prove and show that this will not negatively impact users, it will be easier to approve.

SukantPal:
@bigtimebuddy I have removed the sync barrier so there are no side-effects. I think @ivanpopelyshev would be okay to merge this now. (So this PR now only prevents getLocalBounds from storing its result in this._bounds, preventing corruption this.getBounds(true), via the _localBounds swap).

To prevent unnecessary bounds updates, we will need to implement a system where transform updates are propagated up the scene graph. This will be experimented on another branch.

ivanpopelyshev:
There's only one thing left - `_localBounds` is still created in constructor, and its triggered only by calling `getlocalBounds` of the same element - it better be a lazy variable.

GoodBoyDigital:
Hey @SukantPal , thanks for this! Going to test this branch on one of our more layout heavy games! Will let you know how it goes :D

PR #6366, opened 2020-01-30
Defines for shader and preprocessor

amakaseev:
Added shader defines as optional parameter for use custom defines in programs.
In the future I think to add custom preprocessing.


ivanpopelyshev:
Good for initial draft :)

amakaseev:
Example usage:

  if (!this._shader) {
    let defines = {};

    if (this._texture) {
      defines.TEXTURE = '';
    }

    if (this._alphaCutoff) {
      defines.ALPHA_CUTOFF = '';
    }

    if (this._lights) {
      defines.LIGHTS = '';
    }

    this._shader = new PIXI.Shader(PIXI.Program.from(defaultVertex, defaultFragment, 'default', 
{defines: defines}));
  }

[default.txt](https://github.com/pixijs/pixi.js/files/4134162/default.txt)

packages/core/src/shader/utils/Preprocessor.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * Created by hazed on 30.01.2020.
+ */
```
cursedcoder: Is it needed?

packages/core/src/shader/utils/Preprocessor.ts
```diff
@@ -0,0 +1,78 @@
...
+            else
+            {
+                nameCache[name] = 1;
+            }
+
+            // Setup default define for shader name
+            options.defines.SHADER_NAME = name;
+
+            // Setup custom defines
+            for (const define in options.defines)
```
cursedcoder: I believe you can use `for (const value of options.defines)`

packages/core/src/shader/utils/Preprocessor.ts
```diff
@@ -0,0 +1,78 @@
...
+            {
+                const value = options.defines[define];
+
+                this.vertex = `#define ${define} ${value}\n${this.vertex}`;
+                this.fragment = `#define ${define} ${value}\n${this.fragment}`;
```
cursedcoder: Will it overwrite it on every iteration?

packages/core/src/shader/utils/Preprocessor.ts
bigtimebuddy: Agreed, please remove. We use git history to determine authorship and time stamping.

packages/core/src/shader/utils/Preprocessor.ts
```diff
@@ -0,0 +1,78 @@
...
+        options.isRawShader = options.isRawShader || (this.vertex.substring(0, 8) === '#version');
+
+        if (!options.isRawShader)
+        {
+            name = name.replace(/\s+/g, '-');
+
+            if (nameCache[name])
+            {
+                nameCache[name]++;
+                // TODO: incremented name in shader code breaks ProgramCache functionality
```
GoodBoyDigital: would be good to resolve this todo!

packages/core/src/shader/utils/Preprocessor.ts
amakaseev: ok, its auto-generate by WebStorm

packages/core/src/shader/utils/Preprocessor.ts
```diff
@@ -0,0 +1,78 @@
...
+            options.defines.SHADER_NAME = name;
+
+            // Setup custom defines
+            for (const define in options.defines)
+            {
+                const value = options.defines[define];
+
+                this.vertex = `#define ${define} ${value}\n${this.vertex}`;
+                this.fragment = `#define ${define} ${value}\n${this.fragment}`;
+            }
```
GoodBoyDigital: this is cool, i think a neater (more performant) way would be to generate the defines in an array and concat them to a string:

```
const defineStrings = [];

for (const define in options.defines)
{
    defineStrings.push(`#define ${define}`);
}

const finalDefines =  defineStrings.join(`\n`);
this.vertex = finalDefines + this.vertex;
this. fragment = finalDefines + this. fragment;

```

packages/core/src/shader/utils/Preprocessor.ts
amakaseev: no, defines must have key and value, as example:
```
#define MAX_BONES 64
uniform mat4 boneMatrices[MAX_BONES];
```

PR #6374, opened 2020-02-01
Fix ImageResource load behaviour

eXponenta:
Rejecting `load` when resource loading throw error.

Test added.


eXponenta:
Ups, based on #6373

PR #6375, opened 2020-02-01
Convert Sprite to TypeScript

Zyie:
##### Description of change

Converts the sprite package to TypeScript. 

The main changes are:
- Created `TextureSource` and `SpriteSource` types
- Made `_worldID` in `Transform` be public
##### Pre-Merge Checklist


packages/core/src/textures/Texture.ts
```diff
@@ -10,6 +10,8 @@ import { TextureMatrix } from './TextureMatrix';
 
 const DEFAULT_UVS = new TextureUvs();
 
+export type TextureSource = string|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement|BaseTexture;
```
ivanpopelyshev: you can use `ImageSource` from `@pixi/core` here instead of `HTMLxxx` stuff

ivanpopelyshev: Btw, we have to test if `Sprite.from(resource)` works too

packages/math/src/Transform.ts
```diff
@@ -25,6 +25,7 @@ export class Transform
     public pivot: ObservablePoint;
     public skew: ObservablePoint;
     public _parentID: number;
+    _worldID: number;
```
ivanpopelyshev: see, that's why we need empty modifiers :)

packages/sprite/src/Sprite.ts
```diff
@@ -36,10 +38,37 @@ const indices = new Uint16Array([0, 1, 2, 0, 2, 3]);
  */
 export class Sprite extends Container
 {
+    public blendMode: BLEND_MODES;
+    public shader: Filter|Shader;
```
ivanpopelyshev: Sprites dont have usable `shader` field at the moment. Its old thing from v3. I do not know how did it survive.

packages/sprite/src/Sprite.ts
```diff
@@ -1,13 +1,15 @@
-import { Point, ObservablePoint, Rectangle } from '@pixi/math';
-import { sign } from '@pixi/utils';
-import { Texture } from '@pixi/core';
 import { BLEND_MODES } from '@pixi/constants';
-import { Container } from '@pixi/display';
+import { Shader, Texture, Filter, Renderer, IBaseTextureOptions, TextureSource } from '@pixi/core';
```
ivanpopelyshev: My hands just want to separate it to real `import` and `import type` but we cant do that before new TS is released, also our linter doesnt like two import lines from same module

packages/sprite/src/Sprite.ts
```diff
@@ -271,18 +309,18 @@ export class Sprite extends Container
         {
             // if the sprite is trimmed and is not a tilingsprite then we need to add the extra
             // space before transforming the sprite coords.
-            w1 = trim.x - (anchor._x * orig.width);
+            w1 = trim.x - (anchor.x * orig.width);
```
ivanpopelyshev: See the problem? @finscn wont be happy that we removed his "optimizations". That's a kind of hotspot and it uses extra  property calls. LEts confirm that @GoodBoyDigital is ok with that.

Anchor is always ObservablePoint so maybe we need to make `_x` and `_y` visible, its not the only place we use them.

packages/core/src/textures/Texture.ts
Zyie: Ok I've updated the type. I've not really written any tests, is that something you can help me with?

packages/sprite/src/Sprite.ts
Zyie: I've made `_x` and `_y` internal and reverted these changes

packages/sprite/src/Sprite.ts
Zyie: cool, I've removed shader

packages/core/src/textures/Texture.ts
ivanpopelyshev: I dont mean to write test, just to check if it works in pixi-playground. I can do it tomorrow

ivanpopelyshev: Never mind, its part of our discussion about texture resource parameters.

PR #6376, opened 2020-02-01
Convert App to TypeScript

Zyie:
##### Description of change

Converts app package to TypeScript.

The main changes are:
- Moved `IApplicationOptions` out of `global.d.ts` and now extends `IRendererOptionsAuto 
- Created `IApplicationPlugin` interface

There is a change that will need to be made once `CanvasRenderer` is converted. I've documented this with TODO's, but essentially there is some unnecessary casting happening to make TypeScript happy.

##### Pre-Merge Checklist


packages/app/src/Application.ts
```diff
@@ -136,7 +138,9 @@ export class Application
      */
     get screen(): Rectangle
     {
-        return this.renderer.screen;
+        // TODO: Since CanvasRenderer has not been converted it thinks screen doesn't exists. This can be removed once
+        // converted
+        return (this.renderer as any).screen;
```
eXponenta: Renderer extend from AbstractRenderer, you can cast to it.

packages/app/src/Application.ts
```diff
@@ -126,7 +126,9 @@ export class Application
      */
     get view(): HTMLCanvasElement
     {
-        return this.renderer.view;
+        // TODO: Since CanvasRenderer has not been converted it thinks view doesn't exists. This can be removed once
+        // converted
+        return (this.renderer as any).view;
```
eXponenta: (this.renderer as any)  => (this.renderer as AbstractRenderer)

packages/ticker/src/TickerPlugin.ts
```diff
@@ -1,5 +1,6 @@
 import { Ticker } from './Ticker';
 import { UPDATE_PRIORITY } from './const';
+import { IApplicationOptions } from '../../app/src/Application';
```
bigtimebuddy: No relative importing of packages/types. This makes the structure of the project more brittle. Simply do `export * from './Application'` in the src/index.ts of app and then import this from `@pixi/app`

packages/app/src/Application.ts
```diff
@@ -1,5 +1,20 @@
-import { Container } from '@pixi/display';
-import { autoDetectRenderer } from '@pixi/core';
+import { Container, IDestroyOptions } from '@pixi/display';
+import { autoDetectRenderer, Renderer, IRendererOptionsAuto, AbstractRenderer } from '@pixi/core';
+import { CanvasRenderer } from '@pixi/canvas-renderer';
```
ivanpopelyshev: I still dont understand is it ok to reference types from modules without direct link and how will that affect TS apps that dont use `pixi.js-legacy` . Guess we'll find out later :) At least CanvasRenderer is not build into `pixi.js` - that one i have chacked.

bigtimebuddy: I'm not actually sure this is going to fly when we generate types. We may need a more generic Renderer interface. For building the project it is fine, because node's resolve. But because `@pixi/canvas-renderer` is not a dependency, this will create issues for users later on. @zyie, do you want to try to resolve this now or later?

bigtimebuddy: Another thought, what if we don't use CanvasRenderer as a type anywhere and just use Renderer? CanvasRenderer is a subset of Renderer functionality.

Zyie: I've removed `CanvasRenderer` and replaced it with `AbstractRenderer`.

I can look into a better solution over the next couple of days and do a separate PR to fix this and #6379

PR #6379, opened 2020-02-01
Convert Accessibility to TypeScript

Zyie:
##### Description of change

Converts accessibility to TypeScript.

Fairly simple conversion, just adding another mixin for DisplayObject and a new interface for an accessible HTMLElement.

I have removed the for loop that set its children's `div` to null in `destroy`. I couldn't find any example of `div` being set so i believe this is not needed?
##### Pre-Merge Checklist


adireddy:
# [Codecov](https://codecov.io/gh/pixijs/pixi.js/pull/6379?src=pr&el=h1) Report
> Merging [#6379](https://codecov.io/gh/pixijs/pixi.js/pull/6379?src=pr&el=desc) into [dev](https://codecov.io/gh/pixijs/pixi.js/commit/ad4230b32f03cea4076c01953e97b7237d049a20&el=desc) will **increase** coverage by `1.45%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/pixijs/pixi.js/pull/6379?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##              dev    #6379      +/-   ##
==========================================
+ Coverage   74.70%   76.16%   +1.45%     
==========================================
  Files         105      103       -2     
  Lines        6053     5840     -213     
==========================================
- Hits         4522     4448      -74     
+ Misses       1531     1392     -139     
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/pixijs/pixi.js/pull/6379?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/pixijs/pixi.js/pull/6379?src=pr&el=footer). Last update [bcd040e...90109d5](https://codecov.io/gh/pixijs/pixi.js/pull/6379?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

packages/accessibility/src/AccessibilityManager.ts
```diff
@@ -548,15 +579,17 @@ export class AccessibilityManager
...
-        if (!e.target.getAttribute('aria-live', 'off'))
+        if (!(e.target as Element).getAttribute('aria-live'))
         {
-            e.target.setAttribute('aria-live', 'polite');
+            (e.target as Element).setAttribute('aria-live', 'polite');
         }
-        const interactionManager = this.renderer.plugins.interaction;
 
-        interactionManager.dispatchEvent(e.target.displayObject, 'mouseout', interactionManager.eventData);
+        // TODO: Remove casting when CanvasRenderer is converted
```
ivanpopelyshev: Yes.

packages/display/src/DisplayObject.ts
```diff
@@ -3,14 +3,15 @@ import { DEG_TO_RAD, IPoint, Matrix, ObservablePoint, Point, RAD_TO_DEG, Rectang
...
+import { IAccessibleTarget } from '@pixi/accessibility';
 
 export interface IDestroyOptions {
     children?: boolean;
     texture?: boolean;
     baseTexture?: boolean;
 }
 
-export interface DisplayObject extends InteractiveTarget, EventEmitter {}
+export interface DisplayObject extends InteractiveTarget, IAccessibleTarget, EventEmitter {}
```
ivanpopelyshev: Another cross-module reference to check when we convert all the modules.

packages/accessibility/src/AccessibilityManager.ts
```diff
@@ -32,16 +34,30 @@ const DIV_HOOK_ZINDEX = 2;
  */
 export class AccessibilityManager
 {
+    public readonly isActive: boolean;
```
ivanpopelyshev: @GoodBoyDigital @Zyie lets make those two readonly prop + private field.

eXponenta: Private + getter is more usable in this case becouse ot reassing a few times.

Zyie: Ok I've changed both of these to be private and have a getter.

ivanpopelyshev:
@Zyie `private` shouldn't be `readonly`

Zyie:
> @Zyie `private` shouldn't be `readonly`. That's why we want `private` field - no need for `any` in constructor and other places.

Yeah my bad I popped the wrong stash without realising

ivanpopelyshev:
Now you can change`(this as any).isActive = true` to `this._isActive = true`

Zyie:
> Now you can change`(this as any).isActive = true` to `this._isActive = true`

Done

ivanpopelyshev:
"Amputate left hand. 
I said left. 
I said hand!"

packages/accessibility/src/AccessibilityManager.ts
```diff
@@ -1,7 +1,9 @@
+import { CanvasRenderer } from '@pixi/canvas-renderer';
```
bigtimebuddy: We need to find a better solution. Importing CanvasRenderer like this is going to mess up the dependency tree for a lot of packages that need to know about the renderer. My suggestion is that we just use `AbstractRenderer | Renderer`. That should cover CanvasRenderer without having to import it explicitly.

Zyie: Ok I've updated this to remove `CanvasRenderer` for `AbstractRenderer`

PR #6381, opened 2020-02-03
Convert Extract to TypeScript

Zyie:
##### Pre-Merge Checklist


bigtimebuddy:
Do you think it makes sense to create a IRendererPlugin interface that Extract, Prepare, etc that they can all implement? This would basically just take renderer as a single constructor argument. Not sure if there are any other methods called by Renderer. I like hardening plugin types for users to implement to.

Zyie:
> Do you think it makes sense to create a IRendererPlugin interface that Extract, Prepare, etc that they can all implement? This would basically just take renderer as a single constructor argument. Not sure if there are any other methods called by Renderer. I like hardening plugin types for users to implement to.

I didn't think it was possible to implement an interface that defines a constructor? Could you point to an example for me.

We could create a base class that plugins extend from if an interface doesn't work?

bigtimebuddy:
Would probably look something like this:

```typescript
interface IRendererPluginConstructor {
  new (renderer: Renderer): IRendererPlugin;
}

interface IRendererPlugin {
  destroy(): void;
}
```
So the Renderer [here](https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/Renderer.ts#L491) would be:
```
registerPlugin(static registerPlugin(pluginName: string, ctor: IRendererPluginConstructor): void
```

bigtimebuddy:
Is this good to go, @Zyie? Was there something else you wanted to do?

Zyie:
> Is this good to go, @Zyie? Was there something else you wanted to do?

Sorry got super busy last week. Just updated this to add the `IRendererPlugin` like you suggested.

So this is now good to go

PR #6382, opened 2020-02-03
Convert Mesh to TypeScript

Zyie:
##### Description of change

Converts Mesh to TypeScript. Nice and simple
##### Pre-Merge Checklist


packages/mesh/src/Mesh.ts
```diff
@@ -399,7 +427,7 @@ export class Mesh extends Container
         }
         else
         {
-            this.uvs = geomUvs.data;
+            this.uvs = geomUvs.data as Float32Array;
```
ivanpopelyshev: We can use `ITypedArray` in such places in future, but its fine to convert it if we know certain type.

packages/mesh/src/Mesh.ts
```diff
@@ -477,20 +505,21 @@ export class Mesh extends Container
             this.geometry.dispose();
         }
 
-        this.geometry = null;
+        (this as any).geometry = null;
```
ivanpopelyshev: @eXponenta changed geometry to private field + readonly prop, but for `destroy` both ways are fine and I want to underline that by approving this PR :)

packages/mesh/src/MeshGeometry.ts
```diff
@@ -20,12 +20,17 @@ import { Buffer, Geometry } from '@pixi/core';
...
+    // @ts-ignore
+    private _updateId: number;
+
     /**
      * @param {Float32Array|number[]} vertices - Positional data on geometry.
      * @param {Float32Array|number[]} uvs - Texture UVs.
      * @param {Uint16Array|number[]} index - IndexBuffer
      */
-    constructor(vertices, uvs, index)
+    constructor(vertices: IArrayBuffer, uvs: IArrayBuffer, index: IArrayBuffer)
```
ivanpopelyshev: yay! That's why I created IArrayBuffer.

Zyie:
@ivanpopelyshev I do have a question for this pr. 

Inside `MeshGeometry` there is `_updateId`. I couldn't find any reference to it so is it needed?

PR #6383, opened 2020-02-03
Convert Filters to TypeScript

Zyie:
##### Description of change

Converts all filters to TypeScript and another fairly simple conversion.

Main two changes are:
- Adds a `ArrayFixed` type which defines an array of fixed length
- Adds `ColorMatrix` type that is a `ArrayFixed` with a of size 20

##### Pre-Merge Checklist


packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
```diff
@@ -40,14 +42,14 @@ export class ColorMatrixFilter extends Filter
      * @param {boolean} multiply - if true, current matrix and matrix are multiplied. If false,
      *  just set the current matrix with @param matrix
      */
-    _loadMatrix(matrix, multiply = false)
+    private _loadMatrix(matrix: number[], multiply = false): void
```
bigtimebuddy: This class might benefit from a more strict matrix with a fixed length:

```typescript
// T stands for (content) type, L stands for the array fixed length
type ArrayFixed<T, L extends number> = [ T, ...Array<T> ] & { length: L }
type ColorMatrix = ArrayFixed<number, 20>;
```

packages/filters/filter-displacement/src/DisplacementFilter.ts
```diff
@@ -1,7 +1,8 @@
-import { Filter } from '@pixi/core';
-import { Matrix, Point } from '@pixi/math';
-import vertex from './displacement.vert';
+import { CLEAR_MODES } from '@pixi/constants';
```
bigtimebuddy: Similar to the application conversion comment. It feels weird to not explicitly declaring constants as a new dependency. Unlike the canvas question in app this, however, will ultimately resolve when we generate types because constants is a dependency of core.

packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
Zyie: Ok added this!

packages/filters/filter-displacement/src/DisplacementFilter.ts
Zyie: Are we good to leave this and fix in a later Pr with the others?

packages/core/src/index.ts
```diff
@@ -45,3 +45,6 @@ export * from './geometry/Attribute';
 export * from './geometry/Buffer';
 export * from './geometry/Geometry';
 export * from './geometry/ViewableBuffer';
+
+// T stands for (content) type, L stands for the array fixed length
+export type ArrayFixed<T, L extends number> = [ T, ...Array<T> ] & { length: L };
```
bigtimebuddy: Can we move this to something like **packages/utils/src/types/**, seems like a more appropriate place for these generic types

packages/filters/filter-displacement/src/DisplacementFilter.ts
bigtimebuddy: Sure

packages/core/src/filters/FilterSystem.ts
```diff
@@ -389,7 +389,7 @@ export class FilterSystem extends System
      * @param {number} [resolution] override resolution of the renderTexture
      * @returns {PIXI.RenderTexture}
      */
-    getFilterTexture(input: RenderTexture, resolution: number): RenderTexture
+    getFilterTexture(input?: RenderTexture, resolution?: number): RenderTexture
```
ivanpopelyshev: ok, agree. The point is to force `input` as first param, but its not required. Some old filters pass `resolution` as first.

packages/core/src/index.ts
ivanpopelyshev: should be in utils, yes.

packages/filters/filter-blur/src/BlurFilter.ts
```diff
@@ -151,7 +156,7 @@ export class BlurFilter extends Filter
      * @member {number}
      * @default PIXI.BLEND_MODES.NORMAL
      */
-    get blendMode()
+    get blendMode(): number
```
ivanpopelyshev: type is `BLEND_MODES`

packages/filters/filter-displacement/src/DisplacementFilter.ts
```diff
@@ -63,15 +68,17 @@ export class DisplacementFilter extends Filter
...
+    ): void
     {
         // fill maskMatrix with _normalized sprite texture coords_
         this.uniforms.filterMatrix = filterManager.calculateSpriteMatrix(this.maskMatrix, this.maskSprite);
         this.uniforms.scale.x = this.scale.x;
         this.uniforms.scale.y = this.scale.y;
 
         // Extract rotation from world transform
-        const wt = this.maskSprite.transform.worldTransform;
+        const wt = this.maskSprite.worldTransform;
```
ivanpopelyshev: switching to a property? OK, I don't see a problem, filters are heavy anyway :) Oh, its because of my interface. Its fine, I agree with those changes.

packages/core/src/index.ts
Zyie: Ok it's been added there now

PR #6385, opened 2020-02-05
Convert Loaders to TS

eXponenta:
**Cross-package issues:**
- Spritesheet test was be changed because there is incompatibility types.
Will be restored after Spritesheet converting.

New documentation added for Loader#on* fields.


ivanpopelyshev:
OMG poor `IResourceMetadata`. I wouldn't be able to add spine fields there because there's no interface merging :(

eXponenta:
I don't know how right resolve multi extending , but extending from EE used only for backward compatibility.

bigtimebuddy:
What about this for extending Loader and EE https://stackoverflow.com/questions/26948400/typescript-how-to-extend-two-classes/26961710#26961710

ivanpopelyshev:
In stackoverflow both parents are important. In our case one of them (EE) is a deprecation.

eXponenta:
> What about this for extending Loader and EE https://stackoverflow.com/questions/26948400/typescript-how-to-extend-two-classes/26961710#26961710

Is not great idea, because there is a static fields.
And types will be broken. Real types, not from JSDoc

packages/loaders/src/LoaderResource.ts
```diff
@@ -13,14 +13,16 @@ import { Spritesheet } from '@pixi/spritesheet';
 export interface IResourceMetadata extends Resource.IMetadata {
     imageMetadata?: any;
 }
-export class LoaderResource extends Resource
+export interface ILoaderResource extends Resource
```
eXponenta: I changed `class ILoaderResource` into interface `ILoaderResource` bacause we don't construct class directly, and when we call `resource instanceof ResourceLoader` it is _false_, because **resource** instance of **Resource**
We not need realy class for typesafe,we can use interface.

bigtimebuddy:
Thanks @eXponenta

PR #6389, opened 2020-02-06
Convert spritesheet to TS

eXponenta:
Awaits #6385 because `LoaderResource` type is required


packages/spritesheet/src/SpritesheetLoader.ts
```diff
@@ -38,14 +40,14 @@ export class SpritesheetLoader
 
         const loadOptions = {
             crossOrigin: resource.crossOrigin,
-            metadata: resource.metadata.imageMetadata,
+            metadata: (resource.metadata as any).imageMetadata,
```
ivanpopelyshev: oh no. We cant override metadata anymore! :(

eXponenta: All okay while we use v3 of resource loader
https://github.com/pixijs/pixi.js/blob/dev-spritesheet-ts-r/packages/loaders/package.json#L29

eXponenta: Then i exposed fields on ResourceLoader class

https://github.com/pixijs/pixi.js/pull/6385/commits/614337808e4329c12e49524cdb770bc4223539a1#diff-aa234d8720b6d5fbb3e622efebb73118R13-R15

packages/spritesheet/src/Spritesheet.ts
```diff
@@ -1,7 +1,9 @@
 import { Rectangle } from '@pixi/math';
-import { Texture } from '@pixi/core';
+import { Texture, BaseTexture, resources } from '@pixi/core';
 import { getResolutionOfUrl } from '@pixi/utils';
 
+type Dict<T> = {[key: string]: T};
```
bigtimebuddy: Could you add this to utils in the src/types folder alongside ArrayFixed?

ivanpopelyshev:
Are we adding Dict now? We survived moving core without it

bigtimebuddy:
It’s a helpful utility, it doesn’t hurt anything to add it.

PR #6390, opened 2020-02-06
Convert Text to TypeScript

qtiki:
##### Description of change

Convert Text to TypeScript

##### Pre-Merge Checklist


packages/text/src/TextStyle.ts
```diff
@@ -773,7 +847,7 @@ function getColor(color)
  * @param {Array} array2 Second array to compare
  * @return {boolean} Do the arrays contain the same values in the same order
  */
-function areArraysEqual(array1, array2)
+function areArraysEqual(array1: any[], array2: any[]): boolean
```
eXponenta: I think it should be generic function

packages/text/src/TextStyle.ts
```diff
@@ -57,6 +89,38 @@ const genericFontFamilies = [
  */
 export class TextStyle
```
eXponenta: I think it should implement ITextStyle

packages/text/src/TextMetrics.ts
```diff
@@ -539,7 +572,7 @@ export class TextMetrics
      * @param  {boolean}  breakWords  The style attr break words
      * @return {boolean} whether to break word or not
      */
-    static canBreakWords(token, breakWords)
+    static canBreakWords(token: string, breakWords: TextStyle['breakWords']): boolean
```
eXponenta: I think types should be explicit instead of lookups.

packages/text/src/TextStyle.ts
```diff
@@ -57,6 +89,38 @@ const genericFontFamilies = [
...
 {
+    public styleID: number;
+    
+    protected _align: ITextStyle['align'];
```
eXponenta: Don't use lookup types.

packages/text/src/Text.ts
```diff
@@ -383,7 +401,7 @@ export class Text extends Sprite
      * @param {PIXI.Rectangle} rect - The output rectangle.
      * @return {PIXI.Rectangle} The bounds.
      */
-    getLocalBounds(rect)
+    getLocalBounds(rect: Rectangle): Rectangle
```
eXponenta: 'public' modifier is required!

empty modifier only for 'internal' (not public API, but used outside of type, such as inside Renderer )

eXponenta:
Cool start!
But:
- empty modifier used as 'public, for internal usage', other fields/methods/props require modifiers.
- I think we shouldn't use lookup types (aka ITextStyle ['font']) 
- TextStyle should implement ITextStyle, or convince me why this isn't necessary.

packages/text/src/Text.ts
```diff
@@ -41,12 +44,24 @@ const defaultDestroyOptions = {
...
+    protected _autoResolution: boolean;
+    protected _styleListener: () => void;
+
     /**
      * @param {string} text - The string that you would like the text to display
      * @param {object|PIXI.TextStyle} [style] - The style parameters
      * @param {HTMLCanvasElement} [canvas] - The canvas element for drawing text
      */
-    constructor(text, style, canvas)
+    constructor(text: string, style: TextStyle, canvas: HTMLCanvasElement)
```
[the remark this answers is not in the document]
qtiki: Yes it should say `Partial<ITextStyle>|TextStyle` like elsewhere. I missed that.

packages/text/src/TextStyle.ts
```diff
@@ -4,7 +4,39 @@
 import { TEXT_GRADIENT } from './const';
 import { hex2string } from '@pixi/utils';
 
-const defaultStyle = {
+export interface ITextStyle {
```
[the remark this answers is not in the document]
qtiki: Yes. I made the fields mandatory to force all of the values to be provided in the default values. I've been using `Partial<ITextStyle>` in places where they're optional. If using Partial is not something you would like to have there I can definitely change that.

packages/text/src/TextStyle.ts
```diff
@@ -796,14 +870,16 @@ function areArraysEqual(array1, array2)
...
+
 /**
  * Utility function to ensure that object properties are copied by value, and not by reference
  * @private
  * @param {Object} target Target object to copy properties into
  * @param {Object} source Source object for the properties to copy
  * @param {string} propertyObj Object containing properties names we want to loop over
  */
-function deepCopyProperties(target, source, propertyObj) {
+function deepCopyProperties(target: DeepCopyObject, source: DeepCopyObject, propertyObj: DeepCopyObject): void {
```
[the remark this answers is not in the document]
qtiki: I'll fix that.

packages/text/src/TextStyle.ts
qtiki: I agree. I originally had that there but for some reason I removed it and forgot to put it back in.

packages/text/src/TextStyle.ts
qtiki: I can get rid of those. I agree that it's not really nice for classes to reference each other, but given that TextStyle implements ITextStyle I figured it would make sense to specify the types in only one place. But I'll change those to not use lookup, I'll just need to declare a bunch of extra types for the interface so I don't have to type for example `'normal'|'bold'|'bolder'|'lighter'|'100'|'200'|'300'|'400'|'500'|'600'|'700'|'800'|'900'` more than once.

packages/text/src/TextStyle.ts
qtiki: Changing this to `function areArraysEqual<T1, T2>(array1: T1[], array2: T2[]): boolean` results in a problem with the !== comparison: `This condition will always return 'true' since the types 'T1' and 'T2' have no overlap.`

I can change it to `function areArraysEqual<T>(array1: T[], array2: T[]): boolean` but then it cannot be called with different types of arrays. Then again maybe you don't need to because if they're different types then they definitely are not equal. I'll go with this one.

packages/text/src/TextStyle.ts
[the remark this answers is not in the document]
qtiki: Also having the fields optional wouldn't enforce `TextStyle implements ITextStyle` properly now that I added it.

packages/text/src/Text.ts
qtiki: Ah okay. I missed that there is a semantic difference for this, I'll check the visibility for all the methods.

packages/text/src/TextStyle.ts
bigtimebuddy: Create separate types for things like this maybe? TextStyleAlign, TextStyleWeight, etc

qtiki: Yep that's what I did. I didn't have the TextStyle prefix but I just added that to all of them - I think that makes it a bit clearer. I'll push in a bit.

packages/text/src/TextStyle.ts
eXponenta: Why need to compare arrays with different type?

packages/text/src/Text.ts
```diff
@@ -534,7 +548,7 @@ export class Text extends Sprite
      * @param {boolean} [options.texture=true] - Should it destroy the current texture of the sprite as well
      * @param {boolean} [options.baseTexture=true] - Should it destroy the base texture of the sprite as well
      */
-    destroy(options)
+    public destroy(options: IDestroyOptions): void
```
Zyie: this should be `options: IDestroyOptions|boolean`

qtiki: Good catch! Pushed a fix.

bigtimebuddy:
@Zyie you good on this?

Zyie:
> @Zyie you good on this?

seems good to me

PR #6396, opened 2020-02-10
Convert mesh-extras to TypeScript

Zyie:
##### Pre-Merge Checklist

Simple enough conversion. 

Fixes a few types in `Mesh` and made `data` optional in `Buffer.update`


packages/mesh/src/MeshGeometry.ts
```diff
@@ -26,11 +26,11 @@ export class MeshGeometry extends Geometry
...
     /**
-     * @param {Float32Array|number[]} vertices - Positional data on geometry.
-     * @param {Float32Array|number[]} uvs - Texture UVs.
-     * @param {Uint16Array|number[]} index - IndexBuffer
+     * @param {Float32Array|number[]} [vertices] - Positional data on geometry.
+     * @param {Float32Array|number[]} [uvs] - Texture UVs.
+     * @param {Uint16Array|number[]} [index] - IndexBuffer
      */
-    constructor(vertices: IArrayBuffer, uvs: IArrayBuffer, index: IArrayBuffer)
+    constructor(vertices?: IArrayBuffer, uvs?: IArrayBuffer, index?: IArrayBuffer)
```
ivanpopelyshev: It is weird without default params, we had ones in v4

https://github.com/pixijs/pixi.js/blob/v4.x/src/mesh/Mesh.js#L40

I can justify `vertices?` that because PlaneGeometry uses that constructor. Its fine to set up actual data later.

PR #6398, opened 2020-02-10
Convert sprite-tiling to TypeScript

Zyie:
##### Description of change

Ok so slight complication with this one and not sure the best way round it.

The `TilingSprite.from` method isnt compatible with `Sprite.from` as `Sprite.from` takes a texture and some options whereas `TilingSprite.from` takes a texture, width, height.

The solution i have committed here is:
`static from(source: TextureSource, options: ISize): TilingSprite`

If anyone has better ideas let me know!

##### Pre-Merge Checklist


packages/sprite-tiling/src/TilingSprite.ts
```diff
@@ -265,9 +263,9 @@ export class TilingSprite extends Sprite
      * @param {number} height - the height of the tiling sprite
      * @return {PIXI.TilingSprite} The newly created texture
      */
-    static from(source, width, height)
+    static from(source: TextureSource, options: ISize): TilingSprite
     {
-        return new TilingSprite(Texture.from(source), width, height);
+        return new TilingSprite(Texture.from(source), options.width, options.height);
```
bigtimebuddy: I think this is a better API, but you need to provide backward compatibility. For instance:
```js
// Deprecated
if (typeof options === 'number') {
  deprecation('5.3.0', 'TilingSprite.from use options instead of width and height args');
  options = { width: options, height: arguments[2] } as ISize;
}
```

packages/sprite-tiling/src/TilingSprite.ts
```diff
@@ -1,7 +1,9 @@
-import { TextureMatrix, Texture } from '@pixi/core';
-import { Point, Rectangle, Transform } from '@pixi/math';
-import { TextureCache } from '@pixi/utils';
+import { Renderer, Texture, TextureMatrix, TextureSource, IBaseTextureOptions } from '@pixi/core';
+import { IDestroyOptions } from '@pixi/display/src';
+import { IPoint, Point, Rectangle, Transform } from '@pixi/math';
 import { Sprite } from '@pixi/sprite';
+import { TextureCache } from '@pixi/utils';
+import { ObservablePoint, ISize } from 'pixi.js';
```
eXponenta: from `@pixi/math`

packages/sprite-tiling/src/TilingSprite.ts
```diff
@@ -1,7 +1,9 @@
-import { TextureMatrix, Texture } from '@pixi/core';
-import { Point, Rectangle, Transform } from '@pixi/math';
-import { TextureCache } from '@pixi/utils';
+import { Renderer, Texture, TextureMatrix, TextureSource, IBaseTextureOptions } from '@pixi/core';
+import { IDestroyOptions } from '@pixi/display/src';
```
eXponenta: from `@pixi/display`

packages/sprite-tiling/src/TilingSprite.ts
Zyie: good catch!

packages/sprite-tiling/src/TilingSprite.ts
Zyie: Ok i've added this. Do you guys think it would be a good idea to update the other `.from` methods in this class and deprecate them to keep it consistent? 

for example
`static fromFrame(frameId: string, width: number, height: number): TilingSprite`

packages/sprite-tiling/src/TilingSprite.ts
```diff
@@ -280,7 +283,7 @@ export class TilingSprite extends Sprite
      * @param {number} height - the height of the tiling sprite
      * @return {PIXI.TilingSprite} A new TilingSprite using a texture from the texture cache matching the frameId
      */
-    static fromFrame(frameId, width, height)
+    static fromFrame(frameId: string, width: number, height: number): TilingSprite
```
themoonrat: @bigtimebuddy  - fromFrame and fromImage ... we deprecated these for Sprites and Textures in favour of just 'from' for v5, but looks like this got missed. Is there a case for doing that here too? Or wait until post TS conversion?

bigtimebuddy: I would rather do that as a follow up PR, but you’re right that was missed.

PR #6400, opened 2020-02-11
[WIP] Decouple preprocessor from Program

ivanpopelyshev:
Work in progress.

Preprocessors (aka `ProgramTemplate`) are factories Programs.

PixiJS vanilla will have two of them - default injects `precision`, `name`, raw does not modify code.

Program constructor should be protected, and only factory should be able to call it.

BatchShaderGenerator should be preprocessors too.

Theoretically, if we port ThreeJS or any other 3d renderer to PixiJS core, it should have its own ShaderLib as a preprocessor.

### DIFF
--- a/packages/core/src/batch/BatchShaderGenerator.ts
+++ b/packages/core/src/batch/BatchShaderGenerator.ts
@@ -1,5 +1,5 @@
 import { Shader } from '../shader/Shader';
-import { Program } from '../shader/Program';
+import { Program, defaultProgramTemplate } from '../shader/Program';
 import { UniformGroup } from '../shader/UniformGroup';
 import { Matrix } from '@pixi/math';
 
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
I know this isn't done, but for naming I'm not a big fan of the "Pixi" prefix (e.g. "PixiProgramTemplate"). Just call it "ProgramTemplate" without the Pixi.

ivanpopelyshev:
> I know this isn't done, but for naming I'm not a big fan of the "Pixi" prefix (e.g. "PixiProgramTemplate"). Just call it "ProgramTemplate" without the Pixi.

Its default program template. Maybe `DefaultProgramTemplate`

GoodBoyDigital:
nice, collecting a few thoughts on this.. watch this space!

ivanpopelyshev:
One of results im expecting - 
1. remove double-compilation of shader (test context + renderer)
2. debug info about all shaders used in the game, allows to create shader loader code for game build. Pre-create all shaders before app started.

PR #6405, opened 2020-02-11
add cube texture again

Zyie:
##### Description of change

In #6340 `CubeTexture` was removed as it's `.from` function was not compatible with `BaseTexture`.

@GoodBoyDigital has a use for this class so I'm adding it back in. My solution doesn't solve the underlying issue but at least allows the project to build.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -13,6 +13,7 @@ export * from './Renderer';
 export * from './AbstractRenderer';
 export * from './framebuffer/Framebuffer';
 export * from './framebuffer/GLFramebuffer';
+export * from './textures/CubeTexture';
 export * from './textures/Texture';
 export * from './textures/BaseTexture';
 export * from './textures/GLTexture';
--- a/packages/core/src/textures/BaseTexture.ts
+++ b/packages/core/src/textures/BaseTexture.ts
@@ -602,7 +602,7 @@ export class BaseTexture extends EventEmitter
      * @param {boolean} [strict] Enforce strict-mode, see {@link PIXI.settings.STRICT_TEXTURE_CACHE}.
      * @returns {PIXI.BaseTexture} The new base texture.
      */
-    static from(source: ImageSource|string, options: IBaseTextureOptions,
+    static from(source: ImageSource|string, options?: IBaseTextureOptions,
         strict = settings.STRICT_TEXTURE_CACHE): BaseTexture
     {
         const isFrame = typeof source === 'string';
--- a/packages/core/src/textures/CubeTexture.ts
+++ b/packages/core/src/textures/CubeTexture.ts
@@ -19,7 +19,7 @@ export class CubeTexture extends BaseTexture
      *        on the options available to each resource.
      * @returns {PIXI.CubeTexture} new cube texture
      */
-    static from(resources: any, options: ICubeResourceOptions): CubeTexture
+    static from(resources: any, options?: ICubeResourceOptions): CubeTexture
     {
         return new CubeTexture(new CubeResource(resources, options));
     }
--- a/packages/core/src/textures/CubeTexture.ts
+++ b/packages/core/src/textures/CubeTexture.ts
@@ -1,0 +1,26 @@
+import { BaseTexture } from './BaseTexture';
+import { CubeResource, ICubeResourceOptions } from './resources/CubeResource';
+
+/**
+ * A Texture that depends on six other resources.
+ *
+ * @class
+ * @extends PIXI.BaseTexture
+ * @memberof PIXI
+ */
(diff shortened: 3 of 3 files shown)


GoodBoyDigital:
Yo! Yes had a catch up with @Zyie about this one. Appreciate there is an issue about overriding from with typescript. CubeTexture is pretty important as one of the base 'lego bricks of the midlevel Pixi api. 

Potentially we could make CubeTexture compose BaseTexture like Texture does?

bigtimebuddy:
I’m a little confused by this. CubeTexture is just BaseTexture but there’s nothing to inherit or extend. It’s just a static constructor. Will this CubeTexture have more things in the future? Alternatively, could we add a fromCube method (like fromBuffer) on BaseTexture?

ivanpopelyshev:
Its strange that `deprecation` code was lost.

Another thing is that we have a piece of code in `assemble.js` specially for `from`, beware: https://github.com/pixijs/pixi.js/blob/dev/types/assemble.js#L35

Maybe its time to remove it.

I'm with @bigtimebuddy on that one. I understand why we need `BaseRenderTexture` , there are a few extra methods. But here? Extra class?

GoodBoyDigital:
Yo, from a developer of pixi point of view I totally agree that a class that does nothing other than exist is a bit weird. 

Having said that, from the lens of someone working with a midlevel WebGL API i think its a really intuitive class to have as in pretty much all other APIs the class would exist iyswim?

```
const enviromentMap = CubeTexture.from([...]);

const myShader = Shader.from(vert, frag, {
    enviromentMap
})
```

I am open to the idea of it composing a baseTexture instead, we could then add some nice helper functions over it like `setPositiveX(tex)`, `setNegativeX(tex)` or something like that to justify its existence :D

ivanpopelyshev:
I cant believe im saying this: @GoodBoyDigital, put it into a plugin.

GoodBoyDigital:
haha, Pixi midlevel api is not a plugin :P

The parts that allow people to do pretty much everything with this layer in WebGL(1) include:

Shader
Geometry
FrameBuffer
State
Texture
CubeTexture

With the above basic blocks we reflect pretty much all WebGL fundamentals and you can make any kind of WebGL!

Appreciate that CubeTexture does not have an immediate practical use in our high level 2d api, but I do think it has a home in the midlevel api.

GoodBoyDigital:
or instead of a class how about we make from a bit smarter ? :D

```const cubeTextue = Texture.from(['1'.png, '2.png', '3.png', '4.png', '5.png', '6.png'])```

**Comment by bigtimebuddy** [comment]
Personally, I'd like to see a more substantial PR that utilizes CubeTexture as a feature more broadly. There's no display objects to associate with CubeTexture, no examples. It's not obvious how a developer might utilize it. The missing piece for me is **more information and documentation** about what it is and how to use it, `A Texture that depends on six other resources.` feels insufficient.

Part of v5 refactor of BaseTexture was to have BaseTexture be agnostic to media/resource types. That's why we don't have a VideoTexture. CubeResource already seems like the mid-level API feature. So if `CubeTexture.from`, why not `VideoTexture.from`? Is this only about the convenience and readability for creating a new CubeResource? That doesn't seem mid-level, seems more high-level.

In general, I'm bullish about wanting to reduce the API surface of PixiJS. It has historically been difficult to maintain a large public API. I don't want to add something like this unless it's a legit feature or it's able to serve a broader developer audience. The tiny number of developers who want Cube-like resources, can already do this today:

```js
const texture = new PIXI.BaseTexture(new PIXI.resources.CubeResource([...]))
```

bigtimebuddy:
@GoodBoyDigital I'm happy to help make a PR to add CubeResource to `autoDetectResource` which would accomplish `Texture.from` thing.

ivanpopelyshev:
I explained @GoodBoyDigital three options I see:

1. add CubeTexture back
2. add `test()`  static method to resource
3. move all things to cubemap package and then allow him to do whatever he wants there

@GoodBoyDigital chose 2 for now.

I think 3. will be better option, we can make PR later. If our only user of CubeTexture wants same abstraction with handy methods like RenderTexture/BaseRenderTexture, I don't see why not make it - our new infrastructure makes it easy, and core size will be even smaller after that.

GoodBoyDigital:
nice, Had a good catch up with Ivan and your points are true too Matt!

Agree that option 3 given time could be pretty cool and make better use of CubeTexture and explaining how it could be useful.

adding it to `autoDetectResource` i really like!

GoodBoyDigital:
Nice one for updating PR @ivanpopelyshev 👍

GoodBoyDigital:
happy to close in favour of: #6409 👍

PR #6409, opened 2020-02-12
Support auto-detection of CubeResource types

ivanpopelyshev:
Alternative for #6405.

@GoodBoyDigital please confirm that it works for your project.


bigtimebuddy:
Could you add a unit test for this? Just make sure we get a CubeResource back?

bigtimebuddy:
we could use ArrayFixed in utils for an array of a defined length for CubeResourse content

packages/core/test/CubeResource.js
```diff
@@ -11,4 +17,11 @@ describe('PIXI.resources.CubeResource', function ()
             new CubeResource(8);
         }).to.throw(Error, /invalid length/i);
     });
+    it('should be created through BaseTexture.from()', function ()
+    {
+        const path1 = this.baseTexUrl;
+        const baseTex = BaseTexture.from([path1, path1, path1, path1, path1, path1]);
+
+        expect(baseTex.resource instanceof CubeResource).to.be.true;
```
cursedcoder: what about `expect(baseTex.resource).to.be.instanceof(CubeResource)`?

bigtimebuddy: Yes, better!

ivanpopelyshev:
Yep, I approve the fixed size in type.

PR #6415, opened 2020-02-16
Application and Renderer resize improvements

### LINKED ISSUE #6081: Feature request: resizeTo redraw
### Expected Behavior
If a PIXI `Application` has a `resizeTo` element, it would be nice if the element positions inside the application would also update when resizing. Plus some throttling on the resize event would help avoid performance problems.
If this functionality is too complicated to implement (or already present), a best practice example of a responsive application would be very helpful.

### Current Behavior
Only the canvas is resized.

### Environment

- **`pixi.js` version**: 5.1.3

bigtimebuddy:
I've been meaning to tackle this one for awhile, since no one seemed keen on adding the PR for it.

Fixes #6081

### Added

* Introduces `resize` event to the Renderer. Called immediate after view is resized can be used for responsive layouts, or other resizing events.
* Adds throttle resizing on the Application when using `resizeTo`, this will prevent unnecessary extra event calls triggered by the window 'resize'.

### Example

https://jsfiddle.net/bigtimebuddy/u05c1s28/

### DIFF
--- a/packages/app/src/ResizePlugin.ts
+++ b/packages/app/src/ResizePlugin.ts
@@ -9,8 +9,12 @@ import { IApplicationOptions } from './Application';
  */
 export class ResizePlugin
 {
+    public static _resizeId: number;
     public static _resizeTo: Window|HTMLElement;
+    public static resizeTo: Window|HTMLElement;
     public static resize: () => void;
+    public static queueResize: () => void;
+    public static cancelResize: () => void;
     public static renderer: Renderer|CanvasRenderer;
 
     /**
@@ -21,15 +25,22 @@ export class ResizePlugin
      */
     static init(options?: IApplicationOptions): void
     {
+        /**
+         * The HTML element or window to automatically resize the 
+         * renderer's view element to match width and height.
+         * @type {Window|HTMLElement}
+         * @name resizeTo
+         * @memberof PIXI.Application#
+         */
         Object.defineProperty(this, 'resizeTo',
             {
-                set(dom)
+                set(dom: Window|HTMLElement)
                 {
-                    window.removeEventListener('resize', this.resize);
+                    window.removeEventListener('resize', this.queueResize);
                     this._resizeTo = dom;
                     if (dom)
                     {
-                        window.addEventListener('resize', this.resize);
+                        window.addEventListener('resize', this.queueResize);
                         this.resize();
                     }
                 },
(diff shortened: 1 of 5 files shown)


themoonrat:
Had a play with a slight modified version of your fiddle: https://jsfiddle.net/w63ahoqd/

Can confirm that no matter how crazy I went on the window resizing, it occurs just ones per render

image

packages/app/src/ResizePlugin.ts
```diff
@@ -9,8 +9,12 @@ import { IApplicationOptions } from './Application';
  */
 export class ResizePlugin
 {
+    public static _resizeId: number;
```
Zyie: should `_resizeId`, `_resizeTo`, `queueResize`, and `cancelResize` be private?

`_resizeId` has `_` so it seems private
`_resizeTo` has a get/set
`queueResize`, and `cancelResize` have the `@private` tag in the docs

bigtimebuddy: Good question. I was following the lead on the access modifiers, but I guess these can be private. I think the typings for the app plugins need a little more thought.

bigtimebuddy:
Added a feature based on feedback from @GoodBoyDigital to set the amount of time to throttle instead of just using RAF. I agree this is more useful for producing more stable results, especially if you're doing something expensive on resize.

bigtimebuddy:
Uncertain if we should set a default for `resizeThrottle`, since `100` seems more reasonable than `0`, but wonder what other folks think?

ivanpopelyshev:
Usually, people use `setInterval`. Of course your solution is better, but it involves many lines of code.

themoonrat:
The issue I have with this latest change is that if the screen keeps resizing, then nothing is ever updated.
https://jsfiddle.net/opxhkytg/1/

This has a 1 second resizeThrottle setting. Keep slowly resizing the window, and you'll notice that the resize event _never_ kicks in, since a new window resize keeps resetting the timeout.

This setting should be a throttle rather than a debounce, imo.

eXponenta:
But there not case when animated resize is required, if so, then this architecture apriory wrong.

bigtimebuddy:
I think you're right @themoonrat. If I do this as only a throttle like I had before, users can still implement their own debounce with the `resize` event. That's probably the most minimal implementation.

packages/app/src/ResizePlugin.ts
```diff
@@ -9,9 +9,13 @@ import { IApplicationOptions } from './Application';
  */
 export class ResizePlugin
```
cursedcoder: This plugin is hard to understand with all its static/this context mixed.

cursedcoder:
Not sure if it's the right thing for the plugin to be singleton-ish.
Was there a reason?

packages/app/src/ResizePlugin.ts
bigtimebuddy: yeah these are static methods where the context is the Application instance. Loader plugins are similar.

bigtimebuddy:
It’s not really a singleton. No one really calls this class directly. It comes preinStalled, it’s mostly a code organization. Could’ve been done a bunch of ways, but this is how it evolved

PR #6418, opened 2020-02-17
Travis

eXponenta:
Update travis config for preventing downing by xvfb installation fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 services:
-    -xvfb
+    - xvfb
 
 node_js:
     - "10"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,32 +1,24 @@
...
-      - g++-4.8
+env:
+  - CXX=g++-4.8
 
 install:
-    - npm install xvfb-maybe
     - npm install
     - npm install codecov -g
 
-cache:
```
bigtimebuddy: Could you not remove cache directories? This is useful for doing faster runs on dev, master, and long running PRs

bigtimebuddy:
Thank you for this @eXponenta, just one comment

.travis.yml
bigtimebuddy: https://docs.travis-ci.com/user/caching/#npm-cache

maybe this should be
```yml
cache: npm
```

bigtimebuddy: Oops I missed this:

```
Please note that as of July 2019, npm is cached by default on Travis CI
```

PR #6423, opened 2020-02-19
Fix: Floating point imprecision could break vertical text gradient generation

TheWyo:
##### Description of change
In Chrome, I'd noticed some of my text in a project didn't always have the same gradient on each line. Turns out the seams between tiled gradients could sometimes have floating point imprecision, effectively swapping the start/end gradient colours around and messing a line or two up. This PR caps the gradient to the last value (as color stops are generated in order anyway) to prevent this happening.

[Playground demo (Open in chrome, change bool at start to toggle fix)](https://pixiplayground.com/#/edit/8I-EO0eRlShU4YHvuD6bc)

Happy to backport this for v4 as well if wanted.

##### Pre-Merge Checklist

- Demo provided.
- Do we have a way of testing visual stuff?
- Can't verify, tests currently failing already.

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -490,6 +490,13 @@ export class Text extends Sprite
             // ['#FF0000', '#00FF00', '#0000FF'] over 2 lines would create stops at 0.125, 0.25, 0.375, 0.625, 0.75, 0.875
             totalIterations = (fill.length + 1) * lines.length;
             currentIteration = 0;
+
+            // There's potential for floating point precision issues at the seams between gradient repeats.
+            // The loop below generates the stops in order, so track the last generated one to prevent
+            // floating point precision from making us go the teeniest bit backwards, resulting in
+            // the first and last colors getting swapped.
+            let lastIterationStop = 0;
+
             for (let i = 0; i < lines.length; i++)
             {
                 currentIteration += 1;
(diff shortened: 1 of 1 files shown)


themoonrat:
Hello @TheWyo ! Thanks for the PR and for the playground
When I try to use your playground, in the console I see

"Uncaught TypeError: Failed to execute 'createLinearGradient' on 'CanvasRenderingContext2D': The provided double value is non-finite."

Could you fix this please? Code looks good, would just like to see for myself :)

packages/text/src/Text.ts
```diff
@@ -503,8 +510,13 @@ export class Text extends Sprite
                     {
                         stop = currentIteration / totalIterations;
                     }
-                    gradient.addColorStop(stop, fill[j]);
+
+                    // Prevent color stop generation going backwards from floating point imprecision
+                    const clampedStop = Math.max(lastIterationStop, stop);
```
themoonrat: do we also want to make sure that we don't have a point that's greater than 1? Or do you feel that's not possible to achieve even with the rounding issues?

TheWyo:
> Hello @TheWyo ! Thanks for the PR and for the playground
> When I try to use your playground, in the console I see
> 
> "Uncaught TypeError: Failed to execute 'createLinearGradient' on 'CanvasRenderingContext2D': The provided double value is non-finite."
> 
> Could you fix this please? Code looks good, would just like to see for myself :)

My bad, I was curious if the same code worked in v4, didn't realise saving that setting saved it to the same url as opposed to making a copy. Should work for you now.

packages/text/src/Text.ts
TheWyo: Trying to add a stop >1 does throw a DOMException from addColorStop apparently, so might be a good idea for safety's sake.

PR #6427, opened 2020-02-20
Text destroy fix

soylomass:
##### Description of change

set canvas width and height to 0 to make sure it's GC'd by some versions of Safari

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -55,6 +55,7 @@ export class Text extends Sprite
     protected _resolution: number;
     protected _autoResolution: boolean;
     protected _styleListener: () => void;
+    private _ownCanvas: boolean;
 
     /**
      * @param {string} text - The string that you would like the text to display
@@ -63,7 +64,13 @@ export class Text extends Sprite
      */
     constructor(text: string, style: Partial<ITextStyle>|TextStyle, canvas: HTMLCanvasElement)
     {
-        canvas = canvas || document.createElement('canvas');
+        let ownCanvas = false;
+
+        if (!canvas)
+        {
+            canvas = document.createElement('canvas');
+            ownCanvas = true;
+        }
 
         canvas.width = 3;
         canvas.height = 3;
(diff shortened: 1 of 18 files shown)


soylomass:
I only reproduced it on iOS Safari, but people here https://github.com/openlayers/openlayers/issues/9291 say it exists on Mac too.

themoonrat:
Do we need to do a further sweep of 
`document.createElement('canvas')`
to find other places we just set to null and may need to set width and height to 0 too?

bigtimebuddy:
Good point this may not be the only place where this happens, although definitely the most used/common.

bigtimebuddy:
Here are some other places this fix could also be applied pretty easily:

* CanvasRenderTarget (utils)
* SVGResource (core)
* getTestContext (utils)
* AbstractRenderer (core)
* CanvasPrepare (canvas-prepare)
* canUseNewCanvasBlendModes (canvas-renderer)
* isWebGLSupported (utils)

In the case of this PR (Text) and AbstractRenderer, these both can pass in the canvas to the constructor, in these cases, we shouldn't be setting width and height to `0`, just removing the references. Imagine a user is reusing the canvas, we wouldn't want them to stub their toe on this.

bigtimebuddy:
@soylomass do you want to keep working on this PR or do you want to pass it off?

packages/text/src/Text.ts
```diff
@@ -559,6 +559,9 @@ export class Text extends Sprite
 
         super.destroy(options);
 
+        // set canvas width and height to 0 to make sure it's GC'd by some versions of Safari
```
cursedcoder: Do you have any source for that optimization?
If so please add the link with tsdoc or in comment.

cursedcoder:
Would it make sense to add tsdoc tags for such optimizations in the future?
/cc @bigtimebuddy @themoonrat

soylomass:
> @soylomass do you want to keep working on this PR or do you want to pass it off?

Text was the main problem in my case and this fixes it, but you are right, it should take into account that the canvas might be reutilized.
About your question, I prefer to pass it off as I'm quite busy at the moment.

bigtimebuddy:
No problem. I’m happy to add to it. Thanks for pointing this out, just leave the pr open and I’ll push to it.

bigtimebuddy:
The only problem I have with that approach @themoonrat is that we would be adding an API (permanent) to workaround a browser bug (temporary). I think we need a broader solution than this PR for disposing of internally created canvases across Pixi, but there are a number of ways to do that, ideally not adding a new API and doing it internal-only would be best.

Another approach I was considering was introducing a new DOM package that would consolidate all the browser DOM APIs in once place. There are so many instances of createElement, also not having this DOM package has hurt our ability to do rendering in Node and Workers. This is a much bigger task obviously, but that abstraction might be able to handle disposing internally, in one place.

bigtimebuddy:
I’m okay with this fix for now. Text is one of the worst spots for this issue. But we will likely need follow up PRs to address more broadly.

bigtimebuddy:
Thanks for the PR @soylomass

Issue #6424, opened 2020-02-20
Resizing renderer appears to break extract


JoeHegarty:
### Expected Behavior
Resizing the renderer should not break extract canvas.

### Current Behavior
After resizing the renderer, extracting the canvas causes a vertically flipped mirror image to also appear.
As seen at: https://i.imgur.com/ZfwKIAN.png

### Possible Solution
* I have tried lots of different configurations but resize seems to always cause this.

### Steps to Reproduce
* Optionally enable autoresizing on `Application`
* Enable `preserveDrawBuffer`
* Render at least one
* Resize the window or call `Renderer.resize` if not automatic.
* Run `Renderer.extract.canvas(undefined)`
* Observe the result (perhaps with `toDataURL`)

### Environment

- **`pixi.js` version**:  5.2.1
- **Browser & Version**: Chrome 79
- **OS & Version**: macOS 10.15.3
- **Demo**: https://www.pixiplayground.com/#/edit/jogREV4Vj-gqrZTLI_Hoh

eXponenta:
Can you create minimal demo with reproducing a problem?

JoeHegarty:
https://www.pixiplayground.com/#/edit/jogREV4Vj-gqrZTLI_Hoh

Example here, check the console log to get the data url.

ivanpopelyshev:
Oh, right, it was a part of my poor big PR that fixed many things: https://github.com/pixijs/pixi.js/pull/4632/files#diff-e38c1de4b0f48ed1293bccc38b07e6c1R143

We didnt merge it back then.

@eXponenta care to make a small fix based on it?

JoeHegarty:
Thanks for merging the fix!

PR #6430, opened 2020-02-21
Hot extract fix

eXponenta:
Related to #6424 

Why? Because extract drawn canvas to itself, it is wrong when there is a alpha.

In fix extract creates new CanvasRenderTarget and draw flipped to it.

Before:
https://www.pixiplayground.com/#/edit/pAo7HerhsY5JZTf16ajNq

After:
https://www.pixiplayground.com/#/edit/0vBcRtK80ctpV8pI2bVVn


bigtimebuddy:
Thank you @eXponenta!

PR #6435, opened 2020-02-26
Cubemap renderTexture support

ivanpopelyshev:
1. refactoring - move everything out ArrayResource to parent abstract class
2. handle baseTexture binds differently in ArrayResource/CubeResource
3. check for parentTextureArray when we bind a texture or framebuffer to texture.

Should help @GoodBoyDigital for his cases.

UPD. It works: https://www.pixiplayground.com/#/edit/4RPGjVMWfcWGyxHusdcKE


GoodBoyDigital:
oh! excite! will test...

bigtimebuddy:
I can live with this, but I'm not a huge fan of introducing another abstraction layer. I already feel classes like BaseImageResource overly complicate an otherwise simple resource system and this adds yet another layer. Is there an easier way to achieve the same thing on the resources-side? Either with some static util functions, or having CubeResource override ArrayResource methods?

GoodBoyDigital:
Hey this is cool! Thanks @ivanpopelyshev. We are going to roll this is as it adds a new useful feature 👍 

We discussed and do feel that the resource manager has grown in complexity over time and would definitely benefit from a bit of a refactor, potentially using composition rather than inheritance! 

watch this space!

PR #6443, opened 2020-02-28
Canvas2d transform improvement

ivanpopelyshev:
Our Canvas2d and webgl code regarding render transform differs.

It makes difficult to debug some cases, and it prevents us from refactor all that "temporary parent" code to "pixi/@display" package.

This PR does not add new functionality nor does it change behaviour.  No moving `floor`s around, no doing extra multiplications/divisions, it should be the same code.

I'll add some resolution-related examples a bit later.

+81-158 , yeah, that's a good balance!

### DIFF
--- a/packages/canvas/canvas-renderer/src/CanvasRenderer.js
+++ b/packages/canvas/canvas-renderer/src/CanvasRenderer.js
@@ -141,7 +141,7 @@ export class CanvasRenderer extends AbstractRenderer
      * @param {PIXI.RenderTexture} [renderTexture] - A render texture to be rendered to.
      *  If unset, it will render to the root context.
      * @param {boolean} [clear=false] - Whether to clear the canvas before drawing
-     * @param {PIXI.Matrix} [transform] - A transformation to be applied, supports only translation
+     * @param {PIXI.Matrix} [transform] - A transformation to be applied
      * @param {boolean} [skipUpdateTransform=false] - Whether to skip the update transform
      */
     render(displayObject, renderTexture, clear, transform, skipUpdateTransform)
--- a/packages/canvas/canvas-renderer/src/CanvasRenderer.js
+++ b/packages/canvas/canvas-renderer/src/CanvasRenderer.js
@@ -4,6 +4,9 @@ import { CanvasMaskManager } from './utils/CanvasMaskManager';
 import { mapCanvasBlendModesToPixi } from './utils/mapCanvasBlendModesToPixi';
 import { RENDERER_TYPE, SCALE_MODES, BLEND_MODES } from '@pixi/constants';
 import { settings } from '@pixi/settings';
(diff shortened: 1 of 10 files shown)


ivanpopelyshev:
https://pixijs.io/examples/?v=dev#/demos-basic/cacheAsBitmap.js with `{forceCanvas:true}` doesnt work at the moment. im debugging it.

ivanpopelyshev:
OK, it works now.

packages/canvas/canvas-renderer/src/CanvasRenderer.js
```diff
@@ -247,10 +236,58 @@ export class CanvasRenderer extends AbstractRenderer
...
+        this._projTransform = null;
 
         this.emit('postrender');
     }
 
+    /**
+     * sets matrix of context
+     * called only from render() methods
+     * takes care about resolution
+     * @param {PIXI.Matrix} transform world matrix of current element
```
bigtimebuddy: Could you make this private?

ivanpopelyshev: setBlendMode isn't private

bigtimebuddy: It’s only used internally. No need to be public. Would like to limit the public API

ivanpopelyshev: It shouldn't be private, as our WebGL state-related functions aren't private too.

ivanpopelyshev: We expect from our users to write their own components, right? That override their own `_render` and `_renderCanvas` methods. We expose mid-level API. If we would mark mid-level API differently than high-level, I would be happy to go through all core and do it, but we dont have necessary tools. Same with low-level (internal, only for plugins).

ivanpopelyshev: Of course there's an opinion that canvas2d is legacy and we really shouldn't encourage people to use it, but I dont think like that. Canvas2d will always be here and its needed for better antialiasing, we cant drop the support.

bigtimebuddy: That’s fine, this isn’t a blocker for me. However, from my perspective, making something public means we promise to users: we won’t change this, and if we do we won’t break it. Every new public API is a small maintenance tax. Having a very large public API is a big tax. We should be a little more strict about what is public unless there is a very clear reason for public, not just someone MAY use this someday in a plugin.

ivanpopelyshev: Hm, from that perspective, this thing actually looks good. We can make it part of public API, its formed from all our canvas experience.

PR #6450, opened 2020-03-01
Make core independent from display

ivanpopelyshev:
Continuation of #6443

`@pixi/core` can be independent from `@pixi/display` !

`_lastObjectRendered` can be null after that PR, so I patched `InteractionManager` too.

For now I decided that its ok to have temp parent separate for each possible detached element.

I also couldn't built pixi because of strange `DisplacementFilter` import, I dont know if you guys want to resolve it in one of other PR's


ivanpopelyshev:
Many times people were trying to change that thing. This PR is a delcaration that temp display object stays, but at least we can refactor it so it wont haunt us.

ivanpopelyshev:
I don't know which non-trivial things to add there at the moment. I prefer not to make trivial tests.

bigtimebuddy:
You're reusing this pattern in three places, and I think we can simplify reuse:

```js
const cacheParent = displayObject.enableTempParent();
displayObject.updateTransform();
displayObject.disableTempParent(cacheParent);
```

Maybe make a variant of `updateTransform` (`safeUpdateTransform`? parameter flag?) and remove `enabledTempParent` and `disableTempParent` methods.

ivanpopelyshev:
OK, added a test.

We declined adding a parameter before because of extra if making updateTransform ugly.

no, we cant do that because there might be other operations between updateTransform and disableTempParent. Considering all the cases where i saw temp parent is needed, it will be better if people see those lines. We cant to hide them.

packages/core/package.json
```diff
@@ -29,7 +29,6 @@
   },
   "dependencies": {
     "@pixi/constants": "5.2.1",
-    "@pixi/display": "5.2.1",
```
cursedcoder: :+1:

cursedcoder:
@ivanpopelyshev did you forget to push? Because I can't see any tests.

ivanpopelyshev:
@cursedcoder https://github.com/pixijs/pixi.js/pull/6450/files#diff-b4bbad81d74f4eee9ed95ec693f53c9aR112

ivanpopelyshev:
We have to wait for localBounds fix, then I'll go over this PR again.

PR #6455, opened 2020-03-03
Remove forceFXAA Application and Renderer option

### LINKED ISSUE #6444: Sprite glitch with antialiasing
In some situations a Sprite has a visible corner-to-corner glitch with antialiasing enabled. I was able to reproduce this in the playground by first drawing a full-screen sprite and then another sprite on top of it: https://www.pixiplayground.com/#/edit/Kj8yLiP9bFoPIeMvNzyYe

### Expected Behavior

Drawing a Sprite on top of a previous Sprite should fully cover the old Sprite without any bleeding artifacts.

### Current Behavior

I'm not sure whether the bug is visible with all display resolutions, but it is clearly visible on my QHD display. Here's a 10x zoom of the image where the red Sprite drawn before the white Sprite is bleeding through the polygon edges in the middle: image

### Possible Solution

I'm not sure what is causing this. It could have something to do with all of the Sprites being drawn with the same z-value but I haven't really investigated this any further than the repro-case in the playground.

### Steps to Reproduce

See the linked playground example.

### Environment

- **`pixi.js` version**:  5.2.1
- **Browser & Version**: Chrome Version 80.0.3987.122 (Official Build) (64-bit)
- **OS & Version**:  Windows 10 Pro
- **Running Example**: https://www.pixiplayground.com/#/edit/Kj8yLiP9bFoPIeMvNzyYe
- **Display Card**: Nvidia GTX 1650 (driver 442.19)

bigtimebuddy:
Partially fixes #6444 

We have long discarded the use of `forceFXAA` functionality (last implemented in v3), this removes from docs and typings.

### DIFF
--- a/packages/core/src/Renderer.ts
+++ b/packages/core/src/Renderer.ts
@@ -98,8 +98,6 @@ export class Renderer extends AbstractRenderer
      *   resolutions other than 1.
      * @param {boolean} [options.antialias=false] - Sets antialias. If not available natively then FXAA
      *  antialiasing is used.
-     * @param {boolean} [options.forceFXAA=false] - Forces FXAA antialiasing to be used over native.
-     *  FXAA is faster, but may not always look as great.
      * @param {number} [options.resolution=1] - The resolution / device pixel ratio of the renderer.
      *  The resolution of the renderer retina would be 2.
      * @param {boolean} [options.clearBeforeRender=true] - This sets if the renderer will clear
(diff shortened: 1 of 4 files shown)


packages/core/src/Renderer.ts
```diff
@@ -98,8 +98,6 @@ export class Renderer extends AbstractRenderer
      *   resolutions other than 1.
      * @param {boolean} [options.antialias=false] - Sets antialias. If not available natively then FXAA
```
qtiki: Missed a spot here: _"If not available natively then FXAA antialiasing is used"_.

bigtimebuddy: Yup! Good catch

PR #6462, opened 2020-03-06
Fix: Text gradients don't account for texture padding

TheWyo:
##### Description of change
Texts with gradients don't take the style's padding property into account, and assumed the gradient should be applied to the whole canvas size. On multiline vertical gradients, this caused each successive line to be more offset on the gradient.

For example, the following style with extreme padding to highlight the issue (screenshots from the TextStyle generator):
```json
{
    "fill": [
        "#fff0a1",
        "#face08",
        "#db8c00"
    ],
    "fontWeight": "bold",
    "padding": 20
}
```

Current behaviour (can see the gradient seams for the multiple lines):
GradientPadding-Broken

Fixed behaviour:
GradientPadding-Fixed

Playground demo: https://pixiplayground.com/#/edit/Ko5N0jUy1hAgUalGWisJn

##### Pre-Merge Checklist

  - Of those not already failing

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -476,8 +476,12 @@ export class Text extends Sprite
         // generated with the incorrect dimensions
         const dropShadowCorrection = (style.dropShadow) ? style.dropShadowDistance : 0;
 
-        const width = Math.ceil(this.canvas.width / this._resolution) - dropShadowCorrection;
-        const height = Math.ceil(this.canvas.height / this._resolution) - dropShadowCorrection;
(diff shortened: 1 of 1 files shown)


themoonrat:
Ah you're a legend on this stuff!
I'm happy with this.
There is one last scenario that I find also throws off gradients... when the lineHeight is set. You think this is something you could take a look at in a future PR please?

TheWyo:
Sure, should have some time next week to take a look.
Line height will probably be trickier, might need to actually rewrite the loop for that, because it'll probably break the even-sized-segments thing it has going on currently.

PR #6464, opened 2020-03-09
Public animated sprite update

themoonrat:
New autoUpdate setter getter for AnimatedSprite, that matches behaviour as in the VideoResource autoUpdate setter getter

### DIFF
--- a/packages/sprite-animated/src/AnimatedSprite.ts
+++ b/packages/sprite-animated/src/AnimatedSprite.ts
@@ -49,7 +49,7 @@ export class AnimatedSprite extends Sprite
     private _textures: Texture[];
     private _durations: number[];
     private _autoUpdate: boolean;
-    private _isAutoUpdating: boolean;
+    private _isConnectedToTicker: boolean;
     private _currentTime: number;
 
     /**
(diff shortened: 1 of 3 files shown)


packages/sprite-animated/src/AnimatedSprite.ts
```diff
@@ -76,11 +77,13 @@ export class AnimatedSprite extends Sprite
 
         /**
          * `true` uses PIXI.Ticker.shared to auto update animation time.
+         *
          * @type {boolean}
          * @default true
          * @private
          */
         this._autoUpdate = autoUpdate;
+        this._isAutoUpdating = false;
```
bigtimebuddy: I was a little confused by this at first: why is another boolean needed for tracking auto-updating. But then reading the code made it more obvious. Maybe a JSDoc for this variable would be helpful:
```suggestion
        /**
         * An internal boolean to track if auto-updating is running. Differs from
         * `_playing` (which has nothing to do with auto-updates) and `_autoUpdate`
         * (which is whether it _should_ use the shared ticker). This helps with 
         * handling `autoUpdate` changes when playing or stopped.
         *
         * @type {boolean}
         * @private
         * @default false
         */
        this._isAutoUpdating = false;
```

themoonrat: Updated both the variable name for clarity, and added jsdocs

bigtimebuddy:
Thanks @themoonrat

PR #6467, opened 2020-03-10
V4 backport localbounds

ivanpopelyshev:
Our v5 bounds update has a few good tests that allow us to backport our epic bounds fixes to v4!

Cumulative changes from PR's: #6365 #6334 #6304 

All this code was CTRL+C - ed.

`_localBounds` from v5 is renamed to `_$localBounds` , because name collision with field in v4 Graphics.

@tywang2006 confirmed that there are people who want to use the fix in production.

### DIFF
--- a/src/core/display/Bounds.js
+++ b/src/core/display/Bounds.js
@@ -40,6 +40,15 @@ export default class Bounds
         this.maxY = -Infinity;
 
         this.rect = null;
+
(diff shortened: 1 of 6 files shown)


tywang2006:
Since I update this version, i found the game performance became much better. especially many spine animation run at the same time.

PR #6468, opened 2020-03-11
Сlarify BGR-RGB-RGBA relations

ivanpopelyshev:
Subj.

A small clean-up with changes in documentation.

https://en.wikipedia.org/wiki/RGBA_color_model

I use `word-order is BGR` to represent the fact that our numbers in `tint` have B as their lowest 8 bits.

`_tintRGB` is different , its has `word-order RGB`

People will make a mistake writing 0xBBGGRR instead so I specified `0xRRGGBB` as a hex template.


packages/graphics/src/Graphics.ts
```diff
@@ -1217,12 +1219,8 @@ export class Graphics extends Container
...
                 const b = (tintRGB[2] * batchTint[2]) * 255;
 
-                // TODO Ivan, can this be done in one go?
-                const color = (r << 16) + (g << 8) + (b | 0);
-
-                batch._tintRGB = (color >> 16)
-                        + (color & 0xff00)
-                        + ((color & 0xff) << 16);
+                // assemble word-order RGB
+                batch._tintRGB = (b << 16) + (g << 8) + (r | 0);
```
SukantPal, line 1223: Why `r | 0`, that doesn't change anything?

packages/graphics/src/Graphics.ts
```diff
@@ -1217,12 +1219,8 @@ export class Graphics extends Container
                 const g = (tintRGB[1] * batchTint[1]) * 255;
                 const b = (tintRGB[2] * batchTint[2]) * 255;
```
SukantPal, line 1220: If you are multiplying three floating point values, then the result need not come out to be an integer. Maybe wrap them around with a `Math.floor/round(...)`. Doing that might be necessary because you're using them in bitwise operations afterwards.

packages/graphics/src/Graphics.ts
bigtimebuddy: Is there a bug for this change @ivanpopelyshev? Need some verification here.

ivanpopelyshev: There's no bug.
However, now that I think about it, `round` will work better there.

packages/graphics/src/Graphics.ts
```diff
@@ -370,7 +372,7 @@ export class Graphics extends Container
      * @param {object} [options] - Collection of options for setting line style.
      * @param {number} [options.width=0] - width of the line to draw, will update the objects stored style
      * @param {PIXI.Texture} [options.texture=PIXI.Texture.WHITE] - Texture to use
-     * @param {number} [options.color=0x0] - color of the line to draw, will update the objects stored style.
+     * @param {number} [options.color=0x0] - color (BGR) of the line to draw, will update the objects stored style.
```
bigtimebuddy, line 375:
```suggestion
     * @param {number} [options.color=0x0] - color 0xRRGGBB (BGR) of the line to draw, will update the objects stored style.
```

packages/graphics/src/styles/FillStyle.ts
```diff
@@ -11,6 +11,7 @@ export class FillStyle
 {
     /**
      * The hex color value used when coloring the Graphics object.
+     * Word-order is BGR
```
bigtimebuddy, line 14:
```suggestion
     * Word-order is BGR, hex template is 0xRRGGBB.
```

ivanpopelyshev: its inner field, if its modified - user has to know what is BGR :)

ivanpopelyshev:
> Would be good to understand what it means that BGR being the word order and its implications.

Word order is independent from machine byte order. We dont know how actually those bytes represented in array if we put that integer there.

That topic is hard to explain, the knowledge requires some practice. its very hard to do those explanations when im not a native speaker!

Please, either accept this thing and give a chance to people who encountered those problems , either spend time on it, make better comments.

PR #6469, opened 2020-03-11
Line height & leading support for vertical text gradients

TheWyo:
##### Description of change
Following up from #6462, here's a rewrite of the loop that now also accounts for lineHeight & leading properties. I think this should hopefully wrap up vertical text gradient issues.

[Playground Demo](https://pixiplayground.com/#/edit/mbWwO7cAkmOwuc7jZOUeZ)

Before:
chrome_2020-03-11_16-21-48

After:
chrome_2020-03-11_16-22-49

##### Pre-Merge Checklist


themoonrat:
Code looks good; will take some time soon to try this code in our games.
It's a little bit tricky since I have a very customised version of Text class, so things like this never merge in simply!

bigtimebuddy:
Ping @themoonrat

themoonrat:
I've set a reminder to add this fix in to our libs tomorrow morning! Then I'll let QA have their beady eyes on new builds.

packages/text/src/Text.ts
```diff
@@ -509,55 +505,68 @@ export class Text extends Sprite
...
-            totalIterations = (fill.length + 1) * lines.length;
-            currentIteration = 0;
 
             // There's potential for floating point precision issues at the seams between gradient repeats.
             // The loop below generates the stops in order, so track the last generated one to prevent
             // floating point precision from making us go the teeniest bit backwards, resulting in
             // the first and last colors getting swapped.
             let lastIterationStop = 0;
 
+            const metrics = TextMetrics.measureText(this.text, style, style.wordWrap, this.canvas);
```
themoonrat, line 519: You already have a measurement of the text outside of this function from line 183. So pass this in as a 3rd param to this function on line 247
`context.fillStyle = this._generateFillStyle(style, lines, measured);`
and you've saved a call to measureText

TheWyo: Good idea, no point measuring twice if we can reuse one.

TheWyo:
Yeah, I figured given that lineHeight < fontSize was already broken anyway it made sense to focus on the more obvious use case for lineHeight. You've got to expect some weirdness if you do that really.
It's probably possible to work out the overlap points and build a gradient up so that it just looks like lower lines are above upper ones or something, but as soon as you introduce dropShadow or stroke you'll break the visuals of that anyway as it's all one fillText call in the end. You'd need multiple text passes to resolve that.

themoonrat:
I've been running this code in production for a week now and there have been no reports of any issues.

PR #6472, opened 2020-03-12
Move mixins to global space

ivanpopelyshev:
This PR is awesome.

It will lead to better integration of resource-loader and all my plugins. We'll get proper mixins declaration in `d.ts` files!

This is initial demo - I expose only DisplayObject. More stuff will be exposed later when we discuss this one.

Many mixins will be exposed for canvas packages.


packages/interaction/global.d.ts
```diff
@@ -0,0 +1,60 @@
+declare type InteractionTrackingData = any;
+
+declare type Cursor = 'auto'
```
cursedcoder: sorcery

cursedcoder:
I'll need to test this locally, will let you know tomorrow :)

packages/interaction/global.d.ts
ivanpopelyshev: Those particular lines were moved it from globals. They will be removed as soon as someone converts interaction to TS :)

packages/accessibility/global.d.ts
```diff
@@ -0,0 +1,8 @@
+declare namespace PixiGlobalMixins
```
bigtimebuddy: Why not just call this namespace `PIXI`?

ivanpopelyshev: I agree to change the name to whatever you want except `PIXI`: this thing does not actually exists, its only a TS hack. I dont want it to collide with actual global namespace if we'll have any. 

Also people will type `PIXI.` in IDE, see stub interfaces and they wont understand whats going on.

bigtimebuddy: Mixins?

cursedcoder: GlobalMixins sounds better IMO

ivanpopelyshev: GlobalMixins it is.

bigtimebuddy:
Good one!

PR #6473, opened 2020-03-14
Fix Typescript global type to export type

JacobFischer:
##### Description of change

This should address https://github.com/pixijs/pixi.js/issues/6445

For a reproduction of this bug see my repo here: https://github.com/JacobFischer/pixi-ts-global-error

Simply put, this exports the `PIXI` type instead of not exporting anything which makes them ambient declarations in Typescript that cause `PIXI` to exist as a global in all projects that use pixi.js + Typescript, regardless if pixi is truly a global or not.

##### Pre-Merge Checklist

### DIFF
--- a/types/assemble.js
+++ b/types/assemble.js
@@ -65,11 +65,8 @@ const eventTypes = fs.readFileSync(path.join(__dirname, 'events.d.ts'), 'utf8');
 const loaderTypes = fs.readFileSync(path.join(__dirname, 'loader.d.ts'), 'utf8');
 
 // Lastly, tsd-jsdoc doesn't support this yet, so we'll inject the module
-// declaration at the end, so that users can use ambient namespace typings
-// like `import 'pixi.js';` OR module typings `import * as PIXI from 'pixi.js';`
-const declareModule = `declare module "${bundle}" {
-    export = PIXI;
-}`;
+// declaration at the end
+const declareModule = `export = PIXI;`;
 
 fs.writeFileSync(tempPath, [buffer, eventTypes, loaderTypes, declareModule].join('\n'));
 


ivanpopelyshev:
I removed this declaration from one of vanilla projects and its broken now. 

Of course its inevitable that we will have separate `d.ts` for people who use modern es6 modules and for those who use plain old `tsc` concatenation with `outFile`.

Right now npm modules have only one "types" line in package json.

Cant really accept it before we make a huge article for everyone on how to bundle pixi+plugins in both cases.

JacobFischer:
What vanilla project is that? Perhaps I could expand this PR to fix the types in that project's case too.

I was under the impression that Typescript types should just define the shape of the module they are defining (with the exception of runtime environments like browsers, node, react-native, etc.).

In that case, is it not an error that the current types are making an assumption about runtime environment? No developer should need to change their plugin bundling system, as this change does not affect anything runtime.

Worst case if they were using PIXI globally with Typescript, they will now have to tell TypeScript to expect PIXI to exist globally via something like:

```ts
import * as PIXI from "pixi.js";

export declare global {
  interface Window {
    PixiOrWhateverYouWantToReferToItGlobally: PIXI;
  }
}
```

Best case if they were using PIXI via modules with TypeScript, some errors may be exposed at build time instead of run time, like in my example repo.

And in the cases of plain JS developers without TS, they will not be affected at all with this change.

ivanpopelyshev:
tsconfig
```
    "compilerOptions": {
        "module": "none",
        "target": "es5",
        "outFile": "dist/my_plugin.d.ts",
    },
    "include": [
        "src/**/*"
    ],
```

main file
```
/// <reference types="pixi.js-legacy" />

namespace my_app {
   
}
```

any of these:

https://github.com/pixijs/pixi-spine/
https://github.com/pixijs/pixi-tilemap/
https://github.com/pixijs/pixi-layers/
https://github.com/pixijs/pixi-projection/

Run `npm i ` or `yarn`, go in `node_modules/pixi.js` , in the end of  `d.ts` leave only `export = PIXI` instead of whole declaration. Build will fail.

I cant change those modules because there are many users who just take the pixi build in JS file, use ts.d definitions from npm. That's the whole big problem with ES6 imports/exports.

bigtimebuddy:
Thank you @JacobFischer for putting something together. Appreciate the help improving our TypeScript workflow.

I’m actually really torn about this PR because I know it’s, in general, the correct approach. There are a couples of issues though that are blockers for me.

First, this is a breaking change for some v5 TypeScript users. While it’s easy for users to define their own global, in reality we have a mix of build environments and some users prefer the ambient typings out of the box. This most closely mirrors our documentation and our historical use of always defining a global. Changing it now forces some uses to make a change in their project, like you pointed out. The alternative, we leave it as it, puts a smaller burden on some developers to simply know if they are using module or ambient types. I think the latter feels less burdensome.

Second, we are _almost_ done with our TypeScript conversion. Changing the current JSDocs to typings was always going to be a temporary solution. Likely our next release 5.3 will have full typings for all packages. At that point we’ll need to see how we separate ambient from module types. So this change, even if we accepted it is going to get thrown out very soon.

I think the more important discussion is not if we should merge this PR, but to decide how we want to handle typings for bundles once we auto generate types. I’m not sure the best approach yet.

bigtimebuddy:
One more thing, this will likely break pixiplayground.com: the editor uses the ambient types for auto-completion.

JacobFischer:
These are excellent points. After cloning down those example projects and playing with them, my PR does indeed break them. It seems to once again come down to the fact that they assume `PIXI` will be a global namespace to inject more namespaces into. So this change is not as trivial as I had hoped.

I was not aware PIXI was being converted to TypeScript (yay!). Hopefully those generated types have less issues with users like me that don't assume globals. Given that context, especially as this PR would be thrown out of the codebase soon no matter what, it's probably best for me to close this PR.

Just to give more input for the future PIXI types, I am a strong advocate for never assuming global types (outside of runtime environments, such as @types/node for Node.js apps). PIXI could probably have 2 (or more) modules for users to include: A base that is just a regular module, and then a global version that re-exports the base PIXI as an ambient declaration so it exists globally for users that want it globally without any more configuration on their end.

Just my 2¢ trying to help this excellent project.

bigtimebuddy:
Other projects like pixi-particles have both ambient and module typings. I hope we can do that here too. Thanks for the feedback, I think we are on the same page with this.

PR #6474, opened 2020-03-15
Convert Interaction to TypeScript

KennethGomez:
##### Pre-Merge Checklist


andrewstart:
Hmm, I know the JSDocs say that `InteractionData.target` is a `Sprite`, but this seems like a good opportunity to correct that to just `DisplayObject` which I believe it always actually has been? Thoughts from other people running the TS conversion?

ivanpopelyshev:
DisplayObject can be imported with `import type`.

packages/interaction/src/TreeSearch.ts
```diff
@@ -1,4 +1,10 @@
 import { Point } from '@pixi/math';
+import { Container } from '@pixi/display';
+import { TilingSprite } from '@pixi/sprite-tiling';
+import { Sprite } from '@pixi/sprite';
+import { Graphics } from '@pixi/graphics';
```
bigtimebuddy: You are including new dependencies for interaction (sprite-tiling, sprite and graphics). Interaction shouldn't know about any of these display objects except Container or DisplayObject. Adding additional dependencies makes the package system a lot less flexible and interaction a lot less extendible. Please find another way to check for required APIs, this might require adding new TS class interfaces, which is okay.

KennethGomez: Is this a good way? cb6af43a9be6ac3eb0074978e5a2fc6c166ab4ad

KennethGomez: I think I've fixed it here c2a8a421dbf3491b2462609edb5d54a8b6a795fe

packages/interaction/src/InteractionManager.ts
```diff
@@ -1027,8 +1076,12 @@ export class InteractionManager extends EventEmitter
...
             displayObject.emit(eventString, eventData);
 
-            if (displayObject[eventString])
+            // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+            // @ts-ignore
+            if (displayObject[eventString] && displayObject[eventString] instanceof Function)
             {
+                // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+                // @ts-ignore
                 displayObject[eventString](eventData);
```
KennethGomez: This can be done with 
```javascript
 const event = Object.entries(displayObject).find((entry) => entry[0] === eventString);
```
instead of `@ts-ignore` but I don't think it's a good way since it'd take more time to fix just a type error

bigtimebuddy: The problem with `Object.entries` and `Array.prototype.find` is that these are not supported across all platforms. We need to support IE and older browsers, some that don't support these ES6 features. In general, we have a polyfills packages that contains some things, but for some Object or Array features, we don't bother with polyfilling or transpiling and simply implement in an ES5 way.

bigtimebuddy:
@KennethGomez can you please fix the linting issues so Travis passes?

Zyie:
Just started going through this and there are a couple of things that i noticed right away.

- Can you add the appropriate access modifier for methods.
- Replace `Function` with a more appropriate type e.g. `() => void`
- Make sure you are marking parameters as optional

NOTE: If a method is marked as protected/private in the JSDoc, however in practise its used like a public method then you can leave it blank. We have been using blank modifiers as our own "internal" modifier.

I'll go through this in more depth once these issues have been sorted out.

packages/interaction/src/InteractionManager.ts
```diff
@@ -1130,12 +1184,18 @@ export class InteractionManager extends EventEmitter
...
      * @private
-     * @param {PointerEvent} originalEvent - The DOM event of a pointer button being pressed down
+     * @param {InteractivePointerEvent} originalEvent - The DOM event of a pointer button being pressed down
      */
-    onPointerDown(originalEvent)
+    onPointerDown(originalEvent: InteractivePointerEvent): void
     {
         // if we support touch events, then only use those for touch events, not pointer events
-        if (this.supportsTouchEvents && originalEvent.pointerType === 'touch') return;
+        if (originalEvent instanceof PointerEvent)
```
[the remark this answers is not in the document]
KennethGomez: Because pointerType is not on all events, just in PointerEvent.

packages/interaction/src/TreeSearch.ts
```diff
@@ -84,52 +91,67 @@ export class TreeSearch
         {
             if (hitTest)
             {
-                if (!(displayObject._mask.containsPoint && displayObject._mask.containsPoint(point)))
+                if (displayObject._mask)
                 {
-                    hitTest = false;
+                    if (!(displayObject._mask instanceof Container) && !(displayObject._mask instanceof MaskData))
```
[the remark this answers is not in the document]
KennethGomez: `containsPoint` does not exist on Container.

KennethGomez:
> Just started going through this and there are a couple of things that i noticed right away.
> 
> * Can you add the appropriate access modifier for methods.
> * Replace Function with a more appropriate type e.g. () => void
> * Make sure you are marking parameters as optional
> 
> NOTE: If a method is marked as protected/private in the JSDoc, however in practise its used like a public method then you can leave it blank. We have been using blank modifiers as our own "internal" modifier.
> 
> I'll go through this in more depth once these issues have been sorted out.

Check 750f4516249d92b67c06ab48ad3a289238afbf8a

Zyie:
> Check [750f451](https://github.com/pixijs/pixi.js/commit/750f4516249d92b67c06ab48ad3a289238afbf8a)

I'm confused, there isn't any changes to the method modifiers? All of them are still blank?
e.g. `processPointerMove` should be private

KennethGomez:
> > Check [750f451](https://github.com/pixijs/pixi.js/commit/750f4516249d92b67c06ab48ad3a289238afbf8a)
> 
> I'm confused, there isn't any changes to the method modifiers? All of them are still blank?
> e.g. `processPointerMove` should be private

Oh methods, I understood fields. I will push tomorrow

packages/interaction/src/TreeSearch.ts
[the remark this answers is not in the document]
ivanpopelyshev: No worries, I'll patch this thing in feature PR after this one :)

packages/interaction/src/InteractionManager.ts
```diff
@@ -1736,28 +1834,61 @@ export class InteractionManager extends EventEmitter
             {
                 const touch = event.changedTouches[i];
 
+                // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
cursedcoder: Makes sense to define this per-file

packages/interaction/src/interactiveTarget.ts
```diff
@@ -126,9 +126,10 @@ export const interactiveTarget = {
      * @memberof PIXI.DisplayObject#
      * @private
      */
-    get trackedPointers()
+    get trackedPointers(): Map<number, InteractionTrackingData>
     {
-        if (this._trackedPointers === undefined) this._trackedPointers = {};
+        if (this._trackedPointers === undefined)
+        { this._trackedPointers = new Map<number, InteractionTrackingData>(); }
```
ivanpopelyshev: That's behavious change. Map !== Object because people have to use `map.get(key)` instead of `map[key]` . 

`get` and `set` aren't used in InteractiveManager

https://github.com/pixijs/pixi.js/blob/cd8dffb5cc74d5deb4616ef671a9d140d1897bce/packages/interaction/src/InteractionManager.js#L1204

Its strange that TS allowed you to build this

bigtimebuddy: Also please don’t use Map, it’s not supported in IE 10

cursedcoder:
Won't it be polyfilled?

bigtimebuddy:
Yes, we polyfill a few things and transpile using Babel, but we are not polyfilling Map. In general polyfills can add a lot of weight so it’s easier to not use these polyfills unless they are used across the project like raf, types arrays etc. Anyways, the goal of this PR is to convert to TypeScript and change as little underlying code. Little “improvements”, like using Map, will undoubtedly hurt us later with compatibility.

packages/interaction/src/interactiveTarget.ts
bigtimebuddy: @KennethGomez would you mind taking another look at this ☝️

bigtimebuddy:
Closing in favor #6656.

PR #6477, opened 2020-03-16
return old behaviour of getLocalBounds for cacheAsBitmap and generateTexture

ivanpopelyshev:
#6365 broke cacheAsBitmap because our mixin relies on different behaviour of getLocalBounds()

The thing is, old behaviour of `getLocalBounds()` can be called with `getLocalBounds(undefined, true)`

I wanted to make it right - work with temp parent like in renderer, and restore transform after the fact just to be sure.

But right now its easier to just patch it back.

broken: https://pixijs.io/examples/#/demos-basic/cacheAsBitmap.js

WIP: https://pixijs.io/examples/?v=dev-cacheAsBitmap#/demos-basic/cacheAsBitmap.js

### DIFF
--- a/packages/mixin-cache-as-bitmap/src/index.js
+++ b/packages/mixin-cache-as-bitmap/src/index.js
@@ -296,7 +296,7 @@ DisplayObject.prototype._initCachedDisplayObjectCanvas = function _initCachedDis
     }
 
     // get bounds actually transforms the object for us already!
-    const bounds = this.getLocalBounds(undefined, false);
+    const bounds = this.getLocalBounds(undefined, true);
 
     const cacheAlpha = this.alpha;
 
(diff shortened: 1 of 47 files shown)


ivanpopelyshev:
Work in progress, I'm fixing stuff

packages/mixin-cache-as-bitmap/src/index.js
```diff
@@ -170,7 +170,7 @@ DisplayObject.prototype._initCachedDisplayObject = function _initCachedDisplayOb
     // this function also calls updatetransform on all its children as part of the measuring.
     // This means we don't need to update the transform again in this function
     // TODO pass an object to clone too? saves having to create a new one each time!
-    const bounds = this.getLocalBounds().clone();
+    const bounds = this.getLocalBounds(undefined, true).clone();
```
bigtimebuddy: Why not null instead of undefined? Shorter and still works.

bigtimebuddy:
Can you port this to v4 too?

PR #6479, opened 2020-03-17
Convert text-bitmap package to TypeScript

lloydevans:
##### Description of change

Convert text-bitmap package to TypeScript.

##### Pre-Merge Checklist


bigtimebuddy:
Alright @lloydevans! Just ping us when this is ready for review or if you have any questions. Lots of packages have been completed so far, check out text, which is pretty close to this one.

lloydevans:
@bigtimebuddy Thanks! will do. Yea did have a couple of questions but looking at the text PR and existing packages answered things.

lloydevans:
Hey @bigtimebuddy,

I think I've done a good chunk of the conversion here. Some of it needs slightly bigger refactoring work to get full type coverage. I've opted for casting things to any over refactoring code just to silence typescript errors for now.

BitmapFont.ts - Converted fine for the most part, with the need to cast data going to the auto-selected format parser as any.

BitmapFontData.ts - The JSDoc comments mark properties as readonly, but it is changed from outside of the class in the format parsers. Maybe this only needs to be an interface now?

BitmapFontLoader.ts - Some need for any around the resource object as the custom bitmap font loader adds some additional properties there.

BitmapText.ts - Fine for the most part, but there are two things. One is the `_font` property has a get/set system which can accept a string or object. I think I've seen discussions on this elsewhere, relating to get/set function signatures matching. Did you come to a solution on that? Same goes for the anchor, in JSDoc it is either a number or point, elsewhere I am seeing it is ObservablePoint. Should I make it ObservablePoint here?

BitmapTextStyle.ts - This is a new file I've added with some interfaces for text style. Would you prefer these interface/types live at the top of the BitmapText.ts file?

TextFormat.ts - This file has a few TS issues with the way it's accessing properties on the bitmap font data object from string attribute names. It is also using parseInt to both cast to number and floor I believe, but TypeScript errors passing a number to parseInt. I've cast things to any for now, but could possibly look at doing a bit of refactoring here to avoid that.

XMLFormat.ts - Apart from the wider point about the format classes, this one converted fine.

packages/text-bitmap/src/formats/TextFormat.ts
```diff
@@ -75,39 +75,44 @@ export class TextFormat
...
+            kerning.second = this.parseValue(kerning.second);
+            kerning.amount = this.parseValue(kerning.amount);
         });
 
         return data;
     }
+
+    private static parseValue(value: string|number): number
+    {
+        return typeof value === 'number' ? Math.floor(value) : parseInt(value, 10);
```
bigtimebuddy: Can you explain the necessity for this new change? If this is a functional change we should add a unit-test for it.

lloydevans: @bigtimebuddy It was because TS requires parseInt parameter to be of type string. Would you prefer to silence the TS error in some way instead of the change?

bigtimebuddy: What about just casting, e.g., `parseInt(kerning.amount as string)`

lloydevans: Good idea, I thought that would work but tried it and get this TS error:

`Conversion of type 'number' to type 'string' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.ts(2352)`

Before I cast it as any like so: `data.kerning.forEach((kerning: any) =>`.

bigtimebuddy: If that's the case, then tslint ignore and leave the code like it was before. Would prefer to avoid `any`

packages/text-bitmap/src/BitmapFontLoader.ts
```diff
@@ -41,17 +46,17 @@ export class BitmapFontLoader
...
         // load is the same number as references in the fnt file
-        const completed = (page) =>
+        const completed = (page: any): void =>
         {
             textures[page.metadata.pageFile] = page.texture;
 
             if (Object.keys(textures).length === data.page.length)
             {
-                resource.bitmapFont = BitmapFont.install(data, textures);
+                (resource as any).bitmapFont = BitmapFont.install(data, textures);
```
bigtimebuddy: @ivanpopelyshev or @Zyie how do we mixin properties to loader's Resource?

packages/text-bitmap/src/formats/TextFormat.ts
lloydevans: Ok, will do

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -431,12 +454,12 @@ export class BitmapText extends Container
      *
      * @member {object}
      */
-    get font()
+    public get font(): IBitmapTextFontDescriptor | string
     {
```
lloydevans: This is one of the cases where the get/set types need to be marked differently

packages/text-bitmap/src/formats/TextFormat.ts
```diff
@@ -38,13 +38,13 @@ export class TextFormat
...
             // Extract item name
-            const name = items[i].match(/^[a-z]+/gm)[0];
+            const name = items[i].match(/^[a-z]+/gm)[0] as keyof BitmapFontData;
 
             // Extract item attribute list as string ex.: "width=10"
             const attributeList = items[i].match(/[a-zA-Z]+=([^\s"']+|"([^"]*)")/gm);
 
             // Convert attribute list into an object
-            const itemData = {};
+            const itemData: any = {};
```
bigtimebuddy: Maybe we do something stricter here?
```suggestion
            const itemData: {[name:keyof BitmapFontData]:string|number} = {};
```

lloydevans: Had to make it this  😛 
`const itemData: Partial<{[value in keyof BitmapFontData]: string|number}> = {};`

lloydevans: Typing this bit is kinda tricky. I guess it is parsing some unknown data.

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -20,13 +31,18 @@ const available = {};
  */
 export class BitmapFont
 {
-    constructor(data, textures)
+    public readonly font: string;
+    public readonly size: number;
+    public readonly lineHeight: number;
+    public chars: Dict<IBitmapFontCharacter>;
```
lloydevans: JSDoc marks `chars` as read-only, but it is set to null below.

packages/text-bitmap/src/BitmapFontLoader.ts
bigtimebuddy: from Ivan: 
> Global mixins #6472 were merged , you can make ILoaderResource global mixin.

lloydevans: Awesome, will do

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -161,7 +184,7 @@ export class BitmapText extends Container
      *
      * @private
      */
-    updateText()
+    public updateText(): void
```
Zyie: This and `updateTransform` should be private or internal (blank) if used publicly?

packages/text-bitmap/src/BitmapFont.ts
Zyie: In previous conversion we have preserved the JSDoc readonly when its set to null in `destroy` 

If you take a look inside `AbstractRender` and its `destroy` method you will see how its been handled in the past

packages/text-bitmap/src/BitmapText.ts
lloydevans: Oh yea, I think `updateTransform` needs to be blank and this can be private.

packages/text-bitmap/src/BitmapFont.ts
lloydevans: Oh yes I see, will do

packages/text-bitmap/src/BitmapText.ts
lloydevans: I think I spotted a couple of others which needed changing too, have updated them.

bigtimebuddy:
Thank you @lloydevans!

lloydevans:
@bigtimebuddy No problem! 😄 Any others I can look at?

PR #6482, opened 2020-03-17
Texture.clone should clone orig, frame, trim Rectangles

### LINKED ISSUE #6480: PIXI.Texture.clone() doesn't clone the frame
### Expected Behavior
cloned frame 

### Current Behavior
not cloned

### Possible Solution
this.frame.clone() and if not null also clone this.orig, this.trim

cheers to Ivan

bigtimebuddy:
Fixes #6480

### DIFF
--- a/packages/core/src/textures/Texture.ts
+++ b/packages/core/src/textures/Texture.ts
@@ -312,7 +312,7 @@ export class Texture extends EventEmitter
         return new Texture(this.baseTexture,
             this.frame.clone(),
             this.orig.clone(),
-            this.trim.clone(),
+            this.trim && this.trim.clone(),
             this.rotate,
             this.defaultAnchor
         );
--- a/packages/core/src/textures/Texture.ts
+++ b/packages/core/src/textures/Texture.ts
@@ -309,7 +309,13 @@ export class Texture extends EventEmitter
      */
     clone(): Texture
     {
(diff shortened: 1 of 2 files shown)


packages/core/src/textures/Texture.ts
```diff
@@ -309,7 +309,13 @@ export class Texture extends EventEmitter
      */
     clone(): Texture
     {
-        return new Texture(this.baseTexture, this.frame, this.orig, this.trim, this.rotate, this.defaultAnchor);
+        return new Texture(this.baseTexture,
+            this.frame.clone(),
+            this.orig.clone(),
+            this.trim.clone(),
+            this.rotate,
+            this.defaultAnchor
```
themoonrat: Are we ok with the defaultAnchor copied by reference?

EneaEntertainment: I think trim could be undefined or null perhaps and needs check

bigtimebuddy: defaultAnchor is cloned in the constructor anyways

bigtimebuddy: I added another test check for `defaultAnchor` to make sure it isn't the same reference. Also, made `trim` optional (thanks @EneaEntertainment)

ivanpopelyshev:
In other libs things are usually cloned in constructor, but we cant do that in case someone uses that behaviour.

PR #6484, opened 2020-03-20
[WIP] generate d.ts for packages with @microsoft/api-extractor

zxc122333:
**Do NOT merge yet!**

##### Description of change

Add a build script to generate d.ts for each packages.

 [preview here](https://github.com/zxc122333/pixi.js/blob/dev-pkg-dts-preview/packages/core/lib/core.d.ts)

This PR just for preview generated result. I will reorganize and create new PR after known issues fixed.

If core team prefer to use jsdoc-tsd, just let me know and close this.

Known issues :

1. [microsoft/rushstack#1029](/microsoft/rushstack/issues/1029), but [the fork](/adventure-yunfei/web-build-tools/tree/feature/AE7/local-module-import) in comment works fine. Waiting for merged.
2. declare conflict in `mini-signals` and `resource-loader`. Waiting for`resource-loader@4.0.0` release.
3. Some import like 
   ```
   // in packages/core/src/textures/resources/BaseImageResource.ts
   import type { BaseTexture, Renderer, GLTexture, ImageSource } from '@pixi/core';
   ```
   should be fix to
   ```
   import type { BaseTexture, ImageSource } from '../BaseTexture';
   import type { Renderer } from '../../Renderer';
   import type { GLTexture } from '../GLTexture';
   ```
4. Some jsdoc in interfaced value will be drop. [settings.d.ts](/zxc122333/pixi.js/blob/7048aee914ba514a9bec5f173d17e2d4eff2e3e3/packages/settings/lib/settings.d.ts#L57) 

How to:
1. bootstrap like before
2. clone the @microsoft/api-extractor fork, build it, and `npm link` it.
3. modify `node_modules/mini-signals/package.json`, remove `types` field.
4. `npm run types:packages`

##### Pre-Merge Checklist


package.json
```diff
@@ -53,12 +55,13 @@
...
     "@typescript-eslint/parser": "^2.1.0",
     "floss": "^3.0.0",
+    "fs-extra": "^9.0.0",
     "chai": "~3.5.0",
     "sinon": "^7.3.2",
     "sinon-chai": "^3.3.0",
     "nyc": "^14.1.1",
     "glob": "^7.1.3",
-    "jsdoc": "^3.5.5",
+    "jsdoc": "^3.6.3",
```
bigtimebuddy: Just FYI, we have intentionally not upgraded to 3.6.x of JSDoc because it breaks our docs. We are pinned to 3.5.5 for now. This is a much bigger task to upgrade. We know it doesn't work in Node 12. Eventually we'll transition to typedoc instead.

bigtimebuddy:
Would welcome a PR to fix those import errors in core, for sure! Good spotting that. 

Thank you for getting a start on this @zxc122333. Our TypeScript process is still a work in progress, I'm glad to see so many people helping to push this effort along.

It'll be helpful for some context and to discuss approaches here if you continue to go down this path. Our goal is always to generate types once the TypeScript conversion is complete, we are like 80% of the way there. So you've beat us to the punch! Good job.

I'm not opposed to api-extractor, but maintaining all those config json files in each package definitely seems like they're an easier way to do that. I'd prefer to have type generation happen through our existing Rollup build process, if possible. There are Rollup plugins that generate types and potentially bundle them as well.

zxc122333:
> Would welcome a PR to fix those import errors in core, for sure! Good spotting that.

There it is. #6485 

> I'm not opposed to api-extractor, but maintaining all those config json files in each package definitely seems like they're an easier way to do that.

Yes. 

`api-extractor` provide api for create `ExtractorConfig` programmatically, but there is so many undocumented path arguments need to fill. So I generate lots of `api-extractor.json` to get the preview result quickly.

After they fix microsoft/rushstack#1029, I will finish this.

> I'd prefer to have type generation happen through our existing Rollup build process, if possible. There are Rollup plugins that generate types and potentially bundle them as well.

I prefer too....

In fact, I spent lots of time in Rollup before I started writing a new build script.

I'll keep trying to find a solution.

There are some blocker I found in my previous attempts

1. Most typescript plugin can't share instance between builds like our existing build process (share ts compiler is actually we want).

    The `rollup-plugin-typescript` we used is no longer maintained, and can not emit d.ts.
    The successor `@rollup/plugin-typescript` can not share instance too.

    Separate compiler instance for each package will increase build time greatly.

2. Most non plug-in tools can't match Rollup simply because they r/w files directly and don't provide injection mode.

bigtimebuddy:
That’s helpful information. Thanks.

zxc122333:
[Preview](https://github.com/zxc122333/pixi.js/blob/dev-pkg-dts-preview/packages/core/lib/core.d.ts) is updated after we merged #6485, all ugly things like `import { Program as Program_2 } from '@pixi/core';` are gone. 🎉 

---

Maybe we need to discuss about put type generation into existing Rollup build process.

After some trial and research, I realized that maybe we should put the task of generating `d.ts` file outside of rollup build process.

Like the official document of rollup said, 

> Rollup is a module bundler for JavaScript which compiles small pieces of code into something larger and more complex, such as a library or application.

Rollup is a bundler, not a common build tool or task runner  like GNU-Make/Gulp/Grunt etc. It focuses on packaging many ECMAScript files into one.

Although it has a plug-in system that can handle non es input such as `typescript`/`rollup-plugin string`, it still lacks lots of power about other output types. 

And Rollup team forces the output type to be some JS module schemes, such as `esm`/`cjs`/`iife`, not `d.ts`, `css`, `html`.

Maybe we can live the type generation outside Rollup, like we generate `pixi.js.d.js`?

---


Btw, please take a look at #6488, I need some help or discuss there.

andrewstart:
Because you didn't mention it by name, have you tried `rollup-plugin-ts`? It works okay for me in pixi-animate to build code and bundled declarations in one go, although I haven't tried setting it up to include external modules the way that the full PixiJS build should.

ivanpopelyshev:
```
import type { BaseTexture, ImageSource } from '../BaseTexture';
import type { Renderer } from '../../Renderer';
import type { GLTexture } from '../GLTexture';
```

I'm a bit sad because of it. Our core has multiple semi-modules inside, and I dont see the reason why not import types from other modules through root. However, its not critical, we can think about it later.

zxc122333:
> Because you didn't mention it by name, have you tried `rollup-plugin-ts`? It works okay for me in pixi-animate to build code and bundled declarations in one go, although I haven't tried setting it up to include external modules the way that the full PixiJS build should.

I tried, but it cannot share the instance between each package. It took more than 30 minutes to build all packages. By contrast, the current `rollup-plugin-typescript` on the same machine use 22 seconds. 

Even without generate d.ts , it took more than 90 seconds. 

Another problem, it can't handle the `earcut` module. It will crash by an unhandled exception `Cannot read property 'kind' of undefined` from very deep of ts compiler. No matter generate declarations or not.

In the source of earcut :
```javascript
module.exports = earcut;
module.exports.default = earcut;   // this line
```

---

> ```
> import type { BaseTexture, ImageSource } from '../BaseTexture';
> import type { Renderer } from '../../Renderer';
> import type { GLTexture } from '../GLTexture';
> ```
> 
> I'm a bit sad because of it. Our core has multiple semi-modules inside, and I dont see the reason why not import types from other modules through root. However, its not critical, we can think about it later.

I think its ok except `api-extractor` will translate it to `import { Program as Program_2 } from '@pixi/core';`

If we finally choose `api-extractor`, maybe we can try to request them improve this. Not sure they will accept or not.

`import type { Renderer } from '@pixi/core'` works here because our tsconfig.json has declared `paths: {"@pixi/*": ["packages/*/src"]}`. This is not the standard behavior of ES module.

ivanpopelyshev:
Ugh, guys, I think we should focus on it

bigtimebuddy:
Yes it's on my radar.

bigtimebuddy:
Closing in favor of #6849

PR #6486, opened 2020-03-21
Fix SVG disposed while loading

maximgavrilov:
##### Description of change
If SVGResource was disposed during load, do not draw it on canvas.

##### Pre-Merge Checklist


cursedcoder:
Also, I can't see the green build.

maximgavrilov:
But there no "done" events in this case, because they are suppressed with dispose.

Unit can be something like
```
const resource = new SVGResource(                                   
    path.join(this.resources, 'heart.svg'),                         
);                                                                  

resource.dispose();                                                                    

// ??? wait some time
// without patch it throws error ('canvas is null' in _loadSvg::tempImage.onload)
// with patch nothing is throwed
// maybe make resource.load() promise rejected on dispose? I am not certain with this.
```

bigtimebuddy:
Travis is having issues atm

bigtimebuddy:
Thank you @maximgavrilov!

PR #6487, opened 2020-03-21
Fix AnimatedSprite's current texture index does not match current frame index

JetLua:
##### Description of change

**Note: The page should display 3, which may need to be refreshed many times due to the low probability of bugs.**

demo: https://www.pixiplayground.com/#/edit/oSNBXIXePT__SL3KYhOoI
fixed: https://www.pixiplayground.com/#/edit/7Fjo_WQTq6iloKAAOJ2UN

Because `currentFrame` is computed dynamically, there is always a problem with 
```js
const previousFrame = this.currentFrame;
if (previousFrame !== this.currentFrame) {
  ...
}
...
```

**So I added a new property to save the texture index from the last updateTexture.**

##### Pre-Merge Checklist


themoonrat:
lgtm - but is there a unit test that can cover why this fix works, and the current code doesn't? https://github.com/pixijs/pixi.js/pull/6274 may have introduced this issue whilst fixing another, but it did also include lots of nice unit tests to narrow down expected behaviour

bigtimebuddy:
I agree with @themoonrat. Can you make a test please @JetLua?

JetLua:
@bigtimebuddy @ivanpopelyshev @themoonrat Updated PR and description.

ivanpopelyshev:
I like the test!

JetLua:
> I like the test!
> 
> Wait, I saw 2 and 1 somehow, in fixed version. Do you have pixijs build somewhere that we can reference from that playground? Also, even if you specify settings in playground - modify code, and press "save" or it wont be really saved. Its buggy.

Yes, the custom url did not take effect.

https://static.lufei.so/tmp/pixi.js

bigtimebuddy:
Checks out https://jsfiddle.net/bigtimebuddy/qwo32ph9/5/

PR #6488, opened 2020-03-22
[WIP] Fix some dependencies missing between packages

zxc122333:
**Do NOT merge yet!**

##### Description of change

Fix some dependencies missing between packages.

Since we didn't publish declarations for each packages, everything is ok for now.

When we published, and someone use `npm install @pixi/core` instead of `npm install pixi.js`, those undeclared dependencies will cause some error like `cannot found module @pixi/xxxxx`.

Simply add all package to `dependencies` will cause many dependency cycles, so I move some type-only dependencies to `peerDependencies`.

But I don't think the issue is solved, some examples:

1. `@pixi/spritesheet` is an optional plugin for `@pixi/loader`, but we have a `spritesheet?: Spritesheet;` property inside `ILoaderResource`.

    This caused @pixi/loader -> @pixi/spritesheet -> @pixi/loader 

    Maybe we can define ILoaderResource as a generic type? Not sure about this, don't known what should be at js side.

2. `@pixi/extract` is an optional plugin for `@pixi/core`, we have a `public extract: Extract;` property inside `Render`.

    This caused @pixi/core -> @pixi/extract -> @pixi/core 

    Since we documented it as `renderer.plugins.extract`, can we mark `renderer.extract` as deprecated, and live an `any` for compatibility?

3. Should `ProgramCache`, `TextureCache`, `BaseTextureCache` live in `@pixi/utils`? May be `@pixi/core` is more reasonable?

    This caused @pixi/core -> @pixi/utils -> @pixi/core

4. Since `@pixi/core` can not work without `@pixi/ticker`(used by VideoResource), is it right for `@pixi/ticker` as a plugin? Especially as a plugin for Application?

##### Pre-Merge Checklist

### DIFF
--- a/packages/loaders/package.json
+++ b/packages/loaders/package.json
@@ -23,11 +23,13 @@
     "lib",
     "dist"
   ],
-  "dependencies": {
-    "@pixi/core": "5.2.1",
-    "resource-loader": "^3.0.1"
+  "peerDependencies": {
+    "@pixi/spritesheet": "5.2.1"
   },
-  "devDependencies": {
-    "@pixi/utils": "5.2.1"
+  "dependencies": {
+    "@pixi/app": "5.2.1",
+    "@pixi/core": "5.2.1",
+    "@pixi/utils": "5.2.1",
+    "resource-loader": "^3.0.1"
   }
 }
--- a/packages/ticker/package.json
+++ b/packages/ticker/package.json
@@ -23,6 +23,9 @@
     "lib",
     "dist"
   ],
+  "peerDependencies": {
+    "@pixi/app": "5.2.1"
+  },
   "dependencies": {
     "@pixi/settings": "5.2.1"
   }
--- a/packages/accessibility/package.json
+++ b/packages/accessibility/package.json
@@ -27,6 +27,7 @@
   "dependencies": {
     "@pixi/core": "5.2.1",
     "@pixi/display": "5.2.1",
+    "@pixi/math": "5.2.1",
     "@pixi/utils": "5.2.1"
   },
   "devDependencies": {
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -25,6 +25,7 @@
   ],
   "dependencies": {
     "@pixi/core": "5.2.1",
+    "@pixi/math": "5.2.1",
     "@pixi/display": "5.2.1"
   },
   "devDependencies": {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -30,6 +30,7 @@
   "dependencies": {
     "@pixi/constants": "5.2.1",
     "@pixi/display": "5.2.1",
+    "@pixi/extract": "5.2.1",
     "@pixi/math": "5.2.1",
     "@pixi/runner": "5.2.1",
     "@pixi/settings": "5.2.1",
--- a/packages/extract/package.json
+++ b/packages/extract/package.json
@@ -23,8 +23,11 @@
     "lib",
     "dist"
   ],
+  "peerDependencies": {
+    "@pixi/core": "5.2.1"
+  },
   "dependencies": {
-    "@pixi/core": "5.2.1",
+    "@pixi/display": "5.2.1",
     "@pixi/math": "5.2.1",
     "@pixi/utils": "5.2.1"
   }
--- a/packages/filters/filter-color-matrix/package.json
+++ b/packages/filters/filter-color-matrix/package.json
@@ -25,6 +25,7 @@
     "dist"
   ],
   "dependencies": {
-    "@pixi/core": "5.2.1"
+    "@pixi/core": "5.2.1",
+    "@pixi/utils": "5.2.1"
   }
 }
--- a/packages/filters/filter-displacement/package.json
+++ b/packages/filters/filter-displacement/package.json
@@ -25,6 +25,7 @@
     "dist"
   ],
   "dependencies": {
+    "@pixi/constants": "5.2.1",
     "@pixi/core": "5.2.1",
     "@pixi/math": "5.2.1"
   }
--- a/packages/sprite-animated/package.json
+++ b/packages/sprite-animated/package.json
@@ -25,6 +25,7 @@
   ],
   "dependencies": {
     "@pixi/core": "5.2.1",
+    "@pixi/display": "5.2.1",
     "@pixi/sprite": "5.2.1",
     "@pixi/ticker": "5.2.1"
   }
--- a/packages/text/package.json
+++ b/packages/text/package.json
@@ -25,6 +25,7 @@
   ],
   "dependencies": {
     "@pixi/core": "5.2.1",
+    "@pixi/display": "5.2.1",
     "@pixi/math": "5.2.1",
     "@pixi/settings": "5.2.1",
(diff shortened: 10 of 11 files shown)


zxc122333:
Maybe I should try to finish this PR by following rules, and leave the unresolved problems to the future?

1. dependencies only for test -> devDependencies
2. plugin for some package -> peerDependencies
3. necessary dependence -> dependencies

---

Unresolved problems to be discussed:

1. What should be a plugin? For a example `@pixi/ticker`:

    1. `@pixi/ticker` should not be a plugin, because `VideoResource` is strongly dependent on it. 
        `VideoResource` is not an optional plugin, so `@pixi/ticker` neither.

    2. `@pixi/ticker` should be a plugin, and we should decoupling `VideoResource` from dependence on it.

2. Should a pluggable system reference to its plugin? Example:

    ```typescript
    class Render {
        public extract: Extract;
    }
    ```

3. Can we describe a clear dependency graph between packages? 

    We can render it into an image and put it in document, for someone want to use part instead of whole bundled pixi.js. In this way, they can easily know the dependencies between packages, pick what they need and omit what they don't.

4. Should we make a rule for when we should split code into a separate package?

bigtimebuddy:
I believe `peerDependencies` are not going to work because of a lack of publishing/version support in Lerna. I could be wrong and maybe there's a better workflow around this now. But this needs more investigation.

packages/loaders/package.json
```diff
@@ -23,11 +23,13 @@
     "lib",
     "dist"
   ],
+  "peerDependencies": {
+    "@pixi/spritesheet": "5.2.1"
```
bigtimebuddy: We need a way to decouple spritesheet and loaders. Using a peerDependency is not the solution. Creating a shared typing or interface is better.

packages/ticker/package.json
```diff
@@ -23,6 +23,9 @@
     "lib",
     "dist"
   ],
+  "peerDependencies": {
+    "@pixi/app": "5.2.1"
```
bigtimebuddy: Again, we need a typing solution to decouple. peerDependencies isn't the correct solution here.

ivanpopelyshev:
Global mixins #6472 were merged , you can make ILoaderResource global mixin.

zxc122333:
> Global mixins #6472 were merged , you can make ILoaderResource global mixin.

Very cool #6472, I will create a new PR with global mixin.

zxc122333:
@ivanpopelyshev
I tried and found that GlobalMixin only works at internal pixi.js. When I try to generate type description file for each package, I find that they are not treated as the same namespace.

After some searches, I found this:

https://github.com/Microsoft/TypeScript/wiki/FAQ#why-dont-namespaces-merge-across-different-module-files

and 

https://www.typescriptlang.org/docs/handbook/modules.html#do-not-use-namespaces-in-modules

As the official documents say, the namespace GlobalMixin in `@pixijs/accessibility and` `@pixijs/display` are two separate and independent namespaces. I think it works internal because the compiler treats the entire pixi.js as a single project.

zxc122333:
Btw, any suggestions on how to deal with `ProgramCache`, `TextureCache`, `BaseTextureCache ` in `@pixi/utils`?

This causes cycle `@pixi/core -> @pixi/utils -> @pixi/core`

ivanpopelyshev:
> As the official documents say, the namespace GlobalMixin in `@pixijs/accessibility and` `@pixijs/display` are two separate and independent namespaces.

Those are global namespaces. They worked even before imports/exports were introduced. If they dont work in some microsoft lib, its the problem of said lib that has to be fixed.

zxc122333:
> Those are global namespaces in declaration files. They worked even before imports/exports were introduced. If they dont work in some microsoft lib, its the problem of said lib that has to be fixed.
> 
> It has to be added in resulting `d.ts` file exactly like that, and not as part of module.

I can't find another way to make GlobalMixin work in separated type declaration, except export everything to global with
```
declare module '@pixi/core'
{
      export = PIXI
}
```
like the current bundled release.

I think this is not the expected behavior for those who want to use separated packages. The global variable PIXI still leaking everywhere.

Even I'm trying to build a framework with some part of pixi.js as renderer, end users will see PIXI. Since I won't let PIXI to be global, any code touches PIXI will cause runtime exceptions, instead of compile fail.

Any suggestions i can follow?

Btw, not related to any 'some Microsoft lib'. I just try to simulate the global namespace merge behavior by handwriting some type descriptions.

ivanpopelyshev:
OK, just give me some instructions and I'll try to fix it.

>  I just try to simulate the global namespace merge behavior by handwriting some type descriptions.

it should work. Maybe you are missing something small. If it doesnt work - it'll be a huge problem for all my plugins and supposed resource-loader typings :(

```
- Comrade Zampolit! We have no Ammo!
- But Ivanov, You are a communist!
And machine gun began shoot with renewed vigor.
```

zxc122333:
@ivanpopelyshev I made a small repository for this. 

[github.com/zxc122333/globalmerge](github.com/zxc122333/globalmerge)

ivanpopelyshev:
I looked at repo and i dont see a problem.

`tsc` works for me

```
// This should fail at compile time, but it pass
```

That's the designed behaviour - if you extend global interface in a plugin and include that plugin from any file - classes that extend that interface are modified, you can pass new params!

Imagine we have `ILoaderResource` and or sprite plugin adds `spritesheet` field in it.

If we have `IResourceMetadata` interface that we pass  to `loader.add{{... metadata: myMetadata });`  , `pixi-spine` adds extra possible fields for metadata, we dont have to override `add` method and make new class for `loader`.

Maybe you want

packages/loaders/package.json
MM19871987: i tell you guys all reary that i do not want to see nothing about you guys in mu emails address stop do your onw cloud please migmu11@icloud.com is  the orilligal address and migmu2008@outlook.com is  my  [miguel-murcia] owner

packages/ticker/package.json
MM19871987: go away from the nintendo of my son app

MM19871987:
let my email address migmu11@icloud.com free is no yours facebook account with the name of [miguel-murcia-caravantes] too

zxc122333:
I think I didn't make it clear.

I know the parameter type is exactly what we want, but the problem is not the parameter type.

```
import * as Plugin from '@merge/plugin';

Plugin.sayHelloToPlugin({foo: 'foo', bar: 'bar'})


// This should fail at compile time, but it pass
Merge.sayHelloToCore({foo: 'foo', bar: 'bar'})
```

This should failed because I didn't import anything about `Merge`, the global variable `Merge` doesn't exist.

ivanpopelyshev:
This should work:

core:

```ts
declare namespace GlobalMerge {
  export interface IPluggable {}
}

declare module '@merge/core'{
  export interface Pluggable extends GlobalMerge.IPluggable {
  }
  export class Pluggable {
    foo: string;
  }
  export function sayHelloToCore(target: Pluggable): void;
}
```

plugin:
```ts
declare namespace GlobalMerge {
  export interface IPluggable {
    bar: string;
  }
}

declare module '@merge/plugin' {
  export function sayHelloToPlugin(target: import("@merge/core").Pluggable): void;
}
```

PR #6493, opened 2020-03-24
[WIP] Optimizes bounds calculations in scene graph

SukantPal:
*- For a theoretical understanding of these changes: https://pdfhost.io/v/bnKsyHEN7_Optimization_of_Bounds_Calculations_in_PixiJS_Scene_Graphs2pdf.pdf

This PR is work-in-progress; however, due to its scope **it is ready for review**.

##### Description of change
This change will implement these two things in the white paper:

## Preventing getLocalBounds() from dirtying world transforms in the target's subtree (Using Section 2: Co-existence of non-local transformation matrices)

When calculating the local bounds of a (target) display-object, the world transforms of each node in its subtree will be "saved". Then the world transform w.r.t. to the target display-object will be calculated for each node by prepending them with the target's world-transform-inverse.

This works because the world transform of a node is the multiplication of the all the local transforms from top to bottom. If we want to start from the target (and not the actual root display-object), we **can "cancel" out the local transforms before the target's local transform in that chain**. This is done by prepending with the target's world transform inverse.

After calculating the local bounds, the world transform is restored.

```js
        let worldTransform;

        if (rootInv)
        {
            /* Save actual worldTransform */
            worldTransform = matrixPool.pop() || new Matrix();
            worldTransform.copyFrom(this.transform.worldTransform);

            /* Calculate world transform for local bounds */
            this.transform.worldTransform.prepend(rootInv);
        }

        this._calculateBounds();

        /* Restore */
        if (rootInv)
        {
            // swap instead of copy into
            matrixPool.push(this.transform.worldTransform);
            this.transform.worldTransform = worldTransform;
        }
```

----

## Bounds propagation - don't recalculate bounds if nothing has changed. updateTransform() effectively checks if something changed.

The update-transform method won't increment `_boundsID` (effectively) if the transform itself hasn't changed. However, I changed the name (or repurposed) `_boundsID`.  `_boundsID` just reflects the changes in the intrinsic/content bounds of the display-object. If it is dirty, then the cached bounds are dirty too. However, the cached bounds can be dirty also if

1. the world transform of the node has changed.

2. the local transforms of its (nested) children have changed.

3. `visible`, `renderable`, or `mask` has changed.

For this, `_subtreeBoundsID` is to the rescue. `_subtreeBoundsID` is the sum of the node's `_boundsID` (handling changes in content), `transform._worldID` (handling changes in world transform), and its children's `_subtreeBoundsID` values (handling changes in children). Point 3 (visible, renderable, and mask properties) are handled by incrementing `_boundsID` whenever those values change (via setter).

*- In the paper, I didn't consider some stuff like dependency on visible, renderable, or mask properties. That has been handled properly in this PR.

## Proof: Updates in child boundsID will not cancel each other

In other words, changes downstream of a target node will **always** be reflected in its `_subtreeBoundsID`. This is because `_subtreeBoundsID` is a sum of three **always non-decreasing** things: `_boundsID`, `transform._worldID`, and child `_subtreeBoundsID` values. It is impossible for changes in one of these values to cancel out a change in another.

## Other changes

I've updated the documentation for `Matrix` and `Transform`. I've also added some tests related to the changes.

##### Pre-Merge Checklist


bigtimebuddy:
Could you demonstrate the performance differences between what's current implemented and this approach? Would like to get a sense of the scale of benefit.

SukantPal:
@bigtimebuddy Check this: https://jsfiddle.net/ShukantPal/wy42v7e9/10/. About a 30% better.

If you change `getBounds()` to `getLocalBounds()`, my branch does significantly worse. My branch performs about 67% worse then. I will look at that problem tomorrow 💯.

SukantPal:
@bigtimebuddy This is the test that shows `getLocalBounds()` is slow on my branch right now: https://jsfiddle.net/ShukantPal/c14L0dq6/

SukantPal:
@bigtimebuddy This is a test that shows both `getBounds()` and `getLocalBounds()` being called on after another. My branch is again about +70% faster.

https://jsfiddle.net/ShukantPal/L9f5u4mj/

ivanpopelyshev:
ITs good that you made PR, it sure gives the idea exposure.

SukantPal:
Apart from updateTransform, almost no other behavior has changed. It isn’t even supposed to be overridden and if we want to support “hacks” then we’ll never get any fixes out.

Also, if other modules are waiting on Typescript conversion, this shouldn’t affect them.

bigtimebuddy:
I agree with @SukantPal about breaking behavior. There's a reason we have a public/protected API. Anything that is private or internal only is fair-game. But obviously, if we are going to break well-used plugins we maintain, we should consider workarounds.

Obviously, this need some more review. @GoodBoyDigital said he'd look at it more closely.

ivanpopelyshev:
OK, we talked and this thing is hard to do. I think we can do it without recursive sum, just regular usm of child bounds id. recursive updateTransform  call remains. 

Yes, its not as effective as dirty flags propagation to parents could be, but it should be good enough for our pixijs stage.

The other thing is that we can try add experimental flag to use that feature, because it can have breaking behaviour in a few cases.

And the biggest problem is that `generateTexture` and `cacheAsBitmap` will be broken again, before I refactor them in separate PR. I need my parent-detach API to do that #6450 .

MM19871987:
i do not want to see nothing in my divice about github please this is my familly personal things that github change go to hack to others i let give you the chance to to see my things and trying to see if you can jelp me to do money but what you did just change all the codes and now my icloud account and all that i saved from 2011 is yours and apple and mucrosoft and all the people that have more money the us i we are the owners

MM19871987:
http://www.nowwhere.com.au/lic/NowWhereLic.htm apple put this name to start all to my date

MM19871987:
and againg i do not want to see your emails in my i cloud account no more

MM19871987:
migmu11@icloud.com and migmu2008@outlook.com is my MM19871987

SukantPal:
@englercj Where did those comments come from? Someone was talking about Apple ID & bunch of other stuff!

bigtimebuddy:
This is probably spam. We get it occasionally. I have blocked that user and reported them.

SukantPal:
@GoodBoyDigital @bigtimebuddy 

This PR is slower when the _whole scene graph_ becomes dirty. If the whole scene graph becomes dirty, than getLocalBounds takes twice as long sometimes.  This is because it both calculates the world transforms & prevents world transforms from being overwritten by calculating the relative transforms, which is a bug in current dev. getBounds is also slightly slower for some reason.

HOWEVER, let's say only 3 out of 16 children of the root node are dirty, than the newer version of getBounds() is clearly much faster:

https://jsfiddle.net/ShukantPal/d9smjuf0/

**Summary**: 
This PR brings both correctness & speed to bounds calculations. But it is slower when the whole scene graph is dirty.

If we don't merge this, then we should at least document that getLocalBounds() are illegal in render passes without subsequent updateTransform()s.

englercj:
My machine with that fiddle you linked:

image

SukantPal:
@englercj 

MBP 2018 (Catalina):

Safari:

Screen Shot 2020-05-29 at 5 36 47 PM

Chrome: 

Screen Shot 2020-05-29 at 5 37 20 PM

Firefox:

Screen Shot 2020-05-29 at 5 37 35 PM

Edge:

Screen Shot 2020-05-29 at 5 38 02 PM

PR #6496, opened 2020-03-25
Remove unnecessary fields in Sprite

GreenishFlow:
It looks like AbstractBatchRenderer does not use those fields, only indices.
I made it because of collision with method "start" in my app migrated from v4.


bigtimebuddy:
Thank you @GreenishFlow!

PR #6499, opened 2020-03-26
Convert `canvas-renderer` to typescript

ivanpopelyshev:
A hour ago I wanted to make a draft. 

I woke up when it was done.

Notice how `GlobalMixins` helped in that case.

I moved out `_canvasRenderTarget` field from `BaseRenderTexture` , we can move class `CanvasRenderTarget` from `PIXI.utils` in another PR.

Another idea is that we can move `forcePixels` to globalMixins too, thus preventing people from passing it to `autoDetect` if they didnt use `pixi.js-legacy` , - also for discussion in separate PR.


packages/canvas/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -38,16 +68,16 @@ export class CanvasRenderer extends AbstractRenderer
...
-        super(RENDERER_TYPE.CANVAS, options, arg2, arg3);
+        super(RENDERER_TYPE.CANVAS, options);
 
         /**
          * The root canvas 2d context that everything is drawn with.
          *
          * @member {CanvasRenderingContext2D}
          */
-        this.rootContext = this.view.getContext('2d', { alpha: this.transparent });
+        this.rootContext = this.view.getContext('2d', { alpha: this.transparent }) as any;
```
bigtimebuddy: Why `as any`?

Zyie:
This looks good to me as well. 

Can you add `public` to the class methods? I assume these all are not meant to be `internal`

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
ivanpopelyshev: Because line is too long

bigtimebuddy: So break it into two. Using `any` should be a last-resort for resolving typing issues, not workaround linting

PR #6500, opened 2020-03-26
Working example in SimplePlane docs

tomcolaa:
##### Pre-Merge Checklist

### DIFF
--- a/packages/mesh-extras/src/SimplePlane.ts
+++ b/packages/mesh-extras/src/SimplePlane.ts
@@ -6,10 +6,14 @@ import { PlaneGeometry } from './geometry/PlaneGeometry';
  * The SimplePlane allows you to draw a texture across several points and then manipulate these points
  *
  *```js
- * for (let i = 0; i < 20; i++) {
- *     points.push(new PIXI.Point(i * 50, 0));
- * };
- * let SimplePlane = new PIXI.SimplePlane(PIXI.Texture.from("snake.png"), points);
+ * let plane = new PIXI.SimplePlane(PIXI.Texture.from("snake.png"));
+ * let geometry = new PIXI.Geometry();
+ *
+ * geometry.addAttribute('positions', [0, 0, 100, 0, 100, 100, 0, 100], 2);
+ * geometry.addAttribute('uvs', [0,0,1,0,1,1,0,1], 2);
+ * geometry.addIndex([0,1,2,1,3,2]);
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks @tomcolaa!

packages/mesh-extras/src/SimplePlane.ts
```diff
@@ -6,10 +6,14 @@ import { PlaneGeometry } from './geometry/PlaneGeometry';
...
- * };
- * let SimplePlane = new PIXI.SimplePlane(PIXI.Texture.from("snake.png"), points);
+ * let plane = new PIXI.SimplePlane(PIXI.Texture.from("snake.png"));
+ * let geometry = new PIXI.Geometry();
+ *
+ * geometry.addAttribute('positions', [0, 0, 100, 0, 100, 100, 0, 100], 2);
+ * geometry.addAttribute('uvs', [0,0,1,0,1,1,0,1], 2);
+ * geometry.addIndex([0,1,2,1,3,2]);
+ *
+ * plane.geometry = geometry;
```
bigtimebuddy:
```suggestion
 * let plane = new PIXI.SimplePlane(PIXI.Texture.from("snake.png"));
 * let geometry = plane.geometry;
 *
 * geometry.addAttribute('positions', [0, 0, 100, 0, 100, 100, 0, 100], 2);
 * geometry.addAttribute('uvs', [0,0,1,0,1,1,0,1], 2);
 * geometry.addIndex([0,1,2,1,3,2]);
```

tomcolaa: Hi @bigtimebuddy, I tried that as well but it didn't work. That's why I choose to overwrite plane.geometry. I created a playground to test it: https://www.pixiplayground.com/#/edit/0-90FDBNusGDisJw4vIc7

bigtimebuddy: This is probably not a suitable SimplePlane example then. Probably a better example for Mesh.

ivanpopelyshev:
Exca made demo: https://pixijs.io/examples/#/demos-basic/simpleplane.js we can just take it from there

bigtimebuddy:
I'm going to close this, this doc is not the right approach. I'd suggest something much simpler or removing the SimplePlane example all together from the code. The example Ivan pointed to is a much better example how to use the class.

PR #6502, opened 2020-03-27
Add support for lineJoin, lineCap and miterLimit to Graphics line rendering

SukantPal:
**Progress**: Bevel is now working, round join working, miter limit working, round cap & square cap working (butt was default).

This PR will add the following features:

* Bevel line join
* Round line join
* Miter limit
* Round cap
* Square cap

##### Pre-Merge Checklist


SukantPal:
Oops, I got other commits into this.

SukantPal:
@bigtimebuddy Check this out: https://pixiplayground.com/#/edit/DVoR4RFa01SDSyK94PAOr

SukantPal:
@bigtimebuddy This is the JSFiddle that wasn't working in the last attempt to implement line join: https://jsfiddle.net/ShukantPal/3zkL6cmg/ (after modified it use this branch).

From here: https://github.com/pixijs/pixi.js/pull/5325

bigtimebuddy:
Wow, this is super promising. Could you make sure this is hooked up for canvas-graphics: https://jsfiddle.net/bigtimebuddy/h85a7f0b/

 Should be pretty easy because you're using the same cap/lineJoin names.

bigtimebuddy:
Also looks like some funny issues when closing the paths:
https://jsfiddle.net/bigtimebuddy/g75p8Lo4/

bigtimebuddy:
@SukantPal I made a better visual test for this:

https://jsfiddle.net/bigtimebuddy/h85a7f0b/

SukantPal:
@bigtimebuddy Looks like Travis CI is back. The JSFiddle should now be rendering correctly.

bigtimebuddy:
This is so amazing @SukantPal! Bravo :clap:

API additions looks fine to me. 

I'd like to create some more visual examples just to make sure we don't have any weird cases. I'll work on those.

bigtimebuddy:
@SukantPal it would be nice if the `miterLimit` values for webgl and context2d matched. Right now, they are very different. The default `miterLimit` for canvas is 10.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/miterLimit

bigtimebuddy:
Here's the little utility convert to pixi from that docs page:
You can toggle `forceCanvas` for comparison.
https://jsfiddle.net/bigtimebuddy/L6zp8hdb/

SukantPal:
@bigtimebuddy I fixed the miter limit. Looks like we have a bug with beveling real close.

https://jsfiddle.net/ShukantPal/kjtb2wLr/

bigtimebuddy:
Here's with some extreme line sizes: https://jsfiddle.net/bigtimebuddy/b2Lgzcp9/

Some overlapping issues with WebGL, but in general the shape looks good with context2d

SukantPal:
@bigtimebuddy Now, apart from internal overlapping (which isn't due to this specific PR), there shouldn't be any problems.

SukantPal:
@bigtimebuddy One min actually, maybe not.

bigtimebuddy:
Updated this demo to show both WebGL and context2D side-by-side.

https://jsfiddle.net/bigtimebuddy/L6zp8hdb/

packages/graphics/src/utils/buildLine.ts
```diff
@@ -94,129 +260,253 @@ function buildNonNativeLine(graphicsData: GraphicsData, graphicsGeometry: Graphi
...
+        y0 = verts[(i * 2) + 1];
+
+        x1 = verts[(i + 1) * 2];
+        y1 = verts[((i + 1) * 2) + 1];
+
+        x2 = verts[(i + 2) * 2];
+        y2 = verts[((i + 2) * 2) + 1];
+
+        /* Skip zero area triangles */
+        if ((x0 * (y1 - y2)) + (x1 * (y2 - y0)) + (x2 * (y0 - y1)) === 0)
```
ivanpopelyshev: Lets not do strict equality in float arithmetic. 

Need eps2 here. 
If we use `Math.abs(x1-x2)>eps`. and `eps=1e-3`, then here we have to use `eps2 = eps * eps`

we can put eps in our graphicsutils. @eXponenta @bigtimebuddy can you guys help me a bit here?

bigtimebuddy: Huh? What do you need help with exactly @ivanpopelyshev?

ivanpopelyshev: Is it fine to merge as it is and I'll adjust epsilon in separate PR?

I am sure that we need EPS there. Also we have to put that setting in GraphicsUtils

bigtimebuddy: If this is the only place, seems like an easy fix for @SukantPal. Otherwise, do it as a follow up PR later if you want to do something more extensive.

SukantPal: I added a `settings.EPSILON` property. You can use that if you want to commit directly; otherwise, I'll resolve this tomorrow.

ivanpopelyshev: ok, just remember that if for | x1 - x2 | you use eps, then for squares you have to use eps2. usually.

SukantPal:
@bigtimebuddy @ivanpopelyshev Ready?

bigtimebuddy:
YEAH! :tada:

PR #6503, opened 2020-03-27
Convert canvas-extract to TypeScript

Zyie:
##### Description of change

Nice and simple!


packages/canvas/canvas-extract/global.d.ts
```diff
@@ -0,0 +1,7 @@
+declare namespace GlobalMixins
+{
+    interface CanvasRenderer
+    {
+        extract: import('@pixi/canvas-extract').CanvasExtract;
```
bigtimebuddy: isn't this `plugins.extract`?

Zyie: yes, i didnt see that. Looks like i was too eager to use the new Global Mixins!

Zyie: fixed

bigtimebuddy:
I actually think your instincts were right about the GlobalMixins, however, I think we need them to define CanvasRender plugins and Renderer plugins.

Currently, this is defined pretty loosely here: https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/AbstractRenderer.ts#L36-L39

That means users don't get typings for doing things like `renderer.plugins.extract.canvas()`. 

We should do this as a follow-up fix for all the plugins.

packages/canvas/canvas-extract/src/CanvasExtract.ts
```diff
@@ -1,6 +1,9 @@
 import { RenderTexture } from '@pixi/core';
 import { CanvasRenderTarget } from '@pixi/utils';
 import { Rectangle } from '@pixi/math';
+import type { CanvasRenderer } from '@pixi/canvas-renderer';
+import type { DisplayObject } from '@pixi/display';
+import type { BaseRenderTexture } from '@pixi/core';
```
bigtimebuddy: Can you make sure these dependencies are including in package.json under dependencies?

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -13,6 +13,7 @@ import type {
     RenderTexture,
     BaseRenderTexture,
 } from '@pixi/core';
+import type { CanvasExtract } from '@pixi/canvas-extract';
```
bigtimebuddy: This is not correct. CanvasRenderer.plugins.extract. Also, this creates a circular dependency. CanvasRenderer should know nothing about canvas-extract.

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -44,6 +45,7 @@ export class CanvasRenderer extends AbstractRenderer
     public smoothProperty: string;
     public readonly blendModes: string[];
     public renderingToScreen: boolean;
+    public extract: CanvasExtract;
```
bigtimebuddy: Not needed.

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
eXponenta: It is inessential because `import type` used inside TS only, and striped in build process

bigtimebuddy: Yes, but the type imports will be used when we start generating types.

ivanpopelyshev:
`plugins` can be moved to GlobalMixins in separate PR , this PR is fine as it is. circular type deps? we have them anyway, we can fix stuff a bit later.

bigtimebuddy:
@Zyie can remove CanvasExtract from the CanvasRenderer? Otherwise this is good to go

bigtimebuddy:
I've deprecated CanvasRenderer#extract as part of this PR. This is an anti-pattern with the rest of the Renderer plugins, which hang off the `plugins` object.

PR #6505, opened 2020-03-28
[WIP] Experiment with antialias and stencil

ivanpopelyshev:
As I know, both antialias and stencil are optional in webgl. It might be that device wont be created with them.

Work in progress.

Its not draft because I want it to build.


bigtimebuddy:
Just FYI, all branches are built and deployed. You don’t need a PR to do that.

ivanpopelyshev:
OK, that's trash

PR #6508, opened 2020-03-29
Fix Texture leak with SpritesheetLoader

### LINKED ISSUE #6506: SpriteSheet.destroy(true) does not delete entries from TextureCache
Spritesheet.destroy(true) does not remove all items from the TextureCache.  To reproduce 

```typescript
const app = new Application({
        width: 640,
        height: 480,
        backgroundColor: 0x000000});

Loader.shared.add('assets/spritesheet-png8.json').load(setup);

function setup() {
   const sheet = Loader.shared.resources['assets/spritesheet-png8.json'].spritesheet;
   sheet.destroy(true);
   console.log(utils.TextureCache);
}
```

A potential workaround, for now, is to remove the elements from the cache directly.  

Use the resources from the following [repository](https://github.com/CodeAndWeb/pixijs-sprite-sheets/tree/master/images)

bigtimebuddy:
Fixes #6506

### DIFF
--- a/packages/spritesheet/src/SpritesheetLoader.ts
+++ b/packages/spritesheet/src/SpritesheetLoader.ts
@@ -57,8 +57,13 @@ export class SpritesheetLoader
                 return;
             }
 
+            const { baseTexture } = res.texture;
+
+            // Don't need this Texture anymore
+            res.texture.destroy();
+
             const spritesheet = new Spritesheet(
-                res.texture.baseTexture,
+                baseTexture,
                 resource.data,
                 resource.url
             );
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Is it about TextureCache?

bigtimebuddy:
No problem with TextureCache. The problem is we ignore (but don’t destroy) the Texture created by the TextureLoader—Spritesheet only uses the BaseTexture.

ivanpopelyshev:
Maybe make a test to check if texture remains in TextureCache after spritesheet destroy? I suggest to clean all texture cache before executing it

bigtimebuddy:
I already added a test check for this. You looking for something else?

ivanpopelyshev:
Oh , right. OK, its fine then

Issue #6478, opened 2020-03-16
Graphics' unexpected comportement while tinting with 2d context.


LituOffi:
### Expected Behavior
Only one graphics will be tinted (same as webgl context)
### Current Behavior
Every graphics are tinted
### Possible Solution
??
### Steps to Reproduce
tint some graphics from the same geometry
### Environment

- **`pixi.js` version**: 5.2.1 legacy and dev version
- **Browser & Version**:  chrome 80.0.3987.132
- **OS & Version**:  Windows 10.0.18362.720
- **Running Example**: [here](https://www.pixiplayground.com/#/edit/02gZJCxCpwYf9IqafPZs_)

ivanpopelyshev:
Confirmed. https://www.pixiplayground.com/#/edit/02gZJCxCpwYf9IqafPZs_

bigtimebuddy:
I don't see anything in that playground @ivanpopelyshev

LituOffi:
use my running example @bigtimebuddy

ivanpopelyshev:
@bigtimebuddy somehow `_` isnt treated as a part of url

LituOffi:
up :)

bigtimebuddy:
Yes, looks like there's an issue with tint canvas:
https://github.com/pixijs/pixi.js/blob/dev/packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.js#L329

Tint shouldn't use `geometry.dirty` since tinting is a Graphics-level property. @ivanpopelyshev this looks like an easy fix, right?

ivanpopelyshev:
It requires a bit of refactoring to store transformed tints in Graphics and not GraphicsGeometry. I can do it next week, there is just too much going on right now.

SukantPal:
@bigtimebuddy @ivanpopelyshev  You have to use `geometry.dirty` because `updateGraphicsTint` essentially multiplies the tint with each fill/stroke color.

`_fillTint` and `_lineTint` is essentially a "cache" that `Graphics` is putting into `GraphicsData`. If multiple graphics share that cache, then both of their tints will be updated. To prevent that:

* make `Graphics` store `_fillTint` and `_lineTint` in an object that is tagged by an unique id. Something like `glTexture`. So access like `graphicsData[I].cache[this.id]._fillTint`

* calculate the fill-tint and line-tint whenever you are rendering. It doesn't look like an expensive operation.

Pick your poison :)

[...]

PR #6516, opened 2020-04-02
Remove updateGraphicsTint in lieu of shared geometries, calculate fil…

SukantPal:
Fix for #6478.


SukantPal:
Should be working: https://www.pixiplayground.com/#/edit/02gZJCxCpwYf9IqafPZs_

ivanpopelyshev:
yep, performance should be the same. We can fix performance problems in next PR.

PR #6532, opened 2020-04-07
Fixes SimpleRope on Canvas renderer

themoonrat:
Was trying to set a property that only had a getter
Change the code to match the WebGL renderer code style, using _width instead
https://github.com/pixijs/pixi.js/issues/6531


bigtimebuddy:
TypeScript would have caught @themoonrat’s bug at build time. More of a reason to finish the conversion!

bigtimebuddy:
Thanks!

PR #6539, opened 2020-04-10
Handle extreme radius inputs for rounded rect

bigtimebuddy:
The issue with rendering rounded rectangle was there were some very strange results when the radius was larger than half the smallest size or negative. This should resolve these cases. The rendering mirrors how CSS behaves with `border-radius`.

**Bug (v5.2.1)**
https://jsfiddle.net/bigtimebuddy/hgje6tpy/

**Fixed**
https://jsfiddle.net/bigtimebuddy/at18sdop/


ivanpopelyshev:
I had version of that algo somewhere that also cared about different corner radii.

PR #6481, opened 2020-03-17
Convert prepare package to typescript

motan:
##### Pre-Merge Checklist

- [ ] Tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [ ] Lint process passed (`npm run lint`)
- [ ] Tests passed (`npm run test`)

PR #6540, opened 2020-04-10
Fix issue with preparing Graphics without shader

YuryKuvetski:
_resolveDirectShader is needed in a renderer parameter, otherwise it will fail [there](https://github.com/pixijs/pixi.js/blob/dev/packages/graphics/src/Graphics.ts#L1157)

##### Description of change
Just passing down renderer to the _resolveDirectShader.

##### Pre-Merge Checklist

### DIFF
--- a/packages/prepare/src/Prepare.ts
+++ b/packages/prepare/src/Prepare.ts
@@ -67,7 +67,7 @@ function uploadGraphics(renderer: AbstractRenderer | BasePrepare, item: IDisplay
     // if its not batchable - update vao for particular shader
     if (!geometry.batchable)
     {
-        (renderer as Renderer).geometry.bind(geometry, (item as any)._resolveDirectShader());
+        (renderer as Renderer).geometry.bind(geometry, (item as any)._resolveDirectShader((renderer as Renderer)));
     }
 
     return true;


ivanpopelyshev:
oops :)

packages/prepare/src/Prepare.ts
```diff
@@ -67,7 +67,7 @@ function uploadGraphics(renderer: AbstractRenderer | BasePrepare, item: IDisplay
     // if its not batchable - update vao for particular shader
     if (!geometry.batchable)
     {
-        (renderer as Renderer).geometry.bind(geometry, (item as any)._resolveDirectShader());
+        (renderer as Renderer).geometry.bind(geometry, (item as any)._resolveDirectShader((renderer as Renderer)));
```
bigtimebuddy, line 70: Why is `item` cast as any? We should have caught this issue way sooner.

YuryKuvetski: Some kind of legacy?

bigtimebuddy: I think was just an issue with with this package being converted to TypeScript #6481. We should followup with a PR, that create an Interface for batchable objects.

Issue #6543, opened 2020-04-11
Invalid line rendering in graphics


eXponenta:
Latest `cap&joints` #6502   improvements corrupt line rendering.

Target link:
https://pixijs.io/examples/?v=dev#/graphics/advanced.js

### Expected Behavior
Change pixi version to 5.2.1
### Current Behavior
Change pixi version to Latest dev

![image](https://user-images.githubusercontent.com/12572222/79038960-637c3380-7be6-11ea-97e0-1632d6ee4fc9.png)

### Environment

- **`pixi.js` version**: dev
- **Browser & Version**: all
- **Running Example**: 
https://pixijs.io/examples/?v=dev#/graphics/advanced.js

SukantPal:
These occur when the line join/cap features are used in curves?

bigtimebuddy:
Those features are not used in the examples. So this looks like a regression in normal line rendering.

bigtimebuddy:
Here's an isolated example @SukantPal https://jsfiddle.net/bigtimebuddy/12k7on59/

SukantPal:
@bigtimebuddy @eXponenta You can give feedback on #6545

bigtimebuddy:
Thanks for fixing this @SukantPal. Closing.

PR #6545, opened 2020-04-11
Fixed inner miter point adjustment when on opposite side of corner

SukantPal:
#6543

### DIFF
--- a/packages/graphics/src/utils/buildLine.ts
+++ b/packages/graphics/src/utils/buildLine.ts
@@ -359,12 +359,12 @@ function buildNonNativeLine(graphicsData: GraphicsData, graphicsGeometry: Graphi
         let omx = x1 - ((px - x1) * outerWeight);
         let omy = y1 - ((py - y1) * outerWeight);
 
-        const p0ix = x1 - (perpx * innerWeight);
-        const p0iy = y1 - (perpy * innerWeight);
+        const p0ix = x0 - (perpx * innerWeight);
+        const p0iy = y0 - (perpy * innerWeight);
         const p2ix = x2 - (perp1x * innerWeight);
         const p2iy = y2 - (perp1y * innerWeight);
-        const p0ox = x1 + (perpx * outerWeight);
-        const p0oy = y1 + (perpy * outerWeight);
+        const p0ox = x0 + (perpx * outerWeight);
+        const p0oy = y0 + (perpy * outerWeight);
         const p2ox = x2 + (perp1x * outerWeight);
         const p2oy = y2 + (perp1y * outerWeight);
         /* Check if inner miter point is on same side as p1 w.r.t vector p02 */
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Better, but seems like the square in the lower-left is busted now:

https://pixijs.io/examples/?v=fix-6543#/graphics/advanced.js
Screen Shot 2020-04-11 at 2 03 03 PM

https://pixijs.io/examples/?v=v5.2.1#/graphics/advanced.js
Screen Shot 2020-04-11 at 2 03 23 PM

SukantPal:
@bigtimebuddy Was the square busted before this PR too? Or is it because of this change?

bigtimebuddy:
Looks like because of this change.

bigtimebuddy:
@SukantPal any ideas?

SukantPal:
@bigtimebuddy I know where the problem is coming from. The math is incorrect. I’ll get this sorted tomorrow. It is because of the code where I prevent the inner miter point from being on the opposite side of the corner.

SukantPal:
@bigtimebuddy I also want to present a formal (mathematical) proof that my fix will work and why.

ivanpopelyshev:
Proofs are good things, but please do make refs , where those formulas exist in other renderers. 

Nothing bad in peeking at other people stuff to check for possible mistakes and places where you are actually smarter ;)

SukantPal:
@ivanpopelyshev I'll try.

SukantPal:
@bigtimebuddy Why are the tests failing in CI? Something related to `isMobile` & passing on my machine.

SukantPal:
Okay, this bug was related to the code that "prevented the inner miter point from being on the opposite side of the corner". By doing so, you eliminate a rare class of bugs that arise at high line widths. This animation will show you what happens when the width gets to large eventually: https://jsfiddle.net/ShukantPal/fhkd2r85/. Specifically, focus on the "inner" point of the joint. It rapidly rises up, as you would expect, however, it overshoots the first & last points, looking like this:

Screen Shot 2020-04-13 at 6 59 55 PM

Let me take an example in a Cartesian plane (no y-inversion): connect the points (0, 0), (10, 100), (20, 0) with a variable line width.

* At width = 10, you will find that the inner edges will intersect at (10, 0).

* Increasing the width from 10 will result the inner edges meeting at a y-value below 0. This means on the opposite side of the joint (the joint is (10, 100) on +ve y side).

I have setup a guide showing the math behind this here: https://www.desmos.com/calculator/gtyqjkoad6

SukantPal:
To prevent the inner miter joint from going onto the other side, we can take check for "same-sidedness" of the it and the joint (p1) w.r.t the line joining p0 & p2. This guide explains how that works:

https://www.desmos.com/calculator/zwdrci6rwh

Issue #6511, opened 2020-04-01
A bug by using ismobilejs because of the wrong api uesed


HypnosNova:
### Expected Behavior
no mistakes when compiling
### Current Behavior
![image](https://user-images.githubusercontent.com/7953802/78137334-b40dc700-7457-11ea-9a53-a1d587564a52.png)

### Possible Solution
![image](https://user-images.githubusercontent.com/7953802/78137413-dbfd2a80-7457-11ea-8025-4dd50dc8d387.png)

### Steps to Reproduce
STEP 1: npm i pixi.js and then using 
import * as PIXI from './pixi.js'; 
And just write a hello world demo.
STEP 2: compiler the project and failed

### Environment

- **`pixi.js` version**: _e.g. newest
- **Browser & Version**: _e.g. any bloswers
- **OS & Version**: _e.g. any os
- **Running Example**:

bigtimebuddy:
Please provide a repo the reproduces the issue. Your step are insufficient and the tools you’re using to compile probably matter.

HypnosNova:
Using PIXI with react.
When compiling the whole project and run it on browser, we could see following error:
![image](https://user-images.githubusercontent.com/7953802/78205892-ef4bdc80-74cf-11ea-9d1b-4d8313335523.png)
I add a console to see what ismobilecall is.
And in browser console, we could see ismobilecall is an object that contains several APIs,
![image](https://user-images.githubusercontent.com/7953802/78205995-3639d200-74d0-11ea-86a3-2b7f84370d74.png)
That means after the project compiled, ismobilejs export default an Object instead of a function.

But strangely,  I create a new project and use rollup to compile, nothing went wrong, only in react project it went wrong.
So is it a bug in react tools? Or do you know any solutions to solve it

notorigine:
I believe it is an issue with v1 of ismobilejs - see [issue #115](https://github.com/kaimallea/isMobile/issues/115). 

Using a complete different stack than React (Gulp, ES6), it is also causing an error when running my build process, because ismobilejs isn't transpiled when imported as a dependancy by @pixi/settings (see screenshots).

![Capture d’écran 2020-04-09 à 00 59 45](https://user-images.githubusercontent.com/4939395/78841629-29e8d400-79fe-11ea-863b-cb6f43e58fed.jpg)
[...]

PR #6546, opened 2020-04-12
[RFC] Github Actions

cursedcoder:
So I haven't figured out how travis conditions worked, b/c for me they look similar.

Any suggestions would be highly appreciated.

~This incorporates a bump for jsdoc, so Pixi can properly build with node version > 10.~


bigtimebuddy:
Wow great work! Thanks for doing this and putting in the effort. This has been high on my todo list. Directionally, removing Travis is the correct move.
 
The JSDoc bump is a consequential change. The new version of JSDoc misses a bunch of our documentation like events, and some deprecation. 3.6.x Intoduced a new parsing engine which screwed everything up. My preference would be to pin GitHub actions to node 10 until we can fix all those underlying issues as another PR. Because our typings also depend on JSDoc, this is a nontrivial change and requires a close attention.

Is that ismobilejs change a fix for #6511? Maybe we can do that as a separate PR?

.github/workflows/s3-deploy.yml
```diff
@@ -0,0 +1,54 @@
+name: Deploy to S3
+
+on:
+  push:
+    branches: [ dev ]
+  pull_request:
+    branches: [ dev ]
+  release:
+    types: [ published, created, edited ]
```
englercj, line 9: Probably only want published here, otherwise this will upload anytime we edit the description of the release.

.github/workflows/s3-deploy.yml
```diff
@@ -0,0 +1,54 @@
+name: Deploy to S3
+
+on:
+  push:
+    branches: [ dev ]
+  pull_request:
+    branches: [ dev ]
```
englercj, line 7: Do we have PRs upload to S3 currently? I think we only uploaded pushes to branches on the repo. Doing PRs might be a little dangerous.

.github/workflows/s3-deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - uses: actions/checkout@master
+
+      - uses: jakejarvis/s3-sync-action@master
+        name: Deploy non-tag branches
+        if: github.event_name == 'pull_request'
+        with:
+          args: --acl public-read --follow-symlinks --delete --cache-control "max-age=60"
+        env:
+          AWS_S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
```
englercj, line 23: The bucket isn't secret, it is `pixi.js`

.github/workflows/s3-deploy.yml
```diff
@@ -0,0 +1,54 @@
+name: Deploy to S3
+
+on:
+  push:
+    branches: [ dev ]
```
englercj, line 5: This should be all branches.

.github/workflows/s3-deploy.yml
```diff
@@ -0,0 +1,54 @@
...
+          AWS_S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: 'eu-west-1'
+          SOURCE_DIR: 'dist'
+          DEST_DIR: ${{ github.ref }}
+
+      - uses: jakejarvis/s3-sync-action@master
+        name: Deploy latest release
+        if: github.event_name == 'push'
```
englercj, line 45: This block should also be the release condition, it just uploads to the `release` folder instead of a `<branchname>` folder.

.github/workflows/s3-deploy.yml
```diff
@@ -0,0 +1,54 @@
+name: Deploy to S3
```
englercj, line 1: I don't know if having this as a separate workflow will work. I think as a separate workflow, it will be run separately so the `dist` folder you wanted to uploader won't exist. May have to either rebuild the library in steps, then upload. Or move these steps into the previous workflow and guard them with conditions.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,31 @@
...
+      matrix:
+        node-version: [10.x, 12.x]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm install codecov -g
```
englercj, line 25: Shouldn't be necessary as long as we have this listed in our devDeps

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,31 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ dev ]
```
englercj, line 8: I think this should be all branches.

.github/workflows/s3-deploy.yml
bigtimebuddy: Yes, doing uploads for PRs is dangerous. Should only be for pushes.

cursedcoder:
@bigtimebuddy thanks, will handle mentioned issues.

Regarding ismobilejs – I'm not sure how it worked on travis, probably because of cached node_modules.
On my forked repo travis it was failing with "missing argument", see https://travis-ci.org/github/cursedcoder/pixi.js/builds/673997141 and https://github.com/cursedcoder/pixi.js/runs/580341097?check_suite_focus=true#step:5:314.

```##[error]packages/settings/src/utils/isMobile.ts(6,18): error TS2554: Expected 1 arguments, but got 0.```

@englercj thanks for the review, I'll ensure it also deploys correctly on my fork.

File: .github/workflows/nodejs.yml (hunk 7 above)
Comment (cursedcoder) [reply .github/workflows/nodejs.yml:25]: I think we don't have this in our package.json b/c it's needed to upload coverage to codecov service.

File: .github/workflows/s3-deploy.yml (hunk 3 above)
Comment (cursedcoder) [reply .github/workflows/s3-deploy.yml:23]: @englercj I know that buckets are unique per AWS account, therefore it's not trivial to test it on your fork.

Comment (cursedcoder) [reply .github/workflows/s3-deploy.yml:23]:
```
AWS_S3_BUCKET: ${{ secrets.AWS_S3_BUCKET || 'pixi.js' }}
```
should be better

.github/workflows/s3-deploy.yml
cursedcoder: Mmmh, ok I finally figured this out.
So this another deploy always overwrites `/release` with the latest release.
Will do.

.github/workflows/s3-deploy.yml
englercj: Ah you're using the secrets as a way to parameterize the bucket. That's probably fine.

bigtimebuddy:
Could you update to the latest, we fixed the ismobilejs issue.

cursedcoder:
@bigtimebuddy done.
@englercj 
image

You can see – tags, branches and the latest release are deployed.

englercj:
My final request is that I'd like to combine the two workflows together and instead just guard the deploy steps inline.

I'd rather do this becuase right now they both are building the library anyway. Right now the flow is: install, build, test, if release then install, build, deploy. I'd rather the flow be: install, build, test, optionally deploy.

Here is an example of what I did on gl-tiled:

https://github.com/englercj/gl-tiled/blob/1c0d44e5a02f3ad567bc03ce5e34742f4b435258/.github/workflows/main.yml#L25-L33

cursedcoder:
@englercj thanks, sounds about right.

However, in addition to build-matrix, this will lead to several deploys per build.

I'll dive deeper into the docs, I have few ideas on how to improve that.

englercj:
Currently the matrix only has one entry, so there is only one build.

But you can just add another condition to the `if` to only deploy on a specific matrix entry.

bigtimebuddy:
Also @cursedcoder can you revert changes to package-lock.json? Those changes are unnecessary for this PR and is probably left-overs from bumping jsdoc.

cursedcoder:
@bigtimebuddy @englercj should be OK now.

Please check.

cursedcoder:
@bigtimebuddy make sure to update the repository's secrets before it gets merged.

Thanks.

bigtimebuddy:
@cursedcoder is there anything else I need to setup in the settings other than secrets?

bigtimebuddy:
Also could you fix the build status badge in the readme to not use Travis?

cursedcoder:
@bigtimebuddy added the badge.

Checklist before merging:
- [ ] Add AWS secrets
- [ ] Add codecov token secret

Thinking out loud – can't we instead merge this into `github-actions` and ensure it builds in **this** repository?

bigtimebuddy:
Yes, good idea.

PR #6550, opened 2020-04-13
Enhance event types for EventEmitter

guansss:
Strict type checking for the arguments of events.

#### Enhanced EventEmitter

eventemitter3 introduced [advanced argument types for events](https://github.com/primus/eventemitter3/pull/181) in 4.0.0, but doesn't work well with inheritance, so I managed to tweak it for that purpose.

##### Usage

```typescript
interface ParentEvents extends PIXI.utils.BaseEventTypes {
    parent: [string, number]
}

interface ChildEvents extends ParentEvents {
    child: [number]
}

class Parent extends PIXI.utils.EventEmitter<ParentEvents> { // pass event types to EventEmitter
    method1() {
        this.emit('parent', 'foo', 1); // OK

        this.emit('parent', 'foo', 1, 2); // Error, too many arguments

        this.emit('parent'); // Error, missing arguments

        this.on('parent', this.listener); // OK

        this.on('parent', this.listener1); // Error, missing arg1 in function
    }

    listener(arg0: string, arg1: number) {}

    listener1(arg0: string) {}
}

class Child extends Parent {
    // in subclass, event types should be overridden via property declaration
    declare __events: ChildEvents;

    method2() {
        this.emit('parent', 'foo', 1); // OK

        this.emit('child', 1); // OK

        this.emit('child', 'foo'); // Error, type mismatch
    }
}

// another way to override event types is interface merging
// interface Child {
//     __events: ChildEvents;
// }
```

[...]

PR #6551, opened 2020-04-13
Fix ismobilejs breakage

bigtimebuddy:
This was bumping up in a few places namely: #6550 #6546 #6511

Since the release of ismobilejs@1.1.0, there's no longer a default user-agent/navigator argument.  This was breaking for our users trying to build against `ismobilejs@^1.0.3`

### DIFF
--- a/packages/settings/package.json
+++ b/packages/settings/package.json
@@ -24,6 +24,6 @@
     "dist"
   ],
   "dependencies": {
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Example:
https://jsfiddle.net/bigtimebuddy/8evch59z/

PR #6559, opened 2020-04-15
Fixed example code for RenderTexture.create()

TheOriginalArkless:
I ran into this when looking at the documentation.
Skimming the readme I failed to see guidlines for comment/doc only changes.
I hope this is enough.


bigtimebuddy:
Thank you @TheOriginalArkless!

Issue #6556, opened 2020-04-14
Framebuffer does not release depth buffer when destroyed, causing GPU memory leak


Vanilagy:
I noticed that if you (manually) add a depth texture to a Framebuffer, that texture does NOT get cleaned up by calling destroy() on the framebuffer, although arguably, it should, since it belongs to the buffer.

### Expected Behavior
The depth texture gets destroyed alongside all the color textures.
### Current Behavior
The color textures get destroyed while the depth texture doesn't, causing a memory leak.
### Possible Solution
Add destruction of the depth texture when destroying the framebuffer.
### Steps to Reproduce
I created a commented Fiddle that can help with reproducing the problem:
https://jsfiddle.net/ncj95301/

Click "Start", then check your VRAM usage, it should go up quite quickly.
### Environment

- **`pixi.js` version**: _5.2.1_
- **Browser & Version**: _Chrome 80.0.3987.163_
- **OS & Version**: _Windows 10.0.18362 Build 18362_
- **Running Example**: _https://jsfiddle.net/ncj95301/_

ivanpopelyshev:
Ha! First user of DepthTexture that I see ! I really dont know the demo it was used before.. OK then, time to find that bug.

Yes, it blongs to the framebuffer: https://github.com/pixijs/pixi.js/blob/dev/packages/core/src/framebuffer/FramebufferSystem.ts#L517 . Yes, as you said, in webgl its separate, but in PixiJS it appears only with framebuffer, its fine to dispose it there.

Vanilagy:
Haha yeah man, people who DepthTexture might be rare, but we exist! I'm actually surprised they're rarely used, since depth buffers have always been an important part of computer graphics.

SukantPal:
@bigtimebuddy This is actually a structural problem with Pixi. The depth-texture is not owned by the framebuffer; the color textures aren't freed either. If you call `RenderTexture#destroy`, it destroys itself (which happens to be the only color texture) and disposes the framebuffer.

SukantPal:
@bigtimebuddy I suggest that `RenderTexture#destroy` actually does the work of destroying the depth texture. See my proposed fix: https://github.com/pixijs/pixi.js/pull/6561

SukantPal:
@Vanilagy Please verify the fix: https://jsfiddle.net/ShukantPal/6jqsfa7v/

Vanilagy:
Verified, the leak is gone.

PR #6561, opened 2020-04-15
RenderTexture#destroy destroys the depth texture as well.

SukantPal:
#6556

### DIFF
--- a/packages/core/src/framebuffer/Framebuffer.ts
+++ b/packages/core/src/framebuffer/Framebuffer.ts
@@ -197,10 +197,25 @@ export class Framebuffer
     }
 
     /**
-     * disposes WebGL resources that are connected to this geometry
+     * Disposes WebGL resources that are connected to this geometry
      */
     dispose(): void
     {
         this.disposeRunner.emit(this, false);
+    }
+
+    /**
+     * Destroys and removes the depth texture added to this framebuffer.
+     */
+    destroyDepthTexture(): void
+    {
+        if (this.depthTexture)
+        {
+            this.depthTexture.destroy();
+            this.depthTexture = null;
+
+            ++this.dirtyId;
+            ++this.dirtyFormat;
+        }
     }
 }
--- a/packages/core/src/framebuffer/FramebufferSystem.ts
+++ b/packages/core/src/framebuffer/FramebufferSystem.ts
@@ -538,6 +538,7 @@ export class FramebufferSystem extends System
         if (!contextLost)
         {
             gl.deleteFramebuffer(fbo.framebuffer);
+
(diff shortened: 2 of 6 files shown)


ivanpopelyshev:
Maybe it would be better to move it to `dispose()`. Technically its the same thing for Framebuffer,  because its never recovering.. hm...

SukantPal:
@ivanpopelyshev Dispose just removes the framebuffer from the GPU. Destroy will destroy the data itself.

We shouldn't move this to dispose. `Framebuffer#dispose` disposes the depth texture already by disposing the framebuffer-object, which holds the depth texture on the GPU.

ivanpopelyshev:
> We shouldn't move this to dispose. `Framebuffer#dispose` disposes the depth texture already by disposing the framebuffer-object, which holds the depth texture on the GPU.

Nope, that's the point - depthTexture is not disposed on the GPU automatically. Framebuffer disposes itself and renderbuffer if mounted, but not depthTexture.

```
            gl.deleteFramebuffer(fbo.framebuffer);
            if (fbo.stencil)
            {
                gl.deleteRenderbuffer(fbo.stencil);
            }
```

Oh, I feel that we are missing MSAA case here: `fboFramebuffer` should be also disposed automatically :(

SukantPal:
@ivanpopelyshev Correct, because the depth texture _was not added internally by the framebuffer system_. This is way you cannot run `gl.deleteTexture` on the depth-texture - this is equivalent of destroying (not disposing) the depth texture: https://github.com/pixijs/pixi.js/blob/9cdb526a9bfbc4643d36a106be3f2b16981e516d/packages/core/src/textures/TextureSystem.ts#L353

SukantPal:
The user _might not want to destroy the depth texture and may reuse it after destroying the framebuffer_. That is not the case for the stencil renderbuffer, because it is never manually added to the framebuffer.

ivanpopelyshev:
Agree. But then it might be user problem to destroy the depth texture when he destroys renderTexture and we dont need this PR. 

I don't think it matters, we can change it later, lets merge this one.

SukantPal:
@ivanpopelyshev Agreed, it is all semantics. Putting it in render-texture makes more sense for now, as it is a higher-level API and users might face problems like with 6556.

The user can still "snatch" the depth texture before it gets destroyed by listening to "dispose".

SukantPal:
@bigtimebuddy Done!

bigtimebuddy:
Did you forget to commit RenderTexture? `Error: Cannot find module './RenderTexture'`

SukantPal:
@bigtimebuddy Yup, I forgot a `git add .`

Issue #6535, opened 2020-04-07
drawEllipse and drawCircle broken with setMatrix


georeith:
### Expected Behavior

`drawEllipse` and `drawCircle` can be transformed with `setMatrix`

### Current Behavior

`drawEllipse` draws a cone to its origin when `setMatrix` is called before it and causes a translation of the origin.

![image](https://user-images.githubusercontent.com/3090066/78723536-81fed800-7923-11ea-83b2-ecc7ef6daa7c.png)

### Steps to Reproduce

`drawEllipse` in `PIXI.Graphics` after `setMatrix`.

![image](https://user-images.githubusercontent.com/3090066/78721299-2c283100-791f-11ea-9bee-4651b331719b.png)

https://pixiplayground.com/#/edit/IHholQNOTLxizgn8ucSc9

### Environment

- **`pixi.js` version**:  5.2.1
- **Browser & Version**: Chrome 80
- **OS & Version**: Mac OS Mojave
- **Running Example**: https://pixiplayground.com/#/edit/IHholQNOTLxizgn8ucSc9

PR #6562, opened 2020-04-15
Added center to list of "points" in circle-shape so it can be transfo…

SukantPal:
#6535


bigtimebuddy:
Could you provide a verification playground or fiddle for this? Also, looks like there's a test failing `should have no gaps in line border`, the test might need to be revised a little. Basically that test was added because our previous version had a gap in the stroke.

SukantPal:
@bigtimebuddy You commented just when I pushed a fix. This should be ready in a few minutes.

PR #6565, opened 2020-04-15
Removed center from points

SukantPal:
#6562 made a regression in the stroke. The "points" array is used to do a stroke, and adding the center there add a line from the center to circumference. This reverts that and transforms the center in `triangulate` instead.

### DIFF
--- a/packages/graphics/src/GraphicsData.ts
+++ b/packages/graphics/src/GraphicsData.ts
@@ -17,6 +17,7 @@ export class GraphicsData
     type: SHAPES;
     points: number[];
     holes: Array<GraphicsData>;
+
     /**
      *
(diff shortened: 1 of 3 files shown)


SukantPal:
https://jsfiddle.net/ShukantPal/u6p5rvs7/3/

SukantPal:
https://jsfiddle.net/ShukantPal/uq5f4k8o/ is this update.

packages/graphics/src/GraphicsData.ts
```diff
@@ -17,6 +17,7 @@ export class GraphicsData
     type: SHAPES;
     points: number[];
     holes: Array<GraphicsData>;
+
```
bigtimebuddy, line 20: Can you remove this white space?

bigtimebuddy:
Thank you!

Issue #6563, opened 2020-04-15
[BUG]: Graphics.beginTextureFill({ texture: PIXI.Texture.EMPTY }) makes lineStyle disappear


davidye:
### Expected Behavior

Line style still shows up when you call beginTextureFill({ texture: PIXI.Texure.EMPTY })

### Current Behavior

Line style disappears when called with beginTextureFill({ texture: PIXI.Texure.EMPTY })

### Possible Solution

Seems like an edge case

### Steps to Reproduce

Modify demo here: https://pixijs.io/examples/#/graphics/advanced.js

```
// beatifulRect.beginFill(0xFF0000);
beatifulRect.beginTextureFill({ texture: PIXI.Texture.EMPTY});
```

### Environment

- **`pixi.js` version**: 5.2.1
- **Browser & Version**: Chrome 80.0.3987.163
- **OS & Version**: macOS Catalina 10.15.4 (19E266)

bigtimebuddy:
I agree that's weird.

bigtimebuddy:
@davidye After some discussion on the PR #6568, we have decided that this is designed as intended and we won't fix. It is the correct behavior that when an invalid texture is assigned to either the fill or the stroke the whole graphic object (both fill and stroke) will be hidden. Please do not use `Texture.EMPTY` as a texture. If you want some invisible Texture, consider making a 1x1 transparent canvas texture.

If someone would like to make a note in the documentation describing this behavior, PRs welcome.

PR #6566, opened 2020-04-15
Fixed 6563

SukantPal:
#6563 

`Texture.EMPTY` is never "valid" because its base's width & height is zero.

### DIFF
--- a/packages/graphics/src/GraphicsGeometry.ts
+++ b/packages/graphics/src/GraphicsGeometry.ts
@@ -630,8 +630,8 @@ export class GraphicsGeometry extends BatchGeometry
             const fill = data.fillStyle;
             const line = data.lineStyle;
 
-            if (fill && !fill.texture.baseTexture.valid) return false;
-            if (line && !line.texture.baseTexture.valid) return false;
+            if (fill && !fill.texture.baseTexture.valid && fill.texture !== Texture.EMPTY) return false;
+            if (line && !line.texture.baseTexture.valid && line.texture !== Texture.EMPTY) return false;
         }
 
         return true;
--- a/packages/graphics/test/index.js
+++ b/packages/graphics/test/index.js
@@ -1,5 +1,5 @@
 // const MockPointer = require('../interaction/MockPointer');
-const { Renderer, BatchRenderer, Texture } = require('@pixi/core');
+const { Renderer, BatchRenderer, Texture, BaseTexture } = require('@pixi/core');
 const { Graphics, GRAPHICS_CURVES, FillStyle, LineStyle, graphicsUtils } = require('../');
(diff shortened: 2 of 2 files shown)


davidye:
> Texture.EMPTY is never "valid" because its width & height are zero.

This comment may be completely off the mark as I've just started using PIXI and digging into the codebase, but it looks like:

```
console.log(PIXI.Texture.EMPTY.width, PIXI.Texture.EMPTY.height)
> 1 1
```

SukantPal:
@davidye Correct. However, `Texture.EMPTY.valid` is `false` unless you call `update()` on that. This is because the initial width/height is zero.

However, I am not sure when the width/height becomes 1. Thanks for letting me know.

SukantPal:
@davidye I corrected the statement. If you try: `PIXI.Texture.EMPTY.baseTexture.width`, then it should definitely be zero.

SukantPal:
@bigtimebuddy We have a test specifically that forbids this behavior, i.e. you have two choices - this PR or that test. https://github.com/pixijs/pixi.js/blob/9fc2555f9d21380850d6b5f2869fde8f9b1f1f7d/packages/graphics/test/index.js#L807

davidye:
Ah I see! Would it be better to check:

```texture.baseTexture.width !== 0 && texture.baseTexture.height !== 0```

Instead of special casing EMPTY? Or is EMPTY the only texture that satisfies that condition.

**Comment by SukantPal** [comment]
We are already doing that check & `EMPTY` violates that check, preventing rendering. `baseTexture.valid` means that the texture has a non-zero width & height. This check is used to prevent rendering when the texture hasn't loaded.

However, `EMPTY` texture _is loaded and has a zero width & height._ This must be special cased.

**Comment by SukantPal** [comment]
@bigtimebuddy This one should pass now. Ready for review.

**Comment by SukantPal** [comment]
@ivanpopelyshev What changes are you requesting?

**Comment by ivanpopelyshev** [comment]
I dont understand why do we have to support Texture.EMPTY there. User can just specify "alpha:0" fo lineStyle

**Comment by SukantPal** [comment]
@ivanpopelyshev Because @bigtimebuddy requested a fix. If you don't want this fix, you should explicitly _say_ that. Don't just block the fix ambiguously.

**Comment by bigtimebuddy** [comment]
I don't think this is the correct fix, GraphicsGeometry should not know about EMPTY. This also produces the bug:
```
beginTextureFill({ texture: new PIXI.Texture(new PIXI.BaseTexture())}
```
The problem is that invalid fill textures should only hide the fill. I would assume the inverse is also the same for line textures hiding the fill.

SukantPal:
@bigtimebuddy I see what you want with this.

PR #6568, opened 2020-04-15
Allow Graphics rendering even when not all textures are loaded; update again when textures load.

SukantPal:
The previous `validateBatching` would return false only when we didn't need to update the batching; however, it also didn't let the `Graphics` render until all textures had loaded (became `valid`).

I've now changed this to:

* Allow rendering when not all textures have loaded.

* Update when a texture has loaded (when `_validTextureCount` changes).

#6563

### DIFF
--- a/packages/graphics/src/GraphicsGeometry.ts
+++ b/packages/graphics/src/GraphicsGeometry.ts
@@ -80,6 +80,7 @@ export class GraphicsGeometry extends BatchGeometry
     protected shapeIndex: number;
     protected _bounds: Bounds;
     protected boundsDirty: number;
+    protected _validTextureCount: number;
 
     constructor()
     {
@@ -214,6 +215,14 @@ export class GraphicsGeometry extends BatchGeometry
          * @default 0
          */
         this.boundsPadding = 0;
+
+        /**
+         * Number of loaded textures on last update.
+         *
+         * @member {number}
+         * @protected
+         */
+        this._validTextureCount = 0;
 
         this.batchable = false;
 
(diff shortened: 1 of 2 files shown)


SukantPal:
This doesn't work yet.

SukantPal:
@ivanpopelyshev I don't know what to do. If the texture loads to slow (after the first `render()`), then it is never loaded. A solid color takes it place: https://jsfiddle.net/ShukantPal/qm309boe/.

Screen Shot 2020-04-15 at 9 57 16 PM

ivanpopelyshev:
Drat.. yeah, need to subscribe to single texture update like Sprite does. 

If only we could store ALL textures that are referenced in GraphicsGeometry...

SukantPal:
@ivanpopelyshev The texture system doesn't bind a texture until it is valid. Since it never gets uploaded, I don't see why there should be a problem. IDK, will have to think thru this.

bigtimebuddy:
I thought this would be an easy fix, but it's turning to a much larger task. I'm okay saying that we won't fix this issue if the solution is too complex. Maybe simply adding a note in the documentation to let users know if either line or fill texture is invalid, neither will show up. This includes using `Texture.EMPTY`. So maybe we are okay marking this as "working as designed"?

ivanpopelyshev:
Well, we can add one-liner `if (texture === EMPTY) throw new Error("WTH are you doing?");`

bigtimebuddy:
Thanks for your work here and feedback. Closing.

Issue #6571, opened 2020-04-16
JSDoc 3.5.5 Breaking Builds on Node 12


bigtimebuddy:
## Overview 

We have a known problem that **pixi.js** does not build on Node 12+. This is creating issues for us moving forward and more and more users are stubbing their toe on this issue as Node 12 has become the default LTS. Asking people to use nvm to use Node 10, isn't a great solution.

### Problem

The underlying issue is a dependency `requizzle@0.2.1`, this is a child-dependency only of `jsdoc@3.5.5`. The newer version of JSDoc 3.6.x have fixed this issue however upgrading from 3.5.5 to 3.6.x causes issues for us where documentation is missing (e.g., https://github.com/jsdoc/jsdoc/issues/1425) because JSDoc upgrading their parsing engine. I was hoping we'd get through TypeScript and we could move swiftly to [TypeDoc](https://typedoc.org/) instead of JSDoc, but that probably isn't going to happen soon.

I tried to upgrade to requizzle in the package-lock as a temporary fix, but that didn't work. There were some issues with typings generation and other problems (these might be because of Node 12). I gave up after about an hour trying to trace down bugs.

### Potential Fix

In order to upgrade to JSDoc 3.6, someone needs to build the docs and manually check that we aren't dropping documentation. I think we've put this task off long enough and we should try to address. I'm going to attach a bounty to this in hopes that someone would be willing to step up and help. It's a some tooling work and lots of manually checking and updating JSDoc blocks.

Upgrading to the newest JSDoc _could_ also impact our typings, since these are generated using [tsd-jsdoc](https://github.com/englercj/tsd-jsdoc). This typing generation is temporary solution, but we need to make sure this isn't impacted either.

### Related 

Related bugs or mentioned in these issues:

* #5724
* #6247
* #6546

_Note on Bounties: We award users who are first-to-merge and we consider PRs in the order in which they are submitted. Bounties are paid through OpenCollective._

SukantPal:
https://github.com/pixijs/pixi.js/pull/6572

bigtimebuddy:
Great work @SukantPal!

PR #6572, opened 2020-04-16
Upgrade JSDoc to 3.6.4

SukantPal:
#6571 

* Upgrade JSDoc to 3.6.4

* Use `moduleResolution: 'node'` for generating documentation.

* Change to relative path for JSDoc es6 fix plugin.

---
Looks like the JSON got formatted.


SukantPal:
Screen Shot 2020-04-16 at 2 40 17 PM

bigtimebuddy:
Yes, the `~` is confusing, is usually used in docs to denote internal-only. I wonder if the `~` is something wrong with our theme: https://github.com/pixijs/pixi-jsdoc-template. If you could look into that, that'd be awesome.

SukantPal:
@bigtimebuddy Okay, so after debugging this for some time, I've found that it is not due to our template. However, our template can fix this:

SukantPal:
Screen Shot 2020-04-16 at 4 15 37 PM

bigtimebuddy:
Could you also bump Node in .travis.yml to "12" to make sure we can actually test the build on 12?

SukantPal:
@bigtimebuddy For some strange reason, the docs https://pixijs.download/dev-upgrade-jsdoc/docs/PIXI.groupD8_.html are not showing the "~" after building in the cloud. They are generated by JSDoc 3.6.4 however.

ivanpopelyshev:
It would be good to cherry-pick it for master and v4.x if possible

bigtimebuddy:
Give me a bit to go through and verify nothing is funny in the docs.

SukantPal:
@bigtimebuddy Updates?

themoonrat:
I'll take a dive too

SukantPal:
@themoonrat The ArrayResource.bind method has a “@member” tag in its documentation. That is why, probably, it is showing up as a member now.

bigtimebuddy:
The reason we haven’t done this task before is because the issues are very subtle but significant. As @themoonrat has pointed out. And we have a lot of documentation. This requires a level of thoroughness.

bigtimebuddy:
@SukantPal were you able to dig into any of those issue @themoonrat mentioned?

SukantPal:
@bigtimebuddy No yet, apart from the "member" tag on `ArrayResource.bind` I described before.

bigtimebuddy:
With regard to the `~`, I think it's worth digging in to that a little more rather than override in the template. Do you have any idea why this is happening? https://jsdoc.app/about-namepaths.html

Usually the `~` in the namepath means that it's inner. That shouldn't be the case here. These should be all valid class on static namespaces.

bigtimebuddy:
I can live with a lot of the other subtleties that @themoonrat pointed out, however, 1 and 3 seem like the most important to understand what we are doing wrong there and if there's an underlying or systemic issue. Overall, I would like to merge this ASAP. We can continue to patch small fixes. Getting Node 12+ is better than getting this 100% correct.

SukantPal:
@bigtimebuddy 

* For no. 1, we can fix this by "hacking" our template as I pointed out before: https://github.com/pixijs/pixi-jsdoc-template/blob/dfa61e47e3d99f4e1458b6cc235b62ac34585914/tmpl/container.tmpl#L16. The "ancestor" string has the '~' and we can `replace('~', '.')` that.

TBH, I haven't been able to track this down inside JSDoc.

* For no. 3, I fixed this in the latest commit. The `ArrayResource#(un)bind` documentation had a `@member` tag, making it a member rather than a method.

bigtimebuddy:
By 3, I meant why are some thing being marked as overrides that shouldn't.

Also, for 1, I'd rather not hack the template if possible. Can you try to drill into JSDoc and figure out why it's marking our global namespace members as inner?

SukantPal:
@themoonrat I would be happy to fix the ~ issue for symbol separators in our template. The others are serious issues in JSDoc and that’s why will be moving in the future away from it.

SukantPal:
@bigtimebuddy I don’t know this would be applicable, but in the future before you move to TSDoc, you might to consider this https://github.com/SukantPal/webdoc

I’ve been working on a MUCH better documentation generator that will support JavaScript and Typescript. Other goals for it are:

1. Linking to external documentation. So pixijs plugin documentation can be updated to have links to core PIXI types.

2. More flexibility to fix issues or add features since it is a in house project.

Idk, I’m not saying this is the way to go - but something to consider given that I’m able to complete it before you guys decide to move to tsdoc.

bigtimebuddy:
I’m not a fan of changing the template, since there is actually a valid use for ~ and that seems like an unnecessary hack without understating the problem. I’m more interested in figuring out why JSDoc considers everything inner scoped. Maybe it has something to do with our es6 fix? I think raising a new issue for this is fine though.

bigtimebuddy:
@SukantPal could you merge with dev to get Github Actions workflow instead of Travis?

SukantPal:
#6700

PR #6580, opened 2020-04-22
Hotfix for Spritesheet memory leak PR #6508

ivanpopelyshev:
Thanks to #6508 pixi-tilemap `basic.html` example broke. Other people could use `atlasResource.texture` it too, so this is now a breaking change. Now imagine what could happen if they will get **destroyed texture**!

You guys can push changes to this branch, I honestly don't care what you decide about it. At the least please leave `texture = null` here. Please cherry-pick it for `5.2.2`.

### DIFF
--- a/packages/spritesheet/src/Spritesheet.ts
+++ b/packages/spritesheet/src/Spritesheet.ts
@@ -345,10 +345,7 @@ export class Spritesheet
         this.textures = null;
         if (destroyBase)
         {
-            if (this._texture)
-            {
-                this._texture.destroy();
-            }
+            this._texture?.destroy();
             this.baseTexture.destroy();
         }
         this._texture = null;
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Good catch.

There's no reason to deprecate or break this is we don't need to. I changed to have the Spritesheet optionally manage the Texture from the Loader's Resource. Accomplishes the same thing without any breaks.

ivanpopelyshev:
Its about SpritesheetLoader `resource.texture.destroy()` call, not about spritesheet

bigtimebuddy:
Yeah I know. I'm just moving where texture.destroy happens. Accomplishes the same thing.

ivanpopelyshev:
Its here: https://github.com/pixijs/pixi.js/blob/25db781bef261a9f509882ba730c095a6068089d/packages/spritesheet/src/SpritesheetLoader.ts#L63

bigtimebuddy:
That line was added because calling `spritesheet.destroy(true)`, did not remove the Texture created by the TextureLoader.

ivanpopelyshev:
Yes, I Approve!

PR #6581, opened 2020-04-22
Support Github Actions, remove Travis (#6546)

bigtimebuddy:
Follow up PR to #6546 to merge GitHub Actions into dev.

### Verification

Branch Commit:
https://github.com/pixijs/pixi.js/runs/609013961

Deploy:
https://pixijs.download/dev-github-actions/docs/index.html


bigtimebuddy:
I added `CODECOV_TOKEN` to repository secrets. But we probably need to pass that to `codecov --token=secrets.CODECOV_TOKEN`, right @cursedcoder?

bigtimebuddy:
Alright, works!

bigtimebuddy:
The only other improvement I want to make to this is to stop building the project twice. This happens as part of `pretest` and part of `dist`. Would save us some time if we only had to do this once.

bigtimebuddy:
Don't know why some tests are intermittently failing. Need to dig in. It's consistently having an issue with this one. Maybe a timeout?
```
should correctly calculate alpha mask area if filter is present (1073ms)
```

cursedcoder:
Are there any failing tests? Because builds are green.

Also, I'm a bit worried there are 2 builds for 1 PR.

It would be nice to limit it to just 1 build.

bigtimebuddy:
There were tests failing, but they show up green because I re-ran the tests. I think performance is port running Github Actions and some Mocha tests reach the 2second timeout. We might need to adjust that to be a little more forgiving.

The two builds are because are doing builds on branch pushes and on pull-requests pushes. Wouldn't be nice to have one, but the majority of PRs from outside folks and only run once.

cursedcoder:
@bigtimebuddy anything I can help with?

bigtimebuddy:
Removing the extra unnecessary build process reduced the time from ~4m 56s to ~3m 47s

PR #6591, opened 2020-04-26
Map video file extension to correct mime subtype where necessary

### LINKED ISSUE #6146: Wrong mime type for OGG Videos
Try to create texture from video file with .ogv extension.

in core.js:
`mime = mime || ("video/" + ext);`
in that case mime type will be "video/ogv" and video does not play and texture.valid is false

After I changed that to "video/ogg" - video is playing ok

tested in Chrome

mfallmann:
Fixes #6146 

I noticed that some of the other video mime subtypes where non-standard as well so I've added a simple subtype map according to the list at the bottom of this page: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Containers


mfallmann:
Thanks for the feedback, I've implemented the changes as requested by @bigtimebuddy!

bigtimebuddy:
Thanks for the PR @mfallmann :tada:

Issue #6524, opened 2020-04-05
Proposal: Improving text rendering


ShukantPal:
The community has said it - Pixi's text rendering performance needs to improve. This issue is devoted to how we can improve and what I think will be the best way to deliver:

* Signed-distance fields: This technique renders the text into a texture (using Canvas 2D API) and applies a distance transform. In simple terms, the distance transform will set the value of each pixel in the output texture to the distance of that pixel to the nearest outline of the text in the input texture. pixi-sdf-text is an example: https://github.com/PixelsCommander/pixi-sdf-text

* VTMs: Vector texture maps encode curve discontinuities at pixel level in textures.

* Exact bezier curve rendering: This renders the font "as is" by tessellating everything except the curves. The curves are rendered using a special fragment shader (no sampling of the curve is done, it is rendered exactly on the GPU with antialiasing): https://www.microsoft.com/en-us/research/wp-content/uploads/2005/01/p1000-loop.pdf. I created a demo for quadratic bezier curve: https://codepen.io/sukantpal/pen/GRJawBg?editors=0010

@bigtimebuddy and I discussed these methods and we thought the third approach is the best because:

* SDFs require pre-generated atlases. Generating multi-channel SDFs is very complicated (if we want to do it at runtime).

* VTMs are way too complicated for just fonts.

* Exact bezier curves just requires you to render the font as a path like everything else in Graphics.

eXponenta:
It is should not be in core.
It is heavy package and MUST BE implemented outside main package, core team should not worries about implementation of it.

Pixi SDF is right example.

SukantPal:
@eXponenta `PIXI.Text` implementation will be replaced by a higher performance implementation. This has already been decided.

bigtimebuddy:
I'd like to give some more context. I have asked @SukantPal to research alternative Text/BitmapText approaches that have different trade-offs. Particularly, if we can find something that performs better than Text and looks great.

bigtimebuddy:
There are a few pieces of criteria that I'd use judge a good Text display object:

- **Good Runtime Performance** - Rendering text and changing text is low-cost
- **Low memory**  - RAM usage is low
- **Small Dependencies** - dependency footprint file-size is small
- **Great Fidelity** - works well with different resolutions
- **Flexible styling** - support for strokes, drop shadows, gradients
- **Supports Large Glyph Languages** - E.g., Chinese, Japanese, Korean
- **Backward compatible** - Works with with Context2D

[...]

PR #6597, opened 2020-04-30
Text rendering with atlas-based glyph caching 

SukantPal:
Proposal #6524 

##### Description of change
This PR introduces a more efficient method of rendering text in WebGL. Instead of rendering each `PIXI.Text` instance into its own canvas, each letter is rendered into a shared atlas via the "glyph manager".

This, unfortunately, does not work under certain circumstances. For that, the "fallbackMode" flag will make the text render using the old method:

+ when horizontal gradients are used (this is because the color of each letter depends not only on the style but also on its position in the local space)

+ when drop shadows are used. Since Safari has a bug with gradients+shadows, we had a "hack" to correctly render gradients+drop-shadows in our old method. However, you can't use that with shared glyphs (because you can't "throw" the text far away, it might render over another glyph).

+ large strokes. Technically, all strokes are rendered before all fills. Not doing so only works until the strokes don't start overlapping other characters. This is why I limited the stroke size to 10% of the font size. If larger, the fallback mode is enabled.

-----

**How are glyphs "stored"?**

The glyph manager maintains a list of 1024x1024 canvases. Each canvas is divided into a grid of 16x16 tiles. To find space for a glyph, a "rectangle of tiles" is found that has enough space for it. A rectangle merging algorithm is used to find the largest rectangles.

**Have you measured the benefits?**

Yes.

+ When there are multiple Text instances, characters are still shared and that saves canvas memory.

+ When a large number of characters are rendered, the repeated characters aren't rendered twice.

I'll try to create a fiddle for what I've tested.

**Are you garbage collecting characters?**

Yes, see `GlyphManager#prune`. Each glyph has a "references" semaphore. When it becomes zero, the glyph will be garbage collected (see `Text#_updateGlyphs`).

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/GlyphManager.ts
+++ b/packages/text/src/GlyphManager.ts
@@ -123,7 +123,8 @@ export class GlyphManager
             throw new Error('Can\'t store >512px glyph');
         }
 
-        let { width, height } = GlyphManager.measureGlyph(char, style);
+        const metrics = GlyphManager.measureGlyph(char, style);
+        let { width, height } = metrics;
 
         width *= resolution;
         height *= resolution;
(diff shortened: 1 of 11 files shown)


bigtimebuddy:
Can you rebase this to dev, this has all. your old graphics changes that we already merged. Also, can you fix the CI build so we can test the build?

GoodBoyDigital:
slick! just a thought.. but drop shadows and strokes could be achieved with filters ;D

GoodBoyDigital:
* baked in to the texture filters not live ones

SukantPal:
@GoodBoyDigital Yes. I could spend some more time looking into that.

Note: There seems to be a bug in "clearing" glyphs before overwriting. After some time, this pen's (https://codepen.io/sukantpal/pen/OJyxzRg?editors=0010) description (at the bottom) will start getting blue artifacts. I'll have to look into it.

SukantPal:
@GoodBoyDigital Another obstacle in using our own filters is that they would introduce a dependency on `@pixi/filter-*`. I don't know if people are okay with that (and possibility of circular dependency b/w pixi.js bundle & filters).

PR #6600, opened 2020-05-02
BitmapFontFactory

SukantPal:
#6524 

(follow up to #6597)

@bigtimebuddy wanted a simpler API. Instead of a _dynamic_ caching layer beneath `PIXI.Text`, we opted for a _static_ bitmap generation layer beneath `PIXI.BitmapText`. This has no crazy calculations for tiling and all that.

##### Description of change
* `@pixi/text-bitmap` depends on `@pixi/text` for styling stuff. The factory should generate _any_
`ITextStyle`.
##### Pre-Merge Checklist


SukantPal:
@bigtimebuddy Just did. Why not add packages/**/package-lock.json to .gitignore?

SukantPal:
* https://codepen.io/sukantpal/pen/ExVbaEq?editors=0010

* https://codepen.io/sukantpal/pen/abvVzKo?editors=0010

SukantPal:
@bigtimebuddy Regarding performance, try running the demo locally. Although Stats.js shows 12 FPS, Chrome DevTools (locally) shows 55-70 FPS range.

Screen Shot 2020-05-02 at 2 36 43 PM

packages/text-bitmap/src/BitmapFontFactory.ts
```diff
@@ -0,0 +1,221 @@
...
+
+        context.fillStyle = 'rgba(0, 0, 0, 0)';
+    }
+}
+
+export interface IBitmapFontFactoryOptions extends ITextStyle
+{
+    chars: string | string[];
+    resolution: number;
+}
```
bigtimebuddy, line 214: I would change these options to require `chars`.
```suggestion
export interface IBitmapFontFactoryOptions extends Partial<ITextStyle>
{
    chars: string | string[];
    resolution?: number;
}
```

packages/text-bitmap/src/BitmapFontFactory.ts
```diff
@@ -0,0 +1,221 @@
...
+ */
+export class BitmapFontFactory
+{
+    /**
+     * Generates a bitmap-font for the given style and character set. This does not support
+     * kernings yet.
+     *
+     * @param {PIXI.IBitmapFontFactoryOptions} options
+     */
+    public static create(options: Partial<IBitmapFontFactoryOptions> = {}): BitmapFont
```
bigtimebuddy, line 27: Then you don't need Partial here and options should be non-optional.
```suggestion
    public static create(options: IBitmapFontFactoryOptions): BitmapFont
```

packages/text-bitmap/src/BitmapFontFactory.ts
```diff
@@ -0,0 +1,221 @@
...
+                file: '',
+            };
+
+            positionX += (width + (2 * BMT_PADDING)) * resolution;
+        }
+
+        return new BitmapFont(fontData, textures);
+    }
+
+    private static render(
```
bigtimebuddy, line 138: Even if this is internal, could you add comments to all functions? It's helpful to know what it does without reading the code.

bigtimebuddy: Also, how much of this overlaps from Text's rendering?

packages/text-bitmap/src/BitmapFontFactory.ts
```diff
@@ -0,0 +1,221 @@
...
+ * @memberof PIXI
+ * @class
+ */
+export class BitmapFontFactory
+{
+    /**
+     * Generates a bitmap-font for the given style and character set. This does not support
+     * kernings yet.
+     *
+     * @param {PIXI.IBitmapFontFactoryOptions} options
```
bigtimebuddy, line 25:
```suggestion
     * @param {PIXI.IBitmapFontFactoryOptions} options - Includes all `ITextStyle` properties.
     * @param {string|string[]} options.chars - Character included in the font set.
     * @param {number} [options.resolution] - Optional resolution to define.
     * @return {PIXI.BitmapFont} Font generated by style options.
```

GoodBoyDigital:
This is awesome! 

Agree that functionality being on BitmapText is the best place!

- standard Text - Flexible
- bitmap Text - Speedy

What you have done is basically elevated bitmap text so that it is easy for everyone to use not just people who have hard to find and figure out font generation software!

How do you guys feel about simplifying the API even further and removing the need to actually create the font before its used? As well as exposing the BitmapFontUtils I think _also_ doing it under the hood would be the best developer experience :)

```

// under here - we generate the texture and store in a cache  based on the text style passed in (defaults filled in minus not visual bits like alignment).

const text = new PIXI.BitmapText('Hello world! This is the new PIXI.BitmapText()', {
    font: 'serif',
    align: 'center' });
```

bigtimebuddy:
I’m not sure that the constructor for BitmapText also works for this without being a breaking change or too overloaded to catch errors. I think create and apply are two distinct steps requiring two lines of code.

For instance, there is not a great way to reconcile the font name if we can use the constructor. There are a couple font states for BitmapText: available system font, installed bitmap font, or missing font. And even if you have a bitmap font, when to regenerate seems like a question you may not want to do inside the constructor.

We could maybe simplify the API and do BitmapFont.create instead of BitmapFontFactory. But I think this needs to be two steps UNLESS we introduce a new display object like SimpleBitmapText, or something like that. Then we can do all the font generation automagically.

GoodBoyDigital:
> I’m not sure that the constructor for BitmapText also works for this without being a breaking change or too overloaded to catch errors. I think create and apply are two distinct steps requiring two lines of code.
> 
> For instance, there is not a great way to reconcile the font name if we can use the constructor. There are a couple font states for BitmapText: available system font, installed bitmap font, or missing font. And even if you have a bitmap font, when to regenerate seems like a question you may not want to do inside the constructor.
> 
> We could maybe simplify the API and do BitmapFont.create instead of BitmapFontFactory. But I think this needs to be two steps UNLESS we introduce a new display object like SimpleBitmapText, or something like that. Then we can do all the font generation automagically.

`automagically` love that! currently have my head in the whole zero config style thinking. If we can assume intent for 90% of users but keep it open enough to let people get specific then I think thats the sweet spot.

appreciate if it's a breaking change then maybe changing the API for BitmapText might be better off as a v6 thing...

something like?

```

// I don't know pixi, but daaaanng i love that I can use it so easily..

// new bitmap text...
const text = new BitmapText('omg so easy to use pixi fonts!', {
    font:  'Arial`,
    size: 32  
});

// same as current text..
const textAgain = new Text('omg still so easy to use pixi fonts!', {
    font:  'Arial`,
    size: 32  
});

// im a pixi power user yo! I know this stuff inside and out and want to control whats happening !

const font = BitmapFont.create({
    font:  'Arial`,
    size: 33  
}) // (or from for consistency?)

const text = new BitmapText('omg so easy to use pixi fonts!', font); 

```

Like you say @bigtimebuddy , i guess the issue lies in if we don't want a breaking API change.. what does the interim API look like. creating a new class may be best here? Also, overloading constructors.. generally thats ok right?

This new text could accept `BitmapFont | TextStyle | Object`

GoodBoyDigital:
ooh oh so `BitmapFont.from` could be a good way to go?

```

// make one on the fly..
const font = BitmapFont.from({
    font:  'Arial`,
    size: 33  
}) 

or get one you already loaded..
const font = BitmapFont.from('font-preloaded');

```


```

**Comment by bigtimebuddy** [comment]
Let’s start with BitmapFont.from and then look at future PRs to build convenience on top, add a new display object or break BitmapText to support this path as first class. Does that sound good @Goodboydigital?

```js
BitmapFont.from({ chars:”0123456789”, fontFamily: “custom”, resolution: 2 }):
const counter = new BitmapText(“0”, { fontFamily: “custom”});
```

GoodBoyDigital:
cool, sounds good 👍

packages/text-bitmap/src/BitmapFontFactory.ts
SukantPal: A lot of `drawGlyph` comes from `Text#drawLetterSpacing` and `updateText`. However, it doesn't have the two-pass shadow rendering (to bypass a bug in WebKit).

---
I changed `render` to a local function named `drawGlyph`.

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,281 @@ export class BitmapFont
...
+     *     font: "Arial",
+     *     fontSize: 12,
+     *     strokeThickness: 2,
+     *     fill: "purple"
+     * });
+     *
+     * const text = new PIXI.BitmapText("This is an example using Pixi's BitmapText+Font", font);
+     */
+    public static from(options: IBitmapFontFactoryOptions): BitmapFont
+    {
```
bigtimebuddy, line 263: One small tweak. Could we optionally provide the name (maybe part of options) of the font so that you could register it as part of the `from` call? That seems consistent with `Texture.from` that adds to the TextureCache.

SukantPal:
@bigtimebuddy Will do right now! I can't seem to get the next build here (https://pixijs.download/feature/bitmap-font-factory/pixi.js) - the `BitmapFont.from` method isn't showing. I cleared the cache on Firefox/Safari and it doesn't show up.

Could you see if it comes up?

SukantPal:
@bigtimebuddy Could you deploy this branch so that the demos can be updated?

SukantPal:
@bigtimebuddy I'm not sure how we would add unit tests for checking whether the glyphs rendered at the right "resolution".

A `TextStyle` object is constructed from the passed options so we should be using the defaults. How do you want to test that as well?

bigtimebuddy:
@cursedcoder do you know what `Deploy Non-Tag Branch` is not getting invoked with these branch changes? https://github.com/pixijs/pixi.js/pull/6600/checks?check_run_id=647208137

bigtimebuddy:
For resolution @SukantPal, render two glyphs at different resolutions and compare the texture sizes. 

We should probably be using the same defaults or, probably better, internally convert into a TextStyle object.

SukantPal:
@bigtimebuddy

* The texture sizes are same as the base textures resolution changes so texture frame stays same. The difference would be in the same real sizes.

* the .from method creates a TextStyle object already from the passed options. But that object isn’t “saved” anywhere.

bigtimebuddy:
- comparing real sizes seems good
- sounds good! I missed that

SukantPal:
* Except textures don't have a real size. You can only compare their base-texture's resolution. I just added a test for that.

bigtimebuddy:
I’ll work to fix the deployment. Not sure why that’s not working.

bigtimebuddy:
I created another branch **dev-bitmap-font-from**
https://pixijs.download/dev-bitmap-font-from/pixi.min.js

cursedcoder:
@bigtimebuddy not sure about it either.

However, it is clear that only "pull-request" event is triggered (this one isn't deployed), I assume it may be of `feature/` prefix.

cursedcoder:
@bigtimebuddy 
image

So we need `**`

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,293 @@ export class BitmapFont
...
+                id,
+                file: '',
+            };
+
+            positionX += (width + (2 * BMT_PADDING)) * resolution;
+        }
+
+        const font = new BitmapFont(fontData, textures);
+
+        if (name)
```
bigtimebuddy, line 424: What do you think should be the behavior trying to set a name that's already in use? We could throw an error or automatically destroy the existing font? There's a potential memory leak here if we don't destroy all the textures and someone uses BitmapFont.from over and over again to overwrite.

SukantPal: I could add a throws clause to prevent name collision (and a test for that). Do you want another API for unregistering these fonts?

bigtimebuddy: We already have `BitmapFont.uninstall`. I think a throw that lets users know they should uninstall sounds good. And yes, a test for that.

SukantPal: @bigtimebuddy I added a test for name collision (that checks if an error is thrown).

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -18,15 +21,18 @@ interface IBitmapFontCharacter {
 // Internal map of available fonts, by name
 const available: Dict<BitmapFont> = {};
 
+// Default dimensions of the bitmap font atlas texture
+const BMT_DIMEN = 1024;
```
themoonrat, line 25: Is there a reason the default size is 1024 and note 2048, since all browsers support 2048x2048 textures? Or is it because it's likely for general usage 1024x1024 will cover the glyphs required for English?

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -18,15 +21,18 @@ interface IBitmapFontCharacter {
...
+
+// Default padding b/w glyphs in the bitmap font atlas
+const BMT_PADDING = 4;
```
themoonrat, line 28: It would be nice if these settings weren't hard coded in here, but could be adjusted by developers. Maybe they're part of the settings passed in via 'from'? So can be adjusted on a case by case basis.

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,298 @@ export class BitmapFont
...
+     *     fontSize: 12,
+     *     strokeThickness: 2,
+     *     fill: "purple"
+     * });
+     *
+     * const text = new PIXI.BitmapText("This is an example using Pixi's BitmapText+Font", font);
+     *
+     * const title = new PIXI.BitmapText("This is the title", { font: "TitleFont" });
+     */
+    public static from(options: IBitmapFontFactoryOptions): BitmapFont
```
themoonrat, line 265: I think it'd be nice if the options for the bitmap font, such as chars, name etc, were separate from the text style options, such as FontFamily, fontSize etc. Then it makes it easier to transition from PIXI.Text to PIXI.BitmapText

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,298 @@ export class BitmapFont
...
+        let positionY = 0;
+
+        let canvas;
+        let context;
+        let baseTexture;
+        let maxCharHeight = 0;
+
+        for (let i = 0; i < chars.length; i++)
+        {
+            if (positionY >= BMT_DIMEN - ((style.fontSize + BMT_PADDING) * resolution))
```
themoonrat, line 354: It'd be nice to know, as a dev, how efficient we've been with the atlases.  If all we've added is numbers, then we could have a massive area not used, and we could reduce the size of the atlas. Or if we've used 3 or 4 atlases, we could know to increase the atlas size to say 2048x2048 and beyond. Not sure if it's beyond the scope of this PR or not...  but I know as soon as I start using the extra functionality here, it's gonna be one of the first questions I'm going to have!

packages/text-bitmap/src/utils/generateFillStyle.ts
```diff
@@ -0,0 +1,148 @@
...
+ */
+export function generateFillStyle(
+    canvas: HTMLCanvasElement,
+    context: CanvasRenderingContext2D,
+    style: TextStyle,
+    resolution: number,
+    lines: string[],
+    metrics: TextMetrics
+): string|CanvasGradient|CanvasPattern
+{
```
themoonrat, line 19: I'm worried that we have repeated code in generateFillStyle between here and 'Text'?

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,298 @@ export class BitmapFont
...
+    const char = metrics.text;
+    const fontProperties = metrics.fontProperties;
+
+    context.translate(x, y);
+    context.scale(resolution, resolution);
+
+    const tx = style.strokeThickness / 2;
+    const ty = -(style.strokeThickness / 2);
+
+    context.font = style.toFontString();
```
themoonrat, line 468: I'm worried that we have repeated code in drawGlyph between here and 'Text'?

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,298 @@ export class BitmapFont
...
+                    }
+                }
+                // Handle a character set string
+                else
+                {
+                    flatChars.push(...elem.split(''));
+                }
+            }
+
+            chars = flatChars;
```
themoonrat, line 320: My other concern, that'd be nice to be at least known with some sort of debug flag, or logging, is when BitmapText is going to try to draw a character that is _not_ listed. I'm going to be using this for all sorts of fast counting up currency meters and such, so what if I left out a Euro symbol, by accident? Currently, it'll just silently miss out that char when rendering, but I'd really like to know that I missed something

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,298 @@ export class BitmapFont
...
+     * @param {PIXI.IBitmapFontFactoryOptions} options - includes all `ITextStyle` properties.
+     * @param {string|string[]} options.chars - characters included in the font set. You can also
+     *      use ranges. For example, `[['a', 'z'], ['A', 'Z'], "!@#$%^&*()~`{}[]"]. Don't forget to
+     *      include spaces ' ' in your character set!
+     * @param {number} [options.resolution=1] - optional resolution to render the glyphs at
+     * @return {PIXI.BitmapFont} font generated by style options.
+     * @static
+     * @example
+     * const font = PIXI.BitmapFont.from({
+     *     chars: [['a', 'z'], ['A', 'Z'], "!#$%^&*()~`{}[]"]
```
themoonrat, line 253: It'd be nice to provide either some defaults, or some guidance for defaults. Like
chars: 'ASCII' to auto add all ascii chars, or docs in here to say 'here is the regex to add all ascii characters
chars: 'numeric' to add all numbers plus known separators, or docs in here to say 'here is the regex to add all numbers'
Docs probably the easier route, but I think getting in some common scenarios now would save lots of questions in the future

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,298 @@ export class BitmapFont
     {
         return available;
     }
+
+    /**
+     * Generates a bitmap-font for the given style and character set. This does not support
+     * kernings yet.
```
themoonrat, line 242: when using this for meters, one of the biggest requirements is that each character takes up the same width. So when you're counting up numbers, they all stay in the same spot. If kerning isn't support yet, could there be a way to hack in some consistent kernings for, say, each numeric character, docced for now?

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: I was thinking about this too. Then we could pass in a shared TextStyle object instead of converting a raw object.

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: I propose a future PR to refactor the shared bit better.

packages/text-bitmap/src/utils/generateFillStyle.ts
bigtimebuddy: Same here, let’s not balloon scope. We can find a better way to shared some of the internal rendering guys. Maybe adding TODO in the code would be a good reminder.

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: Is there any reason we couldn’t trim the canvas after we render glyphs to it?

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: This is BitmapText behavior and probably another PR. Since this PR doesn’t touch BitmapText and only uses it for rendering.

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: BitmapFont.ALPHA
BitmapFont.NUMERIC
BitmapFont.ALPHA_NUMERIC

Convenience things like that?

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: Good call

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: Seems like an arbitrary-ish magic number, right @SukantPal?

SukantPal: @bigtimebuddy Correct. Should I add an option for this in `IBitmapFontOptions`.

packages/text-bitmap/src/BitmapFont.ts
themoonrat: Yeah, I'm ok with that. That's why in my basic implementation, I actually _do_ use PIXI.Text to render each letter, then take the texture and use it, so that you guarantee it's the same

packages/text-bitmap/src/BitmapFont.ts
themoonrat: There is the canvas trim util we have; so could default to 2048x2048, to fit as much in as possible at a size we know all devices support, then trim down anything unused at the end?

packages/text-bitmap/src/BitmapFont.ts
SukantPal: @themoonrat But then don't you have one texture per glyph?

themoonrat: @SukantPal - I do, that's very much the downside of what I have, and why I've had an outstanding JIRA for nearly 3 years to attempt to do what you've got here! :)

packages/text-bitmap/src/BitmapFont.ts
SukantPal: @themoonrat Sorry on this one. I tried searching about kernings and how they work but I couldn't find any good documentation. If you could point me to something, that would be helpful 😃

packages/text-bitmap/src/BitmapFont.ts
themoonrat: Yeah. They refer to regexes themselves in the background. We could have a collection of those, so I could go

options.chars = [ BitmapFont.ALPHA_UPPERCASE, BitmapFont.NUMERIC, BitmapFont.SEPARATORS ]

If I wanted chars just for a high score board for example

packages/text-bitmap/src/BitmapFont.ts
themoonrat: I'll take a look tomorrow :)

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: I’m also okay with a follow up PR to add some syntax sugar on top of setting character ranges. I don’t think it changes the base functionality here.

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -226,4 +236,298 @@ export class BitmapFont
...
+    {
+        const name = options.name;
+        let chars = options.chars;
+
+        if (name && BitmapFont.available[name])
+        {
+            throw new Error(`[Name Collision]: The name "${name}" has already been taken by another bitmap font`);
+        }
+
+        // Split the chars string into individual characters
```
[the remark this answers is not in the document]
SukantPal: I named it `processChars` 😅

packages/text-bitmap/src/BitmapFont.ts
SukantPal: @themoonrat I added `ALPHA`, `NUMERIC`, `ALPHANUMERIC`, and `ASCII`. @bigtimebuddy Do you think we should add more (like lower/uppercase, separators, ...)?

packages/text-bitmap/src/BitmapFont.ts
SukantPal: I added `textureWidth`, `textureHeight` to `IBitmapFontOptions`. Also, `BMT_DIMEN` is 2048 as suggested by Dave.

packages/text-bitmap/src/utils/generateFillStyle.ts
SukantPal: I added a TODO comment 👍

packages/text-bitmap/src/BitmapFont.ts
SukantPal: @themoonrat Where is the canvas util you're referring to?

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: I think that’s okay for now. We can add more easily if need be. Let’s see what kinda feedback we get once we release this.

packages/text-bitmap/src/BitmapFont.ts
bigtimebuddy: PIXI.utils.trimCanvas

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -237,6 +237,41 @@ export class BitmapFont
...
+    public static ALPHANUMERIC = [['a', 'z'], ['A', 'Z'], ['0', '9']];
+
+    /**
+     * This character set consists of all the ASCII table.
+     * @readonly
+     * @static
+     * @member {string[][]}
+     * @see http://www.asciitable.com/
+     */
+    public static ASCII = [['\0', '~']];
```
englercj: These are marked as readonly in the docs, but not in the Typescript, is that intentional?

bigtimebuddy:
@themoonrat, would you mind reviewing this again? I think we are in good shape. I think layout concerns (like monospacing) should be handled as a new feature for BitmapText, since ultimately that class positions each glyph. Overriding the font data seems less flexible to solve that problem.

themoonrat:
Had another look - happy with the code!
Just now having a play with the usage of the class to ensure it behaves as expected

themoonrat:
Creating a font with no chars given at all should be an error.

The xAdvance being created for each char are to decimal places; no bitmap generator would ever give data in this way, so these should be rounded in some way. Might get blurry text when rendering to decimal positions.

The 'page' in the data for each char just gives the id of the char itself. It should say which sheet # the text is found on. So if generating takes up 2 base textures, the page should either be '0' or '1'. Useful to know how many textures this font generation is requiring.

Should these Textures and BaseTexture be added to the caches so we know they've been created? Like they are for textures created in the Text class?

Visual issue with drop shadows AND stroke on text: https://www.pixiplayground.com/#/edit/nvZj_usqdMGPmDbfLIFUR
Issues if Italic is used: https://www.pixiplayground.com/#/edit/Y6pgxZRlVMR5qND9igixc

SukantPal:
@bigtimebuddy  @themoonrat Hey Dave, I wanted to first touch upon the issue with drop-shadows. I think the "buggy" look of the  stroke + drop-shadow **will still stay there because the browser renders it like that**.

* `PIXI.Text` does not suffer from the stroke being hidden by the drop-shadow **because it uses two passes - and the first pass renders the drop-shadow separately**.

* This fiddle shows a stroke + drop-shadow used directly on a Canvas2D context. It isn't perfect - and that's why BitmapText won't be either. https://jsfiddle.net/ShukantPal/qtxozr1s/

* This a BitmapText (top) vs Text (comparison):
Screen Shot 2020-05-12 at 10 59 39 AM

* https://jsfiddle.net/ShukantPal/qtxozr1s/ - this is using the Canvas2D API directly

Screen Shot 2020-05-12 at 11 00 47 AM

* You can see that BitmapText looks more like the Canvas 2D render. This is the atlas generated by the BitmapText for your pixi-playground:

Screen Shot 2020-05-12 at 10 08 19 AM

* Since we render glyphs one-by-one the sometimes the stroke is rendered over the drop-shadow. This accounts for those subtle differences b/w BitmapText and Canvas2D.

---
* BitmapText **cannot** handle scenarios where characters overlap. This is because
  * all glyphs strokes are supposed to be rendered before all glyph fills
  * but we are rendering each glyph (stroke+fill) sequentially
  * and we don't want to cache the stroke & fill separately (too complicated, takes 2x space, BitmapFontData doesn't support it)

* Side note: Your pixi-playground links has a little blurry BitmapText (w.r.t the canvas text) because **you didn't set the resolution of the BitmapText**. If you add a `resolution: window.devicePixelRatio || 1` to the `IBitmapFontOptions`, it would look better.

themoonrat:
Regarding drop shadows, yeps, I understand what you're saying, that's fine. No different to if you created a bitmap font now with drop shadows, then tries to place them close to each other so they overlapped :) I worried that it was being caused by the the glyphs in the atlas themselves perhaps overlapping due to the drop shadow.

SukantPal:
@themoonrat Okay, so:

* italics should be working now

* if no characters are resolved, then an error is thrown

* I round up texture frames for glyphs **and** xadvance

SukantPal:
@themoonrat Regarding the pages => char mapping issue, I need more information on how BitmapFont uses those pages & the textures array. TBH, I'm confused b/w (x,y) & (xoffset, yoffset) in the font data.

themoonrat:
So, taking a look at https://pixijs.io/examples/#/text/bitmap-text.js
https://pixijs.io/examples/examples/assets/bitmap-font/desyrel.xml describing https://pixijs.io/examples/examples/assets/bitmap-font/desyrel.png

In this bitmap font, there is only 1 image atlas, desyrel.png, so the page is '0' for all of the characters. 

<pages>
<page id="0" file="desyrel.png"/>
</pages>

If there were 2 image atlases, say, a desyrel1.png and a desyrel2.png, then some of the characters would have a page 0, and some a page 1.

So in this scenario, where you do 
`// Create new atlas once current has filled up`
increment a pageIndex and use that as the page value

SukantPal:
@themoonrat Okay, but another question: what is `(x, y)` and `(xoffset, yoffset)`?

themoonrat:
When bitmap fonts are generated, in the xml, the x, y, width and height are to tightly draw a box over each glyph in the image. Now, how do you then know how to draw each glyph with the same baseline? That's where yoffset comes in. It lets you know how much to adjust the y position when drawing that glpyh. So a lowercase z will have a bigger y offset than an uppercase Z

http://www.angelcode.com/products/bmfont/doc/render_text.html

In the atlases produced here, each item drawn is with the same baseline, so that kinda solves not needing to have a yoffset. The fact that when doing our measuring, width wise there is empty spacing around the edge, solves not needing to have a xoffset.

But that does mean that they work in slightly different ways. With regular bitmap fonts, each glyph data is tight around the character with xoffset and yoffset to normalise their positions. Whereas this 'from' method, each glyph data is _not_ tight around each character, and instead has the padding build it.
I think we're ok with that? But if you measure the of a Sprite for an Arial 30px bitmap, or Arial 30px generated, those widths would be quite different, so I don't know if that's a potential compatibility issue. I'm not brave enough to make that call :D

SukantPal:
@themoonrat OK, I'll try changing the pages mapping to per-BaseTexture instead of per-glyph 😄

themoonrat:
PS, I found a way to effectively turn all numbers fixed width
https://www.pixiplayground.com/#/edit/-VOO6_S1qQoVcbvBDUwi9
So it's easily to go into the font data and edit without requiring any extra support Pixi wise :)

SukantPal:
@bigtimebuddy @themoonrat I also got the issue with glyph pages resolved. You should see fewer different pages in the whole character set. This should be it!

themoonrat:
Looks good!

My only remaining question is for @bigtimebuddy, regarding whether these generated base textures & textures should be added to the BaseTextureCache and TextureCache. Normal Text has it's textures added there, which I find useful to be able to track asset usage. 

I believe it's just the 'adding' to the cache required, as the 'destroy' function in BitmapFont would already handle the removing? Tho the destroy function for bitmap text never destroys the base texture?

But yeah, just clarity on what to do that wise. I'm super happy with how the class works now

bigtimebuddy:
@themoonrat, this implementation is somewhat consistent with current approach to handling BitmapFonts, that each glyph Texture is created by and destroyed by BitmapFont. It's the responsibility of the font to maintain the references (i.e., not added to TextureCache). 

The BaseTextures (page textures), however, _are_ added to the cache if it's create from the Loader. `BitmapFont.from` here doesn't do that. How to reconcile?

In general, Pixi should follow a "campsite rule", which is, if a class creates garbage, it should also destroy it. Graphics, for example, only destroys geometry if it was created internally and not passed by reference.

That's a longwinded way to say that I think we need to fix this. I believe BitmapFont should maintain reference to page textures and we should provide a way in the `destroy` method (like a boolean similar to Sprite) whether to destroy the page textures as well. This change seems a little out of scope for this PR. However, the page textures created with `BitmapFont.from` should _probably_ be added to the caches so that it's consistent (however poorly executed) with BitmapFonts via the Loader.

SukantPal:
@bigtimebuddy A g r e e.

bigtimebuddy:
Are you proposing adding a static destroy method `BitmapFont.destroy`? We don't really have that pattern anywhere in Pixi. Everything is an instance destroy.

Destroying actually is probably more important for this workflow in-case you want to build something on top of this PR that auto-generates the font based on characters. I think I'll do a follow-up to improve this. I think your work here is done @SukantPal. You've put a lot of effort into this and we super appreciate it.

Scrolling back, here are some future PRs related to this:

* Make BitmapFont destroy cleanup base textures from Loader or from Bitmap.from
* Improve ergonomics around regenerating fonts (destroy to remove from available)
* Clean-up code duplication of generateFillStyle

SukantPal:
@bigtimebuddy I was talking about an instance destroy, not static. But it is really a trivial difference.

themoonrat:
Happy for follow up PRs to handle the missing bits.
Thanks @SukantPal :)

PR #6607, opened 2020-05-06
Build for all push branches including ones with slashes

bigtimebuddy:
Follow-up to #6600 comment about not building branches that are `feature/*` or basically contain any `/`. We want to build on all branches pushes, regardless of name.


bigtimebuddy:
Looks like this change worked because we are now getting push and pull_requests now.

PR #6610, opened 2020-05-07
expose getTestContext 

Fxlr8:
##### Description of change

`getTestContext` function is now a method of `testContext` object.
This allows to change this function outside of PIXI.
In my case I need this feature to make `pixi.js v5` work with `react-native`, `expo` and `expo-gl`
Here is my [thread about this problem ](https://www.html5gamedevs.com/topic/45339-pixijs-expo-gettestcontext-workaround/?tab=comments#comment-250961)
##### Pre-Merge Checklist


bigtimebuddy:
Could you please revert the version bump?

Fxlr8:
Done. Sorry for that. I had no idea that Lerna would mess up the git repo and even push a commit under the hood.

Fxlr8:
I don't see `testContext` inside PIXI object in `bundles/pixi.js/index.d.ts`. I had exported it from `core/src/shader/utils/index.ts`, then re-exported them in `core/index.ts`. I can find it in dist js files, but not in PIXI types. VSCode hints don't recognize `PIXI.testContext` either. I don't know how to test it correctly. I can't npm link the project, because it is a monorepo and some paths break. How do you test pixi in a third party project without publishing it?

bigtimebuddy:
My understanding is that you want to override `getTestContext` so that you can get it to work in these other non-browser environments. And the only way you can override it is if you attach it to a class and don't make `getTestContext` a direct export. Is that it?

Supporting these other non-browser environments is not within the scope of this project at the moment. There are lots of other places where PixiJS is using DOM APIs like `document.createElement`, etc. How are you handling those?

Basically, I'm not a huge fan of this change. It feels like a hack to support environments that we're unable to fully support at the moment. Also, in general, we are very careful of needlessly expanding the API, which is already rather large. This exposes some private internals that aren't useful for the vast majority of users.

I think these other environments may be important and have value for PixiJS, but can we discuss first-class support would look like? A bunch of monkey-patches, polyfills and hacks is not sustainable.

packages/core/src/shader/Program.ts
```diff
@@ -45,10 +45,10 @@ export class Program
     public vertexSrc: string;
     public fragmentSrc: string;
     nameCache: any;
-    glPrograms: { [ key: number ]: GLProgram};
+    glPrograms: { [key: number]: GLProgram };
```
bigtimebuddy, line 48: Can you please revert all these whitespace changes? They are not relevant to your change and we'd rather manage whitespace with eslint.

packages/core/src/shader/utils/testContext.ts
```diff
@@ -0,0 +1,61 @@
+import { settings } from '../../settings';
```
bigtimebuddy, line 1: It's not always easy to tell from GitHub, but can you make sure you do a `git mv` on **getTextContext.ts** to **testContext.ts**? Creating a new file destroys the git history, which is bad for blaming.

packages/core/src/shader/Program.ts
Fxlr8: Ok, no problem. This was messed up by my editor. But pixi's lint-fix was fine with this whitespaces, so I decided to keep them. Maybe we should add a lint rule for them?

bigtimebuddy: Probably, but that should be another PR.

packages/core/src/shader/utils/testContext.ts
Fxlr8: Sure. Types are not generated correctly, so it is not ready to merge anyways.

Fxlr8:
@bigtimebuddy you are right. I should try my best to mock the environment to PIXI first. Closing for now

bigtimebuddy:
I've had a plan to build a `@pixi/dom` package which will consolidate all the DOM calls in one place with a default Browser adapter installed. This would make it easier to override in non-browser contexts like Node and some of the places you mentioned. I think that would also solve your issue. Happy if someone else wants to take a stab at this!

Fxlr8:
This would be amazing. I will definitely do it if I fail with the polyfill

PR #6622, opened 2020-05-14
Convert canvas-particles to TS

SerG-Y:
#### Description of change
Converted canvas-particles to TS

##### Pre-Merge Checklist


packages/canvas/canvas-particles/src/ParticleContainer.ts
```diff
@@ -32,7 +33,7 @@ ParticleContainer.prototype.renderCanvas = function renderCanvas(renderer)
 
     for (let i = 0; i < this.children.length; ++i)
     {
-        const child = this.children[i];
+        const child: any = this.children[i];
```
bigtimebuddy, line 36: What about using Sprite here instead of any?

SerG-Y: Yes, it's a good idea

SerG-Y: Fixed

packages/canvas/canvas-particles/src/ParticleContainer.ts
```diff
@@ -1,4 +1,6 @@
 import { ParticleContainer } from '@pixi/particles';
+import { Sprite } from '@pixi/sprite';
```
bigtimebuddy, line 2:
```suggestion
import type { Sprite } from '@pixi/sprite';
```

PR #6630, opened 2020-05-19
Convert mixin-cache-as-bitmap to TS

SerG-Y:
#### Description of change
Converted mixin-cache-as-bitmap to TS

##### Pre-Merge Checklist


SerG-Y:
Hi guys,
is it okay to cast some variables to `any` to be able to access some private fields (ex. `_render`, `_tempDisplayObjectParent `, `_boundsID`)?

@bigtimebuddy @ivanpopelyshev @themoonrat

bigtimebuddy:
Let's try to avoid the any casts. What's better is something like this so that we can maintain the internal properties in once place and we get the correct type.

```typescript
interface RendererInternals extends Renderer {
  _tempDisplayObjectParent: Container;
}
this.parent = (renderer as RendererInternals)._tempDisplayObjectParent;
```

SerG-Y:
@bigtimebuddy it's not possible in this case because typescript doesn't allow to have the same property in one class private and public in another class/interface
image

bigtimebuddy:
Then maybe don’t extend.

ivanpopelyshev:
`_tempDisplayObjectParent` - just `as any` it, I'm gonna remove that object in another PR soon.

PR #6640, opened 2020-05-22
Improve API usage of BitmapFont.from

bigtimebuddy:
## Overview

This change is mostly about improving and streamlining ergonomics using the new `BitmapFont.from` API (#6600).

Also, did some minor code organization and documentation fixes.

**Before**
Previously you optionally defined the font name and no chars were set by default.

```js
const font = BitmapFont.from({ chars: BitmapFont.ALPHANUMERIC }, {
    fill: "#808080",
    fontFamily: "Roboto-Black",
    fontSize: 33,
});
BitmapFont.available.TitleFont = font;
```

**After**
Now, font is installed, name is required, and chars are optional (defaults to ALPHANUMERIC).
```js
BitmapFont.from("TitleFont", {
    fill: "#808080",
    fontFamily: "Roboto-Black",
    fontSize: 33,
});
```

Now, if you want to change something on the fly, regenerating the font is as easy as calling `BitmapFont.from` again with some new style options.

```js
const style = new TextStyle({
    fill: "#808080",
    fontFamily: "Roboto-Black",
    fontSize: 33,
});
BitmapFont.from("TitleFont", style);
style.fontSize = 14;
BitmapFont.from("TitleFont", style); // rerenders font!
```

### Demo

https://jsfiddle.net/bigtimebuddy/wt9szqp6/

### Changed

* ~Both these signatures are now supported `BitmapFont.from(name, style)` and `BitmapFont.from(options, style)`~ Changed API signature to be `BitmapFont.from(name, style, options)`
* Destroying a BitmapFont now also destroys the page textures, not just the glyph textures
* Field `name` is required for `BitmapFont.from` for ease of use.
* When calling `BitmapFont.from`, it will override any existing font by default with the same name, making it easier to regenerate the font.
* Exposed default options via `BitmapFont.defaultOptions` in case users want to change this globally.
* Moved utilities `drawGlyph` and `processCharData` to their own utilities. Still need a cleanup on `generateFillStyle` and `drawGlyph` with Text. Probably better for a follow-up PR.


packages/text-bitmap/test/BitmapFont.js
```diff
@@ -35,16 +56,9 @@ describe('BitmapFont', function ()
             expect(fontRes2.chars[id].texture.baseTexture.resolution).to.equal(2);
         });
 
-        it('should throw an error when name collision is detected in bitmap font registry', function ()
```
SukantPal, line 38: Maybe we could change this test to detect that the font was "replaced" instead of deleting it.

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -85,6 +150,15 @@ export class BitmapFont
...
         this.chars = {};
 
+        /**
+         * The map of base page textures (i.e., sheets of glyphs).
+         *
+         * @member {object}
+         * @readonly
+         * @private
+         */
+        this.pagesTextures = pagesTextures;
```
SukantPal, line 160: Since page <-> texture have a one-to-one meaning, I think it should be `pageTextures`. Just a thought - https://softwareengineering.stackexchange.com/a/344761/307206

bigtimebuddy: Sure will rename.

packages/text-bitmap/test/BitmapFont.js
bigtimebuddy: Done.

bigtimebuddy:
One other thing I discovered was the use of Array.prototype.find which is not supported in IE will need to replace that with something more compatible.

packages/text-bitmap/src/utils/drawGlyph.ts
```diff
@@ -0,0 +1,90 @@
+import { generateFillStyle } from './generateFillStyle';
+import { hex2rgb, string2hex } from '@pixi/utils';
+import type { TextMetrics, TextStyle } from '@pixi/text';
+
+// TODO: Prevent code duplication b/w drawGlyph & Text#updateText
```
GoodBoyDigital, line 5: oh yes! this is great idea!
we could create a cache that is mapped by style :D
love the idea of storing standard fonts as B/W and then tinting them! 👍

bigtimebuddy:
Good shout @themoonrate. Fixed.

PR #6641, opened 2020-05-24
BitmapText optimisation

GoodBoyDigital:
##### Description of change

As BitmapFont has now become much more accessible to our users thanks to recent efforts from the team, figured we could do with doing a bit of an optimisation pass for how bitmap fonts are rendering!

Main thing is that instead of rendering each letter as a sprite, we now render them as a mesh. This has a few benefits straight out of the box!

- lower memory footprint as we don't needs loads of sprites anymore, just a mesh or two
- faster creation `updateText`
    - pooling objects in this function.
    - including the meshes.
    - mesh buffers only reallocated if size grows.
- faster runtime `render`
    - as the text is either batched automatically if small. 
    - OR if its massive it is rendered in a single draw call (super fast, as no per character update is required)
    - tinting no longer required an `updateText` call as we leverage the mesh tint functionality.
- Ensure support for multipage fonts still works (made things a bit trickier :P)
- Added a couple of interfaces to some internal BitmapText bits.
- No surface API changes, all under the hood stuff.
- Character limit is about 11,000 - which is a lot of text :)

##### Pre-Merge Checklist


SukantPal:
@GoodBoyDigital I see Matt's commits in this PR, maybe you need to rebase it?

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -301,53 +331,199 @@ export class BitmapText extends Container
...
-            else
+
+            pagesMeshData[baseTextureUid].total++;
+        }
+
+        const currentMeshes = this._currentMeshes;
+
+        for (let i = 0; i < currentMeshes.length; i++)
+        {
+            if (meshesToAdd.indexOf(currentMeshes[i]) === -1)
```
SukantPal, line 391: `pagesMeshData` was never saved (i.e. was initialized with `{}`), so I'm having trouble seeing how `meshesToAdd` and `currentMeshes` having something in common.

SukantPal: If this is correct, maybe some inline comments on how it works could be beneficial.

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -301,53 +331,199 @@ export class BitmapText extends Container
...
-                c = new Sprite(chars[i].texture);
-                c.roundPixels = this.roundPixels;
-                this._glyphs.push(c);
+                this.removeChild(currentMeshes[i]);
             }
+        }
+
+        for (let i = 0; i < meshesToAdd.length; i++)
+        {
+            if (meshesToAdd[i].parent !== this)
```
SukantPal, line 399: Same here. `pagesMeshData` wasn't saved (initialized with `{}` inside `updateText`), hence all mesh-data (for each base-texture) was created again, so how can any mesh inside `meshesToAdd` have been a child of this BitmapText already?

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -301,53 +331,199 @@ export class BitmapText extends Container
...
+
+            vertexBuffer.update();
+            textureBuffer.update();
+            indexBuffer.update();
+        }
+
+        // return back to the pool..
+        for (const i in pagesMeshData)
+        {
+            pageMeshDataPool.push(pagesMeshData[i]);
```
SukantPal, line 520: If you return a mesh-data to the (shared) pool, then another BitmapText could snatch it in its updateText. Am I wrong here?i 

Let's say you have bitmap-texts B1 and B2:

1. B1.updateText() --> it returns child meshes's data to pagesMeshDataPool
2. B2.updateText() --> it grabs from pagesMeshDataPool and then `addChild` to itself

---
Again, inline comments here how it works would be helpful.

packages/text-bitmap/src/BitmapFont.ts
```diff
@@ -335,6 +335,7 @@ export class BitmapFont
      *
      * const title = new PIXI.BitmapText("This is the title", { font: "TitleFont" });
      */
+
```
SukantPal, line 338: I think this new-line wasn't intended (and shouldn't be there anyway).

bigtimebuddy:
This is awesome @GoodBoyDigital! Super excited about our new and improved BitmapText.

Some tests might need to be looked at, these are failing:

1) PIXI.BitmapText
       should render text even if there are unsupported characters:
2) PIXI.BitmapText
       should support font without page reference:

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -1,29 +1,49 @@
-import { Container } from '@pixi/display';
 import { ObservablePoint, Point } from '@pixi/math';
 import { settings } from '@pixi/settings';
-import { Sprite } from '@pixi/sprite';
+import { Mesh, MeshGeometry, MeshMaterial } from '@pixi/mesh';
```
bigtimebuddy, line 3: Could you make sure you tweak the dependencies in the package.json of text-bitmap to reflect this change?

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -1,29 +1,49 @@
...
  * The primary advantage of this class over Text is that all of your textures are pre-generated and loading,
  * meaning that rendering is fast, and changing text has no performance implications.
  *
- * The primary disadvantage is that you need to preload the bitmap font assets, and thus the styling is set in stone.
  * Supporting character sets other than latin, such as CJK languages, may be impractical due to the number of characters.
  *
  * To split a line you can use '\n', '\r' or '\r\n' in your string.
  *
- * You can generate the fnt files using
+ * Pixi can can auto generate the fonts for you or you can generate the fnt files using
```
bigtimebuddy, line 46:
```suggestion
 * PixiJS can auto-generate fonts on-the-fly using BitmapFont or use fnt files provided by:
```

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -38,13 +58,13 @@ import type { BitmapFontData } from './BitmapFontData';
  * @extends PIXI.Container
  * @memberof PIXI
  */
+
```
bigtimebuddy, line 61:
```suggestion
```

themoonrat:
Question: How does this effect the Canvas renderer?

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -301,53 +331,199 @@ export class BitmapText extends Container
...
+                let pageMeshData = pageMeshDataPool.pop();
+
+                if (!pageMeshData)
+                {
+                    const geometry = new MeshGeometry();
+                    const material = new MeshMaterial(Texture.EMPTY);
+
+                    const mesh = new Mesh(geometry, material);
+
+                    pageMeshData = {
```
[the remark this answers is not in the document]
GoodBoyDigital: heya! The main reason for declaring the whole object is so that we don't change its structure behind the scenes immediately in the next few lines. I think this keeps it a tad more js compiler efficient :D

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -241,12 +257,26 @@ export class BitmapText extends Container
...
 
-            chars.push({
-                texture: charData.texture,
-                line,
-                charCode,
-                position: new Point(pos.x + charData.xOffset + (this._letterSpacing / 2), pos.y + charData.yOffset),
-            });
+            let charRenderData = charRenderDataPool.pop();
+
+            if (!charRenderData)
```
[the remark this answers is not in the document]
GoodBoyDigital: same as above, I think declaring the full structure is better for comiler. Will adjust to use the `||` thought, thats much cleaner 👍

packages/text-bitmap/src/BitmapText.ts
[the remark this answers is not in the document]
SukantPal: You mean its faster in the browser?

packages/text-bitmap/src/BitmapText.ts
GoodBoyDigital: Good question! the thing we need to keep track of is the mesh's as they need to be added removed (potentially!) on a text change.

For that the `meshesToAdd` is where all the meshes we are that must be added to the stage are added. We then compare that with `_currentMeshes` and remove ones that shouldn't be there, whilst adding ones that should.

Then after we have done that, `meshesToAdd` becomes the new `_currentMeshes` list.

Will add some comments to make a bit clearer 👍

packages/text-bitmap/src/BitmapText.ts
GoodBoyDigital: haha! great catch, definitely something fishy there! Will tweak a bit..

packages/text-bitmap/src/BitmapText.ts
GoodBoyDigital: yeah, something smelly! going to maybe store pageMeshData instead of Mesh and we should be good..

GoodBoyDigital:
Hey @SukantPal , thanks for the feedback man! Push some changes. Let me know if its all cool 👍

GoodBoyDigital:
@bigtimebuddy , will look at those tests now..

GoodBoyDigital:
> Question: How does this effect the Canvas renderer?

Good question. i will test, it should work but I guess would technically be slower perf than the old method. If it turns out to be a deal breaker and we should keep the previous method around and move it into Canvas Renderer.

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -241,12 +256,21 @@ export class BitmapText extends Container
...
+                position: new Point(),
+            };
+
+            charRenderData.texture = charData.texture;
+            charRenderData.line = line;
+            charRenderData.charCode = charCode;
+            charRenderData.position.x = pos.x + charData.xOffset + (this._letterSpacing / 2);
+            charRenderData.position.y = pos.y + charData.yOffset;
+
+            chars.push(charRenderData);
```
SukantPal, line 272: What do ya think about optimizing this one as well?

Instead of pushing each character to the (initially) zero-length array, do something like:

```js
// textLength is a _good_ heuristic of the actual no. of chars
const chars: CharRenderData = new Array(textLength);
let charsRendered = 0;

for (let i = 0; i < textLength; i++) {
     if (<skipCondition>) {
         continue;
     }  

     chars[charsRendered] = charREnderData;
     ++charsRenderered;
}

// Make sure undefined slots aren't at the end of the array, if needed
chars.length = charsRendered;
```

packages/text-bitmap/test/BitmapText.js
```diff
@@ -71,7 +71,7 @@ describe('PIXI.BitmapText', function ()
             font: this.font.font,
         });
 
-        expect(text.children.length).to.equal(4);
+        expect(text._activePagesMeshData[0].total).to.equal(4);
     });
```
SukantPal, line 75: I think a test where you sinon-spy on the addChild/removeChild and verify it works as intended would finish this one solidly.

For example:
* BitmapFont(ALPHANUMERIC, fontSize: 8) (make sure it outputs only one page with small enough font size)
* BitmapText("HelloWorld")
* updateText() -> addChild 1 times (because only one page, right?)
* updateText() again -> addChild 0 times
* change text to "dlorWolleH"
* updateText() -> addChild/removeChild 0 times
* change text to ""
* updateText() -> removeChild 1 time

GoodBoyDigital: Good shout! Will add this in later today 👍 
what is sinon-spy? (I'm test newbie, looking to improve!)

SukantPal: See https://sinonjs.org/releases/v9.0.2/spies/ (sinon.spy())

We are using sinon-spy in our tests already: https://github.com/pixijs/pixi.js/blob/2daa8749845e23525f10607f402c434390bf4b43/packages/ticker/test/Ticker.js#L99 is an example.

bigtimebuddy:
Seems to still render fine on canvas: 
https://pixijs.download/feature/turbo-bitmap-fonts/pixi-legacy.min.js
https://jsfiddle.net/bigtimebuddy/hzrx6p47/

packages/text-bitmap/test/BitmapText.js
GoodBoyDigital: thanks, will take a look!

GoodBoyDigital:
Ok, good to go now! Final updates and new unit test

- fixed batching issue
- added unit test for adding / removing children
- fixed issue found when making unit test :)
- don't call updateText on creation

packages/text-bitmap/src/BitmapText.ts
SukantPal: @GoodBoyDigital I'm leaving this one on your will. It would be cool to do this optimization but not necessarily required.

SukantPal: ^^^

PR #6643, opened 2020-05-24
Change the default BitmapFont texture size

bigtimebuddy:
Follow-up to #6600 

The default size for generating BitmapFonts was 2048x2048, which is a _very_ large image. Even if most of that is empty space, it's a big memory footprint for the GPU. I turned it down to something a little more reasonable: 512x512. 

@SukantPal said that he might work on a heuristic to determine the optimal size based on the font size and character count. Until then, let's turn it down a bit.

Even if not all glyphs fit, they are spread across multiple pages, for instance: 
https://jsfiddle.net/bigtimebuddy/et2boa51/


GoodBoyDigital:
yeah, good shout! 

we could get an accurate enough size by  `biggest character size * number of chars` to render!

PR #6645, opened 2020-05-25
Fix documented parameters

SukantPal:
##### Description of change

This fixes cases where the no. of documented parameters is greater than those actually taken.

##### Pre-Merge Checklist


SukantPal:
@bigtimebuddy I didn't pay attention to that. I'll fix it.

PR #6649, opened 2020-05-26
Fix docs for graphicsUtils namespace and members

### LINKED ISSUE #6648: PIXI.Star returns an error on using.
### Expected Behavior
The star is drawn.
### Current Behavior
The star is not shown and return this error : Uncaught TypeError: PIXI.Star is not a constructor.
### Possible Solution
Re-implement the PIXI.Star class. 
### Steps to Reproduce
Put new PIXI.Star(0,0,50,25); and it will result an error.
### Environment

- **`pixi.js` version**: _5.2.4_
- **Browser & Version**: _Chrome 81.0.4044.138_
- **Running Example**:  https://www.pixiplayground.com/#/edit/FEFF3OlTl28XK0vNIPFBG

bigtimebuddy:
Fixes #6648 - Star polygon class is part of `graphicsUtils` namespace not root.

Also, members of this class documented incorrectly. Fixed the names.

```
exports.BATCH_POOL
exports.DRAW_CALL_POOL
exports.FILL_COMMANDS
```

Current dev (broken)
http://pixijs.download/dev/docs/PIXI.graphicsUtils.html

This PR (fixed)
http://pixijs.download/dev-graphics-docs/docs/PIXI.graphicsUtils.html

### DIFF
--- a/packages/graphics/src/utils/index.ts
+++ b/packages/graphics/src/utils/index.ts
@@ -56,6 +56,6 @@ export const BATCH_POOL: Array<BatchPart> = [];
  * Draw call pool, stores unused draw calls for preventing allocations.
  *
  * @memberof PIXI.graphicsUtils
- * @name {Array<PIXI.BatchDrawCall>} DRAW_CALL_POOL
+ * @member {Array<PIXI.BatchDrawCall>} DRAW_CALL_POOL
  */
 export const DRAW_CALL_POOL: Array<BatchDrawCall> = [];
--- a/packages/graphics/src/utils/index.ts
+++ b/packages/graphics/src/utils/index.ts
@@ -1,7 +1,8 @@
 /**
  * Generalized convenience utilities for Graphics.
  *
- * @namespace PIXI.graphicsUtils
+ * @namespace graphicsUtils
+ * @memberof PIXI
  */
 
 import { buildPoly } from './buildPoly';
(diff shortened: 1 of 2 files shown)


packages/graphics/src/utils/index.ts
```diff
@@ -1,7 +1,8 @@
 /**
  * Generalized convenience utilities for Graphics.
  *
- * @namespace PIXI.graphicsUtils
+ * @namespace graphicsUtils
+ * @memberof PIXI
```
SukantPal, line 5: This is a better approach than using `@namespace PIXI.graphicsUtils`!

packages/graphics/src/utils/index.ts
```diff
@@ -47,14 +48,14 @@ export const FILL_COMMANDS: Record<SHAPES, IShapeBuildCommand> = {
  * Batch pool, stores unused batches for preventing allocations.
  *
  * @memberof PIXI.graphicsUtils
- * @type {Array<PIXI.graphicsUtils.BatchPart>}
+ * @member {Array<PIXI.graphicsUtils.BatchPart>} BATCH_POOL
```
SukantPal, line 51: I think renaming the documentation here is unnecessary but works :)

bigtimebuddy: JSDoc has issues with ES6 exports 😦 This was a case where it inferred the name as `exports.BATCH_POOL` instead of just `BATCH_POOL`. Having a better doc parser would help or, what I've seen in other places as a good stylistic choice, is don't do `export class`, `export let`, etc as inline, and do your exports at the bottom of the document `export { BATCH_POOL }`. It's easier to see exports all in one place.

PR #6656, opened 2020-05-27
convert interaction package to TypeScript

Zyie:
##### Description of change

So the previous attempt at this, #6474, seems to have stalled.

It had a few issues where it was rewriting parts of the code base so i have done a more straightforward conversion.

Didn't run into many issues so was a fairly easy conversion.

##### Pre-Merge Checklist


packages/interaction/src/interactiveTarget.ts
```diff
@@ -1,3 +1,56 @@
...
+
+export interface IHitArea {
+    contains(x: number, y: number): boolean;
+}
+
+export interface InteractiveTarget {
+    interactive: boolean;
+    interactiveChildren: boolean;
+    hitArea: IHitArea;
+    cursor: Cursor;
```
andrewstart, line 48: Cursor can actually be any string - it is a dictionary key or any valid CSS value for `cursor` (including the url() stuff). I really need to set up an example of all the capabilities.

PR #6658, opened 2020-05-27
Convert canvas-sprite to TypeScript

SerG-Y:
##### Description of change
Converted canvas-sprite to TypeScript

##### Pre-Merge Checklist


packages/canvas/canvas-sprite/src/CanvasSpriteRenderer.ts
```diff
@@ -131,9 +135,9 @@ export class CanvasSpriteRenderer
 
         if (sprite.tint !== 0xFFFFFF)
         {
-            if (sprite._cachedTint !== sprite.tint || sprite._tintedCanvas.tintId !== sprite._texture._updateID)
+            if ((sprite as any)._cachedTint !== sprite.tint || sprite._tintedCanvas.tintId !== sprite._texture._updateID)
```
bigtimebuddy, line 138: In these cases, I think I'd rather make `_cachedTint` an internal public property (still private from the docs). Makes this a lot less brittle.

packages/canvas/canvas-sprite/src/CanvasSpriteRenderer.ts
```diff
@@ -76,9 +80,9 @@ export class CanvasSpriteRenderer
         // If smoothingEnabled is supported and we need to change the smoothing property for sprite texture
         const smoothingEnabled = texture.baseTexture.scaleMode === SCALE_MODES.LINEAR;
 
-        if (renderer.smoothProperty && renderer.context[renderer.smoothProperty] !== smoothingEnabled)
+        if (renderer.smoothProperty && (renderer.context as any)[renderer.smoothProperty] !== smoothingEnabled)
```
bigtimebuddy, line 83: What is `context` here? Can we avoid any?

Zyie: `context` is the `ContextSystem`. I believe it is set to `any` as we are trying to access a property dynamically. I never found a good way of doing this is TypeScript. 

This sort of thing is littered around the code base

If anyone knows how to fix this sort of error then we can tidy up all occurrences of this
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'CanvasRenderingContext2D'.
  No index signature with a parameter of type 'string' was found on type 'CanvasRenderingContext2D'.
```

packages/canvas/canvas-sprite/src/CanvasSpriteRenderer.ts
SerG-Y: @bigtimebuddy how do you suggest to do it? Because it's not possible to extend Sprite (incorrectly extends interface Sprite property _cachedTint is protected and can't be public in another interface).

packages/canvas/canvas-sprite/src/CanvasSpriteRenderer.ts
bigtimebuddy: Instead of using a string you and do `keyof SomeType` which makes it possible to index like this.

SerG-Y: @bigtimebuddy Cool idea, thanks, but in the next case I need to cast `context` to `any` 

```ts
(context as any)[renderer.smoothProperty as imageSmoothingKey] = smoothingEnabled;
```
because in another way I will get an error `TS2540: Cannot assign to 'canvas' because it is a read-only property.`

packages/canvas/canvas-sprite/src/CanvasSpriteRenderer.ts
bigtimebuddy: Looks like @Zyie did this already in #6665, if you want to see for reference. Given the choice of removing the access modifier (i.e., `protected`) or casting as `any`, it's much better to do the former.

packages/canvas/canvas-sprite/src/CanvasSpriteRenderer.ts
bigtimebuddy: There are [TypeScript utilities](https://www.typescriptlang.org/docs/handbook/utility-types.html) for dealing with this. So you could do something like `keyof Omit<SomeClass, 'canvas'>` to exclude certain properties (blacklist)  or `keyof Pick<SomeClass, 'smoothing'>` (whitelist).

SerG-Y: @bigtimebuddy fixed, thanks a lot for your help)

PR #6679, opened 2020-06-04
Add types for packages

d-shorkin:
##### Pre-Merge Checklist


bigtimebuddy:
The best approach here is to generate typings from source which we are working on. We don’t publish src folders so this won’t work when published. Thanks for your contribution, but am going to close this.

Issue #6670, opened 2020-05-28
Projection system's semantics are wrong


SukantPal:
There are two stages of projection in WebGL:

1. **model** to **clip** space: The sourceFrame is transformed into a rectangle (-1, -1, 1, 1)

2. **clip** to **screen** space: The clip space rectangle is then transformed onto the viewport/canvas frame. This _is done by the WebGL runtime_.

Take this example: https://codepen.io/sukantpal/pen/OJyeOBx?editors=0010

I am changing the width/height of the source-frame, yet it has no effect on the scene. This is because we are calculating the projection scale _based on the destination frame_ instead of the source-frame:

https://github.com/pixijs/pixi.js/blob/616e6a219ca86ca23b65db8f746e94809782acc6/packages/core/src/projection/ProjectionSystem.ts#L114

* https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_model_view_projection: "The projection matrix is used to convert world space coordinates into clip space coordinates" - that clearly means it is unrelated to the destination frame.

### Expected Behavior

Changing the source-frame in the projection system should change the scale of the projection matrix.

### Current Behavior

Nothing happens.

### Possible Solution

I think the expected behavior for this would be to:

* calculate the scale based off of the source-frame itself

* and then update `framebuffer.setViewport` with the destination frame

PR #6683, opened 2020-06-05
Use only sourceFrame to calculate projectionMatrix

SukantPal:
#6670 

Screen Shot 2020-06-05 at 5 25 02 PM

The destination frame should have nothing to do in the projection-matrix.

##### Pre-Merge Checklist


ivanpopelyshev:
We really need one invariant in that case.

viewport size is a multiply to resolution and sourceFrame x,y is always in the corner of pixel. 

If we dont have that invariant, filters might go wrong

SukantPal:
@ivanpopelyshev I figured out what will have to change. See, I'm using this patch at work and our filters did have to change. I made a modified filter-system & after debugging just now - I found the problem.

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/filters/FilterSystem.ts#L301

That line needs to _explicity_ set the destination frame. Otherwise, the patched projection-system will project the source-frame onto all the of texture.

**Fix-1 in Filter System**:
Screen Shot 2020-06-07 at 7 57 02 PM

```js
this.renderer.renderTexture.bind(filterTexture, 
    filterTexture ? filterTexture.filterFrame : null,
    filterTexture && filterTexture.filterFrame ? new Rectangle(0, 0, filterTexture.filterFrame.width, filterTexture.filterFrame.height) : null);
```

**Fix-2 in RenderTexture System**:

This fix would **also bring a comment in the source code to consistency with the actual code**.

**The default destinationFrame in the render-texture system is the whole render-texture; we can change that to a rectangle of the same size as the sourceFrame at (0, 0), instead**

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/renderTexture/RenderTextureSystem.ts#L96

```js
tempRect.width = sourceFrame ? sourceFrame.width : baseTexture.realWidth;
tempRect.height = sourceFrame ? sourceFrame.height : baseTexture.realHeight;
```

Here, the documentation says the default destinationFrame has the same size as the sourceFrame. **Clearly, this is not currently true in the code. This patch merely makes the documentation true**

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/renderTexture/RenderTextureSystem.ts#L78

The net result would be same as the diagram shown previously.
___
@ivanpopelyshev I'm not sure which fix would be the least disrupting. This is something you have more experience with in the community. _BUT_, I think we **need to get this right once and for all**.

SukantPal:
@ivanpopelyshev I figured out what will have to change. See, I'm using this patch at work and our filters did have to change. I made a modified filter-system & after debugging just now - I found the problem.

1. (uncomment the comment already there)

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/filters/FilterSystem.ts#L193

2. (see fix below)

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/filters/FilterSystem.ts#L301

That line needs to _explicity_ set the destination frame. Otherwise, the patched projection-system will project the source-frame onto all the of texture.

**Fix in Filter System**:
Screen Shot 2020-06-07 at 8 18 56 PM

```js
this.renderer.renderTexture.bind(filterTexture, 
    filterTexture ? filterTexture.filterFrame : null,
    filterTexture && filterTexture.filterFrame ? new Rectangle(0, 0, filterTexture.filterFrame.width, filterTexture.filterFrame.height) : null);
```

SukantPal:
@ivanpopelyshev @bigtimebuddy OR, an alternative fix for this is in the render-texture system.

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/renderTexture/RenderTextureSystem.ts#L78

That comment says that the default destinationFrame has a size equal to that of the sourceFrame, but **that is clearly not the case in the code**:

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/renderTexture/RenderTextureSystem.ts#L96

https://github.com/pixijs/pixi.js/blob/cf794b6432bc71bd2d77f1a6f4d855bc6da3cf5d/packages/core/src/renderTexture/RenderTextureSystem.ts#L120

**The default destinationFrame size is that of the render-texture or screen actually**. 

The fix would be to change that to:

```js
tempRect.width = sourceFrame ? sourceFrame.width : baseTexture.realWidth;
tempRect.height = sourceFrame ? sourceFrame.height : baseTexture.realHeight;
```

This wouldn't change the documented behavior. It would actually correct the implementation to behave as it is documented.

___
I think this is the better fix. That's because it would fix the problem for everyone, not just the filter system.

GoodBoyDigital:
hey @SukantPal ! this looks cool.

I think my original use case was the following:

```
const baseRenderTexture = new BaseRenderTexture((width:1024,height:1024});

const tile1 = new RenderTexture(baseRenderTexture, new Rectangle(512, 512, 512, 512));

const g = new Graphics().beginFill(0xFF0000).drawRect(0,0,512, 512)

renderer.render(g, tile1)
```

this.would end up drawing the rectangle in the bottom left of the base texture iyswim?

If it still works like that, im happy :D

SukantPal:
@SukantPal This PR should be doubly good for you then - the current `dev` wouldn't have done that for you. That's because the current render-texture system's default **destination frame is not the render-texture's frame**.

https://codepen.io/sukantpal/pen/PoZjpQw?editors=1010

### Changes/fixes in this PR

+ the render-texture system will use the render-texture's frame as the default destination frame
+ the projection-system will scale the source-frame if the destination-frame has a different width/height

### 1

In the example pen, if you switch to the `dev` branch the `graphics` rectangle **will not render to the center of the texture, rather it will take all of the texture's space**.

### 2

The `graphics` rectangle is 512x512, but the render-texture's frame is 256x256.

* In the current branch, the whole graphics will not be rendered. Only the top-left corner of the graphics will be rendered because the projection system doesn't correctly scale it down. (all three color bands won't be visible)

* In the fix/projection-system branch, all three colors will be visible since the projection-matrix **will have a x,y scale of 1/512 instead of 1/256**. This the correct behaviour, since user expects **all of the source-frame to be rendered only in all of the destination-frame**.

packages/core/src/projection/ProjectionSystem.ts
```diff
@@ -98,21 +101,21 @@ export class ProjectionSystem extends System
...
      * Updates the projection matrix based on a projection frame (which is a rectangle)
      *
-     * @param {PIXI.Rectangle} destinationFrame - The destination frame.
+     * @param {PIXI.Rectangle}[destinationFrame] - The destination frame.
      * @param {PIXI.Rectangle} sourceFrame - The source frame.
      * @param {Number} resolution - Resolution
      * @param {boolean} root - If is root
      */
-    calculateProjection(destinationFrame: Rectangle, sourceFrame: Rectangle, resolution: number, root: boolean): void
+    calculateProjection(_: Rectangle, sourceFrame: Rectangle, __: number, root: boolean): void
```
bigtimebuddy, line 109: It's good to keep the names, but we can just prepend with an underscore to ignore from linting for unused params.
```suggestion
    calculateProjection(_destinationFrame: Rectangle, sourceFrame: Rectangle, _resolution: number, root: boolean): void
```

packages/core/src/renderTexture/RenderTextureSystem.ts
```diff
@@ -73,80 +76,82 @@ export class RenderTextureSystem extends System
...
      * @param {PIXI.Rectangle} [sourceFrame] - part of screen that is mapped to the renderTexture
      * @param {PIXI.Rectangle} [destinationFrame] - part of renderTexture, by default it has the same size as sourceFrame
      */
     bind(renderTexture: RenderTexture = null, sourceFrame?: Rectangle, destinationFrame?: Rectangle): void
     {
-        this.current = renderTexture;
-
         const renderer = this.renderer;
 
+        this.current = renderTexture;
```
ivanpopelyshev, line 88: When we use `current` we usualy compare if its the same as before. in RenderTextureSystem we have to bind it anyway, update uniforms and do other stuff. Add a comment about it here please :)

SukantPal: You can add a your commit here, no problem 😃.

ivanpopelyshev: OK, I'm sure you'll add it before the merge ;)

If someone later adds "optimization" here `if (current === renderTexture) return;` that'll trigger several disasters.

PR #6694, opened 2020-06-12
Formalize the Spritesheet format in TypeScript

bigtimebuddy:
We should strictly define the Spritesheet format we're expecting. Remove the `any` references in Spritesheet.

### DIFF
--- a/packages/spritesheet/src/Spritesheet.ts
+++ b/packages/spritesheet/src/Spritesheet.ts
@@ -33,7 +33,7 @@ export interface ISpritesheetFrameData {
  */
 export interface ISpritesheetData {
     frames: Dict<ISpritesheetFrameData>;
-    animations: Dict<string[]>;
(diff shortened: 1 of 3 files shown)


packages/spritesheet/src/Spritesheet.ts
```diff
@@ -3,6 +3,41 @@ import { Texture, BaseTexture } from '@pixi/core';
...
+    };
+    anchor?: IPointData;
+}
+
+/**
+ * Atlas format.
+ */
+export interface ISpritesheetData {
+    frames: Dict<ISpritesheetFrameData>;
+    animations: Dict<string[]>;
```
andrewstart, line 36: Should animations be optional? It seems like a more advanced thing.

bigtimebuddy: Good call @andrewstart, missed that one.

PR #6700, opened 2020-06-13
Deprecate JSDoc & use webdoc for documentation generation

SukantPal:
This should allow us to build on Node 12 & will help us make our documentation better in the future.

Other changes to the template include:

* Inherited members are separated in another pane
* TypeScript files are linked instead of compiled JavaScript files

### Reviewers

Please compare:

1. (webdoc) https://pixijs.download/future-webdoc-integration/docs/index.html
2. (JSDoc) https://pixijs.download/dev/docs/index.html

and find any remaining mistakes in the documentation. **There will be some things missing & I just need help finding 'em 😄**

### Why so many changes?

webdoc checks for documentation errors even in private/ignored documentation comments. I've fixed all those as well.

Fear not, there should be no change in the code.


package.json
```diff
@@ -27,7 +25,7 @@
     "build": "npm run build:dev",
     "build:prod": "cross-env NODE_ENV=production rollup -c",
     "build:dev": "rollup -c",
-    "postbuild": "npm run types",
+    "postbuild": "echo Port-TSD-JSDOC-TO_WEBDOC",
```
bigtimebuddy, line 28: Can you not break this? We still need to generate types with tsd-jsdoc.

webdoc.conf.json
```diff
@@ -3,15 +3,22 @@
         "allowUnknownTags": false
     },
     "source": {
-        "include": [
-            "dist/compiled"
-        ],
+        "includePattern": [
+            "packages/*/src/**/*.ts",
+            "bundles/pixi.js/src/*.js",
```
bigtimebuddy, line 8: No `*.js` anymore.

webdoc.conf.json
```diff
@@ -3,15 +3,22 @@
...
+            "bundles/pixi.js/src/*.ts",
+            "packages/filters/*/src/**/*.(ts|js)",
+            "packages/canvas/*/src/**/*.(ts|js)"],
```
bigtimebuddy, line 11: Same thing here, no `js` files anymore.

bigtimebuddy:
Oh darn, the problem is that Node 10 is required for JSDoc 3.5.5, which is why this is failing now.

bigtimebuddy:
@SukantPal I guess we have a few options:

1). Roll-back to Node 10 and stay on JSDoc 3.5.5 until we can generate typings without tsd-jsdoc. Would just continue to support our existing pipeline for typings.
2). Bump JSDoc to 3.6.x and *hope* that we don't muck up anything to bad. This seems ripe for more errors and harder to check for issues.
3). Hand-roll your own version of tsd-webdoc. This seems like the worst option because it just replaces a short-term solution with another short-term solution.

Option 1 probably is the most conservative approach. The follow-up PR(s) to auto-generate typings could bump to Node 12 and remove tsd-jsdoc and all the weird hacks we have to generate typings.

SukantPal:
@bigtimebuddy I've changed the build to use Node 10 & now it's working!

packages/polyfill/package-lock.json
```diff
@@ -0,0 +1,24 @@
+{
```
Zyie, line 1: Pretty sure this shouldn't be committed right? None of the other packages have a lock file

bigtimebuddy:
@SukantPal could you resolve conflicts with dependencies?

bigtimebuddy:
@SukantPal could you resolve the merge conflicts?

PR #6704, opened 2020-06-14
Formalizes the Use of IPointData

bigtimebuddy:
This was @andrewstart's (excellent) suggestion to use `IPointData` instead of `IPoint`, `Point` or `ObservablePoint`. This means we can use _more_ generic objects instead of having to construct one of the Point classes. For instance:

```typescript
const shapes = new PIXI.Graphics();
shapes.containsPoint({ x: 10, y: 10 });
```

### DIFF
--- a/packages/math/src/Matrix.ts
+++ b/packages/math/src/Matrix.ts
@@ -2,7 +2,7 @@ import { Point } from './Point';
 import { PI_2 } from './const';
 
 import type { Transform } from './Transform';
-import type { IPoint } from './IPoint';
+import type { IPointData } from './IPointData';
 
 /**
  * The PixiJS Matrix as a class makes it a lot faster.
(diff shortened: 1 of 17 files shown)


packages/math/src/Matrix.ts
```diff
@@ -168,11 +168,11 @@ export class Matrix
      * Get a new position with the current transformation applied.
      * Can be used to go from a child's coordinate space to the world coordinate space. (e.g. rendering)
      *
-     * @param {PIXI.Point} pos - The origin
+     * @param {PIXI.IPointData} pos - The origin
      * @param {PIXI.Point} [newPos] - The point that the new position is assigned to (allowed to be same as input)
      * @return {PIXI.Point} The new point, transformed through this matrix
      */
-    apply(pos: IPoint, newPos?: Point): Point
+    apply(pos: IPointData, newPos?: Point): Point
```
andrewstart, line 175: Could we change apply and applyInverse to `apply<P extends IPointData = Point>(pos: IPointData, newPos?:P):P`? That would allow us to have the same generic typing for `toLocal` and `toGlobal` on DisplayObject.

andrewstart: `InteractionData.getLocalPosition()` could also benefit from this change (and have the `globalPos` parameter changed to an `IPointData`).

PR #6710, opened 2020-06-16
Upgrade ESLint & TypeScript ESLint Plugins

bigtimebuddy:
**FINALLY**, after all those PRs, we can now upgrade ESLint. This PR fixes a few more tiny issues and add some linting exceptions where necessary. 

In general, this newer version of ESLint + plugins is much stricter and works better parsing with the latest TypeScript version. We still have a few `any` types that should get fixed eventually, but this is a big, much needed upgrade.

Everything still runs: https://pixijs.io/examples/?v=dev-typescript-linting#/demos-basic/container.js


packages/core/src/textures/TextureGCSystem.ts
```diff
@@ -139,7 +139,7 @@ export class TextureGCSystem extends System
         const tm = this.renderer.texture;
 
         // only destroy non generated textures
-        if (displayObject._texture && displayObject._texture._glRenderTargets)
+        if (displayObject._texture?._glRenderTargets)
```
bigtimebuddy, line 142: Anyone know what the heck is `_glRenderTargets`, I couldn't find a reference to it anywhere in the code. @GoodBoyDigital or @ivan?

GoodBoyDigital: displayObject._texture?.framebuffer

bigtimebuddy: :+1:

GoodBoyDigital: _glRenderTargets <-- is super old check to see if the texture was render texture. but got missed in a past refactor

packages/core/src/shader/ShaderSystem.ts
```diff
@@ -66,7 +66,7 @@ export class ShaderSystem extends System
      *
      * @private
      */
-    private systemCheck(): void
+    systemCheck(): void
```
themoonrat, line 69: Everything looks fine - my only comment is here. Why would a function be docced as private, yet not be private in typescript?

bigtimebuddy: There’s not really a concept of internal in TypeScript but if there was, this would be it.  We’ve made things public even though they are only referenced externally (eg _tingRGB). This function is overridden by the unsafe-eval package so to keep from throwing a linting error it’s easier to make it public even if it never shows up in the docs. I think there’s a way we can exclude stuff like this from typings when we get there.

Issue #6674, opened 2020-06-01
CanvasRenderer RenderTexture not cleared


Article19:
There appears to be an issue with clearing the context when rendering a container to a RenderTexture while using the CanvasRenderer (PIXI-legacy v5.2.4).

Consider this example:
https://www.pixiplayground.com/#/edit/gM7jGSOAvFhZEeJPKeGI4

Click anywhere to toggle the visibility of one of the bunnies.

Using the CanvasRenderer the context is not cleared, and the bunnies remain visible. Using the WebGL renderer (switch forceCanvas to false) this works as expected. 

I believe the issue may be here:
if (clear !== undefined ? clear : this.clearBeforeRender) {
	if (this.renderingToScreen) {
		if (this.transparent) {
			context.clearRect(0, 0, this.width, this.height);
		}
		else {
			context.fillStyle = this._backgroundColorString;
			context.fillRect(0, 0, this.width, this.height);
		}
	} // else {
	// TODO: implement background for CanvasRenderTarget or RenderTexture?
	// }
}

The 'TODO' needs to get done. It appears there is no context management for the case where we are not renderingToScreen (as I believe is the case when we are rendering a Container to a RenderTexture).

SukantPal:
@Article19 Nice catch! This is a problem specific to canvas-rendering into a `RenderTexture`.

1. The API for `render` is inconsistent b/w the WebGL & canvas renderers.

* `clear` is false by default for canvas rendering: https://github.com/pixijs/pixi.js/blob/e43cfe31b3569049e61fcf33fd5786ebe3e3583f/packages/canvas/canvas-renderer/src/CanvasRenderer.ts#L199

* `clear` is true by default for WebGL rendering: https://github.com/pixijs/pixi.js/blob/e43cfe31b3569049e61fcf33fd5786ebe3e3583f/packages/core/src/Renderer.ts#L359

If you passed `app.renderer.render(bunny, sprite.texture, true)` it _should've_ worked, except that we didn't implement clearing render-textures in the canvas renderer.

The render-texture should be cleared on this line but hasn't been implemented: 

https://github.com/pixijs/pixi.js/blob/e43cfe31b3569049e61fcf33fd5786ebe3e3583f/packages/canvas/canvas-renderer/src/CanvasRenderer.ts#L282

Article19:
My short-term solution for this was to add the following change to a custom build of pixi-legacy (right after the TODO comment):

[...]

PR #6712, opened 2020-06-16
Implement render-target clear in CanvasRenderer

SukantPal:
### Description of change
Implement render-target clearing in `CanvasRenderer#render`. This was reported in #6674.

Additional fixes:
* Fix the default value of the `clear` argument in the documentation.

### DIFF
--- a/packages/canvas/canvas-renderer/src/CanvasRenderer.ts
+++ b/packages/canvas/canvas-renderer/src/CanvasRenderer.ts
@@ -196,7 +196,7 @@ export class CanvasRenderer extends AbstractRenderer
      * @param {PIXI.DisplayObject} displayObject - The object to be rendered
      * @param {PIXI.RenderTexture} [renderTexture] - A render texture to be rendered to.
      *  If unset, it will render to the root context.
-     * @param {boolean} [clear=false] - Whether to clear the canvas before drawing
+     * @param {boolean} [clear=this.clearBeforeRender] - Whether to clear the canvas before drawing
      * @param {PIXI.Matrix} [transform] - A transformation to be applied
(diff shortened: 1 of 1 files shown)


packages/canvas/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -278,9 +278,20 @@ export class CanvasRenderer extends AbstractRenderer
...
-            // }
+            }
+            else
+            {
+                renderTexture = (renderTexture as BaseRenderTexture);
+                renderTexture._canvasRenderTarget.clear();
+
+                const clearColor = renderTexture.clearColor;
+
+                if (clearColor[0] !== 0 || clearColor[1] !== 0 || clearColor[2] !== 0 || clearColor[3] !== 0)
```
bigtimebuddy, line 289: Don't we only need to check for `clearColor[3] > 0`? We should probably support `[0, 0, 0, 1]` as a clearColor (black).

SukantPal: What do you mean "support as a clearColor"? If `clearColor[3] > 0` (true for [0, 0, 0, 1] black), then will be cleared inside the if-block.

bigtimebuddy: I was suggesting just to simplify the logic. We don't need to check for rgb values, only a to know whether to use clearColor or not.

bigtimebuddy:
Can you just create a very simple fiddle to demo that this works?

SukantPal:
@bigtimebuddy @Article19's playground https://www.pixiplayground.com/#/edit/gM7jGSOAvFhZEeJPKeGI4 now works on this branch. 

Let me know if I should create something more specific.

Issue #6702, opened 2020-06-14
Enhance  "getChildByName" feature


HusakYurii:
@SerG-Y hey!
Have a suggestion to enhance this feature.

https://github.com/pixijs/pixi.js/blob/528b44223bb96e18580e72ff94461abd20ed4857/packages/mixin-get-child-by-name/src/index.ts#L19

Currently it allows to get a child by name but not recursively.

The solution below gives more flexibility for this feature. If you consider it useful, could you add it then please?

```javascript

Container.prototype.getChildByName = function getChildByName(name, isRecursive = true) {
            const findChild = (name, children) => {
                return children.find((child) => child.name === name);
            };

            let found = findChild(name, this.children);

            if (!found && isRecursive) {
                found = this.children.reduce((_, child) => {
                    const result = child.getChildByName(name, isRecursive);
                    if(result) return result;
                }, undefined);
            }

            return found;
};

```

SukantPal:
@HusakYurii I'm concerned about the efficiency of using a recursive search to find display-object by their name.

From a standpoint of performance, the best way to for the user to do this would be to use an external hashmap.

```ts
const displayObjectRegistry: { [id: string]: PIXI.DisplayObject } = {};

function getDisplayObjectByName(name: string): ?PIXI.DisplayObject {
    return displayObjectRegistry[name];
}

Object.defineProperty("name", PIXI.DisplayObject.prototype, {
    get() {
        return this._name;
    },
    set(newName: string) {
        if (displayObjectRegistry[this._name] === this) {
            delete displayObjectRegistry[this._name];
        }

        this._name = newName;
        displayObjectRegistry[newName] = this;
    }
});

PIXI.DisplayObject.prototype.getChildByName = getDisplayObjectByName
```

Of course, there is a little difference in terms of what happens with name collisions here.

PR #6713, opened 2020-06-16
Add option for recursive search in @pixi/mixin-get-child-by-name

SukantPal:
@HusakYurii requested this feature in #6702.

##### Description of change
Implements an option for preorder recursive searches in `@pixi/mixin-get-child-by-name`

##### Pre-Merge Checklist


packages/mixin-get-child-by-name/src/index.ts
```diff
@@ -11,20 +11,36 @@ DisplayObject.prototype.name = null;
...
-    for (let i = 0; i < this.children.length; i++)
+    for (let i = 0, j = this.children.length; i < j; i++)
     {
         if (this.children[i].name === name)
         {
             return this.children[i];
         }
     }
 
+    if (isRecursive)
```
bigtimebuddy, line 32: This could be more succinct if you put the recursive check inside the first loop. Then you don’t have to have two loops.

packages/mixin-get-child-by-name/src/index.ts
```diff
@@ -11,20 +11,36 @@ DisplayObject.prototype.name = null;
...
+    {
+        for (let i = 0, j = this.children.length; i < j; i++)
+        {
+            const target = (this.children[i] as Container).getChildByName(name, true);
```
bigtimebuddy, line 36: Also, children is DisplayObject[] right? Do we need to check for Container just in case? This would fail if a child is not a Container.

packages/mixin-get-child-by-name/test/index.js
```diff
@@ -23,13 +23,27 @@ describe('PIXI.Container#getChildByName', function ()
...
+
+    it('should correctly find a indirect child by its name in recursive search', function ()
+    {
+        const obj = new DisplayObject();
+        const parent = new Container();
+        const grandParent = new Container();
+
+        obj.name = 'foo';
+        parent.addChild(obj);
+        grandParent.addChild(obj);
```
bigtimebuddy, line 45: Shouldn’t this be addChild(parent)?

bigtimebuddy: Also, I think there’s more test coverage we could be doing for this feature. For instance, if a child and a descendant both have the same name, the highest level in the hierarchy should return first.

packages/mixin-get-child-by-name/src/index.ts
```diff
@@ -11,20 +11,36 @@ DisplayObject.prototype.name = null;
 /**
  * Returns the display object in the container.
  *
+ * Recursive searches are done in a preorder traversal.
+ *
  * @method getChildByName
  * @memberof PIXI.Container#
  * @param {string} name - Instance name.
+ * @param {boolean}[isRecursive=false] - Whether to search recursively
```
bigtimebuddy, line 19: My preference is to call this “deep” instead of isRecursive.

packages/mixin-get-child-by-name/src/index.ts
HusakYurii: I assume that children are always Display Objects, otherwise an error will be thrown. Do not know if the casting is necessary there

packages/mixin-get-child-by-name/test/index.js
HusakYurii: Tests should cover few more cases:
1) If you passed a non existing name (2 scenarios with is Recursive = true/false)
2) If you passed an existing name (2 scenarios with is Recursive = true/false)
3) Always return the firs it has found
Hope it helps

packages/mixin-get-child-by-name/src/index.ts
SukantPal: "I think there’s more test coverage we could be doing for this feature. For instance, if a child and a descendant both have the same name, the highest level in the hierarchy should return first."

You need two loops for this type of traversal (preorder). First you have to check all direct children, before searching one depth below.

packages/mixin-get-child-by-name/src/index.ts
```diff
@@ -11,20 +11,43 @@ DisplayObject.prototype.name = null;
...
+
+            const target = (this.children[i] as Container).getChildByName(name, true);
+
+            if (target)
+            {
+                return target;
+            }
+        }
+    }
+
```
bigtimebuddy, line 51: What about something like this to simplify the code a little bit more?

```suggestion
    for (let i = 0, j = this.children.length; i < j; i++)
    {
        if (this.children[i].name === name)
        {
            return this.children[i];
        }

        const child = (this.children[i] as Container);

        if (!deep || !child.getChildByName)
        {
            continue;
        }

        const target = child.getChildByName(name, deep);

        if (target)
        {
            return target;
        }
    }
```

bigtimebuddy:
Thanks @SukantPal!

Issue #6722, opened 2020-06-24
Buggy thick lines rendering


broadsw0rd:
### Expected Behavior

Proper thick lines rendering like in v5.2.4. [Example](https://codepen.io/broadsw0rd/pen/ExPmJmL)

### Current Behavior

Buggy thick lines rendering. [Example](https://codepen.io/broadsw0rd/pen/ExPmJjo)

### Environment

- **`pixi.js` version**: 5.3.0
- **Browser & Version**: Chrome 83
- **OS & Version**: Windows 10
- **Running Example**: https://codepen.io/broadsw0rd/pen/ExPmJjo

bigtimebuddy:
Thanks for the examples. @SukantPal would you mind taking a look at this likely related to our 5.3.0 graphics changes.

bigtimebuddy:
Merged fix.

PR #6723, opened 2020-06-25
Remove "same-side" guard.

SukantPal:
#6722 

My math failed me in achieving this :- https://github.com/pixijs/pixi.js/pull/6545#issuecomment-613144628. I think removing that code is okay for now - too complicated.

##### Description of change
Remove the code that prevents miter points from being on the opposite side.


bigtimebuddy:
Could you find the old examples of graphics we were testing before to make sure we didn’t regress anything?

bigtimebuddy:
I changed this to go to master, will probably push with the IE11 fix (whenever that arrives) as a v5.3.1 release.

SukantPal:
@bigtimebuddy That's a regression. The original purpose of this code was to fix that. But, for some reason, it ended up being too sensitive and affected strokes when it shouldn't have.

bigtimebuddy:
Is it fixable?

SukantPal:
This would take a lot of effort. This dot-product based fix didn't work, although the math was provable. It means I missed something and it doesn't help we don't write the math on how stuff is calculated in the code. 🤷

Issue #6394, opened 2020-02-10
Graphics.setMatrix does not work for Canvas Renderer


apptaro:
### Expected Behavior
Graphics.setMatrix should work for canvas renderer

### Current Behavior
Graphics.setMatrix does not have any effect when using canvas renderer.
Only works for WebGL.

### Steps to Reproduce
https://jsfiddle.net/y5scnf3j/

### Environment

- **`pixi.js` version**: 5.2.1
- **Browser & Version**: Chrome 79
- **OS & Version**: Windows 10

eXponenta:
Improve in 5.3, less version not support this feature.

ivanpopelyshev:
Bad news, @eXponenta , its different `matrix` he's talking about :) Wanna make a PR?

PR #6737, opened 2020-07-02
Fix Graphics.setMatrix for CanvasRenderer (#6394)

JetLua:
fix: #6394 

I just added a matrix multiplication.

before: https://jsfiddle.net/y5scnf3j/
fixed: https://www.pixiplayground.com/#/edit/QTZihAP-tdEGtHnhzewVU

### DIFF
--- a/packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
+++ b/packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
@@ -104,6 +104,11 @@ export class CanvasGraphicsRenderer
             const fillColor = data.fillStyle.color | 0;
             const lineColor = data.lineStyle.color | 0;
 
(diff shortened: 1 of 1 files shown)


JetLua:
It seems not good to call `clone` every time in `render`.

bigtimebuddy:
Hmm `clone` is probably creating a lot of unnecessary garbage. Maybe recycle a temporary object instead?

JetLua:
> Hmm `clone` is probably creating a lot of unnecessary garbage. Maybe recycle a temporary object instead?

Add a private property or a local variable (similar to the following)?

https://github.com/pixijs/pixi.js/blob/86a2185488c8a384e40a98ec117c1adb75cac8c0/packages/mixin-cache-as-bitmap/src/index.ts#L9

Issue #6491, opened 2020-03-23
[BUG] this.material._renderCanvas is not a function at v5.2.1-legacy


Fxlr8:
I use a mesh with a custom shader to render a minimap in my game. When the canvas fallback happens my app crashes with this error.

### Steps to Reproduce
Here is a small demo to reproduce the problem
https://codesandbox.io/s/pixi-canvas-bug-ifhd5

### Environment

- **`pixi.js` version**: _e.g. 5.2.1-legacy_
- **Browser & Version**: _Chrome 80_

ivanpopelyshev:
OK, now I understand what you meant. 

However, I dont understand how do you think it can work at all?  I mean, you provided information. Supose you have that info and canvas2d. Which operations would you use to render your mesh?

Fxlr8:
I think it should render nothing and log a warning. This is much better than crashing the whole app.

ivanpopelyshev:
OK, then it should be easy to make PR for `canvas-mesh` module to check if material actually can render stuff. I can do it later if no one else does.

PR #6746, opened 2020-07-06
fix: Mesh crash in canvasRenderer

JetLua:
fix: #6491 

before: https://jsfiddle.net/JetLu/tz2bnmu7/
fixed: https://jsfiddle.net/JetLu/L6dzc74o/

### DIFF
--- a/packages/canvas/canvas-mesh/src/Mesh.ts
+++ b/packages/canvas/canvas-mesh/src/Mesh.ts
@@ -28,7 +28,7 @@ Mesh.prototype._renderCanvas = function _renderCanvas(renderer: CanvasRenderer):
     else if (!warned)
     {
         warned = true;
-        if (window.console) 
+        if (window.console)
         {
             console.warn('Mesh with custom shaders are not supported in CanvasRenderer.');
         }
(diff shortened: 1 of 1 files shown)


packages/canvas/canvas-mesh/src/Mesh.ts
```diff
@@ -19,7 +21,15 @@ Mesh.prototype._renderCanvas = function _renderCanvas(renderer: CanvasRenderer):
...
 
-    this.material._renderCanvas(renderer, this);
+    if (this.material._renderCanvas)
+    {
+        this.material._renderCanvas(renderer, this);
+    }
+    else if (!warned)
+    {
+        warned = true;
+        console.warn('Mesh is not supported in canvasRenderer.');
```
themoonrat, line 31: The wording isn't quite accurate enough, for me.
You see, Mesh is supported... there is a
`Mesh.prototype._renderCanvas`
after all, and on https://pixijs.io/examples/#/mesh-and-shaders/textured-mesh-basic.js you can put { forceCanvas: true } and it works.
So it needs to be a message more about the custom mesh material created not supporting the canvas renderer, rather than just 'mesh' in general

bigtimebuddy, line 31: We need a window.console check for for running in IE9. And tweaked the message a little.
```suggestion
        if (window.console) 
        {
            console.warn('Mesh with custom shaders are not supported in CanvasRenderer.');
        }
```

JetLua:
Done.

Issue #6083, opened 2019-09-10
generateCanvasTexture generates textures twice the size


naranjamecanica:
I've converted a project to PIXI v5 and since we generate a lot textures from graphics we use the canvas for a visual smoother result. 

For this reason I use the legacy build which uses the @pixi/canvas-renderer etc, which works really well, but there is one strange behaviour: the generated textures are twice the size on a retina display.

### Expected Behavior
Generate texture in size of current renderer.

### Current Behavior
Generates a texture twice the size

### Possible Solution
Set up canvas renderer with right settings? 

### Steps to Reproduce
See: https://www.pixiplayground.com/#/edit/XWQP2rDIDQgi7SO_tawZA

Or should i allways pass resolution = 1 to the `generateCanvasTexture()`

- **`pixi.js` version**: 5.1.2

ivanpopelyshev:
Yep, confirmed on dev branch: https://www.pixiplayground.com/#/edit/mzTlmuzRlCsZA6nAqjygj

bigtimebuddy:
Anyone want to check Ivan’s fix idea?

PR #6747, opened 2020-07-06
fix: generateCanvasTexture generates textures twice the size

JetLua:
fix: #6083 

bug: https://jsfiddle.net/JetLu/Lnzyx135/
fixed: https://jsfiddle.net/JetLu/3ok4n7La/


bigtimebuddy:
Thank you @JetLua

Issue #6744, opened 2020-07-05
Fixed width text renders in wrong size for one frame whenever the text is changed


CyberPhoenix90:
### Expected Behavior
Text renders at the correct size from the start
### Current Behavior
Text renders in wrong size for 1 frame then fixes itself
### Steps to Reproduce
1. Create a PIXI.Text object 
2. Give it a fixed width
3. Change the text length
### Environment

- **`pixi.js` version**: 5.3.0
- **Browser & Version**: Tested and reproduced in: Chrome 83, Firefox 77, Firefox 78, Edge 16
- **OS & Version**: Windows 10
- **Running Example**: https://www.pixiplayground.com/#/edit/Om7cxgWh-gYovRXIPy_Ml

bigtimebuddy:
Not sure what's going on here, but seems like a bug.

Calling `updateText()` immediately after setting the text fixes the problem. I created another example (below) to demonstrate the problem more clearly. The "flash" frame is actually just appending the text and the transform isn't applied until the next frame.

https://jsfiddle.net/bigtimebuddy/9fzxq4eo/

@themoonrat does this seem like a bug to you?

themoonrat:
@bigtimebuddy It's a bug from us
Tracked down to something between 
4.5.4 https://jsfiddle.net/5ydsuqe1/2/
4.5.5 https://jsfiddle.net/5ydsuqe1/3/

4.5.4 is fine, 4.5.5 isn't.

I can't see immediately glancing at the diff what could cause this - https://github.com/pixijs/pixi.js/compare/v4.5.4...v4.5.5

bigtimebuddy:
Nothing seems obvious with that diff so probably something subtle. I added a bounty.

PR #6751, opened 2020-07-07
fix: Text updateTransform cannot be triggered before Render

JetLua:
fix: #6744 

fixed: https://jsfiddle.net/JetLu/w501jpxk/
```js
label.text += '100'
app.render()
```

Only in step `2` will `Text.updateText` be triggered, but it's too late.

`Text.updateText` -> `Text.updateTexture` -> `Sprite._onTextureUpdate` -> `Transform.onChange`

`Text` update `transform` property lags behind `render`.

**This PR may be only a temporary solution.**

image

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -8,7 +8,7 @@ import { TEXT_GRADIENT } from './const';
 import { TextStyle } from './TextStyle';
 import { TextMetrics } from './TextMetrics';
 
-import type { IDestroyOptions } from '@pixi/display';
+import type { IDestroyOptions, DisplayObject } from '@pixi/display';
 import type { Renderer } from '@pixi/core';
 import type { ITextStyle } from './TextStyle';
 
@@ -394,9 +394,8 @@ export class Text extends Sprite
 
         baseTexture.setRealSize(canvas.width, canvas.height, this._resolution);
 
-        // HACK: The scale property is immediately copied to worldTranform for render
-        this.transform.worldTransform.a = this.scale.x;
-        this.transform.worldTransform.d = this.scale.y;
+        // Recursively updates transform of all objects from the root to this one
+        (this as any)._recursivePostUpdateTransform();
 
         this.dirty = false;
     }
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -147,8 +147,8 @@ export class Text extends Sprite
          */
         this._font = '';
 
-        this.text = text;
         this.style = style;
+        this.text = text;
 
         this.localStyleID = -1;
     }
@@ -164,11 +164,6 @@ export class Text extends Sprite
     public updateText(respectDirty: boolean): void
     {
         const style = this._style;
-
-        if (!style)
-        {
-            return;
-        }
 
         // check if style has changed..
         if (this.localStyleID !== style.styleID)
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -147,8 +147,8 @@ export class Text extends Sprite
          */
         this._font = '';
 
-        this.style = style;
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Could you check out the failing test?
https://github.com/pixijs/pixi.js/pull/6751/checks?check_run_id=847661289#step:7:1333

packages/text/src/Text.ts
```diff
@@ -165,6 +165,11 @@ export class Text extends Sprite
     {
         const style = this._style;
 
+        if (!style)
```
bigtimebuddy, line 168: I get the `updateText` in the `set text` method, but why the check for no style here?

JetLua: image
`this.text` triggers the `set` before `this.style`

JetLua: Maybe I can switch their places

JetLua:
> Could you check out the failing test?
> https://github.com/pixijs/pixi.js/pull/6751/checks?check_run_id=847661289#step:7:1333

I changed the content of the `unit-test` because the `width` is `1` when `text` = `\n`

https://jsfiddle.net/JetLu/xnhftc31/

bigtimebuddy:
Would it be possible to add a new unit test for this case? If we do find a better solution, would be good to prevent a regression.

JetLua:
> Would it be possible to add a new unit test for this case? If we do find a better solution, would be good to prevent a regression.

hmm, there's no need to add new unit tests. I'm just correcting that value.

```js
const label = new PIXI.Text("\n");

label.updateText() // Without this line，output: 3

console.log(label.canvas.width) // output: 1
```

bigtimebuddy:
Oh I get it! Awesome.

packages/text/src/Text.ts
```diff
@@ -713,6 +713,7 @@ export class Text extends Sprite
         }
         this._text = text;
         this.dirty = true;
+        this.updateText(true);
```
themoonrat, line 716: I am a bit wary of this change, as I think it'll bring performance issues for fast changing text.

Text in PixiJS has always been 'lazy'. So if you updated text a million times within 1 frame..... it doesn't matter, the texture for text only got updated when it comes through to render.
Now we're saying, every time you change the text, it's going to update that texture immediately.

That's a different performance profile which may negatively effect content out there - I'm thinking of high scores or numbers counting up super fast.

JetLua: It is worth considering. Another option is to call `updateTransform` manually in `updateTexture`.

JetLua:
🆕 : https://jsfiddle.net/JetLu/cpmsntq2/

ivanpopelyshev:
need _recursivePostUpdateTransform or something like that

JetLua:
> need _recursivePostUpdateTransform or something like that

3Q, updated

packages/text/src/Text.ts
```diff
@@ -394,6 +394,9 @@ export class Text extends Sprite
 
         baseTexture.setRealSize(canvas.width, canvas.height, this._resolution);
 
+        // Recursively updates transform of all objects from the root to this one
+        (this as any)._recursivePostUpdateTransform();
```
bigtimebuddy, line 398: Much better fix, however, can you please remove `as any`. Keeping it like this masks these API when we have to refactor things later.

The approach to deal with private APIs that are used internally, is to remove the access modifier (in this case `private`) from the original method. We can use documentation flagging to make sure it's not public. 

https://github.com/pixijs/pixi.js/blob/3f90364d02b11b4fef1645df2bedac3e9b2b121e/packages/display/src/DisplayObject.ts#L283

ivanpopelyshev: yes,. this really should not be private

PR #6753, opened 2020-07-08
Added support for depth mask

jnsmalm:
##### Description of change
Added support for depth mask in state. This means that we will be able to turn on/off writing to the depth buffer when needed.

##### Pre-Merge Checklist


bigtimebuddy:
Thank you for this @jnsmalm :tada:

Issue #6752, opened 2020-07-08
Blur Filter problem on latest dev version.


owendwyer:
The slots example isn't working with the latest development version.

https://pixijs.io/examples/?v=dev#/demos-advanced/slots.js

v5.3.0 works fine.

Seems to be a problem with the blur being applied to the reel.

bigtimebuddy:
@SukantPal This is probably related to #6683. That build produces the error and is the earliest thing after 5.3.0.
https://pixijs.io/examples/?v=fix/projection-system#/demos-advanced/slots.js

Would you mind debugging this?

PR #6754, opened 2020-07-08
Make filter system specify destinationFrame

SukantPal:
#6752 

In the projection system fix, I thought the filter system would work fine. Except then @GoodBoyDigital indicating that the default destination-frame should `renderTexture.frame` instead of a rectangle `(0, 0, sourceFrame.width, sourceFrame.height)`. This caused a discrepancy in the filter-system b/c it specifies a source-frame smaller than the texture; hence, the destination-frame should also be of that size, except at the position (0, 0).

##### Description of change
* The filter system specifies a destination-frame with the same size as the filter-frame/source-frame, positioned at (0, 0). It is assumed that the x,y values of `tempRect` are not modified.


SukantPal:
The offending example: https://pixijs.io/examples/?v=fix/projection-system#/demos-advanced/slots.js should now work fine.

bigtimebuddy:
@SukantPal are you good here or are you still testing?

SukantPal:
I’m done now.

ivanpopelyshev:
Just in case, I tested it on a filter that has a hacky resolution trick: 

release https://jsfiddle.net/Hackerham/qvm1kf73/24/
dev https://jsfiddle.net/Hackerham/qvm1kf73/25/

Works fine.

PR #6761, opened 2020-07-15
Developer can supply the prefix to a pixiid to textures and base textures, to help track them

themoonrat:
##### Description of change
Lets a user supply the prefix to a texture pixiId. Since this is used to then generate the cacheId, it gives the ability to track textures.

For example, I could create a texture from a canvas, and I want to track this as part of a feature.
`texture = Texture.from(canvas, { pixiIdPrefix: 'feature1' });`
I can then track the usage of this texture, and others that are also part of feature1, in terms of if it's been rendered or not, and in terms of memory used in the texture cache.

##### Pre-Merge Checklist


themoonrat:
happy to change the name, pixiIdPrefix is a bit of a mouthful, but I found anything shorted could then become confusing as to what it does in the background in other ways :/

ivanpopelyshev (approved):

bigtimebuddy:
@themoonrat could you resolve conflicts?

themoonrat:
@bigtimebuddy 👍

Issue #6759, opened 2020-07-15
Pixi 5.3.0 does not apply tint to BitmapText when using CanvasRenderer


andrew-luminal:
Hi Pixi team, absolutely love the library! We use it extensively in the visualization component of our product.

I noticed an issue when trying to upgrade to v5.3.0 from v5.2.1. After rewiring all of our `BitmapText` objects to use the new constructor format and switching to `BitmapFont.install`, everything works fine in the WebGL renderer. However, when using `CanvasRenderer`, it seems like the `tint` property is ignored for BitmapFont (we use `CanvasRenderer` to generate large downloadable images on-the-fly, and therefore depend on `pixi.js-legacy`). We use white font images generated by bmglyph so that we can easily and quickly change their color with `tint` at runtime. I created a codepen that uses the on-the-fly generation provided by `BitmapFont.from` and the problem also exists when creating the fonts that way. 

https://codepen.io/andrew-luminal/pen/oNbPJxp

### Expected Behavior

`BitmapText` tints properly in `CanvasRenderer`. Font in above codepen is black.

### Current Behavior

`BitmapText` ignores tint in `CanvasRenderer`. Font in above codepen is white.

### Possible Solution

Unclear, I tried looking through the repo to file a PR, but I had trouble tracking down where `BitmapText` gets rendered specifically on a 2d canvas.

### Steps to Reproduce

```js
const app = new PIXI.Application({
  forceCanvas: true,
  backgroundColor: 0xCCCCCC
});
document.body.appendChild(app.view);

const bf = PIXI.BitmapFont.from('Lato', {
  fontFamily: 'Lato',
  fontSize: 24,
  fill: '#FFFFFF'
}, {})
const bt = new PIXI.BitmapText('This text should be black', {
  fontName: 'Lato',
  fontSize: 24,
  tint: 0x000000
})

app.stage.addChild(bt)
```

### Environment

- **`pixi.js` version**: v5.3.0
- **Browser & Version**: Chrome 83, Safari 13.1.1, Firefox 72.0.2 
- **OS & Version**: Mac OS 10.14.6, 64-bit
- **Running Example**: https://codepen.io/andrew-luminal/pen/oNbPJxp

PR #6764, opened 2020-07-17
Fix Mesh tint for CanvasRenderer

JetLua:
fix: #6759 

fixed: https://jsfiddle.net/JetLu/y3x5s9g7/

It's assigned to @GoodBoyDigital , but I want to try to fix it.

The content of the code is actually copied from [NineSlicePlane._renderCanvas](https://github.com/pixijs/pixi.js/blob/dev/packages/canvas/canvas-mesh/src/NineSlicePlane.ts#L38).


bigtimebuddy:
Can you provide an example to verify it works?

JetLua:
> Can you provide an example to verify it works?

Here it is: https://jsfiddle.net/JetLu/y3x5s9g7/

PR #6768, opened 2020-07-20
Added in uint mappings

Mason-Chou:
##### Description of change
Added in gl command and type mapping for uint. This is needed as the Typescript compiler will error out if you have a uint uniform in your shaders. See below for the error message.

image


ivanpopelyshev:
I'm honored that you understand our WebGL bindings. Its the part where JS is the best , you cant just write those things with same amount of code on strict-typed platforms.

We also make sure that people can customize their WebGL usage, and even make direct calls, like for Texture Resources.

Btw, can you also look if we have to adjust `unsafe-eval` module? Its used for environments like WeChat when generating code is not possible. `generateUniformSync` is dynamic there

Mason-Chou:
@ivanpopelyshev Will do, probably won't get to it till tomorrow.

bigtimebuddy:
For more context on unsafe-eval, see this issue: #5463

Relevant source code:
https://github.com/pixijs/pixi.js/blob/dev/packages/unsafe-eval/src/syncUniforms.ts

Mason-Chou:
@ivanpopelyshev  @bigtimebuddy 

Just got the `unsafe-eval` package done.

bigtimebuddy:
Thank you @Mason-Chou!

PR #6773, opened 2020-07-24
Added TexturePacker's multi pack loading to SpritesheetLoader

miltoncandelero:
### Description of change
Texture Packer includes an option to create multiple spritesheets when your images don't fit into a single one. These extra spritesheets are referenced in the field `related_multi_packs` inside the meta field.
This looks for that field and when it finds it, tries to add all the related spritesheets to the loader.

Heavily based on https://github.com/rocket-ua/pixi-tps-loader/blob/master/src/ResourceLoader.js

Sorry, I don't know how to write tests. Some good tests would be: Make sure it creates the URL correctly and make sure it doesn't try to add the same URL twice.

### Pre-Merge Checklist


miltoncandelero:
Ok, then tests that are needed:

Check if the URL creation for the child resource was correct
Check if the child resource was created
Check that no loop in the loading can exist (do not add an already existing resource)
Check that no child resource is added if the atlas doesn't have the field
Check that no child resource is created if the field is malformed (e.g. if the field is not an array of strings)
Check that no child resource is created if the field references files that are not json

I will try to understand how the test framework works and make them 👍

miltoncandelero:
I am back and I bring tests!

## should not add itself via multipack
multipack atlases are circular, they all reference each other. This test checks that you don't add the same atlas twice.

## should create multipack resources when related_multi_packs field is an array of strings
this is the one that means "it works". It checks if it creates a child resource with the other atlas

## should not create multipack resources when related_multi_packs field is missing or the wrong type
If the atlas was not a multipack, the behaviour remains unchanged. This also accounts for rogue values in the `related_multi_packs` field like it being of the wrong type or being an array of the wrong type;

## should build the multipack URL
this is the most confusing one. The URLs are created with `url.resolve` so it becomes pretty much a test on that.

---

All tests are passing along with the linter 👍 
Tell me if you think we need some other test

packages/spritesheet/src/SpritesheetLoader.ts
```diff
@@ -53,6 +53,39 @@ export class SpritesheetLoader
             return;
         }
 
+        // Check and add the multi atlas
+        // Heavily influenced and based on https://github.com/rocket-ua/pixi-tps-loader/blob/master/src/ResourceLoader.js
+        // eslint-disable-next-line camelcase
+        if (resource.data?.meta?.related_multi_packs && Array.isArray(resource.data?.meta?.related_multi_packs))
+        {
+            for (const item of resource.data.meta.related_multi_packs)
```
bigtimebuddy, line 61: Small suggestion to make this a little less verbose.
```suggestion
        const multiPacks = resource.data?.meta?.related_multi_packs;

        if (multiPacks && Array.isArray(multiPacks))
        {
            for (const item of multiPacks)
```

packages/spritesheet/src/SpritesheetLoader.ts
```diff
@@ -53,6 +53,39 @@ export class SpritesheetLoader
...
+                const itemUrl = url.resolve(resource.url.replace(loader.baseUrl, ''), item);
+
+                // Check if the file wasn't already added as multipacks are redundant
+                if (loader.resources[itemName]
+                    || Object.values(loader.resources).some((r) => url.format(url.parse(r.url)) === itemUrl))
+                {
+                    continue;
+                }
+
+                const options: any = {
```
bigtimebuddy, line 78: Can you do this bit without using `any`? Is there a typing issue you're avoiding or issue you're running into?

packages/spritesheet/test/resources/src/sheetMultipack.tps
```diff
@@ -0,0 +1,249 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bigtimebuddy, line 1: I'm not sure we should add the tps file. I don't think we're adding source files like this in other tests. Thoughts?

bigtimebuddy (approved):
Overall, just a few small comments. Thanks for this @miltoncandelero, always helpful for us to keep up with TexturePacker output variants.

packages/spritesheet/src/SpritesheetLoader.ts
```diff
@@ -56,19 +56,23 @@ export class SpritesheetLoader
         // Check and add the multi atlas
         // Heavily influenced and based on https://github.com/rocket-ua/pixi-tps-loader/blob/master/src/ResourceLoader.js
         // eslint-disable-next-line camelcase
-        if (resource.data?.meta?.related_multi_packs)
+        if (resource.data?.meta?.related_multi_packs && Array.isArray(resource.data?.meta?.related_multi_packs))
```
themoonrat, line 59: You don't need the first part of this of statement, can just go straight to the isArray part, as ?. will return undefined and fail the isArray check if it doesn't exist

themoonrat (changes requested):

packages/spritesheet/test/resources/src/sheetMultipack.tps
miltoncandelero: In that folder, there is already a tps file and the individual sprites. I just followed suit.

packages/spritesheet/src/SpritesheetLoader.ts
miltoncandelero: This was covered in the less verbose refactor from @bigtimebuddy 👍

packages/spritesheet/src/SpritesheetLoader.ts
miltoncandelero: removed the `any` type and found that I had actually a bug on `loadType`

---
Since we are here, this is the code for my loadOptions
```ts
const options = {
    crossOrigin: resource.crossOrigin,
    loadType: LoaderResource.LOAD_TYPE.XHR,
    xhrType: LoaderResource.XHR_RESPONSE_TYPE.JSON,
    parentResource: resource,
};
```

and this is the code for the regular image file for the spritesheet loadOptions
```ts
const loadOptions = {
    crossOrigin: resource.crossOrigin,
    metadata: resource.metadata.imageMetadata,
    parentResource: resource,
};
```

Should I remove the `loadType` and `xhrType` (since it should be automagically guessed) and pass through the metadata field?

bigtimebuddy: Yes, for JSON you don't need to set the load type.

packages/spritesheet/test/resources/src/sheetMultipack.tps
bigtimebuddy: Sounds good.

packages/spritesheet/src/SpritesheetLoader.ts
miltoncandelero: removed them then 👍

themoonrat (approved):

bigtimebuddy:
Thank you @miltoncandelero :tada:

ishaan-puniani:
Hello, 
May I ask you, is there planned release date for this addition? 
Thanks in advance..

bigtimebuddy:
Will be in 5.4.0. Release coming soon, hopefully before the end of the month.

emilbayes:
I think there's a bug in the multi pack loading. If I remove this line it works for me:

https://github.com/pixijs/pixi.js/blob/cf20c00ae94009cd3a190c251933958c106825ce/packages/spritesheet/src/SpritesheetLoader.ts#L70

I have 3 packs referencing each other, `0.json`, `1.json` and `2.json`. If I do:

```js
app.loader.add('0.json')
```

`0.json` is loaded twice because the file is not deduped correctly. What's the purpose of removing the extension?

bigtimebuddy:
@emilbayes thanks for letting us know! Could you please file a new bug? Also, if you wouldn't mind putting together an example of the bug, that will help us verify that it's a legit bug and confirm when it's been fixed. Appreciate your help.

LoudSilencex666:
@bigtimebuddy The bug is still the thing and causes a lot of warnings in the console. 

https://codesandbox.io/s/pixi-spritesheetloader-metadata-forked-n0d3pf?file=/src/index.js

Just check the console. And there is another thing, from new version (6.0.0) of Texture Packer only 1st atlas pack in multipacks has an animation object with all frames :(.

PR #6796, opened 2020-08-05
Compressed Textures Support

SukantPal:
# Brief 🎢 

This work will introduce support for the following compressed texture formats:

| Texture Format                                                    | Description                                 |
------------------------------------------------|---------------------------------|
| DXTC (primarily available on desktop machines) | DirectX Texture Compression (S3TC) |
| PVRTC (for iOS devices)                                     | PowerVR Texture Compression |
| ETC                                                                      | Ericssion Texture Compression |
| ATC                                                                      | |

Additionally, it implements loaders for the following texture **container** file formats:

| Extension | Description              |  Site | Notes |
| ---------- | -------------------- |---- | ------- |
| .dds          | DirectDraw Surface | https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dx-graphics-dds-pguide | support for cubemap, 3D textures not included |
| .ktx           | Khronos Texture Container | https://www.khronos.org/opengles/sdk/tools/KTX/file_format_spec/ | support for cubemap, 3D textures not included |
| .basis       | Basis Universal Supercompressed File Format | https://github.com/BinomialLLC/basis_universal | opt-in package [@pixi/basis](https://github.com/pixijs/pixi.js/tree/feature/compressed-texture-workflow/packages/basis). Fallback to uncompressed textures supported! |
| .json         | PixiJS (😛) Compressed Textures Manifest | https://github.com/bigtimebuddy | Fallback to uncompressed textures supported |

## Proof of Concept Demos 🧪 

* https://repl.it/@SukantPal/Inside-PixiJS-Compressed-Textures#workflow-demo.compressed-textures.json

## The Manifest 📎 

The [CompressedTextureLoader](https://github.com/pixijs/pixi.js/blob/feature/compressed-texture-workflow/packages/loaders/src/CompressedTextureLoader.ts) uses **JSON manifest pointing the texture files for each compression format** to 
load the correct texture file for the device. All the user needs to do is create a JSON file with the following schema:

```js
{
  // This is the ID used to add the textures into the global texture cache. After the texture has been loaded, 
  // for example, the developer can get it via: `Texture.from(<cacheID>)`
  "cacheID": "Icon.png",
 
  // List of texture files for this image
  "textures": [
     // NOTE: The "format" must be a key into PIXI.INTERNAL_FORMATS
     { "src": "Icon.dxt.ktx", "format": "COMPRESSED_RGBA_S3TC_DXT5_EXT" },

     // ... <all the other compressed textures created>

     // Fallback texture image
     { "src": "Icon.fallback.png" }
  ]
}
```

**Support for .ktx, .dds files comes built-in and those files can be used like any other loadable asset**

## Basis Universal Supercompressed Textures 🥇 

In order to use .basis files, developers must install the @pixi/basis package, initialize the transcoder, bind the loader to
the transcoder, and then load files.

The following snippet shows how to bind the `BasisLoader` to the transcoder after initializing it:

```js
// Set to true to use a web-worker for transcoding
const BASIS_WEB_WORKER = false;

if (BASIS_WEB_WORKER)
{
    BasisLoader.loadTranscoder('basis_transcoder.js', 'basis_transcoder.wasm');
}
else
{
    window.BASIS().then((basisLibrary) =>
    {
        basisLibrary.initializeBasis();

        BasisLoader.bindTranscoder(basisLibrary);
    });
}
```

**NOTES:**
* For BASIS_WEB_WORKER mode: It is expected that the server serves **basis_transcoder.js and basis_transcoder.wasm with those specific file names**.
* If you want to transcode basis files without web workers, you must add a `<script>` reference to `basis_transcoder.js`:

```html
// This also expects basis_transcoder.wasm to be served
<script src="basis_transcoder.js" />
```
* The `basis_transcoder.[js|wasm]` assets are shipped in the `@pixi/basis` package at `/node_modules/\@pixi/basis/assets/basis_transcoder.[js|wasm]`. Developers can copy these assets from there.
* **Transcoding to PVRTC fails for non-POT textures. Our loader goes into fallback mode and transcodes to an uncompressed textures - however, this severely affects performance b/c two transcoding requests are sent.**

# Shortcuts 🚀 
| Links |
| ----   |
| [BasisLoader](https://github.com/pixijs/pixi.js/blob/feature/compressed-texture-workflow/packages/basis/src/BasisLoader.ts)
| [CompressedTextureLoader](https://github.com/pixijs/pixi.js/blob/feature/compressed-texture-workflow/packages/loaders/src/CompressedTextureLoader.ts) |
| [DDSLoader](https://github.com/pixijs/pixi.js/blob/feature/compressed-texture-workflow/packages/loaders/src/DDSLoader.ts) |
 |[KTXLoader](https://github.com/pixijs/pixi.js/blob/feature/compressed-texture-workflow/packages/loaders/src/KTXLoader.ts) |
|  [BlobResource](https://github.com/pixijs/pixi.js/blob/feature/compressed-texture-workflow/packages/core/src/textures/resources/BlobResource.ts) |
| [CompressedTextureResource](https://github.com/pixijs/pixi.js/blob/feature/compressed-texture-workflow/packages/core/src/textures/resources/CompressedTextureResource.ts) |

##### Pre-Merge Checklist


package.json
```diff
@@ -74,5 +74,8 @@
     "sinon-chai": "^3.3.0",
     "tsd-jsdoc": "^2.1.3",
     "typescript": "^3.9.5"
+  },
+  "dependencies": {
+    "resource-loader": "^3.0.1"
```
bigtimebuddy, line 79: This should be removed from the top-level.

packages/basis/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/pixijs/pixi.js/issues"
+  },
+  "dependencies": {
+    "@pixi/core": "5.3.2",
+    "@pixi/constants": "5.3.2",
+    "@pixi/runner": "5.3.2",
+    "@pixi/loaders": "5.3.2",
+    "resource-loader": "3.0.1"
```
bigtimebuddy, line 41: Do not include resource-loader here. It's not used as a dependency and everything comes from @pixi/loaders.

packages/basis/src/BasisLoader.ts
```diff
@@ -0,0 +1,511 @@
...
+ *
+ * ```js
+ * // Use this if you to use the default CDN url for @pixi/basis
+ * BasisLoader.loadTranscoder();
+ *
+ * // Use this if you want to serve the transcoder on your own
+ * BasisLoader.loadTranscoder('./basis_transcoder.js', './basis_transcoder.wasm');
+ * ```
+ *
+ * NOTE: This can only be used with web-workers.
```
bigtimebuddy, line 54: Missing `@class` and `@memberof PIXI`

http://pixijs.download/feature/compressed-texture-workflow/docs/BasisLoader.html
that should be in the `PIXI` namespace.

packages/basis/src/BasisLoader.ts
```diff
@@ -0,0 +1,511 @@
...
+export class BasisLoader
+{
+    static basisBinding: BasisBinding;
+    static webGLVersion = 1;
+    static defaultRGBFormat: { basisFormat: BASIS_FORMATS, textureFormat: INTERNAL_FORMATS | TYPES };
+    static defaultRGBAFormat: { basisFormat: BASIS_FORMATS, textureFormat: INTERNAL_FORMATS | TYPES };
+    static fallbackMode = false;
+    static onTranscoderInitializedRunner = new Runner('onTranscoderInitialized');
+
+    static workerPool: TranscoderWorker[] = [];
```
bigtimebuddy, line 65: Can you please add access modifiers `public` `private` etc.

packages/basis/src/BasisLoader.ts
```diff
@@ -0,0 +1,511 @@
...
+        else
+        {
+            next();
+        }
+    }
+
+    /**
+     * Whether a transcoder is bound, either worker-based or main-thread based.
+     */
+    private static isTranscoderAvailable(): boolean
```
bigtimebuddy, line 99: This is only used in one place, just add the check to the if statement above.

packages/core/src/textures/resources/CompressedTextureResource.ts
```diff
@@ -0,0 +1,243 @@
...
+ * ```
+ *
+ * @memberof PIXI
+ */
+export class CompressedTextureResource extends BlobResource
+{
+    format: INTERNAL_FORMATS;
+    width: number;
+    height: number;
+    levels: number;
```
bigtimebuddy, line 83: Missing access modifiers?

packages/core/src/textures/resources/CompressedTextureResource.ts
```diff
@@ -0,0 +1,243 @@
...
+        this._levelBuffers = CompressedTextureResource._createLevelBuffers(
+            this.buffer.uint8View,
+            this.format,
+            this.levels,
+            4, 4, // PVRTC has 8x4 blocks in 2bpp mode
+            this.width,
+            this.height);
+    }
+
+    // Returns the key (to ContextSystem#extensions) for the WebGL extension supporting the compression format
```
bigtimebuddy, line 171: Make a proper jsdoc block here, not inline.

packages/core/src/textures/resources/CompressedTextureResource.ts
```diff
@@ -0,0 +1,243 @@
...
+        }
+        else if (format >= 0x8C92 && format <= 0x87EE)
+        {
+            return 'atc';
+        }
+
+        throw new Error('Invalid (compressed) texture format given!');
+    }
+
+    // Pre-creates buffer views for each mipmap level
```
bigtimebuddy, line 200: Make proper jsdoc-block documentation format

packages/basis/src/TranscoderWorker.ts
```diff
@@ -0,0 +1,200 @@
...
+import { ITranscodeResponse, TranscoderWorkerWrapper } from './TranscoderWorkerWrapper';
+import { BasisLoader } from './BasisLoader';
+
+/**
+ * Worker class for transcoding *.basis files in background threads.
+ *
+ * To enable asynchronous transcoding, you need to provide the URL to the basis_universal transcoding
+ * library.
+ *
+ * @memberof PIXI.BasisLoader
```
bigtimebuddy, line 11: Huh? Shouldn't this be `@memberof PIXI`?

packages/constants/src/index.ts
```diff
@@ -179,6 +179,101 @@ export enum FORMATS {
     DEPTH_STENCIL = 34041,
 }
 
+// webdoc should handle the @property list for us now (test it, & fix it if it doesn't)
+/**
+ * WebGL internal formats, including compressed texture formats provided by extensions
+ *
```
bigtimebuddy, line 185: this is not showing up in the documentation

packages/constants/src/index.ts
```diff
@@ -179,6 +179,101 @@ export enum FORMATS {
...
+    // WEBGL_compressed_texture_atc
+    COMPRESSED_RGB_ATC_WEBGL = 0x8C92,
+    COMPRESSED_RGBA_ATC_EXPLICIT_ALPHA_WEBGL = 0x8C92, // TODO: Probably a bug on the MDN site
+    COMPRESSED_RGBA_ATC_INTERPOLATED_ALPHA_WEBGL = 0x87EE,
+}
+
+/**
+ * Maps the compressed texture formats in {@link PIXI.INTERNAL_FORMATS} to the number of bytes taken by
+ * each texel.
+ *
```
bigtimebuddy, line 232: This is not showing up in the documentation.

packages/core/src/textures/resources/CompressedTextureResource.ts
```diff
@@ -0,0 +1,243 @@
+import { BlobResource } from './BlobResource';
+import { INTERNAL_FORMATS, INTERNAL_FORMAT_TO_BYTES_PER_PIXEL } from '@pixi/constants';
```
bigtimebuddy, line 2: Use `import type` when only using imports for types.

packages/loaders/src/CompressedTextureLoader.ts
```diff
@@ -0,0 +1,186 @@
+import { Loader } from './Loader';
+import { ILoaderResource, LoaderResource } from './LoaderResource';
+import { INTERNAL_FORMATS } from '@pixi/constants';
```
bigtimebuddy, line 3: `import type`

packages/basis/src/TranscoderWorker.ts
SukantPal: `TranscoderWorker` was designed as a "inner class" of `BasisLoader`. It doesn't serve any purpose outside of `BasisLoader`.

The inner class concept comes from Java. If you don't like it, I can move `TranscoderWorker` up into the `PIXI` namespace. The idea was just to not pollute the `PIXI` namespace.

packages/constants/src/index.ts
SukantPal: I was a little lazy there, sorry :). I'll add the \@property tags since the webdoc PR won't get merged before this one.

packages/basis/src/TranscoderWorker.ts
bigtimebuddy: Okay, that's fine. Disregard my comment.

bundles/pixi.js/src/index.ts
```diff
@@ -19,6 +19,7 @@ import { ColorMatrixFilter } from '@pixi/filter-color-matrix';
 import { DisplacementFilter } from '@pixi/filter-displacement';
 import { FXAAFilter } from '@pixi/filter-fxaa';
 import { NoiseFilter } from '@pixi/filter-noise';
+import '@pixi/loaders-compressed-textures';
```
bigtimebuddy, line 22: Copy pasta... this should be exporting the classes form this packages. This would be an import for mixins.

packages/loaders-compressed-textures/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import { CompressedTextureLoader } from './CompressedTextureLoader';
+import { DDSLoader } from './DDSLoader';
+import { KTXLoader } from './KTXLoader';
+import { Loader } from '@pixi/loaders';
+
+// parse any *.json compressed-textures manifest
+Loader.registerPlugin(CompressedTextureLoader);
```
bigtimebuddy, line 7: We do not register plugin automatically like this. This should be done in the bundle.

bigtimebuddy:
There are still a few more things in the Globals section of docs that shouldn't be:
http://pixijs.download/feature/compressed-texture-workflow/docs/global.html

SukantPal:
@ivan I moved the types out of `ContextSystem.ts`. There aren't a lot of lines added now.

P.S. You need to add review comments so they can be marked resolved.

ivanpopelyshev:
What I want is to types related to compressed textures appear only in compressed textures module. They shouldn't appear in default pixi d.ts and package at all.

SukantPal:
@ivanpopelyshev The types in `pixi.d.ts` come from the JSDoc comments anyway, so `WebGLExtensions` is not in `pixi.d.ts`. Are you saying I should create a new package instead of making changes in `ContextSystem`? I don't see any value in making such a big deal for 5-6 additional extensions.

ivanpopelyshev:
We can put all the constants to `const.js` and all the extensions in `ContextSystem`. But the point is - source code has to be readable and separated by features.

We already have too many features in core, for example ArrayResource can be moved out. Lets start from simple thing by not adding extensions there.

If you figure out a way how to separate it, the same way we'll be able to use for a few other features.

SukantPal:
@ivanpopelyshev I think that can be part of a separate refactor. Let's focus on the compressed texture stuff first.

Which constants are you talking about? And extensions? What extensions are you talking about?

ivanpopelyshev:
I think we can move away floating point textures too.

I just want a comment from @GoodBoyDigital or @bigtimebuddy about that issue.

SukantPal:
@bigtimebuddy won't comment for 7 days 😩

GoodBoyDigital:
> I think we can move away floating point textures too.
> 
> I just want a comment from @GoodBoyDigital or @bigtimebuddy about that issue.

Yo, I think for this, we can use a similar concept where we have a list of webGL extension ids we can add to before a context is made. Or possibly an ExtensionsSystem that manages extensions and lets users add extensions as required.
The bit that makes it tricky is our any extension a user wants to install, will also need to be added to our test WebGL context used for parsing shaders.

ivanpopelyshev (approved):
ok then, we'll separate it later.

packages/core/src/textures/resources/BlobResource.ts
```diff
@@ -0,0 +1,98 @@
...
+    {
+        // TODO: Override this method
+    }
+
+    /**
+     * Loads the blob
+     */
+    async load(): Promise<Resource>
+    {
+        return fetch(this.origin)
```
GoodBoyDigital, line 84: lets await stuff, will look nicer :D

packages/loaders-compressed-textures/src/CompressedTextureLoader.ts
```diff
@@ -0,0 +1,188 @@
...
+            const canvas = document.createElement('canvas');
+            const gl = canvas.getContext('webgl');
+
+            if (!gl)
+            {
+                console.error('WebGL not available for compressed textures. Silently failing.');
+
+                return;
+            }
+
```
GoodBoyDigital, line 157: one for later, but feel like managing of all extensions might benefit from being centralised?

SukantPal: Yes.

SukantPal:
@mat I followed up on all your suggestions :D

GoodBoyDigital (approved):
very nice work @SukantPal 👍

ivanpopelyshev (changes requested):
Internal formats in constants package - sorry, that's where I draw the line.

Find a way to move it away to separate package. We have to separate those constants the same way we separate our features.

SukantPal:
@ivanpopelyshev Give a concrete suggestion where you want them to be placed.

bigtimebuddy:
@ivanpopelyshev If you are going to block a PR, please provide more actionable feedback. 

What I think Ivan is saying is that `INTERNAL_FORMATS` and related compressed texture enums that are in constants should be moved to **loaders-compressed-textures** package. However, to make the dependency graph clean, we'd probably also need to move the resource-related stuff in core (e.g. CompressedTextureResource) to that package as well. BlobResource probably stays in core?

The advantage is that users can opt-out of compressed texture features without paying the cost of all those constants. This probably makes some organizational sense. We haven't externalized any resources yet and we lack a good API for installing/registering resources. But it doesn't look like you're using `autoDetectResource` for compressed textures, so this is probably fine. Combining `PIXI.resources` in the bundle is going to be annoying, so I suggest CompressedTextureResources if it's moved to loaders-compressed-textures, be at the root level and not in a "resources" namespace. It mostly functions as an internal class anyways.

englercj:
Doesn't seem like this PR needs to be blocked by organizational changes. It is already very large and the organizational suggestions are easier to address in a separate PR dedicated to getting the package graph correct.

ivanpopelyshev:
@bigtimebuddy is right

From my point of view. Its fine to refactor it later to move stuff away from `ContextSystem`, but those constants definitely should not appear in core package. 

Its a clear sign that something went wrong. I saw it - I blocked it, you can thank me later.

ivanpopelyshev (approved):

bigtimebuddy:
Same here @SukantPal, this need to get updated with the remove `resources` namespace.

SukantPal:
@bigtimebuddy Done!

SukantPal:
It's been one month now 🕐 🕚 🕦 🕥

bigtimebuddy:
I will merge this soon, I want to make sure I have time to merge and then do the follow-up PR. This is mostly for myself, but I plan to:

* Rename the package from **loaders-compressed-textures** to **compressed-textures**
* Move some of the core bits out and into the package, including constants and Resource-related bits
* Find a more generalize way to inject extensions in the ContextSystem

bigtimebuddy:
Thank you @SukantPal for your work on this :+1:

PR #6815, opened 2020-08-16
Document resizeThrottle

jamesplease:
##### Description of change

Documents the `resizeThrottle` of `Application`.

**Open questions:**

- Is there a default value?
- Is this actually used by `pixi.js`, or is it a remnant of when this feature was being developed? It only shows up [a single time in the source code search](https://github.com/pixijs/pixi.js/search?q=resizethrottle&unscoped_q=resizethrottle); it's not clear to me how it's being used, if it is used.

If it's not being used, then perhaps [L18](https://github.com/pixijs/pixi.js/blob/ef8f397faf8b3a877adc8b5b6a009aecc0ab7b33/packages/app/src/Application.ts#L18) needs to be deleted instead.

##### Pre-Merge Checklist

- [ ] ~~Tests and/or benchmarks are included~~


bigtimebuddy:
Yeah, I actually removed that functionality while I was working iterating in the feature, must have slipped through. Resizing now happen on request animation frame, any other throttling is up to the user. 

If you could remove the other resizeThrottle reference, that’d be better.

Issue #6810, opened 2020-08-12
Matrix#decompose does not set the pivot to (0, 0)


SukantPal:
https://codepen.io/sukantpal/pen/gOrPYoK?editors=0010

`Matrix#decompose` does its calculations assuming the pivot is (0, 0); however, it does not set it to that value. The pen linked above shows that setting the transform of a display-object to the same value after setting its pivot will render it slightly off.

SukantPal:
@ivanpopelyshev What if the user does not want it to set the pivot? Can `Matrix#decompose` decompose itself such the pivot remains invariant?

I think this is a matter of translation, as the `pivot` value only affects `tx`, `ty`.

PR #6820, opened 2020-08-19
Account for transform pivot when decomposing matrix

SukantPal:
#6810

##### Description of change
This patch has been battle-tested in the [@pixi-essentials/transformer](https://github.com/SukantPal/pixi-essentials/blob/ffe467dc8da91f3ac36723fbd59395428fbd8ffd/packages/transformer/src/utils/decomposeTransform.ts#L44) package.

Basically, the `Matrix#decompose` method sets the position assuming the pivot is (0, 0). Obviously, this is not always
the case.

There are either two fixes to this: set pivot to (0, 0) always OR calculate the position so that the pivot does not have to
change. I chose the latter b/c the user does not expect the pivot to change when decomposing a matrix, because it is
not a part of the actual transform.

##### Reasoning

The reason why this will work is b/c of the way tx, ty are calculated:

https://github.com/pixijs/pixi.js/blob/8f7ed7662949ea1217b6f3b6af470e855cfe0b13/packages/math/src/Transform.ts#L201

##### Pre-Merge Checklist

### DIFF
--- a/packages/math/src/Matrix.ts
+++ b/packages/math/src/Matrix.ts
@@ -359,6 +359,7 @@ export class Matrix
         const b = this.b;
         const c = this.c;
         const d = this.d;
+        const pivot = transform.pivot;
 
         const skewX = -Math.atan2(-c, d);
         const skewY = Math.atan2(b, a);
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Need tests for this @SukantPal

SukantPal:
@bigtimebuddy Done! I added a test that would fail without this patch.

ivanpopelyshev (approved):

PR #6829, opened 2020-08-22
Allows string2hex to work with css colors

themoonrat:
##### Description of change
Helps re-add compatibility to drop shadows using strings, which was accidentally removed when changes to the method to draw drop shadows in v5.1.1
https://github.com/pixijs/pixi.js/issues/6818

##### Pre-Merge Checklist

Before: https://www.pixiplayground.com/#/edit/154Ys5DNBCcKY6AN4DDMn
After: https://www.pixiplayground.com/#/edit/sWHOLF6HPjCiOoi3FcXvb


Tobbb:
Hello, not sure if this in intentional but the color breaks if hex color is provided with 3 digits, for example: if #fff is provided the color becomes blue https://www.pixiplayground.com/#/edit/qA7w4PUpF1o6blS3bDjKj
should be an easy fix

bigtimebuddy:
Good point @Tobbb. Do you want to add to this PR or do `#RGB` as another fix?

ivanpopelyshev (approved):

PR #6843, opened 2020-08-28
Making the rect param of the getLocalBounds methods of the Spirte and TilingSprite classes optional, using the same approach as in the DisplayObject class.

flashist:
##### Pre-Merge Checklist

### DIFF
--- a/packages/sprite-tiling/src/TilingSprite.ts
+++ b/packages/sprite-tiling/src/TilingSprite.ts
@@ -186,16 +186,6 @@ export class TilingSprite extends Sprite
      */
     public getLocalBounds(rect?: Rectangle): Rectangle
     {
-        if (!rect)
-        {
-            if (!this._localBoundsRect)
-            {
-                this._localBoundsRect = new Rectangle();
-            }
-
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
The optional rect makes sense, but the other change I don't understand. What are you fixing by moving the `if (!rect){}`? Can you show a bug or make a test for this?

flashist:
Thanks for the feedback. I changed the code, left only the changes, which made the rect param optional in the API and added a test for Spirte.getLocalBounds

My original idea was to keep code-approach the same, getting the DisplayObject class as an example. But after your feedback I found out that actually the same checking would be done in the DisplayObject class, if no rect is passed.

ivanpopelyshev (approved):
That is definitely a bug in our API, thank you!

flashist:
@ivanpopelyshev , @bigtimebuddy , thanks a lot! Looking forward to see the changes merged into the official dev branch.

bigtimebuddy:
Thanks for the contribution @flashist. If you see more issues, PRs always welcome!

PR #6855, opened 2020-09-03
Fix bitmap font loader to support both text and xml formats in "fnt" files

### LINKED ISSUE #6840: Bitmap font loader not working with "fnt"
The bitmap font can't be loaded using "fnt" format with the code from the examples:

```javascript
app.loader
    .add('TestFont', './TestFont.fnt')
    .load(onAssetsLoaded);

function onAssetsLoaded() {
    const bitmapFontText = new PIXI.BitmapText('bitmap fonts are supported!\nWoo yay!', { font: 'TestFont', align: 'left' });

    bitmapFontText.x = 50;
    bitmapFontText.y = 200;

    app.stage.addChild(bitmapFontText);
}
```

# Expected Behavior

The text will be rendered.

### Current Behavior

The font is not available. Instead, an error is thrown:

```
BitmapText.ts:113 Uncaught Error: Missing BitmapFont "TestFont"
    at new BitmapText (BitmapText.ts:113)
    at onAssetsLoaded (index.js:11)
    at MiniSignal.dispatch (mini-signals.js:93)
    at Loader._onComplete (Loader.js:623)
    at Loader.js:662
    at next (async.js:33)
    at Loader.SpritesheetLoader.use (SpritesheetLoader.ts:37)
    at Loader.js:643
    at async.js:41
```

### Possible Solution

I have no clue.

### Steps to Reproduce

Clone https://github.com/Prior99/pixi-bitmapfont-loader-reproduce

```bash
yarn
yarn start
```

### Environment

- **`pixi.js` version**: 5.3.0
- **Browser & Version**: Chrome 85.0.4183.49
- **OS & Version**: Arch Linux

SerG-Y:
##### Description of change

Fixes #6840 
##### Pre-Merge Checklist


packages/text-bitmap/src/formats/XMLTextFormat.ts
```diff
@@ -0,0 +1,39 @@
...
+     * Check if resource refers to text xml font data.
+     *
+     * @static
+     * @private
+     * @param {any} data
+     * @return {boolean} True if resource could be treated as font data, false otherwise.
+     */
+    static test(data: unknown): boolean
+    {
+        return typeof data === 'string' && data.indexOf('<font>\n  <info face=') === 0;
```
bigtimebuddy, line 22: I think the check here for `data.indexOf` is a little to strict and makes some assumptions about indenting, formatting etc. Not sure it's safe to assume this.Would be in favor of a RegExp instead or maybe something looser like `data.indexOf('<font>') > -1`

packages/text-bitmap/src/BitmapFontLoader.ts
```diff
@@ -16,16 +16,6 @@ import type { Texture } from '@pixi/core';
...
 export class BitmapFontLoader
 {
-    /**
-     * Called when the plugin is installed.
-     *
-     * @see PIXI.Loader.registerPlugin
-     */
-    public static add(): void
-    {
-        LoaderResource.setExtensionXhrType('fnt', LoaderResource.XHR_RESPONSE_TYPE.DOCUMENT);
```
bigtimebuddy, line 26: I think we should keep this `add()` method but parse `fnt` as `LoaderResource.XHR_RESPONSE_TYPE.TEXT` instead. Would be nice to be explicit and not assume the Loader will do the right thing here and default to text.

packages/text-bitmap/src/formats/index.ts
```diff
@@ -1,10 +1,12 @@
 import { TextFormat } from './TextFormat';
 import { XMLFormat } from './XMLFormat';
+import { XMLTextFormat } from './XMLTextFormat';
 
 // Registered formats, maybe make this extensible in the future?
 const formats = [
     TextFormat,
     XMLFormat,
+    XMLTextFormat,
```
bigtimebuddy, line 9: I'm not super sure of this, but what about renaming `XMLTextFormat` to `XMLStringFormat` or `XMLBufferFormat`? I'd like to disambiguate from the other two formats. Right not it seems like it's a combo of Text + XML, which isn't accurate.

bigtimebuddy (changes requested):
Awesome. Thank you for tackling this! I have a couple pieces of feedback. Also, would be nice to add tests for loading `fnt` files that are both flavors of XML & text.

packages/text-bitmap/src/formats/XMLTextFormat.ts
SerG-Y: Replaced with RegExp. Also added test for loading `fnt` based on text format, XML we already have.

packages/text-bitmap/src/BitmapFontLoader.ts
SerG-Y: Done

packages/text-bitmap/src/formats/index.ts
SerG-Y: yep, totally agree with you, as for me `XMLStringFormat ` will be more clear

bigtimebuddy (approved):
Much better! Thank you

themoonrat (approved):

bigtimebuddy:
Thank you @SerG-Y

PR #6856, opened 2020-09-03
Fixed webGLVersion of context

### LINKED ISSUE #6847: ContextSystem#initFromOptions fails to extract webGLVersion
```js
const view = document.createElement('canvas');

const renderer = autoDetectRenderer({
  context: view.getContext('webgl2'),
  view
})

console.log(renderer.context.webGLVersion); // = 1
```

This happens because the `ContextSystem#createContext` method sets the `webGLVersion`. That code path is not invoked when passing a custom context.

JetLua:
Fixed #6847 

demo: https://www.pixiplayground.com/#/edit/5HPwteo_LzF8BVrFGRa5i

### DIFF
--- a/packages/core/src/context/ContextSystem.ts
+++ b/packages/core/src/context/ContextSystem.ts
@@ -285,6 +285,13 @@ export class ContextSystem extends System
     {
         const attributes = gl.getContextAttributes();
 
+        const isWebGl2 = gl instanceof window.WebGL2RenderingContext;
+
+        if (isWebGl2)
+        {
+            this.webGLVersion = 2;
+        }
+
         // this is going to be fairly simple for now.. but at least we have room to grow!
         if (!attributes.stencil)
         {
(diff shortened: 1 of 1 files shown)


packages/core/src/context/ContextSystem.ts
```diff
@@ -285,6 +285,13 @@ export class ContextSystem extends System
     {
         const attributes = gl.getContextAttributes();
 
+        const isWebGl2 = gl instanceof window.WebGL2RenderingContext;
```
bigtimebuddy, line 288: This will throw an error if `window.WebGL2RenderingContext` is undefined. Which is part of the reason for having the `'WebGL2RenderingContext' in window` check.

bigtimebuddy (approved):
:+1:

ivanpopelyshev (approved):

SukantPal (approved):
LGTM!

bigtimebuddy:
Thank you @JetLua for this fix, good one :+1:

PR #6868, opened 2020-09-12
framebuffer.blit() in StencilSystem.push

### LINKED ISSUE #6844: Multisampled textures not handled at all in PixiJS internally
@ivanpopelyshev When you created multisampled textures with renderbuffers, you introduced an extra step in the render process:

```js
msaaTexture.framebuffer.multisample = 16;

renderer.render(displayObject, msaaTexture);
renderer.framebuffer.blit();
```

This makes it very problematic to the extent that PixiJS itself fails to take care of this extra step in masking.

No blit here:
https://github.com/pixijs/pixi.js/blob/064135e795dbfe7b53248ff457d71cb8536077ac/packages/core/src/mask/StencilSystem.ts#L93

@bigtimebuddy I would propose that we do a blit in `renderer#render` and save the user all future hassle. What do you think?

## Pen :pen:

Here's an example that fails with masking (comment out `graphics.mask = mask`, and it'll work):

https://codepen.io/sukantpal/pen/abNwBoN

SukantPal:
Fixes #6844


ivanpopelyshev:
Demo, please?

SukantPal:
@ivanpopelyshev You can test the pen the issue links to. Just change the script location of PixiJS; all create a new pen later today too for others to see as well.

ivanpopelyshev:
Just a note. In filterSystem:

```js
        if (state.renderTexture.framebuffer.multisample > 1)
        {
            this.renderer.framebuffer.blit();
        }
```

that's more verbose expression but `if` is not necessary. i dont remember why did i put it there

GoodBoyDigital (commented):
seems cool, how about we add a `renderer.flush()`
which does all this for us?

SukantPal:
@GoodBoyDigital Yes, I was thinking 'bout that. Maybe a separate PR, but seems like all of us agree on adding that.

ishaan-puniani:
Hello @GoodBoyDigital  and @bigtimebuddy  can we have a quick review for this one.. Seems like a release blocker for 5.4.0 from about a month. Thanks in advance.

ishaan-puniani (approved):

ishaan-puniani (approved):

GoodBoyDigital (approved):
to unblock I approve :D

Would be great to see a PR that encapsulates this into an cleaner `renderer.flush()` when you get a mo :)

bigtimebuddy (approved):

ivanpopelyshev (approved):
I dont understand why, but it works for the demo.

PR #6877, opened 2020-09-14
Fix VAO & instance array check in WebGL2.

dumganhar:
gl.createVertexArray may not be null in WebGL1 context on some platforms.
Use context.webGLVersion instead.

### DIFF
--- a/packages/core/src/geometry/GeometrySystem.ts
+++ b/packages/core/src/geometry/GeometrySystem.ts
@@ -93,7 +93,7 @@ export class GeometrySystem extends System
         this.CONTEXT_UID = this.renderer.CONTEXT_UID;
 
         // webgl2
-        if (!window.WebGL2RenderingContext && !gl.createVertexArray)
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):

ivanpopelyshev (approved):
holy cow! Where did you encounter that?

dumganhar:
@ivanpopelyshev 
I got this error on Hago (open.ihago.net) platform which developed JS Runtime (not WebView) for JS games.
In the old version of Hago, it exports gl.createVertexArray (not null) even running on a WebGL1 only device.
So the `gl.createVertexArray` checking logic will fail and some logic will go wrong direction.

Here, it's safe to check whether it's a WebGL2 context by `context.webGLVersion` which is definitely right.

GoodBoyDigital (approved):

bigtimebuddy:
Thanks for this fix @dumganhar :tada:

PR #6880, opened 2020-09-15
Fixes to support running pixi in WebWorker

IvanZatravkin:
##### Description of change

- replaced window with self
- added checks to see if global variable (i.e. MouseEvent) is defined before trying to access it
- set canvas to active by default if running in worker (otherwise it won't respond to events)

It was enough to run interactive pixi canvas in web worker for us. All original functionality should be unaffected.

##### Pre-Merge Checklist


bigtimebuddy:
Thank you for this @IvanZatravkin. Could you please make sure the unit-test pass? There are bunch of things failing as a result of this change.

Also, `globalThis` is not supported in [all browsers](https://caniuse.com/?search=globalThis), namely IE9-11. We'll need to provide a polyfill or solution to support on these browsers.

IvanZatravkin:
Changed `globalThis` to `self`, it works fine in every browser, main thread & workers. All tests pass now, shouldn't be a problem for IE either.

packages/interaction/src/InteractionManager.ts
```diff
@@ -1050,6 +1050,10 @@ export class InteractionManager extends EventEmitter
     public setCursorMode(mode: string): void
     {
         mode = mode || 'default';
+        if (!this.interactionDOMElement.style)
```
bigtimebuddy, line 1053: Would be helpful to add a comment here about why you're bailing without a style prop

bigtimebuddy:
Are these really the only `window` references?

packages/interaction/src/InteractionManager.ts
IvanZatravkin: offscreen canvas does not support setting style, changed code to better reflect this logic

IvanZatravkin:
> Are these really the only `window` references?

They were the only ones required to get pixi working, replaced all remaining occurrences too. 

Also, https://github.com/englercj/resource-loader/tree/master/src will require a fix, it depends on window.location and crashes in web workers.  I will create pr later for this one.

ivanpopelyshev:
The problem with resource-loader is that we still use v3 and there's already v4. Make a fix for 3.x branch?

IvanZatravkin:
Ok, sure. I'll do it this weekend.

bigtimebuddy (approved):

themoonrat (approved):

bigtimebuddy:
@IvanZatravkin I'm looking at this PR again and I think we need to address a few things before merging.

First, do you have an example of how to run this in a WebWorker? It would be nice to add this as a unit-test, if possible, if we are trying to legit support running in this environment. Do you think this is something you could add? 

Second, I can definitely see that this is something that would be very easy to regress in the future if someone uses `window`. Maybe we should consider adding an ESLint rule to disallow using `window` or favor `self`. Basically, is there anyway to harden this a little more for future developers?

IvanZatravkin:
1. Well, I can add some basic smoke tests, like "import pixi, create app, draw rectangle on canvas"

2. I can change default env to `worker` instead of `window`, but it will require some more work. Currently I have 48 warnings from es lint. Will this do? (after I fix warnings)

bigtimebuddy:
So yes to smoke tests. And depends on how intense the linking changes are, but I can run it and see. How bad is it?

Just some general thoughts: So maybe there is value in this current PR, even if it is half baked. That is, there are some substantial gotchas like dealing with interactive. But if we would like pixi to *officially* support running in WebWorkers we should find a whole solution not just make it easier to hack in support with weird polyfills.

How can we make this easy for users? Maybe we create a WebWorker bundle that only contains the bits that are support? (Maybe strip out interactive and accessibility?) Alternatively, we've been kicking around this idea for awhile to abstract ALL the DOM calls to a single adapter that users can change. That way we could create pixi that runs in workers, runs in Node, or runs in other JS environments where the DOM is too costly. Im kinda more interested in this approach.

Thoughts?

IvanZatravkin:
Well, I'm new to pixi, but stripping out interactive sounds too restrictive to me, almost to the point where it does not make sense to use web worker at all. 

I was thinking about changing InteractionManager to "register" event listeners internally, and providing user with object with required listeners and some helpers to attach these listeners in different environments. 

Adapter approach sounds nice, but may require a lot of work.


Committed changes to eslint with env set to worker, take a look.

bigtimebuddy:
Can you describe more about your use case for workers? How are you using this or planning on using it?

IvanZatravkin:
We are developing a game, we use pixi for graphics and react for UI overlay. To separate rendering from UI we use different threads. UI is handled in main thread (it is the only option, workers can't have access to DOM) and rendering is done in WebWorker. This way the interface will always stay responsive, even if visuals are lagging and vice versa.

IvanZatravkin:
I'm experimenting with smoke tests atm. Looks like I will need to make some changes to floss too :( 

At the very least I need to pass `nodeIntegrationInWorker: true,` to electron and I found no ways to do that in floss docs. 
And I will need some more time to figure out how to run mocha tests in WebWorker properly, never did it before. 

Texture loading will also need some more work. Currently in our game prototype we load textures using OffscreenCanvas with ImageBitmaps from blobs using fetch.

bigtimebuddy:
I can certainly help with floss. I need to update it anyways to support newer versions of Electron. I'm not sure the best way to run tests in workers, so we will have to figure that out together.

ishaan-puniani:
Hello, I think this is blocking the release of V5.4. 
Someone with write access please have a look at it. For me it is showing conflicts issue with "packages/settings/src/utils/isMobile.ts" and nothing else.

IvanZatravkin:
Hi. 
 
Just checking, what changes do I need to make for this to be merged? I currently don't have enough time to add tests in web workers for floss, maybe in a month or so :( sorry. 
If I need to just fix a few conflicts - I can do it in a day or two.

bigtimebuddy:
I don't think the linting is a good idea here. It's creating more headaches than enforcing the use of self. Let's just revert to the use of `self` over `window` before you change the lint rules.

ishaan-puniani:
@IvanZatravkin  would it be possible for reverting for some changes as suggested by @bigtimebuddy . Or can we postpone this feature to 5.4.1 / 5.5.0 Release ? 
Sorry for interruption, Actually I am waiting for a featured that is already packed in 5.4.0 and I have no idea how to use develop version or Pixi js from cdn to build animation using that feature. So, I really want 5.4.0 to be live as soon as possible.

Thanks in advance.

IvanZatravkin:
Done. Sorry for the delay.

ishaan-puniani:
@IvanZatravkin awesome!! Thanks mate.
Please add a label " Needs Review" as well.

bigtimebuddy:
Thank you @IvanZatravkin!

zszq:
> ##### Description of change
> * replaced window with self
> * added checks to see if global variable (i.e. MouseEvent) is defined before trying to access it
> * set canvas to active by default if running in worker (otherwise it won't respond to events)
> 
> It was enough to run interactive pixi canvas in web worker for us. All original functionality should be unaffected.
> 
> ##### Pre-Merge Checklist
> * [x]  Lint process passed (`npm run lint`)
> * [x]  Tests passed (`npm run test`)

I want to know how this should be used in webworker. I can draw graphics and sprites, but the interactive events have no effect. Is there an example for me to refer to? Thanks!

IvanZatravkin:
@zszq workers don't see DOM events, you need to make an event bridge which will send events from your main thread to web worker and register them with Pixi. 

Something like this: (parts of relevant code we use)
`
w: Worker;

const eventListeners = {
  keydown: (e: KeyboardEvent) => {
    w.postMessage({
      channel: "eventListener",
      event: { type: "keydown", data: cloneEvent(e) },
      messageType: "notify",
    });
  },
...
document.body.addEventListener("keydown", eventListeners.keydown, false);
`

And in web worker you process these events: 
`
const listeners = (function () {
  const listeners: [string, (...args: any[]) => void, boolean][] = [];
  const addEventListener = (...args: any[]) => listeners.push(args as any);
  (self as any).document = {
    createElement(type: any) {
      if (type === "canvas") {
        return new OffscreenCanvas(0, 0);
      } else {
        console.log("CreateElement called with type = ", type);

        return {
          style: {},
          addEventListener,
        };
      }
    },
    body: {
      appendChild() {},
    },
    addEventListener,
  };

  (self as any).window = {
    console: self.console,
    addEventListener,
    navigator: {},
    document: self.document,
    removeEventListener: function () {},
    WebGLRenderingContext: self.WebGL2RenderingContext || self.WebGL2RenderingContext,
    location: {},
  };
  return listeners;
})();

eb.subscribe("eventListener", (e) => {
  const fn = listeners.find(([t]) => t === e.data.event.type);
  if (fn) {
    fn[1](e.data.event.data);
  }
});`

What it does, is captures calls to "addEventListener" and saves callbacks to array. Then when you receive event - you call them with the event info. You will need to add back target and set it to your offscreen canvas.

zszq:
> @zszq workers don't see DOM events, you need to make an event bridge which will send events from your main thread to web worker and register them with Pixi.
> 
> Something like this: (parts of relevant code we use)
> 
> ```
> w: Worker;
> 
> const eventListeners = {
>   keydown: (e: KeyboardEvent) => {
>     w.postMessage({
>       channel: "eventListener",
>       event: { type: "keydown", data: cloneEvent(e) },
>       messageType: "notify",
>     });
>   },
> ...
> document.body.addEventListener("keydown", eventListeners.keydown, false);
> ```
> 
> And in web worker you process these events:
> 
> ```
> const listeners = (function () {
>   const listeners: [string, (...args: any[]) => void, boolean][] = [];
>   const addEventListener = (...args: any[]) => listeners.push(args as any);
>   (self as any).document = {
>     createElement(type: any) {
>       if (type === "canvas") {
>         return new OffscreenCanvas(0, 0);
>       } else {
>         console.log("CreateElement called with type = ", type);
> 
>         return {
>           style: {},
>           addEventListener,
>         };
>       }
>     },
>     body: {
>       appendChild() {},
>     },
>     addEventListener,
>   };
> 
>   (self as any).window = {
>     console: self.console,
>     addEventListener,
>     navigator: {},
>     document: self.document,
>     removeEventListener: function () {},
>     WebGLRenderingContext: self.WebGL2RenderingContext || self.WebGL2RenderingContext,
>     location: {},
>   };
>   return listeners;
> })();
> 
> eb.subscribe("eventListener", (e) => {
>   const fn = listeners.find(([t]) => t === e.data.event.type);
>   if (fn) {
>     fn[1](e.data.event.data);
>   }
> });
> ```
> 
> What it does, is captures calls to "addEventListener" and saves callbacks to array. Then when you receive event - you call them with the event info. You will need to add back target and set it to your offscreen canvas.

It's the same way as I thought, except that I am not very familiar with pixijs and don't know how to establish contact with it. This has given me a lot of help, thank you very much.

zszq:
What it does, is captures calls to "addEventListener" and saves callbacks to array. Then when you receive event - you call them with the event info. You will need to add back target and set it to your offscreen canvas.

Well, after many attempts I failed, ‘You will need to add back target and set it to your offscreen canvas.’ What does this sentence mean? Can there be a simple and complete example? Many thanks!

zszq:
> @IvanZatravkin 
Well, after many attempts I failed, ‘You will need to add back target and set it to your offscreen canvas.’ What does this sentence mean? Can there be a simple and complete example? Many thanks!

IvanZatravkin:
@zszq I may have time on weekend, will see what I can do. If you have semi-working example on github I could probably start with that and make it work.

zszq:
> @zszq I may have time on weekend, will see what I can do. If you have semi-working example on github I could probably start with that and make it work.

Yes, I have a test example, but the event does not take effect.[https://github.com/zszq/pixi_for_worker](https://github.com/zszq/pixi_for_worker)

zszq:
> @IvanZatravkin Yes, I have a test example, but the event does not take. [https://github.com/zszq/pixi_for_worker](https://github.com/zszq/pixi_for_worker)

zszq:
> @IvanZatravkin Yes, I have a test example, but the event does not take effect.[https://github.com/zszq/pixi_for_worker](https://github.com/zszq/pixi_for_worker)

IvanZatravkin:
https://github.com/IvanZatravkin/pixi_for_worker

zszq:
> @zszq https://github.com/IvanZatravkin/pixi_for_worker

It's really good, but it seems that the click event is not supported. I need the mouse to press the bounce event to simulate a click. I don't know if this can meet my needs, I have to do more testing. Anyway, thank you very much!

zszq:
> @IvanZatravkin It's really good, but it seems that the click event is not supported. I need the mouse to press the bounce event to simulate a click. I don't know if this can meet my needs, I have to do more testing. Anyway, thank you very much!

PR #6883, opened 2020-09-15
Support charset=utf8 in SVG Resources

### LINKED ISSUE #6771: "charset=utf8" string in the base64 encoded SVGResource not allowed
Hello,
I can see that when moving from pixi v4 to v5 I have sometimes a strange behaviour with the loading of texture, and some research led me to this line below, which doesn't take into account the potential "charset=utf8" string before the "base64", which is exactly the scenario I'm in, hence the strange behaviour I encountered sometimes.

https://github.com/pixijs/pixi.js/blob/86ec8422dfa859f3fb3aa2f6880ac53b1ab8ce93/packages/core/src/textures/resources/SVGResource.ts#L244

Is is something to be fixed by a PR, or is it better adding it to the documentation?

Cheers.

SerG-Y:
##### Description of change

Fixes #6771 
##### Pre-Merge Checklist


bigtimebuddy (approved):
Clean fix, nice test, good work @SerG-Y

ivanpopelyshev (approved):

bigtimebuddy:
Thank you @SerG-Y

Issue #6885, opened 2020-09-16
Calls to Graphics.drawTorus() fail if prior geometry exists


irongaze:
### Expected Behavior
Calling drawTorus() should correctly draw a torus, regardless of order of calls.

### Current Behavior
If prior geometry exists (prior drawTorus call, even a moveTo) new torus segment will start at last drawn point rather than at first point in torus geometry.

### Possible Solution
PR to follow...

### Steps to Reproduce
Create PIXI.Graphics, draw anything, draw a partial torus:

https://jsfiddle.net/v19qt35f/10/

### Environment

- **`pixi.js` version**: 5.3.3
- **Browser & Version**: Firefox

PR #6886, opened 2020-09-16
Fix PIXI.Graphics.drawTorus() behavior with existing geometry

irongaze:
##### Description of change
Fix for drawTorus() bug #6885

##### Pre-Merge Checklist


ivanpopelyshev (approved):
I didnt know we even have a `drawTorus`

bigtimebuddy (approved):
Thanks @irongaze

bigtimebuddy (changes requested):
Oops, spoke too soon. Better, but I'm seeing this weird artifact:
https://jsfiddle.net/bigtimebuddy/0m891ry4/

bigtimebuddy (approved):
Fixed example

irongaze:
> Oops, spoke too soon. Better, but I'm seeing this weird artifact:
> https://jsfiddle.net/bigtimebuddy/0m891ry4/

Yeah, wasn't sure what the expected outcome should be with a moveTo followed by a draw call.  My naive understanding would be that the torus segment should still be drawn properly, i.e. that the call to drawTorus() should implicitly end any prior drawing geometry.  Was planning on digging in more tonight to figure it out.  moveTo() + drawRect() works fine, for example.  

It seems like there needs to be a more clear delineation between "draw a full shape" and "draw a path segment" in the Graphics API, where full shape calls implicitly close/complete any prior actions.  But perhaps I'm not understanding the API well enough.

bigtimebuddy:
No, you're right @irongaze. The thing that makes drawTorus different is that the other shapes like drawStar, drawRegularPolygon go thru the `drawShape` API which doesn't have the issue. 

Adding a `finishPoly()` call _before_ adding the arcs fixes the problem. Updated and this is what I would expect to happen:
https://jsfiddle.net/bigtimebuddy/0m891ry4/

Issue #6884, opened 2020-09-15
Quality of Text and bitmapText


Bouh:
### Expected Behavior
Text and BitmapText should be visually equal to a Sprite with the same text and same styles.

### Current Behavior
The bitmapText and Text are not equal, and not equal to an Sprite with the same text and style.

-  Black: The image from photoshop with the same default settings found in PIXI.TextStyle ( except font size, see the settings bellow)
-   Red  : BitmapText using bitmapFont
-   White: Text

![image](https://user-images.githubusercontent.com/1670670/93263864-a26e7080-f7a6-11ea-81de-12cb6b5873fe.png)

![image](https://user-images.githubusercontent.com/1670670/93264104-f11c0a80-f7a6-11ea-8557-15aff388dbb4.png)

### Steps to Reproduce
https://jsfiddle.net/Bouh/avogx4n1/2/

### Environment

- **`pixi.js` version**: dev branch and 5.3.3
- **Browser & Version**:  Chrome 85
- **OS & Version**:  w10
- **Running Example**: https://jsfiddle.net/Bouh/avogx4n1/2/

bigtimebuddy:
`BitmapFont.from` does not support kerning where `Text` does. This is a known limitation of using BitmapText & BitmapFont.from.

Bouh:
Ok for [kerning](https://en.wikipedia.org/wiki/Kerning) on x-axis.
What about the lineHeight or the visible offset on y-axis?

bigtimebuddy:
BitmapText and Text use different layout methods and are not expected to perfectly line-up in the way you demonstrate. Photoshop also uses a different rasterizing method, so this is not a given.

Silver-Streak:
@bigtimebuddy Just to confirm, using bitmap font to generate a bitmap text object at 16 px, then scaling up, it looks like this:
![image](https://user-images.githubusercontent.com/24237149/93397387-2411dd00-f83f-11ea-8f27-038db77f43c1.png)

But an image made in an editor, then loaded as a texture/sprite, with the exact same font size (16 px) and scaling, looks like this:
![image](https://user-images.githubusercontent.com/24237149/93397463-44da3280-f83f-11ea-8a4b-50c08b8b8cd1.png)

The Bitmap text is much blurrier than the png loaded as a sprite.

Is that expected behavior?

bigtimebuddy:
Can you please provide an example that we can look at rather than just screenshots?

Silver-Streak:
[...]

PR #6888, opened 2020-09-16
BitmapFont.from generate the kerning information

bigtimebuddy:
Attempt at partially fixing #6884

Basically, when using `BitmapFont.from` we were not making any effort to get kerning information. This would make `BitmapText` look very different from the standard `Text` object.

This PR does not solve xadvance, but gets us a little closer.

In this example `Y` + `e` and `Y` + `o` in Arial is a dramatically example.

**Before**
https://jsfiddle.net/bigtimebuddy/k8jdeczu/
Screen Shot 2020-09-16 at 2 32 06 PM

**After**
https://jsfiddle.net/bigtimebuddy/ufrneabt/
Screen Shot 2020-09-16 at 2 32 25 PM

### DIFF
--- a/packages/text-bitmap/src/BitmapFont.ts
+++ b/packages/text-bitmap/src/BitmapFont.ts
@@ -384,11 +384,11 @@ export class BitmapFont
         let positionX = 0;
         let positionY = 0;
 
-        let canvas;
-        let context;
-        let baseTexture;
+        let canvas: HTMLCanvasElement;
+        let context: CanvasRenderingContext2D;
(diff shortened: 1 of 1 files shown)


ivanpopelyshev (approved):

SukantPal (approved):
The `BitmapFont.from` method has become ~165 lines long, which is borderline unacceptable 😅. I propose, in the future, that this be broken down in a separate internal `BitmapFontFactory.ts` file (without changing the exposed API).

PR #6873, opened 2020-09-12
Handle y-flipping when rendering to screen in RenderTextureSystem

SukantPal:
#6871 

Unfixed:
* https://codepen.io/sukantpal/pen/VwaxgdV?editors=0010

Fixed:
* https://codepen.io/sukantpal/pen/eYZrxXe?editors=0010

- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #6892, opened 2020-09-17
Clear partial texture using scissor test, fix regression in y-flipping PR in cases of resolution multiples

### LINKED ISSUE #6842: Clear just part of BaseRenderTexture
I just want to correctly clear part of BaseRenderTexture.

See [example](https://www.pixiplayground.com/#/edit/QSMI0wvL4soHWq-BsjZyr).

I expect that if I bind a renderTexture with frame to a RenderTextureSystem and [call clear method](https://pixijs.download/dev/docs/packages_core_src_Renderer.ts.html#line401) on it, then only renderTexture will be cleared (not whole BaseRenderTexture).

SukantPal:
Fixes #6842

https://codepen.io/sukantpal/pen/YzqOdPo?editors=1010

##### Description of change
* Use the scissor test when a render-texture only cover a portion of its base-texture
* Fix the regression in the y-flipped #6873 PR. It is really sad nobody caught the error in the calculation - I was using `renderer.screen.height` instead of `renderer.height` (the latter is resolution-multiplied, like the viewportFrame). All the examples were rendering by a wrong offset (if your resolution was not 1): https://pixijs.io/examples/#/demos-basic/container.js

https://github.com/pixijs/pixi.js/blob/b7881eb680afd7cad410821dc70027146bef7729/packages/core/src/renderTexture/RenderTextureSystem.ts#L165

vs

https://github.com/pixijs/pixi.js/blob/b7881eb680afd7cad410821dc70027146bef7729/packages/core/src/mask/ScissorSystem.ts#L116

##### Pre-Merge Checklist


ivanpopelyshev:
Can we please not use `renderer.height` and `renderer.width` anywhere? it should be `renderer.view.width` , or we can store values from gl context backbuffer fields, i forgot the name.  `renderer.width` is ambigious.

bigtimebuddy:
@SukantPal could you follow up on Ivan's feedback here? I think that's valid.

SukantPal:
Followed up :)!

bigtimebuddy (approved):

GoodBoyDigital (approved):
nice catch 👍

bigtimebuddy:
Thanks @SukantPal

PR #6893, opened 2020-09-18
@pixi/display documentation

### LINKED ISSUE #6889: Documentation for DisplayObject
DisplayObject needs to have more verbose documentation. I feel like I should add the following stuff (to the class description):

* `renderable` vs `visible`: [renderable](https://sukantpal.github.io/pixi-api/PIXI.DisplayObject.html#renderable) vs [visible](https://sukantpal.github.io/pixi-api/PIXI.DisplayObject.html#visible) is a hard-to-understand difference. That's because we don't explain how the transforms and bounds calculations work. I also think that **libraries use visible** and applications should use **renderable for culling** (as an official recommendation). The reason being that a library can optimize performance by setting visible=false on objects not being used internally (by saving transform calculations) and culling prevent things from being broken if they don't prevent transform updates of objects being used by libraries.

* Bounds calculation: Explain that bounds calculations updates bounds of **all display-objects in the subtree (children, grandchildren, etc.)**. Explain that **getLocalBounds trashes the world transforms and child bounds**. Explain that **when traversing the bounds of the scene graph - do not use getBounds as that will cause calculations to increase quadratically, rather use displayObject._bounds.getRectangle()**. Explain **content bounds, local bounds, world bounds, render bounds, relative bounds, etc.**

* Transform updates: what, why. Explain **width, height are abstractions over scale**.

* Explain how alpha can cause interaction to stop.

* Example for masking. I had a user forget to use `beginFill` before `drawRect` when creating mask - and didn't get why it wasn't working without it.

* Accessibility - the accessibility mixed-in members are not documented at all

## External recommended libraries related to display-objects to add to documentation

Maybe we should link these in the documentation so people don't have to look around:
* Container
* Graphics
* Sprite
* TilingSprite, AnimatedSprite
* Text, BitmapText
* Mesh, SimpleRope, etc.
* pixi-tilemap
* pixi-layers
* @pixi-essentials/cull
* @pixi-essentials/transformer, @pixi-essentials/conic
* and anymore libraries that export custom display-objects or services related to it

SukantPal:
Closes #6889

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -18,7 +18,194 @@ export interface DisplayObject extends GlobalMixins.DisplayObject, EventEmitter 
 /**
  * The base class for all objects that are rendered on the screen.
  *
- * This is an abstract class and should not be used on its own; rather it should b e extended.
+ * This is an abstract class and can not be used on its own; rather it should be extended.
+ *
+ * ## Display objects implemented in PixiJS
+ *
+ * | Display Object                  | Description                                                           |
+ * | ------------------------------- | --------------------------------------------------------------------- |
+ * | {@link PIXI.Container}          | Adds support for `children` to DisplayObject                          |
+ * | {@link PIXI.Graphics}           | Shape-drawing display object similar to the Canvas API                |
+ * | {@link PIXI.Sprite}             | Draws textures (i.e. images)                                          |
+ * | {@link PIXI.Text}               | Draws text using the Canvas API internally                            |
+ * | {@link PIXI.BitmapText}         | More scaleable solution for text rendering, reusing glyph textures    |
+ * | {@link PIXI.TilingSprite}       | Draws textures/images in a tiled fashion                              |
+ * | {@link PIXI.AnimatedSprite}     | Draws an animation of multiple images                                 |
+ * | {@link PIXI.Mesh}               | Provides a lower-level API for drawing meshes with custom data        |
+ * | {@link PIXI.NineSlicePlane}     | Mesh-related                                                          |
+ * | {@link PIXI.SimpleMesh}         | v4-compatibile mesh                                                   |
+ * | {@link PIXI.SimplePlane}        | Mesh-related                                                          |
+ * | {@link PIXI.SimpleRope}         | Mesh-related                                                          |
+ *
+ * ## External libraries that export or service display objects
+ *
+ * | Package                         | Description                                                          |
+ * | ------------------------------- | -------------------------------------------------------------------- |
+ * | pixi-layers                     | Faster layering of display objects than z-indexing                   |
+ * | pixi-sdf-text                   | Signed distance field text implementation                            |
+ * | @pixi-essentials/transformer    | Interactive interface for dragging, scaling, rotating display objects|
+ *
+ * ## Transforms
+ *
+ * The [transform]{@link DisplayObject#transform} of a display object describes the projection from its
+ * local coordinate space to its parent's local coordinate space. The following properties are derived
+ * from the transform:
+ *
+ * <table>
+ *   <thead>
+ *     <tr>
+ *       <th>Property</th>
+ *       <th>Description</th>
+ *     </tr>
+ *   </thead>
+ *   <tbody>
+ *     <tr>
+ *       <td>[pivot]{@link PIXI.DisplayObject#pivot}</td>
+ *       <td>
+ *         Invariant under rotation, scaling, and skewing. The projection of into the parent's space of the pivot
+ *         is equal to position, regardless of the other three transformations. In other words, It is the center of
+ *         rotation, scaling, and skewing.
+ *       </td>
+ *     </tr>
+ *     <tr>
+ *       <td>[position]{@link PIXI.DisplayObject#position}</td>
+ *       <td>
+ *         Translation. This is the position of the [pivot]{@link PIXI.DisplayObject#pivot} in the parent's local
+ *         space. The default value of the pivot is the origin (0,0). If the top-left corner of your display object
+ *         is (0,0) in its local space, then the position will be its top-left corner in the parent's local space.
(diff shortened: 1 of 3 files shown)


SukantPal:
ReAdY fOr ReViEw guys 😃

packages/display/src/DisplayObject.ts
```diff
@@ -18,7 +18,194 @@ export interface DisplayObject extends GlobalMixins.DisplayObject, EventEmitter
...
+ * | {@link PIXI.BitmapText}         | More scaleable solution for text rendering, reusing glyph textures    |
+ * | {@link PIXI.TilingSprite}       | Draws textures/images in a tiled fashion                              |
+ * | {@link PIXI.AnimatedSprite}     | Draws an animation of multiple images                                 |
+ * | {@link PIXI.Mesh}               | Provides a lower-level API for drawing meshes with custom data        |
+ * | {@link PIXI.NineSlicePlane}     | Mesh-related                                                          |
+ * | {@link PIXI.SimpleMesh}         | v4-compatibile mesh                                                   |
+ * | {@link PIXI.SimplePlane}        | Mesh-related                                                          |
+ * | {@link PIXI.SimpleRope}         | Mesh-related                                                          |
+ *
+ * ## External libraries that export or service display objects
```
bigtimebuddy, line 40: I'm not a fan of listing external libraries like this. It's brittle to reference other down stream dependencies maintained outside this repo.

packages/display/src/DisplayObject.ts
```diff
@@ -665,8 +878,11 @@ export abstract class DisplayObject extends EventEmitter
     }
 
     /**
-     * The scale factor of the object.
-     * Assignment by value since pixi-v4.
+     * The scale factors of this object along the local coordinate axes.
+     *
+     * The default scale is (1, 1).
+     *
+     * NOTE: Assignment by value since pixi-v4.
```
bigtimebuddy, line 885: "pixi-v4"? Suggest "PixiJS v4.x" or better yet use the doc tag `@since`

packages/display/src/DisplayObject.ts
```diff
@@ -681,8 +897,12 @@ export abstract class DisplayObject extends EventEmitter
...
 
     /**
-     * The pivot point of the displayObject that it rotates around.
-     * Assignment by value since pixi-v4.
+     * The center of rotation, scaling, and skewing for this display object in its local space. The `position`
+     * is the projection of `pivot` in the parent's local space.
+     *
+     * By default, the pivot is the origin (0, 0).
+     *
+     * NOTE: Assignment by value since pixi-v4.
```
bigtimebuddy, line 905: Same here, pixi-v4 is not a thing, be more verbose about this.

packages/display/src/DisplayObject.ts
```diff
@@ -18,7 +18,194 @@ export interface DisplayObject extends GlobalMixins.DisplayObject, EventEmitter
...
+ *         ⍺ and skew.y is β, then the line x = 0 will be rotated by ⍺ (y = -x*cot⍺) and the line y = 0 will be
+ *         rotated by β (y = x*tanβ). A line y = x*tanϴ (i.e. a line at angle ϴ to the x-axis in local-space) will
+ *         be rotated by an angle between ⍺ and β.
+ *         </p>
+ *         <p>
+ *         It can be observed that if skew is applied equally to both axes, then it will be equivalent to applying
+ *         one rotation. Indeed, if skew.x = -𝜽 and skew.y = 𝛳, it will produce an equivalent of rotation = 𝛳.
+ *         </p>
+ *         <p>
+ *         Another quite interesting observation is that skew.x, skew.y, rotation are communtative operations. Indeed,
```
bigtimebuddy, line 108: There are quite a few instances here of not code-quoting APIs, please consider `skew.x` whenever the subject is the actual API.

packages/display/src/DisplayObject.ts
```diff
@@ -18,7 +18,194 @@ export interface DisplayObject extends GlobalMixins.DisplayObject, EventEmitter
...
+ * | pixi-sdf-text                   | Signed distance field text implementation                            |
+ * | @pixi-essentials/transformer    | Interactive interface for dragging, scaling, rotating display objects|
+ *
+ * ## Transforms
+ *
+ * The [transform]{@link DisplayObject#transform} of a display object describes the projection from its
+ * local coordinate space to its parent's local coordinate space. The following properties are derived
+ * from the transform:
+ *
+ * <table>
```
bigtimebuddy, line 54: Maintaining more complicated HTML formatting in these comments is kind of annoying to deal with. Would be in favor of using sub-headers (h4, h5, etc) and doing these inline instead. A table doesn't add much her imo.

bigtimebuddy (commented):
Great additional info you're providing. This answers a lot of good specific questions. Most of my comments are very minor.

ivanpopelyshev (approved):

packages/display/src/DisplayObject.ts
SukantPal: @bigtimebuddy I would be in favor of using tables here. It gives the user of a sense of "one block" and "if I know it, I can skip it". I've optimized it for the generated docs, not for our readability 🤔

bigtimebuddy: My point is that it's harder to maintain than markdown and easy to mess up. But I'm not bullish on it if you think it's a better user experience ultimately.

bigtimebuddy (approved):

packages/display/src/DisplayObject.ts
SukantPal: @bigtimebuddy I understand that Markdown is generally better for tables. But this table is very large, and then you can't create paragraphs in Markdown tables. So Markdown would've made it hard to create a table with more than 200 characters per cell.

SukantPal:
cc @irongaze some new material added here that could be useful for the guides

PR #6905, opened 2020-09-24
@function -> @method

SukantPal:
Functions and methods are fundamentally different things, at least for webdoc. A function can be called without a `this` context and is not a member of a class/object. They are documented as separate pages. A method does have a `this` context and only is called using "`context.`"`method()`

The methods added by the cache mixin should be tagged `@method` and not `@function`.


bigtimebuddy:
While you're at it, could you fix these two classes in the docs:

* BasisFile
* IGLUniformData

ivanpopelyshev (approved):

bigtimebuddy (approved):

YaroYaro8 (approved):

YaroYaro8:
Hello

YaroYaro8:
``h

bigtimebuddy:
Thanks @SukantPal. If you can do a docs follow-up with those classes above, that would be great :+1:

PR #6914, opened 2020-09-30
fixes unbind issue when using different targets

GoodBoyDigital:
Noticed that we always assume texture is 2d, rather than using the textures target type.
easy fix 👍


SukantPal (approved):
LGTM!

bigtimebuddy (approved):

ivanpopelyshev (approved):
how the heck we didnt notice that

ivanpopelyshev:
Wait, do we even store boundTexture for different targets?

PR #6923, opened 2020-10-06
Fix  rendering tinted Spines

### LINKED ISSUE #6921: Issue with rendering tinted Spines
I stumble upon two minor problems with rendering spine mesh with tint.

### Expected Behavior
Spine meshes to be rendered correctly 

### Current Behavior
**First one :**
When you try to render a spine mesh with tint on IE 11 (in legacy mode) browser fails with multiple IndexSizeError errors during rendering.

Culprit of the issue is recent change made with https://github.com/pixijs/pixi.js/pull/6783 in _renderDrawTriangle
Because textureSource is now tinted canvas the dimensions are wrong.

**Second one :**
In the same scenario - even the texture is properly generated by getTintedCanvas it is not drawn on the original canvas.

### Possible solution: 
For the IE11 issue if both textureWidth and textureHeight might be set to textureSource.width and textureSource.height. Then the IndexSizeError will not be risen.  

For the second one i have no proposed solution but i suspect the issue is with matrix transform ( it is based on original source which may cause tinted canvas to be drawn outside the clip area.)

### Step to reproduced

Attached codepen contains 2 canvases - one in legacy mode(A)  and one standard(B). 
On canvas A there are 2 spines - one with tint and one without 
The tinted spine is rendering only textures that have no transformation, other one is working fine.

On canvas B is webGL version of the same. 
If pixi library is set to version 5.3.0 only difference will be in canvas A - on the tinted canvas all textures are rendered - those with transformation are without tint 

### Environment

- **`pixi.js` version**: legacy 5.3.3
- **Browser & Version**: All
- **OS & Version**:  All
- **Running Example**: https://codepen.io/ggzz/pen/rNLBBGN

JetLua:
Fix #6921 

demo: https://codepen.io/jerald/pen/vYKBPQQ

It's hard to imagine this being the third revision.

### DIFF
--- a/packages/canvas/canvas-mesh/global.d.ts
+++ b/packages/canvas/canvas-mesh/global.d.ts
@@ -5,6 +5,7 @@ declare namespace GlobalMixins {
         canvasPadding: number;
         _cachedTint: number;
         _tintedCanvas: HTMLCanvasElement;
+        _cachedTexture: import('@pixi/core').Texture;
     }
 
     interface MeshMaterial {
--- a/packages/canvas/canvas-mesh/src/CanvasMeshRenderer.ts
+++ b/packages/canvas/canvas-mesh/src/CanvasMeshRenderer.ts
@@ -120,8 +120,9 @@ export class CanvasMeshRenderer
             if (mesh._cachedTint !== mesh.tint)
             {
                 mesh._cachedTint = mesh.tint;
+                mesh._cachedTexture = mesh._cachedTexture || new Texture(base);
                 mesh._tintedCanvas = canvasUtils.getTintedCanvas(
-                    { texture: new Texture(base) },
+                    { texture: mesh._cachedTexture },
                     mesh.tint
                 ) as HTMLCanvasElement;
             }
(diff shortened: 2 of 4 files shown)


ivanpopelyshev:
Are we sure that it wont tint texture two times if there are two meshes in spine?

bigtimebuddy:
I worry about memory leaks creating a new `Texture` object if the Mesh's tint is constantly changing/animating. Would probably be better to re-use (or destroy the Texture)?

ivanpopelyshev:
> I worry about memory leaks creating a new `Texture` object if the Mesh's tint is constantly changing/animating. Would probably be better to re-use (or destroy the Texture)?

Oh that's usual problem in tilingpsrite/sprite tinting for canvas2d.

JetLua:
I cached `texture` should be better. If `tint` values change frequently, `_tintedCanvas` will still be generated frequently.

bigtimebuddy:
Can you make sure the cache texture is destroyed when the Mesh is destroyed?

ivanpopelyshev:
I dont like the fact that we cache tint of whole atlas, its bad for spine. I can approve PR though because i dont particularly care about that part :)

JetLua:
I can't think of a perfect solution here, just to make up for the previous two PR mistakes. 😿

ivanpopelyshev (approved):

ivanpopelyshev:
Let god sort them out. I'll fix if it becomes a memory problem for user. Btw, maybe spine meshes should have separate cache for that thing, then its definitely `pixi-spine` problem

bigtimebuddy (approved):
Thank you for this and for handling our feedback.

themoonrat (approved):

bigtimebuddy:
Thank you @JetLua :+1:

Issue #6903, opened 2020-09-23
Near colinear line join


lcsamaro:
### Expected Behavior
Fixed size line segment, see example.

### Current Behavior
'Line' drawn outside of segment after the join, see example.

Example output:
![image](https://user-images.githubusercontent.com/16543218/94035459-756b2080-fd99-11ea-8422-20aa2dbbcbcb.png)

### Possible Solution
Maybe numeric precision and line miter joins are making it happen.

### Steps to Reproduce
```
gfx.lineStyle(1, 0xffffff, 1);
gfx.moveTo(200,200);
gfx.lineTo(300,300);
gfx.lineTo(250.01,250);
gfx.endFill();
```

### Environment
- **`pixi.js` version**: 5.3.0
- **Browser & Version**: Chrome, Firefox, Safari
- **OS & Version**: macOS Mojave
- **Running Example**: https://pixiplayground.com/#/edit/CgDn9ejs17nCTqjWas8jy

bigtimebuddy:
That's pretty bad. Thanks for reporting @lcsamaro.

PR #6928, opened 2020-10-07
Fix spikes from wide strokes (#6924)

### LINKED ISSUE #6924: Spikes from strokes when a segment contains a neighboring point
When a path is drawn using `moveTo` and `lineTo` in `PIXI.Graphics`, and the path width is large enough that a segment contains a neighboring point, anomalous spikes can appear.

For instance: If you try to draw the path defined by the black circles and black segments, with a suitably large stroke width, you'll get the triangles & stroke shown here:

Note that a large spike emerges, facing in the opposite direction of the angle in the path. Here's a GIF:

bug

A live version of this demo is available [here](https://jsfiddle.net/Lzy2bohq/2/).

### In the wild

This problem occurs naturally if you shrink a path while keeping its stroke width fixed – for instance, in mapping applications. The spikes in the map below come from this problem:

Screen Shot 2020-10-06 at 5 49 11 PM

### Fix

I have a stop-gap fix implemented, which I will submit as a PR ASAP.

Thanks! ✌️

joshuahhh:
##### Description of change

This fixes #6924, the spikes-behind-wide-joins problem.

**Before:** (miter, bevel, round)
before

**After:** (miter, bevel, round)
after

It's not perfect. In particular, some regions are drawn twice, so strokes with alpha will be a bit mis-drawn. But that was the case before anyway. I believe some documentation already recommends using filters rather than stroke alpha in such cases.

(Note that some edges of the triangulation are incorrectly drawn red in the "round" examples above. This is an unrelated bug, which you can see occurs in both the "before" and "after" cases. If I can isolate it, I will report it as a separate issue.)

##### Pre-Merge Checklist

##### Thanks so much!


JetLua:
Cool !  This PR also fixed #6903 . 

https://pixiplayground.com/#/edit/fxZUnCBQVQhwTi2ePohLv

bigtimebuddy:
Thank you for this PR! Great explanation. Will review it. ping @SukantPal

SukantPal (approved):
Thanks for bringing these changes! In the future, I think we'll need to do work reducing the technical debt in this file.

joshuahhh:
Thanks for the quick responses!

@bigtimebuddy, I recommend reviewing the change with whitespace ignored – it makes much more sense that way: https://github.com/pixijs/pixi.js/pull/6928/files?diff=unified&w=1.

@SukantPal, I agree that `buildNonNativeLine` & friends could use a major rework. (For instance, I honestly have no idea why `indexCount` needs to be tracked the way it is.)

bigtimebuddy (approved):

bigtimebuddy:
Another good one @joshuahhh, you're on a roll! Thanks.

Issue #6925, opened 2020-10-07
Contributing: Unclear how to run a single test


joshuahhh:
Hello! I'm working on a contribution to Pixi, and am writing a test to make sure it works. I see on the [contributing guidelines](https://github.com/pixijs/pixi.js/blob/dev/.github/CONTRIBUTING.md) that `npm test` will build and test the entire project. But this is quite slow (48 seconds), and it is difficult to work on my test this way. Is there some way to run a single test, or the tests for a single package (like `graphics`)? Thanks!

(Sorry if this is the wrong venue to ask a question like this. But the contributing guidelines don't list a way for contributors to get help on things like this. Maybe I should submit that as a second issue. 😄)

bigtimebuddy:
The best way to do this is to add `.only` to a `describe` or `it` call. This Mocha convention will ignore other tests and is the best way to run single tests. Hope that helps.

joshuahhh:
@bigtimebuddy: This still means that the entire project is rebuilt though, right? That's most of the cost – a big non-incremental build.

bigtimebuddy:
Try using `npm run unit-test` instead of `test`. Another pro tip is to do `npm run unit-test -- --debug` to open dev tools for doing debugger and console logging.

joshuahhh:
That's very helpful; thanks! I now know how to iterate quickly writing tests.

(I thought that `test` and `unit-test` were identical. I didn't realize that `test` runs `pretest` first!)

So I think the ideal testing workflow for me is:
1. Run `npm run start` in one terminal, to get incremental compilation going.
2. When desired, run `npm run unit-test` in another terminal to quickly run the tests.

Let me know if I'm missing something; otherwise, I'll put that into the docs. Thanks again.

PR #6929, opened 2020-10-07
Don't use 'pretest' script in package.json

joshuahhh:
Right now, `package.json` contains the lines:
```
  "scripts": {
    ...
    "test": "floss --path test",
    ...
  }
```

You might think that `npm run test` and `npm run unit-test` do the same thing. But they don't, because there's a `"pretest"` script in there as well:

```
  "scripts": {
    ...
    ...
  }
```

Is there a reason to split these up? If not, I think this should be made more explicit and less spooky-action-at-a-distance:

```
  "scripts": {
    ...
    "test": "npm run lint && npm run build && npm run unit-test",
    ...
  }
```

(#6925 documents my trouble figuring this out.)

Thanks!


bigtimebuddy (commented):
The reason this is split with `pretest` is so that you can pass arguments to `floss`. If you chain the arguments together using `&&` I believe that argument will only get passed to the first thing. So this changes means that you couldn't do `npm test -- --debug` for instance.

bigtimebuddy:
Thanks for the suggestion, but I'm going to close this as it's unnecessary and breaks some intentional designed functionality mentioned above. The script `npm test` is used more for the CI-side of things and _does_ build using `pretest` (an npm convention).

PR #6946, opened 2020-10-15
Fix TilingSprite shader on some mobile devices

### LINKED ISSUE #6933: TilingSpriteRenderer buggy shader
### Expected Behavior
TilingSpriteRenderer using WEBGL should render normally.

### Current Behavior
TilingSpriteRenderer is not working properly on some mobile device. For example，"BBK H8A", whose Chinese name is "步步高H8"。
Here is the code:
https://www.pixiplayground.com/#/edit/80xRLERqs4Za6uyUthYRF

It displays properly on most device using webgl renderer.
But on the device mentioned above, it displays like the picture:
image

The first TilingSprite using "https://pixijs.io/examples/examples/assets/bunny.png" as texture become invisible.
But it actually takes the first pixel to render full sprite. If you use an image starts with "blue pixel", it will be all blue.

### Possible Solution
After testing, we found v5.1.6 is working normally, and v5.2.0 shows the bug.
So we use `git bisect` to locate the buggy commit.
It turns out [commit 3680d900c0ba8e05ec836a9cff21d25686714e5c](https://github.com/pixijs/pixi.js/commit/3680d900c0ba8e05ec836a9cff21d25686714e5c) is the problem.

We tried to restore and it works!

But we are not sure it is suitable to do that. Shader is beyond our knowledge.

### Steps to Reproduce
Open the [playground case](https://www.pixiplayground.com/#/edit/80xRLERqs4Za6uyUthYRF
) with buggy device.
Not sure what device may cause it, but the buggy device is using GPU called "Powervr rogue ge8300".
Devices using GPU "Adreno" series works fine.
Iphone works fine.

We use different chrome version and different android version, but they seems have nothing to do with it.
And the code is Shader, so we assume it relevant to GPU.

### Environment

- **`pixi.js` version**: 5.2.4, 5.2.0
- **Browser & Version**: Chrome 80.0.3987.87, Chrome 51  (same device mentioned above).
- **OS & Version**: Android 7.0
- **Running Example**: https://www.pixiplayground.com/#/edit/80xRLERqs4Za6uyUthYRF

bigtimebuddy:
Fixes #6933

Thanks @JohnEastWork for the suggestion and finding the bug.

### Example

https://pixijs.io/examples/?v=fix-6933#/sprite/tiling-sprite.js

### DIFF
--- a/packages/sprite-tiling/src/tilingSprite.frag
+++ b/packages/sprite-tiling/src/tilingSprite.frag
@@ -8,7 +8,7 @@ uniform vec2 uClampOffset;
 
 void main(void)
 {
-    vec2 coord = vTextureCoord - floor(vTextureCoord - uClampOffset);
+    vec2 coord = vTextureCoord + ceil(uClampOffset - vTextureCoord);
     coord = (uMapCoord * vec3(coord, 1.0)).xy;
     coord = clamp(coord, uClampFrame.xy, uClampFrame.zw);
 


themoonrat (approved):

GoodBoyDigital (approved):
nice!

ivanpopelyshev:
tagging @halyavin because it was his idea

PR #6947, opened 2020-10-15
Formalize Debug and Production Builds

bigtimebuddy:
### Overview

The goal of this change is to support adding more code inline that we'd otherwise want to remove from minified/production builds (as well as potentially prod-only code). @SukantPal had suggested this because he wanted to implement `toString` methods on our classes to support better debugging. 

```js
//#if _DEBUG
toString() {
  return `[@pixi/math:ObserverablePoint x=0 y=0]`;
}
//#endif
```

### Changes

Adds support for [JSCC](https://www.npmjs.com/package/rollup-plugin-jscc) for doing conditional compiling. There are lots of ways to do this but this is the most flexible and would also allow us to do string substitutions instead of using Rollup's replace plugin. 2-for-1, yah!

This change reorganizes the structure for built files and introduces minified (production) versions of the CommonJS and ESM builds. 

Previously build files looked like this:

* dist/package.js
* dist/package.min.js
* lib/package.js
* lib/package.es.js

This changes to:

* dist/browser/package.js
* dist/browser/package.min.js
* dist/cjs/package.js
* dist/cjs/package.min.js
* dist/esm/package.js
* dist/esm/package.min.js

Now there are debug and production builds for IIFE, CJS and ESM build flavors.

### Additional Thoughts

* Would be great to add asserts inline too, but did not add that yet. There's another plugin called `rollup-plugin-unassert` that could be used to strip asserts from production.
* Right now, the default `main` and `module` fields in the package.json point to the debug (unminified) builds. Ideally, this would be better to default to the minified-production builds and provide a Rollup or Webpack plugin/guide to support building with debug code. This way everyone would get the leanest version of PixiJS by default.


packages/utils/src/browser/hello.ts
```diff
@@ -32,7 +32,7 @@ export function sayHello(type: string): void
     if (navigator.userAgent.toLowerCase().indexOf('chrome') > -1)
     {
         const args = [
-            `\n %c %c %c PixiJS ${VERSION} - ✰ ${type} ✰  %c  %c  http://www.pixijs.com/  %c %c ♥%c♥%c♥ \n\n`,
+            `\n %c %c %c PixiJS ${VERSION} - ✰ ${type} ✰  %c  %c  https://pixijs.com/  %c %c ♥%c♥%c♥ \n\n`,
```
SukantPal, line 35: https://pixijs.com doesn't load without the "www." domain in front. Maybe a DNS redirect is needed?

packages/utils/src/browser/hello.ts
```diff
@@ -48,7 +48,7 @@ export function sayHello(type: string): void
     }
     else if (window.console)
     {
-        window.console.log(`PixiJS ${VERSION} - ${type} - http://www.pixijs.com/`);
+        window.console.log(`PixiJS ${VERSION} - ${type} - https://pixijs.com/`);
```
SukantPal, line 51: Same here, the URL won't load.

SukantPal (approved):
LGTM!

GoodBoyDigital (approved):
awesome awesome awesome 👍

packages/utils/src/browser/hello.ts
bigtimebuddy: Really gotta fix that DNS. I'll revert this bit.

PR #6949, opened 2020-10-17
Use frame.x|y in getImageData call for CanvasExtract.

### LINKED ISSUE #6937: CanvasExtract.pixels get wrong result with modified frame
### Expected Behavior
CanvasExtract.pixels should return pixels in frame area.

### Current Behavior
When modified frame of RenderTexture, CanvasExtract.pixels always returns area starts with (0, 0), which is top left.
CanvasExtract.canvas works fine.

### Possible Solution
It turns out in the code. `pixels` returns getImageData(0, 0, frame_width, frame_height).
So the area always start from top left.
[CanvasExtract](https://github.com/pixijs/pixi.js/blob/dev/packages/canvas/canvas-extract/src/CanvasExtract.ts#L161)

### Steps to Reproduce
https://www.pixiplayground.com/#/edit/vvSbz0egMw4pOREcojZo2

You can only see 2 bunnies using `forceCanvas: true`, which means using CanvasRenderer.
You can see 3 bunnies using WebGLRenderer.

I use v4.8.9 for enabling CanvasRenderer. In v5 may need to load legacy version.

### Environment

- **`pixi.js` version**: v4.8.9, v5.2.4
- **Browser & Version**: All browsers.
- **OS & Version**: All OS.
- **Running Example**: https://www.pixiplayground.com/#/edit/vvSbz0egMw4pOREcojZo2

SukantPal:
Fixes #6937 

##### Description of change
Passes the renderTexture's frame's (x, y) position to getImageData instead of (0, 0).

##### Pre-Merge Checklist


bigtimebuddy:
Example?

SukantPal:
https://www.pixiplayground.com/#/edit/vvSbz0egMw4pOREcojZo2

bigtimebuddy (approved):
Thanks for that. Is there an easy unit test for this we might add?

SukantPal:
No, unless we have an image comparison testing technique.

ivanpopelyshev:
Im sorry, but WHAT? Do we have that in WebGL?

SukantPal:
@ivanpopelyshev Yes. Do you want to remove it from WebGL and keep canvas same? SURPRISED??

JetLua:
Will this PR be released with `v5.4.0`?

ivanpopelyshev (approved):

ivanpopelyshev:
The only thing wrong here is that its subject for X-Files . There's no rounding on multiplying by resolution.

bigtimebuddy:
Thanks @SukantPal!

PR #6960, opened 2020-10-24
[Feature]: toString methods for debugging

SukantPal:
This PR introduces `toString` methods in classes where the contents can be printed easily as a short string. This has the following benefits:

* `console.log('Prefix: ' + objectToPrint)` will no longer print "Prefix: [object Object]" but rather "Prefix: Some meaningful data". This allows developers to print multiple objects without having to log them separately.
* Logging with toString can prevent unnecessary memory pressure when printing at a high frequency (10s of lines per frame = 1000s of lines per second). That's because the garbage collector can still collect the printed object.

##### Description of change
Added toString methods to the shapes in @pixi/math and then some more classes in @pixi/core.

The format of the printed message is as follows:

```js
"[@pixi/packageName:ClassName propName="value" propName2="value2" ...]"
```

##### Pre-Merge Checklist

- ~[ ] Tests and/or benchmarks are included~
- ~[ ] Documentation is changed or added~ (these methods **will not** be added to the documentation)


bigtimebuddy (approved):
Good one!

SukantPal:
cc @ivanpopelyshev

ivanpopelyshev (approved):

themoonrat:
@bigtimebuddy & @SukantPal Maybe a bit late, but my only comment is... since `toString` is common function in the actual language, that it might be confusing to have our own debug functions also use `toString`. Sometimes. Maybe `toDebugString` is a more consistent function name we could use safely, across the libs without worrying about interfering with proper `toString` functions.

SukantPal:
> @bigtimebuddy & @SukantPal Maybe a bit late, but my only comment is... since `toString` is common function in the actual language, that it might be confusing to have our own debug functions also use `toString`. Sometimes. Maybe `toDebugString` is a more consistent function name we could use safely, across the libs without worrying about interfering with proper `toString` functions.

The toString function is common in the language and that's why we should implement it. Generally, you implement toString for exactly this reason (casting to a string)

bigtimebuddy:
I agree with @SukantPal here. This is the intended purpose of `toString`. Instead of `[object Object]` we'd get something more meaningful.

themoonrat:
Ok 👍

PR #6962, opened 2020-10-26
Exclude deprecation messages from Production builds

bigtimebuddy:
This PR depends on #6961

### Fixed

* Fix production browser bundles to use _production_ packages

### Changed

* Production bundles exclude deprecations messages (more performant, removes ~10KB from bundle size)

### Future Notes

I'd like to create a simple guide for bundle users (Rollup & Webpack) that shows them how to bundle a production version of the library. For instance, here's the Rollup version. Might create a Rollup plugin for this.

```js
// rollup.config.js
import alias from '@rollup/plugin-alias';

const rollupConfig = {
  plugins: [
    alias({
      entries: [{
        find: /^(@pixi\/([^\/]+))$/,
        replacement: '$1/dist/esm/$2.min.js',
      }]
    }),
  ]
}
```

```js
// webpack.config.js
import AliasRegexOverridePlugin from 'alias-regex-webpack-plugin';
 
const webpackConfig = {
  plugins: [
    new AliasRegexhOverridePlugin(
      /^(@pixi\/([^\/]+))$/,
      '$1/dist/esm/$2.min.js'
    )
  ]
}
```

### DIFF
--- a/bundles/pixi.js/src/useDeprecated.ts
+++ b/bundles/pixi.js/src/useDeprecated.ts
@@ -76,7 +76,9 @@ export function useDeprecated(this: any): void
         SVG_SIZE: {
             get(): RegExp
             {
+                // #if _DEBUG
                 deprecation(v5, 'PIXI.utils.SVG_SIZE property has moved to PIXI.SVGResource.SVG_SIZE');
+                // #endif
 
                 return PIXI.SVGResource.SVG_SIZE;
             },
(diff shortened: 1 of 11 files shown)


SukantPal (approved):
Skimmed through it! LGTM!

themoonrat (approved):
lgtm
I guess it would have been easier to stick the #if _DEBUG stuff within the deprecation function itself? So on a production build it'd have been an empty function that did nothing.

Was this a conscious decision to put it outside of the function, so that the release builds are as clean as possible?

bigtimebuddy:
@themoonrat the reason I didn't do that is because I actually want to remove the entire messages/strings from the output, not just ignore the logging. This way we save 10kb.

Issue #6934, opened 2020-10-10
roundPixels property is not working on multiline texts that allign to center.


shaise:
BUG:
when working with bitmap text that is aligned to center and is also multy-line , the roundPixel property does not work properly and the text can become blurry.

SOLUTION:
Simple solution: in BitmapText.ts, line 355:
  ` alignOffset = (maxLineWidth - lineWidths[i]) / 2; `
allignOffset should take into account the roundPixel property and if so, floor the value. (I added a floor and it fixed the text)

Thank you for a Great, Great piece of software!!!!

shai

- **`pixi.js` version**: _e.g. 5.3.3
- **Browser & Version**:  all

hjpvue:
I want to frontline this on a couple of projects. But yeah I know everyone has my logs. It's better that way probably. Still mean though. But then again google filament wouldn't be possible without inter-industry co-operation.

bigtimebuddy:
@shaise thanks for reporting this. Could you provide a simple example that demonstrates this issue? Would be helpful to verify when this is fixed.

shaise:
Hi attache a simple example demonstrating the bug. The first line looks ok, the second is smudged.
[TestRoundPixel.zip](https://github.com/pixijs/pixi.js/files/5360247/TestRoundPixel.zip)

shaise:
I have a fix for this issue + added a 'justify' option to align property of bitmap texts. If there is any interest, I can issue a PR.
align: 'justify' example: (there is a \n after the 'fox' in the middle)
![justify](https://user-images.githubusercontent.com/5813568/96787026-c2beca00-13f9-11eb-8551-8a6e2ef201f4.png)

bigtimebuddy:
Great feature suggestion! PR welcome.

shaise:
Done.
BTW, how do i generate the pixi.min.js? 'npm run build' or 'npm start' generate only pixi.js, and pixi.mjs - which is almost the same size as pixi.js.

bigtimebuddy:
Good one @shaise, will review the PR.

shaise:
Hi. PR changed to include fix only

PR #6965, opened 2020-10-26
Fix roundPixel bug with bitmap texts, issue #6934

shaise:
##### Description of change
Fix roundPixel bug with bitmap texts, issue #6934

##### Pre-Merge Checklist

[TestRoundPixel.zip](https://github.com/pixijs/pixi.js/files/5441697/TestRoundPixel.zip)


packages/text-bitmap/src/BitmapText.ts
```diff
@@ -474,7 +474,14 @@ export class BitmapText extends Container
...
             const char = chars[i];
-            const xPos = (char.position.x + lineAlignOffsets[char.line]) * scale;
+            let offset = char.position.x + lineAlignOffsets[char.line];
+
+            if (this.roundPixels)
+            {
+                offset = Math.floor(offset);
+            }
+
+            const xPos = offset * scale;
```
SukantPal, line 484: What if `scale` is a non-integer?

shaise: When scaling / rotating a font you loose aliasing anyway. The roundPixels is good only if you work with 1.0 scale / 0.0 rotation.

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -474,7 +474,14 @@ export class BitmapText extends Container
         for (let i = 0; i < lenChars; i++)
         {
             const char = chars[i];
-            const xPos = (char.position.x + lineAlignOffsets[char.line]) * scale;
+            let offset = char.position.x + lineAlignOffsets[char.line];
+
+            if (this.roundPixels)
+            {
+                offset = Math.floor(offset);
```
bigtimebuddy, line 481: Any reason for `floor` over `round`?

shaise: You are right. round is better. I have modified the code

packages/text-bitmap/src/BitmapText.ts
bigtimebuddy: This sounds reasonable to me.

bigtimebuddy (approved):

SukantPal (approved):
Thanks @shaise!

bigtimebuddy:
@themoonrat could you take a quick look at this?

themoonrat (approved):

themoonrat:
> @themoonrat could you take a quick look at this?

I'm happy with this. Before commit cfc05d1e91d198f7d1fdddc6ff62f39e5965e353 - bitmap text contained individual for each character, each of which had the sprite's 'roundPixels' property set.

But in that commit, BitmapText got changed to using a Mesh for optimisation reasons, which lost any usage of 'roundPixels' altogether. I'm happy that this PR effectively adds that back in.

themoonrat:
What I'll add, tho, is that the roundPixels is no longer dynamic, like sprites are, where you could turn it on and off dynamically and see it's effects at runtime. Now, it's dependent on roundPixels being set before new text is set. 

Not worth this PR being held up, imo but potential for a follow up pr for public getter and setter on roundPixels to then also call updateText function here.

bigtimebuddy:
That's a good suggestion to use setter/getter for this. Would be nice to have round work after text is set. @shaise is this something you want to add? No worries if you don't, we can add it later.

shaise:
Hi @bigtimebuddy ,

Shouldn't be an issue to add a getter/setter. I will do it and update the PR. 

shai.

shaise:
getter/setter added. attached demo showing the fix in action.
[test_round_pixels.zip](https://github.com/pixijs/pixi.js/files/5451305/test_round_pixels.zip)

themoonrat (approved):

bigtimebuddy:
text
(example above toggling roundPixels)

Looks good @shaise, thanks for that fix :+1:

PR #6968, opened 2020-10-28
Support text alignment 'justify' in BitmapText

shaise:
##### Description of change
Add 'justify' feature to bitmap texts. This aligns the text both right and left using varying spaces between words
roundPixels = true, is recommended for small font sizes. 

##### Pre-Merge Checklist
- [x] Documentation is changed or added - only inside code

test_justify
[TestJustify.zip](https://github.com/pixijs/pixi.js/files/5453257/TestJustify.zip)

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -28,6 +28,7 @@ interface CharRenderData {
     line: number;
     charCode: number;
     position: Point;
+    prevSpaces: number;
 }
 
 const pageMeshDataPool: PageMeshData[] = [];
@@ -87,7 +88,7 @@ export class BitmapText extends Container
      * @param {string} style.fontName - The installed BitmapFont name.
      * @param {number} [style.fontSize] - The size of the font in pixels, e.g. 24. If undefined,
      *.     this will default to the BitmapFont size.
-     * @param {string} [style.align='left'] - Alignment for multiline text ('left', 'center' or 'right'),
+     * @param {string} [style.align='left'] - Alignment for multiline text ('left', 'center', 'right' or 'justify'),
      *      does not affect single line text.
      * @param {number} [style.tint=0xFFFFFF] - The tint color.
      * @param {number} [style.letterSpacing=0] - The amount of spacing between letters.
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
I'm bummed we don't support justify on Text as well. Hopefully someone can add support on that class too.

shaise:
@bigtimebuddy ,

In fact, I was looking at the standard text as well, to see if I can add it there as well, but it is a bit more complicated as the text is not placed char-by-char, but full lines. Playing with that is more tricky, and I don't want to mess it up.

shai

bigtimebuddy:
No worries, that's a more complicated task, for sure.

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -473,7 +488,7 @@ export class BitmapText extends Container
         for (let i = 0; i < lenChars; i++)
         {
             const char = chars[i];
-            let offset = char.position.x + lineAlignOffsets[char.line];
+            let offset = char.position.x + (lineAlignOffsets[char.line] * char.prevSpaces);
```
SukantPal, line 491: This line is now quite confusing (although I understand in the context of this PR, "adding line justify"). The problem is that `prevSpaces` is 1 for all other alignments, and that is not at all obvious. There may be actually character spaces before `char` on that line, but we are ignoring them for non-justify alignments

I would suggest keeping `prevSpaces` the actual number of spaces before the character. Then, add an ternary operator to multiply it only when the alignment is justify.

```suggestion
            let offset = char.position.x + (lineAlignOffsets[char.line] * (this._align === 'justify' ? char.prevSpaces : 1));
```

packages/text-bitmap/src/BitmapText.ts
```diff
@@ -301,6 +308,7 @@ export class BitmapText extends Container
             charRenderData.charCode = charCode;
             charRenderData.position.x = pos.x + charData.xOffset + (this._letterSpacing / 2);
             charRenderData.position.y = pos.y + charData.yOffset;
+            charRenderData.prevSpaces = this._align === 'justify' ? spaceCount : 1;
```
SukantPal, line 311: Following up on the other suggestion, keep `prevSpaces` consistent across alignements:

```suggestion
            charRenderData.prevSpaces = spaceCount;
```

SukantPal (changes requested):
Thanks @shaise. The changes I suggest will help future readers from loosing context in the code.

packages/text-bitmap/src/BitmapText.ts
shaise: Hi @SukantPal ,
Indeed you are correct. I was just thinking that my way might be a bit more efficient. But on second thought, the added comparison is compensated by one less comparison in your second fix, so I will update the PR with your changes.

SukantPal (approved):
LGTM!

PR #6983, opened 2020-11-08
Enable scissor mask support by default

### LINKED ISSUE #6938: Mask performance on mobile devices with pixi5
Hi!

Something happened with masks performance in pixi5, especially on Samsung devices .
When I switched to Pixi5 I found that fps dropped to 5-10 in my app on Samsung A5, S6 and to 20-40 fps on Samsung M31. It recovers when I disable all masks. I use them to hide scrollers content, for some animations and to round corners in some cases (maybe totally 10-15 masks on screen). But it shows 60 fps on much worse devices than this A5.

Two codepens here to see difference. Just tap to add masked bunnies and watch on rotating bunny to detect fps drop (recommend set scale to 0.5 and hide html).
https://codepen.io/Insi04/pen/MWeapRb
https://codepen.io/Insi04/pen/oNLjZPq
Also no mask example to make sure that mask is the key: https://codepen.io/Insi04/pen/abZvEmm

E.g. on Samsung A5 (2017) you can spawn 100+ masked bunnies at v4 with no drops, and just nearly 50 on v5. 
On Samsung S6 it's 100+ vs 25 bunnies.
Other devices also has difference: Asus Zenfone 2 has 400 vs 100 bunnies, iPhone 8 Plus easily spawns 1500 on v4 and nearly 600 on v5 (but I don't see lags in my app, so maybe it's some reserve).

We commonly connect that with GPU ARM Mali which used in Samsung Exynos (heard that they has some features about z-index rendering and Unity has [special guide](https://static.docs.arm.com/100140/0302/arm_guide_for_unity_developers_optimizing_mobile_gaming_graphics_100140_0302_00_en.pdf) for optimisation) but then see the difference on all other devices I've check (Asus Zenfone, iPhone 8+ and so on), so looks like masks was last straw in performance for Mali :)

UPD: I've checked my main app on few more Mali GPUs and all of them has fps dropped to 5-25 just with pixi5  :( 
Huawei Ascend Mate 7, Huawei P40 Lite E, Samsung M31. On Apple devices and android with Snapdragon (Samsung A51) has no changes in performance.

- **`pixi.js` version**: 5.3.3
- **Browser & Version**: Chrome 86.0.4240.75
- **OS & Version**: Android

bigtimebuddy:
Fix #6938

### DIFF
--- a/packages/core/src/mask/MaskSystem.ts
+++ b/packages/core/src/mask/MaskSystem.ts
@@ -53,7 +53,7 @@ export class MaskSystem extends System
          * @member {boolean}
          * @readonly
          */
-        this.enableScissor = false;
+        this.enableScissor = true;
 
         /**
          * Pool of used sprite mask filters


SukantPal (approved):
Nice!

GoodBoyDigital (approved):
welcome back scissor mask! 👍

PR #6994, opened 2020-11-12
[4.5.x] Cherry-picking blend modes bug fix to older branches.

flashist:
Cherrypicking bugfix to older pixi.js version: https://github.com/pixijs/pixi.js/commit/2d50b0526d7aee8075ec3e60eac254172ea0ca26

Discussion: https://github.com/pixijs/pixi.js/issues/6988


bigtimebuddy:
Can you please work on making this pass Travis? See fixes in v4.4.x

flashist:
@bigtimebuddy , Thanks for the reference and guidance: I tried to cherry-pick the mentioned changes, but it looks like it's not enough.

At the beginning there were problems with the dist folder, because it wasn't generated. So I changed the scripts a bit, and the npm run build command is run earlier.

Now it looks like the problem is with smth from another package (@pixi/jsdoc-template). I debugged a bit and found few problems:

- It looks like the command fs.copyFileSync (publis.js:444) is used incorrectly, because the toDir param shows the path to the target dir, but not the target dir + filename (according to the node.js docs it should be dist + filename: https://nodejs.org/api/fs.html#fs_fs_copyfilesync_src_dest_mode )

Maybe there is something with the template module? Should we use some other versions of it?

bigtimebuddy:
Try seeing if there’s a jsdoc template update and bumping that dependency

flashist:
@bigtimebuddy , I tried to remove all ^ dependencies (I faced problems with such dependencies in other projects), but the problem still was there.

Then I tried to fix the problem in the @pixi/jsdoc-template and it looks like it works!

https://github.com/pixijs/pixi-jsdoc-template/pull/27 - could you please check the pull-request there and publish the new version there (it looks like the new version should be 2.6.1, cuz my changes are fixes).

After that I'll be able to use the new version of the jsdoc-template module to fix the 4.5.x version travis build problems.

bigtimebuddy:
Why did the docs template work fine in 4.4.x but not this?

flashist:
@bigtimebuddy , I can't see any mention of the @pixi/jsdoc-template in the 4.4.x branch: https://github.com/pixijs/pixi.js/blob/v4.4.x/package.json

Are you sure there is a dependency to this package?

flashist:
@bigtimebuddy , it looks like the version is ready for publishing. Could you please share approximate dates when you plan to publish it?

flashist:
@bigtimebuddy , @GoodBoyDigital , could you please share your ideas about supporting old versions? Is the work I'm doing here a valuable or this is not important for the team?

If it's not important and the team has no time for it, and decide that spending time on reviewing changes for old versions are not important, please, let me know, I won't wait for updates regarding these updates.

bigtimebuddy:
Thanks for doing the leg work on this PR. It would also be helpful if you could do the same for your other two PRs which are also failing.

I believe that I made it very clear in our previous discussions that this is not a high priority at the moment. If this is time-sensitive for you, consider publishing your own fork. I cannot give you a timetable when this will be done. In the queue, we are currently trying to publish 5.2.5, 5.3.4 and 5.4.0-rc.3 which takes precedence over v4 versions.

flashist:
@bigtimebuddy , I think it would be a good practice to apply changes to 4.6.x and 4.7.x branches, after we finish with this one, to make sure there are no additional problems.

One more thing: I still can't complete the PR (and other PRs), because the template modules requires update too. And after your question about it, it became even more vague, whether what I did with the js-template module and 4.5.x branch are correct solutions or not.

So my plan is:

1. Get an approval of the js-template module
2. Then you (or other maintainers) will be able to publish it.
3. I will be able to use the published version in this PR
4. Get an approval of this PR
5. Publishing 4.5.x branch
6. Applying the same changes to the 4.6.x and 4.7.x branches

About timetables: it was clear that this is not the priority of the team. I just was confused by your question about the js-template module and not receiving any replies back.

flashist:
@bigtimebuddy , really, didn't expect that reviewing such small PRs would require such amount of time.

bigtimebuddy:
While I understand your frustration, I'm closing these PRs since we don't have the capacity to address and there's not a huge appetite for these releases right now.

I'm not sure what it's like where you are, @flashist, but our team is still struggling with the realities of COVID. We have demanding day jobs working remotely, families and children that need us. So yes, it has, in this moment, made this project much less responsive than normal, fewer PR approvals, no legacy support, and fewer releases getting through. I will not apologize for everyone feeling strapped, and I appreciate your understanding.

As we've already discussed, there are a number of ways you can workaround or hack in these one-line changes in your projects from everything to publishing your own shadow versions of pixi.js to using Webpack to find-and-replace the offending blend-mode. Best of luck.

flashist:
@bigtimebuddy , it looks like you have some mood swings:
1) (accept) Firstly, you're polite and willing to help when it comes to cherrypicking the existing bugfix into the 4.4.x versions: https://github.com/pixijs/pixi.js/pull/6972

2) (reject) Then you don't want the spend any time more, cuz it takes more time than you expected: https://github.com/pixijs/pixi.js/issues/6988#issuecomment-724733754

3) (accept) Then after @GoodBoyDigital comes, you creates 4.5.x, 4.6.x and 4.7.x branches to do the same thing as was done with the 4.4.x branch: https://github.com/pixijs/pixi.js/issues/6988#issuecomment-725481967

4) (again accept!) Then you ask for fixing build problems (so you wouldn't need to do this job for publishing the new version): https://github.com/pixijs/pixi.js/pull/6994#issuecomment-731838018

5) (FINAL REJECT!) And then, when all the job is done and all you need to to review 4-5 files, you say "I don't have time for this, all the job you did is useless, I'm closing all PRs, thanks for wasting your time".

Is it really a way an open-source project should be maintained? I would've understood if from the beginning you said: "The first rule of the pixi.js club is WE DON'T SPEND TIME ON OLD VERSIONS", but after all the time you and me put into finding solutions, it's really strange.

Maybe you should be not the only one who is responsible for reviewing and publishing versions?

I kind of understand, that it's just not what you're interested in at the current moment, and as you said, everybody has more important things to do. But asking smb else for doing some work and then just "nah, I'm closing it without reviewing, just because I don't have any time" is a very unprofessional way of maintaining open-source projects.

@GoodBoyDigital , may I ask you to give your feedback about the situation, is it the way the pixi.js lib maintained? Should I just stop trying to prepare older versions with important iOS14 bugfix?

Issue #6993, opened 2020-11-11
GlobalMixins Typings Don't Work


bigtimebuddy:
I published v5.4.0-rc.1 to try to identify issues with package-level typings. Overall it works great. There are some improvements we could make, like moving documentation blocks to the properties declarations in the class (historically these were added to the constructor assignment). 

ANYWAYS, the big issue is GlobalMixins class augmentation doesn't work. This is essential for doing mixins and extending existing classes like DisplayObject and Container. I created a simple project locally and I found that two things worked:

### Example

```typescript
import { Container } from '@pixi/display';
import '@pixi/mixin-cache-as-bitmap';

const stage = new Container();
stage.cacheAsBitmap = true; // <-- this fails
```

### Workarounds

1. Adding reference to the **global.d.ts** seemed to work in registering the GlobalMixins namespace.

```typescript
///<reference path="./global.d.ts" />
```

2. Adding `declare global` to the **index.d.ts** also worked.

```typescript
declare global {  namespace GlobalMixins { } }
```

The problem with both of these approaches is that api-extractor ignores both. We might need to add something custom to inject these to the output declaration, or find another solution.

Thoughts?

cc @ivanpopelyshev @SukantPal

SharasB:
I am having this exact issue in 7.0.2

PR #6999, opened 2020-11-12
inject global mixins into index.d.ts

Zyie:
##### Description of change

So here is a pretty basic solution for #6993 

~~All it does is append the `GlobalMixins` if it exists in `global.d.ts` to the `index.d.ts` file. 
Seems to work pretty well but feel free to close this if you guys come up with a better solution~~

New approach, now just injects a path that points to the `global.d.ts` file

example layout
```
/// <reference path="./global.d.ts" />
```


ivanpopelyshev (approved):

bigtimebuddy:
Yes!!! Thank you.

This feels like a bit of a hack, but good enough to unblock us. Couple of small organizational bits, could you:

* move this to a folder at the root level called scripts. The tools folder is for package-structured things.
* can you add some comments in your script about why this script is necessary? Having some comments like that are helpful so we don't have to dig thru GitHub later. Once api-extractor support this, we'd remove the hack.
* make sure eslint is considering this script.
* optionally, we could use ts-node and write the script in TS. Would be a nice, but unnecessary. Up to you if you wanna do this.

Zyie:
> This feels like a bit of a hack,
Yes definitely is a hack 😄 

> * optionally, we could use ts-node and write the script in TS. Would be a nice, but unnecessary. Up to you if you wanna do this.
Yeah i dont mind doing this!

bigtimebuddy:
Thanks @Zyie, you're my hero.

Zyie:
@bigtimebuddy updated!

bigtimebuddy:
Could you also check the bundles typings? Not sure how we might Mixin those global there too.

Zyie:
@bigtimebuddy ok so did a quick test with `pixi.js` and `pixi.js-legacy`, seems to be working now

all we do is bundle all the global mixins together and add them to the `index.d.ts`

also `canvas-renderer` had some extra declarations in its `global.d.ts` which has now been handled!

bigtimebuddy (approved):
Code looks good, Will try to give this a spin locally. Thanks for handling the feedback.

bigtimebuddy:
@SukantPal do you have any feedback on this approach?

SukantPal:
@bigtimebuddy Why are we appending `global.d.ts` when `///<reference path="./global.d.ts" />` seems to work? Why not just add the reference to the `index.d.ts` in the script? We will have to edit this script whenever `global.d.ts` breaks the definitions (which we may not even notice), as in the case of the canvas package.

Zyie:
> Why are we appending `global.d.ts` when `///<reference path="./global.d.ts" />` seems to work? Why not just add the reference to the `index.d.ts` in the script?

good point. I just did a test with adding

```
/// <reference path="../@pixi/display/global.d.ts" />
/// <reference path="../@pixi/mixin-cache-as-bitmap/global.d.ts" />
```
inside the `index.d.ts` of `pixi.js` and it works!

I'll update this PR to do this instead, its much simpler and less error prone

SukantPal:
@Zyie Could we add this into `@pixi-build-tools/lerna-monorepo-api-extractor`?

https://github.com/SukantPal/pixi-build-tools/tree/master/packages/api-extractor-lerna-monorepo

SukantPal:
If you want, we can bring that into the pixijs org @bigtimebuddy

Zyie:
@SukantPal @bigtimebuddy updated the code to use the path reference instead.

I'll move it over to the build tool if @bigtimebuddy is happy with that, just let me know!

bigtimebuddy:
I think we should try to get this in as-is. And then figure if it makes sense to add to build-tools. Ultimately this seems like something specific to this project but if there's a way to handle it in a general or explicit/config way, what's be good. I'll leave that up to @SukantPal to determine.

SukantPal:
@Zyie We can merge this and I'll do stuff in pixi-build-tools later

SukantPal (approved):

scripts/injectGlobalMixins.ts
```diff
@@ -0,0 +1,103 @@
...
+            pixiLegacyLocation = pkg.location;
+
+            return;
+        }
+
+        const globalDtsPath = path.resolve(basePath, './global.d.ts');
+
+        if (fs.existsSync(globalDtsPath))
+        {
+            const pixiTypeData = `/// <reference path="../${pkg.name}/global.d.ts" />\n`;
```
bigtimebuddy, line 79: Can you see if you can use `/// <reference types="${pkg.name}/global.d.ts" />`? It would be nice to avoid any relative pathing issues (for instance symlinked packages) and using node's resolve instead. Not sure if this works tho.

Zyie: ahhhh!!

this didn't work, however I have found this https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html

so we can do `/// <reference types="@pixi/core" />` inside `pixi.js`

Zyie: @bigtimebuddy updated!

bigtimebuddy:
Thank you @Zyie!

PR #7006, opened 2020-11-17
fix some types in 5.4.0-rc.2

Zyie:
##### Description of change

Fixes a couple of incorrect types after upgrading to to the pre-release

- `.from` function now have optional options
- export `TLoaderResource` as this is the type of `LoaderResource`
- make `context` optional in `Ticker.add`, `Ticker.remove`
- update `RenderTexture.baseTexture` to be type `BaseRenderTexture`
- make `options` optional in `destroy` functions for anything extending `DisplayObejct`

### DIFF
--- a/packages/canvas/canvas-renderer/src/CanvasRenderer.ts
+++ b/packages/canvas/canvas-renderer/src/CanvasRenderer.ts
@@ -360,7 +360,7 @@ export class CanvasRenderer extends AbstractRenderer
      *
      * @param {string} [clearColor] - Clear the canvas with this color, except the canvas is transparent.
      */
-    public clear(clearColor: string): void
+    public clear(clearColor?: string): void
     {
         const context = this.context;
 
--- a/packages/core/src/Renderer.ts
+++ b/packages/core/src/Renderer.ts
@@ -100,7 +100,7 @@ export class Renderer extends AbstractRenderer
      * @static
      * @private
      */
-    static create(options: IRendererOptions): AbstractRenderer
+    static create(options?: IRendererOptions): AbstractRenderer
     {
         if (isWebGLSupported())
         {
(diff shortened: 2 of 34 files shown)


themoonrat (approved):

bigtimebuddy:
What about using generics for RenderTexture... e.g. `extends Texture<BaseRenderTexture>` then you don't have to do some much casting and you can avoid overriding baseTexture. Otherwise this is great fixes.

bigtimebuddy (approved):
Optional comment above. Looks good.

Zyie:
@themoonrat @bigtimebuddy sorry guys i found a lot more, opened this a little early.  If you guys dont mind re-reviewing

@bigtimebuddy Sure ill look into making it a generic

Also does anyone know if `TextStyle.constructor` should be optional? 
`deepCopyProperties(this, style, style);` this part confused me

ivanpopelyshev (approved):

Zyie:
@bigtimebuddy ok made `Texture` generic

ivanpopelyshev:
> ok made Texture generic

wait wait wait WHAT?

yeah, its old idea to make it generic because BaseRenderTexture, but honestly, its rudiment. We had VideoBaseTexture before, but we removed it in favor of resources.

Its better to add interface for BaseRenderTexture fields in RenderTexture itself.

My advice is to make it in separate PR

ivanpopelyshev:
Alternative: add `castToRT()` into both Texture and BaseTexture

Zyie:
@ivanpopelyshev ok I have reverted this change so it doesn't block fixing the rest of the types

bigtimebuddy (approved):

bigtimebuddy:
Thank you @Zyie

Issue #7013, opened 2020-11-17
5.4.0-rc.2 Text type issues


andrewstart:
Issue 1: The constructor arguments for Text have not been marked as optional.
Issue 2: The `IDisplayObjectExtended` interface in the Prepare package disagree with `Text` about the `style` property - `IDisplayObjectExtended` specifies `TextStyle`, and won't allow `Text` to be used because `Text` is more lax with its `TextStyle | Partial<ITextStyle>`

PR #7015, opened 2020-11-18
Some fixes for 5.4.0

### LINKED ISSUE #7009: 5.4.0-rc2 OffscreenCanvas issue
In `InteractionManager.setCursorMode()`, there is a check for `OffscreenCanvas`:
```javascript
// offscreen canvas does not support setting styles
if (this.interactionDOMElement instanceof OffscreenCanvas) {
    return;
}
```
In browsers that don't support OffscreenCanvas yet (Firefox, Safari), this throws an error due to `window.OffscreenCanvas` being undefined. `if (self.OffscreenCanvas && this.interactionDOMElement instanceof OffscreenCanvas)` is the obvious fix, although I don't know if this issue would be present elsewhere in the OffscreenCanvas referencing code.

andrewstart:
Fixes #7009 and part of #7013 (the rest of that is fixed by #7006).

For the OffscreenCanvas change, I realized that bailing early would prevent the non-style cursor functionality from working, so changed it to be a more targeted check.

### DIFF
--- a/packages/interaction/src/InteractionManager.ts
+++ b/packages/interaction/src/InteractionManager.ts
@@ -1050,10 +1050,13 @@ export class InteractionManager extends EventEmitter
     public setCursorMode(mode: string): void
     {
         mode = mode || 'default';
-        // offscreen canvas does not support setting styles
-        if (this.interactionDOMElement instanceof OffscreenCanvas)
+        let applyStyles = true;
+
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Thanks @andrewstart!

ivanpopelyshev (approved):

bigtimebuddy:
Thanks @andrewstart

PR #7018, opened 2020-11-18
Fix memory leak with BitmapText

### LINKED ISSUE #7003: Memory leak with BitmapText
Updating bitmap text by doing bitmaptextObj.text = "newString" now has a memory leak. I believe this began in v5.3.3 and is still present in the newest rc 5.4.0.

https://www.pixiplayground.com/#/edit/AGAuOUo9ogkEvDnM3vxqf

It appears that every time BitmapText wants to update itself, it creates several new textures (line 410 here: https://github.com/pixijs/pixi.js/blame/v5.3.3/packages/text-bitmap/src/BitmapText.ts), which in turn spawn new events on the base texture.  @GoodBoyDigital, looks like you may be familiar with this section of code. See the below screenshots which show that the _events array on the base texture grows very large with many updates.

image

Then after about 20 seconds...
image

JetLua:
Fix #7003 

DEMO: https://jsfiddle.net/JetLu/w58mbhry/

The [original demo](https://www.pixiplayground.com/#/edit/AGAuOUo9ogkEvDnM3vxqf) memory footprint will increase over time.

image

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -913,7 +913,7 @@ export class BitmapText extends Container
         }
     }
 
-    destroy(options: boolean | IDestroyOptions): void
+    destroy(options?: boolean | IDestroyOptions): void
     {
         const { _textureCache } = this;
 
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -420,9 +420,14 @@ export class BitmapText extends Container
                 pageMeshData.vertexCount = 0;
                 pageMeshData.uvsCount = 0;
                 pageMeshData.total = 0;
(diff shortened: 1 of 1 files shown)


themoonrat (approved):

GoodBoyDigital (changes requested):
Heya! Sorry i'm a bit late to the party here. Rather than destroy / create each time - i think a better solution is to pull the texture from a cache? Or perhaps even just create a texture for the mesh in the constructor and reassign the base Texture if it changes?

GoodBoyDigital (approved):
nice one @JetLua  👍

bigtimebuddy:
The Texture should still get destroyed but probably when the BitmapText object is destroyed.

bigtimebuddy:
Thank you @JetLua. Please feel free to submit an expense to [Open Collective](https://opencollective.com/pixijs/expenses/new).

PR #7021, opened 2020-11-22
Fix bitmap font loader to load fonts with random placed args into <info> tag in 5.4.0-rc.2

### LINKED ISSUE #7008: BitmapFonts no longer getting processed by the Loader in 5.4.0-rc.2
My Bitmap Fonts no longer seem to be recognized after being loaded. I can see the .fnt file in Loader.resources, but PIXI.BitmapFont.available returns an empty object. This worked in 5.3.3.

SerG-Y:
##### Description of change

Fix #7008 
- For BitmapFont text format that's XML-based check if <font> tag exist than parse as XML and check if needed tags exist
- Added test for case when args into <info> tag placed randomly

##### Pre-Merge Checklist

### DIFF
--- a/packages/text-bitmap/src/formats/XMLStringFormat.ts
+++ b/packages/text-bitmap/src/formats/XMLStringFormat.ts
@@ -19,9 +19,14 @@ export class XMLStringFormat
      */
     static test(data: unknown): boolean
     {
-        const xmlStringRegExp = new RegExp(/<font>(\s+)?<info\s+face=/);
+        if (typeof data === 'string' && data.includes('<font>'))
+        {
+            const xml = new self.DOMParser().parseFromString(data, 'text/xml');
 
-        return typeof data === 'string' && xmlStringRegExp.test(data);
+            return XMLFormat.test(xml);
+        }
+
+        return false;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/text-bitmap/src/formats/XMLStringFormat.ts
```diff
@@ -19,9 +19,14 @@ export class XMLStringFormat
      */
     static test(data: unknown): boolean
     {
-        const xmlStringRegExp = new RegExp(/<font>(\s+)?<info\s+face=/);
+        if (typeof data === 'string' && data.includes('<font>'))
```
bigtimebuddy, line 22: Could you use `indexOf` instead of `includes`, since `includes` [is not ES5](https://caniuse.com/es6-string-includes) and not supported on IE11.

bigtimebuddy (changes requested):
Small change, otherwise looks good @SerG-Y

bigtimebuddy (approved):
Nice work :+1:

andrewstart (approved):

bigtimebuddy:
Thanks @SerG-Y! Feel free to [submit an expense](https://opencollective.com/pixijs/expenses/new) to Open Collective for $25 for solving this bug.

Issue #6975, opened 2020-11-03
The space that is followed by CJK should not be a breaking space in TextMetrics


huang-yuwei:
### Expected Behavior
The space that is followed by Chinese/Japanese/Korean should not be a breaking space.
Like how it works in DOM as well:
<img width="306" alt="スクリーンショット 2020-11-03 15 50 50" src="https://user-images.githubusercontent.com/44968790/97957181-dac91d00-1ded-11eb-9b26-9fc71180a011.png">
![export - 2020-11-03T160427 408](https://user-images.githubusercontent.com/44968790/97957476-9b4f0080-1dee-11eb-8723-edeb7afde4e4.png)

### Current Behavior
The space will be a breaking space no matter what the following character is. The breaking space works correctly in English, but it is not correct with Chinese/Japanese/Korean.
![export - 2020-11-03T160407 935](https://user-images.githubusercontent.com/44968790/97957348-47dcb280-1dee-11eb-8ac9-61576c58bbaa.png)

### Possible Solution
`TextMetrics.isBreakingSpace` may need to check the next character as well.
If the character is CJK, the space should not be a breaking space.
```
let isUnbreakSpace = false;

if (typeof nextChar === 'string')
{
    const matchedAsCJK = nextChar.match(/[\u3040-\u30ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uff66-\uff9f]/g);

    isUnbreakSpace = !!matchedAsCJK;
}

return (TextMetrics._breakingSpaces.indexOf(char.charCodeAt(0)) >= 0) && !isUnbreakSpace;
```

### Steps to Reproduce
```
const style = new PIXI.TextStyle({
    breakWords: true,
    fontSize: 13,
    fontWeight: "bold",
    lineJoin: "bevel",
    stroke: "#896161",
    whiteSpace: "pre-line",
    wordWrap: true,
    wordWrapWidth: 285
});
const text = new PIXI.Text('テストテキスト テストテキスト テストテキスト テストテキスト ', style);
```

### Environment

- **`pixi.js` version**: _e.g. 5.1.6
- **Browser & Version**: _e.g. Chrome 86
- **OS & Version**: _e.g. Mac OS 10.15

SukantPal:
Reproduction: https://codepen.io/sukantpal/full/dyXKWKy

The bounding box is a little smaller (^) than the box you're showing the "current behavior".

themoonrat:
[...]

PR #7023, opened 2020-11-23
Add overridable TextMetrics.isBreakingSpace for better CJK plugin support

huang-yuwei:
##### Description of change
As the discussion with @themoonrat in #6975, suggest allowing the `isBreakingSpace ` to be able to receive the `nextChar ` for solving the CJK issues in adding additions.

For example, the example plugin(https://codepen.io/huang-yuwei/pen/GRqbEmm) shows how we can solve the #6975 and #4447 via an adding addition.
However, I am not 100% sure whether the CJK related issues should be solved directly in PIXI.js itself.
Therefore, suggest applying the additional parameters for other developers can change via other plugins.

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/TextMetrics.ts
+++ b/packages/text/src/TextMetrics.ts
@@ -502,11 +502,16 @@ export class TextMetrics
     /**
      * Determines if char is a breaking whitespace.
      *
+     * It allows one to determine whether char should be a breaking whitespace
+     * For example certain characters in CJK langs or numbers.
+     * It must return a boolean.
+     *
      * @private
-     * @param  {string}  char - The character
+     * @param  {string}  char     - The character
+     * @param  {string}  [nextChar] - The next character
      * @return {boolean}  True if whitespace, False otherwise.
      */
-    private static isBreakingSpace(char: string): boolean
+    private static isBreakingSpace(char: string, _nextChar?: string): boolean
     {
         if (typeof char !== 'string')
         {
--- a/packages/text/src/TextMetrics.ts
+++ b/packages/text/src/TextMetrics.ts
@@ -506,7 +506,6 @@ export class TextMetrics
      * For example certain characters in CJK langs or numbers.
      * It must return a boolean.
      *
-     * @private
      * @param  {string}  char     - The character
      * @param  {string}  [nextChar] - The next character
      * @return {boolean}  True if whitespace, False otherwise.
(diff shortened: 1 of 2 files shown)


packages/text/src/TextMetrics.ts
```diff
@@ -502,11 +502,16 @@ export class TextMetrics
...
+     * It must return a boolean.
+     *
      * @private
-     * @param  {string}  char - The character
+     * @param  {string}  char     - The character
+     * @param  {string}  [nextChar] - The next character
      * @return {boolean}  True if whitespace, False otherwise.
      */
-    private static isBreakingSpace(char: string): boolean
+    private static isBreakingSpace(char: string, _nextChar?: string): boolean
```
bigtimebuddy, line 514: Can you make this public?

packages/text/test/TextMetrics.js
```diff
@@ -485,6 +485,31 @@ describe('PIXI.TextMetrics', function ()
 
             expect(bool).to.equal(false);
         });
+
+        it('overridable breaking spaces', function ()
+        {
+            const reg = /[あいうえお]/;
+
+            // override breakingSpace
+            TextMetrics.isBreakingSpace = function (char, nextChar)
```
bigtimebuddy, line 494: After this test is over, can you reset isBreakingSpace to the default. Otherwise it may have unintended side effects for other tests.

huang-yuwei:
@bigtimebuddy thank you for the review. I have updated the code 😊

bigtimebuddy (approved):
Code works for me! Thanks for the test too. I'll defer to @themoonrat to evaluate the utility of this.

packages/text/src/TextMetrics.ts
```diff
@@ -502,11 +502,16 @@ export class TextMetrics
     /**
      * Determines if char is a breaking whitespace.
      *
+     * It allows one to determine whether char should be a breaking whitespace
+     * For example certain characters in CJK langs or numbers.
+     * It must return a boolean.
+     *
      * @private
```
themoonrat, line 509: this private tag will also need to be remove

themoonrat (changes requested):
Only 1 minor jsdoc issue, otherwise lgtm. Non breaking change that allows greater plugin customising

packages/text/src/TextMetrics.ts
bigtimebuddy, line 509:
```suggestion
```

bigtimebuddy: Fixed

themoonrat (approved):

bigtimebuddy:
Thank you @huang-yuwei!

packages/text/src/TextMetrics.ts
huang-yuwei: I am sorry that I forgot about this, and thanks @bigtimebuddy for fixing it 🙏

Issue #7030, opened 2020-11-25
measureFont with error on some fonts


jflcarvalho:
### Expected Behavior
![Image 25-11-2020 at 14 11](https://user-images.githubusercontent.com/18697938/100238537-3b98d080-2f28-11eb-8c9f-d31603447e67.jpg)

### Current Behavior
![Image 25-11-2020 at 14 07](https://user-images.githubusercontent.com/18697938/100238034-a39ae700-2f27-11eb-9e64-5b0e93947416.jpg)

### Possible Solution
```javascript
    TextMetrics.BASELINE_MULTIPLIER = 1.75; // 1.75 instead of 1.4

    /**
     * Calculates the ascent, descent and fontSize of a given font-style
     *
     * @static
     * @param {string} font - String representing the style of the font
     * @return {PIXI.IFontMetrics} Font properties object
     */
    TextMetrics.measureFont = function (font) {
         [...]
         var metricsString = TextMetrics.METRICS_STRING + TextMetrics.BASELINE_SYMBOL;
         var textMetrics = context.measureText(metricsString);
     
         if (textMetrics.actualBoundingBoxAscent) {
             properties.ascent = Math.ceil(textMetrics.actualBoundingBoxAscent);
             properties.descent = Math.ceil(textMetrics.actualBoundingBoxDescent);
             properties.fontSize = properties.ascent + properties.descent;
             TextMetrics._fonts[font] = properties;
             return properties;
         }

        var canvas = TextMetrics._canvas;
        var width = Math.ceil(textMetrics.width);
        var baseline = Math.ceil(context.measureText(TextMetrics.BASELINE_SYMBOL).width);
        var height = 2.25 * baseline; // 2.25 instead of 2
        baseline = baseline * TextMetrics.BASELINE_MULTIPLIER | 0;
        [...]
    }
```

### Steps to Reproduce
Converting Webfont to BitmapFont using method:
```javascript
    BitmapFont.from(
                name,
                {
                    fontFamily: BebasNeue-Regular,
                    fontSize: 90,
                    fill: 'white'
                },
                {
                    chars
                }):
```

### Environment

- **`pixi.js` version**:  >= 5.1.
- **Browser & Version**: every
- **OS & Version**: every

PR #7031, opened 2020-11-25
Fix measureFont

jflcarvalho:
ISSUE: #7030  

##### Description of change
Fix measureFont and optimize to use modern browsers functionalities, in some cases the result properties are wrong.

##### Pre-Merge Checklist


themoonrat (changes requested):
I appreciate the attempt to have a fix of this.

A couple of worries I have tho

1: actualBoundingBoxAscent is marked by MDN as 'Experimental. Expect behavior to change in the future.' - https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics/actualBoundingBoxAscent - I'm unsure if this is something we'd want to put into PixiJS until it's a little more mature.
2: This is going to change how existing text looks. We can't afford to have a minor version of PixiJS effect the positioning, as there is a lot of content out there relying on it's current behaviour. Also, this will change formatting just from browser to browser, but even from browser version to browser version. Right now in PixiJS, all text renders the same across all browsers.

So for me, any fix needs to demonstrate that existing text is not effected, only 'buggy' text fixed, and that this is consistent across all browsers.

Cheers!

jflcarvalho:
Hi,

Yeah I understand, the use of actualBoundingBoxAscent it has just a try to improve the method's performance, and can be deleted.

I tried on some developments that I have on my machine with safari and chrome and it has everything at same place as before, but I can't test on every browser :/

With that I just propose a increase on canvas height and Y placement of the textString to solve this issue.

bigtimebuddy:
The TextMetrics are not that precise. That's why we have overrides like Baseline Multiplier so that users and adjust for their particular
typefaces. I believe that describes your case. We choose these defaults as accommodating most (but not all) circumstances. Changing those values is a breaking change. 

Why not create a new variable for the 2x (2.25x) value and then you should be able to solve your issue without the other changes.

jflcarvalho:
As requested the new changes are applied, I have create a new static variable to defining height multiplier of canvas, with the default value being 2.0x.

packages/text/src/TextMetrics.ts
```diff
@@ -643,15 +644,16 @@ export class TextMetrics
...
-        const canvas = TextMetrics._canvas;
         const context = TextMetrics._context;
 
         context.font = font;
-
         const metricsString = TextMetrics.METRICS_STRING + TextMetrics.BASELINE_SYMBOL;
-        const width = Math.ceil(context.measureText(metricsString).width);
+
+        const textMetrics = context.measureText(metricsString);
+        const canvas = TextMetrics._canvas;
```
themoonrat, line 653: slightly pedantic, but if you move
this const canvas = TextMetrics._canvas;
line back to where it was, about the  const context = TextMetrics._context; line, it reduces the diff down to just the changes required, and reads a bit better.

themoonrat (changes requested):
Just a little comment of reducing the diff a bit, but I think this is a good strategy. Keeps current compatibility, whilst allows developers to make the call themselves to measure in a way more appropriate for themselves.
Thanks!

packages/text/src/TextMetrics.ts
jflcarvalho: Yah sure!

jflcarvalho:
@themoonrat the requested changes are done!

themoonrat (approved):

bigtimebuddy (approved):
Thank you!

bigtimebuddy:
Thank you @jflcarvalho!

PR #7032, opened 2020-11-25
Fix serious performance problem in canvas renderer

### LINKED ISSUE #7027: Big Drop in Performance from 5.2.0 to 5.3.3 when using CanvasRenderer
I've discovered that moving from version 5.2.0 to 5.3.3 has caused a major drop in FPS when using the Canvas fallback. The two codepens below demonstrate this. They both create a large number of graphics objects, draw circles into them, and then randomly move them all around each frame. I get about 17fps on 5.3.3, and about 35fps on 5.2.0.

You can tweak CIRCLES_TO_CREATE if one of the tests reaches your monitors refresh rate, or if either test runs too slowly.

I'm not sure exactly which PIXI version caused the drop in performance, in between 5.2.0 and 5.3.3.

5.2.0

5.3.3

Thanks!

ivanpopelyshev:
Fix #7027 

Includes unit test.

Aaaand the "Grand Fail of the Year" award goes to Ivan Igorevich!

### DIFF
--- a/packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
+++ b/packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
@@ -1,3 +1,4 @@
+import { Texture } from '@pixi/core';
 import { SHAPES, Matrix } from '@pixi/math';
 import { canvasUtils } from '@pixi/canvas-renderer';
 
@@ -52,7 +53,7 @@ export class CanvasGraphicsRenderer
     {
         let res;
 
-        if (style.texture)
+        if (style.texture && style.texture.baseTexture !== Texture.WHITE.baseTexture)
         {
             if (style.texture.valid)
             {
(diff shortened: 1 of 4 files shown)


andrewstart (approved):

ivanpopelyshev:
I recommend to backport it to 5.2 too

bigtimebuddy (approved):
Verified this fixed the issue. Thanks for the test too :+1:

GoodBoyDigital (approved):
haha! oops :P nice catch there!

bigtimebuddy:
Thanks for hopping on this so quickly @ivanpopelyshev

vwmberry95:
I updated to 5.3.4 and I'm still seeing FPS cut in half after switching focus.

I have an Android app loading a webview with a site using Pixi.  Stays around 60 FPS at first.  When I switch focus to another fragment with a separate webview not using Pixi and then switch back, it stays around 30 FPS.

PR #7039, opened 2020-11-30
Fix #7035

### LINKED ISSUE #7035: Unexpected behavior when rendering object with filters to texture
### Current Behavior

When rendering an object to a texture using the matrix transform parameter of the render function, the transformation is applied only as one would expect if there are no filters on the object, or, if the parameter is not used, but instead the object is transformed directly. If there are filters on the object, the transformation matrix transforms the object differently; there is even a difference in transformation between different filters, and the number of filters that are applied affects the transformation as well. It looks like the transform is applied again for each filter, at least the scaling part.

image
image
image
image
image
image

### Expected Behavior

If the transformation is applied to the object directly and the matrix transform parameter is not used, everything works as expected. The same should be true if the matrix transform parameter is used.

image
image
image
image
image
image

### Steps to Reproduce

Check out the demo [here](https://www.pixiplayground.com/#/edit/H2h30v8g6Qu7wIqXW1NWu).

### Environment

- **Running Example**: https://www.pixiplayground.com/#/edit/H2h30v8g6Qu7wIqXW1NWu

dev7355608:
##### Description of change

Fixes #7035.

1. The transform must not be applied in the pop phase; only in push.
2. The source frame must be transformed as well if no filter area is given, i.e., the bounds must be transformed.
3. The matrix transforms the object, not the render texture; documentation was changed.

##### Pre-Merge Checklist

- [x] Documentation is changed

### DIFF
--- a/packages/core/src/Renderer.ts
+++ b/packages/core/src/Renderer.ts
@@ -379,7 +379,7 @@ export class Renderer extends AbstractRenderer
      * @param displayObject - The object to be rendered.
      * @param [renderTexture] - The render texture to render to.
      * @param [clear=true] - Should the canvas be cleared before the new render.
-     * @param [transform] - A transform to apply to the render texture before rendering.
+     * @param [transform] - A transform to apply to the object before rendering.
      * @param [skipUpdateTransform=false] - Should we skip the update transform pass?
      */
     render(displayObject: DisplayObject, renderTexture?: RenderTexture,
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
This is awesome, but we have to move rectangle stuff to utils.

ivanpopelyshev:
ideally, this method should be in Bounds, but currently Bounds are in `@pixi/display` and core does not depend on it. We can put it in private system of FilterSystem for now.

SukantPal:
Can you create a demo?

bigtimebuddy:
Thanks @dev7355608 for digging into this and coming up with a solution. I'm going to close this in favor of #7044, but I'll make sure to credit you for the fix once it's merged.

PR #7040, opened 2020-11-30
Fix #7036

### LINKED ISSUE #7036: Objects with filters are rendered in lower quality if the renderer's resolution is greater than 1
### Current Behavior

If the app's (or renderer's) resolution is set to anything other than 1, e.g. if `window.devicePixelRatio` is used and greater than 1, then filters with (default) resolution 1, render to textures of relatively lower resolution. Consequently the visual quality degrades as soon as a filter is applied and the filter's resolution isn't adjusted to the renderer's resolution. This behavior is best shown by the demo below.

image

### Expected Behavior

The filter's resolution should be relative to renderer's/render texture's resolution. Manually setting the resolution of every filter used to the same resolution as the renderer doesn't seem like the right way to go, neither solving the problem by changing the settings globally like this, which at the moment is the best way to do it as far as I can tell:

```js
PIXI.settings.RESOLUTION = window.devicePixelRatio;
PIXI.settings.FILTER_RESOLUTION = window.devicePixelRatio;
```

### Possible Solution

After the resolution is determined that is used for the filters in `FilterSystem.push()`, we need to multiply it by the renderer's/render texture's resolution.

```js
if (renderer.renderTexture.current) {
    // Rendering to a texture...
    const baseTexture = renderer.renderTexture.current.baseTexture as BaseRenderTexture;
    resolution *= baseTexture.resolution;
} else {
    // Rendering to the screen...
    resolution *= renderer.resolution;
}
```

I haven't tested it yet, but it seems like this would work.

### Steps to Reproduce

Check out the demo [here](https://www.pixiplayground.com/#/edit/KXGbyIOJzQBv7MQVT_8QT). To see it, you have to zoom in of course. Zoom in as far as possible and click _Refresh_.

### Environment

- **Running Example**: https://www.pixiplayground.com/#/edit/KXGbyIOJzQBv7MQVT_8QT

dev7355608:
##### Description of change

Fixes #7036.

Changes the filter's resolution from absolute to relative to the current render target's resolution.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/filters/FilterSystem.ts
+++ b/packages/core/src/filters/FilterSystem.ts
@@ -12,6 +12,7 @@ import type { Filter } from './Filter';
 import type { IFilterTarget } from './IFilterTarget';
 import type { ISpriteMaskTarget } from './spriteMask/SpriteMaskFilter';
 import type { RenderTexture } from '../renderTexture/RenderTexture';
+import type { BaseRenderTexture } from '../renderTexture/BaseRenderTexture';
 import type { Renderer } from '../Renderer';
 
 /**
@@ -179,6 +180,15 @@ export class FilterSystem extends System
             autoFit = autoFit || filter.autoFit;
 
             legacy = legacy || filter.legacy;
+        }
+
+        if (renderer.renderTexture.current)
+        {
+            resolution *= (renderer.renderTexture.current.baseTexture as BaseRenderTexture).resolution;
+        }
+        else
+        {
+            resolution *= renderer.resolution;
         }
 
         if (filterStack.length === 1)


ivanpopelyshev:
For now all resolutions were absolute, not relative. I had idea to use multiplication , but its better if @bigtimebuddy clarifies whether we can make another API to track absolute resolution or relative in filters

SukantPal:
I'd actually discourage this b/c resolution, until now, had a very simple definition: the ratio of physical pixels to CSS pixels in the rendering pipeline.

bigtimebuddy:
Thanks for the PR @dev7355608, but I'm going to close this. I don't think it's the right solution is to make the resolution relative to the renderer. it's a pretty significant change in behavior with respect to filter resolutions.

That said, I think how Pixi manages resolution everywhere could be improved. It's a little cumbersome to have static globals, a renderer resolution, texture/resource resolutions, AND local display object resolution properties (e.g. Text). If anyone has a proposal to harmonize this mess to be simpler and more streamlined, would welcome it!

PR #7041, opened 2020-11-30
Fix #7038

### LINKED ISSUE #7038: Incorrect framebuffer viewport caused by rounding errors
This was uncovered after the fix in #7036 was applied. This issue occurs only if the resolution is not 1.

image

The effect gets worse as more filters are applied (something like 0.5-1.0 pixels per active filter).

### Environment

- **Running Example**: https://www.pixiplayground.com/#/edit/4rHbrZxn1pnIv7CYFV47d

dev7355608:
##### Description of change

Fixes #7038.

1. Changed calculations in `Rectangle.ceil()` to improve precision of width and height. This is technically not necessary if (2) is applied.
2. Ensure the viewport frame is on the grid, i.e., integer. `gl.viewport()` converts it's arguments to `GLint`/`GLsizei` as `float` would be in C. That's why tiny rounding errors have an undesirable effect. This is an another bug that has actually the opposite effect of (3), which is not visible in the pictures in #7038.

```js
gl.viewport(64.00001, 63.99999, 128.00001, 127.99999);
console.log(gl.getParameter(gl.VIEWPORT)); // => Int32Array [64, 63, 128, 127]
````
image

3. Fixed a bug in `RenderTexture.resize()` that set the width/height wrong that not only caused the flop texture to double in size, but caused what is depicted in #7038.
4. The canvas source frame has now the correct dimensions in case of non-integer resolutions, because `screen` is slightly too large in that case: `view.width == Math.floor(screen.width * resolution)` and `view.height == Math.floor(screen.height * resolution)`.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/renderTexture/RenderTextureSystem.ts
+++ b/packages/core/src/renderTexture/RenderTextureSystem.ts
@@ -155,8 +155,8 @@ export class RenderTextureSystem extends System
 
             if (!sourceFrame)
             {
-                tempRect.width = renderer.screen.width;
-                tempRect.height = renderer.screen.height;
+                tempRect.width = renderer.view.width / renderer.resolution;
+                tempRect.height = renderer.view.height / renderer.resolution;
 
                 sourceFrame = tempRect;
             }
@@ -181,6 +181,8 @@ export class RenderTextureSystem extends System
         {
             viewportFrame.y = renderer.view.height - (viewportFrame.y + viewportFrame.height);
         }
+
+        viewportFrame.ceil();
 
         this.renderer.framebuffer.bind(framebuffer, viewportFrame);
         this.renderer.projection.update(destinationFrame, sourceFrame, resolution, !framebuffer);
--- a/packages/core/src/renderTexture/RenderTexture.ts
+++ b/packages/core/src/renderTexture/RenderTexture.ts
@@ -133,9 +133,6 @@ export class RenderTexture extends Texture
      */
     resize(width: number, height: number, resizeBaseTexture = true): void
     {
-        width = Math.ceil(width);
-        height = Math.ceil(height);
-
         // TODO - could be not required..
         this.valid = (width > 0 && height > 0);
 
--- a/packages/math/src/shapes/Rectangle.ts
+++ b/packages/math/src/shapes/Rectangle.ts
@@ -236,14 +236,16 @@ export class Rectangle
      */
     ceil(resolution = 1, eps = 0.001): this
     {
-        const x2 = Math.ceil((this.x + this.width - eps) * resolution) / resolution;
-        const y2 = Math.ceil((this.y + this.height - eps) * resolution) / resolution;
+        const x1 = Math.floor((this.x + eps) * resolution);
+        const y1 = Math.floor((this.y + eps) * resolution);
+        const x2 = Math.ceil((this.x + this.width - eps) * resolution);
+        const y2 = Math.ceil((this.y + this.height - eps) * resolution);
 
-        this.x = Math.floor((this.x + eps) * resolution) / resolution;
-        this.y = Math.floor((this.y + eps) * resolution) / resolution;
+        this.x = x1 / resolution;
+        this.y = y1 / resolution;
 
-        this.width = x2 - this.x;
-        this.height = y2 - this.y;
+        this.width = (x2 - x1) / resolution;
+        this.height = (y2 - y1) / resolution;
 
         return this;
     }

### CODE REVIEW

**Comment by ivanpopelyshev** [comment]
Need to find previous commit, where that `ceil` was introduced in that place. There was another demo, I remember

X-Files is my label dedicated to those problems.

**Comment by bigtimebuddy** [comment]
@ivanpopelyshev `ceil` was added here #4722

**Comment by dev7355608** [comment]
Could you give this a higher priority as well? I noticed that it's not always just 1 pixel offset per filter as mentioned in #7038. In the case of just one blur filter the results can be extreme. https://www.pixiplayground.com/#/edit/GTjTs7MveUFknBd-FRR2r

image

File: packages/core/src/renderTexture/RenderTextureSystem.ts [hunk 1]
```diff
@@ -155,8 +155,8 @@ export class RenderTextureSystem extends System
 
             if (!sourceFrame)
             {
-                tempRect.width = renderer.screen.width;
-                tempRect.height = renderer.screen.height;
+                tempRect.width = renderer.view.width / renderer.resolution;
```
SukantPal, line 158: This seems unnecessary - `screen.width` should always equal `view.width / renderer.resolution`.

dev7355608: Now always, because the framebuffer dimensions can't be noninteger of course. For example `width = 101` and `resolution = 1.5`: the framebuffer can't be `151.5`, so it becomes `151` instead (https://www.pixiplayground.com/#/edit/ovoCtZn-zaHjGp1ki5bAJ). That's why `screen.width` can be slightly larger in some cases. Small difference, but, if we omit this change, we get a slightly blurry image.

SukantPal: `renderer.view.width` should be `151.5` in your case. It isn't rounded.

https://github.com/pixijs/pixi.js/blob/3f048344f63a436bad40b0d0542a34bd8dc692b9/packages/core/src/AbstractRenderer.ts#L273

SukantPal: Non-integer resolutions are not supported for this reason anyway. Maybe we should throw an error if the `view.width` turns out to be a non-integer and make that explicit.

dev7355608: > 
> 
>

> Non-integer resolutions are not supported for this reason anyway. Maybe we should throw an error if the `view.width` turns out to be a non-integer and make that explicit.

Isn't `window.devicePixelRatio` used as resolution for the renderer in many cases? That would break a lot of code.

> 
> 
> `renderer.view.width` should be `151.5` in your case. It isn't rounded.
> 
> https://github.com/pixijs/pixi.js/blob/3f048344f63a436bad40b0d0542a34bd8dc692b9/packages/core/src/AbstractRenderer.ts#L273

Check the assert's in the playground example. It's `151`. It isn't rounded by pixi itself but in `HTMLCanvasElement` or elsewhere.

SukantPal: Okay, so (/cc @ivanpopelyshev) I think we should strictly enforce resolutions that cause non-integer framebuffer dimensions.

ivanpopelyshev: > This seems unnecessary - screen.width should always equal view.width / renderer.resolution.

We should support this thing. That's why i mark all those issues as X-Files, it takes a bit of fiddling to get floors/ceils/rounds right

dev7355608: We could/should change `AbstractRenderer.resize()`:

```js
    /**
     * Resizes the screen and canvas as close as possible to the specified width and height.
     * Canvas dimensions are multiplied by resolution and rounded to the nearest integers.
     * The new canvas dimensions divided by the resolution become the new screen dimensions.
     *
     * @param desiredScreenWidth - The desired width of the screen.
     * @param desiredScreenHeight - The desired height of the screen.
     */
    resize(desiredScreenWidth: number, desiredScreenHeight: number): void
    {
        this.view.width = Math.round(desiredScreenWidth * this.resolution);
        this.view.height = Math.round(desiredScreenHeight * this.resolution);

        const screenWidth = this.view.width / this.resolution;
        const screenHeight = this.view.height / this.resolution;

        this.screen.width = screenWidth;
        this.screen.height = screenHeight;

        if (this.autoDensity)
        {
            this.view.style.width = `${screenWidth}px`;
            this.view.style.height = `${screenHeight}px`;
        }

        /**
         * Fired after view has been resized.
         *
         * @event PIXI.Renderer#resize
         * @param {number} screenWidth - The new width of the screen.
         * @param {number} screenHeight - The new height of the screen.
         */
        this.emit('resize', screenWidth, screenHeight);
    }
```

bigtimebuddy:
@ivanpopelyshev @SukantPal what do you want to do about this PR? Been here for awhile would like to come to some conclusion.

dev7355608:
I actually identified a bunch of rounding related issues a while back, but I never got around to push it and eventually forgot. I'm not sure if this PR is the right place to push it to. Let me know how you would like to see it.

Issue #6967, opened 2020-10-27
Scaling container causes it's children to "wiggle/shake" if it has AlphaFilter


IljaDaderko:
I will attach gif with example behaviour that I am seeing, but in essence on some screen widths while I'm animating (I use gsap3 with pixi plugin) scale of a parent container I can see it's children "wiggle" / "shake". I've tested on several devices and browsers and can replicate this. You will see in the gif that I am also animatin opacity, this is done using `AlphaFilter`. As soon as I remove this filter and just run scale animation, everything stays solid / crips with no issues. I can also animate just `alpha` of container, however that causes issues mentioned in #4334. Wiggle / shake issue still persists even if I am not animating AlphaFilter i.e. if it is just applied to container.

### Expected Behavior
Should scale without shake / wiggle despite different screen dimensions, if I am correct?

### Current Behavior
![Screen Recording 2020-10-27 at 22 45 12](https://user-images.githubusercontent.com/3154053/97360046-5256f380-18a6-11eb-9a2e-00c3fe3aaf07.gif)

### Possible Solution
Not sure

### Steps to Reproduce
I wanted to check if this is a known behaviour first, if not I'll make a codesandbox to reproduce

### Environment
- **`pixi.js` version**: 5.3.3
- **Browser & Version**: Chrome 86.0.4240.111, Safari 14.0
- **OS & Version**: macOS Catalina 10.15.6
- **Running Example**: -

IljaDaderko:
Not sure if related, but container who's scale is being animated is also having its AlphaFilter animated at the same time

IljaDaderko:
So, what's interesting is that if I set `PIXI.settings.FILTER_RESOLUTION` to be 2x larger than than my `PIXI.Application` resolution, it kind of looks ok, but I assume there is performance hit here?

bigtimebuddy:
I think this is a duplicate: #3742

IljaDaderko:
Sounds like it, however I applied high precision and am still seeing this issue, perhaps some of my other options are causing it?

```ts
const { devicePixelRatio, innerWidth, innerHeight } = window;

PIXI.settings.PRECISION_FRAGMENT = PIXI.PRECISION.HIGH;
PIXI.settings.FILTER_RESOLUTION = devicePixelRatio;

const app = new PIXI.Application({
  resolution: devicePixelRatio,
  autoDensity: true,
  width: innerWidth,
  height: innerHeight,
  view: document.getElementById('app-canvas'),
[...]

PR #7044, opened 2020-12-01
Fix transform filters

ivanpopelyshev:
Alternative to #7039 

We'll remove new private transformAABB function in favor of `Bounds.addFrameMatrix` after we move `Bounds` to `@pixi/math`

### DIFF
--- a/packages/core/src/filters/FilterSystem.ts
+++ b/packages/core/src/filters/FilterSystem.ts
@@ -174,14 +174,21 @@ export class FilterSystem extends System
 
         state.target = target;
 
-        state.sourceFrame.copyFrom(target.filterArea || target.getBounds(true));
-
-        const transform = renderer.projection.transform;
-
-        if (transform)
+        if (target.filterArea)
         {
-            // TODO: use Bounds.addFrameMatrix instead
-            this.transformAABB(transform, state.sourceFrame);
+            state.sourceFrame.copyFrom(target.filterArea);
+        }
+        else
+        {
+            state.sourceFrame.copyFrom(target.getBounds(true));
+
+            const transform = renderer.projection.transform;
+
+            if (transform)
+            {
+                // TODO: use Bounds.addFrameMatrix instead
+                this.transformAABB(transform, state.sourceFrame);
+            }
         }
 
         state.sourceFrame.pad(padding);
--- a/packages/core/src/filters/FilterSystem.ts
+++ b/packages/core/src/filters/FilterSystem.ts
@@ -2,7 +2,7 @@ import { System } from '../System';
 import { RenderTexturePool } from '../renderTexture/RenderTexturePool';
 import { Quad } from '../utils/Quad';
 import { QuadUv } from '../utils/QuadUv';
-import { Rectangle, Matrix } from '@pixi/math';
+import { Rectangle, Matrix, Point } from '@pixi/math';
 import { UniformGroup } from '../shader/UniformGroup';
 import { DRAW_MODES, CLEAR_MODES } from '@pixi/constants';
 import { deprecation } from '@pixi/utils';
@@ -13,6 +13,9 @@ import type { IFilterTarget } from './IFilterTarget';
 import type { ISpriteMaskTarget } from './spriteMask/SpriteMaskFilter';
 import type { RenderTexture } from '../renderTexture/RenderTexture';
 import type { Renderer } from '../Renderer';
+
+const tempPoints = [new Point(), new Point(), new Point(), new Point()];
+
 /**
  * System plugin to the renderer to manage the filters.
  *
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
One more thing: there was a broken test, i believe its something connected to previous PR's, and it should fix #6967

packages/core/src/filters/FilterSystem.ts
```diff
@@ -173,6 +176,14 @@ export class FilterSystem extends System
 
         state.sourceFrame.copyFrom(target.filterArea || target.getBounds(true));
 
+        const transform = renderer.projection.transform;
+
+        if (transform)
+        {
+            // TODO: use Bounds.addFrameMatrix instead
+            this.transformAABB(transform, state.sourceFrame);
```
dev7355608, line 184: We must not transform the source frame if the filter area is specified if I'm not mistaken. Only the bounds should be transformed.

ivanpopelyshev: Yeah, whether filterArea is global coords or screen (after transform) is certainly a question. Consider we assign it to "renderer.screen"...

ivanpopelyshev: OK, I agree, changing it. Just because i put `filterArea = renderer.screen` as an example. everywhere.

ivanpopelyshev:
Here's example to test filterArea case (broken for now): https://www.pixiplayground.com/#/edit/QKhHOtiDw3BSQ_KFzUfaC

Because I put `element.filterArea = app.screen` as fullscreen filter example everywhere, filterArea is not the same as bounds, its real screen coords, while bounds are global coords.

bigtimebuddy (approved):
**Broken** (v5.3.3)
https://jsfiddle.net/bigtimebuddy/bpufc8hd/

**Fixed** (this PR)
https://jsfiddle.net/bigtimebuddy/s1uzyLb8/

dev7355608:
> `@param [transform] - A transform to apply to the render texture before rendering.`

`Renderer.render`'s documentation needs to be adjusted. The object (or canvas/world space) is transformed, which is the same as the render texture being transformed by the inverse transform.

dev7355608:
I noticed that we might need the transform in `Filter.apply()` if we want to do something similar as in `FilterSystem.calculateSpriteMatrix()`. So it might not be a good idea to set simply set it to null in pop. Maybe expose the transform in `FilterState`.

Untested pseudocode example, might not work as intended. What I want is a matrix that maps to local normalized coordinates. 

```js
public apply(
        filterManager: FilterSystem, input: RenderTexture, output: RenderTexture, clearMode: CLEAR_MODES
    ): void
    {
        const { sourceFrame, destinationFrame, transform, target } = this.activeState;
        const localBounds = target.getLocalBounds();
        const mappedMatrix = new Matrix(destinationFrame.width, 0, 0,
            destinationFrame.height, sourceFrame.x, sourceFrame.y);
        const worldTransform = target.worldTransform.copyTo(Matrix.TEMP_MATRIX);

        // We need the transform here to get the "true" world transform
        if (transform)
        {
            worldTransform.preprend(transform);
        }

        worldTransform.invert();
        mappedMatrix.prepend(worldTransform);
        mappedMatrix.translate(-localBounds.x, -localBounds.y);
        mappedMatrix.scale(1.0 / localBounds.width, 1.0 / localBounds.height);

        // filterMatrix maps vTextureCoord to local normalized coordinates
        this.uniforms.filterMatrix = mappedMatrix;

        ...
}
```

creativefctr:
Guys, I am not an expert on Pixi internals, but can my issue here be relevant to this one?
https://github.com/pixijs/pixi-filters/issues/276

bigtimebuddy:
@ivanpopelyshev could you respond to the feedback here? This is high-priority to get in

SukantPal:
I think the approach to preserving the projection transform is incorrect in this PR. The filter-system should eliminate the need of the projection transform by "unapplying" in push() and "applying" it in pop(). Also, the filterArea should be in the world space because the area (relative to other objects) being filtered shouldn't change under a projection transform.

This is going to be hard to explain, so bear with me.

### Current approach

* The source frame (input) of the filter system is projected according to the projection transform.
* The output frame (of the last filter) is not projected.

The problem with is that the additional adjustments required by @dev7355608. The rendering pipeline works with projected coordinates, instead of world space coordinates. A better approach would that the filter system works in world space and then projects the output texture once it is done.

### My approach

* The projection transform is saved, and then removed from the projection system
* If autoFit is used, the source-frame is inverse projected into world space and intersected with the filter-area/object-bounds
* Rendering goes through in world space into the input texture
* The filtering pipeline runs until the last filter.
* `bindAndClear` sets the projection transform when binding to the output texture (`defaultFilterStack[defaultFilterStack.length - 1].renderTexture`)

I tested and this approach has no issue with the sprite masking, unlike this PR. See the following fiddle that modifies @bigtimebuddy's "successful" AlphaMaskFilter fiddle by adding a sprite mask.

* https://jsfiddle.net/ShukantPal/v3dtcohp/

I'll make another PR with this suggestion.

SukantPal:
@ivanpopelyshev Closing in favor of #7050, as we discussed.

Issue #7011, opened 2020-11-17
5.4.0-rc.2 external library typing issues


andrewstart:
The following issues related to libraries bundled with PixiJS are reported:
```
node_modules/@pixi/utils/index.d.ts(2,37): error TS7016: Could not find a declaration file for module 'earcut'. 'D:/github/pixi-animate/node_modules/earcut/src/earcut.js' implicitly has an 'any' type.
  Try `npm install @types/earcut` if it exists or add a new declaration (.d.ts) file containing `declare module 'earcut';`
node_modules/@pixi/utils/index.d.ts(8,32): error TS7016: Could not find a declaration file for module 'url'. 'D:/github/pixi-animate/node_modules/url/url.js' implicitly has an 'any' type.
  Try `npm install @types/url` if it exists or add a new declaration (.d.ts) file containing `declare module 'url';`
node_modules/mini-signals/typings/mini-signals.d.ts(19,2): error TS2309: An export assignment cannot be used in a module with other exported elements.
node_modules/pixi.js/index.d.ts(11,23): error TS2688: Cannot find type definition file for '@pixi/graphics-extras'.
node_modules/resource-loader/typings/resource-loader.d.ts(70,17): error TS2304: Cannot find name 'Signal'.
```
Additionally:
* the `Loader` class is missing type information for its `once()` method.
* An area of improvement would be to change the type output to import EventEmitter via `import EventEmitter = require('eventemitter3');`, as the ES6 import requires consumers to change their `esModuleInterop` setting to true.

bigtimebuddy:
@Zyie is this something you might be able to look at?

bigtimebuddy:
Fixed

PR #6849, opened 2020-08-30
Use @pixi-build-tools/api-extractor-lerna-monorepo for typings!

SukantPal:
# Brief 🎢 

~~Blocked by #6848~~
Blocked by #6857 

This will use [@pixi-build-tools/api-extractor-lerna-monorepo](https://www.npmjs.com/package/@pixi-build-tools/api-extractor-lerna-monorepo) to generate typings in our monorepo using [@microsoft/api-extractor](https://www.npmjs.com/package/@microsoft/api-extractor).

## Caveats 😢 

* ~~`pixi.js` and `pixi.js-legacy` fail because "Unsupported export defaultVertex"~~

```
Analysis will use the bundled TypeScript version 3.9.5
pixi.js-legacy failed:
Error: Unsupported export "defaultVertex":
/Users/shukantpal/Web Projects/pixi.js/global.d.ts:8:11
    at ExportAnalyzer.fetchAstModuleFromSourceFile (/Users/shukantpal/Web Projects/pixi.js/node_modules/@microsoft/api-extractor/lib/analyzer/ExportAnalyzer.js:74:35)
    at ExportAnalyzer._fetchSpecifierAstModule (/Users/shukantpal/Web Projects/pixi.js/node_modules/@microsoft/api-extractor/lib/analyzer/ExportAnalyzer.js:569:41)
    at ExportAnalyzer.fetchAstModuleFromSourceFile (/Users/shukantpal/Web Projects/pixi.js/node_modules/@microsoft/api-extractor/lib/analyzer/ExportAnalyzer.js:90:65)
    at AstSymbolTable.fetchAstModuleFromWorkingPackage (/Users/shukantpal/Web Projects/pixi.js/node_modules/@microsoft/api-extractor/lib/analyzer/AstSymbolTable.js:57:37)
    at Collector.analyze (/Users/shukantpal/Web Projects/pixi.js/node_modules/@microsoft/api-extractor/lib/collector/Collector.js:117:51)
    at Function.invoke (/Users/shukantpal/Web Projects/pixi.js/node_modules/@microsoft/api-extractor/lib/api/Extractor.js:102:19)
    at /Users/shukantpal/Web Projects/pixi.js/node_modules/@pixi-build-tools/api-extractor-lerna-monorepo/index.js:92:37
    at Array.forEach (<anonymous>)
    at main (/Users/shukantpal/Web Projects/pixi.js/node_modules/@pixi-build-tools/api-extractor-lerna-monorepo/index.js:41:16)
```

[...]

PR #7045, opened 2020-12-02
Fix typescript definition errors

Zyie:
##### Description of change

At the moment this PR partially addresses the issues raised in #7011, However there are a few issues that need ironing out with some help.

### Fixed:
- `Could not find a declaration file for module 'earcut'`
- `Cannot find type definition file for '@pixi/graphics-extras'`
- `IRendererOptions` not being defined in `global.d.ts` for `@pixi/core`
- `onChildrenChanged` in `Container` missing optional parameter

### Loader class is missing type information for its `once()` method
This method appears to be deprecated, I found an old PR that explicitly removed deprecated method from types #6163
Is this something we want to add back in? @bigtimebuddy

### error TS2304: Cannot find name 'Signal'
This error is being caused by a conflict in types between `resource-loader` and `mini-signals`, both of which are external to us.
`resource-loader` seems to be trying to overwrite the types provided by `mini-signals` and it does not like that.

I found this [issue](https://github.com/englercj/resource-loader/issues/118) that ran into the same issue however no solution was posted.

In #6849 `skipLibCheck` was turned on in order to get types to build properly for us.

An alternative to turning off all type checking is to do something like this
```
"paths": {
    "mini-signals": [
        "node_modules/resource-loader/typings/mini-signals.d.ts"
    ]
},
```
This keeps types checking but would once again requires external libraries to include this.

So I'm not sure what the proper/best solution for this would be.

### esModuleInterop
I've implemented the easy fix of swapping `esModuleInterop` out for `allowSyntheticDefaultImports` which means external libraries code wont get transpiled differently but they will still need to turn this on to skip type checking.

As @andrewstart said it would be better to not have to turn on any settings. I tried changing the imports to `require` as suggested however this led to more errors being produced. So if anyone knows how to fix these i'll update the PR

These are the following imports/exports that require `allowSyntheticDefaultImports`
```
import EventEmitter from 'eventemitter3';
export { default as earcut } from 'earcut';
export { default as url } from 'url';
import { default as cssColorNames } from 'css-color-names';
```


andrewstart:
What's here looks good so far. I hadn't realized that `once()` was deprecated on Loader - that isn't too important to fix then. However, since we removed typings of deprecated methods, Typescript now recognizes the `@deprecated` tag in docs and will give a compile-time warning on the use of deprecated methods. Since our deprecations are in a separate file I imagine that including them in typings would be more work than it is worth, and should just be a future wishlist item.

bigtimebuddy:
### Loader class is missing type information for its once() method

We need a better TypeScript deprecation strategy moving forward. My current thinking is that I'd like to continue to hide the deprecation APIs from types. We can re-evaluate in v6 maybe by having deprecations that are more localized in packages. So people _should_ get type errors using `once`, but runtime it should still work.

### error TS2304: Cannot find name 'Signal'

I'm cool with that approach.

### esModuleInterop

I agree that no compilerOption is best. But I'm also, I'm okay with this if it allows us to move forward. I think we should document this somewhere (readme?), since I assume devs will have to add this compiler option if they are bundling.

bigtimebuddy:
To avoid `allowSyntheticDefaultImports` is it possible to `declare module` in the global.d.ts for those external dependencies? Would that silence those build errors?

andrewstart:
I assume that ResourceLoader is trying to overwrite the mini-signal types because it wants to use better, more accurate versions. I haven't looked into how we are pulling it in, but is it possible to not pull in the real mini-signal types and just expose ResourceLoader's version to consumers of PixiJS? Potentially adding a triple-slash reference somewhere if needed.

Zyie:
> To avoid `allowSyntheticDefaultImports` is it possible to `declare module` in the global.d.ts for those external dependencies? Would that silence those build errors?

this seems to work for some modules but causes errors for others like `eventemitter3`.

> I assume that ResourceLoader is trying to overwrite the mini-signal types because it wants to use better, more accurate versions. I haven't looked into how we are pulling it in, but is it possible to not pull in the real mini-signal types and just expose ResourceLoader's version to consumers of PixiJS? Potentially adding a triple-slash reference somewhere if needed.

Yes `resource-loader` made the types better by making the signals generic. I've tried adding triple-slash references but it doesn't seem to work.

bigtimebuddy (approved):

bigtimebuddy:
@andrewstart could you review this again?

andrewstart (approved):

lloydevans:
Hi there, I'm installing Pixi v6.0.1 and finding that eventemitter3 types won't work unless I enable `allowSyntheticDefaultImports` in my project tsconfig. In order to avoid the need for this, I found modifying eventemitter3 import to `import * as EventEmitter from 'eventemitter3';` worked. Is there a reason it isn't imported like this?

bigtimebuddy:
@lloydevans if you want to take a crack at it, by all means! While that resolves the types, it still creates issues at runtime  (unit-tests) and not sure why. It's curious.

lloydevans:
@bigtimebuddy Thanks, oh yea I see the star import indeed only works for types. It seems to be an issue with rollup/typescript and the way the eventemitter3 types are exported. I've tried a few things to fix it but no luck. Perhaps it will not be fixable unless the eventemitter3 module is modified/updated.

PR #7050, opened 2020-12-05
[Fix]: Filter system should work in world space and then restore projection transform.

SukantPal:
See https://github.com/pixijs/pixi.js/pull/7044#issuecomment-739322922

Basically, instead of making the filtering pipeline work in projected space (if `projectionSystem.transform` is applied) - we should make it work in world space. This would simplify all our calculations b/c we don't have to worry about the projection transform. The projection transform is restored when rendering to the final output texture in `bindAndClear`.

Another benefit is that applying filtering in projected space can increase the AABB (think rotation), which is unwanted.

Screen Shot 2020-12-05 at 12 26 18 PM

## Other changes

I and @ivanpopelyshev had a call and we coded together and did the following fixes:

* restore the sourceFrame/destinationFrame binding of the render texture system
* apply the rounding of the filter system's source frame in screen space (by projecting it there, rounding, and projecting it back). We skip rounding if the projection has skew involved, since that invalidates the "rectangleness".
* added the fix in #6892, #6893 (the "fix regression in y-flipping PR") since it wasn't already in the `main` branch but in `development`. (It was creating a problem in our testing so we had to "pre-merge" it)


SukantPal:
@bigtimebuddy's fiddle (modified by adding masking, removing `filterArea`) with #7044 vs this (#7050):

* https://jsfiddle.net/ShukantPal/v3dtcohp/
* https://jsfiddle.net/ShukantPal/rLpnd01q/

dev7355608:
1. Since source frame dimension determines the render textures dimension, the quality will be horrible if apply a zoom-in or skew transform. That's currently not taken into account. We would need to do something like this to get appropriate dimensions for the render texture.
```js
const lt = tempPoints[0];
const lb = tempPoints[1];
const rb = tempPoints[2];

lt.set(state.sourceFrame.left, state.sourceFrame.top);
lb.set(state.sourceFrame.left, state.sourceFrame.bottom);
rb.set(state.sourceFrame.right, state.sourceFrame.bottom);

renderer.projection.transform.apply(lt, lt);
renderer.projection.transform.apply(lb, lb);
renderer.projection.transform.apply(rb, rb);

const width = Math.hypot(lt.x - lb.x, lt.y - lb.y); // width for renderTexture
const height = Math.hypot(rb.x - lb.x, rb.y - lb.y); // height for renderTexture
```

2. The filter area, if set, is transformed, but it shouldn't be.

3. I feel like `this.renderer.projection.transform = null;` should be added at the end of `pop()` for correctness. Doesn't seem to matter at the moment since `state.transform` is the identity transform most of the time. With this new approach every transform could be done in shaders, i.e., `state.transform = target.localTransform` with local bounds etc.

Demo: https://jsfiddle.net/7qg2e15x/

SukantPal:
@dev7355608 I agree with 1; and for 2, that is intentional - the filter area is in world space. The documentation doesn't state filterArea is in screen-space: http://pixijs.download/dev/docs/PIXI.DisplayObject.html#filterArea.

Regarding 3, the `renderer.projection.transform` should not change from before `push` to after `pop`. You can't set it to null at the end of `pop()`.

SukantPal:
@dev7355608 To get the render-texture dimensions, we can just multiply the scaleX & scaleY values instead of transforming the whole source frame.

dev7355608:
@SukantPal 

Regarding 2: I feel like 
```js
// object.transform.setFromMatrix(Matrix.IDENTITY);
app.renderer.render(object, texture, true, transform);
```
and
```js
object.transform.setFromMatrix(transform);
app.renderer.render(object, texture, true);
```
should be equivalent. That why I think the filter area must not be transformed.

Regarding 3: I'm thinking of the follow scenario. I hope it makes sense:
```
push()                projection.transform == matrix1
    child1 _render()  projection.transform == matrix1

    child2 push()     projection.transform == matrix2
    child2 _render()
    child2 pop()      projection.transform not reset to matrix1 here

    child3 _render()  projection.transform != matrix1
pop()
```

SukantPal:
@dev7355608 

I understand your reasoning for 2 now. But I still feel `filterArea` should be transformed because the two cases you show do have a fundamental difference. Think about it like this:

First case: You are moving a sheet of paper (i.e. transforming the world).

Second case: You are moving the contents of the paper (i.e. transforming the stage).

Since the `filterArea` is defined in the sheet of paper, it must be transformed in the first case; but not in the second.

---

For 3, your analogy doesn't make sense, because `FilterSystem.push` doesn't assign a _new_ projection matrix. It would look more like this:

```
push()                    projection.transform === matrix1
    child1_render()  projection.transform === null (rendering from world directly into render-texture, without extra projection)
    
    child2_push()    projection.transform === null (as before)
    child2_render()
    child2_pop()     projection.transform ===  Matrix.IDENTITY (nothing to restore)

    child3_render()  projection.transform === Matrix.IDENTITY
pop()                    projection.transform will be set to matrix1 (restored in bindAndClear())
```

If there is a chance that `bindAndClear` is never called on the output render-texture, we should also restore the `projection.transform` in the end of `pop()`. I think that's what you mean (not set it to `null`).

SukantPal:
> @dev7355608 To get the render-texture dimensions, we can just multiply the scaleX & scaleY values instead of transforming the whole source frame.
> 
> ```js
> const sx = Math.sqrt(transform.a**2 + transform.b**2);
> const sy = Math.sqrt(transform.c**2 + transform.d**2);
> 
> sourceFrame.width = sourceFrame.width * sx;
> sourceFrame.height = sourceFrame.height * sy;
> ```

@dev7355608 I think I made a mistake here. You wouldn't resize `sourceFrame` - rather you would resize the `destinationFrame`, since that is the render-texture's dimensions.

ivanpopelyshev:
filterArea shouldnt be transformed because I put 

```js
stage.filterArea = app.screen;
```

in all of my posts/articles.

dev7355608:
@SukantPal 

Maybe we should add a `filterAreaSpace` field with options `SCREEN`, `WORLD`, and `LOCAL` to make everyone happy. But because of what Ivan mentioned, I think the default should be `SCREEN`.

---

Yes, what I meant was to restore the transform at the end of pop().

ivanpopelyshev:
This is a fix PR, we can add breaking/new_features later. Default should be `SCREEN`.

Another issue is that `ceil()` has to be done in screen pixel coors and i dont see in this PR in which coords ceil() is.

SukantPal:
@ivanpopelyshev In all your posts, did you apply a `transform`?

SukantPal:
@ivanpopelyshev If filterArea is in screen space, you would need to also apply the render-texture's inverse destinationFrame → sourceFrame projection. Since we never did that, it doesn't make sense to do so for `projection.transform` either.

SukantPal:
@ivanpopelyshev We can fix the pixel rounding in screen space in another PR. That's out of scope for here. This PR does it in world space, just like before.

dev7355608:
I found this in the docs. In both cases it refers to screen space.

https://github.com/pixijs/pixi.js/blob/4619c56abb2a5887bff797d92b72d81c36a6d8ed/packages/app/src/Application.ts#L131-L136

https://github.com/pixijs/pixi.js/blob/50289496e8c85166048fe2b5c23c579620a80fe7/packages/filters/filter-alpha/src/AlphaFilter.ts#L15-L21

SukantPal:
@dev7355608 So, 1 doesn't refer to "screen space". It just says the `screen` object is "safe" to there are no side effects of using it as `filterArea`.

In most cases, screen & world coordinates are same. We need to fix that in the docs in case 2 by putting `world` coordinates instead.

SukantPal:
I've seen most use cases of filterArea to be either:

* filtering a portion of a display object (like blurring a part of a texture). Here you would want the filter area to move with the display object, and it makes sense the filter area to be projected.

* applying a filter on the whole screen. Here it would make sense to inverse project the filterArea.

dev7355608:
@SukantPal 
> To use clipping in display coordinates, assign a filterArea to the same container that has this filter. 

It definitely sounds more like screen space than world space (to me at least). Whether that was intended or a mistake is another matter.

dev7355608:
The question whether `filterArea` is screen space or world space might actually be the question whether filters should work in screen or world space.

- Just stage rotation: https://jsfiddle.net/kve5f381/
- Rotation via transform parameter (w/ this PR): https://jsfiddle.net/vw8u4x92/2/

SukantPal:
@dev7355608 Filters apply in world space. That is the intended behavior. You are under the assumption of the following:

> Regarding 2: I feel like
> 
> ```js
> // object.transform.setFromMatrix(Matrix.IDENTITY);
> app.renderer.render(object, texture, true, transform);
> ```
> 
> and
> 
> ```js
> object.transform.setFromMatrix(transform);
> app.renderer.render(object, texture, true);
> ```

That is wrong. Applying the projection is like rendering into a texture, and then rendering that texture onto the screen by applying the projection. If you do that, it will have the same effect as  "rotation w transform parameter" .

Think of it like this - the world is a sheet of paper and its contents are the `stage`. The first case in your example is moving the contents w.r.t. the paper. The second case is moving the sheet of paper itself.

Now, applying a pixelation filter is like downsampling the contents to the grid of the paper. Moving the contents (rotation via worldTransform) will not rotate the pixels (i.e. paper's grid). However, moving the page itself (rotation via projection transform) will rotate the pixels.

ivanpopelyshev:
It was undefened behaviour, and I assumed that because of older cases like `stage.filterArea = app.screen`, we have to make it screen-space.

But now that im thinking about it, we can make a fine addition to the camera.

Imagine you have a camera that manages that `transform`, and also moves `camera.screen` rectangle around. Then you can write `stage.filterArea = camera.screen` and it will work with transforms! Before ,`app` managed resizes and it worked automatically - here it will also work automatically, no need to re-assign anything, because its not copying setter, its just a link.

So I propose that we define `filterArea` in world space for the sake of consistency.

ivanpopelyshev:
And we just solved rounding issue , we even checking it for skew. This PR looks more defined than mine, hope @SukantPal makes new PDF about projections soon :)

dev7355608:
Ok, if filters are supposed to work with world space, then I agree as well.

ivanpopelyshev (approved):

dev7355608:
The resolution problem (zoom & skew) hasn't been addressed yet.

image

SukantPal:
What do you mean by skew? And for the zoom thing, we can increase the size of the source texture in another PR. You’d have to add checks for maximum texture sizes supported by the device too.

On Dec 12, 2020, at 11:19 AM, dev7355608 <notifications@github.com<mailto:notifications@github.com>> wrote:



The resolution problem (zoom & skew) hasn't been addressed yet.

[image]<https://user-images.githubusercontent.com/31905376/101989060-1b923c80-3c9e-11eb-9ae5-7edf591c07db.png>

dev7355608:
Never mind, no problems with skew transforms.

GoodBoyDigital:
hey chaps, would be good to understand the performance impact of this, does the extra maths have an impact at all?

SukantPal:
@GoodBoyDigital Filters are already super-heavy so any CPU-side calculation of O(1) will have negligible affect on rendering efficiency. This PR does not introduce any memory allocations either. Which specific math are you concerned about?

ivanpopelyshev:
@GoodBoyDigital nope.

GoodBoyDigital:
cool, a lot of new code to wrap my head around haha! :D 
Would be good to understand the value this will bring to pixi users. (not suggesting it doesn't, but its not entirely clear to me after reading this PR, blame it on my Sunday brain :P )

SukantPal:
@GoodBoyDigital I think you didn't read the previous threads that have more context.

* https://github.com/pixijs/pixi.js/issues/7035
* https://github.com/pixijs/pixi.js/pull/7039
* https://github.com/pixijs/pixi.js/pull/7044

Basically, the filter system didn't work with a projection transform or a custom source/destination frame when binding the render texture.

An additional fix was fixing how the sourceFrame was "pixel-rounded". It has to be rounded in screen space, not world space (i.e. you have to use the source->destination frame projection + projection transform)

GoodBoyDigital:
thanks will give them a read!

packages/core/src/filters/FilterSystem.ts
```diff
@@ -170,17 +173,28 @@ export class FilterSystem extends System
...
 
         state.sourceFrame.pad(padding);
+
         if (autoFit)
         {
-            state.sourceFrame.fit(this.renderer.renderTexture.sourceFrame);
+            const sourceFrameProjected = this.renderer.renderTexture.sourceFrame.clone();
+
+            // Project source frame into world space (if projection is applied)
+            this.transformAABB(renderer.projection?.transform?.clone().invert() || Matrix.IDENTITY, sourceFrameProjected);
```
GoodBoyDigital, line 185: creating a new matrix each time here! could store the invert?

packages/core/src/filters/FilterSystem.ts
```diff
@@ -470,4 +507,76 @@ export class FilterSystem extends System
...
+    {
+        if (transform)
+        {
+            const { a, b, c, d } = transform;
+
+            // Skip if skew/rotation present in matrix, except for multiple of 90° rotation. If rotation
+            // is a multiple of 90°, then either pair of (b,c) or (a,d) will be (0,0).
+            if ((b !== 0 || c !== 0) && (a !== 0 || d !== 0))
+            {
+                return;
```
GoodBoyDigital, line 559: nice 👍

packages/core/src/filters/FilterSystem.ts
```diff
@@ -470,4 +507,76 @@ export class FilterSystem extends System
...
+            }
+        }
+
+        transform = transform ? transform.clone() : new Matrix();
```
GoodBoyDigital, line 563: should we use a temp matrix here too?

packages/core/src/filters/FilterSystem.ts
```diff
@@ -170,17 +173,28 @@ export class FilterSystem extends System
...
+
+            state.sourceFrame.fit(sourceFrameProjected);
         }
 
-        // round to whole number based on resolution
-        state.sourceFrame.ceil(resolution);
+        // Round sourceFrame in screen space based on render-texture.
+        this.roundFrame(
+            state.sourceFrame,
+            renderer.renderTexture.current ? renderer.renderTexture.current.resolution : renderer.resolution,
```
GoodBoyDigital, line 193: `renderer.renderTexture` is used a lot! could create a local var?

GoodBoyDigital:
yeah this is cool! nice work chaps. left a couple of points!

packages/core/src/filters/FilterSystem.ts
ivanpopelyshev: that's a system,. and i really dont want to name `renderTexture` there, so it wont save much

bigtimebuddy: So call the local var `renderTextureSystem` doesn't matter if it's verbose, will get minified to a single character anyways

packages/core/src/filters/FilterSystem.ts
ivanpopelyshev: yes we should do it

packages/core/src/filters/FilterSystem.ts
SukantPal: Done 👍

packages/core/src/filters/FilterSystem.ts
SukantPal: Done 👍

packages/core/src/filters/FilterSystem.ts
SukantPal: Done 👍

GoodBoyDigital (approved):
👍

bigtimebuddy:
@SukantPal can you rebase to dev please?

PR #7051, opened 2020-12-05
[Fix]: x, y matrix skew comment

### LINKED ISSUE #7019: Math.Matrix.append multiplies matrices instead of appending
I am reading the source code and I ran into [Math.Matrix.append()](https://github.com/pixijs/pixi.js/blob/fc4ca303a8f03eee754e12e0e769c263d8b19465/packages/math/src/Matrix.ts#L270) function which (I think) suppose to add one matrix to another.
In the source code, I can see that it does not append but multiply those matrices. Is there any reason for it?

```
   /**
     * Appends the given Matrix to this Matrix.
     *
     * @param {PIXI.Matrix} matrix - The matrix to append.
     * @return {PIXI.Matrix} This matrix. Good for chaining method calls.
     */
    append(matrix: Matrix): this
    {
        const a1 = this.a;
        const b1 = this.b;
        const c1 = this.c;
        const d1 = this.d;

        this.a = (matrix.a * a1) + (matrix.b * c1);
        this.b = (matrix.a * b1) + (matrix.b * d1);
        this.c = (matrix.c * a1) + (matrix.d * c1);
        this.d = (matrix.c * b1) + (matrix.d * d1);

        this.tx = (matrix.tx * a1) + (matrix.ty * c1) + this.tx;
        this.ty = (matrix.tx * b1) + (matrix.ty * d1) + this.ty;

        return this;
    }

```

SukantPal:
Fixes #7019 (the skew comment)

##### Description of change
Swap the x/y skew comments. See https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix()


bigtimebuddy (approved):

themoonrat (approved):

bigtimebuddy:
Thanks @SukantPal

PR #7062, opened 2020-12-12
[Fix]: Make renderer use alpha in main framebuffer by default

### LINKED ISSUE #6853: Do not default to alpha = false when creating WebGL context
I was trying to find a specific bottleneck in my PixiJS application and I found out that creating the context on my own made a HUGE difference (25-30 ➡️ 60 FPS on a 4K monitor). The difference was that I enabled alpha `alpha: true` when creating the WebGL context.

PixiJS defaults to `alpha: false` if the `transparent` option is not passed:

https://github.com/pixijs/pixi.js/blob/54665ecebe676bd0d89e7ceadabb1badd89c1cbb/packages/core/src/Renderer.ts#L280

### Evidence of this being a performance bottleneck

* https://bugs.chromium.org/p/chromium/issues/detail?id=1045643
* https://github.com/mrdoob/three.js/issues/18338#issuecomment-572531584
* https://bugs.chromium.org/p/chromium/issues/detail?id=1103485

Maybe we want to reevaluate whether `transparent = false` by default?

SukantPal:
Fixes #6853 

##### Description of change

The default value of `transparent` in the renderer options is true now. To prevent canvas from actually looking transparent, the default alpha value in `AbstractRenderer#_backgroundColorRgba` is 1. For the majority of uses, this change will not have any visible affect (but may improve performance b/c non-alpha mode emulation in software is expensive).

##### Pre-Merge Checklist


SukantPal:
Pen: https://codepen.io/sukantpal/pen/LYRxGvW

bigtimebuddy (approved):
Can you double check the Application, Renderer/CanvasRenderer and autoDetectRenderer docs update for transparent?

andrewstart:
Do we also need to change documentation to make it clear how to get a
transparent canvas if you need one?

SukantPal:
* @bigtimebuddy I updated the `transparent` param docs in `Renderer`, `AbstractRenderer`, `autoDetectRenderer`.

* @andrewstart I added a `backgroundAlpha` field to `AbstractRenderer`, so users can make the canvas transparent by setting it to 0 (added this to its doc).

andrewstart (approved):
Thanks, I end up overlaying my canvases over video elements a lot.

ivanpopelyshev (approved):
If we do that, I can finally release stable version of pixi-picture or even integrate it to main pixi repo

Issue #6775, opened 2020-07-25
5.3.1 build with JSDoc error ^ Node 12


crazykay:
## JSDoc 3.5.5 Breaking Builds on Node 12
https://github.com/pixijs/pixi.js/issues/6571

### Possible Solution
upgrade
`"jsdoc": "^3.5.5"`
to
`"jsdoc": "^3.6.4",`

### Environment

- **`pixi.js` version**: 5.3.1
- **Browser & Version**: none
- **OS & Version**: Archlinux
- **Running Example**:  none

bigtimebuddy:
This is a duplicate. We already have a plan to fix for 3.4.0, by moving away from JSDoc entirely. Closing.

Issue #5724, opened 2019-05-24
Can't build docs on Node 12


andresroberto:
### Expected Behavior

After running `npm run docs` I should get the docs built into the `dist/docs` folder.

### Current Behavior

When running the command on Node 12, the current result of running `npm run docs` is:

<img width="1117" alt="Screenshot 2019-05-24 at 17 14 44" src="https://user-images.githubusercontent.com/1273138/58338114-73ace800-7e47-11e9-8265-460c56b35abb.png">

### Possible Solution

In order to solve this issue jsdoc needs to be updated to version 3.6 which supports Node 12. However, jsdoc 3.6 has some breaking changes that cause the docs to be generated in a erroneous way. More research is needed in order to find a way to upgrade and fix the docs afterwards.

### Steps to Reproduce

1. Set node 12 as current node version (for instance install and set it as current node version with nvm).
2. Run the `npm run docs` command.
3. See the error message.
4. Check the `dist/docs` folder does not exist.
5. Running `npm test` also generate the same error message but won't prevent the tests from running. The error shows up because the test script also generates typescript types based on the jsdoc output, therefore running jsdoc is part of the process.

### Environment

- **`pixi.js` version**: 5.0.3
- **Browser & Version**: No Browser but Node 12.2.0
- **OS & Version**: macOS 10.14.5
- **Running Example**: NA

ivanpopelyshev:
Yep, its a known issue of requizzle, we are waiting for fix. Can you check it in their repo if it was solved?

andresroberto:
Yes, they made a fix (https://github.com/hegemonic/requizzle/pull/4) and released it and JSDoc `3.6.0` includes the fix. Unfortunately, upgrading JSDoc breaks the generated documentation as mentioned by @bigtimebuddy (https://github.com/pixijs/pixi.js/pull/5722#issuecomment-495632329).

PR #7065, opened 2020-12-14
Add strict engines support

bigtimebuddy:
This is a non-critical change to define `engines` in the root package.json. I added similar changes for v5.3.x and v5.2.x branches (those old branches explicitly require Node 10 to build docs/types) and thought it would be good to declare supported Node.js versions and npm versions more explicitly. This aligns with the GitHub Actions for testing. In the past, users trying to build have hit issues #6775, #6571, #5724, etc

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict = true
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):
nice

Zyie (approved):

PR #7067, opened 2020-12-14
Deprecates transparent Option

### LINKED ISSUE #7066: Transparent Not Working As Expected
Because of this change #7062 `transparent: true` doesn't do what it's suppose to:
https://pixijs.io/examples/#/demos-basic/transparent-background.js

We need to harmonize:

* Renderer's `backgroundAlpha`
* Application/Renderer's `options.transparent`
* Canvas' context `alpha` property

bigtimebuddy:
Fixes #7066

This clears up the ambiguity of the recent transparent changes by creating two distinct options for Application and Renderer. It's easier to backward support `transparent` by not overloading it with new meaning.

* **`contextAlpha`** the readonly property for creating the `canvas.getContext`
* **`backgroundAlpha`** the amount of background alpha from 0 to 1

The option `transparent` is marked as deprecated but still supports backward compatibility by setting `contextAlpha` and `backgroundAlpha`.

**Broken**
https://pixijs.io/examples/#/demos-basic/transparent-background.js

**Fixed**
https://pixijs.io/examples/?v=dev-context-alpha#/demos-basic/transparent-background.js

### DIFF
--- a/packages/app/src/Application.ts
+++ b/packages/app/src/Application.ts
@@ -50,7 +50,7 @@ export class Application
      * @param {number} [options.width=800] - The width of the renderers view.
      * @param {number} [options.height=600] - The height of the renderers view.
      * @param {HTMLCanvasElement} [options.view] - The canvas to use as a view, optional.
-     * @param {boolean} [options.transparent=false] - If the render view is transparent.
+     * @param {boolean} [options.contextAlpha=true] - Pass-through value for canvas' context `alpha` property.
      * @param {boolean} [options.autoDensity=false] - Resizes renderer view in CSS pixels to allow for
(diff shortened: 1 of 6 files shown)


andrewstart (approved):

SukantPal (approved):
I find adding a context prefix to "alpha" distasteful, especially since other context fields aren't. But I'll be a stoic. 👍

bigtimebuddy:
@SukantPal backgroundAlpha does not work with canvas. Did you test that?

Also, I hear you on the context prefix. On the plus tho, it's unlikely a prop that devs will use much since it has some subtle pref side effects in some environments.

PR #7068, opened 2020-12-15
FAIL_IF_MAJOR_PERFORMANCE_CAVEAT switched to false by default

themoonrat:
##### Description of change
After the discussions in https://github.com/pixijs/pixi.js/issues/7054 - I feel it's a better strategy by default to allow WebGL where possible, and let developers decide if they want to fail on this flag. It seems it's not always an accurate indicator on whether a device will have good performance or not.

##### Pre-Merge Checklist

How docs looks after generation

image


bigtimebuddy (approved):
Nice!

SukantPal (approved):

PR #7073, opened 2020-12-16
Fix: Clamp blur filter to inputClamp

### LINKED ISSUE #7063: Blur Filter bug in 5.4.0-rc.3
Noticing an odd issue with 5.4.0-rc.3. The blur filter appears to be bleeding (this did not happen in 5.3.3 running the same code). 

Here are my blur settings:
blur = new PIXI.filters.BlurFilter();
blur.blur = 0.75;
blur.resolution = 2;
blur.quality = 8;

Included below are two screenshots that demonstrate the issue (top is 5.3.3, bottom is 5.4.0-rc3):

image

image

...note the bits of bleed that show up sporadically. I'm testing on MacOS 10.13.6 w the latest Firefox, Chrome, Safari, and Edge browsers. I'm rendering in WebGL (have not tested canvas fallback as yet - but critical for this to work in WebGL).

SukantPal:
Fixes #7063 

This wasn't a regression in that the offending PR was incorrect. The problem rather lied in the `BlurFilter` incorrectly assuming the filter texture outside of the filtering area on the texture was cleared.

### The Problem

In #7063, @Article19 correctly suggested that the filter texture was bleeding. The `BlurFilter` was sampling texels outside of the filtering area (even outside the padding). Since the `FilterSystem` no longer has to clear the whole texture, those texels could have garbage from previous filtering operations.

### Description of change

The blur filter's texel sampling is clamped to the `inputClamp`, i.e. the padded filtering area.

### Explanation

Technically, the blur filter shouldn't require clamping - we already apply a padding around the filter target. However, the blur filter is also applied inside the padding - which is due the limitation of the `FilterSystem`. Ideally, the filter should run only in the filter target's area.

Blur Filter Bleeding

The above diagram shows different texels the filter calculates the output for. In each case, the filter samples all the texels in the input texture within the radius around the output texel. For any location within the yellow region (filter target's bounds), the texels have the correct values within the blurring radius (filter target inside its bounds, transparent black in filter padding). HOWEVER, for any location in the filter padding itself, the blur filter would sample pixels outside of the filter source frame. Since the blue region could contain garbage, using it in the blur manifests itself as "bleeding".

## Verification of fix

I tested that clamping the filter sampling to `inputClamp` by clearing the area outside of the filtering area with opaque red. Without this fix, you would see a red bleeding:

Screen Shot 2020-12-16 at 9 59 32 AM

And indeed, this fix works - after:

Screen Shot 2020-12-16 at 10 14 47 AM

## Reproduction notes

To reproduce this bug, you must ensure that multiple objects are filtered using the same filter textures. That means the next-power-of-2 filter area (bounds + filter padding) must be equal.

##### Pre-Merge Checklist

### DIFF
--- a/packages/filters/filter-blur/src/generateBlurFragSource.ts
+++ b/packages/filters/filter-blur/src/generateBlurFragSource.ts
@@ -13,6 +13,7 @@ const GAUSSIAN_VALUES: IGAUSSIAN_VALUES = {
 const fragTemplate = [
     'varying vec2 vBlurTexCoords[%size%];',
     'uniform sampler2D uSampler;',
+    'uniform vec4 inputClamp;',
 
     'void main(void)',
     '{',
@@ -22,6 +23,11 @@ const fragTemplate = [
 
 ].join('\n');
 
+const sampleTexelTemplate = `gl_FragColor += texture2D(
+    uSampler, 
+    clamp(vBlurTexCoords[%index%], inputClamp.xy, inputClamp.zw)
+) * %value%;`;
+
 export function generateBlurFragSource(kernelSize: number): string
 {
     const kernel = GAUSSIAN_VALUES[kernelSize];
@@ -30,12 +36,11 @@ export function generateBlurFragSource(kernelSize: number): string
     let fragSource = fragTemplate;
 
     let blurLoop = '';
-    const template = 'gl_FragColor += texture2D(uSampler, vBlurTexCoords[%index%]) * %value%;';
     let value: number;
 
     for (let i = 0; i < kernelSize; i++)
     {
-        let blur = template.replace('%index%', i.toString());
+        let blur = sampleTexelTemplate.replace('%index%', i.toString());
 
         value = i;
 


SukantPal:
@bigtimebuddy  Fixed fiddle: https://jsfiddle.net/ShukantPal/h6a0j2be/

ivanpopelyshev (changes requested):
That with what we did with `CLEAR_MODE`

The idea is that if `clearMode = ON` then whoever made filter doesnt want to spend time adjusting clamp. 

Of course after your fix, BlurFilter can use partial clearing,  `clearMode = AUTO`.

But for many other filters - that's not the case, you have to clear everything anyway.

Then it comes decision what exactly FilterSystem has to do with that `AUTO` mode. 
That's `FilterSystem.forceClear`
For now there were two cases:
1. we just clear everything.
2. we actually draw quad over with `BLEND_MODES.NONE`

You added third option
3. we clear it partially, then draw with regular blendMode.

ivanpopelyshev:
The fix for blurfilter is fine, but it wont help other filters

bigtimebuddy:
Great explanation, but I’m with @ivanpopelyshev on this one this seems like a specific fix to a general problem. @SukantPal you made #6892 which introduced this behavior. I’m more inclined to undo that PR than start mucking with BlurFilters shaders.

SukantPal:
I'll make a fix for the filter system today evening.

PR #7075, opened 2020-12-17
Improve CLEAR_MODES docs

ivanpopelyshev:
As discussed with @SukantPal , I have to clarify this once more.

Most people that write filters dont know about clamping and other tricks that are needed in case we use pow2 temporary renderTextures. Even our `BlurFilter` is affected, @GoodBoyDigital wrote this version for v5 :)

For them, we distinguish two clearing modes - one `CLEAR` that clears everything, and second `BLIT` that just overwrites new content over texture, and leaves right and bottom zone as it was.

### DIFF
--- a/packages/core/src/filters/FilterSystem.ts
+++ b/packages/core/src/filters/FilterSystem.ts
@@ -4,7 +4,7 @@ import { Quad } from '../utils/Quad';
 import { QuadUv } from '../utils/QuadUv';
 import { Rectangle, Matrix } from '@pixi/math';
 import { UniformGroup } from '../shader/UniformGroup';
-import { DRAW_MODES, CLEAR_MODES, BLEND_MODES } from '@pixi/constants';
+import { DRAW_MODES, CLEAR_MODES } from '@pixi/constants';
 import { FilterState } from './FilterState';
 
 import type { Filter } from './Filter';
@@ -344,7 +344,7 @@ export class FilterSystem extends System
             renderTextureSystem.bind(filterTexture);
         }
 
-        const autoClear = stateSystem.blendMode !== BLEND_MODES.NONE || this.forceClear;
+        const autoClear = (stateSystem.stateId & 1) || this.forceClear;
 
         if (clearMode === CLEAR_MODES.CLEAR
             || (clearMode === CLEAR_MODES.BLIT && autoClear))
@@ -365,6 +365,7 @@ export class FilterSystem extends System
     {
         const renderer = this.renderer;
 
+        renderer.state.set(filter.state);
         this.bindAndClear(output, clearMode);
 
         // set the uniforms..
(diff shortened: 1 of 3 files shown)


SukantPal:
@ivanpopelyshev BLIT mode doesn't not clear only the filtering area on the texture right now. If that was the intended purpose, we need to fix it.

bigtimebuddy:
I find this doc change not particularly useful or legible. This need some editing for clarity.

SukantPal (approved):
Hey @bigtimebuddy, I had a chat with @ivanpopelyshev and we cleared up (get it?) what `CLEAR_MODES.BLIT` should be doing. It basically skips clearing the filter texture if the filter has `state.blend = false` - as that will overwrite pixels anyway. Setting `forceClear` on the `FilterSystem` will make this mode clear regardless of the blend mode. That flag may be used in the future to force clearing on devices where that operation makes things faster.

SukantPal (approved):
Hey @bigtimebuddy, I had a chat with @ivanpopelyshev and we cleared up (get it?) what `CLEAR_MODES.BLIT` should be doing. It basically skips clearing the filter texture if the filter uses `BLEND_MODES.NONE` - as that will overwrite pixels anyway. Setting `forceClear` on the `FilterSystem` will make this mode clear regardless of the blend mode. That flag may be used in the future to force clearing on devices where that operation makes things faster.

packages/core/src/filters/FilterSystem.ts
```diff
@@ -319,27 +319,35 @@ export class FilterSystem extends System
...
             destinationFrame.height = filterTexture.filterFrame.height;
 
-            this.renderer.renderTexture.bind(filterTexture, filterTexture.filterFrame, destinationFrame);
+            renderTextureSystem.bind(filterTexture, filterTexture.filterFrame, destinationFrame);
         }
         else
         {
-            this.renderer.renderTexture.bind(filterTexture);
+            renderTextureSystem.bind(filterTexture);
         }
```
GoodBoyDigital, line 346: heya! would be great to comment this line, as its hard to tell whats going on 🥇

bigtimebuddy: This line:
```js
const autoClear = (stateSystem.stateId & 1) || this.forceClear;
```

GoodBoyDigital (approved):
lgtm 👍

packages/core/src/filters/FilterSystem.ts
SukantPal: I'll add a comment when the workday is over (be patient, don't merge this until I do pls even tho it has 2 approvals)

bigtimebuddy: Roger that!

bigtimebuddy (approved):

bigtimebuddy:
Thank you @SukantPal and @ivanpopelyshev for making this better.

PR #7084, opened 2020-12-20
GraphicsGeometry JSDoc function signature

jasonsturges:
JSDoc comment doesn't match function signature: `aloow32Indices` changed to: `allow32Indices`

Also updated "_preventings_" to "_preventing_" in the comment.  

Might want to assure you actually mean "_artefacts_" instead of "_artifacts_".


SukantPal (approved):
Yes, artefacts should be artifacts.

bigtimebuddy (approved):

PR #7093, opened 2020-12-21
Fix: Geometry with different shaders leaked buffers

ivanpopelyshev:
Added special test for it.

Exists for all v5 branches.

Recommend to backport to 5.3.x

Bug was eating memory in production application on thousands of users. Now we can move to hundred thousands :)


ivanpopelyshev:
Minimal fix, because any rearrangement of code will probably trigger someone OCD, there are TODO's comments in those methods, so better to refactor it later in bulk.

bigtimebuddy (approved):

GoodBoyDigital (approved):
cool 👍

bigtimebuddy:
Thanks @ivanpopelyshev

PR #7096, opened 2020-12-22
Add new BufferSystem to Renderer

GoodBoyDigital:
Hey peeps,

So was thinking about how we push pixi forwards using the new Uniform buffer objects stuff. This could be quite useful in reducing gpu calls further down the line!

UniformBuffers are basically handled _exactly_ like the buffers we currently have, but are obviously not related to geometry!

To that end I have separated the buffer logic from `GeometrySystem` and created a new `BufferSystem`. Aside from the fact it paves the way for UBO usage, feels like this is actually a nice separation of concerns!

Fortunately only GeometrySystem uses is for now, so a low risk refactor!

One question for perhaps you @ivanpopelyshev , the ref counting - this only makes sense in the geometry right? Or should that be moved to the buffer system too?


ivanpopelyshev:
With refcounting, buffer is disposed at the same time as the last geometry that has reference on it.

ivanpopelyshev:
When you add UBO's - you'll have to refcount buffers in them too.

GoodBoyDigital:
> With refcounting, buffer is disposed at the same time as the last geometry that has reference on it.

cool, so this works as is?

ivanpopelyshev:
I found a bug in it yesterday, #7093, and added test for it. Your code looks straightforward.

There's a reason we didnt implement RenderBufferSystem for example: it completely covered in FrameBufferSystem.

If you really have time to experiment right now then we have to merge this thing so you can start next PR

packages/core/src/geometry/BufferSystem.ts
```diff
@@ -0,0 +1,149 @@
...
+import type { Renderer } from '../Renderer';
+import type { IRenderingContext } from '../IRenderingContext';
+import type { Buffer } from './Buffer';
+
+/**
+ * System plugin to the renderer to manage buffers.
+ *
+ * @class
+ * @extends PIXI.System
+ * @memberof PIXI.systems
```
bigtimebuddy, line 13: We deprecated `systems` namespace.
```suggestion
 * @memberof PIXI
```

packages/core/src/geometry/BufferSystem.ts
```diff
@@ -0,0 +1,149 @@
+import { System } from '../System';
+import { GLBuffer } from './GLBuffer';
+
+import type { Renderer } from '../Renderer';
+import type { IRenderingContext } from '../IRenderingContext';
+import type { Buffer } from './Buffer';
+
+/**
+ * System plugin to the renderer to manage buffers.
```
bigtimebuddy, line 9: I would describe this description as a _basic what_, which basically it the same as the name of the class.  I think it's always better to describe _how_ it works, _when_ to use it, and _why_ it exists.

packages/core/src/geometry/GeometrySystem.ts
```diff
@@ -521,7 +459,7 @@ export class GeometrySystem extends System
         all = Object.keys(this.managedBuffers);
         for (let i = 0; i < all.length; i++)
         {
-            this.disposeBuffer(this.managedBuffers[all[i]], contextLost);
+            this._buffer.dispose(this.managedBuffers[all[i]], contextLost);
```
bigtimebuddy, line 462: Should this be `this._buffer.managedBuffers` instead of `this.managedBuffers`? These duplicate objects on Buffer and Geometry systems.

packages/core/src/geometry/GeometrySystem.ts
```diff
@@ -78,6 +80,11 @@ export class GeometrySystem extends System
          * @readonly
          */
         this.managedBuffers = {};
+
+        /**
+         * shortcut for the buffer system..
+         */
+        this._buffer = this.renderer.buffer;
```
ivanpopelyshev, line 87: Very bad name for variable, we have `buffer`'s everywhere here. Use bufferSystem, bufSystem or something like that.

packages/core/src/geometry/BufferSystem.ts
```diff
@@ -0,0 +1,149 @@
...
+     */
+    update(buffer:Buffer): void
+    {
+        const { gl, CONTEXT_UID } = this;
+
+        const glBuffer = buffer._glBuffers[CONTEXT_UID];
+
+        if (buffer._updateID !== glBuffer.updateID)
+        {
+            glBuffer.updateID = buffer._updateID;
```
ivanpopelyshev, line 78: If there's no "else" may be use same approach as in https://github.com/pixijs/pixi.js/blob/v5.1.x/packages/core/src/framebuffer/FramebufferSystem.js#L439 , just return out of function in if.

packages/core/src/geometry/GeometrySystem.ts
```diff
@@ -11,6 +11,7 @@ import type { Shader } from '../shader/Shader';
 import type { Program } from '../shader/Program';
 import type { Buffer } from './Buffer';
 import type { Dict } from '@pixi/utils';
+import { BufferSystem } from './BufferSystem';
```
ivanpopelyshev, line 14: that's `import type`

ivanpopelyshev (changes requested):

packages/core/src/geometry/GeometrySystem.ts
GoodBoyDigital: I think consistency is important here - all systems so far are short hand like this.  Out of scope for this one, but we could do a separate PR that converts all systems to be post fixed with System?

packages/core/src/geometry/GeometrySystem.ts
GoodBoyDigital: as managed buffers only really work with geometry, figured I would leave `managedBuffers` there?

GoodBoyDigital: oh.. actually I see what you mean now 👍

packages/core/src/geometry/BufferSystem.ts
```diff
@@ -0,0 +1,173 @@
...
+import type { IRenderingContext } from '../IRenderingContext';
+import type { Buffer } from './Buffer';
+
+/**
+ * System plugin to the renderer to manage buffers.
+ *
+ * WebGL uses Buffers as a way to store objects to the GPU.
+ * This system makes working with them a lot easier.
+ *
+ * Buffers are used in two main places in WebGL
```
SukantPal, line 14: You say they are used in two main places but list three 😆

packages/core/src/geometry/BufferSystem.ts
```diff
@@ -0,0 +1,173 @@
...
+        // TODO fill out...
+        this.CONTEXT_UID = this.renderer.CONTEXT_UID;
+    }
+
+    /**
+     * this binds specified buffer. OOn first run, it will create the webGL buffers for the context too
+     *
+     * @param {PIXI.Buffer} buffer the buffer to bind to the renderer
+     */
+    bind(buffer:Buffer):void
```
SukantPal, line 68: I'm noticing you don't leave any space between the identifier and type annotation. Let's give it some breathing room, just for consistency mate 🍺 (this isn't the only place)

```suggestion
    bind(buffer: Buffer): void
```

packages/core/src/geometry/Geometry.ts
```diff
@@ -199,7 +199,8 @@ export class Geometry
             buffer = new Buffer(buffer);
         }
 
-        buffer.index = true;
+        buffer.type = BUFFER_TYPE.ELEMENT_ARRAY_BUFFER;
```
SukantPal, line 202: Refactoring the `index` property to `type` is a breaking change. Since we're getting v6 out, you're lucky @GoodBoyDigital :P

packages/core/src/geometry/GeometrySystem.ts
SukantPal: I'm not going to block this on this account, but building on the consistency argument - caching a system is nowhere used. I don't think we should be caching dependency systems like this. I find it inelegant, but that may be me :)

SukantPal (changes requested):
I like the concept overall, just some minor tweaks and good to go.

packages/core/src/geometry/BufferSystem.ts
GoodBoyDigital: hmm.. feels like we could use lint auto fix for these ones! Does it fix these for you automatically? (just to understand if its my set up, or the lint rules?)

packages/core/src/geometry/Geometry.ts
GoodBoyDigital: i got your back bro, `index` still exists on `Buffer` for backwards compatibility 👍

packages/core/src/geometry/BufferSystem.ts
GoodBoyDigital: lint rule updated!

ivanpopelyshev (approved):

SukantPal:
I think #7107 should be merged into this branch before approving this.

package.json
```diff
@@ -11,6 +11,7 @@
     "pretest": "npm run lint && npm run build",
     "test": "floss --path test",
     "unit-test": "floss --path test",
+    "unit-test:debug": "floss --path test '--debug'",
```
bigtimebuddy, line 14: On second thought @mat could you remove this? I'm adding a PR to add some other testing bits.

GoodBoyDigital: nice! removed 👍

bigtimebuddy (approved):

bigtimebuddy:
@GoodBoyDigital can you resolve conflicts with your other shader optimization PR?

GoodBoyDigital:
> @GoodBoyDigital can you resolve conflicts with your other shader optimization PR?

sure thing! Done and done!

packages/core/src/geometry/BufferSystem.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * Binds a buffer whilst also binding its range.
+     * This will make the buffer start from the offset supplied rather than 0 when it is read.
+     *
+     * @param buffer - the buffer to bind
+     * @param offset - the offset to bind at (this is blocks of 256). 0 = 0, 1 = 256, 2 = 512 etc
+     * @param index - the base index to bind at, defaults to 0
+     */
+    bindBufferRange(buffer: Buffer, offset?: number, index?: number): void
```
SukantPal, line 114: I'm concerned about the order of the parameters `offset, index`. It would be more intuitive for `index` to come first to be consistent with `bindBufferBase`'s pattern.

Also, lets separate the doc brief (add a line between first & second sentence)

packages/core/src/geometry/Buffer.ts
```diff
@@ -107,6 +115,20 @@ export class Buffer
         this.data = null;
     }
 
+    /**
+     * setting index to true, will set the buffer type to be an ELEMENT_ARRAY_BUFFER
+     * false will set its type to ARRAY_BUFFER
+     * for backwards compatibility
```
SukantPal, line 121:
```suggestion
     * Flags whether this is an index buffer.
     *
     * Index buffers are of type `ELEMENT_ARRAY_BUFFER`. Note that setting this property to false will make
     * the buffer of type `ARRAY_BUFFER`. 
     * 
     * For backwards compatibility.
```

I suggest we also remove the "backwards compat" comment unless we actually plan to deprecate the setter in the future. Having a `index` property seems pretty valid.

packages/core/src/shader/ShaderSystem.ts
```diff
@@ -10,13 +10,14 @@ import type { Program } from './Program';
 import type { UniformGroup } from './UniformGroup';
 import type { Dict } from '@pixi/utils';
 import type { UniformsSyncCallback } from './utils';
+import { generateUniformBufferSync } from './utils/generateUniformBufferSync';
```
SukantPal, line 13: `import`s are placed above `import type`. ESLint rule for this would be dope.

packages/core/src/shader/GLProgram.ts
```diff
@@ -17,6 +17,19 @@ export class GLProgram
     public program: WebGLProgram;
     public uniformData: Dict<any>;
     public uniformGroups: Dict<any>;
+    /**
+     * a hash that stores where ubs are bound to on the program
```
SukantPal, line 21:
```suggestion
     * A hash that stores where UBOs are bound to on the program.
```

packages/core/src/shader/GLProgram.ts
```diff
@@ -17,6 +17,19 @@ export class GLProgram
...
+     */
+    public uniformBufferBindings: Dict<any>;
+    /**
+     * a hash for upload functions, we only generates new ones if they don't already exist
```
SukantPal, line 25:
```suggestion
     * A hash for lazily-generated uniform uploading functions.
```

packages/core/src/shader/GLProgram.ts
```diff
@@ -17,6 +17,19 @@ export class GLProgram
...
+     */
+    public uniformSync: Dict<any>;
+    /**
+     * a place where dirty ticks are stored for groups
```
SukantPal, line 29: Can you clarify what "ticks" mean?

```suggestion
     * A place where dirty ticks are stored for groups.
```

packages/core/src/shader/GLProgram.ts
```diff
@@ -54,6 +71,7 @@ export class GLProgram
     {
         this.uniformData = null;
         this.uniformGroups = null;
+        this.uniformBufferBindings = null;
```
SukantPal, line 74: Ignore this if the omission was intentional.

```suggestion
        this.uniformDirtyGroups = null;
        this.uniformBufferBindings = null;
```

packages/core/src/shader/UniformGroup.ts
```diff
@@ -1,36 +1,73 @@
...
+ * or a generic object that pixi will automatically map to a buffer for you.
+ * For maximum benefits, its best to make Ubo UniformGroups static, and only update them each frame.
+ *
+ * rules of UBO's:
+ * - ubo's only work with WebGL2, so make sure you have a fallback!
+ * - only floats are supported (including vec[2,3,4], mat[2,3,4])
+ * - samplers cannot be used i ubo's (a GPU limitation)
+ * - you must ensure that the object you pass in exactly matches in the shader ubo structure.
+ * Otherwise, weirdness will ensue!
+ * - the name of the ubo object added to the group must match exactly the name of the ubo in the shader.
```
SukantPal, line 27:
```suggestion
 * `UniformGroup` has two modes:
 *
 * 1: Normal mode
 * Normal mode will upload the uniforms with individual function calls as required
 *
 * 2: Uniform buffer mode
 * This mode will treat the uniforms as a uniform buffer. You can pass in either a buffer that you manually handle, or
 * or a generic object that PixiJS will automatically map to a buffer for you.
 * For maximum benefits, make Ubo UniformGroups static, and only update them each frame.
 *
 * Rules of UBOs:
 * - UBOs only work with WebGL2, so make sure you have a fallback!
 * - Only floats are supported (including vec[2,3,4], mat[2,3,4])
 * - Samplers cannot be used in ubo's (a GPU limitation)
 * - You must ensure that the object you pass in exactly matches in the shader ubo structure.
 * Otherwise, weirdness will ensue!
 * - The name of the ubo object added to the group must match exactly the name of the ubo in the shader.
```

packages/core/src/shader/UniformGroup.ts
```diff
@@ -61,20 +98,74 @@ export class UniformGroup
          * @member {boolean}
          */
         this.static = !!_static;
+
+        /**
+         * should this group be treated as a uniform buffer object
```
SukantPal, line 103:
```suggestion
         * Flags whether this group is treated like a uniform buffer object.
```

packages/core/src/shader/UniformGroup.ts
```diff
@@ -61,20 +98,74 @@ export class UniformGroup
...
+    }
+
+    static from(uniforms: Dict<any> | Buffer, _static?: boolean, _ubo?: boolean): UniformGroup
+    {
+        return new UniformGroup(uniforms, _static, _ubo);
     }
 
-    static from(uniforms: Dict<any>, _static?: boolean): UniformGroup
+    /**
+     * A short hand function for creating a static ubo UniformGroup,
```
SukantPal, line 162:
```suggestion
     * A short hand function for creating a static UBO UniformGroup.
```

packages/core/src/shader/UniformGroup.ts
```diff
@@ -1,36 +1,73 @@
...
+    autoManage: boolean;
 
     /**
-     * @param {object} [uniforms] - Custom uniforms to use to augment the built-in ones.
+     * @param {object | Buffer} [uniforms] - Custom uniforms to use to augment the built-in ones. Or a pixi buffer
      * @param {boolean} [_static] - Uniforms wont be changed after creation
+     * @param {boolean} [_ubo] - if true, will treat this uniform group as a uniform buffer object
      */
-    constructor(uniforms: Dict<any>, _static?: boolean)
+    constructor(uniforms: LAYOUT | Buffer, _static?: boolean, _ubo?: boolean)
```
SukantPal, line 69: Why do `_static` and `_ubo` have a `_` prefix?

packages/core/src/shader/utils/generateUniformBufferSync.ts
```diff
@@ -0,0 +1,295 @@
...
+import { mapSize } from '../utils';
+import { IUniformData } from '../Program';
+import { UniformGroup } from '../UniformGroup';
+import { uniformParsers } from './uniformParsers';
+import type { Renderer } from '../../Renderer';
+import type { Buffer } from '../../geometry/Buffer';
+
+export type UniformsSyncCallback = (...args: any[]) => void;
+
+const uboUpdate =  (_ud: any, _uv: any, _renderer: Renderer, _syncData: any, buffer: Buffer) =>
```
SukantPal, line 11: We shouldn't use anonymous arrow functions when unnecessary. It obfuscates stack traces:

```suggestion
function uboUpdate(_ud: any, _uv: any, _renderer: Renderer, _syncData: any, buffer: Buffer)
```

packages/core/src/shader/utils/generateUniformBufferSync.ts
```diff
@@ -0,0 +1,295 @@
...
+    vec4: `
+        data[offset] = v[0];
+        data[offset+1] = v[1];
+        data[offset+2] = v[2];
+        data[offset+3] = v[3];
+    `,
+    mat2: `
+        data[offset] = v[0];
+        data[offset+1] = v[1];
+
```
SukantPal, line 44: I guess we're assuming that the third column of the matrix is always [0, 0, 1]. Same for mat3. It would be nice to make this explicit. But that's not a biggie.

packages/core/src/shader/ShaderSystem.ts
```diff
@@ -179,6 +182,82 @@ export class ShaderSystem implements ISystem
...
+    /**
+     * Will create a function that uploads a uniform buffer using the STD140 standard.
+     * The upload function will then be cached for future calls
+     * If a group is manually managed, then a simple upload function is generated
+     *
+     * @param group - the uniform buffer group to sync
+     * @param glProgram - the gl program to attach the uniform bindings to
+     * @param name - the name of the uniform buffer (must exist on the shader)
+     */
+    createSyncBufferGroup(group: UniformGroup, glProgram: GLProgram, name: string): UniformsSyncCallback
```
SukantPal, line 225:
```suggestion
    protected createSyncBufferGroup(group: UniformGroup, glProgram: GLProgram, name: string): UniformsSyncCallback
```

SukantPal (changes requested):
This is intense; I can't wait to test it out & write about it.

I went through all the code and read the docs to understand what's going on. I think we can make sure our docs maintain a standard of quality; we're PixiJS :) Otherwise, the code itself is very robust. Nice work!

bigtimebuddy:
@GoodBoyDigital build is now passing on actions!

GoodBoyDigital:
> This is intense; I can't wait to test it out & write about it.
> 
> I went through all the code and read the docs to understand what's going on. I think we can make sure our docs maintain a standard of quality; we're PixiJS :) Otherwise, the code itself is very robust. Nice work!

cheers man! Totally agree, dos should be top draw! The written word tends to be my enemy :P so feedback on docs like this is always appreciated! 👍

packages/core/src/geometry/BufferSystem.ts
```diff
@@ -0,0 +1,221 @@
...
+        const { gl, CONTEXT_UID } = this;
+
+        const glBuffer = buffer._glBuffers[CONTEXT_UID] || this.createGLBuffer(buffer);
+
+        gl.bindBuffer(buffer.type, glBuffer.buffer);
+    }
+
+    /**
+     * binds a buffer to a base. A cache is used so a buffer will not be bound again if already bound.
+     * Only used by the uniform buffers
```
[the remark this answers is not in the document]
GoodBoyDigital: agree! transform feedback will definitely be coming in at some point!

packages/core/src/shader/utils/generateUniformBufferSync.ts
GoodBoyDigital: in this instance it should set all values directly from the array for both mat2 and mat3?

packages/core/src/shader/UniformGroup.ts
GoodBoyDigital: possible cos they are optional? 
i'm not really sure :D

packages/core/src/geometry/BufferSystem.ts
GoodBoyDigital: agree! will flip em!

packages/core/src/shader/UniformGroup.ts
GoodBoyDigital: ohh! static is a reserved key word :D

packages/core/src/shader/GLProgram.ts
GoodBoyDigital: added a little more detail in the docs!

bigtimebuddy:
Still seems to be intermittently failing, can you look and see if this actually a bug?

```
Uncaught TypeError: Cannot read property 'refCount' of undefined
      at GeometrySystem.disposeGeometry (/home/runner/work/***/***/packages/core/dist/cjs/core.js:2737:256)
      at GeometrySystem.disposeAll (/home/runner/work/***/***/packages/core/dist/cjs/core.js:2741:298)
      at GeometrySystem.contextChange (/home/runner/work/***/***/packages/core/dist/cjs/core.js:2682:137)
      at Runner.emit (/home/runner/work/***/***/packages/runner/dist/cjs/runner.js:57:578)
      at ContextSystem.handleContextRestored (/home/runner/work/***/***/packages/core/dist/cjs/core.js:2504:175)
```

GoodBoyDigital:
Yo! yeah, seeing some weirdness on my end now too! got this error for interaction tests:

image

@SukantPal any idea what could be going on there?

bigtimebuddy:
That seems weird to me that Interaction would originate this TypeError. Here's the code in  `disposeGeometry` where this is coming from. So `buf` here is undefined. I guess that's because `_glBuffers[this.CONTEXT_UID]` is undefined or else we'd get undefined on `_glBuffers`. 

```typescript
        const buffers = geometry.buffers;
        // ...
        for (let i = 0; i < buffers.length; i++)
        {
            const buf = buffers[i]._glBuffers[this.CONTEXT_UID];

            buf.refCount--;
            if (buf.refCount === 0 && !contextLost)
            {
                this.disposeBuffer(buffers[i], contextLost);
            }
        }
```

GoodBoyDigital:
> That seems weird to me that Interaction would originate this TypeError. Here's the code in `disposeGeometry` where this is coming from. So `buf` here is undefined. I guess that's because `_glBuffers[this.CONTEXT_UID]` is undefined or else we'd get undefined on `_glBuffers`.
> 
> ```ts
>         const buffers = geometry.buffers;
>         // ...
>         for (let i = 0; i < buffers.length; i++)
>         {
>             const buf = buffers[i]._glBuffers[this.CONTEXT_UID];
> 
>             buf.refCount--;
>             if (buf.refCount === 0 && !contextLost)
>             {
>                 this.disposeBuffer(buffers[i], contextLost);
>             }
>         }
> ```

hmmm, perhaps the geometry is being destroyed before its buffer has been created?

bigtimebuddy:
Also, seems like this is triggered by a context restore. Might that have anything to do with it?

GoodBoyDigital:
> Also, seems like this is triggered by a context restore. Might that have anything to do with it?

looks like the context was getting lost right before the destroy is called in the tests!

Added a catch to check gl buffers exist before we destroy em. (if they don't, they were already destroyed so should be all good!)

SukantPal (approved):
Nice & slick!

bigtimebuddy:
Great work @GoodBoyDigital!

PR #7100, opened 2020-12-23
Fix tempRect x,y

SukantPal:
##### Description of change

The `tempRect` x, y values can be non-zero since they are used to store the invert-transformed source frame. That's why the x, y values must be set to zero when storing the texture destination frame.


SukantPal:
Working fiddle:- https://jsfiddle.net/ShukantPal/jpqnh2bk/

bigtimebuddy (approved):

ivanpopelyshev (approved):

bigtimebuddy:
Thanks @SukantPal :tada:

PR #7107, opened 2020-12-27
Uniform Buffer Object integration

GoodBoyDigital:
Hey team!

Here is my first pass at integrating uniform buffer objects. Still working on a few bits, but figured any one interested might like to see where its heading!

main goal is to make using uniform buffers as easy as working with uniforms currently is! 

here is the current api:

```
const srcVert = `

uniform myCoolData { // declaring a ubo..
    mat4 uCoolMatrix;
    float uFloatyMcFloatFace
}

// more shader code...
`

// a new uniform buffer object..
const myCoolData = new UniformBufferGroup({
    uCoolMatrix:new Matrix(),
    uFloatyMcFloatFace: 23,
}}

// build a shader...
const shader = Shader.from(srcVert, srcFrag, {
    myCoolData // will be processed accordingly
})

// modify data..
myCoolData.uniforms.uFloatyMcFloatFace = 23;
```

Like with many pixi aspects, things are handled automatically under the hood. A Buffer is created and properties are automatically managed taking care of the very specific (and initially quite confusing) std40 layout required for uniform buffer objects.

Notes on implementation,

UniformBufferGroups support all float items along with the array versions of them too:

```
float
vec2
vec3
vec4
mat2
mat3
mat4
```

additionally the following pixi types will work:

```
matrix (uploaded as mat3)
point (uploaded as vec2)
rectangle (uploaded as vec4)
```

due to the fact that anything stored in std40 format works, there is no benefit to adding bool / ints as they all take up the same amount of space as a float. It could be done, but will complicate the code as we would require multiple buffer views. Figured best to leave for now!


ivanpopelyshev:
We really need WebGL1 fallback for this thing.

GoodBoyDigital:
i thought about it, for now settled on the idea of exposing the feature and leaving that up to the end user. We can deal with under the hood fallback after api is tight and functional... I do think it would actually be pretty easy! The only tricky bit is the fact we would need to modify the shader, which takes us down a preprocessor path..

ivanpopelyshev:
> The only tricky bit is the fact we would need to modify the shader, which takes us down a preprocessor path..

Oh sh!@

OK, I agree. We need preprocessor first, like I predicted year or two ago.

GoodBoyDigital:
Hey question for peeps.. the newly created `UniformBufferGroup` is _very_ close to `UniformGroup`. Thinking I could merge them? so a uniformGroup could have a ubo mode? what do you think?

ivanpopelyshev:
Agree

GoodBoyDigital:
done! new Syntax to create ubo group:

```

const uboGroup = new UniformGroup({uFloat:23}, true, true);

// or a little shortcut:

const uboGroup = UniformGroup.uboFrom({uFloat:23});

```

ivanpopelyshev (approved):
Tested it on all our examples.

GoodBoyDigital:
yooo @bigtimebuddy! I seem to be failing the lint tests on the CI - but have no errors when running lint locally? Any idea why that may be the case? Cheers man!

bigtimebuddy:
This is very strange indeed. Will take a look.

packages/core/src/shader/utils/generateUniformBufferSync.ts
```diff
@@ -0,0 +1,293 @@
...
+    offset: number,
+    dataLen: number,
+    dirty: number
+}
+
+/**
+ *
+ * logic originally from here: https://github.com/sketchpunk/FunWithWebGL2/blob/master/lesson_022/Shaders.js
+ *
+ * @param uniformData
```
bigtimebuddy, line 106: `@ignore` to hide from public docs?

bigtimebuddy (commented):
Minor doc stuff and questions

packages/core/src/shader/utils/generateUniformBufferSync.ts
GoodBoyDigital: updated 👍

bigtimebuddy (approved):

bigtimebuddy:
@GoodBoyDigital go ahead and merge this into your other PR

PR #7113, opened 2020-12-30
Move member documentation from constructor

bigtimebuddy:
We have some tech debt that needs to be addressed, I'm making this PR to have a conversation about what should be our style regarding docs and TypeScript. I also wanted to expose any Webdoc bugs before enlisting the help of others.

## Goals

* Make the constructor easier to read 
* Expose documentation in typings
* Remove redundant information in docs such as access-modifiers and types

## Style Rules

* Add initial values and member docs to the class-level
* Remove redundant access modifiers (public/private/protected)
* Remove type declarations for all private
* Make docs single line, if possible
* Remove redundant types for members (known issue about this below)

### Blocking Issues

* ~TypeScript types/defaults are not inferred https://github.com/webdoc-labs/webdoc/issues/66~
* ~Readonly is not supported or inferred https://github.com/webdoc-labs/webdoc/issues/67~
* ~Type inference for primitive properties: https://github.com/webdoc-labs/webdoc/issues/72~

### Additional Follow-up

* In AccessibilityManager, there are two instances of `.bind(this)` in the constructor. I'd suggest we use [bind-decorator](https://www.npmjs.com/package/bind-decorator) instead to indicate bound functions. Would make the source a little cleaner.


ivanpopelyshev (approved):

SukantPal:
I published @webdoc@cli@1.1.4 and added the upgrade to this branch. Hopefully, those blocking issues should not pop up again!

bigtimebuddy:
Looks great, thanks @SukantPal 

Verify this works:
http://pixijs.download/fix-member-docs-accessibility/docs/PIXI.AccessibilityManager.html

SukantPal (approved):
LGTM!

PR #7116, opened 2021-01-01
Clear-mode should be optional for Filters

bigtimebuddy:
Small typings bug was discovered upgrading pixi-filters to new types.

The `apply` method for Filters should have an optional `clearMode`. This is defaulted in the FilterSystem class.


themoonrat (approved):

SukantPal (approved):
This will be easier once we strip out typings from the doc comments! (webdoc _should_ infer that the argument is optional)

PR #7118, opened 2021-01-02
Bounds integration test circle radius

jasonsturges:
It appears that `radius` parameter is being passed twice to `drawCircle` in the following tests.

```js
drawCircle(0, 0, 10, 10);
```

That doesn't match the function signature of `drawCircle`:

```ts
public drawCircle(x: number, y: number, radius: number): this
```


bigtimebuddy (approved):
Good catch.

Zyie (approved):

PR #7121, opened 2021-01-03
Remove type coupling between core and display

bigtimebuddy:
### Bug

When attempting to install `@pixi/core` without `@pixi/display`, you get this error. There was previous work to decouple runtime the core and display (#6450), but the types were still creating this unnecessary coupling.

```
node_modules/@pixi/core/index.d.ts:10:31 - error TS7016: Could not find a declaration file for module '@pixi/display'. '/node_modules/@pixi/display/lib/display.js' implicitly has an 'any' type.
  Try `npm install @types/pixi__display` if it exists or add a new declaration (.d.ts) file containing `declare module '@pixi/display';`

10 import { DisplayObject } from '@pixi/display';
```

### Description

Created two interfaces that define the surface area between the Renderer and the DisplayObject/Container. These are called IRenderableObject and IRenderableContainer. Potentially this could be a single interface since we only render Containers, in practice.

### DIFF
--- a/packages/core/src/AbstractRenderer.ts
+++ b/packages/core/src/AbstractRenderer.ts
@@ -2,12 +2,11 @@ import { hex2string, hex2rgb, EventEmitter, deprecation } from '@pixi/utils';
 import { Matrix, Rectangle } from '@pixi/math';
 import { RENDERER_TYPE } from '@pixi/constants';
 import { settings } from '@pixi/settings';
-import { DisplayObject } from '@pixi/display';
 import { RenderTexture } from './renderTexture/RenderTexture';
 
 import type { SCALE_MODES } from '@pixi/constants';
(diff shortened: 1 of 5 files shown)


ivanpopelyshev (approved):
Finally

Zyie (approved):

graphnode:
Hi, FYI this merge broke typing with @pixi/events since the latter one is using a DisplayObject in it's own Renderer interface: https://github.com/pixijs/pixijs/blob/95d15c11f8025eec9c00c6685bf8a1b200778218/packages/events/src/EventSystem.ts#L20

It would probably make more sense for @pixi/events use AbstractRenderer from Core or at least change the interface to use the new IRenderableObject but might break again in the future.

bigtimebuddy:
@graphnode could you please file a new issue?

PR #7122, opened 2021-01-03
Add optional clearMode argument type for Blur and Displacement

bigtimebuddy:
Follow-up to #7116

With the `apply` override, the `clearMode` should support being optional.

### DIFF
--- a/packages/filters/filter-blur/src/BlurFilter.ts
+++ b/packages/filters/filter-blur/src/BlurFilter.ts
@@ -48,9 +48,9 @@ export class BlurFilter extends Filter
      * @param {PIXI.FilterSystem} filterManager - The manager.
      * @param {PIXI.RenderTexture} input - The input target.
      * @param {PIXI.RenderTexture} output - The output target.
-     * @param {PIXI.CLEAR_MODES} clearMode - How to clear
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
Here lies a problem:

for `FilterSystem.bindAndClear`, YES is default , but it has `clear()` in its name and description of default param in docs - that's very **clear**

ok, for `FilterSystem.applyFilter()` , YES is default, but its not mentioned in docs, but its being used at the same level of code as bindAndClear so it should be fine.

for `Filter.apply` i dont think it'll be good to make this parameter optional, because 
1. in current usage, this param is always supplied
2. some people might think they can just ignore it and not pass forward

SukantPal (approved):
This will affect all downstream filters.

bigtimebuddy:
@ivanpopelyshev many of our pixi-filters ignore this parameter, i'm making the typings consistent with our implementation. If you want to do it another way, that's fine but would mean changing things downstream.

bigtimebuddy:
Ivan and I slacked about this. I'm going to close this PR and add the defaults in the filters. Keeping this without a default ensures developers who are calling a filter's `apply` directly (or overriding) know what they are doing. Better to be more explicit about `clearMode` in this case.

We need to create more examples which demonstrate different clear mode behaviors.

PR #7124, opened 2021-01-04
Use @pixi/eslint-config for common style rules across PixiJS projects

bigtimebuddy:
We have an awesome package called [@pixi/eslint-config](https://github.com/pixijs/eslint-config) which defines the ESLint styles used in our TypeScript projects. @SukantPal created this as a fork of the current styles in this repo. Since then, the styles have drifted slightly. I have upstreamed these changes in the eslint-config package. I think externalizing the styles is easier to maintain and provides a better base for other projects.(also it's easy to override in the future, if need be).

Upgrading this exposed a few issue with multiple imports in TypeScript. Those errors have been fixed.


SukantPal (approved):
This was the goal all along!

Zyie (approved):

PR #7146, opened 2021-01-12
Fix documentation and types for Graphics' lineStyle

### LINKED ISSUE #7141: Missing Application plugins & Graphics' lineStyle types
I just updated to PixiJS v6 RC and wanted to check if the updated types in v6 broke our Angular project and I get this errors that certain attributes don't exist anymore on certain types. If I check the v6 docs, the attributes are still there and not deprecated. Then I saw, that the pixi.js.d.ts is missing from node_modules/pixijs and maybe that confuses the typescript compiler.

I get this error from VS Code:

`"Unable to open 'pixi.js.d.ts': Unable to read file '/Users/toms/Coding/WaveModeler/node_modules/pixi.js/pixi.js.d.ts' (Error: Unable to resolve non-existing file '/Users/toms/Coding/WaveModeler/node_modules/pixi.js/pixi.js.d.ts')."`

And a lot of compilation errors like this:
`ERROR in src/app/main/processTabContent/editor/editor.component.ts:102:18 - error TS2339: Property 'loader' does not exist on type 'Application'.
this.pixiApp.loader.load(() => this.pixiApp.renderer.render(this.pixiApp.stage));`

### Environment

- **`pixi.js` version**: 6.0.0-rc
- **OS & Version**: macOS 11.1 Big Sur
- **Angular Version**: Angular 10
- **Typescript Version**: 3.9.7

bigtimebuddy:
Partially fixes #7141 

### Fixes

* Object-based 'lineStyle' was missing from documentation
* Param-based 'lineStyle' was missing from types 

### 🔗  Links

* [Graphics#lineStyle](https://pixijs.download/fix-line-style/docs/PIXI.Graphics.html#lineStyle)

### DIFF
--- a/packages/graphics/src/Graphics.ts
+++ b/packages/graphics/src/Graphics.ts
@@ -333,8 +333,6 @@ export class Graphics extends Container
      * Specifies the line style used for subsequent calls to Graphics methods such as the lineTo()
      * method or the drawCircle() method.
      *
-     * @instance
-     * @method PIXI.Graphics#lineStyle
      * @param {number} [width=0] - width of the line to draw, will update the objects stored style
      * @param {number} [color=0x0] - color of the line to draw, will update the objects stored style
      * @param {number} [alpha=1] - alpha of the line to draw, will update the objects stored style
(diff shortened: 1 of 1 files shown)


Zyie (approved):

themoonrat:
Is one of these to be 'deprecated', like you did recently with the render texture create function? Or are we supporting both for now, and one to be deprecated later?

themoonrat (approved):

bigtimebuddy:
I remember a discussion we had awhile ago about this and we decided not to deprecate intentionally. Mostly because it's kinda easy in most cases to do (width, color) rather than add an options object. It's easy to maintain both for now and we can deprecate in the future if need be.

Issue #7130, opened 2021-01-06
Feature Request: Support line alignment on CanvasRenderer


bigtimebuddy:
Example of the incompatibility: https://jsfiddle.net/bigtimebuddy/j2ep45do/

CanvasRenderer does not support rendering alignment for strokes. This would require some changes to the CanvasGraphicsRenderer to get this to work. Since context2d does not have a stroke option for alignment, this would need to be accomplished with some type of clipping. 

PRs welcome!

denim2x:
I'm working on this

denim2x:
This [demo](https://jsfiddle.net/x9ec8Lg2/2/) shows the result:
![image](https://user-images.githubusercontent.com/8561801/104322905-271b9200-54ee-11eb-98fd-8c687e430097.png)

ivanpopelyshev:
Workaround for max 10 shapes, this thing is very slow. 

We have to allow webgl computation for those shapes and use it as input for canvas line.

denim2x:
Working on it

SukantPal:
I don't like races on bounties

bigtimebuddy:
@SukantPal fwiw, I don't like races on bounties either. I don't like people feel like they are competing. This was the first situation where this has happened. I've always been off the mind that first-to-submit gets considered first, first-to-merge get the bounty.

In this case, I'm also happy to pay bounties to both @denim2x and @SerG-Y because they are both helping move this along and it's a relatively small bounty. I appreciate the enthusiasm and desire to improve the canvas renderer. I want this process to be collaborative, so let's just figure out the best PR for performance & maintainability and go with that.

SerG-Y:
@bigtimebuddy just want to clarify :) I don't need bounties as I pursue another goal while doing a contribution to Pixi, so in any case, we can just pay to @denim2x. 
I created my own PR as I saw the first @denim2x realization and I had another vision how to do it without creating an additional canvas that's more performing solution.

denim2x:
My code should be better performing, also no artifacts are present:

![DEMO result](https://user-images.githubusercontent.com/8561801/105879193-8ad3ae00-600a-11eb-990f-5f6fbef2d1d2.png)

SukantPal:
Looks like this got fixed?

PR #7150, opened 2021-01-14
Added support of line alignment on CanvasRenderer

SerG-Y:
##### Description of change

Implementation of #7130 

Example: 
https://jsfiddle.net/SerG_Y/qycon5Lf/

##### Pre-Merge Checklist


SukantPal:
This fails with a line alignment of 0.75, for example (middle):

Screen Shot 2021-01-15 at 9 35 00 AM

packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
```diff
@@ -73,6 +73,118 @@ export class CanvasGraphicsRenderer
...
+            context.fillRect(rectangle.x, rectangle.y, rectangle.width, rectangle.height);
+        }
+
+        if (lineStyle.visible)
+        {
+            if (innerStrokeAlignment)
+            {
+                context.save();
+                context.rect(rectangle.x, rectangle.y, rectangle.width, rectangle.height);
+                context.clip();
```
SukantPal, line 168: Instead of using clipping for rectangles, how about we draw two rectangles and the `evenodd` fill rule to do a stroke? That would be faster since the Canvas2D implementation won't have to do clipping calculations.

packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
```diff
@@ -73,6 +73,118 @@ export class CanvasGraphicsRenderer
...
+     * draw shape on canvas
+     *
+     * @private
+     * @param {PIXI.FillStyle} fillStyle
+     * @param {PIXI.LineStyle} lineStyle
+     * @param {string|CanvasPattern} contextFillStyle
+     * @param {string|CanvasPattern} contextStrokeStyle
+     * @param {number} worldAlpha
+     */
+    private draw(
```
SukantPal, line 86: I dunno if we should call this `paint` (fill + stroke = paint) since `draw` is for actually pushing a primitive/shape into the rendering pipeline (cc @bigtimebuddy)

SukantPal (changes requested):

SerG-Y:
Hey @SukantPal, @bigtimebuddy
I am sorry, it's my fail because I considered only support alignment exactly like [0, 0.5, 1]. Should we support alignment only in the range [0-1] or it can be any number?

SukantPal:
It can be any number. Usually in (0-1) range

packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
bigtimebuddy: It's private, so it doesn't matter to me. If it were public, I'd scrutinize this more. `paint` is fine too.

bigtimebuddy:
I appreciate how clean this PR is @SerG-Y. Awesome work. My biggest request is the float 0-1. I think for numbers outside that range, we can clamp or consider it undefined behavior.

bigtimebuddy:
I like this but there are some limitations, for example what about polygons? For example, `drawStar` is a closed shape that uses `drawPolygon`.

https://jsfiddle.net/bigtimebuddy/9ga2kod0/

SerG-Y:
Hey @bigtimebuddy, thanks for the review, I appreciate it. It's still in progress. I know about this limitation and I am working on it, hope will be able to deliver it till the end of the week. Also, it has some glitches (rounded rectangle, ellipse) when stroke width bigger than the radius, hope I will be able to fix it as well. Thanks! 
Screen Shot 2021-01-25 at 13 42 56

SerG-Y:
Hey @bigtimebuddy, sorry for so long delay, It a bit harder than I thought for the first time (just support 3 numbers [0, 0.5, 1])😄. I updated my PR, added line alignment support for polygon and fixed some glitches.

I have one question, I see that for polygon in WebGL we not consider is polygon clockwise or not, should we consider it?
When `alignment=1` but stroke positioned inside the shape
Example:
https://jsfiddle.net/SerG_Y/7L3svu8t/

PR examples:
1) Shapes -  https://jsfiddle.net/SerG_Y/qycon5Lf/
2) Polygon - https://jsfiddle.net/SerG_Y/vjgz129b/

SukantPal:
Serhiy, I think we had remember that we decided against checking whether a polygon was clockwise or not for alignment. That’s because it would be impossible to determine the inside/outside of an open polyline. Other thoughts welcome on this.

bigtimebuddy:
Clearly the clockwise bias is still in there for WebGL. I think this is a bug for WebGL and I agree with @SukantPal that we should ignore the orientation. I think addressing that is outside the scope of this PR. We should be trying to do in WebGL what you're doing with canvas here, not the other way around.

SerG-Y:
Hey @bigtimebuddy, @SukantPal will you be able to review it? Maybe I can help somehow (additional examples etc.)?

SukantPal:
Things have slowed a little on the PRs. I'll try to get a review in the next 36 hours. Even my #7213 PR needs more eyes :)

bigtimebuddy (approved):

SukantPal (approved):
👍 Let's see how this does in the wild!

bigtimebuddy:
Thanks @SerG-Y for your hard work on this and for handling all the feedback.

SerG-Y:
@ivanpopelyshev Thanks, man!

PR #7158, opened 2021-01-15

denim2x:
##### Description of change

The `CanvasGraphicsRenderer` now performs the actual drawing onto a secondary off-screen `RenderingContext`; the stroke is drawn by compositing drawing operations according to specified `strokeStyle.alignment`. The result then is finally drawn onto the main `RenderingContext`.

**Performance**:
 - a _shared_ secondary `RenderingContext` is acquired whenever possible - i.e. prior to rendering multiple objects;
 - drawing resorts to the secondary `RenderingContext` _only_ when actually necessary - i.e. when the **stroke** is _visible_ and _either_ the _alignment_ is non-**middle** _or_ the **fill** is also _visible_.

[**DEMO**](https://jsfiddle.net/6wfnm5o1/)

DEMO result

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas/canvas-display/src/Container.ts
+++ b/packages/canvas/canvas-display/src/Container.ts
@@ -28,12 +28,13 @@ Container.prototype.renderCanvas = function renderCanvas(renderer: CanvasRendere
         return;
     }
 
-    // This newly forged CanvasRenderer will provide a secondary RenderingContext for
-    //   certain drawing operations (i.e. strokes); this temporary RenderingContext is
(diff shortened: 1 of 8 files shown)


bigtimebuddy:
My general feedback here @denim2x is that there's _a lot_ going on in the PR. In particular, a lot of unwanted changes to core, new constants, new types, new convenience methods, etc. The blast-radius of this PR is way bigger than it should be. Ultimately, this is providing support for something that should be invisible for users.

Can you accomplish the same thing without keeping your change more localized to canvas-graphics and without all the extra overhead or abstractions?

Issue #7159, opened 2021-01-16
Help Wanted: Reformatting Documentation


bigtimebuddy:
## Overview

Recently, we converted our documentation to use [webdoc](https://github.com/webdoc-labs/webdoc/) instead of JSDoc. The advantage of this switch is that much more can be inferred from the TypeScript. Right now, the documentation has type cruft from JSDoc, making it kinda unnecessarily verbose. Also, the documentation for members was historically placed within the constructor, but we'd like to move it to the root of the class where the TypeScript member is declared.

We are looking for help! These are straightforward fixes, but require many hands. If you have a free moment, please consider updating a class below (keep in mind that not all classes will need converting).

💥  **BONUS POINTS FOR:**

* Adding missing descriptions
* Improving descriptions
* Fixing typos

**Please please please _no code or logic changes_.**

## Goals

* Make the constructor easier to read 
* Expose documentation in typings
* Remove redundant information in docs such as access-modifiers and types

## Style Rules

* Add initial values and member docs to the class-level
* Remove redundant access modifiers (public/private/protected)
* Remove type declarations for all private
* Make docs single line, if possible
* Remove redundant types for members
* Make sure that any default values or types with `PIXI.*` are retained (or else Webdoc won't link)

### Example 
Old format:
```javascript
/**	
 * Setting this to true will visually show the divs.	
 *	
 * @type {boolean}
 * @public
 * @default false
 */	
this.debug = false;
```
New format:
```typescript
/** Setting this to true will visually show the divs. */
public debug = false;
```

Here's an example I created with AccessibilityManager #7113 

### Classes To Convert

- [x] packages/accessibility/src/AccessibilityManager.ts #7113
- [x] packages/app/src/Application.ts #7160
- [x] packages/app/src/ResizePlugin.ts #7160
- [x] packages/basis/src/BasisLoader.ts #7584 
- [x] packages/basis/src/TranscoderWorker.ts #7584 
- [x] packages/canvas/canvas-extract/src/CanvasExtract.ts #7584 
- [x] packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts #7584 
- [x] packages/canvas/canvas-mesh/src/CanvasMeshRenderer.ts #7584 
- [x] packages/canvas/canvas-prepare/src/CanvasPrepare.ts #7584 
- [x] packages/canvas/canvas-renderer/src/CanvasRenderer.ts #7584 
[...]

PR #7160, opened 2021-01-17
Format the application docs to be exported

bigtimebuddy:
### Fixed

* Update the documentation according to: #7159 
* Fixes docs bug with instance methods `start` and `stop` of TickerPlugin
* Renamed typedef `PIXI.Application.Plugin` to match types `PIXI.IApplicationPlugin`
* Fix types for `IApplicationPlugin` to be stricter

### webdoc Bug

* ~~Issue with `@typedef` + TypeScript `interface` https://github.com/webdoc-labs/webdoc/issues/76~~


SukantPal:
I published @webdoc/cli@1.1.5 and fixed the issues. This PR should work fine with it now. I don't understand tho - why do we need `IApplicationPlugin` to be a type definition instead of an interface?

bigtimebuddy:
I'm not sure I understand your question @SukantPal. We used typedef in JSDoc because there was no interface doc type. Is there in webdoc?

SukantPal:
I think JSDoc 3 has [@interface](https://jsdoc.app/tags-interface.html). We already have them on the dev documentation: [IBitmapFontOptions](http://pixijs.download/dev/docs/PIXI.IBitmapFontOptions.html) for example.

webdoc already supports interfaces from Typescript source code. So you don't have to use the interface tag anymore.

```ts
/**
 * @memberof PIXI
 */
export interface IApplicationPlugin {
    /**
     * Called when Application is constructed, scoped to Application instance. Passes in `options` as
     * the only argument, which are Application constructor options.
     *
     * @param options
     */
    init(options: IApplicationOptions): void;

    /**
     * Called when destroying Application, scoped to Application instance
     */
    destroy(): void;
}
```

bigtimebuddy:
That's way better

SukantPal (approved):

Zyie (approved):

PR #7161, opened 2021-01-17

denim2x:
##### Description of change

The `CanvasGraphicsRenderer` now performs the actual drawing onto a secondary off-screen `RenderingContext`; the stroke is drawn by compositing drawing operations according to specified `strokeStyle.alignment`. The result is then finally drawn onto the main `RenderingContext`.

**Performance**:
 - the same off-screen `RenderingContext` is reused between multiple `render()` calls;
 - off-screen rendering is actually used _only_ when necessary (i.e. when special rendering is needed - e.g. for stroke alignment);
 - the off-screen `RenderingContext` is _released_ from memory when rendering is finished.

[**DEMO**](https://jsfiddle.net/6wfnm5o1/)

DEMO result

##### Pre-Merge Checklist

- [ ] Write class `Painter`?

### DIFF
--- a/packages/app/global.d.ts
+++ b/packages/app/global.d.ts
@@ -10,6 +10,6 @@ declare namespace GlobalMixins
     // eslint-disable-next-line @typescript-eslint/no-empty-interface
     interface IApplicationOptions
     {
-        resizeTo: Window|HTMLElement;
+        resizeTo?: Window|HTMLElement;
     }
 }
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
Any idea why there's are weird artifacts on canvas renderer (@ 0.75 and 0.25 alignments)

Screen Shot 2021-01-18 at 3 54 06 PM

PR #7183, opened 2021-01-20
Allow passing BaseTexture options to Loader

### LINKED ISSUE #7148: Feature Request: Support Resource Options passed with Loader
Currently, it's not possible to pass options with Loader for Textures. This can be helpful for things like SVG where you want to pass the scale when you load it. Or you want to set the BaseTexture's scaleMode, etc.

bigtimebuddy:
Fixes #7148

Now BaseTexture options can be passed to the metadata field.

```js
loader.add('bunny', 'bunny.png', { metadata: {
  scaleMode: PIXI.SCALE_MODES.NEAREST,
  resolution: 1,
}});

loader.add('logo', 'logo.svg', { metadata: {
  resourceOptions: {
    scale: 0.25
  }
}});
```

### 🔥  Breaking Change

* This changes the default load type for SVG images. Instead of loading as an `` element, SVGs are loaded as text with resource-loader, and parsed with SVGResource.

### 🔬  Demos

https://jsfiddle.net/bigtimebuddy/gs6d81qk/ (png)
https://jsfiddle.net/bigtimebuddy/qwb6nduv/ (svg)

### Follow-up Needed

* Error handling for Textures & Loader is awful. We need to do a better job of handling internal errors with the `fromLoader` function. I suggest we return a Promise like `fromUrl`, and make sure the Loader does not swallow errors.


ivanpopelyshev (approved):
It works!
Though it needs unit-tests for loader

+1 for error Handling problems

bigtimebuddy:
I have some ideas how to handle resource errors and bubble them up to the loader. Will work on that next. But yes on unit tests. Will add.

bigtimebuddy:
Okay I added unit-tests which uncovered an async error with SVGs. Basically, I needed to wait until the resource was completely finished loading before continuing with the Loader.

ivanpopelyshev:
Good!

SukantPal (approved):
LGTM!

saraOrkide:
How can I zoom in on svg sharply and control the gpu to slow it down?
Or how can I update the resolution or scale value each time I zoom in

ivanpopelyshev:
Its possible, but so far people who did it didnt share :) If you give a demo that needs quality - someone can do that. From grounds up? sorry, no.

ShukantPal:
You could use @pixi-essentials/svg for sharp rendering of SVG files

saraOrkide:
> You could use @pixi-essentials/svg for sharp rendering of SVG files

When the number of svg paths is high and the use tag is used, it becomes very slow and does not support the pattern.

saraOrkide:
> Its possible, but so far people who did it didnt share :) If you give a demo that needs quality - someone can do that. From grounds up? sorry, no.

Well, there is no way I can better display svg with resolution and scaled when zooming

ShukantPal:
I think you can increase the width/height of the SVG, and then create a texture from it at that higher size

PR #7185, opened 2021-01-20

denim2x:
##### Description of change

The `CanvasGraphicsRenderer` now performs the actual drawing onto a secondary off-screen `RenderingContext`; the stroke is drawn by compositing drawing operations according to specified `strokeStyle.alignment`. The result is then finally drawn onto the main `RenderingContext`.

**Performance**:
 - the same off-screen `RenderingContext` is (re)used for all rendering (when `render()` is called on the `CanvasRenderer`);
 - off-screen rendering is _only_ used for blending _correctly_ - i.e. the shape being rendered against the current graphics;
 - the off-screen `RenderingContext` is _released_ when all rendering is done.

[**DEMO**](https://jsfiddle.net/gq8p1uhL/)

DEMO result

##### Pre-Merge Checklist

- [x] Visual artifacts eliminated


bigtimebuddy:
@denim2x Thanks so much for all your work on this effort but we're going to go with #7150 instead. It much simpler and easier to maintain.

PR #7199, opened 2021-01-23
Update documentation for Ticker and CanvasRenderTarget

Zyie:
##### Description of change

Updates documentation for `Ticker`, `TickerListener`, and `CanvasRenderTarget` to be in line with the new style.

There is 1 issue i have found and raised at https://github.com/webdoc-labs/webdoc/issues/79 where default values do not appear when the value is negative.

So `lastTime` in `Ticker` will not show a default


bigtimebuddy (approved):
LGTM! Thanks @Zyie. Also, @SukantPal if you could take a look at that default issue. I remember this also being an issue with default arg values, but I believe you fixed that.

bigtimebuddy:
@SukantPal you can see here that `lastTime` is missing default and type:
```typescript
public lastTime = -1;
```
http://pixijs.download/dev-ticker-docs/docs/PIXI.Ticker.html

ivanpopelyshev (approved):

PR #7211, opened 2021-01-29
Fix for animated sprite not stopping

Owlzy:
##### Description of change

Fixes issue where animated sprite continues to play even when `_playing = false`

##### Pre-Merge Checklist

### DIFF
--- a/packages/sprite-animated/src/AnimatedSprite.ts
+++ b/packages/sprite-animated/src/AnimatedSprite.ts
@@ -260,6 +260,11 @@ export class AnimatedSprite extends Sprite
      */
     update(deltaTime: number): void
     {
+		if (!this._playing)
+        {
+            return;
+        }
+		
         const elapsed = this.animationSpeed * deltaTime;
         const previousFrame = this.currentFrame;
 
--- a/packages/sprite-animated/src/AnimatedSprite.ts
+++ b/packages/sprite-animated/src/AnimatedSprite.ts
@@ -260,7 +260,7 @@ export class AnimatedSprite extends Sprite
      */
     update(deltaTime: number): void
     {
-		if (!this._playing)
+        if (!this._playing)
         {
             return;
         }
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks for this @Owlzy. Can you provide an example of this not work so we can verify?

Owlzy:
> Thanks for this @Owlzy. Can you provide an example of this not work so we can verify?

Sure.  Having made a playground I now see it is specific to autoUpdate = false. 

In this playground can see that animation plays regardless of calling stop().  If loop is true it will continue looping, or if not looping it will play till the last frame and stop.  

[](https://www.pixiplayground.com/#/edit/1EdyVzCi9_Djk1FbjhyGw)

bigtimebuddy:
Got it, this example is help for context. At first, it seemed a little redundant to add this check, but now I see why because you are enabling `autoUpdate`. 

I fine with this, but additional, I think it might be nice to add `play` and `stop` events. So the user could know when to remove from the ticker (which is what happens of `autoUpdate` is true). For example:

```js
const sprite = new AnimatedSprite();
sprite.autoUpdate = false;
sprite.on('play', () => ticker.add(sprite.update, sprite) });
sprite.on('stop', () => ticker.remove(sprite.update, sprite) });
```

Play and stop would allow you to remove from the ticker or event loop and remove that call all together, which means you wouldn't need to defensively add the check in update. I think there's value in both, just a thought.

bigtimebuddy (approved):
Can you please fix the linting? Otherwise, I'm fine with this.

Owlzy:
> Can you please fix the linting? Otherwise, I'm fine with this.

Sorry thought I fixed that already. Should be fixed in the latest commit.

themoonrat (approved):

bigtimebuddy:
Thank you @Owlzy :tada:

PR #7213, opened 2021-01-30
Federated Events API (v6+7)

SukantPal:
RFC #7071 

Ready for review / description needs more editing

The `EventSystem` will eventually be defaulted over the `InteractionManager` in v7. Until then, @pixi/events will be a standalone package, not included in the `pixi.js` bundle (I have to remove that registration in `pixi.js` before merging this PR). @bigtimebuddy suggested we could have a `pixi.js-next` (I prefer `@pixi/next` bundle) that uses EventSystem over the InteractionManager.

## Breaking for

Developers doing the following will have to patch their code:

* referencing the interaction plugin (`renderer.plugins.interaction`)
* assuming the pointer event passed to listeners will always be the same (as our dragging example does in pixijs.io/examples)
* relying on `pointermove` and `mousemove` events firing all over the scene graph instead of just the actual mouse target. The new semantics match that of the DOM; the `moveOnAll` flag can be set to use legacy (InteractionManager default) behavior.

### DisplayObject

* addEventListener
* removeEventListener
* dispatchEvent

### EventSystem + EventBoundary

* hitTest
* dispatchEvent
* mapEvent
* addEventMapping
* pointerenter/mouseenter events
* pointerleave/mouseleave events
* wheel events

##### Pre-Merge Checklist


bundles/pixi.js/src/index.ts
```diff
@@ -2,7 +2,7 @@ import '@pixi/polyfill';
 
 import * as utils from '@pixi/utils';
 import { AccessibilityManager } from '@pixi/accessibility';
-import { InteractionManager } from '@pixi/interaction';
+// import { InteractionManager } from '@pixi/interaction';
```
bigtimebuddy, line 5: Would like to leave **pixi.js** as is for v6.

packages/core/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "@pixi/constants": "6.0.0-rc.3",
+    "@pixi/events": "^6.0.0-rc.3",
```
bigtimebuddy, line 33: Again, to start, I'd like to remove the EventSystem from the renderer. Users can `renderer.addSystem(EventSystem, 'events')` until I get the static installing of systems working with Renderer.

packages/events/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+## Usage
+
+The `pixi.js` and `pixi.js-legacy` bundles add the `EventSystem` to renderers by default.
+
+```ts
+import { EventSystem } from '@pixi/events';
+
+renderer.addSystem('events', EventSystem);
```
bigtimebuddy, line 19:
```suggestion
renderer.addSystem(EventSystem, 'events');
```

packages/events/src/EventBoundary.ts
```diff
@@ -0,0 +1,1651 @@
...
+import type { DisplayObject } from '@pixi/display';
+import type { FederatedEvent } from './FederatedEvent';
+// The maximum iterations used in propagation. This prevent infinite loops.
+const PROPAGATION_LIMIT = 2048;
+
+const tempHitLocation = new Point();
+const tempLocalMapping = new Point();
+
+/**
+ * The tracking data for each display object.
```
bigtimebuddy, line 16: `@memberof PIXI`?

packages/interaction/src/InteractionManager.ts
```diff
@@ -521,212 +521,6 @@ export class InteractionManager extends EventEmitter
          * @param {PIXI.InteractionEvent} event - Interaction event
          */
 
-        /**
```
bigtimebuddy, line 524: Why are you removing these event docs? Maybe we need to add something to this that's mention it's from `@pixi/interaction`? We're going to need to have duplicate events.

SukantPal: Hmm, I can leave them there. It will be a bit weird for each event to be listed twice.

bigtimebuddy: That's why I think we need to list the packages in both because they have different params.

SukantPal:
@bigtimebuddy Can you have a look at the build? Can't figure out why JSCC fails in the `build:prod` script. The events.min.js file is generated

packages/events/src/EventBoundary.ts
SukantPal: I add full `@typedef` docs for now 👍

bundles/pixi.js/src/index.ts
SukantPal: Ah, yes. This was just to demonstrate which examples still work as-is with the EventSystem. I'll revert this.

packages/interaction/src/InteractionManager.ts
SukantPal: Done!

bigtimebuddy:
@SukantPal, the reason you were getting that error is the events dependency was missing from the package.json of **pixi.js**. That's been fixed.

bundles/pixi.js/src/index.ts
SukantPal: Reverted 👍

packages/core/package.json
SukantPal: Reverted 👍

packages/events/src/EventBoundary.ts
```diff
@@ -6,6 +6,7 @@ import { Point } from '@pixi/math';
 import type { Cursor, FederatedEventTarget } from './FederatedEventTarget';
 import type { DisplayObject } from '@pixi/display';
 import type { FederatedEvent } from './FederatedEvent';
+import { EventEmitter } from '@pixi/utils/src';
```
bigtimebuddy, line 9:
```suggestion
import { EventEmitter } from '@pixi/utils';
```

bigtimebuddy: I realize that sometimes VSCode auto-completes this. Be mindful not to do it tho.

SukantPal: Fixed 👍

packages/events/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Installation
+
+```bash
+npm install @pixi/events
+```
+
+## Usage
+
+The `pixi.js` and `pixi.js-legacy` bundles add the `EventSystem` to renderers by default.
```
bigtimebuddy, line 14: Line needs updating.

packages/events/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+```ts
+import { EventSystem } from '@pixi/events';
+
+renderer.addSystem(EventSystem, 'events');
+renderer.render(stage);// Render stage so that it becomes the root target for UI events
```
bigtimebuddy, line 20: Is it required to render once first before adding listeners?

Also maybe use Application (or new Renderer) here for a full working example.

packages/interaction/src/interactiveTarget.ts
```diff
@@ -92,42 +92,8 @@ export interface InteractiveTarget {
  *      );
  */
 export const interactiveTarget: InteractiveTarget = {
-
```
bigtimebuddy, line 95: Can you revert all these changes to interactiveTarget? Events should conflict with it.

packages/events/README.md
SukantPal: Not for registering listeners, but for actually having the events fire - yes (the renderer doesn't know about the stage)

packages/interaction/src/interactiveTarget.ts
SukantPal: I don't see the purpose of duplicating these properties in the documentation? They have the same signature and description?

GoodBoyDigital (approved):
beefy! well documented and looks great! Can't wait to battle test it :D

ivanpopelyshev (approved):
I dont like the fact that this thing is a renderer system.

However, I see that we can add option for app-plugin lovers later.

bigtimebuddy:
Nice work @SukantPal, super excited for this one.

PR #7222, opened 2021-02-04
add support for R32F

joshuahhh:
##### Description of change

This might be somewhat niche, but I wanted a single-channel float texture for my project. So I've added support for that.

An alternative / adjunct to this would be making `internalFormat` an option when creating a `BaseTexture`, so that users could set it by hand for their niche uses. I wouldn't mind that either, but I figured this was the minimal clean way to get what I wanted.

Please let me know what you think. Thanks!

##### Pre-Merge Checklist


joshuahhh:
That's funny; tests seem to be passing on my machine. Let me try something diagnostically...

joshuahhh:
OK that's weird! One of those tests is for the new code path. The other I just added is for an old code path. Both of them should pass. Both pass on my computer (with `npm run unit-test`). Does anyone know why they might be failing on the test server? (Thanks for any help!)

bigtimebuddy:
I'm not sure why it's failing. but it seems like `renderer.gl` is missing `R32F` and `RGBA32F` on GitHub Actions.

One thing worth trying is to console.log props on `gl` to see what the tests here return:

```js
console.log(Object.keys(this.renderer.gl.__proto__).filter(i=> i.startsWith("R")));
```

Also, when I run your tests locally:

```sh
npm run unit-test:debug -- --package=@pixi/core
```
I get this warning in the console:
```
BaseImageResource.ts:96 WebGL: INVALID_VALUE: texImage2D: invalid internalformat
```

For the test `should set internalFormat correctly for red-channel float textures`.

ivanpopelyshev:
> it seems like renderer.gl is missing R32F and RGBA32F on GitHub Actions.

WebGL 1

So we have to test it on webgl1, whats internalFormat for RED in that case if R8 doesnt exist?

ivanpopelyshev:
Usually its handled by custom user BufferResource: https://www.pixiplayground.com/#/edit/xpLx-ElnpPPPXFJUkxoRa

We are sure that we can't handle all the cases. We really dont know where to stop adding formats. That's why I'm afraid of it: https://github.com/playcanvas/engine/blob/master/src/graphics/graphics-device.js

However R32F might be a good general format, shame it does not exist in webgl1

packages/core/test/TextureSystem.js
```diff
@@ -53,6 +53,31 @@ describe('PIXI.TextureSystem', function ()
         expect(glTex.wrapMode).to.equal(WRAP_MODES.CLAMP);
     });
 
+
```
bigtimebuddy, line 56:
```suggestion
```

bigtimebuddy (approved):

bigtimebuddy:
Finally, build is green @joshuahhh, sorry this took so long to sort.

joshuahhh:
Woohoo! Thanks @bigtimebuddy. Haven't been following the details, but I'm happy to see the test situation has been improved! :)

GoodBoyDigital (approved):
nice!

bigtimebuddy:
Thank you @joshuahhh :tada:

PR #7223, opened 2021-02-04
remove extractUniformsFromSrc

joshuahhh:
##### Description of change

Based on a search, I don't think this file is referenced anywhere. To avoid confusion, I suggest we remove it.

(Thanks!)

##### Pre-Merge Checklist


bigtimebuddy (approved):
You're right, don't know how we missed this. Thanks :+1:

themoonrat (approved):

PR #7227, opened 2021-02-05
fix instances of '.to.be.notnull' in tests

joshuahhh:
##### Description of change

A few tests use the construction `.to.be.notnull`. I could be missing something but I don't think `notnull` is defined or does anything. (Unfortunately, chai's cute "you don't need to put parentheses at the end!" use of properties means that this wont trigger an error. 🤷‍♂️)

I've replaced them with `.to.exist`.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good change and more readable.

themoonrat (approved):

bigtimebuddy:
Thanks you @joshuahhh

Issue #7229, opened 2021-02-05
Too strict TypeScript type for second argument of Mesh constructor


lazarljubenovic:
### Expected Behavior

The following code works, and is based on https://pixijs.io/examples/#/mesh-and-shaders/triangle.js

```ts
this.shader = pixi.Shader.from(vert, frag, uniforms)
this.quad = new pixi.Mesh(this.geometry, this.shader)
```

### Current Behavior

TypeScript complains that 

```text
TS2345: Argument of type 'Shader' is not assignable to parameter of type 'MeshMaterial'.
  Type 'Shader' is missing the following properties from type 'MeshMaterial': uvMatrix, batchable, pluginName, _tintRGB, and 7 more.
```

### Possible Solution

Seems like the type of the second argument should be relaxed. It currently accepts `MeshMaterial`, but a `Shader` works as well, at least according to the demo.

### Workaround

As usual, just do a cast.

```ts
this.quad = new pixi.Mesh(this.geometry, this.shader as MeshMaterial)
// or
this.quad = new pixi.Mesh(this.geometry, this.shader as any)
```

### Environment

- **`pixi.js` version**: 6.0.0-rc.2 (using separate packages, not `pixi.js`)
- **Browser & Version**: irrelevant
- **OS & Version**: irrelevant
- **Running Example**: can't create a proper reproduction anywhere because of TypeScript, I can create a repository if needed but it should be trivial to reproduce

bigtimebuddy:
Fix was merged for this. Should be in rc.3 now.

PR #7240, opened 2021-02-09
Fix types for Mesh

SerG-Y:
##### Description of change

Fix types for Mesh constructor #7229 
##### Pre-Merge Checklist

### DIFF
--- a/packages/mesh/src/Mesh.ts
+++ b/packages/mesh/src/Mesh.ts
@@ -34,10 +34,10 @@ export interface Mesh extends GlobalMixins.Mesh {}
  * @extends PIXI.Container
  * @memberof PIXI
  */
-export class Mesh extends Container
+export class Mesh<T extends Shader = MeshMaterial> extends Container
 {
     public readonly geometry: Geometry;
-    public shader: MeshMaterial;
+    public shader: T;
     public state: State;
     public drawMode: DRAW_MODES;
     public start: number;
@@ -62,7 +62,7 @@ export class Mesh extends Container
      *        if no state is provided, uses {@link PIXI.State.for2d} to create a 2D state for PixiJS.
      * @param {number} [drawMode=PIXI.DRAW_MODES.TRIANGLES] - the drawMode, can be any of the PIXI.DRAW_MODES consts
      */
-    constructor(geometry: Geometry, shader: Shader|MeshMaterial, state?: State, drawMode = DRAW_MODES.TRIANGLES)
+    constructor(geometry: Geometry, shader: T, state?: State, drawMode = DRAW_MODES.TRIANGLES)
     {
         super();
 
@@ -82,7 +82,7 @@ export class Mesh extends Container
          * Can be shared between multiple Mesh objects.
          * @member {PIXI.Shader|PIXI.MeshMaterial}
          */
-        this.shader = shader as MeshMaterial;
+        this.shader = shader;
 
         /**
          * Represents the WebGL state the Mesh required to render, excludes shader and geometry. E.g.,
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):

packages/mesh/src/Mesh.ts
```diff
@@ -82,7 +82,7 @@ export class Mesh extends Container
          * Can be shared between multiple Mesh objects.
          * @member {PIXI.Shader|PIXI.MeshMaterial}
          */
-        this.shader = shader;
+        this.shader = shader as MeshMaterial;
```
SukantPal, line 85: Okay, what's the point then :). It's better to change the type of `this.shader` and act accordingly.

SerG-Y: If change the type of `this.shader` to `Shader|MeshMaterial` we will need to cast in a few places then, because `Shader` doesn't have properties like `texture`, `alpha`, 'tint' etc.

SukantPal: Well, that's what we need to do. If `Shader` is a supported mesh shader, then we can't blindly assume it'll work?

SerG-Y: I think we can assume that it will work because for `Shader` or `MeshMaterial` with `program` we call `_renderDefault` that not require `texture`, `tint` etc.

SukantPal: Yeah, so wherever you require the shader to be `MeshMaterial` you need to cast it there.

bigtimebuddy: There's always `Mesh<Shader>` and `Mesh<MeshMaterial>` to prevent casting.

bigtimebuddy: I don't understand what you mean @SukantPal. Can you elaborate

SukantPal:
```ts
class Mesh<T extends Shader = MeshMaterial> {
  material: T
  render() {
    if (this.material instanceof MeshMaterial) {} // implicit cast into MeshMaterial
    else {}
  }
}
```

You'd still have to check if the material is `MeshMaterial` or not.

bigtimebuddy: `instanceof` is not a cast, it's a runtime type check. I think that's fine.

bigtimebuddy: Yeah, I think Generics is the way to go for this. What do you think @SerG-Y?

SerG-Y: I am not sure that it will work in this specific case, just an example we have getters and setters for `tint` and `texture` in `Mesh` class and with generic we will need cast it to `MeshMaterial` in all places. 
Just take a look at the following piece of code from `Mesh`. When we pass `Shader` and set `texture` it will not break anything because of JS but in TS we need cast it to `MeshMaterial` and in the code, we have few places where we should cast it to `MeshMaterial` when we will use Generics. Hope my thoughts are clear for you) What do you think guys?
```js
get texture(): Texture
{
    return this.shader.texture;
}

set texture(value: Texture)
{
    this.shader.texture = value;
}
```

SukantPal:
```ts
/** Null for non-MeshMaterial shaders. */
get texture(): Texture
{
  return 'texture' in this.shader ? this.shader.texture : null;
}
```

bigtimebuddy: These issues you're describing are mostly internal. What's more important here is that externally we handle this Mesh class better. I think generics solves that problem, but yeah we'll need to resolve the internal bits with a combo of casting and/or runtime checking. That's okay, IMO.

SerG-Y: Are you agree with the following type of casting?
```ts
const shader = this.shader as unknown as MeshMaterial;
```
it's not possible directly cast to `MeshMaterial` because of the next error. If you know a better way how to do it, please let me know.
image

bigtimebuddy: Sure, not ideal but that's fine.

bigtimebuddy:
I'm happy with these changes. @SukantPal thoughts?

SukantPal (approved):
LGTM!

bigtimebuddy (approved):

PR #7246, opened 2021-02-15
Add Generics to Texture, BaseTexture for Resource types

bigtimebuddy:
This adds an opt-in stricter type for Texture and BaseTexture using Generics.

Here are some uses cases:

### Better Return Values Using `from`
```js
// current
const texture = Texture.from('path/to/image.png');
const url = (texture.baseTexture.resource as ImageResource).url;

// new!
const texture = Texture.from<ImageResource>('path/to/image.png');
const url = texture.baseTexture.resource.url;
```

### More Casting Flexibility
```js
// current
const url = (sprite.texture.baseTexture.resource as ImageResource).url;

// new!
const texture = sprite.texture as Texture<ImageResource>;
const url = texture.baseTexture.resource.url;
```

### Stricter Resource Option Types

```js
// current
const base = BaseTexture.from("path/to/image.svg", {
  resourceOptions: { anything: 'goes' } // <-- NO TYPE CHECKING HERE
});

// new!
const base = BaseTexture.from<SVGResource, ISVGResourceOptions>("path/to/image.svg", {
  resourceOptions: { } // <-- ONLY VALID ISVGResourceOptions
});
```

### DIFF
--- a/packages/canvas/canvas-mesh/src/CanvasMeshRenderer.ts
+++ b/packages/canvas/canvas-mesh/src/CanvasMeshRenderer.ts
@@ -1,4 +1,4 @@
-import { Texture } from '@pixi/core/src';
+import { Texture } from '@pixi/core';
 import { DRAW_MODES } from '@pixi/constants';
 import { canvasUtils } from '@pixi/canvas-renderer';
 
--- a/packages/core/src/textures/BaseTexture.ts
+++ b/packages/core/src/textures/BaseTexture.ts
@@ -16,7 +16,7 @@ const defaultBufferOptions = {
 
 export type ImageSource = HTMLImageElement|HTMLCanvasElement|HTMLVideoElement|ImageBitmap;
 
-export interface IBaseTextureOptions<T = any> {
+export interface IBaseTextureOptions<RO = any> {
     alphaMode?: ALPHA_MODES;
     mipmap?: MIPMAP_MODES;
     anisotropicLevel?: number;
@@ -28,7 +28,7 @@ export interface IBaseTextureOptions<T = any> {
     type?: TYPES;
     target?: TARGETS;
     resolution?: number;
-    resourceOptions?: T;
+    resourceOptions?: RO;
     pixiIdPrefix?: string;
 }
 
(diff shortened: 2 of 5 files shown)


packages/canvas/canvas-mesh/src/CanvasMeshRenderer.ts
```diff
@@ -1,4 +1,4 @@
-import { Texture } from '@pixi/core/src';
+import { Texture } from '@pixi/core';
```
bigtimebuddy, line 1: I don't know how this was missed 👀

packages/mesh-extras/src/SimpleMesh.ts
```diff
@@ -24,7 +24,7 @@ export class SimpleMesh extends Mesh
      * @param {number} [drawMode] - the drawMode, can be any of the Mesh.DRAW_MODES consts
      */
     constructor(
-        texture = Texture.EMPTY,
+        texture: Texture = Texture.EMPTY,
```
bigtimebuddy, line 27: This change was required, because type inference was being used. Which, IMO, is more of a case for not inferring in situations like this (params).

packages/core/src/textures/BaseTexture.ts
```diff
@@ -42,7 +43,7 @@ export interface BaseTexture extends GlobalMixins.BaseTexture, EventEmitter {}
  * @extends PIXI.utils.EventEmitter
  * @memberof PIXI
```
SukantPal, line 44: webdoc doesn't have it currently, but type parameters will be documented with `@typeParam`. Let's add them right now!

```suggestion
 * @memberof PIXI
 * @typeParam T - The base-texture's resource.
 * @typeParam U - The resource parameters.
```

SukantPal (changes requested):

GoodBoyDigital (approved):
This is awesome! _I love generics_ 

Just throwing this out there, but how do you feel about using more descriptive generics rather than T? Like `RESOURCE_TYPE` make it much easier to quickly understand what the generic is and what it's for?

SukantPal:
@GoodBoyDigital I think `R` would suffice.

GoodBoyDigital:
I never quite got why the generic props are single letter. Always found it a bit strange as I feel it makes it harder to understand whats going on when looking at code. 

But having said that the typescript world tends not to agree with me! I guess the risk is the code ends up getting overly verbose and a bit too wordy! 

Potato patato :)

bigtimebuddy:
Thanks guys. 

* @SukantPal I added the docs.
* @GoodBoyDigital I renamed `T` to `R` for "Resource" and `U` to `RO` for "Resource Options"

SukantPal (approved):
Superb!

PR #7260, opened 2021-02-22
Convert the render functions to take options instead of args

bigtimebuddy:
This has been on my "hit-list" for awhile, but haven't gotten around to it. I find this one of the more troublesome examples in the PixiJS API of using arguments where it should be options. This function is probably one of the most consequential usage.

This will allow us to add more rendering options in the future, if we'd like.

### Problem

The old arguments weren't in a particularly useful order, so sometimes you could end up doing things like this:

```js
renderer.render(cont, undefined, undefined, new Matrix().translate(10, 20));
```

Introducing the arguments does this:

```js
renderer.render(cont, { transform: new Matrix().translate(10, 20) });
```

### Old API

```ts
render(
  container: PIXI.Container,
  renderTexture?: PIXI.RenderTexture,
  clear?: boolean,
  transform?: PIXI.Matrix,
  skipUpdateTransform?: boolean): void;
```
### New API

```ts
render(container: PIXI.Container, options?: {
  renderTexture?: PIXI.RenderTexture,
  clear?: boolean,
  transform?: PIXI.Matrix,
  skipUpdateTransform?: boolean
}): void;
```


GoodBoyDigital:
love this!

GoodBoyDigital (approved):

bigtimebuddy:
Per @andrewstart's suggestion, I removed the argument spread and default object object to avoid creating a lot of garbage every frame. Also, support the old syntax in TypeScript.

packages/core/src/Renderer.ts
```diff
@@ -376,14 +375,30 @@ export class Renderer extends AbstractRenderer
...
+     * @param {PIXI.RenderTexture} [options.renderTexture] - The render texture to render to.
+     * @param {boolean} [options.clear=true] - Should the canvas be cleared before the new render.
+     * @param {PIXI.Matrix} [options.transform] - A transform to apply to the render texture before rendering.
+     * @param {boolean} [options.skipUpdateTransform=false] - Should we skip the update transform pass?
      */
-    render(displayObject: IRenderableObject, renderTexture?: RenderTexture,
-        clear?: boolean, transform?: Matrix, skipUpdateTransform?: boolean): void
+    render(displayObject: IRenderableObject, options: IRendererRenderOptions = {}, ...rest:any[]): void
     {
+        if (options && 'baseTexture' in options)
```
SukantPal: It does seem a bit weird that we're assuming that having a "baseTexture" property means that `options` is a Texture, and therefore the legacy signature should be used. I would suggest using `instanceof Texture` instead. That means developers won't have to sanitize their options of a potential "baseTexture".

bigtimebuddy: Yeah, I just fixed this.

SukantPal (approved):
Perfecto!

PR #7262, opened 2021-02-24
Add cacheAsBitmapResolution override, default to Renderer's resolution

### LINKED ISSUE #7258: CacheAsBitmap on resolution > 1
`cacheAsBitmap` not apply resolution when generate RenderTexture
For devices with resolution > 1 this is produce a pixelation.

### Expected Behavior

CacheAsBItmap save pixel density for cached elements

### Current Behavior
`cacheAsBitmap` reduce resolution for texture 

### Possible Solution
Put resolution property from renderer to render texture
https://github.com/pixijs/pixi.js/blob/dev/packages/mixin-cache-as-bitmap/src/index.ts#L209

image

### Environment

- **`pixi.js` version**:  >5.0.0
- **Browser & Version**: all
- **Running Example**: https://www.pixiplayground.com/#/edit/9fx2o5h41xntmYvbVd-Ed

bigtimebuddy:
Fixes #7258 

### 🔥  Breaking Change

This changes the behavior of `cacheAsBitmap`, resolution is automatically inherited from the Renderer's resolution. And introduces a new property called `cacheAsBitmapResolution` which can be set to override the default resolution.

This is a change in behavior and would be considered a breaking change. Previously the only way to adjust the resolution was by using `PIXI.settings.RESOLUTION`.

```js
// ...
container.cacheAsBitmapResolution = 1;
container.cacheAsBitmap = true;
```

### Demo

https://jsfiddle.net/bigtimebuddy/ckt7njzf/


themoonrat (commented):
If cacheAsBitmap is already true.... should changing the cacheAsBitmapResolution cause it to be regenerated?

bigtimebuddy:
It could! I didn't do that, but it's an easy change.

bigtimebuddy:
@themoonrat done and updated demo

GoodBoyDigital (approved):
👍

themoonrat (approved):

packages/mixin-cache-as-bitmap/src/index.ts
```diff
@@ -52,6 +53,39 @@ export class CacheData
...
+            if (resolution === this._cacheAsBitmapResolution)
+            {
+                return;
+            }
+
+            this._cacheAsBitmapResolution = resolution;
+
+            if (this.cacheAsBitmap)
+            {
+                // Toggle to re-renderer at the new resolution
```
SukantPal, line 82:
```suggestion
                // Toggle to re-render at the new resolution
```
lol

SukantPal (approved):
LGTM w optional suggestion

packages/mixin-cache-as-bitmap/src/index.ts
bigtimebuddy: Haha, thanks!

PR #7268, opened 2021-03-03
Shader compiling optimization and better shader debugging

GoodBoyDigital:
One of those things that has been bugging me for a little while, finally had some time to adjust.

problem:
- we were creating and destroying a gl program each time a new Pixi Program was created as a means to extract the program data. We were then creating it _again_ when the shader was actually bound and used for the first time in the shader.
- this first program thats created was using a separate context via `getTestContext`. Mostly this works fine, until the main context uses any extensions not present on the test context. In this case, things tend to break!

solution
- As the attributeData an uniformData objects generated are only really used internally by pixi, I have tweaked this so that those two are generated when a shader is bound instead of when a shader is created. This means we can use the same program rather than creating and destroying one! 
- This will speed up overall pixi initialisation as only one program will ever be compiled once per program as opposed to two! wee!


bigtimebuddy (approved):

SukantPal (approved):
LGTM! Nice optimization.

GoodBoyDigital:
Hey Guys,

Tested this on android and it seems that removing the attribute Location binding was not a great idea :P

I have a fix, but need a little tidy before I update this PR. Only downside is we will need to link the program twice which is still a lot less than we currently do 👍 

watch this space..

GoodBoyDigital:
Hey team! If you could re-review this that would be ace! 

I fixed the android issue... but also spruced up our error logging for shaders too!

+ line numbers on the shaders when logged
+ highlighting the line with the error

image

ivanpopelyshev (approved):

packages/core/src/filters/Filter.ts
```diff
@@ -189,6 +189,11 @@ export class Filter extends Shader
     public resolution: number;
     public enabled: boolean;
     public autoFit: boolean;
+    /**
+     * Legacy filters use position and uvs from attributes (set by filter system)
+     * @member {boolean}
+     * @readonly
+     */
```
bigtimebuddy, line 196:
```suggestion
    /**
     * Legacy filters use position and uvs from attributes (set by filter system)
     * @readonly
     */
```

bigtimebuddy: member and type not needed for docs.

packages/core/src/geometry/GeometrySystem.ts
```diff
@@ -313,12 +313,18 @@ export class GeometrySystem implements ISystem
      * @param {PIXI.Program} program - Instance of program
      * @param {boolean} [incRefCount=false] - Increment refCount of all geometry buffers
      */
-    protected initGeometryVao(geometry: Geometry, program: Program, incRefCount = true): WebGLVertexArrayObject
+    protected initGeometryVao(geometry: Geometry, shader: Shader, incRefCount = true): WebGLVertexArrayObject
```
bigtimebuddy, line 316: Can up update the doc above this to reflect this change?

packages/core/src/shader/Program.ts
```diff
@@ -47,8 +42,15 @@ export class Program
     nameCache: any;
     glPrograms: { [ key: number ]: GLProgram};
     syncUniforms: any;
+    /**
+     * assigned when a program is first bound to the shader system
+     */
```
bigtimebuddy, line 47: More terser comment works too!
```suggestion
    /** Assigned when a program is first bound to the shader system. */
```

packages/core/src/shader/Program.ts
```diff
@@ -47,8 +42,15 @@ export class Program
...
     attributeData: { [key: string]: IAttributeData};
+    /**
+     * assigned when a program is first bound to the shader system
+     */
```
bigtimebuddy, line 51:
```suggestion
    /** Assigned when a program is first bound to the shader system. */
```

packages/core/src/shader/utils/getAttributeData.ts
```diff
@@ -0,0 +1,38 @@
...
+    const attributes: {[key: string]: IAttributeData} = {};
+
+    const totalAttributes = gl.getProgramParameter(program, gl.ACTIVE_ATTRIBUTES);
+
+    for (let i = 0; i < totalAttributes; i++)
+    {
+        const attribData = gl.getActiveAttrib(program, i);
+        const type = mapType(gl, attribData.type);
+
+        /*eslint-disable */
```
bigtimebuddy, line 25: Why is linting disabled here?

packages/core/src/shader/utils/getUniformData.ts
```diff
@@ -0,0 +1,38 @@
+import { IUniformData } from '../Program';
+import { defaultValue } from './defaultValue';
+import { mapType } from './mapType';
+
+/**
+ * returns the uniform data from the program
+ * @private
+ *
+ * @param {webGL-program} [program] - the webgl program
+ * @param {context} [gl] - the WebGL context
```
bigtimebuddy, line 10: Simpler works too
```suggestion
 * @param program - the webgl program
 * @param gl - the WebGL context
```

packages/core/src/shader/utils/logProgramError.ts
```diff
@@ -0,0 +1,91 @@
+/**
+ * will log a shader error highlighting the lines with the error
+ * also will add numbers along the side.
+ *
+ * @param gl - the WebGLContext
+ * @param shader - the shader to log errors for
+ */
+function logPrettyShaderError(gl:WebGLRenderingContext, shader:WebGLShader):void
```
bigtimebuddy, line 8: Recommend wrapping the contents of this function in the following:
```js
// #if _DEBUG

// #endif
```
This would exclude all this work in the min version. maybe consider a simpler version for min version:

```js
// #if _DEBUG
// ... debugging stuff goes here
// #else
console.error('Shader program error');
// #endif
```

packages/core/src/shader/utils/logProgramError.ts
```diff
@@ -0,0 +1,91 @@
...
+        {
+            logPrettyShaderError(gl, fragmentShader);
+        }
+
+        console.error('Pixi.js Error: Could not initialize shader.');
+
+        // if there is a program info log, log it
+        if (gl.getProgramInfoLog(program) !== '')
+        {
+            console.warn('Pixi.js Warning: gl.getProgramInfoLog()', gl.getProgramInfoLog(program));
```
bigtimebuddy, line 88: 🤔 
```suggestion
            console.warn('PixiJS Warning: gl.getProgramInfoLog()', gl.getProgramInfoLog(program));
```

packages/core/src/geometry/GeometrySystem.ts
GoodBoyDigital: yessir

packages/core/src/shader/utils/getAttributeData.ts
GoodBoyDigital: No reason as far as i can see! Removing..

packages/core/src/shader/utils/logProgramError.ts
GoodBoyDigital: Yo! I think this is important to have even in the min version - as it's critical when working on shaders of any kind in pixi not just the internal ones iyswim?

bigtimebuddy: Okay

packages/core/src/shader/utils/logProgramError.ts
```diff
@@ -0,0 +1,91 @@
...
+        {
+            logPrettyShaderError(gl, vertexShader);
+        }
+
+        if (!gl.getShaderParameter(fragmentShader, gl.COMPILE_STATUS))
+        {
+            logPrettyShaderError(gl, fragmentShader);
+        }
+
+        console.error('Pixi.js Error: Could not initialize shader.');
```
bigtimebuddy, line 83:
```suggestion
        console.error('PixiJS Error: Could not initialize shader.');
```

bigtimebuddy:
Thanks @GoodBoyDigital :tada:

PR #7269, opened 2021-03-03
Remove System inheritance

bigtimebuddy:
The System inheritance is not useful. This change paves the way for #7247 by converting System into an interface. Specifically, I want to make InteractionManager a System, but the inheritance with EventEmitter wouldn't work with that class. This PR supports that and removes an unnecessary layer.


packages/core/src/mask/AbstractMaskSystem.ts
```diff
@@ -10,17 +9,15 @@ import type { Renderer } from '../Renderer';
...
-export class AbstractMaskSystem extends System
+export class AbstractMaskSystem implements ISystem
 {
     protected maskStack: Array<MaskData>;
     protected glConst: number;
     /**
      * @param {PIXI.Renderer} renderer - The renderer this System works for.
      */
-    constructor(renderer: Renderer)
+    constructor(protected renderer: Renderer)
```
GoodBoyDigital, line 19: curious why you have added protected / private in the constructor params?

GoodBoyDigital (approved):
lgtm!

packages/core/src/mask/AbstractMaskSystem.ts
bigtimebuddy: That's a shortcut in TypeScript. These two things are equivalent:

```typescript
constructor(private renderer:Renderer) {}
```

```typescript
private renderer:Renderer;
constructor(renderer:Renderer) {
  this.renderer = renderer;
}
```

GoodBoyDigital: nice everyday's a school day!

ivanpopelyshev (approved):

PR #7271, opened 2021-03-04
Fix missing removeEventListener in ResizePlugin

### LINKED ISSUE #7270: resizeTo doesn't clean up resize event listener on application destroy
### Expected Behavior

Resize event listener should be removed when the application is destroyed

### Current Behavior

The resize events are not removed, see proofs of Firefox and Chrome dev tools:

image

image

### Possible Solution

Manually call removeEventListener in the destroy callback of the ResizePlugin

### Steps to Reproduce

Using the fiddle linked below, create and destroy the application several times and inspect the root element to find old event handlers

### Environment

- **`pixi.js` version**:  5.3.8 (probably not only)
- **Browser & Version**: all
- **OS & Version**: all
- **Running Example**: https://jsfiddle.net/bepg08wt/

epaezrubio:
Fixes #7270

### DIFF
--- a/packages/app/src/ResizePlugin.ts
+++ b/packages/app/src/ResizePlugin.ts
@@ -140,6 +140,7 @@ export class ResizePlugin
      */
     static destroy(): void
     {
+        self.removeEventListener('resize', this.queueResize);
         this.cancelResize();
         this.cancelResize = null;
         this.queueResize = null;


bigtimebuddy (approved):
Good catch :+1:

themoonrat (approved):

bigtimebuddy:
Thank you @epaezrubio, i'll working on cherry picking this for 5.3.x as well.

PR #7273, opened 2021-03-04
Unbind integer textures

ivanpopelyshev:
All sample types in shaders can have `u` or `i` prefix: `usampler2D`, `isamplerCube`

I added those types in `SAMPLER_TYPES`. To ensure that batcher doesn't use sampler of wrong type, and drawcall doesnt crash, TextureSystem unbinds those kind of textures.

Problem: https://www.html5gamedevs.com/topic/46983-mesh-graphics-render-issue/
Spawns error in console: https://www.pixiplayground.com/#/edit/xpLx-ElnpPPPXFJUkxoRa
Doesnt spawn errors: https://www.pixiplayground.com/#/edit/xpLx-ElnpPPPXFJUkxoRa

This PR conflicts with #7222 in `afterEach` thingy


packages/core/src/textures/GLTexture.ts
```diff
@@ -70,5 +71,10 @@ export class GLTexture
          * @member {number}
          */
         this.internalFormat = 5121;
+
+        /**
+         * @member {number}
+         */
```
bigtimebuddy, line 77: Minor comment, this is not necessary, if you add a documentation block above `public samplerType: number` see #7159

packages/core/src/textures/TextureSystem.ts
```diff
@@ -69,6 +70,11 @@ export class TextureSystem extends System
          * @readonly
          */
         this.unknownTexture = new BaseTexture();
+
+        /**
+         * Whether glTexture with int/uint sampler type was uploaded
+         */
```
bigtimebuddy, line 76: Same thing, can you add this to the top instead of in the constructor? No reason to maintain this legacy format for new members.

bigtimebuddy (approved):
I'll defer to someone else (@GoodBoyDigital) for the functionality here, but code quality looks fine. Just a few minor comments.

And welcome back @ivanpopelyshev glad to see PRs from you again.

packages/core/src/textures/GLTexture.ts
ivanpopelyshev: done, can you check please?

packages/core/src/textures/TextureSystem.ts
```diff
@@ -19,6 +19,12 @@ export class TextureSystem extends System
 {
     public boundTextures: BaseTexture[];
     public managedTextures: Array<BaseTexture>;
+    /**
+     * Whether glTexture with int/uint sampler type was uploaded
+     * @member {boolean}
+     * @protected
+     */
+    protected hasIntegerTextures: boolean;
```
SukantPal, line 27: We don't need to add `@member` and `@protected` in the docs for this. webdoc will infer.

packages/core/src/textures/GLTexture.ts
```diff
@@ -70,5 +76,7 @@ export class GLTexture
          * @member {number}
          */
         this.internalFormat = 5121;
+
+        this.samplerType = 0;
```
SukantPal, line 80: I'm not sure why this is always `SAMPLER_TYPES.FLOAT`? Doesn't it depend on the internal format of the texture (if we should derive it from that?).

SukantPal (changes requested):

packages/core/src/textures/GLTexture.ts
ivanpopelyshev: This feature only for WebGL2 and it requires custom texture resource, we delegated all that stuff to user. However, user might not know how pixi batcher works,

SukantPal (approved):
This is a good addition! We should add it to PixiJS Examples.

GoodBoyDigital (approved):
nice!

packages/core/src/textures/GLTexture.ts
```diff
@@ -12,6 +12,12 @@ export class GLTexture
     public wrapMode: number;
     public type: number;
     public internalFormat: number;
+    /**
+     * Type of sampler corresponding to this texture. See {@link PIXI.SAMPLER_TYPES}
+     * @member {number}
+     */
```
bigtimebuddy, line 18:
```suggestion
    /** Type of sampler corresponding to this texture. See {@link PIXI.SAMPLER_TYPES} */
```

packages/core/src/textures/TextureSystem.ts
```diff
@@ -19,6 +19,10 @@ export class TextureSystem implements ISystem
 {
     public boundTextures: BaseTexture[];
     public managedTextures: Array<BaseTexture>;
+    /**
+     * Whether glTexture with int/uint sampler type was uploaded
+     */
```
bigtimebuddy, line 24:
```suggestion
    /** Whether glTexture with int/uint sampler type was uploaded. */
```

packages/core/src/textures/TextureSystem.ts
```diff
@@ -235,6 +242,37 @@ export class TextureSystem implements ISystem
         }
     }
 
+    /**
+     * Ensures that current boundTextures all have FLOAT sampler type,
+     * see {@link PIXI.SAMPLER_TYPES} for explanation.
+     *
+     * @param {number} maxTextures - number of locations to check
```
bigtimebuddy, line 249:
```suggestion
     * @param maxTextures - number of locations to check
```

blakjak44:
Hey guys, seems like something from @ivanpopelyshev's fix is missing. Using v6.0.2 I am still unable to render integer textures along with float textures.

v6.0.2: https://www.pixiplayground.com/#/edit/hMVY0qgOa6lLoNTi_PnEX <-- Not working
@ivanpopelyshev custom build: https://www.pixiplayground.com/#/edit/xpLx-ElnpPPPXFJUkxoRa <-- Works

bigtimebuddy:
That's my fault. Looks like it never actually got cherry-picked to 6.0.2. Will make sure it's 6.0.3.

PR #7277, opened 2021-03-06
high shader

GoodBoyDigital:
yoo! here is a starting point for a shader preprocessor! This is almost directly lifted from our internal game engine as its worked quite well for us and is abstract enough that will easily work with any kind of shaders! 

At its most simplest, it's a templating system that injects code into the shader as is finds tags. 

## Features
- all code injected in are in the form of a 'HighFragment' as TS structure that makes it easy to compile in 
- allows for separate WebGL2 code if required
- header compiler is magic - you can add extensions and they will all be hoisted along with the removal of any duplicate uniforms that may be shared by fragments
- support for uniform buffer objects _and_ non uniform buffer objects
- dynamic headers for WebGL2 that enable it to be backwards compatible with WebGL code
- ability to inject precision and a shader name
- ability to add defines
- a code formatter, so shader looks nice (also makes it easy to test!
- lots of caching layers, so it super fast!
- some tests to make sure it all works!

## Usage 
```
import { HighFragment } from '../HighFragment';

/**
 * bog standard way to set the diffuse color using a texture
 */
export const diffuseStandard:HighFragment = {
    fragment: {
        header: `
            uniform sampler2D uDiffuseMap;
        `,
        main: `
            diffuseColor *= texture2D(uDiffuseMap, uv);
        `,
    },
};

```

these fragments are then passed to our shiny new high shader compiler like so:

```
const fragments = [
    diffuseStandard
]

const phongShaderTemplate = {
    vertex:`
         attribute vec3 aPosition;
         
        void main() {
              gl_Position = vec4(aPosition, 0.);
        }
    `,
    fragment:`
      
        void main() {
              vec4 diffuseColor;

              {{MAIN}}

             gl_FragColor = diffuseColor;
        }
    `
}
const { vertex, fragment } = compileHighShader(isWebGL2, phongShaderTemplate, highFragments);

const shader = Shader.from(vertex, fragment);
```

## Concepts

The idea is that a high fragment does one thing only and we can then reuse them for various things and shaders.
BatchFragments! Standard pixi templates! should give us loads of flexibility! Additionally we can start to leverage UBOs without adding additional complexity to the code base! 

Open to feedback and questions if you guys have any!


bigtimebuddy:
Compile error @GoodBoyDigital:
```
Error: packages/core/src/shader/high-shader/compiler/index.ts(4,15): error TS2307: Cannot find module 
'./convertHeaderToUbo' or its corresponding type declarations.
```

PR #7286, opened 2021-03-10
Remove utils dependency to Node types

### LINKED ISSUE #7283: TypeScript Node types "leak" into a project using PIXI with v6
### Expected Behavior

Adding PIXI to a TypeScript project should not add Node types to the project.

### Current Behavior

Adding PIXI to a TypeScript project adds a dependency to Node types.

For example this project compiles successfully:

**tsconfig.json**
```json
{
  "compilerOptions": {
    "lib": [
      "dom",
      "esnext",
      "scripthost"
    ]
  },
  "include": [
    "src"
  ]
}
```

**src/index.ts**
```ts
const timeout: number = setTimeout(() => {
    console.log('hello world');
}, 100);
```

However after just adding `pixi.js` to the project without modifying the code at all the `setTimeout` return value is no longer assignable to a `number` as it's type changes to `NodeJS.Timeout` and compiling gives the following error:

```
src/index.ts:1:7 - error TS2322: Type 'Timeout' is not assignable to type 'number'.

1 const timeout: number = setTimeout(() => {
        ~~~~~~~
```

### Possible Solution

Remove the type dependency to Node. This is most likely caused by a Node type reference in one of PIXI's dependencies rather than PIXI itself. I haven't yet investigated this in more detail so I can't say for sure what is causing this and how to fix it.

### Steps to Reproduce

Run the following commands to create the empty project and copy `tsconfig.json` and `src/index.ts` from the example above:
```
yarn init
yarn add typescript
yarn add pixi.js
```
After the project is setup run `yarn tsc --noEmit` to get the error.

### Environment

- **TypeScript version**: 4.2.3

qtiki:
##### Description of change

Moved the `utils` Node types dependency to a dev dependency. Had to redeclare types for Node `url` and `querystring` packages for the `utils.url` export.

Fixes #7283

##### Pre-Merge Checklist

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -34,7 +34,6 @@
     "url": "^0.11.0"
   },
   "devDependencies": {
-    "@types/node": "^12.0.0",
     "css-color-names": "^1.0.1"
   }
 }
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -29,12 +29,12 @@
     "@pixi/constants": "6.0.0",
     "@pixi/settings": "6.0.0",
     "@types/earcut": "^2.1.0",
-    "@types/node": "^12.0.0",
     "earcut": "^2.2.2",
     "eventemitter3": "^3.1.0",
     "url": "^0.11.0"
   },
   "devDependencies": {
+    "@types/node": "^12.0.0",
     "css-color-names": "^1.0.1"
   }
 }
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -74,7 +74,7 @@ export { EventEmitter };
  */
 export { default as earcut } from 'earcut';
 
-import { parse, format, resolve } from 'url';
+import { parse, format, resolve } from './url';
 
 /**
  * Node.js compatible URL utilities.
--- a/packages/utils/src/network/determineCrossOrigin.ts
+++ b/packages/utils/src/network/determineCrossOrigin.ts
@@ -1,4 +1,4 @@
-import * as _url from 'url';
+import * as _url from '../url';
 
 let tempAnchor: HTMLAnchorElement|undefined;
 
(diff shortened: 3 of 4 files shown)


packages/utils/package.json
```diff
@@ -29,12 +29,12 @@
     "@pixi/constants": "6.0.0",
     "@pixi/settings": "6.0.0",
     "@types/earcut": "^2.1.0",
-    "@types/node": "^12.0.0",
     "earcut": "^2.2.2",
     "eventemitter3": "^3.1.0",
     "url": "^0.11.0"
   },
   "devDependencies": {
+    "@types/node": "^12.0.0",
```
bigtimebuddy, line 37: Do we even need this anymore?

qtiki: They provide some sanity checking for the assignment of `_xxx` functions to the redeclared `xxx` ones  in the url wrapper file. You can try for example to change the `ResolveFunction` return type to `number` and the compiler will complain about `parse = _parse` as the types are incompatible. Removing the Node types completely would also remove this type check.

qtiki: I guess it's a choice between having those wrappers type checked and having the Node types leaking all over the `utils` package or removing type checking from the wrappers and fixing the Node types leak. It could actually be that removing the Node types would have more pros - It's not like those types will be changed often.

mw-developers:
quicksave-interactive:dev

packages/utils/package.json
bigtimebuddy: Yes exactly. We are just using a very narrow scope of those APIs. I'm for removing node types for less complexity and just casting url internally for more strict types.

qtiki: Agreed. I pushed a commit that removes the Node types completely.

bigtimebuddy (approved):

SukantPal (approved):
🙏

ivanpopelyshev (approved):

bigtimebuddy:
Thanks you @qtiki

PR #7295, opened 2021-03-15
type: Loader.resource is a list of ILoaderResource

AdrienLemaire:
##### Description of change

The current implementation returns errors when trying to get a texture from a resource
```
[tsserver 2339] [E] Property 'texture' does not exist on type 'Resource | {}'
```
After reading the code, I found that pixi.js extends the resource-loader Resource type with the desired properties
```ts
export declare interface ILoaderResource extends GlobalMixins.ILoaderResource, Resource {
    texture?: Texture;
    spritesheet?: Spritesheet;
    textures?: Dict<Texture>;
    metadata: IResourceMetadata;
}
```

##### Pre-Merge Checklist

### DIFF
--- a/packages/loaders/src/Loader.ts
+++ b/packages/loaders/src/Loader.ts
@@ -159,12 +159,11 @@ export class Loader extends ResourceLoader
     /**
      * All the resources for this loader keyed by name.
      *
-     * @member {object<string, Resource>}
+     * @member {object<string, ILoaderResource>}
      */
     resources: {
         [key: string]: import('@pixi/loaders').ILoaderResource;
     };
-
 }
 
 // parse any blob into more usable objects (e.g. Image)
--- a/packages/loaders/src/Loader.ts
+++ b/packages/loaders/src/Loader.ts
@@ -155,6 +155,16 @@ export class Loader extends ResourceLoader
 
         return Loader;
     }
+
+    /**
+     * All the resources for this loader keyed by name.
+     *
+     * @member {object<string, Resource>}
+     */
+    resources: {
+        [key: string]: import('@pixi/loaders').ILoaderResource;
+    };
+
 }
 
 // parse any blob into more usable objects (e.g. Image)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,17614 @@
 {
 	"name": "pixi.js-monorepo",
+	"lockfileVersion": 2,
 	"requires": true,
-	"lockfileVersion": 1,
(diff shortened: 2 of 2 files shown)


packages/loaders/src/Loader.ts
```diff
@@ -155,6 +155,15 @@ export class Loader extends ResourceLoader
...
         return Loader;
     }
+
+    /**
+     * All the resources for this loader keyed by name.
+     *
+     * @member {object<string, ILoaderResource>}
+     */
+    resources: {
+        [key: string]: import('@pixi/loaders').ILoaderResource;
```
ivanpopelyshev, line 165: Referencing ILoaderResource from module itself, are we sure just `ILoaderResource` wont work?

ivanpopelyshev (changes requested):
Please remove package.json from this commit, its not your problem  to handle ;)

packages/loaders/src/Loader.ts
bigtimebuddy: Please don't use this import syntax here. Import type at the start of the file to import the interface. 

I thought that resources was a property of resource-loader. Why is it not inherited?

SukantPal:
@AdrienLemaire We don't commit scoped package-specific `package-lock.json` files into the repository.

packages/loaders/src/Loader.ts
```diff
@@ -155,6 +155,15 @@ export class Loader extends ResourceLoader
 
         return Loader;
     }
+
+    /**
+     * All the resources for this loader keyed by name.
+     *
+     * @member {object<string, ILoaderResource>}
```
SukantPal, line 162: I know `PIXI.ILoaderResource` isn't documented, but all of the API is under `PIXI`. Maybe we should add docs to `PIXI.ILoaderResource` @bigtimebuddy?

```suggestion
     * @member {object<string, PIXI.ILoaderResource>}
```

SukantPal (changes requested):

bigtimebuddy:
Thanks for pointing this type error, @AdrienLemaire. Because we had to make some other fixes, I'm closing this in favor of #7300. I will credit you in the commit though.

AdrienLemaire:
My pleasure.
@ivanpopelyshev for the package-lock.json, it would be helpful to add `package-lock.json` instead of `packages/**/package-lock.json` the gitignore
https://github.com/pixijs/pixi.js/blob/dev/.gitignore#L1

PR #7297, opened 2021-03-15
Remove Inch CI support for docs coverage

bigtimebuddy:
Inch CI does not support TypeScript so this hasn't been working for awhile since we converted the source code. This removes integration with this too as well as the README badge.


Zyie (approved):

SukantPal (approved):
Always good to prune out unused services

PR #7300, opened 2021-03-16
Remove spritesheet dependency from loader, fix resources type

bigtimebuddy:
Replaces #7295

* Remove the `@pixi/spritesheet` dependency from `@pixi/loaders`
* Fix the `resources` type for Loader to be more strict (e.g., contains `texture`) credit: @AdrienLemaire
* Added some additional documentation for Loader interfaces

### DIFF
--- a/packages/loaders/src/Loader.ts
+++ b/packages/loaders/src/Loader.ts
@@ -2,6 +2,7 @@ import { Loader as ResourceLoader, middleware } from 'resource-loader';
 import { TextureLoader } from './TextureLoader';
 
 import type { Resource } from 'resource-loader';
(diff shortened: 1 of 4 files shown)


Zyie (approved):

ivanpopelyshev (approved):

SukantPal:
webdoc doesn’t actually read declaration (global.d.ts) files so I’d be careful placing documentation there.

bigtimebuddy:
Understood on webdoc @SukantPal With these mixins we will probably need to add both global docs (for TS tooltips) and src docs (for webdoc).

PR #7318, opened 2021-03-19
Support running GitHub Actions on Forks and other workflow cleanup (take 2)

bigtimebuddy:
## Overview

Made these changes to the workflow so I could iterate more freely on some of the tooling related to unit-tests and I thought it would be valuable for other users too. 

### Changed

* Allows running GitHub Actions on forks by checking for and excluding steps that require secrets (e.g., codecov, s3 deployment)
* Cleaned up common env variables to be at the top of the workflow
* Externalize AWS_REGION in case someone wants to implement their own deployment on a fork
* Removes Node.js matrix, as we only really need to practically test on a single Node version

### Example

* [Fork Actions](https://github.com/bigtimebuddy/pixi.js/runs/2148901897?check_suite_focus=true) (ignores Codecov & S3)
* [Branch Actions](https://github.com/pixijs/pixi.js/runs/2148997163?check_suite_focus=true) (runs Codecov & S3)

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -12,19 +12,17 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [ 12.x ]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
+    - name: Use Node.js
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Ignore the pending "build (12.x)", I changed the name of the task and that's expected.

ivanpopelyshev (approved):

SukantPal (approved):

PR #7322, opened 2021-03-21
feat(DisplayObject): Add `destroyed` event

ayamada:
##### Description of change

Add `destroyed` event to `DisplayObject.destroy()`.
I need it sometimes.
(`removed` event is almost insufficient.)

##### Pre-Merge Checklist

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -444,6 +444,13 @@ export abstract class DisplayObject extends EventEmitter
      */
 
     /**
+     * Fired when this DisplayObject is destroyed.
+     *
+     * @instance
+     * @event destroyed
+     */
+
+    /**
      * Recalculates the bounds of the display object.
      */
     abstract calculateBounds(): void;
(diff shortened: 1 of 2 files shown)


ivanpopelyshev (approved):
Why not?

bigtimebuddy:
Overall, this is helpful. Good suggestion.

Could you add two tests instead of overloading a single test for removed and destroyed? These should probably be separate tests. 

Also, I would recommend using sinon.spy() instead of the local variable in the test. Makes it easier to test for called events this way.

bigtimebuddy (approved):
Thanks for the fixes!

ayamada:
Thanks!

> Why not?

Because, I want to destroy DisplayObject with related objects that made by me, but `removed` event have possibilities to call more than once. I want to call it only once.

> Could you add two tests instead of overloading a single test for removed and destroyed? These should probably be separate tests.
>
> Also, I would recommend using sinon.spy() instead of the local variable in the test. Makes it easier to test for called events this way.

I fixed it, and pushed.

bigtimebuddy:
Thank you @ayamada :tada:

PR #7327, opened 2021-03-23
Fix BitmapFont destroying respects external Textures

### LINKED ISSUE #7323: BitmapFont.uninstall/destroy also destroying the texture that could be used by other objects
When we use a texture to create a BitmapFont, uninstalling this font will also destroy the texture (and the base texture it seems) - even if the BitmapFont was not created using `BitmapFont.from`. This can create a problem as the texture could be used for something else (another BitmapFont? or even a Sprite?) and will be invalidated (and will crash the game).

### Expected Behavior

Assume we follow this pseudo code:
```
// Load beforehand someFontData and someTexture.

const bitmapFont = BitmapFont.install(someFontData, someTexture);

// Use bitmapFont with a BitmapText: 
const bitmapText = new BitmapText("some text", { fontName: bitmapFont.font });
// Use bitmapText somehow...
// We can destroy the bitmapText when not useful anymore:
bitmapText.destroy(); // So far so good.

BitmapFont.uninstall(bitmapFont.font);
```

At this point, I expect to be able to continue using someTexture. 

### Current Behavior

The call to `uninstall` will actually (in 5.3.7 at least) trigger a `destroy` on someTexture and its base texture.
If the texture is used by another object, this will crash at the texture is not `valid` anymore.

### Possible Solution

@SukantPal talked about a way to mark the texture `managed` by the BitmapFont or not (should be the case for `BitmapFont.from`). If not, it should probably not attempt to destroy its texture, apart if `true` is passed as an argument (following the usual convention for destroy options).

Not sure if it's the best solution (is `managed` enough? Is there a risk that it would require a full "reference counting" implementation in a more complex situation?), but mentioning it because it still sounds like a good start.

### Steps to Reproduce

I've not created an example but I could do it if it's useful. Creating a BitmapFont with a texture and `uninstall`ing it should show the problem, as the texture is not `valid` anymore.

### Environment

- **`pixi.js` version**: 5.3.7
- **Browser & Version**: Chrome 88.0.4324.96
- **OS & Version**: macOS
- **Running Example**: none (can be created on demand)

bigtimebuddy:
Fixes #7323

Unchanged behavior: BitmapFont page Textures are automatically destroyed when:
* ... loaded with `BitmapFontLoader`
* ... created with `BitmapFont.from`

New behavior: BitmapFont page Textures are _preserved_ when:
* ... manually created with `BitmapFont.install(fontData, texture)`

This is consistent with how PixiJS respects objects that are passed externally versus generated internally. Another example of this is Graphic and GraphicsGeometry. If passed to the constructor, Graphics will ignore destroying its geometry.

### DIFF
--- a/packages/text-bitmap/src/BitmapFont.ts
+++ b/packages/text-bitmap/src/BitmapFont.ts
@@ -107,19 +107,23 @@ export class BitmapFont
     public readonly lineHeight: number;
     public readonly chars: Dict<IBitmapFontCharacter>;
     public readonly pageTextures: Dict<Texture>;
-    private _generated = false;
+    private _manageTextures: boolean;
 
     /**
      * @param {PIXI.BitmapFontData} data
      * @param {PIXI.Texture[]|Object.<string, PIXI.Texture>} textures
+     * @param {boolean} manageTextures - Setting to `true` will destroy page textures
+     *        when the font is uninstalled.
      */
-    constructor(data: BitmapFontData, textures: Texture[]|Dict<Texture>)
+    constructor(data: BitmapFontData, textures: Texture[]|Dict<Texture>, manageTextures?: boolean)
     {
         const [info] = data.info;
         const [common] = data.common;
         const [page] = data.page;
         const res = getResolutionOfUrl(page.file);
         const pageTextures: Dict<Texture> = {};
+
+        this._manageTextures = manageTextures;
 
         /**
          * The name of the font face.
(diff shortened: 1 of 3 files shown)


SukantPal (approved):

ivanpopelyshev (approved):

ivanpopelyshev:
We have variable names `managedTextures` in our texture system, but that's ok

bigtimebuddy:
Good suggestion @ivanpopelyshev renamed to `ownsTextures`

PR #7332, opened 2021-03-24
Upgrade @pixi-build-tools/api-extractor-lerna-monorepo to 1.0.3

SukantPal:
This new version will exit with a non-zero code immediately when
@microsoft/api-extractor fails on a specific package. This ensures our
CI doesn't pass when a scoped package's declaration file fails to be
generated.


SukantPal:
@bigtimebuddy Is the package-lock.json diff too much?

bigtimebuddy:
Woah, why is that package lock diff so crazy?

SukantPal:
Not sure, I did “npm ci” before installing the new package. Possibly that?

On Mar 23, 2021, at 11:13 PM, Matt Karl ***@***.******@***.***>> wrote:



Woah, why is that package lock diff so crazy?

bigtimebuddy:
Closing in favor of #7333

PR #7338, opened 2021-03-25
Fix Internal Format Constants for Compressed Textures

### LINKED ISSUE #7330: Black RGBA8_ETC2 compressed textures on device due to wrong magic values
Again: Love for compressed textures!

As of version 6.0.0 ETC2 RGBA textures sadly are all black on device.

There seems to be a typo in https://github.com/pixijs/pixi.js/blob/dev/packages/compressed-textures/src/const.ts#L54-L58
For example `COMPRESSED_RGBA8_ETC2_EAC = 0x9275,` 
whereas the spec says `const GLenum COMPRESSED_RGBA8_ETC2_EAC  = 0x9278;`
(see https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_etc/)
The five formats from RGBA8_ETC2_EAC down all have wrong codes.

Additionally `INTERNAL_FORMAT_TO_BYTES_PER_PIXEL` wrongly specifies 0.5bpp for `RGBA8_ETC2_EAC` and `SRGB8_ALPHA8_ETC2_EAC`. These should be 1.

I tried monkey-patching these values in a compiled version of Pixi and ran it on device. I can confirm that changing the numbers to the correct values fixes the textures showing up as black.

I have ETC-capable devices at hand, so let me know if I can help you test a proper fix.

### Environment

- **Browser & Version**: Mobile Safari
- **OS & Version**: iOS 14.4.1

(also happens on a Moto One Action on Android 10 using Chrome)

bigtimebuddy:
Fixes #7330

Reference:
https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_etc/


hdemmer (approved):
Looks good! 👍

ivanpopelyshev (approved):

SukantPal (approved):

PR #7354, opened 2021-03-30
Populate the Loader resource with Texture objects for basis and compressed-textures

### LINKED ISSUE #7329: KTXLoader quirks and ASTC support
I'm delighted to see built-in compressed texture support in 6.0.0, big thanks!

Using KTXLoader I noticed that:
* After load the `resource.texture` field is not populated which breaks atlases (but is fixable with loader middleware)
* mipmaps are not being used, even if they are present. I suppose this line is at the root of this: https://github.com/pixijs/pixi.js/blob/dev/packages/compressed-textures/src/loaders/registerCompressedTextures.ts#L25

MipMaps were lost on BASIS as well, but I suppose that's a different issue. (Sadly the transcoder is too heavy for us)

Additionally, I'd love to see ASTC support, since it's quite widely available and has good tooling around it.

Let me know if there's anything I can help you with.

### Environment

bigtimebuddy:
Partially fixes #7329

### Changes

* The Loader's resource was not getting populated with the Texture being created, the only way to fetch this was by using the TextureCache.
* Fixes an issue where the first compressed texture resource was creating two Textures
* Properly handle errors internally and passed them to the loader instead of throwing globally
* Supports passing options to the BaseTexture constructor using LoaderResource's metadata
* Fixes baseUrl or path for loading manifest of compressed textures

### DIFF
--- a/packages/basis/src/BasisLoader.ts
+++ b/packages/basis/src/BasisLoader.ts
@@ -160,8 +160,9 @@ export class BasisLoader
         // Should be a valid TYPES, FORMATS for uncompressed basis formats
         const type: TYPES = BASIS_FORMAT_TO_TYPE[resources.basisFormat];
         const format: FORMATS = resources.basisFormat !== BASIS_FORMATS.cTFRGBA32 ? FORMATS.RGB : FORMATS.RGBA;
+        const resourceList = resources as Array<CompressedTextureResource | BufferResource>;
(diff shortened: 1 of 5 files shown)


SukantPal (approved):
LGTM

GoodBoyDigital (approved):

PR #7357, opened 2021-03-30
Remove Lerna bootstrap, replace with npm v7 workspaces

bigtimebuddy:
This PR upgrades the build process to use npm 7 to support workspaces. This allows PixiJS to remove a dependency on using the `bootstrap` functionality of Lerna which makes the `npm install` process smoother, faster and less complicated. Also, workspaces in npm has broader support and supposedly better cross-platform integration.

## Verification

✅   [Successful build](https://github.com/bigtimebuddy/pixi.js/runs/2231908500?)


SukantPal (approved):
Diff compact!

GoodBoyDigital (approved):
nice!

PR #7361, opened 2021-03-31
Types support nullable filters

### LINKED ISSUE #7359: Docs: v6 `filters` no longer nullable, but docs imply it is
- **`pixi.js` version**: 6.0.1

Hi, hopefully this isn't too pendantic, but I noticed a small inaccuracy in the docs.

https://pixijs.download/v6.0.0/docs/PIXI.DisplayObject.html#filters

> To remove filters simply set this property to 'null'.

The property is no longer recognized as `nullable` in TypeScript or your docs generator, and it seems that there's no longer any need to set it to `null` either since setting it to `[]` has the same effect. I would just remove this line, since setting to `[]` to remove filters is something most people would intuit.

bigtimebuddy:
Fixes #7359

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -214,7 +214,7 @@ export abstract class DisplayObject extends EventEmitter
     public visible: boolean;
     public renderable: boolean;
     public filterArea: Rectangle;
-    public filters: Filter[];
+    public filters: Filter[] | null;
     public isSprite: boolean;
     public isMask: boolean;
     public _lastSortedIndex: number;


Zyie (approved):

SukantPal:
I don't think our typings are strict-nullable. We don't use that feature of Typescript so I'm not sure why this is necessary.

bigtimebuddy:
@SukantPal that is true, however, I think we should work toward supporting more strict null-checks. In this case, not having this requires that users `strictNullChecks` be `false`. We should not dictate compiler strictness for downstream users if we can help it. There are some things like `esModuleInterop` that we _can't_ help because of 3rd party libraries. 

I imagine that we may get more of these issues in the future, but I don't think that's a bad thing. It's an easy adjustment and gets us closer to strict null checks.

SukantPal:
I agree, but the proper way to do this will be to enable `strictNullChecks` in our `tsconfig.json`

SukantPal (approved):

bigtimebuddy:
Agreed. I'd be happy if someone wants to take a stab at doing that, trying to work towards `strict: true` in our tsconfig, would be awesome. PR(s) welcome for this!

PR #7370, opened 2021-04-02
Text lineheight and positionY

### LINKED ISSUE #7344: Text's lineHeight implementation does not match HTML
Text lines are current aligned to the top. When for instance the `lineHeight` is bigger than the `fontSize` the gap [gets added to the bottom](https://www.pixiplayground.com/#/edit/AVy4yl6EFkC5H9N4TYMAh).

This is not how HTML or graphics editors work.

Rubén from the [Slack channel](https://pixijs.slack.com/archives/C0XS4MVU0/p1616771712076200?thread_ts=1616746168.071900&cid=C0XS4MVU0) thinks this is due to this code in [Text.ts](https://github.com/pixijs/pixi.js/blob/7b79a9d14edaeac4476437dac19472ecdee683c8/packages/text/src/Text.ts):

`linePositionY = ((style.strokeThickness / 2) + (i * lineHeight)) + fontProperties.ascent;`

Maybe not everyone is building a graphics editor like me in PIXI, but I don't see why PIXI shouldn't be as close as possible to how HTML works.

Bouh:
##### Description of change
Fix #7344

##### Pre-Merge Checklist

I can build pixi.js but i can't find how to get the minified build for test on jsfiddle.net
How ever the result should be equal to this one: https://jsfiddle.net/Bouh/0p1f5kc3/12/

( Oops someone can make this PR a normal PR by removing draft? Thanks )

image

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -263,6 +263,10 @@ export class Text extends Sprite
                 linePositionX = style.strokeThickness / 2;
                 linePositionY = ((style.strokeThickness / 2) + (i * lineHeight)) + fontProperties.ascent;
 
+                if (i === 0)
+                {
+                    linePositionY = linePositionY * 2;
+                }
                 if (style.align === 'right')
                 {
                     linePositionX += maxLineWidth - lineWidths[i];
--- a/packages/text/src/TextMetrics.ts
+++ b/packages/text/src/TextMetrics.ts
@@ -167,13 +167,21 @@ export class TextMetrics
             width += style.dropShadowDistance;
         }
 
-        const lineHeight = style.lineHeight || fontProperties.fontSize + style.strokeThickness;
-        let height = Math.max(lineHeight, fontProperties.fontSize + style.strokeThickness)
+        let lineHeight = style.lineHeight || fontProperties.fontSize + style.strokeThickness;
+        let height = Math.max(lineHeight, (fontProperties.fontSize * 2) + style.strokeThickness)
             + ((lines.length - 1) * (lineHeight + style.leading));
 
         if (style.dropShadow)
         {
             height += style.dropShadowDistance;
+        }
+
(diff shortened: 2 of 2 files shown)


bigtimebuddy (changes requested):
Thanks for submitting, unfortunately this does not pass the sniff test: https://jsfiddle.net/bigtimebuddy/cgupLxr0/

I think this needs more work. The goal is to not match the demo, but to find the best general solution. Using the default value of `0` for `lineHeight` is breaking.

I forked a branch to the project: https://github.com/pixijs/pixi.js/tree/fix/text-line-height so that we could test out a build.

packages/text/src/Text.ts
```diff
@@ -263,6 +263,10 @@ export class Text extends Sprite
                 linePositionX = style.strokeThickness / 2;
                 linePositionY = ((style.strokeThickness / 2) + (i * lineHeight)) + fontProperties.ascent;
 
+                if (i === 0)
+                {
+                    linePositionY = linePositionY * 2;
```
bigtimebuddy, line 268: Could you explain this line? I'm not sure why the doubling.

packages/text/src/TextMetrics.ts
```diff
@@ -167,15 +167,23 @@ export class TextMetrics
             width += style.dropShadowDistance;
         }
 
-        const lineHeight = style.lineHeight || fontProperties.fontSize + style.strokeThickness;
-        let height = Math.max(lineHeight, fontProperties.fontSize + style.strokeThickness)
+        let lineHeight = style.lineHeight || fontProperties.fontSize + style.strokeThickness;
+        let height = Math.max(lineHeight, (fontProperties.fontSize * 2) + style.strokeThickness)
```
bigtimebuddy, line 171: I don't understand why you're doubling the `fontSize`.

Issue #7049, opened 2020-12-04
ROUND_PIXELS issues


dev7355608:
1. Incorrect rounding to grid in `Mesh.calculateVertices()`/`Sprite.calculateVertices()`.
https://github.com/pixijs/pixi.js/blob/8b9845ba423c7ea51efed0d81ae24e332568f400/packages/mesh/src/Mesh.ts#L404
https://github.com/pixijs/pixi.js/blob/8b9845ba423c7ea51efed0d81ae24e332568f400/packages/sprite/src/Sprite.ts#L342
It should be: `vertexData[i] = Math.round(vertexData[i] * resolution) / resolution;`

2. The documentation of `settings.ROUND_PIXELS` isn't quite correct. It says `Math.floor`, but `Math.round` is used everywhere.
https://github.com/pixijs/pixi.js/blob/8b9845ba423c7ea51efed0d81ae24e332568f400/packages/settings/src/settings.ts#L265-L276

3. Is rounding missing in `Mesh._renderDefault()`?

4. Is rounding missing in `Sprite.calculateTrimmedVertices()`?

5. Only `Mesh`, `Sprite`, and `BitmapText` seem to be affected by `ROUND_PIXELS`. Is that intended?

6. Another unrelated thing I just noticed: Is the type of `shader` in `Mesh` `MeshMaterial` or `Shader | MeshMaterial`?
https://github.com/pixijs/pixi.js/blob/8b9845ba423c7ea51efed0d81ae24e332568f400/packages/mesh/src/Mesh.ts#L58-L65
https://github.com/pixijs/pixi.js/blob/8b9845ba423c7ea51efed0d81ae24e332568f400/packages/mesh/src/Mesh.ts#L80-L85

ivanpopelyshev:
Oh , why is it `|0` inside my round? mnust be a bug.

Yes, we dont have implemenation for graphics rounding in v5, we had it in v4.

Shader|MeshMaterial stuff is our bane, we cant figure out what to do with it :)

ivanpopelyshev:
I sent invite to pixijs slack on your email. Lets have a talk :)

PR #7373, opened 2021-04-02
Fix rounding issues

dev7355608:
Continuation of #7041. This PR also addresses other rounding-related issues as well:

- Changed all `*.resize()` functions from  `*.resize(newWidth, newHeight)` to `*.resize(desiredWidth, desiredHeight)`, because not all dimensions are possible, i.e., the framebuffer cannot have fractional pixels. The new resize functions round to the nearest possible dimensions.
- Selected to use `Math.round()` everywhere instead of `Math.floor/ceil()`. It naturally works better than the other candidates in cases of slight imprecisions, e.g. `Math.round((widthInPixels / resolution) * resolution) === widthInPixels` which not always true for `floor` and `ceil` without special care. Of course other choices would be valid as well.
- Made sure that `RenderTexture.width/height` is always precisely the width/height in pixels divided by the resolution of the texture. 
- Added explicit rounding for the integer arguments of `gl.viewport`, `gl.scissor`, `gl.drawImage`. Again, we need to make sure to round especially in cases like  `width * resolution` which might not be integer even if `width === widthInPixels / resolution`: we could get one less than expected when this number is cast to int.
- Removed rounding in a few places where it's probably unnecessary.
- Fixed some of the `ROUND_PIXELS` issues in `Mesh` and `Sprite` (#7049).

##### Pre-Merge Checklist


ivanpopelyshev:
Looks good, but I have to check all those places one or two more times :)

bigtimebuddy:
I'd really prefer this was not an omnibus PR. I know the issues being addressed are related but isolating this into smaller PRs makes it easier to review these issues individually. We don't have a lot of test coverage for these areas (and there are no tests added here) so the consequences can be subtle and difficult to observe.

dev7355608:
PR was split in multiple parts. The source frame rounding doesn't really make sense after all, so you won't find it in the other PRs.

PR #7375, opened 2021-04-03
Fix framebuffer mipmap round

ivanpopelyshev:
Fixes #7374 

Demo: https://pixijs.io/examples/?v=fix-framebuffer-mipmap-round#/filters-basic/blur.js

1. reverted some changes 
2. added scale, like in previous PR
3. added rounding like in dev's PR.

PR #7377, opened 2021-04-04
Hotfix/frame buffer

GoodBoyDigital:
Fix for the frame issue in frame buffer system
added a test too. 👍


ivanpopelyshev:
Look at #7375 , i have necessary TODO:

```
// TODO: might be different for non pow2
```

can you add it too?

bigtimebuddy (approved):
Solid, thanks for fixing this.

SukantPal (approved):

PR #7384, opened 2021-04-05
Round texture, framebuffer, and canvas dimensions properly

### LINKED ISSUE #6510: Resizing renderer with resolution greater than one
I am currently facing an issue regarding scaling the renderer or canvas on a high DPI device (or more specifically, with `devicePixelRatio > 1`). 

### Expected behavior
From my understanding, setting `resolution: devicePixelRatio` together with `autoDensity: true` should allow the canvas to adjust seamlessly to the size of the renderer (resized using `renderer.resize()`). 

### Current behavior
Check out this [fiddle](https://jsfiddle.net/zjo5xpsh/44/) for an example where this fails on my device with windows scaling set to 125%, (i.e. `devicePixelRatio = 1.25`). As you resize the viewport you can see pixels stretching into odd shapes. 
**Note** Resizing works just fine if I turn windows scaling to 100% (i.e. `devicePixelRatio = 1`).

### The long way around
In the jsfiddle there's a block of code commented out at line 40 that does what (from my understanding) the renderer does internally to adjust for the DPI, i.e. multiply the canvas and  renderer sizes by the scaling factor and then use CSS style to scale the canvas back down.

This _does_ seem to improve things, oddly enough (perhaps someone could explain exactly why) but there is still some stretching going on (in both my application and the fiddle, there is _exactly one_ "stretch" line, as opposed to _many_ in the previous case, and _none_ with no scaling).

**Edit 2:** This workaround actually fixes the issue entirely. At the time of writing, I had mistakenly _left in_ the settings for `autoDensity = true` which introduces this single "stretch" (which, arguably, is worth looking into all on its own). For posterity, that bit of code requires `autoDensity = false` and `resolution = 1`, i.e. default settings.

### Environment

- **`pixi.js` version**:  5.2.1
- **Browser & Version**: Vivaldi - Chromium 80.0.3987.158 (64-bit) (also on the latest stable versions of Chrome, Edge and Firefox as of writing this)
- **OS & Version**: Windows 10 v1903
- **Running Example**: https://jsfiddle.net/zjo5xpsh/44/

Apologies if this is just me being a knob, been fighting this one for a couple days now.

**Edit:** Here are some gifs to illustrate the issue. Note the "squishing" and "stretching" of the pixels in the first example.
`devicePixelRatio = 1.25` 

`devicePixelRatio = 1`

dev7355608:
Part 1 of #7373.

- Changed all `*.resize()`/`.setSize()` functions from `*.resize(newWidth, newHeight)` to `*.resize(desiredWidth, desiredHeight)`, because not all dimensions are possible, i.e., the framebuffer cannot have fractional pixels. The new resize functions round to the nearest possible dimensions. It is ensured that `width * resolution` and `height * resolution` is equal to the size of the framebuffer/canvas (up to tiny rounding errors).
- Removed unnecessary/erroneous rounding.

Fixes #6510 and #7038/#7041.

##### Pre-Merge Checklist


ivanpopelyshev (approved):

bigtimebuddy:
Here's the build if anyone wants to test: https://pixijs.download/dev-pr-7384/pixi.min.js

dev7355608:
- Changed the rounding from `round` to `ceil` in `getOptimalTexture()` to ensure that the texture is never smaller than requested. I had concerns that changing it to `ceil` would cause the flop texture in the filter system to potentially double in size because of rounding errors. But it seems like `(s / r) * r` is always less than or equal to `s` if `s` is a power of two. So in this regard `ceil`-ing without epsilon doesn't seem to do any harm.
- I included the fix the viewport frame rounding as described in #7041.
- I noticed that I rounded the padding in `Text.updateText()` which doesn't seem make sense now. So I reverted it.

Paki70:
to 22. huhtik. 2021 klo 3.45 dev7355608 ***@***.***>
kirjoitti:

dev7355608:
- Rounded the `pixelPixel` uniform to ensure it's the same values as `realWidth`/`realHeight`.

bigtimebuddy (approved):

PR #7386, opened 2021-04-05
Fix text linePositionY

### LINKED ISSUE #7344: Text's lineHeight implementation does not match HTML

andrej-ilic:
##### Description of change

Fixes #7344

This way the text lines are not aligned to the top. The result looks similar to the HTML example and it does not seem to break when `lineHeight` is set to 0.

##### Pre-Merge Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,44 +57,44 @@
       }
     },
     "bundles/pixi.js": {
-      "version": "6.0.2",
+      "version": "6.0.1",
       "license": "MIT",
       "dependencies": {
-        "@pixi/accessibility": "6.0.2",
-        "@pixi/app": "6.0.2",
-        "@pixi/compressed-textures": "6.0.2",
-        "@pixi/constants": "6.0.2",
-        "@pixi/core": "6.0.2",
-        "@pixi/display": "6.0.2",
-        "@pixi/extract": "6.0.2",
-        "@pixi/filter-alpha": "6.0.2",
-        "@pixi/filter-blur": "6.0.2",
-        "@pixi/filter-color-matrix": "6.0.2",
-        "@pixi/filter-displacement": "6.0.2",
-        "@pixi/filter-fxaa": "6.0.2",
-        "@pixi/filter-noise": "6.0.2",
-        "@pixi/graphics": "6.0.2",
-        "@pixi/interaction": "6.0.2",
-        "@pixi/loaders": "6.0.2",
-        "@pixi/math": "6.0.2",
-        "@pixi/mesh": "6.0.2",
-        "@pixi/mesh-extras": "6.0.2",
-        "@pixi/mixin-cache-as-bitmap": "6.0.2",
-        "@pixi/mixin-get-child-by-name": "6.0.2",
-        "@pixi/mixin-get-global-position": "6.0.2",
-        "@pixi/particles": "6.0.2",
-        "@pixi/polyfill": "6.0.2",
-        "@pixi/prepare": "6.0.2",
-        "@pixi/runner": "6.0.2",
-        "@pixi/settings": "6.0.2",
-        "@pixi/sprite": "6.0.2",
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Could you please remove the package-lock changes?

andrej-ilic:
Done.

bigtimebuddy:
Thanks @andrej-ilic, this will take a bit to verify.

bigtimebuddy (approved):
This works for me, I didn't see anything unexpected fiddling with the lineHeight.

**Broken: (dev)**
https://jsfiddle.net/bigtimebuddy/t10wnyvf/

**Fixed: (this PR)**
https://jsfiddle.net/bigtimebuddy/7ygvL69p/

**Build:**
https://pixijs.download/dev-pr-7386/pixi.min.js

bigtimebuddy:
@themoonrat could you let us know what you think about this Text change?

SukantPal (approved):

AShenawy:
I understand that this is an issue because it's not like how HTML handles line height. But as per typography rules and definitions (see [here](https://material.io/design/typography/understanding-typography.html#readability)):

> Line height, also known as leading, controls the amount of space between baselines in a block of text. A text’s line height is proportional to its type size.

This means that the current way Pixi handles it is the proper way as far as typography is concerned.

bigtimebuddy:
I agree @AShenawy that vertically centering text using line-height isn't what it's intended for, however, currently with Pixi Text objects there's no way for devs to know where the first baseline is located. It's basically a guess. That's not the case for HTML text and knowing this information can be pretty important for doing layout. This doesn't change the overall approach for line-height _except_ for the first line. 

Here's an example that illustrates the issue. Look add the line and how it stays relative to the text baseline.

**This PR**
https://jsfiddle.net/bigtimebuddy/qp7ejrb6/

**Current Behavior**
https://jsfiddle.net/bigtimebuddy/4wvoje13/

themoonrat (changes requested):
Messaged @bigtimebuddy on Slack with my concerns over this; that it _will_ effect lots of already made content out there. It's a visually breaking change.

Also an issue of text getting cut off with lower value line heights which didn't occur before https://jsfiddle.net/e8r21yqg/ to https://jsfiddle.net/fkjvqp3m/

andrej-ilic:
Yes, I agree. It may be bad to affect so many projects.
The last commit should fix the text getting cut off issue, in case you want to merge this PR.

andrej-ilic:
I made it opt-in.

I the 'Documentation is changed or added' check to the PR.

bigtimebuddy:
@nuthinking, is this good for you?

bigtimebuddy (approved):
Updated example with switch. Seems to work great:
https://jsfiddle.net/bigtimebuddy/7ygvL69p/

Documentation preview here:
https://pixijs.download/dev-pr-7386/docs/PIXI.Text.html#nextLineHeightBehavior

nuthinking:
@bigtimebuddy yes, looks great. Thanks! 🙌

themoonrat (approved):

PR #7395, opened 2021-04-08
Remove application dependency on loader

bigtimebuddy:
Found a small bug when trying to do a minimal install with core, loaders, ticker and utils. There was a dangling app reference in loaders that threw an error. This decouples these types.


Zyie (approved):

ivanpopelyshev (approved):

ivanpopelyshev:
btw, `constants` also belongs to minimal install

PR #7406, opened 2021-04-14
Fix FXAA filter

dev7355608:
The FXAA filter uses `inputPixel` when it should use `inputSize`. [Here](https://www.pixiplayground.com/#/edit/kqPXwNtF5wl2UlvmcrgNM) is an example that shows how it fails when the resolution is something other than 1.

##### Pre-Merge Checklist


bigtimebuddy (approved):

ivanpopelyshev (approved):

SukantPal (approved):
Nice catch!

bigtimebuddy:
Thanks @dev7355608 another great PR :+1:

PR #7413, opened 2021-04-16
Add the export namespace to the declaration file

JetLua:
Want to have a namespace called `PIXI` just like `v5`. It's easy to do some custom configuration in `vscode`.

Now: 
https://unpkg.com/browse/pixi.js@6.0.2/index.d.ts

```ts
export {}
```

This PR will replace it with:

```ts
export as namespace PIXI;
```

### DIFF
--- a/scripts/injectGlobalMixins.ts
+++ b/scripts/injectGlobalMixins.ts
@@ -33,18 +33,13 @@ async function getSortedPackages(): Promise<SimplePackageJson[]>
 function writeToIndex(basePath: string, dataToWrite: string): void
 {
     const indexDtsPath = path.resolve(basePath, './index.d.ts');
-    const file = fs.readFileSync(indexDtsPath, { encoding: 'utf8' }).toString().split('\n');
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Why not add the namespace to the individual packages too, not just the bundles? In that case, you could probably add this to `writeToIndex` function instead.

bigtimebuddy:
I tested this and it works well if you want to define types using the PIXI global. For instance:

```typescript
import type { Texture } from 'pixi.js';
let texture: Texture;
```
could become:
```typescript
let texture: PIXI.Texture;
```

bigtimebuddy:
@JetLua what do you think about adding this for all packages instead of just bundles?

JetLua:
All right. I'll try it tonight.

JetLua:
@bigtimebuddy Done.

bigtimebuddy (approved):
Much cleaner :+1:

Zyie (approved):
looks good!

bigtimebuddy:
I think we should revert this change.

PR #7414, opened 2021-04-16
Documentation Update: Point.ts

jrod-disco:
##### Description of change
First pass at contributing to the documentation update effort. Please let me know if there is anything I should change or could do better.

* `fix` - update documentation according to: #7159

##### Pre-Merge Checklist

### DIFF
--- a/packages/math/src/Point.ts
+++ b/packages/math/src/Point.ts
@@ -2,8 +2,8 @@ import type { IPoint } from './IPoint';
 import type { IPointData } from './IPointData';
 
 /**
- * The Point object represents a location in a two-dimensional coordinate system, where x represents
- * the horizontal axis and y represents the vertical axis.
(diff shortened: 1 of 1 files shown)


jrod-disco:
Please note that type inference was not working as expected on `set` which is why I added types/defaults explicitly in the doc.

bigtimebuddy (approved):
Yes!!! Thank you for helping with this!

SukantPal (approved):
👏

jrod-disco:
I may have blown away the constructor doc. Will check and push again if needed.
Just noticed as I was working on ObservablePoint.

jrod-disco:
Sorry about that. Pretty sure we want to keep the constructor bit in there yeah? Even if it isn't pure inference.

Should be good now.

image

SukantPal (approved):

PR #7415, opened 2021-04-16
Documentation Update: ObservablePoint.ts

jrod-disco:
##### Description of change

Update docs for `ObservablePoint.ts`

* `fix` - update documentation according to: #7159

I made the assumption that we're attempting to document the surface API and there was no need to expose `_x` and ` _y` internals. Please correct me if I'm wrong in this assumption.

_Note: That there seems to be an issue with webdoc not properly inferring types from documentation on getters (and likely setters). Followed workaround provided by @bigtimebuddy here: https://github.com/webdoc-labs/webdoc/issues/100_
  
##### Pre-Merge Checklist


jrod-disco:
Should be good to go. Please let me know if there are any issues.

image

bigtimebuddy:
Thanks for this @jrod-disco 

@SukantPal would be really nice to have a look at these inference issues with webdoc.

bigtimebuddy (approved):

packages/math/src/ObservablePoint.ts
```diff
@@ -2,28 +2,35 @@ import type { IPointData } from './IPointData';
...
     _y: number;
 
     /**
-     * @param {Function} cb - callback when changed
-     * @param {object} scope - owner of callback
+     * Creates a new `ObservablePoint`
+     *
+     * @param cb - callback function triggered when `x` and/or `y` are changed
+     * @param scope - owner of callback
      * @param {number} [x=0] - position of the point on the x axis
```
jrod-disco, line 31: Explicit type / default was needed here. Not sure if there's a better way around this.

packages/math/src/ObservablePoint.ts
```diff
@@ -117,10 +124,8 @@ export class ObservablePoint<T = any> implements IPoint
     }
     // #endif
 
-    /**
-     * The position of the displayObject on the x axis relative to the local coordinates of the parent.
-     *
-     * @member {number}
+    /** Position of the observable point on the x axis
+     * @type {number}
```
jrod-disco, line 128: Again, needed to explicitly set the type of this member. It wasn't inferring on either getter or setter.

SukantPal (approved):
@webdoc/cli 1.2.1 will be getting the fix for interface property type inference.

bigtimebuddy:
Thank you @jrod-disco :tada:

PR #7422, opened 2021-04-17
Fix rendering to multisampled framebuffers

### LINKED ISSUE #6844: Multisampled textures not handled at all in PixiJS internally

dev7355608:
Fixes #6844.

The blit that was added to the stencil system is wrong. Blit copies color, but we are in the stencil system. I don't see why we would need that there. If left in there, an empty mask (empty graphics object as mask) has no effect at all, i.e., the object is completely visible when it should be not.

All attachments of a framebuffer must be multisampled with the same number of samples if one attachment is multisampled. Therefore the depth/stencil buffer needs to be multisampled as well. Because there could be multiple color texture and depth texture attachments, I added the function `canMultisampleFrambuffer` to check if multisampling is possible: there are no multisampled textures in webGL, and the blit function as currently implemented only copies the color attachment 0. So if multisampling is not possible, it is rendered without multisampling silently.

I also removed the if-statement around the blit in the filter system. It's unnecessary: all required checks are done in the blit function.

##### Pre-Merge Checklist


dev7355608:
- Added missing deletion of `msaaBuffer` in `disposeBuffer`.
- `msaaBuffer` is now deleted when it is no longer used after `updateFramebuffer`.
- Added missing resizing of `msaaBuffer` in `resizeFramebuffer`.
- Added missing resizing of `stencil` in `updateFramebuffer`.
- The first color texture needs to be bound in `updateFramebuffer`; otherwise resizing of the render texture fails.
- Update `dirtySize` when calling `updateFramebuffer`, because  `updateFramebuffer` also resizes.
- `resizeFramebuffer` didn't account for `parentTextureArray`. Fixed.

Resizing multisampled framebuffers is now working.

ivanpopelyshev:
OMG.

This is complex behaviour, we need playgrounds/jsfiddles for that, do you agree, @bigtimebuddy  ?

bigtimebuddy:
I agree, we need to de-risk this work with tests/examples or preferably deliver this as _much_ smaller changes that are easier to review. No one will approve this because the risk is too high.

dev7355608:
Here are some tests. The problems are connected/layered, so I can't really demonstrated each problem individually in v6.0.2.

- Masks not working with multisampling because of the incorrect blit in the mask system: https://www.pixiplayground.com/#/edit/ybrmwKaqoVkrl_9_9vCad 
- Incomplete multisampled framebuffer because the stencil buffer isn't multisampled: https://www.pixiplayground.com/#/edit/r9wS_d0bwxw_LyT1bMB__
- Incomplete multisampled framebuffer because of multiple color attachments: https://www.pixiplayground.com/#/edit/3d1xX8PWq8kFmt3cqo1Zy
- Resize not working with multisampling: https://www.pixiplayground.com/#/edit/FGHjQxIoYQD8aqFBqBJRt
- Stencil buffer not resized in `updateFramebuffer` causes failure even if multisampling is not enabled: https://www.pixiplayground.com/#/edit/YNQnguJkaYQeoPGngX3Yt

dev7355608:
When I started to create unit tests for these problems earlier today, I noticed that all my test were failing: because `dirtyId` etc in `GLFramebuffer` are not initialized correctly to `-1`. Consequently, new framebuffers are actually never setup at all when bound. This bug was probably never noticed because `RenderTexture` instantly updates `dirtyId` when adding the color texture. So I included the fix for that bug here; otherwise the unit tests won't work.

I'm going to add unit tests for almost all of the problems I described above tomorrow.

Feu-Secret:
Hi. Is it possible for the multisample resize problem to have its own PR? I had a big headache trying to resolve a problem when resizing multisampled RenderTexture until I saw this PR :smile:

ivanpopelyshev (approved):
So many tests!!! the only way to make it better is to actually merge into pixi and look at users reaction

ShukantPal (approved):
Let it run in the wild 🔥!

bigtimebuddy:
@dev7355608 could you please resolve conflicts here?

PR #7423, opened 2021-04-17
Add support for multisampled filters

dev7355608:
Added the multisample property to filters. It behaves very much like filter resolution.

Since multisample rendering is broken at the moment, you need to apply #7422 first in order to test this PR.

It's also probably a good idea to add getters/setters for the `multisample` property of the framebuffer in `RenderTexture` and make it an option in the constructor of `RenderTexture` to make it more accessible.

##### Pre-Merge Checklist


ivanpopelyshev:
It will eat memory like sponge eats water, that's why i didnt want to add it in initial MSAA PR's. Need to test it with heavy things like BlurFilter, hope they wont re-use this for temporary buffers.

SukantPal (changes requested):
You need to blit the filter in the single-filter case as well (L312 on dev, L317 on your branch)

dev7355608:
> You need to blit the filter in the single-filter case as well (L312 on dev, L317 on your branch)

I don't think that's necessary, because we blit when the `lastState` is popped, and blitting there would be too early, because rendering to `lastState.renderTexture` might not yet complete at that point. When `lastState` is popped, rendering to `lastState.renderTexture` is done, and we blit. In the multi-filter case I also didn't add a blit after the final apply call.

SukantPal:
Okay, that makes sense. Also when there are multiple filter-stacks in the FilterSystem, the lastState.renderTexture would be a render-texture managed by the FilterSystem. You’d expect FilterSystem to blit it, and as you mentioned we do so when lastState is popped. I did not notice this because it isn’t in the diff. But since we’re supporting multisampling now, can you add a comment explaining why we don’t blit immediately after running the last filter and where we actually do blit lastState.renderTexture (which happens when right before the next filter stack is run on L311 in your branch)

SukantPal (approved):
Thanks!

ivanpopelyshev (changes requested):
default should be NONE , always. Otherwise we'll have to add NONE in all multi-pass filters like BlurFilter, they dont need MSAA in middle buffers.

Also consider for those filters input requests MSAA, and there's no-MSAA buffer in pool - framebuffer will be changed, and the whole point of pool is to get rid of those operations.

So, we need multiple pools. 

Workaround for those who want always use MSAA in filters is easy - `renderer.texturePool = myTexturePool;`

dev7355608:
Good points.

I changed the default multisample of `getFilterTexture` to `NONE`. I think this covers the multi-pass filters like the blur filter. But there's still the question what to with the flip and flop texture in the filter system. Is OK that they are both multisampled? Or would we like to use non-multisampled textures as well? And if yes, we have the awkward case that me might need one more texture.

Instead of using multiple pools I thought it might be a good idea to extend the current render texture pool and use separate keys for multisampled textures.

Thoughts? @ivanpopelyshev

ShukantPal:
I think RenderTexturePool can be extended to handle multisample textures alongside regular ones.

Sent from my iPad

ivanpopelyshev:
`getFilterTexture` should return msaa only for `push()`, yes, its obsolete for usage from filters

packages/core/src/filters/FilterSystem.ts
```diff
@@ -337,10 +351,17 @@ export class FilterSystem implements ISystem
 
             filters[i].apply(this, flip, lastState.renderTexture, CLEAR_MODES.BLEND, state);
 
+            if (i > 1 && state.multisample > 1)
```
ivanpopelyshev, line 354: I usually store this thing as a single flag, so its guaranteed that we actually were in previous if. I understand its obvious in this case, but we also have things like `(1*2)+1` in our eslint, so maybe we can be careful here too?

ivanpopelyshev (approved):

PR #7424, opened 2021-04-18
Upgrade to webdoc 1.2.1, remove bug workarounds

SukantPal:
* No need for explicit @protected tags on protected methods.
* Fixed @method tag with PIXI.Graphics#drawMethod inferring it as static instead of instanced scoped.

In a follow-up, you can also strip out the explicit @type tag applied on interface properties.


bigtimebuddy (approved):
Nice!

bigtimebuddy:
Would also be nice to get in this: https://github.com/webdoc-labs/webdoc/issues/110

Zyie (approved):

bigtimebuddy:
Thanks @SukantPal

Issue #7389, opened 2021-04-06
Graphics drawPolygon does not consider line alignment for bounds (drawShape does)


AutomaticGood:
### Expected Behavior

When drawing a polygon, line alignment 0.0 gets ignored.
line alignment 0.0 should draw the line on the inside of the polygon so that the object's width/height is in accordance with the vertices.

### Current Behavior

line alignment 0.0 is ignored. the line is drawn on the outside.

### Steps to Reproduce

Please see pixi playground https://www.pixiplayground.com/#/edit/FzzLfCKL3ZnJAj-8sjkW1

bigtimebuddy:
This looks correct to me. Am I missing something?

https://www.pixiplayground.com/#/edit/RVgzb0qSTKXeZuGEQzs1o

![Screen Shot 2021-04-06 at 9 44 30 AM](https://user-images.githubusercontent.com/864393/113747943-c2bb3980-96bc-11eb-98d1-df14c4d9fe34.png)

andrej-ilic:
The objects are drawn correctly. I think this is the issue:

Here is a 100x100 square with stroke 10 drawn with a drawPolygon (left) and drawRect (right). The width/height of the polygon is  greater than the rectangle.

![image](https://user-images.githubusercontent.com/8559012/113751006-bf02d380-970b-11eb-9918-9e7bbf2edeb1.png)

ivanpopelyshev:
There are so many things in graphics that still dont account for lineAlignment :) 

Just in case, dont even try to take into account things like caps/joints, even Flash didnt do that.

bigtimebuddy:
Oh got it, so the bounds are wrong when considering a polygon with line alignment. That seems like a bug. I agree with @ivanpopelyshev probably a few places that aren't accounting for that. We should attempt to find a cohesive approach.

andrej-ilic:
Changing this line:

https://github.com/pixijs/pixi.js/blob/170e890aad69828c394417cafbb3c0a0d928e8d3/packages/graphics/src/GraphicsGeometry.ts#L941-L944

to `lineWidth = lineWidth * alignment;` fixes the bounds. I will submit a PR later if you think this might be it.

ivanpopelyshev:
Probably should do it only for `outer` case, depends on orientation of polygon, but we can fix it in another PR later :)

PR #7434, opened 2021-04-21
Fix incorrect bounds for drawPolygon with lineAlignment lower than 0.5

andrej-ilic:
Fixes: #7389

When using drawPolygon with lineAlignment value less than 0.5, there is a gap between the bounds and the rectangle. The bounds are correct when using drawRect, though.

Example of bug: https://www.pixiplayground.com/#/edit/9JJqaW3NtUcH-2BRn4zgk

##### Description of change
Fix bounds when using drawPolygon with lineAlignment that is less than 0.5.

##### Pre-Merge Checklist


bigtimebuddy:
A unit-test would be great for this, should be an easy check. The test should fail and dev and pass with this PR.

JetLua:
It would be even better if we could consider orientation of vertices.

https://www.pixiplayground.com/#/edit/0-1pqUSC44SY11WpV_-0h

```js
const vertices = [
    new PIXI.Point(0, 0),
    new PIXI.Point(100, 0),
    new PIXI.Point(100, 100),
    new PIXI.Point(0, 100)
].reverse(); // There is no problem
```

bigtimebuddy (approved):

SukantPal (approved):
Little details!

bigtimebuddy:
Thank you @andrej-ilic

PR #7438, opened 2021-04-22
Fix default type and internalFormat of GLTexture

dev7355608:
### Description of change

Swap the default values of `GLTexture.type` and `GLTexture.internalFormat`, because it's the wrong way around.

```
GL_UNSIGNED_BYTE = 5121
GL_RGBA          = 6408 
```

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/GLTexture.ts
+++ b/packages/core/src/textures/GLTexture.ts
@@ -1,3 +1,5 @@
+import { FORMATS, TYPES } from '@pixi/constants';
+
 /**
  * Internal texture for WebGL context
  * @class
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):

packages/core/src/textures/GLTexture.ts
```diff
@@ -66,13 +66,13 @@ export class GLTexture
          * Type copied from baseTexture
          * @member {number}
          */
-        this.type = 6408;
+        this.type = 5121;
```
SukantPal, line 69: Let's use `import('@pixi/constants').TYPES.UNSIGNED_BYTE` instead of a number literal.

packages/core/src/textures/GLTexture.ts
```diff
@@ -66,13 +66,13 @@ export class GLTexture
...
 
         /**
          * Type copied from baseTexture
          * @member {number}
          */
-        this.internalFormat = 5121;
+        this.internalFormat = 6408;
```
SukantPal, line 75: Let's use `import('@pixi/constants').FORMATS.RGBA` instead of a number literal.

SukantPal (changes requested):

ivanpopelyshev (approved):

SukantPal (approved):

bigtimebuddy:
Thanks @dev7355608

PR #7441, opened 2021-04-22
Multisampling and cacheAsBitmap

dev7355608:
##### Description of change

Added `cacheAsBitmapMultisample`. Default behavior (if `cacheAsBitmapMultisample` is undefined): it renders with `HIGH` samples if the renderer has antialias enabled; if antialias is not enabled, it renders without multisampling.

##### Pre-Merge Checklist


bigtimebuddy (approved):
I think this is a good idea. Makes cacheAsBitmap more powerful.

ivanpopelyshev (changes requested):

packages/mixin-cache-as-bitmap/src/index.ts
```diff
@@ -4,13 +4,15 @@ import { Container, DisplayObject, IDestroyOptions } from '@pixi/display';
...
 import { settings } from '@pixi/settings';
+import { MSAA_QUALITY } from '@pixi/constants';
 import type { CanvasRenderer } from '@pixi/canvas-renderer';
 
 const _tempMatrix = new Matrix();
 
 DisplayObject.prototype._cacheAsBitmap = false;
 DisplayObject.prototype._cacheData = null;
 DisplayObject.prototype._cacheAsBitmapResolution = null;
+DisplayObject.prototype._cacheAsBitmapMultisample = undefined;
```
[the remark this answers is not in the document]
dev7355608: I chose the same default behavior as `cacheAsBitmapResolution`: if not defined explicitly, use what the renderer is using. I think this makes the most sense for cacheAsBitmap and that resolution and multisample have the same default behavior. If we render the object once (or every infrequently), we might as well render it in good quality, I think.

packages/mixin-cache-as-bitmap/src/index.ts
```diff
@@ -247,6 +282,15 @@ DisplayObject.prototype._initCachedDisplayObject = function _initCachedDisplayOb
         resolution: this.cacheAsBitmapResolution || renderer.resolution,
     });
 
+    if (this.cacheAsBitmapMultisample !== undefined)
+    {
+        renderTexture.framebuffer.multisample = this.cacheAsBitmapMultisample;
+    }
+    else if (renderer.gl.getContextAttributes().antialias)
+    {
+        renderTexture.framebuffer.multisample = MSAA_QUALITY.HIGH;
```
SukantPal, line 291: Multisampling was introduced as an experimental feature. I'd be careful in making it default behavior just yet.

SukantPal (approved):
Otherwise nice work

packages/mixin-cache-as-bitmap/src/index.ts
bigtimebuddy: Agreed

packages/mixin-cache-as-bitmap/src/index.ts
```diff
@@ -247,6 +282,15 @@ DisplayObject.prototype._initCachedDisplayObject = function _initCachedDisplayOb
...
+    }
```
bigtimebuddy, line 292: I this this is good, but it is a change in the default behavior and should be considered a breaking change. We can either remove it or provide a static property to opt-in to this behavior, then make it the default in v7. Thoughts @dev7355608?
 
```suggestion
```

dev7355608: I removed it. I think we don't really need an opt-in property. For example, anyone who wants to force multisampling everywhere could simply change the default like this.
```js
DisplayObject.prototype._cacheAsBitmapMultisample = PIXI.MSAA_QUALITY.HIGH;
```

PR #6867, opened 2020-09-12
Account for scaling b/w sourceFrame → destinationFrame in ScissorSystem

SukantPal:
Fixes #6866

-----
:pen: https://codepen.io/sukantpal/pen/mdPLvJG?editors=0010

PR #7442, opened 2021-04-23
Round gl.scissor and gl.viewport arguments

dev7355608:
##### Description of change

Round `x`, `y`, `width`, and `height` before passing them to `gl.scissor()`/`gl.viewport`. Both of these function expect integer, and floats are truncated which isn't ideal. `Math.round()` seems like the best method of rounding here.

I also noticed after #6867 scissoring isn't so clean anymore. Consider the following test. Two squares, one larger than the other, and the larger of them has a mask that is exactly the same shape as the smaller square. They are placed exactly on top of each other. Ideally only the foreground square should be visible. But that's not the case anymore after that change was made to the ScissorSystem. But rounding fixes this as these playgrounds show. And is needs to be `Math.round()` to get the desired behavior.

###### playgounds:
- [5.3.9](https://www.pixiplayground.com/#/edit/yO81YbQ9h8tUoDzWMR0lc) ✔️ 
- [5.4.0-rc1](https://www.pixiplayground.com/#/edit/c-eksCN--0QFlbyhpNyuc) ❌ 
- [6.0.2](https://www.pixiplayground.com/#/edit/_P6R1XmwOSyVWfgBLmbzK): Comparison of rounding ✔️ vs. not rounding ❌ 

##### Pre-Merge Checklist


bigtimebuddy (approved):

ivanpopelyshev (approved):

bigtimebuddy:
Thank you @dev7355608

PR #7450, opened 2021-04-24
feat: Use Fetch+blob for ImageBitmap

### LINKED ISSUE #7436: Decode images in a web worker
While porting my game engine from raw WebGL to PIXI I noticed that I got back those dreaded dropped frames while decoding game backgrounds and spritesheets. While profiling it, I noticed the browser was decoding images in the main thread whenever it dropped frames:

image
In this exampe, the task took 27 ms. Decoding the image (a JPG of 1250x500 pixels) took 19.92 ms, and the rest of texImage2D (preparing and uploading to the GPU) took 5.24 ms. If it wasn't for the decoding part, the task would have fit within the frame budget without issues.

In my original code I used to have a tiny web worker that would decode the images off the main thread using `createImageBitmap`. I was using the same approach that PIXI already uses for the BASIS web worker (declare a function, convert it to string, and use a blob as the URL for the worker).

It caught me by surprise that PIXI didn't do that as well for non-basis images, given how much love there is for solid frame-rates.

Is this something that could be included in the next version of PIXI? If not by default, it would be awesome to have it as a flag that can be turned on for those of us dealing with large images.

In the meantime, can the application code do this with a loader middleware? I didn't see support for `ImageBitmap` in the ResourceType enum, but I couldn't find much documentation for the resource-loader library, so I might be looking at the wrong place.

Thanks!

eXponenta:
Use a Fetch api for `createImageBitmap` for prevent synced decoding of `Image` in `createImageBitmap`
Fix #7436

##### Description of change

Test: https://github.com/pixijs/pixi.js/issues/7436#issuecomment-826065150

##### Pre-Merge Checklist


bigtimebuddy:
I'm okay with this if it works. Fetch and createImageBitmap have roughly the same browser support profiles. We have not used fetch more broadly because IE11, but it's already gated by a check here for createImageBitmap.

https://caniuse.com/createimagebitmap

https://caniuse.com/fetch

eXponenta:
> I'm okay with this if it works. Fetch and createImageBitmap have roughly the same browser support profiles. We have not used fetch more broadly because IE11, but it's already gated by a check here for createImageBitmap.
> 
> https://caniuse.com/createimagebitmap
> 
> https://caniuse.com/fetch

You can check test in linked issue, but yep, because createImageBitmap is more specific that fetch, it shouldn't be source of problems.
fetch is not fully required, because only need for load raw blob that can be passed to createImageBitmap.
fetch used only because shorten that XHR.

I tryed to Blobing a Image, but Blob constructor not allow this case.

bigtimebuddy (approved):

ivanpopelyshev (approved):

bigtimebuddy:
Thank you @eXponenta

PR #7452, opened 2021-04-25
InteractionManager: replace useCapture with options

puxiao:
Related issue: [5099](https://github.com/pixijs/pixi.js/issues/5099)

#### Description of change

**what?**

<br>

**why?**

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

'parameters > options > passive'

> ```
> A Boolean that, if true, indicates that the function specified by listener will never call preventDefault(). If a passive listener does call preventDefault(), the user agent will do nothing other than generate a console warning.
> ```

<br>

`Improving scrolling performance with passive listeners`

> To prevent this problem, some browsers (specifically, Chrome and Firefox) have changed the default value of the `passive` option to `true` for the `touchstart` and `touchmove` events on the document-level nodes [`Window`](https://developer.mozilla.org/en-US/docs/Web/API/Window), [`Document`](https://developer.mozilla.org/en-US/docs/Web/API/Document), and [`Document.body`](https://developer.mozilla.org/en-US/docs/Web/API/Document/body). This prevents the event listener from being called, so it can't block page rendering while the user is scrolling.

<br>

passive_default

<br>

I don't think it's necessary to honor IE anymore. 

**And**

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

`Matching event listeners for removal`

> While `addEventListener()` will let you add the same listener more than once for the same type if the options are different, the only option `removeEventListener()` checks is the `capture`/`useCapture` flag. Its value must match for `removeEventListener()` to match, but the other values don't.

<br>

**So...**

replace `true` with `{ capture: true, passive: false }`

```diff
+	private readonly _eventListenerOptions: { capture: true, passive: false };
+	this._eventListenerOptions = { capture: true, passive: false };

-	xxx.addEventListener('xxx', xxx, true);
+	xxx.addEventListener('xxx', xxx, this._eventListenerOptions);

-	xxx.removeEventListener('xxx', xxx, true);
+	xxx.removeEventListener('xxx', xxx, this._eventListenerOptions);
```


SukantPal (approved):

bigtimebuddy (approved):

bigtimebuddy:
Thanks @puxiao

PR #7453, opened 2021-04-26
Fix dirty check in Mesh.calculateVertices()

dev7355608:
##### Description of change

`Mesh` accepts `Geometry`, but `(geometry as any).vertexDirtyId` is only defined for `MeshGeometry`. So, if `MeshGeometry` is not used, `calculateVertices` does not recalculate `vertexData` when just the geometry is updated. Fixed it by accessing `geometry.buffers[0]._updateID` directly.

Also I noticed that `vertexDirty` is initialized with `0`, but it should be `-1` technically if I'm not mistaken. But it doesn't make a difference because `_transformID` is initialized with `-1`.

##### Pre-Merge Checklist


SukantPal (approved):
Nice catch!

bigtimebuddy (approved):

ivanpopelyshev (approved):
You nailed it!

bigtimebuddy:
Thanks @dev7355608

PR #7455, opened 2021-04-28
Remove misleading description

suyingtao:
##### Description of change
Description of ``BLEND_MODES`` is out of date. The WebGL renderer supports for each blend mode.

### DIFF
--- a/packages/constants/src/index.ts
+++ b/packages/constants/src/index.ts
@@ -53,9 +53,6 @@ export enum BUFFER_BITS {
 /**
  * Various blend modes supported by PIXI.
  *
- * IMPORTANT - The WebGL renderer only supports the NORMAL, ADD, MULTIPLY and SCREEN blend modes.
- * Anything else will silently act like NORMAL.
- *
  * @memberof PIXI
  * @name BLEND_MODES
  * @enum {number}


bigtimebuddy (approved):

SukantPal:
Does it really? cc @ivanpopelyshev (then pixi-picture is obsolete?)

ivanpopelyshev:
pixi-picture works through filters, and i didnt move it to v6 yet.

No, OVERLAY/HARDLIGHT and others still have to go through filters.

However, all composition modes (DST_IN and such) are supported now, we might alter that comment, @suyingtao is right about that.

SukantPal:
Then I'd would ask that the documentation reflect this information instead of simply removing mention of the exceptions.

bigtimebuddy:
Can someone propose a better description about blend modes?

SukantPal:
We just need to list the unsupported blend modes in the description.

bigtimebuddy (changes requested):
@suyingtao Thanks for this, but I think this need to be tweaked. Here are the list of blend modes supported on WebGL only (Canvas can support all):

NORMAL
ADD
MULTIPLY
SCREEN
ADD_NPM
SCREEN_NPM
NONE
SRC_IN
SRC_OUT
SRC_ATOP
DST_OVER
DST_IN
DST_OUT
DST_ATOP
SUBTRACT
ERASE
XOR

bigtimebuddy:
After a lot of inactivity, I'm going to close this. Someone is welcome to do an update here or @suyingtao happy to re-open if you want to make the fixes.

Issue #7444, opened 2021-04-23
getChildByName is set as optional in v6


MrEmanuel:
Updating to pixi v6 we've run in to a minor issue. The method `getChildByName` on `Container` is set as optional. 
Is this method (and other attributes added via GlobalMixins) actually undefined at any point, or is the typing wrong? 
We're using the method frequently in our code base and it's tedious to add falsy checks everywhere. If it's actually not guaranteed to be defined, then we will of course do the checks, but I suspect that it's actually always defined. 

Also, I don't understand why the method is moved to a mixin and not defined on the class directly as it was in v5. 

In summary, this boils down to two questions:
1. Is the `getChildByName` not always guaranteed to be defined and the typing is correct, or not? 
2. What's the reasoning behind the `GlobalMixins` and why was `getChildByName` moved there in v6?

Thanks in advance! 

```ts
// https://github.com/pixijs/pixi.js/blob/dev/packages/mixin-get-child-by-name/global.d.ts
declare namespace GlobalMixins
{
    interface DisplayObject
    {
        name?: string;
    }

    interface Container
    {
        getChildByName?(name: string, isRecursive?: boolean): import('@pixi/display').DisplayObject;
    }
}
```

SukantPal:
The method was always defined as a mixin. This is a typings issue I think.

ivanpopelyshev:
this mixin is moved to separate package, so you can build pixi without it through `pixi-customize`. That's why its optional.

SukantPal:
Yeah, but since the typings is declared in the mixin package, it shouldn’t be optional.

MrEmanuel:
Thanks for the fast response @SukantPal. 
I think there are more of the same problems with other methods added via mixins. 
For instance `getGlobalPosition`. There might be more attributes that we're not using and thus haven't noticed.

ivanpopelyshev:
> I think there are more of the same problems with other methods added via mixins.

We found only one method to make typings for mixins in new ES6 modules. I didnt see any other projects that use it, everyone just ignore the problem.

It took us a year to find it and implement.

The downside - all those methods and variables are optional, because we inject in global interfaces that other pixi classes implement. 

There are two ways to fix that:
1. find us another method to do mixins.
2. make your own fork of pixi, put all mixins in same package.

[...]

PR #7458, opened 2021-04-30
Fix GlobalMixins to be avoid issues with strictNullChecks

bigtimebuddy:
Follow up to #7451

Conversation context #7444

This change makes the global mixins non-optional. Some of these were optional that shouldn't be.


ivanpopelyshev:
I have an idea how that all can affect our classes in a bad way, gonna check it tomorrow, it might be connected to https://github.com/pixijs/spine/issues/387

SukantPal (approved):

Zyie (approved):

PR #7462, opened 2021-05-02
Upgrade webdoc to 1.3.3 (@pixi/webdoc-template to 1.3.1)

SukantPal:
(No effect on how docs look)


SukantPal:
No idea why this is failing @bigtimebuddy

bigtimebuddy:
Replacing with #7472. I believe the package-lock.json got into a bad state by upgrading and downgrading between v1 and v2 of the lock file. This was having some unintended side effects after building. Cleanly reverting helped.

Issue #7477, opened 2021-05-07
Can't load SVG images in Pixi 6


nikolas:
### Expected Behavior
SVG image loads with no issue, as it did in Pixi 5.x.

### Current Behavior
I get this JS error when running my Pixi scene with Pixi 6:
> /~nik/astro-simulations/sun-motion-simulator/%3C?xml%20version=%221.0%22%20encoding=%22UTF-8%22%20standalone=%22no%22?%3E%3Csvg%20%20%20xmlns:dc=%22http://purl.org/dc/elements/1.1/%22%20%20%20xmlns:cc=%22http://creativecommons.org/ns#%22%20%20%20xmlns:rdf=%22http://www.w3.org/1999/02/22-rdf-syntax-ns#%22%20%20%20xmlns:svg=%22http://www. - (concatenated) - 8%22%20standalone=%22no%22?%3E%3Csvg%20%20%20xmlns:dc=%22http://purl.org/dc/elements/1.1/%22%20%20%20xmlns:cc=%22http://creativecommons.org/ns 404 (Not Found)

### Steps to Reproduce
```
const loader = new PIXI.Loader();
loader.add('clock', 'img/clock.svg');
```

### Environment

- **`pixi.js` version**: 6.0.2
- **Browser & Version**: Brave & Firefox
- **OS & Version**: Fedora Linux 33
- **Running Example**: http://nikolas.us.to/~nik/pixi-example/

JS console error example:

![Screenshot_2021-05-07_15-01-31](https://user-images.githubusercontent.com/59292/117496505-25feec80-af45-11eb-9444-5e03856d30c4.png)

bigtimebuddy:
You example is a little complicated, would you mind making a minimal reproducible example? I think this might be a duplicate of #7460

nikolas:
Oh, yeah it is a duplicate of that ticket. Removing the XML prolog fixes this issue.

Here's the example SVG I'm using:
view-source:http://nikolas.us.to/~nik/pixi-example/img/clock.svg

And of course, pixi should be accepting these prologs.

nikolas:
Closing as duplicate of #7460

PR #7478, opened 2021-05-09
Fix SVG validation by supporting XML prolog and leading comments

### LINKED ISSUE #7460: PIXI.Loader cannot load SVG with XML prolog
### Current Behavior
Any SVG with an XML prolog (such as default exports from Boxy SVG) cannot be used with PIXI.Loader. The SVG data is preloaded via XHR and fails the SVGResource.test() funciton. 

SVGs with XML prologs can still be used as normal with PIXI.Texture.from() because in this case it is the SVG URL and not the file data that is passed to SVGResource.test().

### Possible Solution
Update SVGResource to accept SVGs with XML prologs. The test function will require an additional regex test and the load function will also require updates.

### Steps to Reproduce
Load any SVG file with an XML prolog (see below) included at the start of the file using the PIXI.Loader add function.

image

Tested on Pixi.js 6.0.2

SukantPal:
Fixes #7460, #7477

##### Description of change

`SVGResource.test` strips out the prolog and leading comments before checking for the first tag being `<svg>`.

##### Pre-Merge Checklist


bigtimebuddy:
What you're doing seems like a lot of work. Why not replace this:

```js
source.indexOf('<svg') === 0
```

with a more robust regexp:

```js
/^(<\?xml[^?]+\?>)?\s*(<!--[^(-->)]*-->)?\s*\<svg/gm.test(source)
```

This should match the following:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<!-- Hello -->
<svg>
  Hello world
</svg>
```

```xml
<?xml version="1.0" encoding="utf-8" ?>
<svg>
  Hello world
</svg>
```

```xml
<svg>
  Hello world
</svg>
```

SukantPal:
Yep, looks like I wasn't thinking right last night 😅. Simplified it now.

bigtimebuddy:
Why String.match instead of RegExp.test? Test returns Boolean, match returns an array.

packages/core/src/textures/resources/SVGResource.ts
```diff
@@ -245,9 +245,16 @@ export class SVGResource extends BaseImageResource
             // source is SVG data-uri
             || (typeof source === 'string' && (/^data:image\/svg\+xml(;(charset=utf8|utf8))?;base64/).test(source))
             // source is SVG inline
-            || (typeof source === 'string' && source.indexOf('<svg') === 0);
+            || (typeof source === 'string' && !!source.match(SVGResource.SVG_XML));
```
SukantPal, line 248:
```suggestion
            || (typeof source === 'string' && SVGResource.SVG_XML.test(source);
```

packages/core/src/textures/resources/SVGResource.ts
```diff
@@ -245,9 +245,16 @@ export class SVGResource extends BaseImageResource
             // source is SVG data-uri
             || (typeof source === 'string' && (/^data:image\/svg\+xml(;(charset=utf8|utf8))?;base64/).test(source))
             // source is SVG inline
-            || (typeof source === 'string' && source.indexOf('<svg') === 0);
+            || (typeof source === 'string' && SVGResource.SVG_XML.test(source);
```
SukantPal, line 248:
```suggestion
            || (typeof source === 'string' && SVGResource.SVG_XML.test(source));
```

packages/core/src/textures/resources/SVGResource.ts
```diff
@@ -245,9 +245,16 @@ export class SVGResource extends BaseImageResource
...
-            || (typeof source === 'string' && source.indexOf('<svg') === 0);
+            || (typeof source === 'string' && SVGResource.SVG_XML.test(source));
     }
 
+    /**
+     * RegExp for SVG XML document.
+     *
+     * @example &lt;?xml version="1.0" encoding="utf-8" ?&gt;&lt;!-- image/svg --&gt;&lt;svg
+     */
+    static SVG_XML = /^(<\?xml[^?]+\?>)?\s*(<!--[^(-->)]*-->)?\s*\<svg/gm;
```
SukantPal, line 256: Can't use global flag for shared RegExp @bigtimebuddy :(

```suggestion
    static SVG_XML = /^(<\?xml[^?]+\?>)?\s*(<!--[^(-->)]*-->)?\s*\<svg/m;
```

Zyie (approved):

bigtimebuddy:
Thanks @SukantPal!

PR #7495, opened 2021-05-16
Fix roundPixels

### LINKED ISSUE #7049: ROUND_PIXELS issues

dev7355608:
##### Description of change

Fixes #7049.

First off all, I consistently used `Math.round()` everywhere. `getBounds()` doesn't have access to the resolution, i.e., is independent of resolution, but we want it to match what is rendered. Also the renderer's resolution isn't necessarily the same resolution as render target's resolution when rendering to texture. So we can't really round to real pixels easily. Rounding world css pixels won't give crisp results when the resolution is non-integer, or when the camera transform isn't rounded, but that are problems that must be solved manually somehow. I also didn't like how the old pixel rounding used `settings.RESOLUTION` in Mesh, since it isn't necessarily the resolution of the current render target, so I got rid of that. Mesh rounded each point to the world grid, but that can cause [distortions](https://www.pixiplayground.com/#/edit/wm9D-u30IMoltS9hS9Q8j) in a way that is very likely not desired. So I round in local space w.r.t. the scaling factors of the world transform and round the translation part of the word transform.

```js
const { a, b, c, d } = worldTransform;
// extract scaling factors sx and sy
const sx = Math.sqrt((a * a) + (b * b));
const sy = Math.sqrt((c * c) + (d * d));

x = Math.round(x * sx) / sx;
y = Math.round(y * sy) / sy;
```

```js
worldTransform.tx = Math.round(worldTransform.tx);
worldTransform.ty = Math.round(worldTransform.ty);
```

Then each point lands on the world grid exactly, if no rotation is present in the world transform; and if there is rotation, it won't get distorted. So essentially rounding happens in between the scaling part and the rotation part of the world transform.

Added round pixels code (which was missing) to ParticleRenderer (webgl), Mesh (webgl, non-batching), Mesh (canvas2d).

Changed the roundPixels default in ParticleContainer to use `settings.ROUND_PIXELS` like everywhere else. It was `true` before and said something about improved performance, but that wouldn't be true in webgl, I suppose.

`CanvasRenderer.setContextTransform()` now just rounds tx/ty of the word transform (and not tx/ty of the entire context transform matrix with projection) to match the behavior in webgl mode.

One thing I haven't addressed yet: sx/sy are unnecessarily computed every time `CanvasMeshRenderer._renderDrawTriangle()` is called when it only needs to be computed once in `CanvasMeshRenderer.render()`.

canvas2d and webgl pixel rounding should give exactly the same results now.

##### Pre-Merge Checklist

### DIFF
--- a/packages/mesh/src/Mesh.ts
+++ b/packages/mesh/src/Mesh.ts
@@ -1,5 +1,5 @@
 import { Shader, State } from '@pixi/core';
-import { Point, Polygon } from '@pixi/math';
+import { Point, Polygon, Transform } from '@pixi/math';
 import { BLEND_MODES, DRAW_MODES } from '@pixi/constants';
 import { Container } from '@pixi/display';
 import { settings } from '@pixi/settings';
@@ -47,6 +47,7 @@ export class Mesh<T extends Shader = MeshMaterial> extends Container
     private vertexDirty: number;
     private _transformID: number;
     private _roundPixels: boolean;
+    private _roundPixelsID: number;
     private batchUvs: MeshBatchUvs;
 
     // Internal-only properties
@@ -151,6 +152,8 @@ export class Mesh<T extends Shader = MeshMaterial> extends Container
          * @private
          */
         this._roundPixels = settings.ROUND_PIXELS;
+
+        this._roundPixelsID = -1;
 
         /**
          * Batched UV's are cached for atlas textures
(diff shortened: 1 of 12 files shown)


ivanpopelyshev:
Yes, needs review, I'm a bit slow after Covid, and I dont understand this PR, yet

packages/mesh/src/Mesh.ts
```diff
@@ -314,8 +318,37 @@ export class Mesh<T extends Shader = MeshMaterial> extends Container
...
 
+        let verticesBuffer;
+        let vertices;
+
+        if (this._roundPixels)
+        {
+            verticesBuffer = this.geometry.buffers[0];
+            vertices = verticesBuffer.data;
+
+            this.calculateVertices();
```
GoodBoyDigital, line 329: perhaps this could be an issue as we are modifying the user data and rounding. Maybe better to copy to a roundsVerts buffer?

GoodBoyDigital (commented):
This looks cool! Left a couple of comments, would be good to see how perf is affected by these changes too!

ivanpopelyshev:
> would be good to see how perf is affected by these changes too!

Compared to canvas rendering - and i saw how skia does it - nope, shouldnt affect at all

hubol:
I'm very interested in this PR!

PR #7531, opened 2021-06-10
Added `tint(color, multiply)` method to ColorMatrixFilter.

miltoncandelero:
##### Description of change
Added `tint(color, multiply)` method to ColorMatrixFilter. It transforms the matrix to achieve an effect similar to the `tint` field of sprites. This was a feature discussed on slack.

##### Pre-Merge Checklist

(no ColorMatrixFilter specific tests exist)

### DIFF
--- a/packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
+++ b/packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
@@ -142,6 +142,30 @@ export class ColorMatrixFilter extends Filter
     }
 
     /**
+     * Sets each channel on the diagonal of the color matrix.
+     * This achieves a tinting effect.
+     *
+     * @param {number} color - Color of the tint. This is a hex value.
+     * @param {boolean} multiply - if true, current matrix and matrix are multiplied. If false,
+     *  just set the current matrix with @param matrix
+     */
+    public tint(color: number, multiply: boolean): void
+    {
+        const r = (color >> 16) & 0xff;
+        const g = (color >> 8) & 0xff;
+        const b = color & 0xff;
+
+        const matrix: ColorMatrix = [
+            r/255, 0, 0, 0, 0,
+            0, g/255, 0, 0, 0,
+            0, 0, b/255, 0, 0,
+            0, 0, 0, 1, 0,
+        ];
+
+        this._loadMatrix(matrix, multiply);
+    }
+
+    /**
      * Set the matrices in grey scales
      *
      * @param {number} scale - value of the grey (0-1, where 0 is black)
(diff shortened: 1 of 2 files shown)


packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
```diff
@@ -141,6 +141,30 @@ export class ColorMatrixFilter extends Filter
...
 
+    /**
+     * Sets each channel on the diagonal of the color matrix.
+     * This achieves a tinting effect.
+     *
+     * @param {number} color - Color of the tint. This is a hex value.
+     * @param {boolean} multiply - if true, current matrix and matrix are multiplied. If false,
+     *  just set the current matrix with @param matrix
+     */
+    public tint(color: number, multiply: boolean): void
```
bigtimebuddy, line 152: Multiply could probably be optional since `_loadMatrix` sets a default of `false`.
```suggestion
    public tint(color: number, multiply?: boolean): void
```

packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
```diff
@@ -141,6 +141,30 @@ export class ColorMatrixFilter extends Filter
         this._loadMatrix(matrix, multiply);
     }
 
+    /**
+     * Sets each channel on the diagonal of the color matrix.
+     * This achieves a tinting effect.
```
bigtimebuddy, line 146: Would be nice to explain why you'd use this instead of `tint` property found on most display objects? What's the advantage.

bigtimebuddy (approved):
Small comments, but otherwise I'm okay with this change.

packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
miltoncandelero: I followed other methods like `brightness`. But this can be done and makes more sense in false anyway

packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
miltoncandelero: Container doesn't have tint. Only sprites do. That's the whole idea behind this but I could make it more clear.

bigtimebuddy: Sprite, Text, Graphics, Mesh have some tint implementation. Some additional context here for users would be this tint could support any Container object.

packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
bigtimebuddy: I see that `saturate` its optional but most everything isn't. I think this should probably be optional for param for all that allow it. So I'd say make this change and we'll follow up with a fix for the typings here.

packages/filters/filter-color-matrix/src/ColorMatrixFilter.ts
miltoncandelero: Added a small text to explain why and when you should use tint.

ShukantPal (approved):
Nice!

bigtimebuddy:
Thank you @miltoncandelero!

Issue #7324, opened 2021-03-22
PIXI requires unsafe CSP


rubenlg:
### Expected Behavior
PIXI.js works with a strict CSP (e.g. `script-src 'self'`) by default.

### Current Behavior
PIXI.js requires an external module that patches it in order to work with strict CSP (`@pixi/unsafe-eval`). Not being safe by default opens a number of attack vectors on the application of the developer:
* Opens the door to XSS due to PIXI itself. Even if today your use of `new Function` is safe (has anyone conducted a security audit recently?), as code evolves, there is a significant risk that an XSS vulnerability will be introduced by accident.
* A relaxed CSP allowing unsafe eval can't be applied just to PIXI (as far as I know, let me know otherwise), so because of PIXI, the unsafe eval behavior has to be allowed on the entire app. While PIXI engineers can, to some extent, be confident in the security of the PIXI library, you can't be confident about any other code landing on the final application of the developer.

The whole point of CSP is to deal with situations in which you can't reliably ensure the quality of all the sources being pulled into your final application.

### Possible Solution
Make `@pixi/unsafe-eval` the default behavior of PIXI.js, and instead create an external module to make PIXI faster for those devs who need the extra speed **and** know what they are doing. Most likely, the majority of developers won't need the extra speed, but their end users will appreciate the extra security.

### Steps to Reproduce
Add a strict CSP meta tag to your HTML (or a header):
```html
  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self'">
```
Run a PIXI-based application.

### Environment

- **`pixi.js` version**: 5.3.8
- **Browser & Version**: Chrome 89
- **OS & Version**: Ubuntu 20.04.2
- **Running Example**: n/a

Related: #5509

ivanpopelyshev:
PixiJS is about performance. If we cant do `new Function()` , it eats away some JS perf.

> Even if today your use of new Function is safe (has anyone conducted a security audit recently?), as code evolves, there is a significant risk that an XSS vulnerability will be introduced by accident.

Security issue in any WebGL backend (chromium for example) is more likely than we introduce XSS here, and all renderers will be affected that way.

I participated in such bugs in chromium bugtracker.

anyway, Welcome to WebGL.

SukantPal:
Yeah, I don't think we'll switch this in the foreseeable future. A long term solution would be to generate uniform uploading functions at compile time.

rubenlg:
[...]

PR #7545, opened 2021-06-21
Fix extra new Function call in unsafe-eval

bigtimebuddy:
Addresses the suggestion here from @qtiki:  https://github.com/pixijs/pixijs/issues/7324#issuecomment-846383519

### Fixed

Installing `unsafe-eval` should not try to test CSP for `new Function` before patching. This can create an issue where reporting CSP will likely throw a false-positive, which we don't want. It should be implicit that users that are installing `unsafe-eval` package are _already_ opting into its performance trade-off and requiring an extra check to confirm it not helpful. 

I also cleaned up the readme to be clearer.


packages/unsafe-eval/src/install.ts
```diff
@@ -6,52 +6,33 @@ interface PIXICore {
     ShaderSystem: typeof ShaderSystem;
 }
 
-export function install(PIXI: PIXICore): void
+export function install({ ShaderSystem }: PIXICore): void
```
andrewstart, line 9: If you want people to be able to do `install({ ShaderSystem })` as in the readme, you'll have to change the argument type here from `: PIXICore` to `: { ShaderSystem: typeof ShaderSystem }` (or something similar, if that exact syntax is incorrect).

andrewstart: Nevermind, I expanded the code to see it is already that way.

Zyie (approved):

ivanpopelyshev (approved):

themoonrat (approved):

PR #7548, opened 2021-06-23
Adding mixins to @pixi/math

miltoncandelero:
##### Description of change
- Added mixins for all classes in @pixi/math
  - I'm not sure if this is the correct approach when adding mixins. Tell me if we should only mixin some of them.
- Loosened type restriction on the `Polygon` constructor
  - The `Polygon` constructor asked (in one of the overloads) for an array of `IPoint` but only used the `.x` and `.y`. Replaced the `IPoint` with an `IPointData` making it easier to feed the right object to that constructor.

##### Pre-Merge Checklist

  - Not sure how to write tests for mixins (or if possible at all)

---
cc: @ivanpopelyshev would this help you with projection? any change you need?

### DIFF
--- a/packages/math/global.d.ts
+++ b/packages/math/global.d.ts
@@ -1,0 +1,57 @@
+declare namespace GlobalMixins
+{
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface IPoint
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface IPointData
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface Point
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface ObservablePoint
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface Circle
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface Ellipse
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface Polygon
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface Rectangle
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface RoundedRectangle
+    {
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    interface Matrix
+    {
+    }
+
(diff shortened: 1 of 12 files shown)


bigtimebuddy:
Polygon type fix is 👍  and the GlobalMixin implementation here seems correct.

Overall, are you sure you need all these as mixins? I know you're going to work on a math-extras package, but I think we should only add the will actually need. We can always add more, if need be. Point, Rectangle seems like good candidates. Maybe Matrix. Specifically, I really don't think that `IPointData` needs a mixin. Since this is just X/Y object. Also `IPoint` is just a shared interface for `Point` and `ObservablePoint`. Would also question RoundedRectangle, Ellipse, Circle unless there are some specific convenience functions for those.

miltoncandelero:
I think @ivanpopelyshev might need the IPointData to add the Z coordinate to projection...

Anyway, I agree with the idea of only mixin in as needed and then expanding. Let's just make sure we cover mine and Ivan's use case for now.

ivanpopelyshev (approved):
Good enough, It will help me too

miltoncandelero:
@ivanpopelyshev which classes do you need. I'm asking so I can trim the mixins to only what we need for now.

ivanpopelyshev:
Point, ObservablePoint, IPointData. Though rectangle is good too, for pixi-layers and pixi-picture :)

ivanpopelyshev:
oh right Transform - that's a serious one, i need it )

miltoncandelero:
@andrewstart 
I guess you can always extend IPointData with an optional field

```
interface IPointData
{
    z?:number;
}
```

this allows for `z` to not be a thing while still being a thing 🤔 

--- 

@ivanpopelyshev for the initial work in `math-extras` I will extend `Point`, `ObservablePoint` and `Rectangle` so you will have rectangle for picture and layers 👍

bigtimebuddy:
I agree with @andrewstart. Allow for adding a Z prop mixin is going to have issues all over the place. I don't think that is the right approach for adding 2.5D or 3D. And it feels way outside the scope now for math-extras. When @ivanpopelyshev is ready, we can easily support that later, if IPointData is the right interface to mixin. Otherwise, that feels like scope creep. Unless Ivan want to describe more what he's planning...

ivanpopelyshev:
`z?` and some other things i already use in pixi-projection, the problem is - i used it  in old typings :) Now I need them in mixins.

Point, ObservablePoint, IPointData, Rectangle , Transform - that's what i need right now, not in the future :)

miltoncandelero:
Ok, only mixins for Point, ObservablePoint, IPointData, Rectangle, and Transform remain.

This covers both use cases, @ivanpopelyshev's  and mine's

bigtimebuddy (approved):
Nice!

ivanpopelyshev (approved):

PR #7559, opened 2021-06-28
Fix dependency design flaws

bigtimebuddy:
Addresses #7368

### Overview

This change converts the internal-PixiJS `dependencies` to `peerDependencies`. This has created serious issues in developers dependency graph _especially_ when users are upgrading versions of PixiJS. Often the only solution is users nuke their node_module and lock file to upgrade. Converting to peer dependencies should fix much of this pain.

Note: peerDependencies have been the standard usage in [pixijs/filters](https://github.com/pixijs/filters) for awhile now without issue.

Also, this change requires a special script to update the peerDependencies versions because Lerna ignores these when doing a version bump.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16722,7 +16722,7 @@
       "name": "@pixi/accessibility",
       "version": "6.1.0-rc.3",
       "license": "MIT",
-      "dependencies": {
+      "peerDependencies": {
         "@pixi/canvas-renderer": "6.1.0-rc.3",
         "@pixi/core": "6.1.0-rc.3",
         "@pixi/display": "6.1.0-rc.3",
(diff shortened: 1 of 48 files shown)


miltoncandelero (approved):
Looks good. It should improve usage in projects with npm 🤞

ShukantPal (approved):

ivanpopelyshev (approved):

ivanpopelyshev:
Now usual `"dependencies"` should be used only in bundle `package.json` , right?

GoodBoyDigital (approved):
nice one dude 👍

bigtimebuddy:
@ivanpopelyshev yeah, `dependencies` should continue to be used for the bundles

PR #7562, opened 2021-06-30
Wrap mipmap, scaleMode, wrapMode in getter/setter

### LINKED ISSUE #7550: BaseTexture wrapMode, scaleMode, mipmap should be getters/setters that increment dirtyStyleId
## Context

I was using `Graphics` with a texture fill and was surprised when the texture did not tile. I tried reproducing this with a lone `Graphics` but would not (unless you used very large coordinates - which was unrelated).

Screen Shot 2021-06-22 at 6 37 14 PM

Then I tried not rendering the existing tiles as sprites - and the problem went away. It turns out that `GraphicsGeometry` sets the `wrapMode` on the texture here:

https://github.com/pixijs/pixijs/blob/410a49c4f4e938d75a52a5129e98eb771780dc61/packages/graphics/src/GraphicsGeometry.ts#L509

However, if the texture has already been rendered once - the texture style would not be dirty. Since setting wrap-mode doesn't update the `dirtyStyleId`, the texture's wrap-mode wouldn't be updated in WebGL.

### Possible Solution

This can be fixed by

* making `wrapMode` a getter/setter that updates `dirtyStyleId` when you change it
* or calling `texture.baseTexture.update()` in `GraphicsGeometry`

I like the first option.

### Steps to Reproduce

In the following fiddle, I render the stage once with a single sprite (left). Then I add texture-tiling graphics (right) and re-render the scene.

https://jsfiddle.net/ShukantPal/dag0bhq9/

Screen Shot 2021-06-23 at 3 11 30 PM

The "Update texture" button will call `baseTexture.update()` and re-render the scene - and that fixes the problem:

Screen Shot 2021-06-23 at 3 11 52 PM

### Environment

- **`pixi.js` version**: 6.0.4
- **Running Example**: [https://jsfiddle.net/ShukantPal/dag0bhq9](https://jsfiddle.net/ShukantPal/dag0bhq9)

ShukantPal:
Fixes #7550 

Also: move documentation comments to properties

---
I'm back 👋

### DIFF
--- a/packages/core/src/textures/BaseTexture.ts
+++ b/packages/core/src/textures/BaseTexture.ts
@@ -49,33 +49,175 @@ export interface BaseTexture extends GlobalMixins.BaseTexture, EventEmitter {}
  */
 export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions> extends EventEmitter
 {
+    /**
+     * The width of the base texture set when the image has loaded
+     *
+     * @readonly
+     * @default PIXI.settings.RESOLUTION
+     */
     public width: number;
+
+    /**
+     * The height of the base texture set when the image has loaded
+     *
+     * @readonly
+     */
     public height: number;
+
+    /**
+     * The resolution / device pixel ratio of the texture
+     *
+     * @default PIXI.settings.RESOLUTION
+     */
     public resolution: number;
+
+    /**
+     * How to treat premultiplied alpha, see {@link PIXI.ALPHA_MODES}.
+     *
+     * @member {PIXI.ALPHA_MODES}
+     * @default PIXI.ALPHA_MODES.UNPACK
+     */
     public alphaMode?: ALPHA_MODES;
-    public mipmap?: MIPMAP_MODES;
+
+    /**
+     * Anisotropic filtering level of texture
+     *
+     * @member {number}
+     * @default PIXI.settings.ANISOTROPIC_LEVEL
+     */
     public anisotropicLevel?: number;
-    public scaleMode?: SCALE_MODES;
-    public wrapMode?: WRAP_MODES;
+
+    /**
+     * The pixel format of the texture
+     *
+     * @default PIXI.FORMATS.RGBA
+     */
     public format?: FORMATS;
+
+    /**
+     * The type of resource data
+     *
+     * @default PIXI.TYPES.UNSIGNED_BYTE
+     */
     public type?: TYPES;
+
+    /**
+     * The target type
+     *
+     * @default PIXI.TARGETS.TEXTURE_2D
+     */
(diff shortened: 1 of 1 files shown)


bigtimebuddy (commented):
For the setters, probably be worth checking for changes so dirty is only if there's a meaningful change?

ShukantPal:
Indeed, my bad - will add dirty checks.

packages/core/src/textures/BaseTexture.ts
```diff
@@ -381,6 +351,53 @@ export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions>
...
+     * @default PIXI.settings.MIPMAP_TEXTURES
+     */
+    get mipmap(): MIPMAP_MODES
+    {
+        return this._mipmap;
+    }
+    set mipmap(value: MIPMAP_MODES)
+    {
+        this._mipmap = value;
+        this.dirtyStyleId++;
```
bigtimebuddy, line 367:
```suggestion
        if (this._mapmap !== value)
        {
            this._mipmap = value;
            this.dirtyStyleId++;
        }
```

packages/core/src/textures/BaseTexture.ts
```diff
@@ -381,6 +351,53 @@ export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions>
...
+     * @default PIXI.settings.SCALE_MODE
+     */
+    get scaleMode(): SCALE_MODES
+    {
+        return this._scaleMode;
+    }
+    set scaleMode(value: SCALE_MODES)
+    {
+        this._scaleMode = value;
+        this.dirtyStyleId++;
```
bigtimebuddy, line 383:
```suggestion
        if (this._scaleMode !== value)
        {
            this._scaleMode = value;
            this.dirtyStyleId++;
        }
```

packages/core/src/textures/BaseTexture.ts
```diff
@@ -381,6 +351,53 @@ export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions>
...
+     * @default PIXI.settings.WRAP_MODE
+     */
+    get wrapMode(): WRAP_MODES
+    {
+        return this._wrapMode;
+    }
+    set wrapMode(value: WRAP_MODES)
+    {
+        this._wrapMode = value;
+        this.dirtyStyleId++;
```
bigtimebuddy, line 398:
```suggestion
        if (this._wrapMode !== value)
        {
            this._wrapMode = value;
            this.dirtyStyleId++;
        }
```

GoodBoyDigital (approved):
👍  nice

bigtimebuddy (approved):

PR #7563, opened 2021-06-30
Introduce Math Extras package (Vector2 Math, Rectangle operations, and more!)

miltoncandelero:
##### Description of change

- Methods for `Point` and `ObservablePoint`
  - addition
  - subtraction
  - multiplication
  - division
  - scalarMultiplication
  - dotProduct
  - crossProduct
  - normalized
  - magnitude
  - magnitudeSquared
  - projection
  - reflection
  - equals
 
- Methods for `Rectangle`
  - intersects
  - containsRect
  - equals
  - intersection
  - union

- Auxiliary pure functions
  - floatEqual
  - lineIntersection

Vector2 methods are mostly adapted from Three.js and Unity but translated to 2D (with `z` set to zero).  
Rectangle methods are mostly adapted from OpenFL and the old PR #4953 
All functions follow a similar approach as [DisplayObject's `toLocal` and `toGlobal`](https://pixijs.download/dev/docs/PIXI.DisplayObject.html#toGlobal) where you can pass a reference to an object to have the calculations be in place or ignore that parameter at all and receive a new object.  
Some typescript black magic was required, but this should make it so that the method respects the data type of the output object you give, saving you from having to cast after you chain a method.  
The two auxiliary pure functions solve two common problems (not related to the `Point` or `Rectangle` augmentation): [Float comparison](https://floating-point-gui.de/) and [Line intersection](http://paulbourke.net/geometry/pointlineplane/#:~:text=Intersection%20point%20of%20two%20line%20segments%20in%202%20dimensions)

##### Pre-Merge Checklist

- [ ] Documentation added

---

I haven't added documentation yet but I wrote all tests I could think of. (tests are not my forte)

I think this is good enough for a feature freeze. Do you think I missed any helper/utility/test case?


miltoncandelero:
huh, interesting. I saw those compile errors on my vscode but I could build fine so I attributed them to some IDE shenanigans. I will fix it tomorrow before starting with the docs.

eXponenta:
Very long namings.

A lot of libs use:
add, sub, div, addScalar, mulScalar, dot, cross, project, reflect.

Ayfri:
Hi, I have done a lot of these methods on my [`PIXI-Extended`](https://github.com/Ayfri/PIXI-Extended) library (see in the `src/maths` folder). Feel free to copy the code if needed ^^

andrewstart:
It looks like in globals.d.ts, your outPoints and outRects need to be marked as optional. Can the generic typing be changed to `T extends import('@pixi/math').IPointData = Point`, or will Typescript complain about that?

andrewstart:
Beyond Typescript complaining in the build errors (just needs to be cast as T, I think), this looks good to me.

bigtimebuddy:
My preference for method naming in PixiJS, and OOP general, is as verbs (divide, add, multiply, etc) not nouns (division, addition, multiplication, etc). Plus the verbs are generally shorter to @eXponenta's point above.

bigtimebuddy:
On the naming front, I'd also particularly avoid names like `sub`, `mul`, `div`, etc, as these things are not obvious or user-friendly: 1) for developers newer to these operations, and 2) for English-learners who might get tripped by these shortenings. Being descriptive but short is a good balance.

miltoncandelero:
> On the naming front, I'd also particularly avoid names like `sub`, `mul`, `div`, etc, as these things are not obvious or user-friendly: 1) for developers newer to these operations, and 2) for English-learners who might get tripped by these shortenings. Being descriptive but short is a good balance.

So, verbs but without abbreviations. Got it 👍

bigtimebuddy:
Another minor code suggestion here to keep from having assign to the prototypes for every function (which will save on space and make the code more readable), you can do something like this:

```js
const mixins = {
  add() {},
  multiply() {},
  subtract() {},
};

Object.assign(Point.prototype, mixins);
Object.assign{ObservablePoint.prototype, mixins);
```

Overall, the work you've done here is awesome @miltoncandelero. This is sure to be very useful for a lot of developers and it's nice to be able to offer an "official" version of these utils.

ivanpopelyshev:
to make less code, you can define an interface somewhere in module, and use it in `global.d.ts`, like `interface ObservablePoint extends MyInterface`. That'll remove code dupe

miltoncandelero:
> to make less code, you can define an interface somewhere in module, and use it in `global.d.ts`, like `interface ObservablePoint extends MyInterface`. That'll remove code dupe

Would documentation work with that? 🤔

miltoncandelero:
@bigtimebuddy 
> Another minor code suggestion here to keep from having assign to the prototypes for every function (which will save on space and make the code more readable), you can do something like this:
> 
> ```js
> const mixins = {
>   add() {},
>   multiply() {},
>   subtract() {},
> };
> 
> Object.assign(Point.prototype, mixins);
> Object.assign{ObservablePoint.prototype, mixins);
> ```


`webdocs` REALLY hates that pattern 

image

You can see three things: 
- the comments inside the mixin objects are not found on the documentation
- if I assign as variable: `prototype.add = mixin.add` it is interpreted as a variable
- only if I write a function it becomes a method `prototype.multiply = function multiply(....`

bigtimebuddy:
Try using `@memberof`?

miltoncandelero:
yep, @method and @memberof are kinda working. I am now trying to get more than one class inside `@memberof` so I can document Point and ObservablePoint in the same comment

bigtimebuddy:
If we need more out of webdoc for these uses, we can always ping @ShukantPal. JSDoc use to have `@alias` which could do multiple docs, so maybe a feature request?

packages/math-extras/src/pointExtras.ts
```diff
@@ -0,0 +1,143 @@
+import { Point, IPointData, ObservablePoint } from '@pixi/math';
+
+const mixins: any = {
+    /**
+     * Adds `other` to `this` point and outputs into `outPoint` or a new Point
+     * @method add
+     * @memberof PIXI.Point#
+     * @alias PIXI.ObservablePoint.add
```
miltoncandelero, line 8: @ShukantPal is there a way to have this jsdoc be parsed for both `Point` and `ObservablePoint` ?
`@alias` didn't work but maybe there is some other secret word to use?

ShukantPal:
@bigtimebuddy If we need it, I can implement @alias in webdoc; however:

* Why can't `ObservablePoint` just extend `Point`?
* Can we create an explicit mixin?

```
/** @mixin */
const PointLike = {
  /** Description */
  add() {
    // IMPL
  }
};

/**
 * @mixes PointLike
 */
class Point {

}

Object.assign(Point.prototype, PointLike.prototype);
```

packages/math-extras/src/pointExtras.ts
ShukantPal: Right now, no :)

ShukantPal: webdoc can't infer if the prototype assignment is a method or property. It doesn't infer `mixin.add`'s documentation either. I think an @alias feature request would be appropriate.

miltoncandelero:
Ok, I had to copy-paste some documentation but I got webdoc working.

The only thing left would be to copy-paste all that doc again to have IDE documentation support, however, this is not done for other mixins (see pictures, first `add` with no docs, but then `getChildByName` with no docs).
exibit a
exhibit b

Marking this as ready for review and if we need to copy-paste the docs I will do it 👍

bigtimebuddy:
Looks like you followed the same pattern as the other packages, so there should be no problem publishing and no rollup ch ages. This package should get published in the next release.

And to answer your question, this package should be an optional opt-in package like graphics-extras. So it is not included by default in the bundles.

bigtimebuddy (approved):
Overall, this is great work @miltoncandelero. Super pleased with the level of tests and documentation. I think as an optional package, this is complete enough to make a release and get more feedback.

My only small comment would be in graphics-extras we added a comment on the method only being available with the opt-in package... for instance [drawRegularPolygon](http://pixijs.download/dev/docs/PIXI.Graphics.html#drawRegularPolygon). I anticipate without this note, we'll likely be getting issues like "_someMethod_ is undefined!".

packages/math-extras/src/pointExtras.ts
```diff
@@ -0,0 +1,419 @@
...
+     * @param {IPointData} other - The point to multiply with `this`.
+     * @param {IPointData} [outPoint] - A Point-like object in which to store the value,
+     * optional (otherwise will create a new Point).
+     * @returns {IPointData} The `outPoint` reference or a new Point, with the component-wise multiplication.
+     */
+    multiply<T extends IPointData>(other: IPointData, outPoint?: T): T
+    {
+        if (!outPoint)
+        {
+            (outPoint as any) = new Point();
```
ShukantPal, line 104: This behavior of creating new objects for the destination deviates from the current behavior. Is there any specific reason why?

packages/math-extras/src/pointExtras.ts
```diff
@@ -0,0 +1,419 @@
...
+    {
+        /*
+             * Returns the magnitude of the vector that would result
+             * from a regular 3D cross product of the input vectors,
+             * taking their Z values implicitly as 0
+             * (i.e. treating the 2D space as a plane in the 3D space).
+             * The 3D cross product will be perpendicular to that plane,
+             * and thus have 0 X & Y components
+             * (thus the scalar returned is the Z value of the 3D cross product vector).
+             */
```
ShukantPal, line 217:
```suggestion
         * Returns the magnitude of the vector that would result
         * from a regular 3D cross product of the input vectors,
         * taking their Z values implicitly as 0
         * (i.e. treating the 2D space as a plane in the 3D space).
         * The 3D cross product will be perpendicular to that plane,
         * and thus have 0 X & Y components
         * (thus the scalar returned is the Z value of the 3D cross product vector).
         */
```

packages/math-extras/src/pointExtras.ts
```diff
@@ -0,0 +1,419 @@
...
+     * optional (otherwise will create a new Point).
+     * @returns {IPointData} The `this` on `onto` projection.
+     */
+    project<T extends IPointData>(onto: IPointData, outPoint?: T): T
+    {
+        if (!outPoint)
+        {
+            (outPoint as any) = new Point();
+        }
+        // Math says: a Projected over b = [(a·b) / (b·b)] * b;
```
ShukantPal, line 357:
```suggestion
        // Math says: a Projected over b = [(a·b) / (b·b)] * b; We normalize that to the length of onto.
```

packages/math-extras/src/pointExtras.ts
```diff
@@ -0,0 +1,419 @@
...
+        const scalarProjection = ((this.x * onto.x) + (this.y * onto.y)) / ((onto.x * onto.x) + (onto.y * onto.y));
```
ShukantPal, line 358: This actually doesn't do the formula in the comment above. It is actually `(a*b) / (b*b)` and it is the fractional length of the projection along `onto`.

packages/math-extras/src/pointExtras.ts
```diff
@@ -0,0 +1,419 @@
...
+
+        outPoint.x = onto.x * scalarProjection;
+        outPoint.y = onto.y * scalarProjection;
```
ShukantPal, line 361:
```suggestion
        const normalizedScalarProjection = ((this.x * onto.x) + (this.y * onto.y)) / ((onto.x * onto.x) + (onto.y * onto.y));

        outPoint.x = onto.x * normalizedScalarProjection;
        outPoint.y = onto.y * normalizedScalarProjection;
```

ShukantPal (changes requested):

packages/math-extras/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+{
+    if (!outPoint)
+    {
+        (outPoint as any) = new Point();
+    }
+
+    const denominator = (((bEnd.y - bStart.y) * (aEnd.x - aStart.x)) - ((bEnd.x - bStart.x) * (aEnd.y - aStart.y)));
+
+    // If lines are parallel or overlapping, the intersection can be nowhere or everywhere... NaN.
+    if (denominator === 0)
```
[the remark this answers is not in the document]
miltoncandelero: We could, however `settings.EPSILON` isn't a thing. I used `Math.EPSILON` for the`floatEqual` in `packages/math-extras/src/index.ts`
Since that function is from this same package, I think it would be good to use it 🤔

packages/math-extras/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    // Line intersection extends beyond the bounds of the segment.
+    if (!isLine && (ua < 0 || ua > 1 || ub < 0 || ub > 1))
+    {
+        outPoint.x = NaN;
+        outPoint.y = NaN;
+
+        return outPoint;
+    }
+
+    outPoint.x = aStart.x + (ua * (aEnd.x - aStart.x));
```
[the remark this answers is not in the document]
miltoncandelero: Agreed, will make the code a bit more readable

packages/math-extras/test/index.js
```diff
@@ -0,0 +1,951 @@
+const { ObservablePoint, Point, Rectangle } = require('@pixi/math');
```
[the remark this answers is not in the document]
bigtimebuddy: Yeah I know that. Someone will need to resolve conflicts. Should be easy tho.

packages/math-extras/src/pointExtras.ts
```diff
@@ -0,0 +1,419 @@
...
+     * @param {IPointData} other - The point to subtract to `this`.
+     * @param {IPointData} [outPoint] - A Point-like object in which to store the value,
+     * optional (otherwise will create a new Point).
+     * @returns {IPointData} The `outPoint` reference or a new Point, with the result of the subtraction.
+     */
+    subtract<T extends IPointData>(other: IPointData, outPoint?: T): T
+    {
+        if (!outPoint)
+        {
+            (outPoint as any) = new Point();
```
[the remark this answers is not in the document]
miltoncandelero: In short: Typescript is smarter than us and knows that _something that extends IPointData_ it is not exactly a `Point` and gets mad at the assignation.

Long story: `outPoint` is of type `T` and `T extends IPointData`. Since we have a check for falsy, the if only enters if you forgo or manually assign undefined to the `outPoint` parameter. By that point, you don't have a reference to write the result into so we create an object. (We can't create an object of type `T` because `T` might not have a constructor.)

Where is the catch then? You can purposefully do something awful like: `point.subtract<SomethingExtendingIPointData>(anotherPoint, undefined)`. Now your function really wants to return a `SomethingExtendingIPointData` reference but you feed undefined into the `outPoint`! We cant create a `SomethingExtendingIPointData` and `Point` wouldn't fit in there! That is what typescript is saying.

As any is a hack but I couldn't find a better way to explain typescript what's going on :(

packages/math-extras/src/pointExtras.ts
miltoncandelero: This allows for in-place operation if needed. Similar to the `toLocal` and `toGlobal` behaviour.
If the user needs every single drop of ram and performance, we use their provided object. If they don't care we create a new one.

packages/math-extras/src/pointExtras.ts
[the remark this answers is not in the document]
ShukantPal: This might be fixed by overloading:

```
subtract<T extends IPointData>(other: IPointData, outPoint: T): T
subtract(other: IPointData): Point
```

packages/math-extras/src/index.ts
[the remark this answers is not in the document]
ShukantPal: My bad, that's sounds good.

packages/math-extras/src/pointExtras.ts
miltoncandelero: The comment is lifted from the wikipedia page (https://en.wikipedia.org/wiki/Vector_projection) which has this image.
image

And we are actually doing that math, what I called `scalarProjection` is equal to `[(a·b) / (b·b)]` which is then multiplied by `b` in the next lines.

**However** I agree that that is definitely not the scalar projection as the scalar projection is (again from Wikipedia https://en.wikipedia.org/wiki/Scalar_projection#Definition_in_terms_of_a_and_b)
image

Resolving this and accepting your other suggestion to rename the variable as you are right and this would be the `normalizedScalarProjection` or _"the fractional length of the projection along `onto`"_

packages/math-extras/src/pointExtras.ts
miltoncandelero: The end result is not normalized. We just moved around some parameters to avoid 2 `sqrt` calculations that would cancel each other.

image

(see your other conversation for the long explanation, I  don't know how to tag it, sorry 😅 )

packages/math-extras/src/pointExtras.ts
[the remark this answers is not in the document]
miltoncandelero: I could swear I tried that but let me check again 🤔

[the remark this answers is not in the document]
miltoncandelero: Right, that is the problem: This is the implementation of the mixin. I can't do overloads here because this is a literal object that will be merged with the `prototype`.

On the `globals.d.ts` the overload is actually made.

packages/math-extras/src/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    bEnd: IPointData,
+    isLine: boolean,
+    outPoint?: T): T
+{
+    if (!outPoint)
+    {
+        (outPoint as any) = new Point();
+    }
+
+    const denominator = (((bEnd.y - bStart.y) * (aEnd.x - aStart.x)) - ((bEnd.x - bStart.x) * (aEnd.y - aStart.y)));
```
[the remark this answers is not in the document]
miltoncandelero: Ok, math ain't pretty and there is no way for me to do these super cool matrix drawings in code
https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line_segment

The idea is that to find the intersection we express both lines as beziers and find the two parameters needed for the beziers to touch.
To do that, we do some _funky vectory_ math and we end up with two formulas that share the denominator.

aaaand that's why it is called denominator :P

(tbh, I had it like that in my own projects and copypasted it. I had to investigate the name and relearn the formula 😅 )

ShukantPal (approved):
Nice work @miltoncandelero!

bigtimebuddy:
@miltoncandelero FYI, I made some superficial changes to the tests suites in the PR. Mostly to follow the TypeScript based format from #7593 but also separated into different tests suites to make it a little more manageable.

PR #7579, opened 2021-07-04
Expose IPoint in globals instead of Point / ObservablePoint

ivanpopelyshev:
Continuation of #7548

Its better to expose IPoint than Point/ObservablePoint because some of our generics use `IPoint` 

I dont see cases when we have to expose those classes.

### DIFF
--- a/packages/math/global.d.ts
+++ b/packages/math/global.d.ts
@@ -7,12 +7,7 @@ declare namespace GlobalMixins
     }
 
     // eslint-disable-next-line @typescript-eslint/no-empty-interface
-    interface Point
-    {
-    }
-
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
Currently Point mixin affects both Point and ObservablePoint, is that correct?

miltoncandelero:
> Currently Point mixin affects both Point and ObservablePoint, is that correct?

Yes, It will mixin the same code into both.

And about the `IPoint`, it was on my original plan to expose it but it was said that it would increase complexity and be redundant since it is only used for Point and ObservablePoint. https://github.com/pixijs/pixijs/pull/7548#issuecomment-866482307 🤔

PR #7581, opened 2021-07-05
Fix Text vertical gradient so that it handles overlapping lines

### LINKED ISSUE #7540: Text vertical gradient breaks when line height is smaller than font size
text-gradient

While a very small line height where characters overlap is expected to inevitably break the rendering, when the line height is marginally smaller and characters don't overlap, it would be desirable if the rendering doesn't break.

[Demo](https://pixijs.io/pixi-text-style/#%7B%22style%22%3A%7B%22fill%22%3A%5B%22white%22%2C%22red%22%5D%2C%22fontSize%22%3A86%2C%22lineHeight%22%3A66%2C%22padding%22%3A10%7D%2C%22text%22%3A%22Hello%5CnWorld%22%2C%22background%22%3A%22%23000000%22%7D)

ShukantPal:
Fixes #7540 

##### Description of change

This change will make the vertical gradient handle cases in which consecutive lines overlap. It will calculate the midpoint between the prior line's bottom and next line's top and check if they are in increasing order; if not, it will set the gradient stop at their midpoint (for both prior & next line). There are two cases for this - previous line's bottom and current line's top, current line's bottom and next line's top.

##### Proof

(before) [https://pixijs.io/pixi-text-style/#%7B%22s...](https://pixijs.io/pixi-text-style/#%7B%22style%22%3A%7B%22fill%22%3A%5B%22white%22%2C%22red%22%5D%2C%22fontSize%22%3A86%2C%22lineHeight%22%3A50%2C%22padding%22%3A10%7D%2C%22text%22%3A%22Hello%5CnWorld%22%2C%22background%22%3A%22%23000000%22%7D.)

| lineHeight=66px | lineHeight=50px |
| ----------------  | ----------------  |
| Screen Shot 2021-07-05 at 5 16 29 PM | Screen Shot 2021-07-05 at 5 15 50 PM | 

(after) https://www.pixiplayground.com/#/edit/MX5ftft1c0i5Fvb5U1mWz

| lineHeight=66px | lineHeight=50px |
| ----------------  | ----------------  |
| Screen Shot 2021-07-05 at 5 15 13 PM | Screen Shot 2021-07-05 at 5 13 51 PM |

### DIFF
--- a/packages/core/src/textures/BaseTexture.ts
+++ b/packages/core/src/textures/BaseTexture.ts
@@ -49,33 +49,175 @@ export interface BaseTexture extends GlobalMixins.BaseTexture, EventEmitter {}
  */
 export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions> extends EventEmitter
 {
+    /**
+     * The width of the base texture set when the image has loaded
+     *
+     * @readonly
+     * @default PIXI.settings.RESOLUTION
+     */
     public width: number;
+
+    /**
+     * The height of the base texture set when the image has loaded
+     *
+     * @readonly
+     */
     public height: number;
+
+    /**
+     * The resolution / device pixel ratio of the texture
+     *
+     * @default PIXI.settings.RESOLUTION
+     */
     public resolution: number;
+
+    /**
+     * How to treat premultiplied alpha, see {@link PIXI.ALPHA_MODES}.
+     *
+     * @member {PIXI.ALPHA_MODES}
+     * @default PIXI.ALPHA_MODES.UNPACK
+     */
     public alphaMode?: ALPHA_MODES;
-    public mipmap?: MIPMAP_MODES;
+
+    /**
+     * Anisotropic filtering level of texture
+     *
+     * @member {number}
+     * @default PIXI.settings.ANISOTROPIC_LEVEL
+     */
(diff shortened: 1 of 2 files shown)


ivanpopelyshev (approved):

bigtimebuddy:
This is definitely an improvement. 

The only choice I might question is this case:

image

It's kinda weird to split the difference here. My preference would probably be to put the stop at the baseline or the topline, not the mid-point. This way it sees more like an overlap.

Curious if other folks have thoughts though or if it's just me...

ShukantPal:
I'm open to splitting it at anypoint. So we want to split it at the top-point of the line below?

bigtimebuddy:
Probably? I generally view lines below as being on-top anyways. So that makes more sense. Would like to see how it look though.

ShukantPal:
Screen Shot 2021-07-08 at 2 12 18 PM

@bigtimebuddy The rationale for using the midpoint is this - maximizing the chance that the midpoint is actually between the lines. With 66px line-height, using top-line renders incorrectly compared to the (after) shown in original description.

bigtimebuddy:
You're right @ShukantPal. The other version is better.

bigtimebuddy (approved):

PR #7586, opened 2021-07-10
Add support for all texture formats and types

dev7355608:
##### Description of change

Added all remaining texture formats and types.

This should select the "best" (or only) valid internal format for each format-type combination. According to the webgl spec in most cases there is only one valid choice for the format-type combination. In the cases where there are multiple valid choices I selected the one that best matches the type.

##### Pre-Merge Checklist


ivanpopelyshev:
Only as a different package with some kind of mixin for TextureSystem.

That's what I'm afraid of:
https://github.com/playcanvas/engine/blob/d2eaf1d8da3ecd0f824fc14134610baed36dd6f2/src/graphics/graphics-device.js

We do not know all the cases for all those formats. We added texture-resource API so users can hack it themselves. I certainly do not want those ifs in default TextureSystem

dev7355608:
At least some of those combinations should be supported. Recently R32F, RGB32F, RGBA16F, and RGB16F was added (#7222, #7425), but not the other members of that family. It should be consistent and support complete subsets at least. Also confusing: there's `FORMAT.RED`, but it's only working with `TYPES.FLOAT`. Why not support R8 etc?

packages/core/src/textures/TextureSystem.ts
```diff
@@ -301,46 +301,273 @@ export class TextureSystem implements ISystem
...
+
         if (this.webGLVersion !== 2)
         {
             return;
         }
 
         const gl = this.renderer.gl;
 
-        if (texture.type === gl.FLOAT)
+        if (texture.type === TYPES.UNSIGNED_BYTE)
```
bigtimebuddy, line 312: This is crazy verbose. Can we do a map/dictionary here instead?
```js
glTexture.internalFormat = glTypeFormatLookup[texture.type][texture.format];
```

ShukantPal:
I like the overall idea - we should be feature complete for whatever feature we support.

packages/core/src/textures/TextureSystem.ts
dev7355608: Changed it. Let me know if you like it like this.

bigtimebuddy (approved):
I'm cool with this. Comprehensive and easier to read the code.

ivanpopelyshev:
Are those all possible combinations? If you are not sure, please expose `glTypeFormatLookup`, allow people to hack it

ivanpopelyshev:
For example, our blending modes can be changed inside StateSystem: https://github.com/pixijs/pixijs/blob/dev/packages/core/src/state/StateSystem.ts#L106

ShukantPal (changes requested):

packages/core/src/textures/TextureSystem.ts
```diff
@@ -1,13 +1,100 @@
...
 import { removeItems } from '@pixi/utils';
-import { MIPMAP_MODES, WRAP_MODES, SCALE_MODES, TYPES, SAMPLER_TYPES } from '@pixi/constants';
+import { FORMATS, MIPMAP_MODES, WRAP_MODES, SCALE_MODES, TYPES, SAMPLER_TYPES } from '@pixi/constants';
 
 import type { ISystem } from '../ISystem';
 import type { Texture } from './Texture';
 import type { IRenderingContext } from '../IRenderingContext';
 import type { Renderer } from '../Renderer';
 
+const glTypeFormatLookup: { [type: number]: { [format: number]: number } } = ('WebGL2RenderingContext' in self ? {
```
GoodBoyDigital, line 11: in the spirit of not making our class too large, would be cool to move this to an external file!

ShukantPal: Should it be in @pixi/constants?

ivanpopelyshev: No, its gl-related, so it should be in core. Making it the same way as in StateSystem will be fine.

ShukantPal: I fine with that - but @ivanpopelyshev everything in @pixi/constants _is_ gl-related.

packages/core/src/textures/TextureSystem.ts
```diff
@@ -299,47 +386,24 @@ export class TextureSystem implements ISystem
...
         }
-
-        // that's WebGL1 HALF_FLOAT_OES
-        // we have to convert it to WebGL HALF_FLOAT
-        if (texture.type === TYPES.HALF_FLOAT)
+        else
         {
-            glTexture.type = gl.HALF_FLOAT;
-        }
+            glTexture.internalFormat = glTypeFormatLookup[texture.type][texture.format];
```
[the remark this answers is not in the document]
dev7355608: Added fallback to `texture.format` if the lookup fails in any way.

packages/core/src/textures/TextureSystem.ts
dev7355608: I moved it to the an external file similar to the `StateSystem` example.

ShukantPal (approved):

GoodBoyDigital (approved):
lovely! 👍

dev7355608:
If I didn't miss anything, the table should be complete. I also added the webgl 1 table.

ivanpopelyshev (approved):
`glTypeFormatLookup` should be exposed for modding because I'm not sure that are all combination

dev7355608:
> `glTypeFormatLookup` should be exposed for modding because I'm not sure that are all combination

It's now available via `TextureSystem.internalFormats` (though it's protected like `StateSytem.blendModes`).

Not sure it is the best name for it.

packages/core/src/textures/utils/mapTypeAndFormatToInternalFormat.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Returns a lookup table that maps each type-format pair to a compatible internal format.
+ *
+ * @memberof PIXI
+ * @function mapTypeAndFormatToInternalFormat
+ * @private
+ * @param {WebGLRenderingContext} gl - The rendering context.
+ * @return {{ [type: number]: { [format: number]: number } }} Lookup table.
+ */
+export function mapTypeAndFormatToInternalFormat(gl: WebGLRenderingContextBase):
+    { [type: number]: { [format: number]: number } }
```
bigtimebuddy, line 13: Something like this might be a more specific format:
`{ [type: TYPES]: { [format: FORMATS]: number } }`

dev7355608: I tried that first but it got the following error message:
> An index signature parameter type cannot be a union type. Consider using a mapped object type instead.

But it seems to compile without issues nonetheless. I'm going to change it.

bigtimebuddy: Not that critical, just seeing if we could do something more specific.

ShukantPal: The correct way for that would be:

```ts
Record<TYPES, Record<FORMATS, number>>
```

bigtimebuddy: I vaguely thought there were string-only key restrictions on Record template. But I think it supports numbers/symbols too.

PR #7593, opened 2021-07-13
Convert unit-tests to TypeScript

bigtimebuddy:
## Changed

* Floss now uses `ts-node` to run tests, that means TypeScript tests!
* Converts all test files names to new naming convention `*.tests.ts`, to not confuse with src files and to make for easier global matching (e.g., eslint)
* Tests now run _without_ building first, it's about the same time, but a much easier workflow
* Remove bundle tests, these weren't meaningful because they only tested for global object from Rollup

## Fixed

* There were some copy-pasta of `new RenderTexture.create` that was wrong
* There was one use of `PIXI.Container` in a test, that was also wrong

### Example

Just to prove that the test run successfully: https://github.com/pixijs/pixijs/runs/3052743455#step:8:1107

### Future Work

* There is no type-checking on tests, any type checking is done in VSCode or other editor, we only transpile with floss. There are numerous type errors that should get resolved after this.
* ~~We should remove `PIXI` from the unit-test `describe`, not that useful anymore~~ **Fixed**
* ~~Code-coverage using `nyc` is busted using TypeScript need to figure what's up with that~~ **Fixed**

### Links

https://pixijs.download/feature/typescript-tests/coverage/index.html


ShukantPal (approved):
This is good - did you use a script to refactor the unit tests or by hand?

bigtimebuddy:
Mostly find and replace using RegExp in VSCode. And some command line bash to batch git mv operations.

GoodBoyDigital (approved):
❤️

Zyie (approved):

PR #7594, opened 2021-07-13
Add the option to explicitly specify the internal format of a texture

dev7355608:
##### Description of change

I added the optional `internalFormat` property for `BaseTexture` to be able to set the desired internal format of the texture explicitly. The automatically inferred internal format might not be what is needed in all cases. For example, it's not possible to create sRGB textures at the moment.

I needed to move the `INTERNAL_FORMATS` out of `@pixi/compressed-textures` and merge it with the newly-created `INTERNAL_FORMATS` in `@pixi/contants`.

##### Pre-Merge Checklist


packages/core/src/textures/BaseTexture.ts
```diff
@@ -87,6 +89,13 @@ export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions>
      */
     public anisotropicLevel?: number;
 
+    /**
+     * The internal format of the texture
+     *
+     * @member {PIXI.INTERNAL_FORMATS}
+     */
```
bigtimebuddy, line 96:
```suggestion
    /** The internal format of the texture */
```

ivanpopelyshev:
Is there any way, please, just any way to not move all those compressed constants to general constants? seoarate enum for them? Make internalFormat a number and not an enum?

dev7355608:
> Is there any way, please, just any way to not move all those compressed constants to general constants? seoarate enum for them? Make internalFormat a number and not an enum? I don't like how our constants are becoming bloat, both formats and internal formats :(

The compressed constants are currently exposed in `PIXI.INTERNAL_FORMATS`. Ideally they should be in `COMPRESSED_INTERNAL_FORMATS`, I suppose. But that can't be changed without breaking anything. And naming the normal internal formats anything other than `INTERNAL_FORMATS` would be silly. They are both internal formats and easily distinguishable by the `COMPRESSED` prefix. So I don't think it's wrong to lump them together.

Making `internalFormat` a number to solve this problem would be weird considering there are the `PIXI.INTERNAL_FORMATS`. Very counterintuitive as well as inconvenient.

bigtimebuddy:
I don't really get this. Can you please describe what problem is solved by introducing another path for explicitly defining the format? Who is this for?

ivanpopelyshev:
Usually format is set in texture resource, based on resource properties. People make custom texture resource, that's a learning experience.

This commit will make setting internalFormat for buffered resource easier, just `baseTexture.fromBuffer(...)` and then `baseTexture.internalFormat`. However, maybe its better to make more custom resources, like I originally wanted? Thank you for reminding me that.

I also had an idea that we can move many resources away from the core, like ArrayResource.

@dev7355608 what do you say about better Resource instead, and moving INTERNAL_FORMATS in separate package, with custom resource?

bigtimebuddy:
If this is mostly about buffer-related resources, why not provide a BaseTexture resource option for it for construction? The fact it's a writable member on BaseTextures means there's a lot that can go wrong state-wise. Also, the documentation for this member needs more context about _why_ you'd need to use this property not just _what_ it is.

ivanpopelyshev (changes requested):
Lets do it through resource

dev7355608:
I'm not very familiar with resources. How would one specify the internal format of a render texture with resources?

ivanpopelyshev:
Oh, render texture.. why do we need internal format for renderTexture? except floats?

dev7355608:
Why shouldn't I be able to choose the internal format of my render texture. I suppose the only really relevant use case would sRGB textures, because it changes how the colors are loaded/saved, and it's not just about the precision of the format.

ivanpopelyshev:
hm, then we might add sRGB as a mode for BaseTexture

ivanpopelyshev:
I just like it when code reflecting usage and not just GL constants/classes.  That's why I rejected an attempt to bring renderbuffers to pixi - they exist only internally and its difficult to explain them to high-level users.

dev7355608:
> I just like it when code reflecting usage and not just GL constants/classes. That's why I rejected an attempt to bring renderbuffers to pixi - they exist only internally and its difficult to explain them to high-level users.

I don't see how the `internalFormat` field would hurt anyone considering it's entirely optional and high-level users that don't want to don't have to concern themselves with it. Also format and type are already GL constants: there's no abstraction over it. I don't see why internal format wouldn't fit in as well. Why shouldn't I be able to set the internal format myself when I can enable mipmap modes which are also technical internal gl stuff to be honest. The only benefit of not having `internalFormat` is a few less constants.

> If this is mostly about buffer-related resources, why not provide a BaseTexture resource option for it for construction? The fact it's a writable member on BaseTextures means there's a lot that can go wrong state-wise. Also, the documentation for this member needs more context about _why_ you'd need to use this property not just _what_ it is.

The format and type are writable members as well and suffer from the same state-wise problem already. I think it's necessary to call `baseTexture.update()` or `baseTexture.dirtyId` after changing them. They should probably converted to getter/setter and increase `dirtyId` automatically. I agree the documentation should say that the `internalFormat` is optional and that pixi chooses an appropriate type automatically.

dev7355608:
> hm, then we might add sRGB as a mode for BaseTexture. Its a natural case, not like all those other formats :)

And adding sRGB support separately could be somewhat awkward. Add `SRGB`/`SRGBA` to `PIXI.FORMATS`? That would mean mixing formats and internal formats constants. No sure if that is desired. Because we already have `PIXI.INTERNAL_FORMATS` and pixi makes the distinction already; at least it looks it for someone from the from the outside looking in. And having `BaseTexture.srgb = true/false` doesn't look right either.

ivanpopelyshev:
Four years ago I opposed to "wow! core now is only X kilobytes!!!" which was the main point of moving out canvas and naming it "legacy". I had to grit my teeth and accept that paradigm in my heart. Now I see that other people are trying to put as much as possible into the core. Its painful :)

I dont see a a way to move types/formats away from core but at least we can dedicate separate package to all those formats. 

We could merge it and then refactor to separate package, but its almost `6.1.0` and I dont want to separate one PR from another.

ivanpopelyshev:
> And having BaseTexture.srgb = true/false doesn't look right either.

Thats why it might better to make it separate enum. Do you think we'll need extra value there? For example for premultipliedAlpha we changed it to alphaMode which has three values.

dev7355608:
> > And having BaseTexture.srgb = true/false doesn't look right either.
> 
> Thats why it might better to make it separate enum. Do you think we'll need extra value there? For example for premultipliedAlpha we changed it to alphaMode which has three values.

It would have to be something like `BaseTexture.colorSpace`, I imagine. 
```js
enum COLOR_SPACE {
    NONE = 0,
    SRGB = 1,
}
```

Problem is, it might still confuse people, because the default webgl framebuffer is linear (and there seems to be no way of changing that). But when they create a texture and see the color space property, and then set it to `SRGB`, because the image is in fact sRGB, the image is now output in linear space on the screen, because the framebuffer is linear and not sRGB. So if they don't convert from linear to sRGB somehow at the end, it's going to be displayed wrong. So they might really want `NONE`.

ivanpopelyshev:
> default webgl framebuffer is linear

all our framebuffers currently store sRGB color (gamma-ed). Because of that blending modes are all wrong, and shaders are working with sRGB.

Some images might be in different color space, so we need to convert them on upload, usually its done automatically, but there's also `createImageBitmap` we dont actually use by default - and its accepting color space as parameter.

I cannot tell you how to code that thing right now, because it requires research. Maybe while you research it, we'll find out that boolean is not enough, like it was with `alphaMode`.

The good things about enums - you can explain those modes in documentation and provide links to articles that explain them.

dev7355608:
There's only linear and sRGB. If we would want to support proper linear blending/sampling, we would need to swap everything out for sRGB textures (in texture/filter system etc) of course, but render in an separate sRGB framebuffer, and then blit it to default framebuffer. This would be nice feature to have. This could be a renderer parameter or global setting. And then extend the `BaseTexture.colorEncoding` like this.

```js
enum COLOR_ENCODING {
    AUTO = 0,  // linear RGB if renderer is in linear RGB mode, or sRGB if renderer is in sRGB mode (default)
    LINEAR = 1, // force linear RGB
    SRGB = 2, // force sRGB
}
```

ivanpopelyshev:
OK, thats an idea, however we dont use modes on renderer, usually its done through `settings` and assigning in constructor.

How alphaMode works:
1. source npm, let it stay npm
2. source npm, please premultiply it
3. source already pma, dont do anything

we can do the same with srgb - add a mode where we know source is in wrong colorspace

dev7355608:
I assume it's a little different with sRGB. It's about whether we choose to blend/sample etc incorrectly all the way through in sRGB, or correctly in linear mode by setting all our textures to sRGB. The sources are (should be) in any case in sRGB color space; the pixels values are the same. So it's not about the color space of the source being wrong, but if we want our shaders to work with linear RGB values or sRGB values.

packages/core/src/textures/BaseTexture.ts
```diff
@@ -87,6 +89,13 @@ export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions>
      */
     public anisotropicLevel?: number;
 
+    /**
+     * The internal format of the texture
```
GoodBoyDigital, line 93: lets add aline here that lets the user know this will be auto detected if not specified.

packages/core/src/textures/TextureSystem.ts
```diff
@@ -342,6 +342,11 @@ export class TextureSystem implements ISystem
                 glTexture.internalFormat = gl.RGB16F;
             }
         }
+
+        if (this.webGLVersion === 2 && texture.internalFormat)
```
GoodBoyDigital, line 346: just curious, but do internal formats only work with WebGL2? or are there just more available?

As this function is getting a bit larger perhaps we tweak to:

```
glTexture.internalFormat = texture.internalFormat || autodetectInternalFormat(texture) ;
```

GoodBoyDigital (changes requested):
this is cool! left a little feedback

bigtimebuddy:
The workaround without this PR would be do do something like this:

```ts
// Set an internalFormat
BaseTexture.prototype.internalFormat = null;

// Override the initTextureType method
renderer.texture._initTextureType = renderer.texture.initTextureType;
renderer.texture.initTextureType = function (texture, glTexture) {
  if (this.webGLVersion === 2 && texture.internalFormat) {
    glTexture.internalFormat = texture.internalFormat;
    return;
  }
  this._initTextureType(texture, glTexture);
};
```

ivanpopelyshev:
I remind that this thing is needed for sRGB and we need different solution for it anyway.

PR #7595, opened 2021-07-14
Group unit-tests together by package name

bigtimebuddy:
This change groups unit-tests together by package name. This should make it easier to identify the source of problems, especially where some tests names might overlap, in the case of mixins or canvas packages.

## Example Snippet

```
  @pixi/canvas-mesh
    NineSlicePlane
      ✓ should be renderable with renderTexture in canvas

  @pixi/canvas-renderer
    CanvasMaskManager
      ✓ should work on all graphics masks inside container
      ✓ should set correct transform for graphics
    CanvasRenderer
      ✓ should default context to rootContext
      ✓ should allow clear() to work despite no containers added to the renderer
      ✓ should update transform in case of temp parent

  @pixi/app
    Application
      ✓ should generate application
      ✓ register a new plugin, then destroy it
      ✓ should remove canvas when destroyed
      ✓ should not destroy children by default
      ✓ should allow children destroy
      resizeTo
        ✓ should assign resizeTo
        ✓ should force multiple immediate resizes
        ✓ should throttle multiple resizes (62ms)
        ✓ should cancel resize on destroy
        ✓ should resize cancel resize queue
        ✓ should resizeTo with resolution
        ✓ should resizeTo with resolution and autoDensity
```

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -3,7 +3,8 @@ import { execFileSync } from 'child_process';
 import chai from 'chai';
 import sinonChai from 'sinon-chai';
 import fs from 'fs';
-import { FlossOptions } from 'floss';
+import type { FlossOptions } from 'floss';
+import type { PackageResult } from './packages';
 
 // Support for global otpions
 declare global {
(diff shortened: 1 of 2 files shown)


ShukantPal (approved):
This will definitely make the unit tests more organized 🎉 !

Zyie (approved):

PR #7601, opened 2021-07-15
Fix texImage2D/3D internal format and type arguments

dev7355608:
##### Description of change

I identified two of places where `baseTexture.format` is used when it should be `glTexture.internalFormat`.

The change in `DepthResource` is also safe to make since the internal format is set properly now.

##### Pre-Merge Checklist


dev7355608:
I also noticed that sometimes `baseTexture.type` was used and other times `glTexture.type`. This wouldn't be an issue if it wasn't for `TYPES.HALF_FLOAT` being the webgl1 extension constant. So we need to use `glTexture.type` everywhere.

ivanpopelyshev (approved):

ShukantPal (approved):
Nice catch!

bigtimebuddy:
Thanks @dev7355608

PR #7603, opened 2021-07-15
Embed resource-loader package

ivanpopelyshev:
Embed resource-loader and signals.

There are two serious tricks here:

### LoaderResource

```ts
interface LoaderResource extends GlobalMixins.ILoaderResource

namespace LoaderResource {
   //enums
}

export { LoaderResource }
```

### Loader.add

```ts
export interface ILoaderAdd {
    (this: Loader, name: string, url: string, callback?: LoaderResource.OnCompleteSignal): Loader;
    (this: Loader, name: string, url: string, options?: IAddOptions, callback?: LoaderResource.OnCompleteSignal): Loader;
    // the rest of declarations
}

class Loader {
    /* jsdoc */
    add: ILoaderAdd;
}
```

### What is left?

Docs. We have to get through docs and check everything.


bigtimebuddy:
Could you bring over the [unit-tests](https://github.com/englercj/resource-loader/tree/release/3.x/test/unit) too?

bigtimebuddy:
## Important Context

I've been thinking about this task for awhile. I have already chatted with @englercj who gave his blessing for us to do this. We are _super grateful_ for resource-loader and everything it has done for Pixi, so please BIG THANKS to Chad. This is in no way a slight to him or his great work.

There are, however, a few current pain-points as it related to the resource-loader dependency that this PR was designed to help sort:

* Types do not play well, specifically we have an exception related to mini-signals. resource-loader v3 has stricter types, but mini-signals now also has types, so TypeScript generally defaults to the package types (more context: https://github.com/pixijs/pixi.js/pull/7045)
* We want to extend the metadata shape to pass in more options with TypeScript. This is not an ideal implementation atm (more context: https://github.com/pixijs/pixi.js/pull/6385)
* We have an effort underway to get Pixi to run in WebWorkers and Node.js. Since resource-loader references window it will not work in WebWorkers. We could change this to self but the tag-based loading will also not work. We are creating adapters for the environments for things like new Image(), but yeah, not an easy fix, and I’d like Loader to work with our adapter system.
* Documentation for loader continually alludes new users because it lives in another place and is not well integrated in our docs.
* Pixi long-term should probably own all the loading. Pixi already loads things outside resource-loader (like via PIXI.Sprite.from). I would like to harmonize this and use a single mechanism instead of two separate systems.
* @GoodBoyDigital  and I have chatted about wanting to take the API in another direction. Something that’s more Promise-based and less OOP. Again, probably makes sense for the project to own this bit more end-to-end.

I asked @ivanpopelyshev if he'd help with this effort, and he's done a great job here porting v3 to TypeScript and removing any dependencies.

ivanpopelyshev:
Of course in the two places I removed `!=` today were problems, uncovered thanks to the tests :)

bigtimebuddy:
@ShukantPal it's kind of a bummer that we don't support `@callback` in webdoc like typedefs and enums. Here's a case where documenting the callbacks for signals is useful.

bigtimebuddy:
Docs are now building @ivanpopelyshev 

https://pixijs.download/resource-loader/docs/index.html

bigtimebuddy (approved):

Zyie (approved):

ivanpopelyshev:
`ILoaderResource` -> `LoaderResource` is kinda  breaking change , for example for `pixi-spine`, but i think its better to break it now.

ivanpopelyshev:
OK, ready. new functionality can be added in next PR

ShukantPal (approved):

PR #7612, opened 2021-07-19
Fix minor documentation errors

bigtimebuddy:
There were a few private classes popping up in the docs, as well as some classes not part of `PIXI`.  This tidies this up.

Before: https://pixijs.download/dev/docs/index.html
After: https://pixijs.download/fix/docs-cleanup/docs/index.html

### DIFF
--- a/packages/basis/src/Basis.ts
+++ b/packages/basis/src/Basis.ts
@@ -89,7 +89,7 @@ export const BASIS_FORMATS_ALPHA: { [id: number]: boolean } = {
  * Binding to C++ {@code BasisFile} wrapper class.
  *
  * @see https://github.com/BinomialLLC/basis_universal/blob/master/webgl/transcoder/basis_wrappers.cpp
- * @ignore
+ * @private
  */
 export declare class BasisFile
 {
--- a/packages/core/src/shader/GLProgram.ts
+++ b/packages/core/src/shader/GLProgram.ts
@@ -1,5 +1,8 @@
 import type { Dict } from '@pixi/utils';
 
+/**
+ * @private
+ */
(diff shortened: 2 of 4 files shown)


packages/basis/src/Basis.ts
```diff
@@ -89,7 +89,7 @@ export const BASIS_FORMATS_ALPHA: { [id: number]: boolean } = {
  * Binding to C++ {@code BasisFile} wrapper class.
  *
  * @see https://github.com/BinomialLLC/basis_universal/blob/master/webgl/transcoder/basis_wrappers.cpp
- * @ignore
```
bigtimebuddy, line 92: @ShukantPal is this a webdoc regression? Seems like exporting a class doesn't exclude it from showing up. I had to use `@private` here, but I would expect `@ignore` to work too.

Zyie (approved):

packages/basis/src/Basis.ts
ShukantPal: Interesting - it might be related to using `declare`, but not sure.

bigtimebuddy: I also saw this behavior working on resource-loader PR. AsyncQueue.ts file, only `@private` worked not `@ignore` and this didn't have `declare`.

ShukantPal (approved):
@bigtimebuddy If you have time, please open an issue at webdoc with a repro :)

packages/basis/src/Basis.ts
bigtimebuddy: @ShukantPal https://github.com/webdoc-labs/webdoc/issues/132

PR #7618, opened 2021-07-20
Set sprite mask filter's resolution and multisample properly

dev7355608:
##### Description of change

Currently objects with a sprite mask are rendered without multisampling contrary to scissor and stencil masks, which is probably undesirable. With this change the sprite mask filter's resolution and multisample are set according to the current render target, and therefore the sprite mask no longer affects the quality of the image.

##### Pre-Merge Checklist


ivanpopelyshev:
I prepared special object to store those settings , `MaskData`

```js
container.mask = new MaskData(sprite);
container.mask.quality  = ...
```

In future that will allow to move out "prerender" phase where masks/filters bounds and settings are computed.

https://github.com/pixijs/pixijs/blob/dev/packages/core/src/mask/MaskData.ts

we can add quality there. Its bad that we dont have "AUTO" setting for qual;ity,but maybe we can just make it `null | MSAA_QUALITY` ?

bigtimebuddy:
Can you show a demo of this?

dev7355608:
Demo: https://www.pixiplayground.com/#/edit/pmHOJkBW8aJftM4QhRz_i

I think the default should be that the quality isn't affected. That would be what I expect from a mask. I didn't notice the `MaskData` class, but I agree that an option to turn off multisampling is nice to have since there are certainly objects that don't need it. Not sure, how to put this option into `MaskData` since it is a sprite-only thing. Maybe even expose all kinds of filter options including multisample and resolution with `maskData.filterOptions`.

bigtimebuddy:
Let's talk about default behavior...

Pixi is biased toward default behavior being the fastest. Performance vs quality, we tend to choose performance and allow developers to opt-in to quality. This is why antialias is off by default, why resolution is not devicePixelRatio, etc.

I'm okay changing default behavior but if it's a quality vs performance choice, we should choose perf.

dev7355608:
Ok, I changed the default to `NONE`. The same choice was made in the case of `cacheAsBitmapMultisample` after all.

I added `MaskData.resolution` and `MaskData.multisample`. Let me know what you think. I'm not sure if the default of `MaskData.multisample` should be `settings.FILTER_MULTISAMPLE` or `NONE`.

ivanpopelyshev:
I think FILTER_MULTISAMPLE as default is fine

ivanpopelyshev (approved):

ShukantPal (approved):

PR #7619, opened 2021-07-20
Allow non-RGBA8 multisampled framebuffers

dev7355608:
##### Description of change

The msaa buffer's internal format is currently hard-coded to RGBA8. With this change the msaa buffer's internal format is set to the same internal format of the first color texture.

Also the msaa buffer is no longer initialized and attached if the first color texture is not present.

##### Pre-Merge Checklist


ShukantPal (approved):
LGTM!

GoodBoyDigital (approved):
lgtm! 👍

SuperSodaSea (approved):
LGTM! @bigtimebuddy Maybe now is a good time to merge this old PR?

By the way, I think we can add support of multisampling for multiple render target in the future, instead of just for `i === 0`.

PR #7621, opened 2021-07-20
GitHub Actions Optimizations and Bug Fixes

bigtimebuddy:
Our Actions were taking awhile (9-12 minutes) mainly because we were building types every time. This is an expensive operation so I made some tweaks to the build system to streamline CI a little. Now it's ~5-7 minutes.

### Changed

* Building types _only_ happens before publishing and when doing a local release (just to be safe)
* Instead of building types, replaces workflow with a type-check (`-noEmit`)
* Test linting now happens as part of the coverage operation

### Fixed

* Lint step got removed from the build process, this adds it back to the `dist` script


ShukantPal (approved):
LGTM!

bigtimebuddy:
Also, the package.json scripts have become a mess. There are too many options and this needs to get cleaned-up. Will do that later.

ivanpopelyshev (approved):

PR #7622, opened 2021-07-20
Fix keys for multisampled textures in RenderTexturePool

dev7355608:
##### Description of change

I noticed two problems:

1. The pool wasn't returning multisampled fullscreen textures, but simply the normal fullscreen texture. I changed `SCREEN_KEY` and assigned `-1` to the normal fullscreen texture, `-2` to the 2-sampled texture, `-4` to the 4-sampled texture, and so on.
2. The bit operations of the key calculation can result in negative values, which not only clash with the solution of the first problem, but also with `key += multisample * 0x100000000`. This was fixed by casting it to an unsigned int after the bit operations.

##### Pre-Merge Checklist


ivanpopelyshev (approved):

GoodBoyDigital (approved):

dev7355608:
I forgot: object keys are always strings. So the screen-sized textures would never be destroyed on resize. Fixed that by casting the key to a number. Also added a test case.

ShukantPal (approved):

PR #7623, opened 2021-07-21
Fix bug in FramebufferSystem.forceStencil

dev7355608:
##### Description of change

This fixes a severe bug in `Framebuffer.forceStencil` that in combination with filters had the effect that stencil masks didn't work at all. `forceStencil` calls `enableStencil`, which invalidates the framebuffer. So when the first stencil mask is drawn, `forceStencil` is called, and the current framebuffer is invalidated. Then the filter textures are bound and afterwards the render target again. But since it was invalided, `updateFramebuffer` is called: the stencil buffer is reinitialized and therefore wiped.

Before the multisample fix PR `updateFramebuffer` failed to resize/update the stencil buffer. That bug canceled the bug in `forceStencil`. `forceStencil` must not invalidate the framebuffer, but only create and attach the stencil buffer if necessary. This was fixed by replacing `framebuffer.enableStencil()` by `framebuffer.stencil = true`.

I also made some non-critical adjustments to `resizeFramebuffer` and `updateFramebuffer`:

1. Made some changes to `resizeFramebuffer` to match with the behavior of `updateFramebuffer` more closely.
2. Cleaned the stencil buffer mess up in `updateFramebuffer`: the stencil buffer is now properly deleted if it isn't used anymore, and it's only created if necessary and no depth texture is attached.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/framebuffer/FramebufferSystem.ts
+++ b/packages/core/src/framebuffer/FramebufferSystem.ts
@@ -332,7 +332,14 @@ export class FramebufferSystem implements ISystem
 
         const colorTextures = framebuffer.colorTextures;
 
-        for (let i = 0; i < colorTextures.length; i++)
+        let count = colorTextures.length;
+
+        if (!gl.drawBuffers)
+        {
+            count = Math.min(count, 1);
+        }
+
+        for (let i = 0; i < count; i++)
         {
             const texture = colorTextures[i];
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):

ivanpopelyshev (approved):

bigtimebuddy:
Thank you @dev7355608

PR #7440, opened 2021-04-22
Add multisample shortcut and parameters

dev7355608:
##### Description of change

- Added getter and setter shortcut for `renderTexture.framebuffer.multisample`.
- Added multisample option to `RenderTexture.create()` and `RenderTexture`/`BaseRenderTexture` constructors.
- Added multisample parameter to `AbtractRenderer.generateTexture()`.

##### Pre-Merge Checklist

- [x] Documentation is changed or added
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #7629, opened 2021-07-21
Convert optional generateTexture args to options

bigtimebuddy:
Follow-up to #7440, converts the options for `generateTexture` into object instead of parameters. 

### Old Style

```ts
renderer.generateTexture(displayObject, scaleMode, resolution, region)
```

### New School

With the addition of `multisample` option, this should become an object, having more than a few options is quite cumbersome.

```ts
renderer.generateTexture(displayObject, { scaleMode, resolution, region, multisample });
```

This is much more flexible and allows you to only set a single parameter without having to add a bunch of undefineds.

```ts
renderer.generateTexture(displayObject, { multisample: MSAA_QUALITY.MEDIUM });
renderer.generateTexture(displayObject, { region: new Rectangle(100, 100, 200, 500) });
```

### DIFF
--- a/packages/core/src/AbstractRenderer.ts
+++ b/packages/core/src/AbstractRenderer.ts
@@ -44,6 +44,13 @@ export interface IRendererRenderOptions {
     skipUpdateTransform?: boolean;
 }
 
+export interface IGenerateTextureOptions {
+    scaleMode?: SCALE_MODES;
+    resolution?: number;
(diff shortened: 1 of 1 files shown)


ivanpopelyshev (approved):
**FINALLY**!

ShukantPal (approved):
👏

PR #7632, opened 2021-07-21
Add multisample property to Renderer

dev7355608:
##### Description of change

This adds the multisample property to `Renderer`, which returns the number of msaa samples of the canvas.

The documentation of `gl.getParameter(gl.SAMPLES)` is somewhat limited (at least I didn't find much with respect to the webgl canvas), but I tested it in Firefox and Chrome, and it works.

This is much better and much more convenient than the following, which I used in #7618 and proposed in #7441. Also since the `antialias` parameter is merely a hint that can be ignored by the browser the following code isn't even correct technically.

```js
if (renderer.gl.getContextAttributes().antialias) {
    multisample = MSAA_QUALITY.HIGH;
} else {
    multisample = MSAA_QUALITY.NONE;
}
```

Also the default framebuffer msaa samples at least on my machine in Firefox and Chrome is 4; so HIGH samples would have been to much in the code above.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/Renderer.ts
+++ b/packages/core/src/Renderer.ts
@@ -341,7 +341,30 @@ export class Renderer extends AbstractRenderer
 
     protected contextChange(): void
     {
-        const samples = this.gl.getParameter(this.gl.SAMPLES);
+        const gl = this.gl;
+
+        let samples;
+
+        if (this.context.webGLVersion === 1)
+        {
+            const framebuffer = gl.getParameter(gl.FRAMEBUFFER_BINDING);
+
+            gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+
+            samples = gl.getParameter(gl.SAMPLES);
+
+            gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
+        }
+        else
+        {
+            const framebuffer = gl.getParameter(gl.DRAW_FRAMEBUFFER_BINDING);
+
+            gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
+
+            samples = gl.getParameter(gl.SAMPLES);
+
+            gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, framebuffer);
+        }
 
         if (samples >= MSAA_QUALITY.HIGH)
         {
--- a/packages/core/src/Renderer.ts
+++ b/packages/core/src/Renderer.ts
@@ -14,7 +14,7 @@ import { ShaderSystem } from './shader/ShaderSystem';
 import { ContextSystem } from './context/ContextSystem';
 import { BatchSystem } from './batch/BatchSystem';
 import { TextureGCSystem } from './textures/TextureGCSystem';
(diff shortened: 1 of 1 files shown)


ivanpopelyshev (approved):

dev7355608:
`gl.getParameter(gl.SAMPLES)` gets the sample count of the current framebuffer. So I bind the default framebuffer first, and then rebind the framebuffer that was bound before.

packages/core/src/Renderer.ts
```diff
@@ -329,6 +339,51 @@ export class Renderer extends AbstractRenderer
...
+
+            samples = gl.getParameter(gl.SAMPLES);
+
+            gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
+        }
+        else
+        {
+            const framebuffer = gl.getParameter(gl.DRAW_FRAMEBUFFER_BINDING);
+
+            gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
```
GoodBoyDigital, line 362: Bit of duplicated code here! should we wrap the if WebGLVersion around this line only?

GoodBoyDigital (approved):
lgtm! only one suggestion

packages/core/src/Renderer.ts
dev7355608: `DRAW_FRAMEBUFFER` and `DRAW_FRAMEBUFFER_BINDING` are webGL2 only, and we don't want to touch the read buffer with webGL2. We would have to wrap around the second bind as well, but that would be more lines of code:

```js
let framebuffer;

if (this.context.webGLVersion === 1)
{
    framebuffer = gl.getParameter(gl.FRAMEBUFFER_BINDING);

    gl.bindFramebuffer(gl.FRAMEBUFFER, null);
}
else
{
    framebuffer = gl.getParameter(gl.DRAW_FRAMEBUFFER_BINDING);

    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
}

samples = gl.getParameter(gl.SAMPLES);

if (this.context.webGLVersion === 1)
{
    gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
}
else
{
    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, framebuffer);
}
```

Maybe like this assuming the constants would be undefined in webGL1:

```js
const framebuffer = gl.getParameter(gl.DRAW_FRAMEBUFFER_BINDING ?? gl.FRAMEBUFFER_BINDING);

gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER ?? gl.FRAMEBUFFER, null);

samples = gl.getParameter(gl.SAMPLES);

gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER ?? gl.FRAMEBUFFER, framebuffer);
```

ivanpopelyshev:
we dont actually need framebuffer re-binding there, just binding null is fine. If there are two libs on same context, its expected that they both reset webgl state before they start the frame.

PR #7633, opened 2021-07-22
Blit in Renderer.generateTexture

### LINKED ISSUE #7628: Multisample to generateTexture doesn't work without blit
Related to #7440 

That PR adds a multisample parameter to `generateTexture` on AbstractRenderer, however, it does not render without calling `blit`. This seems like a bad API choice, and we should handle the blit within `generateTexture`.

Works: https://jsfiddle.net/bigtimebuddy/ftd2caj9/
Doesn't work: https://jsfiddle.net/bigtimebuddy/1wmp06na/

@dev7355608 could you fix this?

dev7355608:
##### Description of change

Fixes #7628.

I probably made a mess overriding the method: I copied all the method definitions from `AbtractRenderer`. But I noticed that the new version `generateTexture` doesn't show up in the `Renderer` [docs](https://pixijs.download/dev/docs/PIXI.Renderer.html); only the deprecated version for some reason.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/Renderer.ts
+++ b/packages/core/src/Renderer.ts
@@ -16,15 +16,17 @@ import { BatchSystem } from './batch/BatchSystem';
 import { TextureGCSystem } from './textures/TextureGCSystem';
 import { RENDERER_TYPE } from '@pixi/constants';
 import { UniformGroup } from './shader/UniformGroup';
-import { Matrix } from '@pixi/math';
+import { Matrix, Rectangle } from '@pixi/math';
 import { Runner } from '@pixi/runner';
 import { BufferSystem } from './geometry/BufferSystem';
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Can you remove the duplicate docs? That seems redundant when all it seems like is needed is the override.

dev7355608:
Removed the docs, but something should be done so that `generateTexture` shows up in the docs of `Renderer` and not only the deprecated version. Maybe there's a better solution that copy & paste.

bigtimebuddy:
That's a bug with webdoc, I'll file an issue for it. Cc @ShukantPal

bigtimebuddy (approved):

GoodBoyDigital (approved):

bigtimebuddy:
Thanks @dev7355608

PR #7635, opened 2021-07-22
Bind empty if texture is invalid

dev7355608:
##### Description of change

Binding an invalid texture does nothing at the moment, i.e. the texture bindings don't change. Consequently, the last bound texture is used instead: [example](https://www.pixiplayground.com/#/edit/BLiI4gXywBluF0E6IJXJa). I assume this is not intended. It should bind an empty texture.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/TextureSystem.ts
+++ b/packages/core/src/textures/TextureSystem.ts
@@ -140,46 +140,38 @@ export class TextureSystem implements ISystem
     {
         const { gl } = this;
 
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):
👍

bigtimebuddy (approved):

bigtimebuddy:
Thanks for the fix @dev7355608

PR #7648, opened 2021-07-28
Add option for custom sprite mask filters

dev7355608:
##### Description of change

I created the option for custom sprite mask filters in order to allow any kind of masking textures and logic. I added the `filter` property to `MaskData` where the user can specify a custom filter to be used in the mask system.

Potentially problematic: a custom filter instance should not be set in multiple mask data with different sprites; if the objects are nested the `maskSprite` property would be overridden by the second push.

I also fixed a potential memory leak: the mask system's sprite mask filters never released the reference to the sprite in `popSpriteMask`.

##### Pre-Merge Checklist


bigtimebuddy:
Could you provide a demo of this work?

ivanpopelyshev:
I need something like that because i already have custom masks for pixi-projection  and pixi-picture. I'll look at it later )

dev7355608:
Demo: https://sprite-mask-filter.dev7355608.repl.co/

dev7355608:
The filter in `MaskData` is now stored wrapped in an array so that we can use it in `pushSpriteMask` and avoid repeated array creation there.

> Potentially problematic: a custom filter instance should not be set in multiple mask data with different sprites; if the objects are nested the `maskSprite` property would be overridden by the second push.

Fixed that by restoring the sprite of the current sprite mask filter in `MaskSystem.pop`. Added a test for this scenario as well.

dev7355608:
The mask system respects the enabled property of the custom filter now.

dev7355608:
If `enabled` is added to `MaskData`, should we respect the custom filter's `enabled` state? We don't do it for `resolution`/`multisample` and use whatever `MaskData` specifies. Maybe we should revert that last commit for consistency.

ivanpopelyshev:
Yes, revert it and i'll approve it

ivanpopelyshev (approved):

GoodBoyDigital (approved):

bigtimebuddy:
@dev7355608 could you resolve these conflicts?

dev7355608:
> @dev7355608 could you resolve these conflicts?

Done!

PR #7652, opened 2021-07-30
@pixi/events: removeEvents() now sets eventsAdded to false

miltoncandelero:
##### Description of change
`removeEvents` was not setting the `eventsAdded` back to false and thus, the first check on `addEvents` was returning early, making it impossible to re-add the events.

##### Pre-Merge Checklist


ShukantPal (approved):
Good find @miltoncandelero! Thanks for reviewing the events system.

bigtimebuddy (approved):

PR #7661, opened 2021-08-03
Bool uniform lazy upload

ivanpopelyshev:
This happened in test educational project im working on: bool is re-uploaded every time. 

Yes, we can add this check to ints too, we can auto-generate auto-generator for all the arrays, but for now I need a single bool :)

See this `uniform1i` every time? its a bool, and its the same every drawcall

image


packages/unsafe-eval/src/syncUniforms.ts
```diff
@@ -46,7 +46,14 @@ const GLSL_TO_SINGLE_SETTERS = {
...
-    bool(gl: IRenderingContext, location: WebGLUniformLocation, _cv: any, value: number): void { gl.uniform1i(location, value); },
+    bool(gl: IRenderingContext, location: WebGLUniformLocation, cv: any, v: boolean): void
+    {
+        if (cv !== v)
+        {
+            cv.v = v;
+            gl.uniform1i(location, Number(v));
+        }
+    },
     bvec2(gl: IRenderingContext, location: WebGLUniformLocation, _cv: any, value: number[]): void { gl.uniform2i(location, value[0], value[1]); },
```
ShukantPal, line 50: I guess we can do this optimization for all of these data types theoretically?

ShukantPal (approved):
Nice one

GoodBoyDigital (approved):
nice! yeah, we could apply this to them all if we want!

bigtimebuddy (approved):

PR #7664, opened 2021-08-03
Cleanup dependencies, don't require package order

bigtimebuddy:
This is a non-critical clean-up.

### Changes

* Removed internal `devDependencies`, since now tests are run against src instead of build files (previously dev dependencies were required by Lerna to build in a specific order)
* Removed the requirement on internal Lerna infrastructure to sort packages, since it's no longer that necessary (want to eventually upgrade or remove Lerna, this removes our dependencies more on this tool)
* Replaced `@lerna/project` with `workspaces-run` since this is more general and has typings and basically does the same thing


ivanpopelyshev:
remind me, why did we need to sort them?

bigtimebuddy:
We sorted the packages because the tests were consuming the built (output) packages. This meant that in order to run, say, tests in graphics, you needed to build core, display, etc first. Since converting unit-tests to TypeScript, this is no longer required.

Pi-Bouf (approved):

ivanpopelyshev (approved):

ShukantPal (approved):

PR #7669, opened 2021-08-04
Fix #7663 loader signal typings

### LINKED ISSUE #7663: Loader.OnCompleteSignal wrong type
The type of `Loader.OnCompleteSignal` suggests that a single LoaderResource object is provided in the callback.  In fact, it is a collection of resources.  It also appears that the comment for it is incorrect.
```
/**
 * When the loader starts loading resources it dispatches this callback.
 * @param loader - The loader that has started loading resources.
 */
export type OnCompleteSignal = (loader: Loader, resource: LoaderResource) => void;
```

### Environment

- **`pixi.js` version**: 6.1.0

ivanpopelyshev:
Subj.

Btw we really have to check if simple 404 goes to onError with correct arguments

### DIFF
--- a/packages/loaders/src/Loader.ts
+++ b/packages/loaders/src/Loader.ts
@@ -789,13 +789,13 @@ namespace Loader
      * When the loader starts loading resources it dispatches this callback.
      * @param loader - The loader that has started loading resources.
      */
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):
Thanks for fixing these two :+1:

Zyie (approved):

PR #7684, opened 2021-08-08
Remove CanvasRenderer types from mixin-cache-as-bitmap and accessibility

### LINKED ISSUE #7683: CanvasRenderer should not be a type dependency of accessibility and mixin-cache-as-bitmap
### Current Behavior
After commit 402ee9a has moved dependencies to peerDependencies, `@pixi/canvas-renderer` is only the dependency of `pixi.js-legacy` and will not be automatically installed by running `npm install pixi.js`. However `@pixi/accessibility` and `@pixi/mixin-cache-as-bitmap` are still using `PIXI.CanvasRenderer` as type.

https://github.com/pixijs/pixijs/blob/01a69a1465b875cb7b97a10caa9c35d9505397ff/packages/accessibility/src/AccessibilityManager.ts#L48
https://github.com/pixijs/pixijs/blob/01a69a1465b875cb7b97a10caa9c35d9505397ff/packages/mixin-cache-as-bitmap/src/index.ts#L29
https://github.com/pixijs/pixijs/blob/01a69a1465b875cb7b97a10caa9c35d9505397ff/packages/mixin-cache-as-bitmap/global.d.ts#L15

These error will not be reported when compiling typescript but running TypeDoc.

image

### Steps to Reproduce
Run `typedoc`

### Environment

- **`pixi.js` version**: 6.1.1

bigtimebuddy:
Fix #7683

### DIFF
--- a/packages/accessibility/package.json
+++ b/packages/accessibility/package.json
@@ -25,7 +25,6 @@
     "*.d.ts"
   ],
   "peerDependencies": {
-    "@pixi/canvas-renderer": "6.1.1",
     "@pixi/core": "6.1.1",
     "@pixi/display": "6.1.1",
     "@pixi/utils": "6.1.1"
--- a/packages/accessibility/src/AccessibilityManager.ts
+++ b/packages/accessibility/src/AccessibilityManager.ts
@@ -4,8 +4,7 @@ import { accessibleTarget } from './accessibleTarget';
 
 import type { Rectangle } from '@pixi/math';
 import type { Container } from '@pixi/display';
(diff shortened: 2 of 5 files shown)


ivanpopelyshev:
We have to check if it works with project that uses `pixi.js-legacy` :)

bigtimebuddy:
I think this'll be okay, but I'm fine doing a pre-release of this just to make sure.

ivanpopelyshev (approved):

Zyie (approved):

PR #7757, opened 2021-08-28
Remove StencilSystem._getBitwiseMask

dev7355608:
##### Description of change

The bit mask returned by `_getBitwiseMask` doesn't seem to make a lot of sense. Maybe it's a leftover from a previous implementation. So I removed it.

Bonus bug fix: `gl.clearStencil(0)` needs to be before `gl.clear(gl.STENCIL_BUFFER_BIT)` not after.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/mask/StencilSystem.ts
+++ b/packages/core/src/mask/StencilSystem.ts
@@ -56,7 +56,7 @@ export class StencilSystem extends AbstractMaskSystem
 
         // Increment the reference stencil value where the new mask overlaps with the old ones.
         gl.colorMask(false, false, false, false);
-        gl.stencilFunc(gl.EQUAL, prevMaskCount, this._getBitwiseMask());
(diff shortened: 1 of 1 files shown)


ShukantPal:
Why doesn't `_getBitwiseMask` make sense? Is it unnecessarily specific?

dev7355608:
> Why doesn't `_getBitwiseMask` make sense? Is it unnecessarily specific?

It doesn't serve any purpose at the moment. It's always large enough. It does the same thing as all the bits 1 mask. The bit mask gets one bit larger for each stencil mask on the stack; the values in the stencil buffer are just incremented/decremented though.

ShukantPal (approved):

GoodBoyDigital (approved):
yeah, this looks like it used to be when we had the super complex graphics objects (that rendered geometry using stencil buffer and triangle fan)

bigtimebuddy (approved):

bigtimebuddy:
Thanks @dev7355608 for making our codebase better.

PR #7759, opened 2021-08-29
Fix blit framebuffer

dev7355608:
##### Description of change

- I added the missing disposal of `blitFramebuffer`.
- If the current framebuffer doesn't have a color texture attachment, there's no reason to blit, because there's no target. Currently a color texture would be created and added to `blitFramebuffer`. That would lead to problems if the color texture was attached later for some reason. And nobody disposes of that created texture as well, since framebuffers don't dispose of their attachments.
- I added code that checks whether the color texture changed.
- Resizing `blitFramebuffer` was working before this change even though `resizeFramebuffer` was not called on it. It seems like it isn't necessary, but I changed it anyway for correctness. I didn't use the `resize` method, because `blitFramebuffer` doesn't own the color texture and therefore doesn't need to resize the color texture.

##### Pre-Merge Checklist


ivanpopelyshev (approved):

ShukantPal (approved):
Nice

bigtimebuddy:
Another good one @dev7355608 :+1:

PR #7760, opened 2021-08-29
Skip flush in renderAdvanced if the container doesn't have an enabled…

dev7355608:
##### Description of change

I think we can skip the flushes in `renderAdvanced` if the container doesn't have an enabled filter or enabled mask.

##### Pre-Merge Checklist


ShukantPal:
`renderAdvanced` should only be called when a filter or mask is active on the display object. I don't think we should have extra checks for that.

dev7355608:
> `renderAdvanced` should only be called when a filter or mask is active on the display object. I don't think we should have extra checks for that.

But `renderAdvanced` is also called if the mask and filters are disabled:
https://github.com/pixijs/pixijs/blob/7ee2be5cf88bb990166db10cfb017e37aa2cbbb1/packages/display/src/Container.ts#L586-L589

We could also improve the check in `render` instead.

ShukantPal:
I suppose that’d be the proper patch.

dev7355608:
Then the question is: do we move this block in `render`?
https://github.com/pixijs/pixijs/blob/7ee2be5cf88bb990166db10cfb017e37aa2cbbb1/packages/display/src/Container.ts#L618-L631

This would be a breaking change probably.

ShukantPal:
Hmm, never mind. I see what you're doing here. If there are no filters enabled and no masking is done, we could do an early return right?

dev7355608:
We could also move the flushes out of `renderAdvanced` and add them to `FilterSystem`/`MaskSystem`'s `push`/`pop`, which should make more sense as well. `Container` is currently flushing for the filter/mask system, but these systems can do it themselves.

dev7355608:
> Hmm, never mind. I see what you're doing here. If there are no filters enabled and no masking is done, we could do an early return right?

If all filters and the mask are disabled, the container could be rendered non-advanced and should not trigger flushes.

ShukantPal (approved):
Nice one. Any reason for being in draft?

dev7355608:
> Nice one. Any reason for being in draft?

I wasn't sure if this is the best solution. I'd prefer to do the flushing in the filter/mask system instead of in Container.

GoodBoyDigital (approved):

dev7355608:
There were two mistakes in my code:
1. I accidently moved the first flush after the filters are pushed.
2. We need to flush if `autoDetect` is enabled even if the mask type is `NONE`.

This check is very long and ugly. We should really let the mask/filter system do the flushing.

ShukantPal (approved):

PR #7769, opened 2021-09-02
Convert Filter resolution from property to accessor

mcdenhoed:
##### Description of change
This PR is an attempt to fix this issue that I've had with the PIXI Advanced Blur Filter (documented here: https://github.com/pixijs/filters/issues/301). I describe the issue in detail there, but here's a short summary:

`Filter` has a public property named `resolution`.  `AdvancedBloomFilter` overrides this property with an accessor named `resolution`. `AdvancedBloomFilter` maintains a private `_resolution` property which is set by this accessor (along with some other side-effects, which seems to be the whole purpose of using an accessor instead of a property).

However, as of https://github.com/microsoft/TypeScript/pull/37894, `tsc` now treats overriding a property with an accessor as an error.

This PR converts the definition of `resolution` in the base `Filter` class to an accessor to avoid this error.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/filters/Filter.ts
+++ b/packages/core/src/filters/Filter.ts
@@ -187,7 +187,6 @@ import type { Dict } from '@pixi/utils';
 export class Filter extends Shader
 {
     public padding: number;
-    public resolution: number;
     public multisample: MSAA_QUALITY;
     public enabled: boolean;
     public autoFit: boolean;
@@ -197,6 +196,8 @@ export class Filter extends Shader
      */
     public legacy: boolean;
     state: State;
+    
+    private _resolution: number;
     /**
      * @param {string} [vertexSrc] - The source of the vertex shader.
      * @param {string} [fragmentSrc] - The source of the fragment shader.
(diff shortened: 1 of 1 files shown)


packages/core/src/filters/Filter.ts
```diff
@@ -197,6 +196,8 @@ export class Filter extends Shader
      */
     public legacy: boolean;
     state: State;
+
+    private _resolution: number;
```
ShukantPal, line 200: As you note @mcdenhoed, `AdvancedBloomFilter` maintains a `_resolution` property already. Using the same property name and making it private will break that downstream dependency. Instead, let's make this `protected`. This will allow overriding accessors to use the same property.

```suggestion
    protected _resolution: number;
```

ShukantPal (changes requested):

packages/core/src/filters/Filter.ts
mcdenhoed: Ahh, yeah, that's a good catch. Thanks!

ShukantPal (approved):

bigtimebuddy (approved):

GoodBoyDigital (approved):

bigtimebuddy:
Thank you @mcdenhoed!

PR #7783, opened 2021-09-07
Move texture.valid check

owendwyer:
##### Description of change
Move
`if (!texture.valid) return; `
so that it comes before the call to 
`texture._frame.width `

This prevents a crash when destroying the texture of a sprite currently being rendered.

##### Pre-Merge Checklist


bigtimebuddy:
Seems reasonable. Do you think you could add a unit-test for this change?

ivanpopelyshev (approved):

ivanpopelyshev:
So, after this, we have invariant "sprite texture can be destroyed, sprite just wont show anything". This thing was born because somehow WebGL implementation allows it. I thought that it doesnt :)

owendwyer:
Hope this test is okay.  And thanks a lot!

bigtimebuddy (approved):

bigtimebuddy:
Thanks @owendwyer :tada:

PR #7786, opened 2021-09-10
Reformat StateSystem.ts documentation

ShukantPal:
##### Description of change

#7159


ivanpopelyshev (approved):

packages/core/src/state/StateSystem.ts
```diff
@@ -12,89 +12,79 @@ const DEPTH_TEST = 3;
...
 const DEPTH_MASK = 5;
 
-/**
- * System plugin to the renderer to manage WebGL state machines.
- *
- * @class
- * @extends PIXI.System
- * @memberof PIXI
- */
+/** System plugin to the renderer to manage WebGL state machines. */
```
bigtimebuddy, line 15: This requires `@memberof PIXI` or else it will not show up in the correct namespace.

https://pixijs.download/chore/docs-StateSystem/docs/StateSystem.html

ShukantPal:
Ready for review! @bigtimebuddy

packages/core/src/state/StateSystem.ts
```diff
@@ -15,86 +15,80 @@ const DEPTH_MASK = 5;
 /**
  * System plugin to the renderer to manage WebGL state machines.
  *
- * @class
- * @extends PIXI.System
- * @memberof PIXI
+ * @namespace PIXI
```
bigtimebuddy, line 18: Should be memberof not namespace

bigtimebuddy (changes requested):

packages/core/src/state/StateSystem.ts
ShukantPal: Fixed, sorry!

bigtimebuddy (approved):

PR #7807, opened 2021-09-19
Fix Texture.from not accepting a BaseTexture as texture source

### LINKED ISSUE #7746: Texture.from doesn't work with BaseTexture
Hi, I was reading **PIXI.Texture** documentation and in the description of `from()` method I've found that we should be able to create a new **Texture** from **BaseTexture**, but unfortunately, this is not possible.

Is it mistake in documentation or we should fix the method to support this option?

To be honest, in my opinion, this option is not super useful as we could just call `new PIXI.Texture(baseTexture)
`
### Expected Behavior
According to documentation (https://pixijs.download/dev/docs/PIXI.Texture.html), using `PIXI.Texture.from `we should be able to create a new **Texture** from **BaseTexture**

### Current Behavior
Currently, we are not able to create new **Texture** from **BaseTexture** using `PIXI.Texture.from` method.
The method is creating a new **BaseTexture** and using passed **BaseTexture** as **Resource**.

### Steps to Reproduce
1) create **BaseTexture**
2) try to create a new **Texture** from created BaseTexture (by method `PIXI.Texture.from`)

example: https://codesandbox.io/s/pixi-texture-from-base-sxvmc

### Environment

- **`pixi.js` version**: 6.1.2

ShukantPal:
##### Description of change

Fixes #7746. Makes a special case for BaseTexture sources.


ivanpopelyshev (approved):
One day we'll remove all those "cacheId" things into separate package.

bigtimebuddy (approved):

bigtimebuddy:
Thanks @ShukantPal

PR #7808, opened 2021-09-19
Fix uniforms sync for primitives and cache all array uniforms properly

ShukantPal:
##### Description of change

Builds on #7661


ShukantPal:
Note that setting `cv.v` was a mistake - `v` meant nothing on the value being cached. Instead, I created a `cu` variable that points to the uniform data "object" holding the value, and then set `cu.value` to update the cached value. See `IUniformData` for reference.

ivanpopelyshev (approved):

GoodBoyDigital (approved):
nice!

Issue #7826, opened 2021-09-27
GeometrySystem's shader compatibility check is unnecessary and restrictive


dev7355608:
It says here that the geometry must have all attributes that the shader requires. But that's not true, see https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/vertexAttrib. So technically the entire check is unnecessary, and it prevents one from using `vertexAttrib`. This limitation is easily bypassed by patching `PIXI.GeometrySystem.prototype.checkCompatibility = function () { };` though.

https://github.com/pixijs/pixijs/blob/bf8f1edfb5da55133d581498760017c1ab886026/packages/core/src/geometry/GeometrySystem.ts#L225-L244

PR #7827, opened 2021-09-27
Fix GeometrySystem.getSignature

dev7355608:
##### Description of change

The signature currently does not include something very essential: the location where the attribute is bound. So two different VAOs can end up with the same signature, which can end very badly. For example, two shader with two attributes in different locations:

```glsl
// signature: g-XXX-aAttribute1-aAttribute2

layout(location = 0) in vec2 aAttribute1;
layout(location = 1) in vec3 aAttribute2;
```
```glsl
// same signature: incompatible VAO is used, does not render correctly

layout(location = 1) in vec2 aAttribute1;
layout(location = 0) in vec3 aAttribute2;
```

Also when the shader-geometry compatibility check is disabled (#7826), and, for example, a geometry with only one attribute `aAttributeB` is used, this example can fail as well:

```glsl
// signature: g-XXX-aAttributeB

attribute vec2 aAttributeA; // location 0, because "aAttributeA" < "aAttributeB"
attribute vec3 aAttributeB; // location 1
```
```glsl
// the signature is the same as above, but that VAO has aAttributeB is in location 1 and is incompatible

attribute vec3 aAttributeB; // location 0
```

##### Pre-Merge Checklist


ivanpopelyshev (approved):

ShukantPal (changes requested):
Can you add a unit test based on the example you provide in the PR's description?

dev7355608:
> Can you add a unit test based on the example you provide in the PR's description?

I cannot add a unit test right now that would fail, because the first example needs #7828 to be merged first, and the second example requires `GeometrySystem.checkCompatibility` to be removed first (#7826).

I made a change to the second example, because it was technically not a case where it would fail: attributes are bound in lexicographical ascending order at the moment and didn't choose the attribute names correctly when I was writing the example.

dev7355608:
Added the unit tests.

Pi-Bouf (approved):

GoodBoyDigital (approved):
👍

bigtimebuddy (approved):

bigtimebuddy:
Fixes #7976

PR #7828, opened 2021-09-27
Fix attribute location

dev7355608:
##### Description of change

Not sure why the attribute location are manually bound instead of using the locations the compiler/linker chose (`getAttribLocation`). Shaders with layout location specifiers do not work because of this (layout location specifiers cannot be overridden): https://www.pixiplayground.com/#/edit/GleYmYB2SrVtbDyw5Tlhv.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/shader/utils/generateProgram.ts
+++ b/packages/core/src/shader/utils/generateProgram.ts
@@ -34,6 +34,25 @@ export function generateProgram(gl: IRenderingContext, program: Program): GLProg
     program.attributeData = getAttributeData(webGLProgram, gl);
     program.uniformData = getUniformData(webGLProgram, gl);
 
+    // GLSL 1.00: bind attributes sorted by name in ascending order
+    // GLSL 3.00: don't change the attribute locations that where chosen by the compiler
+    //            or assigned by the layout specifier in the shader source code
+    if (!(/^[ \t]*#[ \t]*version[ \t]+300[ \t]+es[ \t]*$/m).test(program.vertexSrc))
+    {
+        const keys = Object.keys(program.attributeData);
+
+        keys.sort((a, b) => (a > b) ? 1 : -1); // eslint-disable-line no-confusing-arrow
+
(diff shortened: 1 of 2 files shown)


ShukantPal (approved):
LGTM. This may have some side effects we can't see right now.

ivanpopelyshev (approved):
Wait, we didnt have `getAttribLocation` in pixi core, at all?

ivanpopelyshev:
Yes, agree, we need this for specifiers. btw in WebGPU they are specified in shader, always.

GoodBoyDigital:
Just throwing this out there.. but would be good to understand what the use cases for manually setting the locations might be?

dev7355608:
> Just throwing this out there.. but would be good to understand what the use cases for manually setting the locations might be?

We can force the locations to a specific layout and make sure two different shaders work with the same VAO. But this only works at the moment if both shaders use the same attributes. If one uses just a subset, pixi creates two different VAOs, but one of those VAOs could be used for both shaders. But pixi doesn't detect that at the moment.

```glsl
layout(location = 0) in vec2 aAttribute1;
layout(location = 1) in vec3 aAttribute2;
```
```glsl
layout(location = 1) in vec3 aAttribute2;
```

If pixi detected that, this could solve multiple VAO problem mentioned here, though it requires the user to setup the locations correctly themselves, but this is very easy to do. Automatically modifying the locations at run time is probably a very messy.
https://github.com/pixijs/pixijs/blob/cf828cf40c36a6d8cd2af7c4a79287238f3437b9/packages/core/src/geometry/GeometrySystem.ts#L160-L164

dev7355608:
Attributes locations of GLSL 1.00 programs now bound by pixi again (as before) to guarantee the same location layout for shaders with the same active attributes. So this change affects only GLSL 3.00 programs now.

GoodBoyDigital (approved):
nice, yeah I like that this only affects GLSL 3.00!

bigtimebuddy:
Thanks @dev7355608

PR #7847, opened 2021-10-04
Fix setting precision of shader

dev7355608:
##### Description of change

`setPrecision` always prepended a precision statement, because the shader name was prepended first. I'm not sure which statement would define the precision of the shader in this case. Also we never entered the second branch in `setPrecision`, which could leave a `highp` in there that isn't supported. So I fixed that. `setPrecision` also potentially changed highp integers to mediump in the second branch, but the documentation only mentions floats, so I changed that as well.

```glsl
precision highp float;
#define SHADER_NAME pixi-shader-3
precision lowp float;

attribute vec2 aVertexPosition;
attribute vec2 aVertexPosition2;

uniform mat3 projectionMatrix;
uniform mat3 translationMatrix;

void main() {
    gl_Position = vec4((projectionMatrix * translationMatrix * vec3(aVertexPosition + aVertexPosition2, 1.0)).xy, 0.0, 1.0);
}
```

I also improved the `#version` detection in the program constructor. I split the sources in two parts and insert the shader name and precision statement in-between. Shaders with version statement are no longer handled differently than shaders without.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/shader/Program.ts
+++ b/packages/core/src/shader/Program.ts
@@ -75,32 +75,26 @@ export class Program
         this.vertexSrc = this.vertexSrc.trim();
         this.fragmentSrc = this.fragmentSrc.trim();
 
-        const splitRegex = /^(?:.|\n|\r)*?[ \t]*#[ \t]*version[ \t].*(?:\r\n|\n|\r|$)/;
-        let vertexSrcHead = this.vertexSrc.match(splitRegex)?.[0] ?? '';
-        let vertexSrcBody = this.vertexSrc.substring(vertexSrcHead.length);
-        let fragmentSrcHead = this.fragmentSrc.match(splitRegex)?.[0] ?? '';
-        let fragmentSrcBody = this.fragmentSrc.substring(fragmentSrcHead.length);
-
-        name = name.replace(/\s+/g, '-');
-
-        if (nameCache[name])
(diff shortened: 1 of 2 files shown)


ivanpopelyshev (approved):

ivanpopelyshev:
Possible alternative approach: https://github.com/pixijs/tilemap/blob/7972c60b1154e6cfc5f1344300706f69b3b386eb/src/tilemap.frag#L2-L6

The important thing is, whatever we do, people will have to check how their stuff works on meidump because if uniform is highp in one shader and mediump in another, it wont link.

ivanpopelyshev (changes requested):
actually, no, I cant accept that.

Currently, all shaderws with `#version` wont be modified.
My concern is that there has to be a way to use shader as user wants, without our preprocessor. 

@GoodBoyDigital is doing new preprocessor API, for several months already - then we'll be able to make better changes to the algo.

Without this invariant I cant accept this thing.

dev7355608:
I see, and a new preprocessor API sounds good. Do you want me to revert the `#version` part, but keep the changes to `setPrecision`?

ivanpopelyshev:
Yep. If there's `#version` - we dont modify shader at all.

ShukantPal (changes requested):
Can this be tested?

PR #7878, opened 2021-10-15
feat: [7877] Allow loading SVG with viewBox only

zero41120:
##### Description of change
- Update width/height `SVGResource`'s error handler.
- Instead of reporting 0 and abort, fallback to parse the base64 and get `viewBox`'s width and height
- If viewBox is also not available, then abort.
- Added 2 unit test
  - Get size from `SVGResource.parseWidthHeightFromViewBox(svgString)` 
  - Get scaled size from `SVGResource.parseWidthHeightFromViewBox(svgString, size)` 
image

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/resources/SVGResource.ts
+++ b/packages/core/src/textures/resources/SVGResource.ts
@@ -116,6 +116,8 @@ export class SVGResource extends BaseImageResource
                 resolve(this);
             };
 
+            const svgString = this.svg;
+
             // Convert SVG inline string to data-uri
             if ((/^\<svg/).test(this.svg.trim()))
             {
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
@zero41120 thanks so much for this contribution and apologies it's taken so long to get a resolution on. I had a chat with @GoodBoyDigital about this and I'm going to close this PR.

This is to workaround a browser specific issue, which is to infer width/height from viewBox on Firefox. Firefox already helpfully provides a warning for this case. The easiest fix, that doesn't rely on any code changes, is to add width and height to your SVG to prevent the warning. We don't think Pixi should be handling invalid data in this case, especially when it adds a lot of complexity to the SVGResource. It's better to nudge developers to do the right thing with their assets and make sure they are valid (even if one browser has more strict requirements).

If someone needs to process lots of SVGs that are formatted this way (without width/height), I'd suggest a node.js script or Webpack plugin or some other build-environment solution which can parse the SVG correctly.

Issue #7855, opened 2021-10-08
Nested masked objects break when going outside viewport (in certain conditions)


jmlee2k:
The included playground demo explains it best, but to sum it up, when the following conditions are true, the mask on container C will be disabled when it moves out of the viewport:

 * scissor masking is enabled
 * masked container (C) is a child of a masked object (P)
 * the parent container (P) contains at least one other object

This feels somewhat similar to #5974 (which was fixed a long time ago) but is obviously not exactly the same case.

### Expected Behavior
Mask will never disable
### Current Behavior
Mask disables when object leaves viewport

### Possible Solution
Disable scissor masking

### Steps to Reproduce
https://www.pixiplayground.com/#/edit/2qJVNoyRHDVPMtaTJ4hCY

### Environment
- **`pixi.js` version**: 6.0.0 and up
- **Browser & Version**: Tested in Windows 10 on FF, Chrome, and Edge, all have the same issue
- **Running Example**: https://www.pixiplayground.com/#/edit/2qJVNoyRHDVPMtaTJ4hCY

jmlee2k:
I've tried it on a few windows machines, here's how it looks:
![pixi6-mask-issue](https://user-images.githubusercontent.com/7396545/136977370-db9a480b-af84-474f-965d-08cf0d1b52fe.gif)

robertswan:
This also impacts all iPhones with iOS 15 (so all the latest iPhone 13s and anyone else upgrading their OS).
Confirmed with the running example link above across various iOS 15 vs non 15 iphones.

eXponenta:
@robertswan 
can you provide a 100% reproduced bug?

iOS 15.1b4 (dev) not reproduced:
https://youtu.be/VhTxJptVs4Q

You should check a your bug on unstable version of iOS too, because stable (haha) ios 15 is not stable for some devices.
Same and with Safari 15, i know a lot of hates around it.

jmlee2k:
@eXponenta - Sure, it makes sense that it was introduced in Safari 15.  In terms of Windows, I've tried disabling ANGLE, and the same result occurs (at least in Firefox), so maybe more than one issue is at play here.

ivanpopelyshev:
> > @robertswan can you provide a 100% reproduced bug?
> 
> My apologies as I wish I could. Our products are complicated and I don't have source for the games that are bugged. I might try hacking in the PR changes manually. I'm also working across language and time zone barriers while managing this.
> 
[...]

PR #7895, opened 2021-10-20
Do not render elements in empty masks

ivanpopelyshev:
V6 fix for #7855

This is more permanent solution: API allows to remove whole subtree from render. Scissor handler calculates whether intersection of scissor rect and screen is 0, in that case it doesnt bother scissor at all, evading w=0,h=0 bug on iOS and ANGLE with integrated video.

There will be new tests for this.

Bugged: https://www.pixiplayground.com/#/edit/2qJVNoyRHDVPMtaTJ4hCY
Fixed: https://www.pixiplayground.com/#/edit/NXMa4bNf_v9iVfX0IxNKb

Can be tested on iOS15 or any integrated videocard with ANGLE

### DIFF
--- a/packages/core/src/mask/MaskSystem.ts
+++ b/packages/core/src/mask/MaskSystem.ts
@@ -100,7 +100,7 @@ export class MaskSystem implements ISystem
      * NOTE: The batch renderer should be flushed beforehand to prevent pending renders from being masked.
      *
      * @param {PIXI.DisplayObject} target - Display Object to push the mask to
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
@GoodBoyDigital what do you think about MaskHandler? 

Maybe add a runner like that somewhere? Im sure we need third param because that shows that `Container` is ready to handle the result `DONT_RENDER`, because several people copied implementations of `renderAdvanced` and i want backwards compatibility for them.

ivanpopelyshev:
OK, agree, i'll just make easier separate PR tomorrow

ivanpopelyshev:
Closed in favor of #7896

PR #7941, opened 2021-11-02
Fix: Split addChild arguments into two to prevent rest type errors.

### LINKED ISSUE #7938: AddChild in DisplayObject doesn't allow spreading anymore
Hello, after updating to 6.2.0 I'm experiencing this behavior.

### Expected Behavior

When using a spread operator to add to a DisplayObject like
`const containers = [new Container(), new Container()]`
`a.stage.addChild(...containers)`
I expect no ts issues
### Current Behavior

Currently getting
A spread argument must either have a tuple type or be passed to a rest parameter.ts(2556)

### Possible Solution

Possibly it is due to this change in addChild

addChild<T extends DisplayObject>(...children: [T, ...DisplayObject[]]): T; (6.2.0)
addChild<T extends DisplayObject[]>(...children: T): T[0]; (6.1.3) 

### Environment

- **`pixi.js` version**:  6.2.0
- **OS & Version**:  Mac OS Big Sur

ShukantPal:
Fixes #7938

### DIFF
--- a/packages/display/src/Container.ts
+++ b/packages/display/src/Container.ts
@@ -142,46 +142,42 @@ export class Container extends DisplayObject
      * @param {...PIXI.DisplayObject} children - The DisplayObject(s) to add to the container
      * @return {PIXI.DisplayObject} The first child that was added.
      */
-    addChild<T extends DisplayObject>(...children: [T, ...DisplayObject[]]): T
+    addChild<T extends DisplayObject>(child: T, ...additional: DisplayObject[]): T
     {
         // if there is only one argument we can bypass looping through the them
-        if (children.length > 1)
+        if (additional.length > 0)
         {
             // loop through the array and add all children
-            for (let i = 0; i < children.length; i++)
+            for (let i = 0; i < additional.length; i++)
             {
                 // eslint-disable-next-line prefer-rest-params
-                this.addChild(children[i]);
+                this.addChild(additional[i]);
             }
         }
-        else
+
+        // if the child has a parent then lets remove it as PixiJS objects can only exist in one place
+        if (child.parent)
         {
-            const child = children[0];
-            // if the child has a parent then lets remove it as PixiJS objects can only exist in one place
-
-            if (child.parent)
-            {
-                child.parent.removeChild(child);
-            }
-
-            child.parent = this;
-            this.sortDirty = true;
-
-            // ensure child transform will be recalculated
-            child.transform._parentID = -1;
-
-            this.children.push(child);
-
-            // ensure bounds will be recalculated
-            this._boundsID++;
-
-            // TODO - lets either do all callbacks or all events.. not both!
-            this.onChildrenChange(this.children.length - 1);
-            this.emit('childAdded', child, this, this.children.length - 1);
-            child.emit('added', this);
+            child.parent.removeChild(child);
         }
 
-        return children[0];
+        child.parent = this;
+        this.sortDirty = true;
+
+        // ensure child transform will be recalculated
+        child.transform._parentID = -1;
+
+        this.children.push(child);
+
+        // ensure bounds will be recalculated
+        this._boundsID++;
+
+        // TODO - lets either do all callbacks or all events.. not both!
+        this.onChildrenChange(this.children.length - 1);
+        this.emit('childAdded', child, this, this.children.length - 1);
+        child.emit('added', this);
+
+        return child;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


eXponenta:
why not use method overriding?
```
addChild<T extends DisplayObject>(child: T): T
addChild<T extends DisplayObject>(...additional: DisplayObject[]): T
addChild<T extends DisplayObject>(....args: any):T {

}
```

something like this.
this save previous implementation.

This is type bug that not should affect onto method body.

ShukantPal:
@eXponenta The first signature would be redundant. Passing in a non-T DisplayObject as the first child would be legal as well - and that's what we're trying to solve.

bigtimebuddy (changes requested):
This still does not fix the error that was report. [See playground](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAiCWEAO4CeB5ARgKwKbABdoBvAXwFgAoUSGAYQHsA7AseJ3AJ2lwA8DcTACYwEyNFjyESVaHOhghQugAt4IIQB4AKj36CRcRCjAYc+AgD4AFMDUaAXNG0AaaADpPiofALxmYCBOYiZmUgQA2gC6AJROusSy8smcuAQArpxM0HbqQgDcSXIUlCU0UNAAykicvrh6AsL0zKzsXCQlVMDMEETdLGwc3AC80BwA7tCMA22c1jGF1D1EqRDpIAQADNCj-a1D7t6qefOL-b3Qq+sEAIw7OS2DXIdKxxrWE1U1dfMLXcuXXBrDYAJnueyenBeynsQg+uEmIQk5kI8zcnyRpkkFjRYwRRnEWJRBF+ZwBVw2AGZwY9ZtC3nDPtVagJcRjjMjwmz8ZiwjiYn9KP8mBdchpUtlRhEmd9WTF0fjmT8YlEySKVkDrgAWGkzA5HWHWTzuMVCCULIA).

Part of this is a user typing issue. For doing spread arguments like this, you need to provide a more specific type.
```ts
// Doesn't work
const children = [new Sprite(), new Sprite()];
// Works
const children: [Sprite, Sprite] = [new Sprite(), new Sprite()];
```
I suggest we go back to the original type version for addChild but 1) throw an error if the children is empty, and 2) document it.

ShukantPal:
I guess this won't solve the problem then. Should we revert to the original signature?

bigtimebuddy:
@leonardoraele and @MercurialUroboros what do you think we should do?

eXponenta:
We can use nonEmptyArray declaration
https://gcanti.github.io/fp-ts/modules/NonEmptyArray.ts.html

MercurialUroboros:
> We can use nonEmptyArray type, and prevent call function without arguments.
> 
> https://gcanti.github.io/fp-ts/modules/NonEmptyArray.ts.html

The current implementation already mutated towards a non-empty array solution as far as I can see, there are different implementations of it, but it shouldn't change the end result.

I by far prefer Option A because option B can get ugly if the array has more than 3 elements.

I don't have any more useful ideas so far to prevent empty arguments but still be spreadable, will think later today 🤔

bigtimebuddy:
I think reverting is the best option and I made a PR for that #7945

MercurialUroboros:
> I think reverting is the best option and I made a PR for that #7945

Yup, sorry can't think of anything else, perhaps typescript might improve the inferring in these scenarios, or even better, better typings.

bigtimebuddy:
Going to close this in favor of reverting.

leonardoraele:
@bigtimebuddy Maybe revert to the signature but update the body of the method to validate the arguments (i.e. do nothing in case of empty array) so there is no runtime error.

leonardoraele:
Not sure if it is a satisfactory solution, but as of Typescript 3.4, you can add `as const` to the end of the expression to convert it into a tuple. This solves the issue.
```typescript
const containers = [new Container(), new Container()] as const;
app.stage.addChild(...containers);
```

bigtimebuddy:
In general, Pixi has taken a position to not validate or gracefully handle side-effects like this in production. One example (can't find the PR) was a user who accidentally set `Infinity` for the radius of a circle in Graphics. This crashed the browser. We don't put up guard-rails for stuff like this because it introduces a lot of case checking and bloat. For our development version we can leave in asserts and these are stripped out if you use the minified version.

I'd be more inclined to assert here than silently abort for an empty array. IMO we should not allow accepting an empty array for this function. That should be the responsibility of the developer to handle.

Great suggestion using `as const`, wasn't aware of that one!

MercurialUroboros:
Tha `as const` can be useful, but it also transforms the array in a read-only property, that would disallow functions that mutate the array.

Not sure how I feel about that, so far I never had the need to mutate an array that I already made part of an `addChild` .. but as a rule of thumb I don't like being limited 👀

PR #7969, opened 2021-11-07
Add state and blendMode property to MaskData

dev7355608:
##### Description of change

Added `state` and `blendMode` properties to `MaskData` for sprite masks.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/mask/MaskData.ts
+++ b/packages/core/src/mask/MaskData.ts
@@ -1,6 +1,7 @@
-import { MASK_TYPES, MSAA_QUALITY } from '@pixi/constants';
+import { BLEND_MODES, MASK_TYPES, MSAA_QUALITY } from '@pixi/constants';
 import { settings } from '@pixi/settings';
 import { ISpriteMaskFilter } from '@pixi/core';
+import { State } from '../state/State';
 
 import type { Rectangle, Matrix } from '@pixi/math';
 import type { IFilterTarget } from '../filters/IFilterTarget';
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Thanks @dev7355608. Could you explain more about why you'd need to provide a blendMode for a MaskData? Not sure I understand the use-case visually.

dev7355608:
@bigtimebuddy Sprite masks are currently always blended with the NORMAL blend mode. A workaround would be this:
```js
object.mask = new PIXI.MaskData(sprite);
object.mask.filter = new PIXI.SpriteMaskFilter();
object.mask.filter.blendMode = PIXI.BLEND_MODES.ADD;
```
Before custom sprite mask filters it was impossible to use a mask in this case, and you would have needed to add a filter directly.
```js
object.filters = [new PIXI.SpriteMaskFilter(sprite)];
object.filters[0].blendMode = PIXI.BLEND_MODES.ADD;
```
If the mask filter's `resolution`/`multisample`/`enabled` are overridden the corresponding properties of `MaskData` , then so should `state`/`blendMode`:
```js
object.mask = new PIXI.MaskData(sprite);
object.mask.blendMode = PIXI.BLEND_MODES.ADD;
```

bigtimebuddy:
@dev7355608 could you provide a visual example of this?

dev7355608:
Imagine any scenario where you want to blend an object with anything other than NORMAL, but also want to apply a sprite mask to the object. The default filter of the sprite mask is always blended with NORMAL, but you might want to MULTIPLY after the sprite mask was applied to the object. As I mentioned above it's possible to work around this by using a custom filter and setting the blend mode of it. But all other filter properties except the state are part of mask data already, so I think, for this reason, it makes sense to add the `state` (and `blendMode`) to `MaskData` as well.

Zyie:
Hey,

PR #7974, opened 2021-11-08
Set PREFER_ENV default to WEBGL2 for all devices (#7899)

### LINKED ISSUE #7899: PREFER_ENV default
The doc says that webgl2 is disabled by default for all *non-apple* mobile devices, but it's actually disabled for all mobile devices.

Also the referenced bug has been fixed two years ago. Maybe we can remove this exception for mobile devices?

https://github.com/pixijs/pixijs/blob/234f39d998a20e7928772a784e664c0b7c0b9c02/packages/core/src/settings.ts#L5-L21

dev7355608:
##### Description of change

Fixes #7899.

##### Pre-Merge Checklist


bigtimebuddy:
I think this change is sensible, but there are a handful of PixiJS developers that need to support older mobile devices, which could be disrupted by this change. That would mean this would be a breaking change.

1. I'd like to keep this open for awhile to get feedback from folks on the breaking nature of this change. If there's not a lot of push-back, then I say we merge it.
2. It would be nice add in the additional context into the docs, something like "if you're running on [insert conditions], please consider using WebGL 1 instead"

bigtimebuddy (approved):
Good, but will wait for next major version.

GoodBoyDigital (approved):
nice! yes, the time is right :D Agree with @bigtimebuddy - a major change 👍

ShukantPal (approved):

PR #7989, opened 2021-11-12
chore: add yarn.lock ignore

sudongyuer:
add yarn.lock ignore

##### Pre-Merge Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # package.lock
 packages/**/package-lock.json
 
+# yarn.lock
+packages/**/yarn.lock
+
 # sublime text files
 *.sublime*
 *.*~*.TMP


qwertywelch:
[Package.json](https://github.com/pixijs/pixijs/blob/dev/package.json) indicates that npm should be used, not yarn.

bigtimebuddy:
I agree with @qwertywelch. @sudongyuer could you give us more rationale for this change?

sudongyuer:
Thanks for your reply, npm is very slowly in china hhhh ^_^ ,yarn is quick

sudongyuer:
In other files ,there has no indicates that npm should be used 
image

bigtimebuddy:
The root level package.json has the npm engine requirement that restricts usage.

ShukantPal:
The existence of package-lock.json in the repository itself says that npm is the package manager we use.

bigtimebuddy:
I'm going to close this, because we don't want to support Yarn workflows with the project. Thanks for the suggestion.

PR #7994, opened 2021-11-14
Fix crash creating cached bitmap when filters is empty array

qwertywelch:
##### Description of change

In initCachedDisplayObject, it adds padding from the first filter applied to the DisplayObject if the filters property is truthy, meaning it would attempt to do so even if filters was an empty array. This change checks that's not an empty array, preventing a crash in such a scenario.

##### Pre-Merge Checklist


bigtimebuddy:
Would it be possible to add a test for this situation?

qwertywelch:
Test added!

bigtimebuddy (approved):
Awesome, thanks for the fix!

Zyie (approved):

bigtimebuddy:
Thanks @qwertywelch!

PR #8008, opened 2021-11-20
Clear stencil buffer in push instead of pop

### LINKED ISSUE #6822: Mask ignored when rendering on RenderTexture (Firefox Linux)
I've added a masked sprite to a container, when I render it on screen the mask works correctly but when I try to render it on a RenderTexture the mask is ignored.
Problem is, I tested it in some environments, here's the list:
|  |  Windows 10 | Manjaro KDE   |
| ------------- |:------:|:-------:|
| Chrome    | Ok    |  Ok   |
| Firefox      | Ok    |  Bug |

I have no clue on the bug's origin.
I've created a [simple playground](https://www.pixiplayground.com/#/edit/iJ6CBxiP5PGlzGXt7zi1s), in the upper half you can see the pixi.js behaviour and in the lower half you can see the RenderTexture behaviour.

### Expected Behavior
image

### Current Behavior
image

### Possible Solution
I don't know of any, maybe using filters will bypass the bug but I haven't checked.

### Steps to Reproduce
Get linux + Firefox (I think?) and use this link below.
https://www.pixiplayground.com/#/edit/iJ6CBxiP5PGlzGXt7zi1s

### Environment

- **`pixi.js` version**: 5.3.3
- **Browser & Version**: Firefox 79.0
- **OS & Version**: 5.4.57-1-MANJARO + KDE Plasma 5.12.4
- **Running Example**: https://www.pixiplayground.com/#/edit/iJ6CBxiP5PGlzGXt7zi1s

dev7355608:
##### Description of change

Fixes #6822.

I moved the stencil buffer clear to push out of pop so that the stencil buffer is cleared immediately after it is created and therefore cleared before it is used the first time. I assume that the stencil buffer may be in some sort of undefined state until it is cleared the first time, and that's why the stencil mask fails the first time something with a mask is rendered to the render texture. I couldn't find anything on default/initial stencil buffer (or render buffer) values, so I assume it's undefined.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/mask/StencilSystem.ts
+++ b/packages/core/src/mask/StencilSystem.ts
@@ -47,6 +47,8 @@ export class StencilSystem extends AbstractMaskSystem
         {
             // force use stencil texture in current framebuffer
             this.renderer.framebuffer.forceStencil();
+            gl.clearStencil(0);
+            gl.clear(gl.STENCIL_BUFFER_BIT);
             gl.enable(gl.STENCIL_TEST);
         }
 
@@ -78,8 +80,6 @@ export class StencilSystem extends AbstractMaskSystem
         {
             // the stack is empty!
             gl.disable(gl.STENCIL_TEST);
-            gl.clearStencil(0);
-            gl.clear(gl.STENCIL_BUFFER_BIT);
         }
         else
         {


bigtimebuddy (approved):

Pi-Bouf (approved):

GoodBoyDigital (approved):
nice 👍

PR #8029, opened 2021-12-05
Fix radius documentation for `PIXI.Circle`

ceoshikhar:
##### Description of change

Fixed the documentation for radius member of `PIXI.Circle`. It said the default value of `radius` is `1` but it is actually `0`. If it's the other way round, that the documentation is correct and the default value is supposed to be `1` instead of `0` then I can make that change in a different PR.

##### Pre-Merge Checklist

### DIFF
--- a/packages/math/src/shapes/Circle.ts
+++ b/packages/math/src/shapes/Circle.ts
@@ -14,7 +14,7 @@ export class Circle
     /** @default 0 */
     public y: number;
 
-    /** @default 1 */
+    /** @default 0 */
     public radius: number;
 
     /**


bigtimebuddy (changes requested):
This seems incorrect. The constructor's default value for radius is `0` not `1`. Where did 1 come from, unless I'm missing something?

ceoshikhar:
You are correct, the default radius is `0` but the code documentation says `1` and even the [circle documentation](https://pixijs.download/release/docs/PIXI.Circle.html) says the default value for radius is `1`. The intent of this PR is to fix the documentation and change default value of radius from `1` to `0`.

bigtimebuddy (approved):
Oops, you're right. I read this backwards! Good fix, thanks @ceoshikhar

GoodBoyDigital (approved):

bigtimebuddy:
Thanks @ceoshikhar

PR #8031, opened 2021-12-06
Fix 8030

### LINKED ISSUE #8030: DisplayObject is displayed at the (0,0) position for 1 tick.
Hi.

I found an issue.

It's running example -> https://jsfiddle.net/kgfnsjLe/153/

I use pixijs and pixi-spine.

I tried to change visible of displayObject to true in event listener of pixi-spine.

But, DisplayObject is displayed at the (0,0) position for 1 tick.

*why*
If visible of displayObject is false, updateTransform does not work.
DisplayObject was updated before pixi-spine be update.
and I change of visible displayObject in event listener of pixi-spine.
So, It's displayed at the (0,0) position before next tick.

If displayObject addChild stage before pixi-spine, problem will not reproduce.
However, I think it's not a good solution.

*solution*
So, I want to override it to call updateTransform when the visible of displayObject changes to true.

Any feedback would be appreciated. :)

### Environment

- **`pixi.js` version**: _5.0.4_
- **Browser & Version**: _Chrome 96_
- **OS & Version**: _macOS 10.15.7_
- **Running Example**: _[url](https://jsfiddle.net/kgfnsjLe/153/)_

youga114:
Description of change
Fixes #8030.

Pre-Merge Checklist
- [x] Lint process passed (npm run lint)
- [x] Tests passed (npm run test)

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -211,7 +211,7 @@ export abstract class DisplayObject extends EventEmitter
     public worldAlpha: number;
     public transform: Transform;
     public alpha: number;
-    public visible: boolean;
+    public _visible: boolean;
     public renderable: boolean;
     public filterArea: Rectangle;
     public filters: Filter[] | null;
@@ -290,7 +290,7 @@ export abstract class DisplayObject extends EventEmitter
          *
          * @member {boolean}
          */
-        this.visible = true;
+        this._visible = true;
 
         /**
          * Can this object be rendered, if false the object will not be drawn but the updateTransform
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thank @youga114. Do you mind adding a unit-test for this fix?

dev7355608:
The problem in the linked issue is that the spine state is updated in `updateTransform` if autoUpdate is enabled, which isn't the best place for it. You're not supposed to change visibility while the transforms are updating, just like you shoundn't add or remove objects to the scene graph at that point. If you want to change visibility, you should disable `autoUpdate` and update your objects manually before rendering, which is the proper way anyway. Or you could use the `renderable` property instead of `visible`, and leave `autoUpdate` enabled.

youga114:
Thanks for the feedback. :)
It was a simple example, but some open sources using pixijs are controlling it as visible. I'll check. 
Thank you again.

youga114:
@dev7355608 

Hi. I have a question.
I've disabled and enabled `renderable` and it works fine.
However, I couldn't confirm that it works fine by disabling `autoUpdate`.
Did you mean disabling `autoUpdate` of spine?

dev7355608:
@youga114 Yes, disable `autoUpdate` of the spines, or set `PIXI.settings.GLOBAL_AUTO_UPDATE` to false, and then update the spines like this, for example: 
```js
app.ticker.add(function (delta) {
    const dt = delta / PIXI.settings.TARGET_FPMS;
    for (const spine of spines) {
         spine.update(dt);
    }
});
```

youga114:
@dev7355608 
Thank you. I understand it. it works fine.
I'm also considering running a function registered in pixi-spine's event-listener on the ticker.
Thanks for the feedback. :)

PR #8044, opened 2021-12-13
Remove skip in FilterSystem.roundFrame

dev7355608:
##### Description of change

I removed the skip in `FilterSystem.roundFrame`, because if the skip conditions are met, no rounding is performed at all, which can't be right. Also passing `null` or the identity for `transform` shouldn't make a difference, but it does. So I propose that we remove the skip for now, because it's simpler than fixing it, unless @ShukantPal or someone else wants to do it.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/filters/FilterSystem.ts
+++ b/packages/core/src/filters/FilterSystem.ts
@@ -590,19 +590,6 @@ export class FilterSystem implements ISystem
         transform?: Matrix
     )
     {
-        if (transform)
-        {
-            const { a, b, c, d } = transform;
-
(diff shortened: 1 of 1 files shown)


ShukantPal:
Hmm @ivanpopelyshev - I think we skipped this here because rounding the destination frame doesn't make sense when rotation or skew is in the matrix. 🤔 .

dev7355608:
Never mind, looking at it now I must have been totally confused when I submitted this PR: I read the skip condition inverted, no wonder it didn't make any sense. Sorry for the false alarm!

PR #8050, opened 2021-12-15
A fallback to self if it is undefined. Fixes #7703

### LINKED ISSUE #7703: 'Self is not defined', SSR (Sveltekit)
I am using pixi.js in Sveltekit. Sveltekit is a flexible framework where you can do SSR, SSG, and SPAs (and mix).

Since pixi.js is a browser only library it gets into trouble with SSR. 

What surprises me is that merely importing the module will crash it, even if nothing is instantiated. This forces you to use [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports). It is an extra step which creates problems when using Typescript since you can't import types without crashing everything (or at least I haven't figured out how).

Is this something that is being worked on? Or is this fundamentally not possible to solve?

### Related? 
[Totally random "ReferenceError: self is not defined" in "isMobile.ts" during tests](https://github.com/pixijs/pixijs/issues/7657)
["window is undefined" in nextjs](https://github.com/pixijs/pixijs/issues/6990)

I created a new ticket since I'm not sure how/if these relate, and I wanted to contribute with a minimal reproducible example.

### Expected Behavior

I expect it to be possible to import the module, but not being able to instantiate classes that depend on browser context.

### Current Behavior

Just the act of importing pixi.js crashes in a non-browser environment, whether you do something with it or not.

### Possible Solution

Don't know really. 

Maybe there are workarounds. I'm new to pixi.js.

### Steps to Reproduce

````
npm init svelte@next my-app
cd my-app
nvm use 16
npm i
npm i pixi.js
````
and then import pixi.js in `index.svelte`

````
<script>
    import { onMount } from 'svelte'
    
    // 1. **********************
    // This does not work.
    // Even though the Application never is being instansiated. 
    // Just importing anything from 'pixi.js' will crash the app.
    // import { Application } from 'pixi.js'

    onMount(async () => {
        // 2. ****************
        // This do work, but creates unnecessary hassle, 
        // especially when you start using Typescript and want to be able to import types
        const { Application } = (await import('pixi.js'))
    })
</script>

<h1>Hello world</h1>

````

### Environment

- **`pixi.js` version**: 6.1.2
- **Browser & Version**: NA
- **OS & Version**: Ubuntu 18.04
- **Running Example**:

### PR DESCRIPTION
**Pull request by JoshuaPelealu**
##### Description of change

Uses a fallback of `globalThis.self` if the default `self` global is undefined, which typically occurs on a Server Side Rendering applications. This ensures, on build and test times, that self will be defined

##### Pre-Merge Checklist

This will fix #7703

### CODE REVIEW

**Comment by bigtimebuddy** [comment]
Why not replace self with globalThis? We'd need to polyfill globalThis for IE, but that should work too, right?

**Comment by JoshuaPelealu** [comment]
Actually yeah, I was thinking of doing this, but I didn't know if you guys would be fine with that

**Comment by JoshuaPelealu** [comment]
Oh and for a side note, I only did this on the files that caused an issue on a SvelteKit SSR Application.

as for the polyfill, I don't know what kind of polyfill you guys are ok with

**Comment by bigtimebuddy** [comment]
I'm open to that, the rest of the `self` references can be found in this PR #6880, which was the thing that converted them from `window` to run in Workers. If you don't mind adding these to this PR, that would be great and less complicated to follow-up later.

Polyfill would need some research to figure out the best solution. I don't know off the top of my head but likely should live in the polyfill package.

**Comment by JoshuaPelealu** [comment]
I'm not on a computer right now, but I'll do that when I get on one

File: packages/polyfill/src/Promise.ts [hunk 1]
```diff
@@ -1,7 +1,7 @@
-import Polyfill from 'promise-polyfill';
+packages/polyfill/src/index.tsimport Polyfill from 'promise-polyfill';
```
bigtimebuddy, line 1:
```suggestion
import Polyfill from 'promise-polyfill';
```

bigtimebuddy:
### Builds

* https://pixijs.download/pr-8050/pixi.min.js
* https://pixijs.download/pr-8050/pixi-legacy.min.js

### Links

* https://pixijs.io/examples/?v=pr-8050

JoshuaPelealu:
> ### Builds
> 
>     * https://pixijs.download/pr-8050/pixi.min.js
> 
>     * https://pixijs.download/pr-8050/pixi-legacy.min.js
> 
> 
> ### Links
> 
>     * https://pixijs.io/examples/?v=pr-8050

Ok, finally had time to test it out, and with the new build on my SSR environtment, it runs perfectly fine now

JoshuaPelealu:
> ### Builds
> 
>     * https://pixijs.download/pr-8050/pixi.min.js
> 
>     * https://pixijs.download/pr-8050/pixi-legacy.min.js
> 
> 
> ### Links
> 
>     * https://pixijs.io/examples/?v=pr-8050

Ok, finally had time to test the build up, and in my environment, they are working awesome!

here's the repo:
https://github.com/JoshuaPelealu/TouchOfClassEvents/tree/updated-pixi

here's the deployment:
https://vercel.com/joshuapelealu/touch-of-class-events/EeyDCdQn5PyWCSsBooLY3ES926gJ

I was too lazy, so i made a script that will replace the build inside the node modules after `npm install` is done.

ShukantPal (approved):

bigtimebuddy:
I think the only thing that this PR is missing is a check if `globalThis` is undefined. Practically, [this is for IE/Edge](https://caniuse.com/mdn-javascript_builtins_globalthis). What about something simple like this:

```js
if (typeof globalThis === 'undefined') {
  self.globalThis = self;
}
```

Anyone want to weigh-in on a lightweight fallback/polyfill for globalThis?

themoonrat:
I don't think that would work for Node versions that doesn't have `globalThis` (earlier than 12) as it doesn't have `self`, only a `global`.

```
if (typeof globalThis === 'undefined') {
  if (typeof self !== 'undefined') { // covers browsers
     self.globalThis = self;
  } else if (typeof global !== 'undefined') { // covers versions of Node < 12
     global.globalThis = global;
  }
}
```
^ So something like that?

JoshuaPelealu:
We can also use core-js as our polyfills libraries as it provides polyfills up to ES-2021.

https://github.com/zloirock/core-js

bigtimebuddy:
I like @themoonrat's suggestion and would prefer to not introduce another dependency.

TrevorHinesley:
Unsure if it's related, but when using the build script created by @JoshuaPelealu every other refresh shows the `self is not defined` error, and I can't do `app.stage.render()` without seeing:

```
Sprite.ts:384 Uncaught TypeError: Cannot read properties of undefined (reading 'batch')
```

Which is this line:

```
renderer.batch.setObjectRenderer(renderer.plugins[this.pluginName]);
```

TrevorHinesley:
After your most recent merge of dev->dev, I ran a build from this PR's branch, then ran my own dist from Github, and am still running into the `self` issues and the render error above. If it's just me, that's ok, I just wanted to make sure you were aware of what I'm seeing on my end. @bigtimebuddy @JoshuaPelealu

JoshuaPelealu:
Hey @TrevorHinesley ,

On the TouchOfClassEvents repo that gave as an example does has a shim and it has a manual script that fixes the problem because it doesn't use this one's branch, 

We are currently still trying to put polyfills for globalThis for this or to be approved, but I haven't had the time because I have school. The repo TouchOfClassEvents is a really quick patch up i did because it is an assignment that was already due January, so i wouldn't base anything off of that repo anymore.

JoshuaPelealu:
@TrevorHinesley and to be clear, this pr is only going to fix the error that is caused because on build, the loader will error out because `self` is not declared in nodejs, the shim will more likely still be necessary to run this in an SSR application. We might or might not find a way to solve this, but for now, the shim will be the best option.

JoshuaPelealu:
@TrevorHinesley 

Alright, so I've cloned the latest branch push and build a dist, and using my SvelteKit template I made a quick app to use pixijs and it works fully fine:
https://github.com/JoshuaPelealu/pixisveltekit

if you copy your `dist` from your pixijs build and move it into `src/pixijs/` and run the app, it will run totally fine... Of course I put the shim in there, but again from my comment above, it is necessary for now until we find a better solution. 
just in case, to run it, you have to `npm i` then `npm run dev` and open `localhost:3000`


>  every other refresh shows the `self is not defined` error, and I can't do `app.stage.render()` without seeing:
> 
> ```
> Sprite.ts:384 Uncaught TypeError: Cannot read properties of undefined (reading 'batch')
> ```
> 
> Which is this line:
> 
> ```
> renderer.batch.setObjectRenderer(renderer.plugins[this.pluginName]);
> ```

This is caused because I setup my app so that it doesn't automatically start (which is my preference), that means it doesn't initialize all the necessary components to run the app yet, to fix this you gotta make sure before you run anything else, you have to run `app.render()` to render the app and run it.


Hope this is helpful, and if you find any bugs or error, tell me so I can find out the root, I really want this to be part of pixijs (to support SSR applications better) so no one else have to go through this. :)

TrevorHinesley:
Thanks for the detailed responses and explanations @JoshuaPelealu. Trying this out now. Part of my problem was I was doing `app.stage.render()` in a section of my logic, which wound up being the wrong syntax (though the docs made it seem like that was a usable function IMO).

JoshuaPelealu:
Hey @bigtimebuddy ,

quick question, how would I go about putting the polyfill for globalThis? Should I put it inside everyfile that uses globalThis, or should I put it it inside a file and import it?

bigtimebuddy:
Yes, it should go in polyfill and probably be the first thing imported, since the polyfills now reference globalThis

bigtimebuddy (approved):

bigtimebuddy:
Thank you @JoshuaPelealu :tada:

Specy:
@bigtimebuddy oh wow i literally just had this issue in next.js, when will this avaiable on npm? sadly im using [react-pixi](https://github.com/inlet/react-pixi) so i'll have to ask the creator to update to this version

PR #8057, opened 2021-12-18
Fix for tiling sprite issue

### LINKED ISSUE #7976: When using 2 TilingSprites with separate textures, the second appears as a single pixel
### Expected Behavior
Both TilingSprites should be fully visible
### Current Behavior
Second TilingSprite shows as a single pixel

### Test cases
in 6.1.0, both TilingSprites display as expected: https://jmlee2k.github.io/pixi6-tilingsprite-test/index-610.html

in 6.2.0, the second tiling sprite only shows as a single pixel: https://jmlee2k.github.io/pixi6-tilingsprite-test/index-620.html

### Environment

- **`pixi.js` version**: 6.2.0
- **Browser & Version**: Chrome, Firefox
- **OS & Version**: windows 10, macOS
- **Running Example**: https://jmlee2k.github.io/pixi6-tilingsprite-test/index-620.html

GoodBoyDigital:
##### Description of change
Tiling like a boss! fixes #7976
 

##### Pre-Merge Checklist

The new PR of allowing custom locations in WebGL 2 is the culprit here :D When making webGL2 shaders, we need make we set the locations to be the correct order also, otherwise we get some geometry mis-matches.

In this instance, the uvs and the positions are a different order when the complex shader (es300) is run before the simple shader (es100).

Perhaps we should consider changing things so that we always deal with attribute locations unless the developer specifies thats the want full control, or perhaps we detect for `layout(location = ` in the shader? 

what do you think @dev7355608 ?

### DIFF
--- a/packages/sprite-tiling/src/sprite-tiling.vert
+++ b/packages/sprite-tiling/src/sprite-tiling.vert
@@ -3,8 +3,8 @@
 
 precision lowp float;
 
-in vec2 aVertexPosition;
-in vec2 aTextureCoord;
+layout(location = 1) in vec2 aVertexPosition;
+layout(location = 0) in vec2 aTextureCoord;
 
 uniform mat3 projectionMatrix;
 uniform mat3 translationMatrix;


GoodBoyDigital:
ooh! ok, just getting my head back into this so forgive the slowness... 

https://github.com/pixijs/pixijs/pull/7827 

should fix this issue also.. at the root :D

PR #8059, opened 2021-12-21
Fix Canvas's TilingSprite not dependent of Texture resolution

### LINKED ISSUE #8055: Canvas's TilingSprite should be rendered independent of Texture resolution
### Expected Behavior
Canvas's TilingSprite should be rendered independent of Texture resolution.

### Current Behavior
Canvas's TilingSprite expands according to the Texture's resolution, so the drawing result is different from WebGL.

image

### Possible Solution
This problem does not occur in v5.3.11, but it seems to occur in v6.0 and v5.4.0-rc.1.
#6869 changes may have something to do with it.

### Steps to Reproduce
https://codesandbox.io/s/pixi-tilingsprite-with-canvasrenderer-4nyb2

### Environment

- **`pixi.js` version**: 6.2.0
- **Browser & Version**: Chrome 96.0.4664.110（Official Build） （x86_64）
- **OS & Version**: macOS 12.1
- **Running Example**: https://codesandbox.io/s/pixi-tilingsprite-with-canvasrenderer-4nyb2

JetLua:
Fixed #8055 

fixed: https://jsfiddle.net/JetLu/oraf9jcw/


bigtimebuddy (approved):
Checks out, good fix @JetLua

themoonrat (approved):

bigtimebuddy:
Thank you @JetLua

PR #8060, opened 2021-12-21
Reformat documentation for AppLoaderPlugin.ts and Loader.ts

omer8374:
#### Description of change

#7159

#### Pre-Merge Checklist


packages/loaders/src/Loader.ts
```diff
@@ -668,9 +631,8 @@ class Loader
     private static _shared: Loader;
     /**
      * If this loader cannot be destroyed.
-     * @default false
      */
-    private _protected: boolean;
+    private _protected = false;
```
ShukantPal, line 635: This is technically a code change.

omer8374: Should I change it back? Or is there something else I need to do?

bigtimebuddy: @omer8374 if you could change and it back and push, that would be great. Thanks. In general, we want these documentation fixes to not touch any code, even if it seems small and trivial.

omer8374: @bigtimebuddy thanks I changed it back.

ShukantPal (approved):

ShukantPal:
Thanks!

Zyie (approved):

bigtimebuddy:
Thank you @omer8374

PR #8085, opened 2022-01-03
Make it easier to use the new federated events API with TypeScript

caasi:
##### Description of change

I am following [Shukant Pal's guide](https://www.shukantpal.com/blog/pixijs/federated-events-api/) to use the new federated events API in my project.

When I add these lines:

```typescript
if (!('events' in app.renderer)) {
  app.renderer.addSystem(EventSystem, 'events')
}
```

tsc warns me that `addSystem()` is missing in the `AbstractRenderer` class. And the `Renderer` interface in `@pixi/events` is not compatible with `AbstractRenderer`.

So I make some changes to resolve them.

##### Pre-Merge Checklist


tears330 (approved):

packages/core/src/AbstractRenderer.ts
```diff
@@ -386,6 +387,15 @@ export abstract class AbstractRenderer extends EventEmitter
         return renderTexture;
     }
 
+    /**
+     * Adds a new system to the renderer.
+     *
+     * @param ClassRef - Class reference
+     * @param [name] - Property name for system
```
bigtimebuddy, line 394: We don't need [name] here you can just use name. Our docs infer optional.

packages/core/src/Renderer.ts
```diff
@@ -370,6 +370,9 @@ export class Renderer extends AbstractRenderer
 
     /**
      * Add a new system to the renderer.
+     *
+     * @extends PIXI.AbstractRenderer#addSystem
```
bigtimebuddy, line 374: You can remove @extends flag, it's not functional here.

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -152,6 +152,16 @@ export class CanvasRenderer extends AbstractRenderer
         this.resize(this.options.width, this.options.height);
     }
 
+    /**
+     * Adds a new system to the renderer. It does nothing in the CanvasRenderer.
+     *
+     * @extends PIXI.AbstractRenderer#addSystem
```
bigtimebuddy, line 158: Also, @extends isn't useful here either. Can be removed.

bigtimebuddy (approved):
Thanks for this, super clean, just a few minor docs notes.

packages/core/src/AbstractRenderer.ts
caasi: Fixed. I misunderstand what a `[]` is in JSDoc.

packages/core/src/Renderer.ts
caasi: Removed.

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
caasi: Removed.

bigtimebuddy:
Thank you @caasi

PR #8087, opened 2022-01-06
Fix Sprite.getLocalBounds corrupting bounds

dev7355608:
##### Description of change

Currently `getLocalBounds` modifies and corrupts the `_bounds` of the sprite so that subsequent `getBounds(true)` calls return the local bounds instead of the global bounds. This PR fixes that.

##### Pre-Merge Checklist

### DIFF
--- a/packages/sprite/src/Sprite.ts
+++ b/packages/sprite/src/Sprite.ts
@@ -1,6 +1,6 @@
 import { BLEND_MODES } from '@pixi/constants';
 import { Texture } from '@pixi/core';
-import { Container } from '@pixi/display';
+import { Bounds, Container } from '@pixi/display';
 import { ObservablePoint, Point, Rectangle } from '@pixi/math';
 import { settings } from '@pixi/settings';
 import { sign } from '@pixi/utils';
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Can you provide a test or reproduction of the underlying issue?

dev7355608:
> Can you provide a test or reproduction of the underlying issue?

https://www.pixiplayground.com/#/edit/rBsga2X9LBZK4DnYr6ejx

packages/sprite/src/Sprite.ts
```diff
@@ -417,10 +417,15 @@ export class Sprite extends Container
...
         if (this.children.length === 0)
         {
-            this._bounds.minX = this._texture.orig.width * -this._anchor._x;
-            this._bounds.minY = this._texture.orig.height * -this._anchor._y;
-            this._bounds.maxX = this._texture.orig.width * (1 - this._anchor._x);
-            this._bounds.maxY = this._texture.orig.height * (1 - this._anchor._y);
+            if (!this._localBounds)
+            {
+                this._localBounds = new Bounds();
+            }
```
bigtimebuddy, line 423: Minor suggestion.
```suggestion
            this._localBounds = this._localBounds || new Bounds();
```

bigtimebuddy:
Great, thanks for the example. We need a unit-test that handles that example for this PR. But other than that, I'm good on this fix.

GoodBoyDigital (approved):
nice catch @dev7355608 ! If you could add a test, so we don't regress this in the future that would be awesome. Thanks!

bigtimebuddy (approved):

bigtimebuddy:
Thank you @dev7355608

PR #8088, opened 2022-01-06
Add Rectangle.intersects to core math

dev7355608:
##### Description of change

Moves the `intersects` method from `@pixi/math-extras` to `@pixi/math`.

##### Pre-Merge Checklist

### DIFF
--- a/packages/math-extras/src/rectangleExtras.ts
+++ b/packages/math-extras/src/rectangleExtras.ts
@@ -1,35 +1,6 @@
 import { Rectangle } from '@pixi/math';
 
 /**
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
Can you give more context on this proposed changed? `intersects` is not used internally so this should be in extras.

dev7355608:
> Can you give more context on this proposed changed? `intersects` is not used internally so this should be in extras.

I want to use `intersects` in #8089. Of course this should be merged if the other PR isn't merged.

bigtimebuddy:
I'm going to close this, as it's part of your other PR already. I don't think we would consider this change outside that PR.

PR #8091, opened 2022-01-06
Render to an empty filter texture if the filter source frame does not intersect the current source frame 

dev7355608:
##### Description of change

Correct me if I'm wrong, but if the filter source frame doesn't intersect the current source frame at all, we can render to an empty texture (instead of the full-sized one) even if auto fit is false.

##### Pre-Merge Checklist


ivanpopelyshev:
What we know from #7895 : we cant actually ignore those trees because of texture GC and other stuff, different behaviour. This is the middle ground.

dev7355608:
> What we know from #7895 : we cant actually ignore those trees because of texture GC and other stuff, different behaviour. This is the middle ground.

With this is change no render calls are actually skipped: we only render to an empty (actually 1x1) filter texture instead of the full-sized one. So it should be fine with regards to the problem you mentioned, right? Unless the user is allowed to change this part of the filter fragment shader:
```js
vec4 filterVertexPosition(void)
{
    vec2 position = aVertexPosition * max(outputFrame.zw, vec2(0.)) + outputFrame.xy;
    return vec4((projectionMatrix * vec3(position, 1.0)).xy, 0.0, 1.0);
}
```
If it is required that all filter implement exactly this, then it shouldn't matter if we render to an empty texture if the filter source frame doesn't intersect the current source frame.

dev7355608:
I included a small improvement in the `autoFit` true case: both width and height are set to zero now if the fitted source frame is empty so that we render to a 1x1 texture and not to a 1x2048 texture for example (`Rectangle.fit` doesn't set both width and height to zero if the intersection is empty).

GoodBoyDigital:
Hey @dev7355608 this seems cool, would be good to know if there is a meaningful improvement to performance with this optimisation?

dev7355608:
With culling (#8089) it could potentially make a great difference because it allows us to skip rendering (`_render`) of offscreen filtered objects without autofit completely. Without culling I'm not sure how big the impact would be, but it should definitely make some difference to render to a 1x1 texture instead of something much bigger. I mainly proposed this change to improve the culling implementation.

dev7355608:
Could we include this change in 6.3.0? Without this change an object with a filter that has `autoFit` set to false cannot be culled. If the filter source frame is not intersecting the current source frame, it doesn't matter that we render to an empty texture, because nothing ends up in the current framebuffer assuming the filter's vertex shader doesn't do something illegal, i.e. something other than this:
```glsl
vec2 position = aVertexPosition * max(outputFrame.zw, vec2(0.)) + outputFrame.xy;
gl_Position = vec4((projectionMatrix * vec3(position, 1.0)).xy, 0.0, 1.0);
// or legacy:
// gl_Position = vec4((projectionMatrix * vec3(aVertexPosition, 1.0)).xy, 0.0, 1.0);
```

bigtimebuddy:
Could you resolve conflicts @dev7355608?

bigtimebuddy (approved):

PR #8134, opened 2022-01-24
typescript add DisplayObject.interactive

Pythux:
##### Description of change

add typescript type to DisplayObject.interactive ([exist in documentation](https://pixijs.download/release/docs/PIXI.DisplayObject.html#interactive) but does not exist on use with typescript

I copied the documentation text in the comment of interactive to conform with it.

##### Pre-Merge Checklist

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -215,6 +215,7 @@ export abstract class DisplayObject extends EventEmitter
     public renderable: boolean;
     public filterArea: Rectangle;
     public filters: Filter[] | null;
+    public interactive: boolean;
     public isSprite: boolean;
     public isMask: boolean;
     public _lastSortedIndex: number;
@@ -353,6 +354,12 @@ export abstract class DisplayObject extends EventEmitter
          * @member {?PIXI.Filter[]}
          */
         this.filters = null;
+
+        /**
+         * Enable interaction events for the DisplayObject.
+         * Touch, pointer and mouse events will not be emitted unless `interactive` is set to `true`.
+         */
+        this.interactive = false;
 
         /**
          * Currently enabled filters


bigtimebuddy (changes requested):
Thanks @Pythux for the PR and welcome!

I don't think this is the correct approach here. The `interactive` property should be provided in the global.d.ts file in the `@pixi/interaction` package. And DisplayObject should extend GlobalMixins.DisplayObject.

Could you provide a bug that demonstrates the issue?

Pythux:
Hi @bigtimebuddy,

I noticed in my IDE (vscode) that `.interactive` does not exist when all I had in dependency was `"pixi.js": "^6.2.1",`.

In the documentation it does exist, and work, I assume now that it is because of `@pixi/interaction` package here by default.

It is when I installed `"@pixi/events": "^6.2.1",` package that my IDE pointed me to `node_modules/@pixi/interaction/index.d.ts` -> `InteractiveTarget` interface

Then making the path from `DisplayObject` to it was not that easy, I stopped at `abstract class DisplayObject extends EventEmitter`.
It is when I discovered `interface DisplayObject extends Omit<GlobalMixins.DisplayObject, keyof EventEmitter>, EventEmitter` with `GlobalMixins.DisplayObject` pointing to many files that I found it in the package interaction: `interface DisplayObject extends InteractiveTarget`.

I'm quite new to typescrite and don't know you could have the same name in the same file for `DisplayObject` as `interface` and `abstact class`.

In the abstact class code, `interactive` appear only in the destroy function, it is not declared nor initialized in it. At the time it appear to me that there are missing declaration and initialisation.
The declaration is in the `GlobalMixins.DisplayObject` of `@pixi/interaction`, I see it now.

It is still kind of difficult that the list of interface the class extends to is not explicitly named, imported nor it's value initialized in the constuctor, I don't know why typescript don't ask for it in the constructor.

I see now also `declare namespace GlobalMixins` in all `global.d.ts` at the root of all packages that could add something to it.


So, a lot of questions but now I see it's typing and my IDE see it too.
Thank you for your quick answer.

There is still the problem when only `pixi.js` is in dependency of your project, I still get the error, even if I could navigate in the typing of `Container` extending `DisplayObject` with `GlobalMixins` getting me to `InteractiveTarget` so it is kind of strange.

```js

const app = new PIXI.Application()
app.stage.interactive = true   // Property 'interactive' does not exist on type 'Container'.ts(2339)
```

bigtimebuddy:
Thanks for digging in. I'm going to close and suggest we work this out as an issue. Mixins structure is complex, as you've seen but it allows us to externalize properties. I'll do a quick test to try to reproduce.

Pythux:
Yes, thank you for your time.

bigtimebuddy:
Here's a simple boilerplate project I created https://github.com/bigtimebuddy/pixi-sound-webpack-example this uses `interactive` property and I get no type issues or problems in VSCode.

Pythux:
I finally found the problem thanks to your example, it was in my `tsconfig.json`, I don't have `"esModuleInterop": true`, the default value is false, causing the type issue.

For an unknown reason my IDE give me the type at some point, making me think that it was linked to `@pixi/events`, but a fresh install give me the expected result.

Thank you again for your exemple and time, I would not have guessed it would be in my `tsconfig.json`

PR #8173, opened 2022-02-07
Fix blitFramebuffer

dev7355608:
`gl.blitFramebuffer` are passed incorrect arguments: the source/destination rectangles are defined by two points and not x, y, width, height.

##### Pre-Merge Checklist


bigtimebuddy:
This seems legit, but I wonder why it hasn't shown up at an issue. Do you think can you reproduce a bug triggered by this? I just want to understand the scope of the problem.

dev7355608:
It goes unnoticed as long as you blit with `src/destPixels.x/y` equal zero, in which case the incorrect arguments match the correct ones. Pixi itself never uses the blit function in a way that could trigger this bug.

bigtimebuddy (approved):

PR #8178, opened 2022-02-09
TreeSearch findHit void function return value used - InteractionManager hit is never set

jasonsturges:
It appears that a void function return value is used.  

`hit` is never set because `findHit()` returns nothing - it does not return `recursiveFindHit()` 

`InteractionManager` function `processInteractive()` is setting `hit` to `TreeSearch` class `findHit()` function

image

However, `TreeSearch` class `findHit()` specifies `void` and returns nothing, despite docs specifying a return

image

Note that the `recursiveFindHit()` function returns a `boolean`

```ts
    /**
     * @return - Returns true if the displayObject hit the point
     */
    public recursiveFindHit(interactionEvent: InteractionEvent, displayObject: DisplayObject,
        func?: InteractionCallback, hitTest?: boolean, interactive?: boolean
    ): boolean
    {
        let hit = false;
        // ...
        return hit;
    }
```

### DIFF
--- a/packages/interaction/src/TreeSearch.ts
+++ b/packages/interaction/src/TreeSearch.ts
@@ -193,8 +193,8 @@ export class TreeSearch
      */
     public findHit(interactionEvent: InteractionEvent, displayObject: DisplayObject,
         func?: InteractionCallback, hitTest?: boolean
-    ): void
+    ): boolean
     {
-        this.recursiveFindHit(interactionEvent, displayObject, func, hitTest, false);
+        return this.recursiveFindHit(interactionEvent, displayObject, func, hitTest, false);
     }
 }


jasonsturges:
Should have submitted an issue instead of a pull request for this...

Not sure what this means for `processInteractive()`, which specifies `void` function signature but it returning `hit(0`

```ts
    public processInteractive(interactionEvent: InteractionEvent, displayObject: DisplayObject,
        func?: InteractionCallback, hitTest?: boolean
    ): void
    {
        const hit = this.search.findHit(interactionEvent, displayObject, func, hitTest);

        const delayedEvents = this.delayedEvents;

        if (!delayedEvents.length)
        {
            return hit;
        }
```

This only passed previously because `hit` was `void`.

jasonsturges:
Going to make this an issue instead, as I'm not helping with this pull request.

jasonsturges:
Sorry for the broken pull request - closing this, and created issue #8179 for advice.

PR #8188, opened 2022-02-15
Fix rounding in getOptimalTexture

dev7355608:
##### Description of change

Currently the render texture pool does not return a screen-sized render texture with some noninteger resolutions because of floating point inaccuracies. I added a small epsilon to the ceil in `getOptimalTexure` to fix that problem.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/renderTexture/RenderTexturePool.ts
+++ b/packages/core/src/renderTexture/RenderTexturePool.ts
@@ -78,8 +78,8 @@ export class RenderTexturePool
     {
         let key;
 
-        minWidth = Math.ceil(minWidth * resolution);
-        minHeight = Math.ceil(minHeight * resolution);
+        minWidth = Math.ceil((minWidth * resolution) - 1e-6);
+        minHeight = Math.ceil((minHeight * resolution) - 1e-6);
 
         if (!this.enableFullScreen || minWidth !== this._pixelsWidth || minHeight !== this._pixelsHeight)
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):

packages/core/src/renderTexture/RenderTexturePool.ts
```diff
@@ -78,8 +78,8 @@ export class RenderTexturePool
     {
         let key;
 
-        minWidth = Math.ceil(minWidth * resolution);
-        minHeight = Math.ceil(minHeight * resolution);
+        minWidth = Math.ceil((minWidth * resolution) - 1e-6);
```
ShukantPal, line 81: Don't we have an EPSILON setting / constant?

bigtimebuddy: We do not have a consistent EPSILON. I found two usages: `Number.EPSILON` is being used in math-extras, and `GRAPHICS_CURVES.epsilon` in the graphics. None of them are consistent. I think this is okay here.

ivanpopelyshev: We have https://github.com/pixijs/pixijs/blob/d1e9cf2a069f07d355b09b84afbb250bb8cc8817/packages/math/src/shapes/Rectangle.ts#L198

PR #8199, opened 2022-02-21
Added safe reference to CanvasRenderingContext2D in Text

### LINKED ISSUE #8197: CanvasRenderingContext2D requirement breaks tests in node environment
I don't know if I would call this a real "issue" but since version 6.2.0, my tests have not been able to run under node with `jsdom`, probably related to PR #7748. 

I get the following error: 

```
layout module > alignJustify() > whitespace and alignJustify() (Issue #171) > Should not throw when align is justify and text is only whitespace.
-----
Error: expect(received).not.toThrow()

Error name:    "ReferenceError"
Error message: "CanvasRenderingContext2D is not defined"

      21 | ): IFontMetrics => {
      22 |   if (forceUpdate) {
    > 23 |     textField.updateText(false);
         |               ^
      24 |     return measureFont(textField.context);
      25 |   } else {
      26 |     const props = measureFont(textField.context);

```
- 6.1.3 works fine
- If I reconfigure my tests to include `jest-canvas-mock` the error goes away but the measurements are all wrong. 
- When I run the tests under `@jest-runner/electron`, everything works fine. 
- This is only really an issue because I can no longer rely on a simple setup without electron to run my jest tests. It also basically renders the vscode-jest extension useless. 

For now I plan to use the slightly older version 6.1.x for my development, but if there's a simple fix, like an option whether to use the CanvasRenderingContext2D, I wonder if we can possibly add it?

### Environment

- **`pixi.js` version**:  6.2.0+
- **Browser & Version**: node@16.13.1, jsdom@^19.0.0, canvas@^2.9.0
- **OS & Version**:  MacOS 12.1

mimshwright:
##### Description of change

Added a check for `undefined` when referencing `CanvasRenderingContext2D`.

This allows the code to continue to run in a node environment and shouldn't break anything in other environments.

Fixes #8197 

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -326,9 +326,10 @@ export class Text extends Sprite
         // Checking that we can use moddern canvas2D api
         // https://developer.chrome.com/origintrials/#/view_trial/3585991203293757441
         // note: this is unstable API, Chrome less 94 use a `textLetterSpacing`, newest use a letterSpacing
-        // eslint-disable-next-line max-len
-        const supportLetterSpacing = 'letterSpacing' in CanvasRenderingContext2D.prototype
-            || 'textLetterSpacing' in CanvasRenderingContext2D.prototype;
+        const supportLetterSpacing = window?.CanvasRenderingContext2D !== undefined
+            ? ('letterSpacing' in CanvasRenderingContext2D.prototype
+            || 'textLetterSpacing' in CanvasRenderingContext2D.prototype)
+            : false;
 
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks for the suggestion @mimshwright but unfortunately we aren't going to consider this change. Supporting node environments is something we're working towards in a much more official way, but this is not how we'd like to do it. I suggest as a workaround that you add something like this to your code before hitting that code:

```js
globalThis.CanvasRenderingContext2D = {};
```

mimshwright:
Understood.

PR #8227, opened 2022-03-17
Rewrite buildCircle to handle circles, ellipses, and rounded rectangles

### LINKED ISSUE #7872: Graphics produces incorrect geometry for rounded rectangles
`RoundedRectangle` is defined by circles (that's what the `contains` method tells us), but the graphics geometry builder uses quadratic bezier curves. Those two don't match. See example: green is the bezier rounded rectangle underneath the circle rounded rectangle in red.

### Environment

- **`pixi.js` version**:  6.1.3
- **Running Example**: https://www.pixiplayground.com/#/edit/UyCPdRdDHSCs8SueL9dki

dev7355608:
##### Description of change

Rounded rectangles are now handled by `buildCircle`. Fixes #7872.

I removed one test that was failing. The comment in the test doesn't make much sense, but the test checks that the first and last points generated by `buildCircle` is the same. `buildCircle.triangulate` expected this, but in the new implementation there's no extra point needed to successfully triangulate. The requirement that the first and last point is the same isn't important anywhere else, as far as I can tell, in fact `buildLine` removes the last point if it matches the first. I couldn't observe gaps in the outline either with the new implementation.

If I did my math right, this is translates to a maximum absolute deviation from the circle of approximately 0.029 for circles, but for ellipses it's approx. double that. So I chose the 0.029 for the new implementation regardless of the type.
```js
let totalSegs = Math.floor(30 * Math.sqrt(circleData.radius))
            || Math.floor(15 * Math.sqrt(width + height));

        totalSegs /= 2.3;
```

##### Pre-Merge Checklist


bigtimebuddy:
I agree with Ivan's comment about this, the visual difference is enough that I'd consider this breaking change.

**This PR**
https://jsfiddle.net/bigtimebuddy/L8fx3rq1/2/

**Dev**
https://jsfiddle.net/bigtimebuddy/gkwqv0z6/

My suggestion to get this in 6.x is to add a feature flag (similar to `Text.nextLineHeightBehavior` which change the line-height to behave more like the DOM). Suggest `Graphics.nextRoundedRectBehavior` which will default to `false` setting to true would use the buildCircle behavior. In v7, we'll remove the old stuff and make this the default.

bigtimebuddy (approved):
Thanks, this works!

PR #8234, opened 2022-03-21
Hide new letterSpacing API behind a flag

### LINKED ISSUE #8208: v6.2.2 letterSpacing is broken with chrome 99.0.48
### Expected Behavior
letterSpacing works in chrome

### Current Behavior
letterSpacing has no effect at all

### Steps to Reproduce
See here: https://www.pixiplayground.com/#/edit/19vqJzdGI3_aKknD2P1d1

### Environment

- **`pixi.js` version**: 6.2.2
- **Browser & Version**: Chrome 99.048
- **OS & Version**:  MacOS 12.2.1
- **Running Example**: https://www.pixiplayground.com/#/edit/19vqJzdGI3_aKknD2P1d1

bigtimebuddy:
Fixes #8208 by hiding the new letterSpacing API behind a feature flag `Text.experimentalLetterSpacing` that way users can opt-in.

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -57,7 +57,7 @@ export class Text extends Sprite
      * In the next major release, we'll enable this by default.
      */
     public static nextLineHeightBehavior = false;
-    
+
     /**
      * New rendering behavior for letter-spacing which uses Chrome's new native API. This will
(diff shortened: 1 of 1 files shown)


ivanpopelyshev:
In future, we can make a special section like `PIXI.settings` but for experimental features? because they're likely switched on/off or be deprecated if something goes wrong

PR #8239, opened 2022-03-24
Fix bounds of polygon drawing without fill

dev7355608:
##### Description of change

If a polygon drawing has no fill, the line alignment is based on the orientation of the input polygon. For these drawings, the `lineWidth` calculation has to be different. We can't be sure about the least amount of padding unless we analyze the curve itself; so we have to assume the worst case, and consequently the bounds could be unnecessarily padded by a full width.

Demo of failure: https://www.pixiplayground.com/#/edit/SM4g0Yp_-ad-NrLFVGtD3

##### Pre-Merge Checklist


ivanpopelyshev:
That's interesting!

PR #8243, opened 2022-03-27
Update README.md of @pixi/settings to fix the example import statement

gorisanson:
##### Description of change

The example import statement on the README.md Usage of @pixi/settings 

before this commit:
```js
import * as settings from '@pixi/settings';
```

after this commit: 
```js
```

I checked it out that the latter is the right one. If someone import like the former, the changed values of pixi settings are not shared by other modules.

##### Pre-Merge Checklist

### DIFF
--- a/packages/settings/README.md
+++ b/packages/settings/README.md
@@ -9,5 +9,5 @@ npm install @pixi/settings
 ## Usage
 
 ```js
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):
Thank you!

PR #6876, opened 2020-09-14
Fix crash on Adreno 3xx GPU devices (e.g. vivo 1606, vivo 1610, Xiaomi Redmi Note 4G).

dumganhar:
The sampler array should not be a hardcode value of 16 since it will be larger than what maxRecommendedTextures returns.
maxRecommendedTextures will return 4 if the android version is smaller than 7.

The crash stack is:

```
09-14 14:41:35.409 14755-14934/? A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 14934 (GLThread 1122)
09-14 14:41:35.519 622-622/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
09-14 14:41:35.519 622-622/? A/DEBUG: Build fingerprint: 'vivo/1606/1606:6.0.1/MMB29M/compiler01052117:user/release-keys'
09-14 14:41:35.519 622-622/? A/DEBUG: Revision: '0'
09-14 14:41:35.519 622-622/? A/DEBUG: ABI: 'arm'
09-14 14:41:35.519 622-622/? A/DEBUG: pid: 14755, tid: 14934, name: GLThread 1122  >>> com.yy.demo <<<
09-14 14:41:35.519 622-622/? A/DEBUG: signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
09-14 14:41:35.569 622-622/? A/DEBUG: Abort message: '=================================================================
    ==14755==ERROR: AddressSanitizer: SEGV on unknown address 0x000003fc (pc 0xae5c89bc bp 0x9cc6fc00 sp 0x9d091170 T16777215)
    ==14755==The signal is caused by a READ memory access.
    ==14755==Hint: address points to the zero page.
        #0 0xae5c89b8  (/system/vendor/lib/egl/libRBGLESv2_adreno.so+0xb39b8)
        #1 0xae59b83a  (/system/vendor/lib/egl/libRBGLESv2_adreno.so+0x8683a)
        #2 0xae57fd4e  (/system/vendor/lib/egl/libRBGLESv2_
09-14 14:41:35.569 622-622/? A/DEBUG:     r0 00000000  r1 00003a56  r2 00000006  r3 9d092978
09-14 14:41:35.569 622-622/? A/DEBUG:     r4 9d092980  r5 9d092930  r6 00000019  r7 0000010c
[...]

PR #8244, opened 2022-03-27
Remove hardcoded plugin name for getting MAX_TEXTURES in Graphics.ts

SergeiSomin:
##### Description of change
Remove hardcoded plugin name for getting MAX_TEXTURES values, because the plugin name should be used from pluginName variable.

##### Pre-Merge Checklist


packages/graphics/src/Graphics.ts
```diff
@@ -1024,7 +1024,7 @@ export class Graphics extends Container
             // but may be more than one plugins for graphics
             if (!DEFAULT_SHADERS[pluginName])
             {
-                const MAX_TEXTURES = renderer.plugins.batch.MAX_TEXTURES;
+                const MAX_TEXTURES = renderer.plugins[pluginName].MAX_TEXTURES;
```
bigtimebuddy, line 1027: Small style improvement:
```suggestion
                const { MAX_TEXTURES } = renderer.plugins[pluginName];
```

bigtimebuddy (commented):
This seems reasonable to me. Can you tell me how you bumped into this error? Would it be possible to create a test that captures this?

SergeiSomin:
hi, thanks, i just tried to understand how pixi rendering works and saw it
I think to test it in theory we could change the renderer plugin to some mocked one and then we could check the amount of samplers in shader if it is the same as MAX_TEXTURES in mocked renderer

bigtimebuddy (approved):
This was introduced here, I think it was just an oversight #6876

bigtimebuddy:
Thank you @SergeiSomin!

PR #8247, opened 2022-03-28
Fix #8114 add Option object to pixels function.

### LINKED ISSUE #8114: extract plugin improvement suggestion
Hello,

This is to suggest an improvement to the extract plugin API.

What I am trying to achieve is fetching a single pixel from an offscreen rendered container. For this purpose, I use the extract image API.

By extracting the image and then rendering it into a canvas I can then read the desired pixel. However, I only need a single pixel from the image and therefore this solution is overkill.

Here is the current usage of the `extract.image` API:
```js
const image = renderer.plugins['extract'].image(container);
```

The `extract.pixels` API is probably faster but still overkill. However it is made unusable by the fact that it is impossible to know the dimensions of the image that the pixels represent:
```js
const pixels = renderer.plugins['extract'].pixels(container);
```
The `pixels` object is an array, it contains no information on width or height.

Ideally, the `extract.pixels` and `extract.image` APIs would accept a scale, a position and a dimension:
```js
const pixels = renderer.plugins['extract'].pixels(spineView.parent, scale, x, y, width, height);
```
The position and dimension would determine which part of the container to clip for extraction purposes. And the scale would define the scaling ratio of the rendering. For instance a 200x300 clipped extraction with a scale of 0.5 would generate a pixel buffer corresponding to a 100x150px dimension.

nouhmaster:
Hi,
We add a Interface Option , this interface content a coordonnée  x,y,scall, height ,width
We have done some tests in the repository ``packages/extract/test/Extract.tests.ts`` and it is corrects.
We would like to know what you think about it and have a review.
Thanks you


packages/extract/src/Extract.ts
```diff
@@ -182,7 +182,9 @@ export class Extract implements IRendererPlugin
      *  to convert. If left empty will use the main renderer
      * @return - One-dimensional array containing the pixel data of the entire texture
      */
-    public pixels(target?: DisplayObject|RenderTexture): Uint8Array
+
+    // any { x: double; y: double; scale : double; Width:double;height :double}
+    public pixels(target?: DisplayObject|RenderTexture, opt?: Options): Uint8Array
```
bigtimebuddy, line 187: Recommend not to abbreviate to "opt", use "options" instead.

packages/core/src/renderTexture/RenderTexture.ts
```diff
@@ -45,7 +45,16 @@ import { deprecation } from '@pixi/utils';
  * ```
  *
  * @memberof PIXI
+ *
  */
+export interface Options {
+    x: number,
+    y: number,
+    height: number,
+    scall: number,
```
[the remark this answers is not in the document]
bigtimebuddy: Also, can we move these options to Extract and not put them on RenderTexture? Also, please consider another name other than "Options" which is a little too generic for an exported type. I'd recommend "PixelExtractOptions"

bigtimebuddy (changes requested):
Thanks for this change! I have a few organizational notes, but otherwise seems good and appreciate the test.

packages/core/src/renderTexture/RenderTexture.ts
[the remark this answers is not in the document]
nouhmaster: I have made the requested changes, I am looking forward your feedback.

packages/extract/src/Extract.ts
```diff
@@ -182,7 +191,9 @@ export class Extract implements IRendererPlugin
      *  to convert. If left empty will use the main renderer
      * @return - One-dimensional array containing the pixel data of the entire texture
      */
-    public pixels(target?: DisplayObject|RenderTexture): Uint8Array
+
+    // any { x: double; y: double; scale : double; Width:double;height :double}
```
bigtimebuddy, line 195: Could you please remove these comments and new line. It's breaks the documentation which needs to immediately precede the method.
```suggestion
```

packages/extract/src/Extract.ts
```diff
@@ -30,6 +30,15 @@ const BYTES_PER_PIXEL = 4;
...
  */
+
+export interface PixelExtractOptions {
+    x: number,
+    y: number,
+    height: number,
+    resolution: number,
+    width: number
+}
+
```
bigtimebuddy, line 41: Can you please move this above the comment above? It's breaking the documentation for this Extract class. 

Also, it would be helpful if you could add comments to PixelExtractOptions. Are all fields required?

packages/core/src/renderTexture/RenderTexture.ts
```diff
@@ -45,7 +45,9 @@ import { deprecation } from '@pixi/utils';
  * ```
  *
  * @memberof PIXI
+ *
  */
+
```
bigtimebuddy, line 50: Unnecessary changes
```suggestion
 */
```

packages/extract/src/Extract.ts
nouhmaster: I add comments , I am looking forward your feedback.

packages/core/src/renderTexture/RenderTexture.ts
```diff
@@ -46,6 +46,7 @@ import { deprecation } from '@pixi/utils';
  *
  * @memberof PIXI
  */
+
```
bigtimebuddy, line 49:
```suggestion
```

packages/extract/src/Extract.ts
```diff
@@ -182,7 +197,8 @@ export class Extract implements IRendererPlugin
      *  to convert. If left empty will use the main renderer
      * @return - One-dimensional array containing the pixel data of the entire texture
      */
-    public pixels(target?: DisplayObject|RenderTexture): Uint8Array
+
```
bigtimebuddy, line 200:
```suggestion
```

bigtimebuddy (approved):
Cleanup

bigtimebuddy:
Thank you @nouhmaster!

packages/extract/test/Extract.tests.ts
```diff
@@ -57,12 +57,13 @@ describe('Extract', function ()
         const extract = renderer.plugins.extract as Extract;
         const renderTexture = RenderTexture.create({ width: 10, height: 10 });
         const sprite = new Sprite(Texture.WHITE);
+        const test: PixelExtractOptions = { x: 1, y: 2, resolution: 5, width: 10, height: 10 };
```
bchevalier, line 60: The name of this variable could be more precise, for instance `renderOpts`.

PR #8251, opened 2022-04-01
Feature: KTXLoader support for loading key-value data from texture

ShukantPal:
##### Description of change

This change is behind the opt-in `settings.KTX_LOAD_KV_DATA` setting. It adds support for the key-value section of the [KTX file format](https://www.khronos.org/registry/KTX/specs/1.0/ktxspec_v1.html#keyAndValue). The key-value section allows for an arbitrary list of UTF-8 string key, binary data value pairs to be stored within the KTX container.

These changes will load the key-value data and assign that map on the `ktxKvData` property of generated base-textures.

### Test it yourself

[test.zip](https://github.com/pixijs/pixijs/files/8394666/test.zip)

You can place the test.ktx file next to an HTML document with the following contents, and serve it locally. Then, open the web page and check its console. There should be a "success" log!

```html
<canvas id="view"></canvas>
<script src="https://pixijs.download/feature/ktx-kv-data/pixi.js"></script>
<script>
    backgroundColor: 0xFFFFFF,
    view: document.getElementById('view'),
    resolution: devicePixelRatio,
    autoDensity: true,
    width: 1024,
    height: 1024
});

PIXI.settings.KTX_LOAD_KV_DATA = true;
PIXI.Loader.shared
    // .add('accelerator.ktx', 'accelerator.ktx')
    // .add('feat.ktx', 'feat.ktx')
    // .add('lookup.ktx', 'lookup.ktx')
    .add('test.ktx', 'test.ktx')
    .load()
    .onComplete.once(() => {
        const texture = PIXI.utils.TextureCache['test.ktx'];
        const kvData = texture.baseTexture.ktxKeyValueData;

        if (!kvData)
            return console.log('Failed - no kvData present');

        const vsl_version = kvData.get('vsl_version');

        if (!vsl_version)
            return console.log('Failed - no vsl_version value');
        if ((new TextDecoder()).decode(vsl_version) !== '1.0.0')
            return console.log('Failed - vsl_version is not 1.0.0');

        return console.log('Success - vsl_version = 1.0.0');
    });
</script>
```


ShukantPal:
@bigtimebuddy I wonder if you know what the CI is failing on. It seems unrelated to the changes I made 🤔 .

bigtimebuddy:
I suspect this might be an npm issue. We are install `>=7` and I think there was a npm v8.6.0 release recently, maybe this is the cause since it's failing on dependencies? Could you test that by changing the Github workflow to install `~8.5.0` instead?

bigtimebuddy:
Related issue https://github.com/npm/cli/issues/4664

bigtimebuddy (commented):
Is there any reason why parsing this data wouldn't be on all the time? Is the `KTX_LOAD_KV_DATA` flag really that meaningful? If this flag is more about experimental usage, then I'd suggest adding it as an opt-in feature flag to the KTXLoader class, e..g. `nextParseKVData`, and not as a setting. Otherwise, you can remove some of the complexity of this PR and just remove the setting.

ShukantPal:
@bigtimebuddy The reason I added an opt-in setting is because it could cause a memory leak if you destroy the resource (since the key-value data is still holding onto the KTX texture buffer). I'm okay with moving it into `KTXLoader` − was really just trying to match the PixiJS pattern for settings 😀.

bigtimebuddy:
In general, I think `settings` are not well-defined, so they've been somewhat misused and it's become the junk-drawer for PixiJS of various flags and configuration properties. I'm in favor of settings that touch numerous places in the code, like `RESOLUTION` or `ROUND_PIXELS` and there's no logical home for them. But a bad example of this is `TARGET_FPMS` because that really only is used in Ticker, and could easily be a Ticker static property instead.

I would say that this use-case is probably more like the latter. Likely only to be used in this one case. KTXLoader seems like a better place for it.

Also, can we talk about the naming... does "kv" stand for "key-value"? If so, just remove that, e.g `LOAD_KTX_DATA`, `ktxData`, etc). Abbreviations should be avoided because they are difficult to learn.

ShukantPal:
@bigtimebuddy I renamed "kv" stuff to "keyValue" and moved the setting as a static field on `KTXLoader`.

bigtimebuddy (approved):

bigtimebuddy:
@ShukantPal could you resolve conflicts?

bigtimebuddy:
Ping @ShukantPal

ShukantPal:
@bigtimebuddy Done!

bigtimebuddy:
Thanks @ShukantPal

Issue #8252, opened 2022-04-01
Inconsistency when setting undefined to BitmapText


miltoncandelero:
Creating a BitmapText with an `undefined` text results in a crash but setting `undefined` after creation doesn't (it assigns an empty string)

### Expected Behavior
Either we crash both or we don't crash on only one

### Current Behavior
`undefined` crashes because we store it directly in the `._text` variable. `.text` is a setter that protects against undefined.

### Possible Solution
Super simple: make the constructor write to `.text` instead of `._text`

### Steps to Reproduce
try rendering this object (provide a font)
`const bitmapFontText = new PIXI.BitmapText(undefined, { fontName: 'Desyrel', fontSize: 55, align: 'left' });`

### Environment

- **`pixi.js` version**: 5+, 6+
- **Browser & Version**: All
- **OS & Version**: All
- **Running Example**: https://www.pixiplayground.com/#/edit/cJevxhf0TXH-N9a1OYkWQ

---

## My biggest question: Do we want to fix this? I can PR this in a couple of days or so
For the record: `Text` doesn't crash when creating and passing `undefined` as text. It assigns an empty string.
This looks like an oversight/bug. I will start working on a PR

Zyie:
Fixed in #8253 

Released in `v6.3.1`

PR #8253, opened 2022-04-01
fixed bitmaptext._text storing undefined values during construction

miltoncandelero:
##### Description of change
Creating a BitmapText with an undefined text results in a crash but setting undefined after creation doesn't (it assigns an empty string)

Fixed by using the same logic for the text setter.

I did not call the setter in `.text` because that setter has other side effects that I was not comfortable triggering.

Fix to #8252 

##### Pre-Merge Checklist

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -208,7 +208,7 @@ export class BitmapText extends Container
         this._tint = tint;
         this._fontName = fontName;
         this._fontSize = fontSize || BitmapFont.available[fontName].size;
-        this._text = text;
+        this._text = String(text === null || text === undefined ? '' : text);
         this._maxWidth = maxWidth;
         this._maxLineHeight = 0;
         this._letterSpacing = letterSpacing;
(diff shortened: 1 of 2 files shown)


packages/text-bitmap/src/BitmapText.ts
```diff
@@ -208,7 +208,7 @@ export class BitmapText extends Container
         this._tint = tint;
         this._fontName = fontName;
         this._fontSize = fontSize || BitmapFont.available[fontName].size;
-        this._text = text;
+        this._text = String(text === null || text === undefined ? '' : text);
```
bigtimebuddy, line 211: Why not this instead?
```suggestion
        this.text = text;
```

miltoncandelero: that would excecute the following

```ts
        text = String(text === null || text === undefined ? '' : text);

        if (this._text === text)
        {
            return;
        }
        this._text = text;
        this.dirty = true;
```

at first glance the text get's turned into an empty string first, compared against undefined and then stored. That wouldn't be an issue. I just wasn't sure it was a good idea to tie the dirty flag (and any other future change to the setter) to the constructor.

However it can totally be done if you think it is not an issue 👍

bigtimebuddy: `dirty` is also set to true in the constructor, so I don't think that's an issue

bigtimebuddy (approved):

PR #8274, opened 2022-04-09
Chore: Upgrade to TypeScript 4.3.0 and make setters for ObservablePoints accept IPointData

ShukantPal:
Fixes #8258

PR #8275, opened 2022-04-09
Chore: Add build:types to dist workflow

ShukantPal:
#8274 passed the build even though `build:types` failed locally. This should prevent such a regression.


bigtimebuddy (approved):
Thanks!

PR #8301, opened 2022-04-22
Fix texture typings

AshsHub:
## Description
Allow for the `source` and `url` parameters in the `Texture.from` and `Texture.fromURL` respectively to receive an array. Originally the function already allowed for the creation of a cube map texture using an array of six texture sources, but the typing didn't.

## Pre-Merge Checklist


bigtimebuddy:
Nice fix, thank you @AshsHub.

Could you double check the Texture tests to see if there's an example of loading a cube texture? I think there might be, but doesn't hurt to add.

AshsHub:
Created new tests and moved existing tests to new relevant `Suite`

bigtimebuddy (approved):

bigtimebuddy:
Thank you @AshsHub

PR #8307, opened 2022-04-27
Chore: refactor DOM-based side effects

### LINKED ISSUE #6990: "window is undefined" in nextjs
- FYI: going forward this will likely continue to be an issue due to the popularity of nextjs.
- The root cause is that pixi does not work in node.js so it blows up when run in SSR. 
- This is related to an old bug with pixi.js v4: https://github.com/pixijs/pixi.js/issues/3224
- The workaround is to use dynamic imports to remove it from the SSR phase. See https://medium.com/frontend-digest/why-is-window-not-defined-in-nextjs-44daf7b4604e
- Unfortunately, this is not an ideal solution.
- The use case is a React component that uses pixi.js under the hood. https://gitlab.com/robotandkid/awesome-react-cutscenes
- Because of the dynamic import, this component is unable to manage it's SSR lifecycle unless it includes nextjs-specific code. 
- But really all that needs to happen is for pixi to abort loading if the window is not present. (Something like wrapping everything in a giant `if (typeof window === 'undefined')` :-) )

### Expected Behavior

- pixi errors gracefully when the `window` object is not present

### Current Behavior

- pixi cannot run in SSR

### Possible Solution

- Something like wrapping everything in a giant `if (typeof window === 'undefined')` :-) 

### Environment

- pixi.js - v5.3.3

- **`pixi.js` version**: _e.g. 4.7.1_
- **Browser & Version**: _e.g. Chrome 67_
- **OS & Version**: _e.g. Ubuntu 18.04_
- **Running Example**: _url_

bigtimebuddy:
Fixes #6990

This change removes all the DOM calls at instantiation-time. As we work towards cleaning up the code to support Node.js rendering, this is an important first step to make sure DOM dependencies aren't called immediately. All code can now cleanly be require/import'd without barfing.

### DIFF
--- a/packages/text/src/TextMetrics.ts
+++ b/packages/text/src/TextMetrics.ts
@@ -57,7 +57,7 @@ export class TextMetrics
     public static HEIGHT_MULTIPLIER: number;
 
     // TODO: These should be protected but they're initialized outside of the class.
-    public static _canvas: HTMLCanvasElement|OffscreenCanvas;
+    private static __canvas: HTMLCanvasElement|OffscreenCanvas;
     public static _context: CanvasRenderingContext2D|OffscreenCanvasRenderingContext2D;
     public static _fonts: { [font: string]: IFontMetrics };
     public static _newlines: number[];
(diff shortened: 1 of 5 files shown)


packages/text/src/TextMetrics.ts
```diff
@@ -57,8 +57,8 @@ export class TextMetrics
     public static HEIGHT_MULTIPLIER: number;
 
     // TODO: These should be protected but they're initialized outside of the class.
```
andrewstart, line 59: Should the new getters be protected, per this comment? They are no longer initialized outside the class.

bigtimebuddy: I'll do the cleanup in a new PR. I didn't want to change the API in this one and be 100% backwards, but I agree that it should be fixed. I'll at least update the TODO comment.

PR #8311, opened 2022-04-30
Fix triangulation of circular shape with radius 0

dev7355608:
A circle with radius 0 produces this geometry in 6.3.0:
```js
g.geometry.points
// Array(4) [ 200, 300, undefined, undefined ]
g.geometry.indices
// Array(3) [ 1, 0, 1 ]
```

##### Pre-Merge Checklist


bigtimebuddy:
Test?

dev7355608:
The `undefined` (uploaded as NaN) seem to be harmless (https://www.pixiplayground.com/#/edit/RT8Yj3NSoQxI1BeKVcDE2).

dev7355608:
We should fix this in 6.5.0. The new bounds calculation of graphics is potentially affected by the `undefined` values in `points`.

bigtimebuddy (approved):

PR #8312, opened 2022-04-30
Don't render rectangles with negative width/height

dev7355608:
##### Description of change

Rectangles with negative width/height have always been rendered inverted, but technically such rectangle is empty according to the implementation of `Rectangle.contains`. Therefore it should not be rendered. Circles with negative radius are also not rendered (even before the new implementation of `buildCircle`).

This is probably a breaking change.

##### Pre-Merge Checklist


bigtimebuddy:
Also... tests?

dev7355608:
Demo: https://www.pixiplayground.com/#/edit/SxA5Du1KEcFLwp1_Ks8Pr

bigtimebuddy:
We have a lot of issues like this all over Pixi. This falls into the category of undefined behavior and I don't think bailing is the right solution. What would be better is to add an assert. These are removed in the prod/min version and are useful for out of bounds or invalid inputs. Negative numbers here shouldn't be allowed, the fact that it maybe worked before is not consequential.

bigtimebuddy:
Had a chat with @GoodBoyDigital and @Zyie about this one.

What about normalizing negative values instead? For instance:

```js
drawRect(0, 0, -200, -100)
// equalivalent to
drawRect(-200, -100, 200, 100)
```

dev7355608:
`drawRect(x, y, w, h)` should not differ from `drawShape(new Rectangle(x, y, w, h))`, but if `drawRect`/`drawShape` modifies the dimensions, it modifies the rectangle argument (the shapes passed to graphics are not cloned). So in consequence the rectangles would need to be equivalent:
```js
new Rectangle(0, 0, -200, -100)
new Rectangle(-200, -100, 200, 100)
```

bigtimebuddy:
Yeah exactly. conceptually we should formally support negative values. One example of this is clicking and dragging a marquee. Drawing this with Graphics is much more handy to do if you can use negative values.

bigtimebuddy (approved):
I changed my mind about this. I think this is fine for v7.

PR #8314, opened 2022-05-02
Fix text updates

### LINKED ISSUE #8296: Text sometimes Draws at Wrong Location when doing Render to Texture
### Expected Behavior
left and right output look the same

### Current Behavior
rendering on the right-side has text in the wrong place
image

### Possible Solution
this happens with Text, but not Graphic or Sprite, so appears to be a bug with PIXI.Text
experimentally, if in Text.updateTexture, I remove the call to this._recursivePostUpdateTransform(); it works, but I don't think that is a correct fix.  somehow, the translation of a parent is applied when it shouldn't be.

### Steps to Reproduce
Basically, just text in a container with a parent that isn't at 0, 0 and then render that container to a Texture.
live playground with the code (<50 lines)
https://www.pixiplayground.com/#/edit/A9i0edGpz2lNDbARzQwe3

### Environment
- PIXI.JS 6.2.2
- Browser Chrome Version 100.0.4896.127 (Official Build) (arm64)
- OS MacOS 12.0
- Running Example https://www.pixiplayground.com/#/edit/A9i0edGpz2lNDbARzQwe3

dev7355608:
##### Description of change

Fixes #8296.

I removed `_recursivePostUpdateTransform`: this function must never be called during rendering because transforms should never change while rendering, and because it can mess up all the transforms when rendering a subtree. So the text is updated in `updateTransform` now. The text is also still updated in `_render` but (assuming `updateTransform` was called before) this should only trigger an update if the resolution changed. Therefore the width/height after this update should be almost the same (+-1 pixel) and doesn't need a transform updates (technically it could, but we cannot/shouldn't update the transform while rendering). Because `updateText` doesn't update the transform anymore, I had to override `getBounds` and force a transform update if the texture changed.

##### Pre-Merge Checklist


JetLua:
[This](https://github.com/pixijs/pixijs/pull/6751) is the old PR and you can test the old bug by the way.

dev7355608:
The test of the old PR succeeds.

bigtimebuddy (approved):
Good fix @dev7355608

PR #8298, opened 2022-04-21
Favor Discord for PixiJS community chat

bigtimebuddy:
We have decided to shift our online community focus to Discord away from Slack and Gitter.

PR #8319, opened 2022-05-04
The discord url was an expired one

miltoncandelero:
Changed the URL with a never-expire one

Bug found by a user in a comment of #8298


bigtimebuddy (approved):
There are

epaezrubio:
Thanks for the quickfix!

PR #8330, opened 2022-05-11
Update color uniform of MeshMaterial if alpha mode changes

dev7355608:
##### Description of change

The color uniform has to be updated if the texture is changed and with it the alpha mode of the texture.

##### Pre-Merge Checklist


GoodBoyDigital (approved):
nice catch! thank you!

PR #8331, opened 2022-05-11
Remove AbstractRenderer, move its functionality to Systems

GoodBoyDigital:
##### Description of change

# **Problem**
Pixi uses systems to compose its logic and inheritance (abstract renderer). Two methods of sharing functionality is a not a very good code structure!

# **Solution**
We remove the abstract render, and replace it with a set of new systems. These systems can be shared by both the canvas and the WebGL renderer. (And webGPU in the future!)

**All renderers**

Introduction of `SystemManager` - a base for managing systems (used by both renderers). This is super general, just a few helpers to manage adding systems and runners. 

`AbstractRenderer` is no more! It has been replaced by `IRenderer` interface. The actual functionality has been split out into the following systems:

`BackgroundSystem` - for background color management
`PluginSystem` - managing plugins
`StartupSystem` - starts up our renderers
`GenerateTextureSystem` - generates texture code

**WebGL renderer**
`RenderSystem` - functionality for rendering a Pixi Display object
`MultisampleSystem` - has the multisample code that was sitting in the Renderer previously.

**Canvas renderers**
Much like `Renderer`,  `CanvasRenderer` is now also a collection of systems and implements `IRenderer`. Its nice to see them both aligned! Canvas does need more work as I think the Context System could be split out a little more, but this is fine for now. New Systems include:

`CanvasContextSystem` - manages active contexts, blendModes and context transforms
`CanvasMaskSystem` - manages Masks (just a tweaked CanvasMaskManager)
`CanvasRenderSystem` - functionality for rendering a Pixi Display Object

Notes for Reviewers..
- Please take note of the docs, I moved a lot around!
- I am planning on renaming RenderSystem and ViewSystem to something better
- @dev7355608 could you check [this change](https://github.com/pixijs/pixijs/blob/01f4810c11338091f1cd28e70cbeadd87a738c4f/packages/core/src/render/RenderSystem.ts#L107) is ok pls

I have really tried to follow the success we found having one system care about one thing (even if its small). That way our code is easy to manage and will scale in a much better way!

This PR passes all current tests (only a few tweaks!) and should be _**non breaking**_ despite the size of it!

**Next Steps**
Replace Plugins so that they are added as systems under the hood (this is going to be super easy and not break any existing plugins :D). This will bring us to a single flexible way to manage Pixi and adding additional functionality 

##### Pre-Merge Checklist


packages/accessibility/src/AccessibilityManager.ts
```diff
@@ -4,8 +4,8 @@ import { accessibleTarget } from './accessibleTarget';
 
 import type { Rectangle } from '@pixi/math';
 import type { Container } from '@pixi/display';
-import type { Renderer, AbstractRenderer } from '@pixi/core';
 import type { IAccessibleHTMLElement } from './accessibleTarget';
+import { IRenderer } from 'packages/core/src/IRenderer';
```
bigtimebuddy, line 8: Please be mindful that VSCode doesn't auto-import correctly for monorepo. You'll need to fix this in a few places it seems.
```suggestion
import type { IRenderer } from '@pixi/core';
```

packages/core/src/startup/StartupSystem.ts
```diff
@@ -0,0 +1,48 @@
...
+import type { IRendererPlugins } from '../plugin/PluginSystem';
+import { IRenderer } from '../IRenderer';
+import { ISystem } from '../system/ISystem';
+
+// TODO this can be infered by good use of generics in the future..
+export interface StartupOptions extends Record<string, unknown> {
+    _plugin: IRendererPlugins,
+    background: BackgroundOptions,
+    _view: ViewOptions,
+}
```
Zyie, line 13: I see the todo to make this a generic which is a good idea

However, this is missing an optional `ContextOptions` for use in `Renderer`

packages/core/src/IRenderer.ts
```diff
@@ -0,0 +1,117 @@
...
+    transparent?: boolean;
+    autoDensity?: boolean;
+    antialias?: boolean;
+    resolution?: number;
+    preserveDrawingBuffer?: boolean;
+    clearBeforeRender?: boolean;
+    backgroundColor?: number;
+    backgroundAlpha?: number;
+    powerPreference?: WebGLPowerPreference;
+    context?: IRenderingContext;
```
Zyie, line 63: These used to live in `AbstractRenderer` so would need getter/setters so its not a breaking change

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -182,289 +246,113 @@ export class CanvasRenderer extends AbstractRenderer
...
+
+     get lastObjectRendered(): IRenderableObject
+     {
+         return this._renderer.lastObjectRendered;
+     }
+
+     get renderingToScreen(): boolean
+     {
+         return this._renderer.renderingToScreen;
+     }
```
Zyie, line 356: `CanvasRenderer` is missing the getter for `blendModes`, `maskManager`, `refresh`, `rootContext`, `smoothProperty`, `setBlendMode`, `invalidateBlendMode`, `setContextTrasform`

I'm guessing `maskMaanger` should be a getter for `CanvasMaskSystem`?

Zyie (commented):
There seems to be a few places missing documentation that previously existed

Also some of these comments are based off making this a non breaking change

packages/core/src/startup/StartupSystem.ts
GoodBoyDigital: cool, added!

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
GoodBoyDigital: getters added..

packages/canvas/canvas-graphics/src/CanvasGraphicsRenderer.ts
```diff
@@ -1,10 +1,11 @@
 import { Texture } from '@pixi/core';
 import { SHAPES, Matrix } from '@pixi/math';
-import { canvasUtils, CrossPlatformCanvasRenderingContext2D } from '@pixi/canvas-renderer';
+import { canvasUtils } from '@pixi/canvas-renderer';
 import type { CanvasRenderer } from '@pixi/canvas-renderer';
 import type { FillStyle, Graphics, GraphicsData, LineStyle } from '@pixi/graphics';
 import type { Circle, Ellipse, Polygon, Rectangle, RoundedRectangle } from '@pixi/math';
 import { PolygonUtils } from './utils/PolygonUtils';
+import { CrossPlatformCanvasRenderingContext2D } from 'packages/canvas/canvas-renderer/src/CanvasContextSystem';
```
Zyie, line 8: Looks like this is the only broken import left to fix

packages/core/src/IRenderer.ts
GoodBoyDigital: updated!

packages/core/src/render/RenderSystem.ts
```diff
@@ -0,0 +1,126 @@
...
+
+        displayObject.render(renderer);
+
+        // apply transform..
+        renderer.batch.currentRenderer.flush();
+
+        if (renderTexture)
+        {
+            // @dev7355608 please confirm this is ok to be here?
+            renderer.framebuffer.blit();
```
dev7355608, line 107: The blit should be optional, because there are cases where you might want to blit to some other texture instead (see #6844). If the default is set to true, we need to remove the blit in `Renderer#generateTexture` and cache-as-bitmap to avoid an unnecessary second blit. The documentation of the `multisample` property also needs to be updated.

GoodBoyDigital: thanks @dev7355608 ! I will update

GoodBoyDigital: done! mind taking a second look pls @dev7355608 ? Cheers!

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
Zyie: `blendModes`, `maskManager`, `refresh` are missing documentation from `CanvasRenderer`

`preserveDrawingBuffer` and `useContextAlpha` are missing

packages/core/src/system/SystemManager.ts
```diff
@@ -0,0 +1,154 @@
...
+            const systemName = systemHashKeys.find((systemId) => this._systemsHash[systemId] === system);
+
+            system[runner.name](options[systemName]);
+        });
+    }
+
+    /**
+     * destroy the all runners and systems. Its apps job to
+     */
+    destroy(): void
```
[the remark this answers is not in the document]
GoodBoyDigital: ooh! this guide is useful - will follow it as i move forward!

packages/canvas/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -182,289 +247,284 @@ export class CanvasRenderer extends AbstractRenderer
...
+     {
+         return this.background.clearBeforeRender;
+     }
+
+     // deprecated zone..
+
+     get blendModes(): string[]
+     {
+         // eslint-disable-next-line max-len
+         deprecation('6.4.0', 'Renderer#preserveDrawingBuffer has been deprecated, we cannot truly know this unless pixi created the context');
```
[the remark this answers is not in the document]
GoodBoyDigital: fixed

packages/core/src/Renderer.ts
```diff
@@ -615,6 +499,230 @@ export class Renderer extends AbstractRenderer
...
+        // eslint-disable-next-line max-len
+        deprecation('6.4.0', 'renderer.backgroundAlpha has been deprecated, use renderer.background.alpha instead.');
+
+        this.background.alpha = value;
+    }
+
+    get powerPreference(): WebGLPowerPreference
+    {
+        // eslint-disable-next-line max-len
+        deprecation('6.4.0', 'renderer.powerPreference has been deprecated, we can only know this if pixi creates the context');
```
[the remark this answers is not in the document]
bigtimebuddy: That's right. This means they are removed from the minified build.

packages/core/src/system/SystemManager.ts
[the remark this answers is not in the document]
bigtimebuddy: I wish we could lint the docs for style, but yes. I think those guidelines are still useful.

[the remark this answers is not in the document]
Zyie: we could try using [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc)

I'll do a bit of experimenting to see if its viable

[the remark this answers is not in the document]
Zyie: I created a draft PR to show what it would look like to use this plugin here #8342

packages/core/src/view/ViewSystem.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Measurements of the screen. (0, 0, screenWidth, screenHeight).
+     *
+     * Its safe to use as filterArea or hitArea for the whole stage.
+     *
+     * @member {PIXI.Rectangle}
+     */
+
+    public screen: Rectangle;
```
bigtimebuddy, line 50:
```suggestion
    public screen: Rectangle;

```

packages/core/src/Renderer.ts
```diff
@@ -615,6 +499,246 @@ export class Renderer extends AbstractRenderer
...
+        return this._view.view.width;
+    }
+
+    /**
+     * Same as view.height, actual number of pixels in the canvas by vertical.
+     * @default 600
+     */
+    get height(): number
+    {
+        return this._view.view.height;
```
bigtimebuddy, line 532: I agree that this is kinda weird `view.view`. I think ViewSystem makes a lot of sense actually and probably would like to keep that. What about renaming the `view` property in ViewSystem instead? Maybe something like `element` (or alternatively `canvas`). 

```
renderer._view.element.width
renderer._view.element.height
```

"View" is much more general term and includes things like resolution, the screen-size, resizing etc. Where the actual rendering surface (HTMLCanvasElement, NodeCanvas, etc) can be more specific.

GoodBoyDigital: like that! renamed to element.

packages/accessibility/src/AccessibilityManager.ts
```diff
@@ -280,15 +280,15 @@ export class AccessibilityManager
 
         this.androidUpdateCount = now + this.androidUpdateFrequency;
 
-        if (!(this.renderer as Renderer).renderingToScreen)
+        if (!(this.renderer as IRenderer).renderingToScreen)
```
bigtimebuddy, line 283: Is this casting even necessary anymore?

packages/canvas/canvas-renderer/src/CanvasObjectRendererSystem.ts
```diff
@@ -0,0 +1,184 @@
...
+
+        _context.activeResolution = rootResolution;
+        _context._projTransform = null;
+
+        renderer.emit('postrender');
+    }
+
+    public destroy(): void
+    {
+        // ka boom!
```
bigtimebuddy, line 182: Can you remove references here to lastObjectRenderer and renderer?

bigtimebuddy (approved):
Overall, I think this is a significant improvements. There will undoubtedly be follow up to this. Off the top of my head this would be:

* Remove PluginSystem and replace with PipelineSystem
* Migrate Events, Extract, Accessibility and Prepare be Systems
* Deprecate Interaction package
* Move other plugins to pipelines (e.g. BatchRenderer, TilingSpriteRenderer, etc)
* Add GlobalMixins for Renderer, CanvasRenderer to make them extensible
* Make the `systemConfig` (at least the systems) be static so that plugins can be added before instantiation
* Remove v6 deprecations

bigtimebuddy:
Also, @GoodBoyDigital could you resolve these conflicts?

packages/canvas/canvas-renderer/src/CanvasObjectRendererSystem.ts
GoodBoyDigital: done!

bigtimebuddy:
Sorry @GoodBoyDigital, I have another gnarly merge after I updated to dev. I tried to resolve it, but I think you or @Zyie should.

PR #8341, opened 2022-05-13
fix typescript eslint rules

Zyie:
~~This is based off of [this PR](https://github.com/pixijs/eslint-config/pull/6) being approved
I've auto fixed a couple of classes to demonstrate that type of changes it will be making~~

Update: 

PR has been update to fix all outstanding issues after upgrading to v4 of the eslint config


.eslintrc.json
```diff
@@ -2,6 +2,67 @@
...
+    "project": ["./tsconfig.eslint.json"]
+  },
+  "overrides": [
+    {
+      "files": ["*.ts"],
+      "extends": ["plugin:@typescript-eslint/recommended"],
+      "rules": {
+        "@typescript-eslint/no-explicit-any": 0,
+        "@typescript-eslint/interface-name-prefix": 0,
+
```
Zyie, line 15: These changes will be removed if the eslint PR get merged

bigtimebuddy: Made version v4.0.0 of eslint-config with these changes. Can you update?

.eslintrc.json
```diff
@@ -2,6 +2,18 @@
...
+  },
+  "overrides": [
+    {
+      "files": ["*.ts"],
+      "rules": {
+        "@typescript-eslint/comma-dangle": [1, "always-multiline"],
+        "comma-dangle": "off"
+      }
+    }
+  ],
```
Zyie, line 16: The `@typescript-eslint/comma-dangle` rule requires the `parserOptions` for project aware rules and since the config is different per project i've added in manually here instead of in the eslint config

.eslintrc.json
Zyie: Updated!

packages/events/test/EventBoundary.tests.ts
```diff
@@ -16,7 +16,7 @@ describe('EventBoundary', function ()
         container.addChild(new Graphics().drawRect(80, 0, 100, 100));
 
         const target = container.addChild(
-            new Graphics().drawRect(150, 0, 100, 100)
+            new Graphics().drawRect(150, 0, 100, 100),
```
bigtimebuddy, line 19: This seems wrong to me, comma dangle shouldn't be on function calls.

packages/filters/filter-blur/src/BlurFilterPass.ts
```diff
@@ -58,7 +58,7 @@ export class BlurFilterPass extends Filter
      * @param clearMode - How to clear
      */
     public apply(
-        filterManager: FilterSystem, input: RenderTexture, output: RenderTexture, clearMode: CLEAR_MODES
+        filterManager: FilterSystem, input: RenderTexture, output: RenderTexture, clearMode: CLEAR_MODES,
```
bigtimebuddy, line 61: Yeah, same here: no a fan of arguments having comma dangle.

packages/events/test/EventBoundary.tests.ts
Zyie: yep this was my mistake. Fixed in [this PR](https://github.com/pixijs/eslint-config/pull/7)

packages/compressed-textures/src/loaders/DDSLoader.ts
```diff
@@ -63,7 +63,8 @@ const DDS_DX10_FIELDS = {
  */
 // This is way over-blown for us! Lend us a hand, and remove the ones that aren't used (but set the remaining
 // ones to their correct value)
-enum DXGI_FORMAT {
+enum DXGI_FORMAT
+    {
```
bigtimebuddy, line 67: Why are enums getting this extra indent on the next line?

bigtimebuddy (commented):
With the exception of the enum thing this is great. Good work cleaning this up

packages/compressed-textures/src/loaders/DDSLoader.ts
bigtimebuddy: Seems like a known bug https://github.com/typescript-eslint/typescript-eslint/issues/3225

bigtimebuddy:
Would like to get this one in first before the jsdoc linting. Do you think you could add exceptions for the enum and we could call it?

Zyie:
> Would like to get this one in first before the jsdoc linting. Do you think you could add exceptions for the enum and we could call it?

Yeah I was looking around yesterday/today to see if there was a solution and couldn't find anything. I'll add the exceptions

bigtimebuddy (approved):

bigtimebuddy:
Thanks @Zyie

andrewstart:
I also have run into that enum rule - the impression I got from looking up how to work around it was that using `@typescript-eslint/indent` was not recommended in general (even though it is so nice otherwise).

PR #8344, opened 2022-05-13
Fix for 'attributes is null' exception on mobile (fixes #8216)

### LINKED ISSUE #8216: TypeError: Cannot read properties of null (reading 'stencil') on mobile render
### Expected Behavior
On mobile render
when i am destroying pixi app and creating new instance on same canvas element i have last frame without errors;

### Current Behavior
For mobile render:
when i am destroying pixi app and creating new instance on same canvas element i have clean canvas and error
 **Uncaught TypeError: Cannot read properties of null (reading 'stencil')**
For desktop render:
when i am destroying pixi app and creating new instance on same canvas element i have last frame;

### Steps to Reproduce
open https://codesandbox.io/s/crazy-snow-i0e7p3?file=/src/index.js
click on open in new tab
in new tab toggle device toolbar 
then reload page
and you will get error  **Uncaught TypeError: Cannot read properties of null (reading 'stencil')**
### Environment

- **`pixi.js` version**: _e.g. 6.0.4 - 6.2.2_
- **Browser & Version**: _e.g. Chrome 98_ (mobile render)
- **OS & Version**: _e.g. MacOS BigSur 11.3.1
- **Running Example**: https://codesandbox.io/s/crazy-snow-i0e7p3?file=/src/index.js

jaxreiff:
##### Description of change
When context is lost in some scenarios (e.g. re-rendering canvas in a react component), `gl.getContextAttributes()` can return null, causing an exception while checking the `stencil` attribute to log this warning. This only occurs on some mobile environments (it can be reproduced in desktop browser "Responsive Design Mode" however).

This change adds an additional check to make sure `context` is defined before checking the `stencil` attribute. If context is lost it simply doesn't log the warning as the warning wouldn't necessarily be accurate.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/context/ContextSystem.ts
+++ b/packages/core/src/context/ContextSystem.ts
@@ -280,7 +280,7 @@ export class ContextSystem implements ISystem
         }
 
         // this is going to be fairly simple for now.. but at least we have room to grow!
-        if (!attributes.stencil)
+        if (attributes && !attributes.stencil)
         {
             /* eslint-disable max-len, no-console */
             console.warn('Provided WebGL context does not have a stencil buffer, masks may not render correctly');


bigtimebuddy (approved):

bigtimebuddy:
Thank you @jaxreiff :tada:

PR #8349, opened 2022-05-15
Fix generateTexture

### LINKED ISSUE #8323: Renderer.generateTexture used wrong default region
### Expected Behavior
render.generateTexture(Obj) should create texture with exact bounds of the obj even the obj is a root element with translation of position

### Current Behavior
If an obj is root element and it has tx or ty in transformation, getLocalBounds(null,true) is not the correct default region for its content thus creating wrong texture

### Steps to Reproduce
Here is a playground for reproduce: https://www.pixiplayground.com/#/edit/R5sqRDgqDm4pqdjNOc3-G

### Environment

- **`pixi.js` version**: 6.3.2
- **Browser & Version**: Chrome
- **OS & Version**: MacOS
- **Running Example**:  https://www.pixiplayground.com/#/edit/R5sqRDgqDm4pqdjNOc3-G

dev7355608:
##### Description of change

Fixes #8323.

- Translation:
  - Before: https://www.pixiplayground.com/#/edit/4RNT5xAO7Ib3to95keGwX
  - After: https://www.pixiplayground.com/#/edit/_skWND2rC6Vn4xQrwp4t1
- Scale:
  - Before: https://www.pixiplayground.com/#/edit/Xn7Ul6dFFCvOJurzNjEDn
  - After: https://www.pixiplayground.com/#/edit/do--rno0zbg5bfma3b5UI
- Rotation:
  - Before: https://www.pixiplayground.com/#/edit/mrpVuY-vo7yEWN-ZzaKdK
  - After: https://www.pixiplayground.com/#/edit/3hC6AWUZN0M316a2iuVrG

If the transform of the display object is anything other than the identity, the current implementation of `generateTexture` creates incorrect results. We need to temporarily change the transform of the display object to the identity, because we want to render in local space.

I also removed the `clear: false`. Don't you have to clear a new framebuffer/texture at least once before you use can it? Correct me if I'm wrong though. I'm not sure whether the spec guarantees that a framebuffer is initialized with zeros. We had the problem with stencil buffer where we used it before clearing it, and that wasn't working an some machines. That's why I'd prefer to clear: just to be safe.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/AbstractRenderer.ts
+++ b/packages/core/src/AbstractRenderer.ts
@@ -1,5 +1,5 @@
 import { hex2string, hex2rgb, EventEmitter, deprecation } from '@pixi/utils';
-import { Matrix, Rectangle } from '@pixi/math';
+import { Matrix, Rectangle, Transform } from '@pixi/math';
 import { MSAA_QUALITY, RENDERER_TYPE } from '@pixi/constants';
 import { settings } from '@pixi/settings';
 import { RenderTexture } from './renderTexture/RenderTexture';
@@ -10,6 +10,7 @@ import type { IRenderingContext } from './IRenderingContext';
 import type { IRenderableContainer, IRenderableObject } from './IRenderableObject';
 
 const tempMatrix = new Matrix();
+const tempTransform = new Transform();
 
 export interface IRendererOptions extends GlobalMixins.IRendererOptions
 {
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):
lgtm!

bigtimebuddy:
This is a good change, but is breaking which is why we decided to put it in v7.0.0. Undoubtedly people are working around this issue now and don't want to blow people up.

Also, v7 is going to happen soon, so we won't have to wait too long for this.

PR #8353, opened 2022-05-19
Add color masks

dev7355608:
##### Description of change

I added color masks that restrict rendering to the selected color channels. This is useful to render something different in each channel of a render texture. It was possible before with `gl.colorMask` but that only worked as long as no stencil mask were involved. First I thought about adding  `colorMask` to `State`, but that doesn't really work well, because `colorMask` is used for stencil masks and the mask object's state would override the color mask set by stencil system. So adding the color mask as a mask seems more natural. The color mask stack multiplicatively as one would expect, which wouldn't be possible if it was part of the state.
```
Parent: R G _ A
Child:  _ G B A
----------------
Mask:   _ G _ A
```

I tried my best adding this feature cleanly it to `MaskData`, which was designed around a mask object, but of course color mask don't need a mask object.

Demo: https://www.pixiplayground.com/#/edit/H-_PmqoWbCBSbuxhXXYqX

##### Pre-Merge Checklist


GoodBoyDigital (approved):
thanks for this @dev7355608 !

PR #8357, opened 2022-05-22
Feature: added "resolution" property in "BitmapText" as in "Text"

Aysen69:
##### Description of change
Added "resolution" property in "BitmapText" as in "Text" to take more control of MSDF rendering

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -82,6 +82,16 @@ export class BitmapText extends Container
     public dirty: boolean;
 
     /**
+     * The resolution / device pixel ratio of the canvas.
+     *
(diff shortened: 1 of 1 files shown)


bigtimebuddy (commented):
Awesome! Great add. Can you add a unit test or two to make sure resolution is handled correctly ?

Aysen69:
added the unit test

bigtimebuddy (approved):
Looks good!

miltoncandelero:
Is there a way to return to auto resolution after setting it manually? Do we want/care for that?

Aysen69:
I don't think so. Because "Text" don't.

Aysen69:
Otherwise, we'd need do the same in "Text" class.

bigtimebuddy:
Thanks @Aysen69 !

PR #8364, opened 2022-05-27
Assets Package

GoodBoyDigital:
##### Description of change
Shiny new Assets Manager

This is draft as still got a bunch of stuff to tie up
but made it visible for peeps!
 
##### Pre-Merge Checklist


packages/assets/src/loader/Loader.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Use this to add any plugins to the loadAssets function to use
+     *
+     * @param newParsers - a array of parsers to add to the loader, or just a single one
+     */
+    public addParser(...newParsers: LoaderParser[]): void
+    {
+        this.plugins.push(...newParsers);
+    }
```
andrewstart, line 47: Suggestion: standardize on either "plugins" or "parsers", instead of mixing terms.

bigtimebuddy:
My biggest concern is that I'd like to introduce this in v6 and then make it the default in v7 (removing loaders). Can we make this package optional in v6 and move this PR to `dev`?

packages/assets/src/loader/Loader.ts
GoodBoyDigital: thanks @andrewstart ! yeah, that some legacy bits there! going with parsers!

miltoncandelero:
Is this the end of the TextureCache? 🤔

GoodBoyDigital:
closing as pushing stuff again onto dev..

GoodBoyDigital:
replaced with: https://github.com/pixijs/pixijs/pull/8365

PR #8370, opened 2022-05-30
Update tests to fix TS issues

Zyie:
This PR updates all the tests to fix the TS/ESLint issues

todo: 
- get type checking working through `package.json` command
- clean up remaining TS issues


packages/canvas/canvas-graphics/test/CanvasGraphicsRenderer.tests.ts
```diff
@@ -4,17 +4,17 @@ import { CanvasRenderer } from '@pixi/canvas-renderer';
...
+    it('should not create pattern for White Texture', () =>
     {
         const renderer = new CanvasRenderer({ width: 1, height: 1 });
 
         renderer.plugins.graphics = new CanvasGraphicsRenderer(renderer);
 
         const graphics = new Graphics();
 
-        graphics.beginFill({ texture: Texture.WHITE });
+        graphics.beginTextureFill({ texture: Texture.WHITE });
```
Zyie, line 17: Not entirely sure what the intent for this test was. The usage suggests it was meant to be `beginTextureFill`

packages/core/test/FramebufferSystem.tests.ts
```diff
@@ -3,93 +3,95 @@ import { MSAA_QUALITY } from '@pixi/constants';
 import { Rectangle } from '@pixi/math';
 import { expect } from 'chai';
 
-describe('FramebufferSystem', function ()
+describe('FramebufferSystem', () =>
 {
-    before(function ()
+    let renderer: Renderer;
```
bigtimebuddy, line 8: One of the benefits of using Mocha is that the scope is maintained for tests. So you can set properties on `before` or `beforeEach`. Then the tests can use that same scope. Using fat-arrow functions removes this scoping benefit. So you end up having to do a lot of variable scoping like this. It was intentional to leave the use of `function` in tests and not use fat-arrows. Could you explain why you feel this approach is beneficial?

packages/core/test/TextureSystem.tests.ts
```diff
@@ -100,7 +103,7 @@ describe('TextureSystem', function ()
             glTexture.samplerType = SAMPLER_TYPES.INT;
             if (renderer.context.webGLVersion === 2)
             {
-                glTexture.internalFormat = renderer.context.gl.RGBA32I;
+                glTexture.internalFormat = renderer.context['gl'].RGBA32I;
```
bigtimebuddy, line 106: These variable are using manual look up (e.g., `['gl']`) because they are private, right?

GoodBoyDigital (approved):

packages/core/test/TextureSystem.tests.ts
Zyie: That is correct!

packages/core/test/FramebufferSystem.tests.ts
Zyie: The main reason I used fat arrows is because jest doesn't support scoping like this

Also these added variables would always have the `any` type, so I had to keep casting the variable which was less clean

Issue #8368, opened 2022-05-29
PIXI types not loading in ESM project


zwade:
Circa the latest version of pixi, I am unable to have typescript resolve the pixi import because it cannot find the associated types file. This is likely due to it resolving the import to `dist/cjs/pixi.js`, and attempting to find the `.d.ts` in that folder instead of `index.d.ts`.

### Expected Behavior

pixi.js is able to be included by typescript in ESM mode

### Current Behavior

Typescript is unable to find the types for it

### Possible Solution

This can be fixed via adjustments to the `package.json` (in my case at least, not sure if there are other cases I'm not thinking about)

For more detail, see [this page](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#package-json-exports-imports-and-self-referencing) from the ts 4.7 release notes

```json
{
  "exports": {
    ".": {
       "import": {
          "default": "dist/esm/pixi.js",
          "types": "index.d.ts"
       },
       "require": {
          "default": "dist/cjs/pixi.js",
          "types": "index.d.ts"
       }
    }
  },
  "types": "index.d.ts"
}
```

### Steps to Reproduce

Create a new project with ts 4.7 and pixi. In the compiler options, set `module` to `NodeNext` and in the `package.json` set `type` to `module`. Create a new typescript file and import something from `pixi.js`. Typescript will throw an error.

### Environment

- **`pixi.js` version**: 6.3.2
- **`typescript` version**: 4.7.2
- **Browser & Version**: N/A
- **OS & Version**: MacOS 12.1
- **Running Example**: N/A (could probably throw one together if needed)

zwade:
(Oh also, I'm happy to make a PR for this if needed)

bigtimebuddy:
I don't think there is much harm in this as it is backward compatible. This all seem rather new but would unblock 4.7+ TS users.

So yeah, PR welcome. Thanks for offering.

zwade:
Oh, are you ok with me applying this change to all packages, or would you prefer me to just do it at the top level `bundles/pixi.js` package? (apologies, I'm not really sure what pixi's dev practices look like)

bigtimebuddy:
All packages, please!

zwade:
Resolved in #8371

PR #8371, opened 2022-05-30
chore: Add "types" and "exports" to package.json (#8368)

zwade:
##### Description of change

This CR adjusts the `package.json` to make PIXI easier to use in an ESM environment with TypeScript.

TypeScript 4.7 introduces official (albeit beta) support for ESM in TypeScript. In connection with modern bundling tools such as Vite, it presents an exciting opportunity for front-end development. Unfortunately, the current version of PIXI packages do not behave properly in an ESM environment due to adjustments in how TypeScript resolves packages and types.

To be more specific, when running TS in ESM mode against the current version of PIXI, it resolves the code to the CJS bundle and the types to a non-existent index file in the CJS directory. To get it to correctly resolve both the types and the entrypoint, the `package.json` needs to include an `"exports"` key that specifies both types and entrypoint for ESM and CJS. 

For more details on this feature, see [they TypeScript 4.7 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing) or [the node.js documentation](https://nodejs.org/api/packages.html#packages_exports). Additionally, for fallback reasons, it's recommend to explicitly include the `"types"` field which specifies exactly where the types are located in the package.
 
This PR adds both of those keys to the 50 applicable PIXI packages, as well as the 2 associated bundle packages. 

##### Pre-Merge Checklist

### DIFF
--- a/bundles/pixi.js-legacy/package.json
+++ b/bundles/pixi.js-legacy/package.json
@@ -17,6 +17,19 @@
   "bundleOutput": {
     "name": "PIXI"
   },
+  "types": "index.d.ts",
+  "exports": {
+    ".": {
+      "import": {
+          "default": "./dist/cjs/pixi-legacy.js",
+          "types": "./index.d.ts"
+      },
+      "require": {
+          "default": "./dist/esm/pixi-legacy.js",
+          "types": "./index.d.ts"
+      }
+    }
+  },
   "homepage": "http://www.pixijs.com/",
   "bugs": "https://github.com/pixijs/pixi.js/issues",
   "license": "MIT",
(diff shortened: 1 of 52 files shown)


zwade:
Not sure if there is anything applicable on the test & documentation side of things, but let me know if I should add something!

andrewstart (approved):

jramstedt:
After this change, I'm getting following error with webpack:

Module not found: Error: Default condition should be last one

bigtimebuddy:
@zwade Can you look into this?

zwade:
Yeah, taking a look now

PR #8376, opened 2022-06-01
Fix scissor mask bugs

dev7355608:
##### Description of change

This fixes two bugs:
- `maskObject.getBounds` must skip transform updates. Same reason as #8296.
  - Before: https://www.pixiplayground.com/#/edit/b6M2O5ylXdDYqGdYfRTB5
  - After: https://www.pixiplayground.com/#/edit/KwXtA13MO-k1vDmjuvQ6z
- If the same mask is used multiple times at different levels, `_scissorRect` gets corrupted and the scissor mask is not restored properly on pop.
  - Before: https://www.pixiplayground.com/#/edit/DUNO0XixBNeJKvSVCnR8K
  - After: https://www.pixiplayground.com/#/edit/cS-ujtnS4kKpisF3mNra8
 
I also removed the two `renderable` lines. They don't seem to have any purpose. `getBounds` doesn't require the object to be renderable.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/mask/MaskData.ts
+++ b/packages/core/src/mask/MaskData.ts
@@ -12,7 +12,7 @@ export interface IMaskTarget extends IFilterTarget
     isSprite?: boolean;
     worldTransform: Matrix;
     isFastRect?(): boolean;
-    getBounds(skipUpdate?: boolean): Rectangle;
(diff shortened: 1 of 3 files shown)


GoodBoyDigital (approved):
nice! lgtm @dev7355608 !

bigtimebuddy:
Thanks @dev7355608

Issue #8379, opened 2022-06-02
Module not found after upgrading to Pixi v6.4.0


Julien-Marcou:
### Current Behavior

After upgrading from PixiJS v6.3.2 to v6.4.0, when building/serving an Angular project, I got the following error:
`Error: Module not found: Error: Default condition should be last one`

The problem seems to occur with WebPack and/or TypeScript, not only with Angular

### Expected Behavior

There should be no error

### Possible Solution

Downgrading PixiJS to v6.3.2 fixes the problem

### Environment

- **`PixiJS` version**: 6.4.0
- **`Angular` version**: 13.3.11
- **`TypeScript` version**: 4.6.4

### TSConfig

`module` is set to `es2020` and setting it to `esnext` doesn't solve the problem

jramstedt:
Problem is in package.json files. (#8371)
"default" should be the last one in "import" and "require".

bigtimebuddy:
Fixed by #8381

bigtimebuddy:
Making the release now

zwade:
@Julien-Marcou when you get some time, can you confirm that 6.4.2 fixes this for you? Thanks!

Julien-Marcou:
Yes, everything works fine now 👍

PR #8381, opened 2022-06-02
fix: Resolves issue where "exports" broke webpack (#8379)

zwade:
##### Description of change
Fixes the bug reported in #8379 and introduced in #8371 

General idea, as within the second-level of the "exports" key (where "import" and "require" are specified), the third level (where "types" and "default" are specified) is expected to be sorted in order of specificity, with "default" coming last. Although not all tools enforce this, webpack does and will refuse to compile if this is improperly managed. 

The fix swaps the order of these to prevent such issues. 

Embarrassingly enough, while resolving this I noticed that in the original version of this CR, I had additionally swapped the order the "esm" and "cjs" references. As part of this fix, I also resolved that issue. 

##### Pre-Merge Checklist

Additionally, I linked this against a local webpack project and was able to get past the initial resolution stage.

### DIFF
--- a/bundles/pixi.js-legacy/package.json
+++ b/bundles/pixi.js-legacy/package.json
@@ -21,12 +21,12 @@
   "exports": {
     ".": {
       "import": {
-        "default": "./dist/cjs/pixi-legacy.js",
-        "types": "./index.d.ts"
+        "types": "./index.d.ts",
+        "default": "./dist/esm/pixi-legacy.js"
       },
       "require": {
-        "default": "./dist/esm/pixi-legacy.js",
-        "types": "./index.d.ts"
+        "types": "./index.d.ts",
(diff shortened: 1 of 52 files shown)


zwade:
Huge apologies for pushing the original PR out in a semi-functional state. WRT this PR, is there a good way to link the full monorepo against a local project for testing? Linking the main project is easy, but linking all the subprojects as well seems a bit more challenging.

andrewstart (approved):
I have to say, Node/other tools requiring order to matter in a non-array data object is disappointing.

bigtimebuddy:
Thanks for looking into this @zwade appreciate your responsiveness.

zwade:
Yeah absolutely, and apologies again for the bug itself

PR #8388, opened 2022-06-04
Fix extract pixels

dev7355608:
##### Description of change

There a multiple problems in the new extract pixels parameter:
1. `options.x`/`options.y` are never used.
2. If the target is a render texture, `options` isn't used at all.
3. Passing `resolution` with the options doesn't really make sense: the resolution is a property of the renderer / render texture.

So I made the following changes:
1. I fixed the bugs above.
2. I deprecated `PixelsExtractOptions`. The parameter is a `Rectangle` now instead.
3. I added the parameter to `extract.canvas` as well.
4. I made some changes to the rounding of the frame dimensions. They are now `round`-ed instead of `floor`-ed / cast to integer. `extract.canvas` and `extract.pixels` had different rounding before this change.

Before: https://www.pixiplayground.com/#/edit/M5ZxsbemrAs5clp0jfFSc
After: https://www.pixiplayground.com/#/edit/vBlbFzhawm26XxU1Zx9X0

##### Pre-Merge Checklist

### DIFF
--- a/packages/extract/src/Extract.ts
+++ b/packages/extract/src/Extract.ts
@@ -11,6 +11,7 @@ const BYTES_PER_PIXEL = 4;
 /**
  * this interface is used to extract only  a single pixel of Render Texture or Display Object
  * if you use this Interface all fields is required
+ * @deprecated
  * @example
  * test: PixelExtractOptions = { x: 15, y: 20, resolution: 4, width: 10, height: 10 }
  */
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Can you also add parity to CanvasExtract?

bigtimebuddy:
Nice! Thanks. This is definitely an improvement.

GoodBoyDigital (approved):
nice! thanks @dev7355608

Issue #8105, opened 2022-01-17
Mask might not be applied correctly when using Extract


ymegane:
### Expected Behavior

Can export the Container being drawn.

### Current Behavior

It seems that the mask applied to the moved Container is not applied to the correct position.

Drawing:
![image](https://user-images.githubusercontent.com/2368078/149755274-611c15a9-9aee-468c-9786-53fd926769e8.png)

Exported result:
![screenshot (17)](https://user-images.githubusercontent.com/2368078/149755398-fe969aa5-250d-44e2-9b5b-57f7eb9dfc59.png)

### Possible Solution

What I investigated to reproduce:
This problem does not occur in v6.1.3, but in v6.2.0.
Also, after investigating from the change diff, it seems that the change in #7896 has an effect.
(When I revert adc5217a2e129ce7a4f5534b865141ffb6a5d385, it works as expected.)

### Steps to Reproduce

1. Set x and y of Container to non-zero.
2. Apply the mask to the child Sprite of that Container
3. Export the Container as a canvas using extract.

Reproduce example (using v6.2.1)
https://codesandbox.io/s/extract-with-mask-problem-q9tqn

Expected behavior example (using v6.1.3):
https://codesandbox.io/s/extract-with-mask-problem-expected-behavior-2xmyv

### Environment

- **`pixi.js` version**: 6.2.1
- **Browser & Version**:  Chrome 97.0.4692.71
- **OS & Version**: _e.g. macOS 12.1
- **Running Example**: https://codesandbox.io/s/extract-with-mask-problem-q9tqn

ymegane:
This issue was resolved in #8376 and I have confirmed that the problem does not occur in v6.5.0.
Thanks!

PR #8392, opened 2022-06-07
Fixing bounds calculation with Scissor Mask

michaeljherrmann:
##### Description of change

We started having issues with the positioning of masks in v6.2.0, I narrowed it down to commit https://github.com/pixijs/pixijs/commit/adc5217a2e129ce7a4f5534b865141ffb6a5d385. The issue seems to be fixed by skipping update when calculating bounds of the mask object. According to the [getBounds documentation](https://pixijs.download/dev/docs/PIXI.DisplayObject.html#getBounds) `skipUpdate` should be `true` within a render() call, which I believe is the case for where this mask code is running.

I don't understand PIXI internals enough to be sure this is the right thing to do, but I think it might be so I'm hoping for some feedback on this. 

I think this will address https://github.com/pixijs/pixijs/issues/8105 as well.

##### Pre-Merge Checklist


bigtimebuddy:
@ivanpopelyshev could you weigh-in on this?

dev7355608:
This is going to be fixed by #8376 already.

bigtimebuddy:
Closing in favor of #8376

PR #8397, opened 2022-06-10
First-Class Support for Rendering with Node.js

Zyie:
Ok I believe that this is now ready!

Changes: 
- Created the `@pixi/node` bundle. This holds everything needed to run in the node environment. I've left this packages inside `bundles/pixi.js-node` folder for now as its consistent with this other bundles
- Set minimum version of node to be v16. Update github actions to match this
- `WorkerManager`, `BasisParser`, `makeAbsoluteUrl` and `LoaderParser` all had logic that would call the DOM before changing the adapter took place. I have now moved these bits of logic to be inline
- I've added a temporary hack in the `rollup.config.js` file to allow the node package to be compiled to ES6 as the `canvas` package doesn't like ES5. This hack can be removed in the `next` branch

Working: 
- Sprites
- Pixi Text
- Bitmap Text
- Spritesheets
- Graphics
- Meshes
- Filters
- SVG

TODO in follow ups:
- [ ] get some tests going. I think the visual testing would be a good fit for this package 
- [ ] webp / avif images would not be supported with the current loader. However there appears to be a workaround for this [here](https://github.com/pixijs/pixijs/discussions/8326#discussioncomment-2916176)


packages/adapter-node/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@pixi/adapter-node",
+  "version": "6.4.2",
+  "main": "dist/cjs/adapter-node.js",
+  "module": "dist/esm/adapter-node.js",
+  "bundle": "dist/browser/adapter-node.js",
```
bigtimebuddy, line 6: Again, not needed here as bundle is browser-only.
```suggestion
```

packages/adapter-node/src/LoadNodeFont.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const validFonts = ['woff', 'woff2', 'ttf', 'otf'];
+
+/**
+ * loads our textures!
+ * this makes use of imageBitmaps where available.
+ * We load the ImageBitmap on a different thread using CentralDispatch
+ * We can then use the ImageBitmap as a source for a Pixi Texture
+ */
+export const loadNodeFont = {
```
bigtimebuddy, line 12: Could you change the casing of the filename to match the export? LoadNodeFont vs loadNodeFont.

bigtimebuddy: And same for LoadNodeTexture.ts

packages/adapter-node/src/LoadNodeTexture.ts
```diff
@@ -0,0 +1,38 @@
...
+ * loads our textures!
+ * this makes use of imageBitmaps where available.
+ * We load the ImageBitmap on a different thread using CentralDispatch
+ * We can then use the ImageBitmap as a source for a Pixi Texture
+ */
+export const loadNodeTexture = {
+    test(url: string): boolean
+    {
+        const tempURL = url.split('?')[0];
+        const extension = tempURL.split('.').pop();
```
bigtimebuddy, line 18: You should use the Node URL object here to parse the extension. It'll be much more accurate and less prone to issues parsing.

```js
import { URL } from 'url';
import * as path from 'path';

const url = new URL(url);
const extension = path.extension(url.pathname);
```

packages/adapter-node/src/NodeCanvasElement.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    ctx.putImageData(data, 0, 0);
+
+    return ctx;
+}
+export class NodeCanvasElement extends Canvas
+{
+    private __event__: EventEmitter;
+    private __gl__: WebGLRenderingContext;
+    private __contextType__: ContextIds;
```
bigtimebuddy, line 49: We don't really have this naming pattern anywhere in Pixi. Would suggest more simple `_event`, `_gl` and `_contextType`. Unless there's a reason for doing this?

packages/adapter-node/src/NodeCanvasElement.ts
```diff
@@ -0,0 +1,222 @@
...
+    {
+        super(width, height, type);
+        this.__event__ = new EventEmitter();
+        this.style = {};
+    }
+
+    get width()
+    {
+        return super.width;
+    }
```
bigtimebuddy, line 62: Does TS require the getter here if it's just using parent's getter?

packages/adapter-node/src/temp/misc.ts
```diff
@@ -0,0 +1,69 @@
+/**
+ * Return font face name from a file name
+ * Ex.: 'fonts/tital-one.woff' turns into 'Titan One'
+ * @param url - File url
+ */
+export function getFontFamilyName(url: string): string
```
bigtimebuddy, line 6: Alternatively, what about using fontkit to extract the name from the file?

packages/adapter-node/src/temp/misc.ts
```diff
@@ -0,0 +1,69 @@
...
+        .split(' ')
+        .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+        .join(' ');
+
+    return nameTitleCase;
+}
+
+// Split a filename into [root, dir, basename, ext], unix version
+// 'root' is just a slash, or nothing.
+const splitPathRe
```
bigtimebuddy, line 25: You have node APIs here, there's no need to maintain this complicated regular expressions if you can use the path API for free.

packages/settings/src/settings.ts
```diff
@@ -63,6 +65,7 @@ export interface ISettings
  */
 export const settings: ISettings = {
 
+    ADAPTER: BrowserAdapter,
```
bigtimebuddy, line 68: Would be good to add some docs here for this property, IAdapter and the Node and Browser adapters.

bigtimebuddy (commented):
Awesome start. I made an initial pass as reviewing.

bigtimebuddy:
Also, what do you think about `@pixi/node` for a package name? I know it kinda breaks the bundle naming pattern here, but I'm not a fan of `pixi.js-*`. This decisions was mostly made because of legacy in v5, but we recently discussed maybe removing legacy in v7 to move canvas back to the main bundle.

packages/adapter-node/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import { NodeCanvasResource } from './NodeCanvasResource';
+import { loadNodeTexture } from './LoadNodeTexture';
+import { loadNodeFont } from './LoadNodeFont';
+import { loadJSON } from './temp/LoadJSON';
+import { loadSpritesheet } from './temp/LoadSpritesheet';
+
+INSTALLED.length = 0;
+INSTALLED.push(NodeCanvasResource);
+
+export const NodeAdapter = {
```
GoodBoyDigital, line 15: I think adaptor is cool :D

I think we can simplify it to only a few functions! Might be best to discuss as I may be missing some details! This would result in less places where the Adaptor code needs to be used, and leverage the fact we have a pretty much a fully working canvas in there! 

createCanvas <-- keep
getContext <-- suggest we remove - this can be removed if we add that logic to the NodeCanvasElement (`canvas. getContext`)
getWebGLRenderingContext <--  suggest we remove, as we only use this in the `SCISSOR_TEST` checks . Lets modify the `ScissorSystem` class to to not store the glCOnst and instead access it using `gl.SCISSOR_TEST. we can use:

```
  protected contextChange(gl: IRenderingContext): void
    {
        this.glConst = gl.SCISSOR_TEST;
    }
```
getNavigator <-- keep

GoodBoyDigital (commented):
yeah this is cool man! 

left a few comments! Two main points:
- I think we can simplify the Adaptor interface to only include functions that only exist on the dom. Basically moving the others to our node canvas implementation. 
- Im _sure_ there is a nicer way to upload textures.. but just not sure what that is without playing around with it a bit! That can be explored though!

Anyhoo, nice work on this one! Its going to be a game changer!

bigtimebuddy:
One other small note, it would be very good to be explicit about what version of Node.js we're allowing (in the engines field in package.json for the bundle and adapter).

packages/adapter-node/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@pixi/adapter-node",
+  "version": "6.4.2",
+  "main": "dist/cjs/adapter-node.js",
```
bigtimebuddy, line 4: Why did you remove the exports and module field? This still maybe bundled.

SmartArray:
Great work so far!

Please add to the list above: `Update CI`

One question: When trying to build the project, I get `[!] Error: 'ExtensionType' is not exported by ../../node_modules/@pixi/core/dist/esm/core.js, imported by packages/ticker/src/TickerPlugin.ts`

Is that because this branch is lagging behind the `dev` branch?
Thank you so much for your work. I would love to help because I really need that feature. 😍

Zyie:
Hey @SmartArray I've just finished updating everything so let me know if you run into any more issues!

SmartArray:
> Hey @SmartArray I've just finished updating everything so let me know if you run into any more issues!

Thank you so much for the documentation updates. I will look into them.

Can you give me some hints on how to test this branch into a project of mine?
I built the project, then added 
- pixijs/packages/basis
- pixijs/packages/assets
and pixijs/bundles/pixi.js-legacy

to my project. Anything that I am missing?

Zyie:
> Can you give me some hints on how to test this branch into a project of mine? I built the project, then added

You should only need to link the `@pixi/node` package that can be found inside `bundles/pixi.js-node`
Once linked just import everything from `@pixi/node` e.g. `import { Container } from '@pixi/node'`

The example in the `README` should be able to help you but here is the main part of it:
```
import express from 'express';
import * as PIXI from '@pixi/node';

const app = express();
const port = 3000;

app.use(express.static(__dirname))
app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})

// This package requires the new asset loader to be used.
// Initialize the new assets loader
await PIXI.Assets.init({
    basePath: 'http://localhost:3000/'
});

// load a sprite
const bunnyTexture = await PIXI.Assets.load('assets/bunny.png');
const bunny = PIXI.Sprite.from(bunnyTexture);
```

SmartArray:
Yes, linking works very well. Thank you for helping me to set things up.

Will give feedback soon!

SmartArray:
This PR adds support for base64 encoded urls: https://github.com/pixijs/pixijs/pull/8470

SmartArray:
@Zyie Your WIP PR works perfectly so far. Thanks to you I had some serious progress in a personal project of mine today.
Continuing to build on it for now.

<details>
  <summary>Maybe Offtopic Feedback</summary>
I have one point to make though, but maybe it's not related to this PR at all.

The canvas renderer extension doesn't take into account the background of the canvas. Also it renders renders out of bound content, as far as I noticed.

A quick workaround for this is adding a custom sprite to the game, setting its width and height, its tint, etc.

Do you think I should raise an issue for this? Or is this the intended behaviour?
</details>

rollup.config.js
```diff
@@ -138,6 +140,27 @@ async function main()
...
+
+            tempPlugins.splice(5, 1);
+            tempPlugins.splice(5, 0, ts);
+            tempPlugins.pop();
+            tempProdPlugins.splice(5, 1);
+            tempProdPlugins.splice(5, 0, ts);
+            tempProdPlugins.splice(tempProdPlugins.length - 2, 1);
+            tempBundlePlugins.splice(6, 1);
+            tempBundlePlugins.splice(6, 0, ts);
+            tempBundlePlugins.splice(tempBundlePlugins.length - 2, 1);
```
bigtimebuddy, line 161: 🤔 

I think there are two important pieces of adding a good hack:

* It can be easily removed
* It's clear why it's being added

This change creates a bunch of temp variables making it annoying to extract, and it is opaque as to what's happening here because of all the popping and splicing.

A better and more transparent approach would be to copy-paste the plugin configuration from above and override the variables. And there should be a comment here that this is only for the Node package and can be removed in v7. 

Also, for this hack we do not need bundle plugins, as we don't generate the bundle for `@pixi/node`.

bundles/pixi.js-node/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+    // Install loader plugins
+    loadNodeTexture,
+    loadNodeFont,
+    loadNodeBitmapFont
+);
+
+// reset installed resources and remove resize plugin from Application
+INSTALLED.length = 0;
+INSTALLED.push(NodeCanvasResource);
+Application['_plugins'].shift(); // eslint-disable-line dot-notation
```
bigtimebuddy, line 50: A better way to remove the ResizePlugin is to export it from `@pixi/app` and use the extensions.remove. Much better and more explicit than shifting.

bundles/pixi.js-node/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+ *     .drawRect(40, 40, 200, 200);
+ *
+ * // Add a blur filter
+ * rect.filters = [new PIXI.filters.BlurFilter()];
+ *
+ * // Display rectangle
+ * app.stage.addChild(rect);
+ * document.body.appendChild(app.view);
+ * @namespace PIXI.filters
+ */
```
bigtimebuddy, line 91: Copy-pasta, we don't need this duplicated.
```suggestion
```

bundles/pixi.js-node/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+Application['_plugins'].shift(); // eslint-disable-line dot-notation
+
+/**
+ * String of the current PIXI version.
+ * @static
+ * @constant
+ * @memberof PIXI
+ * @name VERSION
+ * @type {string}
+ */
```
bigtimebuddy, line 59: Copy-pasta, unnecessary.

```suggestion
```

bundles/pixi.js-node/README.md
```diff
@@ -0,0 +1,98 @@
+<div align="center">
+  <h1>@pixi/node</h1>
+</div>
+<br>
```
bigtimebuddy, line 4: Lets keep it simple

```suggestion
# @pixi/node
```

bundles/pixi.js-node/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+app.use(express.static(__dirname))
+app.listen(port, () => {
+  console.log(`Example app listening on port ${port}`)
+})
+
+// This package requires the new asset loader to be used.
+// Initialize the new assets loader
+await PIXI.Assets.init({
+    basePath: 'http://localhost:3000/'
```
bigtimebuddy, line 55: Do absolute file paths not work?

bundles/pixi.js-node/README.md
```diff
@@ -0,0 +1,98 @@
...
+To build from source you will need to make sure you have the following dependencies installed and then reinstall:
+
+`brew install pkg-config cairo pango libpng jpeg giflib librsvg`
+
+## Basic Usage Example
+This example uses `express`, however, you can use any framework you like.
+
+```js
+import express from 'express';
+import * as PIXI from '@pixi/node';
```
bigtimebuddy, line 42: Would like to get away from `PIXI`
```suggestion
import { Application, Sprite, Assets } from '@pixi/node';
```

bundles/pixi.js-node/package.json
```diff
@@ -0,0 +1,80 @@
...
+  },
+  "dependencies": {
+    "@types/gl": "^4.1.0",
+    "@types/xml2js": "^0.4.11",
+    "canvas": "^2.9.1",
+    "cross-fetch": "^3.1.5",
+    "gl": "^5.0.0",
+    "xml2js": "^0.4.23"
+  },
+  "peerDependencies": {
```
bigtimebuddy, line 43: For bundles the packages should be dependencies, not peerDependencies.
```suggestion
    "xml2js": "^0.4.23",
```

bigtimebuddy (commented):
Super excited for this!

bundles/pixi.js-node/README.md
SmartArray: I think the loader expects the filename to be a URL. When loading, it tries to fetch the resource.
That's why I added #8470 (which I noticed isn't fully complete yet, because we need to add a ResourceLoader to INSTALLED, is that correct? Otherwise I am getting `Unrecognized source type to auto-detect Resource`)

bigtimebuddy:
@Zyie could you resolve the conflicts? Sorry about this, I did a squash on assets.

Zyie:
Hey @bigtimebuddy  I've got back round to this and have addressed the feedback

- added support for local file paths (no more express needed)
- made changes to the `rollup.config.json` hack to make it more clear
- exported `ResizePlugin` from `@pixi/app` so that `extensions` can remove it
- split loading base64 string in the `loadNodeTexture` into a new loader called `loadNodeBase64`

bundles/pixi.js-node/src/adapter/polyfills.ts
```diff
@@ -1,6 +1,52 @@
-import fetch from 'cross-fetch';
+import { fetch, Response, Request } from 'cross-fetch';
 
-globalThis.fetch = fetch;
+import fs from 'fs';
+import path from 'path';
+
+// eslint-disable-next-line func-names
+globalThis.fetch = function (url, options)
```
bigtimebuddy, line 7: In Node.js 18, native `fetch` will be standard. Setting `globalThis.fetch` will always override the internal `fetch`, which is polluting the global space. This should probably be a [ponyfill](https://github.com/sindresorhus/ponyfill) instead. 

Node's fetch will not support local paths, so this is a non-standard implementation of fetch, therefore it should probably be something else.

bigtimebuddy: Maybe we need an adapter entry for fetch?

Zyie: added!

bigtimebuddy (approved):
I'm good on this @Zyie, anything else we're waiting on?

Zyie:
> I'm good on this @Zyie, anything else we're waiting on?

Nope, I think we are good to go on this!

bigtimebuddy:
# :tada::tada::tada:

PR #8400, opened 2022-06-13
Add missing MaskData checks

dev7355608:
##### Description of change

Added `MaskData` checks where they were missing.

##### Pre-Merge Checklist

### DIFF
--- a/packages/events/src/EventBoundary.ts
+++ b/packages/events/src/EventBoundary.ts
@@ -497,9 +497,10 @@ export class EventBoundary
 
         if (displayObject._mask)
         {
-            const mask = displayObject._mask as any;
+            const maskObject = ((displayObject._mask as any).isMaskData
+                ? (displayObject._mask as any).maskObject : displayObject._mask);
 
-            if (!(mask.containsPoint && mask.containsPoint(location)))
+            if (maskObject && !(maskObject.containsPoint && maskObject.containsPoint(location)))
             {
                 return true;
             }
--- a/packages/events/src/EventBoundary.ts
+++ b/packages/events/src/EventBoundary.ts
@@ -500,7 +500,7 @@ export class EventBoundary
             const maskObject = ((displayObject._mask as any).isMaskData
                 ? (displayObject._mask as any).maskObject : displayObject._mask);
 
-            if (maskObject && !(maskObject.containsPoint && maskObject.containsPoint(location)))
+            if (maskObject && !maskObject.containsPoint?.(location))
             {
                 return true;
             }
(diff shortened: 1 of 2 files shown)


packages/events/src/EventBoundary.ts
```diff
@@ -497,9 +497,10 @@ export class EventBoundary
 
         if (displayObject._mask)
         {
-            const mask = displayObject._mask as any;
+            const maskObject = ((displayObject._mask as any).isMaskData
+                ? (displayObject._mask as any).maskObject : displayObject._mask);
 
-            if (!(mask.containsPoint && mask.containsPoint(location)))
+            if (maskObject && !(maskObject.containsPoint && maskObject.containsPoint(location)))
```
bigtimebuddy, line 503: Can we do something like this to be more terse?
```suggestion
            if (maskObject && !(maskObject.containsPoint?.(location)))
```

packages/interaction/src/TreeSearch.ts
```diff
@@ -87,7 +87,10 @@ export class TreeSearch
         {
             if (hitTest)
             {
-                if (!((displayObject._mask as any).containsPoint && (displayObject._mask as any).containsPoint(point)))
+                const maskObject = ((displayObject._mask as any).isMaskData
+                    ? (displayObject._mask as any).maskObject : displayObject._mask) as any;
+
+                if (maskObject && !(maskObject.containsPoint && maskObject.containsPoint(point)))
```
bigtimebuddy, line 93:
```suggestion
                if (maskObject && !(maskObject.containsPoint?.(point)))
```

packages/events/src/EventBoundary.ts
```diff
@@ -497,9 +497,10 @@ export class EventBoundary
 
         if (displayObject._mask)
         {
-            const mask = displayObject._mask as any;
+            const maskObject = ((displayObject._mask as any).isMaskData
+                ? (displayObject._mask as any).maskObject : displayObject._mask);
```
bigtimebuddy, line 501: The double-any cast here is kinda ugly. Can we either use real types rather than making this worse?

bigtimebuddy: Why not check for `instanceof MaskData` or `'isMaskData` in displayObject._mask`? Trying to avoid the casting.

dev7355608: We have similarly ugly mask data checks elsewhere already. Maybe we should add the getter `DisplayObject#maskObject`? Or just a helper method to extract the mask object.

bigtimebuddy: I think a normalized, non public getter (eg maskObject) sounds like a good idea. But if there are more places than this, another PR is probably better for that cleanup bit.

bigtimebuddy (approved):

Issue #8281, opened 2022-04-11
Wrong hit area in shadow DOM


pwalczyszyn:
### Expected Behavior

Hit area should be properly calculated when `pixi` is embedded in custom elements with shadow DOM enabled.

### Current Behavior

Hit area is not taking into consideration pixi app position when hosted in custom elements with shadow DOM enabled. This also impacts `click` and `pointerup` events which are missing.

### Steps to Reproduce

In the link below I included a demo that shows custom element with and without shadow DOM enabled. In the first example you can see (some not all) pointer events are registered above the bunny.

### Environment

- **`pixi.js` version**: 6.3.0
- **Browser & Version**: Chrome  100.0.4896.75 
- **OS & Version**: macOS 12.0.1
- **Running Example**: https://stackblitz.com/edit/pixi-shadow-dom-bug?file=src%2Fmy-element.ts

da411d:
Same issue

Zyie:
Fixed in #8406 

Will be released with `v6.5.0`

Issue #7165, opened 2021-01-18
Interactions fail in shadow DOM of web component


soletan:
### Expected Behavior

I'd love to use PIXI.js in context of a web component (made with stenciljs) for rendering an interactive map combining PIXI and leaflet. Markers should be clickable and draggable, thus need to have all interaction events working properly. For ease of implementation there should be a sole covering handler per event on stage or some other container for capturing and handling clicks etc.

### Current Behavior

1. Clicks on interactive display objects don't work. 
2. Clicking an interactive stage doesn't even trigger pointerdown events.

### Possible Solution

1. Click on display object: following some investigating I conclude that this is due to [*up-event's target referring to the wrapping component (host element) rather than PIXI's canvas](https://github.com/pixijs/pixi.js/blob/cdfbc8be6cfbd521c00afc69c996cad48d386ed1/packages/interaction/src/InteractionManager.ts#L1348) (a.k.a. interactionDOMElement) ... thus it assumes *up event having occurred outside of its view and thus doesn't click, but considers interaction sort of cancelled.
2. Click on stage isn't recognized at all: haven't investigated this one, but I assume this is kind of intentional, though stage is a PIXI.Container which is a display object as well. (isn't it?)

### Steps to Reproduce

1. Open fiddle at https://jsfiddle.net/soletan/r9kLstd8/40/
2. Click on red box and observe event handlers logging in on-screen console (or failing to do so).
3. Click on black stage and observe console logging again (though there is no logging).

### Environment

- **`pixi.js` version**: 6.0.0-rc.2
- **Browser & Version**: Firefox 84
- **OS & Version**: Windows 10
- **Running Example**: https://jsfiddle.net/soletan/r9kLstd8/40/

andrewstart:
Containers, while display objects, don't innately have a hit area the way a Sprite or something does. Unless you set its `hitArea` property, if you aren't clicking on one of its children then you are just clicking through it. If you do set a `hitArea`, then that will be used for all hit testing and children will not be considered (by design).

I don't know enough about shadow DOMs to help with the other problem, although your investigation sounds correct.

soletan:
Regarding the hitarea: I see.

[...]

PR #8406, opened 2022-06-16
Fix Interactivity in ShadowDOM

cjgammon:
##### Description of change
This is a small change to access the event target even if inside a shadowDOM context.  Before this 'click' and 'pointerup' events would not trigger inside a web component.  See relevant issues listed below.  The fix is implemented by accessing the first entry of `composedPath()` if it is available.  Otherwise it defaults to the previous implementation of `originalEvent.target`.  I didn't add tests to this as I'm unsure how to add a web component test case given the current test set up, but if you have any suggestions or recommendations as to how to create a test I would be happy to do so.

https://github.com/pixijs/pixijs/issues/8281
https://github.com/pixijs/pixijs/issues/7165

##### Pre-Merge Checklist

### DIFF
--- a/packages/events/src/EventSystem.ts
+++ b/packages/events/src/EventSystem.ts
@@ -258,7 +258,15 @@ export class EventSystem
         // if we support touch events, then only use those for touch events, not pointer events
         if (this.supportsTouchEvents && (nativeEvent as PointerEvent).pointerType === 'touch') return;
 
-        const outside = nativeEvent.target !== this.domElement ? 'outside' : '';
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Thanks @cjgammon for the fix. One small request, could you add this to the events package? This will be replacing interaction very soon.

bigtimebuddy (approved):
LGTM. Thanks @cjgammon

bigtimebuddy:
Thank you @cjgammon

PR #8408, opened 2022-06-17
Fixing CanvasSpriteRenderer behaving differently to WebGL counterpart

drspoik:
##### Description of change
Fixing CanvasSpriteRenderer behaving differently to WebGL counterpart in regards to BaseTexture trimming. If a sprite's texture had custom frame, orig and trim, when using the canvas rendering, the resulting texture would have a wrong scaling. These fixes make it, that both canvas rendering and WebGL behave the same in this instance.

Bug can be reproduced here
https://www.pixiplayground.com/#/edit/8RST1GAiGaqPDUHDmKh8k

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas-sprite/src/CanvasSpriteRenderer.ts
+++ b/packages/canvas-sprite/src/CanvasSpriteRenderer.ts
@@ -113,6 +113,7 @@ export class CanvasSpriteRenderer
             dy = 0;
 
             const h = destWidth;
+
             destWidth = destHeight;
             destHeight = h;
         }
--- a/packages/canvas-sprite/src/CanvasSpriteRenderer.ts
+++ b/packages/canvas-sprite/src/CanvasSpriteRenderer.ts
@@ -56,8 +56,17 @@ export class CanvasSpriteRenderer
             return;
         }
 
-        const width = texture._frame.width;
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
@drspoik Good one. Thank you so much for the contribution too. We love to see the delta between Canvas rendering and WebGL rendering reduced.

Do you think you would be able to add a unit-test or two for this? Seems like it should be pretty straightforward. There's information in the [contributing guide](https://github.com/pixijs/pixijs/blob/dev/.github/CONTRIBUTING.md#testing-your-change) about running unit-test for different packages and there are already tests for CanvasSpriteRenderer [here](https://github.com/pixijs/pixijs/blob/dev/packages/canvas-sprite/test/CanvasSpriteRenderer.tests.ts).

drspoik:
Hi @bigtimebuddy, I hope I did it right, I never really contributed to a repo before

bigtimebuddy:
Looks like some linting issues. You can run run `npm run lintfix` locally which should auto-fix most of these.

bigtimebuddy (approved):

bigtimebuddy (approved):

PR #8414, opened 2022-06-20
implement visual testing

Zyie:
## Visual Testing
This PR introduces the start of visual testing for Pixi with a couple of basic tests to begin with

### Generating Tests

The core concept is that you generate files with the `*.scene.ts` extension e.g. `TEST.scene.ts` 
These files should look something like this

```ts
import { Container } from '@pixi/display';
import { Sprite } from '@pixi/sprite';
import { Texture } from '@pixi/core';
import type { TestScene } from '../../../test/renderTests';

export const scene: TestScene = {
    it: 'should render a grid of bunnies with a tint',
    create: async (scene: Container) =>
    {
        const tex = await Texture.fromURL('https://pixijs.io/examples/examples/assets/bunny.png');
        const tints = [0xFFFFFF, 0xFF0000, 0x00FF00, 0x0000FF];

        // Create a 5x5 grid of bunnies
        for (let i = 0; i < 25; i++)
        {
            const bunny = Sprite.from(tex);

            bunny.x = (i % 5) * 115;
            bunny.y = Math.floor(i / 5) * 115;
            bunny.tint = tints[i % tints.length];
            scene.addChild(bunny);
        }
    },
};
```
#### Test Object
The test object is made up of these elements

- `it` - some text to describe the test
- `create` - An async function that builds the scene
- `id?` - Name to save the image as. By default images are saved as there filename
- `options?` - Options to pass to the renderer on creation. `width` and `height` are set to 512 by default 
- `pixelMatch?` - The number of pixels that can be different. Defaults to 40
- `skip?` - Boolean for if you want to skip over the test. Similar to `it.skip`

### Running Tests

A couple of new scripts have been added 

- `test:scene` - tests all `*.scene.ts` files
- `test:scene --packages=@pixi/math` - also supports running on individual packages
- `test:scene:debug` - tests all scenes but opening the debug window
- `test:main` - runs normal testing
- `test` - runs both normal and scene testing together

### Output

Each snapshot is generated and placed inside the `test/snapshot` folder.

Diffs of the images are generate for each test and placed inside an `.artifacts` folder. I've added an extra step in the github workflow to also upload these diffs so it is easier to see on failing tests

### Future Work

The next steps for this would be to get it automated on something like [BrowserStack](https://www.browserstack.com/) in order for the testing to be done on multiple browsers and devices. 

BrowserStack offer an open source licence which would give us unlimited testing


.github/workflows/nodejs.yml
```diff
@@ -40,6 +40,11 @@ jobs:
         run: npm run coverage:ci
       env:
         NODE_ENV: production
+    - name: Upload Test Artifacts
+      - uses: actions/upload-artifact@v3
```
bigtimebuddy, line 44:
```suggestion
      uses: actions/upload-artifact@v3
```

bigtimebuddy:
How do you re-generate the snapshots?

Zyie:
> How do you re-generate the snapshots?

Just delete the snapshot in the `test/snapshot` folder

> The problem was always that we never had enough energy put into actually creating the tests

I'm happy to lead the charge on this one. If you have a list of tests you think would be useful i can implement them in this PR or a follow up PR

bigtimebuddy:
Also, I think it's useful to have slightly narrower and smaller visual tests. For instance 64 x 64 (or 128 x 128) size for testing individual features. I'll be easier to understand the regressions if the tests are more granular, and the tests will run faster with less memory not having to compare so many values.

Here's a list off the top of my head:

### Graphics

[ ] Basic Rectangle
[ ] Basic Circle
[ ] Basic Rounded Rectangle
[ ] Basic Polygon
[ ] Basic Line
[ ] Basic Arc
[ ] Basic Bezier Curve
[ ] Basic Quadratic Curve
[ ] Rectangle with stroke
[ ] Rectangle with stroke & alignment -1
[ ] Rectangle with stroke & alignment 1
[ ] Rectangle with Texture fill
[ ] Rectangle with Texture fill & transform matrix on Texture
[ ] Line with Texture fill
[ ] Rectangle with separate stroke and fill texture fills
[ ] Rectangle with Holes
[ ] Graphics with Shared Geometry
[ ] Graphics with Mask & Holes
[ ] w/ cacheAsBitmap positioned 0,0
[ ] w/ cacheAsBitmap negative coordinates
[ ] w/ multisample set to High
[ ] w/ cacheAsBitmapMultisample set to High
[ ] Adaptive curves (Circle rendered at 1px radius and scaled up)

### Sprite

[ ] Linear Scale mode
[ ] Nearest Scale mode
[ ] Rotation
[ ] Arbitrary skew
[ ] Anchor Point
[ ] Custom Pivot Point
[ ] Tint applied
[ ] w/ Alpha Filter
[ ] Non-1 resolution
[ ] Opacity

### Text

[ ] Stroke
[ ] Drop shadow
[ ] Letter-spacing
[ ] Line-height
[ ] Word Wrapping
[ ] alignment w/ anchor
[ ] Emojis
[ ] Tinting
[ ] w/ Alpha Filter

### BitmapFont & BitmapText

[ ] Dynamic size
[ ] Emojis
[ ] Letter spacing
[ ] Line Height

### Container

[ ] Nested w/ scale/transforms
[ ] Nested w/ opacity
[ ] Nested w/ tinting
[ ] Nested w/  masking with Graphics
[ ] Nested w/ masking with fastRect
[ ] Nested w/ masking with sprite mask
[ ] Custom pivot point
[ ] Nested w/ custom pivot points
[ ] Nested w/ AlphaFilter on root

PR #8420, opened 2022-06-21
Remove AbstractBatchRenderer and BatchPluginFactory

bigtimebuddy:
We are trying to reduce the levels of inheritance for PixiJS and the BatchRenderer was unnecessarily complex. 

This PR removes the AbstractBatchRenderer and BatchPluginFactory classes and just has a single BatchRenderer class which is also a plugin.

If someone want add a different BatchRenderer, they can extend or modify the properties to the existing batch instance e.g., `renderer.plugins.batch.setShaderGenerator({ fragment })`


ivanpopelyshev:
I'm ok with that

GoodBoyDigital (approved):
lgtm!

PR #8423, opened 2022-06-22
System Extensions

bigtimebuddy:
Converts systems for Renderer and CanvasRenderer into extensions. Systems are now installed statically which helps unblock a handful of fixes. 

## Future Changes Unblocked

* Make EventSystem something that can be automatically installed in the bundle
* Fully deprecate interaction package
* Convert accessibility, extract, canvas-extract, prepare, canvas-prepare into systems

### DIFF
--- a/packages/core/src/renderTexture/RenderTextureSystem.ts
+++ b/packages/core/src/renderTexture/RenderTextureSystem.ts
@@ -7,6 +7,7 @@ import type { RenderTexture } from './RenderTexture';
 import type { BaseRenderTexture } from './BaseRenderTexture';
 import type { MaskData } from '../mask/MaskData';
 import type { ISize } from '@pixi/math';
+import { ExtensionMetadata, ExtensionType } from '../extensions';
 
 // Temporary rectangle for assigned sourceFrame or destinationFrame
 const tempRect = new Rectangle();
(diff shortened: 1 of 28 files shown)


packages/core/src/renderTexture/RenderTextureSystem.ts
```diff
@@ -34,7 +35,13 @@ const tempRect2 = new Rectangle();
...
 export class RenderTextureSystem implements ISystem
 {
-/* eslint-enable max-len */
+    /** @ignore */
+    static extension: ExtensionMetadata = {
+        type: ExtensionType.RendererSystem,
+        name: 'renderTexture',
+    };
+
+    /* eslint-enable max-len */
```
Zyie, line 44:
```suggestion
```

Zyie (approved):

packages/core/src/renderTexture/RenderTextureSystem.ts
bigtimebuddy: This is used to ignore the long table above

Zyie: :facepalm:

yep i misread that one

PR #8425, opened 2022-06-23
Remove InteractionManager, EventSystem as default

bigtimebuddy:
In the next major version, we are removing the interaction package (InteractionManager) and replace it with the events (EventSystem) package instead. This package is superior in many ways to InteractionManager and supports for more DOM-like behaviors like bubbling.

## TODO

* [ ] Replace interaction calls in AccessibilityManager with class to events


bigtimebuddy:
@ShukantPal I could use your advice how best to upgrade these bits in AccessibilityManager. How would you do this with events?

https://github.com/pixijs/pixijs/blob/9794ece302bb5bbb4bcc84eab4adb0c43708c34c/packages/accessibility/src/AccessibilityManager.ts#L502-L511

miltoncandelero:
remind me again why are we not adding those three events to the new system? 🤔

bigtimebuddy:
@miltoncandelero huh? Those events (click, pointertap, tap) are in the EventSystem.

bigtimebuddy:
https://github.com/pixijs/pixijs/blob/9794ece302bb5bbb4bcc84eab4adb0c43708c34c/packages/events/src/EventBoundary.ts#L916-L927

ShukantPal:
@bigtimebuddy Not sure what's going on exactly in the AccessibilityManager, but you should be able to create a synthetic event and dispatch it from this:

https://github.com/pixijs/pixijs/blob/f215d87878f42939812ca0d76fa0784163b94697/packages/events/src/EventBoundary.ts#L236

You'd have to set the event target before it propagates, however.

ShukantPal (approved):

bigtimebuddy:
Runnable here: https://jsfiddle.net/bigtimebuddy/knamsz12/

Zyie (approved):

tiye:
while I can't say I'm against this change... is any plugin or dirty library that could help me implement the fancy dragging feature in the old version? Chrome does not have such complicated events systems but there is `pointermove` that still offer help on touch screens, at least I could make use of.

bigtimebuddy:
@tiye please consider opening up a new issue about what you're looking for, not sure I fully understand from your description. It would be helpful to know or see an example of v6/InteractionManager code that you're looking to replicate in v7.

tiye:
@bigtimebuddy https://github.com/pixijs/pixijs/discussions/8813

PR #8427, opened 2022-06-23
Convert Prepare, Extract to Systems

bigtimebuddy:
Converts the following Renderer and CanvasRenderer plugins to systems:

* Prepare
* Extract
* CanvasPrepare
* CanvasExtract

There are a few benefits for doing this:

* Get closer to disambiguating plugins as systems or pipelines
* These extensions are now strictly typed, plugin is just an any map

This change is backward-compatible with a deprecation warning.

### DIFF
--- a/packages/extract/test/Extract.tests.ts
+++ b/packages/extract/test/Extract.tests.ts
@@ -26,7 +26,7 @@ describe('Extract', () =>
     {
         const renderer = new Renderer();
         const sprite = new Sprite(Texture.WHITE);
-        const extract = renderer.extract as Extract;
+        const extract = renderer.extract;
 
         expect(extract.canvas(sprite)).toBeInstanceOf(HTMLCanvasElement);
         expect(extract.base64(sprite)).toBeString();


packages/extract/test/Extract.tests.ts
```diff
@@ -25,7 +26,7 @@ describe('Extract', () =>
     {
         const renderer = new Renderer();
         const sprite = new Sprite(Texture.WHITE);
-        const extract = renderer.plugins.extract as Extract;
+        const extract = renderer.extract as Extract;
```
Zyie, line 29: Testing locally I don't think these need to be cast anymore

```suggestion
        const extract = renderer.extract;
```

packages/extract/test/Extract.tests.ts
```diff
@@ -39,7 +40,7 @@ describe('Extract', () =>
     it('should extract with no arguments', () =>
     {
         const renderer = new Renderer();
-        const extract = renderer.plugins.extract as Extract;
+        const extract = renderer.extract as Extract;
```
Zyie, line 43:
```suggestion
        const extract = renderer.extract;
```

packages/extract/test/Extract.tests.ts
```diff
@@ -52,7 +53,7 @@ describe('Extract', () =>
     it('should extract a render texture', () =>
     {
         const renderer = new Renderer();
-        const extract = renderer.plugins.extract as Extract;
+        const extract = renderer.extract as Extract;
```
Zyie, line 56:
```suggestion
        const extract = renderer.extract;
```

bigtimebuddy:
Follow-up to this will be making sure prepare and extract show up on the Renderer/CanvasRenderer documentation. Currently, docs are blocked because of an issue with webdocs that @ShukantPal is sorting.

Issue #8429, opened 2022-06-23
Inconsistency with pointer events in new EventSystem


miltoncandelero:
`pointertap` doesn't fire on mouse inputs while `pointerdown` and `pointerup` do fire on mouse input.

### Expected Behavior
consistency, either `pointer` events work as they did before (by generalizing mouse and touch) or become a third  case where it's not mouse nor touch. But consistency.

### Current Behavior
`pointerdown` and `pointerup` fire on mouse but `pointertap` doesn't

### Possible Solution
https://github.com/pixijs/pixijs/blob/9794ece302bb5bbb4bcc84eab4adb0c43708c34c/packages/events/src/EventBoundary.ts#L916-L927

`pointertap` is on an extra `else` clause.

bigtimebuddy:
Yeah so probably this?

```ts
this.dispatchEvent(clickEvent, 'pointertap');

if (clickEvent.pointerType === 'mouse')
{
    this.dispatchEvent(clickEvent, 'click');
}
else if (clickEvent.pointerType === 'touch')
{
    this.dispatchEvent(clickEvent, 'tap');
}
```

miltoncandelero:
maybe the otherway arround for consistency with the old API? just because?
this is a really low hanging fruit, lemme pr in just a bit...

![image](https://user-images.githubusercontent.com/15253010/175426887-d830ea14-08cc-41d6-bf46-2a503b770346.png)

Zyie:
Fixed in #8431

PR #8431, opened 2022-06-23
Firing a pointertap after click or tap

miltoncandelero:
##### Description of change
Forcing a `pointertap` to be fired after click or tap events (or any other pointing device input)

solves #8429 

##### Pre-Merge Checklist


bigtimebuddy (approved):
LGTM

bigtimebuddy:
@ShukantPal any concerns?

ShukantPal:
I think it was like this because `pointertap` isn't defined the [Pointer Events API](https://www.w3.org/TR/pointerevents3/).

ShukantPal:
I would suggest deprecating `pointertap` and making `click` work on touch devices to match the DOM behavior 🤷.

bigtimebuddy:
For now, I'd like to try to match to interaction manager to make the transition easier for folks. Normalizing around pointers makes sense now as it is less breaking and internally consistent.

However, I'd like to introduce a breaking change later on to make events as close to the DOM as possible, which was always the intention anyways. Any convenience (pointertap) or normalization (default pointers) should be built on-top of events. I think we can explore this more in v7.

PR #8433, opened 2022-06-24
Add generic typing for container children

### LINKED ISSUE #8430: Feat: Add generic children type for Container
It would be great to be able to set the Type of all children for the Container class.

At the moment we can already do that for its methods (addChild, removeChild, ...) but it would be better to type the entire class when we know the Container will only contain one type of child.

# Before

```ts
class Container extends DisplayObject {
  readonly children: DisplayObject[];
  addChild<T extends DisplayObject[]>(...children: T): T[0];
}
```

# After

```ts
class Container<T extends DisplayObject = DisplayObject> extends DisplayObject {
  readonly children: T[];
  addChild<U extends T[]>(...children: U): U[0];
}
```

Julien-Marcou:
## Description of change

The `Container` class now uses a generic type to describe its children (closes #8430).

This is a backward compatible change.

A `Container` act as it did before, while a `Container<Type>` will narrow its children to `Type[]` (`Type` must extend `DisplayObject`).

It allows to limit/narrow what type of child can be added to/retrieved from a container, allowing to know the type of a given child without casting it.

## Current Behavior

```ts
const child = container.addChild(new Sprite()); // Sprite
const sameChild = container.getChildAt(0); // DisplayObject
```

## New Behavior

Backward compatible:

```ts
const child = container.addChild(new Sprite()); // Sprite
const sameChild = container.getChildAt(0); // DisplayObject
```

Better narrowing:

```ts
const container = new Container<Sprite>();
const child = container.addChild(new Sprite()); // Sprite
const sameChild = container.getChildAt(0); // Sprite
```

Ability to limit what can be added to the container:

```ts
container.addChild(new Graphics()); // Ok

const spriteContainer = new Container<Sprite>();
spriteContainer.addChild(new Graphics()); // Argument of type 'Graphics' is not assignable to parameter of type 'Sprite'.
```

## Limitation

Unfortunately, it's not possible to apply the generic type for the `getChildByName()` as it can retrieve child recursively.

So it will return a `DisplayObject` by default:

```ts
const container = new Container<Sprite>();
// ...
container.getChildByName('test'); // DisplayObject
```

But you still can give it a type if you don't want to cast the returned value:

```ts
const container = new Container<Sprite>();
// ...
container.getChildByName<Sprite>('test'); // Sprite
```

## Pre-Merge Checklist

For some reason, I was no able to perform the tests, it ran indefinitely without any progress. I tried `npm start` & `npm run build` before running `npm run test` and `npm run test:pkg -- --packages=@pixi/display @pixi/mixin-get-child-by-name` but nothing happened


bigtimebuddy:
Love this change.

One small suggestion to add to this PR would be to do this:

https://github.com/pixijs/pixijs/blob/f215d87878f42939812ca0d76fa0784163b94697/packages/particle-container/src/ParticleContainer.ts#L44

Change to this:

```ts
export class ParticleContainer extends Container<Sprite>
```

This would help enforce ONLY Sprites for ParticleContainer.

bigtimebuddy:
Yes, ParticleContainer has some constraints which means only Sprites and Sprites that share a single texture or atlas.

bigtimebuddy (approved):
Great PR! Equal parts useful and removes a lot of messing internal casting.

Julien-Marcou:
The `ParticleRenderer` methods require children to be DisplayObjects:

https://github.com/pixijs/pixijs/blob/2f583cb62a67c43eaa0182c7c3cda8c0a27ddf6b/packages/particle-container/src/ParticleRenderer.ts#L247-L250

Which force us to cast them to Sprite in some cases:

https://github.com/pixijs/pixijs/blob/2f583cb62a67c43eaa0182c7c3cda8c0a27ddf6b/packages/particle-container/src/ParticleRenderer.ts#L259

Do you think we should change the `children`'s type to `Sprite[]`, or is it too big of a change for this PR ?

bigtimebuddy:
I think that's okay actually. No one uses the ParticleRenderer directly, it's more of a contract between the Renderer and ParticleContainer. Breaking that is not likely to impact anyone.

Also, I just noticed something, could you make sure `@pixi/sprite` is listed in the particle-container's package.json?

bigtimebuddy:
Thank you @Julien-Marcou!

PR #8434, opened 2022-06-24
add consistent import type

Zyie:
The lint rule just forces things to be imported as a type if it is only used as one
Its auto fixable so doesn't add any extra work for the developer

I'm also happy to fix all the messy conflicts this PR brings if you want to add it


bigtimebuddy (approved):
Nice one!

PR #8435, opened 2022-06-24
Asset tweaks

Zyie:
I've gone through and made a couple of changes for the assets PR

- Fixed the examples documentation that had inconsistencies with using ` and '
- Added the `PIXI` prefix to the examples
- Added `removeURLParser ` to complement `addURLParser`
- Exported everything except for `Loader` and `Resolver`
- Added a bunch of `index.ts` files
- Moved a couple of files around into new folders e.g. `utils/url/isAbsoluteURL.ts`


Zyie:
Yeah I'm not opposed to the idea!

It would be nice to keep it consistent throughout the code. So maybe there is an eslint plugin that can help, just like with `import type`

GoodBoyDigital (approved):
nice

PR #8442, opened 2022-06-25
Denote that the geometry parameter of the Graphics constructor is optional

aaclayton:
The `geometry` param of the `Graphics` constructor is currently documented as required which creates IDE warnings like the following:

image

This change denotes `geometry` as optional.

## Testing Done
```
npm run docs
```
image

### DIFF
--- a/packages/graphics/src/Graphics.ts
+++ b/packages/graphics/src/Graphics.ts
@@ -155,7 +155,7 @@ export class Graphics extends Container
     }
 
     /**
-     * @param geometry - Geometry to use, if omitted will create a new GraphicsGeometry instance.
+     * @param [geometry] - Geometry to use, if omitted will create a new GraphicsGeometry instance.
      */
     constructor(geometry: GraphicsGeometry = null)
     {


bigtimebuddy:
We probably have a lot of place in our code where optional parameters aren't specified like this. Generally, using TypeScript is the most accurate for getting this information and doing type checking. I assume you are not using TypeScript?

aaclayton:
> We probably have a lot of place in our code where optional parameters aren't specified like this. Generally, using TypeScript is the most accurate for getting this information and doing type checking. I assume you are not using TypeScript?

Hey Matt, allow me to provide a little extra perspective and perhaps you have advice. We use PIXI in our project and I use WebStorm as my IDE of choice. Our project is not TypeScript. I have PIXI configured to provide type hinting in the IDE using the feature that WebStorm (and other JetBrains IDEs) provide: https://www.jetbrains.com/help/webstorm/configuring-javascript-libraries.html#ws_js_custom_third_party_library

To my knowledge, the only way to configure this so that it properly produces type hinting for PIXI is by configuring the external library to point at the full browser source of the library (as it is used in our environment):

image

The type hinting I'm using in my dev environment is based on the code documentation in the `node_modules/pixi.js/dist/browser/pixi.js` file rather than TypeScript types. Perhaps there is a way to teach the IDE to use the TypeScript types instead of the browser source - but pointing the IDE towards `node_modules/pixi/index.d.ts` instead fails to produce correct type hinting.

Not sure if you have any guidance here, and I certainly am not intending to open a can of worms where the overall docstrings of PIXI need to be redone, but the `new PIXI.Graphics()` constructor is something that shows up a lot in our codebase so it was low hanging fruit to get this particular documentation issue changed.

I do think this goes beyond solely type hinting in an IDE though, because your built web documentation at <https://pixijs.download/release/docs/PIXI.Graphics.html> is based on the docstrings and does not denote that `geometry` is an optional parameter of the `Graphics` constructor.

Any thoughts?

bigtimebuddy:
That is helpful context.

The optional-ness of a parameter is inferred by webdoc, our documentation tool. So we do not declare this explicitly.

Also, it seems like you are using the browser bundle. Our types are designed to work with bundling tools (webpack, parcel, etc) where you import from the package via the commonjs pr esm module (probably why pixi's types aren't getting used by your IDE). Providing browser types is not something we've implemented and our doc types aren't JSDoc format.

This workflow you're doing is less common which is why we haven't made more effort to support it.

As for solutions, I don't think this fix is going to satisfy your use case. I know you'll have more issues because of how intentionally terse webdoc is. The best solution is to find a way to trick your IDE into using the types but unfortunately I'm not sure how'd you'd do that.

bigtimebuddy:
@ShukantPal do you know why this params is non optional? I'd assume it should be.

aaclayton:
Difference of opinion, perhaps, but I don't think our use case is so non-standard as to be out of bounds for what should be supported by the official PIXI documentation.

> The optional-ness of a parameter is inferred by webdoc, our documentation tool. So we do not declare this explicitly.

Perhaps the correct change is that the constructor needs to be annotated then?
```diff
- constructor(geometry: GraphicsGeometry = null)
+ constructor(geometry?: GraphicsGeometry = null)
```
You can see from the logic inside the constructor that `geometry` is optional.

bigtimebuddy:
Yes, that's correct. I think this is a bug with our doc tool.

ShukantPal:
@bigtimebuddy Adding this to my internal tracker - most likely because webdoc was inferring optionality based in the `?` and not if there is a default value.

bigtimebuddy:
I'm going to close this issue @aaclayton. There are two issues here:

1). We will fix the documentation with an update to webdoc to make sure default value parameters are displayed as optional. Thank you for letting us know that was not consistent.
2). As for your WebStorm workflow, unfortunately, we are not supporting the browser bundle with types (using the global namespace object `PIXI`) and we will not be reformatting our docs to add type information. I recommend that you either find a solution yourself (you may consider looking at #7413 for hints about how to do this) or consider using a code bundler (Webpack, Parcel, Rollup, etc) for a more paved-path approach.

PR #8444, opened 2022-06-26
Feature/assets compressed textures

GoodBoyDigital:
##### Description of change
Assets now supports compressed textures (DDS, KTX, Basis).
To achieve this i separated out the loader + parsing logic. So we now have three new class:

`parseDDS` - returns a texture resource
`parseKTX` -  returns a texture resource
`BasisParser` - returns a texture resource, but also comes with the set up functions required.

The current loader class and the new loader class both make use of them.

I also made a couple of changes:
`TranscoderWorker.onTranscoderInitialized` is now a promise rather than a runner (its just a nice flow here!)

The Asset version of the texture parsers will only parse the first texture as multiple textures mean that we should really be creating texture arrays or cube textures. (let me know if i'm way off the mark here @ShukantPal !)

I also created a test for loading the basis file. But it was crashing in electron (although i can confirm it works in chrome).

I have also introduced the concept of cacheParsers! 
This has resulted in a very clean way to manage assets overall. 

Loader - loads ONLY
Cache - caches ONLY
Assets links the two!

Now when caching something like `Cache.set("mySpriteSheet", spriteSheet);`
Will result in the cache understanding that is should cache all the sprite sheet textures as well. 

This is also neatly tied to remove too so `Cache.remove("mySpriteSheet");` will remove every asset it added from the cache. 

##### Pre-Merge Checklist


packages/assets/src/Assets.ts
```diff
@@ -611,9 +612,16 @@ export class AssetsClass
             loadWebFont,
         );
 
+        this.cache.addParser(
+            cacheSpritesheet,
+            cacheTextureArray
+        );
```
Zyie, line 618: All these parsers are only added on reset. Shouldn't they be added in the constructor as well?

packages/assets/src/cache/Cache.ts
```diff
@@ -1,20 +1,62 @@
...
+ *
+ * With this cache you can add parsers that will take the object and convert it to a list of assets that can be cached.
+ * for example a cacheSprite Sheet parser will add al of the textures found with in its sprite sheet directly to the cache.
+ *
+ * This give devs the flexibility to cache any type of object how every we want.
  *
  * It is not intended that this class is created by developers - it is part of the Asset package.
  * This is the first major system of PixiJS' main Assets class.
+ *
+ * This cache
```
Zyie, line 18: Half finished sentence?

packages/assets/src/cache/Cache.ts
```diff
@@ -1,20 +1,62 @@
...
+            const index = this.parsers.indexOf(parser);
+
+            if (index >= 0) this.parsers.splice(index, 1);
+        }
+    }
 
     /** Clear all entries. */
     public reset(): void
     {
+        this.parsers.length = 0;
```
Zyie, line 59: You also need to clear the `_cacheMap`?

packages/assets/src/cache/Cache.ts
```diff
@@ -44,43 +86,111 @@ class CacheClass
...
+                {
+                    BaseTexture.addToCache(texture.baseTexture, key);
+                }
 
-            Texture.addToCache(texture, key);
+                Texture.addToCache(texture, key);
+            });
         }
 
         //
```
Zyie, line 161:
```suggestion
```

packages/assets/test/loader-compressed.tests.ts
```diff
@@ -0,0 +1,54 @@
...
+        // eslint-disable-next-line max-len
+        const texture: Texture = await loader.load(`https://pixijs.io/compressed-textures-example/images/airplane-boeing_JPG_BC3_1.DDS`);
+
+        expect(texture.baseTexture.valid).toBe(true);
+        expect(texture.width).toBe(1000);
+        expect(texture.height).toBe(664);
+    });
+
+    // CRASHES JEST - ELECTRON
+    // it.only('should load a a Basis image', async () =>
```
Zyie, line 36: What is this about?

Zyie (changes requested):

packages/assets/test/loader-compressed.tests.ts
GoodBoyDigital: mentioned this above, for some reason loading basis crashes electron :D

packages/assets/src/cache/Cache.ts
GoodBoyDigital: i do!

packages/assets/src/cache/Cache.ts
GoodBoyDigital: :P

packages/assets/src/cache/parsers/index.ts
```diff
@@ -1,3 +1,3 @@
-export * from './cacheSpriteSheet';
+export * from './cacheSpritesheet';
```
bigtimebuddy, line 1: He found it! Good job @Zyie

Zyie (approved):

PR #8447, opened 2022-06-27
Adjust asset loaders

Zyie:
This PR adds the functionality of the old loaders to the new loaders

Bitmap Font Changes:
- Can now load `fnt` and `txt` fonts
- Can load split fonts
- Can load xml string fonts
- Added more tests

Spritesheet Changes:
- Can now load multi packed spritesheets

Web Font Changes:
- You can now parse in all the options that the `FontFace` API accepts
- You can now provide a `family` override 

Other Changes:
- Added text loader
- Reorganised the test assets folder
- Export the `TextFormat`, `XMLFormat` and 'XMLStringFormat` from `text-bitmap` for use is loader
- Updated the server path for loading files so it works both locally and in github actions
- Cleaned up/Added some documentation

TODO in follow up:
- go over type implementation. There are a few areas where we can be smarter about types and reduce casting


packages/assets/src/loader/parsers/loadWebFont.ts
```diff
@@ -50,18 +60,22 @@ export const loadWebFont = {
...
 
             for (let i = 0; i < weights.length; i++)
             {
                 const weight = weights[i];
 
-                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-                // @ts-ignore
                 const font = new FontFace(name, `url(${url})`, {
                     weight,
+                    variant: options.data?.variant,
```
GoodBoyDigital, line 73: could this be 

```
const data = ...options.data ?? {};

////
{
 ...data,
weight
}

```

GoodBoyDigital (approved):
nice!

PR #8458, opened 2022-06-30
Fix incorrect `saturate` blend mode mapping

### LINKED ISSUE #8457: Incorrect `saturate` blend mode mapping
### Expected Behavior

The `BLEND_MODES.SATURATION` should be mapped to the string `saturation` ([MDN Doc: globalCompositeOperation](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation))

### Current Behavior

It is currently mapped to `saturate`.

https://github.com/pixijs/pixijs/blob/de63538047d615e7230b4300a2ed2f33f623e0e6/packages/canvas-renderer/src/utils/mapCanvasBlendModesToPixi.ts#L30

### Environment

- **`pixi.js` version**: _6.4.2_

Julien-Marcou:
Fixes #8457

##### Description of change

- Replace incorrect `saturate` value for blend mode mapping with `saturation`
- ~Add `dom` lib to the `tsconfig.json` so we can use the `GlobalCompositeOperation` type to avoid this kind of error~ (needs ts 4.6+)

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas-renderer/src/utils/mapCanvasBlendModesToPixi.ts
+++ b/packages/canvas-renderer/src/utils/mapCanvasBlendModesToPixi.ts
@@ -9,6 +9,7 @@ import { canUseNewCanvasBlendModes } from './canUseNewCanvasBlendModes';
  * @param {string[]} [array=[]] - The array to output into.
  * @returns {string[]} Mapped modes.
  */
+// TODO after upgrading to typeScript 4.6, replace `string[]` with `GlobalCompositeOperation[]`
 export function mapCanvasBlendModesToPixi(array: string[] = []): string[]
 {
     if (canUseNewCanvasBlendModes())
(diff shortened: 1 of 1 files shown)


Zyie:
@bigtimebuddy the type errors can be fixed by updating to the latest version of typescript. However the api-extractor is breaking in the latest version so we cant update just yet

I've raised it as an issue on there repo so hopefully it will get fixed

We could add the `GlobalCompositeOperation` to our `global.d.ts` file, or just leave the type as `string[]` until its fixed?

bigtimebuddy:
@Julien-Marcou I'm all for making the minimal change "saturate" => "saturation" 

We'll have to fix the types issue later. It would be helpful to ad a TODO comment to make sure we fix that.

Julien-Marcou:
> @bigtimebuddy the type errors can be fixed by updating to the latest version of typescript

That's what I was seeing ^^ The `dom` lib is only needed for ESLint, but for TypeScript typing to work, we also need to upgrade to TypeScript 4.6+

I'll leave the `string` typing and add a todo.

bigtimebuddy (approved):

bigtimebuddy:
Thanks @Julien-Marcou

PR #8464, opened 2022-07-04
Fix webfont loader

Zyie:
The webfont loader should return either a single font face or an array as we load multiple different weights

### DIFF
--- a/packages/assets/src/loader/parsers/loadWebFont.ts
+++ b/packages/assets/src/loader/parsers/loadWebFont.ts
@@ -94,13 +94,7 @@ export const loadWebFont = {
 
     unload(font: FontFace[]): void
     {
-        if (Array.isArray(font))
-        {
-            font.forEach((t) => document.fonts.delete(t));
-        }
-        else
-        {
-            document.fonts.delete(font);
-        }
+        (Array.isArray(font) ? font : [font])
+            .forEach((t) => document.fonts.delete(t));
     }
 } as LoaderParser<FontFace | FontFace[]>;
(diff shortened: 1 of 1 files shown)


GoodBoyDigital (approved):

packages/assets/src/loader/parsers/loadWebFont.ts
```diff
@@ -80,19 +81,26 @@ export const loadWebFont = {
...
-            console.warn('[loadWebFont] FontFace API is not supported. Skipping loading font');
-        }
+
+        console.warn('[loadWebFont] FontFace API is not supported. Skipping loading font');
 
         return null;
     },
 
-    unload(font: FontFace): void
+    unload(font: FontFace[]): void
```
Julien-Marcou, line 95: Shouldn't that be `FontFace | FontFace[]`?

packages/assets/src/loader/parsers/loadWebFont.ts
```diff
@@ -80,19 +81,26 @@ export const loadWebFont = {
...
     {
-        document.fonts.delete(font);
+        if (Array.isArray(font))
+        {
+            font.forEach((t) => document.fonts.delete(t));
+        }
+        else
+        {
+            document.fonts.delete(font);
+        }
```
bigtimebuddy, line 104: Slightly more terse way to describe the same thing:
```suggestion
        (Array.isArray(font) ? font : [font])
            .forEach((t) => document.fonts.delete(t));
```

bigtimebuddy (approved):

PR #8470, opened 2022-07-06
Added support for Data URLs (base64 only)

SmartArray:
### base64-url-encoded images
This PR adds support for base64 encoded images. Now the new AssetLoader can be used as follows: 

```
const texture = await PIXI.Assets.load('data:image/png;base64,iVBORw0KGgoAAAA…');
```

### DIFF
--- a/bundles/pixi.js-node/src/adapter/loadNodeTexture.ts
+++ b/bundles/pixi.js-node/src/adapter/loadNodeTexture.ts
@@ -7,6 +7,18 @@ import path from 'path';
 import { NodeCanvasElement } from './NodeCanvasElement';
 
 const validImages = ['.jpg', '.png', '.jpeg', '.svg'];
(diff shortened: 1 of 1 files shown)


bundles/pixi.js-node/src/adapter/loadNodeTexture.ts
```diff
@@ -7,14 +7,26 @@ import path from 'path';
 import { NodeCanvasElement } from './NodeCanvasElement';
 
 const validImages = ['.jpg', '.png', '.jpeg', '.svg'];
+const validMimes = ['image/png', 'image/jpg', 'image/jpeg', 'image/svg'];
+
+function isSupportedDataURL(url)
```
bigtimebuddy, line 12:
```suggestion
function isSupportedDataURL(url: string): boolean
```

SmartArray:
Thank you @bigtimebuddy

PR #8472, opened 2022-07-07
Use esbuild and separate targets for browser and module output

bigtimebuddy:
### Changes

* Modules (commonjs and esm) now outputs ES2020
* Browser now outputs ES2017
* Use esbuild with Rollup for better performance

See #8454 cc @Julien-Marcou 

This shaves about 1 minute off our Build for Distribution step in our CI


bigtimebuddy:
Related: I'm thinking about eliminating commonjs all together and setting type=module for all packages. All modern build tools and runtimes understand esmodules, do we need commonjs anymore?

Julien-Marcou:
LGTM, build time when down from 3min to 30sec on my laptop 🚀

> I'm thinking about eliminating commonjs all together

A lot of nodejs packages are still using commonjs, but I'm all for ESM.

bigtimebuddy:
Reading more on the topic tonight, the interop between cjs and mjs is a real pain, it's worth having both probably. And it doesn't really save us much other than simplify the build step.

andrewstart (approved):

Zyie (approved):

PR #8478, opened 2022-07-09
Upgrade to earcut 2.2.4

dev7355608:
> Changelog
>
> 2.2.4 (Jul 5, 2022)
>
> - Improved performance by 10–15%.
> - Fixed another rare race condition that could lead to an infinite loop.
> 
> 2.2.3 (Jul 8, 2021)
> 
> - Fixed a rare race condition that could lead to an infinite loop.

Shouldn't the latest version that matches the version constraint be automatically included before publishing? PIXI 6.4.2 was shipped with earcut 2.2.2, while 2.2.3 was already a year old and matches ^2.2.2.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10333,7 +10333,7 @@
       "license": "BSD-3-Clause"
     },
     "node_modules/earcut": {
-      "version": "2.2.2",
+      "version": "2.2.4",
       "license": "ISC"
     },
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):

bigtimebuddy:
You need to update the lock file too.

PR #8482, opened 2022-07-12
Add detection parsers

Zyie:
## FormatDetectionParser
This PR introduces the idea of a `FormatDetectionParser`

This parser consists of 3 main things

- `test()` -> returns whether or not a format is supported e.g. webp/avif/basis etc
- `add()` -> gives the parser the chance to add whatever formats its wants if the test passes
- `remove()` -> allows the parser to remove any formats it finds is not supported

The reason we need add and remove is for the 2 different use cases

#### Removing:
```
// User specifies what formats they want
await Assets.init({
    formats: ['webp', 'avif', 'png']
}
```
In this case we would want to remove any formats that are not supported

#### Adding:
```
extensions.add(MyNewAwesomeFormat);
await Assets.init()
```
In this case we would want to add the users if the test passes to the default set of formats

## resolveCompressedTextureUrl

This new resolver parser adds a user friendly way of selecting what compressed texture to use
You can now use the `FORMAT.ktx` naming convention to get the correct texture
```
{
    "name": "profile-abel",
    "srcs": [
        "asset.s3tc.ktx",
        "asset.etc.ktx",
        "asset.pvrtc.ktx",
        "asset.png"
    ]
}
```

This is the list of available formats
```
s3tc.ktx
s3tc_sRGB.ktx
etc.ktx
etc1.ktx
pvrt.ktx
atc.ktx
astc.ktx
```


packages/assets/src/loader/parsers/loadCompressedTextures.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        let basePath = dirname(loadAsset.src);
+
+        if (basePath && basePath.lastIndexOf('/') !== (basePath.length - 1))
+        {
+            basePath += '/';
+        }
+
+        Assets.resolver.add(asset.cacheID, basePath + textureURL);
```
Zyie, line 98: @GoodBoyDigital Is there a better way to do this? It feels wrong having to add `Assets`

For context, the manifest allows you to set a `cacheID` to reference the texture instead of the url as the url will change depending on the device

Here is the example manifest
```
{
    "cacheID":"asset",
     "textures":[
            { "src":"asset.fallback.png" },
            { "format":"COMPRESSED_RGBA_S3TC_DXT5_EXT", "src":"asset.s3tc.ktx" },
            { "format":"COMPRESSED_RGBA8_ETC2_EAC", "src":"asset.etc.ktx" },
            { "format":"RGBA_PVRTC_4BPPV1_IMG", "src":"asset.pvrtc.ktx" }
        ]
}
```

Zyie:
@bigtimebuddy updated this PR to use the new extensions changes

I also changed the base so that it can be merged after #8488, to avoid a new set of conflicts

PR #8485, opened 2022-07-13
Eliminate redundant handling code in extensions

bigtimebuddy:
Noticed a few places for re-use and this could use some cleanup.

### Changed

* Adds `extensions.handleByList`
* Adds `extensions.handleByMap`
* Allow chaining for all `extensions` methods
* Adds tests for `extensions`

### DIFF
--- a/packages/assets/src/Assets.ts
+++ b/packages/assets/src/Assets.ts
@@ -774,21 +774,10 @@ export class AssetsClass
 export const Assets = new AssetsClass();
 
 // Handle registration of extensions
-extensions.handle(
-    ExtensionType.LoadParser,
-    (extension) => { Assets.loader.addParser(extension.ref); },
-    (extension) => { Assets.loader.removeParser(extension.ref); }
-);
-extensions.handle(
-    ExtensionType.ResolveParser,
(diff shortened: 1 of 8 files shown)


packages/assets/src/Assets.ts
```diff
@@ -774,21 +774,10 @@ export class AssetsClass
...
-    (extension) => { Assets.resolver.addUrlParser(extension.ref); },
-    (extension) => { Assets.resolver.removeUrlParser(extension.ref); }
-);
-extensions.handle(
-    ExtensionType.CacheParser,
-    (extension) => { Assets.cache.addParser(extension.ref); },
-    (extension) => { Assets.cache.removeParser(extension.ref); }
-);
+extensions
+    .handleByList(ExtensionType.LoadParser, Assets.loader.parsers)
```
bigtimebuddy, line 778: @Zyie do we still need the add remove parser methods?

Zyie: We use them in the tests but I think that was just left over from before everything got moved to extensions

So no i don't think these would be used anymore

I can do a fast follow PR to update these tests and remove the functions if you want?

PR #8506, opened 2022-07-17
Bundle ismobilejs into settings

### LINKED ISSUE #8502: Blocking bug with @pixi/node, unable to use with non-CommonJS modules
I've been experimenting with the `@pixi/node` package available in v6.5.0-rc.2. I have no problems when it's using CommonJS (e.g., `require('@pixi/node')`), but I am having issues using modules (e.g., `import * as PIXI from '@pixi/node'`).

Here's a simple example:

https://github.com/bigtimebuddy/pixi-node-example/blob/module/index.mjs

Running this produces the following error:

```
import { Extract } from '@pixi/extract';
         ^^^^^^^
SyntaxError: Named export 'Extract' not found. The requested module '@pixi/extract' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@pixi/extract';
const { Extract } = pkg;

    at ModuleJob._instantiate (internal/modules/esm/module_job.js:124:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:179:5)
    at async Loader.import (internal/modules/esm/loader.js:178:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
    at async handleMainPromise (internal/modules/run_main.js:59:12)
```

Having all packages be both CommonJS and ESM seems to confuse Node, even with the proper type. I have not found a way to force dependencies to use modules. Kinda stuck, any ideas?

A workaround would be to bundle everything (except the 3rd party dependencies like canvas, gl, etc) so that it's a single file. Providing cjs and mjs files seems like it would work. 

----

EDIT: think I figured out what needs to happen here: 

### TODO

* [x] Add `exports` to assets (#8504)
* [x] Bundle ismobilejs or add to adapter (#8506)
* [ ] Rename esm extension to ".mjs" for all packages (#8505)

bigtimebuddy:
Partially fixes #8502 

This bundles the `ismobilejs` dependency, which avoids import issues because of the non-Node.js nature of the package. Honestly, bundling this into settings it's a bad idea because it's 1) very small and 2) is stable for our purposes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13953,6 +13953,7 @@
     },
     "node_modules/ismobilejs": {
       "version": "1.1.1",
+      "dev": true,
       "license": "MIT"
     },
     "node_modules/isobject": {
@@ -25140,7 +25141,7 @@
       "name": "@pixi/settings",
       "version": "6.5.0-rc.2",
       "license": "MIT",
-      "dependencies": {
+      "devDependencies": {
         "ismobilejs": "^1.1.0"
       }
     },
@@ -34741,7 +34742,8 @@
       "version": "2.0.0"
     },
     "ismobilejs": {
-      "version": "1.1.1"
+      "version": "1.1.1",
+      "dev": true
     },
     "isobject": {
       "version": "3.0.1",
--- a/packages/settings/package.json
+++ b/packages/settings/package.json
@@ -37,7 +37,7 @@
     "dist",
     "*.d.ts"
   ],
-  "dependencies": {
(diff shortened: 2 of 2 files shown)


GoodBoyDigital (approved):
nice catch! to confirm - moving something to a devDependencies mean it will get bundled in?

bigtimebuddy:
That's right, I built that logic into our Rollup script.

PR #8513, opened 2022-07-18
Fix jest not exiting properly when killing the http server

### LINKED ISSUE #8512: Tests run indefinitely
When running Pixi's tests from WSL2 (I suspect it might be related to Windows, like described in this [Stack Overflow Post](https://stackoverflow.com/questions/23706055/why-can-i-not-kill-my-child-process-in-nodejs-on-windows)), Jest prints the following warning in the console:

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```
image

Julien-Marcou:
#### Description of change

Fixes #8512

It uses `spawn` instead of `exec` so the `kill()` signal is received by the child process (looks like it's a bug on Windows, I don't have access to a native Linux distribution right now to test if this problem was present on Linux too).

I also replaced the `setTimeout` of 1sec on the setup, instead it uses the `stdout` of the child process to detect when the server is ready (which is much faster ~200ms on my laptop, and more reliable than just wait)

#### Pre-Merge Checklist

### DIFF
--- a/test/jest-global-setup.ts
+++ b/test/jest-global-setup.ts
@@ -1,19 +1,30 @@
-import { exec } from 'child_process';
-import path from 'path';
+import { spawn } from 'child_process';
+import { join } from 'path';
 
 // eslint-disable-next-line func-names
 module.exports = async function ()
 {
     if (!process.env.GITHUB_ACTIONS)
     {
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Nice :+1:

PR #8518, opened 2022-07-21
Fix LoaderResource tests

Julien-Marcou:
Another attempt to fix the LoaderResource tests (even though it's deprecated and will be replaced) 😄

#### Description of change

- Rename `packages/loaders/test/LoaderResource.test.ts` to `packages/loaders/test/LoaderResource.tests.ts` so that jest picks up the file
- Add `sinon` dependency (used to fake XHR request for the LoaderResource tests)
- Fix the LoaderResource tests
- Re-enable some skipped tests
- Remove no longer needed `@types/chai` dependency

#### Pre-Merge Checklist


bigtimebuddy:
What a hero! Thank you for fixing this.

One small suggestion: could you do a `git mv` for the LoaderResource.test.ts file? Looks like a delete and insert here.

bigtimebuddy (approved):
Just the `git mv` comment, but not blocking since we are throwing this away anyways.

Julien-Marcou:
Ha! Weird, it is a rename, we can see it in the first commit, github shows it correctly, but in the squashed result it doesn't pick it up, I guess there is too much diff with the second commit so the squash doesn't pick the rename anymore

Julien-Marcou:
Yep, even with a dedicated commit, github doesn't pick it up. The only way to make it work would be to not squash this PR I think 😅

bigtimebuddy:
Thank you!

Issue #8432, opened 2022-06-24
Bitmap font is shaking/jittering on newer pixi versions


tsopeh:
### Expected Behavior

I expect for bitmap text to be rendered without shaking, as in previous pixi versions (e.g. 5.3.12). The bitmap font is guaranteed to be monospaced since every glyph has the same `xadvance`.

### Current Behavior

The text appears to be slightly shaking when changing the value. In the demo I provided, I used the version `6.4.2` as a reference, but I believe that the issues started appearing from the version `6.3.0`.

This is possibly related to #8246.

### Steps to Reproduce

The demo code can be found here: https://drive.google.com/file/d/1IVs3gt8ImxiDDhxTLMRyzzx56nr0SXIa/view?usp=sharing
You can toggle pixijs version in the `index.html`.

### Environment

- **`pixi.js` version**: _6.3.0 – 6.4.2_

lazarljubenovic:
We're also experiencing this issue upon trying to upgrade from `6.0.0-rc.2` to the newest version.

tsopeh:
Here's some video samples that illustrates the issue in a real application:

1. Without jittering (good behaviour on older versions)

https://user-images.githubusercontent.com/37545996/178275581-7b8d2c0e-0976-41d2-bb16-0ea89d2c816d.mp4

2. With jittering (bad behaviour on newer versions)

https://user-images.githubusercontent.com/37545996/178275651-32038511-f1e1-472e-8903-d87ede9a22bf.mp4

tsopeh:
Thanks for the explanation and a workaround @bigtimebuddy.

grebinka:
xOffset and xAdvance are added while calculating lastLineWidth. But only xAdvance should be.
That's why everything works with xOffset = 0
I've added pull request to fix this problem (#8536)

PR #7933, opened 2021-10-31
Fix incorrect BitmapText text width calculation (#7758)

kjarrio:
##### Description of change
Fix for BitmapText width calculation when using monospace fonts. Fixes issue #7758 - and doesn't break PR #7479 / #7245

##### Pre-Merge Checklist
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #7479, opened 2021-05-09
Fix lineWidth calculation in BitmapText#updateText

SukantPal:
Fixes #7245

| Before              |  After                       |
| -------------- | -------------------|
| Screen Shot 2021-05-09 at 12 06 18 PM | Screen Shot 2021-05-09 at 12 06 40 PM |

##### Test program

```js
const app = new PIXI.Application({ backgroundColor: 0x1099bb });
document.body.appendChild(app.view);

app.loader
    .add('desyrel', 'https://pixijs.io/examples/examples/assets/bitmap-font/desyrel.xml')
    .load(onAssetsLoaded);

function onAssetsLoaded() {
    const bitmapFontText = new PIXI.BitmapText('T\nY\nq\nb', { fontName: 'Desyrel', align: 'left' });

    bitmapFontText.x = 20;
    bitmapFontText.y = 200;

    app.stage.addChild(bitmapFontText);

    const graphics = new PIXI.Graphics();

    // Rectangle
    graphics.beginFill(0xDE3249, .87);
    graphics.drawRect(bitmapFontText.x, 0, bitmapFontText.textWidth, 10000);
    graphics.endFill();

    app.stage.addChild(graphics);
}

```

##### Description of change

Update `lastLineWidth` based on the rendered position of the character plus its texture width, instead of the next character's layout position.

PR #8536, opened 2022-07-27
Fix incorrect BitmapText text width (#8432)

grebinka:
##### Description of change

charRenderData.position.x has already include xOffset so you should decrease xAdvance by it

##### Pre-Merge Checklist

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -300,7 +300,7 @@ export class BitmapText extends Container
 
             chars.push(charRenderData);
 
-            lastLineWidth = charRenderData.position.x + Math.max(charData.xAdvance, charData.texture.orig.width);
+            lastLineWidth = charRenderData.position.x + Math.max(charData.xAdvance - charData.xOffset, charData.texture.orig.width);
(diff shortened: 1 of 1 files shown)


tsopeh:
I can confirm that this fix resolves [my problem](https://github.com/pixijs/pixijs/issues/8432#issuecomment-1180415626) described in #8432.

bigtimebuddy:
@grebinka thanks for the fix. Would you mind creating a preview of this? Would be helpful to see some examples for verification.

grebinka:
@bigtimebuddy could you give me any example of preview please? I actually don't exactly understand what it should be like.

grebinka:
Also I don't understand why not just sum char's xAdvance whith _letterSpacing. I guess everybody expect such behavior.

bigtimebuddy:
@grebinka Examples are the easiest way for us to understand the visual impact of this change. The implementation in the PR seems correct, but it would be helpful to have a few examples. There may be some demos you may be able to repurpose in #7479, #7933 or searching for BitmapText in issues/pull-requests.

And here's is a build you can use: https://pixijs.download/pr-8536/pixi.min.js

bigtimebuddy (approved):
I was able to verify examples that this fixed this jittering in the attached issue.

PR #8541, opened 2022-07-28
Fix createImageBitmap with null alphaMode

aheubusch:
##### Description of change
In commit 02a186048e4cd7a24d517c4840af2e7578c68e14 `premultiplyAlpha` was changed to `alphaMode` and it broke the `createBitmap` use case when no alphaMode is specified (which is the default). Before a undefined

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/resources/ImageResource.ts
+++ b/packages/core/src/textures/resources/ImageResource.ts
@@ -209,7 +209,8 @@ export class ImageResource extends BaseImageResource
             .then((blob) => createImageBitmap(blob,
                 0, 0, source.width, source.height,
                 {
-                    premultiplyAlpha: this.alphaMode === ALPHA_MODES.UNPACK ? 'premultiply' : 'none',
+                    premultiplyAlpha: this.alphaMode === null || this.alphaMode === ALPHA_MODES.UNPACK
+                        ? 'premultiply' : 'none',
                 }))
             .then((bitmap: ImageBitmap) =>
             {


bigtimebuddy (changes requested):
I don't think this is the correct place to handle this. I'd consider changing this instead:

https://github.com/pixijs/pixijs/blob/a60bcf0e9320db3f0f031863f7f3bb2c269c9117/packages/core/src/textures/resources/ImageResource.ts#L110

The `null` here should probably be `ALPHA_MODES.UNPACK` instead.

packages/core/src/textures/resources/ImageResource.ts
```diff
@@ -45,7 +45,7 @@ export class ImageResource extends BaseImageResource
     /**
      * Controls texture alphaMode field
      * Copies from options
-     * Default is `null`, copies option from baseTexture
```
bigtimebuddy, line 48: I'm sorry, I missed this behavior, maybe this isn't the right fix after all...

Can you clarify the problem a little more or provide an example when this happens?

aheubusch: Will try my best. The source image is assumed to be unpremultiplied alpha/straight alpha. WebGL works with PMA though. Therefor the default for alphaMode "is" PIXI.ALPHA_MODES.UNPACK/PIXI.ALPHA_MODES.PREMULTIPLY_ON_UPLOAD. Excepts its only the case for BaseTexture and not for ImageResource.
https://github.com/pixijs/pixijs/blob/fd9c9305b5722e09d8e9b0927dd2b093668ced60/packages/core/src/textures/BaseTexture.ts#L244
When the image is uploaded to the gpu webgl converts the image to a pma texture.
https://github.com/pixijs/pixijs/blob/fd9c9305b5722e09d8e9b0927dd2b093668ced60/packages/core/src/textures/resources/BaseImageResource.ts#L94

With `ImageBitmap` (from `CreateBitmapImage`) some `gl.pixelStorei` values are ignored. https://registry.khronos.org/webgl/specs/latest/1.0/#6.10
> If the TexImageSource is an ImageBitmap, then these three parameters will be ignored. Instead the equivalent [ImageBitmapOptions](https://html.spec.whatwg.org/multipage/webappapis.html#imagebitmapoptions) should be used to create an ImageBitmap with the desired format.

Because `alphaMode` is null when the `ImageBitmap` is created the image is not converted to pma.
https://github.com/pixijs/pixijs/blob/fd9c9305b5722e09d8e9b0927dd2b093668ced60/packages/core/src/textures/resources/ImageResource.ts#L210-L214

SuperSodaSea: I found the same problem when I was working on `ImageBitmapResource`. In short:

1. `BaseTexture`'s default `alphaMode` is `ALPHA_MODES.PREMULTIPLY_ON_UPLOAD`, and Pixi.js will assume the texture is a premultiplied one.
2. `ImageResource`'s default `alphaMode` is `null`, result in `premultiplyAlpha: 'none'` when `createImageBitmap`, so the `ImageBitmap` is actually non-premultiplied.
3. So the `ImageBitmap` is actually non-premultiplied, but `BaseTexture` assumed it is premultiplied, result in wrong rendering result.

So in order to fix this, our goal is to create the `ImageBitmap` as a premultiplied one when `ImageResource.alphaMode` is not specified. But if we directly change the default value of `ImageResource.alphaMode` (current commit https://github.com/pixijs/pixijs/commit/032cc9a770e58d3fbcac235ca98883cb0a080093), it will fail on these [unit tests](https://github.com/pixijs/pixijs/actions/runs/2777392142/jobs/4431819464#step:8:690):

- ImageResource › alphaMode behaviour › should override BaseTexture alphaMode if specified
- ImageResource › alphaMode behaviour › should not override BaseTexture alphaMode if not specified

So in my opinion it is better to leave `alphaMode` to be `null` and just set `premultiplyAlpha: 'premultiply'` to make the `ImageBitmap` premultiplied, just like the original commit a60bcf0e9320db3f0f031863f7f3bb2c269c9117, except a line wrap is needed since eslint warns you there is a limit for the line length of 125. Just like the new `ImageBitmapResource` I written in #8698:

https://github.com/pixijs/pixijs/blob/8f38e040badac265082f92f7155e7faffa6e33c0/packages/core/src/textures/resources/ImageBitmapResource.ts#L114-L117

@bigtimebuddy What about your opinion?
And
@aheubusch are you willing to complete this PR? If you don't have time, I can fix it in other PR.

bigtimebuddy: @SuperSodaSea that makes sense. Thanks for looking into it. It's cool with me if you want to make another PR.

bigtimebuddy (approved):

bigtimebuddy:
Thanks @aheubusch!

PR #8558, opened 2022-08-06
Refactor Extensions to Self-Install

bigtimebuddy:
The goal of this PR is to remove all of the explicit registrations of extensions. Instead, plugins _self-add_ themselves. All the developer needs to do is import the package. This makes it _much simpler_ to use packages without having hook up all of the correct pieces.

This is a **breaking change** as current v6 could would no longer work or would harmlessly add the extension twice.

### Example

In v6:

```js
import { AccessibilityManager } from '@pixi/accessibility';
import { extensions } from '@pixi/core';
extensions.add(AccessibilityManager);
```

Now becomes:

```js
import '@pixi/accessibility';
```


packages/assets/test/detections.tests.ts
```diff
@@ -1,31 +1,21 @@
...
     });
 
     it('should remove any unsupported formats', async () =>
     {
-        extensions.add(detectCompressedTextures);
         detectCompressedTextures.test = jest.fn(async () => false);
         await Assets.init();
         expect(Assets.resolver['_preferredOrder'][0].params.format).toEqual(
-            ['avif', 'webp', 'png', 'jpg', 'jpeg']
+            ['webp', 'avif', 'png', 'jpg', 'jpeg']
```
Zyie, line 18: I think this highlights a problem. load/detection parsers are order specific sometimes.

For example, the detection parsers add themselves into the array in the order they are called, but now `webp` is being called before `avif`, meaning the priority has changed.

This could also be an issue for the loaders as you could have a generic `loadTxt` and `loadSpecificTxt` and you would want the specific version to be added after the generic one.

Maybe this means these shouldn't be extensions, or the extensions API needs a `addBefore`/`addAfter`, or a priority system?

bigtimebuddy: In v7 there already is a priority flag on ExtensionMetadata. I can fix that.

bigtimebuddy: I reworked this a little bit so that all the detections had a priority. I also created an explicit detection for populating the default formats, the unshift was doing some funny business.

Zyie (approved):

PR #8559, opened 2022-08-06
Be able to load base64 assets

talhaozdemir:
##### Description of change

A line of check that controls url if it includes base64 string in it.

##### Pre-Merge Checklist


packages/assets/src/loader/parsers/loadTexture.ts
```diff
@@ -46,7 +46,7 @@ export const loadTextures = {
         const tempURL = url.split('?')[0];
         const extension = tempURL.split('.').pop();
 
-        return validImages.includes(extension);
+        return url.includes('data:image') || validImages.includes(extension);
```
bigtimebuddy, line 49: This could return a false-positive with "data:image/svg+xml", we probably need something stricter that considers the valid types too. 

Also, includes() is a little too ambiguous. Should probably use indexOf or RegExp instead.

bigtimebuddy (changes requested):
Thanks for the PR. This needs a little work, and it would be great to add some tests to cover this very common use case.

packages/assets/src/loader/parsers/loadTexture.ts
talhaozdemir: > This could return a false-positive with "data:image/svg+xml", we probably need something stricter that considers the valid types too.
> 
> Also, includes() is a little too ambiguous. Should probably use indexOf or RegExp instead.

I have thought it like:

`    test: function (url) {
        var tempURL = url.split('?')[0];
        var extension = tempURL.split('.').pop();

        var isValidBase64;
        for (var i = 0; i < validImages$4.length; i++) {
            if (url.indexOf('data:image/' + validImages$4[i]) > -1) {
                isValidBase64 = true;
                break;
            }
        }

        return isValidBase64 || validImages$4.includes(extension);
    }`

and this snippet is working well. If it is okay I can create a PR for it.

talhaozdemir:
> Thanks for the PR. This needs a little work, and it would be great to add some tests to cover this very common use case.

Sure I will, but where should I add these use cases?

packages/assets/src/loader/parsers/loadTexture.ts
bigtimebuddy: This is better! The only change I would make is === 0 instead of > -1.

bigtimebuddy:
The packages/assets/tests folder has some tests you can follow as guidance.

bigtimebuddy (commented):
Please do not auto-format. Can you revert all these style changes?

talhaozdemir:
Sorry for that, sure I will.

bigtimebuddy (approved):
Nice 👍

talhaozdemir:
By the way I cannot run tests from the terminal, I get this error all the time.

image

Zyie (approved):

bigtimebuddy:
Thank you @talhaozdemir

bigtimebuddy:
Oops, I missed this, but in the future PRs should be made to **dev** not **main** unless it's a critical/blocking issue.

PR #8565, opened 2022-08-09
Fix loadBundle progress

Zyie:
`loadBundle`'s `onProgress` event would fire for the individual assets per bundle, resulting in something like this when loading multiple bundles

`0.3, 0.6, 1, 0.5, 1`

~~This fix only counts when all the bundle loads and produces the expected output~~

~~`0.5, 1`~~

This fix now uses the total assets of all bundles. If 2 bundles are loaded, each containing 2 assets it produces this

`0.25, 0.5, 0.75, 1`

### DIFF
--- a/packages/assets/src/Assets.ts
+++ b/packages/assets/src/Assets.ts
@@ -523,13 +523,17 @@ export class AssetsClass
 
         const out: Record<string, Record<string, any>> = {};
 
-        const promises = Object.keys(resolveResults).map((bundleId) =>
+        const keys = Object.keys(resolveResults);
+        let count = 0;
+        const total = keys.length;
+        const promises = keys.map((bundleId) =>
         {
             const resolveResult = resolveResults[bundleId];
 
-            return this._mapLoadToResolve(resolveResult, onProgress)
+            return this._mapLoadToResolve(resolveResult)
                 .then((resolveResult) =>
                 {
+                    if (onProgress) onProgress(++count / total);
                     out[bundleId] = resolveResult;
                 });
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Wouldn't we want something like this instead?

0.15, 0.3, 0.5, 0.75, 1

onProgress for the bundles completed seems less helpful.

Zyie:
> Wouldn't we want something like this instead?
> 
> 0.15, 0.3, 0.5, 0.75, 1
> 
> onProgress for the bundles completed seems less helpful.

That is a good point 😄 updating it now

Zyie:
@bigtimebuddy updated

packages/assets/src/Assets.ts
```diff
@@ -523,11 +523,20 @@ export class AssetsClass
 
         const out: Record<string, Record<string, any>> = {};
 
-        const promises = Object.keys(resolveResults).map((bundleId) =>
+        const keys = Object.keys(resolveResults);
+        let count = 0;
+        let total = 0;
+        const _onProgress = () =>
+        {
+            if (onProgress) onProgress(++count / total);
```
bigtimebuddy, line 531: What about this?

```suggestion
            onProgress?.(++count / total);
```

bigtimebuddy (commented):
What do you think about onProgress calling an initial value of 0? Could be helpful for folks animating a progress bar.

bigtimebuddy (approved):

GoodBoyDigital (approved):

PR #8568, opened 2022-08-09
Support in Assets text-based loading of fnt BitmapFonts

### LINKED ISSUE #8557: Assets loader fails to load non-XML bitmap fonts from bundle
In the new Assets loader, documentation and the code's list of supported bitmap font types are listed as `['.xml', '.fnt']`. In the `loadBitmapFont` code, it appears to support both formats in `testParse` and `parse`, but `load` is only passing the data loaded into the XML parser, producing undefined results for non-XML .fnt bitmap fonts.

https://github.com/pixijs/pixijs/blob/fd9c9305b5722e09d8e9b0927dd2b093668ced60/packages/assets/src/loader/parsers/loadBitmapFont.ts#L64-L73

**Steps to reproduce**
Use Assets.addBundle, then Assets.loadBundle referencing a bundle that includes a non-XML .fnt bitmap font.
It errors, trying to reference bitmap font properties on an undefined object returned from the asset loader.

**Environment**
pixi.js v6.5.1

bigtimebuddy:
Fixes #8557

### DIFF
--- a/packages/assets/src/loader/parsers/loadBitmapFont.ts
+++ b/packages/assets/src/loader/parsers/loadBitmapFont.ts
@@ -2,32 +2,13 @@ import type { Texture } from '@pixi/core';
 import { ExtensionType } from '@pixi/core';
 import { settings } from '@pixi/settings';
 import type { BitmapFontData } from '@pixi/text-bitmap';
-import { BitmapFont, TextFormat, XMLFormat, XMLStringFormat } from '@pixi/text-bitmap';
+import { BitmapFont, TextFormat, XMLStringFormat } from '@pixi/text-bitmap';
 import { dirname, extname, join } from '../../utils/path';
 
 import type { Loader } from '../Loader';
 import type { LoadAsset } from '../types';
 import type { LoaderParser } from './LoaderParser';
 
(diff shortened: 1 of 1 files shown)


packages/assets/src/loader/parsers/loadBitmapFont.ts
```diff
@@ -41,35 +22,37 @@ export const loadBitmapFont = {
...
+    async load(url: string, _options: LoadAsset): Promise<string>
     {
         const response = await settings.ADAPTER.fetch(url);
 
-        const text = await response.text();
-
-        const data = new window.DOMParser().parseFromString(text, 'text/xml');
-
-        return await _loadBitmap(url, XMLFormat.parse(data), loader);
+        return response.text();
```
bigtimebuddy, line 55: If someone is trying to load an XML file that is _not_ a font, this means they would get back an XML string instead of an XMLDocument. We may need a generic loader like the json/txt loaders. Using XML for data isn't super common these days, so I'm not that concerned about it.

Zyie: I don't fully remember how the unloading works but returning a non `BitmapFont` might cause an issue as we call `font.destroy()` in `unload`

I will test this out and report back

Zyie (approved):

PR #8569, opened 2022-08-09
Fix Resolver basePath setter doc saying it will append (-> prepend)

qwertywelch:
##### Description of change

The Assets Resolver basePath is prepended, not appended. Fixed the setter docs to say "prepend" instead of "append".

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good catch 👍

PR #8573, opened 2022-08-09
Loosen drawPolygon path type with IPointData

mhazy:
##### Description of change

Sync things up with Polygon's points type (https://github.com/pixijs/pixijs/blob/dev/packages/math/src/shapes/Polygon.ts#L23) and allows the caller to pass objects satisfying `{ x: number, y: number }`.

Docs output:

image


bigtimebuddy (approved):
Nice improvement 👍

bigtimebuddy:
Thank you @mhazy

PR #8574, opened 2022-08-10
Add svg loader

Zyie:
- Adds an SVG loader
- Abstracts some duplicated logic between all the different texture loaders
- Fixes the `LoadAsset` type for all the texture loaders


packages/assets/src/loader/parsers/textures/loadSVG.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { LoadAsset } from '../../types';
+
+import type { LoaderParser } from '../LoaderParser';
+import { loadTextures } from './loadTexture';
+import { createTexture } from './utils/createTexture';
+
+export type SVGTextureData = {
+    baseTexture: BaseTexture;
+    svgResource: ISVGResourceOptions;
+};
```
bigtimebuddy, line 16: I don't understand the `baseTexture`, shouldn't this be `IBaseTextureOptions`? What about this instead since IBaseTextureOptions contains `resourceOptions`

```suggestion
export type SVGTextureData = IBaseTextureOptions;
```

Zyie: yeah you are correct. I've updated it!

bigtimebuddy (approved):

Issue #8566, opened 2022-08-09
Migration Loader to Assets ( feedbacks & bug )


Makio64:
Hello,

I saw the warning about Loader being deprecated in favor of Assets so I switched.

Assets is really nice but I got confuse by fews points : 

- Assets is not part of 'pixi.js' and I have to install extra repo '@pixi/assets' + '@pixi/basis' ( I'm wondering if I should used from 'pixi.js' at all actually.. )
- Assets.add / load works a bit differently from Loader.add / load and it tooks me fews minuts to figure it out
- There is a bug out of the box : if I load '/bunny.png' from 'https://mysite/page1/' it try to load 'https://mysite/page1/bunny.png' when I expected 'https://mysite/bunny.png', It was the previous behavior with Loader. So we have to manually do something like : `Assets.resolver.basePath = window.location.origin+'/'`

Maybe a Migration guide would be nice ?

For the rest, I'm very happy with this new Assets API, very clean, congrats! 

### Environment

- **`pixi.js` version**: 6.5.1

bigtimebuddy:
I think the root-relative pathing issue is a bug. @Zyie what do you think?

Zyie:
> I think the root-relative pathing issue is a bug. @Zyie what do you think?

Yeah i don't think root relative pathing was ever accounted for

Issue #8564, opened 2022-08-08
@pixi/node - Assets doesn't support local Windows paths


Slynchy:
### Expected Behavior
Calling `Assets.load(path)` as detailed in the README file should load files as expected in the Node environment.

### Current Behavior
`Loader.load` fails to find a file that definitely exists, with the following error:
```
C:\Users\samly\Documents\Code\fake-foldername>ts-node test/brokenLoader.ts
PixiJS 6.5.1 - WebGL 1 - http://www.pixijs.com/
C:\Circle.png exists: true
C:\Users\samly\Documents\Code\fake-foldername\node_modules\@pixi\assets\src\loader\Loader.ts:147
                    throw new Error(`[Loader.load] Failed to load ${url}.\n${e}`);
                          ^
Error: [Loader.load] Failed to load /C:\Circle.png.
File not found: \C:\Circle.png
    at Loader.<anonymous> (C:\Users\samly\Documents\Code\fake-foldername\node_modules\@pixi\assets\src\loader\Loader.ts:147:27)
    at step (C:\Users\samly\Documents\Code\fake-foldername\node_modules\node_modules\tslib\tslib.es6.js:100:23)
    at Object.throw (C:\Users\samly\Documents\Code\fake-foldername\node_modules\node_modules\tslib\tslib.es6.js:81:53)
    at rejected (C:\Users\samly\Documents\Code\fake-foldername\node_modules\node_modules\tslib\tslib.es6.js:72:65)
```

### Steps to Reproduce
```
import { Application, Assets, Resource, Texture } from "@pixi/node";
import fs from "fs";

async function main() {
    await Assets.init();
    const app = new Application();
    const debugPath = `C:\\Circle.png`;

    console.log(`${
        debugPath
    } exists: ${
        fs.existsSync(debugPath) // true
    }`);

    const temp = await Assets.load(
        debugPath
    ) as Texture<Resource>; // crashes
}

main();
```

### Environment

- **`node` version**: v16.16.0
- **`npm` version**: 8.5.5
- **`pixi.js` version**: 6.5.1
- **Browser & Version**: n/a
- **OS & Version**: Windows 10

bigtimebuddy:
I'd assume this is probably poor support for Windows pathing. Thanks for reporting.

Slynchy:
A temporary workaround:

```
const PIXI = require("@pixi/node");
const crossFetch = require("cross-fetch");

PIXI.NodeAdapter.fetch = (url, options) => {
[...]

PR #8583, opened 2022-08-12
Fix window paths on Node and add root relative pathing support

Zyie:
This PR fixes 2 issues:

- #8566
- #8564

For root-relative paths I added a new options to the adapter called `getRootPath` as these will be different for different adapters.
For windows paths I changed the checks being made for check if a path is absolute or not.

Previous Behaviour:

- Windows Node: 
  - `basePath` = `/`
  - `rootPath` = N/A
- Unix Node: 
  - `basePath` = `/`
  - `rootPath` = N/A
- Browser: 
  - `basePath` = `document.baseURI || window.location.href`
  - `rootPath` = N/A

New Behaviour:

- Windows Node: 
  - `basePath` = `process.cwd()` e.g. `C:\\Users\\foo\\bar`
  - `rootPath` = `C:\\`
- Unix Node: 
  - `basePath` = `process.cwd()` e.g. `/Users/foo/bar`
  - `rootPath` = `''`
- Browser: 
  - `basePath` = `document.baseURI || window.location.href`
  - `rootPath` = `window.location.origin`

This change to `basePath` in node would be a breaking change but makes more sense now that `rootPath` is being used.

I do think this is a band-aid fix as I think that all path code should be split out of assets but that is a bigger change that could be done in a separate PR?

### DIFF
--- a/bundles/pixi.js-node/src/adapter/adapter.ts
+++ b/bundles/pixi.js-node/src/adapter/adapter.ts
@@ -6,7 +6,6 @@ import { NodeCanvasElement } from './NodeCanvasElement';
 
 import fs from 'fs';
 import path from 'path';
-import { isAbsoluteUrl } from '@pixi/assets';
 
 export const NodeAdapter = {
     /**
@@ -20,8 +19,9 @@ export const NodeAdapter = {
     getWebGLRenderingContext: () => gl as unknown as typeof WebGLRenderingContext,
     /** Returns the fake user agent string of `node` */
     getNavigator: () => ({ userAgent: 'node' }),
-    /** Returns an empty base url */
-    getBaseUrl: () => '',
+    /** Returns the path from which the process is being run */
+    getBaseUrl: () => process.cwd(),
+    /** Returns the root of the path from which the process is being run */
     getRootUrl: () =>
     {
         if (process.platform === 'win32')
@@ -35,7 +35,8 @@ export const NodeAdapter = {
     {
         const request = new Request(url, options);
 
-        if (isAbsoluteUrl(request.url))
+        // check if urls starts with http(s) as only these are supported by node-fetch
+        if ((/^http(s)*:/).test(request.url))
         {
             return fetch(url, request);
         }
--- a/bundles/pixi.js-node/src/adapter/adapter.ts
+++ b/bundles/pixi.js-node/src/adapter/adapter.ts
@@ -6,6 +6,7 @@ import { NodeCanvasElement } from './NodeCanvasElement';
 
 import fs from 'fs';
 import path from 'path';
+import { isAbsoluteUrl } from '@pixi/assets';
 
 export const NodeAdapter = {
     /**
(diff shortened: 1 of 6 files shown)


bundles/pixi.js-node/src/adapter/adapter.ts
```diff
@@ -19,45 +19,50 @@ export const NodeAdapter = {
...
+        return '';
+    },
     fetch: (url: RequestInfo, options?: RequestInit) =>
     {
         const request = new Request(url, options);
 
-        // if url is not absolute path then use fs to attempt to read the url
-        try
+        // check if urls starts with http(s) as only these are supported by node-fetch
+        if ((/^http(s)*:/).test(request.url))
```
bigtimebuddy, line 39: RegExp could be a little cleaner.
```suggestion
        if ((/^https?:/).test(request.url))
```

packages/assets/src/utils/url/makeAbsoluteUrl.ts
```diff
@@ -18,15 +20,25 @@ let baseUrl: string;
  * If it's not, we convert it
  * @param url - The URL to test
  * @param customBaseUrl - The base URL to use
+ * @param customRootUrl - The root URL to use
  */
-export function makeAbsoluteUrl(url: string, customBaseUrl?: string): string
+export function makeAbsoluteUrl(url: string, customBaseUrl?: string, customRootUrl?: string): string
```
bigtimebuddy, line 25: This definitely needs some unit-tests to handle the different permutations of platforms and requests.

packages/assets/src/utils/url/makeAbsoluteUrl.ts
```diff
@@ -2,14 +2,16 @@ import { settings } from '@pixi/settings';
 import { isAbsoluteUrl } from './isAbsoluteUrl';
 import { urlJoin } from './urlJoin';
 
-export function getBaseUrl(url: string): string
+export function removeUrlParams(url: string): string
```
bigtimebuddy, line 5: Also, some unit-tests here would be helpful.

bundles/pixi.js-node/src/adapter/adapter.ts
```diff
@@ -19,45 +19,50 @@ export const NodeAdapter = {
...
+        return '';
+    },
     fetch: (url: RequestInfo, options?: RequestInit) =>
     {
         const request = new Request(url, options);
 
-        // if url is not absolute path then use fs to attempt to read the url
-        try
+        // check if urls starts with http(s) as only these are supported by node-fetch
+        if ((/^https?:/).test(request.url))
```
bigtimebuddy, line 39: What about data-URIs?

packages/assets/src/utils/url/isAbsoluteUrl.ts
```diff
@@ -1,18 +1,23 @@
...
-    // Don't match Windows paths `c:\`
-    if ((/^[a-zA-Z]:\\/).test(url))
+    // If running in node then an absolute URL will start will either a `/` on unix or disk drive on windows
+    if (settings.ADAPTER.getNavigator().userAgent === 'node')
     {
-        return false;
+        if ((/^[a-zA-Z][a-zA-Z\d+\-.]*:/).test(url)) return true;
+        if (process.platform === 'win32') return (/^[a-zA-Z]:(\\|\/)/).test(url);
+
+        return url.startsWith('/');
```
bigtimebuddy, line 20: This seems like not the right place for this code. Part of the rationale of having an adapter is that is can externalize platform specifics (i.e., we should not have `process.platform` references in assets).

 An alternative approach:

```js
return settings.ADAPTER?.isAbsoluteUrl(url) ?? (/^[a-zA-Z][a-zA-Z\d+\-.]*:/).test(url);
```

Zyie:
closing in favor of #8605

PR #8584, opened 2022-08-13
Handle URL protocols in asset path join function of loadBitmapFont parser

### LINKED ISSUE #8571: Assets loader crafting invalid URLs for BitmapFont page images on Safari 14
### Expected Behavior
Assets loader should load BitmapFont images using valid URLs as it does on other browsers (Chrome, latest Safari, etc).

Given a pixi.js app running on https://example.com, with Assets initialized using...
```typescript
await Assets.init({ basePath: 'https://assets.example.com/' });
```
...it should load BitmapFont image files with URLs like:

`(basePath)/(image path)`
`https://assets.example.com/font-img.png`

### Current Behavior
On Mobile Safari on iOS 14 / Safari on macOS Big Sur (and probably older), dev tools report that BitmapFont page images are loaded with URLs like this:

`(origin)/(basePath minus protocol)/(image path)`
`https://example.com/assets.example.com/font-img.png`

And not being the correct path, fail to load.

### Possible Solution
I tried looking through the many path transforming parts of the relevant code but could not find any parts that isolated performed differently on the different browser versions. Though there must be something that works differently on this older Safari version; I'm guessing the "https:/" is getting stripped somehow, and treated as a relative URL. Perhaps a regex expression is getting evaluated differently?

### Steps to Reproduce
```typescript
await Assets.init({ basePath: 'https://assets.example.com/' });
```
Load a BitmapFont with the Asset loader on Safari on iOS 14, notice it fails due to incorrectly crafted URLs. Happening to me when loading bundle-referenced fonts, but I think it happens for non-bundled fonts too.

### Environment

- **`pixi.js` version**: 6.5.1
- **Browser & Version**: Safari 14
- **OS & Version**: macOS Big Sur / iOS 14 (& Older)

qwertywelch:
##### Description of change

Fixes #8571 

The Assets parser for BitmapFonts loads the font page textures with URLs created by joining the parent directory of the font xml/fnt file's source URL with the page image filename.
https://github.com/pixijs/pixijs/blob/fdbdc45b6a95bd47145e3a7267fe2a69a1be4ebb/packages/assets/src/loader/parsers/loadBitmapFont.ts#L38-L46
The [join](https://github.com/pixijs/pixijs/blob/fdbdc45b6a95bd47145e3a7267fe2a69a1be4ebb/packages/assets/src/utils/path.ts#L48) function used did not account for protocols in the path, leading to malformed URLs like `https:/example.com/myfont.png`. These URLs are treated as valid in Chrome, however in older versions of Safari (tested on iOS 14.5), loading such a URL actually tries to load `https://example.com/example.com/myfont.png`.

In this fix, join now extracts the protocol (if present in the first part of the passed path), not passing it to the normalize function which would mangle it. I also fixed the join function's docblock to reflect what it does.

I don't think it's possible to add a new test for this as the differing behavior is platform-specific.

##### Pre-Merge Checklist

### DIFF
--- a/packages/assets/src/utils/path.ts
+++ b/packages/assets/src/utils/path.ts
@@ -41,13 +41,14 @@ export function extname(path: string): string
 }
 
 /**
- * Returns the extension of the path, from the last occurrence of the . (period) character to the end of
- * string in the last portion of the path.
+ * Joins together parts of a URL or file path.
  * @param parts - The path parts to join
  */
 export function join(...parts: string[]): string
 {
     let path = '';
+    let protocol = '';
+    const protocolDelim = '://';
 
     for (let i = 0; i < parts.length; i++)
     {
@@ -57,6 +58,22 @@ export function join(...parts: string[]): string
         {
             if (!path)
             {
+                if (i === 0)
+                {
+                    // if the first part of the path has a URL protocol, extract it so it's not passed
+                    // to the normalize function
+                    const protocolIndex = segment.indexOf(protocolDelim);
+
+                    if (protocolIndex !== -1)
+                    {
+                        const protocolEnd = protocolIndex + protocolDelim.length;
+
+                        protocol = segment.substring(0, protocolEnd);
+                        path += segment.substring(protocolEnd);
+
+                        continue;
+                    }
+                }
                 path += segment;
             }
             else
@@ -66,7 +83,7 @@ export function join(...parts: string[]): string
         }
     }
 
-    return normalize(path);
+    return protocol + normalize(path);
 }
 
 /**


Zyie (commented):
This seems to work pretty good

Could you add some basic tests for these. Maybe something like this:

```ts
it('should manipulate paths correctly', () =>
{
    expect(basename('http://example.com/page-1/sprite.png')).toEqual('sprite.png');
    // TODO: this does not work
    // expect(basename('http://example.com/page-1/sprite.png', '.png')).toEqual('sprite');
    
    expect(extname('http://example.com/page-1/sprite.png')).toEqual('.png');
    
    expect(dirname('http://example.com/page-1/sprite.png')).toEqual('http://example.com/page-1');
    
    expect(join('http://example.com/page-1/', 'sprite.png')).toEqual('http://example.com/page-1/sprite.png');
    expect(join('https://example.com/page-1/', 'sprite.png')).toEqual('https://example.com/page-1/sprite.png');
    expect(join('http://example.com/page-1/', '../sprite.png')).toEqual('http://example.com/sprite.png');
    expect(join('file:///example.com/page-1/', 'sprite.png')).toEqual('file:///example.com/page-1/sprite.png');
});
```

bigtimebuddy:
Closing, as this has been superseded by #8605.

PR #8585, opened 2022-08-13
Fix sync of uniform arrays of size 1

dev7355608:
##### Description of change

Uniform arrays of size 1 have been uploaded as if they weren't arrays. For some reason it actually uploads the value correctly (at least for me) even if you nest the value in arbitrary many arrays, for example: https://pixiplayground.com/#/edit/z1WGvDZvLccs-YCgS9G73. But of course the check against the cached value doesn't work. 

So with this change uniform arrays of size 1 are uploaded via the single setters, but the value in the array is unpacked.

Test: The bunny should disappear after 1 second because the alpha is set to 0.
- Before: https://www.pixiplayground.com/#/edit/Js-DXHFKdn0oGCGRc0hwB
- After: https://www.pixiplayground.com/#/edit/4NU7sDjIS7kywGNzWNlTd

##### Pre-Merge Checklist


GoodBoyDigital:
you think it might be best to get arrays of length one to work like arrays? Would lead to simpler code and less checks in the generated code?

dev7355608:
I chose the upload via single getter to take advantage of caching. But arrays of size 1 are probably really rare and it doesn't really matter. If you prefer, we can upload the arrays the same way regardless of size.

GoodBoyDigital:
yea, if we follow the uploading arrays as arrays regardless of size. Keeps things nice and consitent!

GoodBoyDigital (approved):
nice work @dev7355608! 👍

PR #8605, opened 2022-08-19
Add cross compatible path utils

Zyie:
This PR attempts to tackle the problem of trying to supports paths for both node and browsers and fixes some bugs in the assets package.

## Path Module
I've created our own `path` that is based on nodes `path` module which should reduce the complexity of the codebase and make it easier to maintain.

Differences between nodes path and our path:
- All paths are normalized to use `/` as the separator. e.g. `path.join('C:\\foo', './bar')` will be normalized to `C:/foo/bar`.
- URLs and data URLs are treated as absolute paths.
    ```ts
    path.isAbsolute('https://example.com/') // true`
    path.isAbsolute('data:image/png;base64,abc') // true
    path.isAbsolute('file:///example') // true
    path.isAbsolute('/example') // true
    path.isAbsolute('C:/example') // true
    ```
- The root of a URL is the considered to be the domain of the URL
    ```ts
    path.rootname('https://example.com/foo/bar') // 'https://example.com'
    ```
    - This allows us to join paths like so
    ```ts
    path.join('https://example.com/foo/bar', '../../../../baz') // 'https://example.com/baz'
    ```

Additional Helpers:

- `path.toPosix` converts a path to use `/` as the separator.
- `path.isUrl` determines if a path is a URL.
- `path.isDataUrl` determines if a path is a data URL.
- `path.toAbsolute` converts a path to an absolute path.
- `path.rootname` returns the root of the path. e.g. `http://example.com`, `file:///`, `/`, `C:/`
- `path.hasProtocol` determines if a path has a protocol. e.g. `http://`, `file://`, `C:/`
- `path.getProtocol` returns the protocol of a path. e.g. `http://`, `file://`, `C:/`

## Other Fixes
This PR also fixes the following issues:
- [Assets doesn't support root relative URLs](#8566)
- [Assets doesn't support local Windows paths](#8564)
- [Invalid URLs for BitmapFont](#8571)
- Old `basename` function didn't take into account the extension name. Leading to fonts have names such as `Outfit.woff` instead of `Outfit`

This PR should replace #8584 as it incorporates the changes into the `normalize` function of the `path` module.

### Breaking Changes
- The new default `basePath` for node is `process.cwd()`


packages/assets/test/loader.tests.ts
```diff
@@ -361,7 +361,7 @@ describe('Loader', () =>
 
         document.fonts.forEach((f: FontFace) =>
         {
-            if (f.family === 'Outfit.woff2')
+            if (f.family === 'Outfit')
```
Zyie, line 364: This shows the fix for the `basename` issue

bigtimebuddy:
In general, this approach is a much better way to normalize all the pathing cross-platform. My only concern is the maintenance cost of a forked path module. Is there not a possibility to say override/extend rather than fork?

Zyie:
> In general, this approach is a much better way to normalize all the pathing cross-platform. My only concern is the maintenance cost of a forked path module. Is there not a possibility to say override/extend rather than fork?

I'm not sure how we would go about extending/overriding?

I think that the module is pretty small so I don't think the maintenance would be a big deal. 
Also nodes path module has been stable for a while, so i foresee most of the issues being related to the changes made to support URL paths

bigtimebuddy (approved):

PR #8615, opened 2022-08-25
Fixing CanvasSpriteRenderer to correctly render sprites with rotated base textures

smlmyck:
##### Description of change

During an update to the PIXI.js versions in use where I work, we observed newly broken canvas renderer rendering for textures where the `BaseTexture` is rotated - for example, if some textures were atlassed in a spritesheet and the texture packer decided to rotate a texture for whatever reason.

For example, this is screenshot from a canvas renderer running v6.5.2, where the red check pattern is taller than it is wide, the blue pattern is wider than it is tall, but they're [in a texture atlas where the blue pattern has been rotated](https://cdn.glitch.global/b31e8b57-49fc-4abc-8e9a-011e8602fd1f/canvas-atlas.png?v=1661423801783).

image

Prior to #8408 (v.6.5.0), this was not an issue.

I've prepared a demo website showing the issue in detail over here: https://living-checker-skull.glitch.me/

The `destWidth`/`destHeight` swapping in the `if (texture.rotate)` appears to be unneeded, as the `width`/`height` from the `frame` is still accurate. Removing the swap (as this PR does) fixes the rendering back to the pre-#8408 behaviour for this case.

The added unit test pass in a cherry picked 6.4.2 (with the `toEquals` replaced with `to.equals` anyhow), and fail against v6.5.2, but pass with the code change applied.

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas-sprite/src/CanvasSpriteRenderer.ts
+++ b/packages/canvas-sprite/src/CanvasSpriteRenderer.ts
@@ -119,11 +119,6 @@ export class CanvasSpriteRenderer
             // the anchor has already been applied above, so lets set it to zero
             dx = 0;
             dy = 0;
-
-            const h = destWidth;
-
-            destWidth = destHeight;
-            destHeight = h;
         }
 
         dx -= destWidth / 2;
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Nice work @smlmyck, thank you.

bigtimebuddy:
@drspoik do you want to take a look at this?

bigtimebuddy:
I'm making a build, should be available in a few minutes here:

https://pixijs.download/pr-8615/pixi.min.js

bigtimebuddy:
Example in #8408 checks out with this: https://www.pixiplayground.com/#/edit/pvmT1YGcfSSb15CrR8WvS

bigtimebuddy:
Thank you @smlmyck

PR #8616, opened 2022-08-25
Split asset parsers into individual packages

Zyie:
This PR splits all the parsers into individual packages e.g text-bitmap, spritesheet, basis, compressed-textures

I also added `@pixi/assets` to the main bundle and deleted `@pixi/loader`


Zyie:
The failing test has just revealed a bug with the bitmap loader. It doesn't parse a split fonts pages in the correct order all the time

Working on a fix for that now

bigtimebuddy:
Even when it did pass, there's an issue with [webdoc](https://github.com/pixijs/pixijs/runs/8023507066?check_suite_focus=true#step:11:843). This seems new.

bigtimebuddy:
Otherwise, I'm so excited for this PR!

Zyie:
> Even when it did pass, there's an issue with [webdoc](https://github.com/pixijs/pixijs/runs/8023507066?check_suite_focus=true#step:11:843). This seems new.

ah, will also look into this as well

Zyie:
@bigtimebuddy all fixed

PR #8622, opened 2022-08-27
Add OffscreenCanvas and Web Worker support

### LINKED ISSUE #7123: Pixi should support OffscreenCanvas Web Worker environments and other "headless" environments.
I'm trying to use Pixi.js in a webworker with the [OffscreenCanvas reference API](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) It fails at runtime with a `window undefined error`. 

Screen Shot 2021-01-02 at 4 47 58 PM

I am aware of the other Github issues that explain how [Pixi does not want to support Node.js](https://github.com/pixijs/pixi.js/issues/3224) ([etc](https://github.com/pixijs/pixi.js/issues/4771)) environments because the Pixi team feels Pixi is a browser only tool and don't want to write isomorphic code but, WebWorker is a browser enviroment. I even tried to use the recommended `pixi-shim` and it failed expecting a `window` object as well (lol):

Screen Shot 2021-01-02 at 6 35 48 PM

`OffscreenCanvas` Web worker environments are a "browser environment" so what are Pixis plan to support this enviroment?

Other WebGL rendering libraries like [Regl](http://regl.party/) support this pattern by [accepting a reference to a WebGL context passed directly to the initialization code](https://github.com/regl-project/regl/blob/gh-pages/API.md#from-a-webgl-context). I am making assumptions here but feels like PIXI could follow that pattern?

SuperSodaSea:
This PR attempts to add OffscreenCanvas and Web Worker support to Pixi.js.

This should fix #7123.

##### Description of change

- New interface [PIXI.ICanvas](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/settings/src/canvas.ts), the common interface for HTMLCanvasElement and OffscreenCanvas
    - Change many usages of HTMLCanvasElement to PIXI.ICanvas, such as PIXI.Renderer.view
- ~~New [PIXI.environment](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/settings/src/utils/environment.ts) for environment detection~~
    - ~~[PIXI.AccessibilityManager](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/accessibility/src/AccessibilityManager.ts) now only added to extensions in browser main thread, not in worker thread~~
- ~~New [PIXI.WebWorkerAdapter](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/settings/src/adapter.ts) for Web Worker~~
    - ~~Default value for [PIXI.settings.ADAPTER](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/settings/src/settings.ts) will be PIXI.WebWorkerAdapter in worker thread~~
- Other small changes to make things work for OffscreenCanvas / Web Worker

##### Todo List

There are still some small things to do (see Todo List), so I made this PR in draft state for everyone's opinion. :wink:

- [ ] Implement base64 for OffscreenCanvas in [PIXI.CanvasExtract](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/canvas-extract/src/CanvasExtract.ts) and [PIXI.Extract](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/extract/src/Extract.ts)
- [ ] Implement convertTintToImage for OffscreenCanvas / Web Worker in [PIXI.canvasUtils](https://github.com/SuperSodaSea/pixijs/blob/web-worker/packages/canvas-renderer/src/canvasUtils.ts)
- [ ] Add tests for OffscreenCanvas / Web Worker

##### Pre-Merge Checklist

### DIFF
--- a/packages/accessibility/src/AccessibilityManager.ts
+++ b/packages/accessibility/src/AccessibilityManager.ts
@@ -1,13 +1,13 @@
+import { ExtensionType, extensions } from '@pixi/core';
 import { DisplayObject } from '@pixi/display';
-import { isMobile, removeItems } from '@pixi/utils';
+import { FederatedEvent } from '@pixi/events';
+import { environment, isMobile, removeItems } from '@pixi/utils';
 import { accessibleTarget } from './accessibleTarget';
 
-import type { Rectangle } from '@pixi/math';
-import type { Container } from '@pixi/display';
-import { ExtensionType, extensions } from '@pixi/core';
-import type { IAccessibleHTMLElement } from './accessibleTarget';
 import type { IRenderer, ExtensionMetadata } from '@pixi/core';
-import { FederatedEvent } from '@pixi/events';
+import type { Container } from '@pixi/display';
+import type { Rectangle } from '@pixi/math';
+import type { IAccessibleHTMLElement } from './accessibleTarget';
 
 // add some extra variables to the container..
 DisplayObject.mixin(accessibleTarget);
@@ -87,6 +87,15 @@ export class AccessibilityManager
      */
     constructor(renderer: IRenderer)
     {
+        if (!environment.browser)
+        {
+            throw new Error('AccessibilityManager is only supported in browser main thread');
+        }
+        if (typeof HTMLCanvasElement === 'undefined' || !(renderer.view instanceof HTMLCanvasElement))
+        {
+            throw new Error('AccessibilityManager is only supported on HTMLCanvasElement');
+        }
+
         this._hookDiv = null;
 
         if (isMobile.tablet || isMobile.phone)
@@ -292,7 +301,7 @@ export class AccessibilityManager
             this.updateAccessibleObjects(this.renderer.lastObjectRendered as Container);
         }
 
-        const { left, top, width, height } = this.renderer.view.getBoundingClientRect();
+        const { left, top, width, height } = (this.renderer.view as HTMLCanvasElement).getBoundingClientRect();
         const { width: viewWidth, height: viewHeight, resolution } = this.renderer;
 
         const sx = (width / viewWidth) * resolution;
@@ -596,4 +605,7 @@ export class AccessibilityManager
     }
 }
 
-extensions.add(AccessibilityManager);
+if (environment.browser)
+{
+    extensions.add(AccessibilityManager);
+}
--- a/packages/accessibility/src/AccessibilityManager.ts
+++ b/packages/accessibility/src/AccessibilityManager.ts
@@ -91,7 +91,7 @@ export class AccessibilityManager
         {
             throw new Error('AccessibilityManager is only supported in browser main thread');
         }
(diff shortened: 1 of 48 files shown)


bigtimebuddy:
Thank you for this @SuperSodaSea.

I know that @Zyie was already in-progress on the Webworker thing and would like for him to weigh-in on this. 

There are two issues that would be good to address separately: 1) General support for OffscreenCanvas, and 2) Support for running in Webworker threads. Tackling this as two issues would be better even though 2 requires 1. You've done a great job here, but it's a big PR and more difficult to review.

OffscreenCanvas should probably be done in a way you've handled here: create some type interface like ICanvas, and fixing the documentation. We are already opportunistically using OffscreenCanvas (e.g., TextMetrics) if it's available and get a good performance boost from that. Having broader support for this approach might have similar gains (thinking of BitmapFont and other places that create temporary canvases).

As for the webworker, I actually don't think this is the right approach (`environment` in settings, casing when installing AccessibilityManager, multiple adapters in settings). Our approach for Node has been to provide a separate bundle for this environment. The benefit of this is that we can exclude packages which are not relevant like accessibility or interactions/events. Composing is a simpler approach, (e.g., add `@pixi/webworker` bundle (minus events and accessibility) and maybe we separate the adapter into a new package `@pixi/adapter-webworker` (and `@pixi/adapater-browser`, `@pixi/adapter-node`). My preference is to compose another bundle, rather than trying to make the code runs everywhere.

SuperSodaSea:
> Thank you for this @SuperSodaSea.
> 
> I know that @Zyie was already in-progress on the Webworker thing and would like for him to weigh-in on this.
> 
> There are two issues that would be good to address separately: 1) General support for OffscreenCanvas, and 2) Support for running in Webworker threads. Tackling this as two issues would be better even though 2 requires 1. You've done a great job here, but it's a big PR and more difficult to review.
> 
> OffscreenCanvas should probably be done in a way you've handled here: create some type interface like ICanvas, and fixing the documentation. We are already opportunistically using OffscreenCanvas (e.g., TextMetrics) if it's available and get a good performance boost from that. Having broader support for this approach might have similar gains (thinking of BitmapFont and other places that create temporary canvases).
> 
> As for the webworker, I actually don't think this is the right approach (`environment` in settings, casing when installing AccessibilityManager, multiple adapters in settings). Our approach for Node has been to provide a separate bundle for this environment. The benefit of this is that we can exclude packages which are not relevant like accessibility or interactions/events. Composing is a simpler approach, (e.g., add `@pixi/webworker` bundle (minus events and accessibility) and maybe we separate the adapter into a new package `@pixi/adapter-webworker` (and `@pixi/adapater-browser`, `@pixi/adapter-node`). My preference is to compose another bundle, rather than trying to make code that runs everywhere.

I agree with your opinion. I'll try to split this PR into two PRs: one for OffscreenCanvas / ICanvas, another for Web Workers.

SuperSodaSea:
Moved to new PR #8628.

PR #8628, opened 2022-09-04
Introduce interface ICanvas and ICanvasRendererContext2D

SuperSodaSea:
This PR introduce interface ICanvas and ICanvasRendererContext2D, attempts to support OffscreenCanvas in Pixi.js.

Related PR: #7123

##### Description of change

- New interface [PIXI.ICanvas](https://github.com/SuperSodaSea/pixijs/blob/icanvas/packages/settings/src/ICanvas.ts) in `@pixi/settings`, the common interface for HTMLCanvasElement, OffscreenCanvas, and other custom canvas classes.
- New interface [PIXI.ICanvasRenderingContext2D](https://github.com/SuperSodaSea/pixijs/blob/icanvas/packages/settings/src/ICanvasRenderingContext2D.ts) in `@pixi/settings`, the common interface for CanvasRenderingContext2D, OffscreenCanvasRenderingContext2D, and other custom canvas 2D context.
- Add tests for OffscreenCanvas in [Application.tests.ts](https://github.com/SuperSodaSea/pixijs/blob/icanvas/packages/app/test/Application.tests.ts) and  [Renderer.tests.ts](https://github.com/SuperSodaSea/pixijs/blob/icanvas/packages/core/test/Renderer.tests.ts)

##### Breaking Changes

- Most of the HTMLCanvasElement types are replaced with PIXI.ICanvas, such as `PIXI.Renderer.view`
- Most of the CanvasRenderingContext2D types are replaced with PIXI.ICanvasRenderingContext2D

##### Todo List

- [ ] Still need to manually call `PIXI.extensions.remove(PIXI.AccessibilityManager)` to use OffscreenCanvas as `Renderer.view` in browser
- [ ] OffscreenCanvas don't have toDataURL, only have convertToBlob and transferToImageBitmap
    - [ ] Implement base64 for OffscreenCanvas in [PIXI.CanvasExtract](https://github.com/SuperSodaSea/pixijs/blob/icanvas/packages/canvas-extract/src/CanvasExtract.ts) and [PIXI.Extract](https://github.com/SuperSodaSea/pixijs/blob/icanvas/packages/extract/src/Extract.ts)
    - [ ] Implement convertTintToImage for OffscreenCanvas in [PIXI.canvasUtils](https://github.com/SuperSodaSea/pixijs/blob/icanvas/packages/canvas-renderer/src/canvasUtils.ts)

##### Pre-Merge Checklist


SuperSodaSea:
@bigtimebuddy @Zyie
Having any plan to review this PR?

bigtimebuddy:
@SuperSodaSea we haven't lost sight of this PR. @Zyie and I chatted about it offline and feel like it's definitely worth getting in. I would like to get it in v7 before our final release. It's a lot to review, so I hope to get to it soon. Are you happy with it on your end?

SuperSodaSea:
> @SuperSodaSea we haven't lost sight of this PR. @Zyie and I chatted about it offline and feel like it's definitely worth getting in. I would like to get it in v7 before our final release. It's a lot to review, so I hope to get to it soon. Are you happy with it on your end?

I will wait patiently ;)

By the way, the force pushes was just to keep up with the dev branch, no major changes were made.

bigtimebuddy (commented):
I had a chance to review this PR. 

Overall, I think generalized interfaces are great, especially to help normalize the gap between Node.js and the DOM. These interfaces can clearly define the API surface area between this library and other environments in a way that is very explicit. 

That said, one effect of the implementation here is a lot of additional casting.  If it requires extra casting, it's not really normalizing, it's creating a stricter interface than cannot actually be used internally.

Examples:

```ts
(this.renderer.view as HTMLCanvasElement)

textureSource as CanvasImageSource

cachedCanvas as GlobalMixins.GlobalTintable

(this.renderer.view as any)

(view as any)

(canvas as ImageSource)

this._tintedCanvas = canvasUtils.getTintedCanvas(this, this.tint) as ICanvas;
tempCanvas.context.drawImage(this._tintedCanvas as CanvasImageSource, 0, 0);
```

Adding more optional definitions to the ICanvas might help, for instance `parentNode`. Also, using TypeScript generics to specify stricter return types can be helpful too, for instance: `settings.ADAPTER.createCanvas<HTMLCanvasElement>()`.

I would like to see an approach here which simplifies things internally, not add more complexity. Someone shouldn't read the code and ask, "Why is this casting necessary?". The goal isn't to make TypeScript not complain, it's about simplifying the code for ourselves.

SuperSodaSea:
@bigtimebuddy That make sense, I'll try to simplify the casting in the following days.

SuperSodaSea:
@bigtimebuddy

I've removed most of the redundant casting, for examples: 

- No more `... as any` in the commited code.
- Casting in `(... as any).addEventListener` and `.removeEventListener` is removed by adding `addEventListener` and `removeEventListener` in `ICanvas`.

https://github.com/pixijs/pixijs/blob/4f16e60f3dd681999e37a4663af705ecb616da56/packages/settings/src/ICanvas.ts#L51-L79

- Casting in `... as CanvasImageSource` is removed by override `createPattern` and `drawImage` that takes `ICanvas` as argument in `ICanvasRenderingContext2D`.

https://github.com/pixijs/pixijs/blob/4f16e60f3dd681999e37a4663af705ecb616da56/packages/settings/src/ICanvasRenderingContext2D.ts#L24-L29

- Casting in `... as ImageSource` is removed by adding `ICanvas` to `ImageSource`.

https://github.com/pixijs/pixijs/blob/4f16e60f3dd681999e37a4663af705ecb616da56/packages/core/src/textures/BaseTexture.ts#L19

- Casting in  `... as GlobalMixins.GlobalTintable` is removed by adding `interface GlobalMixins.ICanvas extends GlobalTintable` and `interface ICanvas extends GlobalMixins.ICanvas`.

https://github.com/pixijs/pixijs/blob/4f16e60f3dd681999e37a4663af705ecb616da56/packages/canvas-renderer/global.d.ts#L24-L28
https://github.com/pixijs/pixijs/blob/4f16e60f3dd681999e37a4663af705ecb616da56/packages/settings/src/ICanvas.ts#L20

- Casting in `_tintedCanvas = canvasUtils.getTintedCanvas(...) as ICanvas` is removed by changing type of `_tintedCanvas` from `ICanvas` to `ICanvas | HTMLImageElement` (matches the return type of `canvasUtils.getTintedCanvas`)

https://github.com/pixijs/pixijs/blob/4f16e60f3dd681999e37a4663af705ecb616da56/packages/canvas-mesh/global.d.ts#L9

---

However, for the remaining casting in `(... as HTMLCanvasElement).parentNode`, `.style`, `.getBoundingClientRect(...)` and others, since they only exist in `HTMLCanvasElement`, not in `OffscreenCanvas` or `NodeCanvasElement`, I think these casts are necessary. What about your opinion?

bigtimebuddy:
Thanks for figuring out all these bits. Already the code is looking much better! 

As for the HTMLCanvasElement, here's my thinking, but open to hearing other thoughts!

As PixiJS has grown in complexity and the code base has gotten larger, it's harder to know what the surface area looks like between various environments (DOM). It can actually be _very_ useful to know explicitly what the contract between PixiJS and the DOM looks like. For instance, maybe you want to create a mock canvas for testing. Or maybe you want to extend your own canvas. How do you _know_ what to implement?

### Option 1: Stricter Cast

Creating a much stricter HTMLCanvasElement utility type that's casted internally.... i.e., `IHTMLCanvas`. Then we replace our current HTMLCanvasElement castings with this new type. The benefit is that we can be more rigid about defining how this is actually used by PixiJS:

```ts
interface IHTMLCanvas extends ICanvas {
  readonly style: {
    width: string;
    height: string;
    msContentZooming: 'none';
    msTouchAction: 'none';
    touchAction: 'none';
    cursor: string;
  },
  readonly parentNode?: {
    appendChild(element: HTMLDivElement): void;
    removeChild(canvas: IHTMLCanvasElement): void;
  },
  getBoundingClientRect(): {
    left: number,
    top: number,
    width: number,
    height: number
  },
}
```

I don't think this is too much overhead to maintain and it's a nice helper type in cases where we _only_ care about the HTMLCanvasElement implementation (accessibility and events are two places that come to mind).

### Option 2: More Generic ICanvas

The only place that makes me re-think Option 1 is ViewSystem. There are still a handful of casts here when this should still be pretty generic code. This system is rather core for managing the canvas, resolution, etc. That makes me think `style.width`, `style.height`,  and `parentNode.removeChild` should be part of ICanvas. 

You've already added `addEventListener` and `removeEventListener` as optional methods, I don't think it's really anymore messy to add the HTMLCanvasElement bits too.

----

I think I'm probably leaning toward Option 2 so that we can just do ICanvas everywhere. 

WDYT?

packages/extract/src/Extract.ts
```diff
@@ -75,7 +76,13 @@ export class Extract implements ISystem
...
     public base64(target: DisplayObject | RenderTexture, format?: string, quality?: number): string
     {
-        return this.canvas(target).toDataURL(format, quality);
+        const canvas = this.canvas(target);
+
+        if (canvas.toDataURL !== undefined)
+        {
+            return canvas.toDataURL(format, quality);
+        }
+        throw new Error('TODO: base64 for OffscreenCanvas');
```
bigtimebuddy, line 85: Why not fallback to something like this?
```js
return canvas.convertToBlob().then(blob => new FileReaderSync().readAsDataURL(blob));
```
Since this is for v7, changing the return from `string` to `Promise<string>` is okay with me.

SuperSodaSea: I think we should use `URL.createObjectURL(blob)` here, since `FileReaderSync` is only available in worker thread. Using `OffscreenCanvas` in main thread is possible.

SuperSodaSea: `FileReader.readAsDataURL` instead of `FileReaderSync.readAsDataURL`, I think.

bigtimebuddy: Yeah, my bad, FileReader.

SuperSodaSea:
@bigtimebuddy 

I prefer Option 2, too. See 1a940982ca83b14d581695eb2e35fafd2d637a3d:

https://github.com/pixijs/pixijs/blob/1a940982ca83b14d581695eb2e35fafd2d637a3d/packages/settings/src/ICanvas.ts#L89-L115

packages/extract/src/Extract.ts
SuperSodaSea: Done in de690ddc42a109ec635e59a826f41c1958809a34:
https://github.com/pixijs/pixijs/blob/de690ddc42a109ec635e59a826f41c1958809a34/packages/extract/src/Extract.ts#L77-L99

SuperSodaSea:
@bigtimebuddy

While changing `Extract.base64()` to an async method is simple, the `canvasUtils.getTintedCanvas` seems to be more tricky, since this method is used in synchronous rendering functions. What about your opinion?

https://github.com/pixijs/pixijs/blob/de690ddc42a109ec635e59a826f41c1958809a34/packages/canvas-renderer/src/canvasUtils.ts#L58-L73

bigtimebuddy:
I agree that converting canvasUtils to async is probably way too complicated. However, with the canvasUtils usage, I don't think we should be throwing here. I'd suggest this instead:

```js
if (canvasUtils.convertTintToImage && canvas.toDataURL !== undefined) 
{
  // extract data-uri as image
}
else
{
  // use the canvas directly
}
```

SuperSodaSea:
> I agree that converting canvasUtils to async is probably way too complicated. However, with the canvasUtils usage, I don't think we should be throwing here. I'd suggest this instead:
> 
> ```js
> if (canvasUtils.convertTintToImage && canvas.toDataURL !== undefined) 
> {
>   // extract data-uri as image
> }
> else
> {
>   // use the canvas directly
> }
> ```

Fixed in cf17960e676c97a5d542c430e3bc5439335b8624.

bigtimebuddy (approved):
This is excellent work @SuperSodaSea. Thank you for being so open to receiving and responding to feedback.

bigtimebuddy:
Yes @SuperSodaSea, please go ahead and make the PR for `@pixi/worker`

Issue #8621, opened 2022-08-27
Typings for displayobject/interaction seems broken


jnsmalm:
Something has happened with typings for DisplayObject/Interaction.

I have this in typescript, and it doesn't compile:

```
let object = new Container()
object.interactive = true
```

Getting error: Property 'interactive' does not exist on type 'Container<DisplayObject>'

PixiJS 6.5.1

bigtimebuddy:
Cannot reproduce this. Here's the example:
https://github.com/bigtimebuddy/pixi-typescript-webpack-example

```
npm install
npm run types
```

Please make sure you [upgrade correctly](https://github.com/pixijs/pixijs/wiki/Upgrading-PixiJS).

jnsmalm:
Sorry about that, I was missing "esModuleInterop": true in the tsconfig.

Issue #8545, opened 2022-07-30
Missing DisplayObject properties types?


rachoac:
After upgrading from pixijs version 6.2.2 to the latest (6.5.1 at time of writing), i've noticed several typescript errors relating to the `interactive` and `name` attributes on DisplayObject:

https://pixijs.download/release/docs/PIXI.DisplayObject.html#interactive
https://pixijs.download/release/docs/PIXI.DisplayObject.html#name

The Graphics object extends DisplayObject, but now the type definitions seem to no longer encompass these attributes, leading to errors like this in my codebase:

<img width="620" alt="image" src="https://user-images.githubusercontent.com/334520/181925166-47b412e7-f428-456e-996b-245dcfbae8df.png">

<img width="549" alt="image" src="https://user-images.githubusercontent.com/334520/181925567-05d37418-eb92-4af2-be9c-5fce5500b665.png">

(DyanmicTextContainer extends Graphics)

These attributes do seem to exist however when I inspect the object runtime. Is this a matter of outdated types? Or are these attributes now deprecated and are going away? If so, is there an upgrade path?

bigtimebuddy:
Sounds like #8186 or #7816? This can also happen from an improper upgrade with npm.

rachoac:
It total ways @bigtimebuddy ! thank you

Issue #8186, opened 2022-02-13
any Property 'interactive' does not exist on type 'Sprite'


jmatte123:
I am getting an error in typescript where there is no property Interactive on a Sprite.  Looking at the type definition in my node modules I can't even find the field on the DisplayObject.  Am I missing something here?
![Screen Shot 2022-02-13 at 2 00 18 PM](https://user-images.githubusercontent.com/24372751/153772738-f46deefe-6bd6-475f-ba67-3d6c917c93d9.png)
![Screen Shot 2022-02-13 at 2 02 49 PM](https://user-images.githubusercontent.com/24372751/153772747-40a09e03-5fe8-4721-b18d-02112e7eeac7.png)
```
// tsconfig.json

{
  "compilerOptions": {
    "strict": true,
    "target": "ES6",
    "outDir": "dist",
    // Required for PixiJS
    "esModuleInterop": true,
    "moduleResolution": "node",
    "baseUrl": "./",
    "paths": {
      "mini-signals": [
        "./node_modules/resource-loader/typings/mini-signals.d.ts"
      ]
    }
  },
  "include": [
    "src/**/*.ts",
    "node_modules/@types"
  ],
}
```

```
// package.json

{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "scripts": {
    "test": "jest",
    "start": "parcel src/index.html",
    "build": "parcel build src/index.html --public-url ./",
    "build_serve": "npm run build && http-server ./dist",
  },
  "author": "tester",
  "dependencies": {
    "parcel-bundler": "^1.12.5",
    "pixi.js": "^6.2.2"
  },
  "devDependencies": {
    "http-server": "~0.12.3",
    "typescript": "~4.0.5"
  }
}

```

- **`pixi.js` version** 6.2.2

bigtimebuddy:
Did you recently upgrade pixi.js?

jmatte123:
I just started playing around with it a couple of days ago, I did `npm i pixi.js`

jmatte123:
`"allowSyntheticDefaultImports": true` worked!  Thank you!

gterminator:
problem still persist on 7.2.4

gterminator:
> As you are using angular aside, maybe you could try adding `"allowSyntheticDefaultImports": true` into `angularCompilerOptions` in `tsconfig.json`. The result `tsconfig.json` should like here below.
> 
> ```json
> {
>   // Some settings
>   "compilerOptions": {
[...]

Issue #7816, opened 2021-09-20
TS2339: missing properties 'interactive', 'on' for PIXI.Container


abodelot:
Hi, after upgrading from pixi 5 to pixi 6, I got a typescript error about event-related properties not defined anymore on `PIXI.Container` class.

Minimal code example:
```ts
import * as PIXI from 'pixi.js';

class Foobar extends PIXI.Container {
  constructor() {
    super();
    this.interactive = true;
    this.on('pointertap', () => {
      console.log('hello');
    });
  }
}
```

With `pixi.js 6.1.3`, `tsc` gives the following error:

```
src/Foobar.ts:6:10 - error TS2339: Property 'interactive' does not exist on type 'Foobar'.
6     this.interactive = true;
           ~~~~~~~~~~~
src/Foobar.ts:7:10 - error TS2339: Property 'on' does not exist on type 'Foobar'.
7     this.on('pointertap', () => {
           ~~
```

This seems like a false positive, because those properties work as intended when ignoring `tsc` errors.

They are also listed in documentation: https://pixijs.download/release/docs/PIXI.Container.html#interactive

With `pixi.js 5.2.1`, `tsc` does not give any error with the same code.

- **typescript**: 4.4.3
- **ts-loader**: 9.2.5
- **OS & Version**: Ubuntu 20.04

bigtimebuddy:
Please make sure v5 is _completely_ removed from your package-lock.json or yarn.lock file. Having a mix of v5 and v6 will break types in this way.

bigtimebuddy:
Closing this, as it's likely duplicate of either #7505 or #7685

abodelot:
Thank you @bigtimebuddy for fast reply!

Although unrelated to pixi version upgrade, you provided the fix in linked issues (`esModuleInterop: true` in tsconfig :tada:).

PR #8630, opened 2022-09-06
Abandon the usage of peerDependencies

bigtimebuddy:
### Context

The usage of `peerDependencies` have been a huge problem plaguing PixiJS v6. Different package managers resolve them differently causing lots of double-installed packages, especially when upgrading or using 3rd-party PixiJS plugins. Also, our packages use peers in a somewhat complex graph way making debugging this difficult. Here are some examples how this manifests for users: #8621, #8186, #8545 or #7816. We even had to make this [wiki](https://github.com/pixijs/pixijs/wiki/Upgrading-PixiJS) about how to upgrade.

### Overview

This change abandons `peerDependencies` completely in describing the relationships between packages.

For our bundles (`pixi.js`, `pixi.js-legacy` and `@pixi/node`) this change will not impact users. If anything, it will make doing upgrades much simpler and with fewer headaches.

For those using _individual packages_, it's important that you install the complete set of requirements. For this we will also update https://pixijs.io/customize as well as add `pixiRequirements` field in the **package.json** to be able to script resolution (maybe we can create a simple CLI in core to do this check, e.g., `pixi-doctor`).

The consequence of missing a package, is that build-tools or runtime fail at a missing package. This is _much_ easier for a user to understand. For instance, compare the above linked issues to a TypeScript error like this:

```
error TS2307: Cannot find module '@pixi/graphics' or its corresponding type declarations.
```

The solution here is easy: `yarn add @pixi/graphics` or `npm install @pixi/graphics`. 

This change isn't perfect and I wish we could get peers to work, but this approach is easier for users to self-debug and has fewer mysterious double-install side effects (missing types, missing runtime plugins, etc). Also, it's a change that we can continue to build upon to make the ecosystem work with all package managers more gracefully.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25486,147 +25486,77 @@
     "packages/accessibility": {
       "name": "@pixi/accessibility",
       "version": "6.5.2",
-      "license": "MIT",
-      "peerDependencies": {
-        "@pixi/core": "6.5.2",
-        "@pixi/display": "6.5.2",
-        "@pixi/events": "6.5.2"
-      }
+      "license": "MIT"
     },
     "packages/app": {
       "name": "@pixi/app",
       "version": "6.5.2",
-      "license": "MIT",
-      "peerDependencies": {
-        "@pixi/core": "6.5.2",
-        "@pixi/display": "6.5.2"
-      }
+      "license": "MIT"
     },
     "packages/assets": {
       "name": "@pixi/assets",
       "version": "6.5.2",
-      "license": "MIT",
-      "peerDependencies": {
-        "@pixi/core": "6.5.2"
-      }
+      "license": "MIT"
     },
     "packages/basis": {
       "name": "@pixi/basis",
       "version": "6.5.2",
-      "license": "MIT",
-      "peerDependencies": {
-        "@pixi/assets": "6.5.2",
-        "@pixi/compressed-textures": "6.5.2",
-        "@pixi/core": "6.5.2"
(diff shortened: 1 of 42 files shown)


miltoncandelero:
Does this have anything in place to prevent mixing versions?
The worst case would be mixing a v6 with a v5 package, but you can also have a mismatch between minors or patches 🤔

bigtimebuddy:
We always publish every package for every version, regardless of changes. It is strongly recommended that you use all of the same version and not cherry-pick, since all the same versions are tested together. Again, this isn't an issue for bundles, which has explicit dependencies versions.

This change doesn't _prevent_ someone from mixing and matching. In fact, maybe there's a valid use-case for this if you, say, want a small bug fix in `@pixi/sprite`, but don't want to upgrade everything else. We shouldn't recommend or support this, but I don't think we can rely on package-managers to enforce the current versions correctly. We've seen issues with this when plugins have a fuzzy-match (e.g., `^6.0.0`) against Pixi's internal strict peer version (e.g. `6.4.2`). When it doesn't reconcile peers correctly, it will often instance doubles.

Version correctness is maybe a warning that a CLI checking tool might point out.

bigtimebuddy:
One other thing I'll add, is that using packages falls a little more into the advance, non-casual user territory. Our documentation, for example, is not aligned (yet) around the packages. You have to be a little savvier around not using a bundle in order to save some code size cost.

PR #8639, opened 2022-09-10
Feat(sprited-animated): make currentFrame mutable

### LINKED ISSUE #6098: Please make currentFrame editable
I can't even fathom why it's readonly in the first place.  Would have made things so much easier for me.

YunYouJun:
##### Description of change

Make `sprite-animated` `currentFrame` be mutable

close #6098

##### Pre-Merge Checklist

### DIFF
--- a/packages/sprite-animated/src/AnimatedSprite.ts
+++ b/packages/sprite-animated/src/AnimatedSprite.ts
@@ -406,10 +406,7 @@ export class AnimatedSprite extends Sprite
         this.updateTexture();
     }
 
-    /**
-     * The AnimatedSprites current frame index.
-     * @readonly
-     */
+    /** The AnimatedSprites current frame index. */
     get currentFrame(): number
     {
         let currentFrame = Math.floor(this._currentTime) % this._textures.length;
@@ -420,6 +417,18 @@ export class AnimatedSprite extends Sprite
         }
 
         return currentFrame;
+    }
+
+    set currentFrame(value: number)
+    {
+        const previousFrame = this.currentFrame;
+
+        this._currentTime = value;
+
+        if (previousFrame !== this.currentFrame)
+        {
+            this.updateTexture();
+        }
     }
 
     /**
--- a/packages/sprite-animated/src/AnimatedSprite.ts
+++ b/packages/sprite-animated/src/AnimatedSprite.ts
@@ -176,15 +176,7 @@ export class AnimatedSprite extends Sprite
     public gotoAndStop(frameNumber: number): void
     {
         this.stop();
-
-        const previousFrame = this.currentFrame;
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
This setter is the same thing as gotoAndStop/gotoAndPlay. Is it still necessary? 

Also, the getter and setter don't exactly agree. The getter always returns a normal frame index, the setter will accept anything >= 0. Should we check the bounds of the currentFrame input?

YunYouJun:
> This setter is the same thing as gotoAndStop/gotoAndPlay. Is it still necessary?

When I want to change the frame index, my first reaction is to change `currentFrame`.

So in my opinion, they should be equivalent. Like this:

```ts
animatedSprite.gotoAndStop(10)

// same as

animatedSprite.stop()
animatedSprite.currentFrame = 10
```

```ts
animatedSprite.gotoAndPlay(10)

// same as

animatedSprite.currentFrame = 10
animatedSprite.play()
```

So I think the difference is that when you only set `currentFrame`, you can manually control `stop()` or `play()` from elsewhere.

This change is also fully compatible, so it does not affect any existing actions of the user.

---

> Also, the getter and setter don't exactly agree. The getter always returns a normal frame index, the setter will accept anything >= 0. Should we check the bounds of the currentFrame input?

Here I'm referring to `gotoAndPlay` and `gotoAndStop`.

```ts
    /**
     * Goes to a specific frame and begins playing the AnimatedSprite.
     * @param frameNumber - Frame index to start at.
     */
    public gotoAndPlay(frameNumber: number): void
    {
        const previousFrame = this.currentFrame;

        if (previousFrame !== this.currentFrame)
        {
            this.updateTexture();
        }

        this.play();
    }
```

Looks like they didn't do anything about the bounds either?

If you have any ideas on how to change it, just let me know. I'll be happy to change it.

bigtimebuddy:
Okay I think your rationale makes sense. I also think adding currentFrame would allow users to control their own playback.

I have a few suggestions:

1. Remove the redundant code in gotoAndStop/gotoAndPlay and have them use currentFrame directly.
2. Add a bounds check and throw and error in currentFrame if out of bounds.
3. Add unit tests for put of bounds throwing, making sure getter and setter return the same thing, and anything else you can think of.

bigtimebuddy:
@YunYouJun let me know if there's anything you need. Happy to help get this in.

YunYouJun:
Sorry for just finishing this revision due to other busy work.

I have made some changes as you suggested, is this what you want?
Do I need to change anything else?

bigtimebuddy (approved):
This is great, you nailed it! Thanks.

bigtimebuddy:
Thank you!

PR #8643, opened 2022-09-13
fix node WebGLRenderingContext

michaeljherrmann:
##### Description of change

I'm not sure if this is the approach you want to go with, but I've created this PR to highlight a bug I found with the new NodeAdapter.

The default export from the `gl` library doesn't give us a `WebGLRenderingContext`, instead it gives us a method `createContext`. You can see that [here](https://github.com/stackgl/headless-gl/blob/master/src/javascript/node-index.js#L129).
This means that in the current PIXI code, operations like [this one](https://github.com/michaeljherrmann/pixijs/blob/4975264a6ec65121751cbee2ce2cc77a80d34b9a/packages/core/src/mask/StencilSystem.ts#L28) are broken, because `STENCIL_TEST` is `undefined` (`.getWebGLRenderingContext()` is returning the function `createContext` and not an actual context).

So at first I thought we could just do `createContext()` but the dimension arguments are required (otherwise we get `null`). So I just pass 1x1 to the method. I think this is creating a "bound" context, while normally the `window.WebGLRenderingContext` is "unbound", but I think that might be okay? This is how the [docs](https://github.com/stackgl/headless-gl#example-1) in `headless-gl` demonstrate it.

Alternative:
I believe we could instead import from `gl/src/javascript/webgl-rendering-context` and just instantiate the `WebGLRenderingContext` class. It seems a bit more fragile though because I don't think that's part of the official export.

Just wanted to flag this, thanks for all of the node development! It's great!

##### Pre-Merge Checklist

### DIFF
--- a/bundles/pixi.js-node/src/adapter/adapter.ts
+++ b/bundles/pixi.js-node/src/adapter/adapter.ts
@@ -1,7 +1,7 @@
 import type { IAdapter } from '@pixi/settings';
 import { settings } from '@pixi/settings';
 import { fetch, Request, Response } from 'cross-fetch';
-import gl from 'gl';
+import createContext from 'gl';
 import { NodeCanvasElement } from './NodeCanvasElement';
 
 import fs from 'fs';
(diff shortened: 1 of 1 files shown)


michaeljherrmann:
cc @Zyie (I think you did a lot of the node development)

Zyie (approved):
Overall i think this fix using `createContext` is fine for now and better than the alternative direct import

So I'm happy to add this and see if the PR you created goes anywhere and update again

bigtimebuddy:
Thank you @michaeljherrmann!

PR #8649, opened 2022-09-15
pointer over broken on touch devices

GordonTombola:
##### Description of change
prevent pointer over firing on touch pointers as pointerOut doesn't fire on touch, this leads to the activeInteractionData filling up every time you touch the screen

##### Pre-Merge Checklist

### DIFF
--- a/packages/interaction/src/InteractionManager.ts
+++ b/packages/interaction/src/InteractionManager.ts
@@ -1596,6 +1596,8 @@ export class InteractionManager extends EventEmitter
      */
     private onPointerOver(originalEvent: InteractivePointerEvent): void
     {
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Thanks for this @GordonTombola. Could you provide a little more context about when this occurs? Which platforms/browsers?

GordonTombola:
> Thanks for this @GordonTombola. Could you provide a little more context about when this occurs? Which platforms/browsers?

I've had this happen across both iOS and Android,
iOS - iphone 5c (safari)
iOS - iphone 7 plus (safari)
android - Pixel 4 (chrome) 

as an example I dirty patched it in my project via this method and it is working as expected now:
image

bigtimebuddy (approved):

bigtimebuddy:
Thanks @GordonTombola

PR #8652, opened 2022-09-15
Fix MSDF gray squares

ivanpopelyshev:
Bug: https://codesandbox.io/s/pixi-v6-bitmap-text-gray-square-rsp91x

Fixed: https://codesandbox.io/s/pixi-v6-bitmap-text-gray-square-fix-bf5k4l

Zoom out, see how gray squares appear

Two lines fix:
1. uFWidth should be 1.0 minimum, because if median is 0 then result should be 0 and not >0
2. mipmaps dont help with SDF textures, because SDF is linear and guarantees exact result if `scale > 1.0 / spread`. however if its bigger then even mipmaps wont help us, there's just not enough info.


bigtimebuddy (approved):

bigtimebuddy:
Thanks for the example

bigtimebuddy:
Thanks Ivan!

PR #8653, opened 2022-09-15
Add new AssetExtension

bigtimebuddy:
When trying to add support in `@pixi/sound` for the new Assets package, I noticed an opportunity to simplify the plugin API for Assets. Currently, there are 4 different extension types:

* `ExtensionType.LoadParser` (LoaderParser)
* `ExtensionType.ResolveParser` (ResolveURLParser)
* `ExtensionType.CacheParser` (CacheParser)
* `ExtensionType.DetectionParser` (FormatDetectionParser)

When you are trying to support a _single_ asset type (e.g. Spritesheet, Basis file, etc), it's somewhat cumbersome to implement all these pieces separately, since they often all need to be added to make a thing work. This PR combines all these extensions into one mega extension. This is basically syntax sugar to make it easier to co-locate common asset code. I've implemented this on `@pixi/spritesheet` there are probably other opportunities we can convert later.

```js
import { extensions, ExtensionType } from '@pixi/core';
import type { AssetExtension } from '@pixi/assets';

const myExtension = {
  extension: ExtensionType.Asset,
  loader: {
    // implement LoaderParser here 
  },
  resolver: {
    // implement ResolveURLParser here
  },
  cache: {
    // implement CacheParser here
  },
  detection: {
    // implement FormatDetectionParser here
  }
} as AssetExtension;

extensions.add(myExtension);
```

_Note: `loader`, `resolver`, `cache` and `detection` are all optional_


Zyie (approved):
The only other loaders that requires multiple parsers is `basis` and `compressed-textures` i believe

I can tackle this in a separate PR if you want?

PR #8674, opened 2022-09-27
Transfer the .shiftKey property from native -> federated mouse event

bhousel:
##### Description of change
I noticed that FederatedMouseEvent [does include a `shiftKey` property](https://github.com/pixijs/pixijs/blob/dev/packages/events/src/FederatedMouseEvent.ts#L31), but it doesn't seem like this property was being transferred over from the corresponding native MouseEvent.  Simple fix!

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good catch @bhousel

bigtimebuddy:
Thank you!

PR #8683, opened 2022-09-30
Move v6 into LTS, non-latest dist-tag

bigtimebuddy:
This change should be merged to coincide with the final v7 release. Sets the dist-tags for publishing v6 to be `latest-6.x` instead of `latest`. Also v6 no longer will update the `release` build on the CDN.


bigtimebuddy:
This is ready to be reviewed

SuperSodaSea (approved):

ShukantPal (approved):

PR #8698, opened 2022-10-06
Add @pixi/webworker bundle

SuperSodaSea:
This PR adds `@pixi/webworker` bundle for Web Workers. This bundle is basicly `pixijs` with `@pixi/accessibility` & `@pixi/events` removed and `WebWorkerAdapter` added, enough for running small demos in Web Workers.

See also https://github.com/pixijs/examples/pull/158.

Related Issue: https://github.com/pixijs/pixijs/issues/7123

##### Description of change

- Add `@pixi/webworker` bundle
- Add `IAdapter.getFontFaceSet()` to cover the difference between main thread and worker thread (`document.fonts` for main thread, `globalThis.fonts` for worker thread)
- `ImageBitmapResource` now recieve `string` in constructor, in order to support `Texture.from(url)` in Web Workers
- Other minor changes to make basic things work in Web Workers

##### Todo List

- [x] (Fixed) `Texture.from(url)` not working (It use `ImageResource`, which is not working since HTMLImageElement is not available in Web Workers. Perhaps we can make `ImageBitmapResource` recieve `string`). Using `await Assets.load(url)` is OK.

##### Pre-Merge Checklist

### DIFF
--- a/bundles/pixi.js-webworker/src/adapter.ts
+++ b/bundles/pixi.js-webworker/src/adapter.ts
@@ -1,6 +1,6 @@
-import { settings } from '@pixi/settings';
+import { settings } from '@pixi/core';
 
-import type { IAdapter } from '@pixi/settings';
+import type { IAdapter } from '@pixi/core';
 
 export const WebWorkerAdapter = {
     /**
--- a/bundles/pixi.js-webworker/src/adapter.ts
+++ b/bundles/pixi.js-webworker/src/adapter.ts
@@ -1,0 +1,23 @@
+import { settings } from '@pixi/settings';
+
+import type { IAdapter } from '@pixi/settings';
+
+export const WebWorkerAdapter = {
+    /**
+     * Creates a canvas element of the given size.
+     * This canvas is created using the browser's native canvas element.
+     * @param width - width of the canvas
+     * @param height - height of the canvas
+     */
+    createCanvas: (width?: number, height?: number): OffscreenCanvas =>
+        new OffscreenCanvas(width | 0, height | 0),
+    getWebGLRenderingContext: () => WebGLRenderingContext,
+    getNavigator: () => navigator,
+    getBaseUrl: () => globalThis.location.href,
+    getFontFaceSet: () => (globalThis as unknown as WorkerGlobalScope).fonts,
+    fetch: (url: RequestInfo, options?: RequestInit) => fetch(url, options),
+} as IAdapter;
+
+settings.ADAPTER = WebWorkerAdapter;
+
+export { settings };
--- a/packages/core/src/textures/resources/ImageBitmapResource.ts
+++ b/packages/core/src/textures/resources/ImageBitmapResource.ts
@@ -61,7 +61,7 @@ export class ImageBitmapResource extends BaseImageResource
 
         if (typeof source === 'string')
         {
-            super(ImageBitmapResource._DUMMY);
+            super(ImageBitmapResource.EMPTY);
 
             this.url = source;
         }
@@ -183,14 +183,21 @@ export class ImageBitmapResource extends BaseImageResource
             && (typeof source === 'string' || source instanceof ImageBitmap);
     }
 
-    private static __DUMMY: ICanvas;
+    /**
+     * Cached empty placeholder canvas.
+     * @see EMPTY
+     */
+    private static _EMPTY: ICanvas;
 
-    private static get _DUMMY(): ICanvas
+    /**
+     * ImageBitmap cannot be created synchronously, so a empty placeholder canvas is needed when loading from URLs.
+     * Only for internal usage.
+     * @returns The cached placeholder canvas.
+     */
+    private static get EMPTY(): ICanvas
     {
-        if (ImageBitmapResource.__DUMMY) return ImageBitmapResource.__DUMMY;
+        ImageBitmapResource._EMPTY = ImageBitmapResource._EMPTY ?? settings.ADAPTER.createCanvas(0, 0);
 
-        ImageBitmapResource.__DUMMY = settings.ADAPTER.createCanvas(0, 0);
-
-        return ImageBitmapResource.__DUMMY;
+        return ImageBitmapResource._EMPTY;
     }
 }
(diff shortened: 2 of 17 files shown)


andrewstart:
I don't really have the time to test this myself, but does this work in Firefox? Does [`gl.commit()`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/commit) need to be used (where available)?

SuperSodaSea:
> I don't really have the time to test this myself, but does this work in Firefox? Does [`gl.commit()`](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/commit) need to be used (where available)?

@andrewstart The example works just fine in my Firefox 105.
Firefox 105

The HTML Standard doesn't describe `WebGLRenderingContext.commit()` clearly, but it should be similar to [`OffscreenCanvasRenderingContext2D.commit()`](https://html.spec.whatwg.org/multipage/canvas.html#offscreencontext2d-commit):

> **offscreenCanvasRenderingContext2D.[commit](https://html.spec.whatwg.org/multipage/canvas.html#offscreencontext2d-commit)()**
> Copies the rendering context's [bitmap](https://html.spec.whatwg.org/multipage/canvas.html#offscreencontext2d-bitmap) to the bitmap of the [placeholder canvas element](https://html.spec.whatwg.org/multipage/canvas.html#offscreencanvas-placeholder) of the [associated OffscreenCanvas object](https://html.spec.whatwg.org/multipage/canvas.html#associated-offscreencanvas-object). The copy operation is synchronous. Calling this method is not needed for the transfer, since it happens automatically during the [event loop](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop) execution.

It seems that `commit()` is used to force the browser to copy the content from OffscreenCanvas to original canvas (synchronously), which is automatically done by the event loop. That is to say if we use `PIXI.Application`'s ticker (just like how we do in the examples) or something like `setTimeout` or `requestAnimationFrame` to render, the `commit()` is not necessary. It is only necessary to call `commit()` if we manually write a loop and keep rendering without going back to the event loop, for example:

```js
// These codes are running in worker thread

// Won't work, never go back to event loop and nothing displayed
while (true) {
    container.rotation -= 0.01;
    renderer.render();
}

// Work for both Chrome and FireFox, go back to event loop and copy canvas content when wait()
function wait() { return new Promise(resolve => setTimeout(resolve, 0)); }
while (true) {
    container.rotation -= 0.01;
    renderer.render();
    await wait();
}

// Work for Firefox (Chrome don't have gl.commit() currently)
while (true) {
    container.rotation -= 0.01;
    renderer.render();
    renderer.gl.commit();
}
```

**TL;DR**: The `commit()` is not necessary in most the cases. I can still add a method like `IRenderer.commit()` if wanted.

andrewstart:
Thanks for looking into that - this is actually the first I had heard of the feature, and given how recently Firefox added support I just wanted to make sure there wasn't a (purely hypothetical) situation where it worked in Chrome with an older version of the WIP spec but not Firefox with a later version. No changes needed from me.

packages/core/src/textures/resources/ImageBitmapResource.ts
```diff
@@ -1,27 +1,196 @@
...
+    private static __DUMMY: ICanvas;
+
+    private static get _DUMMY(): ICanvas
+    {
+        if (ImageBitmapResource.__DUMMY) return ImageBitmapResource.__DUMMY;
+
+        ImageBitmapResource.__DUMMY = settings.ADAPTER.createCanvas(0, 0);
+
+        return ImageBitmapResource.__DUMMY;
     }
```
bigtimebuddy, line 195: "Dummy" naming could be improved :) Also comments are helpful. 

```suggestion
    /** Temporary empty canvas to use when loading URLs */
    private static _EMPTY: ICanvas;
   
    /** Get cached temp canvas */
    private static get EMPTY(): ICanvas
    {
        ImageBitmapResource._EMPTY = ImageBitmapResource._EMPTY ?? settings.ADAPTER.createCanvas(0, 0);

        return ImageBitmapResource._EMPTY;
    }
```

SuperSodaSea: Done in ca8527757166676cf4d8e4482d072bd513469093.

bundles/pixi.js-webworker/src/adapter.ts
```diff
@@ -0,0 +1,23 @@
+import { settings } from '@pixi/settings';
+
+import type { IAdapter } from '@pixi/settings';
```
bigtimebuddy, line 3: We should import from core instead. settings/constants/ticker/utils/runner are all internal packages only to core.

```suggestion
import { settings } from '@pixi/core';

import type { IAdapter } from '@pixi/core';
```

packages/core/src/textures/resources/ImageBitmapResource.ts
```diff
@@ -1,27 +1,203 @@
+import { ALPHA_MODES } from '@pixi/constants';
+import { settings } from '@pixi/settings';
```
bigtimebuddy, line 2:
```suggestion
import { ALPHA_MODES, settings } from '@pixi/core';
```

packages/core/src/textures/resources/ImageResource.ts
```diff
@@ -1,14 +1,14 @@
-import { BaseImageResource } from './BaseImageResource';
-import { settings } from '@pixi/settings';
 import { ALPHA_MODES } from '@pixi/constants';
+import { settings } from '@pixi/settings';
```
bigtimebuddy, line 2:
```suggestion
import { ALPHA_MODES, settings } from '@pixi/core';
```

bigtimebuddy (commented):
Need to sort some imports, but otherwise looks great.

packages/core/src/textures/resources/ImageBitmapResource.ts
SuperSodaSea: `ImageResource` and `ImageBitmapResource` are inside of `@pixi/core`, they can't import `@pixi/core` itself.

bigtimebuddy: Yeah, my bad!

bigtimebuddy (approved):

bigtimebuddy:
Thanks @SuperSodaSea! Another great PR

PR #8699, opened 2022-10-06
Fix bitmapText frag alpha

ivanpopelyshev:
Trying to fix #8652 . 

`mipmaps dont help with SDF textures, because SDF is linear and guarantees exact result if scale > 1.0 / spread. however if its bigger then even mipmaps wont help us, there's just not enough info.`

this might be wrong. Needs testing from @insraq

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -626,7 +626,7 @@ export class BitmapText extends Container
 
             for (const mesh of this._activePagesMeshData)
             {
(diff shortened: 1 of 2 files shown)


ivanpopelyshev:
@bigtimebuddy Technically, previous solution could work, but I dont understand why it doesnt work with resolution != 1 and I fail to reproduce it on my computer. Thus, I'm trying to make different solution and asking for a test.

bigtimebuddy (approved):
These two examples both check out
https://codesandbox.io/s/pixi-v6-bitmap-text-gray-square-fix-forked-371e3j
https://codesandbox.io/s/elated-dijkstra-z9md8j

PR #8700, opened 2022-10-07
WIP: Added willReadFrequently to text metrics context

saitonakamura:
##### Description of change

Getting a warning from latest Chromium:

>Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently

In our case the warning is triggered from `getImageData` usage in `TextMetrics`. I plan to do a write a benchmark for it, but if you already have one, it'd be awesome to use it. I checked that ios Safari doesn't crash despite not supporting it, but I'm not sure if it's enough (maybe we need to `try` and fallback like in `_canvas` getter with `OffscreenCanvas`).

I'll create an issue with repro later

##### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/TextMetrics.ts
+++ b/packages/text/src/TextMetrics.ts
@@ -745,7 +745,7 @@ export class TextMetrics
     {
         if (!TextMetrics.__context)
         {
-            TextMetrics.__context = TextMetrics._canvas.getContext('2d');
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
I appreciate the contribution @saitonakamura, but I'm going to close in favor of #8746.

PR #8717, opened 2022-10-11
Replaces mandatory console banner with opt-in debug info

bigtimebuddy:
## Overview

See #8705 for context. 

This change removes the opt-out decorative console banner announcing PixiJS. Instead, it replaces this banner with an opt-in ~~`debug`~~ `hello` option that will print out the current version, renderer flavor ~~and collapsed startup options~~ without styling.

* Deprecates `utils.skipHello`
* Deprecates `utils.sayHello`

### Before

```js
```
Screen Shot 2022-10-11 at 9 28 27 AM

### After

```js
const renderer = new Renderer({ hello: true });
```

Screen Shot 2022-10-12 at 4 15 28 PM

## Example

https://jsfiddle.net/bigtimebuddy/r3tm0918/


GoodBoyDigital (approved):
👍

bigtimebuddy:
@domis86 console

domis86:
> @domis86 console

Ok - but where is it in code? i only see this but it doesnt look like is printing options:

```
        if (options.hello)
        {
            // eslint-disable-next-line no-console
            console.log(`PixiJS ${'$_VERSION'} - ${renderer.rendererLogId} - https://pixijs.com`);
        }
```

SuperSodaSea:
> > @domis86 console
> 
> Ok - but where is it in code? i only see this but it doesnt look like is printing options:
> 
> ```
>         if (options.hello)
>         {
>             // eslint-disable-next-line no-console
>             console.log(`PixiJS ${'$_VERSION'} - ${renderer.rendererLogId} - https://pixijs.com`);
>         }
> ```

It seems that it was removed in this commit: https://github.com/pixijs/pixijs/commit/2f49ed6e825994435379b1368d1742115286c2e0#diff-f16c80a3ef80af56cb7eaa973638efb31d86ad30cba8a5157c949985713074d9L53-R54

bigtimebuddy:
Oh yeah, @GoodBoyDigital and I discussed that. I removed including those options. We want to think about debugging in the future more broadly, so I narrowed the scope of this change to just replace the banner.

PR #8719, opened 2022-10-11
Recycle PageMeshData when BitmapText is destroyed

SuperSodaSea:
##### Description of change

When `BitmapText.destroy()` is called, push the `PageMeshData` that are still in `this._activePagesMeshData` to `pageMeshDataPool`, so that more `PageMeshData` can be reused.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good fix, thanks!

PR #8741, opened 2022-10-16
Add isImageBitmapSupported in WorkerManager

### LINKED ISSUE #8738: New Assets Loader is not Working in Unity IOS Ad Testing App
We are making playable ads with PixiJS, after I updated loader system to pixi's new asset manager, the application cannot pass loading state. So I put console's in the assets.mjs file and I get different results with IOS and Android.

### Steps to Reproduce

Step 1- This is how I load image in my loader function
image

Step 2- I put some console messages for debugging
image
and in __generator function
image

Step 3- I got different results with IOS and Android
For IOS
image
For Android
image
And this is the Windows Chrome console
image

The problem appears from the sent function in __generator, what IOS doesn't like is the question

### Environment
PixiJS v6.5.6

SuperSodaSea:
Add `isImageBitmapSupported` method in `WorkerManager`, so that the `loadTextures.load()` won't fail on browser that support `createImageBitmap` in main thread but not worker threads.

Fixes #8738.

##### Description of change

- Add `isImageBitmapSupported` method in `WorkerManager`
- `WorkerManager.loadImageBitmap` will `reject(e)` if error happens, instead of silently `resolve(null)`
- Check `response.ok` after `fetch()` and throw Error in order to get better error information
    - Before
    image
    - After
    image

##### Pre-Merge Checklist

### DIFF
--- a/packages/assets/src/loader/parsers/WorkerManager.ts
+++ b/packages/assets/src/loader/parsers/WorkerManager.ts
@@ -1,33 +1,79 @@
 let UUID = 0;
 let MAX_WORKERS: number;
 
+// 1x1 White PNG Data URL
+const WHITE_PNG = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42m'
+    + 'P8/x8AAwMCAO+ip1sAAAAASUVORK5CYII=';
+const checkImageBitmapCode = {
+    id: 'checkImageBitmap',
+    code: `
+    async function checkImageBitmap()
+    {
+        try
+        {
+            if (typeof createImageBitmap !== 'function') return false;
+
+            const response = await fetch('${WHITE_PNG}');
+            const imageBlob =  await response.blob();
+            const imageBitmap = await createImageBitmap(imageBlob);
+
+            return imageBitmap.width === 1 && imageBitmap.height === 1;
+        }
+        catch (e)
+        {
+            return false;
+        }
+    }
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
In the future, consider making a branch on the repo instead of using your fork. The advantage is that for every local branch, a build gets deployed. That helpful for validating/testing. I'll review this tomorrow.

SuperSodaSea:
> In the future, consider making a branch on the repo instead of using your fork. The advantage is that for every local branch, a build gets deployed. That helpful for validating/testing. I'll review this tomorrow.

Got it!

GoodBoyDigital (approved):
yeah this is cool @SuperSodaSea ! I like that we are parsing the error back to the main thread now too! could be useful for other things too!

SuperSodaSea:
Maybe in the future we can consider using some Rollup plugin to build the worker code, rather than hardcode it as a string.

GoodBoyDigital:
> Maybe in the future we can consider using some Rollup plugin to build the worker code, rather than hardcode it as a string.

I hear that! The first attempt of this kind of loader used webpack to bundle separate files - I was much easier to work with, but the code ended up being more bloated than it needed to be so went for the simple route. 

Happy to try again - as long as the output can be compiled into a string inlined into the codebase. Keen to not have users be responsible for import extra files for webworkers.

bigtimebuddy:
We are already using the `rollup-plugin-string` package to import `*.frag` and `*.vert`. I suggest we add a convention like `*.worker.ts` to externalize the code. Inlining it is not great because the code becomes hidden to TS and ESLint.

PR #8750, opened 2022-10-18
Chore: Update document for v6.x

### LINKED ISSUE #8740: Can't use the new Assets loader with pixi.js
### Expected Behavior
The [docs](https://pixijs.download/dev/docs/PIXI.Assets.html) show how to use the assets loader:

```ts
```

### Current Behavior
This doesn't work:

image

At runtime: `Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/pixi__js.js?v=25d6e02b' does not provide an export named 'Assets' (at plugin.ts:8:23)`

This is on pixi 6.5.6 which as far as I know is the default.

### Other attempt with `@pixi/assets`

Seeing that, I digged around and found this repo which seems to be in the process of converting to a monorepo (though I couldn't find any docs about that). I then found the `@pixi/assets` package and added it to my project.

This worked up to a point: When creating animated sprites using textures from the `Assets` package I get a strange error:

image

This happens because `this._texture` in `AnimatedSprite` is undefined, which in turns is caused by the following:

image

image

The textures coming from the `Assets` loader are not instances of the same base texture class as the one being checked inside `AnimatedSprite`.

This leads me to believe the reason this doesn't work is that using the `@pixi/assets` package on top of the normal `pixi.js` package probably ends up with two versions of the same files being imported, and thus two versions of the base texture class, maybe. Though when I looked at the monorepo, the `Assets` package seems to have no dependencies, so I don't know.

### Possible Solution

I don't know, I just expected to be able to import `Assets` from `pixi.js` as said in the docs.

I wouldn't mind using the monorepo package otherwise, but this is undocumented and I don't know if it's intended to be used.

### Steps to Reproduce

* Try to import `Assets` from `pixi.js`
* It doesn't exist

### Environment

- **`pixi.js` version**: 6.5.6
- **Browser & Version**: Chrome Version 106.0.5249.119 (Official Build) (arm64)
- **OS & Version**: MacOS
- **Note**: I am running this using [Vite](https://vitejs.dev). Though I don't think that should matter as this is standard import behaviour.

SuperSodaSea:
##### Description of change

Add document of how to install Assets and EventSystem for v6.x:
- https://pixijs.download/chore/v6-docs/docs/PIXI.Assets.html
- https://pixijs.download/chore/v6-docs/docs/PIXI.EventSystem.html

Also some minor revision in document is included.

Also fix the type issue caused by previous commit 2befc6864846c43928bac68e4feeca9c684bb8ea.

Fixes #8740.

##### Pre-Merge Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ import * as PIXI from 'pixi.js'
 #### CDN Install (via cdnjs)
 
 ```html
-<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.7/pixi.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.7/browser/pixi.min.js"></script>
 ```
 
 _Note: `6.5.7` can be replaced by any [released](https://github.com/pixijs/pixi.js/releases) version._
--- a/README.md
+++ b/README.md
@@ -65,13 +65,19 @@ There is no default export. The correct way to import PixiJS is:
 import * as PIXI from 'pixi.js'
 ```
 
-#### CDN Install (via cdnjs)
+#### CDN Install
+
+Via jsDelivr:
 
 ```html
-<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.7/browser/pixi.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/pixi.js@6.x/dist/browser/pixi.min.js"></script>
 ```
 
-_Note: `6.5.7` can be replaced by any [released](https://github.com/pixijs/pixi.js/releases) version._
+Or via unpkg:
+
+```html
+<script src="https://unpkg.com/pixi.js@6.x/dist/browser/pixi.min.js"></script>
+```
 
 ### Demos ###
 
--- a/README.md
+++ b/README.md
@@ -68,10 +68,10 @@ import * as PIXI from 'pixi.js'
 #### CDN Install (via cdnjs)
 
 ```html
-<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/5.1.3/pixi.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.7/pixi.min.js"></script>
 ```
 
-_Note: `5.1.3` can be replaced by any [released](https://github.com/pixijs/pixi.js/releases) version._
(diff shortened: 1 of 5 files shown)


bigtimebuddy (approved):
Great clean-up work. I think the docs here are super clear and address the original concern.

SuperSodaSea:
@bigtimebuddy By the way, I find that the number / icon before list items in the document is missing (`li { list-style-type: none; }` somewhere in `main.css`). Should this be fixed?

Incorrect:
image

Correct:
image

bigtimebuddy:
Yeah that's probably a mistake or a hold over from the JSDoc template.

bigtimebuddy:
If you're interested in improving the webdoc template, here are a few things that have been on my todo list:

- ability to toggle visibility of inherited members (particularly in the search, it's very busy)
- version chooser and warning about not being on a release version
- big: not all `@example` code is colorized
- ability to mark APIs as stable or experiment (not just deprecated). The goal is to send a stronger single around things that won't change

README.md
```diff
@@ -68,7 +68,7 @@ import * as PIXI from 'pixi.js'
 #### CDN Install (via cdnjs)
 
 ```html
-<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.7/pixi.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.7/browser/pixi.min.js"></script>
```
bigtimebuddy, line 71: This link isn't working.
Alternatively: https://unpkg.com/pixi.js/dist/browser/pixi.min.js

SuperSodaSea: 6.5.7 is not out yet :) [https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.6/browser/pixi.min.js](https://cdnjs.cloudflare.com/ajax/libs/pixi.js/6.5.6/browser/pixi.min.js) is available!

SuperSodaSea: Perhaps we can provide multiple CDN choices here?

bigtimebuddy: I like unpkg.com and jsdelivr.net because you don't need to provide the version to get the latest tag. At least for our purposes, it makes these readme's a little more maintainable.

https://cdn.jsdelivr.net/npm/pixi.js/dist/browser/pixi.min.js
https://unpkg.com/pixi.js/dist/browser/pixi.min.js

bigtimebuddy: Nice!

PR #8763, opened 2022-10-20
Missing peers causing unnecessary bundling

### LINKED ISSUE #8762: Module Size Regression - Extra bundling
We've continued to have issues with bundle-size regression in v6.5 and v7. There are some things that are getting double-bundled when running `npm run build:prod`.

https://bundlephobia.com/package/pixi.js@6.5.7

bigtimebuddy:
Fixes #8762

Also, adds a script to catch bundling errors in the future.

## Comparison

**pixi.js/dist/cjs/pixi.js**

**Before:** 1792850 bytes
**After:** 1608698 bytes


SuperSodaSea:
By the way, how can I see the cjs bundle size? I can only find the bundle for browser, and for cjs I can only find a small file with `require()`s in it  :thinking:

bigtimebuddy:
That's correct, the esm and cjs formats should be simple modules. Only the browser bundles are all rolled up with dependencies.

SuperSodaSea:
I still wondered why there is a huge gap between 6.4.2 & 6.5.1 on [https://bundlephobia.com/package/pixi.js@6.5.7](https://bundlephobia.com/package/pixi.js@6.5.7) ... I checked [https://unpkg.com/pixi.js@6.4.2/dist/browser/pixi.js](https://unpkg.com/pixi.js@6.4.2/dist/browser/pixi.js) and [https://unpkg.com/pixi.js@6.5.1/dist/browser/pixi.js](https://unpkg.com/pixi.js@6.5.1/dist/browser/pixi.js), and there is no big difference (1787508 bytes vs 1768869 bytes)

bigtimebuddy:
Yeah that's right, this issue didn't affect the overall size of the browser build, only the esm/cjs builds. So if you were using Webpack/Parcel/etc to build your project, you were paying the cost of some of these package inlining packages twice.

SuperSodaSea (approved):

SuperSodaSea (commented):
Seems that the `package-lock.json` is also needed to be updated?

PR #8768, opened 2022-10-21
Auto update BitmapText when its font is replaced

### LINKED ISSUE #8758: The size of bitmap text can't be updated after regenerating bitmap font
### Expected Behavior
A resized bitmap text after regenerating bitmap font.

### Current Behavior
The size of old bitmap text can't be updated while new bitmap text can be created with a correct size.

### Steps to Reproduce
```jsx
    width: window.innerWidth,
    height: window.innerHeight,
    backgroundColor: 0x000000
});

PIXI.BitmapFont.from("Arial", {
  fontFamily: "Arial",
  fontSize: 50,
  fill: 0xffffff,
}, {
  chars: ['1'],
});
let bitmapText = new PIXI.BitmapText("1", {
  fontName: "Arial",
});
bitmapText.position.set(500, 500);
app.stage.addChild(bitmapText);

setTimeout(() => {
    PIXI.BitmapFont.from("Arial", {
        fontFamily: "Arial",
        fontSize: 100,
        fill: 0xffffff,
    });
    let bitmapText2 = new PIXI.BitmapText("1", {
        fontName: "Arial",
    });
    bitmapText2.position.set(400, 400);
    app.stage.addChild(bitmapText2);
    bitmapText.updateText();
}, 1000);
```

### Environment

- **`pixi.js` version**: _above 5.3.0_
- **Browser & Version**: _Chrome 106.0.5249.119 & Edge 106.0.1370.47 & Firefox 106.0_

SuperSodaSea:
Fixes #8758.

##### Description of change

- The `BitmapText` will trigger `updateText()` automatically when `validate()` if the `BitmapFont` it use is replaced
    - See test `should update after font is replaced`
- If `BitmapText.fontSize` is `undefined` and the `BitmapFont` it uses is replaced, it will use the new size of `BitmapFont`, instead of the old value when it is constructed
    - See test `should update fontSize when font is replaced if fontSize is undefined`
    - See test `should not update fontSize when font is replaced if fontSize is defined`
- `BitmapFont.dirty` will become `false` after `updateText()`, so that if user calls `updateText()` externally, there won't be duplicated call to `updateText()` inside `BitmapFont`
    - See test `should unset dirty after updateText`

Example: https://pixiplayground.com/#/edit/EzMFsKC9YZe8pdhMOMrwg
- Old: Broken after replacing the font; if manually call `updateText()` the font size won't change
- New: Works fine and no need to call `updateText()`
    - `bitmapText1` will change its size
    - `bitmapText2` won't change its size since its `fontSize` is specified
    - `bitmapText3` will change its family

##### Pre-Merge Checklist

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -134,6 +134,12 @@ export class BitmapText extends Container
     protected _anchor: ObservablePoint;
 
     /**
+     * Private tracker for the current font.
+     * @private
+     */
+    protected _font: BitmapFont | null;
+
+    /**
      * Private tracker for the current font name.
      * @private
      */
@@ -143,7 +149,7 @@ export class BitmapText extends Container
      * Private tracker for the current font size.
      * @private
      */
-    protected _fontSize: number;
+    protected _fontSize?: number;
 
     /**
      * Private tracker for the current text align.
@@ -200,8 +206,9 @@ export class BitmapText extends Container
         this._textHeight = 0;
         this._align = align;
         this._tint = tint;
+        this._font = null;
         this._fontName = fontName;
-        this._fontSize = fontSize || BitmapFont.available[fontName].size;
+        this._fontSize = fontSize;
         this.text = text;
         this._maxWidth = maxWidth;
         this._maxLineHeight = 0;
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Do you think you could add a few tests to verify the new behavior?

SuperSodaSea:
> Do you think you could add a few tests to verify the new behavior?

I'll try :)

SuperSodaSea:
4 tests are provided!

bigtimebuddy:
Nice work, thanks!

PR #8770, opened 2022-10-22
Add missing events to FederatedEventMap

spencerflem:
##### Description of change
Adds the pointercancel event. https://developer.mozilla.org/en-US/docs/Web/API/Element/pointercancel_event 
Because touchcancel exists, it seems like pointercancel should be there.

##### Pre-Merge Checklist

### DIFF
--- a/packages/events/src/FederatedEventMap.ts
+++ b/packages/events/src/FederatedEventMap.ts
@@ -4,8 +4,8 @@ import type { FederatedWheelEvent } from './FederatedWheelEvent';
 export type FederatedEventMap = {
     click: FederatedPointerEvent;
     mousedown: FederatedPointerEvent;
-    mouseenter: FederatedPointerEvent
-    mouseleave: FederatedPointerEvent
+    mouseenter: FederatedPointerEvent;
+    mouseleave: FederatedPointerEvent;
     mousemove: FederatedPointerEvent;
     mouseout: FederatedPointerEvent;
     mouseover: FederatedPointerEvent;
--- a/packages/events/src/FederatedEventMap.ts
+++ b/packages/events/src/FederatedEventMap.ts
@@ -16,6 +16,7 @@ export type FederatedEventMap = {
     pointertap: FederatedPointerEvent;
     pointerup: FederatedPointerEvent;
     pointerupoutside: FederatedPointerEvent;
+    pointercancel: FederatedPointerEvent;
     rightclick: FederatedPointerEvent;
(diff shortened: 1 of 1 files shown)


packages/events/src/FederatedEventMap.ts
```diff
@@ -16,6 +16,7 @@ export type FederatedEventMap = {
     pointertap: FederatedPointerEvent;
     pointerup: FederatedPointerEvent;
     pointerupoutside: FederatedPointerEvent;
+    pointercancel: FederatedPointerEvent;
```
SuperSodaSea, line 19: Nice catch! And I think it is better to move this line to between `mouseupoutside` and `pointerdown` (so that the map is in alphabetical order).

By the way, I checked [EventBoundary.ts](https://github.com/pixijs/pixijs/blob/3ba43e50030641e2cb028c7c265e80ce0fdf6fff/packages/events/src/EventBoundary.ts) and found that the `mouseenter`, `mouseleave`, `pointerenter`, `pointerleave` are also missing. Can you add them?

spencerflem: added!
I'm not sure the consequences - but just making sure,
the mouseenter, mousedown etc. events are supposed to be FederatedPointerEvents and not FederatedMouseEvents?

SuperSodaSea: Seems that `FederatedMouseEvents` is only used as the base class of `FederatedPointerEvents` and `FederatedWheelEvent`, so I think it is OK to use `FederatedPointerEvents`.

packages/events/src/FederatedEventMap.ts
```diff
@@ -4,12 +4,17 @@ import type { FederatedWheelEvent } from './FederatedWheelEvent';
 export type FederatedEventMap = {
     click: FederatedPointerEvent;
     mousedown: FederatedPointerEvent;
+    mouseenter: FederatedPointerEvent
+    mouseleave: FederatedPointerEvent
```
SuperSodaSea, line 8:
```suggestion
    mouseenter: FederatedPointerEvent;
    mouseleave: FederatedPointerEvent;
```

spencerflem: Good catch! Must have run the linter wrong because it didn't find this 😬

SuperSodaSea (approved):
LGTM!

ShukantPal (approved):

bigtimebuddy:
Thank you @spencerflem!

PR #8779, opened 2022-10-24
fix import for TS

Dmitiry1921:
##### Pre-Merge Checklist


bigtimebuddy:
If you're having issues with this, try adding `"allowSyntheticDefaultImports": true,` to your tsconfig.

bigtimebuddy:
I don't think this is the right approach. If you can sort it out, happy to re-open.

PR #8781, opened 2022-10-24
Change default value of MIPMAP_TEXTURES to ON

### LINKED ISSUE #8775: When the width and height of texture are not integer power of 2, the result of BitmapFont.from looks worse
### Expected Behavior
When given texture sizes are similar, their effects should be similar

### Current Behavior
When the width and height of texture are not integer power of 2, the effects of generated bitmap font look worse after changing scale.
image
1. Texture size: 511x511
2. Texture size: 512x512
3. Texture size: 513x513

### Steps to Reproduce
https://www.pixiplayground.com/#/edit/4-KX0SsB4-Cv5UYesU4dS

### Environment

- **`pixi.js` version**: _above 5.3.0_
- **Browser & Version**: _Chrome 106.0.5249.119 & Edge 106.0.1370.47 & Firefox 106.0_

SuperSodaSea:
##### Description of change

Change default value of `MIPMAP_TEXTURES` from `POW2` to `ON`, since PixiJS can automatically disable mipmap for NPOT textures on WebGL 1. Also some document update included.

Fixes #8775.

##### Pre-Merge Checklist

### DIFF
--- a/packages/constants/src/index.ts
+++ b/packages/constants/src/index.ts
@@ -340,18 +340,19 @@ export enum WRAP_MODES
  * The {@link PIXI.settings.MIPMAP_TEXTURES} affects default texture filtering.
  * Mipmaps are generated for a baseTexture if its `mipmap` field is `ON`,
  * or its `POW2` and texture dimensions are powers of 2.
- * Due to platform restriction, `ON` option will work like `POW2` for webgl-1.
+ * Since WebGL 1 don't support mipmap for non-power-of-two textures,
+ * `ON` option will work like `POW2` for WebGL 1.
  *
  * This property only affects WebGL.
  * @name MIPMAP_MODES
  * @memberof PIXI
  * @static
  * @enum {number}
- * @property {number} OFF - No mipmaps
- * @property {number} POW2 - Generate mipmaps if texture dimensions are pow2
- * @property {number} ON - Always generate mipmaps
- * @property {number} ON_MANUAL - Use mipmaps, but do not auto-generate them; this is used with a resource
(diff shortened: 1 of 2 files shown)


SuperSodaSea:
@bigtimebuddy Do you think this change is OK for v6.x, or only for v7?

GoodBoyDigital:
Hey @SuperSodaSea , thanks for the PR!

What would be your thoughts around mipmaps being turned off by default for all textures rather than on? Just thinking about the memory bandwidth we could save people..

bigtimebuddy:
I'd agree with Mat on this one. The negative memory impact seems substantial enough to keep it as-is.

For the original issue, the user has two very viable options:

1) Enable MIPMAPS by for all textures `settings.MIPMAP_TEXTURES = MIPMAP_MODES.ON`
2) Only use POT sizes for BitmapFonts

SuperSodaSea:
That make sence. So should I keep the current default value `POW2` or change it to `OFF`?

bigtimebuddy:
There are numerous places in Pixi with defaults where we need to balance good rendering with fast performance. This is one of those places. I would vote to keep it as-is (POW2) since I think it's representative of the best practices for Pixi (eg using pow2 spritesheets) over arbitrary sized assets (background images, text). It may not be the best for all scenarios, but that's why we expose a global default and have a local mipmap option on BaseTexture to override per asset.

SuperSodaSea:
OK, now I reverted the default value to `POW2` and only keep the necessary changes in documents.

bigtimebuddy (approved):

PR #8787, opened 2022-10-26
Change `cacheAsBitmapMultisample` default to `null`

dev7355608:
Changes the default to `null` (multisample of the renderer) to match the behavior of `cacheAsBitmapResolution`, which is by default the renderer's resolution.

This default behavior was part of #7441 originally but then changed to `MSAA_QUALITY.NONE` to be reevaluated for v7.


bigtimebuddy (approved):
Good fix. Thanks.

PR #8805, opened 2022-10-31
Make @pixi/unsafe-eval self-installed

SuperSodaSea:
Since one rarely import `@pixi/unsafe-eval` without installing it, I'd like to make `@pixi/unsafe-eval` self-installed.

##### Description of change

- `@pixi/unsafe-eval` is self-installed when imported, so that `install()` no longer needs to be called manually
    - Deprecate `install()` in `@pixi/unsafe-eval`
- `unsafeEvalSupported()` is exported from `@pixi/core`
    - `@pixi/unsafe-eval` is only installed if `unsafeEvalSupported()` returns `false`

##### Pre-Merge Checklist


bigtimebuddy (approved):
I was thinking this same thing when I was looking at this recently! Thanks for this!

bigtimebuddy:
Out of curiosity, do you use this package in any projects or on any platforms?

SuperSodaSea:
> Out of curiosity, do you use this package in any projects or on any platforms?

In my personal project, which is still in a very early stage :)

SuperSodaSea:
Another example of using `@pixi/unsafe-eval` is when using PixiJS in Electron. Electron will show a CSP warning if `unsafe-eval` is enabled (See https://content-security-policy.com/examples/electron/). I don't like any warning in the console, so I need to disable `unsafe-eval` and use `@pixi/unsafe-eval`.

packages/unsafe-eval/src/install.ts
```diff
@@ -1,39 +1,53 @@
...
+ * @param _core
+ * @deprecated since 7.1.0
+ */
+export function install(_core: PIXICore): void
 {
-    if (!ShaderSystem)
-    {
-        throw new Error('Unable to patch ShaderSystem, class not found.');
-    }
+    utils.deprecation('7.1.0', 'install() has been deprecated, @pixi/unsafe-eval is self-installed since 7.1.0');
```
bigtimebuddy, line 19: One small thing , can you wrap this in `// #if _DEBUG` comment? This makes sure it's excluded from the mini field build.

SuperSodaSea: Sure, I will fix this later.

SuperSodaSea: Fixed!

PR #8807, opened 2022-11-01
Fix compatibility between `zone.js` and `FederatedEventTarget`

### LINKED ISSUE #8794: New EventSystem in v7 is conflicting with zone.js in Angular projects
### Expected Behavior

I should be able to use PixiJS in an Angular Project

### Current Behavior

I get the following error when importing `pixi.js` v7.0.0 in an Angular projet:

```sh
Error: node_modules/@pixi/events/lib/FederatedEventTarget.d.ts:11:18 - error TS2320: Interface 'FederatedEventTarget' cannot simultaneously extend types 'EventEmitter<string | symbol, any>' and 'EventTarget'.
  Named property 'removeAllListeners' of types 'EventEmitter<string | symbol, any>' and 'EventTarget' are not identical.

11 export interface FederatedEventTarget extends utils.EventEmitter, EventTarget {
                    ~~~~~~~~~~~~~~~~~~~~
```

It turns out `zone.js` is monkey patching the native EventTarget interface (https://github.com/angular/angular/blob/37ba6104498202b671f5a5a6bbfacc4df501788b/packages/zone.js/lib/zone.api.extensions.ts#L29) to add a `removeAllListeners?(eventName?: string): void;` method, which conflicts with the `removeAllListeners(event?: EventEmitter.EventNames<EventTypes>): this;` method on the EventEmitter from the `eventemitter3` package.

In an Angular projects (or any project using zone.js), it's impossible to create an interface that extends `EventEmitter` (from `eventemitter3`) and `EventTarget` at the same time.

It's possible to disable the monkey patch by adding this line to the Angular project `(window as ZoneGlobalConfigurations).__Zone_disable_EventTarget = true;` but the typing would still make the compilation fail.

### Possible Solution

Still searching for one, does the `FederatedEventTarget` interface need to extend the `EventTarget` ?

Right now I can bypass the problem by not using the new EventSystem. So I have to avoid importing `pixi.js` or `@pixi/events` and only import the other packages individually (`@pixi/core`, `@pixi/sprite`, ...)

### Steps to Reproduce

- Create an angular project
- Install `pixi.js`
- Import `pixi.js` or `@pixi/events` somewhere
- Serve the project

Here is a reproduction repository: https://github.com/Julien-Marcou/angular-pixijs-issue-8794

### Environment

- **`pixi.js` version**: 7.0.0
- **`angular` version**: 14.2.8
- **`zone.js` version**: 0.11.8

Julien-Marcou:
##### Description of change

It overrides the signature of the `removeAllListeners` method on the `FederatedEventTarget` interface to make it compatible with Angular projects.

Fixes #8794

##### Pre-Merge Checklist

### DIFF
--- a/packages/events/src/FederatedEventTarget.ts
+++ b/packages/events/src/FederatedEventTarget.ts
@@ -70,6 +70,15 @@ export interface FederatedEventTarget extends utils.EventEmitter, EventTarget
 
     /** The hit-area specifies the area for which pointer events should be captured by this event target. */
     hitArea: IHitArea | null;
+
+    // In Angular projects, zone.js is monkey patching the `EventTarget`
+    // by adding its own `removeAllListeners(event?: string): void;` method,
+    // so we have to override this signature when extending both `EventTarget` and `utils.EventEmitter`
+    // to make it compatible with Angular projects
+    // @see https://github.com/pixijs/pixijs/issues/8794
+
+    /** Remove all listeners, or those of the specified event. */
+    removeAllListeners(event?: string | symbol): this;
 }
 
 type AddListenerOptions = boolean | AddEventListenerOptions;


bigtimebuddy (approved):
This seems reasonable and thanks for doing all the leg work on this. Also, I appreciate the thorough comment.

PR #8816, opened 2022-11-02
Fix negative coords for roundPixels

ivanpopelyshev:
How the heck we did miss it?

There are two problems in this code:

If sprite coord X < 0 it gets blurred, because rounding works differently for Left and Right parts, left is Ceil, right is Floor. Same with Y. I guess people who work with texts complained about that but we didnt understand the cause.

Second problem is that the code works only for resolution < 1, to round low-quality versions of sprites.

My suggestion is to merge this to v7, and then decide whether we want both fixes for v6 , or only remove (| 0) part, which fixes the negative values.

And Yes, `X-Files` label is back!

### DIFF
--- a/packages/mesh/src/Mesh.ts
+++ b/packages/mesh/src/Mesh.ts
@@ -374,7 +374,7 @@ export class Mesh<T extends Shader = MeshMaterial> extends Container
 
             for (let i = 0; i < vertexData.length; ++i)
             {
-                vertexData[i] = Math.round((vertexData[i] * resolution | 0) / resolution);
+                vertexData[i] = Math.round(vertexData[i] * resolution) / resolution;
             }
         }
 
--- a/packages/sprite/src/Sprite.ts
+++ b/packages/sprite/src/Sprite.ts
@@ -289,7 +289,7 @@ export class Sprite extends Container
 
             for (let i = 0; i < vertexData.length; ++i)
             {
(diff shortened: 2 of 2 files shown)


bigtimebuddy:
Do you have an example of this not working?

yudinvictor:
We have a texture document (PIXI.Sprite from source svg). In the first screenshot the quality is good. In the second one (after the texture piece went over the upper border of the canvas), the quality has dropped drastically.
image
image

Closer demonstration of blurring 
image
image

yudinvictor:
You can see it better in the video


https://user-images.githubusercontent.com/43467675/199558422-6b63e4cc-a2ec-478a-9695-a3e71bf2da9f.mp4

bigtimebuddy:
@yudinvictor Thanks for these. Do you have some code we can run?

ivanpopelyshev:
Here, adjust `SCALE` parameter for your device to see it: 

https://www.pixiplayground.com/#/edit/RgDZsrV_WrESQQ1-P9Q5F

bigtimebuddy (approved):
This LGTM. Here's the v7 version (fixed): 
https://jsfiddle.net/bigtimebuddy/b72L0o6n/

bigtimebuddy:
Thanks @ivanpopelyshev

PR #8820, opened 2022-11-03
Add view type as a template for IRenderer, Application

bigtimebuddy:
### Current Issue

Since we are using ICanvas as a generic interface for OffscreenCanvas, HTMLCanvasElement and Node's canvas, we are getting a type error doing something pretty basic:

```ts
document.body.appendChild(app.view); // ERROR: Argument of type `ICanvas` is not assignable to type 'Node'
```

#### Workaround

Current workaround is to cast:

```ts
document.body.appendChild(app.view as HTMLCanvasElement);
```

### Fix

Use a template to specify the type of view.

```ts
const app = new Application<HTMLCanvasElement>();
```

```ts
const renderer = autoDetectRenderer<HTMLCanvasElement>();
```

### DIFF
--- a/packages/app/src/Application.ts
+++ b/packages/app/src/Application.ts
@@ -45,7 +45,7 @@ export interface Application extends GlobalMixins.Application {}
  * @class
  * @memberof PIXI
  */
-export class Application<VIEW = ICanvas>
+export class Application<VIEW extends ICanvas = ICanvas>
 {
     /** Collection of installed plugins. */
     static _plugins: IApplicationPlugin[] = [];
--- a/packages/app/src/Application.ts
+++ b/packages/app/src/Application.ts
@@ -45,7 +45,7 @@ export interface Application extends GlobalMixins.Application {}
  * @class
  * @memberof PIXI
  */
-export class Application
+export class Application<VIEW = ICanvas>
 {
     /** Collection of installed plugins. */
     static _plugins: IApplicationPlugin[] = [];
@@ -60,7 +60,7 @@ export class Application
      * WebGL renderer if available, otherwise CanvasRenderer.
      * @member {PIXI.Renderer|PIXI.CanvasRenderer}
      */
-    public renderer: IRenderer;
+    public renderer: IRenderer<VIEW>;
 
     /**
      * @param {object} [options] - The optional renderer parameters.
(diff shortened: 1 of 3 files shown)


Julien-Marcou (approved):
Would be nice if the `Application` could automatically type itself to `Application<OffscreenCanvas>`, `Application<HTMLCanvasElement>` or `Application<NodeCanvasElement>` depending on the context, but I'm not sure that's possible.

packages/app/src/Application.ts
```diff
@@ -45,7 +45,7 @@ export interface Application extends GlobalMixins.Application {}
  * @class
  * @memberof PIXI
  */
-export class Application
+export class Application<VIEW = ICanvas>
```
Julien-Marcou, line 48: Just realised, this should be `VIEW extends ICanvas = ICanvas` otherwise we would be able to create an Application with anything (e.g. `new Application<string>()`), but then you get the following error when creating an Application with `new Application<NodeCanvasElement>()`:

```
Type 'NodeCanvasElement' does not satisfy the constraint 'ICanvas'.
  The types returned by 'getContext(...)' are incompatible between these types.
    Type 'CanvasRenderingContext2D | WebGLRenderingContext' is not assignable to type 'ICanvasRenderingContext2D'.
```

packages/core/src/IRenderer.ts
```diff
@@ -111,7 +111,7 @@ export interface IRendererRenderOptions
  * Starard Interface for a Pixi renderer.
  * @memberof PIXI
  */
-export interface IRenderer extends SystemManager, GlobalMixins.IRenderer
+export interface IRenderer<VIEW = ICanvas> extends SystemManager, GlobalMixins.IRenderer
```
Julien-Marcou, line 114: Same with other `VIEW = ICanvas` that should be `VIEW extends ICanvas = ICanvas`

Julien-Marcou (changes requested):

packages/app/src/Application.ts
SuperSodaSea: Oh, the `NodeCanvasElement` problem is there since I made interface `ICanvas`. The `NodeCanvasElement` has some `// @ts-expect-error` in its current implementation (which is bad and unnecessary in my opinion), so maybe we need some breaking changes to make `NodeCanvasElement` implicit convertible to `ICanvas`.

SuperSodaSea: See PR #8823.

Julien-Marcou (approved):

Julien-Marcou:
LGTM!

SuperSodaSea (approved):
LGTM!

weiyidechicheng:
good

weiyidechicheng:
LGTM！

PR #8823, opened 2022-11-05
Fix: Make NodeCanvasElement compatible with ICanvas

SuperSodaSea:
##### Description of change

- Make `NodeCanvasElement` compatible with `ICanvas`, so that `NodeCanvasElement` can be convert into `ICanvas` without `as unknown`.
    - Before: `new NodeCanvasElement(width, height) as unknown as ICanvas`
    - After: `new NodeCanvasElement(width, height) as ICanvas`
- Add `ContextSettings` as the type of `ICanvas.getContext`'s `options` argument.

##### Pre-Merge Checklist

### DIFF
--- a/bundles/pixi.js-node/src/adapter/NodeCanvasElement.ts
+++ b/bundles/pixi.js-node/src/adapter/NodeCanvasElement.ts
@@ -56,7 +56,7 @@ type TempCtx = WebGLRenderingContext & {
  * @class
  * @memberof PIXI
  */
-export class NodeCanvasElement extends Canvas implements ICanvas
+export class NodeCanvasElement extends canvasModule.Canvas implements ICanvas
 {
     public style: Record<string, any>;
     private _gl: WebGLRenderingContext;
--- a/bundles/pixi.js-node/src/adapter/NodeCanvasElement.ts
+++ b/bundles/pixi.js-node/src/adapter/NodeCanvasElement.ts
@@ -132,6 +132,8 @@ export class NodeCanvasElement extends canvasModule.Canvas implements ICanvas
     }
 
     // @ts-expect-error - overriding getContext
+    override getContext(contextId: ContextIds, options?: any): RenderingContext | null;
+    // @ts-expect-error - overriding getContext
     override getContext(
         type: ContextIds,
         options?: NodeCanvasRenderingContext2DSettings | WebGLContextAttributes
--- a/bundles/pixi.js-node/src/adapter/NodeCanvasElement.ts
+++ b/bundles/pixi.js-node/src/adapter/NodeCanvasElement.ts
@@ -56,7 +56,7 @@ type TempCtx = WebGLRenderingContext & {
  * @class
  * @memberof PIXI
  */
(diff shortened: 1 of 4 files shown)


SuperSodaSea:
In the future, I'd like to change `Canvas` into a member of `NodeCanvasElement`, rather than `NodeCanvasElement extends Canvas`, so that we can get rid of the `// @ts-expect-error` for overriding `Canvas`. Since this will be a breaking change, I think it is better to do it in another PR for v7.1.0.

bigtimebuddy (approved):
Thanks

Julien-Marcou:
Although I'm fine with this change, we could get a better typing by defining each signature separately, it would allow us to directly know which context is supported or not.

In `ICanvas.ts`:

```ts
export type ContextOptions = CanvasRenderingContext2DSettings | ImageBitmapRenderingContext | WebGLContextAttributes;
```

In `NodeCanvasElement.ts`:


```ts
import type { ContextIds, ContextOptions, ICanvas } from '@pixi/core';

// ...

    // @ts-expect-error - overriding getContext
    override getContext(
        contextId: '2d',
        options?: NodeCanvasRenderingContext2DSettings | CanvasRenderingContext2DSettings
    ): CanvasRenderingContext2D | null;
    // @ts-expect-error - overriding getContext
    override getContext(
        contextId: 'bitmaprenderer',
        options?: NodeCanvasRenderingContext2DSettings | ImageBitmapRenderingContextSettings
    ): null;
    // @ts-expect-error - overriding getContext
    override getContext(
        contextId: 'webgl' | 'experimental-webgl',
        options?: NodeCanvasRenderingContext2DSettings | WebGLContextAttributes
    ): WebGLRenderingContext | null;
    // @ts-expect-error - overriding getContext
    override getContext(
        contextId: 'webgl2' | 'experimental-webgl2',
        options?: NodeCanvasRenderingContext2DSettings | WebGLContextAttributes
    ): null;
    // @ts-expect-error - overriding getContext
    override getContext(
        contextId: ContextIds,
        options?: NodeCanvasRenderingContext2DSettings | ContextOptions,
    ): RenderingContext | null
    {
        switch (contextId)
        {
```

Which gives:

```ts
const canvas = new NodeCanvasElement();
const context = canvas .getContext('webgl2'); // context is null

const canvas = new NodeCanvasElement();
const context = canvas .getContext('2d'); // context is CanvasRenderingContext2D

canvas as ICanvas; // Good!
```

SuperSodaSea:
@Julien-Marcou Fixed in 44ff6ff! Since the name `ContextOptions` is already used in [`ContextSystem`](https://github.com/pixijs/pixijs/blob/44ff6ff7c865990f336abfe73734bfe965d33630/packages/core/src/context/ContextSystem.ts#L19-L31), I use the name `ContextSettings` instead (and this name also matches `CanvasRenderingContext2DSettings` and `ImageBitmapRenderingContextSettings`).

Julien-Marcou (approved):

Julien-Marcou:
LGTM!

bigtimebuddy:
Thank you @SuperSodaSea

PR #8825, opened 2022-11-06
Add missing resolution setter for Renderer and CanvasRenderer

### LINKED ISSUE #8792: No way to change renderer resolution in 7.0?
Previously resolution could be set after initialization, but in 7.0 that doesn't appear to be possible anymore: https://github.com/pixijs/pixijs/blob/4cbadba3e46c4987f505e76a7f4e844d95a120cc/packages/core/src/Renderer.ts#L481

A browser's window.devicePixelRatio can dynamically change due to page zoom so it's necessary to update resolution between renders without having to reinitialize everything.

### Environment

- **`pixi.js` version**: 7.0.0

### LINKED ISSUE #8804: Adjusting Scale Breaks Click Events
### Expected Behavior
When you adjust the renderer resolution click events on sprite should still work when you click the sprite

### Current Behavior
When you adjust the renderer resolution click events fire when you click the possition the sprite would be if the resolution was 1

### Possible Solution
?

### Steps to Reproduce
Create a PIXI App
set the 
renderer.resolution = 0.5

Add a sprite to stage
Add click event on sprite
Click sprite - Event does not fire
Click where sprite would be at 100% resolution - Event Fires

### Environment

- **`pixi.js` version**:  6.5.8
- **Browser & Version**: Chrome 107
- **Running Example**: 
```
app.renderer.resolution = 0.5

let bol = false;

// create a texture from an image path
const texture = PIXI.Texture.from('examples/assets/flowerTop.png');

// create a second texture
const secondTexture = PIXI.Texture.from('examples/assets/eggHead.png');

// create a new Sprite using the texture
const dude = new PIXI.Sprite(texture);

// center the sprites anchor point
dude.anchor.set(0.5);

// move the sprite to the center of the screen
dude.x = app.screen.width / 2;
dude.y = app.screen.height / 2;

app.stage.addChild(dude);

// make the sprite interactive
dude.interactive = true;
dude.buttonMode = true;

dude.on('pointertap', () => {
    bol = !bol;
    if (bol) {
        dude.texture = secondTexture;
    } else {
        dude.texture = texture;
    }
});

app.ticker.add(() => {
    // just for fun, let's rotate mr rabbit a little
    dude.rotation += 0.1;
});

```

bigtimebuddy:
Closes #8792
Closes #8804

There are uses cases like automatically adjusting the resolution based on the browser zoom. This automatically adjusts the Event's resolution as well.

### :gift: Added

* Adds setter for Renderer and CanvasRenderer's `resolution`

### DIFF
--- a/packages/canvas-renderer/src/CanvasRenderer.ts
+++ b/packages/canvas-renderer/src/CanvasRenderer.ts
@@ -189,7 +189,17 @@ export class CanvasRenderer extends SystemManager<CanvasRenderer> implements IRe
         options = Object.assign({}, settings.RENDER_OPTIONS, options);
 
         const systemConfig = {
-            runners: ['init', 'destroy', 'contextChange', 'reset', 'update', 'postrender', 'prerender', 'resize'],
+            runners: [
+                'init',
+                'destroy',
+                'contextChange',
+                'resolutionChange',
+                'reset',
+                'update',
+                'postrender',
+                'prerender',
+                'resize'
+            ],
             systems: CanvasRenderer.__systems,
             priority: [
                 'textureGenerator',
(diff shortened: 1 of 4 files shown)


Julien-Marcou (approved):
LGTM!

PR #8582, opened 2022-08-12
Fix JSDoc of PIXI.DisplayObject#childRemoved

ryiwamoto:
##### Description of change
I fixed JSDoc of PIXI.DisplayObject#childRemoved. (The event was introduced by https://github.com/pixijs/pixijs/pull/5472)

##### Pre-Merge Checklist
- [x] Documentation is changed or added
- [x] Lint process passed (`npm run lint`)

PR #7718, opened 2021-08-18
Stricter typings for event emitter and federated events

s-r-x:
##### Description of change

Closes #7429
##### Pre-Merge Checklist

- [ ] Tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #8828, opened 2022-11-06
Event emitter types improvements

zOadT:
##### Description of change

Hello,

these are just very minor changes to slightly improve DX:

1. Use labeled tuples
    * A very minor improvement in DX:
Bildschirm­foto 2022-11-06 um 09 51 35
    Before the signature of fn was `fn: (args_0: PIXI.FederatedPointerEvent) => void`
    * The syntax used requires TypeScript 4.1. If that's a problem I can replace it by just writing out the events by hand.

    
2. Add types for `*capture` events
    * These seemed to be missing, I hope this wasn't intentional?

**Note**: The event `removedFrom` doesn't seem to be used anywhere (haven't found it in the docs and neither anywhere in the code). Shall I remove this from the types?

##### Pre-Merge Checklist

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -13,12 +13,12 @@ export interface IDestroyOptions
 
 export interface DisplayObjectEvents extends GlobalMixins.DisplayObjectEvents
 {
-    added: [Container];
-    childAdded: [DisplayObject, Container, number];
-    childRemoved: [DisplayObject, Container, number];
+    added: [container: Container];
+    childAdded: [child: DisplayObject, container: Container, index: number];
+    childRemoved: [child: DisplayObject, container: Container, index: number];
     destroyed: [];
-    removed: [Container];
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
I think you should remove `removedFrom`. I'm not seeing any references to that either. I believe it was added in #7718. @s-r-x, any thoughts on that? Was that added by mistake?

bigtimebuddy (approved):
These are nice type improvements. Thank you!

SuperSodaSea:
> I think you should remove `removedFrom`. I'm not seeing any references to that either. I believe it was added in #7718. @s-r-x, any thoughts on that? Was that added by mistake?

`removedFrom` seems to be the wrong name for `childRemoved`, which is fixed in #8582.

zOadT:
Thank you for the clarification!

SuperSodaSea (approved):
LGTM!

bigtimebuddy:
Thank you @zOadT!

Issue #3085, opened 2016-10-10
SVGs not rendering in Chrome


prime-ralph:
For some reason, the demo pen used in #2510 (https://jsfiddle.net/vwzjzn4h/1/) shows pixi.js unable to render SVG to canvas properly (yes, I changed the js used in that pen to 4.0.3, the latest release, but somehow even as it uses 3.0.x it still does not show the svg).

I tried reproducing it with a fresh pen since I was seeing this in my current project: http://codepen.io/prime-ralph/pen/dpmjdq?editors=0010 (I included a sample png to show I was correctly rendering a PNG to canvas)

Is it a bug or I'm just doing something wrong?
Will try to test a few things more, will be reporting back if I find anything.

And I tried it already with a local SVG, not this one borrowed from @RanzQ. Tested with the latest version of Chrome (53.0.2785.143 m). Will try this with another browser later.

themoonrat:
Trying on the latest dev build of PIXI,
http://jsfiddle.net/f8fu0a84/1/
It flags up an error in the console

> pixi.js:22488 Uncaught Error: The SVG image must have width and height defined (in pixels), canvas API needs them.(…)

ivanpopelyshev:
Can you make example with loader, and render only onload?

prime-ralph:
@ivanpopelyshev Apparently I was calling the render function before the SVG was loaded. Using requestAnimationFrame I was able to render the stage at some point after the SVG was loaded.

Seems like even if you use a loader to load an SVG does not mean it has been loaded as a texture already. Will update pen.

prime-ralph:
Pen has been updated. Leaving it as a basic example of how not to load things :D I just wish this was documented somewhere.

themoonrat:
It would be a great help if you submitted a PR with the added documentation in the relevant location to help the next person who may have this issue.

prime-ralph:
Will do.

Issue #3068, opened 2016-10-05
MovieClip completing conditions


motan:
I need a clip with the following behavior: after some object is hovered the clip starts to play and it is looped. On click or mouseout from that object clip should play until its end and stop. The way I did this is changing loop property for the clip from true to false. In this case update method starts to check if clip is finished and then stops it. But sometimes it stops it immediately. The problem, as I can understand is that the total number of frames is comparing with the _currentTime property. After a few seconds the time becomes bigger than the number of frames and will be never less than it. 

```
else if (this._currentTime >= this._textures.length && !this.loop)
    {
        this.gotoAndStop(this._textures.length - 1);

        if (this.onComplete)
        {
            this.onComplete();
        }
    }
```

So my clip just stops at the frame it is on. Maybe there is another method to make the clip do what it should?

ivanpopelyshev:
Feel free to hack it, its just a JS class. You can also make a PR of it :)

motan:
Ok. I just thought about not to break something)

englercj:
You can always extend the MovieClip to your own class and override the behavior you want to change.

PR #8832, opened 2022-11-07
Chore: Update document

### LINKED ISSUE #8623: Starting with 6.5.x, graphics behave differently than before
I've attached a JS Fiddle to show what can go wrong in existing codebases:

https://jsfiddle.net/awn97mop/9/

The gist of this is, starting with 6.5.x, `PIXI.Graphics` calculate height/width differently than they did before, which I believe is due to [this pull request](https://github.com/pixijs/pixijs/pull/8343) (though I could be wrong about the cause, it feels like a bounds calculation thing!). When graphics are passed a fill of alpha 0 and no line style, they assume the graphics object has a height/width of 0.

I'm not suggesting that this be changed, as _technically_ this does make sense to me. But, it did make upgrading to the newest version of PIXI a little frustrating today as I had to figure out what had changed between versions enough to cause this. 

I'm putting this out there so that perhaps this can be spread through the official channels, or so that anyone else that decides to upgrade to the latest version and experiences a similar "what happened" moment has a point of reference.

The solution for us was to not use a fill with alpha 0 when there was also no linestyle. Not a complicated solution, and probably better in the longrun.

SuperSodaSea:
##### Description of change

- Add document for `GraphicsGeometry.bounds`
    - Closes #8623
- Fix document for `childRemoved` event in `Container`
- Fix issue link in `Texture`'s document
    - I believe the correct link should be #3085 instead of #3068
- Some other small fixes

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/Texture.ts
+++ b/packages/core/src/textures/Texture.ts
@@ -60,7 +60,8 @@ function removeAllHandlers(tex: any): void
  * // sprite1._textureID should not be undefined if the texture has finished processing the SVG file
  * ```
  *
- * You can use a ticker or rAF to ensure your sprites load the finished textures after processing. See issue #3068.
+ * You can use a ticker or rAF to ensure your sprites load the finished textures after processing.
+ * See issue [#3085]{@link https://github.com/pixijs/pixijs/issues/3085}.
  * @memberof PIXI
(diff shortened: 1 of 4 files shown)


bigtimebuddy (approved):
Good one!

PR #8841, opened 2022-11-10
Change "lib/node_modules" to "lib/external" with preserveModules

bigtimebuddy:
Related issue: https://github.com/bigtimebuddy/pixi-svg/issues/33
Issue found: https://codesandbox.io/s/pixijs-examples-events-hit-test-forked-h8eog8

```
ModuleNotFoundError
Could not find module in path: '@pixi/settings/lib/node_modules/ismobilejs/esm/index.js' relative to 
'/node_modules/@pixi/settings/lib/utils/isMobile.js'
```

I believe that codesandbox is having issues with preserveModules in Rollup. Specifically, the external dependencies that are bundled (e.g. ismobilejs) are added to a folder called `node_modules`. This PR changes that folder name to be `external`. Since `node_modules` implies something very specific, I don't think it's appropriate to use that here and may be messing up patterning matching with other tools that expect `node_modules` to be _actual_ modules.

Before:
```
@pixi/settings/lib/node_modules/ismobilejs/esm/index.js
```

After:
```
@pixi/settings/lib/external/ismobilejs/esm/index.js
```

I was not able to confirm codesandbox's issue locally, but I'd like to try to do a test prerelease to validate. Either way, I still think this is a useful change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
         "rollup": "^2.75.7",
         "rollup-plugin-esbuild": "^4.9.1",
         "rollup-plugin-jscc": "^2.0.0",
+        "rollup-plugin-rename-node-modules": "^1.3.1",
         "rollup-plugin-sourcemaps": "^0.4.2",
         "rollup-plugin-string": "^3.0.0",
         "ts-jest": "^26.0.0",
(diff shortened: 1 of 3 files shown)


Zyie (approved):
oooh that interesting!

such a simple fix, hopefully it works for CodeSandBox

bigtimebuddy:
Well yippie, that worked but still have no idea why. Moving on...

PR #8859, opened 2022-11-17
Fix: Handle Spritesheet with looped `related_multi_packs`

### LINKED ISSUE #8833: Loading multiple spritesheet JSONs with related_multi_packs linking to each other can cause PIXI.Assets.load() to deadlock and never resolve
### Expected Behavior

When loading multiple spritesheets that link to each other with related_multi_packs, it should be able to load them all without deadlocking.

### Current Behavior

If you try to load multiple spritesheets that use TexturePacker's related_multi_packs in its JSON file and that link to each other, it will attempt to load them even if they are already being loaded by another spritesheet, causing a deadlock.

```json
{"frames": {

...

"meta": {
	"app": "https://www.codeandweb.com/texturepacker",
	"version": "1.0",
	"image": "my-numbers-0.png",
	"format": "RGBA8888",
	"size": {"w":512,"h":512},
	"scale": "1",
	"related_multi_packs": [ "my-numbers-1.json", "my-numbers-2.json" ],
	"smartupdate": "$TexturePacker:SmartUpdate:ea7b985de3af11d8d7e7bff73b074569:80778496bccc42d13437c459447a7090:06d8674a9e41fa7741f7530a3f8cb3c3$"
}
}
```
```typescript

    const spritesheets: string[] = [
        "assets/my-numbers-0.json",
        "assets/my-numbers-1.json",
        "assets/my-numbers-2.json",
    ]
    console.log("Trying to load spritesheets all in a single load()...");
    await Assets.load(spritesheets);
    console.log("Never reached!");
```

### Possible Solution

Either the user has to carefully make sure to load multi-pack spritesheets one by one, in case it links to another spritesheet and loads it automatically, or the asset loader should implement a check to make sure it is not already loading that spritesheet so it doesn't deadlock itself.

### Steps to Reproduce

1. Download, extract and run the barebones demo project linked below ("npm i", "npm run local:buildRun")
2. Open the webpage and check the console to see that "Never reached!" is never printed.
3. Comment out test_one() and uncomment test_two() to see that the promises never resolve even when loading spritesheets individually all at once.

### Environment

- **`pixi.js` version**: 7.0.3 & also affecting 6.5.8
- **Browser & Version**: Edge Version 107.0.1418.35, Chrome Version 107.0.5304.88 (Official Build) (64-bit), Firefox 106.0.4 (64-bit) 
- **Running Example**: [related_multi_packs_bug_demo.zip](https://github.com/pixijs/pixijs/files/9960244/related_multi_packs_bug_demo.zip)

SuperSodaSea:
#### Description of change

`spritesheetAsset` can now handle looped `related_multi_packs` correctly.

Fixes #8833.

- `spritesheetAssetCache` is introduced to solve this problem
    - Each entry in the cache hold 2 Promises, see below for details
- The loading process is divided in to these steps:
    - Wait for the spritesheet itself being loaded (resolving Promise 0)
    - Wait for all related multi packs to be loaded, recursively (waiting their Promise 0)
    - Fill the spritesheet's `linkedSheets` field (resolving Promise 1)
    - Wait for all related multi packs' `linkedSheets` field being filled (waiting their Promise 1)
- If any spritesheet failed to load, all other spritesheet that directly or indirectly reference it will fail automatically
- :fire: **Breaking change**: Spritesheet with invalid `related_multi_packs` now throw TypeError when loading instead of silently ignoring them

#### Tests
- :heavy_plus_sign: Added:
    - `spritesheetAsset`
        - `should not create multipack resources when related multi packs is missing`
        - `should load looped multi packed spritesheets`
        - `should throw when related multi packs is not an array of string`
        - `should throw when related multi packs do not exist`
        - `should throw when indirect related multi packs do not exist`
        - `should throw when related multi packs is not a spritesheet`
- :heavy_minus_sign: Removed:
    - `spritesheetAsset`
        - `should not create multipack resources when related_multi_packs field is missing or the wrong type`

#### Pre-Merge Checklist


bigtimebuddy (commented):
Thanks for digging into this! The tests are awesome.

My overall impression is that this seems like a lot of plumbing. Is there any opportunity to simplify the cache/cache-entry/parse code? What could make this solution any simpler? One idea is using the initial Spritesheet as the cache object and append to that. Do we really need another static cache?

SuperSodaSea:
@bigtimebuddy Still thinking about how to simplify these codes, may take some time.

SuperSodaSea:
@bigtimebuddy I tried to move some code to a separate function, make the code a little cleaner.

The static cache is necessary in my opinion. Since the user can load multiple Spritesheet at the same time that reference each other (That's how the original problem is triggered!), we need to keep track of the progress of other Spritesheets through this static cache.

Giacom:
Hoping this gets resolved soon, this is a big blocker for us upgrading to v6

bigtimebuddy:
I'm still not crazy about this approach but I have an idea. Going to noodle on it. But this is high priority for 7.1.0.

bigtimebuddy:
First off, @SuperSodaSea thanks for doing the hard-work on this PR to get it to work and address errors users are having. Sorry this took so long to respond to, but @GoodBoyDigital @Zyie and I had a discussion about this PR (and Spritesheets at-large).

The consensus was that, for something simple like loading a Spritesheet, this PR seems very complicated. But more importantly, it fails a few unwritten rules about Assets that we had when conceiving it:

* **Assets load results should be self-contained**
* **Assets load results should be able to be cleanly unloaded**

By waiting for other simultaneous loads and doing dynamic linking cross-Spritesheets, it breaks both of these rules. For instance, when you "unload" a Spritesheet that has shared linkedSheet with another Spritesheet, what do you do? How do you cleanly destroy it? I suppose we could add reference counting, but that adds _more_ complexity. 

The conclusion we reached is that PixiJS shouldn't be responsible for resolving a very complex network (cross-linking objects and recursion checks). Our expectation is that Spritesheet with multi-packs are self-contained and can be fully-destroyed without having side effects on other Spritesheets.

### Suggestion

As an alternative approach:

* Keep the runtime validation of `related_multi_packs` it is good and helpful!
* Add a data-option to opt-out of processing `related_multi_packs` so that users can manually load multipacks themselves... for instance `Assets.load({ src: 'atlas.json', data: { ignoreMultiPacks: true }})`. For someone with a complex Spritesheet dependency graph, this is probably the best option for manually loading and linking themselves (or make their own Spritesheet load parser).
* Provide user-friendly errors for recursion or cross-linking, since we do not support these things and should be explicit about it.
* Encourage users with more complicated situations to utilize `loadBundle` instead of using `related_multi_packs` to load all their assets.

### Future Work

Moving forward, we can improve the Spritesheet object. This object is basically just a container for maintaining references to Textures, a BaseTexture and map of animations. When we load `related_multi_packs`, we should append them into a single Spritesheet object (remove the linkedSheets API). We'd need to support multiple BaseTextures to Spritesheet, but this could be pretty straightforward and help simplify a lot. Would also make it easier to do animations across multiple sheets, which has been an issue.

SuperSodaSea:
I'm going to close this PR. If anyone need this solution, you can copy `packages/spritesheet/src/spritesheetAsset.ts` and use it to replace the original `spritesheetAsset`. (You may need to remove the original `spritesheetAsset` with `extension.remove(spritesheetAsset)`.)

PR #8864, opened 2022-11-19
Fix: Jest running on Windows

SuperSodaSea:
##### Description of change

Previously using `npm test` on Windows will report `Error: spawn http-server ENOENT` due to the difference on the shell between Windows and other Unix-like OS. Since I usually working on Windows, I modified the Jest setup & teardown script to make it working on Windows. Package `tree-kill` is introduced to properly kill the `http-server` on Windows, since `process.kill()` will only kill the shell but not the real `http-server` process.

See https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows for more details.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Nice!

PR #8875, opened 2022-11-21
Include DEBUG statements in module output

### LINKED ISSUE #8874: `#if _DEBUG` blocks are always removed in packages' default exports
The `_DEBUG` is only defined when building plugins / bundles, so the `#if _DEBUG` blocks are always removed in packages' default exports. If users are using PixiJS with a bundler, they will never recieve deprecation warnings in `#if _DEBUG` blocks.

For example, in `7.1.0-alpha`, the user will never recieve the deprecation warning with:

```js
import { ShaderSystem } from '@pixi/core';
import { install } from '@pixi/unsafe-eval';

install({ ShaderSystem });
```

I wondered if there is a good way to solve this problem. Maybe by [adding "development" and "production" fields in "exports" in package.json](https://nodejs.org/api/packages.html#community-conditions-definitions)?

bigtimebuddy:
Close #8874 

Deprecation warnings should be included when building with a bundler.

Users can use plugins (e.g., [@rollup/plugin-strip](https://www.npmjs.com/package/@rollup/plugin-strip)) for removing deprecations warnings, console logs and asserts.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -42,7 +42,7 @@ async function main()
 
     const plugins = [
         rename(),
-        jscc({ values: { _VERSION: repo.version } }),
+        jscc({ values: { _VERSION: repo.version, _DEBUG: true } }),
         esbuild({ target: moduleTarget }),
         ...commonPlugins
     ];


SuperSodaSea (approved):
LGTM :)

PR #8876, opened 2022-11-21
Added `.on<event>` checking for display objects

### LINKED ISSUE #8837: `pointertap` differences when upgrading to v7
When upgrading from v6.5.3 to v7.0.3, I've noticed that the way our codebase has used the `pointertap` event has changed.

Previously, we would set `object.pointertap = function(){}`, which would ensure that only a single pointertap would exist at a time. From the docs, it looks like we're now supposed to do `.on('pointertap', function(){})`? Does this mean that there's no way to reproduce the behavior that existed in v6 and previous?

If the answer is "no", then that's fine, we'll find a way to refactor things, but the docs aren't super clear about this, nor can I find anything in the migration guide for v7!

- **`pixi.js` version**: 7.0.3
- **Browser & Version**: 107.0.5304.91
- **OS & Version**: Fedora 36

adam-savard:
##### Description of change

Previous to 7.x, we had the ability to set single events to properties like `.pointertap` on `DisplayObjects`. This PR adds that functionality back, but with the prefix `on`, so it's more like the DOM (`.onpointertap` for example).

Any events added through listeners are unaffected, though the set properties now execute before any of them.

##### Pre-Merge Checklist

Closes #8837


bigtimebuddy (changes requested):
This change lacks 1) types and 2) documentation. Recommend that DisplayObject have optional types. These can simply refer to the event declaration to keep from duplicating.

adam-savard:
@bigtimebuddy I did modify the JS Docs for the function; is there another spot that this feature should be documented?

Also, the code is basically unchanged from the previous implementation. Is there a better way to do this?

bigtimebuddy:
The change you added will only work runtime, but not with TypeScript.

Looking specifically for these things to be added [here to `FederatedDisplayObject`](https://github.com/pixijs/pixijs/blob/dev/packages/events/src/FederatedEventTarget.ts).

For instance, TypeScript needs these all the types declared explicitly, sort like this:

```ts
type EventHandler<T = FederatedPointerEvent> = (event: T) => void;

onclick:EventHandler = null;
onmousedown:EventHandler = null;
onmouseenter:EventHandler = null;
onmouseleave:EventHandler = null;
onmousemove:EventHandler = null;
onmouseout:EventHandler = null;
onmouseover:EventHandler = null;
onmouseup:EventHandler = null;
onmouseupoutside:EventHandler = null;
onpointercancel:EventHandler = null;
onpointerdown:EventHandler = null;
onpointerenter:EventHandler = null;
onpointerleave:EventHandler = null;
onpointermove:EventHandler = null;
onpointerout:EventHandler = null;
onpointerover:EventHandler = null;
onpointertap:EventHandler = null;
onpointerup:EventHandler = null;
onpointerupoutside:EventHandler = null;
onrightclick:EventHandler = null;
onrightdown:EventHandler = null;
onrightup:EventHandler = null;
onrightupoutside:EventHandler = null;
ontap:EventHandler = null;
ontouchcancel:EventHandler = null;
ontouchend:EventHandler = null;
ontouchendoutside:EventHandler = null;
ontouchmove:EventHandler = null;
ontouchstart:EventHandler = null;
onwheel:EventHandler<FederatedWheelEvent> = null;
```

And each of these needs documentation (`npm run docs` to validate)

```
/**
 * Description
 * @memberof PIXI.DisplayObject
 * @default null
 */
onclick:EventHandler = null;
```

adam-savard:
Ah, I was wondering if that's what you were driving at. We use a vanilla JS dev environment, so TS is new to me. Fair enough! I'll get on that.

packages/events/src/EventBoundary.ts
```diff
@@ -534,12 +534,21 @@ export class EventBoundary
...
      */
     protected notifyTarget(e: FederatedEvent, type?: string): void
     {
         type = type ?? e.type;
+
+        // call the `on${type}` for the current target if it exists
+        if ((e.currentTarget as any)[`on${type}`])
+        {
+            (e.currentTarget as any)[`on${type}`](e);
+        }
```
bigtimebuddy, line 551: I think we can remove the `as any` cast here and replace it with something like this:
```suggestion
        e.currentTarget[`on${type}`]?.(e);
```

adam-savard: I'd be fine with that personally, though ESLint isn't super happy about it. I get the following:

```
Element implicitly has an 'any' type because expression of type '`on${string}`' can't be used to index type 'FederatedEventTarget'
```

That was only avoided using the `as any` cast. I'm all for a more concise solution, though! Should that error be ignored, or should we think of another solution? (Thanks for your patience in this!)

bigtimebuddy: No problem! Thanks for taking the feedback. 

You could try cast  `on${string}` maybe as `as keyof IFederatedDisplayObject`

adam-savard: Sorry for the delay, but there were additional challenges in getting this to be a bit more type safe.

What I ended up with was the following:

```javascript
       const displayObject = (e.currentTarget as unknown as IFederatedDisplayObject);

        (displayObject[`on${type}` as keyof IFederatedDisplayObject] as EventHandler)?.(e as FederatedPointerEvent);
```

I the cast to `unknown` wasn't ideal either, but it's not `any`, so there's that. It had to be split across 2 lines because of the line length limit (avoiding more warnings).

Hopefully this is more in line with best practices?

bigtimebuddy:
This is looking good @adam-savard!

I made a few changes here: https://github.com/pixijs/pixijs/tree/pr_8876 if you wouldn't mind incorporating those. https://github.com/pixijs/pixijs/commit/18075c33e95fa7d9ddbd8773e1acbde91d27963f

* Moved the on* types up a level to the FederatedEventTarget (instead of IFederatedEventTarget) this is so that they could be picked up by the global mixins (global.d.ts in packages/events)
* Remove duplicate EventHandler (renamed to FederatedEventHandler)
* Added typescript doc for the on* types

adam-savard:
Changes have been implemented!

bigtimebuddy (approved):

bigtimebuddy:
Just for validation:
https://jsfiddle.net/bigtimebuddy/mcf0q5sr/

Checkout. Thanks for your work on this @adam-savard :tada:

PR #8884, opened 2022-11-22
Cleanup deprecation warnings, disambiguate new static settings

bigtimebuddy:
### Chore

* Remove `sharedLoader` from the Application options documentation
* Remove `PIXI.` from deprecation warnings
* Renamed `BatchRenderer.maxTextures` to `BatchRenderer.defaultMaxTextures`
* Renamed `BatchRenderer.batchSize` to `BatchRenderer.defaultBatchSize`
* Renamed `Filter.resolution` to `Filter.defaultResolution`
* Renamed `Filter.multisample` to `Filter.defaultMultisample`
* Decentralize settings types using GlobalMixins


packages/core/src/batch/BatchRenderer.ts
```diff
@@ -295,7 +295,7 @@ export class BatchRenderer extends ObjectRenderer
     get MAX_TEXTURES(): number
     {
         // #if _DEBUG
-        deprecation('7.1.0', 'PIXI.BatchRenderer.MAX_TEXTURES renamed to PIXI.BatchRenderer.maxTextures');
+        deprecation('7.1.0', 'BatchRenderer#MAX_TEXTURES renamed to BatchRenderer#maxTextures');
```
Zyie, line 298: This should be `renamed to BatchRenderer.defaultMaxTextures`?
Also there are '#' instead of '.'

Zyie: ahh! 

got used to seeing static properties in these PR's 😄

Zyie (approved):

packages/core/src/batch/BatchRenderer.ts
bigtimebuddy: Threw in this change for free 😝

bigtimebuddy: TITLE_CASING for members is not great

PR #8887, opened 2022-11-22
Add overload signatures for Assets

SuperSodaSea:
#### Description of change

The `Assets.load`, `Assets.get` and `Loader.load` used to return `Promise<T | Record<string, T>>` regardless of the type of the argument is `string` or `string[]`, this caused me to perform extra cast on the return value when using them. So I add overload signatures to these methods, now they will return a more exact type.

Old:
```ts
const texture = await Assets.load('textures/bunny.png') as Texture;
// type of `await Assets.load<Texture>('textures/bunny.png')` is Texture | Record<string, Texture>
```

New:
```ts
const texture = await Assets.load<Texture>('textures/bunny.png');
// type of `await Assets.load<Texture>('textures/bunny.png')` is Texture
```

The tests that use `Assets` or `Loader` are modified. Also some `import` arrangements are included.

#### Pre-Merge Checklist


bigtimebuddy (approved):
Nice work! Good change.

Issue #5571, opened 2019-04-09
These emoji characters become weird after letterSpacing


RhinoLu:
### Expected Behavior
set letterSpacing should not change this character
![註解-2019-04-10-234205](https://user-images.githubusercontent.com/349025/55893380-b8a7f280-5bea-11e9-9f31-df370f0ff9aa.gif)

### Current Behavior
Please take a look :
[PixiJS TextStyle](https://pixijs.io/pixi-text-style/#%7B%22style%22%3A%7B%22letterSpacing%22%3A1%7D%2C%22text%22%3A%22%F0%9F%98%80%20%F0%9F%98%81%20%F0%9F%98%82%20%F0%9F%A4%A3%20%F0%9F%98%83%20%F0%9F%98%84%22%7D)

### Possible Solution
https://github.com/pixijs/pixi.js/blob/7476583718bb5a4586a82a071eac37272f14245d/packages/text/src/Text.js#L262
from
```
const characters = String.prototype.split.call(text, '');
let currentPosition = x;
let index = 0;
let current = '';
while (index < text.length) {
  current = characters[index++];
  if (isStroke) {
    this.context.strokeText(current, currentPosition, y);
  }
  else {
    this.context.fillText(current, currentPosition, y);
  }
  currentPosition += this.context.measureText(current).width + letterSpacing;
}
```
to
```
let currentPosition = x;
for (const s of text) {
  if (isStroke) {
    this.context.strokeText(s, currentPosition, y);
  } else {
    this.context.fillText(s, currentPosition, y);
  }
  currentPosition += this.context.measureText(s).width + letterSpacing;
}
```

### Steps to Reproduce
"letterSpacing: 0" is OK, otherwise become weird.

### Environment

- **`pixi.js` version**: 4.8.7
- **Browser & Version**: Chrome 73
- **OS & Version**: Windows10 & Mac 14
- **Running Example**: [PixiJS TextStyle](https://pixijs.io/pixi-text-style/#%7B%22style%22%3A%7B%22letterSpacing%22%3A1%7D%2C%22text%22%3A%22%F0%9F%98%80%20%F0%9F%98%81%20%F0%9F%98%82%20%F0%9F%A4%A3%20%F0%9F%98%83%20%F0%9F%98%84%22%7D)

themoonrat:
Checked your fix, and it looks good! Putting up a PR for it now for v5

themoonrat:
Fixed for next v5 release

[...]

PR #8893, opened 2022-11-23
Feature: Add `Text.graphemeSegmenter` option

SuperSodaSea:
#### Description of change

The `Text` used to use `Array.from` to split strings into code points, which is not enough for characters that is composed of multiple code points. Now that most of the browsers support [`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter) which can split strings into grapheme cluster, we can use it to split strings properly, and fallback to `[...string]` when it is not available. The segmenter function is provided as a static field `Text.graphemeSegmenter: (s: string) => string[]`, and users can set it to their own polyfill when needed.

See https://github.com/pixijs/pixijs/issues/5571#issuecomment-1324614159 for more context.

#### Todo List

- [x] Type for `Intl.Segmenter`

#### Pre-Merge Checklist

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -10,6 +10,27 @@ import type { IDestroyOptions } from '@pixi/display';
 import type { ICanvas, ICanvasRenderingContext2D } from '@pixi/settings';
 import type { ITextStyle } from './TextStyle';
 
+// The type for Intl.Segmenter is only available since TypeScript 4.7.2, so let's make a polyfill for it.
+interface ISegmentData
+{
+    segment: string;
+}
+interface ISegments
+{
+    [Symbol.iterator](): IterableIterator<ISegmentData>;
+}
+interface ISegmenter
+{
+    segment(input: string): ISegments;
+}
+interface IIntl
+{
+    Segmenter?: {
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):
Nice one and helpful documentation :+1:

packages/text/src/Text.ts
```diff
@@ -353,13 +397,7 @@ export class Text extends Sprite
 
         let currentPosition = x;
 
-        // Using Array.from correctly splits characters whilst keeping emoji together.
-        // This is not supported on IE as it requires ES6, so regular text splitting occurs.
-        // This also doesn't account for emoji that are multiple emoji put together to make something else.
-        // Handling all of this would require a big library itself.
-        // https://medium.com/@giltayar/iterating-over-emoji-characters-the-es6-way-f06e4589516
-        // https://github.com/orling/grapheme-splitter
-        const stringArray = Array.from ? Array.from(text) : text.split('');
```
bigtimebuddy, line 362: It's good that we can just use the spread operator here instead of Array.from or split :+1:

PR #8912, opened 2022-11-30
Allow for easier overriding of text resolution 

Zyie:
Added a static `defaultAutoResolution` and `defaultResolution` to `Text` that allows you to more easily set the resolution of your text 

Before: 
```ts
// renderer resolution 1

settings.RESOLUTION = 2;

// resolution is still 1 because of autoResolution
const text = new Text('hi');
```

After:
```ts
// resolution 1

Text.defaultResolution = 2;
Text.defaultAutoResolution = false;

// text resolution is now 2
const text = new Text('hi');
```

Also [here](https://jsfiddle.net/d53t27bg/) is an example of how text becomes crisper when renderer resolution is 1 and text is 2

### DIFF
--- a/packages/text/src/Text.ts
+++ b/packages/text/src/Text.ts
@@ -81,12 +81,24 @@ export class Text extends Sprite
      * // renderer has a resolution of 1
      * const app = new Application();
      *
-     * settings.RESOLUTION = 2;
+     * Text.defaultResolution = 2;
      * Text.defaultAutoResolution = false;
      * // text has a resolution of 2
      * const text = new Text('This is a PixiJS text');
      */
     public static defaultAutoResolution = true;
+
+    /**
+     * If {@link PIXI.Text.defaultAutoResolution} is false, this will be the default resolution of the text.
+     * If not set it will default to {@link PIXI.settings.RESOLUTION}.
+     * @example
+     * Text.defaultResolution = 2;
+     * Text.defaultAutoResolution = false;
+     *
+     * // text has a resolution of 2
+     * const text = new Text('This is a PixiJS text');
+     */
+    public static defaultResolution: number;
 
     /**
      * New rendering behavior for letter-spacing which uses Chrome's new native API. This will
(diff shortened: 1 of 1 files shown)


packages/text/src/Text.ts
```diff
@@ -10,27 +10,6 @@ import type { IDestroyOptions } from '@pixi/display';
 import type { ICanvas, ICanvasRenderingContext2D } from '@pixi/settings';
 import type { ITextStyle } from './TextStyle';
 
-// The type for Intl.Segmenter is only available since TypeScript 4.7.2, so let's make a polyfill for it.
```
bigtimebuddy, line 13: Looks like all the ISegmenter stuff got removed?

packages/text/src/Text.ts
```diff
@@ -397,7 +366,13 @@ export class Text extends Sprite
...
         let currentPosition = x;
 
-        const stringArray = Text.graphemeSegmenter(text);
+        // Using Array.from correctly splits characters whilst keeping emoji together.
+        // This is not supported on IE as it requires ES6, so regular text splitting occurs.
+        // This also doesn't account for emoji that are multiple emoji put together to make something else.
+        // Handling all of this would require a big library itself.
+        // https://medium.com/@giltayar/iterating-over-emoji-characters-the-es6-way-f06e4589516
+        // https://github.com/orling/grapheme-splitter
+        const stringArray = Array.from ? Array.from(text) : text.split('');
```
bigtimebuddy, line 375: Was this intentional to revert this?

packages/text/src/Text.ts
Zyie: oh that strange...

packages/text/src/Text.ts
Zyie: It was not

Something weird has happened here. The only change should be adding the new variable and setting it in the constructor

bigtimebuddy:
I think adding `Text.defaultResolution` as undefined with a fallback to `settings.RESOLUTION` would allow you to do this:

```ts
Text.defaultResolution = 1;
Text.defaultAutoResolution = false;
```

At least this way it feels fully scoped to Text.

bigtimebuddy (approved):
LGTM

PR #8919, opened 2022-12-02
Add tests for AnimatedSprite.onLoop

### LINKED ISSUE #8752: Needs a param as loopCount in AnimatedSprite.onloop() method.
### Expected Behavior
When onloop methed of AnimatedSprite is called, we need to know how many loops are completed in the callback.
### Current Behavior
When onloop methed of AnimatedSprite is called, we don,t know how many loops are completed in the callback.
### Possible Solution
It is better to have a parameter as loopCounts in the callback: AnimatedSprite.onloop(loopCount).
### Environment

- **`pixi.js` version**: _e.g. 4.7.1_
- **Browser & Version**: _e.g. Chrome 67_
- **OS & Version**: _e.g. Ubuntu 18.04_
- **Running Example**: _url_

SuperSodaSea:
#### Description of change

~Add `AnimatedSprite.loopCount`, and `loopCount` parameter for `AnimatedSprite.onLoop`.~

Add tests for `AnimatedSprite.onLoop`.

#### Pre-Merge Checklist

### DIFF
--- a/packages/sprite-animated/src/AnimatedSprite.ts
+++ b/packages/sprite-animated/src/AnimatedSprite.ts
@@ -86,7 +86,7 @@ export class AnimatedSprite extends Sprite
      *     // Looped!
      * };
      */
-    public onLoop?: (loopCount: number) => void;
+    public onLoop?: () => void;
 
     private _playing: boolean;
     private _textures: Texture[];
@@ -110,9 +110,6 @@ export class AnimatedSprite extends Sprite
     /** The texture index that was displayed last time. */
     private _previousFrame: number;
 
-    /** The AnimatedSprite's current loop count. */
-    private _loopCount: number;
-
     /**
      * @param textures - An array of {@link PIXI.Texture} or frame
      *  objects that make up the animation.
@@ -135,7 +132,6 @@ export class AnimatedSprite extends Sprite
         this.onLoop = null;
 
         this._currentTime = 0;
-        this._loopCount = 0;
 
         this._playing = false;
         this._previousFrame = null;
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
I'm not a huge fan of this feature. This seems better handled by a user in their app code. For example, why not:

```js
const anim = new SpriteAnimation();
let loops = 0;
anim.onLoop = () => {
  ++loops;
};
```
Or this could be solved with inheritance by users extending SpriteAnimation.

The problem with having a `loopCount` property is it's not clear when it should be reset. For instance, I had assumed that when you `play` or `stop` it would reset the `loopCount` so that the number is the number of _contiguous_ loops played (right now, it's implemented as a number _anytime_ it is looped). I also think it's weird to have a setter for this property when it's kinda being handled like an event with `onLoop`. 

Overall, would not recommend merging this feature (but @SuperSodaSea you did a great job as usual implementing it!). I'm just not sure this makes the cut, IMO.

That said, AnimatedSprite could use some help! 

**Time-based Playback**

The biggest problem with AnimatedSprite is that it's a simple flipbook. Frames are changed based on the render loop tick. When an animator creates an animation, they usually think in terms of frames-per-second, but that's not at all how it's implemented. I'd love to see an option where playback of the AnimatedSprite is independent of the ticker playback (e.g., see [AnimatedGIF](https://github.com/pixijs/gif/blob/main/src/AnimatedGIF.ts)). This also would allow for frame-dropping when rendering performance is slowed.

**Allow Frame Durations**

Similar to above, allow holding frames for certain millisecond durations. Right now, you have to add extra textures to the list if you want to hold a frame for longer. It would be better to for this to optionally be time-based.

**Blend Frames**

Another feature I'd like to see with AnimatedSprite is cross-fading between frames. For low-frame animations, this can actually look nice to visual smooth between frames. Particularly for effects like smoke and fire, this can look better.

SuperSodaSea:
@bigtimebuddy Thank you for your long opinion! It will be better if you comment in the original issue #8752 with your explanation.
I think we can still keep the tests for `onLoop`, since it wasn't there before.

bigtimebuddy:
Sure, could you modify this to just add the tests? I'll comment on the other issue.

SuperSodaSea:
Done ;)

bigtimebuddy (approved):
Thanks for adjusting!

PR #8920, opened 2022-12-02
Optimize trimCanvas

### LINKED ISSUE #8815: trimCanvas: use better algorithm
Current trim detection go through all pixels, but the better algorithm is to crop rows/columns by there sides.

Example implementation: https://gist.github.com/timdown/021d9c8f2aabc7092df564996f5afbbf

https://github.com/pixijs/pixijs/blob/978eb26fd693cc1c66e164cfd987d072c358b319/packages/utils/src/media/trimCanvas.ts#L20

SuperSodaSea:
#### Description of change

Optimize `trimCanvas`, using the implementation in https://gist.github.com/timdown/021d9c8f2aabc7092df564996f5afbbf. Also a new test for trimming empty canvas is added.

Closes #8815.

#### Pre-Merge Checklist

### DIFF
--- a/packages/utils/test/trimCanvas.tests.ts
+++ b/packages/utils/test/trimCanvas.tests.ts
@@ -2,6 +2,20 @@ import { trimCanvas } from '@pixi/utils';
 
 describe('trimCanvas', () =>
 {
+    it('should trim empty canvas', () =>
+    {
+        const canvas = document.createElement('canvas');
+
+        canvas.width = 100;
+        canvas.height = 50;
+
+        const trimmedImageData = trimCanvas(canvas);
+
+        expect(trimmedImageData.width).toEqual(0);
+        expect(trimmedImageData.height).toEqual(0);
+        expect(trimmedImageData.data).toBe(null);
+    });
+
     it('should trim the canvas', () =>
     {
         const canvas = document.createElement('canvas');
@@ -15,7 +29,7 @@ describe('trimCanvas', () =>
 
         const trimmedImageData = trimCanvas(canvas);
 
-        expect(trimmedImageData.width).toEqual(9);
+        expect(trimmedImageData.width).toEqual(10);
         expect(trimmedImageData.height).toEqual(5);
         expect(trimmedImageData.data).toBeInstanceOf(ImageData);
     });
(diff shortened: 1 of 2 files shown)


packages/utils/test/trimCanvas.tests.ts
```diff
@@ -15,7 +29,7 @@ describe('trimCanvas', () =>
 
         const trimmedImageData = trimCanvas(canvas);
 
-        expect(trimmedImageData.width).toEqual(9);
+        expect(trimmedImageData.width).toEqual(10);
```
SuperSodaSea, line 32: I believe this is a bug in the old implementation. Since it is drawn by `context.fillRect(10, 20, 10, 5)`, the width should be 10.

https://github.com/pixijs/pixijs/blob/e2f27bf276732172dac4c00fef54df5febec8ea4/packages/utils/src/media/trimCanvas.ts#L64-L71

Should be `bound.right < x + 1` here.

Anyway, with the new implementation in this PR, we can get the correct width.

bigtimebuddy: Yeah, good point!

bigtimebuddy:
This is awesome. Would be nice to see a performance comparison? How much faster is this?

SuperSodaSea:
It would be roughly `(all pixels) / (trimmed pixels)` times faster in my opinion. Since it won't be much empty space when trimming text, it can be a huge boost. Is there a good tool / library for making a benchmark?

bigtimebuddy:
I don't know something off-hand. There's a benchmark npm package you could try? I'm not looking for  anything that rigorous, mostly to understand how much faster relative to current.

bigtimebuddy:
Here's a quick performance comparison:

**v7.0.4:** https://jsfiddle.net/bigtimebuddy/04vm18x6/ (15.038s)
**PR:** https://jsfiddle.net/bigtimebuddy/tqxbdfcg/ (14.206s)

This doesn't seem like a significant performance improvement (<10%). Could you run it and see if you see something similar?

bigtimebuddy (approved):
This works, it's less code, and you fixed the off-by-one error. LGTM.

SuperSodaSea:
@bigtimebuddy Similar result here. However, using `ctx.rect(10, 10, 780, 580)` I got 37.21s vs 22.989s, ~30% speed up. The fewer pixels are trimmed, the greater the speed improvement. The remaining time should be the cost of `getImageData`.

PR #8926, opened 2022-12-03
Chore: Cleanup core dependencies

SuperSodaSea:
#### Description of change

Follow up #8619.

Change some imports from `@pixi/math`, `@pixi/settings`, `@pixi/utils` to `@pixi/core`, mainly some `ICanvas` stuff. ~Also some import sorting is included.~

#### Pre-Merge Checklist


bigtimebuddy:
All the import fixes to core are good 👍

I'm not into the import order changes. These are an "unstable style changes"-- someone can easily change it later not knowing what the rule is. The best solution here is to add an eslint rule that enforces some standard. I think this should be another PR.

SuperSodaSea:
@bigtimebuddy Okay, it is reverted.

bigtimebuddy (approved):
Thanks!

PR #8949, opened 2022-12-07
Feature/cache warning fix

### LINKED ISSUE #8613: recall PIXI.Assets.load(...) get warning in console
### Expected Behavior
as the recommended usage shown in the doc, recalling PIXI.Assets.load(...)  should be no warning here
### Current Behavior  &Steps to Reproduce & Environment
warning BaseTexture added to the cache with an id [examples/assets/bunny.png] that already had an entry
image

GoodBoyDigital:
##### Description of change
The Cache warning now only gets called if a different asset is assigned to the id. No biggie if its the same asset!

fixes #8613

##### Pre-Merge Checklist

### DIFF
--- a/packages/assets/test/assets.tests.ts
+++ b/packages/assets/test/assets.tests.ts
@@ -403,7 +403,7 @@ describe('Assets', () =>
         expect(font).toBeInstanceOf(FontFace);
     });
 
-    it.only('should load not show a cache warning if the same asset is loaded twice', async () =>
+    it('should load not show a cache warning if the same asset is loaded twice', async () =>
     {
         await Assets.init({
             basePath,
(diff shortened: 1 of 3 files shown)


packages/assets/test/assets.tests.ts
```diff
@@ -402,4 +402,20 @@ describe('Assets', () =>
 
         expect(font).toBeInstanceOf(FontFace);
     });
+
+    it('should load not show a cache warning if the same asset is loaded twice', async () =>
```
Zyie, line 406:
```suggestion
    it('should not show a cache warning if the same asset is loaded twice', async () =>
```

SuperSodaSea (approved):

PR #8950, opened 2022-12-07
Fix asset ID clashes between bundles

### LINKED ISSUE #8944: PIXI.Assets bundles can have key conflicts in the Resolver even when only one bundle is loaded in.
### Expected Behavior

If there are multiple bundles that contain the same asset key and only one of them is loaded in, then you should be able to dynamically use the same asset key without worrying about conflicts in the Resolver.

```typescript
  
  const manifest = {
      bundles: [
          {
              name: 'flower',
              assets: [
                  {
                      name: 'character',
                      srcs: 'examples/assets/flowerTop.png',
                  },
              ],
          },
          {
              name: 'egg',
              assets: [
                  {
                      name: 'character',
                      srcs: 'examples/assets/eggHead.png',
                  },
              ],
          },
      ]
  };
```

Let's say this is my manifest, I should be able to just unloadBundle and loadBundle to hotswap the assets.

Even if I just do a loadBundle for `flower`, I'll get the `egg` character instead because it's last on the manifest. See my demo below for an example.

### Current Behavior

Currently when you init a manifest, PIXI will add all the bundles to the Resolver. Meaning that if these bundles share a key, which should be fine if you only intend to load one in at a time (i.e: i18n and localisation), it will randomly choose an asset based on which one was processed last and not let you unload and load in bundles with different keys.

### Possible Solution

Categorise the URLs in the Resolver by the bundles they belong to and only allow it to be searched for when that bundle has been loaded in.

### Environment

- **`pixi.js` version**: 7.0.4
- **Browser & Version**: Edge
- **Running Example**: 

```typescript
const app = new PIXI.Application({ background: '#1099bb' });

async function main() {
  
  const manifest = {
      bundles: [
          {
              name: 'flower',
              assets: [
                  {
                      name: 'character',
                      srcs: 'examples/assets/flowerTop.png',
                  },
              ],
          },
          {
              name: 'egg',
              assets: [
                  {
                      name: 'character',
                      srcs: 'examples/assets/eggHead.png',
                  },
              ],
          },
      ]
  };
  
  await PIXI.Assets.init({ manifest });
  
  let isFlower = true;
  const character = new PIXI.Sprite(PIXI.Texture.EMPTY);
  character.anchor.set(0.5);
  character.x = app.screen.width / 2;
  character.y = app.screen.height / 2;

  app.stage.addChild(character);

  character.interactive = true;
  character.cursor = 'pointer';
  
  const resources = await PIXI.Assets.loadBundle("flower");
  
  // BUG: But I get the egg because the egg was processed last for its key
  character.texture = resources.character;
}

main();

```

GoodBoyDigital:
##### Description of change
Tweak bundle resolving to prepend bundle ids to the keys they load. Fixes #8944

Loading a bundle will now always return the correct assets
```js
const manifest = {
            bundles: [
                {
                    name: 'bunny1',
                    assets: [
                        {
                            name: 'character',
                            srcs: 'textures/bunny.png',
                        },
                    ],
                },
                {
                    name: 'bunny2',
                    assets: [
                        {
                            name: 'character',
                            srcs: 'textures/bunny-2.png',
                        },
                    ],
                },
            ]
        };

await Assets.init({ manifest });

const resources = await Assets.loadBundle('bunny1');
// resources.character is 'textures/bunny.png'

const resources2 = await Assets.loadBundle('bunny2');
// resources2.character is 'textures/bunny-2.png'

// loading an asset outside of the bundle will return the last object to be assigned to the key
// bunny2 overrides bunny1 in global keys
// important thing here is that this line will work regardless of if you have loaded a bundle or not..
const texture1 = await Assets.load('character'); //  'textures/bunny-2.png'

// new alias are if you wanna be specific
const texture3 = await Assets.load('bunny1-character'); //  'textures/bunny.png'
const texture3 = await Assets.load('bunny2-character'); //  'textures/bunny-2.png'

```

##### Pre-Merge Checklist


bigtimebuddy (commented):
Could we externalize the `-` bundle separator? I could see it useful to change in some cases to keep from avoiding name conflicts with existing assets   (E.g., `_` or `:`)
My suggestion is add an init option on Assets, `bundleSep`.

GoodBoyDigital:
> Could we externalize the `-` bundle separator? I could see it useful to change in some cases to keep from avoiding name conflicts with existing assets (E.g., `_` or `:`) My suggestion is add an init option on Assets, `bundleSep`.

Good shout! added a couple of static properties to make it more flexible!

packages/assets/src/resolver/Resolver.ts
```diff
@@ -41,6 +41,33 @@ import type { PreferOrder, ResolveAsset, ResolverBundle, ResolverManifest, Resol
  */
 export class Resolver
 {
+    /** The character that is used to connect the bundleId and the assetId when generating a bundle asset id key */
+    public static readonly bundleIdAssetConnector = '-';
```
SuperSodaSea, line 45:
```suggestion
    public staticbundleIdAssetConnector = '-';
```

packages/assets/src/resolver/Resolver.ts
```diff
@@ -41,6 +41,33 @@ import type { PreferOrder, ResolveAsset, ResolverBundle, ResolverManifest, Resol
...
+
+    /**
+     * A function that generates a bundle asset id key from a bundleId and an assetId
+     * @param bundleId - the bundleId
+     * @param assetId  - the assetId
+     * @returns the bundle asset id key
+     */
+    public static readonly generateBundleAssetId: (
```
SuperSodaSea, line 53:
```suggestion
    public static generateBundleAssetId: (
```

packages/assets/src/resolver/Resolver.ts
```diff
@@ -41,6 +41,33 @@ import type { PreferOrder, ResolveAsset, ResolverBundle, ResolverManifest, Resol
...
+    ) => string = (bundleId, assetId) =>
+            `${bundleId}${Resolver.bundleIdAssetConnector}${assetId}`;
+
+    /**
+     * A function that generates an assetId from a bundle asset id key. This is the reverse of generateBundleAssetId
+     * @param bundleId - the bundleId
+     * @param assetBundleId - the bundle asset id key
+     * @returns the assetId
+     */
+    public static readonly generateAssetIdFromBundleAssetId: (
```
SuperSodaSea, line 65:
```suggestion
    public static generateAssetIdFromBundleAssetId: (
```

SuperSodaSea (commented):
It seems that these static members should not be `readonly`.

GoodBoyDigital:
thanks for the tweak suggestions @SuperSodaSea ! rolled in now 👍

GoodBoyDigital:
> @GoodBoyDigital looks like some failing tests?

god bless tests . will look into later today

GoodBoyDigital:
hey @bigtimebuddy , looks like tests broke due to sprite sheet not being imported (as i had split out the file)
is my fix ok??

packages/assets/test/assets.tests.ts
```diff
@@ -2,6 +2,9 @@ import { Assets } from '@pixi/assets';
 import { BaseTexture, Texture } from '@pixi/core';
 import { Spritesheet } from '@pixi/spritesheet';
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const keepImportAlive = [Spritesheet];
```
GoodBoyDigital, line 6: cc @bigtimebuddy - not sure if there is a more elegant way!

Zyie: Does just doing this work?

`import '@pixi/spritesheet'`

bigtimebuddy:
Good fix. 

My only feedback is I don't like the location of the static properties. No one every create `new Resolver`, so from a user's perspective, Resolver is always static. Static properties make more sense on classes that users actually interact with (Renderer, Text, Container). 

My suggestion is these should be Assets.init properties:

```js
Assets.init({
  resolver: {
    bundleIdentifier: {
      separator: '-',
      join: (bundleId, assetId) => `${bundleId}-${assetId}`,
      find: (bundleId, assetBundleId) => assetBundleId.replace(`${bundleId}-`, ''),
    }
  }
});
```

Also, open to names but something a little verbose would be better. We can document these properties to provide a little more clarity. The name doesn't need to say everything.

GoodBoyDigital:
> Good fix.
> 
> My only feedback is I don't like the location of the static properties. No one ever creates `new Resolver`, so from a user's perspective, Resolver's instance is static and therefore the class is somewhat opaque from users. Static properties make more sense on classes that users actually interact with (Renderer, Text, Container).
> 
> My suggestion is these should be Assets.init properties:
> 
> ```js
> Assets.init({
>   resolver: {
>     bundleIdentifier: {
>       separator: '-',
>       join: (bundleId, assetId) => `${bundleId}-${assetId}`,
>       find: (bundleId, assetBundleId) => assetBundleId.replace(`${bundleId}-`, ''),
>     }
>   }
> });
> ```
> 
> Also, open to names but something a little verbose would be better. We can document these properties to provide a little more clarity. The name doesn't need to say everything.

yeah I like that idea! I have updated the PR to reflect the thinking. 

I think my variable names are probably too verbose, but wander if your suggestions are too short.. 

Open to suggestions - Will see if some new names pop into my brain whilst i sleep :P

GoodBoyDigital:
how about

bundleIdAssetConnector -> connector
generateBundleAssetId -> createBundleAssetId
generateAssetIdFromBundleAssetId -> extractAssetIdFromBundle

bigtimebuddy (approved):

PR #8958, opened 2022-12-09
Fix loading assets with url params

### LINKED ISSUE #8942: The loadBitmapFont parser does not work with URLs that contain GET params, e.g: examples/assets/bitmap-font/desyrel.xml?version=1.0.0
### Expected Behavior

I should be able to load any assets with a URL that contains GET params.

### Current Behavior

The [loadBitmapFont](https://github.com/pixijs/pixijs/blob/dev/packages/text-bitmap/src/loadBitmapFont.ts) parser currently checks and tests to use the parser by seeing if the extension of the URL is inside a list. This fails with URLs with GET params since `extname` will not remove the params.

__Another possible issue__, other parsers only work since they do LIKE / includes() checks, this means it works but it can result in issues where a GET param contains a ".json" string for example.

### Possible Solution

Change all parsers to use a new helper function that properly strips the GET params from the URL to check for.

### Steps to Reproduce

1. Open https://pixijs.io/examples/#/text/bitmap-text.js
2. Add a GET param to the URL, e.g: `examples/assets/bitmap-font/desyrel.xml?version=1.0.0`

### Environment

- **`pixi.js` version**: 7.0.4
- **Browser & Version**: Edge
- **Running Example**: 

```typescript
const app = new PIXI.Application({ background: '#1099bb' });

PIXI.Assets.load('examples/assets/bitmap-font/desyrel.xml?version=1.0.0').then(() => {
    const bitmapFontText = new PIXI.BitmapText(
        'bitmap fonts are supported!\nWoo yay!', {
            fontName: 'Desyrel',
            fontSize: 55,
            align: 'left',
        },
    );

    bitmapFontText.x = 50;
    bitmapFontText.y = 200;

    app.stage.addChild(bitmapFontText);
});
```

Zyie:
`extname`, `basepath`, and `parse` did not take into account url params previously

Fixes #8942


SuperSodaSea (commented):
With this change, maybe we can change some `utils.path.extname(url).includes('.ext')` into `=== '.ext'`, for example here:

https://github.com/pixijs/pixijs/blob/1325282801a17f1974bb9285d351435542257377/packages/assets/src/loader/parsers/loadTxt.ts#L15

Also maybe we can make the extension check case-insensitive?

bigtimebuddy:
I agree with @SuperSodaSea, I don't think `includes` is a sufficient check. For instance ".svgz" vs ".svg" would give a false positive with a check `includes('.svg')`. These aren't super common, but would be good to be a little more precise.

So this:
```js
utils.path.extname(url).includes('.ext')
```
becomes:
```js
utils.path.extname(url).toLowerCase() === '.ext'
```

SuperSodaSea:
Another thing that concerns me is that `?` and `#` can be part of the local file path (for some filesystem), although this is not very common. Maybe we can add an option for that.

bigtimebuddy (approved):

SuperSodaSea (approved):

Issue #8852, opened 2022-11-15
Help Wanted: Enable strictNullChecks with TypeScript


bigtimebuddy:
## Overview

We need some help if you're savvy in TypeScript! We would like to update our TypeScript implementation to enable [`strictNullChecks`](https://www.typescriptlang.org/tsconfig#strictNullChecks). Doing so would add a level of strictness to our code that everyone downstream would benefit from. 

Below are the list of files affected.

### Running

```bash
npm run types -- --strictNullChecks 
```

```
...

Found 1444 errors.
```

### Filter By Package

First, add the path of the package to the `pathPrefixs` array in [`scripts/filterTypeScriptErrors.ts`](https://github.com/pixijs/pixijs/blob/dev/scripts/filterTypeScriptErrors.ts):

https://github.com/pixijs/pixijs/blob/38e2fc3ebe66260c47c334b8602e04b6dd00fc31/scripts/filterTypeScriptErrors.ts#L12-L14

Then run the following command:

```bash
npm run types:strict
```

You will see errors filtered by the specified packages.

If a PR want to enable strictNullChecks in a package, the path of the package should be added to the array in that PR. For further information, you can read the comments in [`scripts/filterTypeScriptErrors.ts`](https://github.com/pixijs/pixijs/blob/dev/scripts/filterTypeScriptErrors.ts), or PR #8965.

### Files Affected

- [x] ~bundles/pixi.js-node/src/adapter/adapter.ts~
- [x] ~bundles/pixi.js-node/src/adapter/loadNodeBitmapFont.ts~
- [x] ~bundles/pixi.js-node/src/adapter/NodeCanvasElement.ts~
- [x] ~bundles/pixi.js-node/src/adapter/NodeCanvasResource.ts~
- [x] bundles/pixi.js-webworker/src/adapter.ts
- [ ] packages/accessibility/src/AccessibilityManager.ts
- [ ] packages/accessibility/src/accessibleTarget.ts
- [ ] packages/accessibility/test/AccessibilityManager.tests.ts
- [ ] packages/app/src/Application.ts
- [ ] packages/app/src/ResizePlugin.ts
- [ ] packages/app/test/Application.tests.ts
- [ ] packages/assets/src/AssetExtension.ts
- [ ] packages/assets/src/Assets.ts
- [ ] packages/assets/src/BackgroundLoader.ts
- [ ] packages/assets/src/cache/Cache.ts
- [ ] packages/assets/src/loader/Loader.ts
[...]

PR #8960, opened 2022-12-09
Add strict null checks to @pixijs/utils

Emobe:
##### Description of change
Adds strict null checks to the @pixijs/utils package with reference to #8852

Not 100% sure if this is the approach that was wanted but this is what I came up with so far. 

##### Pre-Merge Checklist


packages/utils/src/media/trimCanvas.ts
```diff
@@ -38,28 +38,34 @@ export function trimCanvas(canvas: ICanvas): { width: number; height: number; da
...
-    const data = imageData.data;
 
-    let top = 0;
-    let bottom = height - 1;
-    let left = 0;
-    let right = width - 1;
-
-    while (top < height && checkRow(data, width, top)) ++top;
-    if (top === height)
+    if (context !== null)
```
SuperSodaSea, line 42: I prefer throwing a TypeError when `canvas.getContext` returns `null` instead of return `undefined` silently.

packages/utils/src/network/decomposeDataUri.ts
```diff
@@ -2,11 +2,11 @@ import { DATA_URI } from '../const';
...
-    mediaType: string;
-    subType: string;
-    charset: string;
-    encoding: string;
-    data: string;
+    mediaType: string | undefined;
+    subType: string | undefined;
+    charset: string | undefined;
+    encoding: string | undefined;
+    data: string | undefined;
```
SuperSodaSea, line 9: Maybe using `mediaType?: string;` here is better in my opinion.

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -69,27 +75,43 @@ export class CanvasRenderTarget
      * The width of the canvas buffer in pixels.
      * @member {number}
      */
-    get width(): number
+    get width(): number | null
```
SuperSodaSea, line 78: `get width()` returning `null` can be confusing. Maybe we should check if `this.canvas === null` (this means the `CanvasRenderTarget` has been destroyed) and throw a `TypeError`.

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -69,27 +75,43 @@ export class CanvasRenderTarget
...
+        if (this.canvas !== null)
+        {
+            return this.canvas.width;
+        }
+
+        return null;
     }
 
-    set width(val: number)
+    set width(val: number | null)
```
SuperSodaSea, line 88: `set width()` should not accept `null` in my opinion. Also I think it should throw when `this.canvas === null` just like I described in `get width()`.

SuperSodaSea (commented):
This is just part of my opinion about migrating code to `strictNullChecks`, others may have different ideas. CC @bigtimebuddy

By the way, is there a good way of enabling `strictNullChecks` for part of the code?

Emobe:
@SuperSodaSea Thanks for the review. I'm not so familar with the ins and outs of pixi so I appreciate this and will work on the changes.

I'm not sure if there is a "good" way for enabling it for part of the code. 

I think just something similar to the code provided in the issue of #8852 and making sure to run it. It's not really an ideal solution honestly but I don't think there is an official way supported

Emobe:
Added in the changes. Let me know if the error messages are alright or anything else crops up.

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -71,12 +77,20 @@ export class CanvasRenderTarget
      */
     get width(): number
     {
-        return this.canvas.width;
+        if (this.canvas !== null)
+        {
+            return this.canvas.width;
+        }
+
+        throw TypeError('Canvas is null');
```
SuperSodaSea, line 85: `throw new TypeError(...)` can be more consistent.

Also maybe we can add a private method to check if the `CanvasRenderTarget` is destroyed:
```ts
get width(): number
{
    // #ifdef _DEBUG
    this._checkDestroyed();
    // #endif

    return this.canvas.width;
}

// I didn't really test the asserts, hope this will work
private _checkDestroyed(): asserts this is this & { canvas: ICanvas, context: ICanvasRenderingContext2D }
{
    if (this.canvas === null)
    {
        throw new TypeError('The CanvasRenderTarget has already been destroyed');
    }
}
```

packages/utils/test/trimCanvas.tests.ts
```diff
@@ -11,9 +11,9 @@ describe('trimCanvas', () =>
 
         const trimmedImageData = trimCanvas(canvas);
 
-        expect(trimmedImageData.width).toEqual(0);
-        expect(trimmedImageData.height).toEqual(0);
-        expect(trimmedImageData.data).toBe(null);
+        expect(trimmedImageData?.width).toEqual(0);
+        expect(trimmedImageData?.height).toEqual(0);
+        expect(trimmedImageData?.data).toBe(null);
```
SuperSodaSea, line 16: `trimmedImageData?.xxx` is no longer needed here (and below).

packages/utils/test/trimCanvas.tests.ts
```diff
@@ -24,18 +24,25 @@ describe('trimCanvas', () =>
         canvas.width = 100;
         canvas.height = 50;
 
-        context.fillStyle = '#ff0000';
-        context.fillRect(10, 20, 10, 5);
-        context.fillStyle = '#00ff00';
-        context.fillRect(15, 25, 10, 5);
+        if (context !== null)
```
SuperSodaSea, line 27: We can use:

```ts
if (context === null) fail('Failed to create context');
```

And we don't need to check `context?.xxx` below, since `fail()` returns `never` and TypeScript can infer `context` won't be `null` below.

packages/utils/src/media/trimCanvas.ts
```diff
@@ -38,28 +38,33 @@ export function trimCanvas(canvas: ICanvas): { width: number; height: number; da
...
-    };
+        width = right - left + 1;
+        height = bottom - top + 1;
+
+        return {
+            width, height,
+            data: context.getImageData(left, top, width, height),
+        };
+    }
+    throw new TypeError('Canvas rendering context is null');
```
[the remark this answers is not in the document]
Emobe: Sure thing, does make it cleaner and readable

packages/utils/src/media/CanvasRenderTarget.ts
Emobe: Yeah I think this would be a good idea as I can see this having to be used in quite a few places with any more null check changes

Emobe:
Added in the requested changes

packages/utils/src/media/CanvasRenderTarget.ts
Emobe: The method is a good idea as this is used quite a bit

Emobe:
I see there's a CI issue. I'll take a look at this.

SuperSodaSea:
My fault, it should be `#if _DEBUG`

packages/utils/src/browser/isWebGLSupported.ts
```diff
@@ -30,9 +30,9 @@ export function isWebGLSupported(): boolean
                 let gl = (
                     canvas.getContext('webgl', contextOptions)
                     || canvas.getContext('experimental-webgl', contextOptions)
-                ) as WebGLRenderingContext;
+                ) as WebGLRenderingContext | null;
 
-                const success = !!(gl && gl.getContextAttributes().stencil);
+                const success = !!(gl && gl.getContextAttributes()?.stencil);
```
SuperSodaSea, line 35:
```suggestion
                const success = !!gl?.gl.getContextAttributes()?.stencil;
```

Emobe:
> My fault, it should be `#if _DEBUG`

Yeah I just noticed this. Added the change in

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -43,8 +43,12 @@ export class CanvasRenderTarget
...
     clear(): void
     {
-        this.context.setTransform(1, 0, 0, 1, 0, 0);
-        this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);
+        // #if _DEBUG
+        this._checkDestroyed();
+        // #endif
+
+        this.context?.setTransform(1, 0, 0, 1, 0, 0);
+        this.context?.clearRect(0, 0, this.canvas.width, this.canvas.height);
```
SuperSodaSea, line 51: Since we have `assets` in `_checkDestroyed`, `this.context` is also non-null.

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -85,11 +101,27 @@ export class CanvasRenderTarget
...
     set height(val: number)
     {
+        // #if _DEBUG
+        this._checkDestroyed();
+        // #endif
+
         this.canvas.height = Math.round(val);
     }
+
+    private _checkDestroyed(): asserts this is this & { canvas: ICanvas; context: ICanvasRenderingContext2D }
```
SuperSodaSea, line 120: `_checkDestroyed` can also be wrapped in `// #if _DEBUG`.

packages/utils/src/media/trimCanvas.ts
```diff
@@ -29,7 +29,7 @@ function checkColumn(data: Uint8ClampedArray, width: number, x: number, top: num
  * @param {PIXI.ICanvas} canvas - the canvas to trim
  * @returns {object} Trim data
  */
-export function trimCanvas(canvas: ICanvas): { width: number; height: number; data?: ImageData }
+export function trimCanvas(canvas: ICanvas): {width: number, height: number, data?: ImageData | null }
```
SuperSodaSea, line 32:
```suggestion
export function trimCanvas(canvas: ICanvas): { width: number, height: number, data: ImageData | null }
```

packages/utils/src/path.ts
```diff
@@ -194,7 +194,7 @@ export const path: Path = {
 
         if (isFile || isHttp || isWindows)
         {
-            const arr = isFile?.[0] || isHttp?.[0] || isWindows?.[0];
+            const arr = isFile?.[0] || isHttp?.[0] || isWindows?.[0] || '';
```
SuperSodaSea, line 197:
```suggestion
            const arr = (isFile?.[0] || isHttp?.[0] || isWindows?.[0]) as string;
```

Since it won't be `undefined`.

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -10,10 +10,10 @@ import type { ICanvas, ICanvasRenderingContext2D } from '@pixi/settings';
 export class CanvasRenderTarget
 {
     /** The Canvas object that belongs to this CanvasRenderTarget. */
-    public canvas: ICanvas;
+    public canvas: ICanvas | null;
```
SuperSodaSea, line 13: We can use getter to expose a non-null `canvas`:
```ts
private _canvas: ICanvas | null;

/** The Canvas object that belongs to this CanvasRenderTarget. */
public get canvas(): ICanvas
{
    // #if _DEBUG
    this._checkDestroyed();
    // #endif

    return this._canvas;
}
```

And also for `context`. Internal use of `canvas` and `context` need to be change into `_canvas` and `_context` (and also in `asserts` of `_checkDestroyed`). After this we won't need `?.` in `renderTarget.canvas?.xxx`.

SuperSodaSea (commented):
Maybe I'm a little picky, but the rest of the code looks great!

Emobe:
Hah, don't worry about being picky. It's good to iron this stuff out otherwise it nags for the rest of time :grimacing:
Will sort these out

packages/utils/src/browser/isWebGLSupported.ts
Emobe: Is this supposed to be `!!gl?.getContextAttributes()?.stencil;`

I see there is an extra `gl in `gl?.gl...`

SuperSodaSea:
There is an error when running `tsc` with `--strictNullChecks`:

```
packages/utils/src/path.ts(43,18): error TS2454: Variable 'code' is used before being assigned.
```

We can give `code` an initial value, such as `let code = -1`.

https://github.com/pixijs/pixijs/blob/e8796d7a9d6e2faeee5d398e2ca908473da8ba36/packages/utils/src/path.ts#L35-L43

packages/utils/src/media/CanvasRenderTarget.ts
Emobe: I don't think `_checkDestroyed` is actually a type guard as typescript thinks the code can still reach the return value as it's not wrapped in an if statement or something.

```typescript
public get canvas(): ICanvas
{
    // #if _DEBUG
    this._checkDestroyed();
    // #endif

    return this._canvas;
}
 ```
 
This still returns  `Type 'ICanvas | null' is not assignable to type 'ICanvas'.`

Maybe going back to the original if statement type guard would work?

Comment (SuperSodaSea) [reply packages/utils/src/media/CanvasRenderTarget.ts:13]: Seems that the problem is caused by `private`, change it to `protected` may help.

```diff
- private _canvas: ICanvas | null;
+ protected _canvas: ICanvas | null;
```

Emobe: Still the same issue for me when set as protected.

SuperSodaSea: Did you change `canvas` in `asserts` into `_canvas`?

Emobe:
Ahh yes. The test where it checks it is null should check that an error is thrown. Will get on that now

Emobe:
@SuperSodaSea Slightly unrelated issue. When I try to run the tests locally, it just hangs like this forever.

jestissue

Might open an issue if you don't have an immediate answer because it's kind of unrelated for the PR.

Pushed up a change for the test anyway.

packages/utils/test/CanvasRenderTarget.tests.ts
```diff
@@ -16,7 +16,7 @@ describe('CanvasRenderTarget', () =>
 
         renderTarget.destroy();
 
-        expect(renderTarget.canvas).toBeNull();
+        expect(renderTarget.canvas).toThrowError();
```
SuperSodaSea, line 19:
```suggestion
        expect(() => renderTarget.canvas).toThrow(TypeError);
```

Emobe: :man_facepalming:

Emobe: Doneee

SuperSodaSea:
I made PR #8965, with this PR we can check --strictNullChecks more easily (and also on CI) :)

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -28,9 +26,9 @@ export class CanvasRenderTarget
      */
     constructor(width: number, height: number, resolution?: number)
     {
-        this.canvas = settings.ADAPTER.createCanvas();
+        this._canvas = settings.ADAPTER.createCanvas();
 
-        this.context = this.canvas.getContext('2d');
+        this._context = this.canvas.getContext('2d');
```
SuperSodaSea, line 31:
```suggestion
        this._context = this._canvas.getContext('2d');
```

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -43,8 +41,12 @@ export class CanvasRenderTarget
...
     clear(): void
     {
-        this.context.setTransform(1, 0, 0, 1, 0, 0);
-        this.context.clearRect(0, 0, this.canvas.width, this.canvas.height);
+        // #if _DEBUG
+        this._checkDestroyed();
+        // #endif
+
+        this._context.setTransform(1, 0, 0, 1, 0, 0);
+        this._context.clearRect(0, 0, this.canvas.width, this.canvas.height);
```
SuperSodaSea, line 49:
```suggestion
        this._context.clearRect(0, 0, this._canvas.width, this._canvas.height);
```

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -85,11 +99,49 @@ export class CanvasRenderTarget
      */
     get height(): number
     {
+        // #if _DEBUG
+        this._checkDestroyed();
+        // #endif
+
         return this.canvas.height;
```
SuperSodaSea, line 106:
```suggestion
        return this._canvas.height;
```

packages/utils/src/media/CanvasRenderTarget.ts
```diff
@@ -85,11 +99,49 @@ export class CanvasRenderTarget
...
     }
 
     set height(val: number)
     {
+        // #if _DEBUG
+        this._checkDestroyed();
+        // #endif
+
         this.canvas.height = Math.round(val);
```
SuperSodaSea, line 115:
```suggestion
        this._canvas.height = Math.round(val);
```

SuperSodaSea (approved):
Almost LGTM, let's see other's opinion.

Emobe:
Sounds good. Added those changes now too

bigtimebuddy (approved):

SuperSodaSea:
@bigtimebuddy #8978 will have a conflict with this PR, need to merge that PR first, then handle the conflict manually.

SuperSodaSea:
@bigtimebuddy Conflict resolved, ready to merge.

bigtimebuddy:
Thanks for your had work on this @Emobe. Appreciate the help.

PR #8961, opened 2022-12-09
Loose extensions where not getting prioritized correctly

### LINKED ISSUE #8936: Custom ParserLoaders are not ordered according to the extension's priority, which results in being unable to create a parser that can run before or after other parsers.
I have been getting issues adding a Loader Parser which takes a JSON file, checks if it has certain fields and parses it differently. The issue was that the original `loadJSON` extension would Load and Parse it no matter which priority I put on my own Loader Parser. I took a look at the `Assets.loader.parsers` list and my extension would always be at the end of the list, I needed it first so it could bypass it if the JSON was the one I was looking for.

Taking a look at the parsers list, none of them were ordered according to their priority. In the end, I resorted to manually sorting the parsers to fix my issue, like this:

```typescript
        PIXI.Assets.loader.parsers.sort((a, b) => {
            const aValue: number = typeof a.extension == "object" ? a.extension.priority : LoaderParserPriority.Normal;
            const bValue: number = typeof b.extension == "object" ? b.extension.priority : LoaderParserPriority.Normal;
            return Math.sign(bValue - aValue);
        });
```
 

### Expected Behavior

adding an extension with a certain priority should then run before or after other priorities

### Current Behavior

It always adds new loader parser extensions to the end of the Asset.loader.parser 's array.

### Possible Solution

Make sure to sort the parser's array when inserting any new values to it.

### Steps to Reproduce

1. Copy the example I gave below to a script file and try running it.
2. Notice that the JSON it returns is the package.json file, it should be intercepted by the LoaderParser due to the higher priority.
3. Try uncommenting the workaround I included, the LoaderParser will now work correctly.

### Environment

- **`pixi.js` version**:  7.0.4
- **Browser & Version**: Edge
- **Running Example**: 

```typescript
import "./styles.css";
import {
  Application,
  Assets,
  extensions,
} from "pixi.js";
import type { LoadAsset, Loader } from "pixi.js";
import { ExtensionType, LoaderParser, LoaderParserPriority } from "pixi.js/dist/pixi.mjs";


export const loadMyAsset: LoaderParser = {
  extension: {
    name: "Custom Loader Parser",
    priority: LoaderParserPriority.Normal, // Actually will be last priority according to the console.log
    type: ExtensionType.LoadParser,
  },
  test(url: string) { return url.endsWith(".json"); },
  async load(url: string, asset: LoadAsset, loader: Loader) { return { request: url } }
} as LoaderParser;

extensions.add(loadMyAsset);

const app = new Application<HTMLCanvasElement>({
  resizeTo: window,
  background: "#999"
});

async function main() {

  Assets.init();

  // Workaround fix, uncomment to see how it should be
  /*
  Assets.loader.parsers.sort((a, b) => {
    const aValue: number | undefined = typeof a?.extension === "object" ? a?.extension?.priority : LoaderParserPriority.Normal;
    const bValue: number | undefined = typeof b?.extension === "object" ? b?.extension?.priority : LoaderParserPriority.Normal;
    return Math.sign(bValue! - aValue!);
  });
  */

  const json = await Assets.load(
    "https://raw.githubusercontent.com/pixijs/pixijs/dev/package.json"
  );
  console.log(json); // This should have the following json: { request: url }
  console.log(Assets.loader.parsers); // You can see this parser being last in the list
  // and you can see that the Parsers list is not sorted at all by priority
}

main();

```

bigtimebuddy:
Closes #8936 

**Broken (7.0.4)**: https://jsfiddle.net/bigtimebuddy/u68c0xLq/
**Fixed (this PR)**: https://jsfiddle.net/bigtimebuddy/0x21bjcv/


packages/extensions/src/index.ts
```diff
@@ -102,6 +102,16 @@ const normalizeExtension = (ext: ExtensionFormatLoose | any): ExtensionFormat =>
...
 };
 
+/**
+ * Get the priority for an extension.
+ * @ignore
+ * @param ext - Any extension
+ * @param defaultPriority - Fallback priority if none is defined.
+ * @returns The priority for the extension.
+ */
+const normalizePriority = (ext: ExtensionFormatLoose | any, defaultPriority: number): number =>
```
SuperSodaSea, line 112: Just an advice, for top level functions `normalizePriority` and `normalizeExtension`, I prefer `function xxx(...) { ... }` to `const xxx = (...) => { ... }`.

PR #8968, opened 2022-12-12
Add defaultSearchParams option to Assets

GoodBoyDigital:
##### Description of change
Adds support for default url params via the init function.

```
await Assets.init({
            defaultSearchParams: {
                foo: 'bar',
                chicken: 'nuggets',
            },
        });

const bunnyTexture = await Assets.load('bunny.png'); // url = bunny.png?foo=bar&chicken=nuggets
```

##### Pre-Merge Checklist

### DIFF
--- a/packages/assets/src/Assets.ts
+++ b/packages/assets/src/Assets.ts
@@ -24,7 +24,7 @@ export interface AssetInitOptions
     basePath?: string;
 
     /** a default URL parameter string to append to all assets loaded */
-    defaultUrlParameters?: string | Record<string, any>;
+    defaultSearchParams?: string | Record<string, any>;
 
     /**
      * a manifest to tell the asset loader upfront what all your assets are
@@ -256,9 +256,9 @@ export class AssetsClass
 
         this._initialized = true;
 
-        if (options.defaultUrlParameters)
+        if (options.defaultSearchParams)
         {
-            this.resolver.setUrlParameters(options.defaultUrlParameters);
+            this.resolver.setDefaultSearchParams(options.defaultSearchParams);
         }
 
         if (options.basePath)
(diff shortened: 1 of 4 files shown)


GoodBoyDigital:
Oh, i also branched this from PR @Zyie as you already assed the code to strip out url params!

bigtimebuddy:
Can we align around `searchParams` (e.g., `defaultSearchParams`) as a name for this API? There are lots of names for these like queryString, parameters, urlParams, etc, but my preference is to align with the URL object. See:

* https://nodejs.org/api/url.html#urlsearchparams
* https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

GoodBoyDigital:
> Can we align around `searchParams` (e.g., `defaultSearchParams`) as a name for this API? There are lots of names for these like queryString, parameters, urlParams, etc, but my preference is to align with the URL object. See:
> 
> * https://nodejs.org/api/url.html#urlsearchparams
> * https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

good shout! I have renamed to use `searchParams` and relevant variants in the PR.

bigtimebuddy (approved):
Great work @GoodBoyDigital

SuperSodaSea (commented):
One thing that worries me is whether we need to deal with duplicate parameters, such as:

```js
resolver.setDefaultSearchParams('lucky=23');

expect(resolver.resolveUrl('my-image.png?lucky=45')).toBe(/* ??? */);
// 'my-image.png?lucky=23'?
// 'my-image.png?lucky=45'?
// 'my-image.png?lucky=45&lucky=23'?
```

bigtimebuddy:
Duplicate search parameters is not part of the [URI spec](https://www.rfc-editor.org/rfc/rfc3986#section-3.4). Servers can interpret this however they want. While it would be nice, I don't think we should be sanitizing duplicate parameters. This problem should be up to users.

SuperSodaSea (approved):

PR #8972, opened 2022-12-13
Fix: WheelEvent.deltaMode consistency

### LINKED ISSUE #8970: `WheelEvent.deltaMode` changed with Pixi7 only on FireFox
We use a `"wheel"` event listener directly on the `<canvas>` (for UI zooming).

When upgrading from PIXI6->7 we noticed that the wheel-handler (only) on FireFox gets different event settings.
`deltaMode` changed from `0` (`DOM_DELTA_PIXEL`) with PIXI6 to `1` (`DOM_DELTA_LINE`) with PIXI7.

The listener seems not to be connected to PIXI at all? Why only on FireFox? We cannot spot the reason.

```
    width: window.innerWidth,
    height: window.innerHeight,
    backgroundColor: 0x2c3e50
});

console.log("version", PIXI.VERSION);

app.view.addEventListener("wheel", (e) => {
    console.log(e.deltaY, e.deltaMode);
});
```

### Possible Solution
Evaluating `deltaMode` and scaling `deltaY` by a heuristic constant works around the effect, but we'd love to understand the reason.

### Steps to Reproduce
Run the code above in the PIXI playground and switch PIXI version in the settings.
Observe console log in the DevTools when scrolling with the mouse wheel.

*Note* that it only holds when using a mouse-wheel; using the touchpad also triggers the listener but does not show the effect (`deltaMode` is always `0`). 

### Environment

- **`pixi.js` version**: 7.x vs. 6.x
- **Browser & Version**: FireFox 107.0.1 (64-Bit)
- **OS & Version**: seen on MacOS 13.0.1
- **Running Example**: https://www.pixiplayground.com/#/edit/lSDZVVwcTjAf0p1AruozD

SuperSodaSea:
#### Description of change

When `WheelEvent` is triggered by scrolling with mouse wheel, reading `WheelEvent.deltaMode` before `deltaX`/`deltaY`/`deltaZ` on Firefox will result in `WheelEvent.DOM_DELTA_LINE (1)`, while reading `WheelEvent.deltaMode` after `deltaX`/`deltaY`/`deltaZ` on Firefox or reading in any order on other browsers will result in `WheelEvent.DOM_DELTA_PIXEL (0)`.
Therefore, we need to read `WheelEvent.deltaMode` after `deltaX`/`deltaY`/`deltaZ` in order to make its behavior more consistent across browsers.

Fixes #8970.

Example (Use Firefox, scroll the mouse wheel on the right and see result in console):
- **Before**: <https://www.pixiplayground.com/#/edit/-0JdGEPsmMGlNkGUYw0ed>
- **After**: <https://www.pixiplayground.com/#/edit/03E9XWNqQPJMSXzoIfdbb>

#### Pre-Merge Checklist

### DIFF
--- a/packages/events/src/EventSystem.ts
+++ b/packages/events/src/EventSystem.ts
@@ -592,10 +592,17 @@ export class EventSystem
 
         this.transferMouseData(event, nativeEvent);
 
-        event.deltaMode = nativeEvent.deltaMode;
+        // When WheelEvent is triggered by scrolling with mouse wheel, reading WheelEvent.deltaMode
+        // before deltaX/deltaY/deltaZ will result in deltaMode === WheelEvent.DOM_DELTA_LINE (1)
+        // on Firefox, while reading deltaMode after deltaX/deltaY/deltaZ or reading in any order
+        // on other browsers will result in deltaMode === WheelEvent.DOM_DELTA_PIXEL (0).
+        // Therefore, we need to read deltaMode after deltaX/deltaY/deltaZ in order to make its
+        // behavior more consistent across browsers.
+        // @see https://github.com/pixijs/pixijs/issues/8970
         event.deltaX = nativeEvent.deltaX;
         event.deltaY = nativeEvent.deltaY;
         event.deltaZ = nativeEvent.deltaZ;
+        event.deltaMode = nativeEvent.deltaMode;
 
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):
Nice fix, verified that this work :+1:

PR #8978, opened 2022-12-15
Add method to get canvas bounding box to media utils

vikpe:
##### Description of change
Add/expose method for measuring canvas content to media utils (extracted from `trimCanvas()`).

##### Pre-Merge Checklist


vikpe:
Could possibly be used to refactor [`TextMetrics.measureFont()`](https://github.com/pixijs/pixijs/blob/c8d49437a452a3bd6b0d69e6f8b664a6d89436fa/packages/text/src/TextMetrics.ts#L798).

packages/utils/src/media/measureCanvasContent.ts
```diff
@@ -0,0 +1,80 @@
...
+        width: number;
+        height: number;
+    }
+    bounds: {
+        top: number;
+        bottom: number;
+        left: number;
+        right: number;
+    }
+}
```
bigtimebuddy, line 44: Would be nice to make a type here for the result. Also, I'd suggest flattening this, the `size` and `bounds` levels aren't really necessary and this is not an interface that's likely to expand at all so that structure just kinda gets in the way.

Having as a type means that we can document it:

```ts
/**
 * Result when using `utils.measureCanvasContent` to determine
 * the bounds of a canavas' visual content excluding transparency.
 * @memberof PIXI.utils
 */
interface CanvasContent {
    /** Total width in pixels of the canvas */
    width: number;
    /** Total height in pixels of the canvas */
    height: number;
    /** Offset from the top, in pixels, where the canvas' content begins */
    top: number;
    /** Offset from the bottom, in pixels, where the canvas' content begins */
    bottom: number;
    /** Offset from the left, in pixels, where the canvas' content begins */
    left: number;
    /** Offset from the right, in pixels, where the canvas' content begins */
    right: number;
}
```

bigtimebuddy:
Great idea! Helpful utility, I could see this being useful for folks.

packages/utils/src/media/measureCanvasContent.ts
vikpe: Updated!

SuperSodaSea: I am not very satisfied with the wording `Offset from the bottom, in pixels, where the canvas' content begins`, it is more like `height - bottom`.

Also the name `CanvasContent` is not very good in my opinion, since it only contains the pixel boundary information of the canvas, not the real "content".

Or maybe we can use the existing class `Rectangle` in `@pixi/math`?

SuperSodaSea:
By the way, adding `@since 7.1.0` in `measureCanvasContent`'s comment can be helpfel.

packages/utils/src/media/measureCanvasContent.ts
vikpe: feel free to rename/refactor to whatever you see fit.

bigtimebuddy: @SuperSodaSea I'm cool if you want to take a pass on the naming/types. I'm not super opinionated about it. Just having a type that's flatter and documented is my preference.

vikpe: perhaps call it `getCanvasBoundingBox()` returning a `BoundingBox`.

related:
* [Minimum bounding rectangle](https://en.wikipedia.org/wiki/Minimum_bounding_rectangle)
* [Element.getBoundingClientRect()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)

SuperSodaSea:
https://github.com/pixijs/pixijs/tree/pr/8978

I did some refactoring here: d3e7afdc942e852f4f1fe5793f050c1cc365b4de...0f51f56bb30295b9c9d25733e5bb59cdd367ab78
- `measureCanvasContent` -> [`getCanvasBoundingBox`](https://github.com/pixijs/pixijs/blob/0f51f56bb30295b9c9d25733e5bb59cdd367ab78/packages/utils/src/media/getCanvasBoundingBox.ts)
- `interface CanvasContent` -> [`class BoundingBox`](https://github.com/pixijs/pixijs/blob/0f51f56bb30295b9c9d25733e5bb59cdd367ab78/packages/utils/src/media/BoundingBox.ts)
- `right` and `bottom` in `BoundingBox` are now exclusive (`right === left + width`, not `left + width - 1`), which is more common
- Changed some documents

bigtimebuddy (approved):
LGTM!

SuperSodaSea (approved):

PR #9020, opened 2022-12-26
Add CDN Install section in READMEs for browser builds

SuperSodaSea:
#### Description of change

Follow up #8827. Users may be confused about how to use browser build version of plugins (#9019), so I add examples for it in package's `README.md`. This will show up on the <https://npmjs.com> pages, and it can be helpful for users.

(The CDN URL in `/README.md` was wrong, caused by the different directory structure between v6 and v7. Now it is fixed.)

#### Pre-Merge Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ npm install pixi.js
 There is no default export. The correct way to import PixiJS is:
 
 ```js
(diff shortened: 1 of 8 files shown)


bigtimebuddy (approved):
Good one, thank you!

PR #9039, opened 2023-01-04
Deprecate filters namespace

bigtimebuddy:
An issue came up recently using `importmap` with `pixi-filters` (see https://github.com/pixijs/filters/pull/365) that is fixed by this PR.

We created this synthetic namespace `filters` for adding all the display object filters. We have generally removed all of these namespace from project except for filters. Removing filters seems like a good idea and consistent with recent changes.

### Changed

Deprecate the `filters` namespace export for the bundles:

* `pixi.js`
* `pixi.js-legacy`
* `@pixi/node`
* `@pixi/webworker`

### Links

* [API Documentation](https://pixijs.download/deprecate-filters-namespace/docs/PIXI.filters.html)
* [Example](https://jsfiddle.net/bigtimebuddy/15qbzo6e/)


Julien-Marcou (approved):
LGTM!

PR #9047, opened 2023-01-06
Add util method to parse multiple color values into a normalized RGBA array

vikpe:
##### Description of change
Add `color2rgba()`, a method to parse multiple color inputs into a normalized RGBA array. Built on top of [colord](https://github.com/omgovich/colord) (zero dependencies, fast, 1.7 KB gzipped).

**Color** is defined as:
```ts
export type Color = string | number | number[] | Float32Array
| CmykColor | CmykaColor | HslColor | HslaColor | HsvColor | HsvaColor | RgbColor | RgbaColor;
```

- `string` = any CSS color value (color name, #hex, rgb(), hsl(), etc)
- `number` = hex number (`0xff0000`)
- `number[]` and `Float32Array` = array of rgb values in range [0-255]

and the color types are objects defined as:
```js
export type HslColor = {
  h: number;
  s: number;
  l: number;
};
```
(the `*a` versions, such as `HslaColor` follow the same structure but include `a` (alpha) in range `[0-1]`). See [colord](https://github.com/omgovich/colord/blob/master/src/types.ts) for all color type definitions.

**Example usage**
```js
import { utils } from 'pixi.js';
utils.color2rgba('red'); // [1, 0, 0, 1]
utils.color2rgba(0xff0000); // [1, 0, 0, 1]
utils.color2rgba('ff0000'); // [1, 0, 0, 1]
utils.color2rgba('#f00'); // [1, 0, 0, 1]
utils.color2rgba([255, 0, 0, 0.5]); // [1, 0, 0, 0.5]
utils.color2rgba('rgb(255, 0, 0, 0.5)'); // [1, 0, 0, 0.5]
utils.color2rgba({h: 0, s: 100, l: 50, a: 0.5}); // [1, 0, 0, 0.5]
utils.color2rgba({h: 0, s: 100, v: 100, a: 0.5}); // [1, 0, 0, 0.5]
utils.color2rgba({c: 0, m: 100, y: 100, k: 0, a: 0.5}); // [1, 0, 0, 0.5]
```

##### Pre-Merge Checklist


vikpe:
Support for `HWB`, `LCH`, `LAB` and `XYZ` can also be added if asked for.

packages/utils/src/color/color.ts
```diff
@@ -0,0 +1,90 @@
+import { colord, extend } from 'colord';
+import cmykPlugin from 'colord/plugins/cmyk';
+import namesPlugin from 'colord/plugins/names';
+
+import type { AnyColor } from 'colord';
+import type { Color } from '@pixi/utils';
```
bigtimebuddy, line 6: Could you please not import `@pixi/utils` from itself? It can create build warnings for us unnecessarily. Better to import relatively.

packages/utils/src/color/color.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @returns {number[]} [R,G,B,A] array of normalized values
+ */
+export function color2rgba(value: Color): number[]
+{
+    if (Array.isArray(value) && value.length >= 3)
+    {
+        const r = value[0] / 255;
+        const g = value[1] / 255;
+        const b = value[2] / 255;
+        const a = value.length > 3 ? value[3] : 1.0;
```
bigtimebuddy, line 36: Alternatively:
```suggestion
        const a = value[3] ?? 1.0;
```

packages/utils/src/color/color.ts
```diff
@@ -0,0 +1,90 @@
+import { colord, extend } from 'colord';
+import cmykPlugin from 'colord/plugins/cmyk';
+import namesPlugin from 'colord/plugins/names';
+
+import type { AnyColor } from 'colord';
+import type { Color } from '@pixi/utils';
+
+extend([cmykPlugin, namesPlugin]);
```
bigtimebuddy, line 8: I don't think CMYK is necessary for us. WDYT?

bigtimebuddy (commented):
I've thought about this a little bit more, specifically about our conversation about a util vs a new Color class... seeing this, I'm rethinking a Color class. 

I'd like to be able to deprecate all the hex util functions. We should remove `css-color-names` dependency since that's likely doing the same thing as the `namesPlugin` of `colord` and it's probably just adding extra overhead. Also, it's confusing having a lot of random color util functions, and these names don't really align with a lot of the other API (e.g., hex2string, hex2rgb, etc) 

We could do a Color class that combines `color2rgba` (that normalizes the input), the ability to update, and various output functions (for to integer, hex string, rgba, etc). This seems more scalable and flexible. And I see a few opportunities internally where we might use this new object to store persistent colors like tint, text fill, renderer background color, etc.

SuperSodaSea:
@bigtimebuddy I'm also thinking about a new `Color` class recently. For example, `IRendererOptions.backgroundColor` accepts `number | string`, but `Renderer.background.color` only accepts `number`, this can be confusing for users (see #9043). Adding a new `Color` class can be helpful to resolve such inconsistencies, in this way we don't need to call `hex2rgb`, `hex2string`, `rgb2hex` or `string2hex` everywhere.

vikpe:
My main motivation is to have a pixi endorsed color type/util with multiple color space support. For example to align color params in pixi filters.

I'd prefer a color class, with getters and common operations, something like:

```js
const  mycolor = new Color(any);

// color spaces/getters
mycolor.rgb/a
mycolor.cmyk/a
mycolor.hex/a
mycolor.hsl/a
mycolor.hsv/a
mycolor.alpha

// operations
mycolor.mix(otherColor, amount);
mycolor.add(otherColor);
mycolor.subtract(otherColor);
mycolor.multiply(otherColor);
mycolor.divide(otherColor);
mycolor.normalize();
mycolor.grayscale();
mycolor.invert();
mycolor.tint(amount); // mix with white
mycolor.shade(amount); // mix with black 
mycolor.tone(amount); // mix with gray
```

In a filter it would look something like
```js
this.uniforms.uColor = new Color(options.color).normalize().rgba;
```

bigtimebuddy:
@vikpe I like that API and getter options. I probably wouldn't add all those operators unless they are things we can use internally. Otherwise, I think that code is better handled as an optional mixin (e.g., `math-extras`).

Also, we should have a `value` setter same as the constructor arg:

```ts
const mycolor = new Color(any);
mycolor.value = any;
```

SuperSodaSea:
We can rename the `type Color` to something like `type ColorSource`:

```ts
export type ColorSource = string | number | number[] | Float32Array
    | XxxxColor
    | Color; // Should also accept class Color itself

export class Color {
    constructor(value: ColorSource) { /* ... */ }
    set value(value: ColorSource) { /* ... */ }
    // ...
}
```

vikpe:
Ye perhaps the operations is a bit overkill for the color class and can be a separate package.

`normalize` applies to all color spaces and is commonly used so perhaps include it.

vikpe:
btw I'm away for 2 weeks without a computer. I opened the pull request because @bigtimebuddy asked me to 😄

Feel free to close/refactor/commit changes. Should be pretty easy to convert it into a class constructor. Then add some additional tests for the color space getters.

packages/utils/src/color/color.ts
vikpe: Perhaps, but the the cost of supporting cmyk is very small.

bigtimebuddy:
Thanks @vikpe for being flexible.

@SuperSodaSea is this something your like to polish up? If not, I'm happy to do it.

SuperSodaSea:
@bigtimebuddy I have no new idea now, so do whatever you want :)

PR #9051, opened 2023-01-11
Migrate loseContext/restoreContext fix #8835 to v6

yordan-kanchelov:
I am not sure if v6 is still maintained, but felt like the fix from @dev7355608  #8835 could come in handy for the current v6 projects.


SuperSodaSea:
The GitHub Actions are blocked on v6.x branch, I'm fixing it in #9054.

SuperSodaSea (approved):
LGTM, since it looks exactly the same as #8835.

bigtimebuddy:
Thank you @yordan-kanchelov

PR #9055, opened 2023-01-11
Fix defaultSearchParams to append to Spritesheets or BitmapFonts

### LINKED ISSUE #9052: Bug: Assets which load other assets, such as spritesheets, do not append defaultSearchParams to the assets they're loading
### Current Behavior

When you load an asset data file, such as a spritesheet JSON or a bitmap XML file, it will attempt to load the assets listed in there, such as the spritesheet PNG. Problem is, is that it doesn't load it with the defaultSearchParams appended to it.

Here is a network log example, my-numbers.json is a multi spritesheet file.

image

```typescript
  PIXI.Assets.init({
    defaultSearchParams: { version: "1.0.0" },
    manifest: {
      bundles: [
        {
          name: "test_bundle",
          assets: [
            {
              name: "spritesheet",
              srcs: "assets/all-my-numbers.json", // Will load its related image file without the version param
            },
            {
              name: "spritesheet-multi",
              srcs: "assets/my-numbers-0.json", // Will load other spritesheets through the relative_multi property without the version param
            },
            {
              name: "bitmapfont",
              srcs: "assets/bonus_special_highlight_export.xml", // Will load the font image related to this XML without the version param
            },
          ],
        },
      ],
    },
  });

  PIXI.Assets.loadBundle("test_bundle");
```

### Expected Behavior

The network requests should be appending the version to all requests being made by the Assets system.

### Steps to Reproduce

1. Open https://www.pixiplayground.com/#/edit/TEmwE_ExPjqNlkr-NboRO
2. Open the network log
3. Refresh the game
4. See that it append the `version` param to the bitmap font's data file but not to the PNG.

OR

1. Download, extract, and install the example NPM project: [defaultSearchParamIssue.zip](https://github.com/pixijs/pixijs/files/10391882/defaultSearchParamIssue.zip)
5. Run `npm run local:buildRun` to build and run the project.
6. Open the browser and open the network tab to note that the defaultSearchParam is not appended to everything loaded.

### Environment

- **`pixi.js` version**: *7.1.0*
- **Browser & Version**: *All Latest Browsers*
- **Running Example**: *https://www.pixiplayground.com/#/edit/TEmwE_ExPjqNlkr-NboRO & [defaultSearchParamIssue.zip](https://github.com/pixijs/pixijs/files/10391882/defaultSearchParamIssue.zip)*

### Possible Solution

Maybe the resolver should not be handling the defaultSearchParam and it should be appended to every single URL that is used to fetch assets.

### Additional Information

_No response_

GoodBoyDigital:
Appends defaultSearchParams to the assets they're loading by spritesheet or bitmapfont

### DIFF
--- a/packages/assets/src/utils/copySearchParams.ts
+++ b/packages/assets/src/utils/copySearchParams.ts
@@ -1,0 +1,17 @@
+/**
+ * Copies the search params from one url to another
+ * @param targetUrl - the url to copy the search params to
+ * @param sourceUrl - the url container the sear params we want to copy
+ * @returns the url with the search params copied
+ */
+export const copySearchParams = (targetUrl: string, sourceUrl: string) =>
+{
+    const searchParams = sourceUrl.split('?')[1];
+
+    if (searchParams)
+    {
+        targetUrl += `?${searchParams}`;
+    }
+
+    return targetUrl;
+};
--- a/packages/assets/src/utils/copySearchParams.ts
+++ b/packages/assets/src/utils/copySearchParams.ts
@@ -1,7 +1,7 @@
 /**
  * Copies the search params from one url to another
  * @param targetUrl - the url to copy the search params to
- * @param sourceUrl - the url container the sear params we want to copy
+ * @param sourceUrl - the url container the search params we want to copy
  * @returns the url with the search params copied
  */
 export const copySearchParams = (targetUrl: string, sourceUrl: string) =>
--- a/packages/spritesheet/src/spritesheetAsset.ts
+++ b/packages/spritesheet/src/spritesheetAsset.ts
@@ -152,7 +152,7 @@ export const spritesheetAsset = {
 
                     if (searchParams)
                     {
-                        itemUrl = `${itemUrl}?${searchParams}`;
+                        itemUrl += `?${searchParams}`;
                     }
 
                     promises.push(loader.load<Spritesheet>({
--- a/packages/spritesheet/src/spritesheetAsset.ts
+++ b/packages/spritesheet/src/spritesheetAsset.ts
@@ -113,7 +113,7 @@ export const spritesheetAsset = {
 
             if (searchParams)
             {
-                imagePath = `${imagePath}?${searchParams}`;
+                imagePath += `?${searchParams}`;
             }
 
             const assets = await loader.load<Texture>([imagePath]);
--- a/packages/spritesheet/src/spritesheetAsset.ts
+++ b/packages/spritesheet/src/spritesheetAsset.ts
@@ -1,4 +1,4 @@
-import { LoaderParserPriority } from '@pixi/assets';
+import { copySearchParams, LoaderParserPriority } from '@pixi/assets';
 import { extensions, ExtensionType, settings, utils } from '@pixi/core';
 import { Spritesheet } from './Spritesheet';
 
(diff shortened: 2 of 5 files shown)


packages/spritesheet/src/spritesheetAsset.ts
```diff
@@ -134,14 +142,19 @@ export const spritesheetAsset = {
...
 
                     // Check if the file wasn't already added as multipack
                     if (options.data?.ignoreMultiPack)
                     {
                         continue;
                     }
 
+                    if (searchParams)
+                    {
+                        itemUrl = `${itemUrl}?${searchParams}`;
```
bigtimebuddy, line 155:
```suggestion
                        itemUrl += `?${searchParams}`;
```

packages/text-bitmap/src/loadBitmapFont.ts
```diff
@@ -36,10 +36,17 @@ export const loadBitmapFont = {
...
+
         for (let i = 0; i < pages.length; ++i)
         {
             const pageFile = pages[i].file;
-            const imagePath = utils.path.join(utils.path.dirname(src), pageFile);
+            let imagePath = utils.path.join(utils.path.dirname(src), pageFile);
+
+            if (searchParams)
+            {
+                imagePath = `${imagePath}?${searchParams}`;
```
bigtimebuddy, line 48:
```suggestion
                imagePath += `?${searchParams}`;
```

packages/spritesheet/src/spritesheetAsset.ts
```diff
@@ -107,7 +107,15 @@ export const spritesheetAsset = {
...
             }
 
-            const imagePath = basePath + asset.meta.image;
+            const searchParams = options.src.split('?')[1];
+
+            let imagePath = basePath + asset.meta.image;
+
+            if (searchParams)
+            {
+                imagePath = `${imagePath}?${searchParams}`;
```
bigtimebuddy, line 116:
```suggestion
                imagePath += `?${searchParams}`;
```

bigtimebuddy (approved):
This works! I think the only area where I think we could improve would be to create a util. When we get into concatenating strings with `?` things can getting a little hairy. At least having this happen in one place means we can re-use it on other things (e.g., spine assets).

```ts
const copySearchParams = (target, source) => {
  const searchParams = source.split('?')[1];

  if (searchParams) {
    target += `?${searchParams}`;
  }
  return target;
};

**Comment by GoodBoyDigital** [comment]
> This works! I think the only area where I think we could improve would be to create a util. When we get into concatenating strings with `?` things can getting a little hairy. At least having this happen in one place means we can re-use it on other things (e.g., spine assets).
> 
> ```ts
> const copySearchParams = (target, source) => {
>   const searchParams = source.split('?')[1];
> 
>   if (searchParams) {
>     target += `?${searchParams}`;
>   }
>   return target;
> };
> ```

I did think about that, for some reason i was sad about splitting the string multiple times. Its not like this is hot code though, so added the function you suggested. This is much neater now!

File: packages/assets/src/utils/copySearchParams.ts [hunk 4]
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copies the search params from one url to another
+ * @param targetUrl - the url to copy the search params to
+ * @param sourceUrl - the url container the sear params we want to copy
```
bigtimebuddy, line 4:
```suggestion
 * @param sourceUrl - the url container the search params we want to copy
```

bigtimebuddy (approved):

PR #9056, opened 2023-01-11
Add missing data pass-through for bundle assets

### LINKED ISSUE #9053: Feature Request: Add the ability to add metadata to assets in bundles like you can with PIXI.Assets.add
### Description

Currently you can add meta data to the Assets you want to load with `PIXI.Assets.add`:

```typescript
//                                      v Meta-data
PIXI.Assets.add("hero", "player.png", { animations: ["jump", "run", "attack"] } });
```

This is useful for passing data about the file to the related loader, such as adding a list of different animations a spritesheet has or for embedding a Spine .atlas file.

But sadly for bundles there is no way to add meta-data, it would be most useful if there was a new optional property for `ResolverAssetsArray` that would let you pass in meta data from the manifest itself. This will allow people to customise their manifests more to suit their development needs.

Here is what it could look like:

```typescript
const manifest = {
    bundles: [
        {
            name: "game-screen",
            assets: [
                {
                    name: "character",
                    srcs: "robot.png",
                    data: { // New property
                        animations: ["run", "jump", "shoot"]
                    }
                },
            ]
        }
    ]
};
```

GoodBoyDigital:
data now being passed correctly!

test added

fixes #9053

### DIFF
--- a/packages/assets/src/resolver/Resolver.ts
+++ b/packages/assets/src/resolver/Resolver.ts
@@ -315,7 +315,8 @@ export class Resolver
                     const bundleAssetId = this._createBundleAssetId(bundleId, asset.name);
 
                     assetNames.push(bundleAssetId);
-                    this.add([asset.name, bundleAssetId], asset.srcs);
+
+                    this.add([asset.name, bundleAssetId], asset.srcs, asset.data);
                 }
                 else
                 {
--- a/packages/assets/src/resolver/types.ts
+++ b/packages/assets/src/resolver/types.ts
@@ -20,13 +20,14 @@ export interface PreferOrder
  */
 export interface ResolveAsset extends Record<string, any>
 {
-    alias?: string[];
     src: string;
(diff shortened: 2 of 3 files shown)


bigtimebuddy (approved):

miltoncandelero:
This is a part of my locale files parser, what would/should happen if the bundle already had data inside it? 🤔 

```ts
		parse: (value: string): ResolveAsset => {
			const filename = value.split("?")[0].split("/").pop();
			const splitName = filename.split(".").reverse(); // json , i18, locale code, namespace if any

			return {
				data: {
					lng: splitName[2] ?? "en",
					ns: splitName[3] ?? "translation",
				},
				src: value,
			};
		},
```

bigtimebuddy:
@GoodBoyDigital is there anything else to do here to follow-up with @miltoncandelero's comment? Maybe we need a test for that?

PR #9061, opened 2023-01-13
Add new @pixi/color package and Color class

bigtimebuddy:
This PR replaces #9047.

The goal of this PR is to create a normalized utility for managing colors. See [three.js' Color](https://threejs.org/docs/#api/en/math/Color) for reference. This would allow us to create a persistent container for color, especially for things like [filters](https://github.com/pixijs/filters), which readily use various color inputs.

This Color class supports color inputs types, such as:

* Color names (e.g., `"red"`, `"white"`, `"black"`)
* Hex integers (e.g., `0xff0000`, `0xffffff`, `0x000000`)
* Hex integer strings (e.g., `"0xff0000"`, `"0xffffff"`, `"0x000000"`)
* Hex strings (e.g., `"#ff0000"`, `"#ffffff"`, `"#000000"`)
* Hex string w/ alpha (e.g., `"#ff0000ff"`, `"#ffffffff"`, `"#0000000ff"`)
* RGBA objects (e.g., `{ r: 255, g: 0, b: 0 }`)
* RGBA strings (e.g., `"rgb(255, 0, 0)"`, `"rgba(100% 0% 0% / 50%)"`)
* RGBA arrays (e.g., `[1, 0, 0]`, `[1, 1, 1]`, `[0, 0, 0]`)
* RGBA Float32Array (e.g., `new Float32Array([1, 0, 0])`)
* RGBA Uint8Array (e.g., `new Uint8Array([255, 0, 0])`)
* RGBA Uint8ClampedArray (e.g., `new Uint8ClampedArray([255, 0, 0])`)
* HSL objects (e.g., `{ h: 0, s: 100, l: 50 }`)
* HSL strings (e.g., `"hsl(0, 100%, 50%)"`)
* HSV objects (e.g., `{ h: 0, s: 100, v: 100 }`)

### Deprecates

* `utils.rgb2hex` => `new Color(rgb).toNumber()`
* `utils.string2hex` => `new Color(string).toNumber()`
* `utils.hex2string` => `new Color(hex).toHex()`
* `utils.hex2rgb` => `new Color(hex).toRgbArray()`

### Added

* `renderer.background.color` now supports types other than number, similar to `background` options.

### Chores

* Clean up internal deprecations for color manipulations

## Demo

* [Dynamically change background.color](https://jsfiddle.net/bigtimebuddy/gax58kom/)

## Follow-up Changes

* [ ] Allow uniforms to accept Color objects (`uniforms.color = new Color()`)
* [ ] Change tint to accept ColorSource (`sprite.tint = 'red'`)
* [x] Change Graphics fill/stroke to accept ColorSource #9087


packages/utils/src/color/Color.ts
```diff
@@ -0,0 +1,305 @@
...
+            if (typeof value === 'string')
+            {
+                if (value.startsWith('0x'))
+                {
+                    value = value.slice(2);
+                }
+                if (Color.HEX_PATTERN.test(value) && !value.startsWith('#'))
+                {
+                    value = `#${value}`;
+                }
```
SuperSodaSea, line 268:
```suggestion
                if (value.startsWith('0x'))
                {
                    value = `#${value.slice(2)}`;
                }
                else if (Color.HEX_PATTERN.test(value))
                {
                    value = `#${value}`;
                }
```

Original code may accept string like `0x#ffffff`, which I think should be rejected.

Also because the `#` is removed from the regex, we don't need to check it any more.

packages/utils/src/color/Color.ts
```diff
@@ -0,0 +1,305 @@
...
+     * @type {PIXI.ColorSource}
+     */
+    set value(value: ColorSource)
+    {
+        if (this._value !== value)
+        {
+            this._value = value;
+            this.normalize();
+        }
+    }
```
SuperSodaSea, line 88: If `this.normalize()` fails, the `class Color` will be in an invalid state. I prefer not changing `_value` when error happens.

packages/utils/src/color/Color.ts
```diff
@@ -0,0 +1,305 @@
...
+    HsvColor,
+    RgbaColor,
+    RgbColor,
+} from 'colord/types';
+
+extend([namesPlugin]);
+
+/** Different possible value types for Color class */
+export type ColorSource = string | number | number[] | Float32Array
+| HslColor | HslaColor | HsvColor | HsvaColor | RgbColor | RgbaColor;
```
SuperSodaSea, line 18: `Color` can also be a `ColorSource`.

packages/utils/src/color/Color.ts
```diff
@@ -0,0 +1,305 @@
...
+     */
+    setAlpha(alpha: number): this
+    {
+        this._components[3] = alpha;
+
+        return this;
+    }
+
+    /** Get the alpha amount */
+    toAlpha(): number
```
SuperSodaSea, line 197: Maybe `getAlpha`?

SuperSodaSea (commented):
"Add suggestion to batch" can be use to avoid making too much commits.

packages/utils/src/color/Color.ts
```diff
@@ -0,0 +1,305 @@
...
+    /**
+     * Default Color object for static uses
+     * @example
+     * import { utils } from 'pixi.js';
+     * utils.Color.default.setValue(0xffffff).toString(); // '#ffffff'
+     */
+    static readonly default = new Color();
+
+    /** Pattern for hex strings */
+    static readonly HEX_PATTERN = /^#?([a-f0-9]{3}){1,2}([a-f0-9]{2})?$/i;
```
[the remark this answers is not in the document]
bigtimebuddy: This pattern `/^[0-9a-f]{3,8}$/i` would allow for `fffff` (5 chars) or `fffffff` (7 chars), but the original doesn't.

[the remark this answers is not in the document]
SuperSodaSea: `colord` can reject these strings, so no problem here. By the way the original pattern do allow `fffff` (5 chars).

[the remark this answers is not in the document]
SuperSodaSea: [Source of `colord` for reference](https://github.com/omgovich/colord/blob/master/src/colorModels/hex.ts)

packages/utils/src/color/Color.ts
```diff
@@ -0,0 +1,305 @@
...
+    /**
+     * Set the current color source
+     * @type {PIXI.ColorSource}
+     */
+    set value(value: ColorSource)
+    {
+        if (this._value !== value)
+        {
+            this.normalize();
+            this._value = value;
```
SuperSodaSea, line 86: `this.normalize()` need `this._value` as input :( You may make `value` as an argument of `normalize()`, or just merge `normalize()` into `set value` since it is only used here.

bigtimebuddy: Yes, I made it an argument instead.

packages/utils/src/color/Color.ts
bigtimebuddy: I was thinking about adding getters for alpha, r, g, b intead

packages/utils/src/color/Color.ts
bigtimebuddy: Did some more cleanup on this

packages/utils/src/color/Color.ts
```diff
@@ -0,0 +1,321 @@
...
+                normalizeValue(value[2]),
+                value[3] ?? 1.0,
+            ];
+        }
+        else if (typeof value === 'string' || typeof value === 'object')
+        {
+            if (typeof value === 'string' && Color.HEX_PATTERN.test(value))
+            {
+                // Normalize hex string, remove 0x or # prefix
+                const hexString = Color.HEX_PATTERN.exec(value)[2];
```
SuperSodaSea, line 282: This will execute the regex twice, I think we only need to `exec()` once.

SuperSodaSea:
One more thing, maybe we can re-export `Color` and `ColorSource` in `@pixi/core`, so that we can use `Color.xxx` instead of `utils.Color.xxx`?

bigtimebuddy:
I was thinking of moving to a new `@pixi/color` package instead of utils.

GoodBoyDigital:
this is cool! I would 100% add parse these css colors too: `https://www.npmjs.com/package/css-color-names`
So nice to use :)

SuperSodaSea:
@GoodBoyDigital The new package `colord` we use already supports CSS color names, so we do not need `css-color-names` any more :)

packages/utils/package.json
```diff
@@ -42,8 +42,5 @@
     "earcut": "^2.2.4",
     "eventemitter3": "^4.0.0",
     "url": "^0.11.0"
-  },
-  "devDependencies": {
-    "css-color-names": "^1.0.1"
   }
 }
```
SuperSodaSea, line 46: We need to add `@pixi/color` to `@pixi/utils`' dependencies here.

packages/color/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "publishConfig": {
+      "access": "public"
+    },
+    "files": [
+      "lib",
+      "*.d.ts"
+    ],
+    "devDependencies": {
+      "colord": "^2.9.3"
+    }
```
SuperSodaSea, line 37: By the way, why is `colord` (and `css-color-names` in original code) a dev dependency? It should be a normal dependency in my opinion.

packages/color/src/Color.ts
```diff
@@ -0,0 +1,331 @@
...
+ */
+export class Color
+{
+    /**
+     * Default Color object for static uses
+     * @example
+     * import { Color } from 'pixi.js';
+     * Color.default.setValue(0xffffff).toString(); // '#ffffff'
+     */
+    static readonly default = new Color();
```
Zyie, line 42: Should this be named `shared` to match the syntax with `Ticker`?

packages/color/src/Color.ts
```diff
@@ -0,0 +1,331 @@
...
+    {
+        const [r, g, b] = this._components;
+
+        return { r, g, b };
+    }
+
+    /** Get the color as a CSS-style rgba string: `rgba(255,255,255,1.0)`. */
+    toRgbaString(): string
+    {
+        const [r, g, b, a] = this._components;
```
Julien-Marcou, line 154: I would add a new public `toUint8ClampedArray()` method that returns the `_components` as an array of values beween 0 and 255.

This method could be in turned used here with:

```ts
const [r, g, b] = this.toUint8ClampedArray();
const a = this._components[4]:

return `rgba(${r},${g},${b},${a})`;
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,331 @@
...
+        return this._int;
+    }
+
+    /**
+     * Converts a hexadecimal color number to a string.
+     * @example
+     * import { Color } from 'pixi.js';
+     * new Color('white').toString(); // returns "#ffffff"
+     */
+    toString(): string
```
Julien-Marcou, line 195: I would expect implicit string conversion to be lossless (i.e. not losing the alpha channel):

```
expect(`${new Color('#ffffff00')`).toEqual('#ffffff00');
```

I suggest to rename this method to `toHexString()`, rename the `toHexString()` to `toHexAlphaString()` and make the `toString()` method an alias for the `toHexAlphaString()` method.

Or `toRgbHexString()`, `toRgbaHexString()` and make `toString()` an alias for the `toRgbaHexString()`.

WDYT ?

Julien-Marcou (changes requested):

packages/color/src/Color.ts
bigtimebuddy: I don't have strong feelings about these names and would like feedback from others. I was thinking `toString()` as equivalent to `toNumber()`(non-alpha).

packages/color/src/Color.ts
```diff
@@ -0,0 +1,361 @@
...
+            const normalizeValue = (v: number) => (v > 1 ? v / 255 : v);
+
+            components = [
+                normalizeValue(value[0]),
+                normalizeValue(value[1]),
+                normalizeValue(value[2]),
+                value[3] ?? 1.0,
+            ];
+        }
+        else if (value instanceof Uint8Array || value instanceof Uint8ClampedArray)
```
Julien-Marcou, line 305:
```suggestion
        else if ((value instanceof Uint8Array || value instanceof Uint8ClampedArray) && value.length >= 3)
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,361 @@
...
+        {
+            components = [
+                value[0] / 255,
+                value[1] / 255,
+                value[2] / 255,
+                1.0,
```
Julien-Marcou, line 311:
```suggestion
                value[3] / 255 ?? 1.0,
```

Julien-Marcou (changes requested):

packages/color/test/Color.tests.ts
```diff
@@ -0,0 +1,155 @@
...
+            -1,
+            '',
+            null,
+            false,
+            true,
+            {},
+            { foo: 'bar' },
+            '0xfffff',
+            '#ff',
+            'ff'
```
Julien-Marcou, line 19: Maybe add this to the list of invalid colors:

`[1, 1]` & `new Uint8Array([255, 255])`

Julien-Marcou (changes requested):

packages/color/src/Color.ts
SuperSodaSea: If `value[3]` is undefined, then `value[3] / 255` will be `NaN`, and `value[3] / 255 ?? 1.0` will also be `NaN`, instead of `1.0`.

bigtimebuddy: What we want here is `(value[3] ?? 255) / 255,`

Julien-Marcou: Yes, well seen! 😅

bigtimebuddy:
@Julien-Marcou I know different teams use the "Request Changes" feedback option for PRs differently, but in general on this project, we only use this for extreme cases (something dangerous, fundamental, timely, egregious mistakes, etc) since this type of feedback blocks merging. For general, non-critical feedback or nit-picky suggestions, please consider leaving as comments. We always make every effort to address all comments before moving forward and if there are things that are missed, a follow-up PR is always a reasonable and welcome.

Julien-Marcou:
@bigtimebuddy understood 👍

packages/color/src/Color.ts
vikpe: I'd go with something like:

```js
toHex(); // "#ff0000"
toHexa(); // "#ff0000ff"
toRgb();
toRgba();
toNumber();
```
imo `toString()` should expose all properties including alpha, the main purpose is presentation.

```
toString(); // "Color(r: 255, g: 0, b: 0, a: 1)"
```
I don't like assuming preference/usage, e.g. `red` is also a string just like `#ff0000` or `rgb(255,0,0)`.

---

Also I'd use range `0-255` by default and expose a `normalize()` method to change range to `0-1`.

bigtimebuddy: Welcome back @vikpe, thanks for. your thoughts 👍

bigtimebuddy: The reason I used floats internally instead of uints was because most of our uses-cases for color tend to utilize floats (uniforms, shaders, etc). I was trying to avoid having to call `normalize`, which makes for a much more verbose API.

I agree with the toString(), that shouldn't mess with standard JS presentation.

vikpe: Then floats it is, looking good!

packages/core/src/background/BackgroundSystem.ts
```diff
@@ -68,28 +61,24 @@ export class BackgroundSystem implements ISystem<BackgroundOptions>
...
 
     /**
      * The background color to fill if not transparent
-     * @member {number}
+     * @member {number|string|number[]|Float32Array}
      */
     get color(): number
     {
-        return this._backgroundColor;
+        return this._background.toNumber();
```
GoodBoyDigital, line 76: perhaps it might be good to return the original value here?

```
renderer.background.color = 'red'

console.log(renderer.background.color) // 'red'
```

packages/text-bitmap/src/utils/drawGlyph.ts
```diff
@@ -51,11 +51,13 @@ export function drawGlyph(
     if (style.dropShadow)
     {
         const dropShadowColor = style.dropShadowColor;
-        const rgb = utils.hex2rgb(typeof dropShadowColor === 'number' ? dropShadowColor : utils.string2hex(dropShadowColor));
         const dropShadowBlur = style.dropShadowBlur * resolution;
         const dropShadowDistance = style.dropShadowDistance * resolution;
 
-        context.shadowColor = `rgba(${rgb[0] * 255},${rgb[1] * 255},${rgb[2] * 255},${style.dropShadowAlpha})`;
+        context.shadowColor = Color.shared
```
GoodBoyDigital, line 57: this chaining is super nice :)

GoodBoyDigital (approved):
lgtm! 👍

SuperSodaSea:
Also we need a `LICENSE` file for `@pixi/color`.

packages/color/src/Color.ts
```diff
@@ -0,0 +1,361 @@
...
+     * @param value - Input value
+     */
+    private normalize(value: Exclude<ColorSource, Color>): void
+    {
+        let components: number[];
+
+        if ((Array.isArray(value) || value instanceof Float32Array) && value.length >= 3)
+        {
+            // Handle floats or uints
+            const normalizeValue = (v: number) => (v > 1 ? v / 255 : v);
```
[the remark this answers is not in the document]
bigtimebuddy: I fixed this. These two things should not overlap now.

Julien-Marcou (approved):
LGTM!

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+
+        out[0] = Math.round(r * 255);
+        out[1] = Math.round(g * 255);
+        out[2] = Math.round(b * 255);
+
+        return out;
+    }
+
+    /**
+     * Converts color to an [R, G, B] array of normalized floats (numbers from 0.0 to 1.0).
```
vikpe, line 181:
```suggestion
     * Convert to an [R, G, B] array of normalized floats (numbers from 0.0 to 1.0).
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+     * new Color('white').toRgb(); // returns { r: 1, g: 1, b: 1 }
+     */
+    toRgb(): RgbColor
+    {
+        const [r, g, b] = this._components;
+
+        return { r, g, b };
+    }
+
+    /** Get the color as a CSS-style rgba string: `rgba(255,255,255,1.0)`. */
```
vikpe, line 152:
```suggestion
    /** Convert to a CSS-style rgba string: `rgba(255,255,255,1.0)`. */
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+     */
+    toRgba(): RgbaColor
+    {
+        const [r, g, b, a] = this._components;
+
+        return { r, g, b, a };
+    }
+
+    /**
+     * Get RGB color object.
```
vikpe, line 140:
```suggestion
     * Convert to a RGB color object.
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+            this._value = value;
+        }
+    }
+    get value(): Exclude<ColorSource, Color>
+    {
+        return this._value;
+    }
+
+    /**
+     * Get RGBA color object.
```
vikpe, line 127:
```suggestion
     * Convert to a RGBA color object.
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+            Math.min(255, (r / step) * step),
+            Math.min(255, (g / step) * step),
+            Math.min(255, (b / step) * step),
+        ]);
+
+        return this;
+    }
+
+    /**
+     * Converts color to an [R, G, B, A] array of normalized floats (numbers from 0.0 to 1.0).
```
vikpe, line 266:
```suggestion
     * Convert to an [R, G, B, A] array of normalized floats (numbers from 0.0 to 1.0).
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+    HslColor,
+    HsvaColor,
+    HsvColor,
+    RgbaColor,
+    RgbColor,
+} from 'colord/types';
+
+extend([namesPlugin]);
+
+/** Different possible value types for Color class */
```
vikpe, line 16:
```suggestion
/** Possible value types for Color class */
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+export type ColorSource = string | number | number[] | Float32Array | Uint8Array | Uint8ClampedArray
+| HslColor | HslaColor | HsvColor | HsvaColor | RgbColor | RgbaColor | Color;
+
+/**
+ * Converts a color value to an [R,G,B,A] array of normalized values (numbers from 0.0 to 1.0).
```
vikpe, line 21:
```suggestion
 * Color utility class
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+     * @param {PIXI.ColorSource} value - Optional value to use, if not provided, white is used.
+     */
+    constructor(value: ColorSource = 0xffffff)
+    {
+        this._components = new Float32Array(4);
+        this._components.fill(1);
+        this.value = value;
+    }
+
+    /** Read out red value (0 - 1) */
```
vikpe, line 67:
```suggestion
    /** Get red component (0 - 1) */
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+    get red(): number
+    {
+        return this._components[0];
+    }
+
+    /** Read out green value (0 - 1) */
```
vikpe, line 73:
```suggestion
    /** Get green component (0 - 1) */
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+    get green(): number
+    {
+        return this._components[1];
+    }
+
+    /** Read out blue value (0 - 1) */
```
vikpe, line 79:
```suggestion
    /** Get blue component (0 - 1) */
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+    get blue(): number
+    {
+        return this._components[2];
+    }
+
+    /** Read out alpha value (0 - 1) */
```
vikpe, line 85:
```suggestion
    /** Get alpha component (0 - 1) */
```

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+    /** Get the color as a CSS-style rgba string: `rgba(255,255,255,1.0)`. */
+    toRgbaString(): string
+    {
+        const [r, g, b] = this.toUint8ClampedArray();
+
+        return `rgba(${r},${g},${b},${this.alpha})`;
+    }
+
+    /**
+     * Get non-alpha color as clamped uint8 values (0 to 255).
```
vikpe, line 161:
```suggestion
     * Convert to an [R, G, B] array of clamped uint8 values (0 to 255).
```

vikpe (commented):
Update method descriptions/comments.

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+     * @example
+     * import { Color } from 'pixi.js';
+     * new Color('white').toUint8ClampedArray(); // returns [255, 255, 255]
+     * @param {number[]|Uint8ClampedArray} [out] - Output array
+     */
+    toUint8ClampedArray<T extends (number[] | Uint8ClampedArray) = number[]>(out?: T): T
```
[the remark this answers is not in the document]
bigtimebuddy: Yeah, that's better `toUint8RgbArray`

packages/color/src/Color.ts
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Set alpha, suitable for chaining.
+     * @param alpha
+     */
+    setAlpha(alpha: number): this
+    {
+        this._components[3] = alpha;
+
+        return this;
+    }
```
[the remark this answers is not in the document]
bigtimebuddy: I'm trying to not overload on convenience stuff and focus on the essential bits for Pixi. But If there's a useful reason to add setters for individual channels, I'm not opposed to someone adding a follow-up PR.

packages/color/src/Color.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * Rounds the specified color according to the step.
+     * @param step - Number of steps which will be used as a cap when rounding colors
```
vikpe, line 250:
```suggestion
     * @param {number} step - Number of steps which will be used as a cap when rounding colors
```

bigtimebuddy: We actually don't need this. Webdoc (our doc tool) will pick up the type from the TS

packages/color/src/Color.ts
```diff
@@ -0,0 +1,356 @@
...
+    {
+        const alphaValue = Math.round(this._components[3] * 255);
+        const alphaString = alphaValue.toString(16);
+
+        return this.toHex() + '00'.substring(0, 2 - alphaString.length) + alphaString;
+    }
+
+    /**
+     * Set alpha, suitable for chaining.
+     * @param alpha
```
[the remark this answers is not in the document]
bigtimebuddy: Same thing here, we don't need a type. You can see the dock here: https://pixijs.download/feature/color/docs/PIXI.Color.html#setAlpha

packages/color/src/Color.ts
vikpe: Ok, I just noticed inconsistency, sometimes its set and sometimes its not.

bigtimebuddy: Yeah, some of this might be old. Since we converted from JSDoc to webdoc awhile ago and webdoc has gotten better at auto-detecting. That said, some docs don't do a good job of picking up internal types yet (e.g., `PIXI.Color`), but are fine with basic types (e.g., `number`, `string`, `boolean`, `number[]`, etc), in these cases we help webdoc a little by documenting these types.

PR #9063, opened 2023-01-16
Fix: @pixi/node: loadNoadTexture tried to load data-urls

SmartArray:
##### Description of change
`@pixi/node` is unable to load base64 encoded data urls because the wrong handler will try to fetch the file.

`loadNodeBase64.ts` handles loading base64 data URLs already. `loadNodeTexture.ts` must not attempt this because it will try to use `readFileSync` the data url which will cause a rejection.

This PR removes the invalid code.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good clean-up. Thanks.

PR #9067, opened 2023-01-16
Add global move events

Zyie:
This PR add 3 new events `globalpointermove`, `globaltouchmove`, and `globalmousemove`

These move events behave like the old interaction system, and is the equivalent of setting `events.rootBoundary.moveOnAll`

This PR also fixes an issue with `displayObject.addEventListener` types not being correctly assigned in the `global.d.ts` file

### DIFF
--- a/packages/events/src/FederatedEventMap.ts
+++ b/packages/events/src/FederatedEventMap.ts
@@ -35,4 +35,8 @@ export type FederatedEventMap = {
 };
 export type FederatedEventEmitterTypes = {
     [K in keyof FederatedEventMap as K | `${K}capture`]: [event: FederatedEventMap[K]];
+} & {
+    'globalpointermove': [event: FederatedPointerEvent];
+    'globaltouchmove': [event: FederatedPointerEvent];
+    'globalmousemove': [event: FederatedPointerEvent];
 };
--- a/packages/events/src/FederatedEventMap.ts
+++ b/packages/events/src/FederatedEventMap.ts
@@ -33,10 +33,15 @@ export type FederatedEventMap = {
     touchstart: FederatedPointerEvent;
     wheel: FederatedWheelEvent;
 };
+
+export type GlobalFederatedEventMap = {
+    globalmousemove: FederatedPointerEvent;
+    globalpointermove: FederatedPointerEvent;
+    globaltouchmove: FederatedPointerEvent;
+};
+
 export type FederatedEventEmitterTypes = {
     [K in keyof FederatedEventMap as K | `${K}capture`]: [event: FederatedEventMap[K]];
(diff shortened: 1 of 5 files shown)


bigtimebuddy (commented):
Could you add documentation for these events? There are event docs somewhere.

bigtimebuddy (approved):
Looks good :+1:

packages/events/src/FederatedEventMap.ts
```diff
@@ -35,4 +35,8 @@ export type FederatedEventMap = {
 };
 export type FederatedEventEmitterTypes = {
     [K in keyof FederatedEventMap as K | `${K}capture`]: [event: FederatedEventMap[K]];
+} & {
+    'globalpointermove': [event: FederatedPointerEvent];
+    'globaltouchmove': [event: FederatedPointerEvent];
+    'globalmousemove': [event: FederatedPointerEvent];
 };
```
SuperSodaSea, line 42:
```suggestion

export type GlobalFederatedEventMap = {
    globalmousemove: FederatedPointerEvent;
    globalpointermove: FederatedPointerEvent;
    globaltouchmove: FederatedPointerEvent;
};

export type FederatedEventEmitterTypes = {
    [K in keyof FederatedEventMap as K | `${K}capture`]: [event: FederatedEventMap[K]];
} & {
    [K in keyof GlobalFederatedEventMap]: [event: GlobalFederatedEventMap[K]];
};
```

---

Also I wondered if we need to add `GlobalFederatedEventMap` to `IFederatedDisplayObject`:

<https://github.com/pixijs/pixijs/blob/d112742d84da4c1fd61127af3b69519b64a058df/packages/events/src/FederatedEventTarget.ts#L159-L166>

Zyie: I just tested this and it seems that this type never worked in the first place?

I don't seem to get any type information other than the general dom addEventListener typings

Zyie: ok should be all fixed

thanks for tracking this one down for me!

SuperSodaSea (approved):

newtoncodes:
This is crazy slow on spine animations guys! You should revert this ASAP or make a **default** setting to disable it. Animations with many meshes have to be constantly walked through hundreds of times when the mouse moves on something. There is absolutely no point of doing this. I don't have a single pointer or touche event in my application and I had to debug this a whole day to figure out why my animations are running 20 FPS when the mouse is moving...

bigtimebuddy:
@newtoncodes Thanks for reporting. Can you create a new issue so we can track this? You on the latest 7.2.0 prerelease I assume?

PR #9068, opened 2023-01-16
Fix events not firing when mouse does not move

Zyie:
With the new events system events did not fire when the mouse does not move

This PR implements these change by firing a `pointermove` event if the user hasn't moved there mouse.

Most of this logical was directly taken from the old interaction system

todo:

- [ ] fix issue where touch events behave differently to both v6 old interaction & new v7 events

### DIFF
--- a/packages/events/src/EventSystem.ts
+++ b/packages/events/src/EventSystem.ts
@@ -1,4 +1,4 @@
-import { extensions, ExtensionType } from '@pixi/core';
+import { extensions, ExtensionType, Ticker, UPDATE_PRIORITY } from '@pixi/core';
 import { EventBoundary } from './EventBoundary';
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
My biggest concern is that ticker is expensive especially if you don’t need it. I think there’s a meaningful distinction  between static interaction (object doesn’t move while interactive) and dynamic (moving or scaling while interactive). Maybe this could be a new API and we only enable the ticker for dynamic elements.

```js
sprite.interactive = false;
sprite.interactive = ‘static’; // or true
sprite.interactive = ‘dynamic’; // new value
```

Alternative we could add an update method to `renderer.events` which would allow a user to force update (e.g., maybe during a tween).

Zyie:
closing in favor of #9092

PR #9073, opened 2023-01-17
Fix: Prevent filepath to be URLEncoded

SmartArray:
Use the raw url instead of the URLencoded version of cross-fetch. Otherwise `@pixi/node`'s asset loader will be unable to load files with spaces in their filepath

##### Description of change
Currently, if trying to load a local file using `@pixi/node`'s AssetLoader adapter, the loader will be unable to load the file if the filename contains special characters such as spaces as they will be URLencoded.

**Example:**
`/tmp/mario luigi.png` becomes `/tmp/mario%20luigi.png` and the loader will be unable to load the file.

The corresponding fix is easy. When trying to access the file in the adapter, it should not use the url of the request instance but the raw url that was specified when `fetch` was invoked.

##### Pre-Merge Checklist


bundles/pixi.js-node/src/adapter/adapter.ts
```diff
@@ -36,7 +36,7 @@ export const NodeAdapter = {
 
         return new Promise((resolve, reject) =>
         {
-            const filePath = utils.path.normalize(request.url);
+            const filePath = utils.path.normalize(typeof url === 'string' ? url : request.url);
```
SuperSodaSea, line 39:
```suggestion
            const filePath = utils.path.normalize(decodeURI(request.url));
```

Perhaps this is better?

SmartArray: Yes, but only if the Request instance doesn't modify the path. (I haven't looked into that)

SuperSodaSea:
```js
const request = new Request('a%20b c.txt');
console.log(request.url); // a%20b%20c.txt
```

bigtimebuddy: Yeah, I think circumventing `request.url` (not using decodeURI) is probably a better idea, since Request can do thing to resolve paths, add host name, etc.

bigtimebuddy, line 39: My only suggestion is to add a comment, since reading this it's not super obvious why we'd want to use url. Something like:

```
// Request transforms paths and encodeURIs, but for filesystem requests, it's better to use the raw string (path)
```

SmartArray:
@bigtimebuddy Damn, I hope you are not mad. Didn't know you were going to merge this now. The linter wasn't happy. Will split the line now

SmartArray:
Please squash the commits when the checks went through. It's starting to look ridiculous 😂

SuperSodaSea:
@SmartArray We always squash commits when merging PR, so don't worry :)

SmartArray:
Here we go. I hope others will find this useful too.

PR #9089, opened 2023-01-19
Add new types of interaction

Zyie:
## Overview
This PR is the start to an overhaul of `interaction` in pixi which will allow devs fully express different types of interaction

There is now 4 types of interaction:
- `'none'`: does not emit events and ignores all hit testing. Similar to `pointer-events: none` in CSS
- `'auto'`: does not emit events and but is hit tested if parent is interactive. Same as `interactive = false` in v7
- `'static'`: emit events and is hit tested. Same as `interaction = true` in v7
- `'dynamic'`: emits events and is hit tested but will also receive mock interaction events fired from a ticker to allow for interaction when the mouse isn't moving

For backwards compatibility we still allow `interactive = true/false`. 
- `interactive = true` would be the same as `interactive = 'static'`
- `interactive = false` would be the same as `interactive = 'auto'`

NOTE: `dynamic` is still to be implemented. I will do that in a follow up PR to make these reviews easier.

## Questions:

- If you set `interactive = 'none'` on a parent, should we stop hit testing even if a child has `interactive = 'static'`? Currently they are not.
You would have to set `interactiveChildren = false` to stop the children from being hit tested.
e.g.
    ```js
    const sprite = new Sprite();
    container.interactive = 'none';
    sprite.interactive = true;
    sprite.on('pointermove', ()=>{
        console.log('sprite currently still receives this event');
    })
    container.addChild(sprite);
    ```
From what I can tell the DOM would stop the events from reaching the sprite in this case. So should we do the same?

## Todo:
- [x] Masks should be `interactive = 'none'` by default to avoid blocking interaction


Zyie:
@bigtimebuddy @GoodBoyDigital would be good to get your thoughts while i tidy up the last few things!

bigtimebuddy:
I would match the DOM behavior (`pointer-events:none`) with `interactive = 'none'`... so I don't think children should get click. In this case the difference between `interactive = 'none'` and `interactiveChildren = false` would be the the former ignores itself AND children, where the latter only ignores children. Right?

packages/events/src/FederatedEventTarget.ts
```diff
@@ -684,4 +700,34 @@ export const FederatedDisplayObject: IFederatedDisplayObject = {
     }
 };
 
+Object.defineProperties(FederatedDisplayObject, {
+    interactive:  {
+        get()
+        {
+            return this._userSetInteractive;
```
bigtimebuddy, line 707: Being able to set the global default would be very helpful (e.g., `renderer.events.defaultInteraction = 'none'` or `FederatedDisplayObject.defaultInteraction`)

Zyie:
> I would match the DOM behavior (`pointer-events:none`) with `interactive = 'none'`... so I don't think children should get click. 

Just had this thought but if we go with this option it would make setting the global `renderer.events.defaultInteraction = 'none'` very difficult to work with? The stage would be set to `none` and therefore no interaction would take place at all and you would have to crawl through your scene to reset all your containers

We could add an extra type like `passive` which is what `none` does now and change `none` to be explicitly no interaction at all

e.g

- `'none'`: ignores all interaction
- `'passive'`: does not emit events and ignores hit testing on itself. Does allow for events and hit testing on it's children
- `'auto'`: does not emit events and but is hit tested if parent is interactive. Same as interactive = false in v7
- `'static'`: emit events and is hit tested. Same as interaction = true in v7
- `'dynamic'`: emits events and is hit tested but will also receive mock interaction events fired from a ticker to allow for interaction when the mouse isn't moving

bigtimebuddy:
I like it!

GoodBoyDigital:
Yeah, this is cool. i like that `none` removes something entirely from the interaction graph!

Zyie:
Ok `passive` has been implemented

I also added a `isInteractive` function to the DisplayObject as a nice little helper for the common use case of 
```
if (displayObject.interactive) {
    // do something
}
```

TODO:

- [ ] global switch of default type
- [ ] implement `dynamic` functionality (separate PR)
- [ ] make sure masks are not considered for interaction
- [ ]

packages/events/src/FederatedEventTarget.ts
Zyie: I'm struggling to think how this could be implemented without it causing side effects

We don't have access to the `DisplayObjects` constructor, so setting the default value at a fixed point is hard

For example take this potential implementation

```ts
interactive:  {
    get()
    {
        if (!this._hasUserDefinedInteractive)
        {
            this.interactive = EventSystem.defaultInteraction;
        }

        return this._userSetInteractive;
    },
    set(value: boolean | Interactive)
    {
        this._hasUserDefinedInteractive = true;
        this._userSetInteractive = value;
        this._internalInteractive = convertToInteractive(value);
    },
},
```

With the above `interactive` can be a different value depending on when it is accessed, which is not good
e.g.

```ts
EventSystem.defaultInteraction = false;
const graphics = new Graphics();

// if we uncomment this line, the test will pass
// expect(graphics.interactive).toEqual(false);

EventSystem.defaultInteraction = 'dynamic';

const graphics2 = new Graphics();

expect(graphics2.interactive).toEqual('dynamic');

// this will fail because we never accessed the `interactive` property on `graphics` before switching the default
// however it was originally created with the default value of `false` and should not change
expect(graphics.interactive).toEqual(false);
```

Maybe I'm just being dumb here and there is something obvious i am missing 😄

bigtimebuddy: Maybe make the default `_userSetInteractive` undefined then `return this._userSetInteractive ?? EventSystem.defaultInteraction` in the getter.

Zyie: Ok I've got something that works as long as we are ok with `interactive` dynamically changing if the user doesn't explicitly set `interaction` themselves

e.g

```ts
EventSystem.defaultInteraction = false;

const graphics = new Graphics();

// nothing has been explicitly set, so it should be false
expect(graphics.interactive).toEqual(false);
expect(graphics._internalInteractive).toEqual('auto');

EventSystem.defaultInteraction = 'dynamic';

// we have now set the default interaction to dynamic
// we have not explicitly set the state, so this pre existing Graphics should update to be dynamic
expect(graphics.interactive).toEqual('dynamic');
expect(graphics._internalInteractive).toEqual('dynamic');

EventSystem.defaultInteraction = true;
const graphics2 = new Graphics();

// we now explicitly set the default interaction to false
graphics.interactive = false;
expect(graphics.interactive).toEqual(false);
expect(graphics._internalInteractive).toEqual('auto');

// we have NOT set the state on this new Graphics, so it should be true
expect(graphics2.interactive).toEqual(true);
expect(graphics2._internalInteractive).toEqual('static');
```

bigtimebuddy: Alternative idea: what about adding a Renderer/Application constructor option instead to set the default? Changing the default on the fly feels very weird and probably not that practical to me.

thundermarek:
can I ask what is the difference between interactiveChildren = false and interactive = none? In v6 and before I used interactiveChildren to optimize my event trees, right now interactive = none seems to be redundant? Or does it have some slightly different functionality? Anyway is very confusing to have them both, maybe drop interactiveChildren in favor of that interactive = none?

Also if true / false is left for backwards combability maybe do
interactive = true -> 'static'
interactive = false -> 'passive'
as that will be true v6 combability? Yes, it can break active codebases already ported to v7, but I doubt there is a lot of them, and curtenly there is more v6 codebases.

Also also these changes make me happy, after this is merged and released I can finally move on to v7.

md5crypt:
can I ask what is the difference between interactiveChildren = false and interactive = none? In v6 and before I used interactiveChildren to optimize my event trees, right now interactive = none seems to be redundant? Or does it have some slightly different functionality? Anyway is very confusing to have them both, maybe drop interactiveChildren in favor of that interactive = none?

Also if true / false is left for backwards combability maybe do
interactive = true -> 'static'
interactive = false -> 'passive'
as that will be true v6 combability? Yes, it can break active codebases already ported to v7, but I doubt there is a lot of them, and curtenly there is more v6 codebases.

Also also these changes make me happy, after this is merged and released I can finally move on to v7.

bigtimebuddy:
I agree with @md5crypt would be nice to deprecate `interactiveChidlren`.

md5crypt:
Yeah, there actually is, if I want the object to be interactive but it's children not... but that means that interactive = none is redundant with interactive = passive interactiveChildren = false? Then if we are leaving interactiveChildren maybe drop interactive = none? I don't like having 2 ways to do the exact same thing...

Zyie:
> Also if true / false is left for backwards combability maybe do
> interactive = true -> 'static'
> interactive = false -> 'passive'
> as that will be true v6 combability? Yes, it can break active codebases already ported to v7, but I doubt there is a lot of them, and curtenly there is more v6 codebases.

We are going to be adding an option to change what the default setting for `interactive` is so that you can set it to `passive` by default. We are trying not to introduce any breaking changes. We can look at setting `passive` as the default in the next major release

> But maybe there's some use-case for keeping with combo of dynamic or static?

Yeah we want to keep `interactiveChildren` around for the static/dynamic use cases. This will allow you to optimise your interactions if you only care about the top level container getting interaction events


> but that means that interactive = none is redundant with interactive = passive interactiveChildren = false?

This is correct `interactive = 'none'` and `interactive = 'passive'; interactiveChildren = false` are functionally equivalent

However, I like having the shorthand way of doing it and to me feels easily understandable? I don't have a massively strong opinion here though so happy to remove `none`

md5crypt:
how about we go more verbose and add these internal fields instead of `_internalInteractive`:

```
_pruneHitTest: boolean
_pruneChildrenHitTest: boolean
_emitEvents: boolean
_dynamic: boolean
```

then

```
auto:
_pruneHitTest = false
_pruneChildrenHitTest = false
_emitEvents = false
_dynamic = false

passive:
_pruneHitTest = true
_pruneChildrenHitTest = false
_emitEvents = false
_dynamic = false

static
_pruneHitTest = false
_pruneChildrenHitTest = false
_emitEvents = true
_dynamic = false

dynamic
_pruneHitTest = false
_pruneChildrenHitTest = false
_emitEvents = true
_dynamic = true

none
_pruneHitTest = true
_pruneChildrenHitTest = true
_emitEvents = false
_dynamic = false
```

interactive setter can take boolean, string or an object with those 4 keys. Yes, it makes the interactive setter more complex but the actual logic and code inside events system gets simplified. Also end user gets more control. For example now we can have `{pruneHitTest: true, pruneChildrenHitTest: false, emitEvents: false, dynamic: true}` to send the synthetic event down a part of the tree. Also there are no longer two branches in logic doing the same thing.

alternatively to keep things simple maybe add a setter / getter for each field (or just make them public) and make interactive just control a boolean setter / getter for _emitEvents directly which will be fully backwards compatible. I think I actually prefer this solution as its simpler and is easy to understand just by reading the names of the fields.

bigtimebuddy (commented):
This feels like a half-deprecation for the boolean type of `interaction`. For instance `defaultInteraction` or `defaultInteractionType` both allow boolean, but these are new APIs and would remove the boolean type here. boolean should only be allowed for the `interaction` setter.

Also, can we renamed `defaultInteractionType` to `defaultInteraction` in the options? I don't think the "Type" adds anything.

bigtimebuddy (approved):

PR #9092, opened 2023-01-20
Adds EventsTicker to mock events when mouse is stationary

Zyie:
This is an addition to #9089 

This PR adds a new `EventsTicker` that will fire mock events when the mouse is not moving and there is at least 1 `dynamic` element in the scene

This approach does have its drawbacks as we will still fire events even if nothing is actually moving, however this approach is at least simple. Future work could be done to make the ticker smarter in it's decision to turn itself on/off


GoodBoyDigital (approved):
love it!

Issue #9064, opened 2023-01-16
Bug: Doesn't work with a CRA 5.0.1 project


tansanDOTeth:
### Current Behavior

After installing pixi and importing it, this is the following error I am getting:

```
ERROR in ./node_modules/@pixi/core/lib/textures/resources/ImageBitmapResource.mjs
Module build failed (from ./node_modules/babel-loader/lib/index.js):
SyntaxError: /projects/test-project/node_modules/@pixi/core/lib/textures/resources/ImageBitmapResource.mjs: When using '@babel/plugin-transform-parameters', it's not possible to compile `super()` in an arrow function with default or rest parameters without compiling classes.
Please add '@babel/plugin-transform-classes' to your Babel configuration.

   6 |   constructor(source, options) {
   7 |     var __super = (...args) => {
>  8 |       super(...args);
     |       ^^^^^^^^^^^^^^
   9 |     };
  10 |     options = options || {};
  11 |     if (typeof source === "string") {
```

### Expected Behavior

I expect it to run without errors.

### Steps to Reproduce

1. npx create-react-app pixi-test
2. cd pixi-test
3. npm install pixi.js
4. Import pixi
```
import * as PIXI from 'pixi.js';
```
5. npm run start
7. Error shows

### Environment

- **`pixi.js` version**: *e.g. 7.1.0*
- **Browser & Version**: *e.g. Chrome 108*
- **OS & Version**: *e.g. Ubuntu 22.04*
- **Running Example**: *e.g. <https://pixiplayground.com/>*

### Possible Solution

Not really sure

### Additional Information

Seems like babel added this change relatively recently and it was to prevent infinite recursion.

https://github.com/babel/babel/issues/15148
https://github.com/babel/babel/pull/15163

Here is the transpiled code:
```
class ImageBitmapResource extends BaseImageResource {
    constructor(source, options) {
      var __super = (...args) => {
        super(...args);
      };
      options = options || {};
      if (typeof source === "string") {
        __super(ImageBitmapResource.EMPTY);
        this.url = source;
      } else {
        __super(source);
        this.url = null;
      }
      this.crossOrigin = options.crossOrigin ?? true;
      this.alphaMode = typeof options.alphaMode === "number" ? options.alphaMode : null;
      this._load = null;
      if (options.autoLoad !== false) {
        this.load();
      }
    }
```

SuperSodaSea:
Duplicated with #8733 and #8824. This is a limitation of transcription.
[...]

Issue #8824, opened 2022-11-05
Asset.load(image) gets "Maximum call stack size exceeded" error


ivan-work:
I've stuck with loading even a simple asset, could you please help with that?

I've started fresh react-scripts project: npx create-react-app my-app --template typescript

Then I placed 'bunny.png' to /public folder

And wrote Assets.load('bunny.png'); somewhere in the app.

I keep getting error:

```
Uncaught (in promise) Error: [Loader.load] Failed to load http://localhost:3000/bunny.png.
RangeError: Maximum call stack size exceeded
    at Loader.ts:148:1
    at async Promise.all (:3000/index 0)
    at async Loader.load (Loader.ts:153:1)
    at async AssetsClass._mapLoadToResolve (Assets.ts:689:1)
    at async AssetsClass.load
```

Based on my research it happens because for some reason the "ImageBitmapResource" class constructor somehow has recursion and it keeps calling itself.

I've checked it both on my local big project and separate example project: https://github.com/ivan-work/pixi-bug-test

### Environment
- Fresh react-scripts project (npx create-react-app my-app --template typescript)
- **`pixi.js` version**: _e.g. 7.0.2_
- **Browser & Version**: _e.g. Chrome / Firefox_
- **Running Example**: [_url_](https://github.com/ivan-work/pixi-bug-test)

SuperSodaSea:
Seems to be duplicated with #8733. `create-react-app` have [some problems](https://github.com/pixijs/pixijs/issues/8733#issuecomment-1278953941) while transpiling conditional `super` call...

SuperSodaSea:
There's one thing I'm curious about, the `ImageResource` should be used in preference to `ImageBitmapResource` in `Assets.load`... Need some investigation.

ivan-work:
yeah, I think it's a duplicate, thanks for the link

SuperSodaSea:
I found a workaround for this: just delete the `"last 1 safari version"` in `package.json`.

Issue #8733, opened 2022-10-14
Assets bundle load exception


vpmedia:
### Expected Behavior

Assets load successfully (works with v7.0.0-alpha.2)

### Current Behavior

`
   Loader.ts:148 Uncaught (in promise) Error: [Loader.load] Failed to load http://localhost:3000/asset/.....
   RangeError: Maximum call stack size exceeded

    at Loader.ts:148:1

    at async Promise.all (:3000/index 1)

    at async Loader.load (Loader.ts:153:1)

    at async AssetsClass._mapLoadToResolve (Assets.ts:676:1)

    at async Promise.all (:3000/index 0)

    at async AssetsClass.loadBundle (Assets.ts:545:1)

    at async ......
`

### Possible Solution

Downgrade to v.7.0.0-alpha.2

### Steps to Reproduce

`
  await Assets.init({ manifest: "/manifest.json" });
  await Assets.loadBundle("BundleName");
`

### Environment

- **`pixi.js` version**: v7.0.0-beta.3
- **Browser & Version**: Version 106.0.5249.119 (Official Build) (x86_64) (Same issue using Safari, Chrome, Firefox)
- **OS & Version**: MacOS Monterey 12.6
- **Running Example**: _url_

SuperSodaSea:
What is the content of your `manifest.json`?

SuperSodaSea:
Sorry, but I failed to reproduce this... Can you provide more information, such as find the line with `[Loader.load] Failed to load` in the bundle file you are using, and add a `console.log(e)` before it to see the real stack trace of the error?

https://github.com/pixijs/pixijs/blob/5823bec73949aba4c43609a53561c006af7f2517/packages/assets/src/loader/Loader.ts#L147-L148

vpmedia:
If I remove the bundle loading and add one simple
`Assets.add(..) .. Assets.load(..) ..` the same exception is thrown, so it's not connected to the assets bundle format.
I'm having hard time figuring out how to modify the source in the node_modules folder, there is a bunch of package formats (pixi/assets, pixi.js, pixijs.legacy with cjs, esm subfolders).
I'm using create-react-app/webpack to compile my app..

But the thrown exception is re-thrown the way I see `.\n${e}` ..

SuperSodaSea:
I tried `await Assets.init({ manifest: "/manifest.json" }); await Assets.loadBundle("BundleName");` using <script> elememt and using webpack, both of them work just fine.

[...]

PR #9093, opened 2023-01-20
Fix: Workaround for transpilation failure with create-react-app

SuperSodaSea:
#### Description of change

Users using `create-react-app` + PixiJS are continuously facing problem with transpilation (#8733, #8824 and #9064). Let me explain the details.

In #8698 I added code below in `ImageBitmapResource.constructor`:

<https://github.com/pixijs/pixijs/blob/2f12f4ee46fd17ef5b0ab2eb6bc23f1d03295f76/packages/core/src/textures/resources/ImageBitmapResource.ts#L62-L73>

This looks good, and when releasing it will be transpiled by `esbuild`:

```js
var __super = (...args) => {
  super(...args);
};
if (typeof source === "string") {
  __super(ImageBitmapResource.EMPTY);
  this.url = source;
} else {
  __super(source);
  this.url = null;
}
```

([Source code of related logic in esbuild here.](https://github.com/evanw/esbuild/blob/5fe21253ee75fb4c5ea395a0877b2a5ab51c3575/internal/js_parser/js_parser_lower.go#L3064-L3071))

Until then, everything is fine, but users usually need to do their own transpilation, and here is where it can cause problems. Noticing that `__super = (...args) => { super(...args); }` is an arrow function with railing comma in parameters, and have a `super` call in its body. Unfortunately, due to limitation of transpilation, it is impossible to transpile it without transpiling the whole class (using polyfills like `_createClass` in Babel instead of native ES6 classes).

This situation can happen when using the default setting of `create-react-app`, or `@babel/preset-env` with target `since 2017`. Previously Babel wrongly transpiles this into:

```js
var _supercall = function () {
  return _supercall(...arguments);
};
var __super = function () {
  _supercall(...arguments);
};
if (typeof source === "string") {
  __super(ImageBitmapResource.EMPTY);
  this.url = source;
} else {
  __super(source);
  this.url = null;
}
```

This will lead to infinite recursion. So I opened issue babel/babel#15148 and PR babel/babel#15163 to fix this bug in Babel. Now it will report error when transpiling:

```
SyntaxError: .../ImageBitmapResource.mjs: When using '@babel/plugin-transform-parameters', it's not possible to compile `super()` in an arrow function with default or rest parameters without compiling classes.
Please add '@babel/plugin-transform-classes' to your Babel configuration.
```

Users will get an explicit error instead of a confusing infinite recursion, but it is still quite annoying. So in this PR I just move `super()` out of `if()`, solving this issue.

#### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/resources/ImageBitmapResource.ts
+++ b/packages/core/src/textures/resources/ImageBitmapResource.ts
@@ -59,18 +59,21 @@ export class ImageBitmapResource extends BaseImageResource
     {
         options = options || {};
 
+        let baseSource;
+        let url;
+
         if (typeof source === 'string')
         {
-            super(ImageBitmapResource.EMPTY);
-
-            this.url = source;
+            baseSource = ImageBitmapResource.EMPTY;
+            url = source;
         }
         else
         {
-            super(source);
-
-            this.url = null;
+            baseSource = source;
+            url = null;
         }
+        super(baseSource);
+        this.url = url;
 
         this.crossOrigin = options.crossOrigin ?? true;
         this.alphaMode = typeof options.alphaMode === 'number' ? options.alphaMode : null;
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):
Good fix!

PR #9096, opened 2023-01-20
Color API additions: multiply, toLittleEndianNumber

bigtimebuddy:
Builds on #9061 

Adds two new methods:

* `multiply`
* `toLittleEndianNumber`

Cleans-up several instance of these inline color transforms.


packages/color/src/Color.ts
```diff
@@ -207,6 +207,37 @@ export class Color
...
+
+        return (value >> 16) + (value & 0xff00) + ((value & 0xff) << 16);
+    }
+
+    /**
+     * Multiply the colors
+     * @param {PIXI.Color | number[]} value - The color to multiply by, either
+     *   and existing Color object or an [R, G, B] or [R, G, B, A] array of floats.
+     */
+    multiply(value: Color | number[]): this
```
vikpe, line 229: Wouldn't it be better to accept a `ColorSource`?

bigtimebuddy: Yeah, you're probably right. I was trying to avoid `new Color` internally here, but I think it's probably okay.

packages/canvas-graphics/src/CanvasGraphicsRenderer.ts
```diff
@@ -96,9 +96,7 @@ export class CanvasGraphicsRenderer
         let contextFillStyle;
         let contextStrokeStyle;
 
-        const tintR = ((graphics.tint >> 16) & 0xFF) / 255;
-        const tintG = ((graphics.tint >> 8) & 0xFF) / 255;
-        const tintB = (graphics.tint & 0xFF) / 255;
+        const tint = Color.shared.setValue(graphics.tint).toArray();
```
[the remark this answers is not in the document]
bigtimebuddy: Anything that touches rendering is hot code. In this case, this is to avoid extra garbage, as using `shared` is the same thing as using one of those static color util functions. We don't need to be instantiating `new Color` 60fps.

packages/color/src/Color.ts
```diff
@@ -207,6 +207,36 @@ export class Color
...
+        return (value >> 16) + (value & 0xff00) + ((value & 0xff) << 16);
+    }
+
+    /**
+     * Multiply with another color
+     * @param {PIXI.ColorSource} value - The color to multiply by.
+     */
+    multiply(value: ColorSource): this
+    {
+        const [r, g, b, a = 1] = new Color(value)._components;
```
GoodBoyDigital, line 230: could we use the shared on here also?

```suggestion
        const [r, g, b, a = 1] =  Color.shared.setValue(value)._components;
```

packages/graphics/src/Graphics.ts
```diff
@@ -1081,18 +1081,10 @@ export class Graphics extends Container
...
-                const b = (tintRGB[2] * batchTint[2]) * 255;
-
-                // TODO Ivan, can this be done in one go?
-                const color = (r << 16) + (g << 8) + (b | 0);
-
-                batch._tintRGB = (color >> 16)
-                        + (color & 0xff00)
-                        + ((color & 0xff) << 16);
+                batch._tintRGB = Color.shared
+                    .setValue(tintRGB)
```
GoodBoyDigital, line 1085: I really do like this flow chaining :D

packages/graphics/src/GraphicsGeometry.ts
```diff
@@ -808,7 +809,7 @@ export class GraphicsGeometry extends BatchGeometry
         offset = 0): void
     {
         // TODO use the premultiply bits Ivan added
-        const rgb = (color >> 16) + (color & 0xff00) + ((color & 0xff) << 16);
+        const rgb = Color.shared.setValue(color).toLittleEndianNumber();
 
         const rgba = utils.premultiplyTint(rgb, alpha);
```
GoodBoyDigital, line 814: one for the laterbase - `utils.premultiplyTint` <- this could be  Color function also!

GoodBoyDigital (approved):
👍

packages/color/src/Color.ts
bigtimebuddy: I think I tried this and ran into issues if you're already in the shared object. But will try again.

PR #9097, opened 2023-01-20
Support ColorSource in clear color methods

bigtimebuddy:
Builds on #9061 

* Add ColorSource support for BaseRenderTexture's `clearColor`
* Add ColorSource support for RenderTextureSystem's`clear`
* Add ColorSource support for CanvasContextSystem's `clear`
* Remove the convenience getters from BackgroundSystem, replace with `backgroundColor:Color` getter


packages/core/src/renderTexture/BaseRenderTexture.ts
```diff
@@ -97,6 +98,25 @@ export class BaseRenderTexture extends BaseTexture
...
     }
 
+    /** Color when clearning the texture. */
+    set clearColor(value: ColorSource)
+    {
+        this._clear.setValue(value);
+    }
+    get clearColor(): ColorSource
+    {
+        return this._clear;
```
GoodBoyDigital, line 108: should this be ->
```suggestion
        return this._clear.value;
```

GoodBoyDigital (approved):
lgtm! I would suggest renaming `_value` as its a little generic. Perhaps `_backgroundColor` ?

bigtimebuddy:
Good suggestions, thanks @GoodBoyDigital

PR #9099, opened 2023-01-22
Fix: Tests for Assets bundle

SuperSodaSea:
#### Description of change

In #9056, it seems that an `it.only` is introduced in `bundle.test.ts` which accidentally skips all other tests.

<https://github.com/pixijs/pixijs/blob/45afb81c23f1728732f26875998d6cc3fe10790d/packages/assets/test/bundle.tests.ts#L358>

```
Test Suites: 1 skipped, 109 passed, 109 of 110 total
Tests:       12 skipped, 905 passed, 917 total
```

When I was removing the `.only`, there somes some errors in the tests, which I found were caused by `Resolver#reset` not resetting all members of `Resolver`. So I fixed `Resolver#reset` and now the tests are running well.

```
Test Suites: 1 skipped, 109 passed, 109 of 110 total
Tests:       1 skipped, 916 passed, 917 total
```

(The remaining 1 skipped test is caused by [`Basis.tests.ts`](https://github.com/pixijs/pixijs/commit/ab357247e7ac34d89dad6cd0cf3bd06ad9e2e5c4), which is still broken on CI for unknown reason.)

#### Pre-Merge Checklist


bigtimebuddy (approved):

bigtimebuddy:
Do we need a hotfox release with this PR? WDYT?

SuperSodaSea:
> Do we need a hotfix release with this PR? WDYT?

@bigtimebuddy `Resolver#reset` is mainly used for testing, so we don't need to rush to release a new version.

GoodBoyDigital (approved):
nice catch @SuperSodaSea !

Issue #8757, opened 2022-10-20
Export Canvas as PNG does decrease the quality of the image


styriandev:
It seems like the renderer.extract.canvas function is decreasing the quality of the canvas. Is there a possability to change that behaviour by another function or some parameter?

The prepared example that demonstrates that issue is located under [Example (Stackblitz)](https://stackblitz.com/edit/angular-xmu6se-pixijs-dil7ja?file=src/app/app.component.ts). If you check that out you will find a canvas with some example object + a button that exports the canvas as a PNG via renderer.extract.canvas. You will notice that the quality of the exported PNG has some problems if you zoom to the circle:
![image](https://user-images.githubusercontent.com/46619772/196905182-0da56ef4-a606-420d-a2e0-64708bf1d49f.png)

The interesting thing here is that if I export the canvas via right click "Save image as" the quality is much better (there are less pixalated borders):
![image](https://user-images.githubusercontent.com/46619772/196905391-586de69f-a0ae-40af-b90f-f465bb8c39fe.png)

The problem is that this solution is not acceptable as in my application there is also such a button that will trigger the export. Also it seems like the problem is getting bigger with bigger canvas sizes. If I export the canvas as a PNG inside the application I am getting the following pixelated image:
![image](https://user-images.githubusercontent.com/46619772/196906949-573dfbff-ee95-4221-83e1-74560afb6d7e.png)

In the actual canvas that looks like that:
![image](https://user-images.githubusercontent.com/46619772/196907064-7ef50722-5e24-4e0d-ae12-1b80122bba0e.png)

### Expected Behavior
Actually I would expect the exported PNG to look like the canvas, so it should not decrease the quality of the images. 

### Current Behavior
There are pixelated parts which are not pixelated on the canvas. 

### Steps to Reproduce
Open stackblitz, press on the button "Save PNG" to export the pixelated image. Then right click on the canvas, select "Save image as" and compare the 2 images.

[Example (Stackblitz)](https://stackblitz.com/edit/angular-xmu6se-pixijs-dil7ja?file=src/app/app.component.ts) 
### Environment

- **`pixi.js` version**: 6.5.6
- **OS & Version**: Windows 10
- **Running Example**: [Example (Stackblitz)](https://stackblitz.com/edit/angular-xmu6se-pixijs-dil7ja?file=src/app/app.component.ts)

styriandev:
[...]

PR #9102, opened 2023-01-23
Fix: Extract should use multisample

SuperSodaSea:
#### Description of change

Currently the `Extract#canvas` and `Extract#pixels` doesn't consider the `renderer.multisample`, and this results in the extracted image being rendered without anti-aliasing (#8757), so I add support for multisample in Extract. Also the common part of `Extract#canvas` and `Extract#pixels` ate moved to a separate function `Extract#_rawPixels`.

Example (Left: Direct rendering; Right: Extracted image):

- Before: <https://pixiplayground.com/#/edit/dvxrRjI2Kb7SIIz1vNGcA>
- After: <https://pixiplayground.com/#/edit/jhGoI3tGmcRxrQ_hzCdi->

Fix for v6.x is in #9103.

#### Pre-Merge Checklist


bigtimebuddy (approved):
LGTM!

PR #9110, opened 2023-01-25
Fix renderer options not being passed to plugins

Zyie:
In #9010 renderer options were added, however the reference to the options was set after the plugins/systems got initialised, so they could not be used

### DIFF
--- a/packages/canvas-renderer/src/CanvasRenderer.ts
+++ b/packages/canvas-renderer/src/CanvasRenderer.ts
@@ -243,8 +243,8 @@ export class CanvasRenderer extends SystemManager<CanvasRenderer> implements IRe
             }
         };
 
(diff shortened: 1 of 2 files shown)


SuperSodaSea (commented):
https://github.com/pixijs/pixijs/blob/f1171e4b1f65fd4bc1e85b78b2c8047e200a403b/packages/core/src/Renderer.ts#L383-L406

Maybe we don't need this anymore in the future, the systems can take the options from renderer themselves.

bigtimebuddy (approved):

bigtimebuddy:
I think @SuperSodaSea's comment here is a valid one. I think we should just pass the options object (as immutable) and lets the systems pick things from it. That seems way easier to maintain.

PR #9112, opened 2023-01-26
Simplify renderer plugin startup

Zyie:
As a follow up to the feedback on #9110 i've removed much of the logic behind passing options to the render plugins

Now the only plugins receive renderer.options + plugins map


packages/core/src/startup/StartupSystem.ts
```diff
@@ -1,21 +1,13 @@
...
-import type { IRenderer } from '../IRenderer';
+import type { IRenderer, IRendererOptions } from '../IRenderer';
 import type { IRendererPlugins } from '../plugin/PluginSystem';
 import type { ISystem } from '../system/ISystem';
-import type { ContextOptions } from '../systems';
-import type { ViewOptions } from '../view/ViewSystem';
 
-// TODO this can be infered by good use of generics in the future..
-export interface StartupOptions extends Record<string, unknown>
+export interface StartupOptions extends IRendererOptions
```
bigtimebuddy, line 8: Overall, this change is great. Simplification and decentralization is awesome.

The only thing that I might improve, (and totally fine to do this as a follow up PR, if you think it's out of scope):

Organizationally, it would be nice if type concerns were isolated for IRendererOptions. For instance, BackgroundSystem should "own" the option types for background, backgroundAlpha, bsckgroundColor. to make it easier to connect those options to where they are used. Perhaps IRendererOptions could be a union of things like BackgroundSystemOptions, ContextSystemOptions, etc instead of having them centralized in IRendererOptions. In the future we'd like to make systems more composable, this would help that.

packages/canvas-renderer/src/CanvasRenderer.ts
```diff
@@ -227,20 +227,8 @@ export class CanvasRenderer extends SystemManager<CanvasRenderer> implements IRe
...
-                clearBeforeRender: options.clearBeforeRender,
-            },
-            _view: {
-                height: options.height,
-                width: options.width,
-                autoDensity: options.autoDensity,
-                resolution: options.resolution,
-                view: options.view,
-            }
+            plugins: CanvasRenderer.__plugins,
```
bigtimebuddy, line 230: Another idea:

If plugins was removed here we would not need StartupOptions type at all. This is kind of an outlier anyways. Maybe after this.setup call we manually share the reference to plugins:

```js
this._plugins.initPlugins = CanvasRenderer.__plugins;
```

packages/core/src/startup/StartupSystem.ts
SuperSodaSea: Maybe with `GlobalMixins.IRendererOptions`?

bigtimebuddy: I think GlobalMixins is more about shared/external package typings. Most of this stuff is internal-core stuff so probably not necessary.

bigtimebuddy (approved):
I think this is good now. 

I did a few things:

* Decentralized the `IRendererOptions` and `IRendererOptionsAuto` and `IApplicationOptions` (now a union of types)
* Decentralized the `settings.RENDER_OPTIONS` defaults
* Remove redundant docs for `Renderer`, `CanvasRenderer`, `Application`, `autoDetectRenderer`
* ContextSystem, BackgroundSystem, StartupSystem, ViewSystem, etc now manage their option options and don't require  exposing `IRendererOptions`

PR #9119, opened 2023-01-30
Fix glob expansion only working with .add and not with .load

### LINKED ISSUE #9118: Bug: Assets: glob expansion only works with .add and not with .load
### Current Behavior

Glob nomenclature (ex: `file.{avif,webp,png}`) is not being expanded when you use `Assets.load` but it is being expanded with `Assets.add` or `Assets.addBundle`

### Expected Behavior

`Assets.load` should also expand the glob, this could allow for globbing inside other files like the `texture` field of a texture atlas. 

(I know we have a special format detection for texture atlases, but sometimes you don't need many jsons, you only need a texture globbed)

### Steps to Reproduce

```ts
// Error: [Assets] http://localhost:1234/texture.{avif,webp,png} could not be loaded as we don't know how to parse it, ensure the correct parser has being added
Assets.load({src:"./texture.{avif,webp,png}"}).then(asset => {
	console.log(asset)
})

// Works, downloads avif
Assets.add("individual", "./texture.{avif,webp,png}");
Assets.load("individual").then(asset => {
	console.log(asset)
});

// Works, download avif
Assets.addBundle("aux",{
	dummy:"./texture.{avif,webp,png}",
});
Assets.loadBundle("aux").then (asset => {
	console.log(asset)
});
```

### Environment

- **`pixi.js` version**: *e.g. 7.1.1*

### Possible Solution

_No response_

### Additional Information

_No response_

Zyie:
Fixes #9118

This approach now works and correctly loads the `avif` file
```ts
await Assets.load("./texture.{avif,webp,png}")
Assets.get("./texture.{avif,webp,png}")
```


bigtimebuddy (approved):
LGTM

GoodBoyDigital (approved):
love it!

PR #9120, opened 2023-01-30
Chore: Cleanup document for renderer options

### LINKED ISSUE #8073: `notMultiplied` for `useContextAlpha` renderer option missing in the docs
'By accident' I bumped into an example that turns of pre-multiply from the renderer (`transparent: 'notMultiplied'`), so I was wondering what else is possible to set the `transparent` option of the renderer to.

I couldn't find the option in the docs, so looked in the code. To find out the `useContextAlpha` seem to do the same thing, so looks like an alias. 

`useContextAlpha` is in fact mentioned in the docs. However it tells us nothing about a `'notMultiplied'` value and only tells us to not fiddle around with the setting.
image

IMO the `'notMultiplied'` option is an important option to know for people understanding the difference between straight and pre-multiplied and should at least be mentioned there in the docs as a possible value to put in there. 
According to the docs the option accepts only boolean values, but obviously this particular string is accepted too. 
It might be me, but it looks like this option is ONLY useful to use with the `notMultiplied` value in it, as when setting `useContextAlpha` to `false` it looks like the same thing as setting the `backgroundAlpha` to `0`, which would make the option obsolete IMO. But I didn't deep dive in the code, so I might be missing something here.

IMO the `notMultiplied` option should be mentioned in the docs.

- **`pixi.js` version**: 6.2.1

SuperSodaSea:
#### Description of change

Trying to cleanup `IRendererOptions` related documents. Previously these documents were not very consistent in different places, now they are much better. Also I rewritten some of the descriptions.

Document preview:
- `IRendererOptions` [Before](https://pixijs.download/v7.1.1/docs/PIXI.IRendererOptions.html) | [After](https://pixijs.download/chore/renderer-options-docs/docs/PIXI.IRendererOptions.html)
- `Renderer` [Before](https://pixijs.download/v7.1.1/docs/PIXI.Renderer.html#constructor) | [After](https://pixijs.download/chore/renderer-options-docs/docs/PIXI.Renderer.html#constructor)
- `CanvasRenderer` [Before](https://pixijs.download/v7.1.1/docs/PIXI.CanvasRenderer.html#constructor) | [After](https://pixijs.download/chore/renderer-options-docs/docs/PIXI.CanvasRenderer.html#constructor) (- WebGL Only options)
- `autoDetectRenderer` [Before](https://pixijs.download/v7.1.1/docs/PIXI.html#autoDetectRenderer) | [After](https://pixijs.download/chore/renderer-options-docs/docs/PIXI.html#autoDetectRenderer) (+ `forceCanvas`)
- `Application` [Before](https://pixijs.download/v7.1.1/docs/PIXI.Application.html#constructor) | [After](https://pixijs.download/chore/renderer-options-docs/docs/PIXI.Application.html#constructor) (+ `forceCanvas`, `resizeTo`, `autoStart`, `sharedTicker`)
- `settings.RENDER_OPTIONS` [Before](https://pixijs.download/v7.1.1/docs/PIXI.settings.html#RENDER_OPTIONS) | [After](https://pixijs.download/chore/renderer-options-docs/docs/PIXI.settings.html#RENDER_OPTIONS)

Fixes #8073. Cleanup for v6.x branch will be in another PR.

#### Pre-Merge Checklist

### DIFF
--- a/packages/app/src/Application.ts
+++ b/packages/app/src/Application.ts
@@ -62,36 +62,53 @@ export class Application<VIEW extends ICanvas = ICanvas>
     public renderer: IRenderer<VIEW>;
 
     /**
-     * @param {object} [options] - The optional renderer parameters.
-     * @param {boolean} [options.autoStart=true] - Automatically starts the rendering after the construction.
-     *     **Note**: Setting this parameter to false does NOT stop the shared ticker even if you set
-     *     options.sharedTicker to true in case that it is already started. Stop it by your own.
-     * @param {number} [options.width=800] - The width of the renderers view.
-     * @param {number} [options.height=600] - The height of the renderers view.
-     * @param {PIXI.ICanvas} [options.view] - The canvas to use as a view, optional.
-     * @param {boolean} [options.premultipliedAlpha=true] - Set to `false` to disable premultipliedAlpha.
-     * @param {boolean} [options.autoDensity=false] - Resizes renderer view in CSS pixels to allow for
-     *   resolutions other than 1.
-     * @param {boolean} [options.antialias=false] - Sets antialias
-     * @param {boolean} [options.preserveDrawingBuffer=false] - Enables drawing buffer preservation, enable this if you
-     *  need to call toDataURL on the WebGL context.
-     * @param {number} [options.resolution=PIXI.settings.RESOLUTION] - The resolution / device pixel ratio of the renderer.
-     * @param {boolean} [options.forceCanvas=false] - prevents selection of WebGL renderer, even if such is present, this
-     *   option only is available when using **pixi.js-legacy** or **@pixi/canvas-renderer** modules, otherwise
-     *   it is ignored.
-     * @param {number|string} [options.backgroundColor=0x000000] - The background color of the rendered area
-     *  (shown if not transparent). Also, accepts hex strings or color names (e.g., 'white').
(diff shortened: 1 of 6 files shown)


SuperSodaSea:
BTW it seems that `settings.RENDER_OPTIONS.legacy` is not used anywhere in the code (`Filter` do have a `legacy`, but it have nothing to do with the previous one). Should we remove it?

bigtimebuddy:
Yes, please remove legacy. That was used a long time ago with Android browser compatibility. Most of that is now irrelevant.

SuperSodaSea:
Oh, I meant to order them by their function (options for `ViewSystem`, `BackgroundSystem`, `ContextSystem`, etc.). If you insist on alphabetical order, I can change them.

bigtimebuddy:
Yeah, for most people, the systems are an implementation detail they don't care about. I think alpha is more useful for the majority audience.

SuperSodaSea:
I've changed the order of the options in the document to sort by alphabetical order. I keeped the order in `IRendererOptions` so that we can split them into multiple interfaces later as mentioned in <https://github.com/pixijs/pixijs/pull/9112#discussion_r1087834643>. (The member of interface will automatically sorted by Webdoc, so they are already alphabetical in the [generated document](https://pixijs.download/chore/renderer-options-docs/docs/PIXI.IRendererOptions.html).)

bigtimebuddy (approved):

PR #9124, opened 2023-02-01
Fix deprecation docs for global Settings mixins

bigtimebuddy:
One of the deprecated settings in 7.1.0 wasn't updated in display global.d.ts correctly.

### DIFF
--- a/packages/canvas-mesh/global.d.ts
+++ b/packages/canvas-mesh/global.d.ts
@@ -24,7 +24,7 @@ declare namespace GlobalMixins
 
     interface Settings
     {
-        /** @deprecated */
+        /** @deprecated since 7.1.0 */
         MESH_CANVAS_PADDING: number;
     }
 }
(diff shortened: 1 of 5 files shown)


SuperSodaSea (commented):
It seems that `TARGET_FPMS` in `@pixi/ticker` also need a `@deprecated`.
Also I prefer `@deprecated since 7.1.0` (with version information).

bigtimebuddy:
Good catch @SuperSodaSea. I fixed those.

SuperSodaSea (approved):
LGTM!
BTW do we need another branch for this, since we have already merged some features for v7.2?

bigtimebuddy:
For v7.1.x changes, I will cherry-pick these things to the main branch and do releases there. Once v7.2.0 is released and if we need to do further v7.1.x changes, I'll make a release branch like v6.x.

PR #9127, opened 2023-02-02
Import HTMLText API

bigtimebuddy:
## :gift: Added

* Imports existing `@pixi/text-html` as a new package
* Adds `@pixi/text-html` to `pixi.js` and `pixi.js-legacy` bundles

## About

Awhile ago I introduced HTMLText as a plugin: [@pixi/text-html](https://github.com/pixijs/html-text). That plugin is stable and we have resolved all the various issues. After some discussions with @GoodBoyDigital and @Zyie we think this should be standard for PixiJS moving forward because it fills in a few key gaps with Text or BitmapText.

### 👍 Advantages

* Performance is generally much better compared to Text, especially for long text strings
* More flexible options for doing `text-orientation` options with CSS (right-to-left, vertical, etc)
* Emoji support is free
* More flexible inline styles with HTML elements and CSS
* The code is simpler than Text or BitmapText, easier to maintain
* Support for `nowrap` and `pre-wrap` white-space options
* Support for `justify` text alignment

### 👎 Disadvantages

* Doesn't work on Node or Webworkers (yet)
* Doesn't work in Internet Explorer
* Rendering of the text (internal Texture) is async

## 🔗 Links

* [Documentation](https://pixijs.download/feature/text-html/docs/PIXI.HTMLText.html)
* [Demo](https://jsfiddle.net/bigtimebuddy/ybsp6td7/)


SuperSodaSea (commented):
I tried `npm i` and there seems to be an update in `package-lock.json`:

```diff
  "packages/text-html": {
+   "name": "@pixi/text-html",
    "version": "7.1.2",
    "license": "MIT"
  },
```

Does it matters?

SuperSodaSea:
I would like to add a `snapshotResolver` for Jest to make the `.snap` files save to `snapshots` directory instead of `__snapshots__`. If you like this change, I can make a PR for this.

bigtimebuddy:
Yeah, a PR to fix that would be good @SuperSodaSea. Jest's folder naming convention is so goofy.

Zyie (approved):

PR #9128, opened 2023-02-02
Support Texture borders (e.g., NineSlicePlane) with Spritesheets

CodeAndWeb:
##### Description of change
TexturePacker can export the 9-slice configuration of sprites (i.e. the width of the non-scalable borders) to the JSON sprite sheet file. With this change PixiJS can read these values and use them instead of the default 9-slice border width (10 pixels). Border widths passed by the user to the NineSlicePlane constructor always take precedence over these defaults.

##### Pre-Merge Checklist
- [x] Tests are included


packages/core/src/textures/Texture.ts
```diff
@@ -16,6 +16,19 @@ const DEFAULT_UVS = new TextureUvs();
 
 export type TextureSource = string | BaseTexture | ImageSource;
 
+/** INineSliceBorders stores the width of the non-scalable borders of a {@link PIXI.NineSlicePlane} texture */
+export interface INineSliceBorders
```
bigtimebuddy, line 20: Preference here would be to call this `ITextureBorders` instead of `INineSliceBorders`. This shouldn't be couple so close with NineSplicePlane, though I understand that's the primary use-case.

```suggestion
/** Stores the width of the non-scalable borders, for example when used with {@link PIXI.NineSlicePlane} texture */
export interface ITextureBorders
```

packages/core/src/textures/Texture.ts
```diff
@@ -105,6 +118,12 @@ export class Texture<R extends Resource = Resource> extends EventEmitter
      */
     public defaultAnchor: Point;
 
+    /**
+     * Default width of the non-scalable border that is used if 9-slice plane is created with this texture.
+     * (See {@link PIXI.NineSlicePlane})
+     */
+    public defaultNineSliceBorders?: INineSliceBorders;
```
bigtimebuddy, line 125: Same API-naming preference here, decouple from NineSlicePlane.

```suggestion
    public defaultBorders?: ITextureBorders;
```

bigtimebuddy (commented):
Overall this is awesome, just some minor nits about the API names. Thank you @CodeAndWeb!

packages/core/src/textures/Texture.ts
bigtimebuddy: Is there a reason why we wouldn't use a Rectangle here like the other Texture properties (trim, orig, frame, etc)

CodeAndWeb: Internally in TexturePacker we do indeed store the "inner rectangle" instead of border widths, with leftBorder=innerRect.x, rightBorder=sprite.width-innerRect.width-innerRect.y and so on. As PixiJS expects borders and other frameworks like Unity also use border values to configure 9-slice scaling, we thought that it's better to stay with borders.

bigtimebuddy (approved):
LGTM!

Zyie (approved):

kutu:
in json you use short notation of width/height as `w/h`
but for borders you choose long one instead of `l/t/r/b`

PR #9129, opened 2023-02-03
Add snapshotResolver for Jest

SuperSodaSea:
#### Description of change

Rename the path of Jest's snapshots from `__snapshots__` to `snapshots` using `snapshotResolver`.

Jest supports TypeScript snapshot resolver since 27.0.0 (See facebook/jest#8829), but unfortunately our Jest version is locked to 26 due to the limitation of jest-electron. So I keeped `test/jest-snapshot-resolver.js` and it should be quite simple to convert it to TypeScript.

Ref: #9127

#### Pre-Merge Checklist


bigtimebuddy (approved):
Thanks!

PR #9130, opened 2023-02-03
Fix wrong DOM_DELTA_* values by getting them from WheelEvent API

jramstedt:
##### Description of change

DOM_DELTA_* constants were set wrong.
This fixes them by referring them from WheelEvent.

Reference: [https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode)

##### Pre-Merge Checklist


SuperSodaSea (commented):
In my opinion, these constant need to be `static`. This could be a mistake.

Also I think it is better to keep these constants as magic numbers, this will make it easier for us to modify `@pixi/events` to support non-DOM environments later. (In Web Workers, Node.js, etc.)

jramstedt:
I agree they should be static.
I didn't want to change the API too much (just added the readonly)

SuperSodaSea:
@ShukantPal What about your opinion?

ShukantPal:
The original intention was to keep the API as similar to the DOM WheelEvent as possible. I suspect making these static would break that? Up to you if you want to maintain that pattern.

SuperSodaSea:
I checked `WheelEvent`'s definition and I found them to be both dynamic and static properties: 

https://github.com/microsoft/TypeScript/blob/00eb22030f92a0acf15ec7af3a0881c829aa54b0/lib/lib.dom.d.ts#L16751-L16768

```ts
/** Events that occur due to the user moving a mouse wheel or similar input device. */
interface WheelEvent extends MouseEvent {
    readonly deltaMode: number;
    readonly deltaX: number;
    readonly deltaY: number;
    readonly deltaZ: number;
    readonly DOM_DELTA_LINE: number;
    readonly DOM_DELTA_PAGE: number;
    readonly DOM_DELTA_PIXEL: number;
}

declare var WheelEvent: {
    prototype: WheelEvent;
    new(type: string, eventInitDict?: WheelEventInit): WheelEvent;
    readonly DOM_DELTA_LINE: number;
    readonly DOM_DELTA_PAGE: number;
    readonly DOM_DELTA_PIXEL: number;
};
```

So I think we can keep both definitions here.

```ts
    /** Units specified in pixels. */
    static readonly DOM_DELTA_PIXEL = 0;
    /** Units specified in pixels. */
    readonly DOM_DELTA_PIXEL = 0;

    /** Units specified in lines. */
    static readonly DOM_DELTA_LINE = 1;
    /** Units specified in lines. */
    readonly DOM_DELTA_LINE = 1;

    /** Units specified in pages. */
    static readonly DOM_DELTA_PAGE = 2;
    /** Units specified in pages. */
    readonly DOM_DELTA_PAGE = 2;
```

bigtimebuddy:
@jramstedt I think @SuperSodaSea's suggestion above is good to match WheelEvent. Do you think you could update this PR with his suggestions?

bigtimebuddy (approved):

SuperSodaSea (approved):

PR #9136, opened 2023-02-06
Chore: Add npm caching to GitHub Actions

bigtimebuddy:
Seeing if npm cache makes much of a difference

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -28,6 +28,7 @@ jobs:
       with:
         node-version: 16
         registry-url: 'https://registry.npmjs.org'
+        cache: 'npm'
     - name: Install npm
       run: npm install -g npm@8
     - name: Install Dependencies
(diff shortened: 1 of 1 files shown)


vikpe:
the first run will be slower as it has to generate a cache; second run should be faster.

bigtimebuddy:
I'll touch this again to see the delta. 

My comparison, here's a quick sample of the `npm ci` duration on Actions:

* 1:38
* 0:39
* 0:41
* 0:39
* 0:26

If we're looking at optimization opportunities, testing and building are the slowest. Maybe adding more parallelization to the build step would help speed-up.

bigtimebuddy:
After the touch:

* 0:52
* 0:32

Seems inline with priors. Thoughts?

vikpe:
does't seem to yield much of a difference.

bigtimebuddy:
I think article might have some useful stuff:
https://www.rupeshtiwari.com/caching-dependencies-on-github-workflows/

bigtimebuddy:
I'm going to close this as I don't have time to dig in. But if someone wants to try to get caching work for node_modules, that could help us save another minute or so on our CI.

SuperSodaSea:
> It uses [actions/cache](https://github.com/actions/cache) under the hood for caching **global** packages data but requires less configuration settings.

> **Note:** The action does not cache `node_modules`

So in order to cache `node_modules` we need to use `actions/cache` manually.

PR #9147, opened 2023-02-07
Fix: Invalid Color value when using multiply or round methods

### LINKED ISSUE #9146: Bug: `Color` still keeps its original `value` after it is modified
### Current Behavior

`Color` still keeps its original `value` (The `ColorSource` that created it) after it is modified.

### Expected Behavior

`Color.value` should be set to `null` after it is modified by `multiply` or `round`.

### Steps to Reproduce

```js
const color = new PIXI.Color(0xFFFFFF);
console.log(color.toRgbaString()); // rgba(255,255,255,1)
color.multiply(0xFF0000);
console.log(color.toRgbaString()); // rgba(255,0,0,1)
color.setValue(0xFFFFFF);
console.log(color.toRgbaString()); // rgba(255,0,0,1) (The correct result should be rgba(255,255,255,1))
```

### Environment

- **`pixi.js` version**: dev
- **Browser & Version**: (Any)
- **OS & Version**: (Any)
- **Running Example**: <https://pixiplayground.com/#/edit/YoeXFeZ1mzbKy-BCUJnte>

### Possible Solution

The issue is caused by the code below:

https://github.com/pixijs/pixijs/blob/a4409c20d757aaf6a17b5af58362437a87314120/packages/color/src/Color.ts#L163-L167

`Color.value` is meaningless after the `Color` is modified by `multiply` or `round`, so in my opinion it should be set to `null`.

### Additional Information

_No response_

bigtimebuddy:
Closes #9146

Invalidates Color's `value` property (to `null`) when using `multiply` or `round` methods.

### DIFF
--- a/packages/color/src/Color.ts
+++ b/packages/color/src/Color.ts
@@ -94,7 +94,7 @@ export class Color
     private static readonly HEX_PATTERN = /^(#|0x)?(([a-f0-9]{3}){1,2}([a-f0-9]{2})?)$/i;
 
     /** Internal color source, from constructor or set value */
-    private _value: Exclude<ColorSource, Color>;
+    private _value: Exclude<ColorSource, Color> | null;
 
     /** Normalized rgba component, floats from 0-1 */
     private _components: Float32Array;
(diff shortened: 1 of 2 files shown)


SuperSodaSea (approved):
LGTM!

PR #9161, opened 2023-02-14
Docs: update documentation

Zyie:
As i have been going through writing a markdown parser for webdoc I noticed a few issues with the docs

The main change here is that the docs for eums are now inlined. This means things like vscode can now get the descriptions when you hover over them

Also i fixed a whole bunch of `@link` that did not resolve because they were missing the `PIXI` namespace


bigtimebuddy (approved):
Nice PR!

SuperSodaSea (approved):

PR #9166, opened 2023-02-15
Fix: events firing when they shouldn't

### LINKED ISSUE #9156: Bug: huge performance penalty with pointermove with very large Containers
### Current Behavior

Even in non-interactive scenes with large numbers of elements (>>10000) in a `ParticleContainer`, just moving the mouse above the canvas leads to a FPS-drop of >30%.

Profiler shows that Pixi spends a lot of time in the event system, esp. `notifyTarget`:

image

I tried to set `interactive = false; interactiveChildren = false; hitArea = { contains: () => false }` wherever I could without effect.

Mouse wheel in contrast is for free (also seen in profiler).

I suspect that the same holds for `Container` instead of `ParticleContainer`, but I haven't investigated that.

### Expected Behavior

No performance penalty.
As it was in Pixi6.

### Steps to Reproduce

As an example, use Pixi's `ParticleContainer` example: https://pixijs.io/examples/#/demos-basic/particle-container.js

Push the number child elements to the edge where performance gets critical but still is around 60FPS (30000 on my business notebook). Then just moving mouse above the canvas gives a >30% performance drop.

### Environment

- **`pixi.js` version**: 7.1.2
- **Browser & Version**: any
- **OS & Version**: any
- **Running Example**: https://pixijs.io/examples/#/demos-basic/particle-container.js

### Possible Solution

_No response_

### Additional Information

_No response_

Zyie:
This PR fixes 2 issues

- the global move events fired even if a display object is not interactive
- if you set a child to be interactive, the parent would still receive the events even if it was not interactive
    ```
    const sprites = new PIXI.Container();
    sprites.interactive = false

    // previously this would still fire
    sprites.on('pointermove', ()=>{
        console.log('move')
    })

    app.stage.addChild(sprites);

    const totalSprites = 30;

    for (let i = 0; i < totalSprites; i++) {
        const dude = PIXI.Sprite.from('https://pixijs.io/examples/examples/assets/maggot_tiny.png');
        dude.interactive = true
        sprites.addChild(dude);
    }
    ```

This should also help fix #9156

### DIFF
--- a/packages/events/src/EventBoundary.ts
+++ b/packages/events/src/EventBoundary.ts
@@ -359,15 +359,24 @@ export class EventBoundary
 
         const children = target.children;
 
-        if (children)
+        const interactionNone = target._internalInteractive === 'none';
+        const interactionPassive = target._internalInteractive === 'passive' && !target.interactiveChildren;
+        const interactiveChildren = target.interactiveChildren;
+        const shouldIterateChildren = !interactionNone && interactiveChildren && !interactionPassive;
+
+        if (children && children.length > 0)
         {
-            for (let i = 0; i < children.length; i++)
+            if (shouldIterateChildren)
             {
-                this.all(e, type, children[i]);
+                for (let i = 0; i < children.length; i++)
+                {
+                    this.all(e, type, children[i]);
+                }
             }
         }
 
         e.currentTarget = target;
+
         this.notifyTarget(e, type);
     }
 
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Could you fix the failing test so I can test a build?

Zyie:
@bigtimebuddy looking at the failing test, it seems that events meant to bubble to up and fire an event, which is what causes my second point to happen

should that event only bubble up to non interactive targets? e.g. `static`/`dynamic`
or should `auto` be included in that?

bigtimebuddy:
Probably only interactive targets?

bigtimebuddy (approved):
Verified that this does, indeed, fix #9156 
https://jsfiddle.net/bigtimebuddy/w5fLm307/

PR #9167, opened 2023-02-15
Set default value for defaultInteraction to auto

bigtimebuddy:
`defaultInteraction`'s default value was `false` when it should be `auto`, to match documentation.

@Zyie, do you think a `false` value here is really necessary for backward compatibility?

### DIFF
--- a/packages/events/src/EventSystem.ts
+++ b/packages/events/src/EventSystem.ts
@@ -33,7 +33,7 @@ export class EventSystem
         ],
     };
 
-    private static _defaultInteraction: boolean | Interactive;
+    private static _defaultInteraction: Interactive;
 
     /**
      * The default interaction mode for all display objects.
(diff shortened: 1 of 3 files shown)


Zyie:
I think it is necessary as many people will check for it interactive is `true`/`false`

e.g
 ```
if (graphics.interactive) 
{
    ...
}
```

This would now break as the value would be `auto` instead of false

bigtimebuddy:
I find this whole deprecation strategy a little messy. I really wish there was a cleaner break/deprecation. It's helpful when we can just throw away the deprecation later without having to rewrite. I also don't like including a new API like `options.defaultInteraction` and having a mix of types.

I'd love to introduce a successor API to `interactive`. It would make the delineation between the old boolean interactive and the new string types much clearer. `eventMode`? Not super opinionated about the name but something maybe that's more aligned with Events packages, not InteractionManager.

PR #9168, opened 2023-02-15
Feat: Add new config options for loadTextures

### LINKED ISSUE #9158: Feature Request: Add a method to Disable Image loading via WebWorker
### Description

**Feature :**
Add the possibility to load img ressources the old way ( no webworker )

**Reason :**
Of course loading via WebWorker is great but some edge case environment can raise security issue doing so. Therefore it should be an easy way to disable it. 

**Example :** 
When in a WKWebview(Swift/iOS) loading an image in a webworker using a customSchemeURL(custom url protocol like pixijs://myimg.png) raise a security error.

Adding a property to disable the loading via WebWorker would solve this issue and probably other edge cases.

Thanks !

Zyie:
Fixes #9158 

Adds 2 new options for `loadTextures` to allow the user to prefer `createImageBitmap` and set `crossOrigin`

### DIFF
--- a/packages/assets/src/loader/parsers/textures/loadTextures.ts
+++ b/packages/assets/src/loader/parsers/textures/loadTextures.ts
@@ -45,6 +45,20 @@ export async function loadImageBitmap(url: string): Promise<ImageBitmap>
  * this makes use of imageBitmaps where available.
  * We load the ImageBitmap on a different thread using the WorkerManager
  * We can then use the ImageBitmap as a source for a Pixi Texture
+ *
+ * You can customize the behavior of this loader by setting the `config` property.
+ * ```js
+ * // Set the config
+ * import { loadTextures } from '@pixi/assets';
+ * loadTextures.config = {
+ *    // If true we will use a worker to load the ImageBitmap
+ *    preferWorkers: true,
+ *    // If false we will use new Image() instead of createImageBitmap
+ *    // If false then this will also disable the use of workers as it requires createImageBitmap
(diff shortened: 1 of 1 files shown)


bigtimebuddy (commented):
Thanks for this. 

My only question is about the API. We added an Assets setter for `preferWorkers` (I found myself having to use that a lot on sites like jsfiddle, so it was helpful). Is it okay you think to have these be local `config` options on loadTextures, or should we maybe find a way to set these things from `Assets.init` or pass-thru props like `preferWorkers`? I'm not sure what it should be, but it would nice if it were consistent.

IMO, it is to make `loadTextures` (and other loader middleware) be an opaque API that most users don't have to know, like resources or systems.

bigtimebuddy:
We discussed that we are going to make a generic way with `Assets.init` to set these properties so that `Assets` doesn't need to specifically know about the loadTextures extension.

bigtimebuddy (approved):
I'm going to approve this, since the thing we want to add is convenience on top of this. That can be done in another PR.

Makio64:
As someone who use Pixi since the begining I cant follow all the evolutions but having simple & explicit names always help : 

Here some name suggestion as an explicit complement to Assets.init : 
- Assets.setPreferences
- Assets.setLoaderConfig
- Assets.setLoaderPreferences 

Thanks for this PR 👍 
Keep rocking !

bigtimebuddy:
Good suggestion @Makio64. I added this #9184

PR #9169, opened 2023-02-15
Add ability to set Color object on `vec3` or `vec4` uniforms

GoodBoyDigital:
This PR alows devs to pass the color object as a uniform property and it get uploaded accordingly

eg:

```
const fragmentSrc = `
    uniform vec4 inputColor;

    void main() {

    gl_FragColor = vec4(inputColor);
}`;

const color = new Color('yellow');

const uniforms = { 
    inputColor: color,
};
```

### DIFF
--- a/packages/core/src/shader/utils/uniformParsers.ts
+++ b/packages/core/src/shader/utils/uniformParsers.ts
@@ -149,6 +149,62 @@ export const uniformParsers: IUniformParser[] = [
                     data[offset+3] = v.height;
                 `
     },
+    // upload a pixi color as vec4 with caching layer
+    {
+        test: (data: any, uniform: any): boolean =>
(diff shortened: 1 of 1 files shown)


packages/core/src/shader/utils/uniformParsers.ts
```diff
@@ -149,6 +149,62 @@ export const uniformParsers: IUniformParser[] = [
                     data[offset+3] = v.height;
                 `
     },
+    // upload a pixi color as vec4 with caching layer
+    {
+        test: (data: any, uniform: any): boolean =>
+            data.type === 'vec4' && data.size === 1 && !data.isArray && uniform.red !== undefined,
```
bigtimebuddy, line 155: Is there a reason you wouldn't just do `uniform instanceof Color` instead of checking `uniform.red !== undefined`?

vikpe:
Great! 👍

PR #9170, opened 2023-02-15
adds back getLocalPosition from v6

Zyie:
It was mentioned in the discord that getLocalPosition didn't exists anymore

So i've added it back to help with the transition from v6 -> v7

### DIFF
--- a/packages/events/src/FederatedMouseEvent.ts
+++ b/packages/events/src/FederatedMouseEvent.ts
@@ -1,8 +1,8 @@
 import { Point } from '@pixi/core';
 import { FederatedEvent } from './FederatedEvent';
 
-import type { DisplayObject } from 'pixi.js';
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Good one!

packages/events/src/FederatedMouseEvent.ts
```diff
@@ -1,6 +1,9 @@
 import { Point } from '@pixi/core';
 import { FederatedEvent } from './FederatedEvent';
 
+import type { DisplayObject } from 'pixi.js';
```
SuperSodaSea, line 4: Maybe `import type { DisplayObject } from '@pixi/display';`? Though circular references in `import type` won't be a big problem.

Zyie: Good catch!

PR #9172, opened 2023-02-15
Deprecate interaction for eventMode

Zyie:
We have now decided to fully deprecate `interactive` and move to using `eventMode` to describe the new interaction types

For backwards compatibility we map `interactive = false` to `eventMode = 'auto` and `interactive = true` to `eventMode = 'static'`

#9167 should be closed in favour of this as i believe i moved everything over from it

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -716,7 +716,7 @@ export abstract class DisplayObject extends utils.EventEmitter<DisplayObjectEven
         this.filterArea = null;
         this.hitArea = null;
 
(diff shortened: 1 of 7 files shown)


bigtimebuddy (commented):
Looks awesome. Can we change `defaultInteraction` to `defaultEventMode`?

bigtimebuddy:
Now that I think about it, what about eventMode instead of defaultEventMode for the Renderer options? We don't for instance use "defaultScaleMode", "defaultResolution"?

bigtimebuddy (approved):

PR #9181, opened 2023-02-18
Support ColorSource with Mesh & BitmapText tint

bigtimebuddy:
Support ColorSource for Mesh & BitmapText `tint`

```ts
const mesh = new Mesh();
mesh.tint = 'red';
```

Example:
https://jsfiddle.net/bigtimebuddy/4L6yebdx/


SuperSodaSea (commented):
Should we can add tests for this, just like #9180 and other similar PRs?

SuperSodaSea (approved):

Issue #9189, opened 2023-02-20
Bug: The requested module '/npm/@pixi/color@7.2.0-beta/+esm' does not provide an export named 'Color'


8Observer8:
### Current Behavior

Error message: The requested module '/npm/@pixi/color@7.2.0-beta/+esm' does not provide an export named 'Color' 

### Expected Behavior

There should be no errors.

### Steps to Reproduce

- Go to https://www.jsdelivr.com/
- Type: pixi
- Select 7.2.0-beta
- Click on ESM button
- Copy this link: https://cdn.jsdelivr.net/npm/pixi.js@7.2.0-beta/+esm
- Use this link to set up Pixi.js with importmap:

index.html

```html
<!DOCTYPE html>

<html lang="en">

<head>
    <meta charset="UTF-8">
    <title>Example</title>
</head>

<body style="caret-color: transparent;">
    <canvas id="renderCanvas" width="400" height="400"></canvas>

    <!-- Since import maps are not yet supported by all browsers, it is
        necessary to add the polyfill es-module-shims.js
        Source: https://threejs.org/docs/index.html#manual/en/introduction/Installation -->
    <script async src="https://unpkg.com/es-module-shims@1.3.6/dist/es-module-shims.js"></script>
    <script type="importmap">
        {
            "imports": {
                "pixi-js": "https://cdn.jsdelivr.net/npm/pixi.js@7.2.0-beta/+esm",
                "matter-js": "https://cdn.jsdelivr.net/npm/matter-js@0.19.0/+esm"
            }
        }
    </script>

    <script type="module" src="js/main.js"></script>
</body>

</html>
```

js/main.js

```js
// import * as PIXI from "pixi-js";
import { default as PIXI } from "pixi-js";
import { default as Matter } from "matter-js";

console.log(`Matter.js version: ${Matter.version}`);

console.log(PIXI);
```

### Environment

- **`pixi.js` version**: *7.2.0-beta*
- **Browser & Version**: *Chrome 110*
- **OS & Version**: *Windows 10*
- **Running Example**: 
- https://playcode.io/1211309
- https://plnkr.co/edit/cpTRfBpoIdl2ZPto

SuperSodaSea:
> https://cdn.jsdelivr.net/npm/@pixi/color@7.2.0-beta/+esm

```
/**
 * Failed to bundle using Rollup v2.74.1.
[...]

PR #9191, opened 2023-02-20
Ensure colord names plugin is not bundled

bigtimebuddy:
Fixes https://github.com/pixijs/pixijs/issues/9189#issuecomment-1437309750

### DIFF
--- a/bundles/pixi.js-node/package.json
+++ b/bundles/pixi.js-node/package.json
@@ -71,7 +71,7 @@
     "cross-fetch": "^3.1.5",
     "gl": "^6.0.1"
   },
-  "nodeDependencies": [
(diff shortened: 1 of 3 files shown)


SuperSodaSea (commented):
I prefer to map `dependencies` and `peerDependencies` into regexs rather than declaring them manually, which is easy to forget. For example:

```js
Object.keys(dependencies || {}).map(x => new RegExp(`^${x}(/.*)?$`))
```

bigtimebuddy:
Wanna make the PR for that? I was trying to do a quick fix, but your idea is better. Just don't have time at the moment

SuperSodaSea:
Yeah I would love to make a PR for this, maybe later.

PR #9199, opened 2023-02-22
Add support for Color input that extends Number

### LINKED ISSUE #9198: Feature Request: Allow `Color.prototype.normalize` to identify values which are Objects extending the `number` primitive.
### Description

In Foundry VTT, we are using our own `Color` class which is extending the `Number` primitive. Unfortunately, there is some clashes with `Color.prototype.normalize`, which is identifying our value as an `Object` and apply wrong logic.
We have a workaround for that, but maybe something could be done at the PIXI level?

bigtimebuddy:
Closes #9198 

Verify:
https://jsfiddle.net/bigtimebuddy/n83h1kw5/


bigtimebuddy:
cc @Feu-Secret

Zyie (approved):

PR #9200, opened 2023-02-22
Fix: Listeners registered with EventEmitter#once should only trigger once

### LINKED ISSUE #9195: Bug: EventEmitter#once triggers handlers multiple times (like on) in PixiJS v7
### Current Behavior

- The handler is triggered multiple times.

### Expected Behavior

- The handler should only be triggered once.
- Was working with PIXI v6.

### Steps to Reproduce

See the fiddle. Basically, create events with `once` and see the handler be called multiple times.

### Environment

- **`pixi.js` version**: *7.2.0-beta*
- **Browser & Version**: *Chrome 110.0.5481.104 (Official Build) (64-bit)*
- **OS & Version**: *Window 11*
- **Running Example**: *[https://jsfiddle.net/c09xvy27/6/](https://jsfiddle.net/c09xvy27/6/)*

### Possible Solution

_No response_

### Additional Information

_No response_

SuperSodaSea:
#### Description of change

Call `removeListener` on listeners registered with `once` in `EventBoundary#notifyListeners`, so that it can match the behavior of `EventEmitter#emit` (<https://github.com/primus/eventemitter3/blob/b99e3efb9d62392c7c7275d2f98d3d9dbc9b368f/index.js#L159-L216>).

Fixes #9195.

#### Pre-Merge Checklist

### DIFF
--- a/packages/events/src/EventBoundary.ts
+++ b/packages/events/src/EventBoundary.ts
@@ -56,13 +56,16 @@ type TrackingData = {
 };
 
 /**
+ * Internal storage of an event listener in EventEmitter.
+ * @ignore
+ */
+type EmitterListener = { fn(...args: any[]): any, context: any, once: boolean };
+
+/**
  * Internal storage of event listeners in EventEmitter.
  * @ignore
  */
-type EmitterListeners = Record<string,
-| Array<{ fn(...args: any[]): any, context: any }>
-| { fn(...args: any[]): any, context: any }
->;
+type EmitterListeners = Record<string, EmitterListener | EmitterListener[]>;
 
 /**
  * Event boundaries are "barriers" where events coming from an upstream scene are modified before downstream propagation.
@@ -1373,6 +1376,7 @@ export class EventBoundary
 
         if ('fn' in listeners)
         {
+            if (listeners.once) e.currentTarget.removeListener(type, listeners.fn, undefined, true);
             listeners.fn.call(listeners.context, e);
         }
         else
@@ -1382,6 +1386,7 @@ export class EventBoundary
                 i < j && !e.propagationImmediatelyStopped;
                 i++)
             {
+                if (listeners[i].once) e.currentTarget.removeListener(type, listeners[i].fn, undefined, true);
                 listeners[i].fn.call(listeners[i].context, e);
             }
         }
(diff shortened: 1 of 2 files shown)


packages/events/src/EventBoundary.ts
```diff
@@ -1382,6 +1386,7 @@ export class EventBoundary
                 i < j && !e.propagationImmediatelyStopped;
                 i++)
             {
+                if (listeners[i].once) e.currentTarget.removeListener(type, listeners[i].fn, undefined, true);
```
bigtimebuddy, line 1389: Does `removeListener` modify the listeners length? If so, this could potentially skip over listeners:
```js
target.on('click', () => {})
target.once('click', () => {})
target.on('click', () => {}) // <-- skipped?
```

SuperSodaSea: No, `removeListener` create a new array instead of modifying the original one (see the original code <https://github.com/primus/eventemitter3/blob/b99e3efb9d62392c7c7275d2f98d3d9dbc9b368f/index.js#L244-L292>).

SuperSodaSea: The original [`EventEmitter#emit`](https://github.com/primus/eventemitter3/blob/b99e3efb9d62392c7c7275d2f98d3d9dbc9b368f/index.js#L159-L216) do this the same way:

```js
    if (listeners.once) this.removeListener(event, listeners.fn, undefined, true);

    switch (len) {
      case 1: return listeners.fn.call(listeners.context), true;
      case 2: return listeners.fn.call(listeners.context, a1), true;
      case 3: return listeners.fn.call(listeners.context, a1, a2), true;
      case 4: return listeners.fn.call(listeners.context, a1, a2, a3), true;
      case 5: return listeners.fn.call(listeners.context, a1, a2, a3, a4), true;
      case 6: return listeners.fn.call(listeners.context, a1, a2, a3, a4, a5), true;
    }

    for (i = 1, args = new Array(len -1); i < len; i++) {
      args[i - 1] = arguments[i];
    }

    listeners.fn.apply(listeners.context, args);
```

bigtimebuddy (approved):
Thanks for this fix :+1:

PR #9203, opened 2023-02-23
Chore: Remove all dependencies from module bundles

### LINKED ISSUE #9189: Bug: The requested module '/npm/@pixi/color@7.2.0-beta/+esm' does not provide an export named 'Color'

SuperSodaSea:
#### Description of change

Move `ismobilejs` from `devDependencies` to `dependencies`, so that it will not be bundled into `@pixi/settings` and generating invalid source maps (see <https://unpkg.com/browse/@pixi/settings@7.2.0-beta.2/lib/external/ismobilejs/esm/>).

Also remove `rollup-plugin-rename-node-modules`, since there is no bundled dependency now.

Fixes #9189.

#### Pre-Merge Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10817,7 +10817,6 @@
     },
     "node_modules/ismobilejs": {
       "version": "1.1.1",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/isobject": {
@@ -19901,9 +19900,7 @@
       "license": "MIT",
       "dependencies": {
         "@pixi/constants": "file:../constants",
-        "@types/css-font-loading-module": "^0.0.7"
-      },
-      "devDependencies": {
+        "@types/css-font-loading-module": "^0.0.7",
         "ismobilejs": "^1.1.0"
       }
     },
@@ -27566,8 +27563,7 @@
       "version": "2.0.0"
     },
     "ismobilejs": {
-      "version": "1.1.1",
-      "dev": true
+      "version": "1.1.1"
     },
     "isobject": {
       "version": "3.0.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,6 @@
         "rollup": "^2.78.0",
         "rollup-plugin-esbuild": "^4.9.1",
         "rollup-plugin-jscc": "^2.0.0",
-        "rollup-plugin-rename-node-modules": "^1.3.1",
         "rollup-plugin-sourcemaps": "^0.4.2",
         "rollup-plugin-string": "^3.0.0",
         "semver": "^7.3.8",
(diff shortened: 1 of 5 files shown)


SuperSodaSea:
The problem is, when we are not bundling `ismobilejs`, `import isMobileCall from 'ismobilejs'` will have different result for cjs and esm:

- cjs (test:require): `[Function: isMobile]`
- esm (test:import): `{ __esModule: true, default: [Function: isMobile] }`

:thinking:

bigtimebuddy:
@SuperSodaSea we already have a fork of [rollup-plugin-rename-node-modules](https://github.com/pixijs/rollup-plugin-rename-node-modules/) which I'm using in extension scripts. I saw we make the change there and release a new version. The original plugin seems not active.

SuperSodaSea:
@bigtimebuddy Okay, I will try to work on that fork.

SuperSodaSea:
Made a little hack for this, hope it works.
Default exports are bad :(

bigtimebuddy:
Is `ismobilejs` our only other bundled dependency?

SuperSodaSea:
> Is `ismobilejs` our only other bundled dependency?

Yes, I searched for 'external' folder in the build result and it is the only one left.

bigtimebuddy:
Then we could also remove the logic in rollup.config.js for bundling devDependencies? Including: the rename node modules plugin is not needed anymore. Right?

SuperSodaSea:
Right, `rollup-plugin-rename-node-modules` is no longer needed after this. BTW Is there some settings in Rollup to detect if we have bundled unwanted external packages? (Adding `devDependencies` to `external` is not a good idea since these package will not being installed by users, may cause error.)

bigtimebuddy:
Not that I know of

SuperSodaSea:
Perhaps we can verify this by checking if there is a folder named `node_modules` in `dist`.

bigtimebuddy (approved):

bigtimebuddy:
If you want to add some validation script on postbuild, that's fine with me. Maybe something like [path-exists-cli](https://www.npmjs.com/package/path-exists-cli)?

```bash
path-exists packages/*/lib/external && exit 1 || exit 0
```

This will error if external exists in any of the packages.

bigtimebuddy:
I'm happy to merge this as-is and push another beta out. You can add another PR with the check for bundled modules, if you'd like. I don't think it's totally necessary. A better check is probably looking at invalid sourcemaps, which is ultimately the issues with jsdelivr, right?

SuperSodaSea:
Yeah, I agree.

PR #9209, opened 2023-02-24
Resolution isn't readonly in current renderer implementation

baseten:
`Irenderer` currently lists `resolution` as `readonly` but the `Renderer` implementation has a setter for `resolution`: https://github.com/pixijs/pixijs/blob/dev/packages/core/src/Renderer.ts#L459


GoodBoyDigital (approved):
nice! thanks @baseten 👍

PR #9211, opened 2023-02-26
Add strict null checks to packages with no errors

SuperSodaSea:
#### Description of change

Add packages that already have no error with strict null checks to the filter in `scripts/filterTypeScriptErrors.ts`:

- `bundles/pixi.js/`
- `bundles/pixi.js-legacy/`
- `packages/canvas-text/`
- `packages/constants/`
- `packages/filter-alpha/`
- `packages/filter-displacement/`
- `packages/filter-fxaa/`
- `packages/filter-noise/`
- `packages/mixin-get-global-position/`
- `packages/settings/`
- `packages/text-html/`
- `packages/unsafe-eval/`

Also `types:strict` is added back to `npm run dist` (so we can check it easily on CI), but now it will not break the build when error occurs (`types:strict` will not exit with non-zero code).

#### Pre-Merge Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "build:types": "tsc -p tsconfig.types.json && ts-node scripts/fixTypes.ts",
     "postbuild:types": "copyfiles -u 1 \"out/**/*\" . && run-s clean:out",
     "watch": "rollup -cw",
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
Thank you @SuperSodaSea

PR #9217, opened 2023-02-28
Migrate @pixi/node bundle

bigtimebuddy:
This change re-homes the `@pixi/node` bundle to https://github.com/pixijs/node

There are a few reasons for this move:

1. GitHub Actions is slightly slower because it requires extra dependencies and extras steps
2. It's more difficult to `npm install` the pixijs repo locally without stubbing your toe on node bundle dependencies
3. The adapter and extensions APIs are stable enough in v7 to support this move
4. This would allow us to remove a handful of exceptions in our build system for the node bundle
5. This bundle has a small audience  (261 weekly downloads vs 121K for pixi.js) and probably not worth all the overhead of the above to justify the cost


bigtimebuddy:
PR Build times:
* 5m 14s
* 5m 31s
* 6m 40s
* 5m 56s
* 7m 51s

**Average: 6m 14s**

Previous Build times:
* 7m 9s
* 8m 17s
* 7m 53s
* 6m 0s
* 6m 42s

**Average: 7m 12s**


So it's ~1 minute faster on GitHub Actions.

SuperSodaSea (approved):

PR #9218, opened 2023-02-28
Fix: Renderer#type not available

SuperSodaSea:
#### Description of change

I suddenly realized that `app.renderer.type` is always `undefined`. This PR fixes this issue.

#### Pre-Merge Checklist


bigtimebuddy (approved):
Good catch

PR #9222, opened 2023-03-01
Fix: Ensure prepare and extract systems are available from IRenderer

### LINKED ISSUE #9207: Bug: Pixi@7.1.2 renderer.extract is missing in type declarations
### Current Behavior

VSCode:

```ts

const app = new PIXI.Application()
app.renderer.extract // Property 'extract' does not exist on type 'IRenderer<ICanvas>'.ts(2339)
app.renderer.prepare // Same issue

```
It is only type issue, `extract` is present in runtime.

### Expected Behavior

`renderer.extract` and `render.prepare` should be defined according to https://github.com/pixijs/pixijs/wiki/v7-Migration-Guide#extract-and-prepare-systems.

### Steps to Reproduce

Code above in VSCode.

### Environment

- **`pixi.js`**: 7.1.2
- **OS & Version**: MacOS 12.6
- **Running Example**: not reproducible in playground https://pixiplayground.com/#/edit/r-0u8TgLNRyU5thWn2i5K
- **Typescript** 4.9.5

### Possible Solution

Looking through types, found `extract` only here:
```ts
// @pixi/extract/global.d.ts
declare namespace GlobalMixins
{
    interface Renderer
    {
        readonly extract: import('@pixi/extract').Extract;
    }
}
```
Including it via tsconfig does not solve the issue, could it be that declaration merging fails somewhere down the line?

### Additional Information

My `tsconfig.json` for reference:
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "useDefineForClassFields": true,
    "module": "ESNext",
    "lib": ["ESNext", "DOM"],
    "moduleResolution": "Node",
    "jsx": "react-jsx",
    "strict": true,
    "noImplicitAny": true,
    "noImplicitOverride": true,
    "noImplicitThis": true,
    "noUncheckedIndexedAccess": true,
    "resolveJsonModule": true,
    "isolatedModules": true,
    "esModuleInterop": true,
    "noEmit": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "skipLibCheck": true
  },
  "include": ["src"]
}
```

bigtimebuddy:
Closes #9207

### Fixed

* Fix: Ensure `prepare` and `extract` extensions are available from `IRenderer`

### Before
```ts
import { Application, Renderer } from 'pixi.js';

(app.renderer as Renderer).extract.base64(); // throws type error without cast
```

### After
```ts
import { Application } from 'pixi.js';

app.renderer.extract.base64();
```

### DIFF
--- a/packages/canvas-extract/package.json
+++ b/packages/canvas-extract/package.json
@@ -37,6 +37,7 @@
   "peerDependencies": {
     "@pixi/canvas-renderer": "file:../canvas-renderer",
     "@pixi/core": "file:../core",
-    "@pixi/display": "file:../display"
+    "@pixi/display": "file:../display",
+    "@pixi/extract": "file:../extract"
   }
 }
--- a/packages/canvas-extract/src/CanvasExtract.ts
+++ b/packages/canvas-extract/src/CanvasExtract.ts
@@ -3,6 +3,7 @@ import { extensions, ExtensionType, Rectangle, RenderTexture, utils } from '@pix
 import type { CanvasRenderer } from '@pixi/canvas-renderer';
 import type { BaseRenderTexture, ExtensionMetadata, ICanvas, ISystem } from '@pixi/core';
 import type { DisplayObject } from '@pixi/display';
+import type { IExtract } from '@pixi/extract';
 
 const TEMP_RECT = new Rectangle();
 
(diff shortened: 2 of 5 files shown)


SuperSodaSea (approved):
:+1:

PR #9226, opened 2023-03-01
Add ability to override loadParser when loading assets

### LINKED ISSUE #8907: Asset Loader: Could not be loaded as we don't know how to parse it, ensure the correct parser has being added
Using https://pixijs.io/guides/basics/assets.html I get console warning:

"[Assets] https://imagedelivery.net/AS_oRD30zjCHIwJTbRDRrw/023d2eb0-fa3d-4043-5548-06b1b72fab00/700 could not be loaded as we don't know how to parse it, ensure the correct parser has being added"

I'm assuming it expects a file extension in the url like ".png" or ".webp". However I'm using cloudflare for CDN image delivery, and it doesn't allow for file extensions. It sends the most efficient format per their docs: https://developers.cloudflare.com/images/cloudflare-images/serve-images/ 

Any suggestions? i.e.

a) Can I give pixi.js asset loader a hint as to what the file format will be?
b) Can pixi.js figure out the format based on the response the way a browser would if this image url was in the ``?

GoodBoyDigital:
added ability to override and specify your own loadParser when loading elements:

```
       await loader.load({
            src: `${serverPath}textures/bunny.other`,
            loadParser: 'loadTextures' // will force it to be handled as a texture
        });
```

fixes #8907


packages/assets/src/loader/Loader.ts
```diff
@@ -211,4 +242,29 @@ export class Loader
     {
         return this._parsers;
     }
+
+    /** validates our parsers, right now it only checks for name conflicts but we can add more here as required! */
+    private _validateParsers()
```
bigtimebuddy, line 247: There's a function called `extensions.handleByList` at the bottom of **Assets.ts**. This is responsible for receiving the parser extensions and sorting the parsers list.

For instance, one of the problems with this is if an extension is added after load is called once. Not unimaginable, though probably rare.

We could add a custom function to handle the extensions install and update the parserHash at the same time. This seems like the better place to check for duplicates anyways.

packages/assets/src/loader/Loader.ts
```diff
@@ -45,29 +49,51 @@ export class Loader
...
+                        break;
+                    }
+                }
 
-                return null;
+                if (!parser)
+                {
+                    // #if _DEBUG
+                    // eslint-disable-next-line max-len
+                    console.warn(`[Assets] ${url} could not be loaded as we don't know how to parse it, ensure the correct parser has being added`);
```
bigtimebuddy, line 87:
```suggestion
                    console.warn(`[Assets] ${url} could not be loaded as we don't know how to parse it, ensure the correct parser has been added`);
```

packages/assets/src/loader/Loader.ts
```diff
@@ -45,29 +49,51 @@ export class Loader
...
-                if (parser.load && parser.test?.(url, data, this))
+                if (!parser)
                 {
-                    asset = await parser.load(url, data, this);
-                    result.parser = parser;
-
-                    break;
+                    // #if _DEBUG
+                    // eslint-disable-next-line max-len
+                    console.warn(`[Assets] specified loadParser ${data.loadParser} not whilst found attempting to load ${url}`);
```
bigtimebuddy, line 64:
```suggestion
                    console.warn(`[Assets] specified load parser "${data.loadParser}" not found while loading ${url}`);
```

packages/assets/src/loader/Loader.ts
GoodBoyDigital: hmm.. could be a good case for using a Proxy?

bigtimebuddy: That could work too

GoodBoyDigital: nice, yeah - feel like Proxy works well here - keeps things nice and simple!

GoodBoyDigital:
not sure why the TS is throwing an error here. If i make the name optional it works.. but don't really want to do that. Any ideas @bigtimebuddy or @Zyie ?

packages/assets/src/loader/parsers/LoaderParser.ts
```diff
@@ -42,6 +42,9 @@ export interface LoaderParser<ASSET = any, META_DATA = any, CONFIG = Record<stri
     /** A config to adjust the parser */
     config?: CONFIG;
 
+    /** The name of the parser (this can be used when specifying loadParser in a LoadAsset) */
+    name: string;
```
bigtimebuddy, line 46: This should probably be optional, right? `_validateParsers` handles it as optional.

packages/assets/src/loader/Loader.ts
```diff
@@ -206,9 +250,28 @@ export class Loader
...
+                if (this._parserHash[parser.name])
+                {
+                    // #if _DEBUG
+                    console.warn(`[Assets] loadParser name conflict "${parser.name}"`);
+                    // #endif
+                }
+
+                this._parserHash[parser.name] = parser;
+            }
+        }
```
bigtimebuddy, line 275: Could we do something like this instead?

```suggestion
        this._parserHash = this._parsers
            .filter(parser => parser.name)
            .reduce((hash, parser) =>
            {
                if (hash[parser.name])
                {
                    // #if _DEBUG
                    console.warn(`[Assets] loadParser name conflict "${parser.name}"`);
                    // #endif
                }
                
                return { ...hash, [parser.name]: parser };
            }, {});
```

bigtimebuddy (approved):

PR #9228, opened 2023-03-02
Fix interaction performance

Zyie:
This PR fixes the performance issues seen with the global move events

before: https://jsfiddle.net/k69qz4c2/3/
after: https://jsfiddle.net/k69qz4c2/2/

When comparing the two there is no longer lag when the `eventMode` is `passive` or `auto`


GoodBoyDigital (approved):
👍

bigtimebuddy (approved):

PR #9246, opened 2023-03-08
Fix: interaction performance

Zyie:
Changes:

- We can now turn on/off different interactions through renderer settings
I've conceptualised these as `move`, `globalMove`, `click`, and `wheel`
```ts
/**
 * The event types that are enabled by the EventSystem
 * @type {object}
 * @readonly
 * @since 7.2.0
 * @property {boolean} move - Enables pointer events associated with pointer movement:
 * - `pointermove` / `mousemove` / `touchmove`
 * - `pointerout` / `mouseout`
 * - `pointerover` / `mouseover`
 * @property {boolean} globalMove - Enables global pointer move events:
 * - `globalpointermove`
 * - `globalmousemove`
 * - `globaltouchemove`
 * @property {boolean} click - Enables pointer events associated with clicking:
 * - `pointerup` / `mouseup` / `touchend` / 'rightup'
 * - `pointerupoutside` / `mouseupoutside` / `touchendoutside` / 'rightupoutside'
 * - `pointerdown` / 'mousedown' / `touchstart` / 'rightdown'
 * - `click` / `tap`
 * @property {boolean} wheel - Enables wheel events.
 */
public static get defaultEvents()
{
    return this._defaultEvents;
}
```
These should reduce the number of hit tests that are done if you do not care about certain events.

- Previously we were looping through the entire scene 3 times per move event. Now we only loop through the scene once.. This does mean that the global pointer events now happen after mouse or touch events
  - before `pointermove`, `globalpointermove`, `mousemove`, `globalmousemove`, `touchmove`, `globaltouchmove`
  - after `pointermove`, `mousemove`, `touchmove`, `globalpointermove`, `globalmousemove`, `globaltouchmove`

### DIFF
--- a/packages/events/src/EventBoundary.ts
+++ b/packages/events/src/EventBoundary.ts
@@ -5,6 +5,7 @@ import { FederatedPointerEvent } from './FederatedPointerEvent';
 import { FederatedWheelEvent } from './FederatedWheelEvent';
 
 import type { DisplayObject } from '@pixi/display';
+import type { EmitterListeners, TrackingData } from './EventBoundaryTypes';
 import type { FederatedEvent } from './FederatedEvent';
 import type {
     Cursor, EventMode, FederatedEventHandler,
@@ -17,55 +18,6 @@ const PROPAGATION_LIMIT = 2048;
 
 const tempHitLocation = new Point();
 const tempLocalMapping = new Point();
-
-/**
- * The tracking data for each pointer held in the state of an {@link PIXI.EventBoundary}.
- *
- * ```ts
- * pressTargetsByButton: {
- *     [id: number]: FederatedEventTarget[];
- * };
- * clicksByButton: {
- *     [id: number]: {
- *         clickCount: number;
- *         target: FederatedEventTarget;
- *         timeStamp: number;
- *     };
- * };
- * overTargets: FederatedEventTarget[];
- * ```
- * @typedef {object} TrackingData
- * @property {Record.<number, PIXI.FederatedEventTarget>} pressTargetsByButton - The pressed display objects'
- *  propagation paths by each button of the pointer.
- * @property {Record.<number, object>} clicksByButton - Holds clicking data for each button of the pointer.
- * @property {PIXI.DisplayObject[]} overTargets - The DisplayObject propagation path over which the pointer is hovering.
- * @memberof PIXI
- */
-type TrackingData = {
-    pressTargetsByButton: {
-        [id: number]: FederatedEventTarget[];
-    };
-    clicksByButton: {
-        [id: number]: {
-            clickCount: number;
-            target: FederatedEventTarget;
-            timeStamp: number;
-        }
-    };
-    overTargets: FederatedEventTarget[];
-};
-
-/**
- * Internal storage of an event listener in EventEmitter.
- * @ignore
- */
-type EmitterListener = { fn(...args: any[]): any, context: any, once: boolean };
-
-/**
- * Internal storage of event listeners in EventEmitter.
- * @ignore
- */
-type EmitterListeners = Record<string, EmitterListener | EmitterListener[]>;
 
 /**
  * Event boundaries are "barriers" where events coming from an upstream scene are modified before downstream propagation.
@@ -154,6 +106,8 @@ export class EventBoundary
      */
     public moveOnAll = false;
 
+    public allowGlobalPointerEvents = true;
+
     /**
      * Maps event types to forwarding handles for them.
      *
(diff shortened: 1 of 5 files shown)


Zyie:
This all now seems to be working

Before: https://jsfiddle.net/TobiasW/7qf6har0/32
After: https://jsfiddle.net/kv46xsjr/

From my testing we now get roughly the same performance as v6 pixi

packages/events/src/EventSystem.ts
```diff
@@ -45,6 +77,38 @@ export class EventSystem implements ISystem<EventSystemOptions>
...
+     * - `globalmousemove`
+     * - `globaltouchemove`
+     * @property {boolean} click - Enables pointer events associated with clicking:
+     * - `pointerup` / `mouseup` / `touchend` / 'rightup'
+     * - `pointerupoutside` / `mouseupoutside` / `touchendoutside` / 'rightupoutside'
+     * - `pointerdown` / 'mousedown' / `touchstart` / 'rightdown'
+     * - `click` / `tap`
+     * @property {boolean} wheel - Enables wheel events.
+     */
+    public static get defaultEvents()
```
bigtimebuddy, line 107: Can we re-think static usage here? You're only referencing this internally.

For instance, if you wanted to change something _after_ construction, you'd do this:
```ts
import { Renderer, EventSystem } from 'pixi.js';

EventSystem.defaultEvents.wheel = false;
```

I think this is better:

```ts
import { Renderer } from 'pixi.js';

renderer.events.features.wheel = false; // or another name for "features"
```

The benefit is that it allow us to isolate Renderer instance of those people creating multiple renderers at once.

bigtimebuddy: That said, I think `_defaultEvents` (object above) should be public, but internally we should create an object on the instance.

Zyie: I've removed the static member in favour for `private _features`

I don't think we should be exposing these options outside of the renderer initialisation otherwise we have to handle adding / removing the DOM events which currently only happens on `init`

bigtimebuddy: That's fine with me. We can easily add that later, better to keep the public API small as possible

Zyie:
Tidied up the documentation and added the ability to dynamically enable/disable features with `renderer.event.features.move = false`

packages/events/src/EventBoundary.ts
```diff
@@ -136,17 +134,23 @@ export class EventBoundary
...
     /** Whether or not to collect all the interactive elements from the scene. Enabled in `pointermove` */
     private _collectInteractiveElements = false;
 
-    /** @param rootTarget - The holder of the event boundary. */
-    constructor(rootTarget?: DisplayObject)
+    /**
+     * @param rootTarget - The holder of the event boundary.
+     * @param eventSystem - The event system that manages this boundary.
+     */
+    constructor(rootTarget?: DisplayObject, eventSystem?: EventSystem)
```
bigtimebuddy, line 150: Preference here to not passing in the EventSystem. EventBoundary only needs to know about `globalMove`.

Zyie: I think this is required as the user could override `features` entirely
```
renderer.events.features = {
    move: true,
    globalMove: false,
    click: false,
    wheel: false
}
```
So in order to not lose the reference I need to pass the EventSystem?

bigtimebuddy: The goal here is to limit the scope and prevent coupling of EventSystem and EventBoundary. EventBoundary not knowing anything about EventSystem is a much cleaner design.

There's no reason we should allow re-assignment of `features`, only set individual props. I don't think there's anything wrong with that. Devs can use Object.assign, if need be. I think the most common case is turning off a flag.

Second, you could make features a Proxy (like Mat just did in Assets), and handle setting `globalMove` to update a property on EventBoundary:

```js
this.features = new Proxy({...EventSystem.defaultEventFeatures}, {
  set: (target, key, value) => {
    if (key === 'globalMove') {
       this.rootBoundary.globalMoveFeature = value;
    }
    target[key] = value;
    return true;
  }
);
```

Zyie: Yep forgot we could use proxies now!

packages/events/src/EventSystem.ts
```diff
@@ -106,6 +173,25 @@ export class EventSystem implements ISystem<EventSystemOptions>
...
+     *
+     * // to override all features use Object.assign
+     * Object.assign(app.renderer.events.features, {
+     *  move: false,
+     *  globalMove: false,
+     *  click: false,
+     *  wheel: false,
+     * })
+     */
+    public features: EventSystemFeatures;
```
bigtimebuddy, line 193: Making this readonly would prevent user setting this prop (but would still allow members of EventSystemFeatures to be set).
```suggestion
    public readonly features: EventSystemFeatures;
```

bigtimebuddy (approved):
This is great. Thank you for optimizing this.

PR #9255, opened 2023-03-10
Feature: Expose global pointer event data

Zyie:
In pixi v6 there was `interaction.mouse` which allowed you to get the global position of the mouse without listening to an event

This PR exposes this data as `events.pointer`


bigtimebuddy (approved):
Nice!

PR #9261, opened 2023-03-11
Make MSDF fragment shader "alpha" more like it was "coverage"

jakerdy:
## Problem

Hi, i am working on app where i need to draw zoomable node graphs. And in order to make text look good and work fast with big number of nodes, I decided to switch to BitmapText with MSDF Fonts.

After some testing I noticed, that text became less readable compared to PIXI.Text.

I have exactly same problem in one of my previous projects, where I made my own text rendering using MSDF. The problem is in fragment shader that doesn't take in account blending gamma correction. My change make it more valid, not 100% but goodenough from perception point of view.

## Theory

Text rendering, especially on small sizes and low dpi (non-retina) has many pitfals, and one of them - gamma correction. To make text look good and readable, you must use gamma corrected blending. Otherwise, text wouldn't be readable if you change "light-theme" to "dark-theme". Black text on white background always gonna be thicker, than white text on black bg.

Of course, to be correct in 100% cases you need to know, color of text AND background to apply correct gamma correction. But, in 99% of cases when you make GUI's that has good contrast, you could made an asumption that if your text bright, most likely bg will be dark, and wise versa.

More on that problem: 
- [Maxim Shemanarev about Gamma and Text rendering in general](https://agg.sourceforge.net/antigrain.com/research/font_rasterization/index.html#toc0008)
- [About gamma correction in general](https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/)

## Solution
So that you could just take luminance of text (luma of tint color in fram shader), and use it as parameter to lerp between '1' and '1/2.2' gamma. This trick makes alpha more like it was actual coverage of the glyphs. And in most situations you will get same text thickness on black-white and white-black situations.

## Examples
Iv'e made couple examples what will happend if you apply my changes.

```
!!! To get it right, open this examples in 100% scale without resizing on non retina display (96dpi). !!!
Otherwise interpolation will make text blurry.
```

### Synhetic tests: 

PIXI.Text:
00 - Text

PIXI.BitmapText with BitmapFont:
01 - BitmapText

PIXI.BitmapText with MSDF Font (original implementation):
02 - MSDF Current

PIXI.BitmapText with MSDF Font (my implementation):
03 - MSDF Patched

### Real App: 

PIXI.BitmapText with MSDF Font (original implementation):
04 - App with MSDF_Current

PIXI.BitmapText with MSDF Font (my implementation):
05 - App With MSDF_Patched png

PIXI.Text
06 - App Native Text

As you might see, gamma corrected text looks more like `PIXI.Text` which i was used originally.

P.S.
Also there is some problems with layout of MSDF text (all my examples uses same font sizes). Probably, it comes from atlas generator which uses some other metrics, but i didn't dig into that. Not that big problem though, as wrong text thikness...

### DIFF
--- a/packages/text-bitmap/src/shader/msdf.frag
+++ b/packages/text-bitmap/src/shader/msdf.frag
@@ -28,7 +28,11 @@ void main(void) {
     alpha = 1.0;
   }
 
-  // NPM Textures, NPM outputs
-  gl_FragColor = vec4(uColor.rgb, uColor.a * alpha);
+  // Gamma correction for coverage-like alpha
+  float luma = dot(texColor.rgb, vec3(0.299, 0.587, 0.114));
+  float gamma = mix(1.0, 1.0/2.2, luma);
+  float coverage = pow(uColor.a * alpha, gamma);  
 
+  // NPM Textures, NPM outputs
+  gl_FragColor = vec4(uColor.rgb, coverage);
 }
--- a/packages/text-bitmap/src/shader/msdf.frag
+++ b/packages/text-bitmap/src/shader/msdf.frag
@@ -30,7 +30,7 @@ void main(void) {
 
   // Gamma correction for coverage-like alpha
   float luma = dot(uColor.rgb, vec3(0.299, 0.587, 0.114));
-  float gamma = mix(1.0, 1.0/2.2, luma);
+  float gamma = mix(1.0, 1.0 / 2.2, luma);
   float coverage = pow(uColor.a * alpha, gamma);  
 
   // NPM Textures, NPM outputs
--- a/packages/text-bitmap/src/shader/msdf.frag
+++ b/packages/text-bitmap/src/shader/msdf.frag
@@ -29,7 +29,7 @@ void main(void) {
   }
 
   // Gamma correction for coverage-like alpha
-  float luma = dot(texColor.rgb, vec3(0.299, 0.587, 0.114));
+  float luma = dot(uColor.rgb, vec3(0.299, 0.587, 0.114));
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):
I agree this is objectively better. Thanks for the visual examples really helped! Here's another before and after to put a detailed point on it.

detail

jakerdy:
Wait! Don't merge, iv'e made a mistake... In shader, i calculating luma on pixels from atlas, but should actualy calc luma of tint which is in vertex colors... I'll fix it right now. (Screens already updated in main post)

packages/text-bitmap/src/shader/msdf.frag
```diff
@@ -28,7 +28,11 @@ void main(void) {
     alpha = 1.0;
   }
 
-  // NPM Textures, NPM outputs
-  gl_FragColor = vec4(uColor.rgb, uColor.a * alpha);
+  // Gamma correction for coverage-like alpha
+  float luma = dot(uColor.rgb, vec3(0.299, 0.587, 0.114));
+  float gamma = mix(1.0, 1.0/2.2, luma);
```
SuperSodaSea, line 33:
```suggestion
  float gamma = mix(1.0, 1.0 / 2.2, luma);
```

Just a little formatting :)

miltoncandelero:
I can't say that I fully understand the math involved but I understand the idea of perceived brightness so I get the idea.

The text does look sharper and crispier so its a LGTM 🥳

bigtimebuddy:
Thank you @jakerdy!

PR #9262, opened 2023-03-12
Chore: Bump workflow improvements

bigtimebuddy:
### Fixed

* Add confirmation to the bump
* Make sure to fail when attempting to bump using the same version
* Better bump order (patch, minor, major) instead of (major, minor, patch)


SuperSodaSea (approved):
LGTM!

PR #9266, opened 2023-03-14
Add drawRoundedShape to graphics-extras

koteelok:
Added new drawing method to creating any shape with rounded corners.
Very handy, i think.

### Example
image

### Demo
https://codesandbox.io/s/drawroundedpath-examples-oewil0


packages/graphics-extras/src/index.ts
```diff
@@ -14,6 +15,7 @@ export interface IGraphicsExtras
     drawFilletRect: typeof drawFilletRect;
     drawRegularPolygon: typeof drawRegularPolygon;
     drawRoundedPolygon: typeof drawRoundedPolygon;
+    drawRoundedPath: typeof drawRoundedPolygon;
```
SuperSodaSea, line 18:
```suggestion
    drawRoundedPath: typeof drawRoundedPath;
    drawRoundedPolygon: typeof drawRoundedPolygon;
```

SuperSodaSea (commented):
One thing I concern about is that should we use `arc()` or `quadraticCurveTo()`, just like [`drawRoundedPolygon`](https://github.com/pixijs/pixijs/blob/c8a3513c15beb8e3b0dd3ce9d695fab59688f45f/packages/graphics-extras/src/drawRoundedPolygon.ts)? :thinking:

koteelok:
Added choice to use either arc or quadratic curve to make rounded shapes.

bigtimebuddy (approved):
Great contribution @koteelok, thank you!

packages/graphics-extras/src/drawRoundedShape.ts
```diff
@@ -0,0 +1,222 @@
...
+                drawDirection = true;
+            }
+        }
+        else if (angle > 0)
+        {
+            radDirection = -1;
+            drawDirection = true;
+        }
+
+        const pRadius = p2.radius || radius;
```
SuperSodaSea, line 67:
```suggestion
        const pRadius = p2.radius ?? radius;
```

Maybe it would be better to use `??` here, for example if someone wants to set `radius: 0` at a point to make it a sharp corner.

packages/graphics-extras/src/drawRoundedShape.ts
```diff
@@ -0,0 +1,222 @@
...
+            thisPoint,
+            lastPoint,
+            lastOffsetDistance / lastEdgeLength
+        );
+
+        const nextEdgeLength = distance(nextPoint, thisPoint);
+        const nextOffsetDistance = Math.min(
+            nextEdgeLength / 2,
+            thisPoint.radius || radius
+        );
```
SuperSodaSea, line 170:
```suggestion
        const pRadius = thisPoint.radius ?? radius;

        const lastEdgeLength = distance(lastPoint, thisPoint);
        const lastOffsetDistance = Math.min(lastEdgeLength / 2, pRadius);
        const start = pointLerp(
            thisPoint,
            lastPoint,
            lastOffsetDistance / lastEdgeLength
        );

        const nextEdgeLength = distance(nextPoint, thisPoint);
        const nextOffsetDistance = Math.min(nextEdgeLength / 2, pRadius);
```

SuperSodaSea (commented):
It seems that `roundedShapeArc` behaves strangely when radius is large :thinking:

image

koteelok:
> It seems that `roundedShapeArc` behaves strangely when radius is large 🤔
> 
> image

Fixed that in new commit.

koteelok:
Well, the calculations for the first position of the arc were not correct. 🤔
 I fixed that, so everything **_should_** be fine now.

SuperSodaSea (commented):
Looks good now, with one improvement: when `radius <= 0`, we don't need to draw the arcs / quadratic curves, just `lineTo` the corner. This can also eliminate the strange shapes that appear when `radius < 0`.

SuperSodaSea (approved):
I fixed 2 issues myself:

- Wierd shape when radius < 0
- Lines disappear when two corners are too close

Now it looks good for me. Sandbox: <https://codesandbox.io/s/drawroundedshape-example-forked-2wlpcy>

Sadly the automatically generated CodeSandbox is still randomly failing :(

PR #9278, opened 2023-03-17
Add `defaultAutoResolution` to Filter

viniciusbitt:
##### Description of change

When I was developing my application, I noticed that the filters have a resolution of 1 by default, I think it would be more convenient if the resolution by default was equal to the `renderer` resolution like in `Text`.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/filters/Filter.ts
+++ b/packages/core/src/filters/Filter.ts
@@ -213,7 +213,7 @@ export class Filter extends Shader
      * // filter has a resolution of 2
      * const filter = new GlowFilter();
      */
-    public static defaultResolution = 1;
+    public static defaultResolution: number;
 
     /**
      * Default filter samples for any filter.
--- a/packages/core/src/filters/Filter.ts
+++ b/packages/core/src/filters/Filter.ts
@@ -1,4 +1,5 @@
 import { MSAA_QUALITY } from '@pixi/constants';
+import { settings } from '@pixi/settings';
 import { Program } from '../shader/Program';
 import { Shader } from '../shader/Shader';
 import { State } from '../state/State';
@@ -188,8 +189,29 @@ import type { FilterSystem } from './FilterSystem';
 export class Filter extends Shader
 {
     /**
-     * Default filter resolution for any filter.
-     * @static
+     * Override whether or not the resolution of the filter is automatically adjusted to match the resolution
+     * of the renderer.
+     * Setting this to false can allow you to get crisper filter at lower render resolutions.
+     * @example
+     * // renderer has a resolution of 1
+     * const app = new Application();
+     *
+     * Filter.defaultResolution = 2;
+     * Filter.defaultAutoResolution = false;
+     * // filter has a resolution of 2
+     * const filter = new GlowFilter();
+     */
+    public static defaultAutoResolution = true;
+
+    /**
+     * If {@link PIXI.Filter.defaultAutoResolution} is false, this will be the default resolution of the filter.
+     * If not set it will default to {@link PIXI.settings.RESOLUTION}.
+     * @example
+     * Filter.defaultResolution = 2;
+     * Filter.defaultAutoResolution = false;
+     *
+     * // filter has a resolution of 2
+     * const filter = new GlowFilter();
      */
     public static defaultResolution = 1;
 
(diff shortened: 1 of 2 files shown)


packages/core/src/filters/Filter.ts
```diff
@@ -297,6 +327,13 @@ export class Filter extends Shader
 
     set resolution(value: number)
     {
+        this._autoResolution = false;
+
+        if (this._resolution === value)
+        {
+            return;
+        }
+
```
bigtimebuddy, line 336: We don't really need this bail-out, since it's only re-assigning a number.

```suggestion
        this._autoResolution = false;
```

packages/core/src/filters/Filter.ts
```diff
@@ -188,10 +189,31 @@ import type { FilterSystem } from './FilterSystem';
...
+     * @example
+     * // renderer has a resolution of 1
+     * const app = new Application();
+     *
+     * Filter.defaultResolution = 2;
+     * Filter.defaultAutoResolution = false;
+     * // filter has a resolution of 2
+     * const filter = new GlowFilter();
+     */
+    public static defaultAutoResolution = true;
```
bigtimebuddy, line 204: The behavior here of `true` would be a breaking change since now the Filter, by default, with inherit from Renderer. This breaks from the current behavior of defaulting to 1. My suggestion is we set this to `false` and change it to `true` in our next major release.

packages/core/src/filters/Filter.ts
```diff
@@ -230,7 +252,14 @@ export class Filter extends Shader
     /** The WebGL state the filter requires to render. */
     state: State;
 
-    protected _resolution: number;
+    /**
+     * The resolution / device pixel ratio of the canvas.
+     *
+     * This is set to automatically match the renderer resolution by default, but can be overridden by setting manually.
+     * @default PIXI.settings.RESOLUTION
+     */
```
bigtimebuddy, line 260: Could you mark these docs at `@private` so that they aren't included in the docs? These properties are only public for FilterSystem and users should not be changing them.

bigtimebuddy (commented):
I like the parallel design as Text. Just want to be mindful about breaking behavior here that may bite some users.

packages/core/src/filters/Filter.ts
bigtimebuddy: The other context about this is: in general, when faced with a design choice of performance vs being visually-rich, Pixi general prefers performance and opt-in to visually-richness (e.g., resolution, antialias, multisample). Filters can be expensive, especially at full Renderer resolution. That was one of the intentional decisions for defaulting to `1`. Unlike Text, filters "break" batching of Sprites (i.e., rendering with a slower, de-optimized way). For these reasons, I also think `false` is better out-of-the-box performance in many cases.

packages/core/src/filters/Filter.ts
bigtimebuddy: You know what, there's a lot of leaked private properties in the docs. We'll do a clean-up later of these things.

viniciusbitt:
I agree with @bigtimebuddy about the `Filter.detectAutoResolution` defaulting to `false`, about performance can be a big breaking change if it is `true`. Can be merged now since it's `false` by default.

bigtimebuddy (approved):
LGTM! Thank you

dev7355608:
We shouldn't use the renderer's resolution but the resolution of the current render target: so either the resolution of the renderer if we render on screen, or the resolution of the current render texture. We wouldn't want the renderer's resolution to affect filter resolution when we render to a texture, which could have a different resolution than the renderer.

```js
renderTextureSystem.current?.resolution ?? renderer.resolution
```

dev7355608:
We should probably not add `defaultAutoResolution` and `_autoResolution` to track auto resolution. Instead we should use the render target's resolution if `filter.resolution === null`. `MaskData.resolution` already works like that, which is the sprite mask filters resolution.
https://github.com/pixijs/pixijs/blob/311cafdce291068748bd0c0d73fa272ef2b4a474/packages/core/src/mask/MaskData.ts#L48-L53

`cacheAsBitmapResolution` also uses `null` for the renderer's resolution.
https://github.com/pixijs/pixijs/blob/311cafdce291068748bd0c0d73fa272ef2b4a474/packages/mixin-cache-as-bitmap/src/index.ts#L74-L83

bigtimebuddy:
Thanks for the PR @viniciusbitt, but I'm closing this in favor of #9298.

PR #9279, opened 2023-03-20
CodeSandbox CI configuration

baseten:
This PR adds configuration for CodeSandbox CI along with a basic local sandbox template which will be forked when testing PRs.

Because the monorepo packages don't contain their own version numbers an additional script was also necessary to write these so that CodeSandbox CI can publish to its registry.


.codesandbox/ci.json
```diff
@@ -0,0 +1,41 @@
+{
+  "buildCommand": "codesandbox-ci",
+  "packages": [
+    "bundles/pixi.js",
+    "bundles/pixi.js-webworker",
```
Zyie, line 5: we are missing `pixi.js-legacy` here

.codesandbox/ci.json
```diff
@@ -0,0 +1,41 @@
...
+    "packages/mixin-get-global-position",
+    "packages/particle-container",
+    "packages/prepare",
+    "packages/sprite",
+    "packages/sprite-animated",
+    "packages/sprite-tiling",
+    "packages/spritesheet",
+    "packages/text",
+    "packages/text-bitmap",
+    "packages/text-html"
```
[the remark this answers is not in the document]
baseten: Nice, I actually just pushed this change as well 😄 

I was initially trying to avoid spamming the CodeSandbox CI registry with too many modules, but I don't think it will make a difference

baseten:
Annoyingly globs don't seem to work, the CSB registry doesn't have a working pixi version: https://codesandbox.io/s/pixi-js-sandbox-63pdvm

SuperSodaSea:
Version at commit 3a7bf64454f64a9f93466c8d4e3464375063b829 (just publish everything) is working: <https://codesandbox.io/s/pixi-js-sandbox-jurgtz>
I think this is probably a random bug of CodeSandbox.

SuperSodaSea:
It looks like others have encountered this problem:
- <https://github.com/codesandbox/codesandbox-client/issues/3748>
- <https://github.com/codesandbox/codesandbox-client/issues/6825>

SuperSodaSea:
After triggering rebuild x2 it is working again: <https://codesandbox.io/s/pixi-js-sandbox-sco4xi>

bigtimebuddy (approved):
Nice work @baseten

package.json
```diff
@@ -32,6 +32,7 @@
     "watch": "rollup -cw",
     "dist": "run-s lint types types:strict build build:types docs",
     "postdist": "copyfiles -f bundles/*/dist/* dist && copyfiles -f packages/*/dist/* dist/packages",
+    "codesandbox-ci": "run-s dist prepublish-ci",
```
bigtimebuddy, line 35: Would this be more efficient if we remove the doc building, linting and type checking? Is that necessary?

```suggestion
    "codesandbox-ci": "run-s build build:types prepublish-ci",
```

SuperSodaSea:
↑ Should we revert this commit since we have #9282?

baseten:
> 

Yes, it's in pr

baseten:
> ↑ Should we revert this commit since we have #9282?

Yes, in progress already 👍

baseten:
@bigtimebuddy I've made your suggested script change 👍 

If possible, assuming you're happy with https://github.com/pixijs/pixijs/pull/9282 can we hold off merging this until that is merged and then we should be able to see the whole workflow run once this PR is merged 😄

PR #9280, opened 2023-03-20
Fix: Clone and deep equal the Color's value when setting

### LINKED ISSUE #9277: Bug: New 🌈 Color change broke pixi-spine tint
### Current Behavior

It's hard to explain :) Actually, it's better to compare the images.
I am pretty sure it's related to these two lines in SpineBase:
https://github.com/pixijs/spine/blob/7903e9dae81a855a6725d13789f31ca70260ede8/packages/base/src/SpineBase.ts#L328
https://github.com/pixijs/spine/blob/7903e9dae81a855a6725d13789f31ca70260ede8/packages/base/src/SpineBase.ts#L386
Do you have any idea why it broke pixi-spine? 

image

### Expected Behavior

image

### Steps to Reproduce

I think it's easy to reproduce with any spine project that has tint stuff

### Environment

- **`pixi.js` version**: *7.2.0*
- **`pixi-spine` version**: *4.0.3*

### Possible Solution

_No response_

### Additional Information

_No response_

bigtimebuddy:
Closes #9277

Copy the value internally and do a deep compare before re-setting. 

This avoids situations like this:

```js
const value = [0, 0, 0];
const color = new Color(value);
value[0] = 1;
color.setValue(value);
console.assert(color.red === 1); // v7.2.1 would be false, red would still be 0
```

### DIFF
--- a/packages/color/src/Color.ts
+++ b/packages/color/src/Color.ts
@@ -176,7 +176,7 @@ export class Color
         // Support copying from other Color objects
         if (value instanceof Color)
         {
-            this._value = value._value;
+            this._value = this.cloneSource(value._value);
             this._int = value._int;
             this._components.set(value._components);
         }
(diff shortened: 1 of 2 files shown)


packages/color/src/Color.ts
```diff
@@ -176,25 +176,97 @@ export class Color
...
+            this._value = this.cloneSource(value._value);
             this._int = value._int;
             this._components.set(value._components);
         }
         else if (value === null)
         {
             throw new Error('Cannot set PIXI.Color#value to null');
         }
-        else if (this._value !== value)
+        else if (!this.isSourceEqual(this._value, value))
```
SuperSodaSea, line 187:
```suggestion
        else if (this._value === null || !this.isSourceEqual(this._value, value))
```

SuperSodaSea (commented):
`_value` can be `null` here. Thanks `strictNullChecks`!

SuperSodaSea (approved):
LGTM!

PR #9281, opened 2023-03-20
Fix: Ensure input color channels are clamped

bigtimebuddy:
### Fixed

* Fix: Ensure input color channels are clamped 0-1


GoodBoyDigital (approved):
looking good!

PR #9282, opened 2023-03-21
Add trigger workflow to update dev package on beta.pixijs.com

baseten:
This PR adds a github workflow trigger to call the `update-dev` event on `beta.pixijs.com`

I've confirmed the workflow runs as expected in a fork of `pixi.js`, including taking the commit sha from the last commit of the PR rather than the commit where the PR is actually merged:

pixi.js forked repo triggering the action:
Screenshot 2023-03-21 at 16 44 19

beta.pixijs.com receiving the action:
Screenshot 2023-03-21 at 16 49 15

### DIFF
--- a/.github/workflows/trigger-website-build.yml
+++ b/.github/workflows/trigger-website-build.yml
@@ -3,8 +3,8 @@ name: Trigger Website Build
 on:
   pull_request:
     branches:
-      - deploy
-    types: [closed]
+      - dev
+    types: [ closed ]
 
 jobs:
   trigger-website-build:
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-website-build.yml
```diff
@@ -0,0 +1,23 @@
+name: Trigger Website Build
+
+on:
+  pull_request:
+    branches:
+      - dev
```
bigtimebuddy, line 6: We should redeploy on main probably? That is more stable.

bigtimebuddy (approved):

PR #9283, opened 2023-03-21
Fix passing unclamped values to Color

Zyie:
Using a value for `this.alpha` on a Graphics object over 1 would crash the app

A common use case is using a tween to fade the alpha and depending on the ease the number can go above/below 0-1
```
await gsap.to(this, {
    alpha: 1, // alpha to 1
    ease: Back.easeOut.config(1.7),
});
```

### DIFF
--- a/packages/color/src/Color.ts
+++ b/packages/color/src/Color.ts
@@ -8,7 +8,7 @@ import type {
     HsvaColor,
     HsvColor,
     RgbaColor,
-    RgbColor,
+    RgbColor
 } from 'colord/types';
 
 extend([namesPlugin]);
@@ -53,6 +53,8 @@ export type ColorSource = string | number | number[] | Float32Array | Uint8Array
 | HslColor | HslaColor | HsvColor | HsvaColor | RgbColor | RgbaColor | Color |
 // eslint-disable-next-line @typescript-eslint/ban-types
 Number;
+
+type ColorSourceTypedArray = Float32Array | Uint8Array | Uint8ClampedArray;
 
 /**
  * Color utility class.
@@ -471,7 +473,7 @@ export class Color
      */
     setAlpha(alpha: number): this
     {
-        this._components[3] = Math.min(1, Math.max(0, alpha));
+        this._components[3] = this._clamp(alpha, 0, 1);
 
         return this;
     }
@@ -545,7 +547,7 @@ export class Color
             && value.length >= 3 && value.length <= 4)
         {
             // make sure all values are 0 - 1
-            value = value.map((v) => Math.min(1, Math.max(0, v)));
+            value = this._clamp(value, 0, 1);
 
             const [r, g, b, a = 1.0] = value;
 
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):

packages/color/src/Color.ts
```diff
@@ -471,7 +471,7 @@ export class Color
      */
     setAlpha(alpha: number): this
     {
-        this._components[3] = alpha;
+        this._components[3] = Math.min(1, Math.max(0, alpha));
```
bigtimebuddy, line 474: Could you make a clamp function and reuse it here?

Zyie:
@GoodBoyDigital @bigtimebuddy updated to override arrays instead of creating a new one

packages/color/src/Color.ts
```diff
@@ -597,10 +602,31 @@ export class Color
...
         this._int = (((r * 255) << 16) + ((g * 255) << 8) + (b * 255 | 0));
     }
+
+    /**
+     * Clamps values to a range. Will override original values
+     * @param value - Value(s) to clamp
+     * @param min - Minimum value
+     * @param max - Maximum value
+     */
+    private _clamp<T extends number | number[] | ColorSourceTypedArray>(value: T, min: number, max: number): T
```
bigtimebuddy, line 618:
```suggestion
    private _clamp<T extends number | number[] | ColorSourceTypedArray>(value: T, min: number = 0, max: number = 1): T
```

packages/color/src/Color.ts
```diff
@@ -597,10 +602,31 @@ export class Color
     private refreshInt(): void
     {
         // Clamp values to 0 - 1
-        this._components.forEach((value, i) => (this._components[i] = Math.min(Math.max(value, 0), 1)));
+        this._clamp(this._components, 0, 1);
```
bigtimebuddy, line 605:
```suggestion
        this._clamp(this._components);
```

packages/color/src/Color.ts
```diff
@@ -542,10 +544,11 @@ export class Color
         }
         else if ((Array.isArray(value) || value instanceof Float32Array)
             // Can be rgb or rgba
-            && value.length >= 3 && value.length <= 4
-            // make sure all values are 0 - 1
-            && value.every((v) => v <= 1 && v >= 0))
+            && value.length >= 3 && value.length <= 4)
         {
+            // make sure all values are 0 - 1
+            value = this._clamp(value, 0, 1);
```
bigtimebuddy, line 550:
```suggestion
            value = this._clamp(value);
```

packages/color/src/Color.ts
```diff
@@ -471,7 +473,7 @@ export class Color
      */
     setAlpha(alpha: number): this
     {
-        this._components[3] = alpha;
+        this._components[3] = this._clamp(alpha, 0, 1);
```
bigtimebuddy, line 476:
```suggestion
        this._components[3] = this._clamp(alpha);
```

bigtimebuddy (approved):
Minor suggestions to set min/max defaults. Otherwise looks good

PR #9294, opened 2023-03-24
Fix `Color.round`

dev7355608:
The current `Color.round` does absolutely nothing. No rounding at all.

I assume that's how the rounding is supposed to be. Should the alpha component be rounded too?

### DIFF
--- a/packages/color/src/Color.ts
+++ b/packages/color/src/Color.ts
@@ -480,18 +480,16 @@ export class Color
 
     /**
      * Rounds the specified color according to the step. This action is destructive, and will
-     * override the previous `value` property to be `null`.
-     * @param step - Number of steps which will be used as a cap when rounding colors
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Here's the old implementation, for reference:
https://github.com/pixijs/pixijs/blob/39c67e00abe2dd592bf43d23b2559e646ab9791b/packages/canvas-renderer/src/canvasUtils.ts#L270-L281

This is only used when tinting a Sprite with canvas.

bigtimebuddy (approved):

dev7355608:
Is it fine that the alpha channel isn't rounded? The old implementation was just for RGB. `Color.round` could round the alpha channel too without breaking anything, I think. It probably makes more sense if all components are rounded.

bigtimebuddy:
I'm not even sure we need round anymore, TBH

PR #9295, opened 2023-03-24
Avoid object creation in `Color.normalize`

dev7355608:
`Color.shared.setValue(...)` shouldn't produce any garbage if possible as it might be called every frame multiple times.


bigtimebuddy (commented):
Good fix. Can you fix the lints?

dev7355608:
`npm run lint` tells me `warning  It's not necessary to initialize 'r: number' to undefined  no-undef-init` if I initialize with `undefined`. Without explicit initialization it passes: no complaints about "used before being assigned". What should I do?

bigtimebuddy (approved):

PR #9296, opened 2023-03-25
Deprecate the round and roundColor methods

bigtimebuddy:
### Changed

* Deprecates `canvasUtils.roundColor`
* Deprecates `Color.round`

There was an opportunity to simplify the code and remove the rounding (which didn't work anyway prior to #9294). Removes a bunch of redundant inline code for convert an int to hex string.

### Verify

* https://codesandbox.io/s/pixi-js-sandbox-forked-sqlly9

### DIFF
--- a/packages/canvas-renderer/src/canvasUtils.ts
+++ b/packages/canvas-renderer/src/canvasUtils.ts
@@ -263,7 +263,7 @@ export const canvasUtils = {
     roundColor: (color: number): number =>
     {
         // #if _DEBUG
-        utils.deprecation('7.3.0', 'PIXI.canvasUtils.roundColor is deprecated, use PIXI.Color.round instead');
+        utils.deprecation('7.3.0', 'PIXI.canvasUtils.roundColor is deprecated');
         // #endif
 
         return Color.shared
@@ -275,6 +275,7 @@ export const canvasUtils = {
     /**
      * Number of steps which will be used as a cap when rounding colors.
      * @memberof PIXI.canvasUtils
+     * @deprecated since 7.3.0
      * @type {number}
      */
     cacheStepsPerColorChannel: 8,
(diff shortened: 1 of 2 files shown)


packages/canvas-renderer/src/canvasUtils.ts
```diff
@@ -260,14 +255,22 @@ export const canvasUtils = {
...
+     * @deprecated since 7.3.0
+     * @see PIXI.Color.round
      * @param {number} color - the color to round, should be a hex color
      * @returns {number} The rounded color.
      */
     roundColor: (color: number): number =>
-        Color.shared
+    {
+        // #if _DEBUG
+        utils.deprecation('7.3.0', 'PIXI.canvasUtils.roundColor is deprecated, use PIXI.Color.round instead');
```
dev7355608, line 266: If both are deprecated, does this instruction make sense?

packages/canvas-renderer/src/canvasUtils.ts
```diff
@@ -260,14 +255,22 @@ export const canvasUtils = {
...
+        // #endif
+
+        return Color.shared
             .setValue(color)
             .round(canvasUtils.cacheStepsPerColorChannel)
```
dev7355608, line 271: `cacheStepsPerColorChannel` should probably be to be deprecated too, right?

packages/color/src/Color.ts
```diff
@@ -482,6 +482,7 @@ export class Color
      * Rounds the specified color according to the step. This action is destructive, and will
      * override the previous `value` property to be `null`. The alpha component is not rounded.
      * @param steps - Number of steps which will be used as a cap when rounding colors
+     * @deprecated since 7.3.0
      */
     round(steps: number): this
     {
```
dev7355608, line 488: Do we need a deprecation warning here?

bigtimebuddy: The deprecation function is in utils which would make a circular dependency between color and utils. I didn't want to change the package dependencies just for this, it's fine without the warning. No one is using `round` anyways.

dev7355608 (approved):

PR #9304, opened 2023-03-27
Fix: crash in unsafe-eval uniformSync and remove unreachable code

### LINKED ISSUE #9303: Bug: unsafe-eval uniformSync crashes on boolean uniforms
### Current Behavior

With pixi/unsafe-eval imported setting a boolean uniform throws an exception: 
```
TypeError: Cannot create property 'v' on boolean 'false'

          52 |         if (cv !== v)
          53 |         {
        > 54 |             cv.v = v;
             |                 ^
          55 |             gl.uniform1i(location, Number(v));
          56 |         }
          57 |     },

      at bool (packages/unsafe-eval/src/syncUniforms.ts:54:17)
      at Object.syncUniforms (packages/unsafe-eval/src/syncUniforms.ts:222:43)
      at ShaderSystem.syncUniforms (packages/unsafe-eval/src/install.ts:41:17)
      at ShaderSystem.syncUniformGroup (packages/core/src/shader/ShaderSystem.ts:149:18)
      at Object.syncUniforms (packages/unsafe-eval/src/syncUniforms.ts:112:33)
      at ShaderSystem.syncUniforms (packages/unsafe-eval/src/install.ts:41:17)
      at ShaderSystem.syncUniformGroup (packages/core/src/shader/ShaderSystem.ts:149:18)
      at ShaderSystem.bind (packages/core/src/shader/ShaderSystem.ts:117:18)
      at Object.<anonymous> (packages/unsafe-eval/test/unsafe-eval.tests.ts:46:25)
```

This seems to be a leftover of a change made in https://github.com/pixijs/pixijs/commit/662d09a31fb1f6705abf543411b5fa5a2247d50b which was fixed in the core package in https://github.com/pixijs/pixijs/commit/f9cb23670f60f9f3bbd4a3c05d885ce749d11698 but the same section of unsafe-eval was overlooked. The same typo is in the function for float uniforms as well, but that code is unreachable since floats have special handling in syncUniforms. 

Will create a PR with a suggested fix and a little bit of cleanup. For us this issue is fairly critical since we depend on being able to use PIXI with a strict CSP. 

### Expected Behavior

Same behaviour as when running PIXI without importing pixi/unsafe-eval

### Steps to Reproduce

import @pixi/unsafe-eval

Any shader with boolean uniforms should trigger the exception.

### Environment

- **`pixi.js` version**: *7.2.2, also likely in latest versions of pixi.js 5 and 6*
- **Browser & Version**: *e.g. Chrome 111*
- **OS & Version**: *Windows 10*

### Possible Solution

_No response_

### Additional Information

We discovered this while updating a project from PIXI v5.3.3 to v7. Judging by the age of the code this bug has been present for almost 2 years, so it's surprising this has never been reported by anyone else.

olems:
##### Description of change
Fixes crash with boolean uniforms when using the unsafe-eval package.
Fixes #9303 

Also removed some unreachable code that made it confusing to debug. The GLSL_TO_SINGLE_SETTERS function array had a function for floats that had the same typo as the function for bools, but would never be run since floats are handled further down. The same is the case for several other types, so I see no reason to leave them in.
This file could use some work in general, it should be kept in sync with the mainline shader code in the core package. I see there are some types that have special handling in shader/utils/uniformParsers that are not handled in this one. 

Added a basic test just for floats and bools. The test will crash with an exception as described in #9303 if the fix is not applied and passes otherwise. 

##### Pre-Merge Checklist

### DIFF
--- a/packages/unsafe-eval/test/unsafe-eval.tests.ts
+++ b/packages/unsafe-eval/test/unsafe-eval.tests.ts
@@ -32,7 +32,6 @@ void main() {
     afterAll(() =>
     {
         renderer.destroy();
-        renderer = null;
     });
 
     it('should be able to set float and bool uniforms', () =>
--- a/packages/unsafe-eval/test/unsafe-eval.tests.ts
+++ b/packages/unsafe-eval/test/unsafe-eval.tests.ts
@@ -1,0 +1,50 @@
+import { Renderer, Shader, UniformGroup } from '@pixi/core';
+import '@pixi/unsafe-eval';
+
+describe('unsafe-eval', () =>
+{
+    const vertexSrc = `
+
+uniform bool testBool;
+uniform float testFloat;
+
+void main() {
+    bool t1 = testBool;
+    float t2 = testFloat;
+    if(!t1) t2 = 0.0;
+    gl_Position = vec4(t2, t2, t2, t2);
+}`;
+
+    const fragmentSrc = `
+void main() {
+
+    gl_FragColor = vec4(1., 1., 1., 1.) ;
+
+}`;
+
+    let renderer: Renderer;
+
+    beforeAll(() =>
+    {
+        renderer = new Renderer();
+    });
+
+    afterAll(() =>
+    {
+        renderer.destroy();
+        renderer = null;
+    });
+
+    it('should be able to set float and bool uniforms', () =>
+    {
+        const testBool = true;
+        const testFloat = 1.0;
+        const group1 = UniformGroup.from({ testBool });
+        const uniforms = { testFloat, group: group1 };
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):
looks good!

packages/unsafe-eval/test/unsafe-eval.tests.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    beforeAll(() =>
+    {
+        renderer = new Renderer();
+    });
+
+    afterAll(() =>
+    {
+        renderer.destroy();
+        renderer = null;
```
bigtimebuddy, line 35:
```suggestion
```

PR #9305, opened 2023-03-28
fix: cache can-use-blend-mode value

romgrk:
##### Description of change
Cache the value of the can-use-blend-mode check. It shows up in my stack traces more than I'd like to, I have many small canvas elements.

image

##### Pre-Merge Checklist


bigtimebuddy (approved):
Seems like a reasonable optimization. Thanks @romgrk!

PR #9309, opened 2023-03-28
Fix: Add background and autoDensity types to IRenderer

### LINKED ISSUE #9308: Bug: background does not exist on type IRenderer
### Current Behavior

Our application use pixi.js v6 before, several days ago I migrated it to v7 (v7.1.2 actually).
Occurred with a TS error.

our code:
```tsx
this.app = new Application<HTMLCanvasElement>({
      width,
      height,
// ...
    });
```
in v6, I changed the background color with `this.app.renderer.background.color = backgroundColor;`

But in v7 IDE gives me an TS error:
`TS2339: Property 'background' does not exist on type 'IRenderer '.`

But I find that if I add `@ts-ignore`, it still can work. So there might be a type error in `PIXI.Application.IRenderer`.

### Expected Behavior

`PIXI.Application.IRenderer` (or interface `IRenderer`) has the `backgroud` property.

### Steps to Reproduce

```tsx
const app = new Application<HTMLCanvasElement>({
      width: 750,
      height: 1000,
      backgroundColor: 'black'
    });
app.renderer.background.color = 'red' // the background property will raise TS2339: Property 'background' does not exist on type 'IRenderer '.
```

### Environment

### Possible Solution

_No response_

### Additional Information

_No response_

bigtimebuddy:
Closes #9308

* Adds `background` property to IRenderer
* Adds `autoDensity` property to IRenderer

Screen Shot 2023-03-28 at 8 44 52 AM

### DIFF
--- a/packages/core/src/IRenderer.ts
+++ b/packages/core/src/IRenderer.ts
@@ -5,7 +5,13 @@ import type { IRendererPlugins } from './plugin/PluginSystem';
 import type { IGenerateTextureOptions } from './renderTexture/GenerateTextureSystem';
 import type { RenderTexture } from './renderTexture/RenderTexture';
 import type { SystemManager } from './system/SystemManager';
-import type { BackgroundSystemOptions, ContextSystemOptions, StartupSystemOptions, ViewSystemOptions } from './systems';
+import type {
+    BackgroundSystem,
+    BackgroundSystemOptions,
+    ContextSystemOptions,
+    StartupSystemOptions,
+    ViewSystemOptions,
+} from './systems';
 import type { ImageSource } from './textures/BaseTexture';
 
 /**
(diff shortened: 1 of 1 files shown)


SuperSodaSea (commented):
Should we also add other common systems to `IRenderer`, for example `ViewSystem` , `PluginSystem` and `StartupSystem`?

venaissance:
Ha, you're a little ahead of me, https://github.com/pixijs/pixijs/pull/9310

bigtimebuddy:
@SuperSodaSea I don't think Startup or Plugin systems are necessary. These are more internal and don't have any meaningful APIs outside init. ViewSystem has a bunch of convenience things on the renderers like width/height/resolution. All of these are already accounted for in IRenderer except for `autoDensity`, which I added.

bigtimebuddy:
@venaissance I'm sorry! I didn't see that, but thanks for making a contribution.

SuperSodaSea (approved):

PR #9313, opened 2023-03-30
Fix NPM blending

dev7355608:
##### Description of change

The color changes broke NPM.

- PM: https://pixiplayground.com/#/edit/MmH_RO-Edz_6mwGGLlSXT
- NPM: https://pixiplayground.com/#/edit/pd3pf7MP9lB7K15CR6Uj7

##### Pre-Merge Checklist

### DIFF
--- a/packages/color/src/Color.ts
+++ b/packages/color/src/Color.ts
@@ -358,6 +358,19 @@ export class Color
     }
 
     /**
+     * Convert to a hexadecimal number including alpha (signed 32-bit integer).
+     * @example
+     * import { Color } from 'pixi.js';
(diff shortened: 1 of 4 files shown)


dev7355608:
Maybe instead of adding `toNumberWithAlpha()` I should add the `applyToRGB` parameter to `toPremultiplied()` instead like in `premultiply()`. Let me know.

dev7355608:
> Maybe instead of adding `toNumberWithAlpha()` I should add the `applyToRGB` parameter to `toPremultiplied()` instead like in `premultiply()`. Let me know.

I made the change. It makes more sense this way, I think: less changes and the batch renderer test isn't failing because of the difference in rounding.

PR #9315, opened 2023-03-30
Only unpremultiply extracted pixels if the render target is premultiplied

dev7355608:
##### Pre-Merge Checklist


packages/canvas-extract/src/CanvasExtract.ts
```diff
@@ -23,7 +23,7 @@ export class CanvasExtract implements ISystem, IExtract
     };
 
     /** A reference to the current renderer */
-    public renderer: CanvasRenderer | null;
+    public renderer: CanvasRenderer;
```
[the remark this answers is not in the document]
bigtimebuddy: The strict-types are a work in progress. Most of the other systems are in core which hasn't been converted to strict types yet.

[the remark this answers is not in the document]
dev7355608: I see. I reverted the `| null` removal and handled the `Object is possibly 'null'` in `postrender`. Now there's no more complaints from `npm run types:strict`.

packages/core/src/textures/BaseTexture.ts
```diff
@@ -72,32 +72,32 @@ export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions>
      * @member {PIXI.ALPHA_MODES}
      * @default PIXI.ALPHA_MODES.UNPACK
      */
-    public alphaMode?: ALPHA_MODES;
+    public alphaMode: ALPHA_MODES;
```
bigtimebuddy, line 75: One small request, could you create another PR for the optional members? Seems like a separate issue.

dev7355608: It triggered a bunch of strict type warning, so I included it here. But I can split it off.

bigtimebuddy: Yeah, it's helpful when we look back or have to revert to have PRs be single ideas.

Issue #7836, opened 2021-10-01
Video .webm textures not clearing alpha, causing artifacts


Haxxer:
Hello,

After [Issue 3526](https://github.com/pixijs/pixijs/issues/3526) was closed, it appeared as if this was fixed for a time, but it has since returned. Perhaps an update to the way Chromium handles playing .webm files causes this, but in any case this issue still remains.

As per [this discussion](https://github.com/pixijs/pixijs/discussions/7175), various people have tried to raise the issue after it was marked as stale in [Issue 4089](https://github.com/pixijs/pixijs/issues/4089), and after it was missed in a related [issue on the Pixi GL Core Repo](https://github.com/pixijs/pixi-gl-core/issues/27).

From my experimentations, it appears that the current implementation simply does not clear the alpha after rendering frames, causing these rendering artifacts:

![Image](https://i.imgur.com/T2352WF.jpg)

- **Playground examples:** - white background added to demonstrate alpha difference
     - With fix: https://www.pixiplayground.com/#/edit/BSOldwbSTNfJ7e3WRSltM
     - Without fix:  https://www.pixiplayground.com/#/edit/wgmg-EQcyLdFX8EKLWpwV
     - **Credit to JB2A [https://jb2a.com/](https://jb2a.com/) for allowing me to use their asset as an example**
     - Note: I do **not** expect my makeshift patch to be thought of actual implementation, it is **only** for demonstration purposes.

- **`pixi.js` version**: 6.0.4

- **Tested Environments**
     - **Browsers & Versions**:
          - Google Chrome Version 94.0.4606.61 (Official Build) (64-bit)
          - Google Chrome Version 93.0.4577.82 (Official Build) (64-bit)
          - Edge Version 94.0.992.31 (Official build) (64-bit)
     - **OS & Versions**:
          - Windows 10 Pro (64-bit) : Version 20H2, Build 19042.1237
          - Windows 10 Home (64-bit) : Version 10.0.19042, Build 19042
          - Windows 10 Enterprise (64-bit) : Version 19041.1237

Haxxer:
My apologies, this is not due to Pixi at all, but rather the encoding of the .webm asset.

https://www.pixiplayground.com/#/edit/xiB6wlEPquEwU8FFoWQjb

As you can see in this playground example, I'm toggling between:

`baseTexture.alphaMode = PIXI.ALPHA_MODES.NO_PREMULTIPLIED_ALPHA;`

and

[...]

PR #9321, opened 2023-04-01
Add `utils.detectVideoAlphaMode`

dev7355608:
##### Description of change

As you might know, for no good reason, under some circumstances video alpha is already premultiplied by whatever is responsible for uploading the pixels to the GPU. My suspicion is that it is a bug in ANGLE. If I set the ANGLE backend is set to OpenGL, then the video is uploaded as expected: unless the alpha mode is `UNPACK`, the video is not premultiplied on upload. If the ANGLE backend is set to D3D11 in Chrome, then (on my machine) the video is already premultiplied: so, if the alpha mode is `UNPACK`, if is premultiplied twice (but there is another bug: if the video dimensions are 1x1 pixels, then no extra premultiplication is applied in Chrome; in Firefox with the D3D11 backend 1x1 behaves like any other dimension). To workaround this issue I wrote a function that checks whether a video is premultiplied without premultiply on upload, in which case the alpha mode `PMA` is returned instead of `UNPACK`.

This has been a widespread issue for years. But I can't find a bug report for this in the Chrome and ANGLE issue tracker. I think, it might go unnoticed a lot of the times: the developers/artist machine might render it correctly and the user doesn't know what the asset is supposed to look like. For a long time I thought it the assets might have been generated with premultiplied alpha by mistake; that's also what the author of #7836 thought, but the asset is actually encoded with straight alpha as it should be.

I'd like to combine this with #9314, so that the video loader uses `utils.detectVideoAlphaMode()` by default if `alphaMode` isn't specifically requested.

- Before: https://www.pixiplayground.com/#/edit/vhlX3qbIqAWaGvA_cxFZ1
- After: https://www.pixiplayground.com/#/edit/CBwvohPYOWqjoXyr4BvCS

1. View test in Chrome.
2. Go to `chrome://flags`.
3. Search for `Choose ANGLE graphics backend`.
4. Switch between `D3D11` and `OpenGL`.

##### Pre-Merge Checklist


GoodBoyDigital (approved):
nice!

jozefchutka:
Potentially related crbug https://issues.chromium.org/issues/355923583

PR #9324, opened 2023-04-02
Extract performance improvements

dev7355608:
##### Description of change

- 3d2d89c9d6ad1a6fe6bfb78fec5a2a4dc10d9d8f: `canvas.toBlob` instead of `toDataURL` blocks the main thread for a shorter duration.
- 4a2ff59114c9f015b9c96f7f1c04fd4e53c331c2: Improves performance (~50ms => ~42ms in my test). The `getImageData` is unnessecary.
- 4b2ce461ddd34465d27a138ed245c7eaf9d876ca: Improves performance even more (~42ms => ~34ms in my test).

##### Pre-Merge Checklist

### DIFF
--- a/packages/canvas-extract/src/CanvasExtract.ts
+++ b/packages/canvas-extract/src/CanvasExtract.ts
@@ -65,7 +65,7 @@ export class CanvasExtract implements ISystem, IExtract
 
         if (canvas.toBlob !== undefined)
         {
-            return await new Promise<string>((resolve, reject) =>
+            return new Promise<string>((resolve, reject) =>
             {
                 canvas.toBlob((blob) =>
                 {
@@ -85,7 +85,7 @@ export class CanvasExtract implements ISystem, IExtract
         {
             const blob = await canvas.convertToBlob({ type: format, quality });
 
-            return await new Promise<string>((resolve, reject) =>
+            return new Promise<string>((resolve, reject) =>
             {
                 const reader = new FileReader();
 
--- a/packages/canvas-extract/src/CanvasExtract.ts
+++ b/packages/canvas-extract/src/CanvasExtract.ts
@@ -63,6 +63,20 @@ export class CanvasExtract implements ISystem, IExtract
     {
         const canvas = this.canvas(target);
 
+        if (canvas.toBlob !== undefined)
+        {
+            return await new Promise<string>((resolve, reject) =>
+            {
+                canvas.toBlob((blob) =>
+                {
+                    const reader = new FileReader();
+
+                    reader.onload = () => resolve(reader.result as string);
+                    reader.onerror = reject;
+                    reader.readAsDataURL(blob);
+                }, format, quality);
+            });
+        }
         if (canvas.toDataURL !== undefined)
         {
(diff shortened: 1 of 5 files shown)


packages/canvas-extract/src/CanvasExtract.ts
```diff
@@ -63,6 +63,20 @@ export class CanvasExtract implements ISystem, IExtract
     {
         const canvas = this.canvas(target);
 
+        if (canvas.toBlob !== undefined)
+        {
+            return await new Promise<string>((resolve, reject) =>
```
bigtimebuddy, line 68: The await here is unnecessary and creates an additional promise. We should remove it.

```suggestion
            return new Promise<string>((resolve, reject) =>
```

packages/canvas-extract/src/CanvasExtract.ts
```diff
@@ -71,16 +85,18 @@ export class CanvasExtract implements ISystem, IExtract
         {
             const blob = await canvas.convertToBlob({ type: format, quality });
 
-            return await new Promise<string>((resolve) =>
+            return await new Promise<string>((resolve, reject) =>
```
bigtimebuddy, line 88: Same here. 
```suggestion
            return new Promise<string>((resolve, reject) =>
```

packages/extract/src/Extract.ts
```diff
@@ -84,6 +84,20 @@ export class Extract implements ISystem, IExtract
     {
         const canvas = this.canvas(target);
 
+        if (canvas.toBlob !== undefined)
+        {
+            return await new Promise<string>((resolve, reject) =>
```
bigtimebuddy, line 89:
```suggestion
            return new Promise<string>((resolve, reject) =>
```

packages/extract/src/Extract.ts
```diff
@@ -92,16 +106,18 @@ export class Extract implements ISystem, IExtract
         {
             const blob = await canvas.convertToBlob({ type: format, quality });
 
-            return await new Promise<string>((resolve) =>
+            return await new Promise<string>((resolve, reject) =>
```
bigtimebuddy, line 109:
```suggestion
            return new Promise<string>((resolve, reject) =>
```

bigtimebuddy (approved):
Nice one!

dev7355608:
I noticed my y-flipping was flawed: I forgot about the 4 bytes per pixel. There was no test that could have caught this error: so I added a bunch of tests that checks that looks that the extracted pixels/canvas.

I also replaced `postArrayDivide` with a much faster version (more that 3 times faster in my test).

dev7355608:
I added test for `CanvasExtract` too. Notice that `Extract.pixels` from a renderer does not y-flip, while `CanvasExtract.pixels` returns y-flipped results. Is this a bug? Wouldn't it make more sense if `Extract.pixels` y-flipped?

Another difference that bothers me a little bit is that `CanvasExtract.pixels` returns `Uint8ClampedArray` and `Extract.pixels` returns `Uint8Array`.

bigtimebuddy (commented):
Can you fix the strict type issues?

PR #9326, opened 2023-04-02
Add frame parameter to `IExtract.base64` and `IExtract.image`

dev7355608:
##### Description of change

I think `ICanvas.base64` and `ICanvas.image` should have the `frame` parameter that `ICanvas.canvas` and `ICanvas.pixels` have too.

It might be prettier if `frame` was the second parameter for all 4, but then we need to add some overloading to not break anything.

##### Pre-Merge Checklist


bigtimebuddy (approved):
LGTM. Thanks for this one, very helpful.

PR #9330, opened 2023-04-03
Fix `BaseTexture.fromBuffer` creating float texture with the wrong type

dev7355608:
##### Description of change

If `options` was given, the detected `type` never made it to the base texture.

`width` and `height` we don't need to add to the options. The `BufferResource` resizes the texture immediately.

If `type` is in the `options`, it replaces the auto detected `type`.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good catch

PR #9331, opened 2023-04-03
Prevent `BlobResource.load` from loading the data multiple times

dev7355608:
##### Description of change

Another call to `BlobResource.load` shouldn't load the data again.

Changed the return type of all Resource.load implementations to `Promise<this>` to make it consistent.

Added explicit `| null` to the types of some of members/parameters of `BlobResource`

##### Pre-Merge Checklist


GoodBoyDigital (approved):
looks good! more of a personal preference, but i generally find async functions easier to read! Not gonna push it though as its a case of potatoe potato! p :D

PR #9332, opened 2023-04-03
Change `BufferResource.data` type to include all possible typed arrays

dev7355608:
##### Description of change

> /** Source array Cannot be {@code ClampedUint8Array} because it cant be uploaded to WebGL */

The spec allows `Uint8ClampedArray`:
- https://registry.khronos.org/webgl/specs/latest/1.0/#5.14.8
- https://registry.khronos.org/webgl/specs/latest/2.0/#3.7.6

Must have been this bug in Chrome, which has been fixed a long time ago: https://bugs.chromium.org/p/chromium/issues/detail?id=961658

Added all typed array types that are allowed in `texImage2D` as listed here: https://registry.khronos.org/webgl/specs/latest/2.0/#3.7.6

Extended the automatic type detection in `fromBuffer`.

Removed `format: FORMATS.RGBA,` from the default `fromBuffer` options, because that's the default generally.

##### Pre-Merge Checklist


packages/compressed-textures/src/resources/BlobResource.ts
```diff
@@ -22,18 +22,19 @@ export abstract class BlobResource extends BufferResource
...
+     * @param source - the buffer/URL of the texture file
      * @param {PIXI.IBlobOptions} options
      * @param {boolean}[options.autoLoad] - whether to fetch the data immediately;
      *  you can fetch it later via {@link PIXI.BlobResource#load}
      * @param {boolean}[options.width] - the width in pixels.
      * @param {boolean}[options.height] - the height in pixels.
      */
-    constructor(source: string | Uint8Array | Uint32Array | Float32Array,
-        options: IBlobOptions = { width: 1, height: 1, autoLoad: true })
+    constructor(source: string | Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array
```
GoodBoyDigital, line 32: should we consider making a type for :`Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array
        | Float32Array`

Might make all the functions params a little less wordy!

dev7355608: Created the `BufferType` type. Not sure if it's the best name though. Maybe more specific like `BufferResourceDataType` is better?

dev7355608:
I noticed that removing the documentation of `BufferResource.data` (the comment about `Uint8ClampedArray`) made it vanish from the docs. I assume that `data` is supposed to be a public member.

GoodBoyDigital (approved):

PR #9333, opened 2023-04-03
Set `GLTexture.samplerType` based on the internal format

dev7355608:
##### Description of change

The `samplerType` shouldn't be `FLOAT` for integer textures.

##### Pre-Merge Checklist


bigtimebuddy (commented):
Small type suggestions, otherwise LGTM

PR #9336, opened 2023-04-04
Safer multisample detection

### LINKED ISSUE #9269: Bug: Wrong multisample quality detected
### Current Behavior

Property `renderer.multisample` returns 4 (`MSAA_QUALITY.MEDIUM`).

### Expected Behavior

[WebGL Report](https://webglreport.com/?v=2) reports the framebuffer parameter `Max Samples` to be 8. It is therefore expected that the property `renderer.multisample` should return 8 (`MSAA_QUALITY.HIGH`).

### Steps to Reproduce

On a device which has framebuffer parameter `Max Samples` equal to 8, simply read the `renderer.multisample` property.

### Environment

- **`pixi.js` version**: 7.1.4
- **Browser & Version**: Chrome 111.0.5563.64
- **OS & Version**: Ubuntu 22.04

### Possible Solution

The code for `MultisampleSystem` determines the multisample quality by accessing the `gl.SAMPLES` constant, both for WebGL1 and WebGL2. Same thing for `FramebufferSystem` on one occasion. However, WebGL2 has an additional constant used for multisampling — `gl.MAX_SAMPLES`. This is the one that [WebGL Report](https://webglreport.com/?v=2) uses. These two values are different, first one being 4, second being 8. I couldn't find what these two constants represent _exactly_ and what is the difference between them.

If it is decided that `gl.MAX_SAMPLES` is the one to be used, the solution would probably be to read this value instead of `gl.SAMPLES` for WebGL2.

dev7355608:
##### Description of change

Should fix #9269.

Number of samples is chosen based on the internal format. Samples above `MAX_SAMPLES` are filtered out as well as samples that are not valid depth/stencil samples.

Changed the type of `GLFramebuffer.multisample` to `number`, because technically it isn't necessarily a value in `MSAA_QUALITY`.

##### Pre-Merge Checklist


dev7355608:
Before I mark this ready-for-review: Should we change how `detectSamples` works? If the valid samples are `[8, 4]` then with `MSAA_QUALITY.LOW` there's no multisampling. Maybe we should select the minimum of the samples that are greater or equal to the number of requested samples. Then `LOW` would map to 4 in the the case of `[8, 4]`. That might make more sense considering that the samples parameter of `renderbufferStorageMultisample` is just a minimum: the implementation is fre choose a higher number of samples than requested.

dev7355608:
> Before I mark this ready-for-review: Should we change how `detectSamples` works? If the valid samples are `[8, 4]` then with `MSAA_QUALITY.LOW` there's no multisampling. Maybe we should select the minimum of the samples that are greater or equal to the number of requested samples. Then `LOW` would map to 4 in the the case of `[8, 4]`. That might make more sense considering that the samples parameter of `renderbufferStorageMultisample` is just a minimum: the implementation is free choose a higher number of samples than requested.

I'll leave this for another PR.

dev7355608:
I put the `Framebuffer.multisample` changes in a separate PR (#9390) wondering why I didn't in the first place, but now I recall: without the changes here #9390 does of course nothing, because the samples are detected and set in `initFramebuffer`. With the changes here the samples are detected in `updateFramebuffer`, and so it makes sense for `Framebuffer.multisample` to increase `dirtyId`.

packages/core/src/framebuffer/FramebufferSystem.ts
```diff
@@ -98,8 +98,32 @@ export class FramebufferSystem implements ISystem
...
         {
             // WebGL2
-            // cache possible MSAA samples
-            this.msaaSamples = gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES);
+            this.msaaSamples = {};
+
+            // Don't use more than `MAX_SAMPLES`: some devices do not allow to create renderbuffers
+            // with more than `MAX_SAMPLES` despite the fact that it is a valid number of samples
+            // according to `getInternalformatParameter`.
+            const maxSamples = Math.min(gl.getParameter(gl.MAX_SAMPLES) as number, MSAA_QUALITY.HIGH);
```
GoodBoyDigital, line 106: shall we move this to a separate function? FrameBufferSystem is getting pretty beefy!

GoodBoyDigital (approved):
sorry it took so long to get round to this! all makes sense! one tiny comment. Could we move the logic out to a `getMaxSamples()` function? - but other than that good to go!

Zyie:
Hey,

PR #9337, opened 2023-04-04
Default to the renderer's resolution/multisample in `generateTexture`

dev7355608:
##### Description of change

`Extract` uses the renderer's resolution and multisample. I think `extract.pixels(generateTexture(object))` and `extract.pixels(object)` should give the same result.

`IGenerateTextureOptions.resolution` already made the claim that it defaults to the renderer's resolution, but that wasn't the case (`PIXI.settings.RESOLUTION` was the default):
>   The resolution / device pixel ratio of the texture being generated. Optional defaults to Renderer resolution.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/renderTexture/GenerateTextureSystem.ts
+++ b/packages/core/src/renderTexture/GenerateTextureSystem.ts
@@ -73,6 +73,8 @@ export class GenerateTextureSystem implements ISystem
             {
                 width: region.width,
                 height: region.height,
+                resolution: this.renderer.resolution,
+                multisample: this.renderer.multisample,
                 ...textureOptions,
             });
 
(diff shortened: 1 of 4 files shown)


packages/core/src/renderTexture/GenerateTextureSystem.ts
```diff
@@ -2,27 +2,26 @@ import { extensions, ExtensionType } from '@pixi/extensions';
...
+import type { IBaseTextureOptions } from '../textures/BaseTexture';
 
 const tempTransform = new Transform();
 
 // TODO could this just be part of extract?
-export interface IGenerateTextureOptions
+export interface IGenerateTextureOptions extends IBaseTextureOptions
 {
-    /** The scale mode of the texture. Optional, defaults to `PIXI.BaseTexture.defaultOptions.scaleMode`. */
-    scaleMode?: SCALE_MODES;
```
bigtimebuddy, line 17: Why did you remove `scaleMode` from these options?

GoodBoyDigital (approved):

packages/core/src/renderTexture/GenerateTextureSystem.ts
dev7355608: I made `IGenerateTextureOptions` extend `IBaseTextureOptions`.

bigtimebuddy: Oops missed that!. Thanks for clarifying

PR #9338, opened 2023-04-04
Add public `Framebuffer.depth` and `Framebuffer.stencil` properties

dev7355608:
##### Description of change

I thought it might be useful if these properties are publicly accessible and can be changed with the framebuffer automatically updating.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/framebuffer/Framebuffer.ts
+++ b/packages/core/src/framebuffer/Framebuffer.ts
@@ -14,10 +14,16 @@ import type { GLFramebuffer } from './GLFramebuffer';
  */
 export class Framebuffer
 {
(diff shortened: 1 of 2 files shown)


GoodBoyDigital:
Hey @dev7355608 ! This is certainly a technically correct PR, but i can't think of a use case where this would be useful :D
We would prefer users to make separate FB's rather than changing these on the fly I think?

dev7355608:
Yeah, probably unlikely that these properties are changed, but I thought why not: it can't really hurt to have it.

talented634:
Nice

talented634:
Nice

Zyie:
Hey,

PR #9345, opened 2023-04-04
Unattach textures from framebuffer that are no longer used and …

dev7355608:
… don't attach more than `MAX_COLOR_ATTACHMENTS` color textures.

If a color texture or the depth texture is removed, then we need to make sure their attachments are cleared, if they are not overridden by other attachments.

##### Pre-Merge Checklist


bigtimebuddy:
Have you got a clear use-case for this? Are you modifying framebuffers on the fly?

dev7355608:
No, but I thought it should work. It's possible to add color textures after the first bind, but removing attachments isn't possible unless the framebuffer is disposed. Either both adding and removing should be possible after the first bind or neither, right?

Zyie:
Hey,

Issue #9291, opened 2023-03-24
Bug: Wrong hit area in shadow DOM (7.2.2 too)


naramdash:
### Current Behavior

Event occurs in weird position.

https://user-images.githubusercontent.com/24370799/227459296-bedddf41-7085-404f-9aa7-0425b4ae00a9.mp4

### Expected Behavior

Hit area should be properly calculated when pixi is embedded in custom elements with shadow DOM enabled.

### Steps to Reproduce

Append Pixi.view inside shadowRoot.

and

Give the body a margin, but not the custom element.

### Environment

- **`pixi.js` version**: *7.2.2*
- **`@pixi/events` version**: *7.2.2*
- **Browser & Version**: *Chrome 111*
- **OS & Version**: *Windows 11*

### Possible Solution

Maybe https://github.com/pixijs/pixijs/issues/8281/ was same issue of version 6.

Zyie:
Hey!

The code that fixed this last time is still present so it could be a new issue. However we do currently have a few bugs with events in 7.2.2 that will be fixed in the next release

Could you provide a minimal reproduction of the issue. Jsfiddle, codepen, codesandbox etc to give us something to validate the bug and verify it's been fixed

naramdash:
[...]

PR #9354, opened 2023-04-07
check `domElement` is connected to (shadow)DOM when `mapPositionToPoint`

naramdash:
related issue #9291 

##### Description of change

- check `domElement` is connected to (shadow)DOM when `mapPositionToPoint` with polyfill.

The previous code judged a situation(parentElement not exists) in which the client browser was IE11 through this result.

But, a parentElement of very first child of shadowRoot is null.

So, Using isConnected api to verify that it is included in the document regardless of dom or shadow DOM.

For IE11, polyfill code is added.

##### Pre-Merge Checklist


naramdash:
All tests that check function calls fail.

However, this is because the canvas used in the test code is used without a parent element,
So it seems that they all use the rect value corresponding to the IE11 fix.

And during testing, it seems that all calls to the dom element's getBoundingClientRect method fail.

naramdash:
I changed codes related with setting styles of shadowRoot host to avoid type error that TS 4.2 doesn't support `ShadowRoot.adoptedStylesheets`.

then, i ran 
- npm run lint
- npm run test
- npm run build && npm run build:types
- npm run dist

Zyie:
Hey @naramdash 

Pixi doesn't support IE11 anymore, so you can update this PR to remove any IE11 workarounds

naramdash:
@Zyie 

Ok, I removed polyfill codes. and become shorter and performant :)

Zyie (approved):

bigtimebuddy:
Thank you @naramdash :tada:

PR #9357, opened 2023-04-07
Throw an error if the width or height of a framebuffer is 0

dev7355608:
##### Description of change

The `width` and `height` in the `Framebuffer` constructor are not optional, but 0 was mapped to 100, which isn't documented at all. I think it's better if we throw an error.

##### Pre-Merge Checklist


GoodBoyDigital (approved):
👍

bigtimebuddy:
@dev7355608 could you resolve conflicts here?

PR #9359, opened 2023-04-09
Fix: `utils.detectVideoAlphaMode` never resolves if webm isn't supported

dev7355608:
##### Pre-Merge Checklist


bigtimebuddy (commented):
Is there a reason we can't just use https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType here?

packages/utils/src/browser/detectVideoAlphaMode.ts
```diff
@@ -27,64 +27,58 @@ export async function detectVideoAlphaMode(): Promise<ALPHA_MODES>
...
+                wait();
+            });
+        }
+
+        let video: HTMLVideoElement | null = null;
+
+        for (const src of [
+            // WebM VP9
+            // eslint-disable-next-line max-len
+            'data:video/webm;base64,GkXfo59ChoEBQveBAULygQRC84EIQoKEd2VibUKHgQJChYECGFOAZwEAAAAAAAIOEU2bdLpNu4tTq4QVSalmU6yBoU27i1OrhBZUrmtTrIHWTbuMU6uEElTDZ1OsggEnTbuMU6uEHFO7a1OsggH47AEAAAAAAABZAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAVSalmsCrXsYMPQkBNgIxMYXZmNjAuNC4xMDBXQYxMYXZmNjAuNC4xMDBEiYhARAAAAAAAABZUrmvMrgEAAAAAAABD14EBc8WIUzy3iLLKnmKcgQAitZyDdW5kiIEAhoVWX1ZQOYOBASPjg4QCYloA4JSwgQK6gQKagQJTwIEBVbCEVbmBARJUw2dAf3Nzn2PAgGfImUWjh0VOQ09ERVJEh4xMYXZmNjAuNC4xMDBzc9pjwItjxYhTPLeIssqeYmfIpEWjh0VOQ09ERVJEh5dMYXZjNjAuNi4xMDAgbGlidnB4LXZwOWfIokWjiERVUkFUSU9ORIeUMDA6MDA6MDAuMDQwMDAwMDAwAAAfQ7Z1x+eBAKDCoaCBAAAAgkmDQgAAEAAWADgkHBhKAAAgIAARv///iv4AAHWhnaab7oEBpZaCSYNCAAAQABYAOCQcGEoAACAgAEhAHFO7a5G7j7OBALeK94EB8YIBrPCBAw==',
```
bigtimebuddy, line 64: So we could use `canPlayType` for video/webm and only check for VP9

dev7355608: Maybe the VP8 case is unnecessary. Probably extremly unlikely that VP9 isn't supported but VP8 is.

eXponenta:
> Is there a reason we can't just use https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType here?

Browser can answer 'maybe' or '' for random reason, some browser answer that not support h265 but easy play it. 

Very strange api =)

SuperSodaSea (approved):

PR #9362, opened 2023-04-10
Fix: `LineStyle.reset()` does not reset `cap`, `join`, and `miterLimit`

dev7355608:
##### Pre-Merge Checklist


bigtimebuddy (approved):
Good catch!

PR #9363, opened 2023-04-10
Fix: Extract functions no longer throw an error if the frame is empty

dev7355608:
The alternative is that we don't allow empty frames and throw an error ourselves if `frame` is empty that is perhaps more more helpful can the error messages generated by `getImageData` etc.

##### Pre-Merge Checklist


packages/canvas-extract/src/CanvasExtract.ts
```diff
@@ -234,8 +234,8 @@ export class CanvasExtract implements ISystem, IExtract
 
         const x = Math.round(frame.x * resolution);
         const y = Math.round(frame.y * resolution);
-        const width = Math.round(frame.width * resolution);
-        const height = Math.round(frame.height * resolution);
+        const width = Math.max(Math.round(frame.width * resolution), 1);
```
GoodBoyDigital, line 237: i like this! perhaps we set to 1, but also log a warning?

bigtimebuddy:
What would be better is if we can assert the input before doing anything and throw a specific user-friendly error instead of defaulting to 1x1.

bigtimebuddy:
I think ceil would make sense, yes.

dev7355608:
Reverted the throwing of an error: If `generateTexture` doesn't throw if empty, then it should be acceptable for the extraction methods to behave the same way.

Reverted the `ceil` rounding: If the extraction method ceil the frame, then `generateTexture` should probably ceil the width and height of the region too. Ceiling probably makes more sense in both cases, but I'll leave that for another PR.

GoodBoyDigital (approved):

bigtimebuddy:
This has deviated a little from the original title of the PR. How would you rename this now?

dev7355608:
Technically one of the others PR merged before this one already prevented the errors, but didn't mention it. Not sure how to describe the change in a single sentence:
> `CanvasExtract`/`Extract` functions no longer throw an error if the `frame` is empty. If the width/height of the `frame` is less than 1 pixel, it is rounded up to 1 pixel just like the width/height of `region` parameter of `generateTexture` is.

bigtimebuddy:
@dev7355608 could you resolve conflicts so I could merge this?

PR #9373, opened 2023-04-14
Fallback if `createImageBitmap(Blob)` is not supported

dev7355608:
##### Description of change

According to [this](https://bugs.webkit.org/show_bug.cgi?id=182424) `createImageBitmap(Blob)` is not supported in Safari.

- Added `canLoadImageBitmap()`, which tests if `loadImageBitmap()` is supported.
- ~~Added the `testImageOrVideoFormat(dataURL)` helper to test if an image or video format it supported.~~

##### Pre-Merge Checklist


packages/assets/src/detections/parsers/detectAvif.ts
```diff
@@ -9,13 +10,10 @@ export const detectAvif: FormatDetectionParser = {
...
     test: async (): Promise<boolean> =>
     {
-        if (!globalThis.createImageBitmap) return false;
-
         // eslint-disable-next-line max-len
         const avifData = 'data:image/avif;base64,AAAAIGZ0eXBhdmlmAAAAAGF2aWZtaWYxbWlhZk1BMUIAAADybWV0YQAAAAAAAAAoaGRscgAAAAAAAAAAcGljdAAAAAAAAAAAAAAAAGxpYmF2aWYAAAAADnBpdG0AAAAAAAEAAAAeaWxvYwAAAABEAAABAAEAAAABAAABGgAAAB0AAAAoaWluZgAAAAAAAQAAABppbmZlAgAAAAABAABhdjAxQ29sb3IAAAAAamlwcnAAAABLaXBjbwAAABRpc3BlAAAAAAAAAAIAAAACAAAAEHBpeGkAAAAAAwgICAAAAAxhdjFDgQ0MAAAAABNjb2xybmNseAACAAIAAYAAAAAXaXBtYQAAAAAAAAABAAEEAQKDBAAAACVtZGF0EgAKCBgANogQEAwgMg8f8D///8WfhwB8+ErK42A=';
-        const blob = await settings.ADAPTER.fetch(avifData).then((r) => r.blob());
 
-        return createImageBitmap(blob).then(() => true, () => false);
+        return testImageFormat(avifData);
```
bigtimebuddy, line 16: Can't we simply do `new Image` here instead of blob/createImageBitmap? This PR seems like a lot of work to see if AVIF works. We are not trying to test if blob + createImageBitmap is supported only if the format is supported by the browser.

dev7355608: We can't use `Image`, because it isn't available in workers.

Zyie:
Hey,

PR #9374, opened 2023-04-14
Use `requestVideoFrameCallback` to auto update `VideoResource` if possible

dev7355608:
[`requestVideoFrameCallback`](https://wicg.github.io/video-rvfc/) isn't available everywhere unfortunately: https://caniuse.com/?search=requestVideoFrameCallback. If it is, we should use it to avoid unnecessary uploads each frame.
##### Pre-Merge Checklist


GoodBoyDigital (approved):
👍

PR #9390, opened 2023-04-19
Increase `dirtyId` when `Framebuffer.multisample` is changed

dev7355608:
##### Description of change

`Framebuffer.multisample` is now a getter/setter that updates `dirtyId` so that the underlying framebuffer is updated and so the sample count can be changed after the framebuffer was bound for the first time.

##### Pre-Merge Checklist


Zyie:
Hey,

PR #9399, opened 2023-04-24
Move worker codes to seperate files

SuperSodaSea:
#### Description of change

Move worker codes to seperate `*.worker.ts` files, instead of hard-coding them as strings. This can make the code more readable, and visible to TypeScript and ESLint checks.

- New `tsconfig.worker.json` and npm script `types:worker` for workers type checking
- `rollup-plugin-web-worker-loader` for loading workers in Rollup
- `test/jest-webworker-loader.js` for loading workers in Jest

#### TODO

- [x] Replace `rollup-plugin-web-worker-loader` with our own plugin. This plugin generate extra code for Node.js environment that we don't need.
- [x] Move Rollup & Jest plugin code to another new repo? -> <https://github.com/pixijs/webworker-plugins>

#### Pre-Merge Checklist


SuperSodaSea:
@bigtimebuddy Can you please create a new empty <https://github.com/pixijs/webworker-plugins> repository for me? I'm planning to create `@pixi/rollup-plugin-webworker` and `@pixi/jest-plugin-webworker` packages to handle `.worker.ts`.

bigtimebuddy:
Done @SuperSodaSea

SuperSodaSea:
@bigtimebuddy Thanks :)

GoodBoyDigital:
hey @SuperSodaSea ! Just got pointed to this PR as ran in to a similar issue that you are solving here for v8.
@Zyie created a Webworker rollup plugin and works like a champ when building. But does not work in Jest. How far did you get on that part? Would be awesome to get some tests up for webworker based code in jest. 

Cheers!

SuperSodaSea:
@GoodBoyDigital I'm writing a Jest Web Worker plugin (transform) that matches my Rollup Web Worker plugin. The Jest plugin works fine now, but only supports `.worker.ts` without `import`. Since our workers won't be too complex now, I think this won't be a big problem. I'll upload them to <https://github.com/pixijs/webworker-plugins> when they are ready in a few days.

SuperSodaSea:
@bigtimebuddy Seems that there are some problems when releasing the new package :thinking:, can you help me with the problem?

<https://github.com/pixijs/webworker-plugins/actions/runs/6577787015/job/17870168354>

bigtimebuddy:
@SuperSodaSea I added a PR [here](https://github.com/pixijs/webworker-plugins/pull/1). Scoped packages are private by default, so you need to set `access` in the package.json publishConfig. I was able to manually publish 0.1.0.

bigtimebuddy:
@SuperSodaSea is this good to go? Also, could you just verify that sourcemaps still work correctly?

bigtimebuddy:
We upgraded to typescript@5, we might need some additional typing-help for this.

SuperSodaSea:
@bigtimebuddy I'll try to fix it this week.

SuperSodaSea:
@bigtimebuddy It seems we still have to publish `@pixi/webworker-plugins@0.2.0` manually?

bigtimebuddy:
@SuperSodaSea published the latest 0.2.0.

Could you tell me how you did the release? The package-lock.json file didn't have the same version as the package.json which I think was tripping up the publish.

SuperSodaSea:
@bigtimebuddy Oops, I changed the version field in package.json manually.

global.d.ts
```diff
@@ -1,13 +1,27 @@
 // eslint-disable-next-line spaced-comment
 /// <reference types="jest-extended" />
 
-declare module '*.frag' {
+declare module '*.worker.ts'
```
bigtimebuddy, line 4: Noticed you changed the pattern on the import, should this be this instead?
```ts
declare module 'worker:*.ts'
```

packages/core/package.json
```diff
@@ -46,7 +46,6 @@
     "@pixi/runner": "file:../runner",
     "@pixi/settings": "file:../settings",
     "@pixi/ticker": "file:../ticker",
-    "@pixi/utils": "file:../utils",
-    "@types/offscreencanvas": "^2019.6.4"
```
bigtimebuddy, line 50: Why was this dependency removed?

SuperSodaSea: TypeScript 5 already has a built-in OffscreenCanvas interface, so I think this dependency is no longer needed? It is causing some type error when checking the workers.

bigtimebuddy: This should be another PR

bigtimebuddy: Okay

SuperSodaSea:
> Also, could you just verify that sourcemaps still work correctly?

@bigtimebuddy The workers do not have sourcemaps currently, but as our workers are not very large at the moment, this is not a very serious problem. If you think the sourcemap is necessary, I will try to work it out when I have time.

bigtimebuddy:
The workers are inlined as a string, right? I am more worried about it messing up source-maps for the main file then debugging the worker correctly.

SuperSodaSea:
@bigtimebuddy The source map of the main file works pretty well, only the source for the workers is missing. Checked in Chrome & Edge & Firefox.

SuperSodaSea:
> The workers are inlined as a string, right?

Yes, the generated code looks like this:

```js
const WORKER_CODE = `/* ACTUAL WORKER CODE */`;
let WORKER_URL = null;
class WorkerInstance extends Worker {
  constructor() {
    WORKER_URL || (WORKER_URL = URL.createObjectURL(new Blob([WORKER_CODE], { type: "application/javascript" }))), super(WORKER_URL);
  }
}
WorkerInstance.revokeObjectURL = function() {
  WORKER_URL && (URL.revokeObjectURL(WORKER_URL), WORKER_URL = null);
};
export {
  WorkerInstance as default
};
```

bigtimebuddy (approved):
Nice work, thank you @SuperSodaSea

PR #9404, opened 2023-04-26
Add ownsImageBitmap for ImageBitmapResource

### LINKED ISSUE #9403: Bug: Creating a Sprite from an ImageBitmap used to create a previously destroyed Sprite results in an empty Sprite
### Current Behavior

If I create a Sprite from an ImageBitmap, destroy that Sprite with the options `{ children: true, texture: true, baseTexture: true }`, and then try to create a new Sprite from the same ImageBitmap, the new Sprite will be empty when rendered.

### Expected Behavior

I expect to be able to create a new sprite using an ImageBitmap that was used to create a previously destroyed sprite. Some of the ImageBitmaps I am working with are large, and I need to remove them from the graphics card when not in use to prevent context losses. 

### Steps to Reproduce

Here is a playground that replicates the issue: [https://www.pixiplayground.com/#/edit/cW1MglcpTHTH_ViEEj4YJ](https://www.pixiplayground.com/#/edit/cW1MglcpTHTH_ViEEj4YJ)

If that doesn't work, here is a repo that reproduces the issue in a React component: [pixi-bitmap-error-example](https://github.com/rj3d/pixi-bitmap-error-example). If you download the repo you can run `npm install` and then `npm run dev` to start a local development server.

All of the pixi stuff takes place in the file [PixiTest](https://github.com/rj3d/pixi-bitmap-error-example/blob/main/src/pages/PixiTest.tsx) in the following section:

```
const bunnyDel = PIXI.Sprite.from(bitmap);
bunnyDel.destroy({ children: true, texture: true, baseTexture: true })

const bunny = PIXI.Sprite.from(bitmap);

bunny.x = 0;
bunny.y = 0;
        
this.stage.removeChildren()
        
this.stage.addChild(bunny);
        
this.renderer.render(this.rootContainer)
```

If you comment out the two lines that create and destroy the bunnyDel sprite then the bunny sprite renders as expected. However, with the bunnyDel lines in place the bunny sprite renders as empty.

### Environment

- **`pixi.js` version**:7.2.4
- **Browser & Version**: Chrome 112
- **OS & Version**: macOS 12.6.5

SuperSodaSea:
#### Description of change

Add `ownsImageBitmap` for `ImageBitmapResource` and `IImageBitmapResourceOptions`. When it is `true`, the `ImageBitmapResource`'s underlying `ImageBitmap` will be disposed (`this.source.close()` is called); otherwise the `ImageBitmap` will not be disposed.

If the `ownsImageBitmap` option is not set by user:
- When `source` is `ImageBitmap`, `ownsImageBitmap` will be `false` by default.
- When `source` is `string` (a URL), `ownsImageBitmap` will be `true` by default.

`loadTextures` is also updated: when `loadTextures` is using `ImageBitmap`, the `resourceOptions.ownsImageBitmap` will be set to `true` when creating the `BaseTexture` if this option is not set by user.

Fixes #9403.
- Before (disposed): <https://www.pixiplayground.com/#/edit/cW1MglcpTHTH_ViEEj4YJ>
- After (not disposed): <https://www.pixiplayground.com/#/edit/qkDMDCYVMtO4wsVv-RCLp>

#### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/resources/ImageBitmapResource.ts
+++ b/packages/core/src/textures/resources/ImageBitmapResource.ts
@@ -61,7 +61,7 @@ export class ImageBitmapResource extends BaseImageResource
      * Whether the underlying ImageBitmap is owned by the ImageBitmapResource.
      * @see PIXI.IImageBitmapResourceOptions.ownsImageBitmap
      */
-    ownsImageBitmap: boolean;
+    private ownsImageBitmap: boolean;
 
     /**
      * Promise when loading.
--- a/packages/core/src/textures/resources/ImageBitmapResource.ts
+++ b/packages/core/src/textures/resources/ImageBitmapResource.ts
@@ -7,6 +7,10 @@ import type { Renderer } from '../../Renderer';
 import type { BaseTexture } from '../BaseTexture';
 import type { GLTexture } from '../GLTexture';
 
+/**
+ * Options for ImageBitmapResource.
+ * @memberof PIXI
+ */
 export interface IImageBitmapResourceOptions
 {
     /** Start loading process automatically when constructed. */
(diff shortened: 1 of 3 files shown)


packages/core/src/textures/resources/ImageBitmapResource.ts
```diff
@@ -42,35 +57,41 @@ export class ImageBitmapResource extends BaseImageResource
      */
     alphaMode: ALPHA_MODES | null;
 
+    /**
+     * Whether the underlying ImageBitmap is owned by the ImageBitmapResource.
+     * @see PIXI.IImageBitmapResourceOptions.ownsImageBitmap
+     */
+    ownsImageBitmap: boolean;
```
bigtimebuddy, line 64: Could you make this private?

bigtimebuddy (approved):
This looks good to me.

PR #9409, opened 2023-04-28
Don't GC textures that don't have a resource

dev7355608:
##### Description of change

If a texture doesn't have a resource, we shouldn't GC it, because we cannot reupload it. Only textures that are rendered to don't have a resource, but they don't have to be `BaseRenderTexture`s necessarily. For example, textures added to a framebuffer with `addColorTexture`/`addDepthTexture` aren't.

##### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/textures/TextureGCSystem.ts
+++ b/packages/core/src/textures/TextureGCSystem.ts
@@ -137,7 +137,7 @@ export class TextureGCSystem implements ISystem
             const texture = managedTextures[i];
 
             // Only supports non generated textures at the moment!
-            if (!(texture as any).framebuffer && this.count - texture.touched > this.maxIdle)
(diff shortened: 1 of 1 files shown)


SuperSodaSea (approved):

packages/core/src/textures/TextureGCSystem.ts
```diff
@@ -137,7 +137,7 @@ export class TextureGCSystem implements ISystem
             const texture = managedTextures[i];
 
             // Only supports non generated textures at the moment!
-            if (!(texture as any).framebuffer && this.count - texture.touched > this.maxIdle)
+            if (texture.resource && this.count - texture.touched > this.maxIdle)
```
bigtimebuddy, line 140: I'm confused the the title of the PR. Shouldn't it be `!texture.resource`

PR #9415, opened 2023-05-03
Add `BufferResource.unpackAlignment`

dev7355608:
##### Description of change

Without control over the unpack alignment, we cannot upload a buffer resource that has a pixel row size in bytes isn't a multiple of 4, which is the default unpack alignment.

##### Pre-Merge Checklist

### DIFF
--- a/packages/compressed-textures/src/resources/CompressedTextureResource.ts
+++ b/packages/compressed-textures/src/resources/CompressedTextureResource.ts
@@ -149,6 +149,8 @@ export class CompressedTextureResource extends BlobResource
             return false;
         }
 
+        gl.pixelStorei(gl.UNPACK_ALIGNMENT, 4);
+
         for (let i = 0, j = this.levels; i < j; i++)
         {
             const { levelID, levelWidth, levelHeight, levelBuffer } = this._levelBuffers[i];
--- a/packages/core/src/textures/BaseTexture.ts
+++ b/packages/core/src/textures/BaseTexture.ts
@@ -9,7 +9,7 @@ import type { MSAA_QUALITY } from '@pixi/constants';
 import type { ICanvas } from '@pixi/settings';
 import type { GLTexture } from './GLTexture';
 import type { IAutoDetectOptions } from './resources/autoDetectResource';
-import type { BufferType } from './resources/BufferResource';
+import type { BufferType, IBufferResourceOptions } from './resources/BufferResource';
 
 const defaultBufferOptions = {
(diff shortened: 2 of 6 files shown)


packages/core/src/textures/BaseTexture.ts
```diff
@@ -704,11 +704,11 @@ export class BaseTexture<R extends Resource = Resource, RO = IAutoDetectOptions>
      * @returns - The resulting new BaseTexture
      */
     static fromBuffer(buffer: BufferType, width: number, height: number,
-        options?: IBaseTextureOptions): BaseTexture<BufferResource>
+        options?: IBaseTextureOptions<IBufferResourceOptions>): BaseTexture<BufferResource>
     {
         buffer = buffer || new Float32Array(width * height * 4);
 
-        const resource = new BufferResource(buffer, { width, height });
+        const resource = new BufferResource(buffer, Object.assign({ width, height }, options?.resourceOptions));
```
SuperSodaSea, line 711:
```suggestion
        const resource = new BufferResource(buffer, { width, height, ...options?.resourceOptions });
```

Maybe it's more concise to write it this way?

packages/compressed-textures/src/resources/CompressedTextureResource.ts
```diff
@@ -149,6 +149,8 @@ export class CompressedTextureResource extends BlobResource
             return false;
         }
 
+        gl.pixelStorei(gl.UNPACK_ALIGNMENT, 4);
```
SuperSodaSea, line 152: Is this intentional, or should we use `this.unpackAlignment`?

dev7355608: I'm not very familiar with this part code. But the level buffers always have an alignment of 4 (`_createLevelBuffers` is called with `blockWidth/blockHeight` 4). We could change it to `gl.pixelStorei(gl.UNPACK_ALIGNMENT, this.unpackAlignment);`, because the default is 4, but it would probably make more sense if `CompressedLevelBuffer` had an  `levelUnpackAlignment` property as each buffer could (at least in theory) have different alignments. Hardcoding it to 4 definitely doesn't break anything. I wasn't 100% sure what the best solution here would be. I'm open for suggestion.

SuperSodaSea: Yeah I think it is fine to set unpack align to 4 for now. We can add more options later if needed.

SuperSodaSea (approved):

PR #9418, opened 2023-05-05
Replace `url` dependency with modern URL API

MuTsunTsai:
##### Description of change
This is related to #9413, where I proposed to replace `url` dependency with native API.

##### Pre-Merge Checklist

### DIFF
--- a/packages/utils/src/network/determineCrossOrigin.ts
+++ b/packages/utils/src/network/determineCrossOrigin.ts
@@ -21,10 +21,8 @@ export function determineCrossOrigin(url: string, loc: Location = globalThis.loc
 
     const parsedUrl = new URL(url, document.baseURI);
 
-    const samePort = (!parsedUrl.port && loc.port === '') || (parsedUrl.port === loc.port);
-
     // if cross origin
-    if (parsedUrl.hostname !== loc.hostname || !samePort || parsedUrl.protocol !== loc.protocol)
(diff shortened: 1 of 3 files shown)


packages/utils/src/network/determineCrossOrigin.ts
```diff
@@ -23,16 +19,7 @@ export function determineCrossOrigin(url: string, loc: Location = globalThis.loc
...
-    }
-
-    // let the browser determine the full href for the url of this resource and then
-    // parse with the node url lib, we can't use the properties of the anchor element
-    // because they don't work in IE9 :(
-    tempAnchor.href = url;
-    const parsedUrl = _url.parse(tempAnchor.href);
+    const parsedUrl = new URL(url, document.baseURI);
 
     const samePort = (!parsedUrl.port && loc.port === '') || (parsedUrl.port === loc.port);
```
SuperSodaSea, line 24: The `port` of `URL` would be `''` instead of `null` (for package 'url') if port is absent in the URL, so we don't need special treatment for this anymore.

MuTsunTsai: Good observation. Done.

SuperSodaSea (approved):
LGTM!

bigtimebuddy:
Thank you @MuTsunTsai :tada:

PR #9440, opened 2023-05-23
Fix: Remove the asset from the asset cache and the loader promise cache when the texture is destroyed

dev7355608:
We do not want to remove the asset from the cache when the texture is disposed. A texture could be disposed without being destroyed.

We also need to remove the asset from the asset cache.

### DIFF
--- a/packages/assets/src/loader/parsers/textures/utils/createTexture.ts
+++ b/packages/assets/src/loader/parsers/textures/utils/createTexture.ts
@@ -9,7 +9,7 @@ export function createTexture(base: BaseTexture, loader: Loader, url: string)
     const texture = new Texture(base);
 
     // remove the promise from the loader and the url from the cache when the texture is destroyed
-    texture.baseTexture.once('destroy', () =>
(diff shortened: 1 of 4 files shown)


bigtimebuddy:
Would you mind implementing unit-tests for this change?

packages/assets/src/loader/parsers/textures/utils/createTexture.ts
```diff
@@ -8,9 +9,10 @@ export function createTexture(base: BaseTexture, loader: Loader, url: string)
     const texture = new Texture(base);
 
     // make sure to nuke the promise if a texture is destroyed..
-    texture.baseTexture.on('dispose', () =>
+    texture.baseTexture.on('destroy', () =>
```
bigtimebuddy, line 12:
```suggestion
    texture.baseTexture.once('destroy', () =>
```

dev7355608:
Should the make it `destroyed` instead of `destroy`? `DisplayObject` emits `destroyed` and there are no other objects that emit `destroy`.

PR #8806, opened 2022-10-31
VideoResource upload fix

animanathome:
##### Description of change
VideoResource bug fix. Ensures we upload video changes once `play` has finished. When calling `play` on the video element readyState changes from 4 to 1 once playback has ended even though all of the data is buffered. This change takes this into account and allows us to upload video changes such as `currentTime` once playback has ended. See https://codesandbox.io/s/stoic-grass-rkhxl3?file=/src/index.js for working example.

##### Pre-Merge Checklist

- [x] Tests and/or benchmarks are included
- [x] Documentation is changed or added
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #9441, opened 2023-05-23
Fix: Update `VideoResource` on `seeked`

dev7355608:
Reverted #8806: The fix doesn't work in Firefox. It spams a bunch of warnings and produces flickering black frames when the `currentTime` is changed. Open the example provided by #8806 (https://codesandbox.io/s/stoic-grass-rkhxl3?file=/src/index.js) in Firefox to observe this issue.
```
WebGL warning: tex(Sub)Image[23]D: Resource has no data (yet?). Uploading zeros.
WebGL warning: drawElementsInstanced: TEXTURE_2D at unit 0 is incomplete: The dimensions of `level_base` are not all positive. 
```

Instead I added a `seeked` event handler that updates the resource. Note though that the example provided by #8806 does not work after this change because the `currentTime` is updated so fast (every frame) that the `readyState` is 1 in `upload` even though in the previous `seeked` event the `readyState` was 4. I cannot observe any other events on the source in between. Maybe the reason is that the seeks don't finish within a frame and then next seek is requested and they start to overlap (?). If there's enough time in between the seeks it works as expected. Seeking every frame is probably something that isn't really supported by videos.

### DIFF
--- a/packages/core/src/textures/resources/BaseImageResource.ts
+++ b/packages/core/src/textures/resources/BaseImageResource.ts
@@ -85,7 +85,7 @@ export class BaseImageResource extends Resource
         }
         else if (typeof HTMLVideoElement !== 'undefined' && source instanceof HTMLVideoElement)
         {
-            if (source.readyState <= 1 && source.buffered.length === 0)
+            if (source.readyState <= 1)
             {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Nice solution

PR #9444, opened 2023-05-25
Fix: Mark resources of loaded textures as internal so that they are destroyed when unloaded

dev7355608:
Resources of loaded SVGs and videos are not destroyed currently because the resource is created outside of `BaseTexture`. The same is true for the `BufferResource` created in `BaseTexture.fromBuffer`.


bigtimebuddy (approved):
Good catch

PR #9448, opened 2023-05-28
Add Spritesheet type inference

pedr0fontoura:
##### Description of change
Add type inference for the Spritesheet atlas data. It makes using Spritesheets much easier for those using plain objects as an atlas.

- `.toString()` in line 330 is explained at the end of [this](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html) section of the TypeScript docs.
- `S` for the type parameter following existing code style https://github.com/pixijs/pixijs/blob/356abaaad852b248f0aa3f6873f5d7d2a56e3a50/packages/core/src/textures/BaseTexture.ts#L49
##### Pre-Merge Checklist

### DIFF
--- a/packages/spritesheet/src/Spritesheet.ts
+++ b/packages/spritesheet/src/Spritesheet.ts
@@ -114,13 +114,13 @@ export interface ISpritesheetData
  * supported by TexturePacker.
  * @memberof PIXI
  */
-export class Spritesheet
+export class Spritesheet<S extends ISpritesheetData = ISpritesheetData>
 {
     /** The maximum number of Textures to build per process. */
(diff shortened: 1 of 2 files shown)


pedr0fontoura:
Part of the initial CI error slipped right through the `lint`, `test`, and `build` scripts because of the strict check when building for distribution, so I would suggest that running the `dist` script is also mentioned in [CONTRIBUTING.md](https://github.com/pixijs/pixijs/blob/dev/.github/CONTRIBUTING.md) or [PULL_REQUEST_TEMPLATE.md](https://github.com/pixijs/pixijs/blob/dev/.github/PULL_REQUEST_TEMPLATE.md)

Issue #9414, opened 2023-05-03
Bug: CRIT! Safari 16.4+ Array uniforms upload


eXponenta:
### Current Behavior

For some strange APP (NDA) there are unstable behaviour between browsers while syncing a custom uniforms.

While debugging we find this code: 
https://github.com/pixijs/pixijs/blob/5f6b33536dfd4a610e97d4bf94c2b86663e36eb7/packages/core/src/shader/utils/generateUniformsSync.ts#L232

Follow implementation we seen that uniforms is random object, as sample is said:

```
const myCoolData = new UniformBufferGroup({
     uCoolMatrix: new Matrix(),
     uFloatyMcFloatFace: 23,
     uArray: [100, 100],
 }}
 ```
 https://github.com/pixijs/pixijs/blob/5f6b33536dfd4a610e97d4bf94c2b86663e36eb7/packages/core/src/shader/UniformGroup.ts#L59
 
 BUT, Safari 16.4+ has `group` field in array prototype:

 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/group 

and codegenerator begin make emitting uploader block:
<img width="1440" alt="Снимок экрана 2023-05-03 в 13 00 48" src="https://user-images.githubusercontent.com/12572222/235888757-5cf8c38d-c208-4888-acb8-a1a4b07f090e.png">

after that this code is down, because t is array:
<img width="1440" alt="Снимок экрана 2023-05-03 в 13 08 38" src="https://user-images.githubusercontent.com/12572222/235888815-2f07dd2e-4b59-468e-95f9-de1d2b2503f4.png">

### Expected Behavior

No runtime crash in every browsers and platforms

### Steps to Reproduce

...coming soon

### Environment

- **`pixi.js` version**: every version that  has codegeneration for uniform upload 
- **Browser & Version**: Safari 16.4+
- **OS & Version**: MacOS, iOS 16.4+

### Possible Solution

Strict field check
```
if (group.uniforms[i] && 'group' in group.uniforms[i]) {
    ///
}
```

### Additional Information

Call @ErikSom for other information

eXponenta:
@GoodBoyDigital 
@ivanpopelyshev 
@bigtimebuddy

CodingYiyi:
safari 16.4 is nightmare

miltoncandelero:
> ```
> delete Array.prototype.group;
> ```
> 
[...]

PR #9454, opened 2023-06-02
fix: force strict check for group variable in UniformGroup

miltoncandelero:
##### Description of change
This aims to solve #9414 where safari invented the group method in arrays and that made it so arrays now were detected as UniformGroups.

I achieved this by trying to find every place that a UniformGroup is used and where we check for the `group` variable and forced an strict check with true. By doing this, truthy values are not enough, we must have that value set to true.

I would like for somebody with more knowledge on UniformGroup code to see if I missed anything and if my solution makes sense.

### DIFF
--- a/packages/core/src/shader/Shader.ts
+++ b/packages/core/src/shader/Shader.ts
@@ -63,7 +63,7 @@ export class Shader
         {
             const uniform = group.uniforms[i];
 
-            if (uniform.group)
+            if (uniform.group === true)
             {
                 if (this.checkUniformExists(name, uniform))
                 {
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Nice fix. Could you add an inline comment about this? `=== true` is not typical stylistically, so would help someone not stomp on this change in the future.

bigtimebuddy (approved):
LGTM, would like to @GoodBoyDigital to weigh-in as he understands the uniform stuff most deeply.

GoodBoyDigital (commented):
this is cool, i wander would it be less magical and error prone f we just renamed `group` property to `isUniformGroup` ?

Zyie:
> this is cool, i wander would it be less magical and error prone if we just renamed `group` property to `isUniformGroup` ?

We have renamed this in v8, so this can be merged as is and avoid any breaking changes / deprecations

PR #9463, opened 2023-06-08
Chore/destroy-graphics

GoodBoyDigital:
Added a destroy flow for graphics.
Will apply the same flow to the rest of the Renderables if we are cool with this flow!

copilot:all


src/rendering/scene/Container.ts
```diff
@@ -734,6 +737,8 @@ export class Container<T extends View = View> extends EventEmitter<ContainerEven
         this._skew = null;
 
         this.emit('destroyed');
+
+        this.onDestroyed();
```
Zyie, line 741: We call `emit('destroyed')` already
Why do we need the callback?

GoodBoyDigital: i don't really know! I think i was trying to simplify - but just made it more coplex :D will adjust

PR #9465, opened 2023-06-08
Core/destroy-mesh

GoodBoyDigital:
- hook into events for `destroyed`
- renamed sprite to renderable in sprite pipe (for consistency with the other pipes)
- make `Container` call `view.destroy` rather than overriding in the renderable class
- make Renderable extend `EventEmitter`
- fixed up the 'Sprite' destroy cycle
- fixed up the 'Mesh' destroy cycle
- fixed up the 'graphics` destroy cycle

### DIFF
--- a/src/rendering/graphics/shared/GraphicsPipe.ts
+++ b/src/rendering/graphics/shared/GraphicsPipe.ts
@@ -7,11 +7,10 @@ import type { ExtensionMetadata } from '../../../extensions/Extensions';
 import type { PoolItem } from '../../../utils/pool/Pool';
 import type { Instruction } from '../../renderers/shared/instructions/Instruction';
 import type { InstructionSet } from '../../renderers/shared/instructions/InstructionSet';
-import type { RenderPipe } from '../../renderers/shared/instructions/RenderPipe';
+import type { BatchPipe, RenderPipe } from '../../renderers/shared/instructions/RenderPipe';
 import type { Renderable } from '../../renderers/shared/Renderable';
 import type { Shader } from '../../renderers/shared/shader/Shader';
-import type { Renderer } from '../../renderers/types';
-import type { GpuGraphicsContext } from './GraphicsContextSystem';
+import type { GpuGraphicsContext, GraphicsContextSystem } from './GraphicsContextSystem';
 import type { GraphicsView } from './GraphicsView';
 
 export interface GraphicsAdaptor
@@ -25,6 +24,15 @@ export interface GraphicsInstruction extends Instruction
     type: 'graphics';
     renderable: Renderable<GraphicsView>;
 }
+
+export interface GraphicsSystem
+{
+    graphicsContext: GraphicsContextSystem;
+    renderPipes: {
+        batch: BatchPipe
+    }
+}
+
 export class GraphicsPipe implements RenderPipe<GraphicsView>
 {
     /** @ignore */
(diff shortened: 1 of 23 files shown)


src/rendering/graphics/shared/GraphicsPipe.ts
```diff
@@ -189,6 +192,27 @@ export class GraphicsPipe implements RenderPipe<GraphicsView>
...
+        // TODO perhaps manage this outside this pipe? (a bit like how we update / add)
+        renderable.on('destroyed', () =>
+        {
+            this.destroyRenderable(renderable);
+        });
+
         return batches;
     }
+
+    destroyRenderable(renderable: Renderable<GraphicsView>)
```
Zyie, line 204:
```suggestion
    private destroyRenderable(renderable: Renderable<GraphicsView>)
```

src/rendering/mesh/shared/MeshView.ts
```diff
@@ -164,6 +165,29 @@ export class MeshView<GEOMETRY extends MeshGeometry = MeshGeometry>implements Vi
...
     }
 
+    /**
+     * Destroys this sprite renderable and optionally its texture.
+     * @param options - Options parameter. A boolean will act as if all options
+     *  have been set to that value
+     * @param {boolean} [options.texture=false] - Should it destroy the current texture of the renderable as well
+     * @param {boolean} [options.textureSource=false] - Should it destroy the textureSource of the renderable as well
+     */
+    destroy(options: DestroyOptions = false): void
```
Zyie, line 175:
```suggestion
    public destroy(options: DestroyOptions = false): void
```

src/rendering/mesh/shared/MeshView.ts
```diff
@@ -164,6 +165,29 @@ export class MeshView<GEOMETRY extends MeshGeometry = MeshGeometry>implements Vi
...
+        const destroyTexture = typeof options === 'boolean' ? options : options?.texture;
+
+        if (destroyTexture)
+        {
+            const destroyTextureSource = typeof options === 'boolean' ? options : options?.textureSource;
+
+            this._texture.destroy(destroyTextureSource);
+        }
+
+        this._texture = null;
```
Zyie, line 186: Does `owner` also need to be tidied here?

src/rendering/text/Text.ts
```diff
@@ -49,30 +42,4 @@ export class Text extends Container<TextView>
...
-    public destroy(options: DestroyOptions = false): void
-    {
-        if (typeof options === 'boolean')
-        {
-            options = { children: options };
-        }
-
-        options = Object.assign({}, defaultDestroyOptions, options);
-
-        super.destroy(options);
```
Zyie, line 74: We need to assign the default destroy options in the `TextView` now right?

tests/renderering/text/Text.test.ts
```diff
@@ -41,6 +43,48 @@ describe('Text', () =>
 
         text.addChild(child);
         text.destroy({ children: true, texture: true });
-        expect(childDestroyOpts).toEqual({ children: true, texture: true, textureSource: true });
+
+        expect(childDestroyOpts).toEqual({ children: true, texture: true });
```
Zyie, line 47: The original here should be true

src/rendering/text/Text.ts
GoodBoyDigital: i don't think so any more - as it does not extend sprite anymore (eg no texture to manage!)

tests/renderering/text/Text.test.ts
GoodBoyDigital: Should be cool - Text does not extend Sprite any more - so no texture to destroy

Zyie (approved):

PR #9469, opened 2023-06-12
- remove runners from textures

GoodBoyDigital:
replace runners in textures with emit emitters!
added a todo to destroy RenderTargets correctly (this will be in a subsequent PR) 

copilot:all

### DIFF
--- a/src/rendering/renderers/gpu/shader/BindResource.ts
+++ b/src/rendering/renderers/gpu/shader/BindResource.ts
@@ -1,4 +1,4 @@
-import type { Runner } from '../../shared/runner/Runner';
+import type { BindGroup } from './BindGroup';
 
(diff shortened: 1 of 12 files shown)


src/rendering/renderers/gpu/shader/BindResource.ts
```diff
@@ -10,9 +10,12 @@ export interface BindResource
     resourceId: number;
 
     /**
-     * dispatch whenever the underlying resource needs to change
+     * event dispatch whenever the underlying resource needs to change
      * this could be a texture or buffer that has been resized.
      * This is important as it allows the renderer to know that it needs to rebind the resource
      */
-    onResourceChange?: Runner;
+    // onResourceChange?: Runner;
```
Zyie, line 17:
```suggestion
```

PR #9480, opened 2023-06-16
fix(EventSystem): Bind touch events only if not supports pointer events

### LINKED ISSUE #7328: Event hierarchy discrepancy of document.body and stage between mobile and web
I was having issues with event handling in pixijs when I added a pointerdown handler to the body and to certain sprites on the stage. On web versions of pixi, the events fire in the expected order: the Sprite's pointerdown handler, then the body's pointerdown. However, I've found the opposite to occur on mobile: the body's pointerdown is fired first, then the sprite's. I'm not sure if this is a feature of pixi, but the inconsistency across web and mobile versions is inconvenient.

### Expected Behavior

A sprite is added to the PIXI app.stage. The sprite is marked interactive and is given a pointerdown handler that creates an alert indicating it is pressed. The document.body element is also given a pointerdown handler that creates an alert. When the sprite is pressed, the alert indicating it was pressed should pop up first, followed by the body's alert dialog.

### Current Behavior

On mobile, the body's alert message pops up first before the sprite.

### Steps to Reproduce

Run the following playground on mobile
https://www.pixiplayground.com/#/edit/T2G48_gKlANuyvYBLSNb3

### Environment

On web:
HP ProBook 11 G2, Chrome 89

On mobile:
BLU G5, Chrome
an older Samsung Galaxy, Chrome

rnike:
##### Description of change

Fix #7328, the event callback order is different between desktop and mobile, this is because on mobile the EventSystem is skipping the first callback fired by pointer and using the second callback fired by touch.

Bind touch events only if device not supports pointer event can fix the issue.

##### Pre-Merge Checklist


Zyie (approved):

bigtimebuddy:
@Zyie and @rnike Should this change replace #9591?

Zyie:
@bigtimebuddy yeah looks like this should replace the other PR

bigtimebuddy:
Thank you @rnike

Issue #8246, opened 2022-03-28
Unexpected BitmapText text width calculation


ssavova:
BitmapText textWidth has different value for lines with same amount of symbols (monospace font). Actually, according to the documentation, it is stated that at the latest pixi version (6.3.0), this bug must be fixed and the textWidth for characters '11' must be the same as for '33'. 

However, if you observe the **example here**: https://www.pixiplayground.com/#/edit/kvfNbd8URsDHhYWeVp83V 
You can see that the **bug still persist** and there is discrepancy between "11" and "33" textWidth. This discrepancy leads to unexpected and undesired BitmapText "shake". 

**Environment**

**pixi.js version:** > 6.0.0
**Browser & Version:** Chrome/FF
**OS & Version:** Windows
**How does it work now (pixi 6.3.0):** https://www.pixiplayground.com/#/edit/kvfNbd8URsDHhYWeVp83V

bigtimebuddy:
Arial is not a monospaced font, this should not work.

ssavova:
Yes, it`s my fault to consider Arial as monospaced font, but the same example, using pixi ( v5.3.0) behave as expected - no difference between  "11" and "33" textWidth. Could you please explain to me the reason for this behaviour.

bigtimebuddy:
The reason that this issue is happening is because older versions of Pixi did not consider kerning between letters for BitmapFont. This is actually really useful if you want to have a words like `ball` but with the `ll` to be closer together. This is achieved with measuring kernings between letters when the BitmapFont is created. It was introduced here: #6888 (v6.0.0)

If you want to ignore kernings, you can add this to your example:
```js
const { chars } = PIXI.BitmapFont.from(customFont.name, customFont.style, customFont.options);
Object.keys(chars).forEach((code) => (chars[code].kerning = {}));
```

BitmapFont _may_ benefit from having an option to skip kernings being generated. PR is welcome to add this to the BitmapFont.from options.

GoodBoyDigital:
closing as we also have `skipKerning` in both v7 and v8 👍

PR #9496, opened 2023-06-22
Add option to skip kerning generation in BitmapFont

yordan-kanchelov:
This pull request addresses the suggestion in issue [#8246](https://github.com/pixijs/pixijs/issues/8246), adding an option to skip the generation of kernings in the `BitmapFont.from` method.

The brute force kerning method was introduced in PR [#6888](https://github.com/pixijs/pixijs/pull/6888). This method can be computationally intensive, especially for larger character sets. 

By adding an option to bypass this step, developers can reduce the computational overhead when generating bitmap fonts, particularly in cases where the kerning information isn't required.

The changes involve adding a new property in the `IBitmapFontOptions` interface to allow skipping the kerning generation step. If this option is set to `true`, the method will bypass the kerning calculation loop, resulting in a performance gain.

##### Pre-Merge Checklist

- [ ❌ ] Tests and/or benchmarks are included
- [ ✅ ] Documentation is changed or added ( if jsdoc is accounted for documentation ) 
- [ ✅ ] Lint process passed (`npm run lint`)
- [ ✅ ] Tests passed (`npm run test`)

### DIFF
--- a/packages/text-bitmap/src/BitmapFont.ts
+++ b/packages/text-bitmap/src/BitmapFont.ts
@@ -523,7 +523,8 @@ export class BitmapFont
             positionX = Math.ceil(positionX);
         }
 
-        if (!options?.skipKerning) {
+        if (!options?.skipKerning)
+        {
             // Brute-force kerning info, this can be expensive b/c it's an O(n²),
             // but we're using measureText which is native and fast.
             for (let i = 0, len = charsList.length; i < len; i++)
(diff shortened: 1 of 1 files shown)


yordan-kanchelov:
Happy to contribute! 😄 

Could we also add this update to the v6.x branch. Many of my projects still use v6 and It would be very beneficial for us.

bigtimebuddy (approved):

PR #9497, opened 2023-06-22
NPE on INT/destroyed texture

kr0n4ik:
Here is NPE when one bound texture is INT, and another is destroyed.


Zyie:
Hey,

PR #9512, opened 2023-07-03
fix Container.parent type

reececomo:
##### Description of change
Fix incorrect `Container.parent` type to include optionality.

This makes it compatible with strict TypeScript (i.e. `noUncheckedIndexedAccess` and `strictNullChecks`).

Currently this triggers the [@typescript-eslint/no-unnecessary-condition](https://typescript-eslint.io/rules/no-unnecessary-condition) eslint rule because the typing is incorrect.

##### Pre-Merge Checklist
- [x] Tests and/or benchmarks are included - **N/A** type inference change
- [x] Documentation is changed or added - **N/A** self-documenting

### DIFF
--- a/packages/display/src/Container.ts
+++ b/packages/display/src/Container.ts
@@ -90,7 +90,7 @@ export class Container<T extends DisplayObject = DisplayObject> extends DisplayO
      * Will get automatically set to true if a new child is added, or if a child's zIndex changes.
      */
     public sortDirty: boolean;
-    public parent: Container;
+    public parent?: Container;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
@reececomo I think this is a useful change for detached elements or the root container. But it seems there are a lot of type error introduced by this change. Do you think you could look at it: 
https://github.com/pixijs/pixijs/actions/runs/5439198743/jobs/9891370406?pr=9512

reececomo:
I don't have the bandwidth to do this right now, but will leave the Issue up for later

PR #9514, opened 2023-07-03
fix(application): prevent read property of null if the application is destroyed

ronny1020:
##### Description of change

If the application is destroyed, the reading view or screen will cause an error, read property of null. The error is fixed by optional chaining. This PR should not affect any other features. 

##### Pre-Merge Checklist


bigtimebuddy (approved):
Easy enough, thanks!

PR #9528, opened 2023-07-07
Fix: Remove console logs from minified builds

### LINKED ISSUE #9516: Bug: Code inside of `#if _DEBUG ... #endif` is not removed from production build
### Current Behavior

Code inside of `#if _DEBUG ... #endif` is included in the production build.

### Expected Behavior

Code inside of `#if _DEBUG ... #endif` is not included in the production build.

### Steps to Reproduce

In production:
```js
PIXI.Assets.get("invalid.png")
```
https://github.com/pixijs/pixijs/blob/356abaaad852b248f0aa3f6873f5d7d2a56e3a50/packages/assets/src/cache/Cache.ts#L57-L59

Warning:
```
[Assets] Asset id invalid.png was not found in the Cache
```

### Environment

- **`pixi.js` version**: 7.2.4

### Possible Solution

_No response_

### Additional Information

_No response_

bigtimebuddy:
Fixes #9516 

## Fixed

* Strips deprecation warnings and console lots from the minified bundle builds

## Chores

* Remove the use of `rollup-plugin-jscc` and use `esbuild` functionality instead
* Replace `// #if _DEBUG` & `// #endif` syntax with `if (process.env.DEBUG) {}`


bigtimebuddy:
This increases the `pixi.js` bundle size from 462KB to 505KB. Need to investigate further...

bigtimebuddy:
The reason the bundle size is much bigger is because before we use to:
* source -> es2020 (from `lib` folder) -> es2017

But now we are building against the source directly:
* source -> es2017

esbuild inlines a lot more runtime code transpile helpers doing this path (e.g., `__publicField()`). And these helper functions are not de-duped so it adds a lot more overhead ~10%.

bigtimebuddy:
Fixed! I was able to fix this by forcing `useDefineForClassFields` to be `false` (see [esbuild docs](https://esbuild.github.io/content-types/#tsconfig-json))

This shaved off a little in the file-size (which likely is the removal of the logs/deprecation warnings).

462KB (before) -> 452KB (after)

Zyie (approved):

2220759:
ncjksn.bmbd

2220759:
gjnjksgdmlmsblkm;/hmslx.nxjksnb

PR #9530, opened 2023-07-08
Add option 'strict' to Assets to optionally avoid expensive detection methods

olafurkarl:
##### Description of change

New 'strict' option to the AssetInitOptions.texturePreferences object added, which will skip expensive calls to detect file formats.

##### Pre-Merge Checklist

Still working on the above, having some trouble setting up my dev environment. :pray:

### DIFF
--- a/packages/assets/src/Assets.ts
+++ b/packages/assets/src/Assets.ts
@@ -52,6 +52,11 @@ export interface AssetInitOptions
          * ['avif', 'webp', 'png', 'jpg', 'jpeg', 'webm', 'mp4', 'm4v', 'ogv']
          */
         format?: ArrayOr<string>;
+        /**
+         * If true, don't attempt to detect whether browser has preferred formats available.
+         * May result in increased performance as it skips detection step.
+         */
+        strict?: boolean;
     };
 
     /** advanced - override how bundlesIds are generated */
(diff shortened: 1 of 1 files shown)


packages/assets/src/Assets.ts
```diff
@@ -52,6 +52,11 @@ export interface AssetInitOptions
          * ['avif', 'webp', 'png', 'jpg', 'jpeg', 'webm', 'mp4', 'm4v', 'ogv']
          */
         format?: ArrayOr<string>;
+        /**
+         * If true, don't attempt to detect whether browser has preferred formats available.
+         * May result in increased performance as it skips detection step.
+         */
+        strict?: boolean;
```
bigtimebuddy, line 59: How about calling this "skipDetection"? "strict" is a little too generic/non-descriptive.

packages/assets/src/Assets.ts
```diff
@@ -301,14 +306,17 @@ export class AssetsClass
         {
             const formatPref = options.texturePreference?.format;
 
-            formats = (typeof formatPref === 'string') ? [formatPref] : formatPref;
+            formats = typeof formatPref === 'string' ? [formatPref] : formatPref;
 
-            // we should remove any formats that are not supported by the browser
-            for (const detection of this._detections)
+            if (!options.texturePreference?.strict)
```
bigtimebuddy, line 311:
```suggestion
            if (!options.texturePreference?.skipDetection)
```

bigtimebuddy:
I assume this fixes #9531?

I wonder if this is the right solution for the underlying problem. Since it's a performance issue with avif, having a flag to opt-out of detection seems like there maybe a better solution.

olafurkarl:
> I assume this fixes #9531?
> 
> I wonder if this is the right solution for the underlying problem. Since it's a performance issue with avif, having a flag to opt-out of detection seems like there maybe a better solution.

Even with this merged I'd keep #9531 open. :)

It does not address the problem for detections being slow. 

As far as I've found, detecting the various formats will always have some performance impact to startup, though, so I think having a flag to skip them makes sense for cases where they are not needed.

Zyie:
We have narrowed down the performance issue aspect and resolved it in this PR: #9537

olafurkarl:
@Zyie Might it still be a good idea to add this skip option? There's still 200-300 ms to be shaved off from the startup time. :)

bigtimebuddy (changes requested):
Would you please revert all the style/whitespace changes?

Issue #9531, opened 2023-07-09
Bug: Assets.init takes a long time (> 4.5 seconds) on detection step


olafurkarl:
### Current Behavior

When calling `Assets.init` (e.g. through `Assets.load` or by itself), then detection step is taking ~4700 ms in my project.

### Expected Behavior

Should take just a few milliseconds to run detections.

### Steps to Reproduce

Run `Assets.load` or `Assets.init` in latest `chromium`.

### Environment

- **`pixi.js` version**: 7.2.4
- **Browser & Version**: Version 114.0.5735.198 (Official Build) Arch Linux (64-bit)
- **OS & Version**: Up to date arch linux
- **Running Example**: Haven't been able to get it running in a code sandbox, but running `Assets.load` with no assets specified will produce the issue locally.

### Possible Solution

https://github.com/pixijs/pixijs/pull/9530

### Additional Information

The issue does not occur when I use Chrome, so to produce the issue it is probably only happening on certain browser (e.g. latest chromium for me).

olafurkarl:
I've tracked it down further to a specific detection, `detectAvif.ts`
```
// ...
        console.time('detectAvif');
        if (!globalThis.createImageBitmap) return false;

        // eslint-disable-next-line max-len
        const avifData = '<blob redacted for brevity>';
        
        // !! This line takes the bulk of the time !!
        const blob = await settings.ADAPTER.fetch(avifData).then((r) => r.blob());
        console.timeLog('detectAvif', 'after first await') // 4233 ms when I tested

        return createImageBitmap(blob).then(() => true, () => true);
    },
// ...

```

bigtimebuddy:
@Zyie I think you added detectAvif.ts. Any thoughts why this is so slow?

Zyie:
I'm not entirely sure why this would be slow, we have the exact same logic for detecting `webp` files, so i would have thought both would have been slow. My only guess would be an issue with browser implementation of `createImageBitmap` with `avif` files

It might be worth experimenting with using something like `new Image()` instead of `createImageBitmap` to see if that makes a difference

I've also not been able to reproduce this issue so it's hard to debug

olafurkarl:
> It might be worth experimenting with using something like new Image() instead of createImageBitmap to see if that makes a difference

The specific step that is slow is the `await settings.ADAPTER.fetch(avifData)`, before we reach the `createImageBitmap` step.

Zyie:
[...]

PR #9537, opened 2023-07-11
fix image detection speed

Zyie:
I've updated the avif/webp detection to use `new Image()` which appears to fix the issues with them being slow. See #9531


packages/assets/src/detections/parsers/detectWebp.ts
```diff
@@ -12,9 +12,21 @@ export const detectWebp = {
         if (!globalThis.createImageBitmap) return false;
```
bigtimebuddy, line 12: We don't need this `createImageBitmap` check anymore

bigtimebuddy (approved):
Small fix, otherwise looks good.

PR #9539, opened 2023-07-11
Fixes formats not being added correctly

Zyie:
While i was looking at #9530 I noticed we are doing something wrong with the detection logic.

When a user sets `texturePreferrences.format` it completely overrides all default formats. What we should be doing instead is adding these user defined formats to the start of the array, and the others can be used as fallbacks

This issue also meant if you defined a preferred format like this: `['png', 'avif']`, something like pixi sound detections would never be added

This does also mean that using `texturePreferrence.format = ['mp3', 'ogg', 'pmg', 'jpg']` is something you can do, so maybe we should deprecate this option and just use `preferredFormat = [...]` instead, since it has nothing to do with textures explicitly

Also while i was here is did add the `skipDetections` option from #9530

### DIFF
--- a/packages/assets/src/Assets.ts
+++ b/packages/assets/src/Assets.ts
@@ -54,6 +54,12 @@ export interface AssetInitOptions
         format?: ArrayOr<string>;
     };
 
+    /**
+     * If true, don't attempt to detect whether browser has preferred formats available.
+     * May result in increased performance as it skips detection step.
+     */
+    skipDetections?: boolean;
+
     /** advanced - override how bundlesIds are generated */
     bundleIdentifier?: BundleIdentifierOptions;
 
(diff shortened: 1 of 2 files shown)


packages/assets/src/Assets.ts
```diff
@@ -838,6 +821,49 @@ export class AssetsClass
...
+        for (const detection of options.detections)
+        {
+            if (options.skipDetections || await detection.test())
+            {
+                formats = await detection.add(formats);
+            }
+            else if (!options.skipDetections)
+            {
+                formats = await detection.remove(formats);
+            }
```
bigtimebuddy, line 858:
```suggestion
            if (skipDetections || await detection.test())
            {
                formats = await detection.add(formats);
            }
            else if (!skipDetections)
            {
                formats = await detection.remove(formats);
            }
```

packages/assets/src/Assets.ts
```diff
@@ -54,6 +54,12 @@ export interface AssetInitOptions
         format?: ArrayOr<string>;
     };
 
+    /**
+     * If true, don't attempt to detect whether browser has preferred formats available.
+     * May result in increased performance as it skips detection step.
+     */
+    skipDetections?: boolean;
```
bigtimebuddy, line 61: Do you think this should be nested under `texturePreference`?

PR #9551, opened 2023-07-15
Use fetch to load video textures

dev7355608:
Video loading may never resolve, because the video load can be suspended by the browser (`suspend` event). It seems that the browser is free to just suspend the video loading and seems to do that frequently if other assets are loaded at the same time. Spamming `videoElement.load()` in `onsuspend` seems to eventually make the video load, but only after a significant delay. Using `fetch` and a blob URL fixes this issue. We should probably have something in `VideoResource` to handle `suspend` events, but I'm not entirely sure what the correct approach would be.

### DIFF
--- a/packages/assets/src/loader/parsers/textures/loadVideo.ts
+++ b/packages/assets/src/loader/parsers/textures/loadVideo.ts
@@ -79,7 +79,11 @@ export const loadVideo = {
             });
 
             base.resource.src = url;
-            texture = createTexture(base, loader, url, blobURL);
+            texture = createTexture(base, loader, url);
+            texture.baseTexture.once('destroyed', () =>
+            {
+                URL.revokeObjectURL(blobURL);
+            });
         }
         catch (e)
         {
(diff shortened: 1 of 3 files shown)


miltoncandelero:
Is this gets approved... Could we look into same solution for pixi sound? Event "canplaythrough" is really untrustworthy on mobile network 🤔

packages/assets/src/loader/parsers/textures/loadVideo.ts
```diff
@@ -57,23 +57,36 @@ export const loadVideo = {
...
-        base.resource.src = url;
+            const base = new BaseTexture(src, {
+                alphaMode: await utils.detectVideoAlphaMode(),
+                resolution: utils.getResolutionOfUrl(url),
+                ...loadAsset?.data,
+            });
 
-        const texture = createTexture(base, loader, url);
+            base.resource.src = url;
+            texture = createTexture(base, loader, url, blobURL);
```
bigtimebuddy, line 82: Do we need the abstraction on createTexture? Alternatively, we could not pass through the objecturl and instead add another event listener on the BaseTexture destroyed? Keeps all the objecturl stuff contained here.

bigtimebuddy (commented):
Overall, I don't have a problem leaning into object url for this case. Just a small suggestion to simplify.

bigtimebuddy (approved):

PR #9571, opened 2023-07-27
Default Renderer fallback to settings.RESOLUTION

### LINKED ISSUE #9429: Bug: Application no longer uses the global `resolution` setting (after updating from v7.1.3 to v7.2.4)
### Current Behavior

In v7.2.4, I have to do:

```ts
settings.RESOLUTION = window.devicePixelRatio; // I have an hi-dpi screen with a devicePixelRatio of 1.25
settings.RENDER_OPTIONS!.autoDensity = true; // I have to use a "!", otherwise I get an error about RENDER_OPTIONS potentially being undefined, it was not the case before

const application = new Application({
  resolution: settings.RESOLUTION,
});
```

If I don't set the application's resolution option to `1.25`, then the application's renderer has a resolution set to `1` and everything is blurry on my screen.

### Expected Behavior

In v7.1.3, I was able to do:

```ts
settings.RESOLUTION = window.devicePixelRatio; // I have an hi-dpi screen with a devicePixelRatio of 1.25
settings.RENDER_OPTIONS.autoDensity = true;

const application = new Application();
```

Here, the application's renderer has a `resolution` of `1.25` and nothing is blurry on my screen.

### Steps to Reproduce

- Set the global `settings.RESOLUTION` to a value different than 1
- Create a new application

The application's renderer has a resolution of 1 instead of the global resolution value (you can `console.log` the application to see the internal resolution value of the renderer).

### Environment

- **`pixi.js` version**: *7.2.4*
- **Browser & Version**: *Firefox 113* & *Chrome113*

### Possible Solution

Manually set the resolution in the application's options.

### Additional Information

I don't know if it's an intended change or not, but I was not able to figure that out by looking at the changelog between the v7.1.3 & v7.2.4, so I'd rather report it as a bug so people may find this information somewhere while it's being clarified 😅

bigtimebuddy:
Fixes #9429

The default value of `RENDER_OPTIONS.resolution` should be undefined. This way it will fallback to `settings.RESOLUTION` in the ViewSystem constructor.

Broken: https://jsfiddle.net/bigtimebuddy/tmg2u8rL/
Fixed: https://jsfiddle.net/bigtimebuddy/vjLx02tp/

### DIFF
--- a/packages/core/src/view/ViewSystem.ts
+++ b/packages/core/src/view/ViewSystem.ts
@@ -48,7 +48,7 @@ export interface ViewSystemOptions
 export class ViewSystem implements ISystem<ViewSystemOptions, boolean>
 {
     /** @ignore */
-    static defaultOptions = {
+    static defaultOptions: ViewSystemOptions = {
         /**
          * {@link PIXI.IRendererOptions.width}
          * @default 800
@@ -67,7 +67,7 @@ export class ViewSystem implements ISystem<ViewSystemOptions, boolean>
          * @default PIXI.settings.RESOLUTION
          * @memberof PIXI.settings.RENDER_OPTIONS
          */
-        resolution: settings.RESOLUTION,
+        resolution: undefined,
         /**
          * {@link PIXI.IRendererOptions.autoDensity}
          * @default false


GoodBoyDigital (approved):
lgtm!

PR #9580, opened 2023-08-02
Add the `Texture.destroyed` property/event. Log a warning if a `Texture` managed by `Assets` is destroyed but its `BaseTexture` is not

dev7355608:
- Added `Texture.destroyed` property
- Added `destroyed` event to `Texture`
- Forced the destruction of the base texture if the texture of the asset is destroyed. Currently, if the texture is destroyed without the `destroyBase` flag, then the destroyed texture remains in the asset cache and the base texture/resource is not destroyed.

### DIFF
--- a/packages/assets/src/loader/parsers/textures/utils/createTexture.ts
+++ b/packages/assets/src/loader/parsers/textures/utils/createTexture.ts
@@ -21,13 +21,22 @@ export function createTexture(base: BaseTexture, loader: Loader, url: string)
     };
 
     // remove the promise from the loader and the url from the cache when the texture is destroyed
-    texture.baseTexture.once('destroyed', unload);
+    texture.baseTexture.once('destroyed', () =>
+    {
+        if (url in loader.promiseCache)
+        {
+            console.warn('[Assets] A BaseTexture managed by Assets was destroyed instead of unloaded! '
+                + 'Use Assets.unload() instead of destroying the BaseTexture.');
+            unload();
+        }
+    });
     texture.once('destroyed', () =>
     {
         if (!base.destroyed)
         {
+            console.warn('[Assets] A Texture managed by Assets was destroyed instead of unloaded! '
+                + 'Use Assets.unload() instead of destroying the Texture.');
             unload();
-            console.warn('[Assets] A Texture managed by Assets was destroyed but its BaseTexture was not!');
         }
     });
 
--- a/packages/assets/src/loader/parsers/textures/utils/createTexture.ts
+++ b/packages/assets/src/loader/parsers/textures/utils/createTexture.ts
@@ -10,15 +10,24 @@ export function createTexture(base: BaseTexture, loader: Loader, url: string)
     base.resource.internal = true;
 
     const texture = new Texture(base);
-
-    // remove the promise from the loader and the url from the cache when the texture is destroyed
-    texture.baseTexture.once('destroyed', () =>
+    const unload = () =>
     {
         delete loader.promiseCache[url];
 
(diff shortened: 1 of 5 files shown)


bigtimebuddy (dismissed):

packages/assets/src/loader/parsers/textures/utils/createTexture.ts
```diff
@@ -21,6 +21,14 @@ export function createTexture(base: BaseTexture, loader: Loader, url: string)
...
         }
     });
+    // force the destruction of the base texture if the texture is destroyed
+    texture.once('destroyed', () =>
+    {
+        if (!base.destroyed)
+        {
+            base.destroy();
+        }
+    });
```
bigtimebuddy, line 31: The other changes in the PR are fine, but this is a breaking change.

Users should define if they want to destroy the baseTexture (e.g. `texture.destroy(true)`). We don't handle this automatically, it changes the relationship between Textures and BaseTexture.

dev7355608: True, but if the user destroys the `Texture` without the `BaseTexture`, then the asset manager will always return this destroyed `Texture`, and the `BaseTexture` and resource may be never destroyed/released, unless the user calls `baseTexture.destroy()` later (`Assets.unload` cannot destroy it anymore, because the `Texture` is already destroyed). It's should probably not be possible to destroy a `Texture` of the asset manager without destroying the `BaseTexture` in the first place. We could add a warning here when the texture is destroyed without the base. Or we don't destroy the base texture, but remove it from the asset manager. I don't think we would want the asset manager to return destroyed textures. We should probably add a warning in this case as well, because it's very unlikely that the user indented to unload the texture from the asset manager without destroying the resource.

bigtimebuddy: A warning seems more appropriate. If someone is using Assets to manage the lifecycle of their assets (recommended), we don't want them calling destroy manually. They should be using unload instead.

dev7355608: Removed the destruction of the base texture and replaced it the unloading we do when the base is destroyed and added a warning.

bigtimebuddy:
Much better. Thoughts on this @Zyie?

packages/assets/src/loader/parsers/textures/utils/createTexture.ts
```diff
@@ -10,16 +10,25 @@ export function createTexture(base: BaseTexture, loader: Loader, url: string)
...
         delete loader.promiseCache[url];
 
         if (Cache.has(url))
         {
             Cache.remove(url);
         }
+    };
+
+    // remove the promise from the loader and the url from the cache when the texture is destroyed
+    texture.baseTexture.once('destroyed', unload);
```
bigtimebuddy, line 24:
```suggestion
   base.once('destroyed', unload);
```

dev7355608: If the base texture is destroyed, we surely want to remove it from the asset manager, no? Imagine someone destroys only the base texture, but not the texture. We probably don't want the texture with a destroyed base texture to remain in the asset manager.

Maybe we should have a warning to the base texture case as well?

bigtimebuddy: We will have similar problems with bitmap fonts, spritesheets, etc. which generally all have a destroy path too. I think we need a more wholistic approach to this problem: encouraging unload over destroy.

PR #9581, opened 2023-08-02
Set `valid` to false if the `BaseTexture` is destroyed

dev7355608:
Fixed: `BaseTexture`s remained `valid` after destruction.


bigtimebuddy (approved):
LGTM

PR #9595, opened 2023-08-11
Expand spritesheet types

trezy:
##### Description of change
This PR expands the types on spritesheet data. Most people are likely importing their spritesheet atlases from external JSON files, and thus will never run into an issue. That said, this change is valuable for those who are generating spritesheet atlases programmatically at runtime and are running into Typescript errors.

FWIW, none of these types are used within Pixi, but they **are** used within other applications and libraries. If they're not added, we'll have to generate 2 separate structures to support Pixi and other tools.

The following optional keys have been added based on the spritesheet atlas JSON output generated by [Aseprite](https://www.aseprite.org/):

* `meta.app`: A string representing the app used to generate the spritesheet atlas.
* `meta.format`: The color space used in the spritesheet, e.g. RGBA8888.
* `meta.frameTags`: An array of tags (basically keyframes) used in the spritesheet. Typically used to point to collections of frames that represent a particular animation.
* `meta.layers`: Information about the layers used in the source file.
* `meta.slices`: Slices of the sprite, typically used for indicating things like hitboxes. Slices also include the `keys` property, which indicates where the slice is positioned on each sprite frame.

##### Pre-Merge Checklist

### DIFF
--- a/packages/spritesheet/src/Spritesheet.ts
+++ b/packages/spritesheet/src/Spritesheet.ts
@@ -34,12 +34,44 @@ export interface ISpritesheetFrameData
  */
 export interface ISpritesheetData
 {
-    frames: utils.Dict<ISpritesheetFrameData>;
     animations?: utils.Dict<string[]>;
+    frames: utils.Dict<ISpritesheetFrameData>;
     meta: {
+        app?: string;
+        format?: string;
+        frameTags?: {
+            from: number;
+            name: string;
+            to: number;
+            direction: string;
+        }[];
+        image?: string;
+        layers?: {
+            blendMode: string;
+            name: string;
+            opacity: number;
+        }[];
         scale: string;
+        size?: {
+            h: number;
+            w: number;
+        };
+        slices?: {
+            color: string;
+            name: string;
+            keys: {
+                frame: number,
+                bounds: {
+                    x: number;
+                    y: number;
+                    w: number;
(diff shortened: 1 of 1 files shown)


packages/spritesheet/src/Spritesheet.ts
```diff
@@ -34,12 +36,44 @@ export interface ISpritesheetFrameData
...
+            blendMode: string;
+            name: string;
+            opacity: number;
+        }[];
         scale: string;
+        size?: {
+            h: number;
+            w: number;
+        };
+        slices: {
```
bigtimebuddy, line 61:
```suggestion
        slices?: {
```

bigtimebuddy:
Thanks for this fix @trezy. Not sure if `slices` is the only issue, but could you resolve it and make sure that `npm run build:types` and `npm run types` works successfully for you?

packages/spritesheet/src/Spritesheet.ts
trezy: Done in the latest commit. 😁

trezy:
Just fixed the issue and made `slices` optional. `npm run build:types` and `npm run types` both work for me locally. Those commands are probably worth adding to the contributing docs and the PR template, as well. I can make another PR for that. 😁

bigtimebuddy (approved):

PR #9599, opened 2023-08-14
Null Checks

NotARoomba:
##### Description of change
Added null checks to fix issues with ```<React.StrictMode>``` double mounting.
Fixes [this](https://github.com/davidfig/pixi-viewport/issues/438) and [this](https://github.com/davidfig/pixi-viewport/issues/441) issue.

##### Pre-Merge Checklist

### DIFF
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -433,6 +433,7 @@ export abstract class DisplayObject extends utils.EventEmitter<DisplayObjectEven
     /** Recursively updates transform of all objects from the root to this one internal function for toLocal() */
     protected _recursivePostUpdateTransform(): void
     {
+        if (!this.transform) return;
         if (this.parent)
         {
             this.parent._recursivePostUpdateTransform();
--- a/packages/display/src/DisplayObject.ts
+++ b/packages/display/src/DisplayObject.ts
@@ -433,25 +433,23 @@ export abstract class DisplayObject extends utils.EventEmitter<DisplayObjectEven
     /** Recursively updates transform of all objects from the root to this one internal function for toLocal() */
     protected _recursivePostUpdateTransform(): void
     {
-        if (!this.transform) return;
         if (this.parent)
         {
             this.parent._recursivePostUpdateTransform();
-            this.transform.updateTransform(this.parent.transform);
+            this.transform?.updateTransform(this.parent.transform);
         }
         else
         {
-            this.transform.updateTransform(this._tempDisplayObjectParent.transform);
+            this.transform?.updateTransform(this._tempDisplayObjectParent.transform);
         }
     }
 
     /** Updates the object transform for rendering. TODO - Optimization pass! */
     updateTransform(): void
     {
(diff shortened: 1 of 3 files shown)


packages/display/src/DisplayObject.ts
```diff
@@ -447,6 +448,7 @@ export abstract class DisplayObject extends utils.EventEmitter<DisplayObjectEven
     /** Updates the object transform for rendering. TODO - Optimization pass! */
     updateTransform(): void
     {
+        if (!this.transform) return;
         this._boundsID++;
 
         this.transform.updateTransform(this.parent.transform);
```
bigtimebuddy, line 454: What about using optional-chaining instead of an if-statement to bail-out?

```suggestion
        this.transform?.updateTransform(this.parent.transform);
```

NotARoomba: Yes that looks cleaner.

NotARoomba:
```packages/assets/test/bundle.tests.js``` has strange behavior at line ~96 and throws and fails tests randomly sometimes, awaiting for more time seems to have fixed it

File: packages/display/test/DisplayObject.tests.ts [hunk 2]
```diff
@@ -266,6 +266,24 @@ describe('DisplayObject', () =>
...
+        it('should return undefined as transform may be null', () =>
+        {
+            // @ts-expect-error - instantiating DisplayObject
+            const child = new DisplayObject();
+            const container = new Container();
+
+            container.addChild(child);
+
+            container.transform = null;
+            child.transform = null;
```
bigtimebuddy, line 280: No one should be assigning `null` to those transform props. A better test is probably to destroy the DisplayObject & Container. Would assuming this gives the same result, exercises the same paths.
```suggestion
            container.destroy({ children: true });
```

bigtimebuddy:
Thanks for making those fixes.

I read through the attached issues more, and I'm not convinced this project is the best place to resolve the underlying issue. Why not react-pixi handle double-calls gracefully? The optional-chaining, especially when transpiled, could introduce some performance issue to these host code paths, so I'm a little concerned about adding those. Lastly, PixiJS doesn't promise that destroyed objects to gracefully be handled/rendered/added/managed and, in general, that responsibility is on users to manage the lifecycle of their objects. I think, in fact, the preference is failing hard is better than silently ignore.

What might help this is if you could create a reproduction to illustrate the whole problem? I don't think the test here is sufficient describe the whole problem.

bigtimebuddy:
I had a chat with the team about this PR. We decided that we are going to close it for a few reasons:
* **Inappropriate Layer** - solution should be added to pixi-react or another layer that's more closely aligned with React's strict behavior. It's outside the scope of the main project.
* **Hot Code** - this adds optional chaining to hot-code (performance critical) and we'd not have a potential regression in performance
* **Prefer Fail Hard** - We'd prefer that code breaks hard instead of silently failing (or when appropriate throw errors). We generally consider users responsible for managing the life-cycle of their display objects and we feel this change breaks that design intent.

PR #9635, opened 2023-09-07
Add more deprecations

Zyie:


achamas-playco (approved):
Great stuff! Looks good to me, at least from a high level.
A suggestion would be the deprecation changes were in one commit, or even a separate PR?
This would make the actual code changes easier to review.

src/rendering/graphics/shared/Graphics.ts
```diff
@@ -34,4 +38,242 @@ export class Graphics extends Container<GraphicsView>
...
+    }
+
+    // v7 deprecations
+    /**
+     * @param {...any} args
+     * @deprecated since 8.0.0
+     */
+    public drawCircle(...args: Parameters<GraphicsContext['circle']>): this
+    {
+        deprecation('8.0.0', 'Graphics#drawCircle has been renamed to Graphics#circle');
```
bigtimebuddy, line 220: I don't know how you set it up in v8, but in v7, I'm removing deprecation logs (#9637 9528) from the release build by doing this:

```js
if (process.env.DEBUG)
{
  deprecation('8.0.0', '...');
}
```

GoodBoyDigital: Hey @bigtimebuddy , you think if we added process.env.DEBUG into the deprecation function.. it would see it as dead code and just remove the function?

GoodBoyDigital (approved):

src/rendering/graphics/shared/Graphics.ts
bigtimebuddy: Sure, that would remove the actual console log, but we actually want to remove the deprecation call itself or else we are inlining a bunch of strings that don't do anything.

PR #9636, opened 2023-09-07
fix issue with loading asset without an alias

### LINKED ISSUE #9633: Bug: loadParser option in Assets.load doesn't work without specifying alias
### Current Behavior

`loadParser` option in `Assets.load` doesn't work without specifying `alias`.

```ts
PIXI.Assets.load({
    src: 'https://www.pixiplayground.com/',
    loadParser: 'loadTxt',
}).then(data => {
    console.log(typeof(data));
    console.log(data);
});
```

pixi.js 7.3.0-rc.2 Output:

```
object
null
```

### Expected Behavior

Expected Output:

```
string
<!DOCTYPE html>
<html lang="en-US">
    <head>
        ...
```

### Steps to Reproduce

See above.

### Environment

- **`pixi.js` version**: 7.3.0-rc.2
- **Browser & Version**: Chrome 116.0.5845.180
- **Running Example**: <https://www.pixiplayground.com/#/edit/1t-YI25Bz_3OYMj16Gl3z>

### Possible Solution

If we add `alias` to the options, the `Assets.load` can work properly.

```js
// This works
PIXI.Assets.load({
    src: 'https://www.pixiplayground.com/',
    alias: '...',
    loadParser: 'loadTxt',
}).then(...)
```

---

Seems that the problem comes from here:

<https://github.com/pixijs/pixijs/blob/f7e16da2b076f146f8954c708455112d2a2cc304/packages/assets/src/resolver/Resolver.ts#L496>

The `alias` and `name` are both `undefined`, so `aliasesToUse` become `[undefined]`.

<https://github.com/pixijs/pixijs/blob/f7e16da2b076f146f8954c708455112d2a2cc304/packages/assets/src/resolver/Resolver.ts#L550-L553>

... And finally `resolvedAssets` is written into `this._assetMap[undefined]`, causing the issue.

---

A quick fix is:

```diff
- const aliasesToUse = convertToList<string>(alias || name);
+ const aliasesToUse = convertToList<string>(alias || name || src as string);
```

But this only works when `src` is a string.

### Additional Information

_No response_

Zyie:
Fixes #9633 and also fixes the use case where `src` can also be an object containing more src's


bigtimebuddy:
Test?

Zyie:
> Test?

Woops I didn't commit the whole thing, should have tests in now

I also found an issue with it attempting to `add` the same asset multiple times. I've updated the test for this as well to check all use cases

Zyie:
hmm weird... testings are passing locally

going to put this as a draft PR until fixed. Will ping again when ready

packages/assets/test/assets.tests.ts
```diff
@@ -440,33 +440,70 @@ describe('Assets', () =>
...
+                src: 'textures/bunny.4.png',
+            },
+        }));
+        promises.push(Assets.load({
+            src: {
+                src: 'textures/bunny.4.png',
+            },
+        }));
+
+        await Promise.all(promises);
```
bigtimebuddy, line 491: Could we make this a little less verbose?
```suggestion
        await Promise.all([
            Assets.load('textures/bunny.png'),
            Assets.load('textures/bunny.png'),
            Assets.load({ src: 'textures/bunny.1.{png,webp}' }),
            Assets.load({ src: 'textures/bunny.1.{png,webp}' }),
            Assets.load({ src: ['textures/bunny.2.{png,webp}'] }),
            Assets.load({ src: ['textures/bunny.2.{png,webp}'] }),
            Assets.load({
                src: [
                    {
                        src: 'textures/bunny.3.png',
                    },
                    {
                        src: 'textures/bunny.3.webp',
                    }
                ]
            }),
            Assets.load({
                src: [
                    {
                        src: 'textures/bunny.3.png',
                    },
                    {
                        src: 'textures/bunny.3.webp',
                    }
                ]
            }),
            Assets.load({
                src: {
                    src: 'textures/bunny.4.png',
                },
            }),
            Assets.load({
                src: {
                    src: 'textures/bunny.4.png',
                },
            }),
        ]);
```

bigtimebuddy (approved):

PR #9643, opened 2023-09-10
Fix: Round `Sprite` vertices in `calculateTrimmedVertices` if `roundPixels`

dev7355608:
Rounding of vertices is missing in `Sprite.calculateTrimmedVertices`.

### DIFF
--- a/packages/sprite/src/Sprite.ts
+++ b/packages/sprite/src/Sprite.ts
@@ -349,6 +349,16 @@ export class Sprite extends Container
         // xy
         vertexData[6] = (a * w1) + (c * h0) + tx;
         vertexData[7] = (d * h0) + (b * w1) + ty;
+
+        if (this._roundPixels)
+        {
+            const resolution = settings.RESOLUTION;
+
+            for (let i = 0; i < vertexData.length; ++i)
+            {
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
Can you provide and example of what this fixes?

dev7355608:
A sprite with a trimmed texture should have the same bounds as with the original untrimmed texture. So `calculateTrimmedVertices` should also round vertices. Example:
```js
width = height = 4;
frame = new PIXI.Rectangle(0, 0, width, height);
orig = new PIXI.Rectangle(0, 0, width, height);
trim = null; //new PIXI.Rectangle(0, 0, width, height).pad(-1);
texture = new PIXI.Texture(new PIXI.BaseTexture(null, { width, height }), frame, orig, trim);
sprite = new PIXI.Sprite(texture);
sprite.position.set(0.5, 0.5);
sprite.roundPixels = true;
console.log(sprite.getBounds())
```
Without trim: `{x: 1, y: 1, width: 4, height: 4}`
With trim: `{x: 0.5, y: 0.5, width: 4, height: 4}`, but should be the same as without trim.

bigtimebuddy (approved):

bigtimebuddy:
Thank you @dev7355608

Issue #9640, opened 2023-09-07
Bug: Exception when installing a BitmapFont from a XML as a string


D8H:
### Current Behavior

With Pixi 7.2.4, when using `PIXI.BitmapFont.install` with XML as a string. It gives this error:
```
cannot use 'in' operator to search for "getElementsByTag..." in "<?xml version=\"1..."
```

### Expected Behavior

The format should be detected without any exception.

### Steps to Reproduce

This code allows to reproduce the issue:
https://codesandbox.io/s/bitmapfont-from-string-64l8r5

### Environment

- **`pixi.js` version**: *7.2.4*
- **Browser & Version**: *any*
- **OS & Version**: *any*
- **Running Example**: *https://codesandbox.io/s/bitmapfont-from-string-64l8r5/*

### Possible Solution

I applied this change and it seems to work.
```Diff
diff --git a/packages/text-bitmap/src/formats/XMLFormat.ts b/packages/text-bitmap/src/formats/XMLFormat.ts
index b7f0433..2286e17 100644
--- a/packages/text-bitmap/src/formats/XMLFormat.ts
+++ b/packages/text-bitmap/src/formats/XMLFormat.ts
@@ -15,7 +15,8 @@
     {
         const xml = data as Document;
 
-        return 'getElementsByTagName' in xml
+        return typeof data !== 'string'
+            && 'getElementsByTagName' in xml
             && xml.getElementsByTagName('page').length
             && xml.getElementsByTagName('info')[0].getAttribute('face') !== null;
     }
```

PR #9645, opened 2023-09-11
Fix an exception when installing a BitmapFont from an XML as a string

D8H:
When using `PIXI.BitmapFont.install` with XML as a string. It gives this error:
```
cannot use 'in' operator to search for "getElementsByTag..." in "<?xml version=\"1..."
```

This code allows to reproduce the issue:
https://codesandbox.io/s/bitmapfont-from-string-64l8r5

Fixes: https://github.com/pixijs/pixijs/issues/9640

##### Description of change
- Improve type declaration and fix the type error.

##### Pre-Merge Checklist

  - I guess a better type declaration is a better fit to cover this issue than a unit test.
  - Tests don't seem to work on my computer (but `npm run types` is passing)
  image

### DIFF
--- a/packages/text-bitmap/src/formats/TextFormat.ts
+++ b/packages/text-bitmap/src/formats/TextFormat.ts
@@ -54,7 +54,7 @@ export class TextFormat
      * @param data
      * @returns - True if resource could be treated as font data, false otherwise.
      */
-    static test(data: unknown): boolean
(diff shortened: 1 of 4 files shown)


bigtimebuddy (approved):
Nice! Thanks

bigtimebuddy:
Thank you @D8H

PR #9655, opened 2023-09-15
ensure destroy() implemented on all Systems

achamas-playco:
Ensure `destroy()` is implemented for all systems.

### <samp>🤖 Generated by Copilot at fa4041f</samp>

This pull request adds or implements the `destroy` method for various classes that are part of the rendering systems of pixijs. The `destroy` method is used to free up resources and references that are no longer needed by the application. The pull request also imports some type definitions and helpers to enable the destruction operations and fixes a minor issue with the `_currentCursor` property of the `EventSystem` class.

### DIFF
--- a/src/rendering/filters/shared/FilterSystem.ts
+++ b/src/rendering/filters/shared/FilterSystem.ts
@@ -538,8 +538,5 @@ export class FilterSystem implements System
         return mappedMatrix;
     }
 
-    public destroy()
-    {
-        // No critical operations to be called in the destroy method.
-    }
+    public destroy?: () => void;
 }
--- a/src/rendering/filters/shared/FilterSystem.ts
+++ b/src/rendering/filters/shared/FilterSystem.ts
@@ -540,7 +540,6 @@ export class FilterSystem implements System
 
     public destroy()
     {
-        // BOOM!
+        // No critical operations to be called in the destroy method.
     }
 }
-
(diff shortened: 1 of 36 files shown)


src/rendering/filters/shared/FilterSystem.ts
```diff
@@ -540,7 +540,6 @@ export class FilterSystem implements System
 
     public destroy()
     {
-        // BOOM!
+        // No critical operations to be called in the destroy method.
     }
 }
```
Zyie, line 545: If destroy is not needed then we can delete the function and just do `public destroy?: () => void;`

src/rendering/renderers/shared/system/AbstractRenderer.ts
```diff
@@ -304,20 +319,28 @@ export class AbstractRenderer<PIPES, OPTIONS>
...
+        {
+            this._destroySystem(name, options);
+        });
+
+        this._systemsHash = null;
 
+        // destroy all pipes
         writeable.renderPipes = null;
-        writeable.runners = null;
     }
```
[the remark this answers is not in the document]
achamas-playco: Cheers, yes I see that the runners will call the `destroy()` method on the system so good call.

src/rendering/filters/shared/FilterSystem.ts
achamas-playco: Good idea. I applied this to a few other systems with same conditions.

Zyie (approved):

PR #9657, opened 2023-09-15
rename type to renderPipeId

Zyie:
using `type` always confused me as it didn't really mean anything. So i think changing it to `renderPipeId` makes it a bit clearer


GoodBoyDigital (approved):
nice 👍

PR #9658, opened 2023-09-15
add color package

Zyie:
Added the color package back into v8

I have added to the following areas

- container tint
- graphics fill/stroke
- text fill
- ColorMatrixFilter
- RenderTextureSystems for clearColor

Other Changes

- added `ConvertedFillStyle` which should be used in place of `FillStyle` for most things internally
- removed `setFillStyle`/`setStrokeStyle`, we already have getter/setters for these


src/assets/Assets.ts
```diff
@@ -222,12 +217,12 @@ export class AssetsClass
     public resolver: Resolver;
     /**
      * The loader, loads stuff!
-     * @type {PIXI.AssetLoader}
+     * @type {Loader}
```
bigtimebuddy: Do we even need the explicit type here if it's defined below?

src/color/Color.ts
```diff
@@ -0,0 +1,744 @@
+import { colord, extend } from 'colord';
```
achamas-playco, line 1: Would be good to have some tests with this new module since its used all over the place.
Might be something GPT can help with since it's pretty self contained?

achamas-playco (approved):
Looks good. I like the use of the shared Singleton, should help a little with performance in intensive frames.

src/rendering/renderers/gpu/renderTarget/GpuRenderTargetSystem.ts
```diff
@@ -130,7 +129,7 @@ export class GpuRenderTargetSystem implements System
     public getDescriptor(
         renderTarget: RenderTarget,
         clear: CLEAR_OR_BOOL,
-        clearValue: RGBAArray
+        clearValue: ColorSource
```
GoodBoyDigital, line 132: can we keep this as a `RgbaArray` too?

GoodBoyDigital (approved):
lgtm! one tweak to the `getDescriptor` so it accepts an RgbaArray and we are good to go!

PR #9674, opened 2023-09-20
Adds background alias

Zyie:
we have `background` and `backgroundColor` but `background` was not being used


src/rendering/renderers/shared/background/BackgroundSystem.ts
```diff
@@ -95,10 +89,10 @@ export class BackgroundSystem implements System
      */
     public init(options: BackgroundSystemOptions): void
     {
-        options = { ...defaultBackgroundOptions, ...options };
+        options = { ...BackgroundSystem.defaultOptions, ...options };
 
         this.clearBeforeRender = options.clearBeforeRender;
-        this.color = options.backgroundColor || this._backgroundColor; // run bg color setter
```
GoodBoyDigital, line 101: do we need both? should we deprecate `options.background`

src/rendering/renderers/shared/background/BackgroundSystem.ts
```diff
@@ -95,9 +93,14 @@ export class BackgroundSystem implements System
      */
     public init(options: BackgroundSystemOptions): void
     {
-        options = { ...defaultBackgroundOptions, ...options };
+        options = { ...BackgroundSystem.defaultOptions, ...options };
 
         this.clearBeforeRender = options.clearBeforeRender;
+        if (options.background !== undefined)
+        {
+            deprecation(v8_0_0, 'background is deprecated, please use backgroundColor');
```
bigtimebuddy, line 101: I prefer background. I added this because backgroundColor felt so verbose. I'd rather remove that.

GoodBoyDigital (approved):
lgtm!

PR #9683, opened 2023-09-21
use renderer resolution for text by default

Zyie:


bigtimebuddy (commented):
I thought we were going to keep resolution but default to null? And just remove the autoRes bits?

Zyie:
> I thought we were going to keep resolution but default to null? And just remove the autoRes bits?

text resolution does default to null if you don't provide it one?
if none is provided then we use the renderers resolution
if you set `text.resolution` it will use that instead

bigtimebuddy (approved):

PR #9699, opened 2023-09-26
Refactor folder structure

Zyie:
refactors folder structure to reduce the size of the `rendering` folder which became a catch all for everything

I've added a `scene` folder that contains all of the containers, sprites, text etc

Tidied up the structure of folders to always follow `gl`,`gpu`,`shared`, or be flat if everything is `shared`


achamas-playco (approved):

GoodBoyDigital (approved):
lgtm!

PR #9702, opened 2023-09-26
add borders to texture

Zyie:
adds borders property to texture for use with nine slice textures

### DIFF
--- a/src/scene/mesh-extras/NineSliceSprite.ts
+++ b/src/scene/mesh-extras/NineSliceSprite.ts
@@ -70,19 +70,18 @@ export class NineSliceSprite extends Container<MeshView<NineSliceGeometry>>
             options = { texture: options };
         }
 
-        options = { ...NineSliceSprite.defaultOptions, ...options };
-
-        const texture = options.texture;
+        const texture = options.texture ?? NineSliceSprite.defaultOptions.texture;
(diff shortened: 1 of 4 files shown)


src/scene/mesh-extras/NineSliceSprite.ts
```diff
@@ -70,19 +70,18 @@ export class NineSliceSprite extends Container<MeshView<NineSliceGeometry>>
...
         const nineSliceGeometry = new NineSliceGeometry({
             width: texture.width,
             height: texture.height,
             originalWidth: texture.width,
             originalHeight: texture.height,
-            leftWidth: options.leftWidth,
-            topHeight: options.topHeight,
-            rightWidth: options.rightWidth,
-            bottomHeight: options.bottomHeight,
+            leftWidth: options.leftWidth ?? borders?.left ?? NineSliceSprite.defaultOptions.leftWidth,
```
GoodBoyDigital, line 81: do we need the additional `NineSliceSprite.defaultOptions.leftWidth`? 

This get set by the NineSliceGeometry constructor if nothing is passed?

Zyie: We need them if we want to be able to change the default settings of the sprite separately from the geometry, which is currently the case

but that seems weird to me

PR #9709, opened 2023-09-27
add some descender padding

GoodBoyDigital:
fix for cropping on HTML-Text.
This is also present in v7


src/scene/text/html/utils/measureHtmlText..ts
```diff
@@ -41,8 +42,11 @@ export function measureHtmlText(
 
     svgRoot.remove();
 
+    // we probably can find a better way to do this,
```
GoodBoyDigital, line 45:
```suggestion
```

Zyie (approved):

PR #9715, opened 2023-09-28
Fix: loadWebFont: Don't encodeURI if already valid

### LINKED ISSUE #9705: Bug: Support for loading font assets with space breaks file:// protocol usage
### Current Behavior

https://github.com/pixijs/pixijs/commit/06a9562990ea3b0c97a461e61f7f3e54696edb29

This code does not work if file is being loaded using file:// protocol with url already being encoded.
It double encodes making '%20' into '%2520'

### Expected Behavior
Do not reencode already encoded URLs

### Steps to Reproduce
Try loading a file using file:// protocol (in electron app ie.) 
URL you'll get can potentially be:
`’some folder/file.woff'`
When using Assets.load(url) before the url variable reaches the loader it would get transformed to: `'file:///c:/some%20folder/file.woff'`

Using URIencode will now further transform it into:
`'file:///c:/some%2520folder/file.woff'`

Happens locally but I believe if http url given contained spaces it would also occur.

### Environment

- **`pixi.js` version**: *7.1.0*
- **Browser & Version**: * Chrome 108*
- **OS & Version**: *any*

### Possible Solution

Make sure that urls that start with file:// are not reencoded
Or probably better perform encoding only in a single shared płace (possible candidate would be a function that puts together a global path in Assets class)

### Additional Information

_No response_

SuperSodaSea:
##### Description of change

In `loadWebFont`, if the URI is already valid, do not call `encodeURI` to avoid it being accidentally encoded twice.

Fixes #9705.

##### Pre-Merge Checklist

### DIFF
--- a/packages/assets/src/loader/parsers/loadWebFont.ts
+++ b/packages/assets/src/loader/parsers/loadWebFont.ts
@@ -78,6 +78,23 @@ export function getFontFamilyName(url: string): string
     return fontFamilyName;
 }
 
+// See RFC 3986 Chapter 2. Characters
+const validURICharactersRegex = /^[0-9A-Za-z%:/?#\[\]@!\$&'()\*\+,;=\-._~]*$/;
+
+/**
+ * Encode URI only when it contains invalid characters.
+ * @param uri - URI to encode.
+ */
+function encodeURIWhenNeeded(uri: string)
+{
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):

bigtimebuddy:
Thanks @SuperSodaSea

PR #9718, opened 2023-09-29
Change element to canvas

Zyie:
update the app api to be consistent. Now `canvas` is used everywhere instead of the mix between canvas, view, and element


bigtimebuddy:
Nice!

GoodBoyDigital (approved):
n i c e

PR #9722, opened 2023-09-29
Fix passive interaction

Zyie:
removed the `passive` check in `testHitFn` as it is incorrect

we don't care if it is passive or not if the parent is interactive and we already do this check in the recursive functions

this is also an issue in v7, so will look to back port this


bigtimebuddy:
Backport welcome

PR #9737, opened 2023-10-05
Remove settings object

Zyie:
the last remaining settings have been moved to the following places:

- `settings.RESOLUTION` -> `AbstractRenderer.RESOLUTION`
- `setting.FAIL_IF_MAJOR_PERFORMANCE_CAVEAT` -> `AbstractRenderer.FAIL_IF_MAJOR_PERFORMANCE_CAVEAT`
- `settings.ADAPTER` -> `DOMAdapter.get()`/`DOMAdapter.set()`


GoodBoyDigital:
hey @Zyie , this is cool - glad setting are gone!

I think we should have a static defaultOptions property as we do in many other places on AbstractRenderer

```
static defaultOptions = {
    resolution:1,
    failIfMajorPerformanceCaveat:false,
    roundPixels:false
}
```

that way we can set thing 100% with the `renderer.init` too:

```
renderer.init({
    resolution:1,
    failIfMajorPerformanceCaveat:true
    ....
})

```

failIfMajorPerformanceCaveat might need to be passed in to the canUseWebGL function to keep things clean?

GoodBoyDigital (approved):
nice! 👍

PR #9739, opened 2023-10-05
Fix graphics not apply alpha from Color fill

Zyie:


GoodBoyDigital (approved):
nice! and the test 👍 👍

PR #9757, opened 2023-10-13
Add support for different environments

Zyie:


GoodBoyDigital (approved):
looks good! couple of minor changes I think would round this off!

1 - For clarity and consistency, could we update the adapter file names to match the classes they contain? For instance, renaming `webworker.ts` to `WebWorkerAdaptor.ts`.

2 - The function named 'test' within the environment detection seems a bit ambiguous. It doesn't just test but also loads certain code segments. To improve clarity:

Consider splitting the function into two distinct functions, namely `test` and `init` (or alternatively, `load`).

As an alternative approach, we could rename the existing function to something more descriptive, like `testAndLoad`. I definitely lean towards the first option though!

src/rendering/renderers/autoDetectRenderer.ts
```diff
@@ -42,7 +42,7 @@ export async function autoDetectRenderer(options: Partial<AutoDetectOptions>): P
     let RendererClass: new () => Renderer;
 
     await autoDetectEnvironment(
-        options.manageImports ?? false,
+        options.manageImports ?? true,
```
GoodBoyDigital, line 45: is this a correct change?

Zyie: it is

i realised i changed it to false by accident in a previous commit

GoodBoyDigital (approved):

PR #9769, opened 2023-10-19
v8 compressed textures dds support

GoodBoyDigital:
Support for DDS textures in v8

some tiny tweaks to KTX


Zyie (approved):

src/compressed-textures/dds/parseDDS.ts
```diff
@@ -0,0 +1,175 @@
...
+const offPfFlags = 20;
+const offPfFourCC = 21;
+const offRGBBitCount = 22;
+const offRBitMask = 23;
+const offGBitMask = 24;
+const offBBitMask = 25;
+
+export function parseDDS(buffer: ArrayBuffer, supportedFormats: TEXTURE_FORMATS[]): TextureSourceOptions<Uint8Array[]>
+{
+    const header = new Int32Array(buffer, 0, headerLengthInt);
```
achamas-playco, line 59: Just wondering. If this function throws, does the caller have a strategy? How would a user safeguard loading faulty textures at runtime? Can they catch this with `Asset.load`?

achamas-playco (approved):

src/compressed-textures/ktx/loadKTX.ts
```diff
@@ -19,7 +19,7 @@ export const loadKTX = {
         priority: LoaderParserPriority.High,
     },
 
-    name: 'loadKTX2',
+    name: 'loadKTX',
```
achamas-playco, line 22: Other code still seems to be referencing `ktx2*`. Is the idea to just use the latest version for this identifier?

achamas-playco (approved):

src/compressed-textures/dds/parseDDS.ts
GoodBoyDigital: yoo! yes, this error should get caught by the Loader

src/compressed-textures/ktx/loadKTX.ts
GoodBoyDigital: will do a pass to make sure every thing in code named KTX rather than KTX2 as both are supported

PR #9784, opened 2023-10-26
Fix: adapt KTXWorker messageHandlers

YunYouJun:
##### Description of change

KTXWorker `onmessage` listen is global, some tools like vite also send message `type: 'messsage'`.
We should be compatible for it.

image

image
image

messageHandlers[message] will be undefined.

image

##### Pre-Merge Checklist


GoodBoyDigital (approved):
great catch! 👍

PR #9787, opened 2023-10-26
Don't set `sortDirty` to true if the `zIndex` doesn't change

dev7355608:
##### Description of change

We shouldn't set `sortDiry` of the parent container to true if the `zIndex` is not changed by the assignment.

##### Pre-Merge Checklist


aaclayton:
Hey folks. Given the limited scope of this change is it possible it would make a V7 release? Or will this only appear in V8+?

bigtimebuddy (approved):

PR #9788, opened 2023-10-26
document accessibility, app, and assets

Zyie:
this pr tidies up all of the documentation for these packages

I'm trying out a new structure for the docs where classes/interface/method etc are grouped by type e.g `assets`/`app`/`accessibility` and looks like this:

image
image

I would look to modify the doc generation to also add `classes` to the sidebar but will do this once all the docs are complete

Would be good to get any opinions on this structure


bigtimebuddy:
Do you have a link we can browse the docs? I don't see anything here:
https://pixijs.download/docs/aaa/docs/index.html

In general, I think sorting by packages is helpful especially for folks that only want to consume specific exports. I also think having an ungrouped alphabetical list of classes/interfaces/etc in the docs is also useful for folk unfamiliar with the project structure. I don't see that alphabetical list there in your screenshots?

Zyie:
> Do you have a link we can browse the docs? I don't see anything here: https://pixijs.download/docs/aaa/docs/index.html

looks like docs were not outputted to the correct folder for upload, fixed now

> I also think having an ungrouped alphabetical list of classes/interfaces/etc in the docs is also useful for folk unfamiliar with the project structure

there is still the list of classes in the sidebar

One other thing to note is that where ever you see something like `assets.Class` i will be updating the doc generation template to remove the namespace

bigtimebuddy:
How do we deal with namespaces? I'm thinking about `utils`. This is more like a namespace within `@pixi/core`, we don't want to document it as a package. Right?

GoodBoyDigital (approved):
lgtm 👍

PR #9804, opened 2023-10-30
docs for events, filters, and extensions

Zyie:


src/rendering/renderers/shared/system/SystemRunner.ts
```diff
@@ -1,14 +1,14 @@
 /**
- * A Runner is a highly performant and simple alternative to signals. Best used in situations
+ * A SystemRunner is a highly performant and simple alternative to signals. Best used in situations
  * where events are dispatched to many objects at high frequency (say every frame!)
```
GoodBoyDigital, line 3: how about we also add that the systemsRunner is for internal use.

`SystemRunner` is used internally by the renderers as an efficient way for systems to be notified about what the renderer is up to during the rendering phase.`

GoodBoyDigital (approved):
looks good! just a few minor comments!

PR #9813, opened 2023-10-31
docs for maths classes

Zyie:


src/maths/matrix/Matrix.ts
```diff
@@ -509,7 +510,7 @@ export class Matrix
     }
 
     /**
-     * A temp matrix
+     * A static Matrix that can be used to avoid creating new objects.
```
GoodBoyDigital, line 513: `Will always ensure the matrix is reset to identity when requested.`

GoodBoyDigital (approved):
nice!

PR #9818, opened 2023-10-31
v8 add `setDataWithSize` to `Buffer`

GoodBoyDigital:
add a function that lets devs update the data whilst allso setting the size. 

Also added new property, `shrinkToFit` which if false will not shrink the GPU buffer if the data becomes smaller.

added tests to ensure it works like a boss

### DIFF
--- a/src/rendering/renderers/shared/buffer/Buffer.ts
+++ b/src/rendering/renderers/shared/buffer/Buffer.ts
@@ -194,7 +194,7 @@ export class Buffer extends EventEmitter<{
 
     set data(value: TypedArray)
     {
-        this.setData(value, value.length);
+        this.setDataWithSize(value, value.length);
     }
 
     /**
@@ -203,15 +203,12 @@ export class Buffer extends EventEmitter<{
      * @param value - the data to set
      * @param size - the size of the data in bytes
      */
-    public setData(value: TypedArray, size?: number)
+    public setDataWithSize(value: TypedArray, size: number)
     {
         // Increment update ID
         this._updateID++;
 
-        // Calculate size in bytes
-        const sizeInBytes = size ? (size * value.BYTES_PER_ELEMENT) : value.byteLength;
-
-        this._updateSize = sizeInBytes;
+        this._updateSize = (size * value.BYTES_PER_ELEMENT);
 
         // If the data hasn't changed, early return after emitting 'update'
         if (this._data === value)
--- a/src/rendering/renderers/shared/buffer/Buffer.ts
+++ b/src/rendering/renderers/shared/buffer/Buffer.ts
@@ -22,6 +22,15 @@ export interface BufferOptions
     usage: number;
     /** a label for the buffer, this is useful for debugging */
     label?: string;
+    /**
+     * should the GPU buffer bew shrunk when the data becomes smaller?
+     * changing this will cause the buffer to be destroyed and a new one created on the GPU
+     * this can be expensive, especially if the buffer is already big enough!
+     * setting ts to false will prevent the buffer from being shrunk. This will yield better performance
+     * if you are constantly setting data that is changing size often.
+     * @default true
+     */
+    shrinkToFit?: boolean;
 }
 
 export interface BufferDescriptor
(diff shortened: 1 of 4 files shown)


src/rendering/renderers/shared/buffer/Buffer.ts
```diff
@@ -173,24 +194,56 @@ export class Buffer extends EventEmitter<{
...
+        this.setData(value, value.length);
+    }
+
+    /**
+     * Sets the data in the buffer to the given value. This will immediately update the buffer on the GPU.
+     * If you only want to update a subset of the buffer, you can pass in the size of the data.
+     * @param value - the data to set
+     * @param size - the size of the data in bytes
+     */
+    public setData(value: TypedArray, size?: number)
```
Zyie, line 206: Seems weird to have 2 api's for the same thing but with an optional arg

```js
buffer.data = typedArray
buffer.setData(typedArray)
```

we should either remove the setter of `buffer.data`, or change the name of `setData` to something like `setDataWithSize()` and not make the `size` arg optional

GoodBoyDigital: maybe a different name might be better, this function does two things
- sets the data
- sets the size of the data to use.

eg:
```buffer.setData(typedArray, 2)```

File: src/rendering/renderers/shared/buffer/Buffer.ts [hunk 2]
```diff
@@ -22,6 +22,15 @@ export interface BufferOptions
     usage: number;
     /** a label for the buffer, this is useful for debugging */
     label?: string;
+    /**
+     * should the GPU buffer bew shrunk when the data becomes smaller?
+     * changing this will cause the buffer to be destroyed and a new one created on the GPU
+     * this can be expensive, especially if the buffer is already big enough!
+     * setting ts to false will prevent the buffer from being shrunk. This will yield better performance
```
achamas-playco, line 29:
```suggestion
     * setting this to false will prevent the buffer from being shrunk. This will yield better performance
```

src/rendering/renderers/shared/buffer/Buffer.ts
```diff
@@ -135,14 +144,24 @@ export class Buffer extends EventEmitter<{
 
     private _data: TypedArray;
 
+    /**
+     * should the GPU buffer bew shrunk when the data becomes smaller?
```
achamas-playco, line 148:
```suggestion
     * should the GPU buffer be shrunk when the data becomes smaller?
```

src/rendering/renderers/shared/buffer/Buffer.ts
```diff
@@ -135,14 +144,24 @@ export class Buffer extends EventEmitter<{
...
+     * changing this will cause the buffer to be destroyed and a new one created on the GPU
+     * this can be expensive, especially if the buffer is already big enough!
+     * setting ts to false will prevent the buffer from being shrunk. This will yield better performance
```
achamas-playco, line 151:
```suggestion
     * setting this to false will prevent the buffer from being shrunk. This will yield better performance
```

src/rendering/renderers/shared/buffer/Buffer.ts
```diff
@@ -22,6 +22,15 @@ export interface BufferOptions
     usage: number;
     /** a label for the buffer, this is useful for debugging */
     label?: string;
+    /**
+     * should the GPU buffer bew shrunk when the data becomes smaller?
```
achamas-playco, line 26:
```suggestion
     * should the GPU buffer be shrunk when the data becomes smaller?
```

src/rendering/renderers/shared/buffer/Buffer.ts
achamas-playco: I think Sean's suggestion of renaming the function to `setDataWithSize` and making all arguments mandatory might be better, there is a slight chance this may confuse people.

The renamed and mandatory version has a much clearer purpose, with no ambiguity as to which api to use if you just had the data alone.

PR #9823, opened 2023-11-01
docs for utils

Zyie:


src/utils/logging/logDebugTexture.ts
```diff
@@ -1,6 +1,14 @@
 import type { Texture } from '../../rendering/renderers/shared/texture/Texture';
 import type { Renderer } from '../../rendering/renderers/types';
 
+/**
+ * Logs a texture to the console as a base64 image.
+ * This can be very useful for debugging issues with rendering.
+ * @param texture - The texture to log
+ * @param renderer - The renderer to use
+ * @param size - The size of the texture to log in the console
+ * @memberof utils
```
GoodBoyDigital, line 10: we can ignore this function as it actually is used when I need to log a texture mid render (which no one should be doing). 

Users can use `renderer.extract.log(thing)` for most use cases

src/utils/misc/Transform.ts
```diff
@@ -1,20 +1,30 @@
...
 {
+    /** The matrix to use. */
     matrix?: Matrix;
+    /** The observer to use. */
     observer?: {onUpdate: (transform: Transform) => void}
 }
 
-/** Transform that takes care about its versions. */
+/**
+ * Transform that takes care about its versions.
```
GoodBoyDigital, line 19:
```suggestion
 * The Transform class facilitates the manipulation of a 2D transformation matrix through 
 * user-friendly properties: position, scale, rotation, skew, and pivot. 
```

src/utils/misc/Transform.ts
```diff
@@ -60,6 +70,11 @@ export class Transform
     protected dirty = true;
     protected observer: Observer<Transform>;
 
+    /**
+     * @param param0 - Options for the transform.
```
GoodBoyDigital, line 74: lets rename this to `options`

src/utils/misc/Transform.ts
```diff
@@ -77,6 +92,10 @@ export class Transform
         this._sy = 1;
     }
 
+    /**
+     * The local matrix.
```
GoodBoyDigital, line 96:
```suggestion
     * This matrix is computed by combining this Transforms position, scale, rotation, skew, and pivot
     * properties into a single matrix.
```

GoodBoyDigital (commented):
left a few comments - nothing major! If you can tweak, then we are good to merge!

PR #9833, opened 2023-11-02
v8 docs for instructions

GoodBoyDigital:
- rename type to renderPipeId

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/instructions/InstructionSet.ts
+++ b/src/rendering/renderers/shared/instructions/InstructionSet.ts
@@ -2,32 +2,47 @@ import { uid } from '../../../../utils/data/uid';
 
 import type { Instruction } from './Instruction';
 
+/**
+ * A set of instructions that can be executed by the renderer.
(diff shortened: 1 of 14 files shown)


src/rendering/renderers/shared/instructions/InstructionSet.ts
```diff
@@ -2,32 +2,47 @@ import { uid } from '../../../../utils/data/uid';
...
+     * Add an instruction to the set
+     * @param instruction - add an instruction to the set
+     */
     public add(instruction: Instruction)
     {
         this.instructions[this.instructionSize++] = instruction;
     }
 
+    /** Log the instructions to the console (for debugging) */
     public log()
```
Zyie, line 42: does this need to be added to the api docs?

seems like we should keep this as an internal API

GoodBoyDigital: im cool with that. let me. update

achamas-playco (approved):

PR #9839, opened 2023-11-03
v8 fix background alpha bug

### LINKED ISSUE #9816: Bug: [v8] Background alpha doesn't have an effect on WebGPU
### Current Behavior

Setting the background alpha option when initializing the app doesn't change the alpha of the background.

### Expected Behavior

Alpha should blend or completely remove the background based on value.

### Steps to Reproduce

const app = new Application()
await app.init({
backgroundAlpha: 0
})

### Environment

- **`pixi.js` version**:  v8 Beta 7

### Possible Solution

_No response_

### Additional Information

_No response_

GoodBoyDigital:
fixes background alpha for renderers
fixes #9816

### DIFF
--- a/src/rendering/renderers/gpu/renderTarget/GpuRenderTargetSystem.ts
+++ b/src/rendering/renderers/gpu/renderTarget/GpuRenderTargetSystem.ts
@@ -376,8 +376,8 @@ export class GpuRenderTargetSystem implements System
                             | GPUTextureUsage.COPY_DST
                             | GPUTextureUsage.RENDER_ATTACHMENT
                             | GPUTextureUsage.COPY_SRC,
-                        format: 'bgra8unorm',
-                        alphaMode: 'opaque',
+                        format: navigator.gpu.getPreferredCanvasFormat(),
+                        alphaMode: 'premultiplied',
                     });
                 }
                 catch (e)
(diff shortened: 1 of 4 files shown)


src/rendering/renderers/gpu/renderTarget/GpuRenderTargetSystem.ts
```diff
@@ -376,8 +376,8 @@ export class GpuRenderTargetSystem implements System
                             | GPUTextureUsage.COPY_DST
                             | GPUTextureUsage.RENDER_ATTACHMENT
                             | GPUTextureUsage.COPY_SRC,
-                        format: 'bgra8unorm',
-                        alphaMode: 'opaque',
+                        format: navigator.gpu.getPreferredCanvasFormat(),
```
achamas-playco, line 379: We might want to guard against this property not being available, seems like it will crash in Safari currently:
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/gpu#browser_compatibility

GoodBoyDigital: This should be ok as to have got this far in the code as it means the `navigator.gpu` has to exist as its required to create the WebGPU device.

achamas-playco (approved):

PR #9860, opened 2023-11-13
add display tests

achamas-playco:
> NOTE: 📣 This PR has a skipped test with a corresponding follow up ticket, its ready for review and merge.

### Summary
🧪🌐📦

This pull request adds several unit tests for the DisplayObject and Container classes, which are core components of the pixijs library. The tests cover various methods and properties that deal with coordinate transformations, parent-child relationships, and rendering options. The tests also reveal some compatibility issues with the original PIXI.js library that need to be resolved.

> _Testing `DisplayObject`_
> _Methods to transform coordinates_
> _Autumn leaves falling_

### Walkthrough
*  Add unit tests for the `Container` class, a subclass of `DisplayObject` that renders a group of display objects ([link](https://github.com/pixijs/pixijs/pull/9860/files?diff=unified&w=0#diff-17f94aefd5e91e0a9473652d769f645c5d1ba9e7fcdb65d77721bb80e4ba7d7eR1-R359))
* Add unit tests for the `toGlobal` method of the `DisplayObject` class, which converts a point from local to global coordinates ([link](https://github.com/pixijs/pixijs/pull/9860/files?diff=unified&w=0#diff-de88ebb7fdb8e0a5c31643a2b281b0291ca3e756dcaab966c8cfaba85109be9eR1-R32))
* Add unit tests for the `toLocal` method of the `DisplayObject` class, which converts a point from global to local coordinates ([link](https://github.com/pixijs/pixijs/pull/9860/files?diff=unified&w=0#diff-3e8e94210ac251c92bff397a7318b511550d17e0f44201a15469654c2753f321R1-R56))


tests/display/DisplayObject.tests.ts
```diff
@@ -0,0 +1,311 @@
+import { DEG_TO_RAD, RAD_TO_DEG } from '../../src/maths/misc/const';
+import { Container } from '../../src/scene/container/Container';
+import { updateLayerGroupTransforms } from '../../src/scene/container/utils/updateLayerGroupTransforms';
+
+describe('Container', () =>
```
GoodBoyDigital, line 5: Lets rename this File to something else - DisplayObject is not a thing. 
As we could end up with one uber file - perhaps we break container tests up in to smaller groups?

tests/display/Container.tests.ts
```diff
@@ -0,0 +1,937 @@
...
+import { Container } from '../../src/scene/container/Container';
+// import { Graphics } from '../../src/scene/graphics/shared/Graphics';
+import { getRenderer } from '../utils/getRenderer';
+
+import type { WebGLRenderer } from '../../src/rendering/renderers/gl/WebGLRenderer';
+
+// function testAddChild(fn: any)
+// {
+//     return () =>
+//     {
```
Zyie, line 12: any reason for keeping around all this commented code?

tests/display/DisplayObject.tests.ts
achamas-playco: Sorted.

tests/display/Container.tests.ts
achamas-playco: Not at all, thanks for spotting!

Zyie (approved):

tests/display/Container.Visual.tests.ts
```diff
@@ -0,0 +1,177 @@
...
+            rootContainer.addChild(parent);
+            parent.addChild(child);
+
+            parent.alpha = 0.5;
+
+            updateLayerGroupTransforms(rootContainer.layerGroup, true);
+
+            const alpha = ((child.layerColor >> 24) & 0xFF) / 255;
+
+            expect(alpha).toBeCloseTo(0.5, 2);
```
Zyie, line 102: we no longer need to do this

we can just directly test the alpha

tests/display/Container.tests.ts
```diff
@@ -0,0 +1,102 @@
...
+describe('Container', () =>
+{
+    describe('constructor', () =>
+    {
+        it('should initialise properties', () =>
+        {
+            const object = new Container();
+
+            expect(object.alpha).toEqual(1);
+            expect(object.layerColor).toEqual(0xffffffff);
```
Zyie, line 12: layerColor is also now just a normal hex value without alpha

Zyie (changes requested):

tests/display/Container.Visual.tests.ts
achamas-playco: I simplified the test to check the `child.layerAlpha` since its really just testing that parent alpha values are cascaded.

tests/visual/snapshots/webgl-container-scene-ts.png
Zyie, line 1: tests/visual/snapshots/webgl-container-scene-ts.png
tests/visual/snapshots/webgl-container-visibility-scene-ts.png

there are two images here, but only 1 test

Zyie (changes requested):

PR #9872, opened 2023-11-16
Refactor/v8/render-target-system

GoodBoyDigital:
Fixed up the Render Target system

- Allows for frames to be used when binding a render target.
- By default if the render surface passed in is a Texture, it will use that textures frame.
- Created a base `RenderTargetSystem` and a couple of adpators to help it work both with WebGL and WebGPU. (there was a lot of shared code!)
- Fixed up render targets to have there sizes driven by the underlying color texture.


src/rendering/renderers/gl/WebGLRenderer.ts
```diff
@@ -25,8 +25,8 @@ import type { ExtractRendererOptions, ExtractSystemTypes } from '../shared/syste
 import type { GlRenderingContext } from './context/GlRenderingContext';
 
 const DefaultWebGLSystems = [
-    ...SharedSystems,
     GlBackBufferSystem,
+    ...SharedSystems,
```
Zyie, line 29: is there a reason for moving shared systems after back buffer system?

src/rendering/renderers/gpu/WebGPURenderer.ts
```diff
@@ -24,12 +24,14 @@ import type { SystemConstructor } from '../shared/system/System';
 import type { ExtractRendererOptions, ExtractSystemTypes } from '../shared/system/utils/typeUtils';
 
 const DefaultWebGPUSystems = [
+    // we want renderStart to be called first..
+    // before the renderTarget system renderStart..
+    GpuEncoderSystem,
```
Zyie, line 29: we should use `priority` on the extension to manage this?

Zyie (approved):

src/rendering/renderers/gpu/WebGPURenderer.ts
GoodBoyDigital: yes - follow up ticket [here](https://github.com/orgs/pixijs/projects/2/views/4?pane=issue&itemId=44862617)

PR #9887, opened 2023-11-22
add linting for jest tests

Zyie:
Added the `eslint-plugin-jest` with its recommended settings

This should help us avoid committing things like `it.only`

I've also added a `throw` for when the visual tests are pushed to CI with `only: true`.


bigtimebuddy (approved):
Nice!

PR #9890, opened 2023-11-23
Bugfix/v8/hole bounds check

ErikSom:
##### Description of change
Added a check in GraphicsContext to see if a hole is actually inside geometry

** updated the bug playground with newer edge case that this code fixes **
This fixes [this bug](https://www.pixiplayground.com/#/edit/RInLPNH9iVIn8Nha4aJpu)


GoodBoyDigital:
Cleaning up older PRs - Graphics holes should be all good now! Thanks again for this @ErikSom

PR #9892, opened 2023-11-24
enable adaptive bezier smoothing control

achamas-playco:
Provide app option `bezierSmoothness` which sets the `GraphicsContextSystem.defaultOptions.bezierSmoothness` default value. Updated all bezier and quadratic signatures to accept a `smoothness?: number` value, or else the default is used.

This allows the user to either the smoothing once at startup, or per drawing call.

> NOTE: Once shapes are built with current smoothness settings they cannot be changed, but changes to the `GraphicsContextSystem.defaultOptions.bezierSmoothness` will affect future path draw calls.

Here is the result of a smoothing value of `0` (current `next-v8` smoothing level):
image

Here's the current default smoothing of `0.5`:
image

Here's the maximum smoothing possible at `1.0`:
image

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 19ed483</samp>

This pull request adds a new feature to the graphics module that allows users to adjust the smoothness of bezier and quadratic curves in the graphics objects. It introduces a `smoothness` parameter to various classes and methods that draw curves and rounded shapes, and uses an adaptive algorithm to render the curves with different quality and performance levels. It also adds a new test file `Graphics.AdaptiveBezierSmoothing.tests.ts` that contains unit tests for the feature.


tests/accessibility/AccessibilitySystem.tests.ts
achamas-playco, line 1: Attempting to unblock failing tests

src/scene/graphics/shared/path/GraphicsPath.ts
```diff
@@ -363,6 +367,13 @@ export class GraphicsPath
         return this;
     }
 
+    public invalidate()
```
GoodBoyDigital, line 370: yoo, I think we can delete this function?

GoodBoyDigital (approved):
nice work!!

src/scene/graphics/shared/path/GraphicsPath.ts
achamas-playco: Absolutely!

Zyie:
@achamas-playco can we add a visual test that draws quadratic and bezier at different smoothness levels?

Zyie (changes requested):
Just requesting changes so i don't accidently merge this before you fix up the quadratic curve stuff

achamas-playco:
@GoodBoyDigital @Zyie fixed up the visual tests, there wasn't a problem with the quadratic curves, it was a problem with the control points drawing the original shapes in the test. All good now!

Zyie (approved):

GoodBoyDigital (approved):
nice work!

PR #9893, opened 2023-11-24
fix cache warning when using cache parser

Zyie:


GoodBoyDigital (approved):

src/assets/Assets.ts
```diff
@@ -685,8 +685,7 @@ export class AssetsClass
         onProgress?: ProgressCallback
     ): Promise<Record<string, T>>
     {
-        const resolveArray = Object.values(resolveResults) as ResolvedAsset[];
-        const resolveKeys = Object.keys(resolveResults);
+        const resolveArray = [...new Set(Object.values(resolveResults))] as ResolvedAsset[];
```
GoodBoyDigital, line 688: can you explain to a 5 year old (me) what does this line doing? is it removing duplicates?

Zyie: yeah it removes duplicates using a `Set` then puts it back into an array

PR #9896, opened 2023-11-24
Fix type error in `ISpritesheetData`

Nearoo:
##### Description of change

Removes the `?` on the `animations` key of the `ISpritesheetData` type. This is because it results in the type `Record<never, ...>` for the `Spritesheet` class member `animations` [here](https://github.com/Nearoo/pixijs/blob/503605c14078bd66f8a7d0a14bfba2b47b05184b/packages/spritesheet/src/Spritesheet.ts#L180), making it impossible to use `animations` without type error in typescript.

The reason for this is that `animation` of `ISpritesheetData` is of type `Record<keyof S['animations'], Texture[]>`, where, due to the `?`, `keyof S['animations']` will evaluate to `undefined & utils.Dict<string[]>`, which is `never`. 

Since `Spritesheet.animations` is always initialized ([here](https://github.com/Nearoo/pixijs/blob/503605c14078bd66f8a7d0a14bfba2b47b05184b/packages/spritesheet/src/Spritesheet.ts#L227)), it shouldn't cause an issue. However, I don't have a broad overview over the codebase, someone might want to verify that.

##### Pre-Merge Checklist

I didn't add any tests because the error is on a type level, and the change is literally one character.

### DIFF
--- a/packages/spritesheet/src/Spritesheet.ts
+++ b/packages/spritesheet/src/Spritesheet.ts
@@ -224,7 +224,7 @@ export class Spritesheet<S extends ISpritesheetData = ISpritesheetData>
         this._texture = texture instanceof Texture ? texture : null;
         this.baseTexture = texture instanceof BaseTexture ? texture : this._texture.baseTexture;
         this.textures = {} as Record<keyof S['frames'], Texture>;
-        this.animations = {} as Record<keyof S['animations'], Texture[]>;
+        this.animations = {} as Record<keyof NonNullable<S['animations']>, Texture[]>;
         this.data = data;
 
         const resource = this.baseTexture.resource as ImageResource;
--- a/packages/spritesheet/src/Spritesheet.ts
+++ b/packages/spritesheet/src/Spritesheet.ts
@@ -36,7 +36,7 @@ export interface ISpritesheetFrameData
  */
 export interface ISpritesheetData
 {
-    animations: utils.Dict<string[]>;
+    animations?: utils.Dict<string[]>;
     frames: utils.Dict<ISpritesheetFrameData>;
     meta: {
         app?: string;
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
What version of typescript? Do have a reproduction of the type error?

What about using `| undefined` instead? That animations field is optional and you're making it required, which is not correct.

Nearoo:
* typescript version: `5.3.2`

Here's a reproducible example: https://gist.github.com/Nearoo/83c5190c075a12c3e2e8464d43eb5763

It will result in:
```
src/main.ts:32:15 - error TS7053: Element implicitly has an 'any' type because expression of type '"bar"' can't be used to index type 'Record<never, Texture<Resource>[]>'.
  Property 'bar' does not exist on type 'Record<never, Texture<Resource>[]>'.

32 const anims = s.animations['bar']
                 ~~~~~~~~~~~~~~~~~~~
```

I don't think `| undefined` would solve the issue, the type will still be `never`. The issue is that typescript takes the intersection of types for indexed access types, which wasn't always the case. The change was made [here]( https://github.com/microsoft/TypeScript/pull/30769).

I understand your concern though. Maybe a better approach would be to change the type of the `animations` field in `Spritesheet` [here](https://github.com/pixijs/pixijs/blob/503605c14078bd66f8a7d0a14bfba2b47b05184b/packages/spritesheet/src/Spritesheet.ts#L180) to this?

```
animations: Record<keyof NonNullable<S['animations']>, Texture[]>;
```

bigtimebuddy:
Yes, the problem is on the Spritesheet class not the data interface.

Nearoo:
Alright, update my branch. Would be cool if you could merge, we have a lot of `//@ts-ignore` over the codebase :)

bigtimebuddy (approved):

PR #9914, opened 2023-11-30
Chore/v8/culling

GoodBoyDigital:
A new approach at culling that removes the logic from the rendering loop with the idea to give users more control on when culling happens and how frequently

```ts
const container = new GameWorld();
const view = new Rectangle(0, 0, 800, 600);

// whether or not the container can be culled
container.cullable = true;
// a cull area that will be used instead of the bounds of the container
container.cullArea = new Rectangle(0,0,400,400);
// whether or not the containers children can be culled. This can help optimise large scenes
container.cullableChildren = false;

Culler.shared.cull(myContainer, view);
renderer.render(myContainer);
````

There is also a `CullerPlugin` that can be used to automatically call `Culler.shared.cull` every frame if you want the convenience 

```ts
import {extensions, CullerPlugin} from 'pixi.js'
extensions.add(CullerPlugin)
```


src/rendering/renderers/shared/texture/Texture.ts
```diff
@@ -283,7 +378,17 @@ export class Texture extends EventEmitter<{
         return this._source;
     }
 
+    /**
+     * an Empty Texture used internally by the engine
+     * @ignore
+     * @internal
```
Zyie, line 384:
```suggestion
```

Zyie:
I've pushed a change to this PR to update the API a little. I realised that we are trying to solve a very similar problem with the new interaction manager.

@bigtimebuddy it would be good to get your thoughts on this approach!

### API

#### Basic usage
```ts
// ... app setup

const sprite = new Sprite();
sprite.cull = true;
``` 
This will be automatically culled before rendering `app.scene` using `renderer.screen`

#### Custom culler
```ts
// ... app setup

const culler = new Culler();
const sprite = new Sprite();
sprite.culler = culler;
sprite.cull = true;  // now will automatically be added to the custom culler

// call the culler manually
culler.cull(new Rectangle())

renderer.render(scene)
```

### Todo

- add `cullArea` for optimisation
- when culling it can waste time culling a child of a parent that is already been culled

PR #9918, opened 2023-12-01
fix couple of Text bugs

GoodBoyDigital:
word wrap now working
and size triggers and update

also fixed a line height bug in bitmap text


src/scene/text/bitmap/DynamicBitmapFont.ts
```diff
@@ -106,12 +106,18 @@ export class DynamicBitmapFont extends AbstractBitmapFont<DynamicBitmapFont>
         let maxCharHeight = 0;
         let skipTexture = false;
 
+        //  style.lineHeight = 0;
+
```
Zyie, line 110:
```suggestion
```

PR #9921, opened 2023-12-03
v8 optimise container options

GoodBoyDigital:
- fix issue with parent in options
- added some tests

### DIFF
--- a/src/scene/container/Container.ts
+++ b/src/scene/container/Container.ts
@@ -13,7 +13,7 @@ import { onRenderMixin } from './container-mixins/onRenderMixin';
 import { sortMixin } from './container-mixins/sortMixin';
 import { toLocalGlobalMixin } from './container-mixins/toLocalGlobalMixin';
 import { LayerGroup } from './LayerGroup';
-import { definedProps } from './utils/definedProps';
+import { assignPropertiesFromOptions } from './utils/assignPropertiesFromOptions';
 
 import type { PointData } from '../../maths/point/PointData';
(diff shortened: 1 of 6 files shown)


src/scene/container/utils/assignPropertiesFromOptions.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Assigns properties from one object to another, using an optional array of property names to ignore.
+ * @param target - The target object to assign properties to.
+ * @param options - The object to assign properties from.
+ * @param ignore - An object of property names to ignore ({ propToIgnore: true }).
+ */
+export function assignPropertiesFromOptions<T extends Record<string, any>>(
+    target: T,
+    options: T,
+    ignore: Record<string, boolean> = {}
```
bigtimebuddy, line 10: Why not `string[]` instead?

If not, the type should probably be `Record<string, true>`.

src/scene/container/Container.ts
```diff
@@ -515,12 +515,15 @@ export class Container<T extends View = View> extends EventEmitter<ContainerEven
             options.view = undefined;
         }
 
-        Object.assign(this, definedProps(options));
+        assignPropertiesFromOptions(this, options, {
```
bigtimebuddy, line 518: Recommend renaming to `assignWithIgnore`. Since the functionality isn't really about props or options but a generic assign variant.

GoodBoyDigital: good shout! will adjust

Zyie (approved):

PR #9936, opened 2023-12-05
Reduce number of shared systems

Zyie:
Reduced the number of shared system to make it possible to fully customise the renderer more easily

we now have

- `pixi.js/graphics`
- `pixi.js/mesh`
- `pixi.js/text`
- `pixi.js/text-bitmap`
- `pixi.js/text-html`
- `pixi.js/filters`
- `pixi.js/sprite-tiling`

some of these require others to also be added at the same time, e.g `text-bitmap` requires both `graphics` and `text` to be imported. This is a more advanced use case but would be good to get the[pixi customise](https://pixijs.io/customize/) tool working for v8


GoodBoyDigital (approved):
lgtm!

PR #9935, opened 2023-12-05
Add number type support to `scale` field of `ISpritesheetData` 

maliut:
##### Description of change

In type `ISpritesheetData`, the type of `meta.scale` is `string`, which fits the output Json of TexturePacker.

```
"meta": {
	"app": "https://www.codeandweb.com/texturepacker",
	"version": "1.0",
	"image": "building1.png",
	"format": "RGBA8888",
	"size": {"w":128,"h":126},
	"scale": "0.5",
	"smartupdate": "$TexturePacker:SmartUpdate:1cb0f14cbdd5e7bcecd332ecd0eaa9f7:8acde9d234ecca966a410602c71bffad:6046b8eb706ddefaa771c33ceb7cb6d5$"
}
```

However, when using pixijs/assetpack to generate spritesheets, the output `scale` type is `number`.

```
    "meta": {
        "app": "http://github.com/odrick/free-tex-packer-core",
        "version": "0.3.4",
        "image": "spritesheet.png",
        "format": "RGBA8888",
        "size": {
            "w": 134,
            "h": 126
        },
        "scale": 1
    }
```

Pass the Json to `Spritesheet` constructor

```javascript
const sheet = new Spritesheet(baseTexture, data);
```

will cause a Typescript error:

```
TS2345: Argument of type
{ frames: { frame: ISpritesheetFrameData; }; meta: { scale: number; }; }
is not assignable to parameter of type  ISpritesheetData 
The types of  meta.scale  are incompatible between these types.
Type  number  is not assignable to type  string 
```

This PR changes the signature of `scale` to `string | number` to support both cases.

##### Pre-Merge Checklist

- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #9941, opened 2023-12-06
Add number type support to scale field of SpritesheetData

Zyie:
a port of #9935 for v8


src/spritesheet/Spritesheet.ts
```diff
@@ -265,7 +265,7 @@ export class Spritesheet<S extends SpritesheetData = SpritesheetData>
         this.animations = {} as Record<keyof NonNullable<S['animations']>, Texture[]>;
         this.data = data;
 
-        const metaResolution = parseFloat(data.meta.scale);
+        const metaResolution = typeof data.meta.scale === 'number' ? data.meta.scale : parseFloat(data.meta.scale);
```
GoodBoyDigital, line 268: does parseFloat not parse a string and a number?

PR #9947, opened 2023-12-07
Docs: Remove Accessibility Namespace Default Value

bbazukun123:
Resolves [ticket](https://github.com/orgs/pixijs/projects/2/views/4?pane=issue&itemId=46755362).

- JSDoc seems to infer that the `KEY_CODE_TAB` const after the namespace declaration is its default value. Here we can simply ignore it.


Zyie (approved):
nice catch!

PR #9950, opened 2023-12-07
Defer early allocation of canvas in Texture

### LINKED ISSUE #9946: Support for v8 for `pixijs` and `@pixi/webworker` package?
### Description

There seem to be multiple distributions of pixijs:
 * `pixijs`: https://www.npmjs.com/package/pixijs (with CanvasRenderer fallback)
 * `pixi.js`: https://www.npmjs.com/package/pixi.js (WebGL only?)
 * `@pixi/webworker`: https://www.npmjs.com/package/@pixi/webworker

It's unclear to me whether v8 will be available for the Web Worker version. Will there be a beta released?

Furthermore, I'm also unsure whether the Web Worker package contains fallbacks to Canvas2D or, like `pixi.js`, only supports WebGL?

gabrielecirulli:
Attempts to fix #9946.

`Texture.WHITE` is being generated using a Canvas object. Doing so in the global scope of the file means that when we call `DOMAdapter.get().createCanvas()` we might get the wrong adapter. For example, if Pixi is being used in a Web Worker, the DOMAdapter should have been set to a different one than the default. Because this texture is being generated in the global scope however, users of the library don't have time to set the right adapter before the wrong adapter is invoked.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/texture/Texture.ts
+++ b/src/rendering/renderers/shared/texture/Texture.ts
@@ -380,47 +380,48 @@ export class Texture extends EventEmitter<{
         return this._source;
     }
 
+    private static _emptyTexture: Texture;
+    private static _whiteTexture: Texture;
+
     /** an Empty Texture used internally by the engine */
-    public static EMPTY: Texture;
+    public static get EMPTY(): Texture {
+        if (!this._emptyTexture) {
+            const texture = new Texture({});
+            texture.label = "EMPTY";
+            texture.destroy = NOOP;
+            this._emptyTexture = texture;
+        }
+        return this._emptyTexture;
+    }
+
     /** a White texture used internally by the engine */
-    public static WHITE: Texture;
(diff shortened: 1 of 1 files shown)


gabrielecirulli:
This fix does not seem to be effective. There are too many places where `Texture.WHITE` is referenced that cannot all be memoized.

bigtimebuddy:
Thanks for taking a stab at this @gabrielecirulli. This should probably get fixed for v8. I'd lean on @Zyie to figure out how we might better handle or defer those other instances of Texture.WHITE/EMPTY.

gabrielecirulli:
Thanks. Out of curiosity, v8 is in rc, does that mean a release is imminent?

bigtimebuddy:
It means that we are semi-feature locked until the final release. Final release will mostly be bug/type/doc fixes and major issues from here on. We may go thru several rc's.

PR #9954, opened 2023-12-08
Update documentation header content for assets

bbazukun123:
Resolves [ticket](https://github.com/orgs/pixijs/projects/2/views/4?pane=issue&itemId=46876492).

- Add collapsable sections using details tag.
- Tweak and clean up content.
- Styling update is done on the template side, here: https://github.com/pixijs/webdoc-template/pull/30

### DIFF
--- a/src/assets/Assets.ts
+++ b/src/assets/Assets.ts
@@ -71,22 +71,30 @@ export interface AssetInitOptions
  * Only one Asset Class exists accessed via the Global Assets object. See [here]{@link assets.Assets},
  * for full documentation.
  *
- * It has four main responsibilities:
- * <br>     1. Allows users to map URLs to keys and resolve them according to the user's browser capabilities
- * <br>     2. Loads the resources and transforms them into assets that developers understand.
- * <br>     3. Caches the assets and provides a way to access them.
- * <br>     4. Allow developers to unload assets and clear the cache.
+ * <br>
  *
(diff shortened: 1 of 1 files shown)


src/assets/Assets.ts
```diff
@@ -71,22 +71,30 @@ export interface AssetInitOptions
...
+ *
+ * <details id="assets-loading">
+ *
+ * <summary>Assets Loading</summary>
  *
- * Do not be afraid to load things multiple times - under the hood, it will NEVER load anything more than once.
+ * Do not be afraid to load things multiple times - under the hood, it will **NEVER** load anything more than once.
  *
- * For example:
+ * <u>For example:</u>
```
bigtimebuddy, line 97: This is not semantically correct usage of UL. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul

Do you mean to use a header here instead?

bbazukun123: <u> here was intended for an underline but it seems that it is actually a misuse according to [source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u). I have switched to use italic instead for this 🙏

bigtimebuddy: Oops, I read that without my glasses on! Sorry, but I agree using italic is better here.

achamas-playco (approved):

PR #9962, opened 2023-12-08
fix: allow mask to be null or undefined

### LINKED ISSUE #9953: Bug: 8.0.0-rc setting container mask to undefined results in typescript error
### Current Behavior

```
container.mask = undefined;
```

Gives this typescript error: Type 'undefined' is not assignable to type 'number | Container<View>'.

### Expected Behavior

Should not error

### Steps to Reproduce

Run typescript

### Environment

### Possible Solution

_No response_

### Additional Information

_No response_

Zyie:
fixes #9953

### DIFF
--- a/src/scene/container/container-mixins/effectsMixin.ts
+++ b/src/scene/container/container-mixins/effectsMixin.ts
@@ -9,7 +9,7 @@ import type { Effect } from '../Effect';
 
 export interface EffectsMixinConstructor
 {
-    mask?: number | Container;
+    mask?: number | Container | null | undefined;
     filters?: Filter | Filter[];
 }
 export interface EffectsMixin extends Required<EffectsMixinConstructor>
--- a/src/scene/container/container-mixins/effectsMixin.ts
+++ b/src/scene/container/container-mixins/effectsMixin.ts
@@ -9,7 +9,7 @@ import type { Effect } from '../Effect';
 
 export interface EffectsMixinConstructor
 {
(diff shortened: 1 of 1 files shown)


src/scene/container/container-mixins/effectsMixin.ts
```diff
@@ -9,7 +9,7 @@ import type { Effect } from '../Effect';
 
 export interface EffectsMixinConstructor
 {
-    mask?: number | Container;
+    mask?: number | Container | null | undefined;
     filters?: Filter | Filter[];
```
bigtimebuddy, line 13: While we are here, what about `filters`? Shouldn't that allow null/undefined as well?

bigtimebuddy (approved):

src/scene/container/container-mixins/effectsMixin.ts
```diff
@@ -9,7 +9,7 @@ import type { Effect } from '../Effect';
 
 export interface EffectsMixinConstructor
 {
-    mask?: number | Container;
+    mask?: number | Container | null | undefined;
```
achamas-playco, line 12: Being an optional type it implicitly accepts `undefined` so you just need to add `null`.
```suggestion
    mask?: number | Container | null;
```

achamas-playco (changes requested):
Being an optional type it implicitly accepts `undefined` so you just need to add `null`.

achamas-playco (approved):

Issue #9393, opened 2023-04-20
Bug: [Assets] path containing period does not load correctly


SerG-Y:
### Current Behavior

Hi guys.
When I am trying to load the file by url like `http://0.0.0.0:8080/games/game3.0/resources/config.json` pixi.js looking for it in `http://0.0.0.0:8080/games/resources/config.json`. Is it correct behavior?
I think the reason is that `game3.0` in the base path `http://0.0.0.0:8080/games/game3.0` is recognized as an extension name and removed.
https://github.com/pixijs/pixijs/blob/00009c14d64e3b15b900977ee188c3613f1ffa04/packages/utils/src/path.ts#L307

### Expected Behavior

File loads correctly

### Steps to Reproduce

Try to load any file that contains a `dot` in the path.

### Environment

- **`pixi.js` version**: *7.2.3*
- **Browser & Version**: *Chrome 112*

amicoderozer:
I have the same problem, if the build folder contains the build version (e.g. 3.0.0) Pixi doesn't load the assets correctly. If I remove the version, it works. I tried with old pixi version (4.8.5) and it works as expected!

nouakam:
My team encountered the same problem.
Our deadlines coming up quickly, we had to revert to Pixi 6.5 to make things work for our game project.

My understanding is that this issue comes from the new Pixi.Assets module.
We have 3 distinct issues on Pixi 7+
- Unable to load if dots are found in the path
- Unable to load if credentials (username/password) are found in the path
- Unable to load a game wrapped with Cordova (apk). Issue with file:// paths?

SuperSodaSea:
It seems to be introduced in #8614. In my opinion we should not remove the filename based on its "extension" in the `join()` / `toAbsolute()`, but remove it if needed manually before calling them.

adevart:
I also had an issue with this in production builds with version numbers. I used a patch to ignore moving to the parent if the extension found matched a version number pattern in the URL. In the following code, I added a check if there was a version number:

https://github.com/pixijs/pixijs/blob/dev/packages/utils/src/path.ts#L316

```
const urlFileExtension = this.extname(prevArg);
const isVersionNumber = urlFileExtension.match(/\.[0-9]+/);

if (urlFileExtension && !isVersionNumber) {
    joined += `/../${arg}`;
} else {
    joined += `/${arg}`;
}
```

It's unlikely that a file with have a numeric extension so this should work for both scenarios.

[...]

PR #9963, opened 2023-12-08
Fix [Assets] path containing period does not load correctly

SerG-Y:
##### Description of change

Fix for #9393 
I did as @SuperSodaSea suggested:

> In my opinion we should not remove the filename based on its "extension" in the` join()` / `toAbsolute()`, but remove it if needed manually before calling them.

Some tests have been updated as they are no longer relevant.

##### Pre-Merge Checklist

### DIFF
--- a/packages/utils/src/path.ts
+++ b/packages/utils/src/path.ts
@@ -311,19 +311,7 @@ export const path: Path = {
             if (arg.length > 0)
             {
                 if (joined === undefined) joined = arg;
-                else
-                {
-                    const prevArg = segments[i - 1] ?? '';
-
-                    if (this.extname(prevArg))
-                    {
-                        joined += `/../${arg}`;
-                    }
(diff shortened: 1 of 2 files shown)


bigtimebuddy (changes requested):
Thanks for the PR, but this is a breaking change and not suitable for a patch. I think we need to have relative pathing logic, but need to find a way to be less greedy about what `path.join` considers a file extension.

Couple of other ideas how to resolve: 
* Always treat as a directory if path contains a trailing `/`.
* The set of possible positive matches for file extensions seems to be a narrow set of things. Why not check for what we consider _valid_ extensions (maybe expose that list to make it customizable) and treat everything else as a directory. What psycho would would name a directory "my-thing.jpg/"?

bigtimebuddy:
Replacing with #10030

PR #9969, opened 2023-12-12
rename renderer.view to renderer.surface

achamas-playco:
**Context:** Users were requesting the availability of `renderer.view` as per v7 (which is now `.canvas`), however in v8 this was used by the `ViewSystem`, so a balance of finding new names for classes and properties was needed.

* Renamed `ViewSystem` to `CanvasViewSystem` since `view` and `canvas` are associated with `HTMLCanvasElement`
* Renamed `renderer.view` to `renderer.surface` to expose the view (surface) system - `surface` does not collide with `view` or `canvas` which are accessors for the actual canvas
* Added getter for `renderer.view` for backwards compatibility

**Alternative:**
Another potential option is to rename `ViewSurfaceSystem` to `CanvasSystem` and use `renderer.canvas` for the system, keeping `renderer.view` for accessing the `HTMLCanvasElement`, however this would require a breaking change. The pros are it might be a better name for the view system since it mainly deals with a canvas, and users get to keep `.view`.

Thoughts?


GoodBoyDigital:
how about `DefaultCanvasSystem` ?

achamas-playco:
> how about `DefaultCanvasSystem` ?

Hmm, how about `CanvasViewSystem`? The "Default" bit feels extraneous do you think?

Zyie (approved):

Zyie (changes requested):
i accidently clicked approved on this pr instead of a different one, so just requesting changes until i review properly

Zyie:
closing in favour of #10059

PR #9970, opened 2023-12-12
add geometry bounds

GoodBoyDigital:
adds bounds object to geometry


achamas-playco (approved):
Looks good! Added unit test for `getGeometryBounds`.

PR #9971, opened 2023-12-12
v8 Add Get Fast Global Bounds

GoodBoyDigital:
`getGlobalBounds` is great, but somtimes we don't that exact precision (for example when measuring internanly for things like filters)

`getFastGlobalBounds` offers a much faster (over twice as fast)
approach to measuring when we want speed over accuracy.


src/scene/container/bounds/Bounds.ts
```diff
@@ -151,46 +153,51 @@ export class Bounds
...
+        if (y < minY) minY = y;
+        if (x > maxX) maxX = x;
+        if (y > maxY) maxY = y;
 
         this.minX = minX;
         this.minY = minY;
         this.maxX = maxX;
         this.maxY = maxY;
+
+        // console.log('addframe', this.minX, this.minY, this.maxX, this.maxY);
```
achamas-playco, line 190: This command could be removed.
```suggestion
```

src/scene/container/bounds/Bounds.ts
```diff
@@ -2,12 +2,12 @@ import { Matrix } from '../../../maths/matrix/Matrix';
 import { Rectangle } from '../../../maths/shapes/Rectangle';
 
 /** Simple bounds implementation instead of more ambiguous [number, number, number, number] */
-export interface SimpleBounds
+export interface BoundsData
```
achamas-playco, line 5: Ah, much better name.

achamas-playco (approved):
Nice one!

PR #9976, opened 2023-12-13
adjust texture.WHITE creation

GoodBoyDigital:
hey @Zyie! Heard `Texture.WHITE` was giving us jip!

After a chat with @bigtimebuddy, we realised it did not need to be created from a canvas and instead could be created from a Buffer.

Hope this helps to solve the Webworker issues!

### DIFF
--- a/src/rendering/renderers/shared/texture/Texture.ts
+++ b/src/rendering/renderers/shared/texture/Texture.ts
@@ -1,17 +1,15 @@
 import EventEmitter from 'eventemitter3';
 import { Cache } from '../../../../assets/cache/Cache';
-import { DOMAdapter } from '../../../../environment/adapter';
 import { groupD8 } from '../../../../maths/matrix/groupD8';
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):

src/rendering/renderers/shared/texture/Texture.ts
```diff
@@ -387,40 +385,19 @@ export class Texture extends EventEmitter<{
...
-ctx.lineTo(size, 0);
-ctx.lineTo(size, size);
-ctx.closePath();
-ctx.fillStyle = '#ffffff';
-ctx.fill();
-
 Texture.WHITE = new Texture({
-    source: new ImageSource({
-        resource: canvas,
+    source: new BufferImageSource({
```
bigtimebuddy, line 394: Will BufferImageSource work on canvas?

GoodBoyDigital: great question - yes we will should be able make it work with `putImageData`

PR #9980, opened 2023-12-14
enhance app docs

achamas-playco:
NOTE: Not sure if `ApplicationPlugin` is relevant so set to `@ignore`.

copilot:summary

### DIFF
--- a/src/app/Application.ts
+++ b/src/app/Application.ts
@@ -10,16 +10,17 @@ import type { DestroyOptions } from '../scene/container/destroyTypes';
 /**
  * The app module provides a set of classes to use as a starting point when building applications.
  *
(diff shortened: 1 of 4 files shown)


src/app/Application.ts
```diff
@@ -29,24 +30,26 @@ import type { DestroyOptions } from '../scene/container/destroyTypes';
...
+ *    },
+ *    destroy: () =>
  *    {
- *     // destruction code here
+ *       // handle destruction code here
  *    }
  * }
  * @memberof app
+ * @see {@link app.ApplicationOptions}
+ * @ignore
```
achamas-playco, line 46: I cleaned it up but then I searched around and found nothing uses this. The "Plugin" vs "Mixin" vs "Extension" is a little confusing.

GoodBoyDigital (approved):
lgtm!

PR #9984, opened 2023-12-15
enhance assets documentation

achamas-playco:
Enhance docs for assets.

Moved "Assets" info into "assets.html" as it was confusing on the main "assets" page header and more appropriate there. Linked to it.

Removed a lot of references to "This will be added automatically if `pixi.js/assets` is imported"
- There is no clear general explanation of the docs on how to custom load the library
- Feels like this API documentation should just focus on describing what things are rather than build composition details

### DIFF
--- a/src/assets/cache/parsers/cacheTextureArray.ts
+++ b/src/assets/cache/parsers/cacheTextureArray.ts
@@ -5,8 +5,6 @@ import type { CacheParser } from '../CacheParser';
 
 /**
  * Returns an object of textures from an array of textures to be cached
- *
- * This will be added automatically if `pixi.js/assets` is imported
  * @memberof assets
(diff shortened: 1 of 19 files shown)


src/assets/cache/parsers/cacheTextureArray.ts
```diff
@@ -5,8 +5,6 @@ import type { CacheParser } from '../CacheParser';
 
 /**
  * Returns an object of textures from an array of textures to be cached
- *
- * This will be added automatically if `pixi.js/assets` is imported
```
Zyie, line 9: if we remove all of these notes about importing `pixi.js/assets` we should still mention it somewhere that it may be required if you decide to manage your own imports

PR #9985, opened 2023-12-15
enhance color documentation

achamas-playco:
Enhance color documentation.

Ignored `Color.shared` as its a temp object which I believe is more for internal use than something users need to worry about. The object could also be mutated between the user setting its values and internal code running.

### DIFF
--- a/src/color/Color.ts
+++ b/src/color/Color.ts
@@ -6,7 +6,24 @@ import type { AnyColor, HslaColor, HslColor, HsvaColor, HsvColor, RgbaColor, Rgb
 extend([namesPlugin]);
 
 /**
- * Color utilities.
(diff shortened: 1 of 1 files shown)


src/color/Color.ts
```diff
@@ -93,13 +112,15 @@ type ColorSourceTypedArray = Float32Array | Uint8Array | Uint8ClampedArray;
...
  * new Color({ h: 0, s: 100, v: 100, a: 0.5 }).toArray(); // [1, 0, 0, 0.5]
+ * ```
  * @since 7.2.0
  * @memberof color
  */
 export class Color
 {
     /**
      * Default Color object for static uses
+     * @ignore
```
Zyie, line 123: this shouldn't be ignored. `Color.shared` is definitely a public API and useful for just converting colors without creating a new object

also why are all these PR's being merged without review?

PR #9995, opened 2023-12-18
enhance ticker documentation

achamas-playco:
Enhance ticker documentation

### DIFF
--- a/src/ticker/Ticker.ts
+++ b/src/ticker/Ticker.ts
@@ -1,25 +1,38 @@
 import { UPDATE_PRIORITY } from './const';
 import { TickerListener } from './TickerListener';
 
+/**
+ * A callback which can be added to a ticker.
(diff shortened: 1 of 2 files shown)


src/ticker/Ticker.ts
```diff
@@ -1,25 +1,38 @@
...
 
 /**
- * Tickers provide periodic callbacks based on a clock.
+ * {@link ticker.Ticker|Tickers} provide periodic callbacks based on the system clock.
  * Your game update logic will generally be run in response to a tick once per frame.
  * You can have multiple tickers in use at one time.
  * ```js
  * import { Ticker } from 'pixi.js';
  *
+ * const callback = () => {
```
Zyie, line 22: might be worth showing that the `ticker` is passed as the first argument here?

PR #9997, opened 2023-12-18
v8 - WebGL 1 support

GoodBoyDigital:
V8 now supports webGL 1 also! wahoo!

when creating a renderer there is a new property called `preferWebGLVersion` that will default to 2.
If `webGL2` is not supported, it will fall back to `webGL1`.

Everything should work the same with one exception - Antialiasing on textures. A warning will be thrown if you try this combo though and it will simply just not apply any antialiasing. So not breaking at least!

there are a lot of filter changes in this, but they are very small - basically renaming the output color in the shaders to all be the same `finalColor`. So this PR looks scarier than it is!


tests/visual/scenes/graphics/bezier.scene.ts
```diff
@@ -22,21 +22,17 @@ function drawCurve(
 
     const bezierA = new Graphics();
 
-    bezierA.beginPath();
```
Zyie, line 25: Any reason for changing this test in this PR?

It doesn't seem to relate to any of the webgl 1 changes

src/rendering/renderers/shared/renderTarget/GlobalUniformSystem.ts
```diff
@@ -201,7 +199,7 @@ export class GlobalUniformSystem implements System
             worldColorAlpha: { value: new Float32Array(4), type: 'vec4<f32>' },
             uResolution: { value: [0, 0], type: 'vec2<f32>' },
         }, {
-            ubo: true,
+            // ubo: this._renderer.supports.uniformBufferObject,
```
achamas-playco, line 202: Remove?

src/rendering/renderers/shared/startup/HelloSystem.ts
```diff
@@ -24,15 +25,12 @@ export class HelloSystem implements System<HelloSystemOptions>
             ExtensionType.CanvasSystem,
         ],
         name: 'hello',
-        priority: 0,
+        priority: -2,
```
achamas-playco, line 28: Priority change intentional?

achamas-playco (approved):
Amazing! Code looks good.

Will we be expanding our visual tests to cover WebGPU, WebGL1+2?

tests/visual/scenes/graphics/bezier.scene.ts
GoodBoyDigital: not really! Figured it should use the new v8 syntax whilst i was there.

src/rendering/renderers/shared/startup/HelloSystem.ts
GoodBoyDigital: yep! this requires the context to exist to get the correct WebGL version.

ErikSom (approved):

src/rendering/batcher/gl/GlBatchAdaptor.ts
```diff
@@ -60,7 +61,14 @@ export class GlBatchAdaptor implements BatcherAdaptor
 
         renderer.shader.bind(this._shader, this._didUpload);
 
-        renderer.shader.bindUniformBlock(renderer.globalUniforms.uniformGroup, 'globalUniforms', 0);
+        if (renderer.type === RendererType.WEBGPU)
```
Zyie, line 64: why is the `gl` adapter checking for webgpu?

we have the `GpuBatchAdapter`

src/rendering/renderers/gl/context/GlContextSystem.ts
```diff
@@ -197,28 +209,45 @@ export class GlContextSystem implements System<ContextSystemOptions>
...
+            gl = canvas.getContext('webgl', options);
+
+            if (!gl)
+            {
+                // fail, not able to get a context
+                throw new Error('This browser does not support WebGL. Try using the canvas renderer');
+            }
+        }
 
-        this.webGLVersion = 2;
```
Zyie, line 221: you are no longer setting the `webGlVersion` at all

shouldn't this be set to the `preferWebGLVersion`?

src/rendering/batcher/gl/GlBatchAdaptor.ts
GoodBoyDigital: good catch!

GoodBoyDigital: fixed

src/rendering/renderers/shared/renderTarget/GlobalUniformSystem.ts
GoodBoyDigital: yessir!

src/rendering/renderers/gl/context/GlContextSystem.ts
GoodBoyDigital: yo! This gets set in line 194:
```this.webGLVersion = gl instanceof WebGL2RenderingContext ? 2 : 1;```

stuck it there as the gl context may have been provided and we need to set it based in that.

Zyie (approved):

PR #10000, opened 2023-12-19
v8 add support for uint16

GoodBoyDigital:
Adds support for `Uint16Array` as an indexBuffer type

if a batch is smaller than 10000ish (which is most often the case!) 
we should use `Uint16Array` as they are half the size of `Uint32Arrays`.

### DIFF
--- a/src/rendering/batcher/shared/Batcher.ts
+++ b/src/rendering/batcher/shared/Batcher.ts
@@ -7,6 +7,7 @@ import { BatchTextureArray } from './BatchTextureArray';
 import { MAX_TEXTURES } from './const';
 
 import type { BindGroup } from '../../renderers/gpu/shader/BindGroup';
+import type { IndexBufferArray } from '../../renderers/shared/geometry/Geometry';
 import type { Instruction } from '../../renderers/shared/instructions/Instruction';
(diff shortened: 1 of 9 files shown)


bigtimebuddy:
Congrats on issue/pr 10000 :tada:

src/rendering/batcher/shared/Batcher.ts
```diff
@@ -65,7 +65,7 @@ export interface BatchableObject
         index: number,
         textureId: number,
     ) => void;
-    packIndex: (indexBuffer: Uint32Array, index: number, indicesOffset: number) => void;
+    packIndex: (indexBuffer: Uint32Array | Uint16Array, index: number, indicesOffset: number) => void;
```
achamas-playco, line 68: Maybe we should make this a specific type and use instead of the `|` signature. There seem to be enough of them to warrant it.

For example:
```
type IndexBufferArray = Uint32Array | Uint16Array;

packIndex: (indexBuffer: IndexBufferArray, index: number, indicesOffset: number) => void;
```
```

GoodBoyDigital: good shout! PR updated

Zyie (approved):

PR #10004, opened 2023-12-19
Bugfix/v8/gpu-mask

GoodBoyDigital:
finally got to the bottom of the stencil mask WebGPU issue. turns out you need to set the state for both front facing and back facing states! 🤦‍♂️


tests/visual/scenes/mask/circle-mask.scene.ts
```diff
@@ -5,11 +5,13 @@ import type { TestScene } from '../../types';
 
 export const scene: TestScene = {
     it: 'should render circle mask correctly',
+    only: true,
```
Zyie, line 8:
```suggestion
```

Zyie (approved):

PR #10005, opened 2023-12-19
fix depthTexture issue

GoodBoyDigital:
fix issue with setting 

`TextureSource.defaultOptions.autoGenerateMipmaps = true;`

this breaks the depthTextures used in the RenderTarget


Zyie (approved):
can you add a test for this?

achamas-playco (approved):

PR #10006, opened 2023-12-20
cleanup misc docs details

achamas-playco:
Clean up a few details after merging all the doc details and giving them a once over review.

Also update the `@pixi/webdoc-template` to latest `^2.2.2`.

### DIFF
--- a/src/accessibility/AccessibilitySystem.ts
+++ b/src/accessibility/AccessibilitySystem.ts
@@ -21,12 +21,14 @@ import type { isMobileResult } from '../utils/browser/isMobile';
  * This module is a mixin for {@link AbstractRenderer} and will need to be imported if you are managing your own renderer.
  * Usage:
  * ```js
- * import 'pixi.js';
- *
- * container.accessible = true; // see `AccessibleOptions` for more properties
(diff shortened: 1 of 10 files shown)


src/accessibility/AccessibilitySystem.ts
```diff
@@ -21,7 +21,7 @@ import type { isMobileResult } from '../utils/browser/isMobile';
  * This module is a mixin for {@link AbstractRenderer} and will need to be imported if you are managing your own renderer.
  * Usage:
  * ```js
- * import 'pixi.js/accessibility';
+ * import 'pixi.js';
```
Zyie, line 24: the `pixi.js/accessibility` import does exist and should be used if you are managing your own renderer
so this shouldn't be removed

achamas-playco: Ah good point thanks, I was coming at it from more of a general "how to make something accessible" angle.

I've worked it a little more so it separates those two concepts, ie 1) How to include if building your own renderer 2) How to make something accessible.

Zyie (approved):

PR #10008, opened 2023-12-20
Chore/v8/flatten TextStyle dropShadow properties

achamas-playco:
Flatten `TextStyle.dropShadow*` properties from object to separate properties.
Essentially reverts back to v7 format.

### DIFF
--- a/src/scene/text/TextStyle.ts
+++ b/src/scene/text/TextStyle.ts
@@ -275,12 +275,12 @@ export class TextStyle extends EventEmitter<{
             this.fontSize = fullStyle.fontSize as number;
         }
 
-        this._dropShadow = style.dropShadow ?? defaultStyle.dropShadow;
-        this._dropShadowAlpha = style.dropShadowAlpha ?? defaultStyle.dropShadowAlpha;
-        this._dropShadowAngle = style.dropShadowAngle ?? defaultStyle.dropShadowAngle;
-        this._dropShadowBlur = style.dropShadowBlur ?? defaultStyle.dropShadowBlur;
-        this._dropShadowColor = style.dropShadowColor ?? defaultStyle.dropShadowColor;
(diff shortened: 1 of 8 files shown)


tests/visual/scenes/text/text-trim.scene.ts
achamas-playco, line 1: Renamed this as it was the same name as another test and was confusing.

tests/visual/scenes/text/text.scene.ts
achamas-playco, line 1: This wasn't failing locally. I deleted the snapshot and regenerated and the image was exactly the same. For some reason it fails with a relatively high threshold on CI only. I'm not a fan of bumping the threshold like this because it may let other regressions through, though this seemed like the only option.

src/scene/text/TextStyle.ts
```diff
@@ -279,26 +275,12 @@ export class TextStyle extends EventEmitter<{
...
-            }
-            else
-            {
-                this.dropShadow = {
-                    ...TextStyle.defaultTextStyle.dropShadow as TextDropShadow,
-                    ...style.dropShadow as TextDropShadow
-                };
-            }
-        }
+        this._dropShadow = style.dropShadow ?? defaultStyle.dropShadow;
```
GoodBoyDigital, line 278: no need for  the `?? defaultStyle.dropShadow;` bits
this happens above in line 259:

``` const fullStyle = { ...defaultStyle, ...style };```

achamas-playco: Yes! Good catch thanks.

achamas-playco:
@Zyie @GoodBoyDigital pls merge this [PR](https://github.com/orgs/pixijs/projects/2/views/4?pane=issue&itemId=47556362) first, as its based on this current branch, then merge this to `next-v8`.

Zyie:
closing as we are going to take a slightly different approach

PR #10010, opened 2023-12-21
Chore/v8/refactor text

achamas-playco:
Remove `renderMode` concept from `Text` class. Concrete classes for the specific render modes are now preferred.

Refactor existing `Text` functionality into `AbstractText` class and derive 3 main subclasses, which use 3 specific `TextView`:
* `Text` - Canvas based rendering - uses `TextView`
* `BitmapText` - Bitmap font based rendering - uses `BitmapTextView`
* `HTMLText` - HTML text rendering using svg - uses `HTMLTextView`

`TextStyle` is largely untouched though feedback welcome for how to split that into text class specific cases.

Adjusted unit and visual tests to use specific concrete text classes.
Distributed single text documentation into concrete classes.


src/scene/text-bitmap/BitmapTextView.ts
```diff
@@ -0,0 +1,27 @@
...
+        const width = bitmapMeasurement.width * scale;
+        const height = bitmapMeasurement.height * scale;
+
+        bounds.minX = (-anchor._x * width) - padding;
+        bounds.maxX = bounds.minX + width;
+        bounds.minY = (-anchor._y * (height + offset)) - padding;
+        bounds.maxY = bounds.minY + height;
+    }
+
+    get renderPipeId(): string
```
GoodBoyDigital, line 23: no need for a getter here, we could just have it as areadonly string
```public readonly renderPipeId = 'bitmapText'```

GoodBoyDigital (commented):
Thanks for this Ali!

AbstractText should be in a in its own file.. not sure where.. will let you pick!
We should also lift text out scene and add it as top level packages (like v7)

Zyie:
> We should also lift text out scene and add it as top level packages (like v7)
scene is for anything that extends a container, which text does. Why does this need to be moved out?

src/scene/text-bitmap/BitmapTextView.ts
achamas-playco: True! Have updated `TextView`, `BitmapTextView` and `HTMLTextView`.

achamas-playco:
> > We should also lift text out scene and add it as top level packages (like v7)
> 
> @GoodBoyDigital scene is for anything that extends a container, which text does. Why does this need to be moved out?

That's what I also assumed? The docs currently reflect this, ie. if you look at the `scene` namespace you see all the other concrete display object types like `Sprite`, `Graphics` etc, and they all live in `/src/scene`.

All the other display object classes are sub folders of `src/scene` and there was already `/src/scene/text`, `src/scene/text-bitmap` and `src/scene/text-html` folders with existing files, which is the reason behind the current home.

Zyie:
closing as we are going to take a slightly different approach to this

PR #10011, opened 2023-12-21
fixes for webgpu

GoodBoyDigital:
- adjustment to the WebGPU renderer 
- fix gpuProgram pipeline
- fix buffer destroy via events


achamas-playco:
It's still failing on the webgpu rect fill test unfortunately, generating the webgl1 style output only on CI and not locally. We'll need to look into it further.

Zyie (approved):

PR #10022, opened 2023-12-26
Add strict null checks to @pixijs/runner

8times12:
##### Description of change

Ref https://github.com/pixijs/pixijs/issues/8852

To resolve the strict null check error in `packages/runner/src/Runner.ts`, I modified the `destroy` method to set each property value from null to an empty array or an empty string.

I also considered making each property nullable, but decided not to change the type for the following reasons:

- If I make each property nullable, it would require implementing null check processing in the calling code, leading to broader implications.
- There is currently no handling in the calling code to anticipate cases where the property value might be null.

##### Pre-Merge Checklist

### DIFF
--- a/packages/runner/src/Runner.ts
+++ b/packages/runner/src/Runner.ts
@@ -168,8 +168,8 @@ export class Runner
     public destroy(): void
     {
         this.removeAll();
-        this.items = null;
-        this._name = null;
+        this.items = [];
+        this._name = '';
     }
 
     /**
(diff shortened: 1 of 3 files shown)


8times12:
Another approach could be setting a `destroyed` flag when calling the `destroy` method, and throwing errors whenever methods are invoked while this flag is set.
While this might be more appropriate for the post-destroy logic, considering design consistency, similar adjustments would be needed for other classes that implement the `destroy` method.

bigtimebuddy (approved):
This effort went stale for awhile, thanks for bring back strict null checks! :tada:

8times12:
@bigtimebuddy 
Thank you for the review!
I have made the necessary corrections based on your suggestions. Please review it again.

bigtimebuddy:
Thanks @8times12

PR #10025, opened 2023-12-27
flatten TextureSource.style options

achamas-playco:
Flatten the `TextureSource.style` options so they are flat properties for the `TextureSource` constructor. Preserve the use of the `style` object.


src/rendering/renderers/shared/texture/sources/TextureSource.ts
```diff
@@ -241,9 +238,23 @@ export class TextureSource<T extends Record<string, any> = any> extends EventEmi
         this.antialias = options.antialias;
         this.alphaMode = options.alphaMode;
 
-        const style = options.style ?? {};
-
-        this.style = style instanceof TextureStyle ? style : new TextureStyle(style);
+        const styleOptions: TextureStyleOptions = {};
```
bigtimebuddy, line 241: This is pretty verbose. Can you try something like using a `reduce` instead?
```ts
const styleOptions: TextureStyleOptions = Object.entries(options)
  .filter(([key]) => styleKeys.includes(key))
  .filter(([key, value]) => value !== undefined)
  .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {});
```

achamas-playco: I agree absolutely agree in terms of high verbosity, but in the essence of performance it's better by removing the need for function calls.

This is a direction that @GoodBoyDigital has requested on many occasions so I'm following from a style guide point of view. There have even been requests on some PRs to avoid the spread operator for similar reasons.

Mat, what would you prefer?

bigtimebuddy: TextureSource constructor hardly seems like hot code.

achamas-playco: True, though from your suggestion (unless I'm mistaken) the `styleKeys` value is new and would need to be a `const` array of the style keys based on `TextureStyle` class, which would require a new `interface` extracted to ensure strong types, which the `TextureStyle` class would also need to express it `implements`. To me this would increase complexity and moving parts for the sake of a little verbosity.

PR #10029, opened 2023-12-28
set transparent as default renderer background color

achamas-playco:
Fixes this [ticket](https://github.com/orgs/pixijs/projects/2/views/4?pane=issue&itemId=47556362).

Sets a default alpha of `0` to the `BackgroundSystem`. This creates a transparent rendering background.
If a color is passed to either the initial rendering options, or later via `renderer.background.color` then the alpha will become `1`, so the color is visible.

Tested these variations in the blaze playground, added unit tests for each case.

### DIFF
--- a/src/rendering/renderers/shared/background/BackgroundSystem.ts
+++ b/src/rendering/renderers/shared/background/BackgroundSystem.ts
@@ -86,7 +86,7 @@ export class BackgroundSystem implements System
         this._backgroundColor = new Color(0x000000);
 
         this.color = this._backgroundColor; // run bg color setter
-        this.alpha = 1;
+        this.alpha = 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/rendering/renderers/shared/background/BackgroundSystem.ts
```diff
@@ -97,11 +97,11 @@ export class BackgroundSystem implements System
     {
         options = { ...BackgroundSystem.defaultOptions, ...options };
 
+        const optionsHasBackgroundColor = !!(options.background || options.backgroundColor);
+
         this.clearBeforeRender = options.clearBeforeRender;
         this.color = options.background || options.backgroundColor || this._backgroundColor; // run bg color setter
-        this.alpha = options.backgroundAlpha;
-
-        this._backgroundColor.setAlpha(options.backgroundAlpha);
```
achamas-playco, line 104: Line `104` here was extraneous, removed as alpha logic is already handled by line above.

GoodBoyDigital:
i think i might need to look into this one - as it should be that the view background color is only used when rendering the views canvas. and anything else should be transparent.

Zyie:
closing in favor of #10053

PR #8614, opened 2022-08-25
Fix relative path matching

Zyie:
Fixes #8610

The path module didn't take into account a URL ending in an extension e.g `index.html`
it also didn't handle url params inside of a path

Example
```
Website URL:  http://www.exampleurl.com/mypage/index.html?variable=test
Asset Path:   img/logo.png
```
Before
```
Resolved URL: http://www.exampleurl.com/mypage/index.html?variable=test/img/logo.png
```
After
```
Resolved URL: http://www.exampleurl.com/mypage/img/logo.png
```

PR #10030, opened 2023-12-28
Fix: path.join incorrectly joining urls that contain periods

Zyie:
fixes: #9393

https://github.com/pixijs/pixijs/pull/8614 introduced a bug that caused any segment that ended in a file extension to be handled differently. This PR now explicitly allows you to define what extensions should be handled differently with the default being `['.html']`

The `.html` default is because the original issue was around how these two urls are the same when `Assets` does its `path.toAbsolute` when setting a `basePath`
`http://example.com/index.html` and `http://example.com/`

### DIFF
--- a/packages/utils/src/path.ts
+++ b/packages/utils/src/path.ts
@@ -315,7 +315,7 @@ export const path: Path = {
                 {
                     const prevArg = segments[i - 1] ?? '';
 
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Nice work!

PR #10031, opened 2023-12-28
chore: add shader.from

Zyie:
adds `Shader.from` and also makes the types for Shader more explicit

e.g
```ts
// error: must have a gl or gpu program
new Shader({
    resources: {}
});

// error: must have a gl or gpu program
new Shader({
    group: {}
});

// error: cant have resources and groups both defined
new Shader({
    resources: undefined,
    groups: undefined
    glProgram: undefined,
    gpuProgram: undefined,
});
```


GoodBoyDigital (approved):
👍

PR #10032, opened 2023-12-28
Chore: remove ObjectPool and update docs

Zyie:
I've removed the `ObjectPool` that was recently added as we already had the `Pool` class that does exactly the same thing

and a couple of docs have been updated


GoodBoyDigital (approved):
👍

PR #10047, opened 2024-01-03
Add support for 'signal' option to FederatedEventTarget.addEventListener

### LINKED ISSUE #10043: Feature Request: @pixi/events addEventListener does not obey signal option
### Current Behavior

When using an `AbortController` to pass a signal to the event package's `addEventListener` options object, the event continues to fire even after the `abort()` method is called on the controller.

### Expected Behavior

The event listener should be removed when the `abort()` method is called.

### Steps to Reproduce

Given an environment with `pixi.js` and `@pixi/events` available:

```
import * as PIXI from 'pixi.js';
import '@pixi/events';

    width: window.innerWidth,
    height: window.innerHeight,
    backgroundColor: 0x2c3e50
});

PIXI.Assets.load('https://pixijs.io/examples/examples/assets/bunny.png')
            .then((img) => startup(img));

function startup(img)
{
    var bunny = new PIXI.Sprite(img);
    bunny.anchor.set(0.5);
    bunny.x = app.renderer.width / 2;
    bunny.y = app.renderer.height / 2;

    app.stage.addChild(bunny);

    // Event test case
    bunny.eventMode = 'static';
    const abort = new AbortController();
    var count = 0;

    bunny.addEventListener('click', (e) => {
      count = count + 1;
      console.log("Clicked", count, "times!");
      bunny.rotation += 10;
      if (count > 2) {
        abort.abort();
      }
    }, { signal: abort.signal });
}
```

After 3 clicks, the bunny should stop rotating on click. The event continues to fire, however.

### Environment

- **`pixi.js` version**: 7.3.3
- **Browser & Version**: Version 120.0.6099.129 (Official Build) (arm64)
- **OS & Version**: Mac Sonoma 14.1.2
- **Running Example**: Playground doesn't support importing events package :(

### Possible Solution

_No response_

### Additional Information

_No response_

jmcclell:
This change brings the addEventListener implementation more in line with the DOM implementation.

##### Description of change

Adds support for the 'signal' option to the addEventListener method of FederatedEventTarget. This makes this method more congruent with the DOM implementation. Fixes #10043

##### Pre-Merge Checklist


bigtimebuddy (approved):
This works for me. Thank you @jmcclell

bigtimebuddy:
@jmcclell could you resolve conflicts with the once PR?

bigtimebuddy:
Thanks again for these PRs 👍 This was added to v8 too.

jmcclell:
@bigtimebuddy I just realized this contains a nefarious bug. 🤦🏻 In an effort to guarantee (which I _beleive_ the DOM implementation does) that there is no chance the event will fire after `abort()` is called, I wrap the original istener to add a check on the signal itself rather than solely relying on the signal's `abort` event to remove the original listener. Unfortunately, this means that if you want to manually remove the listener, you'd need a reference to that wrapped function as your original listener will no longer match in `off`. 

Here is a test case that demonstrates:

```
    it('should allow manually removing the original event listener when using an AbortSignal', () =>
    {
        const renderer = createRenderer();
        const [stage, graphics] = createScene();
        const eventSpy = jest.fn();

        renderer.render(stage);
        const controller = new AbortController();

        const listener = (e: Event) => {
            expect(e.type).toEqual('click');
            eventSpy();
        };

        graphics.addEventListener('pointertap', listener, { signal: controller.signal });

        const click = () =>
        {
            const event = new PointerEvent('pointerdown', { clientX: 25, clientY: 25 });

            renderer.events.onPointerDown(event);
            const e = new PointerEvent('pointerup', { clientX: 30, clientY: 20 });
            // so it isn't a pointerupoutside

            Object.defineProperty(e, 'target', {
                writable: false,
                value: renderer.view
            });
            renderer.events.onPointerUp(e);
        };

        click(); // Once
        graphics.removeEventListener('pointertap', listener);
        click(); // Twice
        expect(eventSpy).toHaveBeenCalledOnce();
    });

```

For the moment, I might suggest we revert the portion of the change that wraps the listener function and keep the listener to the signal's abort function. In the meantime, I'm trying to see what the DOM implementation actually guarantees. I've always relied on it being as I wrote it, but perhaps I've just been lucky.

jmcclell:
@bigtimebuddy Okay, so given that browsers run JS in a single-threaded context, I believe we are safe to remove the wrapper and that it is entirely superfluous.

If my understanding of the runtime semantics is correct, as soon as the `abort()` method fires the `abort` event, that event is added to the event loop's queue. No other subsequent events will be processed before that event. The processing of that event should allow for the removal of the listener so that any subsequent events that would have otherwise triggered the listener will now be processed without the listener present.

Will ship up a patch. Let me know if my thinking seems correct here. JS is not my primary language. :)

bigtimebuddy:
Yeah, that seems reasonable approach and simplifies the code too.

jmcclell:
PR #10073 pushed. Sorry for the confusion! Appreciate the prompt responses.

jmcclell:
Scratch that. Had some confusing on branches on my end. PR #10074 :)

PR #10052, opened 2024-01-03
Feat: add graphics clone

Zyie:


GoodBoyDigital (approved):
lgtm! a test or two would be good you have the time!

src/scene/graphics/shared/GraphicsContext.ts
```diff
@@ -137,6 +137,23 @@ export class GraphicsContext extends EventEmitter<{
     private _bounds = new Bounds();
     private _boundsDirty = true;
 
+    public clone(): GraphicsContext
```
GoodBoyDigital, line 140: lets add some docs here too!

PR #10054, opened 2024-01-03
chore: publish under pixijs

Zyie:
Just a thought on how we can publish to both `pixi.js` and `pixijs` easily


bigtimebuddy:
I'm pretty sure publishing ignores the package-lock.json, is it really necessary to change that?

Personal preference would be to change the default name to be `pixijs` and backward publish to `pixi.js`, but I'm fine with whatever you decide.

scripts/publish.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        await writeJSON(packagePath, updatedPackage);
+        await writeJSON(packageLockPath, updatedPackageLock);
+
+        // Run npm publish again
+        await spawn('npm', ['publish']);
+
+        // Restore the original package.json and package-lock.json
+        await writeJSON(packagePath, originalPackage);
+        await writeJSON(packageLockPath, originalPackageLock);
```
bigtimebuddy, line 35: I might suggest doing something like this to catch any errors with the publish and make sure everything in the workspace is always reverted.

```suggestion
        try {
             // Run npm publish again
             await spawn('npm', ['publish']);
        }
        finally {         
            // Restore the original package.json and package-lock.json
            await writeJSON(packagePath, originalPackage);
            await writeJSON(packageLockPath, originalPackageLock);
        }
```

bigtimebuddy: Oops just noticed the try/catch above. So, not quite this, but some way to make sure JSON is always reverted in case something goes wrong.

Zyie:
@bigtimebuddy thanks for the feedback, ive updated the script to add the `finally` and made `pixijs` the default

bigtimebuddy (approved):

bigtimebuddy:
Potentially follow-up here is lots of documentation that imports from `pixi.js` and README too.

https://github.com/pixijs/pixijs/blob/a4c018f9868d71861382635722e1132d98c3e2b3/src/app/Application.ts#L19

bigtimebuddy:
I think I've changed my mind on this issue (i.e., "pixi.js" vs "pixijs"). It's an superficial change but creates more headaches in the ecosystem. How do plugins know which package to use? I think we should deprecate "pixijs" package instead.

PR #10060, opened 2024-01-04
Fix: gpu sprite being undefined

Zyie:
Codesandbox would alway error out when it tries to reload the page as `gpuSprite` is undefined.


GoodBoyDigital (approved):
nice!

PR #10062, opened 2024-01-04
Fix: filter issue with skipped filter effects

GoodBoyDigital:
added some tests


tests/scene/filter-effect.tests.ts
```diff
@@ -0,0 +1,71 @@
...
+    {
+        const container = new Container();
+
+        expect(container.effects.length).toBe(0);
+
+        expect(container.filters).toBe(undefined);
+
+        container.filters = [];
+
+        expect(container.effects.length).toBe(1);
```
Zyie, line 18: why would filters have a length of 1 here after being set to an empty array?

tests/scene/filter-effect.tests.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    it('should set filter effects correctly if a modifed array is passed in', async () =>
+    {
+        const container = new Container();
+
+        const filters: Filter[] = [];
+
+        container.filters = filters;
+
+        expect(container.effects.length).toBe(1);
```
Zyie, line 63: same here?

GoodBoyDigital:
thanks for the sanity check!! updated the PR - works better now :)

tests/visual/scenes/container/skip-filter.scene.ts
```diff
@@ -0,0 +1,38 @@
+import { ColorMatrixFilter } from '../../../../src/filters/defaults/color-matrix/ColorMatrixFilter';
+import { NoiseFilter } from '../../../../src/filters/defaults/noise/NoiseFilter';
+import { Container } from '../../../../src/scene/container/Container';
+import { Graphics } from '../../../../src/scene/graphics/shared/Graphics';
+
+import type { TestScene } from '../../types';
+
+export const scene: TestScene = {
+    it: 'should skip filters correctly',
+    only: true,
```
Zyie, line 10:
```suggestion
```

Zyie (approved):

PR #10085, opened 2024-01-08
Only compile filters once where possible

minetoblend:
##### Description of change
Attempts to address some performance issues I encountered when spawning a lot of objects with the same filter at the same time (in my case `AlphaFilter`). The renderer would compile the program for each filter separately, despite them all sharing the same code. Implementation is based on how `TilingSpriteShader` does it, though I'm not 100% if reusing the programs will work across multiple PIXI instances.

I've made a little example that showcases the issue. 
https://www.pixiplayground.com/#/edit/_1N5rSl_W1_lbspwvjOQ3

The performance impact seems to be much bigger in webgl compared to webgpu, spending about 500ms compiling the shaders when creating 100 alpha filter instances on my end.
image

### DIFF
--- a/src/filters/defaults/alpha/AlphaFilter.ts
+++ b/src/filters/defaults/alpha/AlphaFilter.ts
@@ -8,6 +8,9 @@ import source from './alpha.wgsl';
 
 import type { FilterOptions } from '../../Filter';
 
+let gpuProgram: GpuProgram;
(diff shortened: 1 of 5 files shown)


minetoblend:
Closing since this was the wrong approach
 Was supposed to use `GpuProgram.from()` instead which already caches the program.

PR #10094, opened 2024-01-10
added context loss set up to WebGL

GoodBoyDigital:
WebGL now loses context without exploding. You won't even notice it happened!

Added test also


Zyie (approved):

tgrajewski:
I might be wrong, but that doesn't look like a valid code:

```js
        // Restore the context after this event has exited
        setTimeout(() =>
        {
            if (this.gl.isContextLost() && this.extensions.loseContext)
            {
                this.extensions.loseContext.restoreContext();
            }
        }, 0);
```

What I mean specifically is that the function [restoreContext()](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_lose_context/restoreContext) isn't meant to be called in production, it's only for debugging purposes. Which also means that there is no way to restore GPU context on demand, the browser decides about it, not JS application.

PR #10095, opened 2024-01-10
Fix: issues with TextStyle getter/setters and BitmapText default fill

Zyie:
This PR fixes a couple of issues with TextStyle

- the setters did not have the logic that was in the constructor
- dropShadow is now fully supported as either a bool or an object and will apply defaults if a partial style is provided
- stroke now checks for `_originalStroke` instead of `_originalFill`
- when converting from v7 drop shadow options the default settings are now applied 

This PR also fixes the issue where bitmap texts fill was always being set to white

### DIFF
--- a/src/scene/text/Text.ts
+++ b/src/scene/text/Text.ts
@@ -178,9 +178,9 @@ export class Text extends Container implements View
 
         this.style = style;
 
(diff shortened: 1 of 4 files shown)


GoodBoyDigital (approved):
nice

PR #10109, opened 2024-01-12
Chore: split CI into multiple workflows

Zyie:
This PR splits the `node.yml` file up into multiple reusable parts

The main reason for this change is that with visual testing we started hitting the limits on what a single workflow could handle and had errors that are not present locally. These issues went away when we changed to a more powerful runner, however the costs are pretty high.

So the solution is multiple workflows that split the unit and visual testing up

Changes

- added a couple of reusable actions for linting/testing/setup
<br>

- `pull-unit.yml` - runs linting/type checking, runs unit tests, tests building the project
- `pull-visuals.yml` - runs visual tests

<br>

- `push.yml` - will deploy non tagged branches to s3. Will no longer do any sort of validation, only builds and uploads. I figure that the PR workflows handle this and the main reason for the upload is to more easily test docs/builds

<br>

- `release-unit.yml` - same as `pull-unit.yml` without the build step
- `release-visuals.yml` - same as `pull-visuals.yml`
- `release.yml` - This workflow depends on the other two release workflows to be completed, then will build, upload and release the project


GoodBoyDigital (approved):
nice!

bigtimebuddy:
The node/npm versions are spread all over the place as inputs of the setup workflow. It would be easier to manage if you just declared that in one place (or used the input default).

bigtimebuddy:
Otherwise, love the modularity of these workflows. Great job!

PR #10113, opened 2024-01-13
Use `GpuProgram.from` in filters

minetoblend:
Uses `Gpu/GlProgram.from` instead of their constructor for creating shader programs in `AlphaFilter`, `BlendModeFilter`, and `NoiseFilter` to prevent the programs from being compiled separately for each filter instance.


GoodBoyDigital (approved):
thanks @minetoblend !!! 👍

PR #10116, opened 2024-01-15
Feat: add ability to set width/height of Container together

Zyie:
This PR adds the ability to set the width/height of a container in one go which reduces the number of `getLocalBounds` calls that get made

```ts
new Container()
    .setSize(100) // 100, 100
    .setSize(100, 150) // 100, 150
    .setSize({width: 100}) // 100, 100
    .setSize({width: 100, height: 150}) // 100, 150
```

There is also a `getSize()` with an optional `out` property to get the width/height of a container without the additional calls as well


GoodBoyDigital (approved):
👍

PR #10124, opened 2024-01-17
Fix: support for unsafe eval nad std40 uniform buffers

GoodBoyDigital:
- renamed files to better reflect their purpose
- added support for wgsl alongside std40 uniform buffers
- unified uniforms to all work of uniformTypes
- fixed filters that had temp workarounds

- unsafe eval working like a champ now too.
- creates cacheing functions and auto generates the code from our uniform generators - so we have one source of truth


tests/visual/visuals.test.ts
```diff
@@ -51,12 +51,15 @@ describe('Visual Tests', () =>
 {
     scenesToTest.forEach((scene) =>
     {
+        //        if (i > 6 || i < 4) return;
+        //  if (i !== 4) return;
+
```
Zyie, line 56:
```suggestion
```

Zyie (approved):

PR #10125, opened 2024-01-17
Add: Spritesheet option cachePrefix

bigtimebuddy:
This was a request that was made via Discord. 

The use-case is that we have a global `TextureCache` and a global `Assets` cache and collisions of textures can easily happen with an app that's a little more complex. Sometimes there's a need to be more specific when loading multiple atlases that might share the same texture name. This change allows users to optionally provide a cachePrefix that prepended to texture IDs whenever adding to TextureCache or Assets' Cache.

### Added

* Adds `cachePrefix` option to Spritesheet
* Adds options for Spritesheet constructor

### Chores

* General cleanup of Spritesheet tests to unload assets better

### Example

```ts

await Assets.load({
   src: './atlas.json', // contains "background" texture
   data: { cachePrefix: 'atlas.json/' }
});

await Assets.load({
   src: './atlas2.json', // also contains "background" texture
   data: { cachePrefix: 'atlas2.json/' }
});

const background = Assets.get('atlas.json/background');
const background2 = Assets.get('atlas2.json/background');
````

### DIFF
--- a/packages/spritesheet/src/Spritesheet.ts
+++ b/packages/spritesheet/src/Spritesheet.ts
@@ -78,6 +78,26 @@ export interface ISpritesheetData
 }
 
 /**
+ * Options for loading a spritesheet from an atlas.
+ * @memberof PIXI
+ */
+interface SpritesheetOptions<S extends ISpritesheetData = ISpritesheetData>
+{
+    /** Reference to Texture */
+    texture: BaseTexture | Texture;
+    /** JSON data for the atlas. */
+    data: S;
+    /** The filename to consider when determining the resolution of the spritesheet. */
+    resolutionFilename?: string;
+    /**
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/spritesheet/src/Spritesheet.ts [hunk 1]
```diff
@@ -388,7 +428,7 @@ export class Spritesheet<S extends ISpritesheetData = ISpritesheetData>
                 );
 
                 // lets also add the frame to pixi's global cache for 'from' and 'fromLoader' functions
-                Texture.addToCache(this.textures[i], i.toString());
+                Texture.addToCache(this.textures[i], this.cachePrefix + i.toString());
```
aaclayton, line 431: This would certainly resolve my use case of needing to namespace same-named textures in multiple loaded spritesheets, so looks good to me!

Zyie (approved):

PR #10152, opened 2024-01-25
Fix: advanced blend modes not being bundled

Zyie:
advanced blend modes were entirely missing from previous releases

we now have a separate bundle for advanced blend modes to keep the main bundle small


GoodBoyDigital (approved):
👍

Issue #10136, opened 2024-01-19
Bug: v8.0.0-rc.5 Culled containers are culled forever


davidye:
### Current Behavior

Starting a new issue for clarity's sake.

When the scene is culled with Culler.shared.cull(), containers that are culled in prior calls are culled forever unless explicitly reset with container.visible = true.

See playground example: https://www.pixiplayground.com/#/edit/WhU7CVBPsqQn37HwXYpeL

### Expected Behavior

The playground example should render without the need to manually reset the 'visible' property.

### Steps to Reproduce

See playground

### Environment

See playground

Zyie:
hey @davidye i've added a pr to add a new property so it no longer overrides visible. Let me know if this solves your issues

PR #10154, opened 2024-01-25
Fix: culler no longer overrides visible property

Zyie:
fixes: #10136


src/scene/container/Container.ts
```diff
@@ -530,12 +532,12 @@ export class Container extends EventEmitter<ContainerEvents & AnyEvent>
      * @internal
```
GoodBoyDigital, line 532:
```suggestion
     * This property holds three bits: culled, visible, renderable
     * the third bit represents culling (0 = culled, 1 = not culled) 0b100
     * the second bit represents visibility (0 = not visible, 1 = visible) 0b010
     * the first bit represents renderable (0 = renderable, 1 = not renderable) 0b001
     * @internal
```

GoodBoyDigital (approved):
lgtm!

PR #10165, opened 2024-01-30
Fix: types not building correctly

Zyie:
For whatever reason when we build the types we get `/// <reference types="dist" />`
This is not what we want, we want `/// <reference types="@webgpu/types" />`
So i wrote a little script just to replace this out

Also we needed to copy over our `Shader.d.ts` file which defines frag,vert,wgsl files


scripts/types/fixTypes.ts
```diff
@@ -1,32 +1,91 @@
...
+ * - Add reference paths to the mixins
+ * - Replace the dist triple slash reference with the webgpu types
+ * - Copy over the Shaders.d.ts file to the lib folder for
+ *   proper frag,vert,wgsl support
+ * -----------------------------------------------------------------
+ */
+
+import fs from 'fs';
+import glob from 'glob';
+import path from 'path';
```
bigtimebuddy, line 13: Minor future-proof import suggestions.

```suggestion
import * as fs from 'node:fs';
import * as glob from 'glob';
import * as path from 'node:path';
```

scripts/types/fixTypes.ts
```diff
@@ -1,32 +1,91 @@
...
-files.forEach((file) =>
+    const filePath = path.join(process.cwd(), './lib/index.d.ts');
+    const contents = fs.readFileSync(filePath, 'utf8');
+    const updatedContents = `${lines.join('\n')}\n${contents}`;
+
+    fs.writeFileSync(filePath, updatedContents);
+}
+
+/**
+ * For whatever reason when we build the types we get /// <reference types="dist" />
```
bigtimebuddy, line 44: Would be nice to know why this is happening. TS bug or are we implementing something incorrectly?

bigtimebuddy (approved):

scripts/types/fixTypes.ts
GoodBoyDigital: agree! do we have any idea?

GoodBoyDigital (approved):
lgtm

scripts/types/fixTypes.ts
Zyie: I currently have no idea what is causing this, but will add a ticket to look into it some more

PR #10178, opened 2024-02-03
Introduced dts-bundle-generator to produce types declaration bundles

### LINKED ISSUE #7155: PixiJS v6: Request to support generating a single TypeScript definition file
In PixiJS v5, a single definition file is created, that covers the entire library.
This is useful to be able to take and use for 'typings' in an IDE such as Visual Code / WebStorm, even when actually loading an application based on a pre-built browser build - either from bundling it locally, or loading pixi via a http script. A bit like DefinitelyTyped supplied. [Pixi playground](https://www.pixiplayground.com/) uses this for [live typings](https://www.pixiplayground.com/definitions/master/pixi.js.d.ts) in the browser, for example.

However, in PixiJS v6 typings are only provided on a per package basis, with the expectation that PixiJS is a devDependency on your project. Just like v6 does bundles to be used as both imported packages and purely in the browser, it'd be nice if there was an equivalent to both in the way the typings are generated, too.

timocov:
Fixes #7155

##### Description of change

The change introduces an ability to generate typescript typings declaration files.

Below you can find an example of generated bundles:
- `bundles/pixi.js` - [pixi.js.zip](https://github.com/pixijs/pixijs/files/14154206/pixi.js.zip)
- `bundles/pixi.js-legacy` - [pixi.js-legacy.zip](https://github.com/pixijs/pixijs/files/14154207/pixi.js-legacy.zip)

I didn't quite know which bundles you're interested in, but for my best judgement it feels like this is what you wanted (based on the ticket).

In order to create a bundle I needed to make some changes to the codebase, especially for:
- global files which mostly contain declaration of `GlobalMixins` namespace. The changes aren't significant and should be backward-compatible, but I suggest you to take a closer took to decide if you're happy with them
- I'd like to use `src/index.ts` as an input file in each bundle, but apparently because of the compiler bug it doesn't generate correct `/// <reference path` directives based on the code, so I decided to create a separate file that just re-exports everything from index file but adds "correct" references

The command output:

```
$ npm run build:types-bundle

> pixi.js-monorepo@7.4.0 build:types-bundle
> dts-bundle-generator --config dts-config.js

Compiling input files...
Processing bundles/pixi.js/src/typings.d.ts
Processing bundles/pixi.js-legacy/src/typings.d.ts
Writing bundles/pixi.js/src/typings.d.ts -> bundles/pixi.js/dist/pixi.d.ts
Writing bundles/pixi.js-legacy/src/typings.d.ts -> bundles/pixi.js-legacy/dist/pixi-legacy.d.ts
Checking generated files...
Done in 6.10s
```

Some additional notes to consider:
- I added `build:types-bundle` script that runs declaration generation but it is not used anywhere yet. Feel free to propose a suggestion what's the best place/command would be
- Please carefully review changes in `packages/events/global.d.ts`
- In `packages/text-html/src/HTMLText.ts` I had to change the type of an argument in `_renderCanvas` function, otherwise it would fail with the compilation error. It feels like this is what is declared in the base type too, but happy to discuss possible options.

If you're not happy with these changes by any reason (e.g. you don't like a way of declaring globals or anything else) - feel free to cancel the PR.

##### Pre-Merge Checklist


packages/accessibility/global.d.ts
```diff
@@ -1,8 +1,13 @@
...
     {
+        // eslint-disable-next-line @typescript-eslint/no-empty-interface
+        interface DisplayObject extends Partial<import('@pixi/accessibility').IAccessibleTarget>
+        {
 
+        }
     }
 }
+
+export {};
```
timocov, line 13: this turns a file into a "module", which means that to declare a global you need to use `declare global` statement. It just helps dts-bundle-generator differentiate between "local" and "global" declarations. changes in all global files are pretty much the same - wrapping with `declare global` and adding `export {}` (apart from few files)

packages/events/global.d.ts
```diff
@@ -1,47 +1,50 @@
-declare namespace GlobalMixins
-{
-    type FederatedEventEmitterTypes = import('@pixi/events').FederatedEventEmitterTypes;
-    type FederatedEventTarget = import('@pixi/events').FederatedEventTarget;
-    type IFederatedDisplayObject = import('@pixi/events').IFederatedDisplayObject;
```
timocov, line 5: this potentially could be a breaking change. basically that type is just added to a global scope with the same name as it is declared in local, but I don't know how it could be solved with typescript syntax (after bundling it becomes `type A = A` which is an invalid syntax). but considering that you should be able to use these types via exports (or `PIXI`) it shouldn't be a problem?

packages/text-html/src/HTMLText.ts
```diff
@@ -312,7 +313,7 @@ export class HTMLText extends Sprite
      * @private
      * @param {PIXI.CanvasRenderer} renderer - The renderer
      */
-    _renderCanvas(renderer: IRenderer)
+    _renderCanvas(renderer: CanvasRenderer)
```
timocov, line 316: in `packages/canvas-display/global.d.ts` it is declared as the following:

```ts
        interface Container
        {
            _renderCanvas(renderer: import('@pixi/canvas-renderer').CanvasRenderer): void;
        }
```

which makes me think that this change is legit?

tsconfig.json
```diff
@@ -15,6 +15,7 @@
         "resolveJsonModule": true,
         "downlevelIteration": true,
         "baseUrl": "./",
+        "types": ["node", "css-font-loading-module"],
```
timocov, line 18: this significantly speeds up the compilation, it was well noticeable with bundling - by default typescript includes all `@types`, even ones that you don't use, e.g. lots from `@babel`) https://www.typescriptlang.org/tsconfig#types

dts-config.js
```diff
@@ -0,0 +1,36 @@
...
+        '@pixi/colord',
+        'earcut',
+        'eventemitter3',
+    ],
+};
+
+const output = {
+    inlineDeclareGlobals: true,
+    exportReferencedTypes: false,
+    umdModuleName: 'PIXI',
```
timocov, line 12: it is possible that instead of this we might need to use explicit global declarations, something like this:

```ts
import * as PIXIModule from './index';

declare global {
	const PIXI: PIXIModule;
}
```

timocov:
It seems https://github.com/pixijs/pixijs/pull/10286 is merged, so closing this.

Issue #10175, opened 2024-02-03
Bug: Incorrect Text color


EneaEntertainment:
### Current Behavior

When renderMode is `bitmap`, the fill color `0x0` is not used and the result is white text. If fill is set to `'black'` or another non-zero value, the text color is correct.

### Expected Behavior

Text color to be black.

### Steps to Reproduce

Please see: https://www.pixiplayground.com/#/edit/ePzOLCHzqg9TMsKyi9mKS

### Environment

- **`pixi.js` version**: 8.0.0 rc.7

PR #10188, opened 2024-02-06
Fix: allow setting bitmap text fill as 0x0

Zyie:
fixes: #10175


src/scene/text/TextStyle.ts
```diff
@@ -392,7 +392,14 @@ export class TextStyle extends EventEmitter<{
...
         this._originalFill = value;
-        this._fill = convertFillInputToFillStyle(value, GraphicsContext.defaultFillStyle);
+        if (value === 0x0)
+        {
+            this._fill = convertFillInputToFillStyle('black', GraphicsContext.defaultFillStyle);
+        }
+        else
+        {
+            this._fill = convertFillInputToFillStyle(value, GraphicsContext.defaultFillStyle);
+        }
```
bigtimebuddy, line 402: Can we one-line this?

```suggestion
        this._fill = convertFillInputToFillStyle(
            value === 0x0 ? 'black' : value,
            GraphicsContext.defaultFillStyle
        );
```

GoodBoyDigital (approved):
lgtm! Matts suggestion seems good also!

PR #10189, opened 2024-02-06
Fix: texture.update() now only updates uvs and dispatches an event

GoodBoyDigital:
A couple of releases ago, I added a pass through update function. So updating the texture, would call update on the source. This was a bad idea! 

As when you call update on a texture, you should not expect it to re-upload its source (which can be easily acheived by calling `texture.source.update()`

Calling texture update now simple updates the uvs and dispatches an update event. Much better!


tests/renderering/textures/Textures.test.ts
```diff
@@ -92,4 +92,18 @@ describe('Texture', () =>
     {
         expect(Texture.WHITE.source.alphaMode).toBe('premultiply-alpha-on-upload');
     });
+
+    it.only('should call update a texture correctly', () =>
```
Zyie, line 96:
```suggestion
    it('should call update a texture correctly', () =>
```

Zyie (approved):

PR #10191, opened 2024-02-06
Breaking: split `Text` back into `Text`, `BitmapText`, and `HTMLText` 

GoodBoyDigital:
Changed Text to be a little simpler

Rather than renderMode on one Text, we now have `Text`, `BitmapText`, `HTMLText`

This means we can safely adjust the properties over time of the individual classes, without complicating the Text class.

Basically a little closer to how v7 API is!


src/scene/text-html/HtmlText.ts
```diff
@@ -0,0 +1,98 @@
...
+ *     style: {
+ *         fontFamily: 'Arial',
+ *         fontSize: 24,
+ *         fill: 0xff1010,
+ *         align: 'center',
+ *     }
+ * });
+ * @memberof scene
+ */
+export class HtmlText extends AbstractText<HTMLTextStyle, HTMLTextStyleOptions> implements View
```
bigtimebuddy, line 69: Can we keep the casing the same as v7? HTMLText

GoodBoyDigital: yes! good shout

bigtimebuddy (approved):

src/scene/text-bitmap/BitmapFontManager.ts
```diff
@@ -45,7 +77,7 @@ class BitmapFontManagerClass
      */
     public readonly ASCII = [[' ', '~']];
 
-    public defaultOptions: BitmapFontInstallOptions = {
+    public defaultOptions: Omit<BitmapFontInstallOptions, 'style'> = {
```
Zyie, line 80: does this need to be public?

GoodBoyDigital: this is cool as its a singleton

Zyie (approved):

PR #10198, opened 2024-02-08
Fix: pixi.js/unsafe-eval broken in 8.0.0-rc.9

SuperSodaSea:
##### Description of change

It seems that the `pixi.js/unsafe-eval` in v8 has been broken since 8.0.0-rc.6 (by 32cc5583f2023aefd0f0ebbc45c8c4fe8915ed9c):

```
Uncaught (in promise) Error: Current environment does not allow unsafe-eval, please use pixi.js/unsafe-eval module to enable support.
    at GlUboSystem._systemCheck
    ...
```

It turns out that the `UboSystem._systemCheck` is not overrided by an empty function, so the unsafe check is still performed. This PR fixes this issue.

##### Pre-Merge Checklist


SuperSodaSea:
@GoodBoyDigital Should we add some tests to check that `pixi.js/unsafe-eval` is working properly?

GoodBoyDigital (approved):
nice! thanks @SuperSodaSea !!

GoodBoyDigital:
> @GoodBoyDigital Should we add some tests to check that `pixi.js/unsafe-eval` is working properly?

that would be awesome!!

SuperSodaSea:
Tests added!

PR #10214, opened 2024-02-14
Feat: add number setters for transform points

Zyie:
now you can set scale with a number!

```
const mySprite = new Sprite(...);

mySprite.scale = 2; // scaleis now 2!
```


src/scene/container/Container.ts
```diff
@@ -863,14 +862,15 @@ export class Container extends EventEmitter<ContainerEvents & AnyEvent>
...
     {
         return this._position;
     }
 
-    set position(value: PointData)
+    set position(value: PointData | number)
     {
-        this._position.copyFrom(value);
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+        typeof value === 'number' ? this._position.set(value) : this._position.copyFrom(value);
```
bigtimebuddy, line 873: A number to set position seems a little strange IMO. Scale makes a lot of sense though.

Zyie: I've removed the number setter for position

skew currently doesn't have a setter, which is an entirely different issue which i need to fix 😆 

To me setting a number for pivot does make sense since its sort of like anchor, but happy to be overruled on this one

bigtimebuddy: Sure. Also what about anchor too?

src/scene/sprite-tiling/TilingSprite.ts
```diff
@@ -246,19 +247,19 @@ export class TilingSprite extends Container implements View, Instruction
...
     {
         return this._anchor;
     }
 
-    set anchor(value: PointData)
+    set anchor(value: PointData | number)
     {
-        this._anchor.x = value.x;
-        this._anchor.y = value.y;
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
GoodBoyDigital, line 257: should we disable this rule? // eslint-disable-next-line @typescript-eslint/no-unused-expressions

GoodBoyDigital (approved):
👍

bigtimebuddy (approved):

PR #10238, opened 2024-02-21
Fix: Make filters setting immutable

### LINKED ISSUE #10226: Impossible to set sprite filters as an empty array to then add multiple filters
I noticed that it's impossible to pre assign an empty array to a sprite filters, and then add multiple filters to it.

Here is a fiddle : https://jsfiddle.net/pn5Lkjy3/1/

As you can see, the filter isn't displayed at all.

However, this would work (direct filter assignment but then we can't use multiple filters on a sprite) : https://jsfiddle.net/pn5Lkjy3/2/

**Edit** : Even when setting directly app.stage.filters = [filter], it wont work after a certain amount of time (generally if you don't create the shockwave immediatly on the main function).

bigtimebuddy:
Fixes #10226

This change makes the array passed to `filters` immutable with types and at runtime. This is to prevent users from `container.filters.push()` for example.


Zyie (approved):

nasselk:
And how would you add a filter additionally to others (if you don't have reference to all of the others filters) ?

You need to make a copy of the actual filter list, add the new filter to it, and assign the new list as the sprite filters?

bigtimebuddy:
You could add additional filters using the spread operator.

```ts
sprite.filters = [new AlphaFilter()]

// append to existing
sprite.filters = [...sprite.filters, new NoiseFilter()]
```

PR #10243, opened 2024-02-23
Improve the @pixi/runner developer experience

koteelok:
### Description of change

The current implementation of the `runner' package is not always usable with classes and can cause bugs due to overly complicated generics.

I removed default values of `Runner` generics, fixed improper use of `unknown` type and made types smaller to make errors more readable.

Also i found a [bug(?) in BufferSystem](https://github.com/pixijs/pixijs/commit/688ab2647d99b6a38702c760f9334aeeee1f9eab) caused by lack of strict checks that the object contains a callback function. I didn't add these checks myself so as not to change the APIs of this class too much, but this change may help us avoid similar problems in the future. 
Let me know if it's OK to add them, I'll include them in this PR.

#### before
old1

#### now
new1

#### before

old2

#### now

new2

### Pre-Merge Checklist

### DIFF
--- a/packages/core/src/framebuffer/Framebuffer.ts
+++ b/packages/core/src/framebuffer/Framebuffer.ts
@@ -44,7 +44,7 @@ export class Framebuffer
     depthTexture: BaseTexture;
     colorTextures: Array<BaseTexture>;
     glFramebuffers: {[key: string]: GLFramebuffer};
-    disposeRunner: Runner;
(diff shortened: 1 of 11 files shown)


reececomo:
If you use any kind of code/property mangling for production bundles you might want to be passing `this.myFunction.name` instead of raw strings

Zyie:
Hey,

PR #10258, opened 2024-03-01
fix: fix export jpg format error(v8)

yungzhu:
rt


yungzhu:
image

Zyie:
Hey @yungzhu could you update this PR to the latest `next-v8`? 
There seems to be some weird merge issues

src/rendering/renderers/shared/extract/ExtractSystem.ts
```diff
@@ -98,6 +98,8 @@ export class ExtractSystem implements System
 
         const { format, quality } = options;
 
+        const imageFormat = format === 'png' ? 'image/png' : 'image/jpeg';
```
bigtimebuddy, line 101: I don't think this is correct. The format option should be mime type not "png". Can you please create a bug for this?

bigtimebuddy (changes requested):

src/rendering/renderers/shared/extract/ExtractSystem.ts
yungzhu: image
Actually the argument that's passed in is this

bigtimebuddy: The fix here should probably be this instead:

```ts
type Formats = 'image/png' | 'image/jpeg';
```

yungzhu: That's fine too, but png and jpg are easier to use

bigtimebuddy: If that's the case, then probably a map:
```ts
const imageTypes: Record<Formats, string> = {
  png: 'image/png',
  jpg: 'image/jpeg',
  webp: 'image/webp',
};
```
Personally, I think the mime type is more transparent, though slightly more verbose, with how toDataURL and other canvas methods are invoked.

yungzhu: I agree. In fact, I can accept all of them

PR #10278, opened 2024-03-06
Fix: graphics chaining

Zyie:
Fixes `this` not always being returned
also adds docs to the graphics class


GoodBoyDigital (approved):
nice nice!

PR #10280, opened 2024-03-06
Fix: circular deps

GoodBoyDigital:
fixes circular dependancies in the code

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,7 +164,7 @@
     "./lib/index.*",
     "./lib/rendering/init.*",
     "./lib/spritesheet/init.*",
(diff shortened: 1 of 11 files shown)


bigtimebuddy:
I think we should add `failAfterWarnings` ([docs](https://rollupjs.org/command-line-interface/#failafterwarnings)) which should fail the build for future circular dependency warnings.

Zyie:
> I think we should add `failAfterWarnings` ([docs](https://rollupjs.org/command-line-interface/#failafterwarnings)) which should fail the build for future circular dependency warnings.

added!

bigtimebuddy (approved):

PR #10297, opened 2024-03-08
Fix: add missing resize event

Zyie:
adds the missing `resize` event on the renderer from v7


bigtimebuddy (approved):
Looks great. A test would be helpful too.

PR #10316, opened 2024-03-13
Fix: export jpg format error(v8)

yungzhu:
format is 'image/png' or 'image/jpeg'
add a map


src/rendering/renderers/shared/extract/ExtractSystem.ts
```diff
@@ -12,6 +12,12 @@ import type { GenerateTextureOptions } from './GenerateTextureSystem';
 
 type Formats = 'png' | 'jpg'; // Add other formats if needed
 
+const imageTypes: Record<Formats, string> = {
+    png: 'image/png',
+    jpg: 'image/jpeg',
+    // webp: 'image/webp',
```
bigtimebuddy, line 18: Can we enable webp? Was there a reason you disabled?

Also, the Formats type can be simplified to be `typeof keyof imageTypes` to be less redundant.

yungzhu:
type Formats = 'png' | 'jpg'; 
Because the current definition is two, we can add more if needed

src/rendering/renderers/shared/extract/ExtractSystem.ts
yungzhu: type Formats = 'png' | 'jpg';
Because the current definition is two, we can add more if needed

bigtimebuddy: Go ahead and just add webp. It is supported with the browser API

yungzhu: ok

bigtimebuddy (approved):
Nice! Thank you.

PR #10320, opened 2024-03-13
Fix: shader preprocessor to support older chrome versions

dragoon:
Chrome 78 and older do not parse shader source correct when single quote is present inside the comment and produce and error message: shaderSource: string not ASCII

##### Pre-Merge Checklist


GoodBoyDigital:
nice catch @dragoon !

bigtimebuddy (approved):
Wow, good find!

PR #10346, opened 2024-03-20
Fix: performance for setting textures in WebGL

GoodBoyDigital:
GlTextureSystem optimisation:

When first creating a texture and applying its samples - we can skip setting the params that are the same as the defaults. 

this speeds up texture creation. from the cpu side by reducing the need to change the default.

I also set the default to texture style to `repeat`. Which makes more sense now that its supported by WebGL2 and WebGPU for all textures (we also have a catch for this edge case in WebGL 1 already, so its safe!)

It will alls mean textures will just tile by default when being used in shaders - which is more in line with standard expectations. 

This change won't affect anything like sprites etc

### DIFF
--- a/src/rendering/renderers/gl/texture/GlTextureSystem.ts
+++ b/src/rendering/renderers/gl/texture/GlTextureSystem.ts
@@ -223,12 +223,17 @@ export class GlTextureSystem implements System, CanvasGenerator
         }
 
         this.onSourceUpdate(source);
-        this.onStyleChange(source);
+        this.updateStyle(source, false);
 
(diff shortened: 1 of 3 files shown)


ivanpopelyshev:
Btw, I think it saves nothing, adds extra parameter and IF's

GoodBoyDigital:
> Btw, I think it saves nothing, only adds extra parameter and IF's

noted! when profiling this save a bunch of CPU time for standard texture creation, so figure na harm!

Zyie (approved):

Issue #10331, opened 2024-03-14
Bug: [v8] BpuBufferSystem.onBufferDestroy contains null gpuBuffer when destroying via react hook


levi:
### Current Behavior

When destroying my pixi v8 app from a React useEffect hook, I get intermittent crashes related to the GPU buffer no longer existing.

My react component has access to a global instance of my pixi app with a useEffect hook something like this:

```
  useEffect(() => {
    app.load();
    return () => {
      app.unload();
    };
  }, []);
```

And the app class looks something like this:

```
class App {
  [..]
  async load() {
    this._pixi = new Application()
    await this._pixi.init()
  }

  unload() {
    if (this._pixi && this._pixi.renderer) {
      this._pixi.destroy(false)
    }
  }
}
```

Here's the error and stacktrace

```
Unhandled Runtime Error
TypeError: Cannot read properties of null (reading 'destroy')

Source
GpuBufferSystem.onBufferDestroy
javascript/node_modules/pixi.js/lib/rendering/renderers/gpu/buffer/GpuBufferSystem.mjs (67:0)
eval
javascript/node_modules/pixi.js/lib/rendering/renderers/gpu/buffer/GpuBufferSystem.mjs (75:0)
Array.forEach
<anonymous>
GpuBufferSystem.destroy
javascript/node_modules/pixi.js/lib/rendering/renderers/gpu/buffer/GpuBufferSystem.mjs (75:0)
SystemRunner.emit
javascript/node_modules/pixi.js/lib/rendering/renderers/shared/system/SystemRunner.mjs (19:0)
WebGPURenderer.destroy
javascript/node_modules/pixi.js/lib/rendering/renderers/shared/system/AbstractRenderer.mjs (218:0)
_Application.destroy
javascript/node_modules/pixi.js/lib/app/Application.mjs (79:0)
src/lib/canvas/app.ts (350:17) @ destroy

  348 |
  349 | if (this._pixi && this._pixi.renderer) {
> 350 |   this._pixi.destroy(false)
      |             ^
  351 | }
  352 | this._pixi = null
  353 |
```

### Expected Behavior

No crash when using it within a useEffect hook

### Steps to Reproduce

Defined above

### Environment

- **`pixi.js` version**: 8.0.2

### Possible Solution

I assume the canvas is being destroyed before the useEffect hook is being called. I attempted to resolve this by using useLayoutEffect, but there was no difference.

GoodBoyDigital:
[...]

PR #10348, opened 2024-03-20
Alternate way to destory GpuBufferSystem

GoodBoyDigital:
was looking to address #10331 but bit tricky to recreate! even if this does not fix things, its a nicer way to destroy all buffers at once.

### DIFF
--- a/src/rendering/renderers/gpu/buffer/GpuBufferSystem.ts
+++ b/src/rendering/renderers/gpu/buffer/GpuBufferSystem.ts
@@ -111,28 +111,31 @@ export class GpuBufferSystem implements System
      */
     protected onBufferDestroy(buffer: Buffer): void
     {
+        this._managedBuffers.splice(this._managedBuffers.indexOf(buffer), 1);
+
+        this._destroyBuffer(buffer);
+    }
+
+    public destroy(): void
+    {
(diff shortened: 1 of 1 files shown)


TomSlezakowski:
Hello! I just ran into the same issue! I tried the above patch but the issue persists. 

From a little digging, it seems like the order of execution means duplicate `uid`'s are passed in to the `_destroyBuffer` function. Changing `_destroyBuffer` to `if (gpuBuffer) gpuBuffer.destroy();` was enough to get things working again for me.

Here's a log of `destroy()` with the respective buffer & uid.

```
GPUBuffer {size: 192, usage: 40, mapState: 'unmapped', label: 'attribute-batch-buffer'} - UID: 8
GPUBuffer {size: 24, usage: 24, mapState: 'unmapped', label: 'index-batch-buffer'} - UID: 9
GPUBuffer {size: 262140, usage: 76, mapState: 'unmapped', label: ''} - UID: 6
GPUBuffer {size: 262140, usage: 72, mapState: 'unmapped', label: ''} - UID: 7
null - UID: 9
null - UID: 8
```

Hope that helps, and congrats on V8! 🚀

Zyie (approved):

PR #10352, opened 2024-03-21
Fix: uv updates for textures

GoodBoyDigital:
Sprites (and other batchable texture driven views) will now update if a textures uvs are updated.

The simple apoproach of just listening for a texture change via the event emitter was proving to be really slow when many textures are swapped each frame (like 10000s). 

To mitigate this, I created a very simple listener class that adds and removes very quickely. This makes it run much faster when swapping textures on sprites than the previous attempt of adding removing event listeners.


ivanpopelyshev:
Is it potential memory leak for users who dont use "destroy()" for simple elements like Sprite? Which is actually most of our audience

ivanpopelyshev:
Also, adding 100k sprites then removing them , fills the SimpleListen map with null's. When I alloc 100k more, they'll appear in the list again, which is potential problem for things like particle emitters

GoodBoyDigital:
i think leak wise we could be ok - no different than how standard events emitters could potentially memory leak. As the only way to get into this would be to have set the texture from the sprite. We should be cool

GoodBoyDigital:
> Also, adding 100k sprites then removing them , fills the SimpleListen map with null's. When I alloc 100k more, they'll appear in the list again, which is potential problem for things like particle emitters

let me have a think on this, technically true - my emphasis was to not impact people who do not update uvs on textures often (which is most users)

ivanpopelyshev:
> let me have a think on this, technically true - my emphasis was to not impact people who do not update uvs on textures often (which is most users)

What if we add setting to the texture whether it tracks objects? We can even create that `SimpleListen` object lazily, only when user calls something like `ensureListenerUpdates()`

ivanpopelyshev:
In any case, there should be `delete`, not just setting to null. Also its better to be a Map , not object - its faster for small uid's

`SimpleListen` should be named like `ListenByUid` , something that contains `Uid`.

GoodBoyDigital:
Normally i would agree - but delete is slow :)
The goal here is to make add / remove as fast as possible (as this was a huge bottle neck for many sprites swapping textures each frame)

GoodBoyDigital:
agree the naming is whack. Will adjust

GoodBoyDigital:
switching to delete drops fps by 50% for 100000 bunnys cycling through bunny textures

ivanpopelyshev:
> switching to delete drops fps by 50% for 100000 bunnys cycling through bunny textures

What about using "map" ?

GoodBoyDigital:
> > switching to delete drops fps by 50% for 100000 bunnys cycling through bunny textures
> 
> What about using "map" ?

same massive fps drop :/ 50% slower. I wish there was a less derpy way - but we can't lose all that perf for something that most users don't do (mess with texture uvs)

bigtimebuddy:
Why not use a Runner?

GoodBoyDigital:
> Why not use a Runner?

Same problem i'm afraid! the act of adding / reming to the underlining array is whats tanking perf :/

GoodBoyDigital:
closing in favour of #10379

PR #10357, opened 2024-03-21
Fix: NineSliceSprite not rendering correctly when used with a spritesheet texture

GoodBoyDigital:
We had previously adjusted the mesh to take into account sprite sheets.
Nine Slice Sprite was also doing this work - so uvs were being multiplied twice.
easy fix at least!

### DIFF
--- a/src/scene/sprite-nine-slice/NineSliceSprite.ts
+++ b/src/scene/sprite-nine-slice/NineSliceSprite.ts
@@ -252,12 +252,6 @@ export class NineSliceSprite extends Container implements View
         this._roundPixels = value ? 1 : 0;
     }
 
-    /** The texture matrix of the NineSliceSprite. */
-    get textureMatrix()
-    {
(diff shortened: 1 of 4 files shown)


src/scene/sprite-nine-slice/NineSliceSprite.ts
```diff
@@ -252,12 +252,6 @@ export class NineSliceSprite extends Container implements View
         this._roundPixels = value ? 1 : 0;
     }
 
-    /** The texture matrix of the NineSliceSprite. */
-    get textureMatrix()
-    {
-        return this._texture.textureMatrix.mapCoord;
-    }
-
     /** The original width of the texture */
```
Zyie, line 255: any reason for removing this getter?

GoodBoyDigital: no longer required! one less prop to expose in the API

Zyie (approved):

PR #10367, opened 2024-03-22
Fix: Change WebGL to default renderer

GoodBoyDigital:
After seeing the feedback from the commmunity (and seeing some of this first hand) I think it would be best in the short term to make the WebGL renderer the default.

There are two main reasons:

- WebGL is generally more performant at the fundamental level for most of PixiJS's use cases right now. (this makes sense as Browser have had a lot more time to be optimise WebGL than they have WebGPU). An example of this is uploading a texture, WebGL is just that little bit quicker!

- WebGL is more stable. Whilst the chrome implementation seems ok, browsers like FF are a lot more unstable. With quirks and issues at the WebGPU layer.

Not worried about both of these being resolved in time, but for now I think most users would prefer stable over bleeding edge (especially if there is no current perf gains)

Technically a breaking change, but perhaps we are early enough to get away with a minor bump :D

Thought welcome!

### DIFF
--- a/src/rendering/renderers/autoDetectRenderer.ts
+++ b/src/rendering/renderers/autoDetectRenderer.ts
@@ -27,7 +27,7 @@ export interface AutoDetectOptions extends WebGLOptions, WebGPUOptions
     webgl?: Partial<WebGLOptions>;
 }
 
-const renderPriority = ['webgpu', 'webgl', 'canvas'];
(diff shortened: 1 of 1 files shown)


danielbarion (approved):

ivanpopelyshev (approved):
Agree

bigtimebuddy (approved):
Agree! Good call.

PR #10379, opened 2024-03-26
Bugfix: Fix up UV updates in a simpler way

GoodBoyDigital:
Rather than do all the crazy stuff from here #10352 
Instead just add flag to the texture that indicates if its dynamic.

The one change is that we ask the dev to turn it on as required and keep it to false by default. 

Added visual test


ivanpopelyshev (approved):

tests/visual/snapshots/dynamic-texture-scene-ts-webgl1.png
Zyie, line 1: This image doesn't look like the rest!

PR #10383, opened 2024-03-27
Do not use texSubImage2D if not necessary to improve performance

gfodor:
##### Description of change
This PR updates the image uploader to use texImage2D instead of texSubImage2D when possible. In testing on my phone I see a significant performance improvement with this patch.

As per this discussion at least some time in the past it was acknowledged this was a problem: https://issues.chromium.org/issues/40469618#comment85

It's unclear if the fast path has been updated appropriately for texSubImage2D after that comment but in any case the performance improvement on my test device is noticable. (CPU utilization of my app went from 120-150% to 100-120%)

##### Pre-Merge Checklist
- [ ] Tests and/or benchmarks are included (needed?)
- [ ] Documentation is changed or added (needed?)

### DIFF
--- a/src/rendering/renderers/gl/texture/uploaders/glUploadImageResource.ts
+++ b/src/rendering/renderers/gl/texture/uploaders/glUploadImageResource.ts
@@ -69,11 +69,10 @@ export const glUploadImageResource = {
         }
         else if (glWidth === textureWidth || glHeight === textureHeight)
         {
-            gl.texSubImage2D(
-                gl.TEXTURE_2D,
+            gl.texImage2D(
+                glTexture.target,
                 0,
-                0,
-                0,
+                glTexture.internalFormat,
                 glTexture.format,
                 glTexture.type,
                 source.resource as TexImageSource


ivanpopelyshev (changes requested):
There should be a flag.

Regardless, later, we'll need to make immutable textures too - `texStorage2D` makes things faster for sure

gfodor:
Cool. Just for my understanding - what’s the use case or benefit of having
the existing implementation? I didn’t understand why it was that way in the
first place so clearly am missing something.

Also can you advise on the name of the flag, default, etc.

ivanpopelyshev:
Actually, I dont know :) the change was introduced by @GoodBoyDigital long time ago.

GoodBoyDigital:
this is an interesting one. 

**texImage2D:** This function is used to define a two-dimensional texture image. It can initialize a texture from an array of data or from an HTML element (like , <canvas>, or <video>). When you call texImage2D, you're essentially creating a new texture or reinitializing an existing one, which includes allocating memory for the texture. If you are repeatedly calling texImage2D to update a texture with new images or data, it can be less efficient because it involves reallocating texture memory each time it's called.

**texSubImage2D:** This function updates a subset of a two-dimensional texture image, and it can be more efficient for updating textures because it does not require the texture's memory to be reallocated. Instead, it reuses the existing texture's memory allocation. This makes texSubImage2D particularly useful for scenarios where you need to frequently update textures with new data (like streaming video frames into a texture or updating a texture atlas), as it can offer better performance by avoiding unnecessary memory allocations.

on paper we should always strive for texSubImage2D.. but seems that might not be entirely true now?

gfodor:
What are those descriptions from? It’s definitely the case, if the list
serv is to be believed, that in chrome at least the fastest path for video
is hit with texImage2D only

GoodBoyDigital:
i think @ivanpopelyshev has the best idea - lets add a flag to texture source. False by default, but make it true for video textures?
that way users have the options.

@gfodor would you mind updating the PR pls? Thanks man!!

gfodor:
Sure, I can take a spin at that.

GoodBoyDigital:
following up on this one @gfodor - would be great to have this when you get some time! Thanks in advance

hlevring:
@gfodor any chance you can help get this landed?

GoodBoyDigital:
poke @gfodor  - we need you :)

gfodor:
Yikes, sorry I dropped the ball here. I am not working with pixi.js anymore so this slipped by, but can take a look ASAP.

GoodBoyDigital:
thanks for the original @gfodor ! closing in favour of https://github.com/pixijs/pixijs/pull/11867

PR #10387, opened 2024-03-28
BugFix: [v8] The filter is not taking the RenderTexture's frame's offset into account when being applied

### LINKED ISSUE #10368: Bug: [v8] The filter is not taking the RenderTexture's frame's offset into account when being applied
### Current Behavior

Filters are not applied

### Expected Behavior

Filters are applied

### Steps to Reproduce

Playground: https://www.pixiplayground.com/#/edit/cT9UYLBBBtDjoaS5YySA7
As you can see, filter is applied only to the first RenderTexture.

### Environment

- **`pixi.js` version**: *8.0.3*
- **Browser & Version**: *Chrome 122*

### Possible Solution

Temporary workaround: You might be able to bypass this issue by using a single RenderTexture instead of creating multiple from the same source, then setting the transform of the desired DisplayObject you're rendering to it instead of using a new frame.

### Additional Information

The same code worked fine in v7, filters were applied correctly. Issue description and suggested workaround by LunarRaid on Discord.

GoodBoyDigital:
fixes #10368

visual test added also.


ivanpopelyshev (approved):

src/scene/container/bounds/Bounds.ts
```diff
@@ -253,7 +253,7 @@ export class Bounds
      * Resizes the bounds object to include the given rectangle.
      * @param rect - The rectangle to be included.
      */
-    public fit(rect: Rectangle): this
+    public fitRect(rect: Rectangle): this
```
Zyie, line 256: This seems like a needless breaking API change

Can we not rename the other method instead

Zyie (changes requested):

src/scene/container/bounds/Bounds.ts
ivanpopelyshev: agree

GoodBoyDigital: good shout!

Zyie (approved):

PR #10389, opened 2024-03-28
Fix: roundPixels offset 

### LINKED ISSUE #10373: Bug: roundPixels=true causing precision issues
### Current Behavior

Setting roundPixels=true in the Application settings causes rounding issues that are especially apparent at low resolutions. For example, a 16x16 sprite will be rendered at 16x15, or even 15x15 depending on where it's located, causing aliasing artifacts. This problem disappears if the x and y coordinates of the sprite fall on fractional values. The problem also goes away if you use power-of-two dimensions for the renderer, like if you were to set the width and height to 64x64 in the example below.

### Expected Behavior

The sprite in the provided example should appear without aliasing when roundPixels=true.

### Steps to Reproduce

[A very basic example](https://pixiplayground.com/#/edit/ExyVqXY5amHUYt8Vnq9y0)

Change roundPixels to false, set the width and height to 64x64, or set fractional values on the sprite's x and y values to see the difference.

### Environment

- **`pixi.js` version**: 8.0.4
- **Browser & Version**: Edge 122.0.2365.92

### Possible Solution

[src/rendering/high-shader/shader-bits/roundPixelsBit.ts](https://github.com/pixijs/pixijs/blob/dev/src/rendering/high-shader/shader-bits/roundPixelsBit.ts)

Changing the `floor` operators to `round` fixes the issue for me.

GoodBoyDigital:
ensure pixels are rounded and not floored. 

fixes #10373

Clue was in the variable name all along :)

Thanks @FunFetched for the suggestion!

### DIFF
--- a/src/rendering/high-shader/shader-bits/roundPixelsBit.ts
+++ b/src/rendering/high-shader/shader-bits/roundPixelsBit.ts
@@ -4,7 +4,7 @@ export const roundPixelsBit = {
         header: /* wgsl */`
             fn roundPixels(position: vec2<f32>, targetSize: vec2<f32>) -> vec2<f32> 
             {
-                return (floor((position * 0.5 + 0.5) * targetSize) / targetSize) * 2.0 - 1.0;
+                return (round((position * 0.5 + 0.5) * targetSize) / targetSize) * 2.0 - 1.0;
             }
         `
     }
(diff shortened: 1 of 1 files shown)


ivanpopelyshev (changes requested):
Round is bad, do not use round: https://www.w3.org/TR/WGSL/#round-builtin

`Floor(x + 0.5)` is better,

Consider you have sprite from 3.5 to 8.5 - it gets rounded to "4 to 8" which is a pixel smaller. In case of half-pixel we have to round everything to the same side!

Btw v7 had a bug where we used `~~x` for flooring - and it floors values differently for "x>0" and "x<0", so if you ever see float->int conversion for rounding pixels - please get rid of it , use `floor`.

FunFetched:
I did just run into an edge case where the issue still appears to be happening, though not nearly as often. It seems the actual root of the problem lies further up the chain somewhere. Perhaps we're missing a Math.floor somewhere.

GoodBoyDigital:
> Floor(x + 0.5)

great feedback! thanks - will update PR

ivanpopelyshev (approved):

PR #10395, opened 2024-03-30
Fix: NoiseFilter ignoring uniforms

codedpalette:
##### Description of change
NoiseFilter seed and scale uniforms are being ignored in the fragment shader code, it uses constants instead. This change is based on the fragment shader code in pixi v7.

##### Pre-Merge Checklist


GoodBoyDigital:
good catch @codedpalette !

GoodBoyDigital (approved):

Issue #10391, opened 2024-03-29
Bug: crash when WebGL2RenderingContext is not available


elsassph:
### Current Behavior

PixiJS 8 can't start when running in an older browser without webGl 2.0 support: 
```
Uncaught (in promise) ReferenceError: WebGL2RenderingContext is not defined
```

### Expected Behavior

I make apps for connected TVs - a lot of models still don't support webGl 2.0. Some even don't support webGl at all.

It would be great if Pixi 8 would gracefully support older browsers supporting only webGl 1.0.

### Steps to Reproduce

Run any Pixi app on an TV 3+ years old.

### Environment

- **`pixi.js` version**: 8.0
- **Browser & Version**: Chrome/53.0.2785.34 Safari/537.36
- **OS & Version**: webOS 4.10

### Possible Solution

I tried declaring a mock `WebGL2RenderingContext` class (`undefined` also fails further during init) to pass the webGl detection but then gl errors happen and nothing renders:
```
PixiJS Warning:  Provided WebGL context does not support 32 index buffer, large scenes may not render correctly
WebGL: INVALID_ENUM: activeTexture: texture unit out of range
[.Offscreen-For-WebGL-0x123eb0]GL ERROR :GL_INVALID_VALUE : glUniform1iv: texture unit out of range
```

Line of `activeTexture` error is: https://github.com/pixijs/pixijs/blob/dev/src/rendering/renderers/gl/texture/GlTextureSystem.ts#L190

bigtimebuddy:
@GoodBoyDigital you added support for WebGL1 in v8, right?

elsassph:
@bigtimebuddy I see there seem to be support for webGl 1, but it crashes on systems without support for webGl 2.

bigtimebuddy:
Yeah, I gathered that. If I had to guess, we added the internal supported for WebGL1, but probably didn't actually test running in a WebGL1-only environment.

Zyie:
hey @elsassph what version of v8 are you using?

i believe this #10351 should fix the issue and was released in 8.0.4

erikdubbelboer:
@Zyie the issue is still present in 8.0.4. See: https://github.com/pixijs/pixijs/issues/10401

Zyie:
Yeah this is definitely still an issue, we will look into it!

mwx0621:
[...]

PR #10410, opened 2024-04-02
Fix: webgl-1 support

GoodBoyDigital:
fixes an issue where we were looking for a WebGL2Context even if it does not exist.

This PR removes the `getWebGL2RenderingContext`. It then renames `getWebGLRenderingContext` -> `getWebGL1RenderingContext` 
With that instead of checking if somthign is WebGL2, we can check if is not WebGL1.


GoodBoyDigital:
fixes #10391

PR #10430, opened 2024-04-09
Fix: incorrect cache warning check

Zyie:
We were checking the wrong value when testing if an asset is different, resulting in a bunch of warnings for things like spritesheets


bigtimebuddy (approved):

GoodBoyDigital (approved):
nice!!

PR #10469, opened 2024-04-23
Fix: overlay blend mode to be consistent with photoshop

skyfish-qc:
fix the overlay blend mode

The overlay mode uses base as the judgment condition. After modification, the display is consistent with Photoshop.


GoodBoyDigital (approved):
thanks for the fix!

PR #10470, opened 2024-04-23
Update LinearLightBlend.ts

skyfish-qc:
fix linearlight blend mode to consistent with Photoshop
The linear light mode is base+2*top-1. After modification, the display is consistent with that of Photoshop.

### DIFF
--- a/src/advanced-blend-modes/LinearLightBlend.ts
+++ b/src/advanced-blend-modes/LinearLightBlend.ts
@@ -38,7 +38,7 @@ export class LinearLightBlend extends BlendModeFilter
                 }
 
                 float linearLight(float base, float blend) {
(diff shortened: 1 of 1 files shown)


skyfish-qc (commented):
The linear light mode is base+2*top-1. After modification, the display is consistent with that of Photoshop.

bigtimebuddy:
Could you show an image preview of this change?

GoodBoyDigital:
> Could you show an image preview of this change?

that would be great!

skyfish-qc:
(blend <= 0.5) ? linearBurn(base,2.0*blend) : linearBurn(base,2.0*(blend-0.5));

GoodBoyDigital:
closing for now - technically our current solution is correct. more info here as around the propsed solution would be great. Thanks

PR #10486, opened 2024-04-29
Fix: bind to default framebuffer only when copying from main canvas

ShukantPal:
The render target system's copyToTexture assumes that if the source render surface is a canvas, then it must be the main framebuffer. But this is false if someone is trying to copy from a secondary or offscreen canvas.

This change uses a direct comparison with the main canvas to check whether to use the default framebuffer instead.

For a bug repro, see this JSFiddle
https://jsfiddle.net/ShukantPal/vwrh51qs/27/

I've verified locally that this fix causes the gradient to paint correctly. @bigtimebuddy Do we still deploy a public build URL for GitHub branches? (so that I can create a fiddle that works correctly using this branch's build)

##### Pre-Merge Checklist


GoodBoyDigital (approved):
nice!

bigtimebuddy:
Try using the codesandbox link above for testing.

Zyie (approved):
I've verified that the demo works!

PR #10489, opened 2024-04-29
Fix gha smells

ceddy4395:
##### Description of change

Hey! 🙂
I want to contribute the following changes to your workflow:

- Stop running workflows when there is a newer commit in branch
- Avoid starting new workflow whilst the previous one is still running
- Stop running workflows when there is a newer commit in PR
- Use commit hash instead of tags for action versions

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

##### Pre-Merge Checklist

### DIFF
--- a/.github/workflows/pull-unit.yml
+++ b/.github/workflows/pull-unit.yml
@@ -4,12 +4,17 @@ on:
   pull_request:
     branches: [ '**' ]
 
+concurrency: 
+  group: ${{github.workflow}} - ${{github.ref}}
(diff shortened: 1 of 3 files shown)


Zyie:
Thanks for putting this together, but we have decided to go in a slightly different direction and not use the commit hashes
However the concurrency change is a great addition and will be merged in here #10689 instead

Issue #10377, opened 2024-03-26
Performance: [V8] Container.removeChildren performance issue


XiBOR:
### Current Behavior

When deleting a part of a container hierarchy, a recusive search of all elements of that hierarchy is performed, deleting them one after another. This causes a huge performance sag at that point (on the order of hundreds of milliseconds)
This worked much more efficiently in version 7.

### Expected Behavior

I would like the deletion to happen within milliseconds like it did in the pixi 7.

### Steps to Reproduce

Repeat example (with minor edits) on version 7, and get a response of 0 milliseconds

### Environment

- **`pixi.js` version**: *8.0.4*
- **Browser & Version**: *Firefox 125.0b4 (64-bit), Chrome 123.0.6312.59*
- **OS & Version**: *macOS 14.4.1*
- **Running Example**:
```
(async () =>
{
    // Create a new application
    const app = new PIXI.Application();

    // Initialize the application
    await app.init({ background: '#1099bb', resizeTo: window });

    // Append the application canvas to the document body
    document.body.appendChild(app.canvas);

    // Create and add a container to the stage
    const container = new PIXI.Container();

    app.stage.addChild(container);

    // Load the bunny texture
    const texture = await PIXI.Assets.load('https://pixijs.com/assets/bunny.png');

    // Create a 5x5 grid of bunnies in the container
    for (let i = 0; i < 25000; i++)
    {
        const bunny = new PIXI.Sprite(texture);

        bunny.x = (i % 5) * 40;
        bunny.y = Math.floor(i / 5) * 40;
        container.addChild(bunny);
    }

    // Move the container to the center
    container.x = app.screen.width / 2;
    container.y = app.screen.height / 2;

    // Center the bunny sprites in local container coordinates
    container.pivot.x = container.width / 2;
    container.pivot.y = container.height / 2;

    // Listen for animate update
    app.ticker.add((time) =>
    {
        // Continuously rotate the container!
        // * use delta to create frame-independent transform *
        container.rotation -= 0.01 * time.deltaTime;
    });
  
    const start = performance.now()
    app.stage.removeChildren()
    const time = performance.now() - start
    console.log(time)
})();
```

### Possible Solution

If I found correctly, at the point of removal you can see how the approach has changed from removing one pack to removing one after another (with the whole tail shifting)

[...]

PR #10505, opened 2024-05-02
Fix: destroy and removeChild performance

### LINKED ISSUE #10345: Performance: [v8 regression] `Container.destroy()` is very slow on crowded containers
### Current Behavior

Admired developers,

I frequently have to clear crowded containers (>100000 children) and observed that `Container.destroy(true)` can take several seconds. 
Compared to my v7 code, where `destroy()` wasn't noticable at all (using `ParticleContainer`, to be fair).

Clearing the container 'by hand' before calling `destroy()` speeds up the process significantly, which makes me think it might be a bug.

```
myContainer.removeChildren().forEach(child => child.destroy(true));
``` 

### Expected Behavior

`destroy()` shouldn't be slower than clearing a `Container` by hand.

As fast as `ParticleContainer` would be great :).

### Steps to Reproduce

Use playground https://www.pixiplayground.com/#/edit/zFuIcJIth6M6QmC2IAjR9

After 200 frames, the content will be destroyed and starting to reconstruct. 
`largeContainer.destroy(true);` may halt rendering for a couple of seconds.

Setting `MAKE_IT_FAST_AGAIN=true` clears `largeContainer` manually before destruction, bringing up speed again, 

### Environment

- **`pixi.js` version**: 8.0.2
- **Browser & Version**: any
- **OS & Version**: any
- **Running Example**: https://www.pixiplayground.com/#/edit/zFuIcJIth6M6QmC2IAjR9

### Possible Solution

There is a workaround by clearing the container upfront `destroy()`:

```
myContainer.removeChildren().forEach(child => child.destroy(true));
```

### Additional Information

_No response_

GoodBoyDigital:
##### Description of change
The way rendergroup was being stored kind of bugged me as it was not as intuative as it could be. Consequentially, this meant its a bit harder to understand some of the checks that happen in the code around render groups.

Essentially the `renderGroup` property on `Container` could either be the Containers owned render group or its parent render group. Which is kind of weird!

I have modified this so we now have 
- `renderGroup` which is the the containers renderGroup created when `enableRenderGroup` is called
- `parentRenderGroup` a new property which is the render group that is responsible for rendering the container. this is set when a container is added to a rendergroup.

Modifications to RenderGroup
- i remove the `_children` property on renderGroup - we don't actually use it anywhere, so not point pushing and splicing it!
- added `getChildren` which will return them all should someone want to get the list of all renderables in a render group.
- tidied and streamlined some of the functionality of add / removing from a renderGroup to be a bit faster and easier to understand.
- I also added a removeChildren function to renderGroup.

Benchmark:
removeChildren / destroy is now 1000x faster in the demo from this issue: https://github.com/pixijs/pixijs/issues/10345

Will update a demo once a branch is built.

fixes #10345 and #10377

##### Pre-Merge Checklist


src/scene/container/container-mixins/effectsMixin.ts
```diff
@@ -54,11 +54,18 @@ export const effectsMixin: Partial<Container> = {
...
         {
-            this.renderGroup.structureDidChange = true;
+            renderGroup.structureDidChange = true;
         }
 
+        // if (this.renderGroup)
+        // {
+        //     this.renderGroup.structureDidChange = true;
+        // }
+
```
Zyie, line 68: remove?

tests/scene/scene.tests.ts
```diff
@@ -202,15 +243,20 @@ describe('Scene', () =>
...
-        expect(container.renderGroup['_children']).toHaveLength(1);
-        expect(container2.renderGroup['_children']).toHaveLength(3);
+        expect(container.renderGroup.getChildren()).toHaveLength(1);
+        expect(container2.renderGroup.getChildren()).toHaveLength(3);
 
         expect(container.renderGroup).toEqual(container.renderGroup);
-        expect(container2.renderGroup).toEqual(container2.renderGroup);
-        expect(child.renderGroup).toEqual(container2.renderGroup);
+        expect(container2.parentRenderGroup).toEqual(container.renderGroup);
+        // expect(child.renderGroup).toEqual(container2.renderGroup);
```
Zyie, line 251: is this meant to be commented out?

tests/scene/scene.tests.ts
```diff
@@ -223,15 +269,15 @@ describe('Scene', () =>
...
-        expect(container.renderGroup['_children']).toHaveLength(2);
-        expect(container2.renderGroup['_children']).toHaveLength(2);
+        expect(container.renderGroup.getChildren()).toHaveLength(2);
+        expect(container2.renderGroup.getChildren()).toHaveLength(2);
 
         expect(container.renderGroup).toEqual(container.renderGroup);
-        expect(container2.renderGroup).toEqual(container2.renderGroup);
-        expect(child.renderGroup).toEqual(container2.renderGroup);
+        expect(container2.parentRenderGroup).toEqual(container.renderGroup);
+        // expect(child.renderGroup).toEqual(container2.renderGroup);
```
Zyie, line 277: and this one

tests/scene/scene.tests.ts
```diff
@@ -571,22 +574,15 @@ describe('Scene', () =>
...
 
         container.removeChild(container2);
 
-        expect(container.renderGroup.childrenToUpdate[1]).toEqual({
-            list: [child],
-            index: 1
-        });
+        compareUpdateList([child], container.renderGroup, 1);
+
+        expect(true).toBeTrue();
```
Zyie, line 585: this seems wrong

tests/scene/scene.tests.ts
```diff
@@ -502,27 +523,17 @@ describe('Scene', () =>
...
-        });
+        compareUpdateList([child], container.renderGroup, 1);
 
-        expect(container2.renderGroup.childrenToUpdate[1]).toEqual({
-            list: [child3],
-            index: 1
-        });
+        compareUpdateList([child3], container2.renderGroup, 1);
+
+        expect(true).toBeTrue();
```
Zyie, line 536: this seems wrong

tests/scene/scene.tests.ts
```diff
@@ -367,41 +412,21 @@ describe('Scene', () =>
...
-        });
+        compareUpdateList([child, container2], container.renderGroup, 1);
 
-        // child2.x = 50;
+        compareUpdateList([child2, child3], container.renderGroup, 2);
 
-        // expect(container.layerGroup.childrenToUpdate[2]).toEqual({
-        //     list: [child3, child2],
-        //     index: 2
+        expect(true).toBeTrue();
```
Zyie, line 429: ?

tests/scene/scene.tests.ts
```diff
@@ -446,24 +467,24 @@ describe('Scene', () =>
...
-        });
+        compareUpdateList([], container.renderGroup, 1);
 
-        expect(container2.renderGroup.childrenToUpdate[2]).toEqual({
-            list: [child],
-            index: 1
-        });
+        compareUpdateList([child], container2.renderGroup, 2);
+
+        expect(true).toBeTrue();
```
Zyie, line 474: ?

tests/scene/scene.tests.ts
GoodBoyDigital: i know this looks a bit weird - its because the the lint rule specifies that there must be an expect function in the test. The expect is contained in the compareUpdateList function. I will disable the lint rule instead as that is a bit more reasonable!

tests/scene/scene.tests.ts
GoodBoyDigital: solved by disabling the lint rule

tests/scene/scene.tests.ts
GoodBoyDigital: solved by disabling the lint rule

tests/scene/scene.tests.ts
GoodBoyDigital: removed!

Zyie (approved):

PR #10507, opened 2024-05-04
Fix: handling of getRenderTarget when passing in a canvas

ShukantPal:
##### Description of change

Right now, `getRenderTarget(canvas)` returns null because of how the canvas is converted into a `TextureSource`.

In the code, `getCanvasTexture` returns a texture, but the 2 if conditions in `getRenderTarget` only handle render targets and texture sources.

This change updates it so that a canvas is converted into a canvas source, not a texture.


GoodBoyDigital:
nice catch @ShukantPal ! 

would you mind adding a unit test pls?

GoodBoyDigital (approved):

ShukantPal:
@GoodBoyDigital Done!

PR #10508, opened 2024-05-05
Export shapeBuildCommandMap to allow custom shape primitives

ShukantPal:
This is affecting the upgrade of @pixi-essentials/svg to v8 (specifically, for rendering `<path>` elements). :)

### DIFF
--- a/src/scene/graphics/shared/utils/buildContextBatches.ts
+++ b/src/scene/graphics/shared/utils/buildContextBatches.ts
@@ -18,7 +18,7 @@ import type { GpuGraphicsContext } from '../GraphicsContextSystem';
 import type { GraphicsPath } from '../path/GraphicsPath';
 import type { ShapePath } from '../path/ShapePath';
 
-const buildMap: Record<string, ShapeBuildCommand> = {
+export const shapeBuildCommandMap: Record<string, ShapeBuildCommand> = {
     rectangle: buildRectangle,
     polygon: buildPolygon,
     triangle: buildTriangle,
@@ -86,7 +86,7 @@ function addTextureToGeometryData(
 
     const points: number[] = [];
 
-    const build = buildMap.rectangle;
+    const build = shapeBuildCommandMap.rectangle;
 
     const rect = tempRect;
 
@@ -158,7 +158,7 @@ function addShapePathToGeometryData(
 
         const points: number[] = [];
 
-        const build = buildMap[shape.type];
+        const build = shapeBuildCommandMap[shape.type];
 
         // TODO - this can be cached...
         // TODO - THIS IS DONE TWICE!!!!!!
@@ -261,7 +261,7 @@ function getHoleArrays(shape: ShapePath)
         // TODO - need to transform the points via there transform here..
         const holePoints: number[] = [];
 
-        const holeBuilder = buildMap[holePrimitive.type] as ShapeBuildCommand;
+        const holeBuilder = shapeBuildCommandMap[holePrimitive.type] as ShapeBuildCommand;
 
         holeBuilder.build(holePrimitive, holePoints);
 


src/scene/graphics/shared/utils/buildContextBatches.ts
```diff
@@ -18,7 +18,7 @@ import type { GpuGraphicsContext } from '../GraphicsContextSystem';
 import type { GraphicsPath } from '../path/GraphicsPath';
 import type { ShapePath } from '../path/ShapePath';
 
-const buildMap: Record<string, ShapeBuildCommand> = {
+export const shapeBuildCommandMap: Record<string, ShapeBuildCommand> = {
```
bigtimebuddy, line 21: Minor nit: not crazy about the name. I think "Map" specially isn't helpful. I kinda don't like referring to objects informally as "map" when there's a JS Map object. Some alts: "shapeCommands", "shapeBuilders", "buildShapeFuncs"

bigtimebuddy (commented):
Do you thinks this could be another extension type? I like using extensions for any potential extensibility (renderer systems, asset parsers, application plugins), why not shapes?

src/scene/graphics/shared/utils/buildContextBatches.ts
ShukantPal: Could you elaborate on extension type? The `ShapeBuildCommand` is kind of like an extension 🤔 

I've renamed it to `shapeBuilders` 👍

bigtimebuddy: There's a [enum](https://github.com/pixijs/pixijs/blob/dev/src/extensions/Extensions.ts#L40) of extensions that you can add a new entry. This requires two things: 1) handling extensions by type and 2) registering extensions

For example, Application supports extensions and this is how they are added:
https://github.com/pixijs/pixijs/blob/78824e43fe569fdf84206905e5e2c1ce9d9475c3/src/app/Application.ts#L227
There's also a `handleByMap` method.

And here's where the resize and ticker plugins are added to Application:
https://github.com/pixijs/pixijs/blob/78824e43fe569fdf84206905e5e2c1ce9d9475c3/src/app/init.ts#L5-L6

GoodBoyDigital (approved):
lgtm!

src/scene/graphics/shared/utils/buildContextBatches.ts
ShukantPal: @bigtimebuddy Added!

ShukantPal:
@GoodBoyDigital Is there a timeline for the next patch/minor release?

src/scene/graphics/shared/utils/buildContextBatches.ts
```diff
@@ -18,7 +19,7 @@ import type { GpuGraphicsContext } from '../GraphicsContextSystem';
 import type { GraphicsPath } from '../path/GraphicsPath';
 import type { ShapePath } from '../path/ShapePath';
 
-const buildMap: Record<string, ShapeBuildCommand> = {
+export const shapeBuilders: Record<string, ShapeBuildCommand> = {
```
bigtimebuddy, line 22: Could you register the shape extensions (extensions.add) instead of adding them here? Kinda defeats the purpose of making them extensions if you manually add them anyways.

bigtimebuddy (changes requested):

src/scene/graphics/shared/utils/buildContextBatches.ts
Zyie: I've updated the PR to make this change!

bigtimebuddy (approved):
Nice! Thanks.

bigtimebuddy:
@ShukantPal this is merged into dev pending a new minor release. In the meantime, you can test with the published version on npm: `8.1.8-dev.0ff2c97`

ShukantPal:
FYI I see this got into 8.2.0 https://github.com/pixijs/pixijs/blob/v8.2.0/src/scene/graphics/shared/utils/buildContextBatches.ts

But it's not mentioned in the release notes https://github.com/pixijs/pixijs/releases/tag/v8.2.0

Zyie:
> FYI I see this got into 8.2.0 [`v8.2.0`/src/scene/graphics/shared/utils/buildContextBatches.ts](https://github.com/pixijs/pixijs/blob/v8.2.0/src/scene/graphics/shared/utils/buildContextBatches.ts?rgh-link-date=2024-07-01T03%3A46%3A32Z)
> 
> But it's not mentioned in the release notes [`v8.2.0` (release)](https://github.com/pixijs/pixijs/releases/tag/v8.2.0)

I've updated the release notes, looks like the automatic generation missed this one

PR #10520, opened 2024-05-08
feat: add automated published branch releases for dev & main

danielbarion:
## Description of change

I recently implemented a beta-release workflow on the react-tooltip project, maybe it can be interesting for Pixi.js project too.

On ReactTooltip project, we always try to release a beta version on some PR before merging it to let the owner of the issue test and confirm if the issue was solved, this also unblocks projects that are waiting for some fix before we release an official version with the fix.

With the beta release workflow, all commits into the PR will trigger the beta release.

* The beta release always releases a new beta version, following this format: 
`pixi.js@{__VERSION__}-beta.{__PR_NUMBER__}.rc.{__BETA_VERSION__}` -> `pixi.js@v0.0.11-beta.1.rc.1`
`__VERSION__` - current package.json version on this PR
`__PR_NUMBER__` - current PR number
`__BETA_VERSION__` - latest beta version released
* to skip the workflow we need to add the following words to the commit message:
```
[skip ci]
[ci skip]
[no ci]
[skip actions]
[actions skip]
```

https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

### Examples

https://github.com/ReactTooltip/react-tooltip/pull/1179#issuecomment-2101225714

image

https://github.com/ReactTooltip/react-tooltip/actions/runs/9007001657/job/24745795089?pr=1179

The previous beta version is `null` because that's the first beta release for this PR number.
image

https://www.npmjs.com/package/react-tooltip?activeTab=versions

image

Note: ReactTooltip master branch has the CJS code (same as this PR), on V6 branch we have it as ESM version of the code.


bigtimebuddy:
@danielbarion We (@Zyie @GoodBoyDigital and I) had a change to review this change this morning. Overall we love the general idea and thank you for proposing it! Super useful feature to opt-in to bleeding-edge unreleased changes. 

@Zyie pointed out that our codesandbox integration offers some of this (at least testing purposes) for PRs and there is also a mechanism to download a build (I think) using that tool.

That said, we still think there's something here and we have some feedback. If you don't have time to work on these, just let us know someone on the team will likely pick it up.

* Rather than focus on PRs, would like to provide automated releases for _unreleased but merged_ functionality. It's common feedback we get that sometimes users have to wait awhile for a release. This would help elevate that pressure.
* We'd like to not use the term "beta" here for these release, we use "beta" semver for specific, intentional tagged releases.
* Preference would be to have versions formatted as: `8.1.0-[branch].[commit-hash]`, for instance `8.1.3-dev.9afd9d3`. The benefit of doing this is that versions would tie directly to the branch and commit. That's super helpful for debugging and triaging errors.
* Preference for npm dist-tags to use the branch name, e.g., `dev`, that way users could define `"pixi.js": "dev"` in their dependencies or `npm install pixi.js@dev`
* We would only like to focus on `main` and `dev` branches as these are the only branches we cut releases from.

What do you think?

danielbarion:
Hey @bigtimebuddy & team, the initial implementation is the same as we have on ReactTooltip that matches our needs.

Thank you so much for reviewing this suggestion and the detailed feedback, I really appreciate that!

I would like to focus on those changes so the Pixi team can concentrate on other stuff, I have a test repository where I developed this workflow initially to not release the test versions directly on the ReactTooltip project, I'll use it to apply the needs mentioned on your feedback.

I'll keep you guys in the loop.

Thanks!

bigtimebuddy:
Maybe, `Publish Branch`?

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+            #         \`\`\`
+            #         yarn add ${{ env.PACKAGE_NAME}}@${{ env.NEW_VERSION }}
+            #         \`\`\`
+            #         or
+            #         \`\`\`
+            #         npm install ${{ env.PACKAGE_NAME}}@${{ env.NEW_VERSION }}
+            #         \`\`\`
+            #         `
+            #       })
```
danielbarion, line 77: The commented code on this workflow can be used to add a comment on some discussion or PR (currently is implemented for PRs), but please feel free to remove it if not needed

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+            # get the package name so the workflow can be reusable between projects
+            # - name: Get current package.json name
+            #   run: echo "PACKAGE_NAME=$(npm pkg get name | tr -d '"')" >> $GITHUB_ENV
+
+            - name: Setup Branch Release Candidate Version
+              run: echo "NEW_VERSION=$PACKAGE_VERSION-$BRANCH_NAME.$SHORT_SHA" >> $GITHUB_ENV
+
+            - name: Update Package.json Version with Release Candidate Version
+              run: node ./update-branch-version.js --version $NEW_VERSION
```
danielbarion, line 52: I tried a lot to discard the need for this JS file, I did the entire logic on the workflow, but when triggered we received the following error:

image

I tried 16 possible changes to debug and fix it and it doesn't work as expected, but adding it to a JS file and running it through the node command worked as expected

danielbarion:
image

image

https://www.npmjs.com/package/auto-beta-release?activeTab=versions


The code was updated, please let me know if something else is needed, I didn't update the PR description because of our comments, but please let me know your thoughts about it and about the changes

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,77 @@
...
+            BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.head_ref }}
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: "16.x"
+                  registry-url: "https://registry.npmjs.org"
+                  cache: "yarn"
```
bigtimebuddy, line 24: We use npm on the project, not yarn.
```suggestion
                  cache: "npm"
```

.github/workflows/publish-branch.yaml
bigtimebuddy: Yeah I think this is okay to remove.

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,77 @@
...
+            NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+            NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+            BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.head_ref }}
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: "16.x"
```
bigtimebuddy, line 22: Can we do 18? Or better yet use the shared ./.github/actions/setup

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,77 @@
...
+                  registry-url: "https://registry.npmjs.org"
+                  cache: "yarn"
+
+            - name: Install dev dependencies
+              run: yarn install
```
bigtimebuddy, line 27:
```suggestion
              run: npm ci
```

.github/workflows/publish-branch.yaml
bigtimebuddy: This seems very complicated. Why aren't you just doing `run: npm version $NEW_VERSION`?

update-branch-version.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+/**
+ * @NOTE: This can't be done by only using the workflow, because when we try to update
+ * the version to the desired version, we get an error about wrong version format,
+ * but directly on CI it works as expected.
+ */
+const UpdateBranchVersion = async () => {
+    // update the release candidate version on package.json
+    const { error } = await runCommand(`npm version ${version}`);
```
bigtimebuddy, line 26: Preference is to do this in the workflow rather than add a script that basically just invokes one cli command.

.github/workflows/publish-branch.yaml
danielbarion: Because it didn't work, I moved the entire logic to the workflow (because is simple now), but the Action was not able to trigger the npm version command on this format, not even hard coded

.github/workflows/publish-branch.yaml
danielbarion: Not tested, but applied

bigtimebuddy:
@danielbarion I did a quick experiment here that seemed to work for me:

https://github.com/bigtimebuddy/actions-branch-version/blob/a68fa58bdbc26d9e7ef8e662f78aff09aab89a62/.github/workflows/main.yml#L34

Added two args to `npm version`:
* `--no-git-tag-version` - Ignores git tagging which means that it doesn't require git auth/identity
* `--force` - Bump the version even if the git workspace is dirty or has uncommitted changed, this wasn't necessary, but will help prevent issues with, say, npm updates that change the package-lock.json

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+            - name: Get current package.json version
+              run: echo "PACKAGE_VERSION=$(npm pkg get version | tr -d '"')" >> $GITHUB_ENV
+
+            # get the sort SHA and add it into the environment variables
+            - name: Get short SHA
+              run: echo "SHORT_SHA=${GITHUB_SHA::7}" >> $GITHUB_ENV
+
+            - name: Setup Branch Release Candidate Version
+              run: echo "NEW_VERSION=$PACKAGE_VERSION-$BRANCH_NAME.$SHORT_SHA" >> $GITHUB_ENV
```
bigtimebuddy, line 35: I think this can be simplified.
```suggestion
            - name: Setup Branch Release Candidate Version
              run: echo "NEW_VERSION=$PACKAGE_VERSION-$BRANCH_NAME.${GITHUB_SHA::7}" >> $GITHUB_ENV
```

danielbarion:
@bigtimebuddy hmm, interesting, maybe it's the action I was using? I saw the project is using v3 and I was using v4 initially (`uses: actions/setup-node@v3`), or maybe the setup I did previously.

Thanks for letting me know

.github/workflows/publish-branch.yaml
danielbarion: I do agree

danielbarion:
Sorry, I asked review again but I forgot to apply the change, 1 second.

danielbarion:
Updated

bigtimebuddy (commented):
Could you revert the package.json and package-lock and I think we are good to go!

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,40 @@
...
+            - name: Setup Branch Release Candidate Version
+              run: echo "BRANCH_VERSION=$PACKAGE_VERSION-$BRANCH_NAME.${GITHUB_SHA::7}" >> $GITHUB_ENV
+
+            - name: Bump version
+              run: npm version $BRANCH_VERSION --no-git-tag-version --force
+
+            - name: Publish a new branch release candidate version
+              run: npm publish --tag $BRANCH_NAME
+              env:
+                  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
bigtimebuddy, line 40: The `env.NPM_TOKEN` is defined at the top-level of this job, you don't need to define it again here

danielbarion:
@bigtimebuddy NPM sorted the package by itself, should I remove it too?

bigtimebuddy (approved):
Looks good, thank you!

.github/workflows/publish-branch.yaml
```diff
@@ -0,0 +1,38 @@
...
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.head_ref }}
+
+            - name: Setup Project
+              uses: ./.github/actions/setup
+
+            - name: Setup npm credentials file
+              run: echo "//registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN" >> .npmrc
```
bigtimebuddy, line 25: One small thing, I don't believe we need this step. The NODE_AUTH_TOKEN being set in the environment should be sufficient here.

bigtimebuddy (commented):
One small thing.

.github/workflows/publish-branch.yaml
danielbarion: I'll remove it :) 

About the build, on ReactTooltip we have this script on our package.json: `"prepublishOnly": "npm run build"`.

But looks like Pixi.js doesn't have it, do you prefer to add it to the package.json or in the workflow?

bigtimebuddy: I'd say add it to the workflow or else we'd need to fix it double-building on the other workflows. We can clean that up later.

danielbarion: ok, I'll add it again.

danielbarion: Ok, I'll add it into the workflow

Zyie:
@bigtimebuddy @danielbarion is everything all settled with this PR now?

danielbarion:
yes @Zyie

bigtimebuddy:
Yeah, I'm good on this. I think we should merge this into dev and just validate that everyone is working as expected.

danielbarion:
@Zyie @bigtimebuddy 

image

Zyie:
Ok this seems to be working!

https://www.npmjs.com/package/pixi.js/v/8.1.5-dev.141f7e0?activeTab=versions

Thank you @danielbarion for adding this

danielbarion:
@Zyie do you want me to do the PR to main?

Zyie:
> @Zyie do you want me to do the PR to main?

it's cool, this will get merged into main when i do the next release

PR #10536, opened 2024-05-15
fix mask issue when setting to null and back

GoodBoyDigital:
##### Description of change
Fixes a little regression bug when doing the following:

```
 thing.mask = mask;
 thing.mask = null;
 thing.mask = mask; // would throw as mask effect refference was not removed.
```

##### Pre-Merge Checklist

### DIFF
--- a/tests/scene/mask.test.ts
+++ b/tests/scene/mask.test.ts
@@ -1,0 +1,27 @@
+import { Container } from '../../src/scene/container/Container';
+import { Graphics } from '../../src/scene/graphics/shared/Graphics';
+
+describe('Mask Tests', () =>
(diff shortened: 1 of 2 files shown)


tests/scene/mask.test.ts
```diff
@@ -0,0 +1,27 @@
...
+        expect(() =>
+        {
+            container.mask = mask;
+            container.mask = null;
+            container.mask = mask;
+        }).not.toThrow();
+    });
+});
+
+// Test to cover
```
Zyie, line 27:
```suggestion
```

Zyie (approved):

PR #10537, opened 2024-05-15
fix: buffer upload for texture

GoodBoyDigital:
##### Description of change
The upload code for WebGL was a behaving a bit funny when rep-uploading a buffer. This little refactor fixes it and simplifies the code a bit by doing the same kind of upload for webGL1 and 2 (both APIs supported)

I also renamed the `BufferSource.ts` file to `BufferImageSource` to match the internal class name


GoodBoyDigital:
oops! broke some webGL1 tests.. please hold!

bigtimebuddy (approved):

ivanpopelyshev (approved):
That's real fix

PR #10538, opened 2024-05-15
Do a more accurate Max Texture check

GoodBoyDigital:
##### Description of change
MAX_TEXTURES is now set by the GPU. WE do. alittle WebGL check to see how many textures it can support. This also happens for WebGPU. The reason we do a WebGL check is so that we don't need to create a promise and await the result. This keeps the code a lot simpler :D

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/batcher/shared/const.ts
+++ b/src/rendering/batcher/shared/const.ts
@@ -1,1 +1,3 @@
-export const MAX_TEXTURES = 16;
+import { maxRecommendedTextures } from '../../renderers/shared/texture/utils/maxRecommendedTextures';
+
+export const MAX_TEXTURES = maxRecommendedTextures();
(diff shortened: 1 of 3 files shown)


src/rendering/batcher/shared/const.ts
```diff
@@ -1 +1,3 @@
-export const MAX_TEXTURES = 16;
+import { maxRecommendedTextures } from '../../renderers/shared/texture/utils/maxRecommendedTextures';
+
+export const MAX_TEXTURES = maxRecommendedTextures();
```
bigtimebuddy, line 3: Do we even need this export? We are already caching maxTextureUnits, why not just call the maxRecommendedTextures inline?

GoodBoyDigital: i think its a lot easier to read this way - its a good point though, i can remove the caching from the function?

bigtimebuddy: Yeah, I don't think the double cache is necessary.

GoodBoyDigital: done!

bigtimebuddy (approved):

gabrielecirulli:
Please note this PR breaks worker support: https://github.com/pixijs/pixijs/issues/10546

PR #10543, opened 2024-05-17
fix: event mixin types

Zyie:
Ok so this PR was a fun rabbit hole 😄 

# Overview

The main overview of what this PR is actually changing is that i have removed the abstraction of `IFederatedEventTarget` in favour of just using `Container`. `IFederatedContainer` is still mixed into the `Container`, which means that all types should work as before. 

I believe this change reflects a more accurate description of what object is stored in an `FederatedEvent`.

The benefits of this is that we don't need to do any typescript magic that decouples the two different implementations of `EventEmitter` e.g

This
```ts
export interface Container<C extends ContainerChild>
    extends Omit<PixiMixins.Container<C>, keyof EventEmitter<ContainerEvents<C> & AnyEvent>>,
    EventEmitter<ContainerEvents<C> & AnyEvent> { }
```

Becomes this
```ts
export interface Container<C extends ContainerChild>
    extends PixiMixins.Container<C>, EventEmitter<ContainerEvents<C> & AnyEvent> {}
```

## What Is This Fixing

The current implementation revealed a nasty bug with using our `PixiMixins` that is hidden in v7 due to `Container` extending `DisplayObject` and everyone will be attaching there custom properties to `Container`, not `DisplayObject`, because of this we have never seen the issue.

Hopefully i can try and explain it here...

If you try to add your own properties to `Container` in v8 currently like this:

```ts
declare namespace PixiMixins {
    interface Container {
        overrideMethod(): void
		returnThis(): this
    }
}
```
Then you try to override and use these methods like:
```ts
class NewContainer extends Container {
    testProperty: number;
    override overrideMethod(layout: ComputedLayout) {}
}

const container = new NewContainer();
const container2 = container.returnThis();
container.testProperty // type error: does not exist
```

We get two TS errors:

- `Class 'Container` defines instance member property `overrideMethod`, but extended class `NewContainer` defines it as instance member function.
- Property 'testProperty' does not exist on type `Container`

The first issue is that typescript believe the `overrideMethod` is a property on `Container` and not a method. So it wants you to declare it like this
```ts
override overrideMethod = (layout: ComputedLayout) => {}
```
However this isn't accurate as if we wanted to declare it as a property it should be defined like this
```ts
declare namespace PixiMixins {
    interface Container {
        overrideMethod: ()=> void
    }
}
```

The second issue is that `this` is not being respected correctly. It should be returning `NewContainer`, however it always returns `Container`

Now I have zero idea what sort of typescript magic is going on that messes up these types so badly but the change here is minimal and seems to work.

### DIFF
--- a/src/events/EventBoundary.ts
+++ b/src/events/EventBoundary.ts
@@ -12,8 +12,6 @@ import type { EmitterListeners, TrackingData } from './EventBoundaryTypes';
 import type { FederatedEvent } from './FederatedEvent';
 import type {
     Cursor, EventMode, FederatedEventHandler,
-    FederatedEventTarget,
-    IFederatedContainer
 } from './FederatedEventTarget';
 
 // The maximum iterations used in propagation. This prevent infinite loops.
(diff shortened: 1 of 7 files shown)


bigtimebuddy (approved):
I like this change. Flattening the layers of abstractions for events is welcome.

PR #10566, opened 2024-05-21
fix: destroy renderGroups

### LINKED ISSUE #10533: Bug: Memory Leak / Inappropriate cleanUp - stage.renderGroup cleanUp is not there
### Current Behavior

When analysing the memory report I noticed that after closing the view with PIXI app I still see PIXI instances in the memory. Digging deeper into the chain, I noticed a common instance for leaked elements is **renderGroup**. Then I noticed **renderGroup** is cleared for elements in the **Container.removeChild** method:
https://github.com/pixijs/pixijs/blob/78824e43fe569fdf84206905e5e2c1ce9d9475c3/src/scene/container/RenderGroup.ts#L188
However for the **stage** element this method is not called as it has no parent:
https://github.com/pixijs/pixijs/blob/78824e43fe569fdf84206905e5e2c1ce9d9475c3/src/app/Application.ts#L207
https://github.com/pixijs/pixijs/blob/78824e43fe569fdf84206905e5e2c1ce9d9475c3/src/scene/container/container-mixins/childrenHelperMixin.ts#L228
Screenshot 2024-05-13 at 6 38 21 PM

On our end, we call the destroy method for app:
__pixi.destroy(true, {children: true, texture: true, baseTexture: true })_
***_pixi** - PIXI Application instance

### Expected Behavior

**renderGroup** should be cleared for **stage** to remove references to other elements and let GC do its job.

### Steps to Reproduce

Thing I would like to point out is that we are using a slightly unusual approach, we store PIXI on window, perhaps this is the reason why it keeps refs in memory, however it would be cool to clear the renderGroup for stage, this would be valid from my point of view.

### Environment

- **`pixi.js` version**: *8.1.1*
- **Browser & Version**: *124.0.6367.158 (Official Build) (x86_64)*
- **OS & Version**: *MacOS*
- **Running Example**: *<none>*

### Possible Solution

To overcome this issue I remove references to elements manually this way:
_this._pixi.stage.renderGroup.childrenRenderablesToUpdate = {};_
_this._pixi.stage.renderGroup.childrenToUpdate = {};_
_this._pixi.stage.renderGroup.instructionSet = null;_

However it would be great to do it directly in the destroy method of the Application.

### Additional Information

Let me know if I missed something or there is already a solution for this issue. Thanks!

GoodBoyDigital:
##### Description of change
Added a destroy function to renderGroup and also made sure it is called when the container that owns it is destroyed.

##### Pre-Merge Checklist

fixes #10533

### DIFF
--- a/src/scene/container/RenderGroup.ts
+++ b/src/scene/container/RenderGroup.ts
@@ -220,6 +220,17 @@ export class RenderGroup implements Instruction
         }
     }
 
+    public destroy()
+    {
+        this.renderGroupParent = null;
+        this.root = null;
+        (this.childrenRenderablesToUpdate as any) = null;
+        (this.childrenToUpdate as any) = null;
+        (this.renderGroupChildren as any) = null;
+        (this._onRenderContainers as any) = null;
+        this.instructionSet = null;
+    }
+
     public getChildren(out: Container[] = []): Container[]
     {
         const children = this.root.children;
--- a/src/scene/container/Container.ts
+++ b/src/scene/container/Container.ts
@@ -1320,7 +1320,6 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
         // remove children is faster than removeChild..
         const oldChildren = this.removeChildren(0, this.children.length);
 
-        this.removeFromParent();
         this.parent = null;
         this._maskEffect = null;
         this._filterEffect = null;
@@ -1342,6 +1341,12 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
             {
                 oldChildren[i].destroy(options);
             }
+        }
+
+        if (this.renderGroup)
+        {
+            this.renderGroup.destroy();
+            this.renderGroup = null;
         }
     }
 }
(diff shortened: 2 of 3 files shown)


src/scene/container/RenderGroup.ts
```diff
@@ -220,6 +220,17 @@ export class RenderGroup implements Instruction
...
     }
 
+    public destroy()
+    {
+        this.renderGroupParent = null;
+        this.root = null;
+        (this.childrenRenderablesToUpdate as any) = null;
+        (this.childrenToUpdate as any) = null;
+        (this.renderGroupChildren as any) = null;
+        (this._onRenderContainers as any) = null;
```
bigtimebuddy, line 230: Take this or leave it, but I think this will avoid the `as any` casting. 
```suggestion
        Object.assign(this, {
            childrenRenderablesToUpdate: null,
            childrenToUpdate: null,
            renderGroupChildren: null,
            _onRenderContainers: null,
        });
```

bigtimebuddy (approved):

Issue #10366, opened 2024-03-22
[v8] Cannot read property 'addRenderable' of undefined


JetLua:
### Description

I tried using `@pixi/unsafe-eval@7` and `pixi.js@8`, but I found it doesn't work.
```
Current environment does not allow unsafe-eval, please use pixi.js/unsafe-eval module to enable support.
```

bigtimebuddy:
Try including `pixi.js/unsafe-eval`. All packages are now shipped with v8.

JetLua:
> Try including `pixi.js/unsafe-eval`. All packages are now shipped with v8.

@bigtimebuddy Thank you, it's resolved.

When I run the following code in **WeChat**, I encounter an error.

```js

const renderer = new PIXI.WebGLRenderer()

renderer.init({
  canvas,
  antialias: true,
  preferWebGLVersion: 1,
  backgroundColor: 0x171a24,
}).then(() => {
  const g = new PIXI.Graphics()

  g.fill(0xffc333)
  g.rect(0, 0, 100, 100)

  stage.addChild(g)

  renderer.render({container: stage})
})
```

```
Cannot read property 'addRenderable' of undefined
```

I did some basic research and found that the error stems from the following code snippet. Additionally, I added code locally to print variable information.

https://github.com/pixijs/pixijs/blob/d4cb3d10fb8e50be8fc4e41e9d2844831df0a79e/src/scene/container/utils/buildInstructions.ts#L73-L75

```js
console.log(rp)
console.log(container.renderPipeId)
```

output:

<img width="548" alt="image" src="https://github.com/pixijs/pixijs/assets/6738986/9debc883-1e5a-4c68-b871-365c19f9263f">


I found that there is no `graphics` property in `renderPipes`, so `rp[container.renderPipeId]` is `undefined`.

I'm not entirely sure if this is a situation unique to **WeChat**. 

@GoodBoyDigital

GoodBoyDigital:
thanks for sharing! il take a look

GoodBoyDigital:
Hey there! The issue is that the extensions are dynamically loaded when via `autoDetectRenderer`

the quick fix for now would be to do the following:

```
const renderer = await autoDetectRenderer({
    preference:'webgl'
})
```

we will look into adding a `await createWebGLRenderer()` and `await createWebGPURenderer()` functions that install all the plugins and things!

kencanak:
[...]

Issue #10554, opened 2024-05-20
Bug: WebGPURenderer: TypeError: Cannot read properties of null (reading 'features') with NW.js v.0.86


Matt-Maerz:
### Current Behavior

WebGPURenderer is not loaded with nw.js v.0.86

`renderer = new PIXI.WebGPURenderer();`
`    await renderer.init({`
`        width: SETTINGS.RESOLUTION_WIDTH,`
`        height: SETTINGS.RESOLUTION_HEIGHT,`
`        background: 0x000000`
`    });`

Error message:
![error GPURenderer with nwjs](https://github.com/pixijs/pixijs/assets/10009596/175bb83d-726b-4efb-84c4-8556e96df7f6)

### Expected Behavior

WebGPURenderer is loaded

### Steps to Reproduce

Javascript Frontend:

`renderer = new PIXI.WebGPURenderer();`
`    await renderer.init({`
`        width: SETTINGS.RESOLUTION_WIDTH,`
`        height: SETTINGS.RESOLUTION_HEIGHT,`
`        background: 0x000000`
`    });`

Compilation to .exe win64 with https://nwjs.io/

### Environment

- **`pixi.js` version**: *8.1.4*
- **Browser & Version**: *Chrome 119.0.6045.200*
- **OS & Version**: *Windows Version 11 Home 23H2*

Zyie:
Hey, this is a known issue that you will want to use `autoDetectRenderer({ preference: 'webgpu', width. height, background })` for now

https://github.com/pixijs/pixijs/issues/10366

PR #10567, opened 2024-05-22
fix: allow renderers to be created without autoDetectRenderer

Zyie:
Fixes: #10366 and #10554

We now move the loading the environment into `AbstractRenderer`, which populates all of the render pipes, into the `init` call. Tis allows the renderer to be created correctly without using `autoDetectRenderer`

I also tidied up a few interfaces so now the types extend the correct things


GoodBoyDigital (approved):
nice work

PR #10568, opened 2024-05-22
fix: spritesheet resolver not working

Zyie:
the `resolveJsonUrl` is a more general version of the spritesheet resolver so should have a lower priority


bigtimebuddy:
Can you unit-test for this?

bigtimebuddy (approved):

PR #10572, opened 2024-05-23
chore: improve types for asset parsers

Zyie:
This PR improves the types for the asset parsers and keeps the types backwards compatible

- all parsers now use `satisfies` 
- added `AssetExtensionAdvanced` and `LoaderParserAdvanced`, which allow you to have better control over what asset is parsed to each function
- all parsers now have an extension name, making debugging these things so much easier


GoodBoyDigital (approved):
nice!

PR #10581, opened 2024-05-28
fix: fills/strokes not being set correctly

Zyie:
- fixes the issue where setting a `FillPattern` or a `FillGraident` directly did not work e.g
```ts
const gradient = new FillGradient(0, 0, 200, 0);
new TextStyle({
  stroke: gradient,
})
```
- split the types up to clearly separate fill and stroke styles
- moved the types for strokes/fills into a separate file as `GraphicsContext` is already pretty large
- deprecated `FillStyleInputs` in favor of splitting into `FillInput` and `StrokeInput` for better types. Also `FillStyleInputs` contained stroke styles, which meant you got the wrong typings when using `.fill()`

### DIFF
--- a/src/scene/graphics/shared/utils/convertFillInputToFillStyle.ts
+++ b/src/scene/graphics/shared/utils/convertFillInputToFillStyle.ts
@@ -7,76 +7,113 @@ import { FillPattern } from '../fill/FillPattern';
 import type { ColorSource } from '../../../../color/Color';
 import type {
     ConvertedFillStyle,
+    ConvertedStrokeStyle,
+    FillInput,
     FillStyle,
-    FillStyleInputs,
-    PatternFillStyle,
-} from '../GraphicsContext';
+    StrokeInput,
+} from '../FillTypes';
 
-export function convertFillInputToFillStyle(
-    value: FillStyleInputs,
+function isColorLike(value: unknown): value is ColorSource
+{
+    return Color.isColorLike(value as ColorSource);
+}
+
+function isFillPattern(value: unknown): value is FillPattern
+{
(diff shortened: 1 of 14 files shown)


src/scene/graphics/shared/utils/convertFillInputToFillStyle.ts
```diff
@@ -106,6 +140,62 @@ export function convertFillInputToFillStyle(
     style.color = color.toNumber();
     style.matrix = style.matrix ? style.matrix.clone() : null; // todo: lets optimise this!
 
-    // its a regular fill style!
     return style as ConvertedFillStyle;
 }
+
+export function convertFillInputToFillStyle<T extends FillInput>(
```
bigtimebuddy, line 146: I know this func name is old, but it is so verbose. What do you think about `toFillStyle` instead? Especially since you made this templated, it makes even less sense.

bigtimebuddy: Same for `convertStrokeInputToStrokeStyle` too. 

Also, can you add a doc block that explains _why_ this exists (private is final)? Just something about why are we need to do this conversion..

bigtimebuddy (approved):

src/scene/graphics/shared/utils/convertFillInputToFillStyle.ts
Zyie: yeah i did update the name to `toFillStyle` previously, then i reverted it to make the changes smaller. So more than happy to change to this again

GoodBoyDigital (approved):
much nicer!

PR #10582, opened 2024-05-28
fix: FillPattern/FillGradient overriding each other when both are used

Zyie:
Fixes the issue where if you use both a `FillPattern` and a `FillGradient` at the same time, only the first one will render


tests/visual/scenes/text/text-pattern-gradient.scene.ts
```diff
@@ -0,0 +1,45 @@
...
+import { Text } from '../../../../src/scene/text/Text';
+
+import type { Texture } from '../../../../src/rendering/renderers/shared/texture/Texture';
+import type { Container } from '../../../../src/scene/container/Container';
+import type { TestScene } from '../../types';
+
+export const scene: TestScene = {
+    it: 'should render FillPattern and FillGradient in text stroke',
+    pixelMatch: 200,
+    only: true,
```
Zyie, line 13:
```suggestion
```

bigtimebuddy (approved):

GoodBoyDigital (approved):
nice!

Issue #10426, opened 2024-04-08
Bug: V8 text stroke is applied even with width 0. Also strokeThickness does not update at runtime


Unwarped:
### Current Behavior

If you set the strokeThickness to 0, it still applies the text stroke. I made a V7 and V8 playground to show the difference.

[V7 playground](https://www.pixiplayground.com/#/edit/NntdEhYx87jS6EPWGsDyw)
[V8 playground](https://www.pixiplayground.com/#/edit/Isd-hTei9N1B3gyGqGAu7)

You can tell the text looks thicker.

Zyie determined that setting width to 0.00001 or a small decimal number fixed it. Removing stroke or setting to null also works. But 0 does not.

Also, if you try to change strokeThickness at runtime, it does not update. I talked with doormat23 and he believes it is a caching issue. Here is a playground to demonstrate.

[Changing text strokeThickness at runtime Playground](https://www.pixiplayground.com/#/edit/mZCibtN38GSw64fZpmL1L)

### Expected Behavior

With strokeThickness 0 there should be no stroke and the text looks slimmer.

And changing it at runtime should update the thickness

### Steps to Reproduce

See playgrounds above

### Environment

- **`pixi.js` version**: *e.g. 8.0.5*

### Additional Information

Thanks guys!

Unwarped:
Neither issue seems to be fixed when I change the playgrounds version to latest 8.1.1

Unwarped:
This works great thanks guys!!

PR #10583, opened 2024-05-28
fix: stroke and fills not updating when you change property

Zyie:
Fixes: #10426

we now proxy the fill and stroke to make sure we can capture property changes


src/scene/text/TextStyle.ts
```diff
@@ -535,6 +560,13 @@ export class TextStyle extends EventEmitter<{
             }
         });
     }
+
+    private _isFillStyle(value: FillInput): value is FillStyle
+    {
+        // eslint-disable-next-line no-eq-null, eqeqeq
+        return (value != null
```
bigtimebuddy, line 567: Why violate the lint here?

bigtimebuddy (approved):

src/scene/text/TextStyle.ts
```diff
@@ -416,6 +429,19 @@ export class TextStyle extends EventEmitter<{
         if (value === this._originalStroke) return;
 
         this._originalStroke = value;
+
+        // eslint-disable-next-line no-eq-null, eqeqeq
```
bigtimebuddy, line 433: What this old? Lint exception doesn't make sense to me here.
```suggestion
```

src/scene/text/TextStyle.ts
bigtimebuddy: I think being verbose here is okay. Style should only be over ridden as a necessity. You could try using the coalesce operator.

```ts
(value ?? null) !== null
```

GoodBoyDigital (approved):
very cool! nice work!

PR #10585, opened 2024-05-28
feat: add container reparenting

Zyie:
This PR adds `container.reparentChild(child)` and `container.reparentChildAt(child, 1)` which allows you to add a container to a new parent while visually keeping the container looking the same


src/scene/container/container-mixins/childrenHelperMixin.ts
```diff
@@ -231,5 +232,26 @@ export const childrenHelperMixin: Partial<Container> = {
...
+     * Reparent the Container to a new parent Container, keeping the same worldTransform.
+     * @param newParent - The new parent Container to add this Container to.
+     * @param index - The index to add the child to. If not provided, the child will be added to the end of the list.
+     * @memberof scene.Container#
+     */
+    reparent(newParent: Container, index?: number)
+    {
+        const mat = this.worldTransform.clone();
+
+        this.parent.removeChild(this);
```
bigtimebuddy, line 247: How do you want to handle if the.parent is null and someone calls this? Probably an assert? Or  should we use the identity matrix?

Zyie: I've updated to allow for "reparenting" when the container doesn't have a parent as you technically don't need a parent for this math to work. I updated the tests to reflect this

However, now `reparent` feels like it is not the correct name for this anymore?

bigtimebuddy: I agree. I'm also not crazy that reparent is on the child not the container. Maybe an alternative is to create a variant of `addChildAt`, perhaps `insertChildAt` or `applyChildAt`. Also, why not add a third argument to `addChildAt(child, index, inPlace)`?

bigtimebuddy (approved):

Issue #10441, opened 2024-04-11
Bug: Events are firing twice when we have dynamic and static mode


danielbarion:
### Current Behavior

When we have `eventMode = 'static'` and `eventMode = 'dynamic'` in the same app, the events fire twice.

### Expected Behavior

When we have `eventMode = 'static'` and `eventMode = 'dynamic'` in the same app, the events should fire only once.

### Steps to Reproduce

1. Go to https://pixijs.com/8.x/playground?exampleId=events.logger
2. On the following code:
```
    // Enable interactivity everywhere!
    app.stage.eventMode = 'static';
    app.stage.hitArea = app.screen;
    whiteBox.eventMode = 'static';
    blackBox.eventMode = 'static';
```
3. Change any of the event modes to be dynamic:

```
    // Enable interactivity everywhere!
    app.stage.eventMode = 'static';
    app.stage.hitArea = app.screen;
    whiteBox.eventMode = 'dynamic';
    blackBox.eventMode = 'static';
```

Wait it gets reflected into the playground, the events are now firing twice.

[...]

PR #10601, opened 2024-06-03
fix: Event handlers firing twice on non-interacting root display objects

CatchABus:
##### Description of change
Right now, property event handlers like `onclick` can be triggered for non-intaractive display objects. In the case of the root display object, listeners will be called twice.
I added 2 unit tests that seem to catch the case.

This PR is a continuation of #9166
Note: This bug seems similar to #10441 but it's not the same defect.

##### Pre-Merge Checklist


package.json
```diff
@@ -117,7 +117,7 @@
     "ts-jest": "^26.0.0",
     "ts-node": "^9.0.0",
     "tsconfig-paths": "^3.10.1",
-    "typescript": "^5.0.4",
+    "typescript": "~5.3.3",
```
bigtimebuddy, line 120: Could you explain this change?

bigtimebuddy: I'd like to consider this as a separate PR. It's outside the scope of what you're trying to do here and I think for our case, that ESLint warning is benign.

CatchABus: Done! Thank you very much for pointing out! :smile:

bigtimebuddy (approved):
Nice test

PR #10604, opened 2024-06-04
fix: Bug in loading SVG though a spritesheet

AnkushSarkar10:
##### Description of change
Fixed a bug where `asset.data` was undefined when loading a spritesheet which links to an svg as such

(spritessheet.json)
```
{
  "frames": {
    "default": {
         ...
    }
  },
  "meta": {
    "app": "...",
    "version": "...",
    "image": "someSvg.svg",
    "format": "..."
  }
}
``` 

##### Pre-Merge Checklist


src/assets/loader/parsers/textures/loadSVG.ts
```diff
@@ -69,7 +69,8 @@ export const loadSvg: LoaderParser<Texture | GraphicsContext, TextureSourceOptio
         loader: Loader
     ): Promise<Texture | GraphicsContext>
     {
-        if (asset.data.parseAsGraphicsContext ?? this.config.parseAsGraphicsContext)
+        asset.data = asset.data ?? this.config;
+        if (asset.data.parseAsGraphicsContext)
```
bigtimebuddy, line 73: Small suggestion to keep asset immutable.
```suggestion
        const data = asset.data ?? this.config;
        if (data.parseAsGraphicsContext)
```

bigtimebuddy (commented):
Would it be possible to add a simple test for this?

src/assets/loader/parsers/textures/loadSVG.ts
AnkushSarkar10: then if would fail in `loadAsTexture()` line 124 same file. For some reason `asset.data` is not set when loading an svg through a spritesheet (if i load the same svg file using `await Asset.load("svgfile.svg")` asset.data is there). I dont think adding this would break anything (it's just meta data).

AnkushSarkar10: I was trying to add a test case but I am not able to run the test suit for some reason. It just stuck here
image

I ran `npm run start` to build then ran `npm test`.  I don't know what I am missing any suggestions?

I also tried running the test file i need directly, same thing.
image

AnkushSarkar10: > then if would fail in `loadAsTexture()` line 124 same file. For some reason `asset.data` is not set when loading an svg through a spritesheet (if i load the same svg file using `await Asset.load("svgfile.svg")` asset.data is there). I dont think adding this would break anything (it's just meta data).

https://github.com/pixijs/pixijs/blob/54029ce4e39d30db431cd66b8f9161f0880661b9/src/assets/loader/Loader.ts#L111


this line does not add data to asset for some reason

AnkushSarkar10: > Small suggestion to keep asset immutable.

I tried doing that in both `load` and `loadAsTexture`, it fails later. I think its best to use what I did. I don't think it will cause any issues.

bigtimebuddy: If we need to reassign `asset.data` it means there's likely some other problem upstream from this.

AnkushSarkar10: > Small suggestion to keep asset immutable.

done

AnkushSarkar10: @bigtimebuddy let me know if this looks good

AnkushSarkar10: @bigtimebuddy added a test case, but i can't run them so not sure if it's correct (it looks like it should pass)

AnkushSarkar10:
any updates on this @bigtimebuddy @Zyie

AnkushSarkar10:
@bigtimebuddy @Zyie fixed upsteam

AnkushSarkar10:
image

Zyie:
Closing this in favor of #10694 as it more directly fixes the issue, but thank you for pointing me in the right direction!

PR #10627, opened 2024-06-13
fix: dynamic-bitmap-font-fixes

GoodBoyDigital:
##### Description of change
Fixes a few issues with BitmapText

1 - clears the sprite sheet canvas when reusing them from the pool and creating new fonts.
2 - allows for multiple Bitmap Font styles to work with a font. Whilst still optimising for a single BitmapFont in the case of normal filled text. Added a little warning for the user if this abused (as in there are more than 50 dynamically created textures)
3 - set mipmapping on the bitmap font by default - fonts look much nicer! (had to update some of the visual tests because of this)
4 - fix a height issue where stroke width was being applied 2x to the height of text. BitmapText now viually positions itself exactly the same place as regular text.

##### Pre-Merge Checklist


naramdash:
Is this PR related to the issue of dynamically created bitmaptext **looking different on different OSs**?
What I'm experiencing is not between different browsers on the same OS, but on different OSs.

image

Zyie (approved):

PR #10634, opened 2024-06-17
feat: add initialised set up hook

Zyie:
This PR adds a simple hook that exposes the app/renderer to allow for the devtools to work without the need for a user to configure anything

- [devtool docs](https://pixijs.io/devtools/)
- [devtool github](https://github.com/pixijs/devtools/tree/breaking/extensions)


src/utils/global/globalHooks.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    gThis.__PIXI_RENDERER_INIT__?.(renderer);
+}
+
+export function pixiAppCreated(app: Application)
+{
+    const gThis = globalThis as unknown as Globals;
+
+    gThis.__PIXI_APP_INIT__?.(app);
+}
```
bigtimebuddy, line 22: Why not use `declare global` here instead to avoid casting?

```suggestion
declare global
{
    var __PIXI_APP_INIT__:undefined | ((arg: Application | Renderer) => void);
    var __PIXI_RENDERER_INIT__: undefined | ((arg: Application | Renderer) => void);
}

export function pixiRendererCreated(renderer: Renderer)
{
    globalThis.__PIXI_RENDERER_INIT__?.(renderer);
}

export function pixiAppCreated(app: Application)
{
    globalThis.__PIXI_APP_INIT__?.(app);
}
```

Also, not crazy about the names with the "pixi" prefix. Maybe `rendererCreatedHook` and `appCreatedHook`?

bigtimebuddy:
They can be installed by default, I'm thinking more about how to organize this code in a way that's more consistent internally.

bigtimebuddy (approved):
Nice I like it!

PR #10653, opened 2024-06-21
feat: Optimize blur by reducing blur strength on each pass

minetoblend:
##### Description of change
RFC. Instead of using a constant `strength / passes` value for each pass in `BlurFilterPass`, it continuously reduces the strength with each pass, which leads to much less passes needed to get good results.

There's still 2 issues with this currently that I wasn't able to figure out by myself:
- Changing the blur strength on each pass seems to run into a bug on the WebGPU renderer currently, looks like it applies all passes with the strength of the last blur pass, resulting in a much smaller blur strength
- The original blur implementation uses a strength of `strength / passes` for all passes which will result in different combined blur strength depending on how many passes there are. I added a correction step, leading to the appearance being the same regardless of how many passes are used, however this results in the blur strength being different from the original blur. I can try making some adjustments to make it match the original closer.

Comparison with `strength = 15`, `quality = 3`
**Original**
Screenshot 2024-06-21 171557
**PR**
Screenshot 2024-06-21 171541

Comparison with `strength = 15`, `quality = 8`
**Original**
Screenshot 2024-06-21 172003
**PR**
Screenshot 2024-06-21 172025

##### Pre-Merge Checklist


GoodBoyDigital:
this is really cool! il try to find some time to look into the outstanding issues!

GoodBoyDigital:
Hey @minetoblend ! we have not forgoten about this - just super stacked. We love the optimisation here!

GoodBoyDigital:
made a tweak on top of your PR here https://github.com/pixijs/pixijs/pull/11868. Thanks!

PR #10658, opened 2024-06-24
fix: text not updating with renderer resolution

### LINKED ISSUE #10643: Bug: Text's resolution does not update on Renderer's resolution change
### Current Behavior

When the Renderer's `resolution` is changed, Text/HTMLText objects that have `resolution` set to  `null` (auto-resolution), do not redraw with the new resolution. 

Here's an example: 
https://jsfiddle.net/bigtimebuddy/wpfz0x56/

https://github.com/pixijs/pixijs/assets/864393/a0e8e938-ade7-4c53-8245-bf584a43c2f3

### Expected Behavior

Resolution of these Text objects should update with the resolution.

https://github.com/pixijs/pixijs/assets/864393/3559495c-5702-4755-b111-73206c58e2d8

Current workaround is manually manage `resolution` on Text.

https://jsfiddle.net/bigtimebuddy/raw9s8gk/

### Steps to Reproduce

See example above.

### Environment

- **`pixi.js` version**: 8.1.7
- **OS & Version**: macOS
- **Running Example**: https://jsfiddle.net/bigtimebuddy/wpfz0x56/

### Possible Solution

_No response_

### Additional Information

_No response_

Zyie:
fixes #10643

I also had to update the visual tester to allow for any size image to be created


src/scene/text/AbstractText.ts
```diff
@@ -395,6 +396,12 @@ export abstract class AbstractText<
 
     public _getKey(): string
     {
+        if (!this.resolution)
+        {
+            // add rendererResolution to the key to make sure the text is re-rendered when the resolution changes
+            return `${this.text}:${this._style.styleKey}-${this._rendererResolution}`;
```
bigtimebuddy, line 402: You could do this in a single line like this:

```ts
    return `${this.text}:${this._style.styleKey}-${this.resolution ?? this._rendererResolution}`;
```

bigtimebuddy (approved):
Looks good. Thanks for this!

src/scene/text-html/HTMLTextPipe.ts
```diff
@@ -34,6 +34,19 @@ export class HTMLTextPipe implements RenderPipe<HTMLText>
...
+    }
+
+    public resolutionChange()
+    {
+        for (const i in this._gpuText)
+        {
+            const gpuText = this._gpuText[i];
+            const text = gpuText.batchableSprite.renderable as HTMLText;
+
+            text._resolution = text._autoResolution ? this._renderer.resolution : text.resolution;
```
GoodBoyDigital, line 47: mini optimisation - we only need to update the ones with text._autoResolution set to true.

```
if(text._autoResolution)
{
    // will trigger on viewUpdate
    text.resolution = this.renderer.resolution
}
```

src/scene/text/canvas/CanvasTextPipe.ts
```diff
@@ -32,6 +32,19 @@ export class CanvasTextPipe implements RenderPipe<Text>
...
+    public resolutionChange()
+    {
+        for (const i in this._gpuText)
+        {
+            const gpuText = this._gpuText[i];
+            const text = gpuText.batchableSprite.renderable as Text;
+
+            text._resolution = text._autoResolution ? this._renderer.resolution : text.resolution;
+            text.onViewUpdate();
+        }
```
GoodBoyDigital, line 47: can make the same change here too!

```
if(text._autoResolution)
{
    // will trigger on viewUpdate
    text.resolution = this.renderer.resolution
}
```

tests/renderering/text/Text.tests.ts
```diff
@@ -140,12 +140,11 @@ describe('Text', () =>
 
             const text = new Text({ text: 'foo' });
 
-            const key = text._getKey();
-
             container.addChild(text);
-
             renderer.render({ container });
 
+            const key = text._getKey();
```
GoodBoyDigital, line 146: not much of a muchness - but no changes required to this file?

GoodBoyDigital (commented):
loking good! couple of tiny tweaks!

tests/renderering/text/Text.tests.ts
Zyie: the change is required as we now set the resolution in the pipe, so getting the key before its rendered would still give `null` and after it would be `1`

GoodBoyDigital (approved):
nice!

Issue #10131, opened 2024-01-18
Bug: pointerover/pointerout fire twice when eventMode


Slapbox:
### Current Behavior

When the mouse moves over a node with `eventMode='dynamic'` it triggers `pointerover` to fire twice. When the node is in motion and the cursor is stationary, it properly fires `pointerover` only once.

The same behavior affects `pointerout`.

### Expected Behavior

`pointerover` and `pointerout` should fire only once per relevant event.

### Steps to Reproduce

Hover over any moving Sprite that has `eventMode = 'dynamic'` defined.

The issue occurs about 90% of the time.

### Environment

- **`pixi.js` version**: 7.3.3
- **Browser & Version**: 120
- **OS & Version**: Ubuntu 23.10/Windows 11
- **Running Example**: *e.g. <https://pixiplayground.com/>* (Playground errors with `WebGL unsupported in this browser` so I cannot easily provide a repro)

### Possible Solution

Unknown

joanjeremiah:
I would like to work on this.

Slapbox:
@joanjeremiah thanks for your reply!

Do you see what problem I might have in my reproduction attempt? I'm not sure why it doesn't render anything but it seems like the first step to getting this fixed. I found a couple problems but even after fixing them, still nothing renders.

Slapbox:
This also affects Pixi `8.0.0-rc.5`.

Slapbox:
I was able to fix my repro link. I hope somebody might be able to determine why this behavior occurs.

Slapbox:
Friendly bump. It seems really weird that no-one else is seeing (or at least reporting) this behavior.

gabrielliwerant:
I can verify the issue, but it looks like it might be intended functionality, and not a bug. The docs state that it will receive an extra event from the ticker when the mouse is not moving. Have tried using `eventMode = 'static'`? This is the basic mode for interactions and should only fire once per event.

Slapbox:
I didn't read the docs quite the same as you, but perhaps that actually is the right impression?

> 'dynamic': Emits events and is hit tested but will also receive mock interaction events fired from a ticker to allow for interaction when the mouse isn't moving

I read that as, it will fire either or, but not both. I guess it can be interpreted either way.

I did try static but for a graph layout where elements move, it feels bad when the interactivity doesn't occur when the element passes under the cursor.

gabrielliwerant:
[...]

PR #10677, opened 2024-07-01
fix: events firing twice when using dynamic mode

Zyie:
fixes: https://github.com/pixijs/pixijs/issues/10441
fixes: https://github.com/pixijs/pixijs/issues/10131

before: https://www.pixiplayground.com/#/edit/exDzbrsXsLvgdbWm0iZFG
after: https://www.pixiplayground.com/#/edit/cbfMQ_YRzBPNtwevahYkW

I can't find a good way of adding a test for this fix as the way we emulate pointer events in our tests is not the exact same as how a browser does it, due to `pointerId` being different, but the demo above does show the over event only firing once

### DIFF
--- a/src/events/EventTicker.ts
+++ b/src/events/EventTicker.ts
@@ -110,6 +110,8 @@ class EventsTickerClass
         globalThis.document.dispatchEvent(new PointerEvent('pointermove', {
             clientX: rootPointerEvent.clientX,
             clientY: rootPointerEvent.clientY,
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):

Nantris:
Thanks so much for your work on this @Zyie. Is it possible this might be backported to `7.x`? This is causing us serious problems but upgrading to `8.x` right now simply isn't feasible.

PR #10681, opened 2024-07-02
feat: add a renderable GC

GoodBoyDigital:
##### Description of change
This PR introduces a little renderable GC System. The system is designed to identify when renderables have not been rendered for a period of time. If after a that period has passed, the Renderer assets associated with that renderable are released back to a pool that other new renderables can use.

- Currently I have set it to run once every 60 secs.
- Runs independently from the renderer rate
- Minimal runtime changes - just one small assignment and check when instruction set if built per renderable.

Two reasons this is handy:
- If a reference to a renderable (eg a Sprite) is lost - its associated assets will eventually get cleaned up and removed from the renderer.
- Resources will be released as users progress through your app, keeping memory low by reusing resources.

**Breaking change**
The one breaking change here is that i had to modify the pipes to remove / add the destroy listeners. So for custom pipes, this will need to be addressed too. Basically when a renderable is destroyed - the event should also be removed.

**Next steps**
I see a nice path forward with RenderPipes where I will look to refactor them, so that things like creating / destroying the renderer assets are handled automatically - and don't need to be done by the Pipe. This would simplify all Pipes and naturally tie them into the life cycle of renderables.

##### Pre-Merge Checklist


package.json
```diff
@@ -32,7 +32,7 @@
     "watch:build": "run-s build:index build:rollup build:tsc build:dts postbuild",
     "test": "run-s test:unit test:scene",
     "test:unit": "npx jest --silent --testPathIgnorePatterns=tests/visual",
-    "test:debug": "cross-env DEBUG_MODE=1 npx jest --testPathIgnorePatterns=tests/visual",
+    "test:debug": "cross-env DEBUG_MODE=1 npx jest",
```
Zyie, line 35:
```suggestion
    "test:debug": "cross-env DEBUG_MODE=1 npx jest --testPathIgnorePatterns=tests/visual",
```

src/rendering/renderers/shared/texture/RenderableGCSystem.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    public destroy(): void
+    {
+        this.enabled = false;
+        this._renderer = null as any as Renderer;
+        this._managedRenderables.length = 0;
+    }
+}
+
+extensions.add(RenderableGCSystem);
```
Zyie, line 192: this should be added to `SharedSystems`?

GoodBoyDigital: i was doing both! fixed now :)

Zyie (approved):

PR #10683, opened 2024-07-03
fix: restore automatic filter resolution and mask resolution/antialias

### LINKED ISSUE #10671: Bug: Sprite Mask filter should default resolution to Renderer
### Current Behavior

When using a Sprite mask and changing the renderer resolution _after_ initialization, can result in the internal Sprite mask filter having a default resolution of `1` which produces some unsightly quality loss.

Jun-26-2024 15-18-45

Example:
https://jsfiddle.net/bigtimebuddy/6t35ks84/

### Expected Behavior

Sprite masks should always use the default to the Renderer resolution

### Steps to Reproduce

See example.

### Environment

- **`pixi.js` version**: 8.2.1
- **OS & Version**: macOS Latest
- **Running Example**: [*e.g. <https://pixiplayground.com/>*](https://jsfiddle.net/bigtimebuddy/6t35ks84/)

### Possible Solution

_No response_

### Additional Information

_No response_

dev7355608:
- Restores automatic Filter resolution (`null`) that was an option in V7, but now as `'inherit'` instead of `null`.
- Fixes #10671.

##### Pre-Merge Checklist


src/filters/Filter.ts
```diff
@@ -69,10 +69,11 @@ export interface FilterOptions
     blendMode?: BLEND_MODES;
     /**
      * the resolution the filter should be rendered at. The lower the resolution, the more performant
-     * the filter will be, but the lower the quality of the output. (defaults to the renderers resolution)
+     * the filter will be, but the lower the quality of the output. (default 1)
+     * If `null` the resolution of the render target is used.
      * Consider lowering this for things like blurs filters
      */
-    resolution?: number;
+    resolution?: number | null;
```
dev7355608, line 76: Or do we prefer `number | 'inherit'`?

src/filters/FilterSystem.ts
```diff
@@ -182,16 +182,15 @@ export class FilterSystem implements System
         }
         // get GLOBAL bounds of the item we are going to apply the filter to
 
-        const colorTextureSource = renderer.renderTarget.rootRenderTarget.colorTexture.source;
+        const colorTextureSource = renderer.renderTarget.renderTarget.colorTexture.source;
```
dev7355608, line 185: I think we want to base inheriting based on the current render target not the root, right?

src/filters/FilterSystem.ts
```diff
@@ -201,19 +200,16 @@ export class FilterSystem implements System
...
-                }
-                else
-                {
-                    antialias = false;
-                }
+                antialias = true;
+            }
+            else if (filter.antialias === 'inherit')
+            {
+                antialias ||= colorTextureSource.antialias;
```
dev7355608, line 212: The comment of `antialias` says:
> if this is true for any filter, it should be true

The previous logic didn't match the comment. Fixed that.

src/filters/Filter.ts
```diff
@@ -81,9 +82,9 @@ export interface FilterOptions
     padding?: number;
     /**
      * If true the filter will make use of antialiasing. Although it looks better this can have a performance impact.
-     * By default, the filter will detect the antialiasing of the renderer and change this automatically.
```
dev7355608, line 84: Incorrect doc. Default is 'off'.

src/filters/Filter.ts
```diff
@@ -69,10 +69,11 @@ export interface FilterOptions
     blendMode?: BLEND_MODES;
     /**
      * the resolution the filter should be rendered at. The lower the resolution, the more performant
-     * the filter will be, but the lower the quality of the output. (defaults to the renderers resolution)
```
dev7355608, line 72: Incorrect doc. Default is 1. (Though `null` as default would be nicer, I think).

src/filters/FilterSystem.ts
dev7355608: Though it is a bit odd that we use the lowest resolution, but antialiasing if any of the filters has it. I feel like it should be either lowest resolution + antialias if all filters have antialia,s or highest resolution + antialias if any of them has it. In V7 we chose the lowest multisample count.

GoodBoyDigital: good catch - you are correct - should be true only if all filters have antialiasing!

src/filters/Filter.ts
bigtimebuddy: I like inherit, it's easier to reason than null.

src/rendering/mask/alpha/AlphaMaskPipe.ts
```diff
@@ -30,7 +31,9 @@ class AlphaMaskEffect extends FilterEffect implements PoolItem
         super();
 
         this.filters = [new MaskFilter({
-            sprite: new Sprite(Texture.EMPTY)
+            sprite: new Sprite(Texture.EMPTY),
+            resolution: null,
```
bigtimebuddy, line 35: This would be nicer, more readable here:
```suggestion
            resolution: 'inherit',
```

src/filters/Filter.ts
```diff
@@ -181,7 +182,7 @@ export class Filter extends Shader
      * increase the performance of the filter.
      * @default 1
      */
-    public resolution: number;
+    public resolution: number | null;
```
bigtimebuddy, line 185:
```suggestion
    public resolution: number | 'inherit';
```

src/filters/Filter.ts
dev7355608: Made the change.

src/filters/FilterSystem.ts
dev7355608: Fixed.

src/filters/FilterSystem.ts
GoodBoyDigital: Good point - I think technically we would want to take both into account?
- Like if anything in the filter stack does not have antialiasing, then if would be off?
- Find the lowest resolution in the filter stack? 

perhaps we keep track of that as we push / pop?

GoodBoyDigital (commented):
this is cool! one little comment!

src/filters/FilterSystem.ts
dev7355608: I think there are use cases for `filter.antialias = 'on'` even if render does not use antialias. For example an interface layer with graphics needs antialiasing but the scene layer doesn't because it only contains sprite. And a filter within the interface layer with `filter.antialias = 'inherit'` should probably create an antialias filter texture, even though the renderer doesn't have antialiasing enabled.

GoodBoyDigital: Thanks for explaining, thats a pretty cool use case!

GoodBoyDigital (approved):
looks good!

PR #10685, opened 2024-07-03
feat: [WIP] simplified interaction system

Zyie:
# New Input System

This PR introduces the first draft of the `InputSystem`, designed to replace PixiJS's current `EventSystem`. The primary goal of the `InputSystem` is to significantly enhance performance, offering a much faster alternative to our existing event handling system, albeit with reduced DOM compatibility and functionality.

## Motivation

The current `EventSystem` in PixiJS is a powerful tool for creating complex interactions within your applications. However, this comes with a performance cost. This means that for many users, the `EventSystem` is overkill, and they suffer the performance hit without needing the full functionality. We have attempted to mitigate some of these issues by allowing users to turn on/off certain features, but this has proven ineffective.

It has now become clear that building something from the ground up to be faster and more efficient is the best way forward.

## Performance

Currently, the `InputSystem` is between 50% - 250% faster than the `EventSystem` depending on the complexity of the scene. We also have the added benefit of knowing all interactive objects in the scene, which allows us to skip hit testing altogether if there are no interactive objects in the scene.

## Differences

While the `InputSystem` is still based on the `EventSystem`, there are some key differences between the two systems:

- Removed `stopImmediatePropagation`. This added extra complexity and can be controlled by users
- Removed all non-pointer events. This includes `click`, `mousedown`, `mouseup`, `rightdown` etc. We simply have pointer events now. Users can use `event.button` or `event.type` to determine the type of interaction
- Removed `pointerleave`/`pointercancel`/`pointerenter`. Removing these events also reduced complexity
- We now track interactive objects. This means when you set `container.input.enable = true`, this will be added to a list. This change does bring with it a shift for users to tidy up their containers when they are no longer needed as they do not get removed until `container.destroy()` or `container.input.enable = false` is called.

# API

The new API can be broken down into three sections: `InputSystem`, `Container.input`, and `InputEvent`.

## Container Events

The `Container` class has been updated to include a new `input` property. This property is an instance of the `Input` class and is used to manage all input events for the container. 
The following methods are available:

- `enable` - This will enable or disable the input system for the container.
- `globalMove` - This will enable or disable global move events for the container.
- `interactiveChildren` - Same as the `interactiveChildren` property on the `Container` class.
- `cursor` - Same as the `cursor` property on the `Container` class.
- `isPointer` - Similar to `buttonMode` in v6.
- `on` - Same as the `on` method on the `Container` class but with the following events:
    - `pointermove`
    - `globalpointermove`
    - `pointerover`
    - `pointerout`
    - `pointerup`
    - `pointerupoutside`
    - `pointerdown`
    - `pointertap`
  All other events have been removed (e.g., `click`, `mousedown`, `mouseup`, etc.).

```js

container.input.enable = true;
container.input.globalMove = true;
container.input.interactiveChildren = true;
container.input.cursor = 'pointer'; // container.input.isPointer = true;
container.input.on('pointerdown', (event) => {
    console.log('pointerdown', event);
});
```

## Input System
The `InputSystem` has exposed the same API as the `EventSystem` to ensure a smooth transition for users. The following methods are available:

- `autoPreventDefault`
- `cursorStyles`
- `domElement`
- `pointer`
- `setCursor`
- `setTargetElement`
- `mapPositionToPoint`
- `hitTest`

New methods have been added to the `InputSystem` to allow for more control over the system:

- `ignoreHitTestCache` - This will ignore the hit test cache and force a new hit test to be performed.
- `ignoreDynamicEvents` - This will stop the EventTicker from firing events.
- `InputSystem.interactionDirty` - Tells the `InputSystem` that something in the scene has changed and we need to rebuild.

## InputEvent
The `InputEvent` class is a new class that is used to store information about the event that has occurred. This class almost mirrors the `FederatedPointerEvent` class, but with some properties removed:

- `composed` - Was never used in the `EventSystem`.
- `bubbles`/`cancelBubble` - You should use stopPropagation instead.
- `cancelable` - Was never used in the `EventSystem`.
- `eventPhase` - We do not have event phases in the `InputSystem`.
- `stopImmediatePropagation`/`propagationImmediatelyStopped` - We do not support this in the `InputSystem`.
- `manager` - Not needed in the `InputSystem`.

# Compatibility Layer

To ensure that the `InputSystem` can be used as a drop-in replacement for the `EventSystem`, we have created a compatibility layer that will allow users to use the `InputSystem` with the existing `EventSystem` API. This will allow users to transition to the new system more easily.

All you need to do is import it:
```js
import 'pixi.js/input';
```
Now using events like before should work
```js

container.interactive = true;
container.interactiveChildren = true;
container.cursor = 'pointer';
container.on('pointerdown', (event) => {
    console.log('pointerdown', event);
});

### DIFF
--- a/src/input/Input.ts
+++ b/src/input/Input.ts
@@ -1,0 +1,84 @@
+import EventEmitter from 'eventemitter3';
+import { InputSystem } from './InputSystem';
+
+import type { Container } from '../scene/container/Container';
+
+export class Input extends EventEmitter<{
+    pointermove: PointerEvent;
+    globalpointermove: PointerEvent;
+    pointerover: PointerEvent;
+    pointerout: PointerEvent;
+    pointerup: PointerEvent;
+    pointerupoutside: PointerEvent;
+    pointerdown: PointerEvent;
+    pointertap: PointerEvent;
+}>
+{
+    public interactiveChildren = true;
+    public _globalMove = false;
+    public cursor = 'default';
+    public _isButton = false;
+    public hitArea: any = null;
+    protected container: Container = null;
+    protected _enable = false;
+
+    constructor(container: Container)
+    {
+        super();
+        this.container = container;
+
+        this.container.on('destroyed', this.destroy, this);
+    }
+
+    get isButton(): boolean
+    {
+        return this._isButton;
+    }
+
+    set isButton(val)
+    {
+        this._isButton = val;
+        this.cursor = val ? 'pointer' : 'default';
+    }
+
+    get globalMove(): boolean
+    {
+        return this._globalMove;
+    }
+
+    set globalMove(val)
+    {
+        if (val === this._globalMove)
+        {
+            return;
+        }
+
+        this._globalMove = val;
+        if (val) InputSystem.interactionDirty = true;
+    }
+
+    get enable(): boolean
+    {
+        return this._enable;
+    }
+
+    set enable(val)
+    {
+        if (val === this._enable)
+        {
+            return;
+        }
+
+        this._enable = val;
+        val ? InputSystem.addInteractive(this.container) : InputSystem.removeInteractive(this.container);
+    }
+
+    public destroy()
+    {
+        this.container.off('destroyed', this.destroy, this);
+        this.removeAllListeners();
+        this.container = null;
+        this.enable = false;
+        this._globalMove = false;
+    }
+}
--- a/src/input/Input.ts
+++ b/src/input/Input.ts
@@ -17,8 +17,7 @@ export class Input extends EventEmitter<{
     public interactiveChildren = true;
     public _globalMove = false;
     public cursor = 'default';
-    public _isButton = false;
-    public hitArea: any = null;
+    public _isPointer = false;
     protected container: Container = null;
     protected _enable = false;
 
@@ -30,14 +29,14 @@ export class Input extends EventEmitter<{
         this.container.on('destroyed', this.destroy, this);
     }
 
-    get isButton(): boolean
+    get isPointer(): boolean
     {
-        return this._isButton;
+        return this._isPointer;
     }
 
-    set isButton(val)
+    set isPointer(val)
     {
-        this._isButton = val;
+        this._isPointer = val;
         this.cursor = val ? 'pointer' : 'default';
     }
 
--- a/package.json
+++ b/package.json
@@ -176,6 +176,7 @@
     "./lib/compressed-textures/basis/init.*",
     "./lib/events/init.*",
     "./lib/filters/init.*",
+    "./lib/input/init.*",
     "./lib/math-extras/init.*",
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

**Comment by bigtimebuddy** [comment]
This is great! Love all your thinking here and rationale. 

What about having to instantiate the Input class instead of auto-adding it? I know Mat said that instance objects members have a perf cost. Also, this would allow us to add constructor options more easily and be more declarative. Maybe instead of `enabled`, the presence of input assumes true?

```js
import { Container, Input } from 'pixi.js';

container.input = new Input({
  globalMove: true,
  onPointerMove: () => void,
});

// alt

const container = new Container({
  input: new Input({
    globalMove: true,
    onPointerMove: () => void,
  }
});
```
Setting function callbacks was a feature of events. I think that has some value especially if settable as options.

kimjuho-daim:
Would this change also be relevant for interactions in pixi that work in webworker?

It would be nice to have events work even if pixi is running in webworker.

I think it would be possible to pass the event object from the canvas via the postMessage method and have the user manually enter it into the event handler. 

By using postMessage, classes and methods infomations are not sent (e.target.getBoundingRects(), e.target.parent()...). so it would be nice to set params of the event handler to a plain javascript object type.

naramdash:
Would this change also be relevant for interactions in pixi that work in webworker?

It would be nice to have events work even if pixi is running in webworker.

I think it would be possible to pass the event object from the canvas via the postMessage method and have the user manually enter it into the event handler.

By using postMessage, classes and methods infomations are not sent (e.target.getBoundingRects(), e.target.parent()...). so it would be nice to set params of the event handler to a plain javascript object type.

dev7355608:
Is this planned for a 8.x release or 9.0?

bigtimebuddy:
@dev7355608 this will probably be opt-in for v8, default in v9. @Zyie was careful not to overlap the APIs with events.

src/input/Input.ts
```diff
@@ -24,24 +47,68 @@ export class Input extends EventEmitter<{
...
+        this.onpointerover ??= options.onpointerover;
+        this.onpointerout ??= options.onpointerout;
+        this.onpointerup ??= options.onpointerup;
+        this.onpointerupoutside ??= options.onpointerupoutside;
+        this.onpointerdown ??= options.onpointerdown;
+        this.onpointertap ??= options.onpointertap;
+
+        this.globalWheel = options.globalWheel ?? false;
+        this.onwheel ??= options.onwheel;
+        this.onglobalwheel ??= options.onglobalwheel;
```
bigtimebuddy, line 111: Why not use assign here? It's better pattern for defaults. because `target` is a different name the arg should probably be `{ target, ...options }: Options`

```suggestion
        this.container = target;
        Object.assign(this, {
            interactive: false,
            globalMove: false,
            cursor: 'default',
            isPointer: false,
            interactiveChildren: true,
            globalWheel: false,
        }, options);
```

PR #10686, opened 2024-07-04
fix: Filter constructor blendMode option

dev7355608:
##### Description of change

Add missing assignment of the `options.blendMode` to the `Filter` constructor.

##### Pre-Merge Checklist


GoodBoyDigital (approved):
thanks @dev7355608 !!

bigtimebuddy (approved):

src/filters/Filter.ts
```diff
@@ -198,6 +198,7 @@ export class Filter extends Shader
 
         super(options as ShaderWithResources);
 
+        this.blendMode = options.blendMode;
```
Zyie, line 201: This change causes `BlurFilter` to break

We could do something like this 

```js
// Filter.ts
constructor() {
    ...
    if (options.blendMode) this.blendMode = options.blendMode;
}
```

```js
// BlurFilter.ts
constructor(){
    const { strength, quality, blendMode, ...rest } = options;
    super({
        ...rest,
        blendMode: null,
        compatibleRenderers: RendererType.BOTH,
        resources: {}
    });
    
    this.blendMode = blendMode ?? Filter.defaultOptions.blendMode;
}
```

Zyie (changes requested):

src/filters/defaults/blur/BlurFilter.ts
```diff
@@ -209,20 +212,6 @@ export class BlurFilter extends Filter
...
-    get blendMode(): BLEND_MODES
-    {
-        return this.blurYFilter.blendMode;
-    }
-
-    set blendMode(value: BLEND_MODES)
-    {
-        this.blurYFilter.blendMode = value;
-    }
-
```
dev7355608, line 225: Also fixes a bug, I think: In the x-blur-only case the blend mode wouldn't have been applied previously.

Zyie (approved):

PR #10698, opened 2024-07-05
fix: create instructionSet parameter from addToBatch

dev7355608:
##### Description of change

The `addToBatch` method of `BatchPipe` interface requires the instruction set being passed.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/batcher/shared/BatcherPipe.ts
+++ b/src/rendering/batcher/shared/BatcherPipe.ts
@@ -65,7 +65,7 @@ export class BatcherPipe implements InstructionPipe<Batch>, BatchPipe
         this._activeBatch.begin();
     }
 
-    public addToBatch(batchableObject: BatchableObject)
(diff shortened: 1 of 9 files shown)


GoodBoyDigital (commented):
thanks for this! Question - I wander if instead of this route, we modify the interface to not require instructionSet?

dev7355608:
Yeah, perhaps not necessary to pass `instructionSet`. If you really needed the `instructionSet`, you could remember it in `buildStart`: `this._activeInstructionSet = instructionSet`. But if this is done, we wouldn't really need to pass `instructionSet` to `break` and `buildEnd` either technically, I suppose.

GoodBoyDigital:
agreed! Currently, I can't see why `addToBatch` would need the instruction set so lets remove it from the interface for now. We can always change if it turns out i'm wrong :D

GoodBoyDigital (approved):
awesome work as usual! thanks @dev7355608

PR #10702, opened 2024-07-05
feat: support custom batch geometry

dev7355608:
##### Description of change

Support custom batch pipes extending `BatcherPipe` with custom custom geometry.

```ts
class ExampleBatcherPipe extends BatcherPipe {

    public static extension = {
        type: [
            ExtensionType.WebGLPipes,
            ExtensionType.WebGPUPipes,
            ExtensionType.CanvasPipes,
        ],
        name: 'batchExample',
    } as const;

    protected override _createBatcher(): Batcher
    {
        return new Batcher({
           renderPipeId: 'batchExample',
           floatsPerVertex: 12,
        });
    }

    protected override _createGeometry(): Geometry
    {
        return new ExampleBatchGeometry();
    }
}
```

##### Pre-Merge Checklist


src/rendering/batcher/shared/Batcher.ts
```diff
@@ -240,7 +250,13 @@ export class Batcher
         let start = this._batchIndexStart;
 
         let action: BatchAction = 'startBatch';
-        let batch = this._batchPool[this._batchPoolIndex++] || new Batch();
+        let batch = this._batchPool[this._batchPoolIndex++];
+
+        if (!batch)
+        {
+            batch = new Batch();
```
GoodBoyDigital, line 257: should we pass in `this._renderPipeId` as a constructor?

```batch = new Batch(this._renderPipeId);```

**Review by GoodBoyDigital — COMMENTED** [review]
Its great having you back on Pixi @dev7355608 !! 
Taking this in a good direction!

File: src/rendering/batcher/shared/Batcher.ts [hunk 2]
```diff
@@ -107,6 +111,8 @@ export interface BatcherOptions
 export class Batcher
 {
     public static defaultOptions: BatcherOptions = {
+        renderPipeId: 'batch',
+        floatsPerVertex: 6,
         vertexSize: 4,
         indexSize: 6,
```
dev7355608, line 117: Perhaps it might be a good idea to rename those, since the name `vertexSize` is also used to refer to the "number of float32 per vertex", to make just there the naming doesn't cause any confusion.

GoodBoyDigital:
closing this as we have added custom bater support. 👍

PR #10703, opened 2024-07-05
feat: reserve texture units for batching

dev7355608:
##### Description of change

Support reserving a number of texture unit for binding textures that are used by all batched objects.

##### Pre-Merge Checklist


GoodBoyDigital (commented):
heya, I wander if it might not be easier for users if we allow them pass in maxTextures? rather than reserved? 

They have access to `getMaxTexturesPerBatch()` so can figure out the details as they require?

dev7355608:
I think if we have a design where the user doesn't need to think about the max textures that would be good. Because they would always want to reserve a certain number of texture units and use the rest as usual. As it currently is the user would have to call `getMaxTexturesPerBatch` at some point to figure out the reserved locations to bind their textures to. Perhaps there's an argument to reserve texture units from the front instead of the back, e.g. 0 instead of 15. Then the user wouldn't have to call `getMaxTexturesPerBatch` at all.

`getMaxTexturesPerBatch` isn't actually part of the public API at the moment. It's tagged `@private`.

GoodBoyDigital:
Gotcha, my main concern is that a high number of reserved textures could be put in and that  is bigger than available max textures iyswim? eg i stick in 8 reserved textures and a mobile phone might only have 8 available I would end up with 0 for the batcher. (i know its a bit extreme, but hopefully illustrates my point)

Agree that by default we use getMaxTexturesPerBatch internally! I imagine that needing to reserve textures is likely due to some kind of custom rendering / batching work the user is about to embark on - In that situation it feels like we should let the user control this giving them the ability to allocate how many textures they want to use based on how many they have?

We would need to expose `maxTextures` though..  could stick it on the renderer context system?

dev7355608:
Replaced `reservedTextures` by `maxTextures`.

> We would need to expose `maxTextures` though.. could stick it on the renderer context system?

Perhaps the shader system is the more appropriate location: `app.renderer.shader.maxTextures`? I'll think I'll make exposing the maximum textures a separate PR.

GoodBoyDigital:
sounds good! `app.renderer.shader.maxTextures` 👍

Zyie (approved):

PR #10704, opened 2024-07-07
Fix texture matrix calculation

ShukantPal:
The texture matrix calculation was not taking into account textures embedded in a non-default frame inside a resource (e.g. canvas or image).

##### Reproduction

https://jsfiddle.net/ShukantPal/w7z5xjt9/


GoodBoyDigital (approved):
thanks @ShukantPal !

ShukantPal:
@Zyie I see you changed the base branch. I'm curious what branching system we're using now?

Zyie:
> @Zyie I see you changed the base branch. I'm curious what branching system we're using now?

hey, we use `main` for patches and `dev` for features

PR #10710, opened 2024-07-08
chore: rename Container#_onUpdate

dev7355608:
##### Description of change

The name `_onUpdate` is causing problems for our public API (Foundry VTT), where we already have a method `_onUpdate` in a class that extends `PIXI.Container`. Since the function isn't part of the public PIXI API (`@ignore` tag), making it `#`-private (or renaming it) should be not a problem hopefully.

I haven't seen anything `#`-private in the PIXI code base so I'm not sure if there isn't a reason why we cannot do it.

##### Pre-Merge Checklist

### DIFF
--- a/tests/scene/transform-tint.tests.ts
+++ b/tests/scene/transform-tint.tests.ts
@@ -33,7 +33,7 @@ describe('Transform Tints', () =>
     {
         const root = new Container();
 
-        root._onUpdate = jest.fn();
+        (root._position as any)._onUpdate = jest.fn();
 
         root.alpha = 0.5;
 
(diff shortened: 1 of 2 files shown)


tests/scene/transform-tint.tests.ts
```diff
@@ -33,36 +33,36 @@ describe('Transform Tints', () =>
...
+        (root._position as any)._onUpdate = jest.fn();
 
         root.alpha = 0.5;
 
         expect((root.alpha)).toEqual(0.5);
 
         root.alpha = 0.5;
 
-        expect(root._onUpdate).toHaveBeenCalledTimes(1);
+        expect((root._position as any)._onUpdate).toHaveBeenCalledTimes(1);
```
dev7355608, line 44: Ah, this doesn't make sense, because we test `alpha` assignment here. I guess we cannot test `#`-private methods...

GoodBoyDigital:
heya, thanks for the PR. The reason we directly call the onUpdate function is actually fro performance reasons. Its nuts, but this PR slows bunnymark down by about 40%!

dev7355608:
Oh. Was this also the the reason for the design change of `ObservablePoint`? Previously it took a callback, which make it much more flexible compared to the previous design. Or it this related to the usage of a `#`-private functions? I think I once read somewhere that they where slower than normal functions. Either way 40% slow down seems crazy with such a minor change.

GoodBoyDigital:
> Oh. Was this also the the reason for the design change of `ObservablePoint`? Previously it took a callback, which make it much more flexible compared to the current design. Or it this related to the usage of a `#`-private functions? I think I once read somewhere that they where slower than normal functions. Either way 40% slow down seems crazy with such a minor change.

100% the design change was all about the perf. Spent a lot of time trying to make v8 as fast as possible. Which was great - but came at the cost of additional complexity unfortunately! Its more of an engine maintainers problem than an end user one at least, so figured it was worth it for the perf boost!

dev7355608:
I see. I guess that's just bad luck with the name `_onUpdate` for us (Foundry VTT) then. I assume renaming `_onUpdate` to something else isn't an option for PIXI considering that it would be a breaking change for `OberservablePoint` and `Transform` and the observer interface.

Zyie:
We are going to close down this PR as changing the name of `_onUpdate` would be a breaking change for the project and the new implementation has some performance concerns

We would recommend that you update the naming conflict on your project if possible!

PR #10706, opened 2024-07-08
chore: remove dead code from Batcher

dev7355608:
##### Description of change

Perhaps this isn't dead code, but just something under construction.

##### Pre-Merge Checklist

- [ ] Tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [ ] Lint process passed (`npm run lint`)
- [ ] Tests passed (`npm run test`)

PR #10711, opened 2024-07-08
fix: Batch Pooling

GoodBoyDigital:
##### Description of change
Fixes the BatchPooling bug identified in #10706. 
This tweak actually.. you know.. uses the Pool!

##### Pre-Merge Checklist


src/rendering/batcher/shared/Batcher.ts
```diff
@@ -240,7 +240,7 @@ export class Batcher
         let start = this._batchIndexStart;
 
         let action: BatchAction = 'startBatch';
-        let batch = this._batchPool[this._batchPoolIndex++] || new Batch();
+        let batch = this._batchPool[this._batchPoolIndex++] ||= new Batch();
```
dev7355608, line 243: I think we are missing a `.bindGroup = null` in `_finishBatch`. I don't see it being reset anywhere. Not sure if it matters.

src/rendering/batcher/shared/Batcher.ts
```diff
@@ -212,7 +212,7 @@ export class Batcher
         // ++BATCH_TICK;
         const elements = this._elements;
 
-        let textureBatch = this._textureBatchPool[this._textureBatchPoolIndex++] || new BatchTextureArray();
+        let textureBatch = this._textureBatchPool[this._textureBatchPoolIndex++] ||= new BatchTextureArray();
```
dev7355608, line 215: I once read somewhere that out-of-bounds array access and assignment permanently slow down the array in Chrome, and that it is therefore is not recommended. I've seen this in other places in PIXI's code base, so I thought it might be worth bringing up.

src/rendering/batcher/shared/Batcher.ts
dev7355608, line 243: Are we worried about this causing a memory leak? A `TextureSource` might never be GC'd because some `BatchTextureArray` in the pool still references it.

I suppose a similar memory leak concern could have `Batcher#_elements`.

dev7355608 (approved):

src/rendering/batcher/shared/Batcher.ts
GoodBoyDigital: thats a good point! Il add a follow up task to investigate this in all areas

src/rendering/batcher/shared/Batcher.ts
GoodBoyDigital: agree - we have a ticket for this that involves clearing these arrays so references can be released.

PR #10712, opened 2024-07-08
fix: _didChangeId overflows

dev7355608:
##### Description of change

Prevent overflows when `_didChangeId` is incremented.

##### Pre-Merge Checklist


dev7355608:
On related note: Isn't 12 bits way too little here? After 4094 transform updates it also starts overflowing in the the upper 12 bits.
```ts
    /**
     * A value that increments each time the container is modified
     * the first 12 bits represent the container changes (eg transform, alpha, visible etc)
     * the second 12 bits represent:
     *      - for view changes (eg texture swap, geometry change etc)
     *      - containers changes (eg children added, removed etc)
     *
     *  view          container
     * [000000000000][00000000000]
     * @ignore
     */
    public _didChangeId = 0;
```

GoodBoyDigital:
You are totally right, @dev7355608 - i can't quite remember why i had the numbers merged. I have created another PR that simplifies this and splits them into two separate ids   #10771

GoodBoyDigital:
closing in favour of #10771

PR #10718, opened 2024-07-09
fix: Batch Pooling follow-up

dev7355608:
I noticed that we missing adding `||=` in two places in #10711. And also we need to clear `bindGroup` if it comes from the pool, I think.


GoodBoyDigital (approved):
nice!

PR #10721, opened 2024-07-09
chore: add bunnymark to sandbox

Zyie:
Updating the sandbox to use bunnymark so that we can more easily identify major performance hits from PR's


GoodBoyDigital (commented):
this is cool! Lets add two bits:

- lets a add a 'slow' mode (which essentially adds and removing a bunny each frame)
- lets add a `stop` mode (which stops the bunnies from moving)

Two buttons to toggle these and we will be golden! Cheers!

GoodBoyDigital (approved):
looks good!

PR #10725, opened 2024-07-10
chore: add Container#onViewUpdate

dev7355608:
##### Description of change

`_didChangeId` (tagged `@ignore`), `didViewUpdate` (tagged `@private`), `renderGroup` (tagged `@private`), `parentRenderGroup` (tagged `@private`) are all not part of the public API. But I assume every custom subclass of `Container` would need to implement a `onViewUpdate` function, but without access to these private/internal members that is not possible. So I added the `onViewUpdate` to `Container`, which does executes the code that all of the `onViewUpdate` implementations have in common.

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/container/Container.ts
+++ b/src/scene/container/Container.ts
@@ -744,6 +744,21 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
         }
     }
 
+    public onViewUpdate()
+    {
+        this._didChangeId += 1 << 12;
+
+        if (this.didViewUpdate) return;
+        this.didViewUpdate = true;
+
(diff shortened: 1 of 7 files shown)


GoodBoyDigital:
Thanks for this @dev7355608! I wander..  should we have a View class that extends container? There is a fair bit of shared code that could go in there. I have been trying to avoid this in an effort to keep things as flat as possible. Let me do a performance test to see if this has any measurable impact on v8...

GoodBoyDigital:
closing as we created a view container class to handle this. Thanks for showing us the way!

PR #10728, opened 2024-07-10
fix: Mesh shader sharing

dev7355608:
##### Description of change

`Mesh#shader` claims:
> Can be shared between multiple Mesh objects.

While this is true, it doesn't really work if the meshes different textures, because the setter of `Mesh#texture` sets `TextureShader#texture`. So if a shader is shared, `mesh.texture` may not be equal to `mesh.shader.texture`. In PIXI v7 mesh shader sharing was even less possible, because the `tint` was also owned by the shader. But this has changed now with introduction of the `localUniforms`.

With this PR the mesh pipe adapter binds the texture uniforms as `textureUniforms` like `localUniforms`/`globalUniforms` would be. So mesh shader no longer needs to define the `texture` setter, and the shader can be shared among meshes with different textures.

These changes (seem to) work, but I'm not sure they are the optimal implementation. I'm still learning the new API.

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/mesh/gl/GlMeshAdaptor.ts
+++ b/src/scene/mesh/gl/GlMeshAdaptor.ts
@@ -1,11 +1,9 @@
 import { ExtensionType } from '../../../extensions/Extensions';
-import { Matrix } from '../../../maths/matrix/Matrix';
 import { compileHighShaderGlProgram } from '../../../rendering/high-shader/compileHighShaderToProgram';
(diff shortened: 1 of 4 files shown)


dev7355608:
With the changes in #10772 this PR should simplify a lot.

GoodBoyDigital:
closing this one for now - will reopen if the need arises!

PR #10731, opened 2024-07-10
fix: insufficient buffer for graphics

ivanpopelyshev:
The solution: re-use batches in graphics separate from re-using batchers

I spent 3 hours debugging what went wrong in huge application.

Batchers are re-used every frame, however graphics "Batch" can be stored for long time, and it cannot be re-used.

Frame1. Create small geom.
Frame2. Create big geom. 
big geom gets same batches that are used in small geom , and overrides small geom graphicsData.

With my current workload, I dont have time to reproduce it with clean test, I hope someone can do that.

I assign this "critical" label because its potentially breaking any big app using graphics.


GoodBoyDigital:
thanks @ivanpopelyshev will take a look

Zyie:
closing in favor of #10741

PR #10732, opened 2024-07-11
fix: Check if transform is set on lastShape before dereferencing

ShukantPal:
I found this while working on https://github.com/ShukantPal/pixi-essentials/pull/107

This fixes an issue where `_ensurePoly` called after closing a polygon without a transform matrix set. This can happen if `lineTo` is called after `fill`.

For a reproduction, please see:
https://jsfiddle.net/ShukantPal/25g6jrth/


GoodBoyDigital (approved):
nice

PR #10734, opened 2024-07-11
fix: TextureMatrix#clampOffset calculation

dev7355608:
##### Description of change

As I understand `clampOffset` is in real pixels like the `clampMargin` and so needs to be converted first as well before being used in the `uClampFrame` calculation.

##### Pre-Merge Checklist


GoodBoyDigital:
Heya, thanks for the PR would it be possible to validate this with a visual test?

ivanpopelyshev (approved):
Does this fix some of your cases?

Zyie (changes requested):
Just requesting changes until test is added for this PR, so it doesn't accidently get merged

dev7355608:
I added a test, but the test images don't look right given the different results of webgl1 and webgl2, and webgl2/webgpu just look like `clampOffset` is 0. I'm not sure what exactly `clampOffset` is supposed to do and what to expect. It looks like there's some other bug here related to clamp offset. The test confirm though that without this PR, the two sprites with the same texture at different resolutions would not look the same.

ivanpopelyshev:
I confirm that this thing sholud be done using real pixels

GoodBoyDigital (approved):
thanks for providing the test! this PR looks good

Zyie (approved):

dev7355608:
The test images that was merged with this PR is incorrect. If you look at the test images, you see that there must be a bug related to clamp offset (but unrelated to the bug that this is PR fixed) given that the webgl1 image looks completely different.

Zyie:
> The test images that was merged with this PR is incorrect. If you look at the test images, you see that there must be a bug related to clamp offset (but unrelated to the bug that this is PR fixed) given that the webgl1 image looks completely different.

looks like clampOffset is 0 in webgl1 but 13 in webgl2/webgpu

PR #10747, opened 2024-07-12
chore: optimize Mesh instructions

dev7355608:
##### Description of change

I understand correctly, we should be able to the use `Mesh` itself as the instruction.

##### Pre-Merge Checklist


GoodBoyDigital (approved):
nice !

PR #10758, opened 2024-07-15
fix: change renderPipeId types to string

dev7355608:
##### Description of change

Allows subclasses of `Mesh` etc. to change `renderPipeId` without typescript complaining:

```ts
class MyMesh extends Mesh {
    public readonly renderPipeId: string = 'myMesh';
}
```

Also changed `renderPipeId` of `Container` and `View` to readonly. I don't think switching pipes is supported (at the moment).

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/container/RenderContainer.ts
+++ b/src/scene/container/RenderContainer.ts
@@ -83,7 +83,7 @@ export class RenderContainer extends Container implements View, Instruction
     public addBounds: (bounds: Bounds) => void;
 
     public canBundle = false;
-    public renderPipeId = 'customRender';
+    public readonly renderPipeId: string = 'customRender';
 
     /**
      * @param options - The options for the container.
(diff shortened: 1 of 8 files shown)


src/scene/container/RenderContainer.ts
```diff
@@ -86,7 +86,7 @@ export class RenderContainer extends Container implements View, Instruction
     public addBounds: (bounds: Bounds) => void;
 
     public canBundle = false;
-    public renderPipeId = 'customRender';
+    public readonly renderPipeId: string = 'customRender';
```
GoodBoyDigital, line 89: educate an idiot here... but would `public readonly renderPipeId = 'customRender';` not infer a string? thanks!

dev7355608: It becomes 

```ts
public readonly renderPipeId: 'customRender' = 'customRender';
```

The `readonly` affects the inferred type: without it we get `string` as expected.

GoodBoyDigital: ahh! thanks

GoodBoyDigital (approved):
lgtm!

PR #10759, opened 2024-07-15
chore: replace Mesh with Mesh<Geometry> in MeshPipe

dev7355608:
##### Description of change

The `MeshPipe` handles `Mesh<Geometry>` and not just `Mesh` (`Mesh<MeshGeometry>`).

##### Pre-Merge Checklist


src/scene/mesh/shared/MeshPipe.ts
```diff
@@ -89,11 +91,24 @@ export class MeshPipe implements RenderPipe<Mesh>, InstructionPipe<MeshInstructi
...
+
+                meshData.indexSize = geometry.indices.length;
+                meshData.vertexSize = geometry.positions.length;
+            }
+            else
+            {
+                meshData.indexSize = 0;
+                meshData.vertexSize = 0;
+            }
+
```
dev7355608, line 106: I just noticed this in passing: if the `indexSize` and `vertexSize` is not updated in this branch, I expect a second rebuild if the shader from a mesh is removed and at the same time the geometry is changed. Because in the next branch the `indexSize`/`vertexSize` are not equal.

GoodBoyDigital:

PR #10763, opened 2024-07-16
feat: expose world position to end block of the default vertex template

dev7355608:
##### Description of change

The world position can be useful in the `{{end}}` block to calculate the UV coordinates of the position within the screen space so that you can sample from a screen-sized texture within the fragment shader.

```glsl
vScreenUV = worldPosition.xy / uScreenDimensions;
```

I suppose you could just do the following in the end block (calculate the world position yourself), but I'm not sure if the optimizer is smart enough. But I also haven't benchmarked whether the changes of the PR affect performance: I'd think multiplying the matrices first would be less performant, because it's more multiplications in theory, compared to applying each matrix to the vector. But I guess the GPU might just do the matrix multiplication once, because it doesn't depend of the vertex data.
```glsl
vScreenUV = (worldTransformMatrix * modelMatrix * vec3(position, 1.0)).xy / uScreenDimensions;
```

Also removed unnecessary trailing whitespace.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/high-shader/defaultProgramTemplate.ts
+++ b/src/rendering/high-shader/defaultProgramTemplate.ts
@@ -25,17 +25,17 @@ const vertexGPUTemplate = /* wgsl */`
         var uv = aUV;
 
         {{start}}
-        
+
         vColor = vec4<f32>(1., 1., 1., 1.);
 
         {{main}}
 
         vUV = uv;
 
-        var modelViewProjectionMatrix = globalUniforms.uProjectionMatrix * worldTransformMatrix * modelMatrix;
+        var worldPosition = worldTransformMatrix * modelMatrix * vec3<f32>(position, 1.0);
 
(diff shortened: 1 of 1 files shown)


dev7355608:
I realized that the use case for `worldPosition` mentioned above doesn't quite work correctly if `roundPixels = true`, because we want to calculate the screen coordinates based on the rounded position in this case. For perhaps it is necessary because of this to calculate the screen coordinates like so
```glsl
vScreenUV = (gl_Position.xy * uScreenTransform.xy) + uScreenTransform.zw;
```
where `uScreenTransform` was calculated from the projection matrix and the offset of the global uniforms. But the offset isn't publicly exposed at the moment.

dev7355608:
I realized something else: currently the offset is added to the world transform matrix instead of the projection matrix. So currently `worldTransformMatrix * modelMatrix * vec3(position, 1.0)` isn't actually the world position, but the position relative to the filter texture frame position. I propose to add the offset to the projection matrix instead (which was the case in v7, I think). Otherwise there's no way to calculate the world position given that the offset isn't exposed.

GoodBoyDigital:

PR #10764, opened 2024-07-16
chore: compile batch adaptor shaders with ShaderSystem#maxTextures

dev7355608:
##### Description of change

This probably doesn't really change anything in practice, but I though it would be good to get rid of the `getMaxTexturesPerBatch` call and use `renderer.shader.maxTextures` instead.

##### Pre-Merge Checklist


GoodBoyDigital:
heya, thanks for this. Since we have refactored the batcher to allow for custom batching, the use of `getMaxTexturesPerBatch` is handy as it means athe `Batch` class does not need to rely on any renderer information. 
It can also be overwritten with `Batcher.defaultOptions.maxTextures`. Closing for now

PR #10766, opened 2024-07-16
fix: update view if roundPixels is changed and update batches

dev7355608:
##### Description of change

If `Container#roundPixels` is changed, we need to update the view and update the batches.

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/mesh/shared/MeshPipe.ts
+++ b/src/scene/mesh/shared/MeshPipe.ts
@@ -126,6 +126,7 @@ export class MeshPipe implements RenderPipe<Mesh>, InstructionPipe<Mesh>
 
             gpuBatchableMesh.texture = mesh._texture;
             gpuBatchableMesh.geometry = mesh._geometry;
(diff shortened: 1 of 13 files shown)


src/scene/mesh/shared/MeshPipe.ts
```diff
@@ -215,7 +217,7 @@ export class MeshPipe implements RenderPipe<Mesh>, InstructionPipe<Mesh>
         // TODO - make this batchable graphics??
         const gpuMesh: BatchableMesh = BigPool.get(BatchableMesh);
 
-        gpuMesh.mesh = mesh;
```
dev7355608, line 218: `mesh` is assigned a couple of lines below. I assume it is supposed to be a `geometry` assignment here. But the missing `geometry` assignement didn't cause a bug, because we currently always update the batchable mesh in `addRenderable`/`updateRenderable` (there's no `_didMeshUpdate` check like with Sprite).

GoodBoyDigital:
closing this one for now - will reopen if the need arrises! @Zyie - lets maybe just update the docs around round pixels to be 'set once' and then let devs know how to refresh for this edge case

PR #10770, opened 2024-07-17
fix: delete hash entries instead of assigning null

dev7355608:
##### Description of change

If we don't delete, we'll have `null` entries for destroyed objects forever since the uids are not recycled, I think.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/batcher/shared/BatchTextureArray.ts
+++ b/src/rendering/batcher/shared/BatchTextureArray.ts
@@ -29,7 +29,7 @@ export class BatchTextureArray
             const t = this.textures[i];
 
             this.textures[i] = null;
-            this.ids[t.uid] = null;
+            delete this.ids[t.uid];
         }
 
         this.count = 0;
--- a/src/assets/loader/workers/WorkerManager.ts
+++ b/src/assets/loader/workers/WorkerManager.ts
@@ -112,7 +112,7 @@ class WorkerManagerClass
             this._resolveHash[data.uuid].resolve(data.data);
         }
 
-        this._resolveHash[data.uuid] = null;
+        delete this._resolveHash[data.uuid];
     }
 
     private async _run(id: string, args: any[]): Promise<any>
--- a/src/rendering/renderers/gl/buffer/GlBufferSystem.ts
+++ b/src/rendering/renderers/gl/buffer/GlBufferSystem.ts
@@ -195,7 +195,7 @@ export class GlBufferSystem implements System
             gl.deleteBuffer(glBuffer.buffer);
         }
 
-        this._gpuBuffers[buffer.uid] = null;
+        delete this._gpuBuffers[buffer.uid];
     }
 
     /**
--- a/src/rendering/renderers/gl/geometry/GlGeometrySystem.ts
+++ b/src/rendering/renderers/gl/geometry/GlGeometrySystem.ts
@@ -332,7 +332,7 @@ export class GlGeometrySystem implements System
                 }
             }
 
-            this._geometryVaoHash[i] = null;
+            delete this._geometryVaoHash[i];
         }
     }
 
--- a/src/rendering/renderers/gl/geometry/GlGeometrySystem.ts
+++ b/src/rendering/renderers/gl/geometry/GlGeometrySystem.ts
@@ -303,7 +303,7 @@ export class GlGeometrySystem implements System
                 }
             }
 
-            this._geometryVaoHash[geometry.uid] = null;
+            delete this._geometryVaoHash[geometry.uid];
         }
     }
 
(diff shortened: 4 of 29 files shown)


src/rendering/batcher/shared/BatchTextureArray.ts
```diff
@@ -29,7 +29,7 @@ export class BatchTextureArray
             const t = this.textures[i];
 
             this.textures[i] = null;
-            this.ids[t.uid] = null;
```
GoodBoyDigital, line 32: you are not wrong, but delete is a suuppperr slow operation

dev7355608: Yeah, but I imagine if these object have tens of thousands of unnecessary entries that can't be good either. In theory you could probably run out of memory.

GoodBoyDigital: maybe there is like a middle ground - could just reset them every X frames ?

GoodBoyDigital: My main concern is that temoving keys changes the object's shape and can lead to de-optimization.

GoodBoyDigital (changes requested):
Had a think about this - as this is a common thing (especially across systems).

lets use the `SchedulerSystem` to periodically run the following every 60 secs?:

```
const objectToClean

cont newObject = Object.create(null);

for(var i in objectToClean)
{
    const item = objectToClean[i]
    if(item)newObject[i] = item;
}

return newObject;
```

what do you think @dev7355608 ?

for the `BatchTextureArray` thats SUPER hot code. It should not get too big as its only using textures which are generaly fairly stable. Will have a think though - we could do something similar, like periodically clean the batchers...

dev7355608:
I was wondering about Object vs Map. I assume it was benchmarked for these integer key records? I think recall Maps performing better for this kind of use case, especially with a lot of entries. I can't imagine this being great for performance:
```js
for (const i in object)
```
because in theory either the integer key was already converted to a string at the time the entry was put in the object or it is converted when we iterate it. I'm not sure if the optimizer is so smart that no string is created here.
With Maps we don't have this problem that our keys are strings.

dev7355608:
Regarding the `BatchTextureArray`: Considering the number or textures is small, perhaps iterating over the array instead is just as fast that the object access and so we don't need `ids` at all?

```ts
    public getTextureId(texture: TextureSource): number
        const uid = texture.uid;
        const textures = this.textures;
        const count = this.count;

        for (let i = 0; i < count; i++)
        {
            if (textures[i].uid === uid) return i;
        }

        return -1;
    }
```

GoodBoyDigital:
> I was wondering about Object vs Map. I assume it was benchmarked for these integer key records? I think recall Maps performing better for this kind of use case, especially with a lot of entries. I can't imagine this being great for performance:
> 
> ```js
> for (const i in object)
> ```
> 
> because in theory either the integer key was already converted to a string at the time the entry was put in the object or it is converted when we iterate it. I'm not sure if the optimizer is so smart that no string is created here. With Maps we don't have this problem that our keys are strings.

Maps don't tend to cause de-optimisations which is great, but annoyingly reading from them is waaay slower than reading from a standard object..

Every PR creates a bunnymark sandbox now, so it should at least be easy to test this stuff!

GoodBoyDigital:
> Regarding the `BatchTextureArray`: Considering the number or textures is small, perhaps iterating over the array instead is just as fast that the object access and so we don't need `ids` at all?
> 
> ```ts
>     public getTextureId(texture: TextureSource): number
>         const uid = texture.uid;
>         const textures = this.textures;
>         const count = this.count;
> 
>         for (let i = 0; i < count; i++)
>         {
>             if (textures[i].uid === uid) return i;
>         }
> 
>         return -1;
>     }
> ```

could work, we can bench mark this one for sure!

GoodBoyDigital:
hey, had another thought for pipes.. we could attach the objects directly to the renderable itself (eg Sprite / Graphics)
this could be done by adding some kind of `_gpuRenderCache` typed as `unknown` and then the pipe can declare the type it wants. 

That way we don't need the majority of the hashes and we still keep the Renderabels agnostic to Canvas/WebGPU/WebGL?

What doya think?

dev7355608:
The `_gpuRenderCache` property sounds good. But then you can't render the the same container in different renderers. But I'm not sure that is something that is actually supported. Just I thought I had.

GoodBoyDigital:
heya @dev7355608 followed up with this PR #10793

I tried adding the items directly to the sprite - and for some reason its slower?? Perhaps many objects = slower than a single large one?

With that in mind i wend for having the hashes periodically cleaned. I will open an issue for the BatchTextureArray. This one is in the heart of the rendering loop, so keen to find a solution that does not compromise on perf!

GoodBoyDigital:
closing in favour of #10793 . Thanks for identifying and providing the initial fix for this one @dev7355608 !!

PR #10771, opened 2024-07-17
fix: split out update id and view update id

GoodBoyDigital:
added backwards compatability

##### Description of change
Rather than attempting to be overly clever and store two states in a single id, I have split them into two seperate ones for simplicity! There was no real reason any more for them to be merged into one!

Kept `_didChangeId` around for backwards compatablility (eg spine-v8).

Alternative to #10712

##### Pre-Merge Checklist


src/scene/container/Container.ts
```diff
@@ -561,17 +561,34 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
...
+    /**
+     * A value that increments each time the container view is modified
+     * eg texture swap, geometry change etc
+     * @ignore
+     */
+    public _didViewChangeTick = 0;
+
+    /**
+     * We now use the _didContainerChangeTick and _didViewChangeTick to track changes
+     * @deprecated since 8.2.1
```
Zyie, line 578:
```suggestion
     * @deprecated since 8.2.6
```

src/scene/container/Container.ts
```diff
@@ -561,17 +561,34 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
...
-    public _didChangeId = 0;
+    set _didChangeId(value: number)
+    {
+        this._didViewChangeTick = (value >> 12) & 0xFFF; // Extract the upper 12 bits
+        this._didContainerChangeTick = value & 0xFFF; // Extract the lower 12 bits
+    }
+
+    get _didChangeId(): number
+    {
+        return (this._didContainerChangeTick % 0xFFF) | ((this._didViewChangeTick % 0xFFF) << 12);
```
dev7355608, line 589:
```suggestion
        return (this._didContainerChangeTick & 0xFFF) | ((this._didViewChangeTick & 0xFFF) << 12);
```

dev7355608 (changes requested):
I assume there's follow up work planned to address the bounds cache checks (?):
```js
c = new PIXI.Container();
s = new PIXI.Sprite();
c.addChild(s);

console.log(c.getLocalBounds()); //  {minX: 0, minY: 0, maxX: 1, maxY: 1}

for (let i = 0; i < 4096; i++) s.x++; // assuming the overflow into the upper 12 bits is fixed

console.log(c.getLocalBounds()); //  {minX: 0, minY: 0, maxX: 1, maxY: 1}
```

GoodBoyDigital:
> ```js
> c = new PIXI.Container();
> s = new PIXI.Sprite();
> c.addChild(s);
> 
> console.log(c.getLocalBounds()); //  {minX: 0, minY: 0, maxX: 1, maxY: 1}
> 
> for (let i = 0; i < 4096; i++) s.x++; // assuming the overflow into the upper 12 bits is fixed
> 
> console.log(c.getLocalBounds()); //  {minX: 0, minY: 0, maxX: 1, maxY: 1}
> ```

let me check that one now!

GoodBoyDigital:
@dev7355608 i forgot to add both the mask AND the modulo. This should works as intended now! Including wraping correctly. (Added a test)

dev7355608:
Why do we have both? `(x % 0xFFF) & 0xFFF` is equivalent to `x % 0xFFF` for `x >= 0`. And with `x % 0xFFF` instead of `x & 0xFFF` it wraps after 4095 instead 4096 increases. So I expect this to fail now with the changes of this PR:

```js
c = new PIXI.Container();
s = new PIXI.Sprite();
c.addChild(s);

console.log(c.getLocalBounds()); //  {minX: 0, minY: 0, maxX: 1, maxY: 1}

for (let i = 0; i < 4095; i++) s.x++; // assuming the overflow into the upper 12 bits is fixed

console.log(c.getLocalBounds()); //  {minX: 0, minY: 0, maxX: 1, maxY: 1}
```

GoodBoyDigital:
fixed the test :D

tests/scene/checkChildrenDidChange.tests.ts
```diff
@@ -9,13 +9,13 @@ describe('checkChildrenDidChange', () =>
         const container = new Container();
         const changeId = getChangeId(container);
 
-        expect(changeId).toEqual(16777216);
+        expect(changeId).toEqual(0);
 
         container._didContainerChangeTick = 4096;
```
dev7355608, line 14: The test will fail if you change it to 4095. The weird math (`(x % 0xFFF) & 0xFFF`) just makes it wrap 1 earlier.

> (4095 % 0xFFF) & 0xFFF === 0

`(x % 0xFFF) & 0xFFF` isn't any better than `x & 0xFFF`. In fact it's a tiny bit worse.

dev7355608 (changes requested):

tests/scene/checkChildrenDidChange.tests.ts
GoodBoyDigital: oh yeah! Silly me, I'Il change it back to your suggestion!

I think ultimatly the chances of this happening are pretty slim as you would need to do _exactly_ 4096 transforms between calling getLocalBounds().

Went for this approach so that checking for local bounds changes can be as fast as possible for effective caching. 

What do ya think, acceptible risk?

dev7355608: It doesn't feel super low. We should at least document it.

Do we need the upper 8 bits for the `uid`: `((container.uid & 255) << 24)`? If we do need it, I feel like there's another case where we have chance the the bounds could be wrong. After all there's a 1 in 256 chance that the upper 8 bits collide. If we don't need it we could use the 16 instead of 12 bits for each id, which would reduce the chances significantly that we have a collision.

GoodBoyDigital:
Let me take a look today, on reflection maybe I had been focused too much
on perf here! Will have a think around alternate routes.

GoodBoyDigital:
hey @dev7355608 think this will do the trick!
down side is that we need an extra comparison per child when checking if it changed...
but the upside - is that the chance of a collision are now _much_ lower 4096 -> 65535.

please review when you get a mo! thanks!

dev7355608 (approved):

Zyie (approved):

PR #10772, opened 2024-07-18
fix mesh texture on GPU

GoodBoyDigital:
##### Description of change
Provides a fix for #10770. We need to use a unique bind group per texture, luckily we had that funcainaly built into the GPUTextureSystem but it was not being used. Tweaked it and now non-batched meshes that have the default shader will have the correct texture.

Test added

##### Pre-Merge Checklist


src/rendering/renderers/gpu/texture/GpuTextureSystem.ts
```diff
@@ -204,23 +205,33 @@ export class GpuTextureSystem implements System, CanvasGenerator
...
         const source = texture.source;
 
-        const bindGroupId = source.uid;
-
-        this._bindGroupHash[bindGroupId] = new BindGroup({
+        this._bindGroupHash[texture.uid] = new BindGroup({
             0: source,
             1: source.style,
+            2: new UniformGroup({
+                uTextureMatrix: { type: 'mat3x3<f32>', value: texture.textureMatrix.mapCoord },
```
dev7355608, line 230:
```suggestion
                uTextureMatrix: { type: 'mat3x3<f32>', value: texture.textureMatrix.mapCoord },
                uClampFrame: { type: 'vec4<f32>', value: texture.textureMatrix.uClampFrame },
                uClampOffset: { type: 'vec2<f32>', value: texture.textureMatrix.uClampOffset },
```

I think we should be adding the other texture uniforms here too. We'd need it for #10728 at least.

src/rendering/renderers/gpu/texture/GpuTextureSystem.ts
```diff
@@ -204,23 +205,33 @@ export class GpuTextureSystem implements System, CanvasGenerator
         return this._gpuSources[source.uid] || this.initSource(source);
     }
 
+    /**
+     * this returns s bind group for a specific texture, the bind group contains
+     * - the texture source
+     * - the texture style
+     * - the texture matrix
```
dev7355608, line 212:
```suggestion
     * - the texture matrix
     * - the clamp frame
     * - the clamp offset
```

src/rendering/renderers/gpu/texture/GpuTextureSystem.ts
GoodBoyDigital: This would break the mesh WebGPU currently as is the shader needs to have the exact uniform structure in the shader. I do see you point though. 

How about we get this one in as it fixes the rope bug and we follow with this when reviewing the Shared Shader sharing PR?

dev7355608 (approved):

CoreyDotCom:
Waiting to pounce on this fix when it lands. ;) Thanks for fixing.

Zyie (approved):

PR #10774, opened 2024-07-18
fix: attributes locations

dev7355608:
##### Description of change

The geometry should be compatible regardless of the locations that the shader program has assigned.

##### Pre-Merge Checklist


src/rendering/renderers/gpu/pipeline/PipelineSystem.ts
```diff
@@ -277,24 +277,50 @@ export class PipelineSystem implements System
         {
             const attribute = geometry.attributes[attributeKeys[i]];
 
-            keyGen[index++] = attribute.location;
             keyGen[index++] = attribute.offset;
             keyGen[index++] = attribute.format;
             keyGen[index++] = attribute.stride;
+            keyGen[index++] = attribute.instance;
```
dev7355608, line 283: Seems like it should be included since we use it in `_createVertexBufferLayouts`.

src/rendering/renderers/gpu/pipeline/PipelineSystem.ts
```diff
@@ -277,24 +277,50 @@ export class PipelineSystem implements System
...
         }
 
-        const stringKey = keyGen.join('');
+        const stringKey = keyGen.join('|');
```
dev7355608, line 286: Probably needs a separator here.

src/rendering/renderers/gl/shader/program/ensureAttributes.ts
```diff
@@ -27,7 +27,6 @@ export function ensureAttributes(
 
         if (attributeData)
         {
-            attribute.location ??= attributeData.location;
             attribute.format ??= attributeData.format;
             attribute.offset ??= attributeData.offset;
             attribute.instance ??= attributeData.instance;
```
dev7355608, line 32: BTW, do we really need this? If the geometry is used for the first time with a shader that only uses a subset of the attributes this won't work. And if the geometry already specified format offset and instance correctly, we get a warning anyway, even though no filling-in is needed.

GoodBoyDigital (approved):
nice work @dev7355608 ! 
location was added before the inference work, agree its no longer needed as we can get it from the Program.

(the next change would be to infer the types for the uniforms values again (as we do in v7! This would make uniform creation much nicer!).

PR #10776, opened 2024-07-19
fix: ImageSource canvas drawImage supplying width and height

Kethen:
It was observed on nwjs 0.71.0, on pixi 8.1.4, that changing HTMLImageElement.width and HTMLImageElement.height prior to ImageSource creation, would lead to a texture created with a canvas of the new width and height, but image rendered in original size

Supply width and height to drawImage to avoid such behavior

##### Pre-Merge Checklist


GoodBoyDigital (approved):
thanks! looks good!

PR #10778, opened 2024-07-19
feat: filter clear color

dev7355608:
##### Description of change

Currently you have to add a graphics/sprite behind the object, wrap it in a container and apply the filter to the container (or if it is already a container, add the graphics as first child). It would be a bit more convenient and a bit more performant to change the clear color of the filter input texture instead.

The use case is probably pretty niche though, but sometimes you want to apply the filter also to the background color the object is rendered to, if that makes sense.

##### Pre-Merge Checklist


GoodBoyDigital:
Hey @dev7355608 , thanks for this PR. After discussing with the team, we feel that this is a little to niche for us to add to the core. We are mindful about not expanding the API unless we really need to. This situation could be solved by customizing the applyFilter method, or by including the fill color in a custom shader. Thanks man!

PR #10779, opened 2024-07-19
fix: updateQuadBounds padding

dev7355608:
##### Description of change

I noticed this in passing that the padding calculation is wrong here. Currently the padding doesn't but shifts the bounds. Potentially/probably related to reports that text padding is wrong in v8 (?).

##### Pre-Merge Checklist


bigtimebuddy:
We need a test for this

GoodBoyDigital (approved):
nice! Mind adding a test to cover this pls?

ivanpopelyshev:
I dont know whether its my issue , im testing it...

k9p5:
I tested these changes on the dev branch (latest) and it stretched the text, while the Y axis was disproportional.

dev7355608:
I also suspected that something else might be wrong with the text padding, but I didn't look at it further. Maybe padding isn't even the right thing to do for the text. It's not padding what `updateQuadBounds` is doing. Perhaps the parameter is just poorly named. It just looked wrong to me.

ivanpopelyshev:
> I also suspected that something else might be wrong with the text padding, but I didn't look at it further. Maybe padding isn't even the right thing to do for the text. It's not padding what `updateQuadBounds` is doing. Perhaps the parameter is just poorly named. It just looked wrong to me.

Yes, we have to understand what is going on there. Something is fishy and I already think its not padding, because my texts are wrong with padding=0

GoodBoyDigital:
heyall! closing this as replaced by #11182

Issue #10789, opened 2024-07-23
Bug: Visual artifacts when changing HTMLText Size


bigtimebuddy:
### Current Behavior

When changing the fontSize of an HTMLText, there's unexpected hairline artifacts that occurs.  See example:

https://jsfiddle.net/bigtimebuddy/f29z5vod/

![Jul-23-2024 17-27-38](https://github.com/user-attachments/assets/7abd0835-2805-4c9e-8421-99b771183b3c)

### Expected Behavior

No artifacts.

### Steps to Reproduce

https://jsfiddle.net/bigtimebuddy/f29z5vod/

### Environment

- **`pixi.js` version**: 8.2.5
- **Browser & Version**: Chrome Latest
- **OS & Version**: MacOS Latest
- **Running Example**: [*e.g. <https://pixiplayground.com/>*](https://jsfiddle.net/bigtimebuddy/f29z5vod/16/)

### Possible Solution

Maybe this is an internal caching texture issue?

PR #10795, opened 2024-07-25
fix: html text padding and uv issue

GoodBoyDigital:
##### Description of change
After investigating, it turns out that the padding was being applied twice in the creation of the HTML text (once to the outer div and a second time to an inner div). This has now been fixed, so the padding is only applied once when creating the HTML Fragment.

Padding now does not visually change the texts postion - just like Canvas Text. Its used to give extra space if a font is being cropped (which can happen if a font is extremly skewed!)

I also added a little bit of uv padding to the Image to solve #10789. This will stop bleeding from the previous thing rendered by the texture. (This is cool as it saves us having to actually clear the texture!)

##### Pre-Merge Checklist


bigtimebuddy (approved):
Verified the fixes here work. Nice work!
https://jsfiddle.net/bigtimebuddy/fL1mx0u4/

PR #10803, opened 2024-07-29
fix: alphaMode always respected on texture load

davidetan:
##### Description of change
`Assets.load` allows specifying `alphaMode`. However, under the following (default) conditions:
- `webgl`
- `preferWorkers: false`
- `preferCreateImageBitmap: false`

`alphaMode` is ignored because `createImageBitmap` is used without passing the [`premultiplyalpha` option](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap#premultiplyalpha) to the function. 

This PR allows passing the `alphaMode` desired from the user down to the `createImageBitmap` function both in case worker or not is used to load the image.

Adding a test for this is complicated, but I've created a Pixi playground to easily view the difference: https://www.pixiplayground.com/#/edit/4b22JPxGTsa0Thdb3kx4f

If you set the `preferWorkers` preference to `false`, the texture is loaded correctly - the clouds should be more white, like this:
image

Rather than greyish like this:
image

##### Pre-Merge Checklist


GoodBoyDigital (approved):
nice! I pushed a visual test to cover this - pre-multiply alpha is like my Achilles heal with Web gPU programming :P

Zyie (approved):

davidetan:
> nice! I pushed a visual test to cover this - pre-multiply alpha is like my Achilles heal with Web gPU programming :P

Argh, sorry, I didn't think about a visual test for this!
I was thinking about something more complicated with unit tests.
Next time I'll look more thoroughly at existing tests!

PR #10813, opened 2024-08-01
fix: Support options for AnimatedSprite

bigtimebuddy:
### Changes

* Support sprite constructor arguments with AnimatedSprite

### DIFF
--- a/src/scene/sprite-animated/AnimatedSprite.ts
+++ b/src/scene/sprite-animated/AnimatedSprite.ts
@@ -1,7 +1,6 @@
 import { Texture } from '../../rendering/renderers/shared/texture/Texture';
 import { UPDATE_PRIORITY } from '../../ticker/const';
 import { Ticker } from '../../ticker/Ticker';
-import { deprecation } from '../../utils/logging/deprecation';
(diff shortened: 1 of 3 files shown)


src/scene/sprite-animated/AnimatedSprite.ts
```diff
@@ -110,19 +127,39 @@ export class AnimatedSprite extends Sprite
...
     private _previousFrame: number;
-
     /**
-     * @param textures - An array of {@link Texture} or frame
-     *  objects that make up the animation.
-     * @param {boolean} [autoUpdate=true] - Whether to use Ticker.shared to auto update animation time.
+     * @param frames - Collection of textures or frames to use.
+     * @param autoUpdate - Whether to use Ticker.shared to auto update animation time.
+     */
+    constructor(frames: AnimatedSpriteFrames, autoUpdate?: boolean);
```
trezy, line 134: Should we also mark this old constructor as `@deprecated`?

bigtimebuddy: Also, I'd rather be conservative with the deprecation to get it in a patch release.

Zyie (approved):

PR #10825, opened 2024-08-04
fix: autoDensity getter/setter

Zyie:
`autoDensity` in `ViewSystem` has not been implemented correctly and the value is never set

`autoDensity` is also controlled by `this.texture` and all the resizing logic is contained within `CanvasSource` 

so this PR simply adds a getter/setter to expose autoDensity correctly and removes unused code


src/rendering/renderers/shared/view/ViewSystem.ts
```diff
@@ -139,7 +139,14 @@ export class ViewSystem implements System<ViewSystemOptions, TypeOrBool<ViewSyst
...
      */
-    public autoDensity: boolean;
+    public get autoDensity(): boolean
+    {
+        return (this.texture.source as CanvasSource).autoDensity;
+    }
+    public set autoDensity(value: boolean)
+    {
+        (this.texture.source as CanvasSource).autoDensity = value;
+    }
```
bigtimebuddy, line 149: This change inspired #10872

GoodBoyDigital (approved):

PR #10826, opened 2024-08-05
fixed: RenderTexture.create return RenderTexture

DexYang:
##### Description of change
RenderTexture.create should return RenderTexture, otherwise the resize function is no way to use.


Zyie:
hey @DexYang thanks for the PR, however this is a duplicate of #10802

DexYang:
😭really thought I catch on something

Issue #10834, opened 2024-08-06
Bug: RenderableGCSystem adds managed views to its internal Array even when disabled


lunarraid:
### Current Behavior

When setting `renderableGCActive: false` in Application init, the RenderableGCSystem continues to add views to its managed views array. Since it won't even run when disabled, this leads to a memory leak as the Array adds more and more views and never removes them.

### Expected Behavior

RenderableGCSystem should not add managed views when it is disabled, or classes calling `add` on the system should not do so if the system is not enabled.

### Steps to Reproduce

Set `renderableGCActive: false` in Application init. Use debugger to observe that views are still being added to the system.

### Environment

- **`pixi.js` version**: *8.3.0*

Issue #10586, opened 2024-05-28
Bug: Memory leak in Graphics destruction in v8


MattTreichelYeah:
### Current Behavior

Rapidly creating and destroying Graphics in v8 leads to a memory leak in at least Chrome, and eventually crashes with a `RangeError: Invalid array length`. Stack Trace:

```
_initContext (GraphicsContextSystem.ts:255)
updateGpuContext (GraphicsContextSystem.ts:151)
addRenderable (GraphicsPipe.ts:109)
collectAllRenderablesSimple (buildInstructions.ts:87)
collectAllRenderables (buildInstructions.ts:71)
collectAllRenderablesAdvanced (buildInstructions.ts:142)
buildInstructions (buildInstructions.ts:47)
render (RenderGroupSystem.ts:116)
emit (SystemRunner.ts:74)
render (AbstractRenderer.ts:298)
render (Application.ts:155)
emit (TickerListener.ts:64)
update (Ticker.ts:461)
_tick (Ticker.ts:163)
requestAnimationFrame (async)
_tick (Ticker.ts:167)
requestAnimationFrame (async)
_tick (Ticker.ts:167)
requestAnimationFrame (async)
_tick (Ticker.ts:167)
...
```

### Expected Behavior

In Pixi v7, this example is stable.

### Steps to Reproduce

The following example reproduces it, linked here: https://www.pixiplayground.com/#/edit/3ocZP-9nVFphFe1ozWvsQ. Chrome's "Memory" heap slowly balloons and eventually crashes with the RangeError after about 15 minutes for me. Setting it to the latest v7 version and running it in parallel doesn't run into the same problems.

If I change the destroy line to `.destroy()` or `.destroy(true)`, it doesn't seem to make a difference.

```
(async () => {
    const app = new PIXI.Application();
    await app.init({
        width: window.innerWidth,
        height: window.innerHeight,
    })
    document.body.appendChild(app.canvas);

    let rectangles = [];

    app.ticker.add(function()
    {
        for (let i = 0; i < 1000; i++) {
            if (rectangles[i]) rectangles[i].destroy({ children: true, context: true });
            const rectangle = new PIXI.Graphics()
                .rect(
                    0,
                    0,
                    100,
                    100,
                )
                .fill({ color: 0xFF0000 });
            rectangles[i] = rectangle;
            app.stage.addChild(rectangle);
        }
    });
})();
```

### Environment

[...]

PR #10835, opened 2024-08-06
fix: RenderableGCSystem saving reference to destroyed object

Zyie:
The aim of this PR is to solve the issue where you are manually destroying lots of objects before the `RenderableGCSystem` runs e.g #10586

and also solves the issue of if you disable the system the `managedRenderables` array no longer holds a reference to the objects e.g #10834


src/rendering/renderers/shared/texture/RenderableGCSystem.ts
```diff
@@ -186,5 +191,17 @@ export class RenderableGCSystem implements System<RenderableGCSystemOptions>
...
+        this._managedRenderablesIndexes = {};
+    }
+
+    private _removeRenderable(renderable: Container): void
+    {
+        const index = this._managedRenderablesIndexes[renderable.uid];
+
+        if (index !== null)
+        {
+            this._managedRenderables.splice(index, 1);
```
lunarraid, line 203: If you splice here, wouldn't that invalidate the indexes of all the managed renderables ahead of it?

Zyie: you are correct, thats what i get for late night coding 😄

src/rendering/renderers/shared/texture/RenderableGCSystem.ts
```diff
@@ -187,4 +195,14 @@ export class RenderableGCSystem implements System<RenderableGCSystemOptions>
         this._renderer = null as any as Renderer;
         this._managedRenderables.length = 0;
     }
+
+    private _removeRenderable(renderable: Container): void
```
lunarraid, line 199: I might be missing something here, but since `addRenderable()` takes a `Renderable`, why is `_removeRenderable()` taking a `Container`?

Zyie: all events expect the thing passed to be a `Container`. We don't have any fancy types that know what the thing you listening too is

lunarraid:
This solves the issue of the memory leak of referencing Views once they are destroyed, but still adds them to the managed view list, which means it's creating an infinitely expanding Array when disabled, even if the reference itself eventually nulls out. Should we still be adding views if disabled?

src/rendering/renderers/shared/texture/RenderableGCSystem.ts
lunarraid: I suppose that makes sense with the assumption that anything passed into this extends `Container`, though I'm a bit nervous that the add only takes a Renderable. If there's never the possibility of a Renderable that doesn't extend Container then this is probably fine, though seems sketchy type-wise.

Zyie:
> This solves the issue of the memory leak of referencing Views once they are destroyed, but still adds them to the managed view list, which means it's creating an infinitely expanding Array when disabled, even if the reference itself eventually nulls out. Should we still be adding views if disabled?

Hopefully all sorted now!

src/rendering/renderers/shared/texture/RenderableGCSystem.ts
```diff
@@ -130,12 +131,14 @@ export class RenderableGCSystem implements System<RenderableGCSystemOptions>
...
     {
+        if (!this.enabled) return;
+
         renderable._lastUsed = this._now;
 
         if (renderable._lastInstructionTick === -1)
         {
-            // TODO manage index...
             this._managedRenderables.push(renderable);
+            renderable.once('destroyed', this._removeRenderable, this);
```
GoodBoyDigital, line 141: one thought on this (which i guess is a little subjective!) -  do we need to add the event listener? As this will always get cleaned up in the end either way?

lunarraid: For this issue this PR was trying to solve, yes. Memory was being eaten up faster than the custom GC was running. If the question is whether or not we should address the issue itself, I believe it would be confusing if calling `destroy()` on an object still kept a reference to the object around under the hood for 30 seconds. Conceptually, I would expect this to fully destroy the object and remove all references.

GoodBoyDigital (approved):

PR #10870, opened 2024-08-19
fix: Filter jittering using fractional resolutions

### LINKED ISSUE #10809: Bug: BlurFilter shaking when animating non-integer resolution
### Current Behavior

When setting resolution to a non-integer (1.5) the BlurFilter creates this awful shake when animating:
https://jsfiddle.net/bigtimebuddy/58vpjg9e/

Likely a v8 similar bug to #9885

https://github.com/user-attachments/assets/0b5ccea4-0ee5-4746-b1a2-cc250789db02

### Expected Behavior

BlurFilter should not jitter and shake animation

### Steps to Reproduce

* Set application resolution to 1.5
* Set BlurFilter resolution to 1.5 or 'inherit'
* Apply a BlurFilter
* Move the display object with the filter

### Environment

- **`pixi.js` version**: 8.2.6
- **OS & Version**: MacOS Latest
- **Running Example**: https://jsfiddle.net/bigtimebuddy/58vpjg9e/

### Possible Solution

_No response_

### Additional Information

_No response_

GoodBoyDigital:
##### Description of change
Much smoother now! the issue was that the padding was not rounded - which meant that for resolutions that where not whole numbers, a small precision was introduced that scaled with each render pass of a filter. Adjusted the order so that padding is added at the end and rounded.

fixes #10809

##### Pre-Merge Checklist

### DIFF
--- a/src/filters/FilterSystem.ts
+++ b/src/filters/FilterSystem.ts
@@ -251,7 +251,6 @@ export class FilterSystem implements System
         // need to factor in resolutions also..
         bounds.scale(resolution)
             .fitBounds(0, viewPort.width, 0, viewPort.height)
-            .ceil()
             .scale(1 / resolution)
             .pad(padding);
 
@@ -277,8 +276,8 @@ export class FilterSystem implements System
         // bind...
         // get a P02 texture from our pool...
         filterData.inputTexture = TexturePool.getOptimalTexture(
-            bounds.width,
-            bounds.height,
+            Math.ceil(bounds.width),
+            Math.ceil(bounds.height),
             resolution,
             antialias,
         );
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):
:tada: Verified!

**Broken** (v8.3.3): https://jsfiddle.net/bigtimebuddy/58vpjg9e/ 
**Fixed** (this PR): https://jsfiddle.net/bigtimebuddy/0b1kyoum/

cha0s:
I'm not on 8.x yet, but for anyone else still behind, the solution is encoded in the comment above:

> a small precision was introduced that scaled **with each render pass** of a filter.

I fixed this by setting the `quality` parameter to 1 (one render pass). For my use, this has no visible drawback and as a bonus must be faster than doing more passes anyway.

PR #10881, opened 2024-08-22
chore: optimise webGPU to not setVertexBuffer for each attribute if is inter…

GoodBoyDigital:
##### Description of change
This PR optimises the fact that we do not need to bind all slots for interleaved buffers. WebGPU is smart enough to figure out that itself. 

This change means that for most Pixi WebGPU draw calls, we now only call setVertexBuffer once instead of 4 times. 

##### Pre-Merge Checklist


src/rendering/renderers/gpu/GpuEncoderSystem.ts
```diff
@@ -145,9 +145,17 @@ export class GpuEncoderSystem implements System
 
     public setGeometry(geometry: Geometry, program: GpuProgram)
     {
-        for (const i in program.attributeData)
+        // when binding a buffers for geometry, there is no need to bind a buffer more than once if it is interleaved.
+        // which is often the case for Pixi. This is a performance optimisation.
+        // Instead of looping through hte attributes, we instead call getBufferNamesToBind
```
bigtimebuddy, line 150:
```suggestion
        // Instead of looping through the attributes, we instead call getBufferNamesToBind
```

src/rendering/renderers/gpu/pipeline/PipelineSystem.ts
```diff
@@ -312,6 +313,43 @@ export class PipelineSystem implements System
         return program._attributeLocationsKey;
     }
 
+    /**
+     * returns a hash of buffer names mapped to bind locations.
+     * This is used to bind the correct buffer to the correct location in the shader.
+     * @param geometry - The geometry to get the buffer names for
+     * @param program - The program to get the buffer names for
+     * @returns a object of buffer names mapped to the bind location
```
bigtimebuddy, line 321:
```suggestion
     * Returns a hash of buffer names mapped to bind locations.
     * This is used to bind the correct buffer to the correct location in the shader.
     * @param geometry - The geometry where to get the buffer names
     * @param program - The program where to get the buffer names
     * @returns An object of buffer names mapped to the bind location.
```

Zyie (approved):

PR #10885, opened 2024-08-23
feat: allow for custom batchers

GoodBoyDigital:
##### Description of change
Did a pass at making the batcher generic. This is to set it up so that creating custom batchers becomes much easier. This will ultimatly take a few steps. This is the first one!

Essentially it removes the packAttributes / packIndex from the batchable objects and moves it to a batcher instead. This means we can create new batchers and modify the packAttribute function without having to modify any of the other batchable classes! 

I also added a `packQuadAttributes` function to the batchers as this lets us batch the sprites even faster. Without this perf would be slower for spritesas we would need to copy sprite to a format packAttribute undertands and then copy that to the batch. `packQuadAttributes` does this in one step.

This is a draft for people to see where im at. Next steps will be:

- tie move geometry to custom batchers
- allow for new batchers to be registered
- work to remove the batchable objects and have them live on the sprite / mesh directly. This will make it much easier to customise batches as devs will only need to create a new batcher and then set the batch id on the sprite / mesh etc.
- auto generate the batching funcitions  so users will just need to create a shader and make sure the required props live on the renderable and it will work magically :D This will be really cool!

Oh and as a bonus - bunny mark is running a bit faster now too.. so we will get some more perf wins! woop!

##### Pre-Merge Checklist


src/rendering/batcher/shared/Batcher.ts
```diff
@@ -73,30 +77,54 @@ function returnBatchToPool(batch: Batch)
 {
     batchPool[batchPoolIndex++] = batch;
 }
-export interface BatchableObject
+
+export interface BatchableElement
```
Zyie, line 81: All these interfaces should now be documented as we are turning this into a more public api

src/rendering/batcher/shared/Batcher.ts
```diff
@@ -107,24 +135,23 @@ let BATCH_TICK = 0;
  */
 export interface BatcherOptions
 {
-    /** The size of the vertex buffer. */
-    vertexSize?: number;
-    /** The size of the index buffer. */
-    indexSize?: number;
     /** The maximum number of textures per batch. */
     maxTextures?: number;
+    vertexInitialSize?: number;
```
Zyie, line 140: do we need to change the name here? Seems like a avoidable breaking change

src/rendering/batcher/shared/Batcher.ts
```diff
@@ -142,27 +169,72 @@ export class Batcher
...
     private _batchIndexStart: number;
     private _batchIndexSize: number;
 
     /** The maximum number of textures per batch. */
     public readonly maxTextures: number;
 
+    /** The name of the batcher. Must be implemented by subclasses. */
+    public abstract name: string;
+    /** The vertex size of the batcher. Must be implemented by subclasses. */
+    public abstract vertexSize: number;
```
Zyie, line 183: seems like this doesn't need to be public and should be `protected`

src/rendering/batcher/shared/Batcher.ts
```diff
@@ -142,27 +169,72 @@ export class Batcher
...
     constructor(options: BatcherOptions = {})
     {
         Batcher.defaultOptions.maxTextures = Batcher.defaultOptions.maxTextures ?? getMaxTexturesPerBatch();
         options = { ...Batcher.defaultOptions, ...options };
 
-        const { vertexSize, indexSize, maxTextures } = options;
+        const { maxTextures, vertexInitialSize, indexInitialSize } = options;
 
-        this.attributeBuffer = new ViewableBuffer(vertexSize * this._vertexSize * 4);
+        this.attributeBuffer = new ViewableBuffer(vertexInitialSize * 6 * 4);
```
Zyie, line 235: why is this no longer using `vertexSize`?
e.g.
`this.attributeBuffer = new ViewableBuffer(vertexInitialSize * this.vertexSize * 4);`

src/rendering/batcher/shared/DefaultBatcher.ts
```diff
@@ -0,0 +1,193 @@
...
+import { Shader } from '../../renderers/shared/shader/Shader';
+import { getMaxTexturesPerBatch } from '../gl/utils/maxRecommendedTextures';
+import { Batcher } from './Batcher';
+import { BatchGeometry } from './BatchGeometry';
+
+import type { BatchableMeshElement, BatchableQuadElement } from './Batcher';
+
+let defaultShader: Shader = null;
+
+/** The default batcher is used to batch quads and meshes. */
```
Zyie, line 16: if we are ignoring all the other batcher classes we should add that here as well

src/rendering/batcher/shared/DefaultBatcher.ts
```diff
@@ -0,0 +1,193 @@
...
+                resources: {
+                    batchSamplers: getBatchSamplersUniformGroup(maxTextures),
+                }
+            });
+        }
+
+        this.shader = defaultShader;
+    }
+
+    public packAttributes(
```
Zyie, line 75: It would be good to get these documented

src/scene/graphics/shared/BatchableGraphics.ts
```diff
@@ -1,140 +1,93 @@
...
 import type { Texture } from '../../../rendering/renderers/shared/texture/Texture';
 import type { Graphics } from './Graphics';
 
+const identityMatrix = new Matrix();
 /**
  * A batchable graphics object.
  * @ignore
  */
-export class BatchableGraphics implements BatchableObject
+
```
Zyie, line 13:
```suggestion
```

tests/renderering/graphics/Graphics.test.ts
```diff
@@ -242,8 +242,8 @@ describe('Graphics', () =>
 
             const graphicsData = renderer.graphicsContext['_graphicsDataContextHash'][graphics.context.uid];
 
-            expect(graphicsData.geometry.indexBuffer.data.length).toEqual(3 * 6);
-            expect(graphicsData.geometry.buffers[0].data.length).toEqual(3 * 4 * 6);
+            expect(graphicsData.batcher.geometry.indexBuffer.data.length).toEqual(3 * 6);
```
Zyie, line 245: should we add a deprecated getter for `graphicsData.geometry` in case anyone was using it?

src/rendering/batcher/shared/Batcher.ts
GoodBoyDigital: removed this.vertexSize from the multiplicaiton as this is abstract and causing errors. Will leave it for the user to factor in when parsing in `vertexInitialSize`

src/rendering/batcher/shared/Batcher.ts
GoodBoyDigital: moving terminology from vertex to attribute for consistency, i think this is ok. (no more breaking than the other changes iyswi?)

src/rendering/batcher/shared/Batcher.ts
GoodBoyDigital: agree!

src/rendering/batcher/shared/DefaultBatcher.ts
GoodBoyDigital: please clarify - not sure i understand the comment :D

src/rendering/batcher/shared/DefaultBatcher.ts
GoodBoyDigital: agree!

src/scene/graphics/shared/GraphicsContextSystem.ts
```diff
@@ -44,13 +45,27 @@ export class GpuGraphicsContext
...
+    /**
+     * @deprecated since version 8.0.0
+     * Use `batcher.geometry` instead.
+     * @see {Batcher#geometry}
+     */
+    get geometry()
+    {
+        // eslint-disable-next-line max-len
+        deprecation(v8_3_4, 'GraphicsContextRenderData#geometry is deprecated, please use batcher.geometry instead.');
+        // #endif
```
Zyie, line 65:
```suggestion
```

Zyie (approved):

PR #10888, opened 2024-08-25
fix: Color value setter normalizes cloned value

### LINKED ISSUE #10882: Bug: Array color used to tint a Sprite it changes the original color array values
### Current Behavior

Setting the tint property of Sprite if the color is an array, because it is `._normalize` method is applied directly to the passed color but not the cloned one, it causes the original color to be modified causing a side effect on the application using Pixi.js

### Expected Behavior

I expect that the color passed is not modified. After looking at the Color.ts  file I guess that the clone intends to avoid that color is modified from  outside Pixi.js once it has been set. Aligned to that intention normalize should be applied to the cloned color.

### Steps to Reproduce

I posted a pixiplayground link so it is clear how to reproduce it.

https://pixiplayground.com/#/edit/xX09VNY0_G1XCeHTd5iRr

### Environment

- **`pixi.js` version**: *e.g. 8.1.5*
- **Browser & Version**: *e.g. Chrome 116*
- **OS & Version**: *e.g. MacOS 14.6.1*
- **Running Example**: *[https://pixiplayground.com/#/edit/xX09VNY0_G1XCeHTd5iRr](https://pixiplayground.com/#/edit/xX09VNY0_G1XCeHTd5iRr)*

### Possible Solution

On value setter of Color class just call the normalize method on the already cloned color.

```
         else if (this._value === null || !this._isSourceEqual(this._value, value))
         {
-            this._normalize(value);
             this._value = this._cloneSource(value);
+            this._normalize(this._value);
         }
```

### Additional Information

_No response_

rperello:
##### Description of change
On `Color.ts` value setter normalize this._value once cloned so the original value is not modified.

Closes #10882 

##### Pre-Merge Checklist

### DIFF
--- a/src/color/Color.ts
+++ b/src/color/Color.ts
@@ -229,8 +229,8 @@ export class Color
         }
         else if (this._value === null || !this._isSourceEqual(this._value, value))
         {
-            this._normalize(value);
             this._value = this._cloneSource(value);
+            this._normalize(this._value);
         }
     }
     get value(): Exclude<ColorSource, Color> | null
--- a/tests/color/Color.test.ts
+++ b/tests/color/Color.test.ts
@@ -28,7 +28,7 @@ describe('Color', () =>
         });
     });
 
-    it.concurrent('should not throw error for invalid color values', async () =>
+    it.concurrent('should not throw error for invalid color values and not alter the original one', async () =>
     {
         const invalidColorValues: any[] = [
             [1, 1, -1, 1],
@@ -36,9 +36,13 @@ describe('Color', () =>
             { r: 1, g: 1, b: 1, a: 1.1 },
         ];
 
-        invalidColorValues.forEach((value) =>
+        invalidColorValues.forEach((originalValue) =>
         {
-            expect(() => new Color(value as any)).not.toThrow();
+            const value = structuredClone(originalValue);
+
+            expect(() => new Color(value)).not.toThrow();
+
+            expect(value).toEqual(originalValue);
         });
     });
 


bigtimebuddy (commented):
This is great and appreciate the test. One suggestion is if you could move the test to the other Color tests and instead of a Sprite tint, the test should be for the setter for `value`. That ensures it handled accurately everywhere that color is used.

rperello:
> This is great and appreciate the test. One suggestion is if you could move the test to the other Color tests and instead of a Sprite tint, the test should be for the setter for `value`. That ensures it handled accurately everywhere that color is used.

Of course. I appreciate the comment.

rperello:
While working on adding tests for this fix in `Color.test.ts`, as you correctly suggested, my first approach is to create a failing test for the previous code (the one before this fix). In order to happen, the color passed should be altered by normalize, and none of the colors defined in the `it.concurrent('should convert color values to rgba')` was useful for the objective. All those are colors that do not really need a normalize process (nothing would change really the normalize process is not executed). They are totally valid as exposed by the [ColorSource documentation](https://pixijs.download/release/docs/color.html#ColorSource).

Then I realised that  `it.concurrent('should not throw error for invalid color values')` is testing that passing "invalid" (but normalizable) colors doesn't throw an error. My guess is those invalid colors are accepted only if the normalize process converts them into valid colors without throwing an error.

The bug I mentioned was indeed caused by an invalid but "normalizable" color ([255, 0, 0]) that was transformed into [1, 0, 0] by the normalize process, but as normalize is currently done before cloning caused the original passed color to be altered.

My first attempt was to add invalid colors in the mentioned 'should convert color values to rgba' block, but this would also required to add the color (second element in the list) that is expected to be converted to. And I do think this goes beyond the scope as this somehow would be testing what the normalize outputs for invalid colors. This would be another task: Testing normalized outputs from invalid but "normalizable" colors.

So my final decision has been to add an extra expect in 'should not throw an error for invalid color values' to check that the original passed color is not altered.

I think it's important to provide context to my decision.

Let me know what you think about this approach.

bigtimebuddy (approved):
Thanks for the explanation. This works for me 👍

PR #10909, opened 2024-09-05
chore: fix up batch property names

GoodBoyDigital:
extracted default shader

##### Description of change
- Whilst im here in batcher land, figured it would be good to do a pass on some of the batcher property names. anything with a `_` prefix is a prop managed by the batcher, whilst anything else should be user defined.
- I also put some stronger types in for the `BatchableQuadElement` and `BatchableMeshElement`. 
- extracted `DefaultShader` from `DefaultBatcher`, making things a bit more readable!
- created `DefaultBatchableQuadElement` and `DefaultBatchableMeshElement` so they contain the props that onnly they care about. Making`DefaultBatchableElement` and `DefaultBatchableElement` more generic.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/batcher/shared/DefaultBatcher.ts
+++ b/src/rendering/batcher/shared/DefaultBatcher.ts
@@ -17,7 +17,7 @@ interface DefaultBatchElements
 }
 
 export interface DefaultBatchableQuadElement extends BatchableQuadElement, DefaultBatchElements {}
-export interface DefaultBatchableElement extends BatchableMeshElement, DefaultBatchElements {}
+export interface DefaultBatchableMeshElement extends BatchableMeshElement, DefaultBatchElements {}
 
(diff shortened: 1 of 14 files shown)


src/rendering/batcher/shared/DefaultBatcher.ts
```diff
@@ -1,18 +1,29 @@
...
 import { BatchGeometry } from './BatchGeometry';
+import { DefaultShader } from './DefaultShader';
 
+import type { Matrix } from '../../../maths/matrix/Matrix';
+import type { Shader } from '../../renderers/shared/shader/Shader';
 import type { BatchableMeshElement, BatchableQuadElement } from './Batcher';
 
 let defaultShader: Shader = null;
 
+interface DefaultBatchElements
```
Zyie, line 12:
```suggestion
export interface DefaultBatchElements
```

PR #10918, opened 2024-09-11
feat: add version to global hooks

Zyie:
a simple PR to pass the version of pixi to the global hooks. Makes it easier to make changes to the devtool based on the version of pixi


bigtimebuddy (approved):

GoodBoyDigital (approved):
looks good!

ONe tweak - lets move VERSION to its own const and have `sayHello` import it.

Issue #8207, opened 2022-03-03
dose pixijs support reverse mask feature ?


liujinxiao:
is there any plan to achieve **mask reverse feature** ? thanks !

now our team is using pixi.js to achieve mask feature and mask reverse feature, like this

1、this is the original sprite
![image](https://user-images.githubusercontent.com/26742095/156516226-839cb912-7355-4cd6-bb25-a88c88a53aaf.png)

2、this is the sprite with circle mask 
![image](https://user-images.githubusercontent.com/26742095/156516388-edb21475-94d5-46e2-9e98-b4370f148295.png)

3、this is the sprite with circle reverse mask
![image](https://user-images.githubusercontent.com/26742095/156516516-3db0d608-33d6-409c-9693-01cf7ccc4937.png)

liujinxiao:
https://github.com/liujinxiao/pixijs/pull/1/files

edit packages/core/src/filters/spriteMask/spriteMaskFilter.frag file,  maybe resolve the issue?
![image](https://user-images.githubusercontent.com/26742095/156958733-4d0bdffd-9b62-44ce-ace5-cefe565f910e.png)

liujinxiao:
>

liujinxiao:
> You can use custom sprite mask filters like so:
> 
> ```js
> object.mask = new PIXI.MaskData(sprite);
> object.mask.filter = new MyCustomSpriteMaskFilter();
> ```
> 
> But you shouldn't use a mask for reverse masking. That doesn't work. Use a filter:
> 
> ```js
> object.filters = [new MyReverseSpriteMaskFilter()];
> object.filters[0].maskSprite = sprite;
> ```

mask is actually a filter, so i want to edit mask filter to achieve mask reverse feature。this does work.

dev7355608:
Try this:
```js
object.filters = [new PIXI.SpriteMaskFilter(undefined, `\
varying vec2 vMaskCoord;
varying vec2 vTextureCoord;

uniform sampler2D uSampler;
uniform sampler2D mask;
uniform float alpha;
uniform float npmAlpha;
uniform vec4 maskClamp;

void main(void)
{
    float clip = step(3.5,
        step(maskClamp.x, vMaskCoord.x) +
        step(maskClamp.y, vMaskCoord.y) +
        step(vMaskCoord.x, maskClamp.z) +
        step(vMaskCoord.y, maskClamp.w));

    vec4 original = texture2D(uSampler, vTextureCoord);
[...]

PR #10922, opened 2024-09-13
feat: Add ability to inverse mask

katmai7:
##### Description of change
Add ability to inverse mask.

Done:
- [x] Inverse for stencil mask
- [x] Inverse for alpha mask

##### Pre-Merge Checklist

Related issues:
https://github.com/pixijs/pixijs/issues/8207


src/filters/mask/mask.wgsl
```diff
@@ -79,20 +77,22 @@ fn mainFragment(
 
     var maskClamp = filterUniforms.uMaskClamp;
 
-     var clip = step(3.5,
-        step(maskClamp.x, filterUv.x) +
-        step(maskClamp.y, filterUv.y) +
-        step(filterUv.x, maskClamp.z) +
-        step(filterUv.y, maskClamp.w));
+    var clip = step(3.5,
```
enriko-riba, line 80: clip is unused?

katmai7: Oops, I mistakenly removed it from shader.. will return it back

katmai7: Fixed

bigtimebuddy:
Absolutely love this feature. Super useful and is a think that folks often ask for.

Had a chat about this with the team, in terms of the Container API, we'd like to avoid `maskInverse`. We had this similar problem in v7 with `cacheAsBitmap` having to add a bunch of properties like `cacheAsBitmapResolution`, `cacheAsBitmapMultisample`, etc. It's kinda clunky.

Our preference would be to add a `setMask(options)` method---`options` would have `mask` and `inverse` properties. In the future, I could see us adding more pass-thru customizations like `resolution` (for mask filters). Also, we should keep the `mask` setter in Container for backward compatibility and convenience.

bigtimebuddy:
Yeah, setMask as a new method on container.

GoodBoyDigital (approved):
thanks so much for this @katmai7 !!! we have wanted this feature for ages :D

PR #10945, opened 2024-09-27
fix: Reset defaultPrevented in Events

### LINKED ISSUE #9749: Bug: event.defaultPrevented behaves incorrectly on consecutive events
### Current Behavior

When `app.renderer.events.autoPreventDefault = false;` is set and an event listener on some element calls `event.preventDefault()`, all subsequent events always have `event.defaultPrevented` set to `true`.

### Expected Behavior

`event.defaultPrevented` should initially be `false` for every event

### Steps to Reproduce

```ts
const app = new Application({
  width: 900,
  height: 600,
  background: "#00ff00"
});

app.renderer.events.eventMode = "static";

const text = new Text("TEST", { fontSize: 60 });
text.eventMode = "static";
text.onmousedown = (event) => {
  console.log("text listener", event.defaultPrevented);
  if (!event.defaultPrevented) {
    event.preventDefault();
  }
}
```

### Environment

- **`pixi.js` version**: *7.3.1*
- **Browser & Version**: *FF118.0.1*
- **OS & Version**: *Win 10*
- **Running Example**: *https://codesandbox.io/s/pixi-js-forked-q6ng6s?file=/src/index.js*

### Possible Solution

investigate and fix event dispatch logic

### Additional Information

_No response_

STRd6:
Fix #9749

##### Description of change

Reset `defaultPrevented` when allocating events. This makes subsequent events behave correctly after `preventDefault()` has been called.

##### Pre-Merge Checklist

### DIFF
--- a/src/events/EventBoundary.ts
+++ b/src/events/EventBoundary.ts
@@ -1420,6 +1420,7 @@ export class EventBoundary
 
         event.eventPhase = event.NONE;
         event.currentTarget = null;
+        event.defaultPrevented = false;
         event.path = null;
         event.target = null;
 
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Good catch

PR #10949, opened 2024-09-30
feat: add ParticleContainer

GoodBoyDigital:
##### Description of change
Finally got round to implementing a particle container for v8! This one differews slightly in that it does not render `Sprites`, but `Particles`. A `Particle` is just a very basic object that has - tint / alpha / texture / position / anchor / scale - so almost exactly likea sprite!!

The upside of using particles is that this thing can render sooooo much faster. On my lappy - 
- regular bunnies (sprites + container) **200,000 @ 64fps**
- particle bunnies (particle + particle container) **1,100,000 @ 64fps** - The bottle neck for particles at that point is the bunny update logic, not the actual rendering!

The ParticleContainer does retain a similar static / dynamic api when for its properties. The more static attributes you have, the faster it runs. By default only position is dynamic - but this can be adjusted based on your needs. 

##### Pre-Merge Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a `ParticleContainer` class for efficient particle rendering.
	- Added support for both OpenGL and GPU-based rendering with new shader classes and adapters.
	- Implemented new particle update functions for dynamic and static properties.
	- Enhanced particle management with `ParticleBuffer` and `ParticleContainerPipe`.
	- Added new interfaces and classes for particle attributes, including `IParticle` and `Particle`.
	- Introduced a utility function for generating index buffers for quads.

- **Bug Fixes**
	- Improved cleanup methods to ensure proper resource management.

- **Tests**
	- Added a visual test scene for rendering particles with different textures.


src/scene/particle-container/init.ts
```diff
@@ -0,0 +1,7 @@
+import { extensions } from '../../extensions/Extensions';
+import { GlParticleContainerPipe } from './shared/GlParticleContainerPipe';
+import { GpuParticleContainerPipe } from './shared/GpuParticleContainerPipe';
+
+// TODO @Zyie, will this tree shake out for the correct renderer?
+extensions.add(GlParticleContainerPipe);
+extensions.add(GpuParticleContainerPipe);
```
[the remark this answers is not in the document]
GoodBoyDigital: cc @Zyie!

tests/visual/scenes/particle/particle.scene.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Particle } from '../../../../src/scene/particle-container/shared/Particle';
+import { ParticleContainer } from '../../../../src/scene/particle-container/shared/ParticleContainer';
+
+import type { Container } from '../../../../src/scene/container/Container';
+import type { TestScene } from '../../types';
+
+export const scene: TestScene = {
+    it: 'should render particle container',
+    pixelMatch: 200,
+    only: true,
```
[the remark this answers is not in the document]
GoodBoyDigital: good catch!

src/scene/particle-container/shared/GlParticleContainerPipe.ts
```diff
@@ -0,0 +1,21 @@
...
+        type: [
+            ExtensionType.WebGLPipes,
+        ],
+        name: 'particle',
+    } as const;
+
+    constructor(renderer: WebGPURenderer)
+    {
+        super(renderer, new GlParticleContainerAdaptor());
+    }
```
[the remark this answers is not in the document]
GoodBoyDigital: good catch!

src/rendering/renderers/shared/geometry/Geometry.ts
```diff
@@ -61,7 +59,7 @@
     /** an optional label to easily identify the geometry */
     label?: string;
     /** the attributes that make up the geometry */
-    attributes: AttributeOptions;
+    attributes?: AttributeOptions;
```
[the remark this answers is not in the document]
GoodBoyDigital: good catch!

src/scene/particle-container/shared/ParticleContainer.ts
```diff
@@ -0,0 +1,355 @@
...
+    public static defaultOptions: ParticleContainerOptions = {
+        dynamicProperties: {
+            vertex: false, // Indicates if vertex positions are dynamic.
+            position: true, // Indicates if particle positions are dynamic.
+            rotation: false, // Indicates if particle rotations are dynamic.
+            uvs: false, // Indicates if UV coordinates are dynamic.
+            color: false, // Indicates if particle colors are dynamic.
+        },
+        roundPixels: false // Indicates if pixels should be rounded for rendering.
+    };
```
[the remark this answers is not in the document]
GoodBoyDigital: nice catch, vertex is the correct name - instead i updated the type

src/scene/particle-container/shared/Particle.ts
```diff
@@ -0,0 +1,120 @@
...
+    public rotation = 0;
+    /** The color of the particle as a hexadecimal number. */
+    public color = 0xffffffff;
+    /** The texture of the particle. */
+    public texture: Texture;
+
+    private _alpha = 1;
+    private _tint = 0xffffff;
+
+    constructor(texture: Texture)
```
bigtimebuddy, line 75: I'd love to see similar options to Sprite where you can set the texture but also set any attributes like alpha scale etc.

bigtimebuddy:
This is amazing @GoodBoyDigital 🎉

src/scene/particle-container/shared/ParticleBuffer.ts
```diff
@@ -0,0 +1,227 @@
...
+
+            this._size = Math.max(particles.length, this._size * 1.5);
+
+            this.staticAttributeBuffer = new ViewableBuffer(this._size * this._staticStride * 4 * 4);
+            this.dynamicAttributeBuffer = new ViewableBuffer(this._size * this._dynamicStride * 4 * 4);
+            this.indexBuffer = createIndicesForQuads(this._size);
+
+            this.geometry.indexBuffer.setDataWithSize(
+                this.indexBuffer, this.indexBuffer.byteLength, true);
+        }
```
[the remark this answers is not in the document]
GoodBoyDigital: they are updated further along in the function

src/scene/particle-container/shared/Particle.ts
GoodBoyDigital: good shout! let me add an options object

GoodBoyDigital: added!

src/scene/particle-container/shared/Particle.ts
```diff
@@ -0,0 +1,144 @@
...
+    anchorX: number;
+    anchorY: number;
+    rotation: number;
+    color: number;
+    texture: Texture;
+}
+
+export type ParticleOptions = Omit<Partial<IParticle>, 'color'> & {
+    texture: Texture
+    tint?: number;
```
bigtimebuddy, line 33: This could be more flexible:

```suggestion
    tint?: ColorSource;
```

src/scene/particle-container/shared/Particle.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Here is an example of how to create a new particle:
+ *
+ * ```javascript
+ * const particle = new Particle(texture);
+ * particle.x = 100;
+ * particle.y = 100;
+ * particle.scaleX = 0.5;
+ * particle.scaleY = 0.5;
+ * particle.rotation = Math.PI / 2;
+ * particle.color = 0xff0000;
```
bigtimebuddy, line 53: Preference here for options:
```suggestion
 * const particle = new Particle({
 *   texture,
 *   x: 100,
 *   y: 100,
 *   scaleX: 0.5,
 *   scaleY: 0.5,
 *   rotation: Math.PI / 2,
 *   color: 0xff0000,
 * });
```

src/scene/particle-container/shared/Particle.ts
```diff
@@ -0,0 +1,144 @@
...
+            this.x = options.x ?? 0;
+            this.y = options.y ?? 0;
+            this.scaleX = options.scaleX ?? 1;
+            this.scaleY = options.scaleY ?? 1;
+            this.rotation = options.rotation ?? 0;
+            this._alpha = options.alpha ?? 1;
+            this._tint = options.tint ?? 0xffffff;
+
+            this._updateColor();
+        }
```
bigtimebuddy, line 101: This is a more familiar pattern to use static defaultOptions so you don't have to declare the defaults twice (declaring members and setting defaults in constructor).

```suggestion
        if (options instanceof Texture)
        {
            options = { texture: options };
        }
        
        Object.assign(this, Particle.defaultOptions, options);

        this._updateColor();
```

src/scene/particle-container/shared/Particle.ts
```diff
@@ -0,0 +1,144 @@
...
+    public scaleY = 1;
+    /** The rotation of the particle in radians. */
+    public rotation = 0;
+    /** The color of the particle as a hexadecimal number. */
+    public color = 0xffffffff;
+    /** The texture of the particle. */
+    public texture: Texture;
+
+    private _alpha = 1;
+    private _tint = 0xffffff;
```
bigtimebuddy, line 79:
```suggestion
    /** Default options for constructing with options */
    static defaultOptions: {
        anchorX: 0,
        anchorY: 0,
        x: 0,
        y: 0,
        scaleX: 1,
        scaleY: 1,
        rotation: 0,
        tint: 0xffffff,
        alpha: 1,
    };
    
    /** The x-coordinate of the anchor point. */
    public anchorX: number;
    /** The y-coordinate of the anchor point. */
    public anchorY: number;
    /** The x-coordinate of the particle. */
    public x: number;
    /** The y-coordinate of the particle. */
    public y: number;
    /** The scale factor in the x-axis. */
    public scaleX: number;
    /** The scale factor in the y-axis. */
    public scaleY: number;
    /** The rotation of the particle in radians. */
    public rotation: number;
    /** The color of the particle as a hexadecimal number. */
    public color: number;
    /** The texture of the particle. */
    public texture: Texture;

    private _alpha: number;
    private _tint: number;
```

src/scene/particle-container/shared/Particle.ts
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        this._updateColor();
+    }
+
+    private _updateColor()
+    {
+        // combine alpha and tint
+        this.color = this._tint + (((this._alpha * 255) | 0) << 24);
+    }
```
bigtimebuddy, line 143: We might consider adding convenience setters for `scale` and `anchor` to more closely mimic Container-like objects.

src/scene/particle-container/shared/Particle.ts
```diff
@@ -0,0 +1,167 @@
...
+        anchorX: 0,
+        anchorY: 0,
+        x: 0,
+        y: 0,
+        scaleX: 1,
+        scaleY: 1,
+        rotation: 0,
+        tint: 0xffffff,
+        alpha: 1,
+    };
```
Zyie, line 89: @GoodBoyDigital i tested bunny mark with using particles and get ~52 fps with 1 million particles which is identical to not using a default options object

src/scene/particle-container/shared/ParticleContainer.ts
```diff
@@ -0,0 +1,571 @@
...
+     * Calling this method will throw an error. Please use `ParticleContainer.addParticle()` instead.
+     * @param {...any} _children
+     * @throws {Error} Always throws an error as this method is not available.
+     */
+    public override addChild<U extends ContainerChild[]>(..._children: U): U[0]
+    {
+        throw new Error(
+            'ParticleContainer.addChild() is not available. Please use ParticleContainer.addParticle()',
+        );
+    }
```
Zyie, line 439: @bigtimebuddy @GoodBoyDigital 

For `ParticleContainer` I have found that you can't accurately convert the types for `addChild`/`removeChild` etc to return `IParticle` without losing type accuracy for normal containers.

So the current solution is to change the names of these helper function to `addParticle`/ `removeParticle` etc
and then to throw errors if a user tries to use one of the "normal" functions

It's not quite as clean as I would have liked but it does keep type accuracy, any thoughts on this approach?

also as an additional question, should we support the less common helpers such as `swapChildren`/`getChildIndex` etc, or just keep to `addParticle`/`removeParticle`/`addParticleAt`/`removeParticleAt`/`removeParticles`

bigtimebuddy: I think this is a nice solution to not only the type issue but also what might be developer confusion between Container and ParticleContainer. Having a harder API break is helpful, I think.


Might be nice to also have some development errors thrown if the child is not an instanceof Particle. I can definitely see v7 users trying to add Sprites here.

bigtimebuddy: Be conservative about all the container methods. We can always easily add later if folks use this heavily.

src/scene/particle-container/shared/Particle.ts
Zyie: @GoodBoyDigital i have tested this now

https://github.com/pixijs/pixijs/pull/10949#discussion_r1782669137

I basically get the same FPS regardless of which one was used

bigtimebuddy: Would this be more of an issue with performance allocating not runtime fps? Since this is only constructor related. Ultimately, the shape of the object is the same, so I don't think runtime fps is going to be an issue.

src/scene/particle-container/shared/Particle.ts
GoodBoyDigital: woop! great new :D

Zyie (approved):

Issue #10668, opened 2024-06-25
Bug: unable to change resolution of BitmapText


goodbot1356:
### Current Behavior

Changing .resolution of BitmapText does nothing. Chaning .resolution of Text works just great.

### Expected Behavior

Changin .resolution of BitmapText should change its resolution.

### Steps to Reproduce

```


/**
* This is the default playground.
* You should see a bunny spinning in the right preview pane.
* Feel free to use this as a starting point for you own playground!
*/

// Create our application instance
(async () => {
    const app = new PIXI.Application();
    await app.init({
        width: window.innerWidth,
        height: window.innerHeight,
        backgroundColor: 0x2c3e50
    })
    document.body.appendChild(app.canvas);



    const bitmapText = new PIXI.BitmapText({
        text: "TEST TEXT",
        style: {
            fontSize: 150,
            fontFamily: "monospace",
            letterSpacing: 5
        }
    });
    app.stage.addChild(bitmapText);




    const defaultText = new PIXI.Text({
        text: "TEST TEXT",
        style: {
            fontSize: 150,
            fill: "white",
            letterSpacing: 5
        }
    });
    defaultText.position.y = 200;
    app.stage.addChild(defaultText);





    setTimeout(() => {
        bitmapText.resolution = 0.1;
        defaultText.resolution = 0.1;
    }, 2000)
    


    // Listen for animate update
    app.ticker.add(function(ticker)
    {
        // Rotate mr rabbit clockwise
    });
})();


```

### Environment

- **`pixi.js` version**: *8.2.1*
- **Browser & Version**: *Chrome 125*
- **Running Example**: *<https://www.pixiplayground.com/#/edit/CI8pmcR84jj-ZW3UxAkCN/>*

GoodBoyDigital:
heya the best thing todo for now is to use `BitmapFont.install('myFont', {...style, resolution:1})` this way you can control the fonts and resolution more precisely.

bigtimebuddy:
Talked about this with @GoodBoyDigital and @Zyie, I think the solution here is to treat resolution as _readonly_ for BitmapText and override the resolution setter in AbstractText to throw an error. Resolution should be managed by the asset (.fnt file or BitmapFont) for BitmapText. The display object itself has no concept of resolution.

PR #10956, opened 2024-10-02
fix: warn when trying to change BitmapText resolution

Zyie:
fixes: #10668

just adding a warning that you cant change the resolution of a bitmap text

### DIFF
--- a/src/scene/text-bitmap/BitmapText.ts
+++ b/src/scene/text-bitmap/BitmapText.ts
@@ -1,3 +1,4 @@
+import { warn } from '../../utils/logging/warn';
 import { AbstractText, ensureOptions } from '../text/AbstractText';
 import { TextStyle } from '../text/TextStyle';
 import { BitmapFontManager } from './BitmapFontManager';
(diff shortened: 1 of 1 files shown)


src/scene/text-bitmap/BitmapText.ts
```diff
@@ -144,4 +145,21 @@ export class BitmapText extends AbstractText<TextStyle, TextStyleOptions> implem
...
         bounds.maxY = bounds.minY + height;
     }
+
+    /**
+     * The resolution / device pixel ratio of the canvas.
+     * @default 1
+     */
+    override set resolution(_value: number)
+    {
+        warn(
```
bigtimebuddy, line 155: I forget, are we excluding stuff from prod builds? If so, I would only include this in development.

bigtimebuddy (approved):

src/scene/text-bitmap/BitmapText.ts
Zyie: nice catch! its now removed for prod

Issue #10915, opened 2024-09-10
Bug: Text does not update its value when 2 text fields are using shared style


vlasenkoigor:
### Current Behavior

When 2  textfields that uses shared font style and if we trying for the first one to set same text values that 2nd text field currently has, then 1st text field does not render that value. 

### Expected Behavior

Text fields should be updated

### Steps to Reproduce

https://www.pixiplayground.com/#/edit/M6SOE_CJuV6Gjm2su12GR

In this playground 2 selectors have the same range and and same font style, and if one of them has been switched to value that another already has, it's value stays outdated

### Environment

- **`pixi.js` version**: *8x*
- **Browser & Version**: *Chrome *
- **OS & Version**: *Windows*
- **Running Example**: https://www.pixiplayground.com/#/edit/M6SOE_CJuV6Gjm2su12GR

GoodBoyDigital:
thanks for posting and providing this example! going to do a bit of a text pass next!

Issue #10889, opened 2024-08-26
Bug: pixi text fill color not updating promptly


NetheriteJam:
### Current Behavior

When there are 2 pixi Text **with the same text property**, changing text.style.fill of 1 of them does not promptly update on the canvas.
However, when there is a 3rd pixi Text **with the same text property and does the same updates** as the 1st Text, their color updates promptly on the canvas.

### Expected Behavior

changing text.style.fill promply update on the canvas

### Steps to Reproduce

```javascript
import * as pixi from 'pixi.js';

(async () =>
{
    // Create a new application
    const app = new pixi.Application();

	await app.init({
		background: '#91ffde',
	});

	document.body.appendChild(app.canvas);

	const text1 = new pixi.Text({
		text: 'text',
		style: {
			fill: '#000000',
		},
		x: 100,
		y: 100,
	});

	const text2 = new pixi.Text({
		text: 'text',
		style: {
			fill: '#000000',
		},
		x: 100,
		y: 200,
	});

	const text3 = new pixi.Text({
		text: 'TEXT',
		style: {
			fill: '#000000',
		},
		x: 100,
		y: 300,
	});

	window.addEventListener('mousedown', onclick);
	function onclick() {
		if ( text1.style.fill != '#ffffff' ) {
			text1.style.fill = '#ffffff';
			text3.style.fill = '#ffffff';
		} else {
			text1.style.fill = '#000000';
			text3.style.fill = '#000000';
		}
		console.log(text1.style.fill);
	}

	app.stage.addChild(text1);
	app.stage.addChild(text2);
	app.stage.addChild(text3);
})();
```

### Environment

- **`pixi.js` version**: *8.3.3*
- **Browser & Version**: *Chrome Version 128.0.6613.84 (Official Build) (64-bit)*
- **OS & Version**: *Windows 10*
[...]

PR #10959, opened 2024-10-02
fix: text not updating with shared style

Zyie:
fixes: #10915 & #10889

when the text key changes we should be invalidating the text as we now need to update the text to the correct texture


GoodBoyDigital (approved):
like this. less complicated :)

PR #10968, opened 2024-10-04
feat: clean hashes and arrays

Zyie:
Replacement for #10793

In PixiJS we make use of hashes to map data to things like renderables within pipes. To keep things efficinat we do not use 'delete' when an item needs to be removed, instead we simply set the property to null. Over time (with lots of destroy / create) the hashes could grow to be pretty large.

To prevent hashes from growing to large this PR introduces two new methods to the `RenderableGCSystem` that allow it to schedule clean up operations for hashes and arrays.


GoodBoyDigital (approved):
nice!

PR #10973, opened 2024-10-07
fix: Rebuild instruction when "setMask" is called

katmai7:
##### Description of change
If mask `inverse` option changed dynamically, it has no effect.
So we need to mark render group structure as changed in order to trigger instructions rebuild.

##### Pre-Merge Checklist


src/scene/container/container-mixins/effectsMixin.ts
```diff
@@ -92,10 +102,7 @@ export const effectsMixin: Partial<Container> = {
 
         this.effects.splice(index, 1);
 
-        if (this.parentRenderGroup)
-        {
-            this.parentRenderGroup.structureDidChange = true;
```
katmai7, line 97: I assume that when effect is removed it should has same logic as `addEffect` for getting rendering group.
Please correct me if I'm wrong.

katmai7:
Hi @GoodBoyDigital,
It's a fix for inverse mask, could yo please review it?

src/scene/container/container-mixins/effectsMixin.ts
```diff
@@ -48,6 +54,15 @@ export const effectsMixin: Partial<Container> = {
      */
     effects: [],
 
+    _markStructureAsChanged()
```
bigtimebuddy, line 57: Could you add private to this?

```suggestion
    private _markStructureAsChanged()
```

src/scene/container/container-mixins/effectsMixin.ts
```diff
@@ -21,6 +21,11 @@ export interface MaskOptions
     inverse: boolean;
 }
 
+export type MaskOptionsAndMask = MaskOptions &
+{
+    mask: Mask;
+};
```
bigtimebuddy, line 27: Why not interface?

```suggestion
export interface MaskOptionsAndMask extends MaskOptions
{
    mask: Mask;
}
```

GoodBoyDigital (approved):
👍

PR #10979, opened 2024-10-09
fix: custom render pipe crashing with gc

Zyie:
fixes the issue where the `RenderableGCSystem` is calling `destroyRenderable` on the `CusomtRenderPipe` but it does not exist


GoodBoyDigital (approved):
nice

PR #10785, opened 2024-07-19
chore: refactor Views to extend a shared class

GoodBoyDigital:
##### Description of change

Added a new `ViewContainer` Class that contains most of the shared logic for Views. Reducing the amount of code and generally making things a bit nicer!

image

I also simplified how didViewUpdate works - this is now also used by the pipes rather the views (rather than an additonal prop like `didSpriteUpdate`)

A view now just needs to override `updateBounds` as the `boundsDirty` is handled by the shared class.

Tests have been updated too!

Not Happy with `ViewContainer` as the name though.. suggestions welcome :D

##### Pre-Merge Checklist

- [x] Tests and/or benchmarks are included
- [x] Documentation is changed or added
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #10980, opened 2024-10-09
chore: unify onViewUpdate calls

Zyie:
another update from #10785

this one unifies the calls to `onViewUpdate` to be shared between the views


GoodBoyDigital (approved):
nice!

PR #10983, opened 2024-10-11
fix: avoid BitmapText resolution warning in constructor

rnike:
##### Description of change

This warning below always shows when `new BitmapText()`
https://github.com/pixijs/pixijs/blob/abcd2e38a28738045e99ac6bae2f2f4f4d6de090/src/scene/text-bitmap/BitmapText.ts#L153-L161

Because of this
https://github.com/pixijs/pixijs/blob/abcd2e38a28738045e99ac6bae2f2f4f4d6de090/src/scene/text/AbstractText.ts#L117

Don't show the warning if the value is null.

##### Pre-Merge Checklist


bigtimebuddy (approved):

creativoma:
It still happens, even in version 8.5.1 

```
const textTooltip = new PIXI.BitmapText({
  text: `FPS: loading`,
  resolution: devicePixelRatio / 2,
  style,
})
```

_PixiJS Warning:  [BitmapText] dynamically updating the resolution is not supported. Resolution should be managed by the BitmapFont._

bigtimebuddy:
@creativoma That is working as intended. You cannot set resolution with BitmapText.

PR #10984, opened 2024-10-11
remove `math-extras` mixin import

SaintPepsi:
Given that the functionality for `math-extras` is not included in the main `pixi.js` bundle when you're installing `pixi.js`, these types hint to the consumer they should be able to use methods that don't exist.

##### Description of change
Removed `MathExtraMixins.d.ts` mixin from `types/PixiMixins.d.ts`

##### Pre-Merge Checklist

### DIFF
--- a/types/PixiMixins.d.ts
+++ b/types/PixiMixins.d.ts
@@ -11,5 +11,4 @@
 /// <reference path="../src/scene/text-bitmap/TextBitmapMixins.d.ts" />
 /// <reference path="../src/scene/text-html/TextHTMLMixins.d.ts" />
 /// <reference path="../src/scene/particle-container/ParticleMixins.d.ts" />
-/// <reference path="../src/math-extras/MathExtraMixins.d.ts" />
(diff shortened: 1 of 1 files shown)


Zyie:
Hey @SaintPepsi 

`math-extras` is part of pixi.js now without needing to install a separate package. The only thing you need to do is import it

e.g.
`import 'pixi.js/math-extras'`

this mean that we do need the types here so going to close down this PR

ArtisanLin:
> Hey  嘿@SaintPepsi
> 
> `math-extras` is part of pixi.js now without needing to install a separate package. The only thing you need to do is import it`math-extras` 是 pixi.js Now 的一部分，无需安装单独的软件包。您唯一需要做的就是导入它
> 
> e.g. 例如。 `import 'pixi.js/math-extras'``导入 'pixi.js/math-extras'`
> 
> this mean that we do need the types here so going to close down this PR这意味着我们确实需要这里的类型，所以要关闭这个 PR

I think remove the type of math-extra is a good idea, when i haven't import the module, the type gives me a bad guide.

PR #11004, opened 2024-10-19
feat: add pixel line to graphics! (native line from v7)

GoodBoyDigital:
##### Description of change

Adds a new property called 'pixelLine' to lineStyle graphics that will use the native gpu line command to draw the line under the hood. This is exactly like the `native` property we used to have an v7! Only i believe a slightly better name!

Pixel lines are always 1px thick, no matter how far you zoom in or out on them. They can be very useful! This was quite a common feature request for v8. And now we have it, woop! one more checked off the list :)

##### Pre-Merge Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced support for pixel lines in rendering, enhancing line drawing capabilities.
	- Added a new `Topology` type for dynamic drawing modes based on batch properties.
	- Implemented a new `buildPixelLine` function for constructing pixel lines.

- **Bug Fixes**
	- Adjusted parameter handling in various rendering functions to improve topology management.

- **Documentation**
	- Updated comments and formatting for clarity across several files.

- **Tests**
	- Added a new test scene for verifying pixel line rendering functionality.


GoodBoyDigital:
@coderabbit review

GoodBoyDigital:
@coderabbitai review

tests/visual/scenes/graphics/graphics-pixel-line.scene.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { Container } from '../../../../src/scene/container/Container';
+import type { TestScene } from '../../types';
+
+export const scene: TestScene = {
+    it: 'should render a pixel line correctly',
+    create: async (scene: Container, _renderer: Renderer) =>
+    {
+        const graphics = new Graphics();
+
+        graphics
```
bigtimebuddy, line 13: Hmm this doesn't match the snapshot image. I only see the first black line. Am I missing something?

tests/visual/scenes/graphics/graphics-pixel-line.scene.ts
```diff
@@ -0,0 +1,24 @@
...
+            .moveTo(8, 0).lineTo(0, 8).moveTo(12, 20).lineTo(20, 12)
+            .stroke({ color: 0, pixelLine: true })
+            .rect(128 / 4, 128 / 4, 128 / 2, 128 / 2)
```
bigtimebuddy, line 16: I notice the stroke on the rect doesn't close, right? The upper left corner looks open.

bigtimebuddy:
Great solution! Much easier than we discussed. "pixelLine" I'm not 100% on that name, but I'm not sure I have anything better to offer. 🤷

GoodBoyDigital:
> Great solution! Much easier than we discussed. "pixelLine" I'm not 100% on that name, but I'm not sure I have anything better to offer. 🤷

agree! its a step up from 'native' which does not really mean anything unless your a gpu programer.. kind of?

100% open to better suggestions!

tests/visual/scenes/graphics/graphics-pixel-line.scene.ts
GoodBoyDigital: yep - that correct. they should appear as two dashes in the top left

bigtimebuddy:
gpuLine?
lineList?

ivanpopelyshev (changes requested):
I didnt see when people put "native lines" and regular lines in same graphics object.  Are you sure we acnt just add `nativeLine` thingy to the top level of GraphicsGeometry? And in that case render fills also as "lines", just triangulated, like a wiremesh mode. Please confirm that you have the case where it has to be like that. 

The reason is that when i'll be ready to do graphics-smooth stuff, this API might hinder me.

GoodBoyDigital:
> I didnt see when people put "native lines" and regular lines in same graphics object. Are you sure we acnt just add `nativeLine` thingy to the top level of GraphicsGeometry? And in that case render fills also as "lines", just triangulated, like a wiremesh mode. Please confirm that you have the case where it has to be like that.
> 
> The reason is that when i'll be ready to do graphics-smooth stuff, this API might hinder me.

Until i started working on the PR, i would have agreed with you! I initially created a new line class, that just rendered the line. But then i wanted batching... and multple styles per line (eg redline / white line in same object) and I realised that i would basically be copying huge parts of graphics. So instead, but a couple of extra ifs in the code and it was a lot easier. This was actually the shortest path that gives us good batched lines too.

Perhaps we can make the `buildContextBatches` a bit more modular and extensable to accommodate different render / build methods such as graphics smooth?

GoodBoyDigital:
> gpuLine? lineList? mode=normal|gpu?

My thinking is that I kind of want to move away from 'GPU' references as to the user is not really relevant. They are essentially asking for is a 1px width line - keen to capture that in a varialble iyswim.

ivanpopelyshev (approved):

src/scene/graphics/shared/FillTypes.ts
```diff
@@ -39,6 +39,8 @@ export interface StrokeAttributes
     join?: LineJoin;
     /** The miter limit to use. */
     miterLimit?: number;
+    /** If the stroke is a pixel line. */
+    pixelLine?: boolean;
```
Zyie, line 43: This type is also used by `Text`/`BitmapText` for its stroke properties, so we should omit this in `TextStyle.stroke`

GoodBoyDigital: good shout!

Zyie (approved):

PR #11008, opened 2024-10-21
fix: html text height

### LINKED ISSUE #10977: Bug: `CanvasTextMetrics.measureText` change height property of `HTMLText`
### Current Behavior

I've noticed strange thing. I have the `HTMLText` element and i want to measure it's height to adjust position of next element below.
I call `element.height` and get a number.
If I call `CanvasTextMetrcis.measureText(element.text, element.style)` before getting `element.height` - it change the value of `element.height`.

But it's not the most strange thing. 
I have several separate canvases with different content.
If i call `CanvasTextMetrcis.measureText()` in one canvas for random text element, it will cause changes of `element.height` on another canvas too!

### Expected Behavior

Strange side effects shouldn't exist i guess:)

### Steps to Reproduce

My case to complecated to paste it here:(

### Environment

- **`pixi.js` version**: *e.g. 8.2.4*
- **Browser & Version**: *e.g. Chrome 129*
- **OS & Version**: win11*

### Possible Solution

_No response_

### Additional Information

_No response_

GoodBoyDigital:
##### Description of change
fixes #10977. I can no longer see a need to add the descender to the htmlText when measuring it. Removing it gives a much closer value to the Canvas measure text function. @bigtimebuddy , let me know if this looks ok!

##### Pre-Merge Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "watch:build": "run-s build:index build:rollup build:tsc build:dts postbuild",
     "test": "run-s test:unit test:scene",
     "test:unit": "npx jest --silent --testPathIgnorePatterns=tests/visual",
-    "test:debug": "cross-env DEBUG_MODE=1 npx jest --testPathIgnorePatterns=tests/visual",
+    "test:debug": "cross-env DEBUG_MODE=1 npx jest",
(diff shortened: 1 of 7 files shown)


bigtimebuddy (approved):
This sounds reasonable to me.

PR #11020, opened 2024-10-30
fix: guard renderable access

rigor789:
Noticing this error on iOS in WKWebView occasionally when suspending/resuming the app:

```
TypeError: undefined is not an object (evaluating 'this.renderable.groupBlendMode')
```

image

##### Description of change

Added guards whenever accessing `renderable` which can be `null` in some cases.

Marking this as a draft until I get a chance to test whether or not this fixes the issue completely, or if it's a side-effect of another issue.

##### Pre-Merge Checklist


GoodBoyDigital:
thanks for sharing this @rigor789 , i would like to understand the root of this issue a little better as the renderables should not be null. Will get back to you here!

rigor789:
@GoodBoyDigital definitely, keeping it a draft here as I do think it may be a side-effect of another issue... I'll post anything else I can find here, thanks!

GoodBoyDigital:
heya, just following up here:
- are you destroying things at all when this happens?
- if you add the line `renderer.renderableGC.enabled = false` does it fix the issue?
- a long shot - but do you have a minimal example i could debug pls?

Thanks!

rigor789:
@GoodBoyDigital Hey, sorry for the slow response, I am not destroying things as this happens, will try turning gc off and see if that makes a difference.

I'll try to reproduce it in a minimal example, though I would say the chances are pretty slim, as it's somewhat inconsistent in the prod (soon to be) project.

I'm a a bit swamped with work on the project, but I'll do my best to get something useful back here as soon as I can! Thanks for keeping tabs on this, really appreciated!

GoodBoyDigital:
> @GoodBoyDigital Hey, sorry for the slow response, I am not destroying things as this happens, will try turning gc off and see if that makes a difference.
> 
> I'll try to reproduce it in a minimal example, though I would say the chances are pretty slim, as it's somewhat inconsistent in the prod (soon to be) project.
> 
> I'm a a bit swamped with work on the project, but I'll do my best to get something useful back here as soon as I can! Thanks for keeping tabs on this, really appreciated!

hey, no worries I know how it goes :) 
Im 99% sure it will be th gc and turning it off will fix it at least!

GoodBoyDigital:
heya! i a have an incoming fix for this - closing for now :)

PR #11031, opened 2024-11-04
feat: cacheAsTexture

GoodBoyDigital:
##### Description of change
With this PR, containers can have a `cacheAsTexture` property. It behaves similarly to `cacheAsBitmap` from v7, but I changed the name as we no longer refer to textures as bitmaps.

Setting `cacheAsTexture` to `true` will render the container as a texture, and that texture will be displayed instead of rendering the container directly. This offers significant performance gains for complex, static containers.

Under the hood, the container is assigned a render group, allowing it to be rendered as a texture with a dynamic set of instructions.

Calling `container.updateRenderTexture()` will re-render the container on the next frame, a very fast process with minimal overhead (apart from the render itself).

As an optimization, the textures used come from the texture pool (the same pool that filter textures use), which speeds up `cacheAsTexture` initialization—especially if the container already has an assigned render group. 

Example

```
const myComplexContainer = new Container();

// add loads of stuff! to it..

myComplexContainer.cacheAsTexture = true;

// or

myComplexContainer.enableCacheAsTexture({antialias: true, resolution:2});

```
 

##### Pre-Merge Checklist

## Release Notes

- **New Features**
	- Introduced caching functionality for containers to improve rendering performance.
	- Added a new method for copying bounds in the `Bounds` class.
	- Implemented a new test scene to verify cached texture rendering.

- **Improvements**
	- Enhanced rendering logic with new methods for managing cached textures in the `RenderGroup` and `RenderGroupSystem` classes.
	- Updated method names for clarity and improved control flow in rendering processes.

- **Bug Fixes**
	- Refined validation checks and error handling in rendering methods.

- **Chores**
	- Removed deprecated functions and unnecessary imports to streamline code.


GoodBoyDigital:
@coderabbitai review

bigtimebuddy:
Just to present an alt API concept of using a single function instead of  `get cacheAsTexture`, `set cacheAsTexture`, `enableCacheAsTexture` and `disableCacheAsTexture`.

```ts
container.cacheAsTexture(true);
container.cacheAsTexture(); // return true
container.cacheAsTexture({ resolution: 2, antialias: true });
container.cacheAsTexture(false);
```

GoodBoyDigital:
@coderabbitai review

GoodBoyDigital:
> Just to present an alt API concept of using a single function instead of `get cacheAsTexture`, `set cacheAsTexture`, `enableCacheAsTexture` and `disableCacheAsTexture`.
> 
> ```ts
> container.cacheAsTexture(true);
> container.cacheAsTexture(); // return true
> container.cacheAsTexture({ resolution: 2, antialias: true });
> container.cacheAsTexture(false);
> ```

interesting! im very open to spending a little time getting this api right. I think this is a better direction!
my only reservations would be 

``` container.cacheAsTexture(); // return true```

As i think this looks me like an action - as in calling it like this would enable cacheAsBitmap. But the other use cases look good!

**Comment by bigtimebuddy** [comment]
Yeah, I think we can get rid of the getter for undefined argument. Not sure that's necessary.

**Comment by bigtimebuddy** [comment]
Are you going to rework any of the APIs like we discussed above an only create a single `cacheAsTexture` method?

**Comment by GoodBoyDigital** [comment]
> Are you going to rework any of the APIs like we discussed above an only create a single `cacheAsTexture` method?

yes indeed! mr @Zyie will be taking that one!

File: src/rendering/renderers/shared/system/AbstractRenderer.ts [hunk 1]
```diff
@@ -301,6 +301,10 @@ export class AbstractRenderer<
             options.transform = options.container.localTransform;
         }
 
+        // lets ensure this object is a render group so we can render it!
+        // the renderer only likes to render - render groups.
+        options.container.enableRenderGroup();
```
Zyie, line 306: Would we need to disable the render group after the render?

src/scene/container/RenderGroupSystem.ts
```diff
@@ -63,49 +62,12 @@ export class RenderGroupSystem implements System
             container.renderGroup.localTransform.copyFrom(transform);
         }
 
+        //  this._assignTop(container.renderGroup, null);
```
Zyie, line 65:
```suggestion
```

src/scene/container/container-mixins/cacheAsTextureMixin.ts
```diff
@@ -0,0 +1,82 @@
...
+     * // you're done?
+     * container.cacheAsTexture(false);
+     *
+     * ```
+     */
+    cacheAsTexture(val: boolean | CacheAsTextureOptions): void
+    {
+        if (typeof val === 'boolean' && val === false)
+        {
+            this.disableRenderGroup();
```
Zyie, line 64: Do we need to call `disableCacheAsTexture` on the render group here?
I noticed the function is implemented but never called

Zyie (commented):
@bigtimebuddy API has been updated for you to review

bigtimebuddy (approved):

src/scene/container/container-mixins/cacheAsTextureMixin.ts
GoodBoyDigital: I added it for just such an occasion!

src/rendering/renderers/shared/system/AbstractRenderer.ts
GoodBoyDigital: no, not for now - we are assuming that things will generally be rendered multiple times, setting it to false after a render will cause everything to rebuild each frame. 

But i do agree that this is kind of a hidden side effect. Lets think of a better way to handle this outside of this PR

PR #11032, opened 2024-11-04
fix: stroke should use alignment when extending bounds

dapi303:
##### Description of change
When stroke instruction is used, bounds are extended by width of the stroke.
It should use alignment not hardcoded 0.5 multiplier.

Before this change, the element has increased size even if the alignment is set to 1 and stroke is inside of the original shape which should not affect bounds.

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/graphics/shared/GraphicsContext.ts
+++ b/src/scene/graphics/shared/GraphicsContext.ts
@@ -1046,7 +1046,9 @@ export class GraphicsContext extends EventEmitter<{
             {
                 const data = instruction.data as StrokeInstruction['data'];
 
-                const padding = data.style.width / 2;
+                const alignment = data.style.alignment;
+
+                const padding = (data.style.width * (1 - alignment));
 
                 const _bounds = data.path.bounds;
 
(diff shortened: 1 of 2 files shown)


bigtimebuddy:
@GoodBoyDigital @Zyie I thought in v6 (#8343) we did away with this type of stroke bounds padding and instead used the line vertices to ensure miterlimit and line joins were handled correctly. Did this get undone in v8?

GoodBoyDigital (approved):
nice!

GoodBoyDigital:
> @GoodBoyDigital @Zyie I thought in v6 (#8343) we did away with this type of stroke bounds padding and instead used the line vertices to ensure miterlimit and line joins were handled correctly. Did this get undone in v8?

We don't actually have access to the verticies in v8 as the rendering of the graphics is decoupled from Graphics itself. The measuring process should work ideally on say, the canvas renderer also (which does not exist just yet :P) which would not require the lines to be tessellated to render.

bigtimebuddy (approved):
Thank you @dapi303!

PR #11054, opened 2024-11-12
fix: Use multiply instead of mix colors when propagating tint to children

lunarraid:
##### Description of change
Since `.tint` on containers is a multiply operation, this changes tint stacking to multiply the colors when propagating to children instead of mixing by averaging. 

##### Pre-Merge Checklist


GoodBoyDigital (approved):
lovely PR! Thanks @lunarraid 👍

PR #11057, opened 2024-11-14
feat: Add some new functions to get global tint / alpha / transform

GoodBoyDigital:
##### Description of change

Adds three new functions to Container to make it easier to work with global transforms / tints / alphas.
In keeping with the `getGlobalBounds` style functions we now have:

**getGlobalTransform**
```
const skipUpdate = false;
const outMatrix = new Matrix();

const globalTransform = container.getGlobalTransform(outMatrix, skipUpdate); // writes and returns outMatrix;
```
**getGlobalTint**
```
const skipUpdate = false;
const globalTint = container.getGlobalTint(skipUpdate); // returns rgb color
```
**getGlobalAlpha**
```
const skipUpdate = false;
const globalAlpha = container.getGlobalAlpha(skipUpdate); // returns alpha as number;
```

All options are optinal and by defualt skipUpdate will be false.

if `skipUpdate` is true - it will be faster but may be outdated - uses the last rendered data
if `skipUpdate` is false - it will be 100% accurate but slower - recalculates transform chain

##### Pre-Merge Checklist


src/scene/container/Container.ts
```diff
@@ -852,7 +855,64 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
...
+     * const preciseTransform = container.getGlobalTransform();
+     *
+     * // Faster but may be outdated - uses cached transform
+     * const cachedTransform = container.getGlobalTransform(undefined, true);
+     *
+     * // Reuse existing matrix
+     * const existingMatrix = new Matrix();
+     * container.getGlobalTransform(existingMatrix);
+     */
+    public getGlobalTransform(matrix?: Matrix, skipUpdate?: boolean): Matrix
```
bigtimebuddy, line 876: What do you think about this arguments order? Would skipUpdate then matrix make more or less sense? I'm not sure I have a POV but am asking. If both args are equally optional, then options may be better choice.

bigtimebuddy (commented):
Overall, I like the method names, just a question about the args in the transform method.

src/scene/container/Container.ts
GoodBoyDigital: I was following the getGlobalBounds sig - but would be open to changing them all to be option objects if we think thats a better direction! 

I would like to encourage devs to pass in a matrix - so that a new one is not created each time the function is called.

bigtimebuddy: I am okay with this I think unless @Zyie has any other thoughts.

bigtimebuddy (approved):

PR #11058, opened 2024-11-14
fix: If using multiview rendering, call ensureCanvasSize before rendering

### LINKED ISSUE #10972: Bug: When resizing renderer with `multiView: true`, need to ensure the render canvas size also
### Current Behavior

We're using `multiView` rendering for our app, re #10913 
(h/t @Zyie , thanks!!)

I realized that when we resize our app, we also need to manually resize the rendering canvas..  
(i.e. If our app resizes larger, the rendered output remains cut off at the original size)

### Expected Behavior

I'd expect a `renderer.resize()` to trigger `ensureCanvasSize()` automatically.

### Steps to Reproduce

see solution section below - without the `ensureCanvasSize` line, the rendered output may be cut off.

### Environment

- **`pixi.js` version**: *e.g. 8.4.1*

### Possible Solution

Here's our workaround:

```js
// 'w', 'h' contain the new size
// 'ourcanvas' is our main canvas

// Resize pixi canvas
const renderer = this.pixi.renderer;
renderer.resize(w, h);

// needed for multiview?
// If we are using the WebGL renderer and have multiView enabled,
// Pixi will render to a different canvas before copying to the target canvas.
// The render canvas may need a resize too.
if (renderer.type === PIXI.RendererType.WEBGL && renderer.context.multiView) {
  renderer.context.ensureCanvasSize(ourcanvas);
}

```

### Additional Information

Would love to send a pull request, but I don't know where this would belong, if it's something you want - there is already a bunch of code to handle resizing.

bhousel:
(fixes #10972)
Thanks for pointing me in the right direction @GoodBoyDigital  👍 
I confirmed that this fixes the resizing issue with my local build of Pixi.

##### Pre-Merge Checklist


bhousel:
> I just pushed up a little change that makes this check only happens at the start of a render, rather than each time a render target is changed (this happens at the start, but also when we use filters etc)

oh thanks! I didn't see your comment before... 
This is very interesting because some of our testers have reported an issue where the filters seem to be using the wrong dimensions.. I haven't been able to reproduce it but maybe your fix covers this case?

filter cut off

bhousel:
> Let me know if this still works for you locally! Thanks man!

I tested this locally and it looks good, thanks!

GoodBoyDigital (approved):

GoodBoyDigital:
> > I just pushed up a little change that makes this check only happens at the start of a render, rather than each time a render target is changed (this happens at the start, but also when we use filters etc)
> 
> oh thanks! I didn't see your comment before... This is very interesting because some of our testers have reported an issue where the filters seem to be using the wrong dimensions.. I haven't been able to reproduce it but maybe your fix covers this case?
> 
> filter cut off

interesting, mind opening a seperate issue for this it the issue is still present please? Thanks!

Issue #11059, opened 2024-11-14
Bug: GradientFill leaks memory


Liam-Hegarty:
### Current Behavior

I am seeing memory leak whilst trying to dynamically GradientFill a graphics object (think of a skyline with a day/night cycle). I have made a quick example here:

https://www.pixiplayground.com/#/edit/y_qhrsKrn7b6lwHZ0_mX2

In my chrome developer tools [Version 131.0.6778.69 (Official Build) (arm64)] I see two things:

1. Memory usage steadily increasing
2. A growing list of detached canvas elements `<canvas width=256 height=256>`

I've attached a couple of things in case you can't replicate, a heap snapshot and an allocation timeline:

[Archive.zip](https://github.com/user-attachments/files/17756204/Archive.zip)

It's definitely the fill command on L31 that causes the leak, commenting it out stops the above issues.

Would be amazing if you could resolve this as it makes my game crash.

### Expected Behavior

No memory usage increase

### Steps to Reproduce

Continually fill a graphics element with a gradient fill. (eg: https://www.pixiplayground.com/#/edit/y_qhrsKrn7b6lwHZ0_mX2)

### Environment

- **`pixi.js` version**: *8.5.2*
- **Browser & Version**: *131.0.6778.69 (Official Build) (arm64)*
- **OS & Version**: *MacOS Sonoma 14.0*
- **Running Example**: *https://www.pixiplayground.com/#/edit/y_qhrsKrn7b6lwHZ0_mX2*

bigtimebuddy:
Duplicate of #10936?

Liam-Hegarty:
Absolutely, not sure how I missed that

PR #11061, opened 2024-11-14
fix: FillGradient leak

### LINKED ISSUE #10936: Bug: FillGradient fill causes memory leak (creates one texture per frame)
### Current Behavior

When using FillGradient my WebGL context crashes after a certain amount of time. Beforehand the fps drop significantly and using the DevTools Performance recording, it's possible to observe more and more skipped frames without any real scripting load.

### Expected Behavior

No crashes and stable fps.

### Steps to Reproduce

1. Open example page: https://pixijs.com/8.x/examples/graphics/fill-gradient
2. Wait for 30-60s

I've experienced a similar behavior in my more complex app using the pixi-react beta, but since the example crashes my webgl context even faster I think it might be a better reproducer. Though it could be connected to the OS/Vulkan, so I hope you will be able to reproduce the issue.

### Environment

- **`pixi.js` version**: 8.4.1
- **Browser & Version**: Vivaldi 6.9.3447.46 (Stable channel) stable (64-bit) 
- **OS & Version**: Linux Mint 21.3
- Attached [the vivaldi://gpu report](https://github.com/user-attachments/files/17106052/about-gpu-2024-09-23T23-53-08-124Z.txt), including the error logs at the bottom, that have happened within roughly 30s after opening the example page

### Possible Solution

_No response_

### Additional Information

Might be connected to: https://stackoverflow.com/questions/78953390/pixijs-fillgradient-memory-leak

GoodBoyDigital:
##### Description of change
Silly us, we were creating the texture for the gradient each frame. This is no longer the case, create once use - often!

fixes #10936 and #11059 

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/graphics/shared/fill/FillGradient.ts
+++ b/src/scene/graphics/shared/fill/FillGradient.ts
@@ -63,6 +63,8 @@ export class FillGradient implements CanvasGradient
     // TODO move to the system!
     public buildLinearGradient(): void
     {
+        if (this.texture) return;
+
         const defaultSize = FillGradient.defaultTextureSize;
 
         const { gradientStops } = this;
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Easy enough! Thanks @GoodBoyDigital

braebo:
I'm trying to install the fixed dev branch to stop it from crashing our app, but I'm having issues... I guess I need to clone and build it locally?

Is there a planned release schedule that I can use to mark the fix release date in my calendar?

s10y10:
I used to modify x0, y0, x1, y1 and colorStop in the Ticker.update. After this change, is there any good way to dynamically modify the properties of fillGradient?

PR #11075, opened 2024-11-19
feat: enhance how gradients are applied to text objects

GoodBoyDigital:
##### Description of change

# Text Gradient Improvements

This PR enhances how gradients are applied to text objects, making them more intuitive and flexible.

## Changes
- Gradients now default to 'local' space, stretching to fill the entire text object
- Special handling for vertical gradients
  - Automatically applies gradient per text line
  - Maintains consistent gradient appearance across wrapped text
- Retains option for 'global' space gradients via `textureSpace` property

vertical:

```
const grad = new PIXI.FillGradient(0, 0, 0, 1);
grad.addColorStop(0, 'green');
grad.addColorStop(1, 'purple');
```

image

non-vertical:

```
const grad = new PIXI.FillGradient(0, 0, 1, 1);
grad.addColorStop(0, 'green');
grad.addColorStop(1, 'purple');
```

image

##### Pre-Merge Checklist


src/scene/graphics/shared/fill/FillGradient.ts
```diff
@@ -4,105 +4,304 @@ import { Matrix } from '../../../../maths/matrix/Matrix';
...
+     */
+    y0?: number;
+    /** The radius of the inner circle where the gradient begins. */
+    r0?: number;
+    /** The x coordinate of the center point of the outer circle where the gradient ends. In local coordinates by default. */
+    x1?: number;
+    /** The y coordinate of the center point of the outer circle where the gradient ends. In local coordinates by default. */
+    y1?: number;
+    /** The radius of the outer circle where the gradient ends. */
+    r1?: number;
```
bigtimebuddy, line 75: I find context2d's radial gradient API to be terrible. It's not practically useful and doesn't map on well to other renderers. What I really want from this API is elliptical gradients with rotation. From a practical design stand-point, I think using elliptical gradients are far more useful than `createRadialGradient`. 

All that is to say, I'm not in love that our interface here is a pass-thru for `createRadialGradient` without the ability to use elliptical gradients. Also the `x0`, `y0`, etc prop names aren't very user-friendly. My preference would be use to `center: PointData` (and same with linear start/end points to use PointData).

src/scene/graphics/shared/fill/FillGradient.ts
```diff
@@ -134,17 +332,160 @@ export class FillGradient implements CanvasGradient
...
+    }
+}
+
+function getCanvas(width: number, height: number): CanvasAndContext
+{
+    const canvas = DOMAdapter.get().createCanvas();
+
+    canvas.width = width;
+    canvas.height = height;
+
```
bigtimebuddy, line 443:
```suggestion
    const canvas = DOMAdapter.get().createCanvas(width, height);
```

src/scene/graphics/shared/fill/FillGradient.ts
bigtimebuddy: Nice, the names are better and more user-friendly! 

However, I still don't agree than we should pass-thru for `createRadialGradient` with just nicer names. Those controls (inner and outer) are _narrowly_ useful. Here's an example of what you can do with this. I think this is an edge-case.

Screenshot 2025-01-27 at 1 51 47 PM

No modern design tools that support radial gradient render gradients render like this (inner and outer concept). Figma, for instance, the emphasis is on transforming gradients to create a variety of elliptical and rotated gradients.

Jan-27-2025 13-57-26

```ts
/** The center of the inner circle */
innerCenter?: PointLike;
/* The radius of the inner circle */
innerRadius?: number;
/* The center of the outer circle */
center: PointLike;
/* The radius of the outer circle */
radius: number;

// Would add these:
/** Transform the gradient, defaults to { x: 1, y: 1 } */
scale?: PointLike;
/** Rotation in degrees, useful for elliptical gradients, applied after scale */
rotation?: number;
```

I would require center and radius and then `innerCenter` and `innerRadius` can still support that edge-case:

```ts
const x0 = innerCenter?.x ?? center.x;
const y0 = innerCenter?.y ?? center.y;
const r0 = innerRadius ?? 0;
const x1 = center.x;
const y1 = center.y;
const r1 = radius;

ctx.createRadialGradient(x0, y0, r0, x1, y1, r1);
```

GoodBoyDigital: oh, yes that makes sense! let me see if i can incorporate those two values. 
il keep inner and outter as they are useful for rendering svg gradients correctly. But i will also make it so that outerCenter matches innerCenter if its not defined. Infact I'll rename innerCenter -> center too!

GoodBoyDigital:
Hey @bigtimebuddy , I believe this is in a place to review

Overall I think this is cool! Im on the fence about scale and rotation as I can't apply this to text unfortunately only to graphics (canvas limitations). Well technically I could - but it would involve a lot more work that I don't thing is worth it right now I beleive.

Apart from that, functionally its all there though - We just need to make sure we are happy with the new API. I feel like its getting kind of big, but maybe thats cos I have stared at it too long! Let me know your thoughts @bigtimebuddy !

Thanks!

src/scene/graphics/shared/fill/FillGradient.ts
```diff
@@ -4,111 +4,364 @@ import { Matrix } from '../../../../maths/matrix/Matrix';
...
+ *
+ * For linear gradients, color stops define colors and positions (0 to 1) along a line from start point (x0,y0)
+ * to end point (x1,y1).
+ *
+ * For radial gradients, color stops define colors between two circles - an inner circle centered at (x0,y0) with radius r0,
+ * and an outer circle centered at (x1,y1) with radius r1.
+ * @example
+ * ```ts
+ * // Create a vertical linear gradient from red to blue
+ * const linearGradient = new FillGradient({
```
bigtimebuddy, line 115: These examples don't seem to reflect the API changes

bigtimebuddy (approved):
Great work! And thanks for talking through the API, appreciate your patience.

src/scene/graphics/shared/fill/FillGradient.ts
```diff
@@ -4,111 +4,364 @@ import { Matrix } from '../../../../maths/matrix/Matrix';
...
+ *     textureSpace: 'global'  // Use world coordinates
+ * });
+ *
+ * // Create an offset radial gradient
+ * const offsetRadial = new FillGradient({
+ *     type: 'radial',
+ *     x0: 0.3, y0: 0.3,  // Inner circle offset from center
+ *     r0: 0.1,           // Small inner circle
+ *     x1: 0.5, y1: 0.5,  // Outer circle centered
+ *     r1: 0.5,           // Large outer circle
```
bigtimebuddy, line 160:
```suggestion
 * const linearGradient = new FillGradient({
 *     type: 'linear',
 *     start: { x: 0, y: 0 },  // Start at top
 *     end: { x: 0, y: 1 },    // End at bottom
 *     colorStops: [
 *         { offset: 0, color: 'red' },   // Red at start
 *         { offset: 1, color: 'blue' }   // Blue at end
 *     ],
 *     textureSpace: 'local'
 * });
 *
 * // Create a radial gradient from yellow center to green edge
 * const radialGradient = new FillGradient({
 *     type: 'radial',
 *     center: { x: 0.5, y: 0.5 },
 *     innerRadius: 0,
 *     outerCenter: { x: 0.5, y: 0.5 },
 *     outerRadius: 0.5,
 *     colorStops: [
 *         { offset: 0, color: 'yellow' }, // Center color
 *         { offset: 1, color: 'green' }   // Edge color
 *     ],
 *     textureSpace: 'local'
 * });
 *
 * // Create a rainbow linear gradient in global coordinates
 * const globalGradient = new FillGradient({
 *     type: 'linear',
 *     start: { x: 0, y: 0 },
 *     end: { x: 100, y: 0 },
 *     colorStops: [
 *         { offset: 0, color: 0xff0000 },    // Red
 *         { offset: 0.33, color: 0x00ff00 }, // Green
 *         { offset: 0.66, color: 0x0000ff }, // Blue
 *         { offset: 1, color: 0xff00ff }     // Purple
 *     ],
 *     textureSpace: 'global'  // Use world coordinates
 * });
 *
 * // Create an offset radial gradient
 * const offsetRadial = new FillGradient({
 *     type: 'radial',
 *     center: { x: 0.3, y: 0.3 },
 *     innerRadius: 0.1,
 *     outerCenter: { x: 0.5, y: 0.5 },
 *     outerRadius: 0.5,
```

bigtimebuddy (commented):
Are start/end in real pixel values or normalized floats? I see both being used here.

Zyie:
> Are start/end in real pixel values or normalized floats? I see both being used here.

If you are using `textureSpace: 'local'` then the values are normalised, if `'global'` then its real pixel values. I added a comment to the examples to help explain this

src/scene/text-html/HTMLText.ts
```diff
@@ -1,5 +1,5 @@
 import { AbstractText, ensureTextOptions } from '../text/AbstractText';
-import { HTMLTextStyle } from './HtmlTextStyle';
+import { HTMLTextStyle } from './HTMLTextStyle';
```
bigtimebuddy, line 2: @Zyie we should add [forceConsistentCasingInFileNames](https://www.typescriptlang.org/tsconfig/#forceConsistentCasingInFileNames) to our tsconfig.json to make sure that this is caught with a type check.

PR #11080, opened 2024-11-21
feat: support SVG non-zero fills

GoodBoyDigital:
##### Description of change
Add support for non-zero fills in svg. This is the most common method and turns out we have most of the functionality to support this already! A few tweaks and now most SVGs will render correctly!

before:
image
now:
image

##### Pre-Merge Checklist


bigtimebuddy (approved):
Nice!

PR #11083, opened 2024-11-22
Fix: Correct color format in `addColorStop` function for Android WebV…

AlexanderShushunov:
Android WebView does not support the `#RRGGBBAA` color format in the `addColorStop` method. Using this format results in the following error:

```
SyntaxError: Failed to execute 'addColorStop' on 'CanvasGradient': The value provided ('#000000ff') could not be parsed as a color.
```
For further details, refer to the Chromium issue tracker: [Chromium Issue #40222657](https://issues.chromium.org/issues/40222657).

### DIFF
--- a/src/scene/graphics/shared/fill/FillGradient.ts
+++ b/src/scene/graphics/shared/fill/FillGradient.ts
@@ -54,7 +54,7 @@ export class FillGradient implements CanvasGradient
 
     public addColorStop(offset: number, color: ColorSource): this
     {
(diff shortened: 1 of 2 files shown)


bigtimebuddy (commented):
There are a lot of places where we are using toHexa. I think we need a more general solution here.

bigtimebuddy:
The hacky solution might be to do:

```js
// use this for browsers not supporting #RRGGBBAA
Color.prototype.toHexa = function toHexa() {
  return this.toRgbaString();
};
```

GoodBoyDigital:

PR #11094, opened 2024-11-26
Feature/layers

GoodBoyDigital:
##### Description of change
The RenderLayer API allows developers to control the rendering order of objects independently of their logical parent-child relationships in the scene graph. This decouples object transformations from their rendering order, leading to more flexible and manageable scene graphs. 

Key Concepts:
- RenderLayers define where objects are drawn in the render stack but do not affect their transformations or hierarchy.
- Objects maintain their logical parent for transformations.
- Developers explicitly assign objects to layers using `renderLayer.add` and remove them using `renderLayer.remove`.

API Details:
1. **Creating a RenderLayer**:
   ```js
   const layer = new RenderLayer();
   app.stage.addChild(layer);
   ```
2. **Adding Objects to a Layer**:
   ```js
   const rect = new PIXI.Graphics();
   container.addChild(rect);
   layer.add(rect);
   ```
3. **Removing Objects from a Layer**:
   ```js
   layer.remove(rect);
   ```
4. **Re-Adding Objects to Layers**:
   ```js
   container.addChild(rect);
   layer.add(rect);
   ```
5. **Layer Position in Scene Graph**:
   ```js
   const backgroundLayer = new RenderLayer();
   const uiLayer = new RenderLayer();
   app.stage.addChild(backgroundLayer);
   app.stage.addChild(world);
   app.stage.addChild(uiLayer);
   ```

This API has been tagged experimental and may have some known issues, such as interaction and hit testing not accounting for visual render order, and potential issues with filters and masks. Despite these, I think this API is highly useful even as it is now.

##### Pre-Merge Checklist


src/scene/layers/__tests__/RenderLayer.test.ts
```diff
@@ -0,0 +1,144 @@
+import { Container } from '../../container/Container';
```
bigtimebuddy, line 1:
```suggestion
import { Container } from '~/scene';
```

bigtimebuddy (approved):
While experimental, I think this is well-considered and well-documented. I believe this is enough to get more feedback in our next feature release. Great work.

ivanpopelyshev:
I remembered sorting test for pixi-layers: make many containers with three elements: 
Character sprite
His nickname (label on top)
Shadow

All three elements are in different render layers.
Need a way to sort those containers by Y coord that way all three layers will change order inside.

src/scene/layers/RenderLayer.ts
```diff
@@ -0,0 +1,267 @@
...
+        this.sortFunction = options.sortFunction;
+    }
+
+    /**
+     * Add an object to this render layer. The object will be rendered as part of this layer
+     * while maintaining its original parent in the scene graph.
+     * If the object already belongs to a layer, it will be removed from the old layer before being added to this one.
+     * @param child - The object to add to this layer
+     */
+    public attach(child: Container)
```
Zyie, line 156: should this be the same API as `addChild` which accepts multiple children e.g `layer.attach(a, b, c)`

Zyie (approved):

PR #11095, opened 2024-11-26
fix: bitmap font texture takes into account incoming texture rotation

raejuli:
##### Description of change

Bitmap font takes into account the incoming texture's rotation property so the resulting bitmap characters render in the correct orientation.

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/text-bitmap/BitmapFont.ts
+++ b/src/scene/text-bitmap/BitmapFont.ts
@@ -47,6 +47,7 @@ export class BitmapFont extends AbstractBitmapFont<BitmapFont>
             const {
                 frame: textureFrame,
                 source: textureSource,
+                rotate: textureRotate
             } = textures[charData.page];
 
             const frameReal = new Rectangle(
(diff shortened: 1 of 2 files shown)


src/scene/text-bitmap/BitmapFont.ts
```diff
@@ -47,6 +47,7 @@ export class BitmapFont extends AbstractBitmapFont<BitmapFont>
             const {
                 frame: textureFrame,
                 source: textureSource,
+                rotate: textureRotate
```
GoodBoyDigital, line 50: hey @raejuli can't seem to see where `textureRotate` is declared?

GoodBoyDigital (changes requested):
thanks for this - looks like there may be an issue as we cannot see where `textureRotate` is declared? Mind taking a look please?

@Zyie can help fix up the tests!

GoodBoyDigital:
closing as inactive - @raejuli please update if you get a chance. Thanks!

PR #11102, opened 2024-11-28
fix: Better undefined check in beginFill deprecation

### LINKED ISSUE #11101: Bug: Setting alpha to 0. in graphics.beginFill has the same effect as setting it to 1.
### Current Behavior

Setting alpha to 0. in graphics.beginFill has the same effect as setting it to 1. 

### Expected Behavior

Setting alpha to 0. in graphics.beginFill should be the limit of setting it to 0.1, 0.01, ...

### Steps to Reproduce

```
(async () => {
    await app.init({
        width: window.innerWidth,
        height: window.innerHeight
    })

    app.stage.addChild(graphics);

    graphics.beginFill(0xFF0000, 0.)

    graphics.drawRect(0, 0, 300, 200);

    graphics.endFill();

})();
```

### Environment

- **`pixi.js` version**: 8.6.1
- **Browser & Version**: firefox, chrome
- **Running Example**: https://www.pixiplayground.com/#/edit/5Fw4IN9ZZ-yEo68SXvO1P

### Possible Solution

_No response_

### Additional Information

I am aware that this API is deprecated and it looks lie like the new interface gets it right. However I am very sad about the breakage - I am a visual artist and I lost the opportunity to do several performances because things were inexplicably broken after a pixi upgrade.

bigtimebuddy:
Closes #11101

### DIFF
--- a/src/scene/graphics/shared/Graphics.ts
+++ b/src/scene/graphics/shared/Graphics.ts
@@ -754,8 +754,8 @@ export class Graphics extends ViewContainer implements Instruction
         const fillStyle: Partial<FillStyle> = {};
 
         // avoid undefined assignment
-        color && (fillStyle.color = color);
-        alpha && (fillStyle.alpha = alpha);
+        if (color !== undefined) fillStyle.color = color;
+        if (alpha !== undefined) fillStyle.alpha = alpha;
 
         this.context.fillStyle = fillStyle;
 


bigtimebuddy:
Verified: https://codesandbox.io/p/sandbox/pixi-js-sandbox-forked-sqhdvf (should see black screen)

Zyie (approved):

PR #11119, opened 2024-12-02
fix: Memory leak during canvas extraction

mcdu11:
When I tried to shard an image from a very large container, there was a serious memory leak

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/extract/ExtractSystem.ts
+++ b/src/rendering/renderers/shared/extract/ExtractSystem.ts
@@ -230,7 +230,7 @@ export class ExtractSystem implements System
 
         const canvas = renderer.texture.generateCanvas(texture);
 
-        texture.destroy();
+        texture.destroy(true);
 
         return canvas;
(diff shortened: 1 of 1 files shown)


src/rendering/renderers/shared/extract/ExtractSystem.ts
```diff
@@ -230,7 +230,7 @@ export class ExtractSystem implements System
 
         const canvas = renderer.texture.generateCanvas(texture);
 
-        texture.destroy();
+        texture.destroy(true);
```
bigtimebuddy, line 233: Good catch. This is likely an issue with pixels() below too. Would you mind adding that as well?

mcdu11: Sure! I missed it

bigtimebuddy (approved):

GoodBoyDigital (approved):

PR #11127, opened 2024-12-04
chore: update `Assets.ts` comment for src file ext

furic:
##### Description of change
Small comment fix, just adding the missing `.` in `bunny{png,webp}`, which should match the `bunny.png` and `bunny.webp` in the same comment.

(Currently, it's kinda confusing, I thought we need to put file ext like that and Pixi will add them automatically, i.e. `bunny{png,webp}` to `bunny.png` and `bunny.webp` but it's not)

This should fix the [online docs](https://pixijs.download/dev/docs/assets.Assets.html) too.

##### Pre-Merge Checklist

### DIFF
--- a/src/assets/Assets.ts
+++ b/src/assets/Assets.ts
@@ -397,7 +397,7 @@ export class AssetsClass
      * // passing options to to the object
      * Assets.add({
      *     alias: 'bunnyBooBooSmooth',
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):

bigtimebuddy:
Thank you @furic!

PR #11151, opened 2024-12-12
fix: unload family font caching

JenovaSolier:
- Since changed font name when caching, can't find that font from it font family name.

##### Pre-Merge Checklist


GoodBoyDigital (approved):
nice catch! Thanks for this!
bonus points if you could add a test :)

PR #11152, opened 2024-12-12
feat: allow for graphics.fill(texture)

GoodBoyDigital:
##### Description of change

You can now pass a texture directly to a graphics!

```
const texture = await PIXI.Assets.load('assets/bg_scene_rotate.jpg') ;

const g = new Graphics()
    .circle(50,50,50)
    .fill(texture);
```

image

- also switched default TextureSpace to be `local` this is MUCH more intuative space to work in for devs. 
- added some tests too!

##### Pre-Merge Checklist


bigtimebuddy (approved):
Awesome ergonomic improvement!

src/scene/graphics/shared/GraphicsContext.ts
```diff
@@ -81,7 +81,7 @@ export class GraphicsContext extends EventEmitter<{
         /** The fill pattern to use. */
         fill: null,
         /** The fill units to use. */
-        textureSpace: 'global',
+        textureSpace: 'local',
```
Zyie, line 84: This change doesn't seem like it should be here?

bigtimebuddy:
```suggestion
        textureSpace: 'global',
```

src/scene/graphics/shared/utils/__tests__/toFillStyle.test.ts
```diff
@@ -0,0 +1,139 @@
+import { Color } from '../../../../../color/Color';
+import { Texture } from '../../../../../rendering/renderers/shared/texture/Texture';
```
bigtimebuddy, line 2: Recommend using the src alias `~`
```suggestion
import { Color } from '~/color/Color';
import { Texture } from '~/rendering/renderers/shared/texture/Texture';
```

Zyie (approved):

Issue #11148, opened 2024-12-11
Feature Request: Add Texture Options to the sprite asset when loading a spritesheet


sam007mac:
### Description

There is no way (or at least, no convenient way) to set texture options on a texture when that sprite is loading as part of a spritesheet.

From the docs:
```js
// passing options to to the object
 Assets.add({
     alias: 'bunnyBooBooSmooth',
     src: 'bunny{png,webp}',
     data: { scaleMode: SCALE_MODES.NEAREST }, // Base texture options
 });
```
_As a side note, this code is deprecated as 'SCALE_MODES' has since been deprecated._

This works when loading in single images as assets. However when loading in a spritesheet these are the only "data" options:
```ts
{
     texture?: Texture;
     imageFilename?: string;
     ignoreMultiPack?: boolean;
}
```

A workaround to this issue involves loading the spritesheet image asset first with the texture options, and then adding it as the `texture` in the spritesheet's data property after the fact. This is poor because:
1. It disconnects the spritesheet image from the spritesheet json file. Normally the json file has an "image" property which links to the sprite's location so it knows where to load it, but if the image has to be loaded first then this can't be used without some method of first reading the json file to find the image location, loading the image, then loading the json file again.
2. It doesn't behave well with bundled assets. You won't be able to have a bundle load all at once if you have to use the already-loaded Texture as the "texture" property of the json file's data property.

An alternative workaround is to manually find and set the values in the TextureSource after it has already loaded but this isn't required when loading in normal images so it definitely feels like forcing a workaround.

There are a couple potential solutions, the best would be adding a ` { textureOptions?: string } ` property to the loadJSON data that is added to the image when it loads.

GoodBoyDigital:
this is a great idea! Do you have the bandwitch to submit a PR? Thanks in advance!

mayakwd:
It appears to be fixed already.

Zyie:
> It appears to be fixed already.

Ahh good catch! 
Closing this issue

PR #11163, opened 2024-12-17
feat: added Texture Options to spritesheet loader

sam007mac:
##### Description of change

I've added the ability to set TextureOptions on sprites loaded via a spritesheet, as outlined in this Feature Request: #11148 

##### Pre-Merge Checklist

I was having difficulty with the test suite but I tested my changes locally and it functioned as intended with normal and multipacked spritesheets.


src/spritesheet/spritesheetAsset.ts
```diff
@@ -148,7 +158,16 @@ export const spritesheetAsset = {
             {
                 const imagePath = copySearchParams(basePath + (imageFilename ?? asset.meta.image), options.src);
 
-                const assets = await loader.load<Texture>([imagePath]);
+                let assets;
+
+                if (textureOptions)
```
GoodBoyDigital, line 163: I don't think we even need the `if (textureOptions)` check here as textureOptions can be null:

replacing the original line
`assets = await loader.load<Texture>([imagePath]);`

with
`assets = await loader.load<Texture>([{ src: imagePath, data: textureOptions}]);`

should work!

GoodBoyDigital (commented):
thanks so much for this @sam007mac ! tiny tweak requested!

bigtimebuddy (commented):
A test here would be useful to make sure the options are actually set.

src/spritesheet/spritesheetAsset.ts
sam007mac: It looks like that check is still needed, removing it caused the SVG spritesheet test to fail. 
I could modify the SVG script to fix that issue but I don't know if that's in the scope of this PR, what are your thoughts?

sam007mac:
I've added a test for checking the texture options.
I've also fixed the issue causing loadSVG to fail because of my change but I don't know if that should be under a separate PR or issue report, it's only partly related.

bigtimebuddy (approved):
Good fix! Thank you for the test.

GoodBoyDigital (approved):
thanks @sam007mac 👍

PR #11170, opened 2024-12-20
feat: Add GIF decoding and playback 

bigtimebuddy:
This change imports https://github.com/pixijs-userland/gif (and git history) as an opt-in package `pixi.js/gif` that can be used to parse and playback gif animations.

## Example

https://jsfiddle.net/bigtimebuddy/ejcuoL20/

## Usage Example

```ts
import 'pixi.js/gif';

const gif = await Assets.load('./path/to/file.gif');
```


bigtimebuddy:
* I'm not opposed to generating individual Textures as load time if you think the added memory cost outweighs than the performance of a single canvas texture.
* The reason I chose not to use AnimatedSprite is because it doesn't support unique frame lengths, which is a property supported by GIF that I'm supporting (see FrameObject interface). I think we'd need some more flexibility in AnimatedSprite to accomplish this time-based playback rather than frame/tick-based playback.

GoodBoyDigital:
> * I'm not opposed to generating individual Textures as load time if you think the added memory cost outweighs than the performance of a single canvas texture.
> * The reason I chose not to use AnimatedSprite is because it doesn't support unique frame lengths, which is a property supported by GIF that I'm supporting (see FrameObject interface). I think we'd need some more flexibility in AnimatedSprite to accomplish this time-based playback rather than frame/tick-based playback.

RE second point, i agree - would need some more changes.. not entierly sure it would be worth it!

thinking about the first point a little more. Would it be better to treat the loader as returning a 'gifData' (im sure there is a better name!) essentially a resource that can then be handled by the GifSprite. A bit like a sprite sheet?

```ts
const gifData = await Assets.load('./path/to/file.gif');

// create a scene item like:
const gifOne = new GifSprite(gifData);

// or like this:
const gifTwo = GifSprite.from('./path/to/file.gif');
``` 

I know it would be an extra line to create, but i think it follows the pixi paradigm a bit better (load a resource - share it across scene elements) this would make it work in a similar way to how sprites do, plus give users access to the data object to do more advanced things if they want eg:

```ts
// or mess with the textures and do somthing fancy!
const textures = gifData.textures;

for(var i = 0; i < textures.length; i++)
{
    const sprite = new Sprite(textures[i]);
    
    stage.addChild(sprite);

    sprite.x = i * 100;
}

**Comment by GoodBoyDigital** [comment]
> I like that design a lot more. Will work on tweaking this.

awesome! Thanks Matt!

**Comment by bigtimebuddy** [comment]
@GoodBoyDigital made those changes:

```ts
import { AnimatedGIF } from 'pixi.js/gif';

const source = await Assets.load('./path/to/file.gif');

// Source supports multiple instances, like Textures
const gif1 = new AnimatedGIF(source);
const gif2 = new AnimatedGIF(source);

```

bigtimebuddy:
Here are the current API names, which I'm not in love with:
* AnimatedGIF
* AnimatedGIFSource
* AnimatedGIFAsset

Thinking about an alt:
* GIFAnimation or GIFSprite
* GIFTimeline
* GIFAsset

GoodBoyDigital:
> Here are the current API names, which I'm not in love with:
> 
> * AnimatedGIF
> * AnimatedGIFSource
> * AnimatedGIFAsset
> 
> Thinking about an alt:
> 
> * GifAnimation or GifSprite
> * GifTimeline
> * GifAsset

I really like the new ones! The 'Animated' part is not really giving us more info as a users will know gifs are animated :D
I like: 

- GifSprite
- GifAsset 
- GifSource

src/gif/AnimatedGIFSource.ts
```diff
@@ -0,0 +1,215 @@
...
+    /** Animation frames */
+    public readonly frames: AnimatedGIFFrame[];
+
+    /** Textures */
+    public readonly textures: Texture<CanvasSource>[];
+
+    /** Total number of frames in the animation */
+    public readonly totalFrames: number;
+
+    constructor(width: number, height: number, frames: AnimatedGIFFrame[])
```
GoodBoyDigital, line 56: small one - wander if witdh and height could be optional? As we can get the first frame?
actually.. do we need width and height any more?

GoodBoyDigital (approved):

src/gif/AnimatedGIFSource.ts
bigtimebuddy: Yeah, good catch. Removed width/height args

bigtimebuddy:
Renamed to these:

- GifSprite
- GifAsset
- GifSource

Zyie (approved):

PR #11182, opened 2025-01-07
chore: replace updateQuadBounds with updateTextBounds in text

GoodBoyDigital:
##### Description of change
- replaced updateQuadBounds with updateTextBounds in text renderering pipes
- Updated CanvasTextPipe and HTMLTextPipe to use the new updateTextBounds function for better handling of text bounds.
- Removed padding parameter from updateQuadBounds calls, simplifying the bounds update logic.
- Introduced updateTextBounds function to manage text-specific bounds adjustments, including padding compensation.
- Refactored Sprite visualBounds getter to remove unnecessary padding argument in updateQuadBounds.

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/text/utils/updateTextBounds.ts
+++ b/src/scene/text/utils/updateTextBounds.ts
@@ -1,0 +1,29 @@
+import { updateQuadBounds } from '../../../utils/data/updateQuadBounds';
+import { type BatchableSprite } from '../../sprite/BatchableSprite';
+import { type AbstractText } from '../AbstractText';
+
+/**
+ * Updates the bounds of the given batchable sprite based on the provided text object.
+ *
+ * This function adjusts the bounds of the batchable sprite to match the dimensions
+ * and anchor point of the text's texture. Additionally, it compensates for any padding
+ * specified in the text's style to ensure the text is rendered correctly on screen.
+ * @param {BatchableSprite} batchableSprite - The sprite whose bounds need to be updated.
+ * @param {AbstractText} text - The text object containing the texture and style information.
+ */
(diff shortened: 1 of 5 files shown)


src/utils/__tests__/updateQuadBounds.test.ts
```diff
@@ -0,0 +1,80 @@
+import { type ObservablePoint } from '../../maths/point/ObservablePoint';
+import { type Texture } from '../../rendering/renderers/shared/texture/Texture';
```
bigtimebuddy, line 2:
```suggestion
import { type ObservablePoint } from '~/maths';
import { type Texture } from '~/rendering';
```

src/scene/text/utils/updateTextBounds.ts
```diff
@@ -0,0 +1,33 @@
...
+    // the texture itself. However, we don't want this padding to affect the text's
+    // actual position on screen.
+    // To compensate, we shift the render position back by the padding amount,
+    // ensuring the text appears exactly where intended while maintaining the
+    // buffer zone around it.
+
+    bounds.minX -= padding;
+    bounds.minY -= padding;
+    bounds.maxX -= padding;
+    bounds.maxY -= padding;
```
bigtimebuddy, line 32: Should this be?

```ts
bounds.maxX += padding;
bounds.maxY += padding;
```

GoodBoyDigital: good question! We are applying an offset to the bounds rather than padding the bounds here, so this this is all good! It gets confusing as the padding is the offset value we need! ( to account for the extra padding in the texture)

Zyie (approved):

PR #11183, opened 2025-01-07
chore: ignore type errors when watching

Zyie:
##### Description of change

* Added a new `build:silent` script that uses `tsc-silent` to suppress all errors during the TypeScript compilation process.
* Updated the `watch` and `watch:lib` scripts to use the new `build:silent` script instead of the previous build scripts.

This should stop the watch command from crashing when you are developing


package.json
```diff
@@ -24,17 +24,16 @@
     "clean:uploads": "rimraf .pr_uploads .s3_uploads --glob",
     "prebuild": "npm run clean",
     "build": "run-s build:pkg build:rollup build:tsc build:dts build:dts-bundle",
+    "build:silent": "tsc-silent -p tsconfig.types.json --suppress @ && run-s build:pkg build:rollup build:dts",
```
bigtimebuddy, line 27: I find these scripts to be a little complicated. Can I suggest the following instead as it's a little simpler doesn't involve changing the watch commands...

```json
{
  "build:tsc:silent": "tsc-silent -p tsconfig.types.json --suppress",
  "watch:build": "run-s build:tsc:silent build:pkg build:rollup build:dts"
}
```

bigtimebuddy (approved):

GoodBoyDigital (approved):

PR #11209, opened 2025-01-17
fix: DDS BC format need Alignment Blocks

justjuangui:
Each dimension must be aligned to a multiple of 4 for BC formats. If a dimension isn’t a multiple of 4, round up to the nearest multiple of 4

##### Description of change
Each dimension must be aligned to a multiple of 4 for BC formats. If a dimension isn’t a multiple of 4, round up to the nearest multiple of 4.
image

Current calculated values in dev
image

Renderer only show the dds image in maximum size, but using any of mipmap no work because the warning
File used:
[acolyteofchayulaascendancy.zip](https://github.com/user-attachments/files/18455270/acolyteofchayulaascendancy.zip)


GoodBoyDigital (approved):
nice catch!

PR #11211, opened 2025-01-19
fix: loadTextures.load add onerror for Image case

midiusRed:
Added error handling when using Image to upload images. Currently, in this mode, Promise will never be allowed (for example, in the case of an incorrect url)

### DIFF
--- a/src/assets/loader/parsers/textures/loadTextures.ts
+++ b/src/assets/loader/parsers/textures/loadTextures.ts
@@ -147,8 +147,9 @@ export const loadTextures: LoaderParser<Texture, TextureSourceOptions, LoadTextu
                 }
                 else
                 {
-                    src.onload = (): void =>
+                    src.onerror = src.onload = (): void =>
                     {
+                        src.onerror = src.onload = null;
                         resolve(src);
                     };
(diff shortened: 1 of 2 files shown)


src/assets/loader/parsers/textures/loadTextures.ts
```diff
@@ -147,8 +147,9 @@ export const loadTextures: LoaderParser<Texture, TextureSourceOptions, LoadTextu
                 }
                 else
                 {
-                    src.onload = (): void =>
+                    src.onerror = src.onload = (): void =>
```
bigtimebuddy, line 150: Wouldn't we want to reject onerror instead of resolve?

bigtimebuddy (commented):
Thanks for this. Would be useful to get a unit test too.

src/assets/loader/parsers/textures/loadTextures.ts
midiusRed: Yes, you're right, it's the right thing to do to reject.

bigtimebuddy (approved):
Good one! Thanks for the test.

PR #11214, opened 2025-01-20
feat: update Earcut to v3.0.0

LukeAbby:
I want to update earcut to v3.0.0 because I have a project that makes their own `earcut` global that conflicts with the 2.0.0 era global `earcut` variable with TypeScript typings.

Wrangling with Jest unfortunately took most of the time because apparently its support for node module dependencies written in ESM is lacking. If anyone knows of a better way I'm all ears!

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+    plugins: ['@babel/plugin-transform-modules-commonjs'],
+};
--- a/jest.config.js
+++ b/jest.config.js
@@ -13,7 +13,9 @@ module.exports = {
         '\\.vert$': 'jest-raw-loader',
         '\\.frag$': 'jest-raw-loader',
         '\\.wgsl$': 'jest-raw-loader',
+        '\\.js$': 'babel-jest'
     },
+    transformIgnorePatterns: ['/node_modules/(?!earcut|@types/earcut)'],
     moduleNameMapper: {
         '^worker:(.*)$': '$1',
         '^~/(.*)$': '<rootDir>/src/$1',
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,8 @@
         "parse-svg-path": "^0.1.2"
       },
       "devDependencies": {
+        "@babel/core": "7.22",
+        "@babel/plugin-transform-modules-commonjs": "^7.26.3",
         "@pixi/eslint-config": "^6.0.0",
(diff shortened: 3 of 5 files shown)


bigtimebuddy:
What are the big difference between v2 and v3 of earcut?

LukeAbby:
There aren't many. Here's the differences copy and pasted from the release notes:
- Fix edge cases related to touching inner rings. https://github.com/mapbox/earcut/pull/177
- Slightly improve performance. https://github.com/mapbox/earcut/pull/177
- ⚠️ Breaking: publish earcut as a ES module, dropping support for CommonJS (use v2.2.4 if you need compatibility).
- ⚠️ Breaking: use modern ES syntax, dropping support for IE11 (you can still transpile on your end to keep support).
- ⚠️ Breaking: the legacy UMD browser bundle now exposes the main function as earcut.default.

Granted there's also the entirety of 2.1.5 to 2.2.4 that don't have any release notes at all. The only JS file is less than 700 lines long so if you want to look at [2.1.4](https://github.com/mapbox/earcut/blob/dcc22eb30bb02938d1e36f7ececcd066b5baf650/src/earcut.js) compared to [3.0.1](https://github.com/mapbox/earcut/blob/4814c8c97ee3965d03e2c7e5f68a0dc4700458c1/src/earcut.js) you can!

I was really thinking this'd be a tiny bump in the dependencies, the fact that Jest was so annoying to grapple with was an unexpected development. I just hope it won't be a blocker to getting this merged, if it is I really would like to work with you in any way to work through this.

bigtimebuddy (approved):

LukeAbby:
@Zyie would you mind taking a look at if this PR can be merged? Hopefully not being impatient, just want to make sure it hasn't slipped under your radar.

mourner:
> Granted there's also the entirety of 2.1.5 to 2.2.4 that don't have any release notes at all.

They do, they were just kept in a different place (I moved them from being kept in the readme to using GitHub releases): https://github.com/mapbox/earcut/tree/v2.2.4#changelog

Zyie:
Hey @LukeAbby,

Thanks for this! I’m planning to merge it in for the next minor release. While it looks safe, there’s a small chance it could impact users, so I’d prefer to include it in 8.10.0 rather than a patch. A minor version bump signals a bit more strongly that there’s a change they might want to be aware of.

LukeAbby:
@mourner Good to know! And thanks for the library.

PR #11217, opened 2025-01-21
fix: events firing PointerEvent when not supported

midiusRed:
1) Emulating the absence of a PointerEvent:
window.PointerEvent = null;

2) We created the sprite in eventMode = 'dynamic'

3) We get an exception, because the Event Timer creates a PointerEvent


bigtimebuddy (approved):
Good catch. Also, a test for this would be great!

PR #11226, opened 2025-01-23
fix: gif export

Zyie:
We are currently not exporting the new gif classes and this PR allows for them to be imported like this:

```ts
import { GifSprite } from 'pixi.js/gif' // asset extension is added automatically 

// use gif
```


bigtimebuddy (approved):
I think this is my preference

PR #11234, opened 2025-01-24
match uv size if non is provided

### LINKED ISSUE #11207: Bug: Mesh breaks when changing geometry size around 200 vertices
### Current Behavior

This is a weird bug, and apologies in advance if I'm doing something I shouldn't be.

When a `PIXI.Mesh` is created, is seems that the vertex buffer size is locked in.  
If the `mesh.geometry` is replaced later, and the new geometry differs in size, it can trigger a crash in the MeshPipe code, 
or "GL_INVALID_OPERATION: Vertex buffer is not big enough for the draw call"

We've noticed that the issues occur if the mesh starts out with <200 verts and the later exceeds that number - or if it starts out with >200 verts and that number later drops below 200.

### Expected Behavior

I'd expect to be able to replace the `mesh.geometry` and have it work.

### Steps to Reproduce

Here is a playground example that demonstrates the issue.  
Click "One More" to add a triangle, the vert count will exceed 200 and the mesh will break.
Click "One Less" to bring the count back below 200 and get the mesh back.

https://www.pixiplayground.com/#/edit/go_kw5sXDdM6s0T0hwCbj

### Environment

- **`pixi.js` version**:  v8.any

### Possible Solution

In our code, we are working around the issue by just creating a `new Mesh()` if we detect the vertex count changing in this way.

### Additional Information

Related: https://github.com/facebook/Rapid/issues/1636

GoodBoyDigital:
##### Description of change

Fixes a little issue, where if only positions are provided when creating a `MeshGeometry`  the uvs are an incorrect size and the geometry does not render correctly.

fixes #11207 

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/mesh/shared/MeshGeometry.ts
+++ b/src/scene/mesh/shared/MeshGeometry.ts
@@ -14,7 +14,7 @@ export interface MeshGeometryOptions
 {
     /** The positions of the mesh. */
     positions?: Float32Array;
-    /** The UVs of the mesh. */
+    /** The UVs of the mesh. If not provided, they will be filled with 0 and match the size of the positions. */
     uvs?: Float32Array;
     /** The indices of the mesh. */
     indices?: Uint32Array;
--- a/src/scene/mesh/shared/MeshGeometry.ts
+++ b/src/scene/mesh/shared/MeshGeometry.ts
@@ -63,7 +63,21 @@ export class MeshGeometry extends Geometry
         options = { ...MeshGeometry.defaultOptions, ...options };
 
         const positions = options.positions || new Float32Array([0, 0, 1, 0, 1, 1, 0, 1]);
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks for digging in and fixing this! 
Do similar changes need to go into the setters also, to make sure that the buffer lengths match up?

I think I can make sure going forward to provide some uvs, just in case.  
(The original mesh was being used for a mask anyway, so this wasn't something I was thinking about).

GoodBoyDigital:
> Thanks for digging in and fixing this! Do similar changes need to go into the setters also, to make sure that the buffer lengths match up?
> 
> I think I can make sure going forward to provide some uvs, just in case. (The original mesh was being used for a mask anyway, so this wasn't something I was thinking about).

I think im going to leave this as is for now. It will start to get really bloated and with all the validation code! I'll update the docs instead!

Zyie (approved):

PR #11245, opened 2025-01-29
fix: increase didViewChangeTick on removeChildren call

dapi303:
##### Description of change
When elements are removed from the container by the removeChildren function,
_didViewChangeTick should be updated to drop boundary cache on next call.

##### Pre-Merge Checklist


GoodBoyDigital (approved):
nice catch! appreciate the tests too 👍

PR #11246, opened 2025-01-29
fix: texture space incorrectly set to global

Zyie:
This PR updates the default texture space to be `local` instead of `global`. This makes it much easier to use fills and gradients as the coordinates are normalised.

To get the old behaviour all API's accept `textureSpace: 'global'`, so the update path is simple.


bigtimebuddy (approved):

GoodBoyDigital (approved):
nice!

PR #11253, opened 2025-02-04
feat: add mixins for leaf nodes

Zyie:
##### Description of change

This PR expands upon the mixin API for containers and adds support for using mixins on leaf nodes.

While developing an extension for pixi i found the need to apply specific properties to individual leaf nodes and this PR allows for that to happen much more seamlessly than putting doc comments stating it only work for xyz and using `ContainerOptions`.

I also found that `Sprite.mixin` would apply properties to the `Container` which meant all leaf nodes would get the mixin. This has been fixed by overriding the mixin function per leaf node

##### Pre-Merge Checklist


bigtimebuddy:
I love this idea.

What do you think about a general util instead of these separate static functions? Maybe something on extensions?

```ts
import { extensions } from "pixi.js";

extensions.mixin(Sprite, {})
```

Also is there a better type than Dict<any>.

Zyie:
> What do you think about a general util instead of these separate static functions? Maybe something on extensions?
> 
> ```ts
> import { extensions } from "pixi.js";
> 
> extensions.mixin(Sprite, {})
> ```

Yeah this is great! I will update

> Also is there a better type than `Dict<any>`?

I can look into this

src/scene/container/Container.ts
```diff
@@ -1395,15 +1400,15 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
...
-Container.mixin(onRenderMixin);
-Container.mixin(measureMixin);
-Container.mixin(effectsMixin);
-Container.mixin(findMixin);
-Container.mixin(sortMixin);
-Container.mixin(cullingMixin);
-Container.mixin(cacheAsTextureMixin);
-Container.mixin(getGlobalMixin);
-Container.mixin(collectRenderablesMixin);
+extensions.mixin(Container, childrenHelperMixin);
```
bigtimebuddy, line 1403: What if the args for mixin uses a rest args so you could do this (Target, ...sources[])? It would make it less verbose here.

bigtimebuddy (approved):
Great work! Thanks for adding tests.

PR #11263, opened 2025-02-06
feat: Texture 2D Array Support

justjuangui:
##### Description of change
DDS have support for texture array (in DX10 / 11) like layers [Reference](https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dx-graphics-dds-pguide#using-texture-arrays-in-direct3d-1011)

This PR enable support for this kind of resource in Windows PC and MAC that have support for texture compress BC (not supported in mobiles).

This introduce new components to keep isolated the changes from the new feature:
- Sprite2DArray: This component support the new attribute "LayerId" and work in batchable mode
- Default2DArrayBatcher & Batch2DArrayGeometry: This components are responsible to send the "LayerId" in the vertex attribute in batch mode
- Default2DArrayShader: This components is aware of the new attribute '"LayerId", use the new shaders that replace texture_2d to texture_2d_array

Enhanced Texture and texture options to contain the attribute "viewDimension" to keep agnostic between renderes.

This PR support for both WebGL2.0 and WEBGPU

The test case is [HERE](https://justjuangui.github.io/)
and the source code is [https://github.com/justjuangui/justjuangui.github.io](url)


GoodBoyDigital:
this is cool!

bigtimebuddy:
Epic PR! Thanks so much for putting this together. It seems like you have an excellent grasp on both WebGL and WebGPU as well as Pixi to pull this off.

I have some concerns however. First, it's a lot of code, especially for something that many users may not use. Can you describe the benefits and tradeoffs of using 2D Array textures? Why would a developer consider this feature? We are very mindful about bloat in the project and adding high-value features, if you could speak to that might give us more context.

We are trying to build Pixi for a general developer audience and that sometimes means not adding everything, making some functionality opt-in and even externalizing some functionality as plugins. My initial instinct was to do the latter: minimally support 2D Arrays in Pixi but the bulk of the code (Sprite2DArray/Pipe) should live in an external project. Open to other ideas.

justjuangui:
> Epic PR! Thanks so much for putting this together. It seems like you have an excellent grasp on both WebGL and WebGPU as well as Pixi to pull this off.
> 
> I have some concerns however. First, it's a lot of code, especially for something that many users may not use. Can you describe the benefits and tradeoffs of using 2D Array textures? Why would a developer consider this feature? We are very mindful about bloat in the project and adding high-value features, if you could speak to that might give us more context.
> 
> We are trying to build Pixi for a general developer audience and that sometimes means not adding everything, making some functionality opt-in and even externalizing some functionality as plugins. My initial instinct was to do the latter: minimally support 2D Arrays in Pixi but the bulk of the code (Sprite2DArray/Pipe) should live in an external project. Open to other ideas.

First at all thx to review, validate and take some time to give me some advice about. This is my first time in Graphic world,I've only learned the basics this year with [PoB Project ](https://github.com/PathOfBuildingCommunity/PathOfBuilding-PoE2). So im not super expert about it :).

Like you said this PR is very specific use case, due that this only work with DDS being said that only windows computer machine and some OSX machine will be able to use. In my mind i was targeting Embbed Desktop apps that use Pixijs for graphic side and want to use DDS.

I tried at the beggining to build like plugin but some code was need to add in the heart of code. I will try to remove things that can be builded with plugin system but for another things lke pipe (src/rendering/renderers/shared/system/SharedSystems.ts) i was forced to hardcoded and create the pipe, I didnt see how to "add" with plugin system, maybe i was blind at that time.

If you think this can be useful for future and comunity i will spent time to decouple the code.

Thx in advance

GoodBoyDigital:
closing this one for now - as we have fresh PR solving for this!

Issue #11244, opened 2025-01-29
Bug: HTMLText inside RenderGroup gets GC'd incorrectly


lightspeedlucas:
### Current Behavior

An `HTMLText` instance inside a `Container` that was set up with `isRenderGroup: true` disappears after a while.

### Expected Behavior

It should remain visible regardless of its parent's render group status.

### Steps to Reproduce

[Playground](https://pixiplayground.com/#/edit/l7BGZ8rvM-EnsnOXxL8J0)

```js
// Force collection after 2 seconds so it's quickly reproducible
PIXI.RenderableGCSystem.defaultOptions.renderableGCMaxUnusedTime = 2000;
PIXI.RenderableGCSystem.defaultOptions.renderableGCFrequency = 100;

const app = new PIXI.Application();
await app.init({ resizeTo: window });
document.body.appendChild(app.canvas);

new PIXI.Container({
isRenderGroup: true, // Commenting this line avoids the issue
parent: app.stage,
children: [
    new PIXI.HTMLText({
    text: 'Hello world! This will disappear after 2 seconds.',
    style: { fontSize: 16, fill: 'white' },
    }),
]
});
```

### Environment

- **`pixi.js` version**: 8.7.3
- **Running Example**: [Playground](https://pixiplayground.com/#/edit/l7BGZ8rvM-EnsnOXxL8J0)

### Possible Solution

`RenderableGCSystem` compares `renderGroup.gcTick` to `renderGroup.instructionSet.gcTick`. The latter seems to be updated correctly, but the former is only updated for `Application.stage`. Not sure if it should be updated for every render group or if we should change that comparison to always use the stage's `gcTick` instead.

I've worked around it locally by replacing `_updateInstructionGCTick`'s implementation to update both `gcTick` values.

Zyie:
This should be fixed by #11346

PR #11268, opened 2025-02-07
fix: html text being removed by GC

Zyie:
fixes: #11244

This PR fixes two issue, one being that if you made a new render group the `gcTick` was never being updated which meant the renderables inside of it would be constantly tidied up and the second issue is we now make sure that we regenerate the html text texture after it has been tidied up by the renderable gc.


Zyie:
closing as this has been fixed in #11346

PR #11269, opened 2025-02-07
Update CanvasTextSystem.ts

GoodBoyDigital:
##### Description of change

Small use case - 
```const texture = renderer.canvasText.getTexture(...)` <-- gets a text texture

I added a complimentry

```renderer.canvasText.returnTexture(texture)```

Devs could destroy the texture, but we all like pooling here :D

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/text/canvas/CanvasTextSystem.ts
+++ b/src/scene/text/canvas/CanvasTextSystem.ts
@@ -173,6 +173,17 @@ export class CanvasTextSystem implements System
         this._activeTextures[textKey].usageCount++;
     }
 
+    public returnTexture(texture: Texture)
+    {
+        const source = texture.source;
+
+        source.resource = null;
+        source.uploadMethodId = 'unknown';
+        source.alphaMode = 'no-premultiply-alpha';
+
+        TexturePool.returnTexture(texture);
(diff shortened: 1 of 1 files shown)


bigtimebuddy (commented):
Not huge fan of `returnTexture` as a name. I think `return` as in "return a library book" is challenging with `return` as something a function does. 

I know it's already part of TexturePool, but I think we want something that's more like the opposite of `getTexture`, like `putTexture`? Or something more descriptive like `poolTexture`?

GoodBoyDigital:
> Not huge fan of `returnTexture` as a name. I think `return` as in "return a library book" is challenging with `return` as something a function does.
> 
> I know it's already part of TexturePool, but I think we want something that's more like the opposite of `getTexture`, like `putTexture`? Or something more descriptive like `poolTexture`?

I don't disagree... :D was thinking more about constancy here as I use `return` generally across Pixi for these kinds of things. Perhaps we can change these all together in another PR (with depreciation?)

bigtimebuddy:
Yeah, I'm okay with doing that as a followup PR

Zyie (approved):

PR #11272, opened 2025-02-11
chore: Rename "returnTexture" to "addTexture"

bigtimebuddy:
Follow-up to #11269 

### Changed

* Renames `returnTexture` to `addTexture` in `CanvasTextSystem`
* Renames `returnTexture` to `addTexture` in `TexturePool`

### Rationale

The method name `returnTexture` is somewhat confusing because the method doesn't actually `return` any value. The verb here means "return this to the pool" but is a little overload in the context of a JavaScript function. A better approach is `addTexture`, as it "add this to the pool". We use this verb heavily in PixiJS already e.g., "addChild", "addGroup", "addAttribute", "addManifest", etc.

### DIFF
--- a/src/filters/FilterSystem.ts
+++ b/src/filters/FilterSystem.ts
@@ -376,7 +376,7 @@ export class FilterSystem implements System
             filters[0].apply(this, inputTexture, filterData.previousRenderSurface, false);
 
             // return the texture to the pool so we can reuse the next frame
-            TexturePool.returnTexture(inputTexture);
+            TexturePool.addTexture(inputTexture);
         }
         else
         {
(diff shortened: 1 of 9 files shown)


bigtimebuddy:
In doing some searching, the prior art I used for this decision was Panda3d TexturePool class which also uses `addTexture`:
https://docs.panda3d.org/1.10/python/reference/panda3d.core.TexturePool

GoodBoyDigital (approved):
nice! Thanks Matt!

ivanpopelyshev (changes requested):
Usuall addXXX() is the beginning of work, here its the end of work. I use this method often, amd with "add" all my code stops being readable.

Previously, we had problems with "point1.copy(point2)" and we added methods like "copyFrom" and "copyTo" instead.

Better to make more clarification, like "returnTextureToPool()"

bigtimebuddy:
Other option I considered was "put" as something opposite with "get". Since these are often use in the same lifecycle.

- TexturePool.getTexture
- TexturePool.putTexture

Some other ideas I considered.

- remitTexture
- sendBackTexture
- storeTexture

ivanpopelyshev (approved):
All three options are fine, "put" looks good

lunarraid:
Just throwing my two cents in, but for pools I've always used 'obtain' and 'release'. Is this a public API, though? Will it break backwards compatibility on things just for a name change because we don't like it?

ivanpopelyshev:
> Just throwing my two cents in, but for pools I've always used 'obtain' and 'release'. Is this a public API, though? Will it break backwards compatibility on things just for a name change because we don't like it?

deprecation will be there, so it shouldn't be a problem

PR #11282, opened 2025-02-14
feat: textStyle filters

GoodBoyDigital:
##### Description of change

Ever want to bake a filter into a Text at creation time..? Yes? Well me too! This new feature allows filters to part a text style!

```
const blurFilter = new BlurFilter();

const text = new Text({
        text: 'HI',
        style: {
            fontSize: 100,
            fill: 'white',
            filters: [blurFilter], // add some filters to the style
        }
    });
```

When setting filters this way, the filter is actually baked into the texture at creation time, rather than having to be calculated at run time, eliminating the runtime cost.

This unlocks the power of filters to make text look extra awesome, but without the performance overhead of dynamically applying the filter each frame. For example you could use a nice pixi outline filter along with a nice soft dropshadow filter rather than being limited by current text style options.

Additionally this gives us access to a new feature on the `filterSystem` called `generateFilteredTexture` that  takes a texture and a list of filters, applies the filters to the texture and returns the resulting texture.

```
const blurFilter = new BlurFilter();

myTextureWithFilters = renderer.filters.generateFilteredTexture({texture, filters:[blurFilter]});
```

This did require a bit of filter refactoring, but the tests made sure that everything still looks and works well. This has been tested on our games which makes extensive use of filters!

##### Pre-Merge Checklist


src/scene/text/TextStyle.ts
```diff
@@ -130,6 +131,13 @@ export interface TextStyleOptions
     wordWrap?: boolean;
     /** The width at which text will wrap, it needs wordWrap to be set to true */
     wordWrapWidth?: number;
+    /**
+     * An optional filter or array of filters to apply to the text, allowing for advanced visual effects.
+     * These filters will be applied to the text as it is created, resulting in faster rendering for static text
+     * compared to applying the filter directly to the text object (which would be applied at run time).
+     * @default null
```
Zyie, line 138: The default value would be `undefined` right?

src/rendering/renderers/gl/renderTarget/GlRenderTargetAdaptor.ts
```diff
@@ -165,7 +165,7 @@ export class GlRenderTargetAdaptor implements RenderTargetAdaptor<GlRenderTarget
         // we are rendering to the main canvas..
         const colorTexture = renderTarget.colorTexture;
 
-        if (CanvasSource.test(colorTexture.resource))
+        if (colorTexture instanceof CanvasSource)
```
Zyie, line 168: What was the purpose of this change? I know we have talked about it before but I can't remember it properly

src/filters/FilterSystem.ts
```diff
@@ -68,17 +68,83 @@ export interface FilterInstruction extends Instruction
...
+
+    /**
+     * The texture to which the filter is applied.
+     * @type {Texture}
+     */
+    public inputTexture: Texture = null;
+
+    /**
+     * The back texture used for blending, if required.
+     * @type {Texture | undefined}
```
Zyie, line 91:
```suggestion
     * @type {Texture | null}
```

src/filters/FilterSystem.ts
```diff
@@ -68,17 +68,83 @@ export interface FilterInstruction extends Instruction
...
+
+    /**
+     * The render surface where the output of the filter is rendered.
+     * @type {RenderSurface}
+     */
+    public outputRenderSurface: RenderSurface = null;
+
+    /**
+     * The offset of the output render surface.
+     * @type {{ x: number, y: number }}
```
Zyie, line 127: We should use the point data type for x/y values to keep it consistent with the rest of pixi

```suggestion
     * @type {PointData}
```

src/filters/FilterSystem.ts
```diff
@@ -68,17 +68,83 @@ export interface FilterInstruction extends Instruction
...
+     */
+    public outputOffset = { x: 0, y: 0 };
```
Zyie, line 129: We should use the point data type for x/y values to keep it consistent with the rest of pixi

```suggestion
     * @type {PointData}
     */
    public outputOffset: PointData = { x: 0, y: 0 };
```

src/filters/FilterSystem.ts
```diff
@@ -321,101 +327,151 @@ export class FilterSystem implements System
...
         }
 
-        this._activeFilterData = filterData;
+        const globalResolution = rootResolution;
+        const offsetX = 0;
+        const offsetY = 0;
 
-        const inputTexture = filterData.inputTexture;
+        filterData.outputOffset.x = -bounds.minX;// - offsetX;
+        filterData.outputOffset.y = -bounds.minY;// - offsetY;
```
Zyie, line 401:
```suggestion
        filterData.outputOffset.x = -bounds.minX;
        filterData.outputOffset.y = -bounds.minY;
```

src/filters/FilterSystem.ts
```diff
@@ -660,4 +666,226 @@ export class FilterSystem implements System
...
+        // set on our globalUniform Bind Group
+
+        // update the resources on the bind group...
+        this._globalFilterBindGroup.setResource(inputTexture.source.style, 2);
+        this._globalFilterBindGroup.setResource(filterData.backTexture.source, 3);
+
+        if (filters.length === 1)
+        {
+            // render a single filter...
+            // this.applyFilter(filters[0], inputTexture, filterData.previousRenderSurface, false);
```
Zyie, line 689:
```suggestion
```

Zyie (commented):
With such a large change to the filter system we should probably just verify this doesn't break any of the filters for [pixi filters](https://github.com/pixijs/filters/tree/main)

src/scene/text/TextStyle.ts
```diff
@@ -527,9 +544,31 @@ export class TextStyle extends EventEmitter<{
             whiteSpace: this.whiteSpace,
             wordWrap: this.wordWrap,
             wordWrapWidth: this.wordWrapWidth,
+            filters: this._filters,
```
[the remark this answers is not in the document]
GoodBoyDigital: yes! will destructure

src/scene/text/TextStyle.ts
GoodBoyDigital: yes!

src/rendering/renderers/gl/renderTarget/GlRenderTargetAdaptor.ts
GoodBoyDigital: good questions - this change allows all canvas sources to have framebuffers attached to them (not just the ones we use for rendering to the screen) . In this instance it means we cna directly modify the text canvas texture, without needing to duplicate or make a new one

src/scene/text/canvas/CanvasTextSystem.ts
```diff
@@ -218,10 +234,17 @@ export class CanvasTextSystem implements System
...
      * @param canvasAndContext
      */
-    public renderTextToCanvas(text: string, style: TextStyle, resolution: number, canvasAndContext: CanvasAndContext): void
+    public renderTextToCanvas(
+        text: string,
+        style: TextStyle,
+        padding: number,
+        resolution: number,
+        canvasAndContext: CanvasAndContext
+    ): void
```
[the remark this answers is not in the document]
GoodBoyDigital: deprecated this instead, left the API as is as its internal

src/scene/text/TextStyle.ts
```diff
@@ -527,9 +544,31 @@ export class TextStyle extends EventEmitter<{
...
         });
     }
 
+    /**
+     * Returns the final padding for the text style, taking into account any filters applied.
+     * Used internally for correct measurements
+     * @internal
+     * @returns {number} The final padding for the text style.
+     */
+    public getFinalPadding(): number
```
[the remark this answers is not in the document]
GoodBoyDigital: sure

PR #11283, opened 2025-02-14
ShapeBuilder.build > boolean

midiusRed:
ShapeBuilder includes 2 methods: build, triangulate. Using the build Rectangle example, we see that there is a check (width >= 0 && height >= 0). In this case, the points are not added, and then triangulate (which adds undefined) is performed and BatchableGraphics is created for the discarded primitive. I suggest an option in which build decides what to do with the primitive, and if it is incorrect, false will be returned. In this case, we will not add BatchableGraphics and run this primitive in pipe.

### DIFF
--- a/src/scene/graphics/shared/buildCommands/ShapeBuildCommand.ts
+++ b/src/scene/graphics/shared/buildCommands/ShapeBuildCommand.ts
@@ -4,7 +4,7 @@ import type { ShapePrimitive } from '../../../../maths/shapes/ShapePrimitive';
 export interface ShapeBuildCommand<T extends ShapePrimitive = ShapePrimitive>
 {
     extension: ExtensionMetadataDetails
-    build(shape: T, points: number[]): void;
+    build(shape: T, points: number[]): boolean;
     triangulate(
         points: number[],
         vertices: number[],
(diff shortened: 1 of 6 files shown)


GoodBoyDigital:
ooh, yes - i like the idea here!

GoodBoyDigital:
would you be able to add some tests around this please?

midiusRed:
Added tests, finalized the code and fixed the problem with GpuContext.isBatchable when the 'batch' mode is set

GoodBoyDigital (approved):
nice work @midiusRed !

src/scene/graphics/shared/__tests__/GraphicsContextSystem.test.ts
```diff
@@ -0,0 +1,104 @@
+import { GraphicsContext } from '../GraphicsContext';
+import { GraphicsContextSystem } from '../GraphicsContextSystem';
+import { getWebGLRenderer } from '@test-utils';
+
+describe('GraphicsContextSystem', () =>
+{
+    describe('ShapeBuildCommand', () =>
+    {
+        it('no skip', async () =>
```
bigtimebuddy, line 9: Can we make more descriptive text to describe what this test does rather than `no skip`? Same for the other's below. It's difficult to know what this is testing.

PR #11286, opened 2025-02-16
Geometry improvement: Part 1. Sugar for Geometry Attributes API

ivanpopelyshev:
Those changes are used in `tesera.io` project for since june 2024, project contains heavy usage of different geometries. 

This is first part, second will be about MultiDrawBuffer and swapping buffers of geometry on-fly.

### Breaking change: All attributes require vertex format.

In practice, if format is not specified when geometry is written the first time, probability of error in code is very high, and its one of "why the object is not shown on screen" problems. By specifying format we can spawn more warnings for user if something goes wrong.

That restriction adds more code for attributes specification, however it is mitigated with following sugar.

### Specifying VertexBuffer for attributes

Interleaving geometry is most common type, so here's new way with better readability:

```
const proto_geom = new Geometry({
  vertexBuffer: new Float32Array([1, 2, 3, 4, 5, 6]),
  attributes: {
    aPosition: 'float32x2',
    extra: { buffer: new Float32Array(1), format: 'float32' },
    aTexRegion: 'float32x4'
  }
});
```

here we have two buffers, one main (vertexBuffer) and some extra. For attributes of main buffer only format is required.
This also helps to access main buffer , instead of using `geometry.buffers[0]` each time, `geometry.vertexBuffer` is accessible.

### Instancing improvement

The problem with instancing is that for every attribute we have to specify a field, but its actually the same for all attributes with same buffer.

```
new Geometry({
  instanceParams: { instanced: true }
  // ...
});
```

All attributes for main vertex buffer are automatically specified as instanced. It is not possible to use only one boolean due to following MultiDrawBuffer PR's that will need more params in the object.

### Memory optimization: attributes sharing, geometry prototype

Every time when I read geometry creation code i have that itch: specify attributes only one time , then clone all those geometries. This is not possible because all "Attribute" objects have different buffers. Here I fixed it:

```
const geom1 = new Geometry({
  attributes: {},
  proto: proto_geom,
  vertexBuffer: new Float32Array([6, 7, 8, 9, 10, 11, 12]),
});
```

We can now take whole "attributes" object from another geometry and use different the vertex/index buffers. To swap another buffer `attributes: { extra: new Buffer(...) }` can be specified.

### Fixes of actual errors

- webGPU buffer bindings were wrong. Due to heavy work on the project I forgot exact case, I just remember that its very easy to hit it.

### Possible improvements for PR

- merge new "Geometry API" test with main test
- move "ensureAttributes" to geometry itself, entirely
- add "divisor" to instanceParams
- allow addAttribute to swap buffer, if it wasnt specified in "geometry with prototype" constructor
- improve documentation of the class to specify new usages
- add PipelineSystem test that shows that previous code of caching bindings does not work


ivanpopelyshev:
Here's one snipper from my project:

```ts
static proto = new PIXI.Geometry({
    vertexBuffer: new PIXI.Buffer({ data: null, glStatic: true }),
    instanced: true,
    topology: 'triangle-strip',
    vertexPerInstance: 4,
    attributes: {
        a_position: 'float32x3',
        a_color: 'sint32',
        a_axis_hor: 'float32x3',
        a_flags: 'sint32',
        a_axis_vert: 'float32x3',
        a_element_2: 'sint32',
        a_uv: 'sint32x4',
        a_quad: {
            buffer: QuadGeometry.quadBuf,
            format: 'float32x3'
        },
        a_element: {
            buffer: TerrainGeometry17.placeholderBuf,
            format: 'sint32'
        }
    }
});
    ```

schmooky:
New geometry will ease usge of instanced geoms in gambling, both performance wise and developer experience wise   it will allow for creation of smooth graphics port to pixi v8 and achieving particles with geoms that don't clunk of the cpu too much 🔥

Asjpik:
By enforcing vertex format specifications, we're eliminating a whole class of silent rendering errors. The new API provides more intuitive, flexible buffer management, enables memory-efficient geometry cloning, and simplifies instancing – all while maintaining performance. Adopting this PR means saying goodbye to cryptic rendering issues and hello to a more robust, developer-friendly graphics implementation.

Issue #11247, opened 2025-01-31
Bug: ASTC textures are not properly restored after WebGL context loss


vforsh:
### Current Behavior

Hello,

Here is a repo with a minimal reproducible example - https://github.com/vforsh/pixi-8-astc-black-screen

When compressed textures are present on stage during WebGL context loss, the following warning occurs during context restoration:
`WebGL: INVALID_ENUM: compressedTexImage2D: invalid format`

After this warning, the textures are replaced with black rectangles. While we're using ASTC 4x4 textures (in KTX container format), I suspect this issue might affect other compressed texture formats as well.

Please let me know if you need any additional information.

### Expected Behavior

No WebGL warning on context restoration and sprites that use compressed textures are displayed the same way as they were before the context loss.

### Steps to Reproduce

1. Add a sprite that uses a compressed texture (e.g. ASTC 4x4)
2. Force a WebGL context loss via `renderer.context.forceContextLoss()`
3. Notice the `WebGL: INVALID_ENUM: compressedTexImage2D: invalid format` in the browser's console

### Environment

- **`pixi.js` version**: 8.7.3 & 7.4.2
- **Browser & Version**: latest Chrome & Safari
- **Running Example**: https://vforsh.github.io/pixi-8-astc-black-screen/

Zyie:
fixed in 8.9.0

PR #11299, opened 2025-02-21
fix: Restore gl extensions state on context restore

ddenisyuk:
##### Description of change
After restoring the context, you usually need to reactivate or reinitialize any WebGL extensions manually because they are not automatically restored.

This is related to https://github.com/pixijs/pixijs/issues/11247 and possibly another issues concerning context restoration.

##### Pre-Merge Checklist


schmooky:
That would help a lot, we face the same issues on extensions

bigtimebuddy (approved):

PR #11308, opened 2025-02-26
fix: don't use alias for `PointData` import

kosmotema:
##### Description of change

Right now, aliases are not converted to relative imports during build. This causes invalid imports to appear in the final .d.ts files. One of them is the `PointData` import in `FillGradient` (you can see it [here](https://unpkg.com/browse/pixi.js@8.8.1/lib/scene/graphics/shared/fill/FillGradient.d.ts#L3), for example).

So I replaced the alias import of `PointData` with a relative one and also made it fully type-only

##### Pre-Merge Checklist


bigtimebuddy (approved):
Thank you!

PR #11314, opened 2025-03-02
feat: optimizing the BitmapFontManager.getLayout call

midiusRed:
If you look at the logic of BitmapText (and SDF in particular), you can see the call to the layout operation of the same text several times (getBitmapTextLayout). This is a very good place to use the cache and, if possible, not create many new structures when composing text, but this is for the future.

Specifically, the current solution uses the idea of reducing the layout to one time, while we will immediately calculate the data for bounds. The cost of such a calculation is minimal. In addition, the size of the text (bounds) is the data that is needed for proper layout, and in most cases it always leads to repeated unnecessary layout. Moreover, the size is usually required before the first rendering, and in this case there is also a desire to save on calling getBitmapTextLayout again.

Additionally, we get isolation of the text layout from pipe. And the last one is the initGpuText function, which previously required a layout, without in any way marking that it has already been done.

I also changed the aesthetic code, which I just did not have - "should set width and height on the constructor"

### DIFF
--- a/src/scene/text-bitmap/BitmapText.ts
+++ b/src/scene/text-bitmap/BitmapText.ts
@@ -2,6 +2,7 @@ import { warn } from '../../utils/logging/warn';
 import { AbstractText, ensureTextOptions } from '../text/AbstractText';
 import { TextStyle } from '../text/TextStyle';
 import { BitmapFontManager } from './BitmapFontManager';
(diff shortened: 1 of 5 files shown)


GoodBoyDigital:
hey @midiusRed just following up here - closing this for now as we have another PR that will introduce caching here! thanks

PR #11316, opened 2025-03-05
fix: move maxAnisotropy property from `source.style` to `source`

JulianKarhof:
##### Description of change
Moves the maxAnisotropy property from `texture.source.style.maxAnisotropy` to `texture.source.maxAnisotropy`.

##### Pre-Merge Checklist


bigtimebuddy (commented):
Thanks for the change!

This is a breaking change as it eliminates `maxAnisotropy` on TextureStyle. I'll let others weigh-in on the scope of that breaking as I've never used this prop myself.

That said, there are some minimal things that we should do to help devs with this breaking change: 1) keep `maxAnisotropy` on TextureStyleOptions eliminate the comment, but give a `@deprecated` comment instead, and 2) keep the getter and setter but provide a deprecation warning (both in comments and runtime using `deprecation()` function) to let folks know that the location of the property has moved.

GoodBoyDigital:
Hey @JulianKarhof thanks for you.PR here!

I was thinking even more simple pass through prop like the others:

```
    get maxAnisotropy(): SCALE_MODE
    {
        return this._style. maxAnisotropy;
    }

    set maxAnisotropy(value: SCALE_MODE)
    {
        this._style. maxAnisotropy = value;
    }
```

bigtimebuddy:
That's a better approach :+1:

JulianKarhof:
Thanks for your replies, that makes a lot of sense! Just out of curiosity though, are deprecated fields on pixijs ever fully removed? In that case it would seem weird passing through the deprecated property right?

bigtimebuddy:
We remove deprecate properties at the next major release. So something deprecated now would be removed in v9.

JulianKarhof:
If i forward the properties like this, i can't really add a runtime deprecation warning to the getter/setter on `TextureStyle` because it would also be logged when the non-deprecated ones on `TextureSource` are called.

GoodBoyDigital:
hey! sorry its taken so long to get round to this. Would be good to better understand the reason we want to depricate the value on textureStyle? 

Currently this PR would break WebGPU as TextureStyle is directly injested by the WebGPU Texture samplers. Cool with keeping this new pass thgouth though!  

thanks!!

Zyie:
closing in favor of #11588

PR #11340, opened 2025-03-19
feat: DOMContainer

GoodBoyDigital:
##### Description of change

# DOM Container Support

This PR adds support for integrating DOM elements within the PixiJS scene graph, allowing HTML elements to be positioned and transformed alongside other PixiJS objects.

## Features
- New `DOMContainer` class for wrapping HTML elements
- New `DOMPipe` for managing DOM elements in the scene
- Proper transform support (position, scale, rotation, skew)
- Anchor point support
- Opacity/visibility handling
- Automatic cleanup on destroy

## Performance Considerations
- Uses CSS transforms instead of top/left for better performance
- Efficiently handles DOM element positioning using translate
- Minimal DOM operations by tracking element state

## Example Usage
```js
import 'pixi.js/dom'

// Create a DOM element
const element = document.createElement('div');
element.innerHTML = 'Hello World!';

// Create a DOM container
const domContainer = new DOMContainer({ element });

// Add it to your scene
app.stage.addChild(domContainer);

// Position and transform like any other DisplayObject
domContainer.x = 100;
domContainer.y = 100;
domContainer.rotation = 0.5;
domContainer.anchor.set(0.5);
```

## Notes
- DOM elements are rendered in a separate layer above the canvas
- Elements maintain proper stacking order and transforms
- Full support for PixiJS transform hierarchy
- Handles window resize and canvas position changes

## Open questions
- Is `DOMContainer` the best name? 
- ~@Zyie , will this PR cause issues with offscreen canvas or other things that don't have access to the dom?~
- ~should this be imported - or included by default?~

##### Pre-Merge Checklist


src/rendering/renderers/shared/system/SharedSystems.ts
```diff
@@ -38,6 +39,7 @@ export const SharedRenderPipes = [
     BlendModePipe,
     BatcherPipe,
     SpritePipe,
+    DomPipe,
```
Zyie, line 42: Since this using the DOM we need to treat this like `AccessibilitySystem` and only import it in the browser bundle

bigtimebuddy (approved):
Looks great just nit feedback.

Zyie (approved):

PR #11346, opened 2025-03-22
chore: remove renderable hashes

GoodBoyDigital:
##### Description of change
Moved though all the renderables and removed the hashes found on the render pipes. Instead this PR now stores  `_gpuData` on the renderables themselves. This means that a lot of code could be removed around how renderables are destroyed and ultimatly we end up with a simpler system! 

This also removed the need for the `RenderableGCSystem` which can be depricated / removed in another PR

##### Pre-Merge Checklist


ivanpopelyshev (approved):
Finally.

Zyie (approved):

PR #11352, opened 2025-03-26
fix: Inverse stencil mask should not modify stencil buffer during normal draws

### LINKED ISSUE #11301: Bug: Inverse mask ignores the render order and draws on top
### Current Behavior

WebGL renderer: Inverse masked graphics object gets drawn on top of everything, ignoring the expected render order. On the WebGPU renderer it works as expected.

### Expected Behavior

Inverse mask graphics object should be drawn in the order it's added to the scene graph.

### Steps to Reproduce

[https://www.pixiplayground.com/#/edit/ZhxUi6l2lD_4npqwn8ve8](url)

Comment/Uncomment the line that switches renderer type and see the different render order of the circle and square.

### Environment

- **`pixi.js` version**: 8.8.0
- **Browser & Version**: Chrome 133
- **Running Example**: *e.g. [https://www.pixiplayground.com/#/edit/ZhxUi6l2lD_4npqwn8ve8](url)

### Possible Solution

_No response_

### Additional Information

_No response_

Mysak0CZ:
fixes #11301 

##### Description of change

The current implementation of inverse masks works by incrementing stencil buffer where mask is, and then draws only on pixels that don't match the mask depth.
The `passOp`, however, was set to `replace`, which caused the stencil buffer to be set to the mask value in all areas where something was drawn.
This effectively caused the effect that only first drawn thing was visible per-pixel.

This simple fix changes the `passOp` to be `"keep"` - same as non-inverse variant, and updates the relevant test to test for this usecase.

##### Pre-Merge Checklist


Mysak0CZ:
Hello @GoodBoyDigital! Could I please ask you to review this tiny fix? Thank you in advance!

PR #11374, opened 2025-04-17
fix default values on removeParticles logic

### LINKED ISSUE #11357: Bug: The "removeParticles" function has wrong default values.
### Current Behavior

When I use removeParticles without endIndex, it doesn't work.

### Expected Behavior

According to the docs, endIndex (the ending position) defaults to the container's size.

### Steps to Reproduce

[pixiplayground](https://www.pixiplayground.com/#/edit/6wVxRNW4E6PQjix8zacpe)

### Environment

### Possible Solution

_No response_

### Additional Information

_No response_

PaulKokhanov1:
##### Description of change

Fix the `removeParticles` logic to match the description provided. We are now setting the endIndex to default to the container length

Closes #11357 

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/particle-container/shared/ParticleContainer.ts
+++ b/src/scene/particle-container/shared/ParticleContainer.ts
@@ -300,10 +300,13 @@ export class ParticleContainer extends ViewContainer implements Instruction
      * @param endIndex - The ending position. Default value is size of the container.
      * @returns - List of removed particles
      */
-    public removeParticles(beginIndex?: number, endIndex?: number)
-    {
-        const children = this.particleChildren.splice(beginIndex, endIndex);
-
+    public removeParticles(beginIndex?: number, endIndex?: number) {
+        // Default beginIndex to 0 if not provided
+        const start = beginIndex ?? 0;
+        // Default endIndex to the length of particleChildren if not provided
+        const end = endIndex ?? this.particleChildren.length;
+        // Remove the correct range
+        const children = this.particleChildren.splice(start, end);
         this.onViewUpdate();
 
         return children;


src/scene/particle-container/shared/ParticleContainer.ts
```diff
@@ -302,7 +302,12 @@ export class ParticleContainer extends ViewContainer implements Instruction
      */
     public removeParticles(beginIndex?: number, endIndex?: number)
     {
-        const children = this.particleChildren.splice(beginIndex, endIndex);
+        // Default beginIndex to 0 if not provided
+        const start = beginIndex ?? 0;
+        // Default endIndex to the length of particleChildren if not provided
+        const end = endIndex ?? this.particleChildren.length;
+        // Remove the correct range
+        const children = this.particleChildren.splice(start, end);
```
bigtimebuddy, line 310: Alternatively: 
```suggestion
        // Default beginIndex to 0 if not provided
        beginIndex ??= 0;
        // Default endIndex to the length of particleChildren if not provided
        endIndex ??= this.particleChildren.length;
        // Remove the correct range
        const children = this.particleChildren.splice(beginIndex, endIndex);
```

susiwen8: This should be considered a breaking change because users may have removeParticles() calls in production. If their pixijs version isn't locked, their next release could exhibit unintentional behavior if they upgrade without thorough testing.

GoodBoyDigital: tweaked this based on your feedback @bigtimebuddy . 
Thanks for youre insight @susiwen8! , whilst it is kind of a breaking change, it also is fixing incorrect behaviour. With that in mind, I think on this occasion its ok not to consider this breaking as the use case would be quite strange!

src/scene/particle-container/shared/ParticleContainer.ts
```diff
@@ -302,12 +302,11 @@ export class ParticleContainer extends ViewContainer implements Instruction
      */
     public removeParticles(beginIndex?: number, endIndex?: number)
     {
-        // Default beginIndex to 0 if not provided
-        const start = beginIndex ?? 0;
-        // Default endIndex to the length of particleChildren if not provided
-        const end = endIndex ?? this.particleChildren.length;
+        beginIndex ||= 0;
+        endIndex ||= this.particleChildren.length;
```
bigtimebuddy, line 306: We should use ??= to handle undefined. Or else passing endIndex of 0 would use the particleChildren.
```suggestion
        beginIndex ??= 0;
        endIndex ??= this.particleChildren.length;
```

Zyie (approved):

PR #11384, opened 2025-04-22
fix: render layer detachment when destroying.

### LINKED ISSUE #11373: Bug: Destroying parent which destroys children should also remove children from any render layers
### Current Behavior

When the parent is destroyed, resulting in children destroyed, we should also remove children from any render layers else we keep getting the warning

```
Container must be added to both layer and scene graph. Layers only handle render order - the scene graph is required for transforms (addChild) 
```

Not doing so will require keeping track of every transient child container which has been added to render layers and manually removing them from render layers

### Expected Behavior

Destroyed children should not give "Container must be added to both layer and scene graph. Layers only handle render order - the scene graph is required for transforms (addChild)" warning

### Steps to Reproduce

NA

### Environment

- **`pixi.js` version**: *8.9.1*

### Possible Solution

Currently i always have to do this

```typescript
destroy(options?: DestroyOptions): void {
    // Detach children from render layers (prevent stray warnings)
    for (const child of this.children) {
        child.parentRenderLayer.detach(child);
    }
}
```

### Additional Information

_No response_

GoodBoyDigital:
##### Description of change

Fix issue where when destroying a container - the children were not being removed from the render layers. 
Added test too.

fixes #11373 

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/container/container-mixins/childrenHelperMixin.ts
+++ b/src/scene/container/container-mixins/childrenHelperMixin.ts
@@ -61,7 +61,14 @@ export const childrenHelperMixin: ChildrenHelperMixin<ContainerChild> = {
 
             for (let i = 0; i < removed.length; ++i)
             {
-                this.emit('childRemoved', removed[i], this, i);
+                const child = removed[i];
+
+                if (child.parentRenderLayer)
+                {
+                    child.parentRenderLayer.detach(child);
+                }
+
+                this.emit('childRemoved', child, this, i);
                 removed[i].emit('removed', this);
             }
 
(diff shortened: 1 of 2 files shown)


src/scene/container/container-mixins/childrenHelperMixin.ts
```diff
@@ -61,7 +61,14 @@ export const childrenHelperMixin: ChildrenHelperMixin<ContainerChild> = {
 
             for (let i = 0; i < removed.length; ++i)
             {
-                this.emit('childRemoved', removed[i], this, i);
+                const child = removed[i];
+
+                if (child.parentRenderLayer)
+                {
+                    child.parentRenderLayer.detach(child);
+                }
```
bigtimebuddy, line 69:
```suggestion
                child.parentRenderLayer?.detach(child);
```

bigtimebuddy (approved):

PR #11400, opened 2025-05-02
fix: Dom Container on scaled canvas

GoodBoyDigital:
##### Description of change
This change fixes an issue in `DOMPipe` where DOM elements managed by `DOMContainer` would not be positioned correctly if the parent canvas element was scaled using CSS properties like `width` or `height`.

Previously, the container `div` for DOM elements was only translated based on the canvas `offsetLeft` and `offsetTop`. This fix introduces scaling to the container `div`'s transform. The scale factor is calculated based on the ratio of the canvas's CSS `style.height` to its `canvas.height`, multiplied by the renderer's resolution. This ensures that the DOM elements align correctly with the PixiJS scene, even when the canvas itself is scaled via CSS.

##### Pre-Merge Checklist

### DIFF
--- a/src/dom/DOMPipe.ts
+++ b/src/dom/DOMPipe.ts
@@ -122,7 +122,10 @@ export class DOMPipe implements RenderPipe<DOMContainer>
             canvas.parentNode?.appendChild(this._domElement);
         }
 
-        this._domElement.style.transform = `translate(${canvas.offsetLeft}px, ${canvas.offsetTop}px)`;
+        const scale = (parseFloat(canvas.style.height) / canvas.height) * this._renderer.resolution;
+
+        // scale according to the canvas scale and translate
+        this._domElement.style.transform = `translate(${canvas.offsetLeft}px, ${canvas.offsetTop}px) scale(${scale})`;
 
         for (let i = 0; i < attachedDomElements.length; i++)
         {


src/dom/DOMPipe.ts
```diff
@@ -122,7 +122,10 @@ export class DOMPipe implements RenderPipe<DOMContainer>
             canvas.parentNode?.appendChild(this._domElement);
         }
 
-        this._domElement.style.transform = `translate(${canvas.offsetLeft}px, ${canvas.offsetTop}px)`;
+        const scale = (parseFloat(canvas.style.height) / canvas.height) * this._renderer.resolution;
+
+        // scale according to the canvas scale and translate
+        this._domElement.style.transform = `translate(${canvas.offsetLeft}px, ${canvas.offsetTop}px) scale(${scale})`;
```
Zyie, line 128: Do we not need to also check the width here?

```ts
const scaleX = (parseFloat(canvas.style.width) / canvas.width) * this._renderer.resolution;
const scaleY = (parseFloat(canvas.style.height) / canvas.height) * this._renderer.resolution;

// scale according to the canvas scale and translate
this._domElement.style.transform = `translate(${canvas.offsetLeft}px, ${canvas.offsetTop}px) scale(${scaleX}, ${scaleY})`;

Zyie:
Could we have a test for this?

src/dom/DOMPipe.ts
GoodBoyDigital: we 100% should !

GoodBoyDigital:
> Could we have a test for this?

yes! test added 👍

Zyie (approved):

PR #11401, opened 2025-05-02
feat: add onRender option to Container interface and corresponding test

GoodBoyDigital:
##### Description of change
This change introduces the ability to provide an `onRender` callback function directly within the `Container` constructor options.

Previously, the `onRender` property could only be set after the `Container` instance was created. This update adds the `onRender` field to the `ContainerOptions` interface and ensures it is assigned during object instantiation. A corresponding unit test has been added to verify this new constructor behavior.

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/container/Container.ts
+++ b/src/scene/container/Container.ts
@@ -152,6 +152,8 @@ export interface ContainerOptions<C extends ContainerChild = ContainerChild> ext
     y?: number;
     /** @see scene.Container#boundArea */
     boundsArea?: Rectangle;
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):

Zyie (changes requested):
I'm not sure this PR is needed, we already have the `OnRenderMixinConstructor` interface that adds the `onRender` type to the containers constructor.

I removed the type change and the new test still passed

GoodBoyDigital:
my bad! must have been having a bad types day :D

PR #11417, opened 2025-05-09
feat: add a limits system to WebGL and WebGPU renderers

### LINKED ISSUE #11389: Bug: WebGPU broken on some devices
### Current Behavior

When following the Pixijs examples on some devices, adding `preference: "webgpu"` causes the rendering to fail.

This can be replicated on Google Pixel 9s as well as Chrome browsers on Ubuntu 24.04 systems when WebGPU is enabled by the chrome flags.

Viewing the console logs will show this:

Image

### Expected Behavior

The expected behavior is for it to render on all devices with WebGPU

### Steps to Reproduce

See above (Pixijs examples on pixel 9 or Ubuntu Chrome with WebGPU enabled)

### Environment

- **`pixi.js` version**: *8.9.1.*
- **Browser & Version**: *latest chrome*
- **OS & Version**: *Google Pixel 9 or Ubuntu 24.04*
- **Running Example**: *Pixijs containers example*

### Possible Solution

The root cause of this issue is that the WebGPU batcher in Pixijs is setting `maxTextures` based on WebGL's  `MAX_TEXTURE_IMAGE_UNITS` parameter rather than WebGPU's `device.limits.maxSampledTexturesPerShaderStage`

This falsely assumes that these two values are equal. However, there are systems where this isn't true. For example, our Pixel 9's and my Ubuntu Chrome browser with the unsafe WebGPU feature flag.

You can see if yours are different values easily here: https://jsfiddle.net/hmbf9skw/1/

On my Pixel 9 that can't render:

Image

On my friend's Samsung Galaxy S24 that renders successfully:

Image

I believe the culprit code is:

https://github.com/pixijs/pixijs/blob/9d177e2d2d3c33df2b0ae7afef180c5028e4d3af/src/rendering/batcher/gpu/getTextureBatchBindGroup.ts#L31

As well as potentially:

https://github.com/pixijs/pixijs/blob/9d177e2d2d3c33df2b0ae7afef180c5028e4d3af/src/rendering/batcher/shared/Batcher.ts#L358

TLDR: The codebase needs to be updated to use the WebGPU limits for WebGPU instead of the WebGL parameters.

### Additional Information

Potentially related tickets:

#11160 
#11388 
#10375

GoodBoyDigital:
##### Description of change

Added a new limits system for both WebGL and WebGPU renderers. This system provides a unified and convenient way to access GPU feature limits—such as the maximum number of textures—even though the method of querying these limits differs between the two APIs.

For example, maxTextures is now accessed through this shared system rather than using separate logic for each renderer. This change allows us to handle limits generically and expand the system easily in the future.

Important Note:
As part of this update, I had to refactor several areas of the codebase that previously relied on a global WebGL-specific method to get maxTextures. As discussed in #11389, WebGL and WebGPU can return different values for the same properties—which is a bit odd—but this makes a unified system even more important.

One trade-off is that we now have to wait until the rendering context (WebGL) or device (WebGPU) is ready before accessing these limits. Since WebGPU setup is asynchronous, anything that depends on these values must be initialized accordingly.

fixes #11389

##### Pre-Merge Checklist


ivanpopelyshev (approved):
Do it, and then I'll add buffer size and texture size for WebGPU.

Zyie (approved):

Zyie (approved):

PR #11423, opened 2025-05-16
chore: documentation overhaul with typedoc

Zyie:
This PR overhauls our documentation to use TypeDoc and removes the documentation quirks from using webdoc

This also introduces the idea that anything that has been exported should explicitly state whether or not it is meant for internal or public use. This is done by using the `@internal` tag. This should help us clearly define our public API and give us more freedom to change the internals of pixi

To help ensure our docs remain consistent there is also a custom eslint plugin that forces a tag to be applied to anything that has been exported

https://pixijs.download/feat/typedoc/docs/index.html

Follow up tasks:
- apply styling to match current docs
- add redirects to help with migrating over to new docs structure
- add back google analytics tracking
- implement `@experimental` tags
- implement `@standard` / `@advanced` tags


GoodBoyDigital (approved):
this is great work Sean!

bigtimebuddy:
Awesome job @Zyie! I'm glad we got the big lift here out of the way in this PR. Do you want to capture a project board with those remaining tasks?

Issue #10633, opened 2024-06-16
Bug: v8.x Failure to play Video source on iOS 17.5.1 (iPhone + iPad both tested)


Liquidream:
### Current Behavior

When using `Assets.load` to load a video resource in PixiJS v8.x - the video sprite does not render.

### Expected Behavior

When using `Assets.load` to load a video resource in PixiJS v8.x - the video sprite should render, even if it doesn't autoplay.

### Steps to Reproduce

1. Open the [Video playground example](https://pixijs.com/8.x/playground?exampleId=sprite.video) on an iOS device
2. Click the ▶️ button,
3. Nothing happens

NOTE: Play button doesn't disappear, so think there must be an error happening, but I'm unable to see what it is.

### Environment

- **`pixi.js` version**: *e.g. 7.1.0*
- **Browser & Version**: *e.g. Chrome 108*
- **OS & Version**: *e.g. Ubuntu 22.04*
- **Running Example**: *e.g. <https://pixiplayground.com/>*

### Additional Information

Issue recreated using current v8.x [Video playground example](https://pixijs.com/8.x/playground?exampleId=sprite.video) on iPhone AND iPad, both running iOS 17.5.1.
(v7.x of the playground example works fine)

bigtimebuddy:
I cannot reproduce on iOS 17.5.1. Video plays as expected.

Liquidream:
> I cannot reproduce on iOS 17.5.1. Video plays as expected.

Interesting. Ok thanks, I'll check to see if it's somehow network related (only other thing I can think of)

bigtimebuddy:
Yeah please report back if you find something. Are these recent iPad and iPhone versions?

bigtimebuddy:
Also, try some other WebGL content to see if is Pixi + environment or just environment.

Liquidream:
Thanks for the additional thoughts, @bigtimebuddy.

> Are these recent iPad and iPhone versions?

Recent iPhone (13) with on iOS 17.5.1 works fine - as does another similar device on BrowserStack.
But my iPad (6th Gen) on same iOS 17.5.1 is not playing the video in any of the browsers (inc. private mode).

> Also, try some other WebGL content to see if is Pixi + environment or just environment.

All other demos seem to work just fine (inc. others that load lots of remote assets), it's just the video demo.
Yet the v7.x version of video demo plays fine.  Really weird.

I'm sure it must be something else "special" about my iPad, just not tracked down "what" yet 😅

[...]

PR #11451, opened 2025-05-29
fix: iOS issue loading video when autoplay is false and preload is true.

jgelens:
##### Description of change

This fixes bug #10633, tested on both an iPad and iPhone. load() is not called on the video element at all when autoload is false and preload is true.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Thank you @jgelens

bigtimebuddy:
@jgelens sorry, the CI needs to be approved to run for outside contributors. Kicked that off.

PR #11476, opened 2025-06-11
Feat: Add PaddingSides to text

ivanpopelyshev:
New CSS-like API feature: `textStyle.padding = [top, right, bottom, left]` 

Improve code quality by affecting texture frame/orig/trim , remove override methods like `setSize` because they are the same as Sprite now

### DIFF
--- a/src/filters/PaddingSides.ts
+++ b/src/filters/PaddingSides.ts
@@ -1,0 +1,164 @@
+import type { Filter } from './Filter';
+
+/**
+ * class used for paddings of text styles and filters
+ * @category filters
+ * @advanced
+ */
+export class PaddingSides
+{
+    public top: number = 0;
+    public right: number = 0;
+    public bottom: number = 0;
+    public left: number = 0;
+
+    get horizontal()
+    {
+        return this.left + this.right;
+    }
+
+    get vertical()
+    {
+        return this.top + this.bottom;
+    }
+
+    public copyFrom(sides: IPaddingSidesLike): this
+    {
+        if (!sides)
+        {
+            this.top = 0;
+            this.right = 0;
+            this.bottom = 0;
+            this.left = 0;
+        }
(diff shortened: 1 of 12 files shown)


ivanpopelyshev:
I'm working on fixing "setting width/height on constructor" problem

src/filters/PaddingSides.ts
```diff
@@ -0,0 +1,164 @@
+import type { Filter } from './Filter';
+
+/**
+ * class used for paddings of text styles and filters
+ * @category filters
+ * @advanced
+ */
+export class PaddingSides
```
bigtimebuddy, line 8: Can this not be a class? The usages below are mostly normalizing a value or doing some compute. This can be function and doesn't add much value as a class.

src/filters/PaddingSides.ts
```diff
@@ -0,0 +1,164 @@
...
+            this.right = sides;
+            this.bottom = sides;
+            this.left = sides;
+        }
+        else if (sides instanceof Array)
+        {
+            this.top = (sides as any)[0];
+            this.right = (sides as any)[1];
+            this.bottom = (sides as any)[2];
+            this.left = (sides as any)[3];
```
bigtimebuddy, line 46: Please do not cast any here. Consider adding a type-guard function instead to keep from using any.

```ts
function isNumberList(value: unknown): value is [number, number, number, number] {
  return value instanceof Array
    && value.length === 4
    && value.every(v => typeof v === 'number');
}
```

src/scene/text/AbstractText.ts
```diff
@@ -478,124 +476,6 @@ export abstract class AbstractText<
         this.onViewUpdate();
     }
 
-    /**
```
bigtimebuddy, line 481: I'm not sure I understand why you're removing width, height, getSize and setSize from here. Can you clarify?

PR #11496, opened 2025-06-24
feat: adds text splitting functionality

Zyie:
Adds the ability to split Text and BitmapText objects into characters, words, and lines.
This provides a more granular control over text elements, allowing for advanced animations and effects.

We now have two new classes `SplitText` and `SplitBitmapText`. I decided to split these up to allow for as much tree shaking to be preserved as possible. If we went with the one class approach you would always be bundling `Text` and `BitmapText`.

These classes also support dynamic updates to styles and text. Changing text does mean we destroy all the children and make them again. Changes to style are safe as we just update the position/style without needing to create a new instance.

examples

Split Existing Text
```ts
const result = SplitBitmapText.from(myBitmapText);
const result = SplitBitmapText.from(myBitmapText, {...));
```

New Instance
```ts
const result = new SplitBitmapText({
  text: 'Hello World',
  style: {...},
  lineAnchor: 0.5,  // sets origin to be centered for easy animations
  wordAnchor: 0.5,
  charAnchor: 0.5,
  autoSplit: true
});
```

Working with Results
```ts
const result = SplitBitmapText.from(myBitmapText);

// Animate characters
result.chars.forEach((char, i) => {
  char.alpha = 0;
  // Fade in each character sequentially
  gsap.to(char, {
    alpha: 1,
    delay: i * 0.1
  });
});
```

demo: https://codesandbox.io/p/sandbox/pixi-js-sandbox-forked-chgp88
demo2: https://codesandbox.io/p/devbox/pixi-js-sandbox-forked-yly7k3
demo3: https://codesandbox.io/p/devbox/pixi-js-sandbox-forked-j4swyl
demo4: https://codesandbox.io/p/devbox/pixi-js-sandbox-forked-sfny96
demo5: https://codesandbox.io/p/devbox/pixi-js-sandbox-forked-cctmf8


bigtimebuddy:
This is a very cool idea and well documented! Excellent bar of quality as usual @Zyie.

Would you mind creating a quick demo (maybe with animation) to showcase this?

I have one general concern which is about destroying lifecycle. Since the split result is a generic object, I don't see an easy way to destroy the extra text elements and containers. Our contract generally is: _if Pixi instantiates it, Pixi destroys it_. How would a user do that here? Also `replace` removes it from the display tree, but does not destroy, right?

Another API approach might be to introduce a new display class to manage the element more directly (maybe that extends Container). Something like `TextAnimator`, `TextSegmenter`. This gives us more flexibility in the future and maintain the current contract with objects created and objects destroyed.

```js
const myText = new Text({
  text: 'Hello World',
  style: {},
});

const seg = new TextSegmenter({
  element: myText,
  replace: true,
});

seg.chars.forEach((char, i) => {
  char.alpha = 0;
  // Fade in each character sequentially
  gsap.to(char, {
    alpha: 1,
    delay: i * 0.1
  });
});

seg.destroy();
```

Zyie:
> Would you mind creating a quick demo (maybe with animation) to showcase this?

Here is a demo link: 
https://codesandbox.io/p/sandbox/pixi-js-sandbox-forked-chgp88 
https://codesandbox.io/p/devbox/pixi-js-sandbox-forked-yly7k3




> I have one general concern which is about destroying lifecycle. Since the split result is a generic object, I don't see an easy way to destroy the extra text elements and containers. Our contract generally is: _if Pixi instantiates it, Pixi destroys it_. How would a user do that here?

So the result from `Text.split` returns an object like this:

```ts
interface Result {
  container: Container;
  lines: Container[];
  words: Container[];
  chars: Text[];
}
```
So to destroy the new text you would use `result.container.destroy(true)`.


> Also `replace` removes it from the display tree, but does not destroy, right?

That is correct. It doesn't destroy the original text. My thinking was that you might want to split the text up for an animation, then swap it back for the original.



> Another API approach might be to introduce a new display class to manage the element more directly (maybe that extends Container). Something like `TextAnimator`, `TextSegmenter`. This gives us more flexibility in the future and maintain the current contract with objects created and objects destroyed.
> 
> ```js
> const seg = new TextSegmenter({
>   element: myText,
>   replace: true,
> });
> seg.destroy();
> ```

I'm not against this approach either. It aligns with what i had in mind originally before going down the current path.
@GoodBoyDigital any thoughts?

Also just as a note that this is all inspired by the gsap plugin that does this for html text: https://gsap.com/docs/v3/Plugins/SplitText/

Zyie:
Me and Mat had a quick chat about this today and agree that we should go down the `TextSegmenter` route. I'll update the PR accordingly!

bigtimebuddy:
Cool. The benefit of an explicit class destroy is that it will also handle all the dereferencing for lines, words, chars arrays.

Zyie:
@bigtimebuddy ready for another review

bigtimebuddy (approved):

src/scene/text-split/AbstractSplitText.ts
```diff
@@ -0,0 +1,554 @@
...
+     * text.autoSplit = false;
+     * text.text = 'Manual Update';
+     * text.split();
+     * ```
+     */
+    set text(value: string)
+    {
+        this._originalText = value;
+        this.lines.forEach((line) => line.destroy({ children: true }));
+        this.lines = [];
```
GoodBoyDigital, line 336: this.lines.length = 0
```suggestion
        this.lines = [];
```

src/scene/text-split/AbstractSplitText.ts
```diff
@@ -0,0 +1,554 @@
...
+        style ||= {};
+
+        this._style = new TextStyle(style);
+
+        // tidy up word/line containers, characters can be reused
+        this.words.forEach((word) => word.destroy());
+        this.words = [];
+
+        this.lines.forEach((line) => line.destroy());
+        this.lines = [];
```
GoodBoyDigital, line 510:
```suggestion
        this.lines.length = 0;
```

Issue #11465, opened 2025-06-06
Bug: Heavy performance regression with many Text with 8.10.0


joergplewe:
### Current Behavior

Hi! 

I experienced a severe performance breakdown when updating from 8.9.2 to 8.10.0. My app went from a solid 60fps to 5fps.
This even affected other browser tabs.

I suspect the problem to be somewhere around the `Text` object. 

### Expected Behavior

No performance regression

### Steps to Reproduce

https://www.pixiplayground.com/#/edit/JN1VsNYeHKpt0o2GRU7ik

Try with 8.9.2 and 8.10.0

### Environment

- **`pixi.js` version**: *8.10.0*
- **Browser & Version**: *any*
- **OS & Version**: *MacOS and Windows*
- **Running Example**: *[playground](https://www.pixiplayground.com/#/edit/JN1VsNYeHKpt0o2GRU7ik)*

GoodBoyDigital:
thanks for sharing @joergplewe - going to take a look into this now!

GoodBoyDigital:
hey! This is an interesting one. In the recent update, we removed a bunch of code that managed shared textures between text with the same style as comparing text styles was getting kind of hard and expensive.

So now creating 10 instances of a Text will result in 10 Textures being created. 

Figured the need for 10k texts would be a bit of an edge case. To combat this we actually have a handy API for generating Text Textures that can be used by a dev for more advanced things:

```
// create a text texture
const texture = app.renderer.canvasText.getTexture({
        text:"X", 
        style:{ fontSize: 24, fontFamily: "Verdana", fill: 0xffffff, }
});

const COUNT = 5000; 

for (let i = 0; i < COUNT; i++) {
    const textSprite = new PIXI.Sprite({
        texture,
        anchor: 0.5,
        x:(i % 100) * 25,
        y: Math.floor(i / 100) * 25
    });

    container.addChild(textSprite);
}
```

[example](https://www.pixiplayground.com/#/edit/Z3ZG0WqYIs4VnDZmLRzRm)

Keen to know devs thoughts on this!

joergplewe:
My "edge case" even uses 100000 `Text`s (Symbols coming from a font).
I think I did what you suggest prior to V8, when text caching was introduced.
And I think I could go back to that.

The only thing (right now) is, that in my code at that certain place I have no access to the renderer. I assume having something like a static `Text.createTexture(options)` is not feasible (bc. you don't know about antialias, resolution, ...)?

[...]

PR #11505, opened 2025-06-27
feat: simplified caching for text

Zyie:
**Reintroducing the caching logic for text we had in Pixi <= 8.9, but with a new, simplified approach.**

### Key Changes:

* **Simplified Caching Mechanism:**
  This new implementation removes the somewhat fragile style-key tracking logic. Instead, it uses a much simpler `uid + tick` system to manage cached textures.
  This reduces the code complexity while still providing the benefits of texture caching when reusing the same text style.

  * This approach strikes a balance between performance and maintainability. While it's not as perfectly efficient as tracking every individual style change, it offers a middle ground that's far easier to maintain.
  For example, this would address use cases like [#11465](https://github.com/pixijs/pixijs/issues/11465).


GoodBoyDigital (commented):
I think we should either use a getter for `styleKey`
or rename the function to `getStyleKey` to remove any ambiguity.

GoodBoyDigital:
I would be happy to merge in

- Reintroduced measureText Cache with LRU
- Frozen Filters Array on TextStyle

right now 👍 

Not against the additional caching on text - its just a _lot_ of code to add back in and want to make sure it justifies the use cases!

niusia-ua:
Hey guys. Thank you for your work.

When can we expect the release of this feature? Text objects are integral part of my app.

I just want to know whether should I implement the caching of text textures myself if this PR will take more time to merge, or it is reasonable to wait for the release.

Zyie:
> When can we expect the release of this feature? Text objects are integral part of my app.

Hey, we are planning on getting this in for the next release, which should happen next week

niusia-ua:
Awesome!

GoodBoyDigital (approved):

PR #11520, opened 2025-07-07
fix: HTMLText incorrectly handling font weights

Zyie:
This PR addresses an issue where HTMLText was not correctly applying font weights specified in tag styles.

Key changes:
- Ensures that all fonts weights are generated by `getFontCss`
- Ensures that we use the font face data for CSS generation.
- Updates font caching to support multiple font faces per URL.
- Updates font unloading to correctly remove fonts from the cache and FontFaceSet.


bigtimebuddy (approved):
Nice work 👍

Issue #11440, opened 2025-05-23
Bug: BitmapFontManager.install miss override the fill value if any fill in TextStyle


spd789562:
### Current Behavior

Any Object like fill when `BitmapFontManager.install`, it will just be replaced with white fill when actually used in BitmapText.

![Image](https://github.com/user-attachments/assets/7ad895e1-e47b-42f2-af9f-d16346b86005)

### Expected Behavior

Gradient and any FillPattern should work properly

### Steps to Reproduce

1. BitmapFontManager.install any font with `fill: FillGradient | FillPattern`
```js
    const texture = await Assets.load('https://pixijs.com/assets/bunny.png');
    const pattern = new FillPattern(texture);
    BitmapFontManager.install({
        name: 'TestPatternFont',
        style: {
            fontFamily: 'Arial',
            fontSize: 48,
            fill: pattern,
        }
    })
```
2. Using it in BitmapText
```js
    const text = new BitmapText({
        text: 'Testing preinstall BitmapText',
        style: {
           fontFamily: 'TestPatternFont',
           fontSize: 48,
        }
    })
```

### Environment

- **`pixi.js` version**: 8.8.0
- **Browser & Version**: all
[...]

PR #11523, opened 2025-07-08
fix: Installed bitmap text showing wrong fill

Zyie:
Fixes: #11440

Updates the logic used to determine when bitmap text styles can use tinting instead of baking colors.

Includes a new visual test case to verify correct rendering of bitmap text with gradient fills and tinting.

### DIFF
--- a/src/scene/text-bitmap/BitmapFontManager.ts
+++ b/src/scene/text-bitmap/BitmapFontManager.ts
@@ -213,7 +213,7 @@ export interface BitmapFontInstallOptions
      *         fontSize: 24,
      *         fill: 0xFFFFFF  // Must be white for tinting
      *     },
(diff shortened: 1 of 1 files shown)


Zyie:
> What is the weird artifact in the test image under the "t"?

That is a different bug that i am going to be looking into. This is caused when you define a `fontSize` that is different to the one that is installed with a stroke.

We do have other tests with the same artifact e.g: https://github.com/pixijs/pixijs/blob/dev/tests/visual/snapshots/bitmap-install-layout-text-scene-ts-webgpu.png

GoodBoyDigital (approved):

Zyie:
@bigtimebuddy I've added a `allowFillOverride` option to the install options like we previously talked about. Let me know what you think!

Issue #11537, opened 2025-07-15
Bug: DOMContainer scaling breaks if canvas width/height is percentage


Treborbob:
### Current Behavior

If your canvas heigh/width is set to "100%" it breaks the sacaling of DOMContainer, and those elements appear very small.

### Expected Behavior

DOMContainer elements should appear at the correct scale

### Steps to Reproduce

You can see the effect by editing the DOMContainer example in the playground here: https://pixijs.com/8.x/examples/basic/dom-container

```js
import { Application, DOMContainer } from 'pixi.js';

(async () => {
  const app = new Application();

  await app.init({ backgroundColor: 0x1099bb, resizeTo: window });

// Force canvas to be percentage height/width
  app.canvas.style.width = '100%';
  app.canvas.style.height = '100%';

  document.body.appendChild(app.canvas);
  document.body.style = 'margin: 0; overflow: hidden';

  // Create a DOM element
  const element = document.createElement('textarea');

  element.value = 'Type here...';

  // Create a DOM container
  const domContainer = new DOMContainer({
    element,
    x: app.screen.width / 2,
    y: app.screen.height / 2,
    anchor: 0.5,
  });

  // Add it to your scene
  app.stage.addChild(domContainer);

  app.ticker.add(() => {
    // Rotate the DOM container
    domContainer.rotation += 0.01;
  });
})();
```

### Environment

- **`pixi.js` version**: 8.11.0
- **Browser & Version**: Firefox 140.0.4
- **OS & Version**: MacOS 15.5
- **Running Example**: https://pixijs.com/8.x/examples/basic/dom-container (and update the JS with the above)

### Possible Solution

The issue seems to be from `DOMPipe.postrender():`

```ts
const sx = (parseFloat(canvas.style.width) / canvas.width) * this._renderer.resolution;
const sy = (parseFloat(canvas.style.height) / canvas.height) * this._renderer.resolution;
```

**Suggested Fix**

Replace:

```ts
const sx = (parseFloat(canvas.style.width) / canvas.width) * this._renderer.resolution;
const sy = (parseFloat(canvas.style.height) / canvas.height) * this._renderer.resolution;
```

with:

```ts
const boundingRect = canvas.getBoundingClientRect();
const sx = (boundingRect.width / canvas.width) * this._renderer.resolution;
const sy = (boundingRect.height / canvas.height) * this._renderer.resolution;
```

PR #11539, opened 2025-07-15
fix: DOMContainer scaling if canvas width/height is percentage

Zyie:
Fixes: #11537

Addresses issues with incorrect DOMContainer positioning when the canvas is scaled.

Introduces CanvasTransformSync to manage the DOM element's transform based on canvas size and position, utilizing
ResizeObserver for efficient updates. Uses a fallback to the Ticker is ResizeObserver is not found (been supported since 2020)

Ensures correct scaling and translation of DOM elements relative to the canvas, resolving discrepancies in rendering.

### DIFF
--- a/src/dom/DOMPipe.ts
+++ b/src/dom/DOMPipe.ts
@@ -1,5 +1,5 @@
 import { ExtensionType } from '../extensions/Extensions';
-import { CanvasTransformObserver } from './CanvasTransformObserver';
+import { CanvasObserver } from './CanvasObserver';
 import { type DOMContainer } from './DOMContainer';
 
 import type { InstructionSet } from '../rendering/renderers/shared/instructions/InstructionSet';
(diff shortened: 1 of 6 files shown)


src/dom/DOMPipe.ts
```diff
@@ -97,18 +106,8 @@ export class DOMPipe implements RenderPipe<DOMContainer>
...
-            canvas.parentNode?.appendChild(this._domElement);
-        }
-
-        const sx = (parseFloat(canvas.style.width) / canvas.width) * this._renderer.resolution;
-        const sy = (parseFloat(canvas.style.height) / canvas.height) * this._renderer.resolution;
-        // scale according to the canvas scale and translate
-
-        this._domElement.style.transform = `translate(${canvas.offsetLeft}px, ${canvas.offsetTop}px) scale(${sx}, ${sy})`;
+        // Ensure the canvas is set
+        this._canvasTransformSync.canvas = this._renderer.view.canvas as HTMLCanvasElement;
```
GoodBoyDigital, line 110: I think we can simplify this to leveraging `init` hook in the systems 

 we can create `CanvasTransformSync`

```
  this._canvasTransformSync = new CanvasTransformSync({
            domElement: this._domElement,
            renderer: this._renderer, // canvas / view will be ready
  });
```

that way `CanvasTransformSync` does not need an init function and can be set up in the constructor simplifying the code flow.

Issue #11218, opened 2025-01-21
Bug: Memory leak when using FillGradient


FacePlace:
### Current Behavior

When using `FillGradient`, the `CanvasRenderingContext2D` doesn't get removed from memory, causing the leak.

![Image](https://github.com/user-attachments/assets/67f0d320-b563-488d-9fd7-e334982940d5)

### Expected Behavior

Everything related to `FillGradient` should be destroyed to free memory.

### Steps to Reproduce

Create many objects that utilize `FillGradient`, and then destroy them. In the provided example, toggle `enableMemoryLeak` to use solid color instead of a gradient and avoid the leak.

### Environment

- **`pixi.js` version**: *8.6.6*
- **Browser & Version**: *Chrome 132.0.6834.84*
- **OS & Version**: *Windows 10*
- **Running Example**: https://www.pixiplayground.com/#/edit/ocT0NMQDJQyvBW-OU8os4

bigtimebuddy:
Good catch. FillGradient needs a `destroy` method that clean's up the internal `texture` reference created. I'm guessing Graphics' `clear` and `destroy` should call this cleanup.

PR #11543, opened 2025-07-16
fix: resource leak in Graphics and fills

Zyie:
Fixes: #11218

Ensures proper destruction of resources associated with Graphics and fills, preventing memory leaks.

This change introduces `destroy` methods to `FillGradient` and `FillPattern` to release their textures and other resources.

The GraphicsContext now utilizes these destroy methods when cleaning up fills during its own destruction.

A new test case is added to verify the correct destruction behavior.

### DIFF
--- a/src/scene/graphics/__tests__/Graphics.Destroy.test.ts
+++ b/src/scene/graphics/__tests__/Graphics.Destroy.test.ts
@@ -1,4 +1,5 @@
 import { Container } from '../../container/Container';
+import { FillGradient } from '../shared/fill/FillGradient';
 import { Graphics } from '../shared/Graphics';
 import { GraphicsContext } from '../shared/GraphicsContext';
 import '../init';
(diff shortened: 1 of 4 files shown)


bigtimebuddy (commented):
What if a gradient fill is shared between multiple graphics?

Zyie:
> What if a gradient fill is shared between multiple graphics?

This PR doesn't change how this was previously handled. Just does a better job of tidying it all up.
If you called `graphic.destroy(true)` and had multiple graphics using the same gradient then it would have destroyed the texture previously due to this line:

https://github.com/pixijs/pixijs/blob/7d91234eb506457fb6e612b02b31cdf6dae7062f/src/scene/graphics/shared/GraphicsContext.ts#L1218

We don't really support using the same gradient across multiple graphics unless you are sharing the same context:

```ts
const g1 = new Graphics().fill(gradient);
const g2 = new Graphics({context: g1.context});

g1.destroy(); // doesn't affect g2
g1.destroy({context: true, texture: true}) // would affect both
```

bigtimebuddy (approved):

Issue #10896, opened 2024-08-28
Bug: When the URL address of a video file does not have an extension, using 'Assets. load' cannot load resources！ Help !


raitawyc:
### Current Behavior

I have encountered a problem:
The video resources are stored on the OSS server, and the URL of the video does not have an extension, so I cannot load the video resources. May I ask how to configure access to video resources without extensions?

My current configuration is as follows:

```
const textureVideo = await Assets.load({
src: ’https://xxx/oss/xxx-video/123456789‘,
data: {
preload: true,
autoPlay: true
}，
loadParser: 'loadVideo'
});

console.log(textureVideo);
```

Nothing was printed out

### Expected Behavior

I hope the video can be loaded。

If the video files I store on the server do not have an extension, how can I configure them to load the video files?

### Steps to Reproduce

When debugging the following TS files, I found that when the URL does not include the extension of the video file, the type of the source tag cannot be assigned, so the video resources cannot be loaded correctly

npm/pixi.js@8.x/src/assets/loader/parsers/textures/loadVideoTextures.ts

        **// Determine MIME type
        let mime: string | undefined;

        if (url.startsWith('data:'))
        {
            mime = url.slice(5, url.indexOf(';'));
        }
        else if (!url.startsWith('blob:'))
        {
            const ext = url.split('?')[0].slice(url.lastIndexOf('.') + 1).toLowerCase();

            mime = VideoSource.MIME_TYPES[ext] || `video/${ext}`;
        }

        sourceElement.src = url;

        if (mime)
        {
            sourceElement.type = mime;
        }**

### Environment

- **`pixi.js` version**: *e.g. 8.3.4*
- **Browser & Version**: *e.g. Chrome 127*
- **OS & Version**: *e.g. windows11*

nw-b:
Experiencing this issue as well, on latest Pixi 8.7.2 For some assets we append a cache busting query string, which works with normal textures but not videos.

ie. https://assets.example.com/myvideo.mp4?t=2025-01-25T18:40:36.000000Z

- With the `?t=` query string, `Assets.load` promise never resolves or rejects, nor attempts to load anything.
- Without the query string, the video loads as expected.

`Assets.load` does not seem to choke on the query string for any other kinds of assets.

raitawyc:
[...]

PR #11546, opened 2025-07-16
fix: video texture loading with no extension

Zyie:
Fixes: #10896

Allows specifying a MIME type for video textures during loading. This ensures correct handling of video data, particularly when the URL doesn't explicitly define it.

```ts
const textureVideo = await Assets.load({
  src: 'https://xxx/oss/xxx-video/123456789',
  data: {
    mime: 'video/mp4'
  }，
  loadParser: 'loadVideo'
});
```


bigtimebuddy (approved):
Nice!

raitawyc:
perfect！Thank you for solving my problem, even though a year has passed. Hehe

Issue #11492, opened 2025-06-19
Bug: v8.10.1 -> Text stoping updates while using WebGPU renderer


zatx-private:
### Current Behavior

Have updated Pixi from 8.9.2 to 8.10.1 and noticed, that Text updating only once or twice, after that bounds updating correctly but the drawn text itself is not updating 

### Expected Behavior

Text updating every time after setting new value 

### Steps to Reproduce

Make simple demo
https://www.pixiplayground.com/#/edit/YPkb-7ADtrintsfHMnHz1

### Environment

- **`pixi.js` version**: *e.g. 8.10.1*
- **Browser & Version**: *e.g. Chrome 137*
- **OS & Version**: *e.g. Mac OS Tahoe 26.0 Beta*
- **Running Example**: *e.g. <https://pixiplayground.com/>*

### Additional Information

Noticed current behavior after this commit
https://github.com/pixijs/pixijs/commit/7602157ab63bc4be61751fbcc7df7735ea4a24e2

bcazur:
I've noticed the same issue. It is working in Safari and not working in Chrome. The bug I started from was style not updating - see a modified version of your sample here: https://www.pixiplayground.com/#/edit/QNbf_v9NLJYqhzJFM3P4K 

Thanks for raising it!

AndreMicheletti:
Same is happening here, `pixi.js@8.11.0`

NerdHorse:
Hi @bigtimebuddy  I tried to reverse-engineer the problem and ended up identifying this issue.
The onViewUpdate from ViewContainer isn’t executing the full flow because it stops at the didViewUpdate condition.
I’m not sure if it’s related, but in AbstractText, you check didViewUpdate, and if it’s false, you set didTextUpdate to true.
If I remove the condition, onViewUpdate works.
Well, I don’t think that’s the solution, as it probably impacts the renderer’s performance,but it’s a lead for your investigation.![image](https://github.com/user-attachments/assets/55d428ee-3870-4bca-8c61-2d408a062c84)![image](https://github.com/user-attachments/assets/07a2c10c-f361-4c75-a3fc-4a7523a7a548)

AndreMicheletti:
Hi, adding to the discussion, I just found a _-not optimal-_ workaround, that is to create a new Text object everytime the content changes.

But the thing I noticed is: if you **destroy** the old Text object, the solutions does not work. But if you only **visible=false** it, it works.

And if you add a **timeout** to destroy the object, it then works.

[...]

Issue #11294, opened 2025-02-20
Bug: BitmapText not update whlie being invisible.


phutaekwondo:
### Current Behavior

bitmapText show "**second set text**"

### Expected Behavior

bitmapText show "**first set text**"

### Steps to Reproduce

**In short**: 
set bitmapText visible to false; 
set bitmapText text; 
set bitmapText text; 
set bitmapText visible to true;
delay a bit each step.

**In not so short**:
Just create a simple BitmapText.

Then run the code:

```typescript
const delay = async () =>
{
  const duration = 1000; //ms
  return new Promise(resolve => setTimeout(resolve, duration));
}

await delay();
bitmapText.visible = false;

await delay();
bitmapText.text = "first set text";

await delay();
bitmapText.text = "second set text";

await delay();
bitmapText.visible = true;

```

### Environment

- **`pixi.js` version**: *8.8.0*
- **Browser & Version**: *Chrome 133.0.6943.99*
- **OS & Version**: *Windows 10 Pro 22H2 (OS Build 19045.5487)*
- **Running Example**: *https://github.com/phutaekwondo/pixijs-bitmap-text-bug-re-produce*

### Possible Solution

Assume we have a `bitmapText` (an instance of `BitmapText`). set the `bitmapText.visible` to **false**, then set `bitmapText.text` to something.

In `BitmaptextPipe::validateRenderable(bitmapText: BitmapText)`
```typescript
    public validateRenderable(bitmapText: BitmapText): boolean
    {
        const graphicsRenderable = this._getGpuBitmapText(bitmapText);

        if (bitmapText._didTextUpdate)
        {
            bitmapText._didTextUpdate = false;

            this._updateContext(bitmapText, graphicsRenderable);
        }

        return this._renderer.renderPipes.graphics.validateRenderable(graphicsRenderable);

        // TODO - need to shift all the verts in the graphicsData to the new anchor

        // update the anchor...
    }
```

It is updating context of the `bitmapText `and set `bitmapText._didTextUpdate` to **false**, but if `bitmapText `being invisible, the property `bitmapText.didViewUpdate` still be **true**.
and the next time we set `bitmapText.text`, the `bitmapText._didTextUpdate` won't be set to **true**, and won't be updated. because in `AbstractText::onViewUpdate()` it is checking `didViewUpdate` to set **true** to `_didTextUpdate`
```typescript
    public override onViewUpdate()
    {
        if (!this.didViewUpdate) this._didTextUpdate = true;
[...]

PR #11547, opened 2025-07-16
fix: `Text` and `BitmapText` that doesn't render it's new value on some condition

privatestefans:
##### Description of change
- `Text` doesn’t update correctly when using WebGPU. This happens because a new texture is created everytime the text changed, and replacing the original texture that still being used for rendering. The new texture doesn't replace the original GPU texture view, and the batch bind doesn't seems refreshed for the new texture. Because of that, the old texture still got rendered although the text already updated.
  - Fix: #11492
‎ 
- `BitmapText` doesn't update correctly when it's not visible. This occurs because `_didTextUpdate` already set to `false` even it's haven't rendered yet. This can be resolved by removing the one inside `validateRenderable` function, since the `addRenderable` function also sets `_didTextUpdate` to false and it correctly render the text.
  - Fix: #11294
‎ 
- Playground with the bug: https://pixiplayground.com/#/edit/LbxZkSCezG5SKwQ2TGgEw

##### Pre-Merge Checklist


AndreMicheletti:
Hi, thanks for the PR @privatestefans ! Hope this gets shipped soon

bigtimebuddy (approved):

src/rendering/renderers/gpu/texture/GpuTextureSystem.ts
```diff
@@ -70,6 +70,10 @@ export class GpuTextureSystem implements System, CanvasGenerator
 
     public initSource(source: TextureSource): GPUTexture
     {
+        let gpuTexture = this._gpuSources[source.uid];
```
GoodBoyDigital, line 73: heyhey! whilst this is a fixe, the deeper issue is that initSource should not be called if the texture already exists.

GoodBoyDigital (commented):
appreciate the proposed fix @privatestefans! I would like us to figure out why initSource is being called if the texture is already ready. 

Sorry to be a pain :P

privatestefans:
no worries, I also understand it should not be called every time the text updates. but after digging to the source code, I feel the FPS still drop significantly when we update many text. while this fix improve from 4 FPS to 12FPS, I feel it's better to have cache system for Text -- and I just found having back the cache system like https://github.com/pixijs/pixijs/pull/11505 would actually fix the performance for Text.

btw feel free to modify or close this PR in case if #11505 is preferred, I hope you can find something useful here to accelerate the fix anyway :D

privatestefans:
The `initSource` for WebGL seems like simply bind the new texture tho:
https://github.com/pixijs/pixijs/blob/b19a5b3a0ec0119cc6ad70cbc5e85f343216e5ef/src/rendering/renderers/gl/texture/GlTextureSystem.ts#L101-L104

But for WebGPU it replaced the GPU Texture in `_gpuSources` that still being used for rendering the text and causing the bug where the new text not rendered after being updated.
https://github.com/pixijs/pixijs/blob/b19a5b3a0ec0119cc6ad70cbc5e85f343216e5ef/src/rendering/renderers/gpu/texture/GpuTextureSystem.ts#L103-L105

GoodBoyDigital:
thanks for the context @privatestefans ! this is all super helpful.

GoodBoyDigital (approved):
looks good! I pushed a _tiny_ tweak to bring it inline with the `GLTextureSystem`. Thanks for this fix!

PR #11557, opened 2025-07-17
feat: Refactors asset parser configuration

Zyie:
Migrates from `loadParser` to `parser` for asset configuration.

This change introduces an `id` field to the `LoaderParser` interface and updates the asset loading logic to use this `id` for parser identification, while maintaining backwards compatibility with the `name` field.

New name are:
`json`
`svg`
`text`
`video`
`web-font`
`bitmap-font`
`spritesheet`
`texture`
`basis`
`dds`
`ktx2`
`ktx

### DIFF
--- a/src/assets/loader/Loader.ts
+++ b/src/assets/loader/Loader.ts
@@ -75,10 +75,18 @@ export class Loader
                 // they have? lovely, lets use it
                 parser = this._parserHash[data.parser || data.loadParser];
 
+                // #if _DEBUG
+                if (data.loadParser)
+                {
+                    warn(
+                        `[Assets] "loadParser" is deprecated, use "parser" instead for ${url}`
+                    );
+                }
+                // #endif
+
                 if (!parser)
                 {
                     // #if _DEBUG
-
                     warn(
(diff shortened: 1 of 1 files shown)


src/assets/types.ts
```diff
@@ -27,6 +19,33 @@ export type LoadParserName =
...
+ * ```ts
+ * Assets.load({
+ *     src: assetToLoad,
+ *     parser: 'json' // Use the built-in JSON parser
+ * });
+ * ```
+ * @category assets
+ * @advanced
+ */
+export type AssetParserName =
```
bigtimebuddy, line 34: Nit maybe "AssetParser" instead?

src/assets/loader/Loader.ts
```diff
@@ -70,16 +70,18 @@ export class Loader
             let parser: LoaderParser = null;
 
             // first check to see if the user has specified a parser
-            if (data.loadParser)
+            if (data.parser || data.loadParser)
             {
```
bigtimebuddy, line 74: Can you add a deprecation log for using loadParser?

bigtimebuddy (approved):
Looks great, so much easier to remember! Thanks for doing this.

PR #11563, opened 2025-07-17
chore: update issue templates

Zyie:
Updates the bug reporting process by introducing a bug reproduction guideline document and updates the bug report issue template.

This idea here was to better enforce a minimal reproduction when an issue is created.


bigtimebuddy (approved):
Looks great. Just a small concern about PixiPlayground

Issue #11545, opened 2025-07-16
Feature Request: Use `DOMAdapter.get().Image` instead of `globalThis.Image`


UlyssesZh:
### Description

Many places use something like `new Image()`, which does not work in Node.js environment, leading to bugs such as pixijs-userland/node#4. We can fix this by using the `DOMAdapter`.

Zyie:
Hey @UlyssesZh, this is a good idea

What would be used in pixi node as a replacement for `Image`?

UlyssesZh:
```js
import { Image } from 'canvas'
```

UlyssesZh:
Also there are a bunch of `document.createElement('canvas')` being used. Maybe replace with `DOMAdapter.get().createCanvas()`?

PR #11565, opened 2025-07-18
feat: use DomAdapter for new Image

Zyie:
Fixes: #11545

Ensures image creation uses the environment adapter. This change standardizes image creation across different environments (browser, web worker, node) by using the adapter pattern


src/environment/adapter.ts
```diff
@@ -14,6 +14,8 @@ export interface Adapter
 {
     /** Returns a canvas object that can be used to create a webgl context. */
     createCanvas: (width?: number, height?: number) => ICanvas;
+    /** Returns an HTMLImageElement that can be used to create a texture. */
+    createImage: () => HTMLImageElement;
```
bigtimebuddy, line 18: Do we not need an abstraction here like ICanvas?

UlyssesZh:
```typescript
/**
 * Interface for HTMLImageElement.
 * @category environment
 * @advanced
 */
export interface IImage extends EventTarget
{
    /** Whether or not the image has completely loaded. */
    readonly complete: boolean;

    /** The Cross-Origin Resource Sharing (CORS) setting to use when retrieving the image. */
    crossOrigin: string | null;

    /** The URL of the image which is currently presented in the <img> element it represents. */
    readonly currentSrc: string;

    /** The width. */
    width: number;

    /** The height. */
    height: number;

    /** The address or URL of the a media resource that is to be considered. */
    src: string;

    /** Returns a Promise that resolves once the image is decoded. */
    decode(): Promise<void>;

    onload: ((this: GlobalEventHandlers, ev: Event) => any) | null;
    onerror: ((this: GlobalEventHandlers, ev: Event) => any) | null;
}
```

Zyie:
@UlyssesZh Thanks for providing the `IImage` abstraction!

UlyssesZh:
I think some types also needs to be changed from `HTMLImageElement` to `IImage`. For example, in `ICanvasRenderingContext2D.createPattern` and `ICanvasRenderingContext2D.drawImage`, change the type of `image` argument to `CanvasImageSource | ICanvas | IImage`; in `ExtractSystem.image`, change the return type to `Promise<IImage>`. Type casting `as HTMLImageElement` is then unnecessary. Some documentation needs to be changed accordingly.

bigtimebuddy (approved):
Very helpful. The double-I bothers me, but I don't have anything better to offer for names.

One thing that might be worth considering in a follow up is disallowing except in DOMAdpater things like `document.createElement` or `new Image` with some lint rules. This is an easy thing to regress and we don't have a good tests it.

Zyie:
> The double-I bothers me, but I don't have anything better to offer for names.

I also don't like the double-I
What about any of these?:

`IWebImage`
`IHTMLImage`
`IHTMLImageElement`

bigtimebuddy:
`ImageLike`? More like `PointLike` but less like `ICanvas`

Zyie:
> `ImageLike`? More like `PointLike` but less like `ICanvas`

I like `ImageLike`, should we also consider deprecating `ICanvas` for `CanvasLike`?
I can do a follow up PR for that if so

bigtimebuddy:
Yeah, good call on CanvasLike

Issue #11556, opened 2025-07-17
Bug: Text trim doesn't work correctly if resolution of text set to 2


verkkis-dev:
### Current Behavior

At the moment if you set a resolution of text with trim to 2, text is not shown correctly.

### Expected Behavior

Text is in correct position

### Steps to Reproduce

https://www.pixiplayground.com/#/edit/1_O-iPcfJaF7ueK7CK1P2

### Environment

- **`pixi.js` version**: latest
- **Browser & Version**: Chrome latest
- **OS & Version**: Windows 11
- **Running Example**: https://www.pixiplayground.com/#/edit/1_O-iPcfJaF7ueK7CK1P2

PR #11570, opened 2025-07-20
fix: trimmed text rendereing with non default resolution

mayakwd:
Fixes: #11556

##### Description of change
The trimmed frame is now scaled adequately to the expected size.
image

##### Pre-Merge Checklist


Zyie:
example of fix: https://stackblitz.com/edit/zppxjesi?file=src%2Findex.js

PR #11571, opened 2025-07-22
chore: Enables StackBlitz integration

Zyie:
Refactors the project to support StackBlitz instead of CodeSandbox due to instability.

- Migrates from CodeSandbox to StackBlitz.
- Updates build and publish scripts for StackBlitz compatibility.
- Adds StackBlitz configuration files.


Zyie:
> What happened with codesandbox?

For the past couple of months, it’s been crashing about 75% of the time when I open a link.

PR #11577, opened 2025-07-24
fix: alpha mask filter broken with render group cache-as-texture

silenaker:
##### Description of change

1. The `AlphaMaskPipe` calculates the bounds of the mask using world coordinates instead of the local coordinates of the cache-as-texture render group.
2. `getFastGlobalBounds` should calculate local bounds only for render groups; otherwise, it may cause unpredictable bugs.
3. The `FilterSystem` applies `clipToViewport` with the screen root viewport instead of the cache-as-texture render group's viewport, causing unexpected bugs.
4. The `Pool` does not account for duplicate object returns, leading to erratic bugs and unit test failures.
5. The `TexturePool#getOptimalTexture` method does not dispatch texture update event when the texture frame is updated. As a result, the `TextureMatrix` that depends on it fails to timely update its state (`TextureMatrix#mapCoord`), leading to errors when calculating `uFilterMatrix` uniform in `MaskFilter#apply`.
6. Unlike `StencilMask`, `AlphaMask` should not have `addBounds` and `addLocalBounds` methods, as they lead to incorrect bounds calculation on masked objects. No changes were committed since users can implement a new AlphaMask class to fix this.

##### Pre-Merge Checklist

### DIFF
--- a/src/filters/FilterSystem.ts
+++ b/src/filters/FilterSystem.ts
@@ -230,11 +230,13 @@ export class FilterSystem implements System
         if (instruction.container)
         {
             const renderGroup = instruction.container.renderGroup || instruction.container.parentRenderGroup;
+            const closestCacheAsTexture = renderGroup.isCachedAsTexture
+                ? renderGroup : renderGroup._parentCacheAsTextureRenderGroup;
 
-            if (renderGroup._parentCacheAsTextureRenderGroup)
+            if (closestCacheAsTexture)
             {
-                const frame = renderGroup._parentCacheAsTextureRenderGroup.texture.frame;
-                const resolution = renderGroup._parentCacheAsTextureRenderGroup.texture._source._resolution;
+                const frame = closestCacheAsTexture.texture.frame;
+                const resolution = closestCacheAsTexture.texture._source._resolution;
 
                 viewport = new Rectangle();
                 viewport.x = ((frame.x * resolution) + 0.5) | 0;


Zyie:
hey @silenaker are you able to take a look at the failing visual tests?

silenaker:
> hey @silenaker are you able to take a look at the failing visual tests?



> hey @silenaker are you able to take a look at the failing visual tests?

The latest commit passed the visual tests locally, but it's still failing in the automated pipeline. Could you help me run it to verify? thanks

src/filters/FilterSystem.ts
```diff
@@ -225,7 +225,23 @@ export class FilterSystem implements System
 
         this._calculateFilterArea(instruction, bounds);
 
-        this._calculateFilterBounds(filterData, renderer.renderTarget.rootViewPort, rootAntialias, rootResolution, 1);
+        let viewport = renderer.renderTarget.rootViewPort;
+
+        const renderGroup = instruction.container.renderGroup || instruction.container.parentRenderGroup;
```
mayakwd, line 230: `instruction.container` is not always available. This line will break execution and produce "TypeError".
`FilterInstruction` contains either container or renderables.

mayakwd (commented):
Overall, this fix appears to be a solution for a specific case involving one filter. Of course, the mask is a specific case to begin with, so there's nothing wrong with that.

However, it may be worth revisiting this pull request and considering all filters and bounds calculations for RenderGroup.

Currently, filters “break” in unexpected places when the user uses `cacheAsTexture`.

silenaker:
> Overall, this fix appears to be a solution for a specific case involving one filter. Of course, the mask is a specific case to begin with, so there's nothing wrong with that.

This isn’t a fix targeting a specific filter. The `FilterSystem` already takes cache-as-texture render groups into account. For it to work correctly in such cases, two conditions must be met. The first—applying the inverse transform to global bounds—has already been handled. However, it missed a subtle but essential detail: the root viewport also needs to be adjusted accordingly. This PR addresses that missing piece.

> However, it may be worth revisiting this pull request and considering all filters and bounds calculations for RenderGroup.

On the surface, this PR fixes an issue where alpha mask doesn’t work properly inside cache-as-texture render groups. But while addressing that, I encountered several deeper issues in core components like `getFastGlobalBounds`, `TexturePool`, and `Pool`, which had hard-to-detect bugs that were difficult to test. The `FilterSystem` was not my only focus—it just happens to be one of the affected components. Given its importance, I took the time to review its implementation as well.

During the review, I noticed several inconsistencies in how it handles nested filters, resolution, and antialias settings. The logic there isn't entirely clear and contains some conflicting intentions. I attempted to address these issues too, but the changes were too extensive and out of scope for this PR.

As a result, I’ve only included the minimal changes necessary to keep this PR focused and stable and I'll create new PRs to address these not directly related issues.

PR #11582, opened 2025-07-25
fix: Remove deprecation version constants

albinkong:
##### Description of change
Currently the pixi.js node package exports 2 constants that represent internal use variables only used when calling the `deprecation` function (`v8_0_0` and `v8_3_4`).

In concept this is not an issue but sometimes my IDE, and I imagine other's as well, decide to autocomplete `0` to `v8_0_0`. As these serve no purpose other than to keep two specific versions as constants I propose they are removed.

##### Pre-Merge Checklist

### DIFF
--- a/src/utils/logging/deprecation.ts
+++ b/src/utils/logging/deprecation.ts
@@ -6,11 +6,13 @@ const warnings: Dict<boolean> = {};
 /**
  * deprecation name for version 8.0.0
  * @ignore
+ * @internal
  */
 export const v8_0_0 = '8.0.0';
 /**
  * deprecation name for version 8.1.0
  * @ignore
+ * @internal
  */
 export const v8_3_4 = '8.3.4';
 


mayakwd:
Hi @albinkong. I understand your concern, but there is a much simpler solution that preserves developer experience for maintainers and does not interfere with the user's library experience. 

You can add `@internal` to ts-doc, and those two constants won't be exported from the package. Such an approach is less invasive and satisfies your concern.

albinkong:
Ah, @mayakwd. I was not aware of this.

I can definitely make the change if that would be preferred. It could even be that the `@ignore` comments in the existing code where added to do what `@internal` does.

mayakwd:
If it's okay for you, let's just mark them as `@internal`.

albinkong:
I've updated the PR. Thanks for the feedback!

Issue #11311, opened 2025-02-28
Bug: Advanced blend modes don't adjust to renderer resolution / window pixel ratio


ax9880:
### Current Behavior

Unlike blend modes like 'add' or 'multiply', advanced blend modes like 'overlay' or 'color-dodge' don't adjust or resize automatically when the renderer resolution or window pixel ratio is different to 1 or powers of two, resulting in the blend mode being applied to only a smaller part of the texture or sprite.

I first noticed this issue when testing a pixi.js program on a phone with a non-standard `window.devicePixelRatio`, but it is also present on desktop if you manually change the resolution.

### Expected Behavior

An advanced blend mode applied to a sprite should cover the whole texture or sprite when the renderer resolution or `window.devicePixelRatio` is different to 1 or powers of two.

### Steps to Reproduce

Go to https://pixijs.com/8.x/examples/basic/blend-modes and replace the code with the following:
```js
import { Application, Assets, Sprite } from 'pixi.js';
import 'pixi.js/advanced-blend-modes';

(async () =>
{
    const app = new Application();

    await app.init({
        antialias: true,
        backgroundColor: '#bbbbbb',
        resizeTo: window,
        useBackBuffer: true,
        resolution: 2.625, // Example of a non-standard resolution, Google Pixel 7
    });

    document.body.appendChild(app.canvas);

    const pandaTexture = await Assets.load(`https://pixijs.com/assets/panda.png`);

    const sprite = new Sprite({
        texture: pandaTexture,
        blendMode: 'overlay'
    });

    app.stage.addChild(sprite);

    // For comparison
    const spriteWithAddBlendMode = new Sprite({
        texture: pandaTexture,
        blendMode: 'add'
    });

    spriteWithAddBlendMode.position.x = sprite.width;

    app.stage.addChild(spriteWithAddBlendMode);
})();
```

You will notice that the part of the left sprite is completely black and has not applied the overlay blend mode.

### Environment

- **`pixi.js` version**: 8.8.1
- **Browser & Version**: Firefox 135
- **OS & Version**: Ubuntu 20
- **Running Example**: See above

### Possible Solution

A workaround is to instantiate the blend mode, set its resolution, and add it to the sprite's filters instead of setting the `blendMode`:
```js
import { Application, Assets, Sprite, OverlayBlend } from 'pixi.js';

// ...

const overlayBlend = new OverlayBlend();
overlayBlend.resolution = app.renderer.resolution;

const sprite = new Sprite({
[...]

PR #11584, opened 2025-07-26
fix: set default blend mode filter's resolution to inherit

privatestefans:
##### Description of change
Blend mode filter now inherit render target's resolution, instead default to 1.
This will fix advanced blend mode like `overlay, color-burn, soft-light` when `devicePixelRatio` or the renderer’s resolution is not 1 or a power of two.

##### Pre-Merge Checklist

##### Related
- Fixes: #11311


bigtimebuddy:
I'm not sure using inherit as the default filter resolution is the best fix. The underlying problem seems like a mismatch between filter and renderer resolution, which we need to support. This fix is masking that problem.

Also, would probably consider this a breaking change since it will increase memory usage unexpectedly for users. I'd, however, be open to reconsidering 1 as the default filter resolution in the next major version.

privatestefans:
Thanks for the feedback!

As far as I know, if we're using `inherit`, it will automatically use the render target's resolution as the filter resolution.
And because `BlendModeFilter` does grab a snapshot of the area where it's being rendered and pass it to the shader, I think the default filter resolution should be `inherit` to ensure the filter matches the resolution of the captured content.

But I understand your concern, and I comes up with 2 solution:
1. Instead of modify the default in the `Filter` class, we can set `inherit` only for `BlendModeFilter`. or maybe...
2. Only change the filter resolution if the back texture `resolution` is not power of two. The filter texture may looks downscaled on final render, but may unnoticeable for device with high device pixel ratio and memory usage won’t be significantly impacted.

I would love to get your input on this.

privatestefans:
Playground to test:
- [Before](https://pixijs.com/8.x/playground?state=ImltcG9ydCB7IEFwcGxpY2F0aW9uLCBBc3NldHMsIENvbnRhaW5lciwgU3ByaXRlLCBUZXh0LCBUZXh0dXJlIH0gZnJvbSAncGl4aS5qcyc7XG5pbXBvcnQgJ3BpeGkuanMvYWR2YW5jZWQtYmxlbmQtbW9kZXMnO1xuXG4oYXN5bmMgKCkgPT4ge1xuICBjb25zdCBhcHAgPSBuZXcgQXBwbGljYXRpb24oKTtcblxuICBhd2FpdCBhcHAuaW5pdCh7XG4gICAgYW50aWFsaWFzOiB0cnVlLFxuICAgIGJhY2tncm91bmRDb2xvcjogJ3doaXRlJyxcbiAgICByZXNpemVUbzogd2luZG93LFxuICAgIHVzZUJhY2tCdWZmZXI6IHRydWUsXG4gICAgcmVzb2x1dGlvbjogMS41LCAvLyBDaGFuZ2UgdGhpcyB0byAxLCAgMS41LCAgMiwgIDIuNiwgIDMsICA0XG4gIH0pO1xuXG4gIGRvY3VtZW50LmJvZHkuYXBwZW5kQ2hpbGQoYXBwLmNhbnZhcyk7XG5cbiAgY29uc3QgbWFpbkNvbnRhaW5lciA9IG5ldyBDb250YWluZXIoKTtcbiAgbWFpbkNvbnRhaW5lci5zY2FsZSA9IDAuODtcbiAgYXBwLnN0YWdlLmFkZENoaWxkKG1haW5Db250YWluZXIpO1xuXG4gIGNvbnN0IHBhbmRhVGV4dHVyZSA9IGF3YWl0IEFzc2V0cy5sb2FkKGBodHRwczovL3BpeGlqcy5jb20vYXNzZXRzL3BhbmRhLnBuZ2ApO1xuICBjb25zdCByYWluYm93R3JhZGllbnQgPSBhd2FpdCBBc3NldHMubG9hZChgaHR0cHM6Ly9waXhpanMuY29tL2Fzc2V0cy9yYWluYm93LWdyYWRpZW50LnBuZ2ApO1xuXG4gIGNvbnN0IGFsbEJsZW5kTW9kZXMgPSBbXG4gICAgJ25vcm1hbCcsXG4gICAgJ2FkZCcsXG4gICAgJ3NjcmVlbicsXG4gICAgJ2RhcmtlbicsXG4gICAgJ2xpZ2h0ZW4nLFxuICAgICdjb2xvci1kb2RnZScsXG4gICAgJ2NvbG9yLWJ1cm4nLFxuICAgICdsaW5lYXItYnVybicsXG4gICAgJ2xpbmVhci1kb2RnZScsXG4gICAgJ2xpbmVhci1saWdodCcsXG4gICAgJ2hhcmQtbGlnaHQnLFxuICAgICdzb2Z0LWxpZ2h0JyxcbiAgICAncGluLWxpZ2h0JyxcbiAgICAnZGlmZmVyZW5jZScsXG4gICAgJ2V4Y2x1c2lvbicsXG4gICAgJ292ZXJsYXknLFxuICAgICdzYXR1cmF0aW9uJyxcbiAgICAnY29sb3InLFxuICAgICdsdW1pbm9zaXR5JyxcbiAgICAnYWRkLW5wbScsXG4gICAgJ3N1YnRyYWN0JyxcbiAgICAnZGl2aWRlJyxcbiAgICAndml2aWQtbGlnaHQnLFxuICAgICdoYXJkLW1peCcsXG4gICAgJ25lZ2F0aW9uJyxcbiAgXTtcblxuICBjb25zdCBzaXplID0gODAwIC8gNTtcblxuICBjb25zdCBwYW5kYXMgPSBbXTtcblxuICBmb3IgKGxldCBpID0gMDsgaSA8IGFsbEJsZW5kTW9kZXMubGVuZ3RoOyBpKyspIHtcbiAgICBjb25zdCBjb250YWluZXIgPSBuZXcgQ29udGFpbmVyKCk7XG5cbiAgICBjb25zdCBzcHJpdGUgPSBuZXcgU3ByaXRlKHtcbiAgICAgIHRleHR1cmU6IHBhbmRhVGV4dHVyZSxcbiAgICAgIHdpZHRoOiAxMDAsXG4gICAgICBoZWlnaHQ6IDEwMCxcbiAgICAgIGFuY2hvcjogMC41LFxuICAgICAgcG9zaXRpb246IHsgeDogc2l6ZSAvIDIsIHk6IHNpemUgLyAyIH0sXG4gICAgfSk7XG5cbiAgICBwYW5kYXMucHVzaChzcHJpdGUpO1xuXG4gICAgY29uc3Qgc3ByaXRlMiA9IG5ldyBTcHJpdGUoe1xuICAgICAgdGV4dHVyZTogcmFpbmJvd0dyYWRpZW50LFxuICAgICAgd2lkdGg6IHNpemUsXG4gICAgICBoZWlnaHQ6IHNpemUsXG4gICAgICBibGVuZE1vZGU6IGFsbEJsZW5kTW9kZXNbaV0sXG4gICAgfSk7XG5cbiAgICBjb250YWluZXIuYWRkQ2hpbGQoc3ByaXRlLCBzcHJpdGUyKTtcblxuICAgIGNvbnN0IHRleHQgPSBuZXcgVGV4dCh7XG4gICAgICB0ZXh0OiBhbGxCbGVuZE1vZGVzW2ldLFxuICAgICAgc3R5bGU6IHtcbiAgICAgICAgZm9udFNpemU6IDE2LFxuICAgICAgICBmb250RmFtaWx5OiAnc2hvcnQtc3RhY2snLFxuICAgICAgfSxcbiAgICB9KTtcblxuICAgIC8vIEFkZCBibGVuZCBtb2RlIHRleHQgbGFiZWxzXG4gICAgdGV4dC54ID0gc2l6ZSAvIDIgLSB0ZXh0LndpZHRoIC8gMjtcbiAgICB0ZXh0LnkgPSBzaXplIC0gdGV4dC5oZWlnaHQ7XG4gICAgY29uc3QgdGV4dEJhY2tncm91bmQgPSBuZXcgU3ByaXRlKFRleHR1cmUuV0hJVEUpO1xuXG4gICAgdGV4dEJhY2tncm91bmQueCA9IHRleHQueCAtIDI7XG4gICAgdGV4dEJhY2tncm91bmQueSA9IHRleHQueTtcbiAgICB0ZXh0QmFja2dyb3VuZC53aWR0aCA9IHRleHQud2lkdGggKyA0O1xuICAgIHRleHRCYWNrZ3JvdW5kLmhlaWdodCA9IHRleHQuaGVpZ2h0ICsgNDtcbiAgICBjb250YWluZXIuYWRkQ2hpbGQodGV4dEJhY2tncm91bmQsIHRleHQpO1xuXG4gICAgbWFpbkNvbnRhaW5lci5hZGRDaGlsZChjb250YWluZXIpO1xuXG4gICAgY29udGFpbmVyLnggPSAoaSAlIDUpICogc2l6ZTtcbiAgICBjb250YWluZXIueSA9IE1hdGguZmxvb3IoaSAvIDUpICogc2l6ZTtcbiAgfVxuXG4gIGFwcC50aWNrZXIuYWRkKCgpID0%2BIHtcbiAgICBwYW5kYXMuZm9yRWFjaCgocGFuZGEsIGkpID0%2BIHtcbiAgICAgIHBhbmRhLnJvdGF0aW9uICs9IDAuMDEgKiAoaSAlIDIgPyAxIDogLTEpO1xuICAgIH0pO1xuICB9KTtcbn0pKCk7XG4i)
- [After](https://stackblitz.com/edit/bo1x8rvw?file=src%2Findex.js)

bigtimebuddy:
The context for why we chose 1 as the default resolution instead of the renderer resolution was more about performance. Filters generally hurt runtime performance in two ways: deoptimize batching and can use a lot of memory. For Pixi, we generally set defaults that are designed to perform the best sometimes at the expensive of fidelity. This was the reasons we chose resolution: 1.

I'd like @GoodBoyDigital or @Zyie to weigh-in on this one.

bigtimebuddy:
@privatestefans Thanks for your patience here. I had a chat with the team and this is what we decided:

* We agreed that `"inherit"` _should_ be the new default `resolution` for _all_ Filters. However, as it is a breaking change, we feel uncomfortable pushing this change in v8. Our next major release (currently unscheduled) we will be changing this.
* For now, the best way to solve this is education, so we'd be in favor of telling users, maybe in the docs, that setting `Filter.defaultOptions.resolution = "inherit"` before creating any filters is the best way to be future-compatible.

privatestefans:
Thanks for taking the time to discuss with the team.
I understand the decision, and it sounds like a thoughtful approach.

I think for this PR, I'll revert to my original proposal so it can be merged later or for future reference.

Apologies for the delayed response 😅

PR #11585, opened 2025-07-27
fix: ensure correct parser for blob image paths in bitmap font loader

ragePowered:
#### Description of change
This pull request extends BitmapFont support in PixiJS 8 to allow using Blob URLs as texture sources.

Previously, loadBitmapFont relied solely on file extensions to determine the appropriate loader, which fails for blob URLs (since they have no extension).
This change assumes that any file referenced within the .fnt file is a texture (e.g. PNG, JPG, WEBP, AVIF, or a Blob representing one of these formats), and treats it accordingly.

#### Potential impact:
This may affect cases where compressed textures (e.g. .ktx, .basis, etc.) are used as .fnt texture pages. Their handling is not explicitly addressed here and might require additional logic.

#### Pre-Merge Checklist

### DIFF
--- a/src/scene/text-bitmap/asset/loadBitmapFont.ts
+++ b/src/scene/text-bitmap/asset/loadBitmapFont.ts
@@ -96,7 +96,8 @@ export const loadBitmapFont = {
 
             textureUrls.push({
                 src: imagePath,
-                data: textureOptions
+                data: textureOptions,
+                parser: imagePath.startsWith('blob:') ? 'texture' : undefined
             });
         }
 


src/scene/text-bitmap/asset/loadBitmapFont.ts
```diff
@@ -96,7 +96,8 @@ export const loadBitmapFont = {
 
             textureUrls.push({
                 src: imagePath,
-                data: textureOptions
+                data: textureOptions,
+                parser: imagePath.startsWith('blob:') ? 'texture' : undefined
```
bigtimebuddy, line 100: It would be much more useful if users could declare the parser to use rather than assuming every blob is a texture. 

```ts
Assets.load({ src: "path/to/font.fnt", data: { imageParser: "ktx" } })
```

This way it would easily support ktx, basis and other compressed texture types.

Zyie:
going to close this one down as there hasn't been any activity from the author in a while

Issue #11597, opened 2025-08-04
Container.cullArea uses global coordinates, despite docs saying it uses container's local space coordinates


jujurocket:
### PixiJS version

8.11.0

### Link to minimal reproduction

https://www.pixiplayground.com/#/edit/KdXGl2p6By75TFZh5OYTS

### Steps to reproduce

- turn on culling
- add a child sprite
- set child to cullable
- set child cullArea to `new Rectangle(0, 0, 10, 10)` for example
- place child way out of view of the scene, for example `child.position.x = 10000`
- `child.culled` will still be `false`

### What is expected?

When culling, for every container's `cullArea` to be transformed from local container coordinates to global coordinates before doing cull checking against the view.

[As per the docs:](https://pixijs.download/release/docs/scene.Container.html#cullarea)

> Custom shape used for culling calculations instead of object bounds. Defined in local space coordinates relative to the object.

### What is actually happening?

Every container's `cullArea` is not transformed, and therefore the cullArea is not relative to the container. 

Which means  the cullArea will not move or transform when its corresponding container moves or transforms, unless the cullArea is updated manually on every change.

This is usually not desirable as if an object moves out of view, you would expect the cullArea to operate based on that movement.

### Environment

All

### Any additional comments?

Even if the docs are wrong and this is indeed the desired behaviour and design (which seems unlikely to me), could we at least have the option to specify that the cullArea should work relative the the coordinates local space?

jujurocket:
I have submitted my version of a fix here: https://github.com/pixijs/pixijs/pull/11598

PR #11598, opened 2025-08-04
project container.cullArea to global coordinates before doing cull check

jujurocket:
##### Description of change

fixes https://github.com/pixijs/pixijs/issues/11597

When an object's `cullArea` is defined, during culling the cullArea will be transformed from local coordinate space to global space ([which it should do according to the docs](https://pixijs.download/release/docs/scene.Container.html#cullarea)).

I fixed it in a bit of a crude manner, so I might have missed some usecases. Please feel free to make changes or suggest any.

It is currently failing the following test in `Culler.test.ts`:
```
'cullable container should be rendered if bounds intersects the frame'
```

If this is indeed the desired behaviour then I will go ahead and change the test as well.

##### Pre-Merge Checklist

- [] Tests passed (`npm run test`)


mayakwd:
@jujurocket , have you had a chance to look into failing unit tests?

src/culling/Culler.ts
```diff
@@ -96,7 +97,34 @@ export class Culler
     {
         if (container.cullable && container.measurable && container.includeInBuild)
         {
-            const bounds = container.cullArea ?? getGlobalBounds(container, skipUpdateTransform, tempBounds);
+            let bounds = new Bounds();
```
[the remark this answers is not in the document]
jujurocket: Thanks for this suggestion, I have applied it. This was useful for me to learn how PIXI keeps on top of it's memory allocation, thank you!

jujurocket:
The test `"cullable container with cullArea should be rendered if the bounds intersect the frame"` was failing because my previous solution failed to account for rotation or other transforms applied to the parent container other than just translation.

While I was trying to figure it out however I took the time to create a test scene to visualise it:
https://www.pixiplayground.com/#/edit/PYOm13pQio3looudvu51n

The tiny red pixel in the top left is how the cullArea behaved before my fix. Because it was global, its position and rotation were unaffected by the position and rotation of its parent. It tests for `container.culled === false` and so it passes because the cullArea is intersecting the culling view rectangle by one pixel.

The green rectangle is how it behaves after my fix, and as you can see it just about touches the corner of the view area.

Was this test written as an edge case by design? The cull area really does only just barely touch the view rectangle corner.

src/culling/Culler.ts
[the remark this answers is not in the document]
jujurocket: I realised that my solution did not work for all kinds of transforms applied to the parent container, have adjusted my approach to apply the container's transforms to the cullArea rectangle.

I tried to apply your suggestion and I think I managed to avoid unnecessary allocations but please let me know if I missed anything.

mayakwd (approved):

mayakwd:
Thank you @jujurocket!

src/culling/Culler.ts
```diff
@@ -96,13 +99,28 @@ export class Culler
...
+                tempRectangle.height = view.height;
 
-            // check view intersection..
-            container.culled = bounds.x >= view.x + view.width
-                || bounds.y >= view.y + view.height
-                || bounds.x + bounds.width <= view.x
-                || bounds.y + bounds.height <= view.y;
+                container.culled = !tempRectangle.intersects(
+                    container.cullArea,
+                    container.getGlobalTransform(Matrix.shared, skipUpdateTransform)
```
GoodBoyDigital, line 111: hello! we should be able to grab the transform and avoid getGlobalTransform  via 

`container.worldTransform` <- this should be a ready to go global matrix 👍

jujurocket: @GoodBoyDigital could we get a follow-up on this? Happy to go either way, just wanted to make sure this flag is not being overlooked

GoodBoyDigital: heya - sorry for the later reply! slammed at the mo :D 
it should yes, my thinking is that al the transforms should have already happened and we don't need to recalcualte them?

jujurocket: no worries!

Ok I am not as familiar with the codebase so happy to go with your suggestion, was just making sure it was not being overlooked 👍

jujurocket:
@GoodBoyDigital applied your suggestion and it seems to work the same 👍 

to you and @mayakwd thank you both very much for your help!

mayakwd:
Everything is green, well done @jujurocket! 
@GoodBoyDigital, do you think we can merge it?

jujurocket:
@GoodBoyDigital please let me know if there's anything I can do to help get this one over the line 🙂

PR #11606, opened 2025-08-12
fix: avoid instanceof

EliasWatson:
##### Description of change

Originated from https://github.com/pixijs/pixi-react/pull/621

The `instanceof` operator does not work across JS realms (aka a pop-out window). I have based my changes on https://jakearchibald.com/2017/arrays-symbols-realms/.

This involves a lot of repetitive code, so I was hoping I could do this using a TypeScript decorator or a build plugin, but both would cause the `is<Class>` method to not appear in the type definition from what I could tell.

##### TODO
- [x] Replace all `instanceof` usages for pixi.js classes
- [ ] Replace all `instanceof` usages for built-in classes
- [ ] Create a test that fails on the `main` branch and ensure it passes for this PR

##### Pre-Merge Checklist

### DIFF
--- a/src/accessibility/AccessibilitySystem.ts
+++ b/src/accessibility/AccessibilitySystem.ts
@@ -11,6 +11,8 @@ import type { Renderer } from '../rendering/renderers/types';
 import type { Container } from '../scene/container/Container';
 import type { isMobileResult } from '../utils/browser/isMobile';
 
+const typeSymbol = Symbol.for('pixijs.AccessibilitySystem');
+
 /** @ignore */
 const KEY_CODE_TAB = 9;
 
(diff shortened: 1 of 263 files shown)


mayakwd:
Hi! That's an interesting suggestion. 

Could you please elaborate on a real world problem, to justify these changes?

It is hard to imagine that one by any reason decide to pass the AccessibilitySystem instance from iframe to the parent.

These changes feels a bit artificial.

bigtimebuddy:
@EliasWatson thanks so much for this suggestion and PR. We talked about it as a team and feel like this approach, agreeing with @mayakwd, is really heavy-handed (~4700+ lines) and difficult to maintain for the solve it provides. Going to close this PR, but are open to lighter-touch approaches that could achieve the same thing that don't require decorating every class in the codebase.

PR #11616, opened 2025-08-14
fix: reset bounds dirty flag in GraphicsContext to prevent recalculations on each access

mayakwd:
##### Description of change
Currently, every attempt to access the Graphics(GraphicsContext) boundaries causes them to be recalculated, regardless of whether the content has changed.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good catch!

Issue #9491, opened 2023-06-21
Bug: deprecation implementation is messing up remote logging systems


pie6k:
### Current Behavior

Current behavior of `deprecation` function uses console.groupCollapsed, which uses nicely styled text and then logs stack with ~3 last parts of the stack truncated:

```ts
        if (console.groupCollapsed)
        {
            console.groupCollapsed(
                '%cPixiJS Deprecation Warning: %c%s',
                'color:#614108;background:#fffbe6',
                'font-weight:normal;color:#614108;background:#fffbe6',
                `${message}\nDeprecated since v${version}`
            );
            console.warn(stack);
            console.groupEnd();
        }
        else
        {
            console.warn('PixiJS Deprecation Warning: ', `${message}\nDeprecated since v${version}`);
            console.warn(stack);
        }
```

I understand and appreciate the intention of implementing it this way, however it is causing quite a lot of hussle on our side, especially with remote crash analytics such as Sentry.

I also believe this, as a part of a library breaks a common sense rule to use native tooling such as console.warn and native behavior of it (where console.warn by itself manages showing stack trace)

Result in our crash analytics:

<img width="312" alt="CleanShot 2023-06-21 at 02 10 30@2x" src="https://github.com/pixijs/pixijs/assets/7311462/8cc3f425-8ad8-4e29-a411-c92a2c99ffec">

In general our warnings registry has literally thousands of misterious logs starting with `at <HERE SOMETHING>` without descriptive messages.

After digging a bit, here are my findings:
- groupCollapsed is not properly reported at all (it is mostly Sentry fail, but it results in the actual message not being reported at all which makes debugging quite painful, especially initially where I had no idea what is it)
- then stack without 3 last items is reported as a content of the warning. I believe this is a bad practice. I think it should rely on native console.warn stack reporting. Content of the warning should be 'static', whereas if stack trace is part of the warning content - it is quite dynamic, resulting in hundreds or thousands of new warning entries in crash analytics. It also means each time I release a new update of my app - I re-bundle it, so the app is minified differently, so the names of functions are different and as a result - logged warning content is different -> resulting in new warning entries in crash analytics after every app update, even if I already marked them as resolved / ignored in Sentry.

### Expected Behavior

I'd expect at least an opt-in option to use regular console.warning instead:

- without custom HTML or formatting tricks
- without using groupCollapsed
- relying on native console.warn stack reporting instead of logging it manually as warning (I know it will not have ~3 last stack steps excluded, which is fine)

[...]

Issue #10216, opened 2024-02-15
Feature Request: Make deprecation() warnings optional


reececomo:
### Description

I totally understand the motivation for noisy loud warnings, but it would be fabulous if these warnings could be toggled off somehow. It's very noisy in the clients.

Relates to https://github.com/pixijs/pixijs/issues/9491

reececomo:
![Screenshot 2024-02-15 180239](https://github.com/pixijs/pixijs/assets/7390781/d2072a05-9203-47bb-8460-da3082759589)

inb4 update your code

PR #11617, opened 2025-08-15
feat: suppression and colorization options for deprecation message

mayakwd:
##### Description of change
Fixes: #9491, #10216

- Added `quiet` and `noColor` options to control global deprecation message behavior for more flexible usage.
- Replaced `warn` method with `deprecation` in `TilingSprite`.

I chose names `quiet` and `noColor` instead of `silence` and `simple` (suggested in the issue), just because they are pretty common in software development (especially in CLI).  

Introduced API:
```ts
import { deprecation } from 'pixi.js';
// Supresses deprecation warnings
deprecation.quiet = true;
// Removes color formatting from deprecation messages
deprecation.noColor = true;
```

##### Pre-Merge Checklist

### DIFF
--- a/src/utils/logging/deprecation.ts
+++ b/src/utils/logging/deprecation.ts
@@ -32,7 +32,7 @@ interface DeprecationOptions
      * @default true
      * @standard
      */
-    colorful: boolean;
+    noColor: boolean;
 }
 
 /** @internal */
@@ -40,7 +40,7 @@ export type DeprecationFn = ((version: string, message: string, ignoreDepth?: nu
 
 const deprecationState: DeprecationOptions = {
     quiet: false,
-    colorful: true
+    noColor: true
 };
 
 /**
@@ -73,7 +73,7 @@ export const deprecation: DeprecationFn = ((version: string, message: string, ig
     let stack = new Error().stack;
 
     const deprecationMessage = `${message}\nDeprecated since v${version}`;
-    const useGroup = typeof console.groupCollapsed === 'function' && deprecationState.colorful;
+    const useGroup = typeof console.groupCollapsed === 'function' && deprecationState.noColor;
 
     // Handle IE < 10 and Safari < 6
     if (typeof stack === 'undefined')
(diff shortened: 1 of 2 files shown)


bigtimebuddy (commented):
Did you consider adding the deprecation settings directly on the function? Could be simpler without the defineProperties or the extra types:

```ts
/**
 * @standard
 */
export function deprecation() {
    // Ignore duplicate
    if (deprecation.quiet || warnings[message])
    {
        return;
    }
}

/**
 * Indicates whether deprecation messages should be formatted with colors or as plain text.
 * @default true
 * @standard
 */
deprecation.colorful = true;

/**
 * Indicates whether deprecation messages must be suppressed.
 * @default false
 * @standard
 */
deprecation.quiet = false;
```

bigtimebuddy:
Nit, for flags like this, IMO it's more intuitive to opt-in (set to `true`) to the non-default behavior rather than opting-out (setting to `false`). i.e., `colorful` having to set to `false`. Another option would be inverting it to be `noColor`, similar as the ANSI `NO_COLOR=1` environment variable.

Otherwise, I like the name `quiet`. Shorter and more standard.

mayakwd:
> Did you consider adding the deprecation settings directly on the function? Could be simpler without the defineProperties or the extra types

That's an option that can reduce the boilerplate. The only issue with such implementation is that "quiet" and "colorful" will be exported as "vars" and part of the "deprecation" namespace (the type system will merge that with the exported function). 

This is fine on its own, but ts-docs will not be exported to d.ts, as this case is considered an untyped augmentation used for backward compatibility with JS.

> Nit, for flags like this, IMO it's more intuitive to opt-in (set to true)

Fully agree with that, I'll change colorful to "noColor".

src/utils/logging/deprecation.ts
```diff
@@ -16,10 +14,50 @@ export const v8_0_0 = '8.0.0';
...
  * A stack track for warnings is given; useful for tracking-down where
  * deprecated methods/properties/classes are being used within the code.
+ *
+ * Deprecation messages can be configured globally:
+ * ```ts
+ * // Suppress all deprecation messages
+ * deprecation.quiet = true;
+ *
+ * // Put plain text to console instead of colorful messages
+ * deprecation.colorful = false;
```
bigtimebuddy, line 59:
```suggestion
 * deprecation.noColor = true;
```

PR #11631, opened 2025-08-22
feat: Use browser garbage collection to cleanup images and textures

Mysak0CZ:
##### Description of change

According to both WebGL and WebGPU specifications, the underlying device resources are released also when matching JS object obtained though their APIs are GC'd.
This PR changes the system abstraction layers to use `WeakMap` with `Buffer`/`TextureSource` objects as keys, instead of their unique numeric ids. As such it allows JS to garbage collect the GL/GPU buffers/textures when the Pixi `Buffer`/`TextureSource` is garbage collected, even if `.destroy()` was not properly called.

This change doesn't affect current behavior - explicit destroy is still used in all cases where Pixi's object is explicitly destroyed.

The only exception is `destroyAll()` on the systems, as it can no longer iterate through the weak map to destroy all things without holding a strong reference to them (preventing garbage collection), but I didn't find this method used anywhere, and in fact it crashes in most cases before this change (to reproduce crash create object, destroy it manually, and then call `destoryAll()`.

This change should fix all GPU-resource memory leaks - both PIXI-caused and from user errors, but I didn't test specific issues; only on private projects.

##### Pre-Merge Checklist


dragonbaba:
1. Under the v8 engine specification, querying and storing using weakmap is much slower than an object without a prototype 
2. Even if you use weakmap here, then it is only dereferenced in various systems, and it does not guarantee that all buffers and textures are not referenced elsewhere. Modern garbage collection systems are no longer tagged by count references, but are only collected when an object is completely inaccessible. 
3. Memory leaks need to be troubleshoot for multiple reasons, which is too complicated

Mysak0CZ:
Hello @dragonbaba
> 1. Under the v8 engine specification, querying and storing using weakmap is much slower than an object without a prototype

Can I ask for a link to that part of specification that says that? Also I would note that `WeakMap` is not like `WeakRef`, which relies on hooking into GC logic.
That said if you make a benchmark I can also rework this to store the objects using a `Symbol` on the logical objects, which will achieve almost the same result (except making `destroyAll` methods not possible at all), Or using a `FinalizationRegistry`.

> 2. Even if you use weakmap here, then it is only dereferenced in various systems, and it does not guarantee that all buffers and textures are not referenced elsewhere. Modern garbage collection systems are no longer tagged by count references, but are only collected when an object is completely inaccessible.

There are two points to address here:
I use WeakMap in these places, as they are the only places that directly hold a reference to underlaying WebGL/WebGPU resources. Pixi.js itself doesn't reference these buffer or texture objects elsewhere. This means that unless the library consumer manually digs into the systems and accesses these _private_ properties, there will not be any more references to them and they will get GC'd after the matching logical `Buffer`/`TextureSource` class is GC'd.

Second part is about refcounts. First of all you imply that refcounting can collect objects even while they are accessible, which isn't true. Object being reachable implies it has non-zero refcount, so it won't be collected - refcounts have the opposite problem of objects in cycles not getting collected as they have non-zero refcount even if not actually reachable from GC roots.
I am, however, not sure how this has anything to do with this PR.

> 3. Memory leaks need to be troubleshoot for multiple reasons, which is too complicated

Yes, memory leaks are troublesome to troubleshoot. This is especially true if language/API requires an explicit operation to free some resource. For normal objects in JS that isn't the case - loosing all references is enough to let GC handle it.
In this case, however, simply creating e.g. a `Mesh` object will allocate resources, but only loosing it would not free all associated resources - you need to explicitly call `.destroy()` to clean it up.
This PR updates this behavior such that the explicit `.destroy()` call is no longer required for cleaning up resources held on the GPU side of things (which is often more limiting than any resources in RAM). So while it doesn't guarantee there won't be any leaks without proper calls to `.destroy()`, it is certainly a step that moves Pixi's resource logic in the direction JS programmers are more used to - GC doing its work instead of need for an explicit resource management.

dragonbaba:
> 你好
> 
> > 1. 在 v8 引擎规范下，使用 weakmap 查询和存储比没有原型的对象慢得多
> 
> 我可以要求提供规范中说明这一点的那部分的链接吗？另外，我还要指出，它不像依赖于 GC 逻辑。也就是说，如果你做一个基准测试，我也可以重新设计它以在逻辑对象上使用 a 来存储对象，这将实现几乎相同的结果（除了使方法根本不可能），或者使用 .`WeakMap``WeakRef``Symbol``destroyAll``FinalizationRegistry`
> 
> > 2. 即使你在这里使用weakmap，那么它也只会在各种系统中被取消引用，并且不能保证所有的缓冲区和纹理都不会在其他地方被引用。现代垃圾回收系统不再由计数引用标记，而是仅在对象完全无法访问时才被收集。
> 
> 这里有两点需要解决：我在这些地方使用 WeakMap，因为它们是唯一直接包含对底层 WebGL/WebGPU 资源的引用的地方。Pixi.js本身不会在其他地方引用这些缓冲区或纹理对象。这意味着，除非库使用者手动挖掘系统并访问这些_私有_属性，否则不会再有对它们的引用，并且在匹配的逻辑/类被 GC 后，它们将得到 GC。`Buffer``TextureSource`
> 
> 第二部分是关于引用计数的。首先，你暗示引用计数可以收集对象，即使它们是可访问的，但事实并非如此。对象可访问意味着它具有非零的引用计数，因此它不会被收集 - 引用计数有相反的问题，即循环中的对象不会被收集，因为它们具有非零引用计数，即使实际上无法从 GC 根访问。然而，我不确定这与这个公关有什么关系。
> 
> > 3. 内存泄漏需要排查，原因多重，太复杂了
> 
> 是的，内存泄漏很难排除。如果语言/API 需要显式作来释放某些资源，则尤其如此。对于 JS 中的普通对象，情况并非如此 - 丢失所有引用足以让 GC 处理它。然而，在这种情况下，简单地创建一个对象将分配资源，但仅丢失它不会释放所有关联的资源 - 您需要显式调用来清理它。此 PR 更新此行为，以便不再需要显式调用来清理 GPU 端保存的资源（这通常比 RAM 中的任何资源都更具限制性）。因此，虽然它不能保证如果没有正确调用 ，就不会有任何泄漏，但这肯定是将 Pixi 的资源逻辑朝着 JS 程序员更习惯的方向移动的一步——GC 完成其工作，而不是需要显式的资源管理。`Mesh``.destroy()``.destroy()``.destroy()`

I may have misunderstood, I did not say that the garbage collection system of modern browsers is a reference count, which is a garbage collection system from a long time ago. When I say that a JavaScript object becomes inaccessible, it means that it has no other JavaScript objects and variables associated with its memory address. I think you can systematically understand what garbage collection is, and there will be primary and secondary garbage collectors in the browser. Here's a link, https://en.wikipedia.org/wiki/Garbage_collection_(computer_science)
That's why I say it doesn't make sense for you to use weakmap, because the problem is really complicated, and just using weakmap doesn't improve them.
If you don't believe me, then you can do some testing.
Test with FinalizationRegistry to see which is more effective between your method and the original pixi method, the FinalizationRegistry object accepts a callback and triggers the current function when the current object is garbage collected
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry

Mysak0CZ:
> I may have misunderstood, I did not say that the garbage collection system of modern browsers is a reference count, which is a garbage collection system from a long time ago. When I say that a JavaScript object becomes inaccessible, it means that it has no other JavaScript objects and variables associated with its memory address. I think you can systematically understand what garbage collection is, and there will be primary and secondary garbage collectors in the browser. Here's a link, [en.wikipedia.org/wiki/Garbage_collection_(computer_science)](https://en.wikipedia.org/wiki/Garbage_collection_(computer_science)) That's why I say it doesn't make sense for you to use weakmap, because the problem is really complicated, and just using weakmap doesn't improve them. If you don't believe me, then you can do some testing. Test with FinalizationRegistry to see which is more effective between your method and the original pixi method, the FinalizationRegistry object accepts a callback and triggers the current function when the current object is garbage collected [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry)

Sure, here is an example that shows how using `WeakMap` here _does_ improve things: https://www.pixiplayground.com/#/edit/01nHQL6ZVFXO33TCk9DJO

<details>
<summary>Content of the playground example</summary>

```js
(async () => {
    await app.init({
        width: window.innerWidth,
        height: window.innerHeight,
        backgroundColor: 0x2c3e50
    })

    // Container for our meshes (this is one scene within the app)
    container.enableRenderGroup();

    // Initialize a few meshes
    // This simulates loading several complex objects in a normal scene
    for (let i = 0; i < 10_000; i++) {
        // Make a texture and a mesh (normally each mesh would have different texture, faking that for testing)
        const texture = await PIXI.Assets.load('https://pixijs.io/examples/examples/assets/bunny.png');
        if ((i % 100) === 0)
            PIXI.Assets.reset(); // Simulate loading differnet assets


        const vertices = new Float32Array(2 * 256);
        const uvs = new Float32Array(2 * 256);
        const indices = new Uint32Array((2 * 3 * 15 * 15));

        // Generate vertices and triangles (usually loaded from model file)
        for (let y = 0; y <= 15; y++) {
            for (let x = 0; x <= 15; x++) {
                const index = 2 * (y * 16 + x);
                vertices[index] = 50 * x;
                vertices[index + 1] = 50 * y;
                uvs[index] = x / 15;
                uvs[index + 1] = y / 15;
            }
        }

        let j = 0;
        for (let y = 0; y < 15; y++) {
            for (let x = 0; x < 15; x++) {
                const vertexA = (y * 16 + x);
                const vertexB = (y * 16 + x + 1);
                const vertexC = ((y + 1) * 16 + x + 1);
                const vertexD = ((y + 1) * 16 + x);

                indices[j++] = vertexA;
                indices[j++] = vertexB;
                indices[j++] = vertexC;
                indices[j++] = vertexC;
                indices[j++] = vertexD;
                indices[j++] = vertexA;
            }
        }

        // Create the mesh
        const mesh = new PIXI.MeshSimple({
            texture,
            vertices,
            uvs,
            indices,
        });
        container.addChild(mesh);
    }

    // Let app render

    // After a while delete the container (e.g. during scene change)
    // Common mistake: Forget to delete the meshes
    setTimeout(() => {
        container.destroy();
    }, 3_000);
})();

```

</details>

Results with current Pixi.js build (v8.12.0) using memory snapshot, filtering for `WebGL`:
image

Results with this patch:
image

dragonbaba:
snapshot
Things seem to be getting interesting, friend, just adding a few lines of code to correct glbuffer leaks without using weakmap
For each mesh class
if(mesh._geometry){
  // destroy gpu buffer  
  mesh._geometry.destroy(true);
}

Mysak0CZ:
> Things seem to be getting interesting, friend, just adding a few lines of code to correct glbuffer leaks without using weakmap For each mesh class if(mesh._geometry){ // destroy gpu buffer mesh._geometry.destroy(true); }

Then it appears you are completely missing the goal of this PR.
What I am trying to do is move in a direction where explicit `.destroy()` call is no longer required (though still possible and beneficial), allowing for more ergonomic use that is closer to how JS is commonly used; especially so in environments where it might be hard to know when a display object will no longer be used and when it is only waiting to be reused.

Explicit resource management is a hard problem after all - which is why we have a Garbage Collection in the first place. So what is the point of having it if the API _requires_ you to do it explicitly anyway?

bigtimebuddy:
@Mysak0CZ Automatic garbage collection of display objects (i.e., not calling `.destroy()`) is a non-goal of the project currently. That approach requires more conversation to align as it would touch a lot of existing choices. I'd suggest starting a discussion thread on the topic where we can weigh the pros/cons and get more input before proposing a solution.

> what is the point of having it if the API requires you to do it explicitly anyway?

Good question! From my experience, we _want_ developers to destroy things they instantiate for mostly one reason: JS's garbage collection sweeps cause noticeable performance hiccups and a manually destroying lifecycle provides a little more control over managing that experience. Devs in the past have even complained about Pixi not preforming well when, in fact, they are not allocating and releasing memory effectively for 60fps.

General recommendations around Garbage Collection for Pixi can be found here: https://pixijs.com/8.x/guides/concepts/garbage-collection

Zyie:
Going to close this on down as we went in a different direction for handling gpu data in recent releases

PR #11632, opened 2025-08-22
fix: svg edges trimmed decimal resolutions

### LINKED ISSUE #11625: Bug: SVG edges being trimmed at particular resolutions (Pixi 8.12.0)
### PixiJS version

8.12.0

https://www.pixiplayground.com/#/edit/m34tGtxHcWeDOis3Xamyy

### Steps to reproduce

Load an SVG with particular decimal data.resolution values

```
const texture = await PIXI.Assets.load({
    alias: 'myAlias1',
    src: fullSvgURI,
    data: {
        resolution: 1 // this works correctly
        // resolution: 1.0125 // this trims the right and bottom edges
        // resolution: 1.03 // this adds extra artefacts to the right and bottom edges
        // resolution: 1.04 // this seems like it is fine?
    }
});
```

### What is expected?

The two circles should both be completely round even though they have different resolutions
Image

The circle on the left has a resolution of 1 and is correct
The circle on the right has a resolution of 1.0125 and has the right and bottom edges trimmed slightly
Image

This is a zoomed in version to see the difference a little easier
Image

### Environment

- **Browser & Version**: *Chrome 138.0.7204.93*, *Firefox 140.0.2*, *Safari 18.5*
- **OS & Version**: *macOS 15.5*

### Other Notes

- Setting `roundPixels: true` or `antialias: true` in `app.init()` appears to have no effect 

```
await app.init({
    width: window.innerWidth,
    height: window.innerHeight,
    backgroundColor: 0x2c3e50,
    roundPixels: true // toggling this to true has no effect
    antialias: true // toggling this to true has no effect
```

- Setting `sprite.roundPixels` seems to also not have any effect

```
const brokenSprite = new PIXI.Sprite(texture2);
brokenSprite.roundPixels = true; // Toggling this to true has no effect
```

### Possible Solution
In `loadSVG.ts` I was able to fork and make the following change from:
```
const canvas = DOMAdapter.get().createCanvas(width * resolution, height * resolution);
const context = canvas.getContext('2d');
context.drawImage(image as CanvasImageSource, 0, 0, width * resolution, height * resolution);
```
to:
```
const roundedWidth = Math.ceil(width * resolution);
const roundedHeight = Math.ceil(height * resolution);
const canvas = DOMAdapter.get().createCanvas(roundedWidth, roundedHeight);
const context = canvas.getContext('2d');
context.drawImage(image as CanvasImageSource, 0, 0, roundedWidth, roundedHeight);
```
This seems to work for my particular scenario. I'm not sure about the performance impact for everyone else. I'm also not sure if this should be tied to `roundPixels: true` so it only happens if you opt-in

creativoma:
##### Description of change
Fixes SVG edges being trimmed when using decimal resolutions (#11625).

Canvas dimensions were being truncated with decimal resolutions (e.g., 1.0125, 1.5x), causing edge trimming. The fix ensures canvas dimensions are always integers by applying `Math.ceil()` to prevent fractional pixel dimensions that cause rendering artifacts.

**Closes #11625**

**Changes:**
- Modified `loadSVG.ts` to always use `Math.ceil()` for canvas dimensions
- Added high-quality image smoothing for better rendering quality
- Simplified logic based on maintainer feedback for better code clarity

##### Pre-Merge Checklist


src/assets/loader/parsers/textures/loadSVG.ts
```diff
@@ -114,9 +114,20 @@ async function loadAsTexture(
     const width = asset.data?.width ?? image.width;
     const height = asset.data?.height ?? image.height;
     const resolution = asset.data?.resolution || getResolutionOfUrl(url);
-    const canvas = DOMAdapter.get().createCanvas(width * resolution, height * resolution);
+
+    // Only apply ceiling and padding for decimal resolutions to avoid affecting integer resolutions
+    const hasDecimalResolution = resolution % 1 !== 0;
```
bigtimebuddy, line 119: I'm not sure I understand this based off the comment. Why not ceil everything? What case is that not desirable?

bigtimebuddy (commented):
Overall, awesome PR and tests, just have one question.

src/assets/loader/parsers/textures/loadSVG.ts
creativoma: We only ceil decimal resolutions (where truncation causes edge trimming) while preserving exact multiplication for integer resolutions that already work correctly. What do you think? 🙂

creativoma: You're absolutely right! It's much easier to always use `Math.ceil()` I just uploaded the change. 😄

bigtimebuddy (approved):

PR #11635, opened 2025-08-24
fix: allow canvas source options for gif asset

### LINKED ISSUE #11633: GifSprite scale mode "nearest" doesn't work
### PixiJS version

8.11.0

https://stackblitz.com/edit/pixijs-v8-ixxfc8wd?file=src%2Fmain.js

### Steps to reproduce

1. Create a GifSource to a gif
2. Create a GifSprite with the source
3. Set `scaleMode` to "nearest"
4. Attach sprite to a container
5. Scale up sprite

### What is expected?

The zoomed pixels should be interpolated via nearest, and look like pixel art.

The zoomed pixels are blurry, like they are interpolated via linear.

### Environment

- Chrome Version 139.0.7258.138 (Official Build) (64-bit)
- Ubuntu 24.04.3

### Any additional comments?

_No response_

bigtimebuddy:
Fixes #11633

### Overview

Currently `scaleMode` is an option for creating GifSprite objects, but is not implemented (looks like it never was). Because normal Sprites don't support `scaleMode` (this is something that lives on the texture source), I have removed this property. The correct way to set source options is through the loader instead.

**Before: (incorrect implementation)**

```ts
import { Assets, GifSprite } from "pixi.js";

const source = await Assets.load("path/to/some.gif");
const sprite = new GifSprite({
  source,
  scaleMode: "nearest"
});
```

**After: (correct implementation)**

```ts
import { Assets, GifSprite } from "pixi.js";

const source = await Assets.load({
  src: "path/to/some.gif",
  data: {
    scaleMode: "nearest"
  }
});
const sprite = new GifSprite(source);
```

### Changes

* Removes `scaleMode` options from GifSpriteOptions 
* Supports CanvasSourceOptions when loading GifAssets
* Fixes incorrect `AssetExtension` type implementation

### DIFF
--- a/src/gif/GifSprite.ts
+++ b/src/gif/GifSprite.ts
@@ -4,8 +4,6 @@ import { UPDATE_PRIORITY } from '../ticker/const';
 import { Ticker } from '../ticker/Ticker';
 import { GifSource } from './GifSource';
 
-import type { SCALE_MODE } from '../rendering/renderers/shared/texture/const';
-
 /**
  * Configuration options for creating a GifSprite instance.
  *
@@ -22,7 +20,6 @@ import type { SCALE_MODE } from '../rendering/renderers/shared/texture/const';
  *     animationSpeed: 1.5,      // 50% faster than normal
  *     loop: true,               // Loop the animation
  *     autoPlay: true,           // Start playing immediately
- *     scaleMode: 'nearest',     // Pixel art style scaling
  *     onComplete: () => {       // Called when non-looping animation ends
  *         console.log('Animation complete!');
  *     }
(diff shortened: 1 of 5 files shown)


src/gif/GifSprite.ts
```diff
@@ -22,7 +20,6 @@ import type { SCALE_MODE } from '../rendering/renderers/shared/texture/const';
  *     animationSpeed: 1.5,      // 50% faster than normal
  *     loop: true,               // Loop the animation
  *     autoPlay: true,           // Start playing immediately
- *     scaleMode: 'nearest',     // Pixel art style scaling
```
mayakwd, line 25: It's a breaking change (minor, yet breaking) for strictly typed projects. 
Maybe we can introduce a deprecation warning and update the documentation, explaining how to apply nearest scale mode on load?

mayakwd: For example, we can add an example to the top level of the class documentation.

bigtimebuddy: Good feedback. Added two examples and a deprecation for scaleMode in options.

mayakwd (approved):

PR #11650, opened 2025-08-31
fix: cast index type to number as it required by setVertexBuffer

katmai7:
##### Description of change
`setVertexBuffer` expect to get a number as index but right now it's a string. 
It works fine in browser, but in ReactNative with react-native-wgpu it leads to an error.

##### Pre-Merge Checklist


bigtimebuddy (approved):
Thanks!

PR #11659, opened 2025-09-04
fix: application destroy for graphics

Zyie:
Originally reported on discord: [message](https://discord.com/channels/734147990985375826/1143191340230914068/1413101088680317030)

Fixes issue with graphics double destroying the batcher. The batcher is already getting tidied up as its part of its own pool

Adds a test case to verify that all children, including graphics, sprites, and text, are destroyed when the destroy method is called with the appropriate options.

### DIFF
--- a/src/app/__tests__/Application.test.ts
+++ b/src/app/__tests__/Application.test.ts
@@ -1,3 +1,8 @@
+import { Texture } from '../../rendering/renderers/shared/texture/Texture';
+import { Graphics } from '../../scene/graphics/shared/Graphics';
+import { Sprite } from '../../scene/sprite/Sprite';
+import { Text } from '../../scene/text/Text';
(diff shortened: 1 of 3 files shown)


bigtimebuddy (approved):

fakob:
Thanks for the fix @Zyie ! The initial errors were fixed by your hot-fix. Though I am still getting these errors which look similar. I lack the deep knowledge of it. What do you think?

Screenshot 2025-09-14 at 11 35 16

Zyie:
> Thanks for the fix @Zyie ! The initial errors were fixed by your hot-fix. Though I am still getting these errors which look similar. I lack the deep knowledge of it. What do you think?

Hey @fakob can you create a new issue with an example of this and I will take a look!

fakob:
I was not able to reproduce this yet in a stackblitz example. Will keep trying. Until then I have rolled back to 8.12.0 where this does not happen.

PR #11581, opened 2025-07-24
fix: memory leaks in renderer systems

Zyie:
Fixes: #11550 #11560 #11573

Addresses memory leaks by:

- Adding new `PoolCollector` singleton that allows us to register pools/caches that can be cleared up on destroy
- Adds new `fullCleanup` option to `renderer.destroy()`
- Adding destroy callbacks to various instruction pipes and systems to ensure proper resource cleanup.
- Clearing references to avoid circular dependencies and facilitate garbage collection.

I'm not sold on the naming of anything here so feel free to give some suggestions

PR #11663, opened 2025-09-06
fix: check if object exists before call destroy on it

katmai7:
##### Description of change
If Graphics cannot be batched, there is will be an error on destroy, because `_batcher ` or `_batch` will be null.

I did review [PR](https://github.com/pixijs/pixijs/pull/11581/files) where it was introduced and look like it's only the place where error can happened.

##### Pre-Merge Checklist

- [] Tests passed (`npm run test`)

### DIFF
--- a/src/scene/graphics/shared/BatchableGraphics.ts
+++ b/src/scene/graphics/shared/BatchableGraphics.ts
@@ -114,9 +114,15 @@ export class BatchableGraphics implements DefaultBatchableMeshElement
         this.texture = null;
         this.geometryData = null;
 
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
This conflicts with #11659 so need to clarify the approach before merging this.

Zyie:
#11659 fixes the underlying issue here. The graphics object shouldn't destroy the batcher, instead the `GlobalResourceRegistry` should handle tidying up pooled batches

Thanks for putting this together but I'm going to close this one down in favor of the other PR!

Issue #11613, opened 2025-08-14
Destroying unrendered prerendered `Text` throws error


darzarzycki:
### PixiJS version

8.12.0

### Link to minimal reproduction

https://codesandbox.io/p/sandbox/cnjwkk

### Steps to reproduce

1. Create a `Text` object
2. Prerender it using prepare plugin
3. Destroy it before it is actually rendered.

One way would be:
```typescript
    const text = new Text({ text: "Some text" });
    await app.renderer.prepare.upload(text);
    text.destroy();
```

Another one:
```typescript
    const text = new Text({ text: "Some text" });
    text.visible = false;
    await app.renderer.prepare.upload(text);
    app.stage.addChild(text);
    requestAnimationFrame(() => {
        text.destroy();
    });
```

### What is expected?

`Text` object can be prerendered and destroyed without the requirement to be actually rendered.

### What is actually happening?

An error is thrown.

### Environment

- **Browser & Version**: *Chrome 139.0.7258.127*
- **OS & Version**: *Windows 11*

PR #11672, opened 2025-09-09
fix: destroy of prepared but not rendered text should not throw an error

mayakwd:
##### Description of change
Fixes: #11613

- BatchableText now verifies that the texture exists before trying to return it.
- PrepareSystem now does not throw an error if it's destroyed before all items from the queue are uploaded.

##### Pre-Merge Checklist

### DIFF
--- a/src/prepare/PrepareBase.ts
+++ b/src/prepare/PrepareBase.ts
@@ -170,7 +170,7 @@ export abstract class PrepareBase
     /** called per frame by the ticker, defer processing to next tick */
     private readonly _tick = () =>
     {
-        if (!this._destroyed) return;
+        if (this._destroyed) return;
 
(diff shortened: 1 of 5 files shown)


mayakwd:
Hm, there is an issue with the Jest timeouts. 
Which is not related (I guess?) to the pull request.

I'm getting them randomly in the local environment as well. I was able to fix it by setting `maxConcurrency` and `maxWorkers` in the Jest config.

mayakwd:
Hm, looks like it's related to my changes. Because tests failed for PrepareSystem only, I'll check.

mayakwd:
What a shame:
- `if (!this._destroyed) return` 
- instead of `if (this.destroyed) return` 😵‍💫

bigtimebuddy:
We've all been there!

PR #11680, opened 2025-09-16
fix: incorrect accessibility div when recycled from the pool (#11679)

### LINKED ISSUE #11679: Bug: Accessibility debug divs use the wrong HTML element type when recycled from the pool (Pixi 8)
### PixiJS version

8.13.2

https://www.pixiplayground.com/#/edit/EqTZ6saVVgjLYonGgTRQE

### Steps to reproduce

- Enable accessibility behaviour

```
await app.init({
    ...
    accessibilityOptions: {
        enabledByDefault: true
    }
});
```

- Create an accessible button (a Container with `accessible: true` and `accessibleType: 'button'`)
- Render the frame
- Disable the button's accessibility (`accessible: false`)
- Render the frame
- Create an accessible header (a Container with `accessible: true` and `accessibleType: 'h1'`)
- Render the frame
- Accessible header will be rendered as a `<button>` instead of a `<h1>`. This also affects how it behaves in macOS Voice Over

In [this demo](https://www.pixiplayground.com/#/edit/EqTZ6saVVgjLYonGgTRQE):
- tab between the buttons
- select and enter press "Button A"
- the mock popup will show, enter press the "Close popup" button
- keyboard and Voice Over navigation will now be broken for "Header B" and "Button A"
- (it will still working correctly for the other element)

https://github.com/user-attachments/assets/8de9115b-4194-46f6-8b21-471f5bd75d54

### What is expected?

When it comes to tab and macOS Voice Over navigation:
- Accessible buttons should always be treated as buttons
- Accessible headers should always be treated as headers

Image

Image

If you create a button, disable it, then create a header:
It incorrectly recycles the disabled `<button>` element and uses it for the header

Image

Image

This video shows how Voice Over navigation breaks too

https://github.com/user-attachments/assets/3ab07486-db72-44c5-afae-4f487b491053

### Environment

- **Browser & Version**: *Chrome 139.0.7258.139*, *Firefox 141.0.3*, *Safari 18.5*
- **OS & Version**: *macOS 15.5*

### Any additional comments?

In AccessibilitySystem, disabled divs are returned to a pool. When a new div is required it pops a div from the pool. The problem occurs if a `<h1>` element is required but a `<button>` is popped from the pool

This is what the elements look like when it breaks. Note how the element for "h1" is a `<button>` and the element for "Button A" is a `<h1>`

Image

This is how it should look

Image

brentmc:
##### Description of change
Fixes screen reader/keyboard navigation when a mixture of HTML elements are used [#11679](https://github.com/pixijs/pixijs/issues/11679)

Previously, all disabled accessibility divs were added to the one pool. It would pop and reuse a `<button>` element and use it for a header instead of creating a `<h1>` element

Now instead of one pool, I've created a Record of pools to pull from
e.g. 
`this._pools['button'].pop()`
`this._pools['h1'].pop()`

Closes #11679 

##### Pre-Merge Checklist

### DIFF
--- a/src/accessibility/AccessibilitySystem.ts
+++ b/src/accessibility/AccessibilitySystem.ts
@@ -162,7 +162,7 @@ export class AccessibilitySystem implements System<AccessibilitySystemOptions>
     private _div: HTMLElement | null = null;
 
     /** A simple pool for storing divs. */
-    private _pool: AccessibleHTMLElement[] = [];
+    private _pools: Record<string, AccessibleHTMLElement[]> = {};
 
     /** This is a tick used to check if an object is no longer being rendered. */
     private _renderId = 0;
@@ -176,6 +176,9 @@ export class AccessibilitySystem implements System<AccessibilitySystemOptions>
     /**  The frequency to update the div elements. */
     private readonly _androidUpdateFrequency = 500; // 2fps
     private _canvasObserver: CanvasObserver;
+
+    // eslint-disable-next-line @typescript-eslint/prefer-readonly
+    private _isRunningTests: boolean = false;
 
     // eslint-disable-next-line jsdoc/require-param
     /**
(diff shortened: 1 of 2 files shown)


brentmc:
The `AccessibilitySystem.postrender()` function was returning early if:
`if ((!this._renderer.renderingToScreen || !this._renderer.view.canvas)`

However, in my tests I needed it to run all the way through so it had a chance to create divs, return them to the pool and then reuse them to assert the change actually fixed the issue. 

I added `this._isRunningTests` as a private variable which I then toggled with `system['_isRunningTests'] = true;`. I really dislike this approach but I wasn't sure how else to skip the early return without creating a public variable/getter function that would then become available to other developers when it is really only needed in this very specific case. 

What do you guys think?

PR #11687, opened 2025-09-23
fix: Allow IRenderLayer as part of the ContainerOptions children array

albinkong:
##### Description of change

Currently it is possible to provide a `IRenderLayer` as an argument to the `Container`'s `addChild` but it is not possible to provide it as an element of `children` in `ContainerOptions`.

Before this patch the following gives a typescript error:

```ts
new Container({
  children: [
    new RenderLayer(),
  ]
})
```

while

```ts
container.addChild(new RenderLayer());
```
works without typescript type error. With this change both of these cases work.

This change will also have the following side effect:

```ts
// Before patch
const parentContainer = new Container<ContainerChild>();
for (const childContainer of parentContainer.children) {
  // Calling container method on childContainer is always possible
  childContainer.getGlobalTint()
}

// After patch
const parentContainer = new Container<ContainerChild>();
for (const childContainer of parentContainer.children) {
  // Calling container method on childContainer is not possible as
  // childContainer is either ContainerChild or IRenderLayer
  childContainer.getGlobalTint()
}
```

I am open to the idea of making the change _only_ to `ContainerOptions` rather than both to the `ContainerOptions` _and_ `Container["children"]`. This would not cause the issue above, but it would continue to be misleading.

An alternative solution could be to expose `RenderLayer` as a class that extends `Container` rather than [hiding that fact](https://github.com/pixijs/pixijs/blob/8eee29cd4ceb3b5f402e3ddce3679e96e8c6cf5f/src/scene/layers/RenderLayer.ts#L521) but I do not know what that would entail.

##### Pre-Merge Checklist

### DIFF
--- a/src/events/FederatedEventTarget.ts
+++ b/src/events/FederatedEventTarget.ts
@@ -1,4 +1,3 @@
-import { type IRenderLayer } from '../scene/layers/RenderLayer.js';
 import { EventSystem } from './EventSystem';
 import { FederatedEvent } from './FederatedEvent';
 
@@ -1094,7 +1093,7 @@ export interface IFederatedContainer extends FederatedOptions
     readonly parent?: Container | null;
 
     /** The children of this event target. */
-    readonly children?: ReadonlyArray<Container | IRenderLayer>;
+    readonly children?: ReadonlyArray<Container>;
 
     /** @private */
     _internalEventMode: EventMode;
--- a/src/events/FederatedEventTarget.ts
+++ b/src/events/FederatedEventTarget.ts
@@ -1,3 +1,4 @@
+import { type IRenderLayer } from '../scene/layers/RenderLayer.js';
 import { EventSystem } from './EventSystem';
 import { FederatedEvent } from './FederatedEvent';
 
@@ -1093,7 +1094,7 @@ export interface IFederatedContainer extends FederatedOptions
     readonly parent?: Container | null;
 
     /** The children of this event target. */
-    readonly children?: ReadonlyArray<Container>;
+    readonly children?: ReadonlyArray<Container | IRenderLayer>;
 
     /** @private */
     _internalEventMode: EventMode;
--- a/src/scene/container/Container.ts
+++ b/src/scene/container/Container.ts
@@ -301,7 +301,7 @@ export interface ContainerOptions<C extends ContainerChild = ContainerChild> ext
      * @see {@link Container#addChild} For adding children
      * @see {@link Container#removeChild} For removing children
      */
-    children?: C[];
+    children?: (C | IRenderLayer)[];
     /**
      * The display object container that contains this display object.
      * This represents the parent-child relationship in the display tree.
@@ -700,7 +700,7 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
      * @see {@link Container#addChild} For adding children
      * @see {@link Container#removeChild} For removing children
      */
-    public children: C[] = [];
+    public children: (C | IRenderLayer)[] = [];
     /**
      * The display object container that contains this display object.
      * This represents the parent-child relationship in the display tree.
--- a/src/scene/container/Container.ts
+++ b/src/scene/container/Container.ts
@@ -9,7 +9,7 @@ import { uid } from '../../utils/data/uid';
 import { deprecation, v8_0_0 } from '../../utils/logging/deprecation';
 import { warn } from '../../utils/logging/warn';
 import { BigPool } from '../../utils/pool/PoolGroup';
-import { type IRenderLayer } from '../layers/RenderLayer';
+import { type RenderLayer } from '../layers/RenderLayer';
 import { cacheAsTextureMixin } from './container-mixins/cacheAsTextureMixin';
 import { childrenHelperMixin } from './container-mixins/childrenHelperMixin';
 import { collectRenderablesMixin } from './container-mixins/collectRenderablesMixin';
@@ -301,7 +301,7 @@ export interface ContainerOptions<C extends ContainerChild = ContainerChild> ext
      * @see {@link Container#addChild} For adding children
      * @see {@link Container#removeChild} For removing children
      */
-    children?: (C | IRenderLayer)[];
+    children?: C[];
     /**
      * The display object container that contains this display object.
      * This represents the parent-child relationship in the display tree.
@@ -737,7 +737,7 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
      * @readonly
      * @advanced
      */
-    public parentRenderLayer: IRenderLayer;
+    public parentRenderLayer: RenderLayer | null = null;
 
     // / /////////////Transform related props//////////////
 
@@ -990,7 +990,7 @@ export class Container<C extends ContainerChild = ContainerChild> extends EventE
      * @see {@link Container#removeChild} For removing children
      * @see {@link Container#addChildAt} For adding at specific index
      */
-    public addChild<U extends(C | IRenderLayer)[]>(...children: U): U[0]
+    public addChild<U extends C[]>(...children: U): U[0]
     {
         // #if _DEBUG
         if (!this.allowChildren)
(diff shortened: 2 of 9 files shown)


albinkong:
I will wait with potentially correcting all the places where `container.children` is used to after comments on the implementation has been made.

bigtimebuddy:
I think types is the correct way to solve this problem and not inheritance. 

Had you tried changing the `ContainerChild` type rather than modifying the children or options?

```ts
export type ContainerChild = Container | IRenderLayer;
```

albinkong:
Thank you for your comment!

> Had you tried changing the `ContainerChild` type rather than modifying the children or options?

I did not initially but I did now and it introduces 24 typescript errors in `childrenHelperMixin.ts` and more in other files as well. Perhaps it is a more complete solution, but it has the same/similar issue as I describe with the `children` property no longer being `Container` only.

bigtimebuddy:
Curious, it seems like RenderLayer does extend Container:

https://github.com/pixijs/pixijs/blob/8eee29cd4ceb3b5f402e3ddce3679e96e8c6cf5f/src/scene/layers/RenderLayer.ts#L98C1-L98C41

But it's casting the constructor as IRenderLayer:

https://github.com/pixijs/pixijs/blob/8eee29cd4ceb3b5f402e3ddce3679e96e8c6cf5f/src/scene/layers/RenderLayer.ts#L520-L522

albinkong:
Initially I was going to just undo that, but it is something that feels deliberate and odd enough to not be undone without understanding why it was done in the first place. The commit that introduced it (23d12630d18836fc6e28b2dec334a88d38406402) did not provide any hints. When digging in the PR I found [this commit](https://github.com/pixijs/pixijs/pull/11094/commits/c73280653fb5c10a9f7754ff6a408837215fc084#diff-eb241c6855f309938a01a98131dd7323ee7b672ad965b1639fcfe0f67c575f59), but I can not discern any reasons for it.

Perhaps @GoodBoyDigital remembers?

mayakwd:
> Initially I was going to just undo that, but it is something that feels deliberate and odd enough to not be undone without understanding why it was done in the first place. The commit that introduced it ([23d1263](https://github.com/pixijs/pixijs/commit/23d12630d18836fc6e28b2dec334a88d38406402)) did not provide any hints. When digging in the PR I found [this commit](https://github.com/pixijs/pixijs/pull/11094/commits/c73280653fb5c10a9f7754ff6a408837215fc084#diff-eb241c6855f309938a01a98131dd7323ee7b672ad965b1639fcfe0f67c575f59), but I can not discern any reasons for it.

I will try to shed some light on the essence of these changes:
- The IRenderLayer interface was introduced to safely integrate `RenderLayer` into the existing architecture, while protecting the user from using `addChild`, `removeChild`, `sortChildren`, etc., as these APIs do not make sense for RenderLayer.
- If you look closely at the Container.addChild code, you will notice that any child passed is automatically cast to `Container`.

As a result, we got an external facade `IRenderLayer` that lacks child management methods, yet is a `Container` in its core.

Now let's think about possible solutions:

- We might consider a concept that aligns a shared API between Container, RenderLayer, and Graphics, of course (because we don't want Graphics to be a Container). So it can sound like a `SceneNode`, `Renderable`, or even a `DisplayObject`!
The eighth version of Pixi.js was designed (I assume) with the idea that everything is a Container. So dropping the DisplayObject was an essential design choice. But there are always edge cases.

- We can extract such a type or interface and apply it to every renderable entity. Still, the refactoring will affect every corner of the entire codebase, and we may find ourselves with an interface or type that has become an abomination.

- The other option is to bring back RenderLayer as a class, dropping `IRenderLayer` interface, and override `addChild`, `removeChild`, and other container-related methods with a code block that throws an error, as is done in ParticleContainer. **Pros:** It's a good compromise, with RenderLayer being a Container. **Cons:** we are not solving type-safety issues and introducing runtime checks.

- And the last option that comes to my mind, and that you already mentioned - update the `ContainerOptions` and put `(C | IRenderLayer)[]` as type for the `children` field. It is **the safest** of existing options with minimal changes, yet we won't achieve strict type safety.

It's a good idea to make minimal changes that will achieve the desired result. All other options are far too overwhelming compared to the benefits they offer (at least in the context of this issue).

bigtimebuddy:
I'm in favor of bring back DisplayObject as a shared interface. I had proposed an idea here to change the inheritance: https://github.com/pixijs/pixijs/discussions/8437 I still don't love things that have no business having children extend Container.

mayakwd:
I agree. 

The only thing I'm concerned about is that there will be a massive refactoring with a lot of breaking changes. So, we might end up with v9.

albinkong:
Great insight @mayakwd!

I agree that it would be outside of the scope of this change to completely redo the render object architecture. 😆 

I will reduce the change to solve my immediate problem with the constructor argument `children` not accepting `IRenderLayer` and retain the issues with `Container<T>.children` technically not being truthful.

albinkong:
I've continued to do things with the render layers and I have discovered that `Container.parentRenderLayer` is typed as `IRenderLayer` but should be either `IRenderLayer | undefined` or even [`IRenderLayer | undefined | null`](https://github.com/pixijs/pixijs/blob/70575309521ab841d36aa5ac26e3cac57ab6ac22/src/scene/layers/RenderLayer.ts#L294).

Should I add this change here or open a separate PR?

bigtimebuddy:
There are some challenges with this PR that I think are type regressions:

Before, defining the Container generic would enforce containers with the same child type. For instance, if you added a different child. This PR would not enforce this error.

```ts
const container = new Container<Sprite>();
container.addChild(new RenderLayer()); // <--- No Type Error
```

Also, with this change, the children are not considered the same type, so you'd need to cast the result for getChild, children, etc. 

```ts
const container = new Container<Sprite>();
container.addChild(new Sprite());
const child = container.children[0]!; // <--- Child is now typed as Sprite | IRenderLayer
```

I'm talking with @GoodBoyDigital and @Zyie today and I'll get them to weigh-in.

bigtimebuddy:
Okay, update on this! I talked with the team and we have some consensus around the approach forward:

First, the solution for solving the specific type problem here is to **remove casting RenderLayer as IRenderLayer** and considering it just something that extends Container. In addition, we'd do what we do with ParticleContainer (which is also a Container but disallows using things like `addChild`), which is to override and throw runtime errors on disallowed methods. Example: https://github.com/pixijs/pixijs/blob/70575309521ab841d36aa5ac26e3cac57ab6ac22/src/scene/particle-container/shared/ParticleContainer.ts#L592-L597

@albinkong is that a change that you want to take on?

Second, the longer-term fix (probably breaking-change) is to re-add DisplayObject which would contain things like transform, bounds measuring, etc and Container's children would extend DisplayObject. This would allow us to remove the previously mentioned `addChild`, `removeChild`, etc methods from things like ParticleContainer and RenderLayer. Originally, we removed DisplayObject in v8 because there's some performance cost for having deeper prototype chains in JS. But we can measure this to see the impact before we commit.

albinkong:
I could look into it, but I just want to verify that the issues with `addChild` accepting `IRenderLayer` actually where introduced here. The workaround I did for my initial problem was to add the render layer with `addChild` rather than the `children` array in the constructor.

My initial change also modifies what the `Container.children` array contained, but after 7f2568452aaa52eb64ea759a8102a962c39f9e7a that should no longer be the case.

albinkong:
I have done the changes but not pushed them here yet. Would it be better if I push it as a change to this PR or as a separate PR?

I've attached the change if anyone wants to provide feedback before I push it anywhere.
[issue-11687-proposed-change.patch](https://github.com/user-attachments/files/22656841/issue-11687-proposed-change.patch)

bigtimebuddy:
Just push to this PR. That'll be fine

albinkong:
Pushed.

I kept a few of the ["hidden" properties](https://github.com/pixijs/pixijs/blob/70575309521ab841d36aa5ac26e3cac57ab6ac22/src/scene/layers/RenderLayer.ts#L89) `@internal` in the `RenderLayer` but elected to not add overrides for all the `EventEmitter` methods.

bigtimebuddy (approved):
Nice work!

albinkong:
Thanks for the feedback and guidance!

Issue #11550, opened 2025-07-16
Bug: Memory leak regression in PIXI.Graphics (WebGL)


VeselyT:
### Current Behavior

A memory leak occurs when a PIXI.Graphics object is repeatedly cleared and redrawn (e.g., using quadraticCurveTo). With each redraw, new WebGLBuffer objects are allocated, but the old ones are not deallocated, causing memory usage to grow indefinitely.

This issue is a regression, as it does not occur in version 8.9.2 but is present in subsequent versions. A reproducible example is provided below.

### Expected Behavior

Memory usage should remain stable when a PIXI.Graphics object is repeatedly cleared and redrawn.

### Steps to Reproduce

- Version 8.11.0 with memory leak is [here](https://www.pixiplayground.com/#/edit/6BYsJ99Aotqt55blcfhiS)
- Version 8.9.2 without memory leak is [here](https://www.pixiplayground.com/#/edit/7-ISVz7gGkQAaIZGlZiuG) 

1. Open the provided reproduction page and wait 10 seconds for the application to run and then destroy itself.
2. After it's destroyed, take a heap snapshot using your browser's developer tools.
3. When you filter the snapshot for WebGLBuffer, you will see that numerous buffer objects are still retained in memory.
4. Note the time logged in the console for the destroy operation. It is significantly slower (160ms vs. 3ms) because it gets stuck in WEBGL_lose_context.loseContext(). In our production app, this can cause the page to freeze for up to 30 seconds if it was open for the long time.

<img width="813" height="143" alt="Heap snapshot with memory leak" src="https://github.com/user-attachments/assets/df6e4ac1-5883-4d19-bc6a-49521a998ee9" />

<img width="924" height="155" alt="Heap snapshot without leak" src="https://github.com/user-attachments/assets/813af619-54d2-452e-bb61-27d36be994a2" />

### Environment

- WebGL renderer 
- **`pixi.js` version**: *8.11.0*
- **Browser & Version**: *Chrome 138*
- **OS & Version**: *macOS 15.5*

VeselyT:
I noticed that GraphicsContextRenderData seems to allocate a DefaultBatcher when it's obtained from the BigPool, but I couldn't find where the batcher is destroyed upon being returned. Could this be the source of the leak? I hope this suggestion is helpful!

PR #11712, opened 2025-10-13
fix: graphics memory leak and rendering resource cleanup

Zyie:
Fixes: #11550

- Improves memory management for renderable objects by introducing a `nullCount` tracking mechanism within the `RenderableGCSystem`. This enables immediate cleanup of managed hash tables when their null entry count exceeds a configurable threshold, preventing memory bloat between full garbage collection cycles.
- Refines `Batcher` and `DefaultBatcher` destruction, enabling conditional shader and geometry. This was the main culprit for the graphics memory leak.
- Ensures more comprehensive nulling of internal references in `Geometry`, `Shader`, and `ViewableBuffer` when they are destroyed, preventing memory leaks.

Examples
- [8.9.2](https://www.pixiplayground.com/#/edit/VW1vv3hw8DVkWeGpS-jtQ)
- [8.11.0](https://stackblitz.com/edit/ahnorc5k?file=src%2Findex.js)
- [This PR](https://stackblitz.com/edit/ahnorc5k-eskxfefe?file=src%2Findex.js)

### DIFF
--- a/src/rendering/renderers/gpu/buffer/GpuBufferSystem.ts
+++ b/src/rendering/renderers/gpu/buffer/GpuBufferSystem.ts
@@ -27,12 +27,10 @@ export class GpuBufferSystem implements System
     private _gpuBufferBinding: ManagedItem;
 
     private _gpu: GPU;
-    private _renderer: WebGPURenderer;
 
     constructor(renderer: WebGPURenderer)
     {
-        this._renderer = renderer;
-        this._gpuBufferBinding = renderer.renderableGC.addManagedHash(this, '_gpuBuffers');
+        this._gpuBufferBinding = renderer.renderableGC.addManagedHash(this, '_gpuBuffers', true);
     }
 
     protected contextChange(gpu: GPU): void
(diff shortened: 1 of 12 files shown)


Zyie:
Hey @VeselyT are you able to check this PR in your application?

VeselyT:
@Zyie thanks for this work. I tested this briefly and it seems to work for us. I do not see any leaked resources as previously. We will perform some more testing on our staging server and I will report back in couple days how it works. Thanks again.

VeselyT:
@Zyie it seems to work fine for our use case

src/rendering/renderers/gpu/buffer/GpuBufferSystem.ts
```diff
@@ -105,6 +106,7 @@ export class GpuBufferSystem implements System
 
     protected onBufferChange(buffer: Buffer)
     {
+        if (!this._gpuBuffers) return;
```
GoodBoyDigital: we should try to avoid needing this check. Is this a renderer.destroy issue?

Issue #11711, opened 2025-10-12
SplitBitmapText requires "fill: 'white'" when using a custom bitmap font, otherwise the text is black


lennybakkalian:
### PixiJS version

8.14.0

### Link to minimal reproduction

https://stackblitz.com/edit/pixijs-v8-jtangtjo?file=src%2Fmain.js

### Steps to reproduce

Create an instance of BitmapText without specifying a fill style property.

Create an instance of SplitBitmapText using the same font and style configuration, also without specifying a fill style property.

### What is expected?

Both the BitmapText and SplitBitmapText objects should correctly render the text using the custom font's texture when no fill is specified.

Actual Behavior
The BitmapText renders correctly.
The SplitBitmapText renders the text as solid black.

### What is actually happening?

The SplitBitmapText object displays text as completely black when a custom bitmap font is used, unless a fill style property is explicitly provided in the configuration. This behavior differs from the standard BitmapText object, which correctly renders the text using the bitmap texture without needing an explicit fill value.

### Environment

- **Browser & Version**: Chrome 140.0.7339.214
- **OS & Version**: macOS Sequoia v15.7.1

PR #11721, opened 2025-10-17
fix: SplitBitmapText requires "fill: 'white'"

GiorgiMaziashvili:
##### Description of change

Added a default `fill = 0xffffff` to `SplitBitmapText` since it was inheriting the default black color.

Fix: #11711

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/text-split/SplitBitmapText.ts
+++ b/src/scene/text-split/SplitBitmapText.ts
@@ -170,7 +170,7 @@ export class SplitBitmapText extends AbstractSplitText<BitmapText>
             ...config,
         };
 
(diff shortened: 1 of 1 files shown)


src/scene/text-split/SplitBitmapText.ts
```diff
@@ -170,6 +170,9 @@ export class SplitBitmapText extends AbstractSplitText<BitmapText>
             ...config,
         };
 
+        completeOptions.style ??= completeOptions.style || {};
```
bigtimebuddy: But this could be simpler:
```suggestion
        completeOptions.style ??= {};
```

PR #11729, opened 2025-10-23
chore: add PR performance benchmark workflow

Zyie:
Implements an automated workflow to run performance benchmarks on pull requests.
Compares the PR's performance against the main development branch baseline.

### DIFF
--- a/.github/workflows/benchmark-pr.yml
+++ b/.github/workflows/benchmark-pr.yml
@@ -10,7 +10,7 @@ concurrency:
 
 jobs:
   benchmark:
-    runs-on: blacksmith-4vcpu-ubuntu-2404
+    runs-on: macos-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


bigtimebuddy:
What is the reason the the performance benchmark in another repo? Why not add it here?

Zyie:
> What is the reason the the performance benchmark in another repo? Why not add it here?

I was mostly thinking to avoid bloating the main repo as we add more benchmarks
I happy to move it over though if you like?

bigtimebuddy:
The challenge of keeping it external is that it creates an upstream dependency on our repo. Say for instance, that we change an interface or some API it would require us to update the benchmark or we'd be blocked. Because of that coupling, I would say that it makes more sense to keep it in this repo.

Zyie:
going to convert this one back to a draft as the consistency disappears with the more tests that are run

PR #11747, opened 2025-11-09
fix: parse resolution and format from URLs in object sources in assets resolver

stereopasa:
hey
noticed that with produced manifest from assetpack i don't get expected result for resolving the asset.
probably that could be fixed on assetpack level, looks like test says it's expected output now https://github.com/pixijs/assetpack/blob/main/packages/assetpack/test/manifest/Manifest.test.ts#L419

##### Description of change
Previously, when adding assets with object sources (e.g., 
`{ resolution: 2, src: 'sprite@2x.webp.json' }`), the Resolver would 
not run URL parsers on the nested `src.src` string. This caused the 
`resolveJsonUrl` parser to be skipped, preventing proper extraction of 
resolution and format metadata from the URL pattern.

Changes:
- Added URL parsing for `src.src` strings when source is an object
- Refactored duplicate parser lookup logic into `parseUrl()` helper
- Parser results are now properly merged with explicit object properties

Fixes the issue where spritesheet URLs like 'my-sprite@0.5x.webp.json' 
would not have their resolution (0.5x) and format (webp) extracted when 
provided as object sources, causing incorrect asset resolution based on 
preferences.

Test: Resolver.test.ts - "should parse resolution and format from json 
URLs for both string and object sources"

##### Pre-Merge Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pixi.js",
-  "version": "8.14.0",
+  "version": "8.14.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "pixi.js",
-      "version": "8.14.0",
+      "version": "8.14.1",
       "license": "MIT",
       "dependencies": {
         "@pixi/colord": "^2.9.6",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pixi.js",
-  "version": "8.14.0",
+  "version": "8.14.1",
   "author": "PixiJS Team",
   "homepage": "http://pixijs.com/",
   "bugs": "https://github.com/pixijs/pixijs/issues",


src/assets/resolver/Resolver.ts
```diff
@@ -506,6 +506,25 @@ export class Resolver
...
+                    const parser = this._parsers[i];
+
+                    if (parser.test(url))
+                    {
+                        result = parser.parse(url);
+                        break;
+                    }
+                }
+
+                return result;
```
bigtimebuddy: Can we reduce this?
```suggestion
                const parser = this._parsers.find((p) => p.test(url));
                return parser?.parse(url);
```

bigtimebuddy: Yeah, no worries. Lots of older code in the project. This is not performance critical code, so we don't need to be quite as verbose.

Zyie (approved):
Thanks for this @stereopasa!

PR #11748, opened 2025-11-09
fix: filters with blendRequired but no gpuProgram not rendering correctly

jimhigson:
filters with blendRequired=true work if a glProgram is given, but no gpuProgram

Closes https://github.com/pixijs/pixijs/issues/11745

##### Description of change

See [this bug report](https://github.com/pixijs/pixijs/issues/11745) - filters with a blendRequired=true, and a glProgram (but no gpuProgram) were not working.

The reason is the gpuProgram was required to specify the binding, even if it wasn't doing any rendering. After my change, the name `uBackTexture` is bound to slot 3 if a resource called `uBackTexture` exists.

##### Pre-Merge Checklist

- [ ] Tests and/or benchmarks are included
*no new tests added - I couldn't find an existing scene for blendMode=true to base this off*
- [x] Documentation is changed or added
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

PR #11754, opened 2025-11-12
fix: correct uBackTexture binding for filters with blendRequired

### LINKED ISSUE #11745: Filters with backbuffers seem not to work as expected
### PixiJS version

8.14.0, 8.3.4

https://pixiplayground.com/#/edit/q0w7pEykjfdlJSXL6aJEz

### Steps to reproduce

Please try my reproduction. Clicking these text: 

Image

Will insert into the template for the frag. The code in the frag template is very similar to the [built-in blend](https://github.com/pixijs/pixijs/blob/15bd3d9d2d99e5cd9aa5817ae75c929d93fcf58d/src/filters/blend-modes/blend-template.frag) - it samples from `uTexture`, and `uBackTexture` and blends in different ways.

What I'm finding both in [my retro game](https://blockstack.ing) and this reproduction is unexpected behaviour - in my game, if I sample from uTexture (even if I multiply the result by zero to destroy the information), it inserts uTexture completely into the display.

### What is expected?

Click 'back' - see back only (front sprite's shader samples from its backbuffer and returns it unmodified)
'back inv' should be the same as above but with inverted colours (actually inverts the front sprite)
Click 'front' - see front as normal only (front sprite's shader samples from its utexture and returns it unmodified)
Click 'mixed' - see the two mixed

In this example, using the backBuffer for `finalColor` doesn't show what is behind the front sprite, it instead renders only the front sprite but stretched.

The example for `interlaceChannels` does `'vec4(textureCol.r, backCol.g, textureCol.b, 1.0)'` - however this too just shows the front sprite.

Mixed is the same as front, despite doing `mix(back,front,0.5)`

### Environment

- Chrome 141.0.7390.123
- MacOS 26.0.1
- Apple Silicon

Also seen on Windows 11, x64 arch

### Any additional comments?

`uBackTexture` is not well documented, so I could have the wrong idea here. However, my shader in this example (and the more complex one in my actual game) both show seemingly impossible behaviour where simply reading from the `uTexture` seems to override the output.

In fact, googling `uBackTexture pixi.js` currently brings up nothing relevant to using this backbuffer.

I'm happy to contribute to the docs if this is found to be a misunderstanding on my part.

GoodBoyDigital:
##### Description of change
Fixes #11745

When filters had `blendRequired` set to true, the `uBackTexture` was being bound to the wrong texture unit (3 instead of 1), causing unexpected behavior where the back buffer would not display correctly.

The issue was that filters expecting to sample from the back buffer (the pixels behind the filtered object) were instead seeing stretched or incorrect textures. This was because the texture unit binding was set to 3 when it should have been 1.

##### Pre-Merge Checklist

### DIFF
--- a/src/filters/Filter.ts
+++ b/src/filters/Filter.ts
@@ -208,7 +208,14 @@ export class Filter extends Shader
         this.blendRequired = options.blendRequired;
         this.clipToViewport = options.clipToViewport;
 
+        // this is where the filter system will attach the filter texture
         this.addResource('uTexture', 0, 1);
+
+        if (options.blendRequired)
+        {
+            // this is where the filter system will attach the back texture
+            this.addResource('uBackTexture', 0, 3);
+        }
     }
 
     /**


jimhigson:
@GoodBoyDigital  -  Can I suggest pulling [this change](https://github.com/pixijs/pixijs/pull/11748/files#diff-16a8bd89885c0eef247f2ef075c5530fcb8853538837907733c29bbc95e76d57) over from my original PR? Otherwise, there's no way to know the uniform is called `uBackTexture` unless you poke around the pixi source - this string doesn't appear anywhere in the docs currently.

I'm happy to make a small PR with this change

GoodBoyDigital:
> @GoodBoyDigital - Can I suggest pulling [this change](https://github.com/pixijs/pixijs/pull/11748/files#diff-16a8bd89885c0eef247f2ef075c5530fcb8853538837907733c29bbc95e76d57) over from my original PR? Otherwise, there's no way to know the uniform is called `uBackTexture` unless you poke around the pixi source - this string doesn't appear anywhere in the docs currently.
> 
> I'm happy to make a small PR with this change

great idea! please a small PR would be ace - thanks!

jimhigson:
Cool, I'll make one tonight after I finish my day job :-)

jimhigson:
@GoodBoyDigital - very small docs fix here: https://github.com/pixijs/pixijs/pull/11756

Issue #11737, opened 2025-10-28
TypeError: Cannot redefine property: resizeTo after second init


abnerlinan:
### PixiJS version

main

### Link to minimal reproduction

[https://jsfiddle.net/mLo1xkds/](https://jsfiddle.net/mLo1xkds/)

### Steps to reproduce
file: src\app\ResizePlugin.ts
in method: destory()  set this._resizeTo = null;
run init() again,
Object.defineProperty(this, 'resizeTo' will run mistake

### What is expected?

before Object.defineProperty check resizeTo :
 if ((this as any).resizeTo !== undefined) {
        delete (this as any).resizeTo;
  }

### What is actually happening?

up

### Environment

chrome 140

abnerlinan:
> Hello [@abnerlinan](https://github.com/abnerlinan). Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) using [StackBlitz](https://stackblitz.com/edit/pixijs-v8) or [JSFiddle](https://jsfiddle.net/yavg3stL/). Issues marked with `Needs Reproduction` will be closed if they have no activity within 7 days.

https://jsfiddle.net/mLo1xkds/

PR #11759, opened 2025-11-13
fix: enable proper application reinitialization

Zyie:
##### Description of change

fixes: #11737

Ensures the application's core stage is recreated when initialized,
addressing issues where a previously destroyed application could not be
reliably reinitialized.

Also makes plugin properties configurable, allowing them to be redefined
during reinitialization cycles.

A new test case validates this improved behavior.


bigtimebuddy (approved):
LGTM!

PR #11764, opened 2025-11-14
fix: Remove automatically added 'resize' event listener from TextureSource on Texture destroy()

Romans-I-XVI:
##### Description of change

I noticed my manually created Textures weren't being released from memory. I found it was due to a reference being held by TextureSource caused by the automatically added 'resize' event listener.

Example
https://stackblitz.com/edit/pixijs-v8-yo5njdcr?file=src%2Fmain.js

This change removes the listener from TextureSource when Texture destroy() is called 

##### Pre-Merge Checklist


bigtimebuddy (approved):
Good catch!

PR #11768, opened 2025-11-24
fix: Add error handling for message processing in KTX worker

zardoy:
##### Pre-Merge Checklist

Hello! Sorry, opened a PR a bit early via vscode popup without linking to a correct issue.

In short, this is an enhancement improvement so errors would popup correctly to a place where the .load was called, without this it's almost impossible to track what source triggered an issue and code in the main thread would hang forever.

Example of the stack that you would now be able to see:
```
Uncaught (in promise) Error: [Loader.load] Failed to load blob:http://localhost:3001/b4c9f365-a446-4bcd-894f-8d08bfe91fd5.
Error: Unable to transcode basis texture.
    at _Loader._loadAssetWithRetry (Loader.mjs:202:1)
    at async Loader.mjs:122:1
    at async Promise.all (:3001/index 0)
    at async _Loader.load (Loader.mjs:127:1)
    at async AssetsClass._mapLoadToResolve (Assets.mjs:527:1)
    at async AssetsClass.load (Assets.mjs:212:1)
    at async testSeqLoad (main.ts:227:1)
```

### DIFF
--- a/src/assets/loader/Loader.ts
+++ b/src/assets/loader/Loader.ts
@@ -461,7 +461,13 @@ export class Loader
 
                 // strategy 'throw' or exhausted 'retry'
                 if (onError) onError(e as Error, asset);
-                throw new Error(`[Loader.load] Failed to load ${url}.\n${e}`);
+                const error = new Error(`[Loader.load] Failed to load ${url}.\n${e}`);
+
+                if (e instanceof Error && e.stack)
+                {
+                    error.stack = e.stack;
+                }
+                throw error;
             }
         }
     }


zardoy:
im not sure if its worth a dedicated issue
However, as I was debugging it I noticed that memory usage never goes down as I load more textures
I think it would be useful to expose worker and worker code for some adjustments e.g. so you can terminate it or easily patch code for it

bigtimebuddy (approved):

zardoy:
Hey I also added error stack restoration so the stack of the actual error pops up in the console. Alternatively, I could use the `cause` argument of an Error here.

Issue #11769, opened 2025-11-25
Feature Request: V8 Shaded Mesh examples / documentation


laino:
### Description

There seems to be a lack of example code / documentation around setting up up a basic mesh with a shader + uniforms. The ``resources`` entry being a record of any values doesn't make any more obvious what's supposed to go in there either.

For my gl shader it took me a good two hours to figure out that it's supposed to be

```ts
resources: {
    whatever: {
        uTint: { value: [r, g, b], type: "vec3<f32>" },
    }
}
```

rather than

```ts
resources: {
    uTint: { value: [r, g, b], type: "vec3<f32>" }, // wrong, needs to be wrapped in a group like above!
}
```

A basic example of a working vertex shader is also not really to be found in V8's docs, so it also took me a long time to find this as well:

```.vert
in vec2 aPosition;
in vec2 aUV;

out vec2 vUV;

uniform mat3 uProjectionMatrix;
uniform mat3 uWorldTransformMatrix;
uniform mat3 uTransformMatrix;

void main() {
    mat3 mvp = uProjectionMatrix * uWorldTransformMatrix * uTransformMatrix;
    gl_Position = vec4((mvp * vec3(aPosition, 1.0)).xy, 0.0, 1.0);
    vUV = aUV;
}
```

### Full Mesh Shader Example

In the hopes that finding this issue helps the next person, here's some working example code:

```ts
// Vertex shader - transforms quad vertices
const vertexShader = `
in vec2 aPosition;
in vec2 aUV;

out vec2 vUV;

uniform mat3 uProjectionMatrix;
uniform mat3 uWorldTransformMatrix;
uniform mat3 uTransformMatrix;

void main() {
    mat3 mvp = uProjectionMatrix * uWorldTransformMatrix * uTransformMatrix;
    gl_Position = vec4((mvp * vec3(aPosition, 1.0)).xy, 0.0, 1.0);
    vUV = aUV;
}
`;

// Fragment shader - renders 3D sphere with lighting
const fragmentShader = `
in vec2 vUV;
uniform vec3 uTint;

void main() {
    // Convert UV to centered coordinates [-1, 1]
    vec2 uv = vUV * 2.0 - 1.0;
    
    // Calculate distance from center
    float dist = length(uv);
    
    // Discard pixels outside the circle
    if(dist > 1.0) {
        discard;
    }
    
    // Calculate sphere depth
    float z = sqrt(1.0 - dist * dist);
    
    // Simple lighting with normal
[...]

PR #11779, opened 2025-11-30
feat: add shader and filter resource generic

bigtimebuddy:
### Overview

This PR is an attempt at adding a [generic argument to Shader and Filter](https://github.com/pixijs/pixijs/issues/11769#issuecomment-3592559050) to describe the shape of Shader `resources`. This improves internal type-safety a lot for Filters as now resources are properly typed.

#### Before

Screenshot 2025-11-30 at 6 33 56 PM

#### After

Screenshot 2025-11-30 at 6 33 45 PM

### Further improvement

* The options `resources` can be mutated into UniformGroups differing in shape to the Shader's `resources`, so there may need to be separate optional output options. 

cc @laino

### DIFF
--- a/src/filters/Filter.ts
+++ b/src/filters/Filter.ts
@@ -5,9 +5,9 @@ import { State } from '../rendering/renderers/shared/state/State';
 
 import type { RenderSurface } from '../rendering/renderers/shared/renderTarget/RenderTargetSystem';
 import type {
-    GenericResources,
     IShaderWithResources,
     ShaderFromResources,
+    ShaderResources,
     ShaderWithResources
 } from '../rendering/renderers/shared/shader/Shader';
 import type { BLEND_MODES } from '../rendering/renderers/shared/state/const';
@@ -69,7 +69,7 @@ export interface FilterOptions
  * @advanced
  * @see {@link FilterOptions}
  */
-export type FilterWithShader<R = GenericResources> = FilterOptions & IShaderWithResources<R>;
+export type FilterWithShader<R = ShaderResources> = FilterOptions & IShaderWithResources<R>;
 
 /**
  * The antialiasing mode of the filter. This can be either:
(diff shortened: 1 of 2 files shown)


laino:
Just tried this. Currently the inferred types are not correct.

For a shader created with:
```ts
function makeShader(seed: number, radius: number, baseColor: [number, number, number]) {
    return new Shader({
        glProgram: starGLProgram,

        resources: {
            waveUniforms: {
                uTint: { value: baseColor, type: "vec3<f32>" },
                uSeed: { value: seed, type: "f32" },
                uRadius: { value: radius, type: "f32" },
                uRotation: { value: 0, type: "f32" },
                uNoise: { value: 0, type: "f32" },
            }
        }
    });
}

```

I get the following inferred type on ``shader.resources.waveUniforms``:

```ts
{
    //...
    uTint: {
        value: [number, number, number];
        type: string;
    };
    //....
}
```

I believe the correct type should be:

```ts
{
    uniforms: {
        //...
        uTint: [number, number, number];
        //...
    }
}
```

bigtimebuddy:
Yeah, this only works now if you create the UniformGroup yourself and pass that in. This needs work because the Shader mutates the options into UniformGroup objects.

Does this work?

```ts
function makeShader(seed: number, radius: number, baseColor: [number, number, number]) {
    const waveUniforms = new UniformGroup({
        uTint: { value: baseColor, type: "vec3<f32>" },
        uSeed: { value: seed, type: "f32" },
        uRadius: { value: radius, type: "f32" },
        uRotation: { value: 0, type: "f32" },
        uNoise: { value: 0, type: "f32" },
    });
    return new Shader({
        glProgram: starGLProgram,
        resources: {
            waveUniforms,
        }
    });
}

laino:
That works! 👍

PR #11782, opened 2025-12-02
fix: make rendering to multiple textures work (WebGL)

laino:
We need to call gl.drawBuffers to instruct GL to actually use the multiple buffers we bound. In WebGL1 this is available as an extension. This is also per-framebuffer state, so this setup is only required once.

Now something like this will work:

```frag
layout(location = 0) out vec4 fragNormal;
layout(location = 1) out vec4 fragMaterial;
layout(location = 2) out vec4 fragAlbedo;
```

```ts
const target = new RenderTarget({
    colorTextures: [normalTexture, materialTexture, albedoTexture],
});

renderer.render({
    container,
    target,
    clear: true,
    clearColor: [0, 0, 0, 0],
});
```

Before only writing to whatever was in ``(location = 0)`` would do anything.

### DIFF
--- a/src/rendering/renderers/gl/renderTarget/GlRenderTargetAdaptor.ts
+++ b/src/rendering/renderers/gl/renderTarget/GlRenderTargetAdaptor.ts
@@ -314,7 +314,7 @@ export class GlRenderTargetAdaptor implements RenderTargetAdaptor<GlRenderTarget
                 0);// mipLevel);
         });
 
-        if (colorTextures.length > 0)
+        if (colorTextures.length > 1)
         {
             const bufferArray: number[] = [];
 


laino:
Later it might be nice to also have an option to select target textures per-shader, since that was the reason ``drawBuffers`` exists in the first place, rather than GL just using all bound textures of the FB implicitly.

GoodBoyDigital:
heeey! worked on a PR here: https://github.com/pixijs/pixijs/pull/11792

turns out WebGPU was broken too.. 

all fixed now thought! Like your follow up idea. Will have a noddle on how to implement!

Zyie:
closing in favor of #11792

PR #11792, opened 2025-12-08
fix: Add Multiple Render Targets (MRT) support

GoodBoyDigital:
## Description of change

This PR adds support for Multiple Render Targets (MRT) to both WebGL and WebGPU renderers, allowing rendering to multiple color attachments simultaneously.

### WebGL Changes:
- Moved draw buffers setup to the bind method for proper MRT configuration
- Pre-computed draw buffer arrays for performance (supports up to 16 targets)
- Supports both WebGL1 (via WEBGL_draw_buffers extension) and WebGL2
- Proper warnings when WebGL1 lacks MRT support

### WebGPU Changes:
- Added `colorTargetCount` property to `GpuRenderTarget`
- Fixed pipeline caching to include color target count in the global state key
- Updated `getColorTargets()` to create multiple color target states
- Applied write mask to all color targets, not just the first one

### Testing:
- Added comprehensive test scene with 3 color attachments
- Visual test snapshots for WebGL1, WebGL2, and WebGPU
- Generates red, green, and blue gradients to different render targets
- Fixed WGSL vertex shader formatting for proper attribute parsing

## Pre-Merge Checklist

## Implementation Details

The key fix was adding color target count to the pipeline cache key. Previously, pipelines were being cached without considering the number of color targets, causing incorrect pipeline reuse when switching between single-target and multi-target rendering.

Supports 1-4 color attachments with proper state management and caching.


guillaumebrunerie:
Looking at the pull request, is there a good reason why the green and blue gradients are inverted in WebGPU?

Julien-Marcou:
Hey @GoodBoyDigital & @Zyie, I don't know if you have seen the above comment, I just want to double check with you that this is the desired behavior:

The green & blue vertical gradients are inverted between webgl & webgpu

image

pie6k:
Hey! It looks amazing! Thank you!

I'd like to ask some question about how to properly utilize it without some hacky tricks.

I want to implement motion blur.

I already have it working, where I have a custom `Mesh` and this mesh knows its current and prev position and calculates velocity of each fragment pixel and blurs itself.

However, I'd like to experiment with deferred shading here:
I want to render normally, but also ouptut to velocity buffer and add final render pass that applies motion blur using velocity buffer.

My Mesh shader would be aware of it and output to 2 targets. I'd like other things (like normal filters, that output only to 1 target) to not break.

Some other notes:

It could be useful to be able to have multiple "layers" of our output buffers. Or would it work out of the box with render groups?

Take a look at this example: we have a moving webcam on top of (currently) static background
CleanShot 2026-03-18 at 20 26 18@2x

it currently looks as expected - webcam pixels are blurry, but bg pixels stay sharp everywhere. However, if we'd have a single velocity buffer and a single 'motion blur pass' - we'd have to blur pixels of the final image, so background pixels would likely also be blurred. In this case I'd hope to separate those layers inside render groups to keep the result as it is currently. How should I implement this?

In general, I really like it, but TBH when digging really deep into pixi, I often feel quite lost as there are no solid and up-to-date docs about advanced topics like those or some recommendation or examples.

I did check your PR code and the test example, but it feels a bit hacky:
- you create render texture, than render into that texture, then create 3 sprites with those textures and attach those to the scene
- in real-world scenario, where I have a lot of containers already added to the scene - how should I approach this? Should I keep those attached to the scene, but then to 'cover' them with a sprite with a final texture? Feels hacky and also feels like the stage would first be normally rendered, then I'd render it again to render target and then I'd render it again, but only showing our texture inside sprites. Doesn't feel correct way of multi-pass rendering

Issue #8632, opened 2022-09-07
Why basis texture's alphaMode default 0?


B1anker:
### Expected Behavior
alphaMode = 1
### Current Behavior

<img width="1084" alt="image" src="https://user-images.githubusercontent.com/17514962/188777290-6c143094-cadb-464a-87fa-db6b1318ff5a.png">

### Environment

- **`pixi.js` version**:6.4.2
- **Browser & Version**: _e.g. Chrome 67_
- **OS & Version**: _e.g. Ubuntu 18.04_
- **Running Example**: _url_

ivanpopelyshev:
Which formats of compressed textures that allow alpha do you suggest to transcode from basis?

B1anker:
> Which formats of compressed textures that allow premultiplied alpha do you suggest to transcode from basis?
Because it will increase drawcall.
<img width="899" alt="image" src="https://user-images.githubusercontent.com/17514962/189074037-defcf058-623a-4a48-afd3-d9a7fc6d61c7.png">
，so I replace alphaMode from 0 to 1.
Finally drawcall will be normal.

ivanpopelyshev:
Oh, you mean you have no partially transparent pixels, and alphaMode doesnt affect your case? No, there's no way for that. 

You can make custom batcher with shader that multiplies alpha if needed, and put alphaMode in argument

B1anker:
> Oh, you mean you have no partially transparent pixels, and alphaMode doesnt affect your case? No, there's no way for that.
> 
> You can make custom batcher with shader that multiplies alpha if needed, and put alphaMode in argument

What confuses me is that `@pixi/basis` generating texture(BASISU/ETC1S)'s alphaMode is 0 by default.
Actually the `.basis` has alpha mode.

eXponenta:
@B1anker 
Alpha mode 0 means that texture is  NOT premultiplied, there are not premultiplied compressed formats.

PMA and NPMA textures must be processed differently.

https://pixijs.download/dev/docs/PIXI.html#ALPHA_MODES

See:
https://github.com/KhronosGroup/KTX-Software/issues/137

ivanpopelyshev:
@eXponenta we can do something for case alpha=0,1

kvanderhaeghen-aleacsysonline:
@ivanpopelyshev @eXponenta We have issues with alpha not being premultiplied when loading KTX2 as texture/sprite, is there a way to fix this in pixi 8? (we are on 8.7) Thanks in advance.

[...]

PR #11813, opened 2025-12-21
feat: Add ktx2 batcher and shader to prevent erroneously premultiplying alpha

artemiomorales:
##### Description of change

See #8632

This adds a batcher and shader to handle the alpha correctly, at least for my use case with KTX2 files. I initially branched off of `v8.14.3` and I haven't tested it against `dev` yet, so not sure how if it works with the latest code.

This is very much a draft that I'm not sure belongs in core, though I've opened it in case it's helpful for anyone else.

If this is indeed a change we'd like to make to core, I can see about potentially cleaning it up and/or pushing it further, and of course anyone feel free to run with this if you'd like 🙏

##### Pre-Merge Checklist


dragonbaba:
I currently use a large number of KTX2 and DDS textures for Android and Windows platforms. There's no need to write an additional KTX2 batch processor (in fact, your new batch processor would increase draw calls because it would cause the previously used batch processor to break when called). What you need to do is organize compressed textures and other uncompressed textures into different layers, allowing objects within the same layer to fully utilize batch processing.

Zyie:
closing this down as its been in draft for a while without any updates

PR #11815, opened 2025-12-23
feat: Canvas renderer

krzys:
##### Description of change
Implemented CanvasRenderer and hooked it into the render pipeline.

> **Story time**: I’ve been rewriting our 2D game client from raw Canvas (hand-rolled rendering) to PixiJS, and on weaker devices the performance took a noticeable hit. The bottleneck wasn’t the game logic — it was the rendering path overhead for simple scenes where Canvas used to be “good enough” and very predictable. And well... brought back Canvas render. 
> *Merry Christmas 🎄*

Let me know if you want any changes or additional feature coverage for the CanvasRenderer. What’s implemented is more than enough for our 2D game, so I’m good with it.

##### CanvasRenderer Feature Support Checklist
- [x] Sprite
- [x] AnimatedSprite
- [x] TilingSprite
- [x] Graphics (Rectangle, Circle, Polygon, Rounded Rectangle)
- [x] GraphicsPath
- [x] RotatedTextures
- [x] PremultipliedAlpha (automatic un-premultiplication for Canvas 2D compatibility)
- [x] Tinting
- [x] SVG Gradients
- [x] BlendModes (**partially**: not every one is supported by browsers)
- [x] CanvasText
- [x] BitmapText (**partially**: lacks MSDF/SDF distance field rendering)
- [x] HTMLText (**partially**: standard rendering only)
- [x] Masking (**partially**: Graphics masks only; sprite/texture masks skipped with warn-once; inverse masks ignored with warn-once)
- [x] Batching (translated - canvas draws elements individually)
- [x] Filters (simplified):
  - AlphaFilter (via globalAlpha)
  - BlurFilter (via CSS blur, approximate strength)
  - ColorMatrixFilter (approximate): brightness, contrast, saturate, hue-rotate, grayscale/greyscale, sepia,
    negative (invert)
  - FilterArea clipping supported via clip path
  - Unsupported filters are skipped with warn-once
- [x] ~CustomShaders~ (unsupported)
- [x] ~Meshes~ (unsupported)
- [x] ~CompressedTextures~ (unsupported)

##### Pre-Merge Checklist


bigtimebuddy:
@krzys You're amazing for working on this! I'll go through it more thoroughly, but looks promising. Would you mind taking a look at that failing visual tests?

bigtimebuddy:
@krzys thanks for your patience, we are back after the New Year and the team will take a look at this. 

I noticed two things that didn't seem right:

* Noticed update to `tests/visual/snapshots/graphics-texture-scene-ts-*.png` test images. This doesn't seem correct, I wouldn't expect these to change
* Why are there no canvas test images? I would expect those to be generated and added with this PR for any visual tests?

GoodBoyDigital:
This. Is. Awesome!!!!  Thanks so much for this PR :)


We plan on reviewing this ASAP

krzys:
@bigtimebuddy just added test images.
`tests/visual/snapshots/graphics-texture-scene-ts-*.png` - test was updated to cover the `tint` edge case where black (`0x000000`) is passed as a tint value. Internally, `0` is treated as falsy so it falls back to white.  (commit ec72c823a57277950b82b258871c851fe38b98c4)

bigtimebuddy:
Thanks. We should address that tint test as another PR, but good find!

src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
```diff
@@ -46,82 +47,125 @@ interface RenderTargetAndFrame
...
+ * @memberof rendering
+ * @ignore
+ */
+export interface RenderTargetAdaptor<RENDER_TARGET extends RendererRenderTarget>
 {
+    /**
+     * Initializes the adaptor.
+     * @param {Renderer} renderer - the renderer
+     * @param {RenderTargetSystem} renderTargetSystem - the render target system
+     * @standard
```
Zyie: We can remove all the standard tags here as the interface is ignored

src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
```diff
@@ -150,30 +194,58 @@ export interface RenderTargetAdaptor<RENDER_TARGET extends GlRenderTarget | GpuR
  * @category rendering
  * @advanced
  */
-export class RenderTargetSystem<RENDER_TARGET extends GlRenderTarget | GpuRenderTarget> implements System
+export class RenderTargetSystem<RENDER_TARGET extends RendererRenderTarget> implements System
 {
-    /** When rendering of a scene begins, this is where the root render surface is stored */
+    /**
+     * When rendering of a scene begins, this is where the root render surface is stored
+     * @advanced
```
Zyie: We can remove all the advanced tags here as the class already has the advanced tag on it

src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
```diff
@@ -434,17 +506,17 @@ export class RenderTargetSystem<RENDER_TARGET extends GlRenderTarget | GpuRender
...
-     * @param originSrc - the origin of the copy
-     * @param originSrc.x - the x origin of the copy
-     * @param originSrc.y - the y origin of the copy
-     * @param size - the size of the copy
-     * @param size.width - the width of the copy
-     * @param size.height - the height of the copy
-     * @param originDest - the destination origin (top left to paste from!)
-     * @param originDest.x - the x origin of the paste
-     * @param originDest.y - the y origin of the paste
+     * @param {RenderTarget} sourceRenderSurfaceTexture - the render surface to copy from
```
Zyie: We don't need to define the type, TypeDoc will do that for us

e.g `@param sourceRenderSurfaceTexture -`

src/scene/container/RenderGroupPipe.ts
```diff
@@ -99,12 +97,10 @@ export class RenderGroupPipe implements InstructionPipe<RenderGroup>
...
-            const worldTransformMatrix = tempMatrix
-                .identity()
-                .translate(
-                    -renderGroup._textureBounds.x,
-                    -renderGroup._textureBounds.y
-                );
+            const worldTransformMatrix = new Matrix().translate(
+                -renderGroup._textureBounds.x,
+                -renderGroup._textureBounds.y
+            );
```
Zyie, line 103: We shouldn't need to be creating a new matrix everytime. Was there a reason for this change?

src/scene/particle-container/canvas/CanvasParticleContainerPipe.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ParticleContainerPipe } from '../shared/ParticleContainerPipe';
+import { CanvasParticleContainerAdaptor } from './CanvasParticleContainerAdaptor';
+
+import type { CanvasRenderer } from '../../../rendering/renderers/canvas/CanvasRenderer';
+
+/**
+ * A render pipe for the ParticleContainer that uses the Canvas renderer.
+ * @internal
+ */
+export class CanvasParticleContainerPipe extends ParticleContainerPipe
```
Zyie, line 11: I don't believe this class is needed. Inside `CanvasRenderer` you should be able to update the `DefaultCanvasAdapters` and add the `CanvasParticleContainerAdaptor` directly.

src/scene/text-bitmap/BitmapTextPipe.ts
```diff
@@ -174,7 +178,7 @@ export class BitmapTextPipe implements RenderPipe<BitmapText>
 
                     context.texture(
                         texture,
-                        tint ? tint : 'black',
+                        tint ?? 0xFFFFFF,
```
Zyie: As per the visual tests feedback this change should be reverted

src/scene/text/canvas/utils/getCanvasFillStyle.ts
```diff
@@ -32,15 +32,15 @@ export function getCanvasFillStyle(
     // Solid color fill
     if (fillStyle.texture === Texture.WHITE && !fillStyle.fill)
     {
-        return Color.shared.setValue(fillStyle.color).setAlpha(fillStyle.alpha ?? 1).toHexa();
+        return Color.shared.setValue(fillStyle.color).setAlpha(fillStyle.alpha ?? 1).toRgbaString();
```
Zyie: What is the reason for making these changes here?

src/scene/text/canvas/CanvasTextMetrics.ts
```diff
@@ -271,10 +271,11 @@ export class CanvasTextMetrics
         const fontProperties = CanvasTextMetrics.measureFont(font);
 
         // fallback in case UA disallow canvas data extraction
-        if (fontProperties.fontSize === 0)
+        if (fontProperties.fontSize <= 0)
```
Zyie: As per the visual tests feedback this change should be reverted

src/scene/text/canvas/CanvasTextPipe.ts
```diff
@@ -83,6 +83,21 @@ export class CanvasTextPipe implements RenderPipe<Text>
     {
         const batchableText = this._getGpuText(text);
 
+        if (text._didTextUpdate)
```
Zyie: What is the need for this change? If this was a bug with text rendering then a separate PR for fixing this up would be better!

Zyie (changes requested):
Hey @krzys 

Thank you for putting this together, this is really amazing work!

Since this change is quite large, I’m going to split the review into a few phases to make it more manageable:

- **Phase 1:** General architecture review and cleanup  
- **Phase 2:** Review of the actual logic changes for the canvas renderer  
- **Phase 3:** A final pass over any remaining bits and pieces that need addressing

Let me know if you have any questions, or if you need help with anything!

Below are the changes I would like you to make but don't fit into a specific comment.

---

### Visual Tests

I believe we’re once again running into the issue where our visual tests only work correctly on macOS. This affects text‑related tests, as text rendering differs slightly between operating systems.

What I’d like you to do is revert the changes made to `BitmapTextPipe` and `CanvasTextMetrics`. Doing so passes the tests locally for me.

For local testing, you can update this [line](https://github.com/krzys/pixijs/blob/61cbb8dd88232c2bb91fd8617c40b4c8f44519f8/tests/visual/visuals.test.ts#L140) by increasing the value of `scene.data.pixelMatchLocal ?? 100` to something high enough for all tests to pass for you

This isn’t a perfect solution, but it will allow you to test locally while we rely on CI for the “real” test validation.

---

### Systems / Pipes / Adapters

There’s a recurring pattern I want to call out here rather than commenting on every individual instance.

There are quite a few checks like `if (renderer.type === RendererType.CANVAS)`. 

We should avoid having any canvas renderer specific code inside the WebGL or WebGPU renderers to maximise tree shaking.

If a system, pipe, or adapter isn’t needed, it should be removed from the class’s `extensions` object. 
If it *is* needed, we should either follow the adapter pattern or create a canvas‑specific system (potentially sharing an abstract base system).

src/scene/container/RenderGroupPipe.ts
krzys: Pretty weird issue here.

Without this change:
Image

After:

Image

Zyie: hmm interesting... summoning @GoodBoyDigital!

src/scene/text/canvas/CanvasTextPipe.ts
krzys: Not needed, some leftovers from debugging this cacheAsTexture/text/container-thingy issue https://github.com/pixijs/pixijs/pull/11815#discussion_r2676459736

src/scene/container/RenderGroupPipe.ts
krzys: I think the issue is that `_executeCacheAsTexture` call is repeated during a frame and reuses `tempMatrix`, so `worldTransformMatrix` gets mutated.

src/filters/defaults/color-matrix/ColorMatrixFilter.ts
```diff
@@ -57,6 +57,9 @@ export type ColorMatrix = ArrayFixed<number, 20>;
  */
 export class ColorMatrixFilter extends Filter
 {
+    private _canvasFilterParts: string[] | null = null;
+    private _pendingCanvasFilter: { part: string; multiply: boolean } | null = null;
```
Zyie: We seem to be doing a lot of extra work in this filter to accommodate the canvas renderer. I wonder if there is a different way of going about this

@bigtimebuddy @GoodBoyDigital any thoughts on this?

src/filters/FilterPipe.ts
```diff
@@ -54,6 +54,8 @@ export class FilterPipe implements InstructionPipe<FilterInstruction>
 
     public execute(instruction: FilterInstruction)
     {
+        if (!this._renderer.filter) return;
```
Zyie: In what situation would filters system not exist?

src/scene/text/canvas/utils/getCanvasFillStyle.ts
```diff
@@ -40,7 +40,7 @@ export function getCanvasFillStyle(
         const pattern = context.createPattern(fillStyle.texture.source.resource, 'repeat');
         const tempMatrix = fillStyle.matrix.copyTo(Matrix.shared);
 
-        tempMatrix.scale(fillStyle.texture.frame.width, fillStyle.texture.frame.height);
+        tempMatrix.scale(fillStyle.texture.source.pixelWidth, fillStyle.texture.source.pixelHeight);
```
Zyie, line 47: I think this change is correct, however, we should keep a note that using a fill pattern from a spritesheet doesn't seem to be working as expected anyway

src/scene/view/ViewContainer.ts
```diff
@@ -210,8 +213,27 @@ export abstract class ViewContainer<GPU_DATA extends GPUData = any> extends Cont
...
+            {
+                const renderableName = this.constructor?.name ?? 'ViewContainer';
+
+                warn(
+                    `Renderer missing render pipe "${this.renderPipeId}" for ${renderableName}; `
+                    + 'skipping renderable.'
+                );
+                missingRenderPipeWarnings.add(renderPipeKey);
+            }
+        }
```
Zyie: While this change is certainly helpful, I’m not sure this is the right place to introduce this level of protection, given that this is hot code. In this area, we generally assume a higher level of familiarity from users who are creating custom renderables

Zyie (commented):
Hey @krzys I made a few changes to your branch just to fix up a couple of things. Overall this is looking incredible!

I just have a couple of extra questions for you and our team

src/filters/FilterPipe.ts
krzys: I swear this made sense once, haha. We can delete this line.

src/scene/view/ViewContainer.ts
krzys: That's fair. This was only for DX, I'm fine to delete it.

src/filters/defaults/color-matrix/ColorMatrixFilter.ts
krzys: I considered using SVG filters with `ctx.filter = 'url(#...)'`, but this doesn't work in some current, stable browsers (Firefox and Safari, I think) due to limited support. The current solution only works in simple cases, so it would be great to come up with something better.

Zyie: After discussing this with @GoodBoyDigital we are going to remove the alpha, blur, and color matrix changes and do this as a follow up PR. 

This feature was the only major blocker for us releasing this and since v7 didn't even support filters we can just add this back in once we have thought about it some more

bigtimebuddy:
@krzys Thank you so much! And thanks to @Zyie for getting this over the finish line. Really excited to see how folks will use this.

PR #11816, opened 2025-12-28
feat: Add ExternalSource Texture Source

astralarya:
##### Description of change

Enable using external gpu textures loaded from outside of Pixi (eg. Three.js).

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/gpu/texture/GpuTextureSystem.ts
+++ b/src/rendering/renderers/gpu/texture/GpuTextureSystem.ts
@@ -175,14 +175,20 @@ export class GpuTextureSystem implements System, CanvasGenerator
             const gpuData = source._gpuData[this._renderer.uid] as GPUTextureGpuData;
             const newTexture = source.resource as GPUTexture;
 
-            if (gpuData && gpuData.gpuTexture !== newTexture)
+            if (!gpuData)
+            {
+                if (newTexture)
+                {
+                    this._initSource(source);
+                }
+
+                return;
+            }
+
+            if (gpuData.gpuTexture !== newTexture)
             {
                 this._bindGroupHash[source.uid] = null;
                 gpuData.textureView = null;
-            }
-
-            if (gpuData)
-            {
                 gpuData.gpuTexture = newTexture;
             }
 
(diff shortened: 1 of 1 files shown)


astralarya:
Added support / tests for updating the TextureSource resource.  This is useful for when you want a dynamic external texture.

astralarya:
As a motivating example for why ExternalSource is needed, it is core to enabling my new library for Pixi.js/Three.js interop:
https://github.com/astralarya/pixi-three

It supports realtime zero copy GPUTexture sharing bidirectionally between Three and Pixi.

https://github.com/user-attachments/assets/4248d49d-3f93-47e6-b247-558f90ac3200

astralarya:
I have live examples of this being used on the documentation website for my pixi-three library:
https://astralarium.github.io/pixi-three/

The relevant lines where ExternalSource is being used are here:
https://github.com/astralarium/pixi-three/blob/main/src/three-scene.tsx#L245-L277

bigtimebuddy:
@astralarya thanks for doing this work, looks great! Sorry for the lack of response, holidays and new year has everyone quite busy. @GoodBoyDigital is going to get to this and we are eyeing to include it in the 8.16.0

astralarya:
No worries! I think there may be a few subtleties around lifecycle that should be double checked, as I'm not as familiar with those systems. I think I've figured out a solution that at least works for pixi-three. I've tested in-place updates of the GPUTexture resource on the ExternalSource, resizing the texture, as well as mounting and unmounting the application.

The existing tests do cover the resizing case. Not sure how to make an unmounting test, but there were some texture view lifecycle subtleties I found during manual testing of my library that necessitated the later commits.

GoodBoyDigital:
hey @astralarya thanks for you patience on this one. Was nnoodling on this for while! haha.
As its quite an edge case - i wanted to see if we could reduce the amount of code we had to touch in the renderer and isolate the functionality to the `ExternalSource`.

The key difference in the API is that you now need to pass the renderer on creation, which lets us do things like validate the gpu texture is from the same context - and also set it correctly.

The updated PR can be found [here](https://github.com/pixijs/pixijs/pull/11846)

Does this work ok for you? Would love to get your feedback! thanks!

Zyie:
Closing in favor of #11846

Thank you for all the work you put into this @astralarya! 
I'll make sure you are credited in the release notes for this feature!

PR #11827, opened 2026-01-07
feat: tagged text

Zyie:
### Summary

Adds rich text formatting support to PixiJS through a simple XML-like tag markup system. This allows applying different styles to portions of text within a single `Text` object using the new `tagStyles` property on `TextStyle`.

### Implementation Details

- Nested tags inherit styles from parent tags and can override specific properties
- Tags are only parsed when `tagStyles` has entries and text contains `<`
  - `tagStyles` is `undefined` by default which avoids this being a breaking change 
- Unknown tags are treated as literal text (no error thrown)
- Mismatched closing tags are treated as literal text
- Works with all existing `TextStyle` features (alignment, trimming, bounds, etc.)

#### Basic Usage

```typescript
const text = new Text({
    text: 'Hello <red>World</red>!',
    style: {
        fontSize: 24,
        fill: 'white',
        tagStyles: {
            red: { fill: 'red' }
        }
    }
});
```

#### Multiple Tags with Different Styles

```typescript
const text = new Text({
    text: '<bold>Important:</bold> This is <highlight>highlighted</highlight> text',
    style: {
        fontFamily: 'Arial',
        fontSize: 28,
        fill: 'white',
        tagStyles: {
            bold: { fontWeight: 'bold', fill: 'yellow' },
            highlight: { fill: 'cyan', fontSize: 32 }
        }
    }
});
```

#### Nested Tags with Style Inheritance

```typescript
const text = new Text({
    text: '<outer>Outer text <inner>inner inherits + overrides</inner> back to outer</outer>',
    style: {
        fontSize: 28,
        fill: 'white',
        tagStyles: {
            outer: { fill: 'yellow', fontWeight: 'bold' },
            inner: { fill: 'red', fontStyle: 'italic' }  // inherits bold from outer
        }
    }
});
```


src/scene/text/TextStyle.ts
```diff
@@ -1179,6 +1266,20 @@ export class TextStyle extends EventEmitter<{
         return `${this.uid}-${this._tick}`;
     }
 
+    /**
+     * Returns the CSS font string for this style, cached for performance.
+     * @returns CSS font string
+     */
+    public get fontString(): string
```
GoodBoyDigital: should this be tagged as internal?

GoodBoyDigital (approved):

PR #11833, opened 2026-01-15
fix: improves tree-shaking by optimizing module imports

Zyie:
Improves import organization across the codebase to increase tree-shaking efficiency:
- Moves module-level initialization imports to individual component files, allowing unused modules to be excluded from the final bundle
- Streamlines the `browserAll.ts` and `webworkerAll.ts` files to only import necessary modules
- Dynamically import `BitmapFont` to avoid text being bundled

This change helps reduce the final bundle size by up to 15% on a simple project.

### DIFF
--- a/src/dom/DOMContainer.ts
+++ b/src/dom/DOMContainer.ts
@@ -1,8 +1,12 @@
 /* eslint-disable no-restricted-globals */
+import { extensions } from '../extensions/Extensions';
 import { Point } from '../maths/point/Point';
 import { ViewContainer, type ViewContainerOptions } from '../scene/view/ViewContainer';
+import { DOMPipe } from './DOMPipe';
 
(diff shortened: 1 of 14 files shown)


src/dom/DOMContainer.ts
```diff
@@ -1,9 +1,13 @@
 /* eslint-disable no-restricted-globals */
+import { extensions } from '../extensions/Extensions';
 import { Point } from '../maths/point/Point';
 import { ViewContainer, type ViewContainerOptions } from '../scene/view/ViewContainer';
+import { DOMPipe } from './DOMPipe';
 
 import type { PointData } from '../maths/point/PointData';
 
+extensions.add(DOMPipe);
```
Zyie, line 9: This was added directly as we get circular imports otherwise

bigtimebuddy (approved):

Issue #11620, opened 2025-08-17
Why is the updated texture incorrect when using HTMLText to batch update textures?


daydaylee1227:
### PixiJS version

v8.12.0

### Link to minimal reproduction

https://www.pixiplayground.com/#/edit/8dcXrTzWjQsGRCCS2ez7Y

### Steps to reproduce

This is demo url: https://www.pixiplayground.com/#/edit/8dcXrTzWjQsGRCCS2ez7Y
In actual development, there are scenarios where we need to use this rich text HTMLText node to render rich text. If we select 10 different nodes at the same time and then update their corresponding widths and heights in the same batch, triggering texture regeneration, we find that the corresponding textures are generated incorrectly. In the demo example, we use a container to randomly generate 9 containers, then insert HTMLText nodes, and change the width of these containers through code to achieve our required usage scenario. 

![Image](https://github.com/user-attachments/assets/0eb7ba88-a38d-4def-b370-229af97e2032)

As shown in the image above, when manually simulating this batch processing scenario through code, a texture update disorder issue occurs. Essentially, it's still a problem with incorrect texture rendering when the Container updates, resulting in the Container failing to render the rich text data as expected.

### What is expected?

The expectation is that the HTMLText nodes rendered under the corresponding Container will display the rich text content correctly.

### What is actually happening?

The usage scenario is batch processing HTMLText nodes. For example, when the width and height of the external container change, the textures need to be updated again, such as updating and processing in the same frame. If you check with the SpectorJS plugin, you'll find that the texture generation inside is incorrect.

### Environment

I'm using the latest version of the Chrome browser. My computer is an M4, and I'm also using the latest version of PixiJS.

### Any additional comments?

I believe this issue exists in many versions, and I'm just using one of them for testing.

mayakwd:
Related to: #11014

daydaylee1227:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

daydaylee1227:
@bigtimebuddy @Zyie It seems that someone has reported a similar bug before, and you have also marked it as a bug, which indicates that this scenario does exist. It has been a long time. Regarding this situation, if we want to solve the problem of rich text rendering based on your PixiJS framework, can you provide a specific time for the fix instead of delaying it?

bigtimebuddy:
[...]

PR #11834, opened 2026-01-16
fix: HTMLText prevent texture cache collision on resolution updates

### LINKED ISSUE #11790: HTML Text: incorrect texture reference when batch updating resolution
### PixiJS version

8.14.3

https://www.pixiplayground.com/#/edit/73wQLO-S2oJb-1YqLlvhl

### Steps to reproduce

- Create multiple HTMLText instances
- Update the resolution for all of them
- After a short time, some `HTMLText` instances start rendering with the texture of another text or no texture at all

The issue isn’t 100% deterministic, so in the repro I trigger the resolution update on every tick to make it show up more reliably. My real world app will only update it once, but only sometimes the texture/reference will get lost

### What is expected?

Updating the resolution of multiple `HTMLText` instances should not cause them to lose their internal texture references.

Texture reference is lost/wrong

### Environment

Chromium 139.0.7258.156
macOS 13.5.2

### Any additional comments?

An issue has been reported for a similar bug when updating dimension of text container. See https://github.com/pixijs/pixijs/issues/11620

It has been fixed and release, although the issue is still happening for resolution updates.

subhu339:
This PR fixes an issue where HTMLText instances could render with incorrect or missing textures when updating their resolution in batch or within the same frame.

The root cause was a texture cache collision: the cache key used for text textures did not include the text resolution or auto-resolution state. As a result, multiple HTMLText instances could incorrectly reuse textures generated at a different resolution.
The fix updates AbstractText.styleKey to include both resolution and autoResolution, ensuring that texture caching remains unique and stable when resolution changes dynamically.

Fixes #11790
Related to #11620


bigtimebuddy (changes requested):
Using this build, I created a reproduction from the original issue, but I'm seeing the same flickering result. 

https://stackblitz.com/edit/oyee1q8f?file=src%2Findex.js

Jan-16-2026 11-09-21

subhu339:
Thanks for testing and the reproduction link!
I’ll investigate the remaining flickering using the StackBlitz example and update the PR accordingly

subhu339:
https://github.com/user-attachments/assets/3afbd203-40ac-491d-a3c3-224526ceed90

Update: adjusted HTMLText texture cache key to avoid collisions when resolution
changes.

Tested locally with a minimal repro toggling resolution. Major flickering is
resolved. A very small vertical jitter remains for glyph “5”, which appears
related to font metrics rather than cache behavior.
Video attached.

subhu339:
@bigtimebuddy, Thanks for testing! I’ve updated the fix to further stabilize the HTMLText texture cache key. The major flickering during resolution updates is now resolved.

I re-tested with the StackBlitz repro; only a very minor vertical jitter remains, which seems related to font metrics rather than caching. Please take another look.

subhu339:
@bigtimebuddy please review!

PR #11835, opened 2026-01-16
fix: BitmapFontManager getFont emitting TextStyle update event

### LINKED ISSUE #11723: BitmapFontManagerClass.getFont emit TextStyle update event
### PixiJS version

8.12+

https://jsfiddle.net/j0xtc741/3/

### Steps to reproduce

		let text = new Text({ text:'testst' });
		text.style.on('update', () => {
			console.log('update');
		});
		text.style.lineHeight = 20;
		text.style.stroke = { width:4, color:0xFFFF00 };
		app.stage.addChild(text).position.set(20, 20);
		console.log('111');
		BitmapFontManager.getFont(text.text, text.style);
		console.log('222');

The console will have 111, update, 222.

Stefans.A* 05.08.2025 11:39 fix: `lineHeight` calculation for multi-line layout for `BitmapText` (#11531)
a2c09867b834f7a37782f6adbab996dd20e9ce8c

            const styleCopy = Object.create(style);

            // Override the lineHeight, let the BitmapFont calculate the lineHeight
            // from the fontMetrics instead using a custom lineHeight from BitmapText parameter
            styleCopy.lineHeight = 0;

According to the author's idea, this method provides object cloning. But the fact is that the TextStyle properties are setter/getter. And the styleCopy.lineHeight = 0 code will call setter style.lineHeight but with this === styleCopy. Since the object has a style prototype, all update listeners will receive this event as a side event. We are getting a kind of false positive change. And if there are custom handlers, then this can lead to problems, as in my case.

Interestingly, in 8.14, since checking for changes in the internal state of TextStyle was added, you first need to change this value. Previously, this was not required, thus this problem is partly even more hidden.

### What is expected?

For example like this:
Object.defineProperty(styleCopy, 'lineHeight', {
  value: 0,
  writable: true,
  configurable: true,
  enumerable: true
});

### Environment

8.12+
All browsers

subhu339:
##### Description of change
Fixes an issue where BitmapFontManager.getFont() could emit a TextStyle update event as a side effect.
Instead of assigning lineHeight via the setter on a prototype-based style clone, the value is now defined directly on the cloned style instance to avoid triggering unintended updates.
##### Pre-Merge Checklist
Tests and/or benchmarks are included (behavior verified via manual reproduction)
 Documentation is changed or added (not required for this internal fix)
 Lint process passed (npm run lint)
 Tests passed (npm run test)

Fixes #11723


src/scene/text-bitmap/BitmapFontManager.ts
```diff
@@ -326,16 +326,20 @@ class BitmapFontManagerClass
...
             // from the fontMetrics instead using a custom lineHeight from BitmapText parameter
-            styleCopy.lineHeight = 0;
-
+              Object.defineProperty(styleCopy, 'lineHeight', {
+                  value: 0,
+                  writable: true,
+                  configurable: true,
+                  enumerable: true,
+              }); 
+
```
Zyie: I think the fix can be as simple as `styleCopy['_lineHeight'] = 0` to avoid the update call?

Zyie (changes requested):
hey @subhu339 can we get some tests added for this to verify we no longer emit extra events

subhu339:
Added unit tests to verify that BitmapFontManager.getFont no longer emits a TextStyle update event.
Also updated the implementation to set _lineHeight directly on the cloned style instance, avoiding the setter and preventing unintended updates.

Zyie (approved):

PR #11838, opened 2026-01-17
fix: prevent autoGenerateMipmaps from affecting TexturePool render targets

### LINKED ISSUE #11717: Enabling autoGenerateMipmaps breaks filter screen sampling
### PixiJS version

8.14.0

https://codesandbox.io/p/sandbox/hardcore-rosalind-89vxfs

### Steps to reproduce

Open the link and observe the bunny fading away with the pinch filter when `PIXI.TextureSource.defaultOptions.autoGenerateMipmaps` is set to `true`.

### What is expected?

Setting `PIXI.TextureSource.defaultOptions.autoGenerateMipmaps` to `true` shouldn't affect filter sampling as that doesn't make any sense. The bunny should be fully visible and pinched correctly:

Image

Enabling mipmaps causes the `TexturePool` to create render textures with mipmap levels which are never populated with any real data. This causes filters that do scaled UV sampling to sample invalid or "corrupt" data:

Image

### Environment

- **Browser & Version**: *Chrome Version 141.0.7390.67 (Official Build) (64-bit)*
- **OS & Version**: *Windows 11 Home 26200.6725*

### Any additional comments?

As per #11304 there is no other way to generate mipmaps for `Text` than to enable the mipmap generation by default. While it would be possible to set `PIXI.TexturePool.textureOptions.autoGenerateMipmaps = false` to fix the filter issue, unfortunately the `TexturePool` is also used by the text rendering which will also disable it for `Text` instances.

Ideally I would like to have the mipmaps enabled for text and disabled for filters. That doesn't seem to be possible due to this issue.

doyoonear:
##### Description of change

Fixes #11717 where setting `TextureSource.defaultOptions.autoGenerateMipmaps = true` causes filter rendering to break.

**The Problem:**
When users set `TextureSource.defaultOptions.autoGenerateMipmaps = true`, TexturePool textures (used as render targets for filters) were also affected because `createTexture()` did not explicitly set `autoGenerateMipmaps`. This caused:

1. Pool textures inherited `autoGenerateMipmaps: true` from `defaultOptions`
2. GPU textures were created with multiple mipmap levels
3. Filters render to level 0, but mipmaps are never updated (render targets don't trigger `update` event)
4. GPU samples from uninitialized mipmap levels, causing visual corruption

**The Fix:**
Explicitly set `autoGenerateMipmaps: false` in `TexturePool.createTexture()` to ensure pool textures are never affected by the global default setting.

| Before | 
|:------:|
| 스크린샷 2026-01-17 오후 4 23 54 | 

| After |
|:------:|
| 스크린샷 2026-01-17 오후 4 27 49 |

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/texture/TexturePool.ts
+++ b/src/rendering/renderers/shared/texture/TexturePool.ts
@@ -65,6 +65,7 @@ export class TexturePoolClass
             resolution: 1,
             antialias,
             autoGarbageCollect: false,
+            autoGenerateMipmaps: false,
         });
 
         return new Texture({


bigtimebuddy (changes requested):
Thanks for submitting, but this doesn't actually solve the underlying problem, just masks it. Changing default values is also considered a breaking change, which this is doing.

PR #11840, opened 2026-01-20
fix: Clear page mesh data and update texture references in BitmapText

ViktorVovk:
Good day. This PR aims to fix an issue with rendering bitmap fonts when BitmapText instances are reused from a pool.
I would like to describe the problem I encountered in my project

##### Description of change
We have a scene that contains a UI slider with around 8 different slides, which can be switched using left and right arrows. Each slide is implemented as a separate layer that contains a certain number of BitmapFont objects with different text values. Using layers allows us to easily handle slide transition logic by simply setting the required layer to visible and the others to invisible. It is worth noting that the total number of BitmapText instances is around 57.

After we leave this scene, we call `Scene::unload()`, which in turn calls `destroy()` on all objects in the scene. This eventually leads to `BitmapText::destroy`. As is known, during the execution of this method we populate the `pageMeshDataPool` array, where mesh data is stored for reuse. After leaving the scene, around 57 `_activePagesMeshData` instances remain in the pool.

In the next scene, completely different bitmap fonts reuse instances from this pool, and we encountered an issue where some of them are rendered incorrectly (the result looks terrible). When a mesh is taken from the pool, its WebGL buffers (geometry.buffers) already exist and still contain references to TypedArrays from the previous usage. Performing a deeper cleanup helps to avoid this problem.

I would appreciate your review and feedback. If you see an alternative or better approach to solving this problem, I would be glad to read your comments and suggestions.

##### Pre-Merge Checklist

### DIFF
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -921,14 +921,16 @@ export class BitmapText extends Container
         for (const pageMeshData of this._activePagesMeshData)
         {
             this.removeChild(pageMeshData.mesh);
-            pageMeshData.vertices = null;
-            pageMeshData.uvs = null;
-            pageMeshData.indices = null;
-            pageMeshData.index = 0;
-            pageMeshData.indexCount = 0;
-            pageMeshData.vertexCount = 0;
-            pageMeshData.uvsCount = 0;
-            pageMeshData.total = 0;
+            Object.assign(pageMeshData, {
+                vertices: null,
+                uvs: null,
+                indices: null,
+                index: 0,
+                indexCount: 0,
+                vertexCount: 0,
+                uvsCount: 0,
+                total: 0,
+            });
         }
         pageMeshDataPool.push(...this._activePagesMeshData);
         this._activePagesMeshData = [];
--- a/packages/text-bitmap/src/BitmapText.ts
+++ b/packages/text-bitmap/src/BitmapText.ts
@@ -918,11 +918,19 @@ export class BitmapText extends Container
         const pageMeshDataPool = data.distanceFieldType === 'none'
             ? pageMeshDataDefaultPageMeshData : pageMeshDataMSDFPageMeshData;
 
-        pageMeshDataPool.push(...this._activePagesMeshData);
         for (const pageMeshData of this._activePagesMeshData)
         {
             this.removeChild(pageMeshData.mesh);
+            pageMeshData.vertices = null;
+            pageMeshData.uvs = null;
+            pageMeshData.indices = null;
+            pageMeshData.index = 0;
+            pageMeshData.indexCount = 0;
+            pageMeshData.vertexCount = 0;
+            pageMeshData.uvsCount = 0;
+            pageMeshData.total = 0;
         }
+        pageMeshDataPool.push(...this._activePagesMeshData);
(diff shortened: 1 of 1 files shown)


packages/text-bitmap/src/BitmapText.ts
```diff
@@ -918,11 +918,19 @@ export class BitmapText extends Container
...
         {
             this.removeChild(pageMeshData.mesh);
+            pageMeshData.vertices = null;
+            pageMeshData.uvs = null;
+            pageMeshData.indices = null;
+            pageMeshData.index = 0;
+            pageMeshData.indexCount = 0;
+            pageMeshData.vertexCount = 0;
+            pageMeshData.uvsCount = 0;
+            pageMeshData.total = 0;
```
bigtimebuddy: Nit. We could make this a little less verbose using Object.assign.

```suggestion
            Object.assign(pageMeshData, {
                vertices: null,
                uvs: null,
                indices: null,
                index: 0,
                indexCount: 0,
                vertexCount: 0,
                uvsCount: 0,
                total: 0,
            });

```

ViktorVovk: Thank you for the review. You’re right — that makes sense. I’ve updated the PR.

bigtimebuddy:
Thanks for putting this PR together! Your description makes sense, but do you have a before and after verification? It's helpful to see a reproduction.

ViktorVovk:
> Thanks for putting this PR together! Your description makes sense, but do you have a before and after verification? It's helpful to see a reproduction.

No problem. I’ve created a simple sample application that demonstrates the issue. https://github.com/ViktorVovk/bitmapfonts-sample

The flow is straightforward: the application starts, then a container is created with 10 instances of the same bitmap font, each displaying random text. Next, we click the button at the top — the container is removed from the stage and destroy() is called on its objects. After that, a new container is created with several instances of a different bitmap font, and the result can be observed. If you refresh the application multiple times, you can see that the result is always different.

I would also like to point out some nuances that I did not mention earlier. For the second font (in my real application), we were using compressed textures, and the issue was observed specifically with such textures. The textures were compressed using AssetsPack, and in this sample application there is a script called packScript.js in the root directory that performs this process using the same configuration that I use in my real project. I used this script to generate the textures for this sample.

bigtimebuddy:
Have you tested this in v8 yet? Is it a problem there too?

ViktorVovk:
> Have you tested this in v8 yet? Is it a problem there too?

I apologize for not providing this information earlier.
The issue is reproducible in 
Google Chrome Version 144.0.7559.59 (Official Build) (arm64) as well as 
Safari Version 26.2 (21623.1.14.11.9) on a MacBook with Apple M1 Pro,
and also on the latest iOS version on an iPhone 17.

bigtimebuddy:
No, sorry, I mean does this reproduce this in PixiJS v8.x. I noticed this PR is for v7.x

ViktorVovk:
> No, sorry, I mean does this reproduce this in PixiJS v8.x. I noticed this PR is for v7.x

I checked it in https://github.com/ViktorVovk/bitmapfonts-sample/tree/v8-sample, and there isn't this issue here.


I’m working with **version 7**, and in the current project there is no possibility to migrate to **version 8**.)

PR #11846, opened 2026-01-23
feat: external texture support

GoodBoyDigital:
## Description

Refactors external texture support to use a cleaner architecture where `ExternalSource` requires the renderer in its constructor and pre-populates `_gpuData` directly. This eliminates all special-case checks scattered throughout the texture systems.

**Key improvements:**
- ExternalSource pre-populates _gpuData on construction (no special checks needed)
- Adds updateGPUTexture() method for swapping external textures
- Validates WebGL context ownership on construction
- Removes 5 special-case checks from GlTextureSystem and GpuTextureSystem
- Simplified ExternalSourceOptions (no longer extends TextureSourceOptions)

**API:**
```typescript
// WebGPU - dimensions auto-detected
const texture = new Texture({
    source: new ExternalSource({
        resource: gpuTexture,
        renderer: renderer,
    })
});

// WebGL - dimensions required (WebGLTexture is opaque)
const texture = new Texture({
    source: new ExternalSource({
        resource: glTexture,
        renderer: renderer,
        width: 512,
        height: 512,
    })
});

// Update texture reference
source.updateGPUTexture(newGpuTexture);
```

## Tests

- [x] Visual tests for external texture rendering (WebGL1, WebGL2, WebGPU)
- [x] Visual tests for updating external textures with dimension changes
- [x] Build and type checks pass
- [x] Lint process passed


astralarya:
I am fine with passing the renderer, as I always know which renderer I want to use.  Would it be possible to allow the resource to be null/undefined at the start and then later on it gets assigned a GPUTexture?  That would simplify a lot of code on my end.  Otherwise, this solution looks great and will work for PixiThree.

astralarya:
As an example of the syntax I'm looking for is:

```ts
sprite.texture = new Texture({
  dynamic: true,
  source: new ExternalSource({
    renderer: app.renderer,
    label: "three-scene",
    // resource: unspecified
  }),
});

// ... later on ...

sprite.texture.source.resource = aGPUTexture;
sprite.texture.source.update();
```

GoodBoyDigital:
sure - we can have like a placeholder GPU texture that means you can update it at a later point

GoodBoyDigital:
placeholder added!

Zyie (approved):

astralarya:
Hooray! Although, correct me if I'm wrong, but it looks like `resource` is still required?

astralarya:
I created a PR #11860 that implements the placeholder texture

Issue #11690, opened 2025-09-24
DOMContainer translate is offset when inside a positioned DOM parent


danlite:
### PixiJS version

8.12.0

### Link to minimal reproduction

[...]

PR #11852, opened 2026-01-27
fix: account for root DOM bounds when positioning elements

subhu339:
#11690
##### Description of change

Fixes an offset issue when positioning DOM elements if the PixiJS canvas is not at the top-left of the page.  
The DOM container’s bounding box is now taken into account so elements line up correctly with their display objects.

##### Pre-Merge Checklist

<details>

##### Fixes
- Fixed DOM element positioning offset when the PixiJS canvas is not at the top-left of the page. DOM elements now correctly align with their display objects by accounting for the DOM container's bounding position.

</details>


GoodBoyDigital (changes requested):
Thanks for this PR! A few points
- It would be great if you could provide tests to cover this
- `getBoundingClientRect` is sooo expensive (as it causes the DOM do some really intensive stuff). We shoudl try to find a way to handle this without degrading performance.

thanks!

PR #11853, opened 2026-01-28
fix: allow lineHeight to be set to 0

Riphal:
## Summary

This PR fixes an issue where setting `lineHeight` to `0` in text styles was being ignored because of truthy checks that treated `0` as falsy.

### Changes

- Changed `lineHeight` checks from truthy checks (`if (style.lineHeight)`) to explicit undefined checks (`if (style.lineHeight !== undefined)`)
- Updated the default `lineHeight` value from `0` to `undefined` in `TextStyle.ts`

### Affected components

- `BitmapText`
- `HTMLText`  
- `CanvasText`

### Problem

Previously, setting `lineHeight: 0` would fall back to default values because JavaScript treats `0` as falsy:

// Before: 0 is falsy, so this would use the default
if (style.lineHeight) { ... }

// After: explicit check allows 0 to be a valid value
if (style.lineHeight !== undefined) { ... }


src/scene/text-bitmap/utils/bitmapTextSplit.ts
```diff
@@ -31,7 +31,7 @@ export function bitmapTextSplit(
     const chars: BitmapText[] = [];
     const words: Container[] = [];
     const lines: Container[] = [];
-    const lineHeight = style.lineHeight ? style.lineHeight : font.lineHeight * scale;
+    const lineHeight = style.lineHeight !== undefined ? style.lineHeight : font.lineHeight * scale;
```
bigtimebuddy: What about this instead?

```suggestion
    const lineHeight = style.lineHeight ?? font.lineHeight * scale;
```

src/scene/text/canvas/CanvasTextMetrics.ts
```diff
@@ -303,7 +303,7 @@ export class CanvasTextMetrics
             width += style.dropShadow.distance;
         }
 
-        const lineHeight = style.lineHeight || fontProperties.fontSize;
+        const lineHeight = style.lineHeight !== undefined ? style.lineHeight : fontProperties.fontSize;
```
bigtimebuddy:
```suggestion
        const lineHeight = style.lineHeight ?? fontProperties.fontSize;
```

src/scene/text/TextStyle.ts
```diff
@@ -756,7 +756,7 @@ export class TextStyle extends EventEmitter<{
         fontWeight: 'normal',
         leading: 0,
         letterSpacing: 0,
-        lineHeight: 0,
+        lineHeight: undefined,
```
bigtimebuddy: Could we use null instead of undefined? Eg, stroke

bigtimebuddy (approved):
Might be good to have a visual test for 0 lineHeight, also a minor suggestion, this looks good. Thank you!

src/scene/text/TextStyle.ts
Riphal: Yes will change that today

Riphal:
> Might be good to have a visual test for 0 lineHeight, also a minor suggestion, this looks good. Thank you!

@bigtimebuddy done! It should be good to merge now. I just resolved all conflicts.

bigtimebuddy (approved):

PR #11856, opened 2026-01-29
fix: Video Loader Not Catching Load Errors

### LINKED ISSUE #11845: Video Loader Not Catching Load Errors
### PixiJS version

8.15.0

https://jsfiddle.net/7o5f2wh3/

### Steps to reproduce

`Assets.load()` does not capture errors in loading videos.

```ts
  await Assets.load("fake_file.webp").catch(e => {
    console.log("FAILED TO LOAD WEBP FILE")
  })

  await Assets.load("fake_file.mp4").catch(e => {
    console.log("FAILED TO LOAD MP4 FILE")
  })
```

### What is expected?

The loader, based on the documentation, should return the same kind of error regardless of the file type that failed to load.

For the example above, the console output should be:
```
FAILED TO LOAD WEBP FILE
FAILED TO LOAD MP4 FILE
```

Currently, it is not possible to catch errors in loading videos.

The output of the code example is this:
```
FAILED TO LOAD WEBP FILE
```

### Environment

- **Browser & Version**: *e.g. Firefox 146.0.1*
- **OS & Version**: *e.g. Ubuntu 25.10*

### Any additional comments?

I assume it's a missing event handler. I can see that the preload functionality wires in the `error` handler, but experimenting with forcing preload leads to the same outcome. It's possible that another event altogether is fired on a 404 or network error.

stargazer-2697:
##### Description of change

Fixes #11845 by adding error event handlers to both the video element and source element. Video error events can fire on the `<source>` element, not just the parent `<video>` element, so both need listeners.

##### Pre-Merge Checklist

Automated tests are difficult due to the reliance on DOM events and callbacks. The jsFiddle in the linked issue was used to validate the fix, although it should be noted that the error message is still logged to `console.error` in Chrome 143.0.7499.193 despite bring handled by `Promise::catch`.

### DIFF
--- a/src/assets/loader/parsers/textures/loadVideoTextures.ts
+++ b/src/assets/loader/parsers/textures/loadVideoTextures.ts
@@ -274,12 +274,12 @@ export const loadVideoTextures = {
                 reject(event);
             };
 
-            function cleanup(): void
+            const cleanup = (): void =>
             {
                 videoElement.removeEventListener('canplay', onCanPlay);
                 videoElement.removeEventListener('error', onError);
                 sourceElement.removeEventListener('error', onError);
-            }
+            };
 
             if (options.preload && !options.autoPlay)
             {
--- a/src/assets/loader/parsers/textures/loadVideoTextures.ts
+++ b/src/assets/loader/parsers/textures/loadVideoTextures.ts
@@ -252,13 +252,13 @@ export const loadVideoTextures = {
 
         // this promise will make sure that video is ready to play - as in we have a valid width, height and it can be
         // uploaded to the GPU. Our textures are kind of dumb now, and don't want to handle resizing right now.
-        return new Promise((resolve) =>
+        return new Promise((resolve, reject) =>
         {
             const onCanPlay = async () =>
             {
                 const base = new VideoSource({ ...options, resource: videoElement });
 
-                videoElement.removeEventListener('canplay', onCanPlay);
+                cleanup();
 
                 if (asset.data.preload)
                 {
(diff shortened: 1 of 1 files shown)


src/assets/loader/parsers/textures/loadVideoTextures.ts
```diff
@@ -268,12 +268,27 @@ export const loadVideoTextures = {
...
+                cleanup();
+                reject(event);
+            };
+
+            function cleanup(): void
+            {
+                videoElement.removeEventListener('canplay', onCanPlay);
+                videoElement.removeEventListener('error', onError);
+                sourceElement.removeEventListener('error', onError);
+            }
```
bigtimebuddy: nit, we prefer fat arrows for inline functions to maintain a single context.

```suggestion
            const cleanup = (): void =>
            {
                videoElement.removeEventListener('canplay', onCanPlay);
                videoElement.removeEventListener('error', onError);
                sourceElement.removeEventListener('error', onError);
            };
```

bigtimebuddy (approved):
Thank you!

src/assets/loader/parsers/textures/loadVideoTextures.ts
stargazer-2697: @bigtimebuddy Avoiding the lint error after switching to arrow functions is difficult due to the circular dependencies between the closures. I've updated the code to use a function hoisting style, which is consistent with `preloadVideo` above. Is that acceptable?

https://github.com/pixijs/pixijs/blob/ab7dae3108fb0728e53f3d4828134d867a4eb820/src/assets/loader/parsers/textures/loadVideoTextures.ts#L47-L74

PR #11860, opened 2026-02-01
feat: add placeholder external texture

astralarya:
##### Description of change
Allow `resource` to be undefined and create a default placeholder texture.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/texture/sources/ExternalSource.ts
+++ b/src/rendering/renderers/shared/texture/sources/ExternalSource.ts
@@ -86,8 +86,8 @@ export class ExternalSource extends TextureSource<GPUTexture | WebGLTexture>
         }
 
         // Auto-detect dimensions for GPUTexture (WebGLTexture is opaque, requires explicit dimensions)
-        const width = options.width ?? (resource as GPUTexture)?.width ?? 1;
-        const height = options.height ?? (resource as GPUTexture)?.height ?? 1;
+        const width = options.width ?? (resource as GPUTexture).width;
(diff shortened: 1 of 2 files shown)


src/rendering/renderers/shared/texture/sources/ExternalSource.ts
```diff
@@ -76,7 +76,14 @@ export class ExternalSource extends TextureSource<GPUTexture | WebGLTexture>
...
     {
-        const { resource, renderer, label } = options;
+        const { renderer, label } = options;
+        let { resource } = options;
+
+        // Create a default placeholder texture if resource is null/undefined
+        if (!resource)
+        {
+            resource = ExternalSource._createPlaceholderTexture(renderer);
+        }
```
bigtimebuddy:
```suggestion
        // Create a default placeholder texture if resource is null/undefined
        resource ??= ExternalSource._createPlaceholderTexture(renderer);
```

astralarya:
I added a cleanup on destroy of the placeholder in WebGL.  I think WebGPU should cleanup on its own?

astralarya:
I decided to just cleanup in WebGPU as well

astralarya:
I'm happy with these changes if you think they are ready to merge.

Zyie:
hey @astralarya thanks for this, we had a bad merge from #11846 where the placeholder feature was not committed and added it back here #11861, so this pr is no longer needed

we have since released 8.16.0 which should have everything you need. If not please let us know!
thanks again for your work on this!

PR #11865, opened 2026-02-03
fix: prevent filter corruption with TexturePool mipmap separation

### LINKED ISSUE #11717: Enabling autoGenerateMipmaps breaks filter screen sampling

vkarponen:
##### Description of change

Fixes texture pool to prevent filter sampling corruption when `TextureSource.defaultOptions.autoGenerateMipmaps` is enabled globally for text rendering.

When users set `TextureSource.defaultOptions.autoGenerateMipmaps = true` to enable mipmaps for better text quality, filters were getting render textures with mipmap levels that were never populated with valid data. This caused filters doing scaled UV sampling (like the pinch filter) to sample corrupt/invalid data, resulting in visual artifacts (e.g., sprites fading away).

### Solution

Added an optional `autoGenerateMipmaps` parameter to `TexturePool.getOptimalTexture()` that allows callers to specify whether they need mipmaps. The mipmap setting is included in the pool key, ensuring textures with different mipmap configurations are stored in separate pool buckets and never mixed.

#### 1. Added `autoGenerateMipmaps` parameter to `getOptimalTexture()` ([TexturePool.ts:87-93](src/rendering/renderers/shared/texture/TexturePool.ts#L87-L93))

```typescript
public getOptimalTexture(
    frameWidth: number,
    frameHeight: number,
    resolution = 1,
    antialias: boolean,
    autoGenerateMipmaps = false  // NEW: defaults to false
): Texture
```

**Defaults to `false`** to maintain backward compatibility and ensure render textures (filters, masks, render groups) don't get mipmaps by default.

#### 2. Updated `createTexture()` to accept mipmap parameter ([TexturePool.ts:59](src/rendering/renderers/shared/texture/TexturePool.ts#L59))

The pool now creates textures with the specified mipmap setting instead of inheriting from `this.textureOptions`.

#### 3. Added mipmap bit to pool key ([TexturePool.ts:101-102](src/rendering/renderers/shared/texture/TexturePool.ts#L101-L102))

```typescript
const mipmapBit = autoGenerateMipmaps ? 1 : 0;
const key = (po2Width << 17) + (po2Height << 1) + (antialias ? 1 : 0) + (mipmapBit << 2);
```

This ensures textures with different mipmap settings are stored in separate pool buckets within the same `TexturePool` instance.

#### 4. Updated text rendering to request mipmaps ([getPo2TextureFromSource.ts:40](src/scene/text/utils/getPo2TextureFromSource.ts#L40))

Text rendering now explicitly passes the global mipmap setting when requesting textures:

```typescript
const texture = TexturePool.getOptimalTexture(
    bounds.width,
    bounds.height,
    resolution,
    false,
    TextureSource.defaultOptions.autoGenerateMipmaps  // Respect global setting
);
```

### How It Works

- **Filters, masks, render groups**: Call `getOptimalTexture(w, h, res, aa)` → defaults to `autoGenerateMipmaps: false` → stored in pool key bucket with `mipmapBit=0`
- **Text rendering**: Calls `getOptimalTexture(w, h, res, aa, TextureSource.defaultOptions.autoGenerateMipmaps)` → stored in pool key bucket with `mipmapBit=1` when enabled
- Textures with different mipmap settings never mix because they have different pool keys

### Result

Users can now:
- Set `PIXI.TextureSource.defaultOptions.autoGenerateMipmaps = true` for high-quality text rendering with mipmaps
- Use filters without visual corruption
- Text and filter textures are separated at the pool key level within a single pool instance

### Advantages of This Approach

- **Single pool instance**: Simpler architecture, no module-level state
- **Backward compatible**: Default behavior unchanged (no mipmaps for render textures)
- **Flexible**: Any caller can specify mipmap preference per request
- **Clean API**: Follows existing pooling pattern

### Breaking Changes

None. The new parameter is optional and defaults to `false`, maintaining backward compatibility.

### Tests

Comprehensive test suite added in `src/rendering/renderers/shared/texture/__tests__/TexturePool.test.ts` with **21 tests** covering:

#### Pool Key Generation
- ✅ Unique keys for different texture configurations
- ✅ Pool reuse for identical configurations
- ✅ Mipmap settings don't mix

#### Bit Shift Operations
- ✅ Width encoding at bit position 17
- ✅ Height encoding at bit position 1
- ✅ Antialias flag encoding at bit position 0
- ✅ Mipmap flag encoding at bit position 2
- ✅ All flag combinations work independently

#### Mipmap Separation (Bug Fix Verification)
- ✅ **Critical**: Prevents mipmapped text textures from being reused for filters
- ✅ Both mipmapped and non-mipmapped textures coexist in same pool
- ✅ Separation maintained across multiple pool operations

#### Additional Coverage
- ✅ Default parameter behavior
- ✅ Backward compatibility with existing code
- ✅ Power-of-2 rounding with mipmap separation

##### Pre-Merge Checklist

Fixes #11717


GoodBoyDigital:
Hey, this is a really cool fix! The TexturePool changes are solid - the mipmap bit separation in the pool key is the right approach, and the test coverage is thorough.

One architectural suggestion: rather than having `getPo2TextureFromSource` read from the global `TextureSource.defaultOptions.autoGenerateMipmaps`, what if we made `autoGenerateMipmaps` an explicit option on `CanvasTextOptions`?

Currently the flow has a hidden dependency on global state:
```typescript
// getPo2TextureFromSource.ts:41
TextureSource.defaultOptions.autoGenerateMipmaps  // action at a distance
```

A cleaner approach would be to add it as an explicit option:

```typescript
// Text.ts - add to CanvasTextOptions interface
export interface CanvasTextOptions extends TextOptions
{
    textureStyle?: TextureStyle | TextureStyleOptions;
    /** Whether to generate mipmaps for the text texture. Improves quality when scaled down. */
    autoGenerateMipmaps?: boolean;
}

// getPo2TextureFromSource.ts - accept as parameter
export function getPo2TextureFromSource(
    image: ImageLike | HTMLCanvasElement | ICanvas,
    width: number,
    height: number,
    resolution: number,
    autoGenerateMipmaps = false  // explicit param instead of global read
): Texture

// CanvasTextSystem.ts - pass it through from options
const texture = getPo2TextureFromSource(
    canvasAndContext.canvas, 
    frame.width, 
    frame.height, 
    resolution,
    options.autoGenerateMipmaps
);
```

Benefits:
- **Per-text control**: `new Text({ text: 'Hi', autoGenerateMipmaps: true })` - users can enable mipmaps for specific text instances
- **No global state dependency**: explicit > implicit
- **Consistent with existing patterns**: follows how `textureStyle` is already handled
- **Backward compatible**: defaults to `false`

The TexturePool changes in this PR are great and would remain the same - this is just about how the option flows into text rendering.

What do you think?

vkarponen:
Ok. I have a commit coming up that adds per text instance mipmapping. But in order to stay backwards compatible TextureSource.defaultOptions.autoGenerateMipmaps should afaics still control the Text mipmap generation unless an instance overrides it.

Correct?

vkarponen:
Here is the per text instance mipmapping. The code-rabbit bot is complaining about adding more of using the default, but me and @qtiki feel that would be excessive. What do you think @GoodBoyDigital ?

vkarponen:
Bump @GoodBoyDigital

GoodBoyDigital (approved):

GoodBoyDigital:
sorry for the late response - looks good :D

PR #11866, opened 2026-02-03
chore: auto-find available port for test scene server

GoodBoyDigital:
##### Description of change

When running `npm run test:scene:debug`, the test server now automatically finds an available port instead of hardcoding 8080, eliminating port conflicts. The server startup is moved from Jest global setup into the visual test suite itself.

- Refactored server management into `tests/visual/utils/serverUtils.ts`
- Uses OS-assigned free ports (port 0) for automatic availability detection
- Removed now-unnecessary global setup/teardown files
- Server only starts locally; CI continues using GitHub raw URLs

##### Pre-Merge Checklist

- [x] Tests and/or benchmarks are included
- [ ] Documentation is changed or added
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test:scene`)

---
##### Behavior Changes
- Test scene server now automatically discovers and uses an available port instead of hardcoded port 8080. Port detection uses OS-assigned free ports (port 0). Server starts only for local test runs; CI continues using GitHub raw URLs.

##### Chores
- Removed Jest global setup/teardown lifecycle from `jest.config.js` and deleted corresponding setup/teardown scripts (`scripts/jest/jest-global-setup.ts`, `scripts/jest/jest-global-teardown.ts`).
- Moved HTTP server management to a new utility module `tests/visual/utils/serverUtils.ts` with exported functions: `startServer()`, `stopServer()`, and `getServerPort()`.
- Updated visual tests (`tests/visual/visuals.test.ts`) to manage server lifecycle within test suite using `beforeAll`/`afterAll` hooks instead of global Jest setup.

PR #11874, opened 2026-02-04
chore: Improve test server setup by dynamically allocating port

Zyie:
a minimal change version of #11866

---
##### Chores
- Improved test server setup to dynamically allocate an available port for the HTTP server, reducing port conflicts across developer environments and CI runs
  - Added `get-port` dependency (^7.1.0) for port availability detection
  - Updated Jest global setup to obtain and use a free port via the `SERVER_PORT` environment variable
  - Updated asset URL basePath in visual tests to reference the dynamically allocated port instead of hardcoded 8080


bigtimebuddy (approved):
Nice!

PR #11875, opened 2026-02-04
fix(BitmapText): improve word wrap to match CSS break-word behavior

Riphal:
## Summary

- Add CSS-like break-after support for hyphens and dashes in BitmapText word wrapping
- Fix `breakWords: false` to break long words as a last resort (matching CSS `overflow-wrap: break-word`)
- Fix `breakWords: true` to break at any character (matching CSS `word-break: break-all`)

## Problem

Previously, BitmapText word wrapping had inconsistent behavior compared to CSS:
- `breakWords: true` was breaking ALL words at line boundaries, even short words
- `breakWords: false` would let long words overflow instead of breaking them

## Solution

- **`breakWords: false`** = CSS `overflow-wrap: break-word` (prefer natural breaks at hyphens, break long words as last resort)
- **`breakWords: true`** = CSS `word-break: break-all` (break at any character)

### Break-after characters
- `-` (hyphen), `–` (en-dash), `—` (em-dash), soft hyphen

---
##### Behavior Changes
- `BitmapText` word wrapping now follows CSS-like semantics: `breakWords: false` matches `overflow-wrap: break-word` — prefer natural breaks (e.g., after hyphens/dashes/soft-hyphen) and only break long words as a last resort.
- `breakWords: true` now matches `word-break: break-all` — allow breaking at any character to avoid overflow.
- Added break-after character handling for hyphen (`-`), unicode hyphen (`\u2010`), en-dash (`\u2013`), em-dash (`\u2014`), and soft hyphen (`\u00AD`) so these characters stay with the preceding word when wrapping.

##### Fixes
- Fixed incorrect behavior where `breakWords: true` would break words at every line boundary instead of only when needed.
- Fixed incorrect behavior where `breakWords: false` allowed very long words to overflow instead of breaking them as a fallback.

##### Chores
- Added/updated visual tests and snapshots for bitmap text wrapping (break-all / break-word / hyphen cases) and adjusted an HTML text stroke scene's pixelMatch tolerance.


Zyie:
closing in favor of #11947

PR #11886, opened 2026-02-05
fix: Graphics getLocalBounds stale between ops

### LINKED ISSUE #11870: Graphics `getLocalBounds` returns stale data when called between graphics operations in single frame
### PixiJS version

v8.16.0

https://stackblitz.com/edit/pixijs-v8-seuxtzxp?file=src%2Fmain.js,package.json

### Steps to reproduce

* Trigger new frame
* Call graphics method which updates bounds
* Call getLocalBounds
* Call graphics method which updates bounds
* Call getLocalBounds again

### What is expected?

The latter call to getLocalBounds is updated to reflect the bounds changes in the second graphics method call(s)

The latter call to getLocalBounds uses the cached bounds from the first call

### System Info

```Shell
System:
    OS: Linux 6.18 Manjaro Linux
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 3.24 GB / 15.26 GB
    Container: Yes
    Shell: 5.3.9 - /bin/bash
  Browsers:
    Chromium: 144.0.7559.109
    Firefox: 147.0.2
    Firefox Developer Edition: 147.0.2
```

### Any additional comments?

This appears to be due to this optimization here:
https://github.com/pixijs/pixijs/blob/94156d33426f74df3b8f9ea83eb7f18252d5436b/src/scene/graphics/shared/GraphicsContext.ts#L1085-L1089

The `update` emitted is what triggers the bounds cache invalidation via (onViewUpdate). This is trivially solvable by not having that early return, but I don't know what else that may effect

GoodBoyDigital:
### Overview

Fixes #11870

#### Fixes
- Fixed `Graphics.getLocalBounds()` returning stale cached bounds when called between graphics operations (e.g. between two `fill()` calls) in the same frame. `GraphicsContext.onUpdate()` had an early return that suppressed the `update` event after the first call per frame, preventing bounds cache invalidation on subsequent operations.
- Removed the early-return guard in `onUpdate()` so the event always fires. `ViewContainer.onViewUpdate()` already prevents duplicate rendering work, so this is safe. `onUpdate()` only fires on finalizing operations (`fill()`, `stroke()`, `texture()`, `clear()`), not path-building ops.

##### Pre-Merge Checklist

##### Fixes
- Fixed Graphics.getLocalBounds returning stale cached bounds when called between graphics operations in the same frame. The `update` event now consistently fires after finalizing operations (fill, stroke, texture, clear), ensuring the bounds cache is properly invalidated.


bigtimebuddy (approved):

luxaritas:
Thanks for addressing this!

PR #11888, opened 2026-02-06
fix: WebGPU UBO batch overflow crash by dynamically growing the uniform buffer

Riphal:
## Summary

- Replace the hard crash (`"UniformBufferBatch: ubo batch got too big"`) with dynamic buffer growth — the backing `Float32Array` doubles in size when full, matching the existing `// TODO push a new buffer` intent.
- Synchronize GPU buffer sizes immediately on resize via a callback, preventing WebGPU bind group validation errors (`"Binding offset is larger than the size of Buffer"`).
- Null out `_gpuData` before triggering the buffer `change` event so the old `GPUBuffer` is not destroyed mid-frame — avoids `"Buffer used in submit while destroyed"` errors from bind groups already recorded in the current render pass.

## Test plan

- [x] Render a scene with enough draw calls to exceed the original 65535-float (~256 KB) uniform buffer capacity — should no longer crash
- [x] Verify no WebGPU validation warnings/errors in the browser console
- [x] Confirm visual correctness (no missing/corrupt draws) on scenes both below and above the old buffer limit
- [x] Check that GPU memory usage remains stable across frames (buffer grows once, does not re-allocate every frame)

---

##### Fixes
- Fixed WebGPU UBO batch overflow crash by dynamically growing the uniform buffer when capacity is exceeded, replacing the hard error with automatic doubling of capacity
- Fixed GPU buffer synchronization issues that caused WebGPU validation errors by immediately syncing GPU buffer sizes when the backing buffer resizes
- Fixed potential destruction of GPU buffers mid-frame by nullifying GPU data references before triggering buffer change events

##### Features
- Added optional `onResize` callback to `UboBatch` to notify consumers when the backing buffer resizes:
  ```ts
  batch.onResize = (newData: Float32Array) => {
    // Update GPU resources with new buffer reference
  };
  ```

### DIFF
--- a/src/rendering/renderers/gpu/GpuUniformBatchPipe.ts
+++ b/src/rendering/renderers/gpu/GpuUniformBatchPipe.ts
@@ -51,6 +51,31 @@ export class GpuUniformBatchPipe
                 usage
             }));
         }
+
+        // When the batch buffer grows, immediately update all GPU buffer data references
+        // so that their descriptor.size reflects the new capacity before any BufferResources
+        // are created with offsets into the grown region.
+        this._batchBuffer.onResize = (data: Float32Array) =>
+        {
+            const rendererUid = this._renderer.uid;
+
+            for (let i = 0; i < this._buffers.length; i++)
+            {
+                const buffer = this._buffers[i];
+
+                // Null out the GPU data reference BEFORE setting buffer.data.
+                // Setting buffer.data with a larger array emits a 'change' event, which triggers
+                // GpuBufferSystem.onBufferChange → GCManagedHash.remove. Normally remove() would
+                // destroy the old GPUBuffer. However, the old GPUBuffer may still be referenced by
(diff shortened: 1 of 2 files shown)


GoodBoyDigital:
Thanks for tackling this — the crash definitely needs fixing. However, I think resizing the buffer is the wrong approach here. WebGPU has a hard limit on uniform buffer size (`maxUniformBufferBindingSize` is 65536 bytes by default), so doubling will eventually hit that ceiling and trade one crash for a validation error.

The right fix is what the original TODO suggested: push a new buffer. Instead of growing a single `UboBatch`, we should maintain an array of them. When the current batch is full, allocate the next one. This avoids the resize entirely, which also means no `onResize` callback, no `_gpuData` null trick, and no orphaned GPUBuffers. Each batch stays at a fixed size within device limits, and the pool grows as needed.

Zyie:
closing this down for now as its not the right approach

PR #11889, opened 2026-02-06
fix: Fixed Ticker minFPS setting when maxFPS is 0

Romans-I-XVI:
##### Description of change
I noticed it was impossible to change the Ticker minFPS unless maxFPS is set to a non-zero value first. Since maxFPS is zero by default this causes minFPS to not be set in most use cases .

Example
https://stackblitz.com/edit/pixijs-v8-4aerx4o5?file=src%2Fmain.js

##### Pre-Merge Checklist

---
##### Fixes
- Fixed an issue where `Ticker.minFPS` could not be set when `Ticker.maxFPS` was 0 (the default). The minFPS setter now correctly applies the value directly when maxFPS is unset.

### DIFF
--- a/src/ticker/Ticker.ts
+++ b/src/ticker/Ticker.ts
@@ -770,7 +770,7 @@ export class Ticker
     set minFPS(fps: number)
     {
         // Minimum must be below the maxFPS
-        const minFPS = Math.min(this.maxFPS, fps);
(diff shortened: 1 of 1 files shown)


Zyie:
Thanks for this change, however, we are going to close this one in favor of #11952

PR #11895, opened 2026-02-08
fix: MSDF BitmapText alpha applied twice (squared)

Riphal:
## Summary

Fixes an issue where setting `alpha` on a `BitmapText` using MSDF fonts resulted in the
alpha being much stronger than expected (e.g. setting `alpha = 0.5` produced ~25% opacity
instead of 50%), making it inconsistent with other display objects.

## Root Cause

The `calculateMSDFAlpha` function in the MSDF shader was multiplying the coverage by
`shapeColor.a` (the user's alpha from `vColor`):

```glsl
float coverage = pow(shapeColor.a * alpha, gamma);
```

However, the high-shader fragment template in `defaultProgramTemplate.ts` **always** applies
`vColor` to the output afterwards:

```glsl
finalColor = outColor * vColor;
```

This caused the alpha to be applied **twice**, effectively squaring it (`alpha * alpha`).

## Fix

Removed `shapeColor.a` from the `pow()` call in `calculateMSDFAlpha` (both WGSL and GLSL
variants), so the function returns only the SDF edge coverage. The user's alpha is then
correctly applied once by the shader template.

```glsl
// Before (alpha squared)
float coverage = pow(shapeColor.a * alpha, gamma);

// After (alpha applied once by template)
float coverage = pow(alpha, gamma);
```

---
##### Fixes
- Fixed alpha being applied twice (squared) when setting `alpha` on `BitmapText` with MSDF fonts. The `calculateMSDFAlpha` function was multiplying by `shapeColor.a` inside the gamma correction, but the shader template already applies user alpha via `finalColor = outColor * vColor`. Removed `shapeColor.a` from the `pow()` call in both WGSL and GLSL shader variants to apply user alpha only once.

### DIFF
--- a/src/scene/text/sdfShader/shader-bits/mSDFBit.ts
+++ b/src/scene/text/sdfShader/shader-bits/mSDFBit.ts
@@ -22,9 +22,13 @@ export const mSDFBit = {
                 }
 
                 // Gamma correction for coverage-like alpha
+                // Note: do NOT multiply by shapeColor.a here. The shader
+                // template already applies vColor (which carries the user's
+                // alpha) via: finalColor = outColor * vColor.
+                // Including shapeColor.a here would square the alpha.
                 var luma: f32 = dot(shapeColor.rgb, vec3<f32>(0.299, 0.587, 0.114));
                 var gamma: f32 = mix(1.0, 1.0 / 2.2, luma);
-                var coverage: f32 = pow(shapeColor.a * alpha, gamma);
+                var coverage: f32 = pow(alpha, gamma);
 
                 return coverage;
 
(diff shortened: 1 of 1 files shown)


bigtimebuddy (approved):

Riphal:
Hey @bigtimebuddy , I've opened a new PR that supersedes this one: https://github.com/pixijs/pixijs/pull/11896

The old PR fixed the alpha-squared issue by changing `pow(shapeColor.a * alpha, gamma)` to `pow(alpha, gamma)`, which was correct in removing the double alpha application. However, it still kept the gamma correction (`pow(alpha, gamma)`), which has two problems:

1. **Text still appears thicker than expected when zoomed out.** The gamma correction boosts intermediate alpha values at glyph edges (e.g. `pow(0.3, 0.45) ≈ 0.51`). At small scales, more pixels fall in the SDF transition zone, and the gamma inflates all of the, making the text visually bolder.

2. **The luma calculation uses premultiplied `shapeColor.rgb`.** Since `vColor` is premultiplied in the vertex shader (`rgb = color * alpha`), the luma value changes depending on the object's opacity, which unintentionally alters the gamma correction strength based on alpha.

The new PR removes the gamma correction entirely and instead adds `max(distance, 1.0)` to clamp the screen pixel range, matching the [msdf-atlas-gen reference implementation](https://github.com/Chlumsky/msdf-atlas-gen). This produces cleaner, more consistent edges at all zoom levels without artificially inflating edge opacity.

Zyie:
closing this one down as #11896 supersedes

PR #11896, opened 2026-02-09
fix: MSDF/SDF BitmapText alpha and edge rendering

Riphal:
## Summary

Fixes two issues with MSDF/SDF BitmapText rendering:

1. **Alpha was squared** — `calculateMSDFAlpha` multiplied the coverage by `shapeColor.a`, but the high-shader fragment template already applies `vColor` afterwards (`finalColor = outColor * vColor`). This caused `alpha` to be applied twice, so setting `alpha = 0.5` on a BitmapText produced ~25% opacity instead of 50%.

2. **Text appeared thicker at small scales** — The gamma correction (`pow(shapeColor.a * alpha, gamma)`) boosted intermediate alpha values at glyph edges, making text visually bolder — especially when zoomed out where more pixels fall in the SDF transition zone. Removed the gamma correction and added `max(distance, 1.0)` to clamp the screen pixel range, matching the [msdf-atlas-gen reference implementation](https://github.com/Chlumsky/msdf-atlas-gen) which prevents overly soft edges at small scales.

## Changes

- Removed `shapeColor.a` from coverage calculation (alpha is applied once by the template)
- Removed gamma correction (`pow(alpha, gamma)`) that was inflating edge opacity
- Added `max(distance, 1.0)` clamp to prevent soft/thick edges when zoomed out
- Updated visual test snapshots for MSDF and SDF text

---
##### Fixes
- Fixed alpha being applied twice in MSDF/SDF BitmapText rendering. Alpha was previously multiplied in both the coverage calculation and the shader template, resulting in squared alpha values. Alpha is now applied only once by the shader template.
- Fixed MSDF/SDF text appearing thicker at small scales. Removed gamma correction step and added distance clamping to match the msdf-atlas-gen reference implementation, preventing overly soft/thick edges at small scales.


Zyie:
Hey @Riphal 

I've tried creating visual tests for this but none are revealing any issues
If ou could provide a demo of the issue, that would be helpful

bigtimebuddy (approved):
Tests look good!

bigtimebuddy:
@Zyie look at the snapshots the contour of the text looks much better

PR #11901, opened 2026-02-09
fix: remove listener from old resource in BindGroup.setResource

GoodBoyDigital:
### Overview
Fixes a listener leak in `BindGroup.setResource()` where the change listener was being removed from the new resource instead of the old resource being replaced.

#### Fixes
- When replacing a resource in a bind group, the change listener is now correctly removed from `currentResource` (the old resource) instead of `resource` (the new resource).

##### Pre-Merge Checklist

---
##### Fixes
- Fixed a listener leak in `BindGroup.setResource()` where the change listener was being removed from the new resource instead of the old one.

### DIFF
--- a/src/rendering/renderers/gpu/shader/BindGroup.ts
+++ b/src/rendering/renderers/gpu/shader/BindGroup.ts
@@ -95,11 +95,7 @@ export class BindGroup
 
         if (resource === currentResource) return;
 
-        if (currentResource)
-        {
-            currentResource.off?.('change', this.onResourceChange, this);
-        }
-
+        currentResource?.off?.('change', this.onResourceChange, this);
         resource.on?.('change', this.onResourceChange, this);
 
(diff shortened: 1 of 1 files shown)


Zyie (approved):

src/rendering/renderers/gpu/shader/BindGroup.ts
```diff
@@ -97,7 +97,7 @@ export class BindGroup
 
         if (currentResource)
         {
-            resource.off?.('change', this.onResourceChange, this);
+            currentResource.off?.('change', this.onResourceChange, this);
```
bigtimebuddy: Nit. Optional chaining would get rid of the if block.
```suggestion
            currentResource?.off?.('change', this.onResourceChange, this);
```

bigtimebuddy (approved):

PR #11909, opened 2026-02-12
fix: blur by reducing strength on each pass

### LINKED ISSUE #10654: Bug: Can't alter uniforms between passes in BlurFilterPass with WebGPU renderer
### Current Behavior

Companion to #10653 
The WebGPU renderer is not using the correct `uStrength` uniform for each pass when changing the value between blur passes. It looks like it's using the strength of the last pass for all passes.

Screenshot 2024-06-21 171541
*Left to right: WebGPU, WebGL 1, WebGL 2*

### Expected Behavior

The WebGPU renderer uses the correct `uStrength` uniform for each pass, matching the WebGL renderer.

### Steps to Reproduce

Run `blur-filter.scene.ts` in #10653 

### Environment

- **`pixi.js` version**: #10653
- **Browser & Version**: pixi visual test runner (electron 12.2.3)
- **Running Example**: `blur-filter.scene.ts` in #10653 

### Possible Solution

_No response_

### Additional Information

_No response_

Zyie:
### Overview

Optimizes `BlurFilterPass` by progressively halving blur strength on each pass instead of using a constant `strength / passes` value. Uses the Gaussian variance addition property (`σ² = Σσᵢ²`) to compute the correct initial strength so the combined effect matches the target. Also fixes a WebGPU bug where all blur passes used the last pass's uniform values due to draw batching.

Based on the work by @minetoblend in #10653.

Closes #10653
Closes #10654

#### Behavior Changes
- `BlurFilter` now uses an optimized halving strength scheme by default, which changes the visual output compared to previous versions. Set `legacy: true` to restore the old behavior:
  ```ts
  new BlurFilter({ legacy: true });
  // or globally:
  BlurFilter.defaultOptions.legacy = true;
  ```

##### Pre-Merge Checklist


bigtimebuddy (approved):
Thank you!

GoodBoyDigital (approved):

Issue #11461, opened 2025-06-05
Bug: PIXI 8 Chrome webm Breaks Alpha Channel


suwalski:
### Current Behavior

We've been using PIXI with webm with an alpha channel for several PIXI major versions without issue until PIXI 8.

PIXI running in Chrome seems to turn every pixel of a webm's alpha channel that is not completely transparent into a 1-bit alpha mask.

This is how it looks in Chrome:

![Image](https://github.com/user-attachments/assets/146d6dde-1b79-43da-a8d5-b3b162f5c2c7)

The video renders correctly in PIXI 7. It also renders correctly in Chrome when embedded into plain HTML.

### Expected Behavior

This is how it looks in Firefox:

![Image](https://github.com/user-attachments/assets/abcbbd94-bd58-44b5-9ce5-8e9a7ffab913)

### Steps to Reproduce

Here is a playground URL: [https://www.pixiplayground.com/#/edit/Wb9IdW4G1Bk3xw123blMe](url)

### Environment

- **`pixi.js` version**: *e.g. 8.10.0*
- **Browser & Version**: *e.g. Chrome 137*
- **OS & Version**: *e.g. Multiple; Linux and Windows*
- **Running Example**: *e.g. <https://www.pixiplayground.com/#/edit/Wb9IdW4G1Bk3xw123blMe/>*

### Additional Information

Toggling webgpu doesn't change anything. Forcing the `alphaMode` doesn't seem to help either.

reececomo:
It's rendering correctly on MacOS Chrome

- Chrome 135.0.7049.115 (Official Build) (arm64)
- MacOS Sonama 14.5

<img width="437" alt="Image" src="https://github.com/user-attachments/assets/0dde960e-e99a-4d64-9cc4-449272a8e565" />

```ts
(async () => {
    const app = new PIXI.Application();
    await app.init({
        width: window.innerWidth,
        height: window.innerHeight,
        backgroundColor: 0x2c3e50
    })
    document.body.appendChild(app.canvas);

    // Load the bunny texture
    const texture = await PIXI.Assets.load('https://rideaunautical.ca/temp/fw.webm');

    texture.source.resource.loop = true;

    const videoSprite = new PIXI.Sprite(texture);
    app.stage.addChild(videoSprite);
})();
```

suwalski:
[...]

PR #11911, opened 2026-02-15
fix: respect explicit alphaMode in VideoSource, skip detectVideoAlphaMode()

### LINKED ISSUE #11473: Bug: alphaMode passed into video loader ignored
### Current Behavior

This issue is related to #11461. As per that issue, due to an unreliable `detectVideoAlphaMode()` function, we pass in an `alphaMode` option, which should prevent the alpha mode detection routine from running at all, and apply the desired mode.

as per [src/assets/loader/parsers/textures/loadVideoTextures.ts](https://github.com/pixijs/pixijs/blob/7d91234eb506457fb6e612b02b31cdf6dae7062f/src/assets/loader/parsers/textures/loadVideoTextures.ts#L140):

```ts
    async load(url: string, asset: ResolvedAsset<VideoSourceOptions>, loader: Loader): Promise<Texture>
    {
        // --- Merge default and provided options ---
        const options: VideoSourceOptions = {
            ...VideoSource.defaultOptions,
            resolution: asset.data?.resolution || getResolutionOfUrl(url),
            alphaMode: asset.data?.alphaMode || await detectVideoAlphaMode(),
            ...asset.data,
        };
```

If the `alphaMode` is provided, it uses what was provided rather than detecting it automagically.

However, later on it calls out to the `VideoSource` renderer:

```ts
      const onCanPlay = async () => {
        const base = new VideoSource({ ...options, resource: videoElement });
```

And in this class, the handling is different. It calls out to `detectVideoAlphaMode()` unconditionally. From [src/rendering/renderers/shared/texture/sources/VideoSource.ts](https://github.com/pixijs/pixijs/blob/7d91234eb506457fb6e612b02b31cdf6dae7062f/src/rendering/renderers/shared/texture/sources/VideoSource.ts#L264):

```ts
    public async load(): Promise<this>
...
        this.alphaMode = await detectVideoAlphaMode();
```

This clearly does not utilize the `alphaMode` as provided, overriding it. So, on one hand, the provided `alphaMode` is used, but then it doesn't make its way to the video as shown.

Adding debug prints to ```detectVideoAlphaMode()``` this is shown to be true. When a video is played, this utility function is normally called twice. If `alphaMode` is passed in, it's called once. But that one time makes it ignore what was passed in.

### Expected Behavior

Seeing as `detectVideoAlphaMode()` is pretty heavy, and at the very least quite hacky, it would be great to limit how often it's executed. One time should be enough, and an override should prevent it from ever running.

### Steps to Reproduce

Load a webm video:

```ts
const texture = await PIXI.Assets.load('https://foo.com/bar.webm');
```

### Environment

- **`pixi.js` version**: *e.g. 8.x.x*
- **Browser & Version**: all
- **OS & Version**: all

### Possible Solution

_No response_

### Additional Information

As per https://github.com/pixijs/pixijs/issues/11461#issuecomment-2952715340 it is possible to override this behaviour post-hoc with:

```
texture.source.alphaMode = "no-premultiply-alpha";
```

However, the `detectVideoAlphaMode()` function has already run at least once, and probably twice.

fayanwubian-hue:
## Summary

When `alphaMode` is explicitly provided to `VideoSource` (via `Assets.load` data or constructor options), the detection routine `detectVideoAlphaMode()` should be skipped entirely.

Fixes #11473

## Problem

`VideoSource.load()` unconditionally overwrites `alphaMode` with the detected value at line 264:

```ts
this.alphaMode = await detectVideoAlphaMode();
```

This ignores any user-provided `alphaMode` option, making it impossible to override the unreliable auto-detection (as noted in #11461).

## Solution

1. Track whether `alphaMode` was explicitly provided via a private `_alphaModeExplicit` flag (checked **before** the options merge with defaults)
2. In `load()`, only run `detectVideoAlphaMode()` when no explicit value was given

This is a minimal, backward-compatible fix:
- No explicit `alphaMode` → behavior unchanged (auto-detection runs)
- Explicit `alphaMode` → detection skipped, user value respected

## Testing

```ts
// Before: alphaMode override ignored, detectVideoAlphaMode() always runs
// After: alphaMode override respected, detectVideoAlphaMode() skipped
const texture = await PIXI.Assets.load({
  src: 'video.webm',
  data: { alphaMode: 'no-premultiply-alpha' }
});
// texture.source.alphaMode === 'no-premultiply-alpha' ✓
```

<details>

##### Fixes
- Respect explicit `alphaMode` option in `VideoSource` by skipping `detectVideoAlphaMode()` when an alpha mode is explicitly provided. Previously, any user-supplied `alphaMode` was unconditionally overwritten by auto-detection.
  ```ts
  const source = new VideoSource({
    resource: videoElement,
    alphaMode: 'no-premultiply-alpha'
  });
  // alphaMode is now preserved and detectVideoAlphaMode() is skipped
  ```

</details>

### DIFF
--- a/src/rendering/renderers/shared/texture/sources/VideoSource.ts
+++ b/src/rendering/renderers/shared/texture/sources/VideoSource.ts
@@ -116,6 +116,7 @@ export class VideoSource extends TextureSource<VideoResource>
 
     private _msToNextUpdate: number;
     private _preloadTimeout: number;
+    private _alphaModeExplicit: boolean;
 
     /** Callback when completed with load. */
     private _resolve: (value?: this | PromiseLike<this>) => void;
@@ -130,6 +131,8 @@ export class VideoSource extends TextureSource<VideoResource>
     {
         super(options);
 
+        const alphaModeExplicit = options.alphaMode !== undefined;
+
         // Merge provided options with default ones
         options = {
             ...VideoSource.defaultOptions,
@@ -141,6 +144,7 @@ export class VideoSource extends TextureSource<VideoResource>
         this._updateFPS = options.updateFPS || 0;
         this._msToNextUpdate = 0;
         this.autoPlay = options.autoPlay !== false;
+        this._alphaModeExplicit = alphaModeExplicit;
         this.alphaMode = options.alphaMode ?? 'premultiply-alpha-on-upload';
 
         // Binding for frame updates
@@ -261,7 +265,10 @@ export class VideoSource extends TextureSource<VideoResource>
             this._mediaReady();
         }
 
-        this.alphaMode = await detectVideoAlphaMode();
+        if (!this._alphaModeExplicit)
+        {
+            this.alphaMode = await detectVideoAlphaMode();
+        }
 
         // Create and return the loading promise
         this._load = new Promise((resolve, reject): void =>


Zyie:
Closing this pr down for now as no tests have been provided. Feel free to reopen with additional tests added

PR #11912, opened 2026-02-16
fix: emit removed event when using addChildAt

aSipz:
##### Description of change

In addChildAt, when the child already had a parent (including when re-parenting or moving within the same container), the code removed it from the old parent by mutating the array directly:

`child.parent.children.splice(currentIndex, 1);`

That only changed the children array. It did not:
Emit removal events (removed / childRemoved)
Run the normal removal path (e.g. render group updates, clearing parent refs via the public API)
So listeners and internal state could get out of sync when using addChildAt to move or re-parent a child.

What was changed

The direct splice was replaced with the public removal API:

`child.removeFromParent();`

So when addChildAt has to remove the child from its current parent, it now goes through the same removal logic as removeChild/removeFromParent: events are emitted and render groups and parent refs are updated correctly.

##### Pre-Merge Checklist

<details>

##### Fixes
- Fixed `addChildAt` to properly emit `removed` and `childRemoved` events when re-parenting a child to a different parent. Previously, these events were bypassed due to direct array manipulation.
- `addChildAt` no longer emits removal events when re-indexing a child within the same parent, keeping re-indexing semantically distinct from parent changes.

</details>

### DIFF
--- a/src/scene/container/__tests__/Container.Events.test.ts
+++ b/src/scene/container/__tests__/Container.Events.test.ts
@@ -76,6 +76,64 @@ describe('Container Events', () =>
 
             expect(listener).toHaveBeenCalledTimes(3);
         });
+
+        it('should trigger removed listeners when addChildAt is used', () =>
+        {
+            const removedListener = jest.fn();
+            const child = new Container();
+            const parentA = new Container();
+            const parentB = new Container();
+            const sibling = new Container();
+
+            parentB.addChild(sibling);
+
+            child.on('removed', removedListener);
+
+            parentA.addChild(child);
+
+            parentB.addChildAt(child, 1);
+
+            expect(removedListener).toHaveBeenCalledTimes(1);
+
+            parentB.addChildAt(child, 0);
+
+            expect(removedListener).toHaveBeenCalledTimes(2);
+
+            parentB.setChildIndex(child, 1);
+
+            expect(removedListener).toHaveBeenCalledTimes(3);
+
+            parentA.addChildAt(child, 0);
+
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Could you add tests for this change?

aSipz:
Added tests whether "removed" and "childRemoved" events are emitted correctly.

GoodBoyDigital:
Great catch on the missing `removed`/`childRemoved` events! One suggestion — we could take this a step further by only calling `removeFromParent()` when the child is actually changing parents. When it's the same parent (just re-indexing), a plain splice is sufficient and avoids unnecessary work:

```js
if (child.parent === this)
{
    // Same parent, just re-index - no need to fire removal events
    children.splice(currentIndex, 1);
}
else
{
    // Different parent, full removal with events and cleanup
    child.removeFromParent();
}
```

The reasoning: re-indexing within the same parent isn't semantically a "removal" — the child never leaves. Firing `removed`/`childRemoved` in that case would be misleading to listeners, and it also avoids the overhead of `renderGroup.removeChild()` + `renderGroup.addChild()` round-tripping and the extra `indexOf` call (since we already have `currentIndex`).

This also keeps `setChildIndex` (which delegates to `addChildAt`) from incorrectly emitting removal events.

aSipz:
addChildAt() was updated so that moving a child inside the same container no longer triggers remove/add events or uses removeFromParent(). Only re-parenting (moving to another container) does that. Additional tests were introduced.

GoodBoyDigital (approved):

bigtimebuddy (approved):

bigtimebuddy:
Thank you @aSipz

PR #11921, opened 2026-02-19
feat: Add function for removing aliases from resolver

### LINKED ISSUE #9115: Feature Request: Add a way to let the user remove asset aliases from PIXI.Assets.resolver
### Description

`PIXI.Assets.resolver` currently has no way to clear out any an unwanted aliases. The only other option is to use `reset()` which will clear out every other alias in use.

I'm making this a feature request as it would be useful if there was a way to swap out assets while re-using the same alias, such as for different language files or for getting a newer updated version of a JSON.

Sertion:
##### Description of change
Adding a function for removing an alias from the resolver. Resolves #9115.

##### Pre-Merge Checklist

### DIFF
--- a/src/assets/resolver/Resolver.ts
+++ b/src/assets/resolver/Resolver.ts
@@ -318,6 +318,29 @@ export class Resolver
     }
 
     /**
+     * Removes the specified alias for an asset
+     *
+     * This does not remove or destroy the asset!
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Seems reasonable 👍

Issue #8359, opened 2022-05-24
Incorrect behavior in ColorMatrixFilter.contrast() with multiply=true


rubenlg:
### Expected Behavior
`ColorMatrixFilter.contrast(1, true)` produces the same matrix as `ColorMatrixFilter.contrast(1, false)` when the current matrix is the identity.

### Current Behavior
`ColorMatrixFilter.contrast(1, true)` produces a broken matrix that mostly updates the brightness, even making dark colors brighter.

### Possible Solution
The core of the problem is in `ColorMatrixFilter._loadMatrix` which calls `ColorMatrixFilter._colorMatrix` only when `multiply=true`. Then, `_colorMatrix` divides the rightmost column (the offset) by 255. To be honest, I don't understand the reasoning behind that normalization. I don't see any other code that produces an offset that's in the range 0..255 that needs to be normalized, and even if that existed, calling it only in multiply mode doesn't seem correct.

Dropping the normalization fixes this bug, but I'm unsure about the side effects of this change. I think fixing this safely also requires improving the tests of the filter, to verify the effects of each method on the matrix.

### Steps to Reproduce
```
const node = /* Some PIXI node, it doesn't matter */
const filter = new PIXI.filters.ColorMatrixFilter();
node.filters = [filter];
filter.reset(); /* Not necessary, but to clarify that we start with the identity matrix */
filter.contrast(1, true);
 ```

### Environment

- **`pixi.js` version**: Both 5.3 and 6.3
- **Browser & Version**: Chrome 101 (doesn't matter)
- **OS & Version**: Ubuntu 21.10 (doesn't matter)
- **Running Example**: [_url_](https://www.pixiplayground.com/#/edit/NrJ3oM-_T-rL1gp0R98tY)

rubenlg:
I just noticed that `brightness()` and `saturate()` (and all other methods) have the same issue, because they all end up calling `_colorMatrix` for `multiply=true`. If I want to adjust brightness, contrast and saturation, I can't call the three methods in order with `multiply=true`. It doesn't even work to call the first one with `multiply=false` and the others with `multiply=true`.

As a workaround, I need to do the math myself and set the final matrix.

bigtimebuddy:
I don't, this code hasn't been touched in awhile.

PR #11923, opened 2026-02-20
feat: Split ColorMatrixFilter for more nimble MatrixFilter

Sertion:
##### Description of change
* Split ColorMatrixFilter into two classes:
  * `ColorMatrixFilter` - Works as before
  * `MatrixFilter` - Only applies a 5x4 matrix
* Update ColorMatrixFilter documentation to match functionality
* Add return values to ColorMatrixFilter-methods to match documentation
* Fix issue with matrix offsets breaking with multiply (#8359)

##### Pre-Merge Checklist

<details>

##### Breaking Changes / Behavior Changes
- `ColorMatrixFilter` now extends `MatrixFilter` instead of `Filter`
- `ColorMatrixFilter` constructor parameter is now optional: `constructor(options?: FilterOptions)`
- `ColorMatrixFilter` methods now return `this` for method chaining instead of `void`
  ```ts
  // Now chainable
  filter.brightness(0.5).contrast(1.2).saturate(0.8);
  ```
- Public `matrix` and `alpha` getter/setter removed from `ColorMatrixFilter` (moved to `MatrixFilter`)
  ```ts
  // Before
  const matrix = filter.matrix;
  
  // After
  const matrix = filter.matrix; // Still available through MatrixFilter base
  ```
- `ColorMatrix` type export moved from `ColorMatrixFilter` to `MatrixFilter`
- Color adjustment method signatures updated with `multiply` parameter defaulting to `false`

##### Features
- New `MatrixFilter` class for applying 5x4 color matrices with optional alpha blending
  ```ts
  const filter = new MatrixFilter({ alpha: 0.8 });
  filter.prepend(customMatrix);
  filter.append(anotherMatrix);
  ```
- Added `prepend()` and `append()` methods on `MatrixFilter` for matrix composition
- Public `ColorMatrix` type and `MatrixFilterOptions` interface exported from new `MatrixFilter`
- ColorMatrixFilter methods are now chainable
- `MatrixFilter` exported from main filters module

##### Fixes
- Fixed matrix offset issue when using multiply parameter (addresses #8359)

</details>


bigtimebuddy (commented):
I think this split is a good idea.   Separating the convenience stuff from the matrix operations is a good separation of concerns.

I'm not in love with the naming "MatrixFilter" is a little too generic and is confusingly overlaps with ColorMatrixFilter.

What do you think about keeping ColorMatrixFilter and moving the convenience operations to a new filter maybe called ColorPresetFilter... or something else? It'll be a little annoying to setup the deprecations for backward compatibility, but seems clearer in the filter naming long term.

Sertion:
Thanks for the feedback @bigtimebuddy!

I do not like the name `MatrixFilter` either. It is very generic and does not explain very well what it does. I was thinking it could be split further and having the convenience functions be methods that return a matrix that one can assign with the `prepend` and `append` methods on `MatrixFilter` but I did not want to make a change that big and breaking in a project I'm not a well established voice in.

If I were to move it how would the project prefer such a deprecation should work?

What I'm thinking long term is something like this:
```ts
export class ColorMatrixFilter {} // This PR's MatrixFilter

export function brightnessMatrix(intencity: number): ColorMatrix {}
export function vintageMatrix(): ColorMatrix {}

const filter = new ColorMatrixFilter({
    matrix: brightnessMatrix(0.75),
});
filter.append(vintageMatrix());
```

I am not familiar with Pixijs to know if a similar pattern exists elsewhere or what a similar setup would look like in "the pixijs style".

tests/visual/scenes/filters/color-matrix-filter.scene.ts
```diff
@@ -0,0 +1,96 @@
+import { ColorMatrixFilter } from '~/filters';
+import { type Container, Graphics } from '~/scene';
+
+import type { TestScene } from '../../types';
+
+export const scene: TestScene = {
+    excludeRenderers: ['canvas', 'webgpu'],
```
bigtimebuddy, line 7: Why exclude WebGPU?

Sertion: I'm on Linux and could not make it generate an image without falling back to webgl2.

bigtimebuddy:
There are two things in this PR: a new API and a bugfix. Would it be possible to create a smaller, more narrow PR just to fix the multiply bug while we discuss the right plan for the API change?

Sertion:
Good idea. I have created #11925.

bigtimebuddy:
Did some more thinking about this yesterday. I have a proposal how we might organize this:

### ColorMatrixFilter

Narrowly focused on manipulating a color matrix and the math invovled.

- `constructor({ alpha, matrix })`
- `alpha: number`
- `matrix: ColorMatrix`
- `prepend (matrix: ColorMatrix): this`
- `append (matrix: ColorMatrix): this`

### ColorTransformFilter extends ColorMatrixFilter, PixiMixins.ColorTransformFilter

This class is focused on **general** transforms for color. These are unopinionated and what we might expect on other rendering systems. Would also be nice to add convenience during construction, e.g., `new ColorTransformFilter({ saturate: 0.5 })`

- `constructor({ brightness, tint, grayscale, hue, contrast, saturate, alpha })`
- `brightness(b: number, multiply?: boolean): this`
- `tint(color: ColorSource, multiply?: boolean): this`
- `grayscale(scale: number, multiply?: boolean): this`
- `hue(rotation: number, multiply?: boolean): this`
- `contrast(amount: number, multiply?: boolean): this`
- `saturate(amount: number, multiply?: boolean): this`
- `reset(): void`

### PixiMixins.ColorTransformFilter

These are additional methods that can be _optionally_ included (via a new import) and added to the `ColorTranformFilter` class. These are highly-opinionated and somewhat gimmicky color transforms that not everyone should pay the tax for. Would be nice to transition this to be optional.

- `blackAndWhite(scale: number, multiply?: boolean): this`
- `desaturate(multiply?: boolean): this`
- `negative(multiply?: boolean): this`
- `sepia(multiply?: boolean): this`
- `technicolor(multiply?: boolean): this`
- `polaroid(multiply?: boolean): this`
- `toBGR(multiply?: boolean): this`
- `kodachrome(multiply?: boolean): this`
- `browni(multiply?: boolean): this`
- `vintage(multiply?: boolean): this`
- `colorTone(desaturate? number, toned?: number, lightColor?: ColorSource, darkColor?: ColorSource, multiply?: boolean): this`
- `night(intensity?: number, multiply?: boolean): this`
- `predator(intensity?: number, multiply?: boolean): this`
- `lsd(intensity?: number, multiply?: boolean): this`

@Zyie do you have any thoughts on this proposal?

Sertion:
I implemented parts of the suggestion while I wait for more input.

I've re-named the "old" `ColorMatrixFilter` to `ColorTransformFilter` and re-named the new leaner `MatrixFilter` to `ColorMatrixFilter`.

As is probably obvious this is a breaking change.

I did not add the `PixiMixins.ColorTransformFilter` split or the matrix applications as part of the constructor arguments as it still felt uncertain. There is also the question of order for the constructor arguments as I don't think, for example, `grayscale` and then `tint` has the same result as `tint` then `grayscale`.

It would need to be some sort of array of matrix information objects and that could get messy quick if not thought through.

```ts
// This is just an example
type ColorTransformation = {
    operation: "grayscale";
    scale: number;
} | {
    operation: "tint";
    color: ColorSource;
} | ...

type ColorTransformFilterOptions = {
    ...
    operations?: ColorTransformation[],
    ...
}
```

bigtimebuddy:
Let's chat about how to make this non-breaking. I wonder if we could the ColorTransformFilter methods and inject them into ColorMatrixFilter but wrap with a deprecation warning?

```ts
ColorMatrixFilter.prototoype.brightness = function (this: ColorMatrixFilter, v:number, m:boolean) {
  deprecation("Moved to ColorTransformFilter");
  ColorTransformFilter.prototype.brightness.call(this, v, m);
  return this;
}
```

Would this work?

Sertion:
I added a variant of what you suggested as I could not get `ColorMatrixFilter.prototoype.xxx` to play nice with typescript. If you have an example of how to do it better I would love to learn!

It also creates a dependency loop. :crying_cat_face:

bigtimebuddy:
Try Object.defineProperties instead?

Sertion:
The only way I could figure it out was to implement it in both classes. Otherwise there will always be some sort of dependency loop.

I could look into splitting it into "common transformations" and "artistic transformations" before it goes out so there is no need to immediately deprecate a new set of methods, but I don't know how the proposed pattern is supposed to work.

Are there any places in pixi today that use it that I can use as a reference?

PR #11926, opened 2026-02-23
fix: prevent BitmapFont.install cache key collision

### LINKED ISSUE #11894: Bug: BitmapFont.install: Cache collision on generic '-bitmap' key and redundant key generation
### PixiJS version

v8.16.0

https://jsfiddle.net/bm9yxd2r/

### Steps to reproduce

1. Create a basic PixiJS v8 setup.
2. Call `BitmapFont.install()` with a unique `name` (`'FirstFont'`) and a system `style.fontFamily` (e.g., `'Arial'`).
3. Call `BitmapFont.install()` again with a **different** unique `name` (`'SecondFont'`).
4. Inspect the `Cache` keys using `Array.from(Cache['_cache'].keys())` to see the generated keys.
5. Observe the console warning.

### What is expected?

`BitmapFont.install` should use the provided `name` property as the unique identifier for the cache. It should register the font **only once** under the expected key (e.g., `name-bitmap`) without generating redundant or generic keys like `-bitmap` or `...-bitmap-bitmap`. Installing multiple fonts with different names should not cause a cache key collision, regardless of their underlying `fontFamily`.

1. **Cache Key Collision**: On the second `install` call (`SecondFont`), the console outputs: `PixiJS Warning: [Cache] already has key: -bitmap`
2. **Redundant & Malformed Keys**: Inspecting `Cache` keys reveals that each `install` call generates multiple unintended keys: `["FirstFont-bitmap", "FirstFont-bitmap-bitmap", "-bitmap", "SecondFont-bitmap", "SecondFont-bitmap-bitmap"]`
   - The Ghost Key (`-bitmap`): A generic key is created during the first install, causing the second install to collide.
   - Double Suffixing: Keys like `FirstFont-bitmap-bitmap` suggest the internal logic is incorrectly appending the suffix multiple times.

### System Info

```Shell
System:
    OS: macOS 26.2
    CPU: (8) arm64 Apple M2
    Memory: 147.03 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 144.0.7559.133
    Safari: 26.2
```

### Any additional comments?

_No response_

veeceey:
The `bitmapFontCachePlugin` was generating bogus cache keys when `BitmapFont.install()` was called, causing collisions and warnings.

Two problems:

1. `install()` stores fonts under `${name}-bitmap`, but the cache plugin's `getCacheableAssets` was unconditionally appending `-bitmap` again, producing keys like `FirstFont-bitmap-bitmap`.

2. `DynamicBitmapFont` has an empty `fontFamily` by default, so `${asset.fontFamily}-bitmap` produced a bare `-bitmap` key. When a second font was installed, it collided with this ghost key.

The fix:
- Skip appending `-bitmap` suffix when the key already ends with it
- Only create the `fontFamily`-based key when `fontFamily` is non-empty and not already covered

After the fix, `BitmapFont.install({ name: 'FirstFont', ... })` creates just `FirstFont-bitmap`, and a second install with `SecondFont` creates just `SecondFont-bitmap` — no collision, no duplicate keys.

Fixes #11894

### DIFF
--- a/src/scene/text-bitmap/asset/loadBitmapFont.ts
+++ b/src/scene/text-bitmap/asset/loadBitmapFont.ts
@@ -33,10 +33,22 @@ export const bitmapFontCachePlugin = {
         keys.forEach((key) =>
         {
             out[key] = asset;
-            out[`${key}-bitmap`] = asset;
+            if (!key.endsWith('-bitmap'))
+            {
+                out[`${key}-bitmap`] = asset;
+            }
         });
 
-        out[`${asset.fontFamily}-bitmap`] = asset;
+        if (asset.fontFamily)
+        {
+            const familyKey = `${asset.fontFamily}-bitmap`;
+
+            // Only add fontFamily key if it's not already covered by the input keys
+            if (!out[familyKey])
+            {
+                out[familyKey] = asset;
+            }
+        }
 
         return out;
     }
--- a/src/scene/text-bitmap/asset/loadBitmapFont.ts
+++ b/src/scene/text-bitmap/asset/loadBitmapFont.ts
@@ -33,21 +33,13 @@ export const bitmapFontCachePlugin = {
         keys.forEach((key) =>
         {
             out[key] = asset;
-            if (!key.endsWith('-bitmap'))
-            {
-                out[`${key}-bitmap`] = asset;
-            }
+            out[`${key}-bitmap`] = asset;
         });
 
         if (asset.fontFamily)
         {
-            const familyKey = `${asset.fontFamily}-bitmap`;
-
             // Only add fontFamily key if it's not already covered by the input keys
-            if (!out[familyKey])
-            {
-                out[familyKey] = asset;
-            }
+            out[`${asset.fontFamily}-bitmap`] ??= asset;
         }
 
         return out;
--- a/src/scene/text-bitmap/BitmapFontManager.ts
+++ b/src/scene/text-bitmap/BitmapFontManager.ts
@@ -478,13 +478,15 @@ class BitmapFontManagerClass
             textureStyle: options.textureStyle,
         });
 
+        (font as { fontFamily: string }).fontFamily = name;
+
         const flatChars = resolveCharacters(options.chars);
 
         font.ensureCharacters(flatChars.join(''));
 
-        Cache.set(`${name}-bitmap`, font);
+        Cache.set(name, font);
 
-        font.once('destroy', () => Cache.remove(`${name}-bitmap`));
+        font.once('destroy', () => Cache.remove(name));
 
         return font;
     }
@@ -495,8 +497,7 @@ class BitmapFontManagerClass
      */
     public uninstall(name: string)
     {
-        const cacheKey = `${name}-bitmap`;
-        const font = Cache.get<BitmapFont>(cacheKey);
+        const font = Cache.get<BitmapFont>(`${name}-bitmap`) ?? Cache.get<BitmapFont>(name);
 
         if (font)
         {


veeceey:
### Manual test results

Full unit test suite passes (167/167 suites, 1716/1716 tests):
```
Test Suites: 1 skipped, 167 passed, 167 of 168 total
Tests:       16 skipped, 1716 passed, 1732 total
```

**Before the fix**, calling `BitmapFont.install` twice with different names would produce:
- Keys: `FirstFont-bitmap`, `FirstFont-bitmap-bitmap`, `-bitmap`, `SecondFont-bitmap`, `SecondFont-bitmap-bitmap`
- Warning: `[Cache] already has key: -bitmap`

**After the fix**, each install only creates one key:
- Keys: `FirstFont-bitmap`, `SecondFont-bitmap`
- No warnings

src/scene/text-bitmap/asset/loadBitmapFont.ts
```diff
@@ -33,10 +33,22 @@ export const bitmapFontCachePlugin = {
...
+        if (asset.fontFamily)
+        {
+            const familyKey = `${asset.fontFamily}-bitmap`;
+
+            // Only add fontFamily key if it's not already covered by the input keys
+            if (!out[familyKey])
+            {
+                out[familyKey] = asset;
+            }
+        }
```
albinkong, line 43: This looks like a great fix. You can shorten this with the following by using an nullish coalescing assignment:


```suggestion
        if (asset.fontFamily)
        {
            // Only add fontFamily key if it's not already covered by the input keys
            out[`${asset.fontFamily}-bitmap`] ??= asset;
        }
```

(Sorry for the double comment, the Github UI sometimes does not behave)

src/scene/text-bitmap/asset/loadBitmapFont.ts
```diff
@@ -33,10 +33,22 @@ export const bitmapFontCachePlugin = {
         keys.forEach((key) =>
         {
             out[key] = asset;
-            out[`${key}-bitmap`] = asset;
+            if (!key.endsWith('-bitmap'))
+            {
+                out[`${key}-bitmap`] = asset;
+            }
```
bigtimebuddy: Why is this suffix being appended twice? That seems like not super desirable and this feels like a bandaid/mask of a potential underlying issue.

src/scene/text-bitmap/asset/loadBitmapFont.ts
```diff
@@ -33,10 +33,22 @@ export const bitmapFontCachePlugin = {
...
         });
 
-        out[`${asset.fontFamily}-bitmap`] = asset;
+        if (asset.fontFamily)
```
bigtimebuddy, line 39: When is fontFamily empty?

src/scene/text-bitmap/asset/loadBitmapFont.ts
veeceey: Good question — I agree this shouldn't just be papered over.

The double-suffix issue comes from the interaction between `BitmapFont.install()` and `getCacheableAssets`. Previously, `install()` stored the font under `${name}-bitmap` in the Cache directly. But `getCacheableAssets` also appends `-bitmap` to every key it receives — that's its job, since asset-loader keys don't have the suffix yet. So when `install()` pre-baked `-bitmap` into the name, getCacheableAssets would produce `name-bitmap-bitmap`.

Rather than changing how getCacheableAssets works (which would break the normal asset-loader path where keys genuinely don't have the suffix), the fix I went with is to have `install()` store under just `name` — without the `-bitmap` suffix — so the two paths don't clash. The `uninstall()` method then checks both key forms for backward compatibility.

So it's not really masking the issue — the root cause was that `install()` was adding a suffix that only `getCacheableAssets` should be adding, and the fix is to stop `install()` from doing that.

src/scene/text-bitmap/asset/loadBitmapFont.ts
veeceey: When `BitmapFont.install()` is called, it creates a `DynamicBitmapFont`, which extends `AbstractBitmapFont`. The `fontFamily` property on `AbstractBitmapFont` is initialized to an empty string (`''`) by default (line 108 in AbstractBitmapFont.ts). Unlike `BitmapFont` (which gets `fontFamily` from parsed font data), `DynamicBitmapFont` never sets it.

That's actually the other part of this fix — before my change, `getCacheableAssets` would evaluate `${asset.fontFamily}-bitmap` as just `'-bitmap'` (empty string + suffix), creating a bogus key that would collide as soon as a second font was installed.

I've addressed this two ways: the guard `if (asset.fontFamily)` prevents the empty-string key, and I also set `fontFamily = name` in the `install()` method so the DynamicBitmapFont actually has a meaningful fontFamily going forward.

src/scene/text-bitmap/asset/loadBitmapFont.ts
veeceey: Thanks\! I actually already have `??=` in the updated code — looks like we're on the same page there. Appreciate the review\!

Zyie:
closing this down as it does not solve the root issue. the `test` for `loadBitmapFont` is incorrect

PR #11929, opened 2026-02-25
fix: NineSliceSprite renders outside trimmed frame

### LINKED ISSUE #11706: NineSliceSprite renders outside trimmed frame
### PixiJS version

v8.x

https://stackblitz.com/edit/pixijs-v8-yjsnwcxz?file=src%2Fmain.js

### Steps to reproduce

1. Create a `Spritesheet` with a trimmed frame that does not occupy full texture area. Include `sourceSize` and `spriteSourceSize` to simulate trimming.
2. Create regular `Sprite` and `NineSliceSprite` using the same trimmed frame.
3. Render both sprites side by side and observe the visual difference.

### What is expected?

Both types of sprite should position and clip trimmed content correctly.

`Sprite` uses `spriteSourceSize` and `sourceSize` to clip and position the trimmed content correctly within the original dimensions.
`NineSliceSprite` does not clip the texture to `sourceSize`, so it renders extra pixels that were meant to be outside the visible area (i.e., trimmed away).
This results in unexpected visual output, especially when the trimmed texture contains content outside the intended frame.

### Environment

- **Browser & Version**: *Chrome 140*

### Any additional comments?

_No response_

luxh1121:
## Summary

Fixes #11706

When a `NineSliceSprite` uses a trimmed/atlas texture (e.g. from a `Spritesheet` with `spriteSourceSize` and `sourceSize`), the UV coordinates were computed assuming the texture occupies the full 0–1 UV range. This caused the sprite to render pixels outside the trimmed frame area, unlike `Sprite` which correctly clips to the trimmed region.

### `NineSliceGeometry`
- Added optional `textureUvs` field to `NineSliceGeometryOptions`
- When `textureUvs` is provided, `updateUvs()` computes UV coordinates directly in texture-source space, mapping the nine-slice cut points within the actual frame UV boundaries
- When `textureUvs` is not provided, behavior is unchanged (0–1 UV range)

### `NineSliceSpritePipe`
- Detects when a texture has a non-trivial frame (atlas/trimmed) by checking if the frame covers the entire source
- Passes `texture.uvs` to the geometry so UVs are computed in source space
- Overrides the `uvs` getter in `NineSliceSpriteGpuData` to bypass `TextureMatrix` transformation when UVs are already baked in source space (prevents double-transformation)

### Tests
- Added unit tests verifying correct UV computation with and without `textureUvs`

## How it works

For a trimmed texture with frame UVs `(x0, y0)` to `(x1, y3)` in source-normalized space:
- The nine-slice UV boundaries are mapped to `[x0, x1]` × `[y0, y3]` instead of `[0, 1]` × `[0, 1]`
- Cut points (left/right/top/bottom) are computed as fractions of the frame UV range
- The `NineSliceSpriteGpuData.uvs` getter returns the geometry UVs directly, bypassing `TextureMatrix.multiplyUvs()` which would otherwise double-transform them

For non-trimmed textures, the existing behavior is preserved.

<details>

##### Fixes
- Fixed NineSliceSprite rendering pixels outside the trimmed frame when using trimmed or atlas textures. The sprite now correctly computes UVs in texture-source space for trimmed frames, matching the behavior of regular Sprite.

</details>

### DIFF
--- a/src/scene/sprite-nine-slice/NineSliceGeometry.ts
+++ b/src/scene/sprite-nine-slice/NineSliceGeometry.ts
@@ -28,6 +28,9 @@ export interface NineSliceGeometryOptions
 
     /** The anchor point of the NineSliceSprite. */
     anchor?: PointData
+
+    /** The texture UVs to use for mapping. When provided, UVs are computed directly in texture-source space. */
+    textureUvs?: { x0: number; y0: number; x1: number; y1: number; x2: number; y2: number; x3: number; y3: number }
 }
 
 /**
@@ -71,6 +74,8 @@ export class NineSliceGeometry extends PlaneGeometry
     private _originalHeight: number;
     private _anchorX: any;
     private _anchorY: number;
+    /** @internal */
+    public _textureUvs: { x0: number; y0: number; x1: number; y1: number; x2: number; y2: number; x3: number; y3: number } | null = null;
 
     constructor(options: NineSliceGeometryOptions = {})
     {
@@ -103,6 +108,8 @@ export class NineSliceGeometry extends PlaneGeometry
 
         this._anchorX = options.anchor?.x;
         this._anchorY = options.anchor?.y;
+
+        this._textureUvs = options.textureUvs ?? this._textureUvs;
 
         this.updateUvs();
         this.updatePositions();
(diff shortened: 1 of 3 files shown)


bigtimebuddy:
Thanks for submitting this, however, we already have an approved PR #11919 which solves the same problem.

luxh1121:
Closing this as #11919 already addresses the issue. Thanks for the review!

PR #11930, opened 2026-02-25
fix: correct MeshRope aspect ratio when textureScale > 0

### LINKED ISSUE #10821: Bug: MeshRope does not have correct aspect ratio when using textureScale > 0
### Current Behavior

When setting `textureScale` to a value greater than zero, the MeshRope tiles the texture with an aspect ratio of 1 instead of the texture's aspect ratio.

### Expected Behavior

The MeshRope should tile the texture with the texture's own aspect ratio.

### Steps to Reproduce

In the MeshRope demo on pixi.io, set `textureScale` to 1 in the mesh options. Observe that the snake is tiled in squares.

### Environment

- **`pixi.js` version**: *8.1.*0

### Possible Solution

Right now, the Rope geometry's `_width` param is set to the texture's height, but _width seems to be used as both the geometry width for UV calculation and height of the geometry, hence the 1.0 aspect ratio of height/height. We might need to add in a `_height` property as well and use that in calculations.

### Additional Information

_No response_

luxh1121:
## Summary

Fixes the incorrect 1:1 aspect ratio when using `textureScale > 0` on `MeshRope`.

Fixes #10821

## Problem

When `textureScale > 0`, the UV tiling calculation in `RopeGeometry._build()` used `_width` (rope thickness = `texture.height`) for the horizontal UV scale:

```ts
const textureWidth = this._width * this.textureScale;
```

Since `_width` equals `texture.height`, this produces a 1:1 aspect ratio regardless of the texture's actual dimensions. A 200x50 texture would tile as if it were 50x50.

## Solution

Added a `_textureWidth` property to `RopeGeometry` that stores the texture's horizontal width separately from `_width` (rope thickness). The UV calculation now uses:

```ts
const textureWidth = this._textureWidth * this.textureScale;
```

`MeshRope` passes `texture.width` as `textureWidth` to `RopeGeometry` and keeps it in sync during render updates.

## Backward Compatibility

- When `textureWidth` is not provided, it falls back to `_width` (old behavior)
- When `textureScale` is 0 (stretch mode), behavior is unchanged
- No breaking changes to the public API

## Files Changed

- `src/scene/mesh-simple/RopeGeometry.ts` - Added `_textureWidth` property and `textureWidth` option, fixed UV calculation
- `src/scene/mesh-simple/MeshRope.ts` - Pass `texture.width` as `textureWidth`, sync on render
- `src/scene/mesh-simple/__tests__/RopeGeometry.test.ts` - Unit tests for aspect ratio fix

## Testing

```ts
// Before: texture tiles as squares (1:1 aspect ratio)
// After: texture tiles with correct aspect ratio
const rope = new MeshRope({
  texture: snakeTexture, // e.g. 200x50
  points,
  textureScale: 1
});
```

### DIFF
--- a/src/scene/mesh-simple/MeshRope.ts
+++ b/src/scene/mesh-simple/MeshRope.ts
@@ -196,7 +196,7 @@ export class MeshRope extends Mesh
     {
         const geometry: RopeGeometry = this.geometry as any;
 
-        if (this.autoUpdate || geometry._width !== this.texture.height)
+        if (this.autoUpdate || geometry._width !== this.texture.height || geometry._textureWidth !== this.texture.width)
         {
             geometry._width = this.texture.height;
             geometry._textureWidth = this.texture.width;
--- a/src/scene/mesh-simple/MeshRope.ts
+++ b/src/scene/mesh-simple/MeshRope.ts
@@ -169,7 +169,12 @@ export class MeshRope extends Mesh
     constructor(options: MeshRopeOptions)
     {
         const { texture, points, textureScale, ...rest } = { ...MeshRope.defaultOptions, ...options };
-        const ropeGeometry = new RopeGeometry(definedProps({ width: texture.height, points, textureScale }));
+        const ropeGeometry = new RopeGeometry(definedProps({
+            width: texture.height,
+            points,
+            textureScale,
+            textureWidth: texture.width,
+        }));
 
         if (textureScale > 0)
         {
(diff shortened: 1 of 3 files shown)


luxh1121:
Addressed both CodeRabbit review comments in commit 62c2219:

1. **Texture width change detection (Major):** Added `geometry._textureWidth !== this.texture.width` check in `MeshRope._render()`, so texture width changes are now detected even when `autoUpdate=false`.
2. **Division by zero guard (Minor):** Added fallback `|| 1` for `textureWidth` in `RopeGeometry.update()` to prevent `Infinity`/`NaN` when `_textureWidth * textureScale` evaluates to 0.

Zyie:
@luxh1121 can you create visual test for this and show a before/after for the change please!

bigtimebuddy:
This adds another confusingly named "width" option to the constructor creating a bad DX. This is not the right solution to this issue. Closing

PR #11932, opened 2026-02-26
feat: DOM Container configuration

theKingSTL:
Adds a small dom.wrapper init option so the DOMContainer wrapper div can be styled/configured directly. Supports wrapper className, attributes, and style without relying on brittle CSS selectors. Keeps default behavior unchanged and includes tests for both configured and default paths.

##### Description of change
Adds a dom.wrapper renderer/app init option to configure the internal DOMContainer wrapper div without brittle CSS selectors.Supports wrapper className, attributes, and style, while preserving default behavior when no options are provided.Also updates typings and adds unit tests for both configured and default paths.

##### Pre-Merge Checklist
[x] Tests and/or benchmarks are included
[ ] Documentation is changed or added
[ ] Lint process passed (npm run lint)
[ ] Tests passed (npm run test)


src/dom/DOMPipe.ts
```diff
@@ -7,6 +7,41 @@ import type { InstructionSet } from '../rendering/renderers/shared/instructions/
...
+    /** @standard */
+    attributes?: Record<string, string>;
+    /** @standard */
+    style?: DOMPipeStyleOptions;
+}
+
+/**
+ * Options for the DOM render pipe.
+ * @category - Optional
+ * @internal
```
Zyie: these types shouldn't have the internal tag

src/dom/DOMPipe.ts
```diff
@@ -63,6 +98,17 @@ export class DOMPipe implements RenderPipe<DOMContainer>
     /** Initializes the DOMPipe, setting up the main DOM element and adding it to the document body. */
     public init(): void
     {
+        // `renderer.options` (public) is the canonical home for all init options once
+        // the renderer has finished initialising. `_initOptions` is an internal field
+        // that exists in some versions but does not always carry top-level keys like
+        // `dom` through to pipes. Checking both makes the lookup resilient.
+        const rendererAsAny = this._renderer as any;
+        const domPipeOptions: DOMPipeOptions | undefined
```
Zyie: options should be being passed as the first parameter to the init method, not sure why all this is needed

Zyie (changes requested):
Thanks for putting together this PR
There seems to be linting issues that are causing the CI to fail

bigtimebuddy (commented):
Thanks for the PR.

I'm not convinced this is needed. It's creating some convenience but of which there are easy out of the box alternatives.

Like creating the canvas manually:

```ts
const canvas = document.createElement("canvas"):
canvas.classList.add("pixi")

const app = new Application()
await app.init({ canvas })
```

Or using the DOM API directly after creating the renderer:

```ts
const app = new Application()
await app.init({})

app.canvas.classList.add("pixi")
```

Why not this simpler approaches, which doesn't expand the Pixi API at all?

carlos22:
> Thanks for the PR.
> 
> I'm not convinced this is needed. It's creating some convenience but of which there are easy out of the box alternatives.
> 
> Like creating the canvas manually:
> 
> ```ts
> const canvas = document.createElement("canvas"):
> canvas.classList.add("pixi")
> 
> const app = new Application()
> await app.init({ canvas })
> ```
> 
> Or using the DOM API directly after creating the renderer:
> 
> ```ts
> const app = new Application()
> await app.init({})
> 
> app.canvas.classList.add("pixi")
> ```
> 
> Why not this simpler approaches, which doesn't expand the Pixi API at all?

But that is something different isn't it? The DOM wrapper element is not the `canvas` or is it? In our app we do some obscure targeting of the DOM wrapper via CSS as its just an injected div with not class nothing. So I think its nice to be able to change that. I don't see how your code would change that?! 

EDIT:
We are talking about this [div](https://github.com/theKingSTL/pixijs/blob/f615113919801abaaebae5ca44f54c87853f1e46/src/dom/DOMPipe.ts#L87) and  AFAIK there is no way to access it form outside of pixi or even select it reliable with in an app that has also (dynamic) DOM content.

bigtimebuddy:
Apologies, I misunderstood. 

Why not expose (add a getter) the domElement in the DOMPipe? When you could set attributes or styles to whatever.

carlos22:
> Apologies, I misunderstood.
> 
> Why not expose (add a getter) the domElement in the DOMPipe? When you could set attributes or styles to whatever.

Also a possibility. Similar to `app.canvas` right? But I think that is not so easy due to the plugin/extension architecture?

bigtimebuddy:
This is how one would address that now. 

```ts
app.renderer.renderPipes.dom._domElement
```

We could add a convenience getter, maybe on Application. What about `app.domRootElement`? Open to other name suggestions, "domContainerRoot", "domContainerParent", etc

carlos22:
Naming things is hard ;-) But I also think `domContainerRoot` is good. The type will be Element anyway and the feature is called DOM Containers right?

bigtimebuddy:
I like that. DOMContainer is the feature so that makes sense

Zyie:
going to close this one down in favor of #11974

Issue #11823, opened 2026-01-03
Documentation differs from actual behavior in BackgroundLoader.


tonku321:
### PixiJS version

8.14.3

### Link to minimal reproduction

https://github.com/pixijs/pixijs/blob/98ea728fc887ea93a1131a61cac66ab4fc45eebe/src/assets/BackgroundLoader.ts#L39

### Steps to reproduce

1. [Assets are cached as they complete loading](https://github.com/pixijs/pixijs/blob/98ea728fc887ea93a1131a61cac66ab4fc45eebe/src/assets/BackgroundLoader.ts#L39) is [incorrect](https://github.com/pixijs/pixijs/blob/98ea728fc887ea93a1131a61cac66ab4fc45eebe/src/assets/BackgroundLoader.ts#L127-L148). They are not available in `Assets.cache` after loading completed. `Assets.backgroundLoadBundle` - also say they [are cached](https://github.com/pixijs/pixijs/blob/98ea728fc887ea93a1131a61cac66ab4fc45eebe/src/assets/Assets.ts#L915). I think this is more of a bug than incorrect documentation.
2. [The max assets it will try to load at one time will be 4](https://github.com/pixijs/pixijs/blob/98ea728fc887ea93a1131a61cac66ab4fc45eebe/src/assets/BackgroundLoader.ts#L125C8-L125C64) is incorrect, [actually 1](https://github.com/pixijs/pixijs/blob/98ea728fc887ea93a1131a61cac66ab4fc45eebe/src/assets/BackgroundLoader.ts#L76).
3. Configurable concurrency is wrong, actually [the field is private and readonly](https://github.com/pixijs/pixijs/blob/98ea728fc887ea93a1131a61cac66ab4fc45eebe/src/assets/BackgroundLoader.ts#L58). Also i think it would be great, if it was configurable.

### What is expected?

-

### What is actually happening?

-

### Environment

-

PR #11935, opened 2026-02-27
fix: clarify BackgroundLoader cache behavior in documentation (issue #11823)

imuday984:
##### Description of change
✅ Documentation updated to reflect real behaviour

The “cached immediately” line in the background‑loader docs turned out to be the source of confusion.

In reality the loader just queues assets at low priority.
They only enter the normal cache once the background fetch actually finishes – not at the moment you call Assets.backgroundLoad().

That means:

calling [Assets.load()](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for an asset that’s still in the background queue will not resolve instantly; the loader will bump it to high priority and download it again, pausing the background process.
only when the background loader has completed the individual request will subsequent [Assets.load()](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/loadBundle() calls become instant.
What changed
The top of [background-loader.md](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) now reads:

The background loader queues asset downloads at low priority so they happen between frames while your application runs. Assets added to the background queue are not downloaded immediately – they are fetched one at a time (or up to the configured concurrency limit) while the main application continues. Only once each item has actually finished downloading is it placed into the regular cache.

Important: a background‑queued asset is not instantly available just because you called Assets.backgroundLoad. If you call [Assets.load()](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) or Assets.loadBundle() for the same asset before the background fetch has completed, the loader will treat it as a high‑priority request and perform a fresh download (pausing the background process); the call only resolves instantly when the background load has already finished.

This aligns the docs with the behaviour described in [BackgroundLoader.ts](vscode-file://vscode-app/d:/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and resolves the discrepancy mentioned in issue #11823.

- [✅] Documentation is changed or added

<details>

##### Chores
- Updated Background Loader documentation to clarify that background-queued assets are not immediately downloaded. Assets are fetched sequentially (or up to configured concurrency) and only placed in cache after completion. Added note that calling `Assets.load()` or `Assets.loadBundle()` before background fetch completes triggers a fresh high-priority download, pausing the background process.

</details>

### DIFF
--- a/src/assets/__docs__/background-loader.md
+++ b/src/assets/__docs__/background-loader.md
@@ -6,7 +6,10 @@ category: assets
 
 # Background loader
 
-The background loader queues asset downloads at low priority so they happen between frames while your application runs. Assets loaded this way are cached normally; when you later call `Assets.load()` or `Assets.loadBundle()` for the same asset, it resolves instantly if the background load already finished.
+The background loader queues asset downloads at low priority so they happen between frames while your application runs. Assets added to the background queue are **not** downloaded immediately – they are fetched one at a time (or up to the configured concurrency limit) while the main application continues. Only once each item has actually finished downloading is it placed into the regular cache.
+
+> **Important:** a background‑queued asset is *not* instantly available just because you called `Assets.backgroundLoad`. If you call `Assets.load()` or `Assets.loadBundle()` for the same asset before the background fetch has completed, the loader will treat it as a high‑priority request and perform a fresh download (pausing the background process); the call only resolves instantly when the background load has already finished.
+
 
 ## Loading bundles
 


imuday984:
do let me know whether it is fine or not

Zyie (approved):

PR #11938, opened 2026-02-28
fix: meshrope tiling aspect ratio when textureScale > 0

### LINKED ISSUE #10821: Bug: MeshRope does not have correct aspect ratio when using textureScale > 0

yuliuyi717-ux:
## Summary
- add a dedicated rope UV tiling width so uv distance can use texture width instead of rope thickness
- pass `texture.width` from `MeshRope` into `RopeGeometry` for textureScale tiling calculations
- add regression tests for both the new path and backward-compatible fallback behavior

## Test Plan
- `npm run test:unit -- --testPathPattern=src/scene/mesh-simple/__tests__/RopeGeometry.test.ts`

Fixes #10821

<details>

##### Fixes
- Fixed MeshRope tiling textures with incorrect aspect ratio when `textureScale > 0`. MeshRope now uses the texture's actual aspect ratio instead of 1:1 (square) tiling.

##### Features
- Added optional `textureWidth` parameter to `RopeGeometry` to explicitly specify texture width for UV tiling calculations, independent of rope width. When not provided, falls back to rope width for backward compatibility.
  ```ts
  const geometry = new RopeGeometry({
    width: 50,
    textureWidth: 256, // Texture's actual width
    points: [...],
    textureScale: 1,
  });
  ```

</details>

### DIFF
--- a/src/scene/mesh-simple/MeshRope.ts
+++ b/src/scene/mesh-simple/MeshRope.ts
@@ -169,7 +169,9 @@ export class MeshRope extends Mesh
     constructor(options: MeshRopeOptions)
     {
         const { texture, points, textureScale, ...rest } = { ...MeshRope.defaultOptions, ...options };
-        const ropeGeometry = new RopeGeometry(definedProps({ width: texture.height, points, textureScale }));
+        const ropeGeometry = new RopeGeometry(
+            definedProps({ width: texture.height, textureWidth: texture.width, points, textureScale })
+        );
 
         if (textureScale > 0)
         {
--- a/src/scene/mesh-simple/RopeGeometry.ts
+++ b/src/scene/mesh-simple/RopeGeometry.ts
@@ -20,6 +20,8 @@ export interface RopeGeometryOptions
 {
     /** The width (i.e., thickness) of the rope. */
     width?: number;
+    /** The texture width used to calculate UV tiling distance when textureScale is enabled. */
+    textureWidth?: number;
     /** An array of points that determine the rope. */
     points?: PointData[];
     /**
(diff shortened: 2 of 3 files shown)


yuliuyi717-ux:
Quick follow-up: this PR is ready for maintainer review.

Scope is focused on the tiling aspect-ratio bug when : UV tiling now uses texture width (while rope width still uses texture height), plus regression coverage in .

yuliuyi717-ux:
Quick follow-up (formatting-safe): this PR is ready for maintainer review.

Scope is focused on the tiling aspect-ratio bug when `textureScale > 0`:
- UV tiling now uses texture width
- rope width still uses texture height
- includes regression coverage in `src/scene/mesh-simple/__tests__/RopeGeometry.test.ts`

bigtimebuddy:
Duplicate of #11930 closing

PR #11951, opened 2026-03-06
fix: removeAllListeners() on Renderer.destroy()

taye:
##### Description of change
While profiling memory usage, I noticed some objects not being garbage collected because they were still referenced by event listeners that weren't removed by a renderer after destroying it. This PR resolves the issue by calling `this.removeAllListeners()` in the destroy method.

##### Pre-Merge Checklist


Zyie (approved):
thank you!

PR #11970, opened 2026-03-13
chore: replace pr-title check with inline github-script

Zyie:
### Overview

Replaces the third-party `clowdhaus/actions/pr-title` action with an inline `actions/github-script` implementation. The new script avoids the spam that the old action created, there is now only one comment per PR which gets minimised when the title is correct

#### Chores
- Replaced `clowdhaus/actions/pr-title` with `actions/github-script` for PR title validation
- Added concurrency group to cancel redundant workflow runs
- Bot now posts/updates a warning comment on invalid titles and minimizes it once fixed
- Tightened title regex to require a prefix and non-empty description

##### Pre-Merge Checklist

### DIFF
--- a/.github/workflows/pull-name.yml
+++ b/.github/workflows/pull-name.yml
@@ -10,16 +10,152 @@ on:
       - edited
       - synchronize
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
   prTitle:
     name: Check
+    if: github.repository == 'pixijs/pixijs'
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


GoodBoyDigital:
Wander if this could be llm automated? Like- auto Change it but also post giving user opportunity to change if they don't like it?

bigtimebuddy (commented):
Can this be externalized into its own shared action for our org? I'd like to use this on other projects.

Also can we expand the regex to include more of the conventional commits format such as expanded types and also scope.

bigtimebuddy (commented):
One more thing, don't inline the script like this it hides it from eslint and other static analysis

PR #11974, opened 2026-03-17
feat: provide a getter to the domElement via app.domContainerRoot

carlos22:
##### Description of change
Add a getter to the domElement that is the wrapper/root container for DOMContainers. This way its possible to add a class, or change any aspects of that ` HTMLDivElement`.

Related: https://github.com/pixijs/pixijs/pull/11932

##### Pre-Merge Checklist

### DIFF
--- a/src/app/Application.ts
+++ b/src/app/Application.ts
@@ -358,6 +358,16 @@ export class Application<R extends Renderer = Renderer>
     }
 
     /**
+     * Get the html div element that holds all DOM Container elements.
+     * @readonly
+     * @type {HTMLDivElement}
+     */
+    get domContainerRoot()
+    {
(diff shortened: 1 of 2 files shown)


src/app/Application.ts
```diff
@@ -357,6 +357,15 @@ export class Application<R extends Renderer = Renderer>
         return this.renderer.screen;
     }
 
+    /**
+     * Get the html div element that holds all DOM Container elements.
+     * @readonly
+     * @type {HTMLDivElement}
+     */
+    get domContainerRoot() {
```
bigtimebuddy:
```suggestion
    get domContainerRoot()
    {
```

bigtimebuddy (approved):
I appreciate the simplicity of this approach. Thanks for suggesting it.

PR #11976, opened 2026-03-17
fix: skip pointermove on canvas when pointer is over an overlapping D…

### LINKED ISSUE #10975: Bug: moving mouse over a div also triggers movemouse (or pointermove) event in the underlying pixijs-canvas
### Current Behavior

I have a pixijs-canvas with a div element stacked on top. 
When I move my mouse over the div element, it also triggers a `mousemove` (or actually a `pointermove`) event in the underlying pixijs-canvas.

### Expected Behavior

I would expect moving my mouse over an element does not trigger pointermove or mousemove events in the underlying pixijs-canvas element, just like moving my mouse over the blue square does not trigger mousemove events in the underlying pink square.

### Steps to Reproduce

https://codesandbox.io/p/sandbox/pixi-mousemove-pointer-events-dktsr3
If you open the dev console, you can see that moving the cursor over the blue or pink square also creates events in the underlying pixijs-canvas.
You can also see that moving your mouse over the blue div does not trigger events in the underlying pink div, as I would expect. 

### Environment

- **`pixi.js` version**: *8.4.1*
- **Browser & Version**: *Firefox 130, Chrome 129*
- **OS & Version**: *Pop!_OS 22.04*
- **Running Example**: *https://codesandbox.io/p/sandbox/pixi-mousemove-pointer-events-dktsr3*

### Possible Solution

_No response_

### Additional Information

_No response_

Philipp-Glaubitz:
fix: skip pointermove when pointer is over an overlapping DOM element

`pointermove`/`mousemove` events are registered on `globalThis.document` with `capture: true` so that dragging works outside the canvas. This caused hover events to fire on canvas objects even when the cursor was over an overlapping DOM element (e.g. a UI overlay `<div>`).

Fix: in `_onPointerMove`, resolve the real event target via `composedPath()` (for shadow DOM parity with `_onPointerUp`) and return early when the target is a known non-canvas element and no pointer buttons are held. The `buttons` guard preserves drag-outside-canvas behaviour: while a button is pressed events continue to flow regardless of which element the cursor is over.

Fixes #10975

##### Description of change

When a DOM element (e.g. a UI overlay) is placed on top of a PixiJS canvas, moving the cursor over it incorrectly triggers `pointermove`/`pointerover` events on canvas objects beneath it. This is because move events are captured at the `document` level before the browser routes them to the actual target element.

The fix adds a target check to `_onPointerMove` — the same pattern already used in `_onPointerUp` — that skips processing when the event's real target is not the canvas and no mouse button is held down.

##### Pre-Merge Checklist

### DIFF
--- a/src/events/EventSystem.ts
+++ b/src/events/EventSystem.ts
@@ -666,6 +666,20 @@ export class EventSystem implements System<EventSystemOptions>
         if (!this.features.move) return;
         this.rootBoundary.rootTarget = this.renderer.lastObjectRendered;
 
+        // If the pointer is over an overlapping DOM element (not the canvas) and no buttons are held
+        // (i.e. not dragging), skip processing to avoid spurious hover events on canvas objects.
+        let target = nativeEvent.target;
+
+        if (nativeEvent.composedPath && nativeEvent.composedPath().length > 0)
+        {
+            target = nativeEvent.composedPath()[0];
+        }
+
+        if (target && target !== this.domElement && ('buttons' in nativeEvent ? nativeEvent.buttons : 0) === 0)
+        {
+            return;
+        }
+
         EventsTicker.pointerMoved();
 
         const normalizedEvents = this._normalizeToPointerData(nativeEvent);
--- a/src/events/__tests__/EventSystem.test.ts
+++ b/src/events/__tests__/EventSystem.test.ts
@@ -1219,4 +1219,56 @@ describe('EventSystem', () =>
 
         expect(eventSpy).toHaveBeenCalledTimes(2);
     });
+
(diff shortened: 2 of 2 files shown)


FelixNumworks:
Checking if a button is pressed to avoid tempering with "drag" behaviour is a bit weird.
Why don't you just disable `pointermove` completely ? We already have `globalpointermove` for detecting things outside of the canvas.

FelixNumworks:
I feel that the change should rather be done in [EventBoundary.ts](https://github.com/pixijs/pixijs/blob/fd367781840da88f7d2352cd51b0f1b651931810/src/events/EventBoundary.t)
Probably in the `mapPointerMove` method (i.e. do not map `pointermove`, `pointerover`, etc. if the native event target is not the canvas)

PR #11979, opened 2026-03-18
fix: stroke-only Graphics masks now render correctly on Canvas renderer

### LINKED ISSUE #11969: Stroke mask not working with canvas
### PixiJS version

8.17.0

https://codepen.io/xevin/pen/raMjJZJ?editors=0011

### Steps to reproduce

1. create mask with without fill, like Graphics().circle().stroke() 
2. apply mask to any Sprite
3. check the stroke-mask is working
4. add `preference: "canvas"` to app.init({ })
5. now stroke-mask is like fill mask

### What is expected?

same mask with canvas and webgl

dunno

### System Info

```Shell
Chrome, Firefox
```

### Any additional comments?

**WebGL** (as expected)

Image

**Canvas** (wrong)

Image

DmitriyGolub:
Stroke-only Graphics masks now render correctly on Canvas renderer. Previously,
stroke masks were treated as fill masks because Canvas 2D `clip()` only operates
on filled path region.

The fix converts stroke geometry to fill triangles using the existing `buildLine`
utility (the same approach used in `CanvasGraphicsAdaptor` for non-centered stroke
alignment), then uses those triangles as the clip path.

Fixes #11969

##### Description of change

`CanvasStencilMaskPipe` previously treated `fill` and `stroke` instructions
identically it built a path and called `context.clip()`. Since `clip()` clips
to the interior of the path (the filled region), stroke masks behaved as fill
masks, masking the entire shape area instead of just the stroke outline.

**Changes in `CanvasStencilMaskPipe.ts`:**
- Added `buildStrokeMaskPath()` function that converts stroke to fill geometry
  via `buildLine` + `shapeBuilders` , then constructs a clip path from the resulting
  triangles
- Modified the instruction loop to detect `stroke` actions and route them through
  the new function, while `fill` actions continue through the original code path unchanged

**Fill masks are not affected** — the `else` branch contains the exact same code
as before.

##### Pre-Merge Checklist


bigtimebuddy (commented):
Nice one! Could you add a visual test for this?

DmitriyGolub:
> Nice one! Could you add a visual test for this?

Like that?

bigtimebuddy (approved):

PR #11985, opened 2026-03-24
fix: skip _applyMipRange for single-mip textures (iOS 18.0–18.1)

### LINKED ISSUE #11984: regression: `_applyMipRange` setting `TEXTURE_MAX_LEVEL=0` triggers ANGLE Metal corruption on iOS 18.0–18.1
### PixiJS version

8.16.0, 8.17.0, 8.17.1

https://pixijs.io/filters/examples/

### Steps to reproduce

1) pixi >=8.16.0 on iOS 18.0.x or 18.1.x (Safari/WKWebView). iOS sim or real hardware doesn't matter; it's an ANGLE bug, not a hardware issue
2) Navigate to https://pixijs.io/filters/examples

### What is expected?

Filters example works.

Filters example completely frozen and does not animate. 

https://github.com/user-attachments/assets/0b7dee6f-cb47-425a-8e92-d255bf247a74

### System Info

```Shell
Any iOS device between 18.0 and 18.1 (fixed in 18.2) can repro this, including using the iOS simulator.
```

### Any additional comments?

8.16.0 introduced [`_applyMipRange()`](https://github.com/pixijs/pixijs/blob/6e59c6fd0fb031f661f4d7db99dd44f45f5e4ef1/src/rendering/renderers/gl/texture/GlTextureSystem.ts#L522) which sets `TEXTURE_MAX_LEVEL=0` on single-mip textures. This unfortunately triggers a bug in iOS 18.0–18.1's webkit: it seems that the Metal driver sometimes confuses which texture is attached to which FBO, causing textures to appear swapped, or otherwise fail in strange (but deterministic) ways, including "freezing" (see Filters example) or crashing depending on the specific order of gl calls.

As best I can determine (I am not an expert on ANGLE or Metal, and my investigation here was from careful bisection of iOS versions using the simulator and Pixi versions, then rummaging through webkit source using an LLM), the ANGLE bug seems to be related to `FramebufferMtl::syncState`, which was rewritten as part of landing WebXR in time for the Apple Vision Pro launch. From tracing the code a bit in ANGLE, when `MAX_LEVEL=0`, ANGLE creates Metal texture views with `levelCount=1` which hit the buggy code path. (The GL default of `MAX_LEVEL=1000` seems to avoid it.)

**Suggested fix**

Based on my limited understanding, there's no real downside to just skipping `_applyMipRange()` for single-mip textures, and it seems to fully work around this bug for iOS users. We're shipping this workaround below (via `pnpm patch`) in production and it resolves the issue. The PR that introduced the change already [guards `_allocateEmpty2DMipChain` with `mipLevelCount > 1`](https://github.com/pixijs/pixijs/pull/11801/changes/1a7a262dee90379a81c0ba3ae145bfa68a116dee#diff-e583155ec5dd946ff2bf3d340cb45c20d56745ddc2d5df413faed8cdf0c88d50R352) — it seems to me like the same guard should apply to [`_applyMipRange`](https://github.com/pixijs/pixijs/pull/11801/changes/1a7a262dee90379a81c0ba3ae145bfa68a116dee#diff-e583155ec5dd946ff2bf3d340cb45c20d56745ddc2d5df413faed8cdf0c88d50R410), but someone more familiar with GPU internals should weigh in here... I'm in a bit beyond my depth :) cc @GoodBoyDigital who wrote that bit :D 

```diff
diff --git a/lib/rendering/renderers/gl/texture/GlTextureSystem.mjs b/lib/rendering/renderers/gl/texture/GlTextureSystem.mjs
index ce3e4f8bbd4cba97eb06c2cdadc4e64f4c191d4f..ac5178a22c8b59548bfc8c8e8683a5873ae09237 100644
--- a/lib/rendering/renderers/gl/texture/GlTextureSystem.mjs
+++ b/lib/rendering/renderers/gl/texture/GlTextureSystem.mjs
@@ -213,7 +213,14 @@ class GlTextureSystem {
     } else {
       throw new Error("[GlTextureSystem] Unsupported texture target for empty allocation.");
     }
-    this._applyMipRange(glTexture, source);
+    // Only apply explicit mip range for multi-level textures.
+    // Single-mip textures don't need it (the driver clamps to
+    // allocated levels), and setting MAX_LEVEL=0 triggers an
+    // ANGLE Metal bug on iOS 18.0-18.1 during FBO switching.
+    // https://github.com/pixijs/pixijs/issues/11984
+    if (source.mipLevelCount > 1) {
+      this._applyMipRange(glTexture, source);
+    }
     if (source.autoGenerateMipmaps && source.mipLevelCount > 1) {
       this.onUpdateMipmaps(source, false);
     }
```

GoodBoyDigital:
### Overview
- Guard `_applyMipRange()` with `source.mipLevelCount > 1` so `TEXTURE_MAX_LEVEL=0` is never set on single-mip textures, working around an ANGLE Metal bug on iOS 18.0–18.1 that causes texture corruption and freezing when switching FBOs.

Fixes: #11984

#### Fixes
- Skip `_applyMipRange` for single-mip textures to avoid setting `TEXTURE_MAX_LEVEL=0`, which triggers a Metal driver bug in iOS 18.0–18.1's ANGLE (`FramebufferMtl::syncState` creates texture views with `levelCount=1` hitting a buggy code path). Single-mip textures don't need explicit mip range clamping since the driver clamps to allocated levels by default.

##### Pre-Merge Checklist

---

> [!NOTE]
> **Low Risk**
> Low risk, localized change to WebGL2 texture parameter setup; it only adds an early return for `mipLevelCount <= 1` to avoid a known driver bug.
> 
> **Overview**
> Avoids setting `TEXTURE_MAX_LEVEL` for single-mip textures in WebGL2.
> 
> `GlTextureSystem._applyMipRange()` now returns early when `source.mipLevelCount <= 1`, working around an iOS 18.0–18.1 ANGLE Metal issue triggered by `MAX_LEVEL=0` while leaving multi-mip/partial-mip-chain behavior unchanged.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d563a54b4f4fc13564b3272dfe0daf7c159f642. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/rendering/renderers/gl/texture/GlTextureSystem.ts
+++ b/src/rendering/renderers/gl/texture/GlTextureSystem.ts
@@ -360,12 +360,7 @@ export class GlTextureSystem implements System, CanvasGenerator
 
         // Keep the texture's mip range in sync with the declared mipLevelCount.
         // This is required in WebGL2 for FBO attachments at mipLevel > 0 when using partial mip chains.
-        // Skip for single-mip textures: setting MAX_LEVEL=0 triggers an ANGLE Metal bug
-        // on iOS 18.0–18.1 (https://github.com/pixijs/pixijs/issues/11984).
-        if (source.mipLevelCount > 1)
-        {
-            this._applyMipRange(glTexture, source);
-        }
+        this._applyMipRange(glTexture, source);
 
         if (source.autoGenerateMipmaps && source.mipLevelCount > 1)
         {
@@ -527,6 +522,10 @@ export class GlTextureSystem implements System, CanvasGenerator
     private _applyMipRange(glTexture: GlTexture, source: TextureSource): void
     {
         if (this._renderer.context.webGLVersion !== 2) return;
+
+        // Skip for single-mip textures: setting MAX_LEVEL=0 triggers an ANGLE Metal bug
+        // on iOS 18.0–18.1 (https://github.com/pixijs/pixijs/issues/11984).
+        if (source.mipLevelCount <= 1) return;
 
         const gl = this._gl as WebGL2RenderingContext;
         const maxLevel = Math.max((source.mipLevelCount | 0) - 1, 0);
--- a/src/rendering/renderers/gl/texture/GlTextureSystem.ts
+++ b/src/rendering/renderers/gl/texture/GlTextureSystem.ts
@@ -360,7 +360,12 @@ export class GlTextureSystem implements System, CanvasGenerator
 
         // Keep the texture's mip range in sync with the declared mipLevelCount.
         // This is required in WebGL2 for FBO attachments at mipLevel > 0 when using partial mip chains.
-        this._applyMipRange(glTexture, source);
+        // Skip for single-mip textures: setting MAX_LEVEL=0 triggers an ANGLE Metal bug
+        // on iOS 18.0–18.1 (https://github.com/pixijs/pixijs/issues/11984).
+        if (source.mipLevelCount > 1)
+        {
+            this._applyMipRange(glTexture, source);
+        }
 
         if (source.autoGenerateMipmaps && source.mipLevelCount > 1)
         {


Zyie:
hey @aroman are you able to test this out for us as we do not have the devices to test ourselves 

`npm i https://pkg.pr.new/pixi.js@11985`

bigtimebuddy (approved):

PR #11987, opened 2026-03-24
feat: add mask channel selection for sprite masks

Zyie:
### Overview

Sprite masks previously always combined the red and alpha channels of the mask texture. This adds a `channel` option to `setMask()` that lets users choose between `'red'` (default, preserving existing behavior) and `'alpha'` for masks defined by transparency rather than color. The `'alpha'` channel mode matches how Figma applies sprite masks, making it easier to use mask assets exported from design tools.

#### Features
- Add `channel` option to `setMask()` for selecting which mask texture channel controls visibility
  ```ts
  // Read alpha channel instead of red (matches Figma mask behavior)
  sprite.setMask({ mask: maskSprite, channel: 'alpha' });
  ```

##### Pre-Merge Checklist


bigtimebuddy (approved):
Nice one :+1:

Issue #11752, opened 2025-11-11
Self-intersecting polygon fill behavior doesn't match standard fill rules (nonzero or evenodd)


mathedu4all:
### PixiJS version

v8

### Link to minimal reproduction

https://jsfiddle.net/mzpx8cbt

## Description

When rendering self-intersecting polygons (e.g., a bowtie/hourglass shape), PixiJS v8's Graphics fills the shape inconsistently compared to standard Canvas 2D fill rules (`nonzero` and `evenodd`). The rendering appears to use a different algorithm that produces unexpected results.

## Current Behavior

PixiJS fills only some triangular regions of a self-intersecting polygon, leaving other regions unfilled. This behavior doesn't match either:
- **nonzero winding rule** (which fills the entire shape including the intersection)
- **evenodd rule** (which creates a hole at the intersection)

See comparison screenshot below.

## Expected Behavior

PixiJS should support standard fill rules for self-intersecting polygons:
1. **Default to nonzero** (like Canvas 2D): Fill all regions where winding number ≠ 0
2. **Optionally support evenodd**: Allow users to specify `fill({ color: 0xff0000, fillRule: 'evenodd' })`

### Code

```javascript
import { Application, Graphics } from 'pixi.js';

const app = new Application();
await app.init({ width: 300, height: 300, backgroundColor: 0xffffff });
document.body.appendChild(app.canvas);

// Draw a self-intersecting bowtie shape
const graphics = new Graphics();
graphics.moveTo(50, 50);
graphics.lineTo(200, 200);  // Diagonal 1
graphics.lineTo(200, 50);
graphics.lineTo(50, 200);   // Diagonal 2 (intersects diagonal 1)
graphics.closePath();
graphics.fill({ color: 0xFF6B6B });

app.stage.addChild(graphics);
```

### What is expected?

<img width="2034" height="778" alt="Image" src="https://github.com/user-attachments/assets/e635c66b-4fcd-4a77-98f9-487b5956ef63" />

**Left (PixiJS)**: Only 2 of 4 triangular regions are filled  
**Middle (Canvas 2D nonzero)**: All 4 regions filled  
**Right (Canvas 2D evenodd)**: All 4 regions filled, center is hollow  

## Root Cause Analysis

[...]

Issue #10965, opened 2024-10-04
Bug: Cubic Bezier fill is incorrect


ShukantPal:
### Current Behavior

<img width="214" alt="Screenshot 2024-10-03 at 9 22 51 PM" src="https://github.com/user-attachments/assets/fdf884c7-0c0e-461f-912a-7673bcb99cb8">

### Expected Behavior

This is the output from the browser's SVG renderer.

<img width="179" alt="Screenshot 2024-10-03 at 9 23 06 PM" src="https://github.com/user-attachments/assets/7ee82d86-21d9-461b-87ba-e04eb601e090">

### Steps to Reproduce

https://jsfiddle.net/ShukantPal/fhoepg6m/

### Environment

- **`pixi.js` version**: *8.4.1*
- **Browser & Version**: *Safari*
- **OS & Version**: *macOS Sequoia*
- https://jsfiddle.net/ShukantPal/fhoepg6m/

### Possible Solution

As I've been a bit out of touch, I'd like a core member to review my code to make sure I'm rendering the cubic bezier correctly.

I can try investigating further if confirmed.

bigtimebuddy:
This isn't a problem with Cubic Beziers, you can also reproduce it with lines:
https://jsfiddle.net/bigtimebuddy/9cw7zx5u/

![Screenshot 2024-10-04 at 7 13 08 AM](https://github.com/user-attachments/assets/ab097fe0-eceb-48d8-b2b3-11519219d78c)

Graphics doesn't render these twisted, non-normal fills the same was as SVG. Likely a quirk about how tessellation works with earcut.

Issue #7325, opened 2021-03-22
Wrong fill on polygons


Davabgaryan:
### Expected Behavior
The fill of the shape should be only inside the polygon

### Current Behavior
It fills some outside parts of the polygon and leaves the inside part transparent

What it looks like:
<img width="184" alt="Screen Shot 2021-03-22 at 15 26 42" src="https://user-images.githubusercontent.com/2950370/111983014-15282200-8b23-11eb-839a-0966082a26fc.png">

What is should look like:
<img width="156" alt="Screen Shot 2021-03-22 at 15 26 46" src="https://user-images.githubusercontent.com/2950370/111983040-1f4a2080-8b23-11eb-912a-6954006c0c23.png">

Working example:
https://www.pixiplayground.com/#/edit/C1ZDPGKV5WPWaXMR41H0g

ivanpopelyshev:
Switch `earcut` to `tess`. #6800 contains playground you need: https://www.pixiplayground.com/#/edit/RPpVUBLDu-4WMLffkvVWR

SukantPal:
@Davabgaryan Yup, `Graphics` doesn't work well with self-intersecting shapes. @pixi-essentials/svg has more libtess.js-based path tessellation implementation:-

https://github.com/SukantPal/pixi-essentials/blob/master/packages/svg/src/utils/buildPath.ts

yazbaryan:
> Switch `earcut` to `tess`. #6800 contains playground you need: https://www.pixiplayground.com/#/edit/RPpVUBLDu-4WMLffkvVWR

@ivanpopelyshev I'm getting error in `Tess2.tesselate` method, checked `Sweep.topRightRegion` and saw that `req` is null

```
Uncaught TypeError: Cannot read property 'eUp' of undefined
    at Object.Sweep.topRightRegion (tess2.js:1575)
    at Object.Sweep.checkForIntersect (tess2.js:2015)
    at Object.Sweep.walkDirtyRegions (tess2.js:2126)
    at Object.Sweep.addRightEdges (tess2.js:1761)
    at Object.Sweep.sweepEvent (tess2.js:2413)
    at Object.Sweep.computeInterior (tess2.js:2641)
    at Tesselator.tesselate (tess2.js:3405)
    at Object.Tess2.tesselate (tess2.js:57)
    at Object.triangulate (index_pixi.js:133)
    at GraphicsGeometry.processFill (graphics.js:1893)
```

avpeery:
[...]

Issue #3282, opened 2016-11-14
How to set fillRule of PIXI Graphics ?


finscn:
in HTML5 Canvas , when draw polygon , there are 2 fillRules : 

> fillRule:
> The algorithm by which to determine if a point is inside a path or outside a path.
> Possible values:
> "nonzero": The non-zero winding rule, which is the default rule.
> "evenodd": The even-odd winding rule.

How to do that in PIXI ? and Which algorithm is used in PIXI ?

finscn:
the terminology of  fillRule is 『winding rule』 ：
https://en.wikipedia.org/wiki/Nonzero-rule
https://en.wikipedia.org/wiki/Even-odd_rule

There is an article  talk about it with OpenGL :  http://www.glprogramming.com/red/chapter11.html

finscn:
Could you help me ？ @englercj  @GoodBoyDigital

englercj:
I think the tessellation that article describes is using GLU, which we don't have in WebGL.

finscn:
Thanks.

bobby:
Having the ability to control the Winding Rule would really help when integrating with backend 2d libraries like java.awt.geom.

1j01:
In case this is useful to anyone, I have some working code (not using PIXI but using WebGL) which draws polygons with an even-odd winding rule: see https://github.com/pixijs/pixi.js/issues/891#issuecomment-473567390

It uses [libtess.js](https://github.com/brendankenny/libtess.js/) to convert the polygon to a list of triangles that make up the filled regions.

PR #12000, opened 2026-03-28
feat: Replace earcut with tess2 for correct winding rules and self-intersection support

gmile:
## Summary

Replaces earcut with [tess2](https://github.com/memononen/tess2.js) (GLU tessellator) for polygon triangulation. earcut does not support winding rules or self-intersecting contours — a [known limitation since 2016](https://github.com/pixijs/pixijs/issues/3282). tess2 handles both correctly.

This fixes font glyph rendering (e.g. `%` rendered with one circle filled solid instead of hollow — [reproduction](https://github.com/pixijs/pixijs/issues/11752#issuecomment-4147793344)), self-intersecting SVG paths, and multi-contour winding-based holes.

**What changed (net +230 lines, -143 lines):**

- `triangulateWithHoles.ts` — earcut → tess2, with optional `windingRule` parameter (defaults to `WINDING_NONZERO`)
- `buildContextBatches.ts` — batches all multi-contour fill paths into a single tess2 call, selecting `WINDING_ODD` vs `WINDING_NONZERO` based on `shapePath.signed`
- `SVGParser.ts` — removed ~70 lines of area-based heuristic hole detection (`extractSubpaths`, `calculatePathArea`, `checkForNestedPattern`, fill+cut simulation). tess2 handles both fill rules natively.
- `ShapePath.ts` — removed `containsPolygon` hole detection heuristic (same reason)
- `tess2.d.ts` — TypeScript declarations for the tess2 module
- `jest.config.js` — added tess2 to `transformIgnorePatterns`
- `SVGParser.test.ts` — updated `checkForHoles` expectation for evenodd paths

earcut remains as a public API re-export (`import { earcut } from 'pixi.js'`) for backwards compatibility but is no longer used internally.

**5 failing visual tests added (verified broken with earcut, passing with tess2):**

| Test | Shape | Issue |
|------|-------|-------|
| Bowtie polygon | Self-intersecting quad, edges cross | #11752 |
| Twisted quadrilateral | Crossing edges, Z-shape | #10965 |
| Cubic bezier loop | Self-intersecting curve | #10965 |
| Pentagram | 5 self-intersections | #7325 |
| Figure-8 curve | Two-lobe self-intersecting bezier | — |

Plus 1 regression test (opposite-wound nested rectangles — previously handled by `containsPolygon` heuristic, now handled natively by tess2).

**Trade-offs:**

- tess2 adds ~24KB gzipped to the bundle (vs earcut's ~6KB). This is the cost of correct winding rule support.
- tess2 is slower than earcut for simple convex polygons. For the shapes where earcut produces *wrong* output, this is acceptable.
- tess2 has [known precision issues](https://github.com/memononen/libtess2/issues/33) with very long degenerate contours — uncommon in typical SVG/Graphics usage. A future improvement could use [iTriangle](https://github.com/iShape-Rust/iTriangle) (Rust/WASM, integer math) for better robustness.
- Single-path evenodd (e.g. evenodd pentagram with hollow center) is not yet fixed — requires threading `windingRule` through the `ShapeBuildCommand` interface. Tracked separately.

## Test plan

- [x] Visual tests: 5 failing SVG tests added, all pass after fix
- [x] Regression test: opposite-wound contour hole detection preserved
- [x] Build passes (`npm run build`)
- [x] ESLint passes (lint-staged on every commit)
- [x] Existing `SVGParser.test.ts` updated and passing
- [ ] Full visual test suite (`npm run test:visual`) — needs snapshot regeneration

### DIFF
--- a/.configs/jest.config.js
+++ b/.configs/jest.config.js
@@ -16,7 +16,7 @@ module.exports = {
         '\\.wgsl$': 'jest-raw-loader',
         '\\.js$': ['babel-jest', { plugins: ['@babel/plugin-transform-modules-commonjs'] }]
     },
-    transformIgnorePatterns: ['/node_modules/(?!earcut|@types/earcut)'],
+    transformIgnorePatterns: ['/node_modules/(?!earcut|@types/earcut|tess2)'],
     moduleNameMapper: {
         '^worker:(.*)$': '$1',
         '^~/(.*)$': '<rootDir>/src/$1',
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "gifuct-js": "^2.1.2",
         "ismobilejs": "^1.1.1",
         "parse-svg-path": "^0.1.2",
+        "tess2": "^1.0.0",
         "tiny-lru": "^11.4.7"
       },
       "devDependencies": {
@@ -17378,6 +17379,12 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
+    "node_modules/tess2": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/tess2/-/tess2-1.0.0.tgz",
+      "integrity": "sha512-iSWBSOUoPn3cCT26L5Wi6mvVgL11RV4kReSnVIIPdMN7qNpkL5SLKen5BJcWj+ZTN7kK6JrHBdqTV7vvL8g+9w==",
+      "license": "SGI-B-2.0"
+    },
     "node_modules/test-exclude": {
       "version": "6.0.0",
       "resolved": "https://registry.npmjs.org/test-exclude/-/test-exclude-6.0.0.tgz",
@@ -31479,6 +31486,11 @@
         "ansi-escapes": "^4.2.1",
         "supports-hyperlinks": "^2.0.0"
       }
+    },
+    "tess2": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/tess2/-/tess2-1.0.0.tgz",
+      "integrity": "sha512-iSWBSOUoPn3cCT26L5Wi6mvVgL11RV4kReSnVIIPdMN7qNpkL5SLKen5BJcWj+ZTN7kK6JrHBdqTV7vvL8g+9w=="
     },
     "test-exclude": {
       "version": "6.0.0",
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "gifuct-js": "^2.1.2",
     "ismobilejs": "^1.1.1",
     "parse-svg-path": "^0.1.2",
+    "tess2": "^1.0.0",
     "tiny-lru": "^11.4.7"
   },
   "devDependencies": {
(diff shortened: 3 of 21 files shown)


gmile:
### Future tessellation improvements

For anyone looking at this area in the future — the tessellation landscape has evolved significantly. Some pointers:

- **[iTriangle](https://github.com/iShape-Rust/iTriangle)** (Rust/WASM) — integer-based tessellator that avoids the floating-point precision issues tess2 has with long degenerate contours. Handles self-intersections and winding rules natively. 3-13x faster than earcut at scale. Has [JS/WASM bindings](https://github.com/iShape-Rust/iShape-js).
- **[Vello](https://github.com/linebender/vello)** — GPU compute-centric 2D renderer that bypasses CPU triangulation entirely via prefix-sum tiling. Supports nonzero/evenodd natively. Relevant if PixiJS ever moves tessellation to the WebGPU compute pipeline.
- **Shpagin & Tereschenko (2026)** — "[A Simple, Practical, and Fast Polygon Triangulation in O(n+k log k) Time](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=6146463)" — input-sensitive algorithm that's dramatically faster on nearly-convex polygons.

tess2 is the right choice for now (pure JS, no WASM build step, known quantity), but iTriangle would be a strong candidate for a future opt-in via the `ShapeBuildCommand` extension point.

gmile:
Hi, all! Word of warning: this PR is a result of me using a coding agent to go into a rabbit hole after I encountered a bump with SVG-based text rendering here: https://github.com/pixijs/pixijs/issues/11752#issuecomment-4147793344.

After checking out the tess2 library, I looked around to see if there isn't more issues that this could fix, and again with a help of the coding agent - linked what I could find.

I acknowledge that reviewing this must be the next bottleneck now, so certainly no rush on my end. Open to suggestions with how to best move forward with this PR, if any.

bigtimebuddy:
Thank you for putting this together.

For context, pixi has generally taken the posture of speed first, higher fidelity as opt-in. An example of this is adaptive curves for graphics.

Because of the perf implications of this switch, it would be good to better understand the relative trade-offs between earcut and tess2. Is there a comparison to compare the two with maybe some real world examples?

gmile:
@bigtimebuddy thank you for checking the PR. I'll look into this closer in the coming days, will gather and post perf data to see the trade-offs of replacing the library.

For now, it's clear that sadly tess2 is substantially bigger in size when compared to earcut - though that appears to be the cost of correctness.

Anyway, I'll get back to you once I have more.

bigtimebuddy:
After doing some research on this, the size and perf tradeoffs might be a non-starter for Pixi. Another approach we could take is allowing users to bring their own tessellation library of choice. Tess2 could be an easy drop-in replacement for earcut for those requiring a higher degree of accuracy.

gmile:
Sounds sensible. I'll close this PR so it doesn't linger open for no good reason.

Would you have any pointers where to start to make the triangulator implementation pluggable? Or is that something you rather look into yourself?

bigtimebuddy:
We have a few paths for extensibility (application plugins, systems for renderers, mixins, DOM adapter), but I'm not sure what would be the best here. Whatever it is, it should be a one or two line code change to make that happen. @Zyie WDYT?

PR #12002, opened 2026-03-29
feat: add SmoothGraphics with shader-based HHAA anti-aliasing

Zyie:
### Overview

Adds `SmoothGraphics`, a drop-in alternative to `Graphics` that uses Half-plane Half-Axis Anti-Aliasing (HHAA) in the fragment shader for analytical per-pixel coverage. This produces smooth edges without requiring MSAA, making it ideal for high-quality vector rendering at any scale.

The existing graphics pipeline is refactored into abstract base classes so both `Graphics` and `SmoothGraphics` share ~90% of pipe/system/adaptor logic. `BatcherPipe` gains multi-batcher support so smooth and standard graphics can coexist in the same render group.

#### Features
- Add `SmoothGraphics` class with shader-based anti-aliased rendering (HHAA) for circles, lines, fills, and polygons
- Add `DashLineShader` for smooth dashed line rendering via `graphicsContext.customShader`
- Add `SmoothBatcher` with 14-float-per-vertex format supporting line width, alignment, and scale mode packing
- Support all cap types (butt, square, round), join types (bevel, miter, round), and stroke alignment
- `BatcherPipe` now supports multiple batcher types per instruction set, enabling mixed standard/smooth graphics in a single scene

##### Pre-Merge Checklist


ivanpopelyshev:
Its close but chipped, gonna check whats wrong with input to shaders there. something with joints

изображение

ivanpopelyshev:
General plan to prepare this thing for real production:

- [ ] check webgpu
- [ ] check scaleMode (transform-independent-width) in canvas
- [ ] remove extra pixelLine params from both versions of shader, not needed main lib, it was experimental

the main here is that buffer is too big those can be serious pain, 13 flots per vertex, basically 52 floats per one segment is too much. With v8 batching it eats even more resources, because it multiplies everything on a matrices before upload.

better to implement before it gets into people's projects:
- [ ] quad/stroke instancing (max 9 vertices per stroke? check it)

nebsar:
May I ask when this PR will be ready?

NIKENIT:
Any news?

PR #12009, opened 2026-04-03
fix: avoid play/mediaReady recursion before video dimensions (#11133)

### LINKED ISSUE #11133: Bug: Calling `.play` on a video source when video is not ready leads to call stack exceeded error
### Current Behavior

If you call `.play` immediately on a video resource after setting its srcObject param, the `_onPlayStart()` and `_mediaReady()` methods inside of `VideoSource` will recursively call one another since the video doesn't have a width or height for the initial frame, resulting in a stack overflow.

### Expected Behavior

Calling `.play()` should either result in the video playing when ready, or giving an error that it isn't loaded yet.

### Steps to Reproduce

https://www.pixiplayground.com/#/edit/Xax6J7RUGciVERwNuOsD_

Go the the playground example. Two webcam streams are created, and clicking on the bunny will switch the srcObject between the two streams. Observe a stack overflow when the stream switch occurs. Placing a logpoint inside of the infinite loop can actually prevent the stack overflow (video thread completes if the JS thread takes too long, maybe?).

### Environment

- **`pixi.js` version**: *8.6.2*
- **Browser & Version**: *131.0.6778.86 (Official Build) (arm64)*
- **Running Example**: https://www.pixiplayground.com/#/edit/Xax6J7RUGciVERwNuOsD_

### Possible Solution

I believe one of the two methods is incorrectly assuming that the video _must_ be ready if the `_onPlayStart()` callback is called, leading to the recursion. Another load check may need to be added.

### Additional Information

_No response_

satoren:
##### Description of change

fix #11133 

- Remove _mediaReady() from _onPlayStart when dimensions are missing; it caused unbounded recursion with _mediaReady calling _onPlayStart again.
- Resolve load() only once videoWidth/videoHeight are known.
- Listen for loadedmetadata when not yet valid so MediaStream/srcObject can resize after intrinsic size is reported.
- Fix canplaythrough listener removal (was removing wrong handler).
- Add regression test for the play-before-dimensions case.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/texture/__tests__/VideoSource.test.ts
+++ b/src/rendering/renderers/shared/texture/__tests__/VideoSource.test.ts
@@ -137,5 +137,33 @@ describe('VideoSource', () =>
         expect(canPlaySpy).not.toHaveBeenCalled();
         expect(canPlayThroughSpy).not.toHaveBeenCalled();
     });
+
+    it('should not recurse between _onPlayStart and _mediaReady when playing before dimensions (11133)', () =>
+    {
+        const video = document.createElement('video');
+
+        Object.defineProperty(video, 'videoWidth', { value: 0, configurable: true });
+        Object.defineProperty(video, 'videoHeight', { value: 0, configurable: true });
+        Object.defineProperty(video, 'paused', { value: false, configurable: true });
+        Object.defineProperty(video, 'ended', { value: false, configurable: true });
+
+        const source = new VideoSource({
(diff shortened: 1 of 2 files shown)


satoren:
I was able to verify the code fix for the issue at the following URL.
https://stackblitz.com/edit/c8e5kcab?file=src%2Findex.js

bigtimebuddy (approved):

PR #12012, opened 2026-04-08
fix: guard against null entries and missing render pipes

tiagokeller:
##### Description of change
Adds null guards to three render group code paths that crash when entries in `childrenRenderablesToUpdate.list` are null or when `renderPipes[renderPipeId]` is undefined:                                                    
                                                                                            
- **`RenderGroup.updateRenderable`** — extract pipe lookup, early return if missing                            
- **`RenderGroupSystem._updateRenderables`** — skip null entries in the update list
- **`validateRenderables`** — skip null entries, force full instruction rebuild if pipe is missing (safe       
 recovery instead of crash)                                                                                     

## Reproduction                                                                                                
These are latent on desktop Chrome but crash on Android WebView (tested on Pixel 4a / Adreno 618 with Chrome 145, and Samsung S23 / Adreno 740). Different GC and timing behavior on mobile surfaces null entries and missing pipe states that desktop never hits.                                                                                                                                                                      

Errors observed:
- `TypeError: Cannot read properties of undefined (reading 'updateRenderable')`
- `TypeError: Cannot read properties of undefined (reading 'validateRenderable')`                              

Both originate from unguarded `renderPipes[renderable.renderPipeId]` access. The null list entries cause `clearList()` to break on the first null, leaving stale entries in the array tail that compound the problem on subsequent frames.                                                                                             

## Approach
Minimal defensive guards — no behavioral changes for the happy path. When a pipe is missing in  `validateRenderables`, the fix forces `rebuildRequired = true` to trigger a full instruction rebuild via  `_buildInstructions`, which safely re-syncs the render group state.                   

##### Pre-Merge Checklist

### DIFF
--- a/src/scene/container/RenderGroup.ts
+++ b/src/scene/container/RenderGroup.ts
@@ -333,7 +333,12 @@ export class RenderGroup implements Instruction
     public updateRenderable(renderable: ViewContainer)
     {
         if (renderable.globalDisplayStatus < 0b111) return;
-        this.instructionSet.renderPipes[renderable.renderPipeId].updateRenderable(renderable);
+
+        const pipe = this.instructionSet.renderPipes[renderable.renderPipeId];
+
+        if (!pipe) return;
+
+        pipe.updateRenderable(renderable);
         renderable.didViewUpdate = false;
     }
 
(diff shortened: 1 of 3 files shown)


Zyie:
hey @tiagokeller do you have an example of this we can test?

Zyie:
closing down for now as we have no example to test, feel free to reopen when you are able to provide us an example!

PR #12013, opened 2026-04-09
fix: add null guard for textureSource in getAdjustedBlendModeBlend

Sonali-Sharma-tech:
## Summary

`getAdjustedBlendModeBlend` crashes when `textureSource` is `null`. This happens when `Assets.unload()` destroys a texture's `_source` while the render loop is still running — `Batcher.break()` passes the nulled `texture._source` to `getAdjustedBlendModeBlend`, which accesses `.alphaMode` without a null check.

**Error:**
```
TypeError: Cannot read properties of null (reading 'alphaMode')
```

**Reproduction scenario:**
1. Load a texture and add a sprite to the stage
2. Call `Assets.unload()` on that texture **without** removing the sprite first
3. The next render frame crashes in `getAdjustedBlendModeBlend`

This race condition is especially common on **memory-constrained devices** (e.g., set-top boxes, smart TVs) where textures are aggressively unloaded to stay within memory budgets, while the PixiJS ticker continues rendering.

## Fix

Added a null guard at the top of `getAdjustedBlendModeBlend` — if `textureSource` is null, return the original `blendMode` unchanged. This is safe because null source means the texture is already destroyed, so no alpha-mode adjustment is needed.

## Test plan

- Added unit tests for `getAdjustedBlendModeBlend` covering:
  - `textureSource` is `null` → returns `blendMode` without crashing
  - `textureSource.alphaMode` is `'no-premultiply-alpha'` → returns NPM blend mode
  - `textureSource.alphaMode` is premultiplied → returns original `blendMode`


Zyie:
hey @Sonali-Sharma-tech do you have an example of this bug we can test?

Zyie:
closing down for now as we have no example to test, feel free to reopen when you are able to provide us an example!

PR #12020, opened 2026-04-20
feat: expose WebGPU canvas and pipeline color formats for HDR output

### LINKED ISSUE #12019: Feature Request: HDR canvas output for WebGPU renderer
Proposal: add opt-in HDR canvas output to the WebGPU renderer, using `GPUCanvasConfiguration.toneMapping` with `rgba16float` canvas format (Chrome 131+, Safari 18.x+). This lets fragment shaders output colors beyond 1.0 for genuine HDR luminance on capable displays. SDR displays receive a tone-mapped/clamped fallback.

## Current blocker

Three call sites hardcode `bgra8unorm` with no public override, and all three must agree or WebGPU throws a pipeline/target format mismatch:

1. `GpuRenderTargetAdaptor.initGpuRenderTarget`: the canvas context configure call, `context.configure({ format: 'bgra8unorm', ... })`
2. `GpuStateSystem.getColorTargets`: the pipeline descriptor's `GPUColorTargetState.format`
3. `TextureSource.defaultOptions.format`: default format for offscreen render targets (filter scratch buffers, etc.)

## Proposed API

```ts
await renderer.init({
    hdr: true, // opt-in; default false preserves current behavior
});
```

When enabled:

- canvas context is configured with `rgba16float` plus `toneMapping: { mode: 'extended' }`
- pipeline color target format becomes `rgba16float`
- render-target texture sources default to `rgba16float`

Only render targets change format. Sampled image textures and spritesheets stay in `rgba8unorm`/`bgra8unorm`, so there is no impact on asset pipelines or VRAM cost for sprite data.

Laserwolve:
##### Description of change

Closes #12019.

Three call sites in the WebGPU renderer hardcode `'bgra8unorm'`:

1. `GpuRenderTargetAdaptor.initGpuRenderTarget` (canvas context format)
2. `GpuStateSystem.getColorTargets` (pipeline color target format)
3. `TextureSource.defaultOptions.format` (already publicly mutable)

All three must agree or WebGPU throws a pipeline/target format mismatch, which prevents users from opting into `rgba16float` canvas output required for HDR presentation.

This PR exposes the two non-mutable formats plus an optional canvas tone mapping as public static properties, so a user can opt in before `renderer.init()`:

```ts
import {
    GpuRenderTargetAdaptor,
    GpuStateSystem,
    TextureSource,
} from 'pixi.js';

GpuStateSystem.colorTargetFormat = 'rgba16float';
GpuRenderTargetAdaptor.canvasFormat = 'rgba16float';
GpuRenderTargetAdaptor.canvasToneMapping = { mode: 'extended' };
TextureSource.defaultOptions.format = 'rgba16float';

await app.init({ /* ... */ });
```

Defaults are unchanged, so existing projects see no behavior difference.

This is intentionally minimal. If a higher-level `hdr: true` option on `WebGPUOptions` is preferred, happy to iterate toward that shape in follow-up commits. Wanted to get the lowest-friction escape hatch landed first so the three format values stop needing prototype-level monkeypatching in userland.

##### Pre-Merge Checklist

Static checks (lint, types, index, prune) pass locally. Jest unit tests run inside Electron via `@pixi/jest-electron` and could not be executed in the contributor's headless environment; relying on CI to validate the two added test files.

### DIFF
--- a/src/rendering/renderers/gpu/renderTarget/GpuRenderTargetAdaptor.ts
+++ b/src/rendering/renderers/gpu/renderTarget/GpuRenderTargetAdaptor.ts
@@ -19,6 +19,23 @@ import type { WebGPURenderer } from '../WebGPURenderer';
  */
 export class GpuRenderTargetAdaptor implements RenderTargetAdaptor<GpuRenderTarget>
 {
+    /**
+     * The pixel format used when configuring the WebGPU canvas context. Must
+     * match {@link GpuStateSystem.colorTargetFormat} and any render-target
+     * texture formats. Override before renderer initialization to enable HDR
+     * output (e.g. `'rgba16float'`).
+     * @default 'bgra8unorm'
+     */
+    public static canvasFormat: GPUTextureFormat = 'bgra8unorm';
+
+    /**
+     * Optional tone mapping applied when configuring the WebGPU canvas
+     * context. Set to `{ mode: 'extended' }` to enable HDR output on capable
+     * displays (Chrome 131+, Safari 18.x+). Non-HDR displays receive a
+     * clamped/tone-mapped SDR presentation.
+     */
+    public static canvasToneMapping: GPUCanvasToneMapping | undefined;
+
     private _renderTargetSystem: RenderTargetSystem<GpuRenderTarget>;
     private _renderer: WebGPURenderer<HTMLCanvasElement>;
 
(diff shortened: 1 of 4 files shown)


Laserwolve:
Hey @GoodBoyDigital - tagging you since this seems most relevant to your work. Given your recent work on the WebGPU render-target side (#12050, #12066), you seemed like the right person to ask: any thoughts on the approach here?

Quick recap: this exposes the three hardcoded 'bgra8unorm' format values (canvas context, pipeline color targets, default texture format) as public statics so users can opt into rgba16float for HDR output before renderer.init(). Defaults are untouched, so there's no behavior change for existing projects.

I kept it intentionally minimal as an escape hatch, but I'm happy to rework it into a higher-level hdr: true option on WebGPUOptions if that's the shape you'd prefer.

PR #12030, opened 2026-04-27
fix: BatcherPipe leak on RenderGroup destroy

### LINKED ISSUE #11409: Bug: Possible memory leak when render container to a RenderTexture
### Current Behavior

Attempt 1: Use `app.renderer.extract.base64` on a container.
Attempt 2: Render the container on a `RenderTexture`.

In both cases, one `_defaultBatcher` is leaked on each call according to the memory profile.

### Expected Behavior

No memory leak.

### Steps to Reproduce

Take two memory snapshots with ~5 seconds delay and check the objects allocated between them. There will be ~300 `_defaultBatcher`s.

### Environment

- **`pixi.js` version**: 8.9.2
- **Browser & Version**: Chrome 135
- **OS & Version**: MacOS 15.1.1
- **Running Example**: https://www.pixiplayground.com/#/edit/8Xul5b2kJcCJlR5eBQyY7

### Possible Solution

_No response_

### Additional Information

_No response_

laino:
##### Description of change
BatcherPipe._batchersByInstructionSet cached batchers per InstructionSet but never released them, leaking GPU buffers, geometry, and texture refs after Container destroy.

Add an InstructionPipe.destroyInstructionSet hook fired from InstructionSet.destroy(), called by RenderGroup.destroy(). BatcherPipe implements it to destroy cached batchers and drop the entry. Pooled RenderGroup.reset() is left alone so pool reuse keeps the cache.

Fixes: https://github.com/pixijs/pixijs/issues/11409

##### Pre-Merge Checklist


laino:
I'd love a hotfix for this, but for a memory leak it's maybe not necessary.

laino:
Actually I just realized that renderPipes are ``any`` typed there. ~~Makes me think this should be solved another way.~~

They are used as ``any`` in many places. Not sure who is really supposed to own that value.

GoodBoyDigital (approved):

PR #12039, opened 2026-05-06
fix: stabilize HTMLText visual tests against CI flakiness

Zyie:
### Overview

The HTMLText visual scenes relied on a fixed `setTimeout(350ms)` after `renderer.render()` to wait for async texture generation (font fetch + SVG image load). On slower CI runs the timeout could expire before the texture promise resolved, causing `extract.canvas` to read `Texture.EMPTY` and producing large pixel diffs that randomly failed the suite (e.g. `webgpu-html-text-whitespace-tagged` with a 37851-pixel diff). The tester now walks the stage and awaits every in-flight HTMLText `texturePromise` before extraction, removing the race entirely. The trailing `setTimeout` waits in 37 scenes are no longer needed and have been deleted; the four dynamic scenes keep their intermediate wait since it sequences a property change between two renders, which the tester-level wait can't replace.

#### Fixes
- HTMLText visual tests no longer flake on CI when font/SVG loads exceed the 350ms wait

##### Pre-Merge Checklist

### DIFF
--- a/playground/src/scenes/sceneRunner.ts
+++ b/playground/src/scenes/sceneRunner.ts
@@ -4,6 +4,29 @@ import type { Renderer, RendererOptions } from 'pixi.js';
 import type { TestScene } from '../../../tests/visual/types';
 import type { RenderType } from '../types';
 
+async function waitForPendingHTMLText(container: Container, renderer: Renderer): Promise<void>
+{
+    const promises: Promise<unknown>[] = [];
+
+    const visit = (c: Container): void =>
+    {
+        if (c.renderPipeId === 'htmlText')
+        {
+            const gpuData = (c as unknown as {
+                _gpuData: Record<number, { texturePromise?: Promise<unknown> }>,
+            })._gpuData[renderer.uid];
+
+            if (gpuData?.texturePromise) promises.push(gpuData.texturePromise.catch((): undefined => undefined));
+        }
+
+        for (let i = 0; i < c.children.length; i++) visit(c.children[i]);
+    };
+
+    visit(container);
+
+    if (promises.length) await Promise.all(promises);
+}
+
 const defaultRendererOptions = {
     width: 128,
     height: 128,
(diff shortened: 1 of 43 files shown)


playground/src/scenes/sceneRunner.ts
```diff
@@ -4,6 +4,29 @@ import type { Renderer, RendererOptions } from 'pixi.js';
...
+        if (c.renderPipeId === 'htmlText')
+        {
+            const gpuData = (c as unknown as {
+                _gpuData: Record<number, { texturePromise?: Promise<unknown> }>,
+            })._gpuData[renderer.uid];
+
+            if (gpuData?.texturePromise) promises.push(gpuData.texturePromise.catch((): undefined => undefined));
+        }
+
+        for (let i = 0; i < c.children.length; i++) visit(c.children[i]);
```
bigtimebuddy:
```suggestion
        c.children.forEach(visit);
```

playground/src/scenes/sceneRunner.ts
```diff
@@ -4,6 +4,29 @@ import type { Renderer, RendererOptions } from 'pixi.js';
...
+    };
+
+    visit(container);
+
+    if (promises.length) await Promise.all(promises);
```
bigtimebuddy: If-check not needed, empty array will return immediately
```suggestion
    await Promise.all(promises);
```

tests/visual/tester.ts
```diff
@@ -8,6 +8,38 @@ import { Container, Graphics } from '~/scene';
...
+
+            if (gpuData?.texturePromise) promises.push(gpuData.texturePromise.catch((): undefined => undefined));
+        }
+
+        for (let i = 0; i < c.children.length; i++) visit(c.children[i]);
+    };
+
+    visit(container);
+
+    if (promises.length) await Promise.all(promises);
```
bigtimebuddy:
```suggestion
        c.children.forEach(visit);
    };

    visit(container);

    await Promise.all(promises);
```

GoodBoyDigital (approved):

bigtimebuddy (approved):
Nice. Small code suggestions. 

I wish Pixi had a promise-based API to await for an HTMLText ready. This can be a pain if you're static rendering, but good fix for tests.

PR #12042, opened 2026-05-07
fix: handle null from getShaderSource/getShaderInfoLog on lost context (#12032)

### LINKED ISSUE #12032: WebGL: error thrown on logPrettyShaderError when getShaderSource returns null
### PixiJS version

8.0.0 and above

https://stackblitz.com/edit/pixijs-v8-rkdrxcav?file=src%2Fmain.js

### Steps to reproduce

Simply run the reproduction link and notice the error in the browser console. In Chromium: Uncaught TypeError: Cannot read properties of null (reading 'split'). Firefox: Uncaught (in promise) TypeError: can't access property "split", gl.getShaderSource(...) is null.

### What is expected?

It should gracefully handle null being returned, without throwing. As stated here for example, both getShaderSource and getShaderInfoLog return null on any GL errors: https://registry.khronos.org/webgl/specs/latest/1.0/#5.14.9. So it should preferably take that possibility into account.

It throws an error, crashing. See example:

Image

### System Info

```Shell
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 AMD Ryzen 5 5600H with Radeon Graphics         
    Memory: 2.92 GB / 15.35 GB
Browsers:
    Chrome: 147.0.7727.138
    Edge: Chromium (140.0.3485.54)
    Firefox: 150.0.1 - C:\Program Files\Mozilla Firefox\firefox.exe
```

### Any additional comments?

We worked around it, for now, by making sure WebGL context isn't lost when calling .render.

Zyie:
### Overview

Fixes #12032. When the WebGL context is lost, `gl.getShaderSource` and `gl.getShaderInfoLog` return `null` per the WebGL spec (§5.14.9), causing `logPrettyShaderError` to throw a `TypeError` on `.split` and masking the original render failure. The error path now logs a clear context-lost message and returns early instead.

#### Fixes
- Handle null returned from `getShaderSource`/`getShaderInfoLog` so a lost WebGL context no longer crashes shader-error logging during `render()`.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/gl/shader/program/__tests__/logProgramError.test.ts
+++ b/src/rendering/renderers/gl/shader/program/__tests__/logProgramError.test.ts
@@ -1,0 +1,47 @@
+import { compileShader } from '../compileShader';
+import { logProgramError } from '../logProgramError';
+import { DOMAdapter } from '~/environment';
+
+describe('logProgramError', () =>
+{
+    it('should not throw when WebGL context is lost and shader sources are unavailable', () =>
+    {
+        const canvas = DOMAdapter.get().createCanvas();
+        const gl = canvas.getContext('webgl') as WebGLRenderingContext;
+
+        const vertexShader = compileShader(gl, gl.VERTEX_SHADER, `
+            attribute vec4 aPosition;
+            void main() { gl_Position = aPosition; }
+        `);
+        const fragmentShader = compileShader(gl, gl.FRAGMENT_SHADER, `
+            precision mediump float;
+            void main() { gl_FragColor = vec4(1.0); }
+        `);
+        const program = gl.createProgram();
+
+        gl.attachShader(program, vertexShader);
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):
LGTM. The null-guarding on `getShaderSource`/`getShaderInfoLog` is the right call — a lost WebGL context during shader compilation is itself a cause of compile/link failure, which routes straight into this error path while making the source/log unretrievable. The old `null.split('\n')` masked the real context-loss error with a TypeError in the handler.

Pushed two small tweaks:
- dropped the unreachable `if (shaderLog) console.error(shaderLog)` branch — `getShaderInfoLog` is always null/empty whenever `getShaderSource` is null (same GL failure)
- moved the `shaderLog` declaration back to its original position to keep the diff minimal; `?? ''` guard retained

Verified locally: lint, tsc build, and the new `logProgramError` unit test all pass.

PR #12050, opened 2026-05-14
feat: implement transientAttachment for MSAA RenderTextures for WebGPU

GoodBoyDigital:
### Overview

WebGPU foundation for transient MSAA attachments. Today every MSAA RenderTexture's multisample buffer is allocated with the full `TEXTURE_BINDING | COPY_DST | RENDER_ATTACHMENT | COPY_SRC` bundle and stored out to DRAM at end-of-pass. This PR exposes the API surface to mark MSAA buffers as transient — letting the WebGPU backend skip the post-pass writeback and, on supporting browsers, add `GPUTextureUsage.TRANSIENT_ATTACHMENT` so TBDR mobile drivers can keep contents in tile memory entirely.

Pixi does not auto-set `transient` anywhere — it's pure opt-in. Default behaviour is unchanged for every user who doesn't touch the new flag. Follow-up PRs will add (a) plumbing to set transient on the canvas root via renderer options, and (b) an MSAA reload-from-resolve trick that lets pixi auto-set transient even on internal MSAA buffers without breaking filter pop-back paths.

#### Features
- New `TextureSourceOptions.transient` option + matching `TextureSource.transient` field. When true on an MSAA `TextureSource`, the WebGPU backend uses `storeOp: 'discard'` on the attachment at end-of-pass and OR's `GPUTextureUsage.TRANSIENT_ATTACHMENT` into the texture's usage when the browser exposes it.

```ts
const rt = RenderTexture.create({
    width: 512, height: 512,
    antialias: true,
    transient: true,    // single-pass render, contents won't be loaded back
});
```

- New `renderer.device.extensions: GpuExtensions` struct, mirroring `renderer.context.extensions` on WebGL. First entry: `transientAttachment: boolean`, true when `GPUTextureUsage.TRANSIENT_ATTACHMENT` is exposed by the browser at runtime.

```ts
if (renderer.device.extensions.transientAttachment)
{
    // memoryless MSAA path is available on this device
}
```

#### Chores
- `GpuTextureSystem._initSource` narrows MSAA texture allocations (`sampleCount > 1`) to `RENDER_ATTACHMENT` only. The bundled `TEXTURE_BINDING | COPY_DST | COPY_SRC` bits are dead on MSAA buffers (never sampled, never uploaded, never copied) and are required to be absent when `TRANSIENT_ATTACHMENT` is set. Single-sample path is byte-identical to before.
- `.context` added to eslint ignores so local-only benchmark/scratch files don't trip the linter.

##### Pre-Merge Checklist

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -13,6 +13,7 @@ export default tseslint.config(
     {
         ignores: [
             '.s3_uploads',
+            '.context',
             'out',
             'docs',
             'dist',
--- a/src/rendering/index.ts
+++ b/src/rendering/index.ts
@@ -127,6 +127,7 @@ export * from './renderers/gpu/buffer/UboBatch';
 export * from './renderers/gpu/GpuColorMaskSystem';
 export * from './renderers/gpu/GpuDeviceSystem';
 export * from './renderers/gpu/GpuEncoderSystem';
+export * from './renderers/gpu/GpuExtensions';
 export * from './renderers/gpu/GpuLimitsSystem';
 export * from './renderers/gpu/GpuStencilSystem';
 export * from './renderers/gpu/GpuUboSystem';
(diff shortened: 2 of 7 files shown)


bigtimebuddy (approved):
Nice!

PR #12051, opened 2026-05-16
fix: canvas renderer does not round anchor offset when roundPixels is enabled

Clonex:
##### Description of change

The v8 rewrite introduced `CanvasBatchAdaptor` as a unified canvas batch renderer. `setContextTransform` is called with `roundPixels`, which rounds only `mat.tx` and `mat.ty` (the world translation) - but the anchor offset (`bounds.minX/Y`) is passed to `drawImage` unrounded, so sprites with a non-zero anchor still render at sub-pixel positions.

This PR reintroduces the [explicit `dx | 0` rounding from v7](https://github.com/pixijs/pixijs/blob/v7.x/packages/canvas-sprite/src/CanvasSpriteRenderer.ts#L138) for the `drawX`/`drawY` values passed to `drawImage`.

##### Pre-Merge Checklist


Clonex:
@Zyie I am not sure why the test is failing, it is failing even when I revert the my changes and re-run them 🤔

Zyie (approved):

PR #12053, opened 2026-05-19
feat: add HTML-in-Canvas texture support

Zyie:
### Overview

Adds the `HTMLSource` texture source family, backed by the experimental HTML-in-Canvas browser API. It renders live DOM elements (`HTMLSource`) or immutable `captureElementImage()` snapshots (`ElementImageSource`) into PixiJS textures while the element stays interactive in the browser. Includes WebGL and WebGPU upload paths; the uploaders throw a clear error if the browser API is not enabled.

Ships as an opt-in subpath (`pixi.js/html-source`) so apps that do not need DOM rendering are not paying for it in their main bundle.

#### Breaking Changes / Behavior Changes
- `Texture.from` / `TextureSource.from` resolve a generic HTML element to an `HTMLSource` (and an `ElementImage` snapshot to an `ElementImageSource`) as a lowest-priority fallback. Previously this threw `Could not find a source type`. Requires `import 'pixi.js/html-source'` (or any import from the subpath) to register the sources.
```ts
import 'pixi.js/html-source';

// Previously threw; now returns a Texture backed by HTMLSource
const texture = Texture.from(document.querySelector('#card'));
```

#### Features
- New `HTMLSource` texture source for live DOM elements. The element stays interactive (forms editable, links clickable, CSS animations running) while mirrored to the GPU.
```ts
import { HTMLSource } from 'pixi.js/html-source';

const source = new HTMLSource({ resource: formElement });
const sprite = Sprite.from(source);
```
- New `ElementImageSource` texture source for immutable `ElementImage` snapshots produced by `captureElementImage()`. Ready on construction, never repaints; optional `autoClose` releases the snapshot when the source is destroyed.
```ts
import { ElementImageSource } from 'pixi.js/html-source';
import type { HTMLSourceCanvas } from 'pixi.js/html-source';

const canvas = app.canvas as HTMLSourceCanvas;
const snapshot = canvas.captureElementImage!(element);
const source = new ElementImageSource({ resource: snapshot, autoClose: true });
```
- Automatic owning-canvas inference, `layoutsubtree` setup, `paint`-event auto-update, and an initial paint request — all configurable via `HTMLSourceOptions` (`canvas`, `autoLayout`, `autoUpdate`, `autoRequestPaint`).
- WebGL and WebGPU upload implementations, including premultiplied alpha handling on GPU. A shared `HTMLUploadableSource` contract lets the single `'html'` uploader pair serve both source classes.

#### Package Structure
- `HTMLSource`, `ElementImageSource`, their types, and both GL/GPU uploaders live in `src/html-source/`, mirroring the `src/gif/` layout.
- Dedicated bundle (`dist/packages/html-source.{js,mjs}`) and subpath export (`pixi.js/html-source`).
- `src/html-source/init.ts` registers `HTMLSource`, `ElementImageSource`, and both uploaders via `extensions.add`; the subpath re-exports everything from `index.ts` so a single import covers both registration and the named exports.
- Uploaders register through two new extension types — `ExtensionType.TextureUploaderWebGL` and `ExtensionType.TextureUploaderWebGPU` — consumed by `GlTextureSystem` and `GpuTextureSystem` via `extensions.handleByMap`, so the core renderer stays decoupled from the html-source types and the uploaders remain tree-shakeable.
- Playground Vite config switched to array-form aliases with a regex entry so `pixi.js/<pkg>` subpath imports resolve to `lib/<pkg>/init.mjs` (fixes `pixi.js/gif` too).

#### Public Exports (from `pixi.js/html-source`)
`HTMLSource`, `HTMLSourceOptions`, `HTMLSourceCanvas`, `HTMLSourceResource`, `ElementImageSource`, `ElementImageSourceOptions`, `ElementImage`.

##### Pre-Merge Checklist


bigtimebuddy:
Your description is out-of-date, right? I don't see HTMLSnapshotSource anymore. 

Because HTML-in-Canvas is so bleeding-edge, I think this should probably be opt-in and not included with the core import by default. WDYT?

dagangtj:
Thank you for the review, @bigtimebuddy! You're absolutely right — the PR description is outdated and still references <code>HTMLSnapshotSource</code>, which is no longer part of this PR. I apologize for the confusion.

I completely agree that, given how bleeding-edge HTML-in-Canvas is, it should be opt-in rather than included in the core import by default. That makes a lot of sense for stability and bundle size.

I'm happy to update the PR description to reflect the current changes and adjust the implementation to make the feature opt-in. I'll push an update shortly — thanks again for the thoughtful feedback!

GoodBoyDigital (approved):

Zyie:
@bigtimebuddy description has been updated and i've moved everything to be under `pixi.js/html-source`. ITs all set up the same way as `pixi.js/gif`

bigtimebuddy (approved):

src/html-source/glUploadHTMLResource.ts
```diff
@@ -0,0 +1,104 @@
+import { ExtensionType } from '../extensions/Extensions';
+
+import type { GlRenderingContext } from '../rendering/renderers/gl/context/GlRenderingContext';
+import type { GlTexture } from '../rendering/renderers/gl/texture/GlTexture';
+import type { GLTextureUploader } from '../rendering/renderers/gl/texture/uploaders/GLTextureUploader';
+import type { HTMLSourceResource, HTMLUploadableSource } from './HTMLSourceTypes';
+
+interface GlTexElementImageContext extends GlRenderingContext
+{
+    texElementImage2D?: (
```
ewewraw, line 10: Please note that from Chromium 150+ the implementation of this method is changing, e.g.:

```
  // WebGL old:
  const level = 0;
  const internalFormat = gl.RGBA;
  const srcFormat = gl.RGBA;
  const destType = gl.UNSIGNED_BYTE;
  gl.texElementImage2D(
    gl.TEXTURE_2D,
    level,
    internalFormat,
    srcFormat,
    destType,
    element);

  // WebGL new:
  const internalFormat = gl.RGBA8;
  gl.texElementImage2D(gl.TEXTURE_2D, internalFormat, element);
```

Here's the [spec change](https://github.com/WICG/html-in-canvas/pull/128/changes). 

For more context, this change is addressing the community group feedback. If you're interested in more detail, you can check the discussion [here](https://github.com/KhronosGroup/WebGL/pull/3752).

src/html-source/gpuUploadHTMLResource.ts
```diff
@@ -0,0 +1,64 @@
+import { ExtensionType } from '../extensions/Extensions';
+
+import type { GPU } from '../rendering/renderers/gpu/GpuDeviceSystem';
+import type { GpuTextureUploader } from '../rendering/renderers/gpu/texture/uploaders/GpuTextureUploader';
+import type { HTMLSourceResource, HTMLUploadableSource } from './HTMLSourceTypes';
+
+interface GpuCopyElementImageQueue extends GPUQueue
+{
+    copyElementImageToTexture?: (
```
ewewraw, line 9: Please note that the implementation of copyElementImageToTexture is also [changing](https://github.com/WICG/html-in-canvas/pull/128/changes) from Chromium 150+:

```
  // WebGPU old:
  GPUQueue.copyElementImageToTexture(
    element,
    width,
    height,
    { texture: texture });

  // WebGPU new:
  const sourceDict = { source: element };
  const destDict = {
    destination: { texture: texture },
    width: width,
    height: height
  };
  GPUQueue.copyElementImageToTexture(sourceDict, destDict);
```

And here's the [background](https://github.com/gpuweb/gpuweb/pull/6250).

ewewraw (commented):
Hey @Zyie, here's Natalia, a DevRel from Chrome.
That's great to see PixiJS is adopting HTML-in-Canvas! Please note that from Chromium 150+, there are some changes to the API that are addressing the community group feedback.

bigtimebuddy:
@Zyie could you take a look the the feedback above?

src/html-source/glUploadHTMLResource.ts
Zyie: Hey @ewewraw, thanks for this! We have a PR up to support chrome 150 #12077

PR #12060, opened 2026-05-26
fix: bitmap font base nan

Slynchy:
##### Description of change
This change simply adds a nullish check to the number-as-a-string `base` property of a loaded bitmap font when parsing, and falls-back to "0" if necessary. This fixes an issue introduced in v8 (specifically commit de4b515f) where bitmap fonts lacking this property would just be invisible, since it would perform `(font.lineHeight - NaN)` which would set the `baseLineOffset` property to `NaN`.

Very minor change, not requiring any updated tests or documentations. Tests run and pass perfectly fine.

##### Pre-Merge Checklist


src/scene/text-bitmap/asset/bitmapFontTextParser.ts
```diff
@@ -138,7 +138,7 @@ export const bitmapFontTextParser = {
 
         const map: Record<string, string> = {};
 
-        font.baseLineOffset = font.lineHeight - parseInt(common.base, 10);
+        font.baseLineOffset = font.lineHeight - parseInt(common.base || "0", 10);
```
bigtimebuddy:
```suggestion
        font.baseLineOffset = font.lineHeight - parseInt(common.base ?? '0', 10);
```

Slynchy: Addressed in my latest commit! (didn't see that I could just do it in the PR)

bigtimebuddy:
@Slynchy can you create a simple test for this so we don't have a regression here in the future? Otherwise looks good :+1:

Slynchy:
Hmm, do you think that the test can get away with simply deleting the `base` property after loading the font XML, before proceeding with actually loading/rendering the font for the test? Would be a little weird to have a near-duplicate `desyrel` font in the assets otherwise.

At any rate yes, I'll make the test when I get a chance!

bigtimebuddy:
I'd duplicate the font fixture minus the base field, that sound be fine :+1: 

Thanks for handling this.

Zyie:
closing in favor of https://github.com/pixijs/pixijs/pull/12121

PR #12066, opened 2026-05-28
feat: usage-aware WebGPU textures (narrow transient render-pass textures)

GoodBoyDigital:
## Summary

Pixi's WebGPU backend previously gave every non-compressed texture the same broad usage flags (`TEXTURE_BINDING | COPY_DST | RENDER_ATTACHMENT | COPY_SRC`) regardless of how the texture is actually used. That prevents tile-based mobile GPUs (Apple / Mali / Adreno) from making stronger assumptions about transient render-pass textures — assumptions that unlock tile-resident / memoryless attachments, better compression paths, and reduced bandwidth/sync pressure across many passes.

This PR makes texture usage **specialisable** and narrows the internal **transient** pool textures, while keeping public `RenderTexture` / asset textures broad and flexible by default.

## What changed

**1. Configurable usage API**
- New `TextureUsage` enum (`src/.../shared/texture/const.ts`) mirroring `GPUTextureUsage`, in the same spirit as the existing `BufferUsage` enum.
- New optional `gpuUsage?: number` on `TextureSourceOptions` (+ a `gpuUsage` field on `TextureSource`). Defaults to the broad set, so **existing behaviour is unchanged**. Ignored on WebGL.
- `GpuTextureSystem` now reads `source.gpuUsage` directly instead of deriving usage from `uploadMethodId`. `CompressedSource` overrides with the valid narrow set for compressed formats (`TEXTURE_BINDING | COPY_DST`).

**2. Narrowed transient pool textures**
- `TexturePool.getOptimalTexture` / `getSameSizeTexture` / `createTexture` take a `usage` parameter, **defaulting to the narrowest valid set for transient render targets: `RENDER_ATTACHMENT | TEXTURE_BINDING`**. This covers the bulk of pool traffic — filter input/output/ping-pong, cache-as-texture, and alpha masks.
- Usage is folded into the pool key so a texture created with narrow usage is never reused where wider usage is required. Since pooled dimensions are always powers of two, the key now stores their log2 exponents (freeing bits to pack usage into the existing numeric key — no extra allocation, and it fixes a latent sign-bit overflow in the old `width << 17` packing).
- The two pool consumers that copy into their textures opt into wider usage:
  - **Text glyph atlases** (`getPo2TextureFromSource`) → `TEXTURE_BINDING | COPY_DST | RENDER_ATTACHMENT` (uploaded via `copyExternalImageToTexture`, which the spec requires to have both `COPY_DST` and `RENDER_ATTACHMENT`).
  - **Blend-filter back-textures** (`FilterSystem.getBackTexture`) → `TEXTURE_BINDING | COPY_DST` (`copyTextureToTexture` destination).

## Usage rules reference

| Operation | Consumer | Required usage |
|---|---|---|
| render → sample | filters / cache / masks | `RENDER_ATTACHMENT \| TEXTURE_BINDING` |
| `copyTextureToTexture` dest | blend back-texture, mipmap copy-back | `+ COPY_DST` |
| `copyExternalImageToTexture` | text glyph upload | `COPY_DST \| RENDER_ATTACHMENT` |

## Compatibility

- Default usage on public `TextureSource` / `RenderTexture` is unchanged (broad), so no behavioural change for existing apps.
- WebGL ignores `gpuUsage` entirely.
- Regular asset textures are intentionally **not** narrowed: their minimum is already near-broad (upload forces `COPY_DST | RENDER_ATTACHMENT`, sampling needs `TEXTURE_BINDING`), the only droppable flag is `COPY_SRC`, and dropping it would break `renderer.extract` on arbitrary textures for no tile-residency benefit (they aren't transient attachments). Advanced users can still opt in via `gpuUsage`.

## Testing

- Added `TexturePool` unit tests for usage bucketing (default usage, requested usage, no cross-usage reuse, correct return routing).
- Type-check, lint, and affected unit suites (TexturePool / FilterSystem / filters / text / masks / RenderGroup / texture sources) all green.
- Verified visually on a real WebGPU renderer (Electron debug mode): text, blur/blend filters, and cache-as-texture masks render correctly.


Huaian666:
PixiJS v8's rendering architecture is genuinely impressive - the way it abstracts WebGL/WebGPU into a unified API while maintaining blazing performance is no small feat. The WebGPU texture management work being discussed here shows the team's commitment to staying ahead of the curve. 47k stars well deserved.

By the way, if anyone's into HTML - there's an iOS app called Like Code that lets you upload and run HTML directly on your phone. If you use AI to write some HTML and it runs well, you can optionally save it to your collection, even offline. If it's a game, you can even tweak the code to make yourself invincible haha

It's on the App Store, feel free to check it out!

PR #12067, opened 2026-05-31
chore: document blend mode resolution inheritance

### LINKED ISSUE #11311: Bug: Advanced blend modes don't adjust to renderer resolution / window pixel ratio

MFei8ht:
Fixes pixijs/pixijs#11311.

## Description of change

Adds documentation for the current v8-compatible workaround for advanced blend modes on high-DPI or non-`1` renderer resolutions.

The docs now explain that advanced blend modes rely on filter rendering and that users can set:

```ts
Filter.defaultOptions.resolution = 'inherit';
```

before creating filters or display objects that use advanced blend modes.

## Why docs instead of changing runtime defaults?

Maintainer feedback in #11584 said the team agrees that `"inherit"` should become the future default for filters, but changing it in v8 is uncomfortable because it can increase memory usage and would be breaking. The requested v8 path was education/documentation.

## Testing

Docs-only change. No build run.


bigtimebuddy (approved):
Thank you, good suggestion!

MFei8ht:
Quick follow-up: I updated the PR title to `chore: document blend mode resolution inheritance` so it matches the repo's PR title regex.

No code/docs content changed after your approval. The latest title-check run now appears to be waiting for workflow approval on the fork.

PR #12081, opened 2026-06-10
fix: stop GC from breaking cached bind groups when unloading buffers

### LINKED ISSUE #12080: WebGPU: black screen after GC unloads uniform buffers referenced by cached bind groups
**Current behavior**

On the WebGPU renderer, when the GC system unloads a uniform buffer, cached `GPUBindGroup`s that reference it are never invalidated. Every subsequent `queue.submit()` that draws with one fails validation:

```
[Buffer (unlabeled)] used in submit while destroyed.
```

One stale buffer invalidates the whole command buffer, so the canvas goes black and the warning spams every frame. With default GC settings this surfaces about a minute into a session, and it never recovers unless something else happens to rebuild the bind group.

Two ways to hit it:

- A mesh or filter with a custom uniform group is not rendered for `gcMaxUnusedTime` (offscreen, `visible = false`, etc.), then shown again.
- A mesh with a `UniformGroup` created with `isStatic: true` breaks while it is visible and rendered every frame. Static groups skip the per-draw sync in `UboSystem.updateUniformGroup`, so the buffer's `_gcLastUsed` never refreshes and the GC collects it anyway.

**Why it happens**

- Bind group cache keys are built from each resource's `_resourceId`. A `UniformGroup`'s id is assigned in the constructor and never changes, so `BindGroupSystem` keeps serving the stale cached `GPUBindGroup`.
- `Buffer.unload()` destroys the `GPUBuffer` but does not bump `_resourceId` or emit `change`.
- On unload, `GpuBufferSystem` also removes the buffer's `update`/`change` listeners, so later `buffer.update()` calls do nothing and the GPU buffer is never recreated.

Worth calling out: the engine already solves this exact problem for textures. `TextureSource.unload()` bumps `_resourceId` and emits `change` so cached bind groups re-key and rebuild on next use, and there is a unit test asserting that behavior. Buffers never got the same treatment, which is why GC unloading is transparent for textures but fatal for uniform buffers.

**Reproduction**

Seen on pixi.js 8.19.0, Chrome 149, Windows 11, RTX 3060. GC timings are shortened below so it reproduces in seconds; the defaults reproduce the same thing after 60 to 90 seconds. Serve over localhost, WebGPU needs a secure context.

```html
<!doctype html>
<script src="https://cdn.jsdelivr.net/npm/pixi.js@8.19.0/dist/pixi.min.js"></script>
<script type="module">
const WGSL = `
struct GlobalUniforms {
  uProjectionMatrix: mat3x3<f32>,
  uWorldTransformMatrix: mat3x3<f32>,
  uWorldColorAlpha: vec4<f32>,
  uResolution: vec2<f32>,
};
struct LocalUniforms {
  uTransformMatrix: mat3x3<f32>,
  uColor: vec4<f32>,
  uRound: f32,
};
struct MyUniforms {
  uTint: vec4<f32>,
};
@group(0) @binding(0) var<uniform> globalUniforms : GlobalUniforms;
@group(1) @binding(0) var<uniform> localUniforms : LocalUniforms;
@group(2) @binding(0) var<uniform> myUniforms : MyUniforms;
struct VSOut { @builtin(position) position : vec4<f32>, };
@vertex
fn mainVertex(@location(0) aPosition : vec2<f32>, @location(1) aUV : vec2<f32>) -> VSOut {
  let mvp = globalUniforms.uProjectionMatrix * globalUniforms.uWorldTransformMatrix * localUniforms.uTransformMatrix;
  let p = mvp * vec3<f32>(aPosition, 1.0);
  var o : VSOut;
  o.position = vec4<f32>(p.xy, 0.0, 1.0);
  return o;
}
@fragment
fn mainFragment(in: VSOut) -> @location(0) vec4<f32> {
  return myUniforms.uTint;
}
`;

await app.init({
  width: 320, height: 240,
  preference: 'webgpu',
  background: 0x3355ff,
  gcFrequency: 400,        // default 30000
  gcMaxUnusedTime: 1200,   // default 60000
});

const makeMesh = (tint, isStatic) => new PIXI.Mesh({
  geometry: new PIXI.MeshGeometry({
    positions: new Float32Array([0, 0, 100, 0, 100, 100, 0, 100]),
    uvs: new Float32Array([0, 0, 1, 0, 1, 1, 0, 1]),
    indices: new Uint32Array([0, 1, 2, 0, 2, 3]),
  }),
  shader: PIXI.Shader.from({
    gpu: PIXI.GpuProgram.from({
      vertex: { source: WGSL, entryPoint: 'mainVertex' },
      fragment: { source: WGSL, entryPoint: 'mainFragment' },
    }),
    resources: {
      myUniforms: new PIXI.UniformGroup({
        uTint: { value: new Float32Array(tint), type: 'vec4<f32>' },
      }, { isStatic }),
    },
  }),
});

const hideAndShow = makeMesh([1, 0, 0, 1], false);
app.stage.addChild(hideAndShow);

const alwaysVisible = makeMesh([1, 1, 0, 1], true);
alwaysVisible.position.set(120, 0);
app.stage.addChild(alwaysVisible);

setTimeout(() => { hideAndShow.visible = false; }, 600);
setTimeout(() => { hideAndShow.visible = true; }, 3000);
// By 3s the console is spamming the validation error and the canvas is black.
// The isStatic mesh actually fails first, while still on screen.
</script>
```

**Expected behavior**

GC unloading is transparent, the same as for vertex/index buffers and textures: the resource is recreated on next use.

**Workaround**

`gcActive: false` at init.

Laserwolve:
##### Description of change

Fixes #12080

The GC system can destroy a uniform buffer's `GPUBuffer` while a cached `GPUBindGroup` still references it. Bind group cache keys are built from each resource's `_resourceId`, and a `UniformGroup`'s id never changes, so the stale bind group keeps being served. Every subsequent `queue.submit()` that draws with it fails validation with `used in submit while destroyed` and the whole frame is discarded, leaving the canvas black. Repro in the issue.

Two changes:

- `UboSystem` now marks the buffers it creates with `autoGarbageCollect = false`. Their bind group keys can never re-key, so they must not be collected behind the cache. They are small, and `destroy()` still frees them.
- `Buffer.unload()` now bumps `_resourceId` and emits `change`. `BufferResource` forwards the change, so bind groups holding a buffer directly or through a `BufferResource` re-key and rebuild on next use. The emit happens after the `unload` event, at which point the buffer systems have already detached their listeners, so the GPU buffer is not eagerly recreated during collection.

This is the same invalidation textures already do: `TextureSource.unload()` bumps `_resourceId` and emits `change`, with an existing unit test covering it. This change brings buffers to parity and adds the matching test.

Verified with the repro from the issue (Chrome 149, Windows 11): on 8.19.0 the validation errors start within seconds and nothing renders; on dev built with this change rendering stays correct, and vertex and index buffers still get collected and recreated on demand, so the GC keeps working where it is safe. The Buffer, UboSystem, UniformGroup, BindGroup and GCSystem suites pass locally.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/buffer/Buffer.ts
+++ b/src/rendering/renderers/shared/buffer/Buffer.ts
@@ -314,8 +314,18 @@ export class Buffer extends EventEmitter<{
     /** Unloads the buffer from the GPU */
     public unload()
     {
+        // the GPU buffer is about to be destroyed, so any cached bind group keyed on the
+        // old _resourceId would keep referencing it. Bump the id and emit 'change'
+        // (after 'unload' so the buffer systems have already detached their listeners)
+        // so BufferResources and bind groups re-key and rebuild on next use.
+        // Mirrors TextureSource.unload().
+        this._resourceId = uid('resource');
+
         /** Unloads the GPU data from the view container. */
         this.emit('unload', this);
+
+        if (!this.destroyed) this.emit('change', this);
+
         for (const key in this._gpuData)
         {
             this._gpuData[key]?.destroy();
--- a/src/rendering/renderers/shared/buffer/__tests__/Buffer.test.ts
+++ b/src/rendering/renderers/shared/buffer/__tests__/Buffer.test.ts
@@ -216,6 +216,42 @@ describe('Buffer', () =>
         expect(buffer._updateSize).toBe(2 * 4);
     });
 
+    it('should increment the resourceId and emit change on unload so cached bind groups re-key', () =>
+    {
+        const buffer = new Buffer({
+            data: new Float32Array([1, 2, 3]),
+            usage: 1,
+        });
+
+        const changeObserver = jest.fn();
+
+        buffer.on('change', changeObserver);
+
+        const startingId = buffer._resourceId;
+
+        buffer.unload();
+
+        expect(buffer._resourceId).not.toBe(startingId);
+        expect(changeObserver).toHaveBeenCalledTimes(1);
+    });
+
+    it('should not emit change from unload during destroy', () =>
+    {
+        const buffer = new Buffer({
+            data: new Float32Array([1, 2, 3]),
+            usage: 1,
+        });
+
+        const changeObserver = jest.fn();
+
+        buffer.on('change', changeObserver);
+
+        buffer.destroy();
+
+        // only the change emitted by destroy itself
+        expect(changeObserver).toHaveBeenCalledTimes(1);
+    });
+
     itLocalOnly('should only add add listeners to buffer on first gpu init', async () =>
     {
         const renderer = (await getWebGPURenderer()) as WebGPURenderer;
--- a/src/rendering/renderers/shared/shader/UboSystem.ts
+++ b/src/rendering/renderers/shared/shader/UboSystem.ts
@@ -54,10 +54,18 @@ export class UboSystem implements System
     {
         const uniformData = this.getUniformGroupData(uniformGroup);
 
-        uniformGroup.buffer ||= new Buffer({
-            data: new Float32Array(uniformData.layout.size / 4),
-            usage: BufferUsage.UNIFORM | BufferUsage.COPY_DST,
-        });
+        if (!uniformGroup.buffer)
+        {
+            uniformGroup.buffer = new Buffer({
+                data: new Float32Array(uniformData.layout.size / 4),
+                usage: BufferUsage.UNIFORM | BufferUsage.COPY_DST,
+            });
+
+            // bind group keys are derived from the UniformGroup's _resourceId, which never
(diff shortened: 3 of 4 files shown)


Laserwolve:
Superseded by #12147, which landed the same `Buffer.unload()` re-key + `change` emit and goes further: `UniformGroup` now forwards the buffer's `change` so cached bind groups re-key, and `_gcLastUsed` passes through `UniformGroup`/`BufferResource` to the underlying buffer so `BindGroup._touch` keeps buffers in constant use from being collected at all — which also covers the `isStatic: true` case this PR only worked around by opting UBO buffers out of GC.

Closing in favour of that. Thanks!

PR #12083, opened 2026-06-14
fix: Rectangle containsRect edge containment

### LINKED ISSUE #12082: Rectangle containsRect method returns false for the same rectangle
### PixiJS version

v8.19.0

https://www.pixiplayground.com/#/edit/i4RRnnvbjMfSkLOBDD7Gf

### Steps to reproduce

Pass the same rectangle to the `containsRect` method.

### What is expected?

Rectangle contains itself as stated in method description:
> Rectangles that occupy the same space are considered to be containing each other.

Rectangle does not contain itself; `containsRect` returns `false`.

### System Info

```Shell
System:
    OS: Linux 7.0 Arch Linux
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz
    Memory: 6.13 GB / 15.48 GB
    Container: Yes
    Shell: 5.3.15 - /usr/bin/bash
  Browsers:
    Chromium: 149.0.7827.53
    Firefox: 151.0.4
    Firefox Developer Edition: 151.0.4
```

### Any additional comments?

_No response_

cyphercodes:
##### Description of change
Fixes #12082.

Updates `Rectangle.containsRect` so the contained rectangle's right and bottom edges are inclusive. This matches the documented behavior that rectangles occupying the same space contain each other.

##### Pre-Merge Checklist
- [x] Lint process passed (`NODE_OPTIONS=--max-old-space-size=4096 npm run test:lint`)
- [x] Tests passed (`xvfb-run -a env ELECTRON_DISABLE_SANDBOX=1 npm run test -- unit --testPathPattern=src/maths/__tests__/Rectangle.test.ts --runInBand`)

### DIFF
--- a/src/maths/__tests__/Rectangle.test.ts
+++ b/src/maths/__tests__/Rectangle.test.ts
@@ -474,13 +474,19 @@ describe('Rectangle', () =>
         it('should handle edge touching cases', () =>
         {
             const outer = new Rectangle(0, 0, 100, 100);
-            const edge = new Rectangle(0, 0, 50, 50); // Touches at corner
+            const edge = new Rectangle(0, 0, 50, 50); // Touches at top-left corner
             const edge2 = new Rectangle(0, 25, 50, 50); // Touches at left edge
             const edge3 = new Rectangle(25, 0, 50, 50); // Touches at top edge
+            const edge4 = new Rectangle(50, 25, 50, 50); // Touches at right edge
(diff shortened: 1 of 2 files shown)


bigtimebuddy (approved):
Good catch 👍

src/maths/__tests__/Rectangle.test.ts
rsilivestr, line 490: This test becomes kinda redundant after the change. Or incomplete if left as is.

PR #12104, opened 2026-07-02
feat: object-form bind/push and capturable, replayable bind state

GoodBoyDigital:
> [!NOTE]
> Stacked on #12103 (`feat/copy-texture`) — only the last commit is new here.

#### Features
- `RenderTargetSystem.bind` and `push` now take an options object, in line with the rest of the v8 API. The positional forms remain as deprecated overloads that warn once and are sanitised into options, so existing callers keep byte-identical behavior:
  ```ts
  renderer.renderTarget.bind({ target: texture, clear: false, flipY: true });
  ```
- New `BindState` interface (`target`, `frame`, `mipLevel`, `layer`, `flipY`) describing what a binding *is*, with `BindOptions extends BindState` adding the per-call `clear`/`clearColor` actions. The system now retains the live binding as a `BindState` — the `renderSurface`/`mipLevel`/`layer` getters and `getBindState()` all read from that single source of truth, written once per bind.
- New `getBindState(out?)`: captures the current binding as a `BindOptions` that replays exactly and non-destructively — `clear` is pinned to `CLEAR.NONE`, the frame is deep-copied, and `target`/`frame` are held as passed (a Texture bound without a frame replays through its frame fallback). Save/restore around a temporary bind no longer needs to enumerate bind axes, and survives future ones:
  ```ts
  const saved = renderer.renderTarget.getBindState();

  renderer.renderTarget.bind({ target: scratchTexture, clear: true });
  // ... draw ...
  renderer.renderTarget.bind(saved);

  // or compose: the saved binding, but into mip 1
  renderer.renderTarget.bind({ ...saved, mipLevel: 1 });
  ```
  Pass `out` to reuse one capture object (zero allocation per capture); `bind(getBindState())` is a no-op.

#### Fixes
- `push()` stack entries are now themselves replayable `BindOptions` holding the as-passed surface and a deep-copied frame. This fixes two latent `pop()` bugs: popping back to a Texture pushed without an explicit frame now re-applies its frame fallback (previously it rebound the full po2 source — the TexturePool case), and a caller mutating its reused frame rect can no longer leak into a later pop.
- `pop()` with no previous binding now throws a clear "unbalanced pop" error instead of failing cryptically inside `bind`; it also returns the restored render target, matching `push`.

#### Chores
- `renderStart()` takes a `BindOptions` instead of a field-for-field duplicate inline type.
- All internal engine call sites (FilterSystem, GlBackBufferSystem, AbstractTextSystem, AlphaMaskPipe, RenderGroupPipe) are migrated to the object forms — a test renders across filters, cacheAsTexture, and the back buffer asserting no internal path ever hits the deprecation.
- Unit tests cover positional/object parity (adaptor args, viewport, projection), warn-once deprecations for both methods, the `getBindState` round-trip (including `flipY`/`frontFaceInverted` and WebGPU mip/layer), frame deep-copy independence, poisoned-`out` overwrite and rect reuse, fallback replay, unbound/unbalanced throws, and the pop replay fixes.

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
+++ b/src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
@@ -1,5 +1,6 @@
 import { Matrix } from '../../../../maths/matrix/Matrix';
 import { Rectangle } from '../../../../maths/shapes/Rectangle';
+import { deprecation } from '../../../../utils/logging/deprecation';
 import { warn } from '../../../../utils/logging/warn';
 import { CLEAR } from '../../gl/const';
 import { calculateProjection } from '../../gpu/renderTarget/calculateProjection';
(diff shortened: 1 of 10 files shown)


src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
```diff
@@ -403,12 +440,38 @@ export class RenderTargetSystem<RENDER_TARGET extends RendererRenderTarget> impl
...
+        let options: BindOptions;
+
+        if ('target' in surfaceOrOptions)
+        {
+            options = surfaceOrOptions;
+        }
+        else
+        {
+            // legacy positional call: sanitise the arguments into a BindOptions and carry on
+            deprecation('8.20.0', 'RenderTargetSystem.bind: positional arguments are deprecated, '
```
Zyie, line 453: This should be wrapped in `// #if _DEBUG` statement

src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
```diff
@@ -594,55 +719,94 @@ export class RenderTargetSystem<RENDER_TARGET extends RendererRenderTarget> impl
...
+        let options: BindOptions;
+
+        if ('target' in surfaceOrOptions)
+        {
+            options = surfaceOrOptions;
+        }
+        else
+        {
+            // legacy positional call: sanitise the arguments into a BindOptions and carry on
+            deprecation('8.20.0', 'RenderTargetSystem.push: positional arguments are deprecated, '
```
Zyie, line 772: This should be wrapped in `// #if _DEBUG` statement

src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
Zyie: Done in ba503bee5 — wrapped in `// #if _DEBUG` / `// #endif`.

src/rendering/renderers/shared/renderTarget/RenderTargetSystem.ts
Zyie: Done in ba503bee5 — wrapped in `// #if _DEBUG` / `// #endif`.

Zyie (approved):

PR #12108, opened 2026-07-06
feat: add warnOnce helper for one-time warnings

GoodBoyDigital:
##### Description of change

Adds a `warnOnce(key, ...args)` utility to `src/utils/logging/warn.ts` for cases where we only want to surface a warning a single time (e.g. per-frame or per-object hot paths that would otherwise spam the console).

- Dedupes on an explicit `key` string, mirroring the "fire once" pattern already used in `deprecation.ts` (`Set` of already-fired messages).
- Routes through the existing `warn`, so it keeps the `PixiJS Warning:` prefix and shares the same global 500-warning cap.
- Exported automatically via `export * from './logging/warn'` in `src/utils/index.ts`.

Stacked on top of `feat/bind-objects`.

Usage:

\`\`\`ts
warnOnce('myTexture-npot', 'Texture is not power of two, mipmaps disabled');
\`\`\`

##### Pre-Merge Checklist

- [ ] Lint process passed (\`npm run lint\`)
- [ ] Tests passed (\`npm run test\`)

Made with [Cursor](https://cursor.com)


bigtimebuddy (commented):
Two questions:
- should this always be fired for prod (minified) and dev build? For instance we ignore deprecation from prod bundle.
- deprecation util has a user-facing way to silence warnings, does that seems appropriate here too?

PR #12109, opened 2026-07-08
fix: Text not always rendering, create CanvasPool 2D contexts with willReadFrequently

whizzkid:
### Problem

Text objects intermittently render blank in Firefox: the browser's GPU-accelerated 2D canvas can hand `texImage2D` a stale or blank snapshot of a freshly drawn glyph canvas, so a Text whose display object state is completely healthy ends up with an empty texture on the GPU.

live demo: https://whizzkid.nu/coolgames/pixi-text-canvas-pool-repro.html
On my system the disappearing texts happen all the time, other systems show no issues. An external partner of ours had the same problems in their game, but not in my test.   But this fix solved all the problems we were seeing.

### Fix

Create CanvasPool's 2D contexts with `willReadFrequently: true`. This keeps these canvases in software rasterization, which removes the snapshot race entirely. It also matches how these canvases are actually used: `CanvasTextGenerator` already reads them back with `getImageData` when trimming, which is the exact access pattern `willReadFrequently` exists for.

### Verification

Automated repro (18 Text rows, whole-container pixel checks per regeneration cycle, real Firefox 151):
- pristine v8.19.0: blank rows within ~10 cycles, every run
- with this change: 0 failures across 14,483 cycles / 120 s soak

The same soak also passes when the canvas snapshot path is bypassed by uploading `getImageData` pixels instead, independently confirming the snapshot race as the root cause; this variant was chosen because it adds no per-upload readback cost.

### Performance

Benchmarks (median per text-texture generation, `gl.finish()` inside the timed region):
- Firefox 151: faster across the board (typical label 0.93 ms -> 0.29 ms)
- Chrome 149: unchanged for steady-state rendering, ~7x faster for `trim: true` texts, +0.02 ms for typical labels


whizzkid:
These failed tests also fail for me on master branch when running them locally... so without my change?

whizzkid:
title fixed.
other tests are still failing, but they fail without my fix as well. So I don't know what to do here :)

PR #12112, opened 2026-07-09
fix: canvas account for texture frame origin in graphics texture fills

### LINKED ISSUE #12111: Canvas renderer texture fill spritesheet shifts
### PixiJS version

8.18.1, 8.19.0

https://stackblitz.com/edit/pixijs-v8-6akuprak?file=src%2Fmain.js

### Steps to reproduce

1. Open the reproduction. It renders the same Graphics texture fill twice: first with the
   WebGL renderer, then with the Canvas renderer (`preference: 'canvas'`).
2. The fill uses an atlas sub-texture whose frame origin is NOT (0,0), with
   `textureSpace: 'local'` and a fill matrix mapping the shape onto the frame's UV region.
3. Compare the two rendered squares.

### What is expected?

Image

Both renderers produce the same result: a solid yellow square (the texture's frame region
scaled to the shape), as the WebGL renderer does.

The Canvas renderer draws the fill shifted by the texture's frame origin (modulo the pattern
tile size), so neighboring atlas regions bleed into the shape with visible wrap seams.

Root cause: in `getCanvasStyle` (src/scene/graphics/canvas/CanvasGraphicsAdaptor.ts),
the fill pattern is created via `canvasUtils.getTintedPattern()`, which crops the texture
to its frame — pattern content starts at (0,0). But `patternMatrix` is built as
`textureMatrix.scale(source.pixelWidth, source.pixelHeight)`, i.e. in full-source pixel
coordinates including the frame origin, which is never subtracted before the matrix is
inverted and applied to the pattern. Any texture with frame.x/frame.y != 0 renders shifted.

### System Info

```Shell
System:
    OS: macOS 15.7.4
    CPU: (12) arm64 Apple M2 Pro
    Memory: 214.78 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 150.0.7871.47
    Firefox: 152.0.4
    Safari: 18.6
```

### Any additional comments?

Found in production: a betting-grid game fills rounded-corner cells from a spritesheet via
Graphics texture fills (used instead of masked Sprites for performance). WebGL renders
correctly; the canvas fallback shows every cell's texture shifted with seam lines.

Suggested fix — compensate for the frame origin before applying the pattern transform:

    if (textureMatrix)
    {
        const resolution = texture.source.resolution;

        patternMatrix.translate(-texture.frame.x * resolution, -texture.frame.y * resolution);
    }
    canvasUtils.applyPatternTransform(pattern, patternMatrix);

in this file: https://github.com/pixijs/pixijs/blob/dev/src/scene/graphics/canvas/CanvasGraphicsAdaptor.ts , 247 line

Verified locally against 8.18.1 (via patch-package): canvas output becomes identical to
WebGL. Textures with a frame at (0,0), gradients and FillPatterns are unaffected.

ComplexImg:
Fixes #12111

## Description of change

In the canvas renderer, `Graphics` texture fills built from atlas sub-textures render
shifted and wrapped. `getCanvasStyle()` creates the fill pattern via
`canvasUtils.getTintedPattern()`, which crops the texture to its frame (pattern content
starts at (0,0)), but the pattern matrix is computed in full-source pixel coordinates —
the frame origin is never subtracted. Any texture whose `frame.x/y != 0` renders shifted
by the frame origin modulo the tile size. WebGL renders the same fill correctly, since it
samples the full source with UVs.

This change translates the pattern matrix by `-frame * resolution` before it is inverted
and applied, aligning it with the cropped pattern content. Textures with a frame at (0,0),
gradients, and `FillPattern`s are unaffected.

Note: this fixes fills whose matrix maps the shape onto the texture frame (the common
spritesheet case, tile-aligned). Matrix-less `textureSpace: 'local'` fills of sub-textures
still differ from WebGL because a cropped canvas pattern cannot reproduce full-source
sampling; that would require building the pattern from the full source instead.

## Pre-Merge Checklist
- [x] Tests and/or benchmarks are included (visual regression scene, fails on canvas without the fix)
- [ ] Documentation is changed or added (n/a — bugfix)
- [x] Lint process passed
- [x] Tests passed


src/scene/graphics/canvas/CanvasGraphicsAdaptor.ts
```diff
@@ -246,6 +246,16 @@ function getCanvasStyle(
                 .scale(texture.source.pixelWidth, texture.source.pixelHeight)
             : style.matrix;
 
+        if (textureMatrix)
+        {
+            // getTintedPattern crops the texture to its frame, so the pattern content starts
+            // at (0,0) - but textureMatrix maps into full-source pixel coordinates. Compensate
+            // for the frame origin, otherwise atlas sub-textures render shifted and wrapped.
+            const resolution = texture.source.resolution;
```
bigtimebuddy: nit use deconstruction 

```suggestion
            const { resolution } = texture.source;
```

bigtimebuddy (approved):

Zyie:
closing in favor of #12123

PR #12118, opened 2026-07-11
fix(unsafe-eval/ubo): match eval 4-arg polyfill signature so custom UniformGroups upload under strict CSP (#12117)

### LINKED ISSUE #12117: Bug: unsafe-eval UBO sync polyfill has wrong arity — custom UniformGroup UBOs upload all-zero under strict CSP (blank WebGPU rendering)
### PixiJS version

8.19.0 (also confirmed by inspection at current `dev` HEAD)

Runnable: https://gist.githack.com/gtran-07/d97fbb8803804438d5873870ce3e85f2/raw/index.html
(source: https://gist.github.com/gtran-07/d97fbb8803804438d5873870ce3e85f2 — a single self-contained `index.html`, pixi 8.19.0 from jsdelivr)

### Steps to reproduce

1. Open the reproduction link. It imports `lib/unsafe-eval/ubo/generateUboSyncPolyfill.mjs` (the exact module `pixi.js/unsafe-eval` installs as `_generateUboSync`) and builds a sync function for a minimal uniform set (`alpha: f32`, `count: i32`).
2. The page then calls the generated sync function **exactly as `UboSystem.syncUniformGroup` calls it** — with 4 arguments `(uniforms, data, dataInt32, offset)` — and prints the resulting buffer.
3. Observe the two red FAIL lines:
   - the 4-arg call leaves the buffer **all-zero** (`data[0] = 0`, expected `1`);
   - the i32 uniform is written through the Float32 view (`dataInt32[1] = 1088421888` — the IEEE-754 bits of `7.0` — expected `7`).
4. A control call using the polyfill's own 3-arg shape `(uniforms, data, offset)` writes correctly, showing the polyfill internals are fine — only the signature disagrees with the caller.

Real-world trigger: any app under a strict CSP (`script-src 'self'`, no `'unsafe-eval'`) — where `pixi.js/unsafe-eval` is mandatory — using a custom `UniformGroup` on the WebGPU renderer. Every custom UBO uploads all-zero, so custom-shader content renders blank (we hit this in production: our entire custom WGSL layer disappeared on the WebGPU backend, while native `Graphics` in the same frame rendered fine).

### What is expected?

The polyfill-generated sync function should have the same signature as the eval build — `new Function("uv", "data", "dataInt32", "offset", …)` — so the numeric `offset` is applied and uniform values reach the buffer. Custom `UniformGroup`s should upload identically with and without `unsafe-eval`.

`generateUboSyncPolyfill` (both `…WGSL` and `…STD40` variants) returns a **3-parameter** closure `(uniforms, data, offset)`:
https://github.com/pixijs/pixijs/blob/dev/src/unsafe-eval/ubo/generateUboSyncPolyfill.ts

…but the caller `UboSystem.syncUniformGroup` invokes it with **4 arguments** — `syncFunction(uniformGroup.uniforms, data, dataInt32, offset)` (and the declared type of `syncFunction` is 4-arg too):
https://github.com/pixijs/pixijs/blob/dev/src/rendering/renderers/shared/shader/UboSystem.ts

So the polyfill's `offset` parameter receives the `dataInt32` typed array and the real numeric offset is silently dropped. Every write then computes `data[dataInt32 + elementOffset]` — the typed array coerces to a garbage string key — so writes land as expando properties and the underlying buffer stays all-zero.

**Bonus latent bug found while root-causing:** in `src/unsafe-eval/ubo/uboSyncFunctions.ts`, the `i32` single-value setter writes through `data` (the Float32 view) instead of `dataInt32`:

```js
i32: (_name, data, offset, _uv, v) => { data[offset] = v; },
```

The eval build writes i32 values through `dataInt32`, so under the polyfill a shader-side `i32` reads the float bit pattern instead of the integer. (Masked today because the signature bug zeroes everything first.) A fix should extend `UboUploadFunction` to receive `dataInt32` so integer types can write the correct view.

### System Info

```Shell
  System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-13700H
    Memory: 5.72 GB / 31.69 GB
  Browsers:
    Chrome: 150.0.7871.115
    Edge: Chromium (150.0.4078.48)
```

(Bug is platform-independent — a pure JS arity mismatch; we hit it in Electron/Chromium on the WebGPU backend.)

### Any additional comments?

Suggested fix: have `generateUboSyncPolyfill` return `(uniforms, data, dataInt32, offset) => …` and thread `dataInt32` through the per-uniform upload functions; fix the `i32` setter (and any other integer types) to write through `dataInt32`.

Workaround we ship today: a side-effect module imported right after `pixi.js/unsafe-eval` that wraps `GpuUboSystem._generateUboSync` / `GlUboSystem._generateUboSync` and, when the produced sync has arity 3, adapts the 4-arg call by forwarding the real numeric offset into the third slot (arity-guarded so it no-ops once the signature is fixed upstream).

sanjibani:
Fixes #12117.

## Problem

\`generateUboSyncPolyfill\` returned a 3-arg closure \`(uniforms, data, offset)\` but the eval build's JIT function (in \`compileBufferSync.ts\`) and the caller \`UboSystem.syncUniformGroup\` both pass four arguments \`(uniforms, data, dataInt32, offset)\`. With the arity mismatch, the third positional argument (\`dataInt32\`) landed in the polyfill's \`offset\` slot, the real numeric \`offset\` was silently dropped, and the per-uniform writers computed \`data[typedArrayCoercedToString + elementOffset]\`. Writes landed on expando properties, the underlying buffer stayed all-zero, and every custom UniformGroup under a strict-CSP \`pixi.js/unsafe-eval\` build (plus all \`i32\`/\`u32\`/\`vec*<i32>\`/\`vec*<u32>\` uniforms on the eval build) read garbage.

The minimal reproducer in the issue shows the failure with two lines: \`data[0] = 0\` instead of the expected \`1\`, and the i32 uniform \`7\` written as the IEEE-754 bit pattern \`1088421888\`. Two distinct root causes:

1. The polyfill's returned closure had the wrong arity (3 instead of 4).
2. The auto-generated file \`src/unsafe-eval/ubo/uboSyncFunctions.ts\` was stale relative to its source-of-truth \`src/rendering/renderers/shared/shader/utils/uboSyncFunctions.ts\` (which was updated in #12091 to add \`i32\`/\`u32\` and the integer vector types). The regenerated \`i32\` setter still wrote through the Float32Array view.

## Fix

Three coordinated changes:

1. \`src/unsafe-eval/ubo/generateUboSyncPolyfill.ts\` (hand-written): the returned callback now takes \`(uniforms, data, dataInt32, offset)\` and forwards \`dataInt32\` to each per-uniform writer. The STD40 and WGSL array-setups also take the new \`dataInt32\` parameter to match the widened \`UboUploadFunction\` type.

2. \`scripts/utils/autoGenerateUnsafeEvalFunctions.ts\`: \`UboUploadFunction\` is now 6-arg \`(name, data, offset, uv, v, dataInt32)\` so the regenerated per-type setters (especially \`i32\`/\`u32\`/\`vec*<i32>\`/\`vec*<u32>\`) can write through the Int32Array view.

3. \`src/unsafe-eval/ubo/uboSyncFunctions.ts\` (regenerated, plus the additions the auto-gen script's last run never picked up): the generated \`i32\`/\`u32\`/\`vec*<i32>\`/\`vec*<u32>\` setters now write through \`dataInt32\` (matching the eval build and the source-of-truth \`src/rendering/renderers/shared/shader/utils/uboSyncFunctions.ts\` that was updated in #12091). The previous generated file was stale relative to the source, so the integer setters were writing float bits into the Float32Array view. Underscore prefixes on unused parameters keep \`tsc\`'s \`noUnusedParameters\` happy.

## Tests

Added \`src/unsafe-eval/ubo/__tests__/generateUboSyncPolyfill.test.ts\` covering:

- (a) std140 \`i32\` + \`vec2<u32>\` round-trip at offset 0,
- (b) the same data at a non-zero caller offset (proves the 4th argument is honored; with the old 3-arg polyfill, values were written at byte 0 instead),
- (c) a single \`f32\` at a non-zero offset (Float32 view path stays correct after the signature change),
- (d) the WGSL equivalents.

Verified locally: \`tsc --noEmit -p tsconfig.json --skipLibCheck\` reports 0 errors in the changed files; a direct esbuild bundle of the polyfill modules + a 16-assertion harness (the same scenarios as the new test file) passes all 16 assertions against the rebuilt polyfill.

## Scope

4 files, +469 / -239 lines. The regenerated file is the bulk of the diff (it picks up the \`i32\`/\`u32\`/\`vec*<i32>\`/\`vec*<u32>\` entries that the previous run never pulled in). The hand-written polyfill and the auto-gen script each gain the 6-arg \`dataInt32\` parameter on the relevant signatures.

### DIFF
--- a/scripts/utils/autoGenerateUnsafeEvalFunctions.ts
+++ b/scripts/utils/autoGenerateUnsafeEvalFunctions.ts
@@ -91,11 +91,17 @@ function autoGenerateUboUnsafeEvalFunctions()
 {
     const out: string[] = [header];
 
-    out.push(`export type UboUploadFunction = (name:string, data:Float32Array, offset:number, uv:any, v:any) => void;`);
+    // The 6-arg signature matches the eval build's `UboUploadFunction` shape
+    // (`name, data, offset, uv, v, dataInt32`). The polyfill wrapper in
+    // `generateUboSyncPolyfill.ts` invokes every per-uniform writer with all
+    // six arguments; `dataInt32` lets integer-typed setters (`i32`, `u32`,
+    // `vec*<i32>`, `vec*<u32>`) write through the Int32Array view rather than
+    // aliasing the float bits. See #12117.
+    out.push(`export type UboUploadFunction = (name:string, data:Float32Array, offset:number, uv:any, v:any, dataInt32:Int32Array) => void;`);
 
     function convertToFunction(body: string)
     {
-        return `(name:string, data:Float32Array, offset:number, uv:any, v:any):void =>
+        return `(name:string, data:Float32Array, offset:number, uv:any, v:any, dataInt32:Int32Array):void =>
         {
             ${body}
         }`;
--- a/src/unsafe-eval/ubo/__tests__/generateUboSyncPolyfill.test.ts
+++ b/src/unsafe-eval/ubo/__tests__/generateUboSyncPolyfill.test.ts
@@ -1,0 +1,145 @@
+import { createUboElementsSTD40 } from '../../../rendering/renderers/gl/shader/utils/createUboElementsSTD40';
+import { createUboElementsWGSL } from '../../../rendering/renderers/gpu/shader/utils/createUboElementsWGSL';
+import {
+    generateUboSyncPolyfillSTD40,
+    generateUboSyncPolyfillWGSL
+} from '../generateUboSyncPolyfill';
+
+import type { UniformData } from '~/rendering';
+
+describe('generateUboSyncPolyfillSTD40 (i32/u32 via Int32 view, #12117)', () =>
+{
+    it('should lay out and round-trip i32 + vec2<u32> in a std140 UBO via the Int32 view', () =>
+    {
+        // Reproduces #12117: a custom UniformGroup under a strict-CSP
+        // `pixi.js/unsafe-eval` build must round-trip integer uniforms through
+        // the Int32 view at the caller-supplied byte offset. Earlier the polyfill
+        // returned a 3-arg closure that silently dropped the numeric `offset`
+        // and coerced the Int32Array to a string key, so writes landed on
+        // expando properties and the underlying buffer stayed all-zero.
+        const uniformData: UniformData[] = [
+            { name: 'idx', type: 'i32', value: -7, size: 1 },
+            { name: 'pair', type: 'vec2<u32>', value: new Uint32Array([0x80000000, 0xFFFFFFFF]), size: 1 },
+        ];
+
+        const layout = createUboElementsSTD40(uniformData);
+
+        // std140: i32@byte0, vec2<u32> aligns to 8 → byte8. Trailing pad to 16.
+        expect(layout).toMatchObject({
+            uboElements: [
+                { offset: 0, size: 4 },
+                { offset: 8, size: 8 },
+            ],
+            size: 16,
+        });
+
+        const sync = generateUboSyncPolyfillSTD40(layout.uboElements);
+
+        const data = new Float32Array(layout.size / 4);
+        const dataInt32 = new Int32Array(data.buffer);
+        const u32View = new Uint32Array(data.buffer);
+
+        const uniforms = {
+            idx: uniformData[0].value,
+            pair: uniformData[1].value,
+        };
+
+        sync(uniforms, data, dataInt32, 0);
+
+        expect(dataInt32[0]).toBe(-7);
+        expect(u32View[0]).toBe(0xFFFFFFF9);
+        expect(u32View[2]).toBe(0x80000000);
+        expect(u32View[3]).toBe(0xFFFFFFFF);
+    });
+
+    it('should honor a non-zero caller-supplied byte offset (write at offset=4, not 0)', () =>
+    {
+        // The earlier 3-arg polyfill dropped the 4th argument entirely; writes
+        // always landed at byte 0. With the 4-arg signature, calling
+        // `sync(..., offset=1)` (1 float = 4 bytes) must place the i32 at
+        // byte 4 and the vec2<u32> at byte 12.
+        const uniformData: UniformData[] = [
+            { name: 'idx', type: 'i32', value: -7, size: 1 },
+            { name: 'pair', type: 'vec2<u32>', value: new Uint32Array([0x80000000, 0xFFFFFFFF]), size: 1 },
+        ];
+
+        const layout = createUboElementsSTD40(uniformData);
+        const sync = generateUboSyncPolyfillSTD40(layout.uboElements);
+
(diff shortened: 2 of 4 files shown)


Zyie:
Thanks for the fix — the root cause is correctly diagnosed. Closing for now as CI is failing on lint (the regenerated uboSyncFunctions.ts and the test file's relative imports fail the --max-warnings 0 gate) and the PR title format check. Feel free to reopen once updated.

gtran-07:
Confirming from the reporting side (#12117): I independently verified this fix (`b609b92`) against a real strict-CSP `pixi.js/unsafe-eval` + WebGPU consumer — it resolves the bug completely, both the signature mismatch and the i32-setter bonus finding.

Minimal headless repro (Node, no GPU — the bug is a pure arity mismatch), run against the pkg.pr.new build of this PR vs `8.19.0`:

```js
import 'pixi.js/unsafe-eval';
import { GpuUboSystem } from 'pixi.js';

// UboSystem installs the WGSL polyfill as _generateUboSync
const sync = GpuUboSystem.prototype._generateUboSync([
  { data: { name: 'uScale', type: 'f32', size: 1, value: 0 }, offset: 0 },
  { data: { name: 'uFlag',  type: 'i32', size: 1, value: 0 }, offset: 4 },
]);

const buf = new ArrayBuffer(32);
const data = new Float32Array(buf);
const dataInt32 = new Int32Array(buf);

// call exactly as UboSystem.syncUniformGroup does — FOUR args:
sync({ uScale: 3.5, uFlag: 7 }, data, dataInt32, 0);

// b609b92 : sync.length === 4, data[0] === 3.5, dataInt32[1] === 7   (uploads correctly)
// 8.19.0  : sync.length === 3, data[0] === 0,   dataInt32[1] === 0   (UBO stays all-zero)
```

| build | returned arity | `f32` uniform | `i32` uniform |
|---|---|---|---|
| `b609b92` (this PR) | **4** | `3.5` ✅ | `7` ✅ (via `Int32Array` view) |
| `8.19.0` | 3 | `0` ❌ | `0` ❌ |

Both parts land: the 4-arg signature (custom `UniformGroup`s upload again under strict CSP) and the `i32`/`u32` setters writing through `dataInt32` instead of the `Float32Array` view.

@Zyie / @sanjibani — keen to see this merged; it's the sole reason we ship a runtime `_generateUboSync` shim today. The only remaining blockers look cosmetic — the `--max-warnings 0` lint on the regenerated `uboSyncFunctions.ts` + the test's relative imports, and the PR-title format check. @sanjibani if you're able to reopen with those tidied I'm glad to help verify; otherwise happy to open a follow-up PR crediting you. Thanks for the fix.

PR #12123, opened 2026-07-13
fix: account for frame origin and rotation in graphics texture fills

### LINKED ISSUE #12111: Canvas renderer texture fill spritesheet shifts

Zyie:
### Overview

Graphics texture fills built from atlas/spritesheet sub-textures rendered incorrectly: the canvas renderer ignored the frame origin (shifted, wrapped output), and groupD8-rotated frames filled wrong on every renderer. This PR supersedes #12112, keeping @ComplexImg's frame-origin fix (their commits are preserved) and extending it with rotation support: the WebGL/WebGPU fill matrix now maps through the texture's UVs (which encode origin plus rotation), and the canvas adaptor maps the fill matrix back through the UV-matrix inverse into the un-rotated, frame-cropped pattern's pixel space. Non-rotated textures keep their exact previous numerics. Both visual scenes pass on all four renderers and fail with the fixes reverted.

Known limitation carried over (documented in code, not addressed here): `textureSpace: 'local'` fills of sub-textures map the whole source on WebGL/WebGPU but the cropped frame on canvas; aligning them is a behavioral decision for a future major.

#### Fixes

- Graphics texture fills using atlas sub-textures (`frame` origin ≠ 0) now render correctly on the canvas renderer, matching WebGL/WebGPU
- Graphics texture fills using rotated atlas frames (`texture.rotate` ≠ 0) now render correctly on all renderers

```ts
const rotated = new Texture({ source: atlas.source, frame, rotate: 2 });
graphics.rect(0, 0, 48, 32).fill({ texture: rotated, textureSpace: 'global' });
```

##### Pre-Merge Checklist


bigtimebuddy (approved):
Looks good!

PR #12124, opened 2026-07-13
fix: make GCSystem lazy-hash-replacement tests independent of real elapsed time

### LINKED ISSUE #12084: Flaky test: GCSystem 'lazy hash replacement' is order-dependent
### Summary
The unit test `GCSystem › lazy hash replacement › should accumulate null count from both existing nulls and GC'd resources` (`src/rendering/renderers/shared/__tests__/GCSystem.test.ts`) is **order-dependent**: it passes reliably in isolation but can fail in a full unit-suite run depending on test execution order.

### Reproduce
- `npx jest --testPathPattern=GCSystem` → passes (56/56).
- Full unit run (`npx jest --testPathIgnorePatterns=tests/visual`) → this test intermittently fails, with the hash not being replaced at the expected null-count threshold.

### Expected
Passes regardless of execution order.

### Likely cause
Reliance on shared / module-level state (e.g. global uid counters or cached state leaking between test files) so the GC null-count thresholds depend on what ran beforehand. Adding or removing unrelated test files shifts ordering and flips the result.

### Suggested fix
Make the test self-contained — reset the relevant shared state in `beforeEach`, or assert on deltas rather than absolute counts so it's independent of prior tests.

### Environment
Local; `jest-electron` runner.

santichausis:
### Overview

The `GCSystem › lazy hash replacement` tests were order-dependent: they passed in isolation but could fail intermittently in a full unit-suite run. They computed staleness from real `performance.now()` calls, so under a loaded full-suite run (many parallel test files competing for CPU) the wall-clock delay between marking a resource as recently used and calling `run()` could exceed `maxUnusedTime` (1000ms in these tests) and flip the expected GC outcome. The describe block now mocks `performance.now()` to a fixed value (`beforeEach`/`afterEach`), computing staleness from a controlled clock instead of real elapsed time.

Fixes #12084

#### Chores

- Removed order and system-load dependence from the `GCSystem` lazy-hash-replacement tests by mocking `performance.now()`, eliminating intermittent full-suite failures

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/__tests__/GCSystem.test.ts
+++ b/src/rendering/renderers/shared/__tests__/GCSystem.test.ts
@@ -832,9 +832,21 @@ describe('GCSystem', () =>
 
     describe('lazy hash replacement', () =>
     {
+        // These tests rely on a fixed "now" to decide which resources are stale. Using the real
+        // performance.now() made them order-dependent: under a full-suite run, real elapsed time
+        // between marking a resource as recently-used and calling `run()` can vary with system
+        // load, occasionally pushing it past `maxUnusedTime`. Mocking the clock removes that variance.
+        let nowSpy: jest.SpyInstance;
+
         beforeEach(() =>
         {
             gcSystem.init({ gcActive: true, gcMaxUnusedTime: 1000, gcFrequency: 100 });
+            nowSpy = jest.spyOn(performance, 'now').mockReturnValue(1_000_000);
+        });
+
+        afterEach(() =>
+        {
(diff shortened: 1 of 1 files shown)


santichausis:
Friendly ping — this has been open for a bit with no reviews yet. Let me know if there's anything I can clarify or adjust. Also worth noting: the failing Visual Tests check is an unrelated WebGPU timeout on `nine-slice-trim`, not caused by this change.

bigtimebuddy (approved):
Seems reasonable 👍

PR #12132, opened 2026-08-05
fix: apply render group world color and alpha to ParticleContainer

### LINKED ISSUE #12127: Bug: ParticleContainer ignores the world color/alpha of its RenderGroup (particles don't fade with an ancestor's alpha)
### PixiJS version

8.19.0

https://stackblitz.com/edit/pixijs-v8-8vhqawql?file=src%2Fmain.js

### Steps to reproduce

1. Build two identical subtrees: `fader (alpha = 0.1) -> wrapper -> [ParticleContainer with one Particle, Sprite]`.
2. In the first subtree leave `wrapper` as a plain Container; in the second set `isRenderGroup: true`.
3. Render.

```
import { ParticleContainer, Application, Assets, Container, Particle, Sprite } from 'pixi.js';

(async () => {
  // Create a new application

  // Initialize the application
  await app.init({ autoStart: false, resizeTo: window });

  const texture = await Assets.load('https://pixijs.com/assets/bunny.png');

  for (const [x, isRenderGroup] of [[200, false], [400, true]]) {
      const fader = new Container();
      fader.alpha = 0.1;                       // alpha set ABOVE the render group
      app.stage.addChild(fader);

      const wrapper = new Container({ isRenderGroup });
      fader.addChild(wrapper);

      const particles = new ParticleContainer();
      particles.addParticle(new Particle({ texture, x, y: 200 }));
      wrapper.addChild(particles);

      const sprite = new Sprite(texture);      // reference renderable
      sprite.position.set(x, 300);
      wrapper.addChild(sprite);
  }

  app.start();
})();

```

### What is expected?

Both particles render at alpha 0.1, exactly like the Sprites beside them. The same applies to tint set above the render group.

The particle under `isRenderGroup: true` renders fully opaque, while the Sprite in the same wrapper fades correctly. Without the render group both fade correctly.

### System Info

```Shell
Reproduces on WebGL, WebGPU and Canvas. Chromium 140 / Windows 10 (also reproduces in Electron).
```

### Any additional comments?

Cause (source read, not a guess):

Every other renderable receives the render group's world color through the shared `globalUniforms` bind, and the generated shader applies it (`vColor *= uWorldColorAlpha` in `defaultProgramTemplate`). `ParticleContainerPipe` bypasses that machinery and builds its own `localUniforms` group, manually copying only the *transform* parts of the global data back in:

```js
// scene/particle-container/shared/ParticleContainerPipe.ts -> execute()
const globalUniformData = renderer.globalUniforms.globalUniformData;
transformationMatrix.tx -= globalUniformData.offset.x;
transformationMatrix.ty -= globalUniformData.offset.y;
transformationMatrix.prepend(globalUniformData.projectionMatrix);
uniforms.uResolution = globalUniformData.resolution;
uniforms.uRound = renderer._roundPixels | container._roundPixels;
color32BitToUniform(container.groupColorAlpha, uniforms.uColor, 0);  // globalUniformData.worldColor is never applied
```

`container.groupColorAlpha` is relative to the render group root, so everything above the render group is lost. Outside a render group the two happen to coincide, which is why this is rarely hit.

All three backends share the gap: the GL and WebGPU adaptors go through the pipe's `uColor`, and `CanvasParticleContainerAdaptor` likewise derives its alpha from `container.groupColorAlpha` alone.

Suggested fix — multiply the pushed world color into `uColor` in `ParticleContainerPipe.execute()`, mirroring `vColor *= uWorldColorAlpha`:

```js
color32BitToUniform(container.groupColorAlpha, uniforms.uColor, 0);
color32BitToUniform(globalUniformData.worldColor, tempWorldColor, 0);
for (let i = 0; i < 4; i++) uniforms.uColor[i] *= tempWorldColor[i];
```

(Both are premultiplied, so a componentwise multiply is the correct composition.) The canvas adaptor needs the equivalent on its `groupAlpha`/tint.

Possibly the same root area as #11903 (CanvasRenderer + ParticleContainer + RenderGroup).

TheOnlyJason:
## Description

`ParticleContainer` ignores the world color and alpha of its `RenderGroup`, so particles do not fade or tint with an ancestor's alpha while every other display object does (#12127).

Unlike the other pipes, the particle shader has no `uWorldColorAlpha` uniform — `particles.vert` computes `vColor = vec4(aColor.rgb * aColor.a, aColor.a) * uColor`. `ParticleContainerPipe.execute` wrote only `container.groupColorAlpha` into `uColor`, so the render group's world color never reached the shader.

Confirmed locally before fixing: with an ancestor render group at `alpha = 0.5`, `uColor[3]` was `1.0`; with this change it is `0.5`.

## Changes

- `src/scene/particle-container/shared/ParticleContainerPipe.ts`: compose `globalUniformData.worldColor` into `uColor`, multiplying the alpha channels and the BGR channels separately (reusing the existing `multiplyHexColors` helper).
- `src/scene/particle-container/shared/__tests__/ParticleContainerWorldColor.test.ts`: new regression tests for an ancestor render group's alpha, an ancestor tint, and the unchanged no-ancestor-color case.

## Testing

- `npx jest --config .configs/jest.config.js --testPathPattern "ParticleContainer"` — 10/10 pass.
- `--testPathPattern "src/scene"` — 81 suites, 997 tests pass.
- Full suite: 1921 pass. The one failing suite (`tests/visual/visuals.test.ts`, html-text-stroke snapshots) fails identically on a clean `dev` checkout with these changes stashed, so it is pre-existing and unrelated.
- `eslint` clean on both changed files.

Fixes #12127

### DIFF
--- a/src/scene/particle-container/shared/ParticleContainerPipe.ts
+++ b/src/scene/particle-container/shared/ParticleContainerPipe.ts
@@ -4,6 +4,7 @@ import { UniformGroup } from '../../../rendering/renderers/shared/shader/Uniform
 import { getAdjustedBlendModeBlend } from '../../../rendering/renderers/shared/state/getAdjustedBlendModeBlend';
 import { State } from '../../../rendering/renderers/shared/state/State';
 import { GCManagedHash } from '../../../utils/data/GCManagedHash';
+import { multiplyHexColors } from '../../container/utils/multiplyHexColors';
 import { color32BitToUniform } from '../../graphics/gpu/colorToUniform';
 import { ParticleBuffer } from './ParticleBuffer';
 import { ParticleShader } from './shader/ParticleShader';
@@ -143,8 +144,17 @@ export class ParticleContainerPipe implements RenderPipe<ParticleContainer>
         uniforms.uResolution = globalUniformData.resolution;
         uniforms.uRound = renderer._roundPixels | container._roundPixels;
 
+        // The particle shader has no uWorldColorAlpha, so the render group's world
+        // color/alpha has to be composed into uColor here. Without this, particles
+        // ignore the tint and alpha of any ancestor render group.
+        const groupColorAlpha = container.groupColorAlpha;
+        const worldColorAlpha = globalUniformData.worldColor;
+
+        const alpha = (((groupColorAlpha >>> 24) * (worldColorAlpha >>> 24)) / 255) | 0;
+        const bgr = multiplyHexColors(groupColorAlpha & 0xFFFFFF, worldColorAlpha & 0xFFFFFF);
+
         color32BitToUniform(
-            container.groupColorAlpha,
+            ((alpha << 24) | bgr) >>> 0,
             uniforms.uColor,
             0
         );
--- a/src/scene/particle-container/shared/__tests__/ParticleContainerWorldColor.test.ts
+++ b/src/scene/particle-container/shared/__tests__/ParticleContainerWorldColor.test.ts
@@ -1,0 +1,62 @@
+import { ParticleContainer } from '../ParticleContainer';
+import { getWebGLRenderer } from '@test-utils';
+import { Texture } from '~/rendering';
+import { Container, Particle } from '~/scene';
+
+describe('ParticleContainer world color and alpha', () =>
+{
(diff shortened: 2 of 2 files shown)


GoodBoyDigital (approved):
nice!

PR #12139, opened 2026-08-10
fix: WebGPU ignores clockwiseFrontFace

koteelok:
##### Description of change
Working on custom 3D rendered for pixi.js i stubbled on divergence of how my PBR probes looked on WebGL vs WebGPU. WebGPU ones were always flipped upside down.

State.clockwiseFrontFace never reaches the WebGPU pipeline as a winding: PipelineSystem passes it as state.cullMode (which returns 'front' when the flag is set) and hardcodes frontFace.

Check the fix out: https://stackblitz.com/edit/vitejs-vite-x5vf8n75?file=src%2Fmain.js

##### Pre-Merge Checklist


koteelok:
This is how my test probes looked with WebGL:

image

And with WebGPU (before the fix):

image

WebGPU after the fix:

image

Zyie (approved):

PR #12155, opened 2026-08-20
fix: restore WebGL context on any loss (v7 parity)

### LINKED ISSUE #12134: WebGL Context Restoration Fails in PixiJS v8
### PixiJS version

8.14.0

https://jsfiddle.net/vote138y/6/

### Steps to reproduce

1. Open the PixiJS v8.14.0 example on JSFiddle: [https://jsfiddle.net/vote138y/6/](https://jsfiddle.net/vote138y/6/)
2. **Simulate a real GPU crash** by manually typing `chrome://gpucrash` in the browser's address bar and pressing Enter.
   - *(Note: A real GPU crash is different from the `WEBGL_lose_context` extension, which only simulates a loss for debugging.)*
3. After the page recovers from the GPU crash, observe the context restoration attempt:
   - The browser fires a `webglcontextrestored` event.
   - **First crash:** The scene may recover.
   - **Second crash:** Repeat step 2. The scene does **not** recover.
4. Alternatively, use the **"Lose Context"** button to simulate context loss:
   - Click **"Lose Context"** → scene disappears.
   - Click **"Restore Context"** → an alert appears: `WEBGL_lose_context not supported`.
   - The scene remains blank.

**Comparison with v7:**
In PixiJS v7.4.2, both `chrome://gpucrash` and manual context loss/restore work reliably. The scene recovers every time.  
Reference: [https://jsfiddle.net/0r58Lakx/1/](https://jsfiddle.net/0r58Lakx/1/)

### What is expected?

1. After a real GPU crash (triggered by `chrome://gpucrash`), the WebGL context should restore automatically.
2. The scene should fully recover and display correctly.
3. The `WEBGL_lose_context` extension should remain available for manual testing.
4. Multiple restores (both automatic and manual) should work reliably, as they do in v7.

In PixiJS v8.14.0:

**Scenario 1: Real GPU Crash (via `chrome://gpucrash`)**
1. First crash → context may restore (inconsistent).
2. Second crash → context does **not** restore.
3. The scene remains blank.

**Scenario 2: Manual Context Loss (via `WEBGL_lose_context` extension)**
1. Click **"Lose Context"** → context lost.
2. Click **"Restore Context"** → an alert appears: `WEBGL_lose_context not supported`.
3. The scene does **not** recover.
4. The bug occurs on the **very first manual restore attempt**.

### System Info

```Shell
PixiJS version: 8.14.0 (regression from 7.4.2)
Browser: Chrome 150.0.0.0
OS: Windows 10 Version 22H2 (Build 19045.6456)
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36
```

### Any additional comments?

_No response_

qiao-coding:
# Fix: restore WebGL context on any loss (v7 parity)

Closes #12134

## Problem

Since v8, an externally-triggered WebGL context loss never recovers — the canvas stays permanently blank. v7 restored the context unconditionally; v8 does not.

Reproduction (JSFiddle vote138y/6 from #12134):
- Click "Lose Context" → scene disappears, never comes back.
- "Restore Context" button cannot help: after loss, `gl.getExtension('WEBGL_lose_context')` returns `null`, so the alert "WEBGL_lose_context not supported" fires and the scene stays blank.
- A real GPU crash (`chrome://gpucrash`) may recover once, then fails on a second crash.

## Root cause

`GlContextSystem.handleContextLost` only called `restoreContext()` when `_contextLossForced` was `true` — a flag set only by Pixi's own `forceContextLoss()`. For every other loss path (real GPU crash, or an app calling `loseContext()` directly) the gate never opens, and because the browser **never auto-restores an API-triggered loss**, the context stays lost forever.

The gate was added in PR #10111 to avoid resurrecting a context during teardown. That concern no longer applies: `destroy()` removes the `webglcontextlost` listener **before** calling `loseContext()`, so the handler never runs during teardown. The guard is redundant.

## Fix

Restore the v7 behaviour — attempt `restoreContext()` on any loss, not just Pixi-initiated ones.

In `src/rendering/renderers/gl/context/GlContextSystem.ts`, removed the `_contextLossForced` gate and its field:

```ts
protected handleContextLost(event: WebGLContextEvent): void
{
    event.preventDefault();

    // v8 previously gated this behind `_contextLossForced` (PR #10111) to avoid
    // resurrecting a context mid-teardown — but destroy() removes the
    // webglcontextlost listener *before* calling loseContext(), so that guard
    // is redundant. Restoring on ANY loss (GPU crash, an app calling
    // WEBGL_lose_context.loseContext() directly, or forceContextLoss()) is
    // required because the browser never auto-restores an API-triggered loss
    // (matches v7 behaviour).
    setTimeout(() =>
    {
        if (this.gl.isContextLost())
        {
            this.extensions.loseContext?.restoreContext();
        }
    }, 0);
}
```

Also removed:
- the `private _contextLossForced: boolean;` field
- `this._contextLossForced = true;` in `forceContextLoss()`

Why it's safe: `destroy()` (lines 422-435) removes the `webglcontextlost` listener **before** calling `loseContext()`, so `handleContextLost` never runs during teardown — the original purpose of the gate is already handled by that code, making the gate redundant.

## Tests

Two new test files (run on Electron / real WebGL2):

- `GlContextSystem.test.ts` — external `loseContext()` → `restoreContext()` is called and rendering matches the baseline (red on the pre-fix code, green with the fix).
- `_fiddle_repro.test.ts` — mirrors the JSFiddle repro: automatic recovery with no manual restore, incl. 3 repeated losses; also asserts that `getExtension('WEBGL_lose_context')` is `null` after loss (why the fiddle's Restore button cannot work).

Both fail on the pre-fix code and pass with the fix. Also verified:
- `destroy()` does not trigger `restoreContext()`.
- `forceContextLoss()` still restores — the "sleep / release GPU" feature is unchanged.
- All GL renderer tests pass; lint and type-check clean.

### DIFF
--- a/src/rendering/renderers/gl/context/GlContextSystem.ts
+++ b/src/rendering/renderers/gl/context/GlContextSystem.ts
@@ -174,7 +174,6 @@ export class GlContextSystem implements System<ContextSystemOptions>
     public canvas: ICanvas;
 
     private _renderer: WebGLRenderer;
-    private _contextLossForced: boolean;
 
     /** @param renderer - The renderer this System works for. */
     constructor(renderer: WebGLRenderer)
@@ -400,19 +399,19 @@ export class GlContextSystem implements System<ContextSystemOptions>
     {
         event.preventDefault();
 
-        // only restore if we purposefully nuked it
-        if (this._contextLossForced)
+        // Restore the context after this event has exited.
+        // v8 previously gated this behind `_contextLossForced`  to avoid resurrecting
+        // a context mid-teardown — but destroy() removes the webglcontextlost listener *before*
+        // calling loseContext(), so that guard is redundant. Restoring on ANY loss (GPU crash, an
+        // app calling WEBGL_lose_context.loseContext() directly, or forceContextLoss()) is required
+        // because the browser never auto-restores an API-triggered loss (matches v7 behaviour).
+        setTimeout(() =>
         {
-            this._contextLossForced = false;
-            // Restore the context after this event has exited
-            setTimeout(() =>
+            if (this.gl.isContextLost())
             {
-                if (this.gl.isContextLost())
-                {
-                    this.extensions.loseContext?.restoreContext();
-                }
-            }, 0);
-        }
+                this.extensions.loseContext?.restoreContext();
+            }
+        }, 0);
     }
 
     /** Handles a restored webgl context. */
@@ -447,7 +446,6 @@ export class GlContextSystem implements System<ContextSystemOptions>
     public forceContextLoss(): void
     {
         this.extensions.loseContext?.loseContext();
-        this._contextLossForced = true;
     }
     /**
      * Validate context.
--- a/src/rendering/renderers/gl/context/__tests__/GlContextSystem.test.ts
+++ b/src/rendering/renderers/gl/context/__tests__/GlContextSystem.test.ts
@@ -1,0 +1,94 @@
+import { getWebGLRenderer } from '@test-utils';
+import { Texture } from '~/rendering';
+import { Container, Sprite } from '~/scene';
+
+async function waitFor(condition: () => boolean, timeout = 2000): Promise<void>
+{
+    const start = Date.now();
+
+    while (!condition())
+    {
+        if (Date.now() - start > timeout) throw new Error('Timed out waiting for condition');
+        await new Promise((resolve) => setTimeout(resolve, 10));
+    }
+}
+
+describe('GlContextSystem', () =>
+{
+    it('should restore the context after an externally-triggered context loss', async () =>
+    {
+        const renderer = await getWebGLRenderer();
+
+        try
+        {
+            const ext = renderer.context.extensions.loseContext;
+
+            expect(ext).toBeTruthy();
+
+            // An external context loss — a real GPU crash, or an app calling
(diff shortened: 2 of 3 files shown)


tgrajewski:
The [restoreContext](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_lose_context/restoreContext) is not meant to restore GPU context, it's meant to simulate restoration. You actually can't restore GPU context from JS code, you need to wait for the browser and OS to handle you the GPU back which happens when browser sends the `webglcontextrestored` event.

The restoreContext function and whole [WEBGL_lose_context](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_lose_context) extension are for **debugging** purposes only so you can test that the JS code of your app handles it properly. Excerpt from the documentation:
> The WEBGL_lose_context extension is part of the WebGL API and exposes functions to **simulate** losing and restoring a WebGLRenderingContext.

In other words restoreContext is not meant to be called in production live code, but only during development/debugging.

qiao-coding:
Thanks for the review. To be clear on what this change did: it fixed the `loseContext()` repro in #12134 — but only for a simulated loss. A real GPU loss is recovered by the browser and is unaffected. Since `restoreContext()` is a testing API, wiring it into `handleContextLost` isn't appropriate for production. Closing.

PR #12157, opened 2026-08-20
feat: support TypeScript 5, 6, and 7

### LINKED ISSUE #12130: Feature Request: Support Typescript 6
### Description

PixiJS's WebGPU declarations currently do not compile cleanly with TypeScript 6.0.3.  
  
It would be nice if PixiJS supported TypeScript 6 without requiring downstream projects to patch its declarations.  
  
### Environment  
  
- PixiJS: 8.19.0  
- TypeScript: 6.0.3  
  
### Problem  
  
TypeScript 6.0.3 includes a partial set of WebGPU declarations in its built-in DOM library. PixiJS also unconditionally loads `@webgpu/types` through this reference in `lib/Shaders.d.ts`:  
  
```ts  
/// <reference types="@webgpu/types" />  
```  
  
This produces duplicate and incompatible declarations between TypeScript's DOM types and `@webgpu/types`.
  
### Current workaround  
  
Following the  [`@webgpu/types` documentation](https://github.com/gpuweb/types#integration-into-dom-type-definitions), we use `@types/web` as our DOM declaration source when working with TypeScript 6.0.3.
  
We also patch PixiJS locally to remove the explicit `@webgpu/types` reference from `Shaders.d.ts`. With that patch, the complete WebGPU declarations provided by `@types/web` are used without loading the conflicting supplemental declarations.
### Requested support  
  
It would be helpful if PixiJS could:  
  
- Add TypeScript 6 support.
- Ensure PixiJS’s WebGPU declarations compile without conflicting with WebGPU types supplied by TypeScript or `@types/web`, without requiring downstream patches.
  
Additionally, according to the [`@webgpu/types` documentation](https://github.com/gpuweb/types#integration-into-dom-type-definitions):  
  
> It appears that v6.0.4 and above will include a full set of WebGPU types and thus should not need the supplemental types from this repo.  
  
If PixiJS moves its minimum supported version to TypeScript 6.0.4 or later, it may be possible to remove the explicit `@webgpu/types` reference entirely.

creativoma:
### Overview

PixiJS's published declarations do not compile under TypeScript 6 or 7: they reference `@webgpu/types`, which conflicts with the WebGPU types TypeScript 6 builds into its own `lib.dom`. This fixes that and makes TypeScript 5, 6 and 7 all supported consumer versions. The repo now builds with TypeScript 6 by default; `lib/index.d.ts` is the canonical entry with no `@webgpu/types` reference, and TypeScript 5 consumers are routed to `.legacy.d.ts` wrappers (for `pixi.js`, `pixi.js/gif` and `pixi.js/html-source`) that add the reference back, selected per consumer through the `types@<6.0` export condition and `typesVersions`. A `dts` check compiles a consumer fixture per typed entry point against all three versions so none of them can regress unnoticed.

Closes #12130

#### Features
- TypeScript 5, 6 and 7 are all supported: each version gets the entry point that matches its built-in WebGPU types, and the declarations are checked against 5.9, 6.0 and 7.0 in CI
- TypeScript setup is documented per version in the README, a new TypeScript guide and the `pixijs-create` skill: nothing to do on TypeScript 5; on 6 and 7, `@types/web` in place of `"dom"` and no `@webgpu/types` in `types`
  ```json
  {
    "compilerOptions": {
      "lib": ["esnext"],
      "types": ["@types/web"]
    }
  }
  ```

#### Fixes
- Type declarations compile on TypeScript 6 and 7 with `skipLibCheck: false`; TypeScript 5 keeps working with no setup change, including programs that import only `pixi.js/gif` or `pixi.js/html-source`
- `HTMLCanvasElement` and `OffscreenCanvas` keep satisfying `ICanvas` on TypeScript 6 and 7, whose `lib.dom` lacks the `getContext('webgpu')` overload; the declarations now ship it

#### Chores
- Build the repo with TypeScript 6: `typescript` ^6.0.3, `typedoc` ^0.28.20, `typescript-eslint` ^8.67.0; migrate off the compiler options TypeScript 6 deprecates and 7 removes (`moduleResolution: node`, `baseUrl`, `downlevelIteration`)
- Add the `dts` check (`npm test -- dts`) to the lint CI job: a consumer fixture per typed entry point compiled with TypeScript 5.9, 6.0 and 7.0 under `bundler` and `node10` resolution, plus a guard that `lib/index.d.ts` never regains an `@webgpu/types` reference
- Keep the WebGPU flag namespaces (`GPUTextureUsage` and friends, which TypeScript 6 and 7's `lib.dom` omit) in an internal `types/WebGPUFlags.d.ts` that is not shipped, so consumers with `@webgpu/types` installed get no collision

##### Pre-Merge Checklist


creativoma:
CI is green, including the new `dts` check in the lint job; the TypeScript 6 fetch through `npx` adds a few seconds, no more.

Two calls are yours, easy either way:

1. If you would rather the check ran in the `Build` step instead of lint, it is a one line move.
2. `moduleResolution: node16` still reports 2 errors from `earcut` being ESM only. Identical on an untouched checkout, so I left it alone — happy to take it here or open a separate issue.

Zyie:
hey @creativoma thanks for putting this together! I'm thinking that we should flip this PR to update the project to typescript 6 by default and do the workarounds for supporting typescript 5

creativoma:
@Zyie Sounds good; I built that version locally and the full suite is green on TypeScript 6.0.3 (lint, types, build, 1951 unit, 1666 visual, docs). The flip means moving the repo onto TS6 with @webgpu/types out of the tsconfig, lib/index.d.ts as the canonical entry and a lib/index.legacy.d.ts wrapper adding it back for TS5, plus a few small workarounds for what TS6's lib.dom doesn't cover: the canvas getContext('webgpu') overloads and the WebGPU flag namespaces (GPUTextureUsage and friends), which it omits entirely.

It also needs typedoc@^0.28.20 and typescript-eslint@^8.67, the first releases that accept TS6; ts-jest is fine as is, so no jest migration. I also checked TypeScript 7.0.2 while I was there: same gaps, and the published declarations compile clean, so one shim covers both. 

Three calls I'd like from you before I push: TS6 or TS7 (I'd say 6 until typedoc and typescript-eslint catch up), whether to migrate off the deprecated compiler options or just silence them with ignoreDeprecations, and whether the flag namespaces should ship or stay internal to the repo.

Zyie:
Thanks for looking into this

> Three calls I'd like from you before I push: TS6 or TS7 (I'd say 6 until typedoc and typescript-eslint catch up), whether to migrate off the deprecated compiler options or just silence them with ignoreDeprecations, and whether the flag namespaces should ship or stay internal to the repo.

- Lets use TS6 for now as I'm pretty sure typedoc doesn't currently work with TS7
- Migrate the deprecated options
- I'm not sure what these flag namespaces are? Push up your changes and i'll take a look

creativoma:
@Zyie Pushed the changes; TS 6 by default, deprecated compiler options fully migrated (not silenced), and I've left the WebGPU flag namespaces internal to the repo for now (see types/WebGPUFlags.d.ts and the PR description) since that one's still your call 👍🏻

package.json
```diff
@@ -216,11 +224,13 @@
   "exports": {
     ".": {
       "import": {
-        "types": "./lib/index.d.ts",
+        "types@>=6.0": "./lib/index.d.ts",
+        "types": "./lib/index.legacy.d.ts",
```
bigtimebuddy: Can this be types and types@<6.0? So that the default is the latest?

Zyie:
thanks for this @creativoma! I've pushed up a couple of changes after testing this out with the different TS versions

Zyie (approved):

PR #12188, opened 2026-09-07
fix: clamp blur samples to the input frame

### LINKED ISSUE #12171: Bug: BlurFilter bleeds leftovers from TexturePool — samples outside the frame without clamping to uInputClamp
### PixiJS version

8.20.1

Self-contained single file — save as `repro.html` and open it, no build step and nothing else needed:

```html
<!doctype html>
<meta charset="utf-8" />
<body style="margin:0;background:#101018">
<script type="module">
import { Application, Container, Graphics, BlurFilter } from 'https://cdn.jsdelivr.net/npm/pixi.js@8.20.1/dist/pixi.mjs';

// ?renderer=webgpu to switch — the bleed only shows on WebGL
const preference = new URLSearchParams(location.search).get('renderer') || 'webgl';

await app.init({ width: 1000, height: 560, background: 0x101018, preference });

// Both filter frames land in the same TexturePool bucket (512x512):
//   big:   460 + 2*8  = 476 -> nextPow2 512
//   small: 300 + 2*30 = 420 -> nextPow2 512
const big = new Container();
big.addChild(new Graphics().rect(0, 0, 460, 460).fill(0xffffff));
big.position.set(20, 50);
big.filters = [new BlurFilter({ strength: 8 })];

// The small container holds only two dots in the corners of a 300x300 area.
// Everything between them is empty, so anything else in there is bleed.
const small = new Container();
const dots = new Graphics();
dots.circle(14, 14, 14).fill(0x44ff88);
dots.circle(286, 286, 14).fill(0x44ff88);
small.addChild(dots);
small.position.set(600, 50);
small.filters = [new BlurFilter({ strength: 30 })];

// Order matters: the big one renders first and returns its texture to the pool,
// the small one gets that same texture back — leftovers included.
app.stage.addChild(big, small);
</script>
```

I can put this on JSFiddle/StackBlitz as well if a hosted link is required — say the word.

### Steps to reproduce

Two containers, each with a `BlurFilter`, rendered in the same frame. Their filter frames are different sizes but land in the **same `TexturePool` bucket**:

- big: `460 + 2 * 8` (padding) = 476 → `nextPow2` = **512**
- small: `300 + 2 * 30` (padding) = 420 → `nextPow2` = **512**

The big one is rendered first, so it returns its texture to the pool and the small one gets that same texture back.

### What is expected?

The small container shows two blurred green dots and nothing else.

A bright vertical stripe runs down the **right** edge of the small container's filter frame and a bright horizontal stripe along its **bottom** edge — the white of the big container bleeding in. In a real app, where a filtered container grows or shrinks by one child at a time, the stripes jump around as the frame size changes.

Cause: the blur shader samples the input texture with **no clamp to `uInputClamp`** at all, in GLSL and in WGSL alike:

- `src/filters/defaults/blur/gl/generateBlurFragSource.ts` → `texture(uTexture, vBlurTexCoords[%index%])`
- `src/filters/defaults/blur/gpu/generateBlurProgram.ts` → `textureSample(uTexture, uSampler, offset${i})`

Outside the filter frame the pooled texture still holds whatever the previous, larger user of that bucket left there, and the blur reads it.

This looks like the other leg of #10808. That one was fixed in #10869 by clearing the blur's own intermediate texture; here the leftovers come from the **input** texture that `FilterSystem` hands to the filter, which is only cleared over its frame.

Notes:

- `padding` does not help — it moves the frame edge and the stripes move with it.
- Reproduced on **WebGL**. The same reproduction on `preference: 'webgpu'` looks clean, so the WebGPU path either clears more or lays the pool out differently.
- Filters that do clamp (for example `GlowFilter` from `pixi-filters`) are unaffected, which is what makes the blur look like the odd one out.

### System Info

```Shell
  System:
    OS: Windows 11 10.0.26200
    CPU: (16) x64 AMD Ryzen 7 9800X3D 8-Core Processor
    Memory: 33.74 GB / 61.56 GB
  Browsers:
    Chrome: 152.0.7977.65
    Edge: Chromium (152.0.4191.53)
```

### Any additional comments?

Clamping the samples to `uInputClamp`, the way `pixi-filters`' `GlowFilter` does, would fix it in both backends and would not depend on the pool being clean.

ICOM725:
##### Description of change

Clamp blur texture samples to `uInputClamp` in the GLSL and WGSL generators. This prevents a smaller filtered object from sampling pixels left outside its input frame by a previous user of the pooled texture.

Add six pixel-based regression cases covering pooled-texture reuse in WebGL 1/2 at quality 1/4, and `repeatEdgePixels` at quality 1/4. The initial reproduction failed before the fix with a maximum channel difference of 156 between clean-pool and reused-pool output; it passes with the fix (tolerance: 1).

Fixes #12171.

Verification on Windows:

- `npm run test -- unit --testPathPattern=src/filters --runInBand`: 22 tests passed.
- With `SCENE_FILTER=blur-filter*.scene.ts`, `npm run test -- visual --runInBand`: all 6 existing blur visual tests passed across WebGL 1, WebGL 2 and WebGPU, including legacy mode. No baselines changed.
- `npm run test -- lint types index`: all three type checks and the index check passed. Default lint is blocked by CRLF checkouts of unchanged files (`core.autocrlf=true`).
- ESLint passes on all changed files with no rules disabled. A separate full-repository ESLint run with only `linebreak-style` disabled passes with zero warnings; no lint configuration was changed.
- `git diff --check` passed. The full `npm test` suite and production build were not run.

##### Pre-Merge Checklist

### DIFF
--- a/src/filters/defaults/blur/__tests__/BlurFilterRendering.test.ts
+++ b/src/filters/defaults/blur/__tests__/BlurFilterRendering.test.ts
@@ -1,0 +1,104 @@
+import '~/scene/graphics/init';
+import '~/rendering/init';
+import { BlurFilter } from '../BlurFilter';
+import { getWebGLRenderer } from '@test-utils';
+import { RenderTexture, TexturePool } from '~/rendering';
+import { Container, Graphics } from '~/scene';
+
+describe('BlurFilter rendering', () =>
+{
+    it.each<[1 | 2, number]>([[1, 1], [1, 4], [2, 1], [2, 4]])(
+        'should not sample pooled pixels with WebGL %i and quality %i', async (preferWebGLVersion, quality) =>
+        {
+            const renderer = await getWebGLRenderer({ width: 1000, height: 560, preferWebGLVersion });
+            const output = RenderTexture.create({ width: 1000, height: 560 });
+            const stage = new Container();
+            const big = new Container({ x: 20, y: 50 });
+            const small = new Container({ x: 600, y: 50 });
+            const bigBlur = new BlurFilter({ strength: 8 });
+            const smallBlur = new BlurFilter({ strength: 30, quality });
+
+            big.addChild(new Graphics().rect(0, 0, 460, 460).fill(0xffffff));
+            big.filters = [bigBlur];
+            small.addChild(new Graphics()
+                .circle(14, 14, 14).fill(0x44ff88)
+                .circle(286, 286, 14)
+                .fill(0x44ff88));
+            small.filters = [smallBlur];
+            stage.addChild(small);
+
+            try
+            {
+                TexturePool.clear();
+                renderer.render({ container: stage, target: output, clear: true });
+                const expected = renderer.extract.pixels(output).pixels;
+
+                TexturePool.clear();
+                // Both filter frames use 512x512 pooled textures. Rendering the larger
+                // object first leaves pixels outside the smaller object's filter frame.
+                stage.addChildAt(big, 0);
+                renderer.render({ container: stage, target: output, clear: true });
+                const actual = renderer.extract.pixels(output).pixels;
+                let maxDifference = 0;
+                let maxAlpha = 0;
+
+                for (let y = 0; y < 560; y++)
+                {
+                    for (let x = 540; x < 1000; x++)
+                    {
+                        const offset = ((y * 1000) + x) * 4;
+
+                        maxAlpha = Math.max(maxAlpha, expected[offset + 3]);
+                        for (let channel = 0; channel < 4; channel++)
+                        {
+                            const difference = Math.abs(actual[offset + channel] - expected[offset + channel]);
+
+                            maxDifference = Math.max(maxDifference, difference);
+                        }
+                    }
+                }
+
+                expect(maxAlpha).toBeGreaterThan(0);
+                expect(maxDifference).toBeLessThanOrEqual(1);
+            }
+            finally
+            {
+                stage.destroy({ children: true });
+                bigBlur.destroy();
+                smallBlur.destroy();
+                output.destroy(true);
+                TexturePool.clear();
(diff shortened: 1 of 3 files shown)


GoodBoyDigital (approved):

Saulite933:
> ##### Description of change
> 
> Clamp blur texture samples to `uInputClamp` in the GLSL and WGSL generators. This prevents a smaller filtered object from sampling pixels left outside its input frame by a previous user of the pooled texture.
> 
> Add six pixel-based regression cases covering pooled-texture reuse in WebGL 1/2 at quality 1/4, and `repeatEdgePixels` at quality 1/4. The initial reproduction failed before the fix with a maximum channel difference of 156 between clean-pool and reused-pool output; it passes with the fix (tolerance: 1).
> 
> Fixes #12171.
> 
> Verification on Windows:
> 
> - `npm run test -- unit --testPathPattern=src/filters --runInBand`: 22 tests passed.
> - With `SCENE_FILTER=blur-filter*.scene.ts`, `npm run test -- visual --runInBand`: all 6 existing blur visual tests passed across WebGL 1, WebGL 2 and WebGPU, including legacy mode. No baselines changed.
> - `npm run test -- lint types index`: all three type checks and the index check passed. Default lint is blocked by CRLF checkouts of unchanged files (`core.autocrlf=true`).
> - ESLint passes on all changed files with no rules disabled. A separate full-repository ESLint run with only `linebreak-style` disabled passes with zero warnings; no lint configuration was changed.
> - `git diff --check` passed. The full `npm test` suite and production build were not run.
> 
> ##### Pre-Merge Checklist
> 
> - [x] Tests and/or benchmarks are included
> - [ ] Lint process passed (`npm run lint`)
> - [ ] Tests passed (`npm run test`)
> 

- [x]

PR #12191, opened 2026-09-08
fix: make the GC resource sweep safe under re-entrant unload listeners

Zyie:
### Overview

`GCSystem.run()` compacted `_managedResources` in place while calling `unload()` on stale entries. Every core `unload()` emits `unload` synchronously, and `addResource` registers `removeResource` on that event, so the swap-remove ran underneath the compaction: with `[A, B, C]` and A stale, C was never visited and stayed untracked for good with its `_gcData` intact. Only `gc.addResource` users are affected; core tracks through the hash path. The sweep now walks a snapshot and routes every removal through `removeResource`.

#### Breaking Changes / Behavior Changes
- `GCSystem.runOnResource` (protected) is now `(resource, now): void`; it no longer takes or returns a write index

#### Fixes
- The GC resource sweep no longer loses tracked resources when an `unload` listener removes the resource itself or another tracked one
- A resource collected by the sweep is unloaded before it is untracked, so a listener that touches it during unload cannot block it from being registered again

##### Pre-Merge Checklist

### DIFF
--- a/src/rendering/renderers/shared/GCSystem.ts
+++ b/src/rendering/renderers/shared/GCSystem.ts
@@ -123,7 +123,7 @@ export class GCSystem implements System<GCSystemOptions>
     private _renderer: Renderer;
 
     /** Array of resources being tracked for garbage collection */
-    private readonly _managedResources: GCableEventEmitter[] = [];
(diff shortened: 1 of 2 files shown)


GoodBoyDigital (approved):
removed the slice :D - looking good!

PR #12200, opened 2026-09-14
fix: decompose mirrored matrix into rotation and negative scale

### LINKED ISSUE #11197: Bug: Matrix `decompose()` method doesn't work correctly with mirrored matrix
### Current Behavior

When a `Matrix` with a rotation has a "mirroring" scale, such as -1 across X or Y axis, calling `.decompose()` on the matrix produces skew values, and no rotation value, on the returned `Transform` object.

### Expected Behavior

The basis vectors of the matrix are still orthogonal in this case, so `.decompose()` should return a rotation if the matrix has a rotation. No skew values.

### Steps to Reproduce

1. Create a new `Matrix`
2. Apply a rotation with `.rotate(rad)`
3. Apply a mirror with `.scale(-1, 1)`
4. Create a new `Transform`
5. Call `.decompose(t)`
6. Observe `t.rotation` is 0

### Environment

- **`pixi.js` version**: 8.6.6
- **Browser & Version**: N/A
- **OS & Version**: N/A
- **Running Example**:  https://codesandbox.io/p/sandbox/xenodochial-feistel-tv7rdg

### Possible Solution

The 2nd condition in [this check](https://github.com/pixijs/pixijs/blob/ee65355a7e93f51199e19ecd27157679d7b04bd8/src/maths/matrix/Matrix.ts#L382) should be expanded, though I'm not sure exactly what would be best. In my x-mirror example, `delta` was PI, and the condition failed.

### Additional Information

_No response_

cpruijsen:
`Matrix.decompose()` returns the wrong transform for a mirrored matrix. A rotation combined with a
negative scale on one axis has orthogonal basis vectors, so the existing orthogonal branch is taken,
but that branch assumes a positive scale and produces a rotation of 0 with a positive scale, which
does not reconstruct the matrix.

The two cases are told apart by the angle between the basis vectors: the existing branch handles the
vectors being aligned, and a difference of pi means they are orthogonal but mirrored. In that case the
rotation is `skewY - Math.PI`, the skews are zero, and `scaleX` is negated, which is the
representation `Transform` uses for a mirror.

`src/maths/matrix/Matrix.ts` computes `scaleX` into a local so the mirrored branch can negate it
before it is assigned, rather than assigning the positive root and correcting afterwards.

The tests cover mirroring on each axis, asserting the rotation is not zero and that the decomposed
transform rebuilds the original matrix.

Fixes #11197


GoodBoyDigital (approved):

amoffat:
Thank you @cpruijsen :pray:

PR #12203, opened 2026-09-15
fix: don't destroy the shared GpuProgram on WebGPU renderer destroy

### LINKED ISSUE #12201: WebGPU: destroying one Application breaks Graphics/Mesh rendering in other live Applications (shared cached GpuProgram is destroyed)
### PixiJS version

8.20.1 (also present on `dev`)

https://jsfiddle.net/ZenBlender/5hfLetjg/

### Steps to reproduce

1. Open the reproduction in a browser with WebGPU support. Keep the tab visible, because the ticker only renders while it is.
2. Leave the renderer on `webgpu` and click **Run without fix**. Each run does the following, using a freshly imported copy of PixiJS so runs are independent:
   1. Creates two `Application`s. Each draws a small, batchable rect.
   2. Destroys app A.
   3. Adds a non-batchable `Graphics` (a 300-point polygon) to app B.
3. Check the log below the canvas.
4. For comparison, click **Run with fix**. It applies the change proposed below at runtime before destroying app A. Then switch the renderer to `webgl` and click **Run without fix**. Runs can be repeated in any order.

### What is expected?

App B keeps rendering and draws the blue rect and the green circle. This is what happens with `webgl`, and with `webgpu` when the fix is applied.

With `webgpu` and no fix, app B throws on its next render. That error also stops app B's ticker, so its canvas stays black:

```
TypeError: Cannot destructure property 'source' of 'object null' as it is null.
    at extractAttributesFromGpuProgram (pixi.mjs:17356:44)
    at get attributeData (pixi.mjs:17621:69)
    at PipelineSystem._generateAttributeLocationsKey (pixi.mjs:55055:47)
    at PipelineSystem._createVertexBufferLayouts (pixi.mjs:55091:47)
    at PipelineSystem.getBufferNamesToBind (pixi.mjs:55074:23)
    at GpuEncoderSystem.setGeometry (pixi.mjs:54331:51)
    at GpuGraphicsAdaptor.execute (pixi.mjs:56361:13)
    at GraphicsPipe.execute (pixi.mjs:34645:19)
    at executeInstructions (pixi.mjs:45324:40)
    at RenderGroupSystem.render (pixi.mjs:46102:5)
```

Depending on which of the destroyed program's lazily computed data gets read first, the error can instead come from bind group creation:

```
TypeError: Cannot read properties of null (reading '0')
    at BindGroupSystem._createBindGroup (pixi.mjs:53901:39)
    at BindGroupSystem.getBindGroup (pixi.mjs:53895:50)
    at GpuEncoderSystem.setBindGroup (pixi.mjs:54327:51)
    at GpuGraphicsAdaptor.execute (pixi.mjs:56363:13)
    at GraphicsPipe.execute (pixi.mjs:34645:19)
```

### System Info

```Shell
System:
    OS: macOS 26.6.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 64.00 GB
  Browsers:
    Chrome: 152.0.7977.83
```

### Any additional comments?

**Cause:**

- `compileHighShaderGpuProgram()` returns programs from `GpuProgram.from()`, which uses a module-level `programCache`. Every WebGPU renderer on the page gets the *same* `GpuProgram` instance for the built-in `graphics` and `mesh` shaders.
- When a renderer is destroyed, `GpuGraphicsAdaptor.destroy()` and `GpuMeshAdapter.destroy()` call `shader.destroy(true)`. That destroys the shared program: `vertex`, `fragment`, `layout` and `gpuLayout` are set to `null`, and its cache entry is removed.
- Every other live renderer still holds that destroyed program. As long as the program's lazily computed data (`_attributeData`, `_attributeLocationsKey`, bind group layouts) was never computed, the next non-batched draw reads the nulled fields and throws. In practice, that means no renderer has drawn a non-batchable Graphics or a Mesh with the default shader yet.
- This runs regardless of destroy options: `releaseGlobalResources: false` doesn't help, because the pipes/adaptors are torn down unconditionally in `AbstractRenderer.destroy()`.
- WebGL isn't affected: `GlGraphicsAdaptor` and `GlMeshAdaptor` use `compileHighShaderGlProgram()`, which creates a fresh `new GlProgram(...)` per renderer instead of using the cache.

**Possible fix:**

Don't destroy the cached programs from the adaptors:

```diff
 // src/scene/graphics/gpu/GpuGraphicsAdaptor.ts
 public destroy(): void
 {
-    this.shader.destroy(true);
+    this.shader.destroy(false);
     this.shader = null;
 }

 // src/scene/mesh/gpu/GpuMeshAdapter.ts
 public destroy(): void
 {
-    this._shader.destroy(true);
+    this._shader.destroy(false);
     this._shader = null;
 }
```

The **Run with fix** button applies this change at runtime. We also applied it via `pnpm patch`, and it fixes our multi-instance app. An alternative would be to reference-count cached `GpuProgram`s so they are only destroyed when the last user releases them.

**Context:**

We render multiple `Application`s on one page with `@pixi/react`. When one unmounts, which destroys its app, the remaining apps crash the next time they draw a non-batchable `Graphics`.

Possibly related: #11694 (multiple `Application`s interfering on `destroy()`), #12172 (rendering with a second renderer).

Zyie:
### Overview

Every WebGPU renderer shares the same cached `GpuProgram` for the built-in `graphics` and `mesh` shaders. The GPU graphics and mesh adaptors destroyed that program on renderer teardown, so any other live renderer threw on its next non-batched draw. The adaptors now destroy only their per-renderer shader, matching how every other holder of a cached program already behaves.

Fixes #12201

#### Fixes
- Destroying one WebGPU renderer no longer breaks non-batched `Graphics` and `Mesh` rendering in other live renderers

##### Pre-Merge Checklist


Zyie:
hey @zenblender can you verify if this fix works in your project?

bigtimebuddy (approved):

GoodBoyDigital (approved):

zenblender:
> hey @zenblender can you verify if this fix works in your project?

Thanks @Zyie!
I can confirm now that this bug was happening when I used WebGPU on `8.20.1` but now with `8.21.0` it's fixed!

